PR #7160, opened 2023-10-05
Bump regalloc2 to 0.9.3

elliottt:
Bump regalloc2 to version 0.9.3, which removes a pretty significant number of unnecessary moves in our filetests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2196,9 +2196,9 @@ dependencies = [
 
 [[package]]
 name = "regalloc2"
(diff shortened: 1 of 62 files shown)


cfallin (approved):
Nice! I suspect this must be because of bytecodealliance/regalloc2#161, which allows more bundle merging. Seems like uniformly good changes/improvements, at least eyeballing it.

alexcrichton (approved):

Issue #6090, opened 2023-03-23
Verification of addressing / memory accesses with VeriWasm or similar


cfallin:
I've been working on VeriWasm in Wasmtime the last few weeks, and decided to dump a summary of my investigation so far and ideas for the right approach to take. (I'm doing this partly because I've been asked to context switch back to something else, now that it's clear that VeriWasm proper cannot work and we need a new tool; I estimate it would take ~1 month to build what I describe in the last section below.)

(This may be better as an RFC, and I can clean it up, expand it, and post it as one once I have time to properly do so!)

## Summary

We should include built-in support in Wasmtime to verify that compiled code from Wasm modules is accessing only the memory that it is supposed to access. Such verification would take the form of ahead-of-time static analysis on the code, so that it would not require dynamic checks; and it should trust as little as possible in the existing compiler stack, to increase the confidence we have in its conclusions.

The ideal would be to verify machine code bytes with no other input, but in practice, Wasmtime will have to tell the tool what to assume about details such as runtime data structures; and certain kinds of analysis, like full recovery of a CFG in the presence of indirect branches, may require whole-program analysis that is too expensive to be practical.

This issue proposes a scheme that embeds a notion of "memory capabilities" in CLIF, along with (static) assertions on values (mostly with respect to ranges), and carries these through to the VCode level in Cranelift. The Wasm lowering would be modified to annotate every load/store with an associated memory capability that permits it; and we can then check the uses of these capabilities at the other end of the pipeline, after instruction selection is done. This is inspired by, but in some sense the dual to, the approach taken by VeriWasm.

## History: VeriWasm

In Johnson et al.[^1], the VeriWasm tool is described, with an approach that performs a static analysis of disassembled machine code to prove that all heap accesses in AOT-compiled Wasm code are properly to the Wasm heap, and cannot escape the sandbox. (The tool also checks other properties, such as control-flow integrity, but that is outside the scope of this issue.)

[^1]: Evan Johnson et al. "Довер ́яй, но провер ́яй: SFI Safety for native compiled Wasm." In NDSS 2021.  <https://cseweb.ucsd.edu/~lerner/papers/wasm-sfi-ndss2021.pdf>
      
VeriWasm was designed to verify code produced by [Lucet](https://github.com/bytecodealliance/lucet), another AOT Wasm compiler that has since been subsumed by Wasmtime and EOL'd.

[...]

PR #7165, opened 2023-10-06
Skeleton and initial support for proof-carrying code.

cfallin:
This PR adds basic support for "facts" to annotate SSA values and VCode VRegs. The idea is that these facts describe properties of the values -- currently, "in a given range" or "points to valid memory of a given size" -- and by validating a chain of facts along the whole derivation of an accessed memory address, we can verify that the memory access is safe.

This is the start of work on the ideas described in #6090. It is a new approach that subsumes/replaces "VeriWasm": it replaces the post-hoc derive-all-facts analysis approach (which is slow and brittle) with a cooperative approach in which the codegen frontend generates a "proof witness" of sorts, we preserve that, and we check that in the machine code (actually the pre-regalloc VCode, and then we can use the regalloc's symbolic checker to obtain the same guarantee).

Right now, this includes (i) the "fact" data structures, (ii) support for parsing these facts in CLIF and dumping them back out, (iii) propagation of facts from CLIF to VCode, when no optimization is enabled, and (iv) validation of a few basic facts in VCode for the AArch64 backend. Everything else is still TODO and will come in further PRs.

Right now, this is sufficient to take the CLIF

```
test compile
set enable_pcc=true
target aarch64

function %simple1(i64 vmctx, i32) -> i8 {
block0(v0 ! points_to(0x1_0000_0000): i64, v1 ! max(32, 0xffff_ffff): i32):
    v2 ! max(64, 0xffff_ffff) = uextend.i64 v1
    v3 ! points_to(0x1) = iadd.i64 v0, v2
    v4 = load.i8 safe v3
    return v4
}
```

where the `v0 ! fact` syntax annotates `v0` with a fact, and validate that the `load` is in fact safe. The `safe` memory-flag forces this check; if the `points_to` range is shrunk, or `v1` is changed to an `i64` and its max range is increased, this code will fail to compile with

```
 ERROR cranelift_filetests::concurrent > FAIL: compile
FAIL cranelift/filetests/filetests/pcc/simple.clif: compile

Caused by:
    Proof-carrying-code validation error: MissingFact
```

The idea is to continue further work along these lines:
- Add support for validating facts produced by aarch64 add, shift, ... instructions -- whatever is used in basic address generation.
- Add support for validating amode usage in x64 as well.
- Propagate facts through the egraph layer so they are preserved during optimization.
- Add a richer `points_to` language for facts to describe *what* they point to, so a load of a memory base address (for example) from `vmctx` yields another value with an attached fact (points to a region of this size...). These are the "memory capabilities" described in #6090.
- Emit facts on `vmctx` and on memory-access sequences generated by `cranelift-wasm`, and add the `safe` flag to loads/stores wherever possible. At this point, we should be able to validate single- and multi-memory Wasm modules with static memory configurations.
- Add support for annotations on bounds-check logic. After this, we should be able to validate dynamic memories, and table accesses as well.
- Finally, add a "strong enforcing" mode that disallows all non-`safe` loads/stores, and add whatever memory-capability info is needed so that, e.g., loads of func pointers, VM state, and everything else from `vmctx` is covered.

After my initial exploration and ideas in #6090, these ideas were further discussed with @fitzgen, @elliottt, @jameysharp, and others; and a large part of this PR was


cranelift/codegen/src/ir/pcc.rs
```diff
@@ -0,0 +1,347 @@
...
+                    bit_width: bw_max,
+                    max,
+                },
+            ) if *bw_max >= pointer_width && add_width >= *bw_max => {
+                let computed_region = MemoryRegion {
+                    max: region.max.checked_sub(*max)?,
+                };
+                Some(Fact::PointsTo {
+                    region: computed_region,
+                })
```
fitzgen, line 232: Stylistic nitpick: I'd personally either name `computed_region` just `region` so we can use short-hand struct literal syntax for the field, or even inline the variable into the struct literal field.

cranelift/codegen/src/ir/pcc.rs
```diff
@@ -0,0 +1,347 @@
...
+    pub fn sextend(&self, from_width: u16, to_width: u16) -> Option<Self> {
+        match self {
+            // If we have a defined value in bits 0..bit_width, and
+            // the MSB w.r.t. `from_width` is *not* set, then we can
+            // do the same as `uextend`.
+            Fact::ValueMax { bit_width, max }
+                if *bit_width == from_width && (*max & (1 << (*bit_width - 1)) == 0) =>
+            {
+                self.uextend(from_width, to_width)
+            }
```
fitzgen, line 265: And otherwise we could set it to `bit_width::MAX` -- any reason we shouldn't do that and returning `None` instead is preferred?

cranelift/codegen/src/ir/pcc.rs
```diff
@@ -0,0 +1,347 @@
...
+    ///
+    /// The slicing behavior is needed because this fact can describe
+    /// both an SSA `Value`, whose entire value is well-defined, and a
+    /// `VReg` in VCode, whose bits beyond the type stored in that
+    /// register are don't-care (undefined).
+    ValueMax {
+        /// The bitwidth of bits we care about, from the LSB upward.
+        bit_width: u16,
+        /// The maximum value that the bitslice can take (inclusive).
+        max: u64,
```
fitzgen, line 100: It is sort of already defined by the fact we are using a `u64` for `max`, but it might be good to reiterate that we are interpreting values as unsigned here.

cranelift/codegen/src/ir/pcc.rs
```diff
@@ -0,0 +1,347 @@
...
+}
+
+/// A memory region that can be accessed. This description is attached
+/// to a particular base pointer.
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
+pub struct MemoryRegion {
+    /// Includes both the actual memory bound as well as any guard
+    /// pages. Inclusive, so we can represent the full range of a
+    /// `u64`.
+    pub max: u64,
```
fitzgen, line 119: Ditto regarding unsignedness.

cranelift/codegen/src/ir/pcc.rs
```diff
@@ -0,0 +1,347 @@
...
+            }
+            _ => None,
+        }
+    }
+
+    /// Scales a value with a fact by a known constant.
+    pub fn scale(&self, width: u16, factor: u32) -> Option<Self> {
+        match self {
+            Fact::ValueMax { bit_width, max } if *bit_width == width => {
+                let max = max.checked_mul(factor as u64)?;
```
fitzgen, line 274: Integer conversions a tricky and `as` can be a footgun; let's future proof:

```suggestion
                let max = max.checked_mul(u64::from(factor))?;
```

fitzgen: Again: is it better to return `None` than to do a saturating multiplication? I honestly don't know. But if we are using `None` for "any value" then this essentially means that we have two representations for "any value": `None` and `Some(Fact::ValueMax { max: u64::MAX })`.

fitzgen: On this topic: it might be worth drawing out a diagram of our lattice in the docs for `Fact` or something and a note about what our representation is for the things where it isn't obvious (top/bottom). Something like:

```
///                None
///               /     \
///              /       \
///             /         \
///     Some(ValueMax)  Some(PointsTo)
///             \         /
///              \       /
///               \     /
///            Err(PccError)
///
/// * The top of the lattice, representing any value, is `None`.
///
/// * The bottom of the lattice, representing the lack of any (valid)
///   value, is an error.
///
/// * The intermediate values are either a maximum bound on the value,
///   or that the value is a pointer to a memory region. Various
///   arithmetic operations are valid on various combinations of these
///   facts. Adding two `MaxValue`s will produce a new `MaxValue` with
///   a maximum of the sum of the parts' maximums. Adding a `MaxValue`
///   to a `PointsTo` will produce a new `PointsTo` that is pointing
///   to a smaller memory region. Etc...
```

Although, now that I think about it, I don't think this is even really a lattic? Maybe a semi-lattice. I don't think we have both a bottom and a top... I guess the real bottom would be "no value" for dead code, but we don't really have that given that we aren't propagating facts ourselves, merely checking them.

And I guess I would expect `None` to represent "no value" rather than "any value" so maybe that is why I am hung up on this `None` stuff.

I don't know, again this comment is all over the place and rambly and doesn't have a real concrete suggestion.

What do you think?

cranelift/codegen/src/ir/pcc.rs
```diff
@@ -0,0 +1,347 @@
...
+        // negative offset could underflow, and removes all
+        // claims of constrained range.
+        if offset < 0 {
+            return None;
+        }
+        let offset = u64::try_from(offset).unwrap();
+
+        match self {
+            Fact::ValueMax { bit_width, max } if *bit_width == width => {
+                let max = max.checked_add(offset).unwrap();
```
fitzgen, line 301: Wait why is it okay to unwrap here? Shouldn't this use `?` to propagate?

cranelift/codegen/src/ir/pcc.rs
```diff
@@ -0,0 +1,347 @@
...
+                    bit_width: *bit_width,
+                    max,
+                })
+            }
+            _ => None,
+        }
+    }
+
+    /// Offsets a value with a fact by a known amount.
+    pub fn offset(&self, width: u16, offset: i64) -> Option<Self> {
```
fitzgen, line 288: This is essentially `add` but with a signed offset, right? Can we name it accordingly and/or have it call out to `add` after the check for negative offsets?

cranelift/codegen/src/ir/pcc.rs
```diff
@@ -0,0 +1,347 @@
...
+                let max = max.checked_add(offset).unwrap();
+                Some(Fact::ValueMax {
+                    bit_width: *bit_width,
+                    max,
+                })
+            }
+            Fact::PointsTo {
+                region: MemoryRegion { max },
+            } => {
+                let max = max.checked_sub(offset).unwrap();
```
fitzgen, line 310: Ditto regarding unwrap.

cranelift/codegen/src/ir/pcc.rs
```diff
@@ -0,0 +1,347 @@
...
+            _ => bail!(OutOfBounds),
+        }
+
+        Ok(())
+    }
+}
+
+/// Top-level entry point after compilation: this checks the facts in
+/// VCode.
+pub fn check_facts<B: LowerBackend + TargetIsa>(
```
fitzgen, line 335: Maybe name this `check_vcode_facts` to leave the door open for `check_clif_facts` as well?

cranelift/codegen/src/ir/pcc.rs
```diff
@@ -0,0 +1,347 @@
...
+    _f: &ir::Function,
+    vcode: &VCode<B::MInst>,
+    backend: &B,
+) -> PccResult<()> {
+    for inst in 0..vcode.num_insts() {
```
fitzgen, line 340: At the top of this function is where we would create the `PccContext` thing I mentioned earlier that factors out pointer width (and eventually other stuff) and has the `add` method etc...

fitzgen (approved):
Very excited for this! r=me with comments, some of which can be deferred to follow up PRs.

cranelift/codegen/src/ir/memflags.rs
```diff
@@ -20,10 +29,15 @@ enum FlagBit {
     /// Accesses only the "vmctx" part of abstract state. Used for
     /// alias analysis. Mutually exclusive with "heap" and "table".
     Vmctx,
+    /// Check this load or store for safety when using the
+    /// proof-carrying-code framework. The address must have a
+    /// `PointsTo` fact attached with a sufficiently large valid range
+    /// for the accessed size.
+    Safe,
```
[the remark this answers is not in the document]
fitzgen: Checked vs unchecked maybe? So you set the unchecked bitflag to opt out of PCC checking.

cranelift/codegen/src/ir/pcc.rs
```diff
@@ -0,0 +1,347 @@
...
+//! include an addition, multiplication/shift, or extend operation,
+//! and there is no way to attach facts to the intermediate values
+//! "inside" the instruction, so instead the backend can use
+//! `Fact::add()` and friends to forward-propagate facts.
+//!
+//! TODO:
+//! - Propagate facts through optimization (egraph layer).
+//! - Generate facts in cranelift-wasm frontend when lowering memory ops.
+//! - Implement richer "points-to" facts that describe the pointed-to
+//!   memory, so the loaded values can also have facts.
```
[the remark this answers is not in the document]
jameysharp: I too had wondered if we could making global-value assertions axiomatic to avoid a richer assertion language so I'm glad you asked. I can definitely see downsides but it seems nice as a starting point.

cranelift/codegen/src/ir/memflags.rs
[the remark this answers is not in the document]
cfallin: Ah, yeah, I like the "checked" terminology. I'll go with `checked` for now (with "unchecked" being the default) for backwards compatibility.

cranelift/codegen/src/ir/pcc.rs
```diff
@@ -0,0 +1,347 @@
+//! Proof-carrying code. We attach "facts" to values and then check
+//! that they remain true after compilation.
+//!
+//! A few key design principle of this approach are:
+//!
+//! - The producer of the IR provides the axioms. All "ground truth",
+//!   such as what memory is accessible -- is meant to come by way of
+//!   facts on the function arguments. In some sense, all we are doing
```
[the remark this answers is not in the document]
cfallin: Added!

cranelift/codegen/src/ir/pcc.rs
```diff
@@ -0,0 +1,347 @@
...
+//!
+//!   This does create more verbose IR, but a significant positive
+//!   benefit is that it avoids unnecessary work: we do not build up a
+//!   knowledge base that effectively encodes the integer ranges of
+//!   many or most values in the program. Rather, we only check
+//!   specifically the memory-access sequences. In practice, each such
+//!   sequence is likely to be a carefully-controlled sequence of IR
+//!   operations from, e.g., a sandboxing compiler (such as
+//!   `cranelift-wasm`) so adding annotations here so communicate
+//!   intent (ranges, bounds-checks, and the like) is no problem.
```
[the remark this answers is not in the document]
cfallin: Fixed!

cranelift/codegen/src/ir/pcc.rs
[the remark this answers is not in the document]
cfallin: I've thought about this for a bit too; I think I'd still like to try actually describing memory layout (and I have some thoughts on how to do it while keeping `Fact` cheap with shared "memcaps" that are shared across the func). The basic reason is to avoid growing the trusted base. If we provide an axiom on what the loaded value from `vmctx` is, then we're trusting that load to match what we've stated; but the load-generation machinery is exactly what we're trying to verify with all of this in the first place. Now, arbitrary program logic may create more complex conditions / isel cases than a controlled `vmctx` case, but maybe not; in either case, I'd rather have the machine load instructions actually verified.

cranelift/codegen/src/ir/pcc.rs
```diff
@@ -0,0 +1,347 @@
...
+    /// An operation wraps around, invalidating the stated value
+    /// range.
+    Overflow,
+    /// An input to an operator that produces a fact-annotated value
+    /// does not have a fact describing it, and one is needed.
+    MissingFact,
+    /// A memory access is out of bounds.
+    OutOfBounds,
+    /// Proof-carry-code checking is not implemented for a certain case.
+    Unimplemented,
```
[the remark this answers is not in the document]
cfallin: I opted to split it out into `UnimplementedBackend` and `UnimplementedInst` -- slight preference for programmatic errors rather than strings (which preclude any sort of parsing and make test asserts harder). Hopefully that gets the spirit of this at least, noting what's unimplemented :-)

cranelift/codegen/src/ir/pcc.rs
```diff
@@ -0,0 +1,347 @@
...
+
+macro_rules! bail {
+    ( $err:tt ) => {{
+        return Err(PccError::$err);
+    }};
+}
+
+impl Fact {
+    /// Computes whether a fact "subsumes" (implies) another.
+    pub fn subsumes(&self, other: &Self) -> bool {
```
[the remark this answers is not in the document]
cfallin: Made doc comment clearer!

cranelift/codegen/src/ir/pcc.rs
```diff
@@ -0,0 +1,347 @@
...
+                max_rhs <= max_lhs
+            }
+            _ => false,
+        }
+    }
+
+    /// Computes whatever fact can be known about the sum of two other
+    /// values with attached facts. The add is performed to the given
+    /// bit-width, and the target machine has the given pointer-width.
+    pub fn add(&self, other: &Self, add_width: u16, pointer_width: u16) -> Option<Self> {
```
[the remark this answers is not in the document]

cranelift/codegen/src/ir/pcc.rs

cranelift/codegen/src/ir/pcc.rs
cfallin: Yep, good point, added.

cranelift/codegen/src/ir/pcc.rs

cranelift/codegen/src/ir/pcc.rs

cranelift/codegen/src/ir/pcc.rs
cfallin: Done; this method will return a `Some(..)` in any case we have a `ValueMax`, and at worst return the range of the given bitwidth.

Re: `None` -- I guess the way I think about it is that "we have no fact" (in the same way that the facts attached to values are an array of `Option<Fact>`) -- we just can't say/derive anything. I'm not sure I want to describe this in terms of a (semi-)lattice as we don't have a meet-operation anywhere, and that's pretty intentional -- we aren't doing any fixpoint analysis or inference, we just check what's given. (I guess subsumption technically induces a partial order and a least-upper-bound might exist, but...)

cranelift/codegen/src/ir/pcc.rs
cfallin: Good point; now this returns the "minimal fact" (the range of the whole bitwidth) instead if this overflows.

cranelift/codegen/src/ir/pcc.rs
cfallin: I had thought of that initially, but the difference here is that it's an add-with-constant, and we don't have a `Const` (known exact value) in our facts (yet?). I didn't want to complicate things by adding that and working out all the details. I agree this would be a nice simplification if we do go there eventually!

cranelift/codegen/src/ir/pcc.rs
cfallin: Fixed, to a `?` this time (the "minimal fact" here is a "region of 0 bytes" which isn't very useful).

cranelift/codegen/src/ir/pcc.rs

cranelift/codegen/src/ir/pcc.rs
cfallin: I created this in the backend since some of the context (actually the only piece of context now, the pointer width) is machine-specific.

PR #7167, opened 2023-10-06
fix(wasmtime-cli): wrong linker in wasi http

eduardomourar:
This will fix the error one would get while trying to run the wasmtime CLI with `-S http` because the async functions were being added to the linker. Additionally, we will expose only the necessary functions to ~~allow shadowing in~~ linker when using WASI HTTP in order to prevent the following error:
> import of `wasi:clocks/wall-clock` defined twice

~~Also, to prevent it from breaking in the future, a simple handcrafted test has been added to the CLI test suite.~~


eduardomourar:
I am currently using wasmtime CLI from the main branch, so I don't need this fix in wasmtime 14. Nonetheless, I would recommend backporting it to the v14 release branch.

src/commands/run.rs
```diff
@@ -643,7 +643,8 @@ impl RunCommand {
                         bail!("Cannot enable wasi-http for core wasm modules");
                     }
                     CliLinker::Component(linker) => {
-                        wasmtime_wasi_http::proxy::add_to_linker(linker)?;
+                        linker.allow_shadowing(true);
+                        wasmtime_wasi_http::proxy::sync::add_to_linker(linker)?;
```
alexcrichton, line 647: I think it would be best if `allow_shadowing` was disabled because that's otherwise easy to accidentally run into issues.

There's unfortunately not a great solution for the error you're running into here, so in the meantime I think it would be best to expose individual `add_to_linker` functions per-interface and only invoke the wasi-http-related ones here.

tests/all/cli_tests/wasi-http-component.wat
```diff
@@ -0,0 +1,357 @@
+(component
```
alexcrichton, line 1: Thanks for adding this! This is unfortunately not really possible to change though and without the source used to generate this we can't feasibly change this over time. I think the best solution is to enable the `cli-tests` suite here to use programs from the `test-programs` crate. That's a larger refactoring though and probably best done in a separate PR, so it's ok to skip a test for this PR specifically and rely on local testing.

eduardomourar: i actually built the component out of the original core module `wasi-http` (which was handcrafted before in order to make it more maintainable).

eduardomourar: i agree it will be very hard to keep it up-to-date while we are still iterating over WASI. maybe we can keep the original core module and build a component on the fly with wasm-tools?

src/commands/run.rs
eduardomourar: done. as it is only temporary, i exposed function only in sync module and kept out of docs.

tests/all/cli_tests/wasi-http-component.wat
alexcrichton: Yes that makes sense but it's not clear to me, for example, how to regenerate this test. This means that if there's an update it would be quite difficult to update this. Even the core module will change because the wasi-http interfaces might change.

The "real fix" here is to use source code rather than compiled wat or wasm, but that would require further refactorings to provide the `test-programs` programs to the WASI CLI tests here. I think it's best to have no tests until that's done, and I don't think it would be too large of a refactoring so I can work on it after this PR lands.

eduardomourar: there is no real source code because it was done by hand. it is not hard to make it, though. i just wanted to keep it very simple without a server-side in the cli as it takes longer to run.

alexcrichton: Yes that's all fine, but again I would personally prefer not to check in a large `*.wat` file which we don't know how to regenerate and change. I don't doubt that you've generated this from source, but there's no instructions in-tree about what the generation process is or what the source is. I think it's best to leave this out, manually test this PR, and I'll follow up with a better testing strategy to re-land this.

eduardomourar: just added a test to `test-programs` called `outbound_request_response_build` that does not require a server. i just don't know how to find the path to reference in the wasmtime cli tests file.

alexcrichton: Hm ok so in my opinion this PR can't land with this `*.wat` test. This is even if it's generated from the test you've added in-tree. This is local knowledge on this PR's comment which will be lost basically as soon as the PR lands, meaning that if you or I aren't the next person to touch this file, which is quite likely, then they'll not know what to do. To solve this dilemma one of a few things need to happen:

* This `*.wat` test should be deleted, temporarily. I'll send a follow-up PR to enable referencing wasms from `test-programs` here in this test suite.
* You should, in this PR, delete this `*.wat` and implement the previous bullet in this PR.
* Block this PR on me doing the above refactor, then rebase on it and reference the `test-programs`-generated wasm file.

Personally I prefer the first option because, as you pointed out, we'll want to backport this. In that sense I'd rather not block this PR. I'm happy to defer to you though on which route you'd prefer.

eduardomourar: i have removed the wat file and ignored that particular test case. let me know and i can help re-enabling the test back when ready

crates/test-programs/wasi-http-tests/src/bin/outbound_request_response_build.rs
```diff
@@ -0,0 +1,40 @@
+use wasi_http_tests::bindings::wasi::{cli::stdout::get_stdout, http::types as http_types};
+
+fn print(msg: &[u8]) {
+    let _ = get_stdout().blocking_write_and_flush(&msg);
+}
```
alexcrichton, line 5: For the future this isn't necessary as `println!` should work in guests. If it doesn't that's a bug we should document and fix.

alexcrichton (approved):

crates/test-programs/wasi-http-tests/src/bin/outbound_request_response_build.rs
eduardomourar: If I use println, it gets translated to wasi preview1 but I wanted to exercise the preview2 functions

alexcrichton: Ah that's sort of right but also not quite accurate. It's correct that `println!` uses the Rust standard library which uses wasi preview1, but that's adapted to preview2 with the adapter, meaning that `println!` ends up calling the same preview2 function this one calls.

eduardomourar: But not in the core module, right?

alexcrichton: Sorry I'm not exactly sure what you mean, but when using `println!` the core module produced by rustc will refer to `wasi_snapshot_preview1::fd_write` if that's what you mean. That core import is then hooked up to the adapter, and the adatper imports preview2 instead.

tests/all/cli_tests/wasi-http-component.wat
alexcrichton: To follow up here, https://github.com/bytecodealliance/wasmtime/pull/7182 is the refactoring I had in mind and [this](https://github.com/bytecodealliance/wasmtime/blob/14418300cd98a942d15e383834f533680816ef91/tests/all/cli_tests.rs#L1250-L1268) is the un-ignore of the test

PR #7170, opened 2023-10-06
Update syntax for `trappable_error_type` in `bindgen!`

alexcrichton:
Following through on a suggestion from #7152 this makes the syntax a bit more consistent with the rest of `bindgen!` today. This additionally updates the documentation on the `bindgen!` macro itself.


elliottt (approved):
This is awesome! Thanks for making the syntax for `with` and `trappable_error_type` the same :)

PR #7172, opened 2023-10-06
Support multiple versions in `bindgen!`

alexcrichton:
This commit fixes a bug in the `bindgen!` macro where when faced with multiple packages that differ only in version number invalid bindings were generated. The fix here is to add version number information to package module names if necessary in situations such as this. This required some refactoring internally to have a single source of truth for what the name of a module should be and avoid having it implicitly calculated in two locations.

### DIFF
--- a/crates/wit-bindgen/src/lib.rs
+++ b/crates/wit-bindgen/src/lib.rs
@@ -27,20 +27,22 @@ mod source;
 mod types;
 use source::Source;
 
-struct InterfaceName {
-    /// True when this interface name has been remapped through the use of `with` in the `bindgen!`
-    /// macro invocation.
-    remapped: bool,
-
-    /// The string name for this interface.
-    path: String,
+#[derive(Clone)]
+enum InterfaceName {
+    /// This interface was remapped using `with` to some other Rust code.
+    Remapped {
+        name_at_root: String,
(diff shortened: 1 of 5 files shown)


crates/wit-bindgen/src/lib.rs
```diff
@@ -27,20 +27,22 @@ mod source;
...
-    remapped: bool,
-
-    /// The string name for this interface.
-    path: String,
+#[derive(Clone)]
+enum InterfaceName {
+    /// This interface was remapped using `with` to some other Rust code.
+    Remapped {
+        name_at_root: String,
+        local_path: Vec<String>,
```
elliottt, line 35: It looks like this vector only ever has one or three strings in it, would it make sense to make a more precise type that names those cases?

elliottt (approved):
I think this all makes sense! I had one suggestion about the changes to the `InterfaceName` type, but it's not a blocker.

crates/wit-bindgen/src/lib.rs
alexcrichton: I think that's probably a bit overkill on types for now since nothing uses this as anything other than `.join("::")` currently, but if that changes in the future I think it'd be reasonable to add yeah

PR #7177, opened 2023-10-06
cranelift: Consider functions with TLS values as non leaf

afonso360:
👋 Hey,

@bjorn3 reported an issue with TLS values in the RISC-V backend, when testing it with cg_clif. This presented as a segfault running the cg_clif testsuite.

Digging into it, it looks like we would call the `elf_tls_getaddr` function, which would clobber the `ra` register and before returning from the current function we would **not** restore the correct `ra` value from the stack. This would cause the function to recurse back into the middle of itself and cause a segfault.

This function met all criteria for us to avoid emitting a prologue and epilogue sequence, that would correctly restore the `ra` value. The check that was failing here is the `is_leaf` check. It assumes that a function will only call other functions if it contains any signatures itself.

This PR adds an additional check for TLS values, and assumes that if such a value exists we will be calling a function to retrieve it's address. This is not necessarily true, but it is a good conservative estimate.

More importantly I don't think this fully addresses the issue. I think we can still have a similar issue if we decide to use a libcall instead of native instruction during lowering. I'm not entirely sure how to fix that.

### DIFF
--- a/cranelift/codegen/src/ir/function.rs
+++ b/cranelift/codegen/src/ir/function.rs
@@ -314,7 +314,17 @@ impl FunctionStencil {
     pub fn is_leaf(&self) -> bool {
         // Conservative result: if there's at least one function signature referenced in this
         // function, assume it is not a leaf.
-        self.dfg.signatures.is_empty()
+        let has_signatures = !self.dfg.signatures.is_empty();
+
+        // Under some TLS models, retrieving the address of a TLS variable requires calling a
+        // function. Conservatively assume that any function that references a tls global value
+        // is not a leaf.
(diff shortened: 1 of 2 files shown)


alexcrichton (approved):

bjorn3:
Would it be possible to backport this to cranelift 0.101? That would allow me to ship riscv64 support in cg_clif one month earlier and this shouldn't affect Wasmtime.

alexcrichton:
Certainly yeah, would you be up for sending a PR to the release-14.0.0 branch?

bjorn3:
Done: https://github.com/bytecodealliance/wasmtime/pull/7195

PR #7180, opened 2023-10-06
PCC: add semantics for core add/shift/extend/amode ops on AArch64.

cfallin:
This PR adds verification of facts on values produced by adds, shifts, and extends on AArch64, handling the various combination instructions (adds with builtin extends or shifts, for example), and also adds verification of all addressing modes, including those with builtin extends and shifts.

It also splits the test suite into"succeed" and "fail" sets, and provides cases that PCC should catch.

### DIFF
--- a/cranelift/codegen/src/ir/pcc.rs
+++ b/cranelift/codegen/src/ir/pcc.rs
@@ -76,6 +76,9 @@ pub enum PccError {
     /// An input to an operator that produces a fact-annotated value
     /// does not have a fact describing it, and one is needed.
     MissingFact,
+    /// A derivation of an output fact is unsupported (incorrect or
+    /// not derivable).
+    UnsupportedFact,
     /// A memory access is out of bounds.
     OutOfBounds,
     /// Proof-carrying-code checking is not implemented for a
@@ -310,7 +313,10 @@ impl FactContext {
                 if *bit_width < 64 && max > max_value_for_width(width) {
                     return Some(minimal_fact);
                 }
-                Some(minimal_fact)
+                Some(Fact::ValueMax {
+                    bit_width: *bit_width,
+                    max,
+                })
             }
             _ => None,
         }
--- a/cranelift/codegen/src/ir/pcc.rs
+++ b/cranelift/codegen/src/ir/pcc.rs
@@ -322,6 +322,15 @@ impl FactContext {
         }
     }
 
+    /// Left-shifts a value with a fact by a known constant.
+    pub fn shl(&self, fact: &Fact, width: u16, amount: u16) -> Option<Fact> {
+        if amount >= 32 {
+            return None;
+        }
+        let factor: u32 = 1 << amount;
+        self.scale(fact, width, factor)
+    }
+
     /// Offsets a value with a fact by a known amount.
     pub fn offset(&self, fact: &Fact, width: u16, offset: i64) -> Option<Fact> {
         // If we eventually support two-sided ranges, we can
--- a/cranelift/codegen/src/isa/aarch64/pcc.rs
+++ b/cranelift/codegen/src/isa/aarch64/pcc.rs
@@ -10,7 +10,7 @@ use crate::machinst::Reg;
 use crate::machinst::VCode;
 use crate::trace;
 
-fn get_fact_or_fail(vcode: &VCode<Inst>, reg: Reg) -> PccResult<&Fact> {
+fn get_fact(vcode: &VCode<Inst>, reg: Reg) -> PccResult<&Fact> {
     vcode.vreg_fact(reg.into()).ok_or(PccError::MissingFact)
 }
 
@@ -97,7 +97,13 @@ pub(crate) fn check(inst: &Inst, vcode: &VCode<Inst>) -> PccResult<()> {
             rn,
             flags,
             ..
-        } => check_scalar_addr(&ctx, *flags, *rn, vcode, access_ty.bytes() as u8),
+        } => check_scalar_addr(
+            &ctx,
+            *flags,
+            *rn,
+            vcode,
+            u8::try_from(access_ty.bytes()).unwrap(),
+        ),
 
         // TODO: stores: once we have memcaps, check that we aren't
         // overwriting a field that has a pointee type.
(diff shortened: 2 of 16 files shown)


cranelift/codegen/src/isa/aarch64/pcc.rs
```diff
@@ -113,37 +263,32 @@ fn check_addr(
...
+            let sum = fail_if_missing(ctx.offset(&rn, 64, simm9.value as i64))?;
             ctx.check_address(&sum, size as u32)
         }
         &AMode::UnsignedOffset { rn, uimm12 } => {
-            let rn = vcode.vreg_fact(rn.into()).ok_or(PccError::MissingFact)?;
+            let rn = get_fact_or_fail(vcode, rn)?;
+            // Safety: this will not overflow: `size` should be at
+            // most 32 or 64 for large vector ops, and the `uimm12`'s
+            // value is at most 4095.
             let offset = (uimm12.value as u64) * (size as u64);
```
fitzgen, line 290: Can we do `a.checked_mul(b).unwrap()` to assert that this will not overflow?

cranelift/codegen/src/machinst/lower.rs
```diff
@@ -771,6 +771,19 @@ impl<'func, I: VCodeInst> Lower<'func, I> {
...
+                        // instead. Lookup goes through the alias, but
+                        // we want to preserve whatever was stated
+                        // about the vreg before its producer was
+                        // lowered.
+                        if let Some(fact) =
+                            self.vregs.take_fact(dst.to_virtual_reg().unwrap().into())
+                        {
+                            self.vregs
+                                .set_fact(temp.to_virtual_reg().unwrap().into(), fact);
+                        }
```
fitzgen, line 786: Do we know we won't ever look up facts for the aliased vreg anymore? That is, is it really safe to `take` rather than `clone`?

cranelift/codegen/src/ir/pcc.rs
```diff
@@ -310,7 +313,10 @@ impl FactContext {
                 if *bit_width < 64 && max > max_value_for_width(width) {
                     return Some(minimal_fact);
                 }
-                Some(minimal_fact)
+                Some(Fact::ValueMax {
+                    bit_width: *bit_width,
+                    max,
+                })
```
cfallin, line 319: This was an imprecision in the initial impl -- we had returned the "full range" fact (correct but not useful) rather than the thing we actually computed.

cranelift/filetests/filetests/pcc/fail/add.clif
fitzgen, line 1: It would be nice to (eventually) have expected error messages that we check against for each of these failures.

fitzgen (approved):

cranelift/codegen/src/machinst/lower.rs
cfallin: The lookup follows the alias, so querying the alias will get the fact as well -- we're moving it here so if the aliaser had a fact and the aliasee didn't, the fact gets preserved (that's the common case with the way ISLE lowering works -- we alias the original reg to the new temp the rule returned).

cranelift/filetests/filetests/pcc/fail/add.clif
cfallin: Agreed! I hope to build up a more granular testing thing in general (assert exactly which inst/fact fails, etc)

cranelift/codegen/src/isa/aarch64/pcc.rs
```diff
@@ -63,25 +118,118 @@ pub(crate) fn check(inst: &Inst, vcode: &VCode<Inst>) -> PccResult<()> {
...
+            alu_op: ALUOp::Add,
+            size,
+            rd,
+            rn,
+            rm,
+        } if has_fact(vcode, *rn) && has_fact(vcode, *rm) => {
+            check_output(&ctx, vcode, rd.to_reg(), || {
+                let rn = get_fact_or_fail(vcode, *rn)?;
+                let rm = get_fact_or_fail(vcode, *rm)?;
+                fail_if_missing(ctx.add(rn, rm, size.bits() as u16))
```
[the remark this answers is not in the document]
cfallin: Done!

cranelift/codegen/src/isa/aarch64/pcc.rs
```diff
@@ -63,25 +118,118 @@ pub(crate) fn check(inst: &Inst, vcode: &VCode<Inst>) -> PccResult<()> {
...
+            rd,
+            rn,
+            rm,
+            shiftop,
+        } if shiftop.op() == ShiftOp::LSL && has_fact(vcode, *rn) && has_fact(vcode, *rm) => {
+            check_output(&ctx, vcode, rd.to_reg(), || {
+                let rn = get_fact_or_fail(vcode, *rn)?;
+                let rm = get_fact_or_fail(vcode, *rm)?;
+                let factor = 1 << (shiftop.amt().value() as u32);
+                let rm_shifted = fail_if_missing(ctx.scale(&rm, size.bits() as u16, factor))?;
```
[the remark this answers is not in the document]
cfallin: Done!

cranelift/codegen/src/isa/aarch64/pcc.rs
```diff
@@ -63,25 +118,118 @@ pub(crate) fn check(inst: &Inst, vcode: &VCode<Inst>) -> PccResult<()> {
...
+            alu_op: ALUOp::Lsl,
+            size,
+            rd,
+            rn,
+            immshift,
+        } if has_fact(vcode, *rn) && has_fact(vcode, *rn) => {
+            check_output(&ctx, vcode, rd.to_reg(), || {
+                let rn = get_fact_or_fail(vcode, *rn)?;
+                let factor = 1 << (immshift.value() as u32);
+                fail_if_missing(ctx.scale(&rn, size.bits() as u16, factor))
```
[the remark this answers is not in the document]
cfallin: Done!

cranelift/codegen/src/isa/aarch64/pcc.rs
```diff
@@ -93,18 +241,20 @@ fn check_addr(
...
     }
 
+    trace!("check_addr: {:?}", addr);
+
     match addr {
         &AMode::RegReg { rn, rm } => {
-            let rn = vcode.vreg_fact(rn.into()).ok_or(PccError::MissingFact)?;
-            let rm = vcode.vreg_fact(rm.into()).ok_or(PccError::MissingFact)?;
-            let sum = ctx.add(&rn, &rm, 64).ok_or(PccError::MissingFact)?;
+            let rn = get_fact_or_fail(vcode, rn)?;
```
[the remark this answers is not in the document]
cfallin: Done!

cranelift/codegen/src/isa/aarch64/pcc.rs
cfallin: Yep, good idea.

PR #7181, opened 2023-10-06
riscv64: Improve `f{min,max}` codegen

afonso360:
👋 Hey,

This PR improves our lowerings for `fmin` and `fmax` and moves them to ISLE.

The native `fmin/fmax` instructions in RISC-V almost match the WASM semantics, they correctly order `-0.0 < +0.0`, however they return the non-NaN input if there is one. This does not match the semantics of the instruction so we fixup this case manually.

This codegen can be further improved using the `fminm` / `fmaxm` instructions available in the `Zfa` extension which I'm planing on implementing next.


cranelift/codegen/src/isa/riscv64/lower.isle
```diff
@@ -1478,8 +1478,16 @@
...
 ;;;; Rules for `fmin` ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
+;; RISC-V's `fmin` instruction returns the number input if one of inputs is a
+;; NaN. We handle this by manually checking if one of the inputs is a NaN
+;; and selecting based on that result.
 (rule 0 (lower (has_type (ty_scalar_float ty) (fmin x y)))
-  (gen_float_select (FloatSelectOP.Min) x y ty))
+  (let (;; Check if both inputs are not nan.
+        (is_ordered FCmp (emit_fcmp (FloatCC.Ordered) ty x y))
+        ;; `fadd` returns a nan if NaN of the inputs is a NaN.
```
cfallin, line 1487: s/NaN of/one of/ ?

cfallin (approved):
LGTM, thanks! The `fadd` trick to get a NaN is clever.

It's too bad the semantics are just a tad off -- especially without a true select/cmove instruction to use in the fixup!

Issue #5467, opened 2022-12-18
Cranelift: `iabs.i128` not implemented on AArch64


afonso360:
👋 Hey

### `.clif` Test Case

```
test interpret
test run
target aarch64

function %iabs_i128(i128) -> i128 {
block0(v0: i128):
    v1 = iabs.i128 v0
    return v1
}
; run: %iabs_i128(1) == 1
; run: %iabs_i128(-1) == 1
```

### Steps to Reproduce

* `clif-util test ./the-above.clif`

### Expected Results

The above test to pass

### Actual Results

```
    Finished dev [unoptimized + debuginfo] target(s) in 0.17s
     Running `qemu-aarch64 -L /usr/aarch64-linux-gnu -E LD_LIBRARY_PATH=/usr/aarch64-linux-gnu/lib /home/afonso/git/wasmtime/target/aarch64-unknown-linux-gnu/debug/clif-util test ./lmao.clif`
thread 'worker #1' panicked at 'should be implemented in ISLE: inst = `v1 = iabs.i128 v0`, type = `Some(types::I128)`', cranelift/codegen/src/machinst/lower.rs:753:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
 ERROR cranelift_filetests::concurrent > FAIL: panicked in worker #1: should be implemented in ISLE: inst = `v1 = iabs.i128 v0`, type = `Some(types::I128)`
FAIL ./lmao.clif: panicked in worker #1: should be implemented in ISLE: inst = `v1 = iabs.i128 v0`, type = `Some(types::I128)`
1 tests
Error: 1 failure
```

### Versions and Environment

Cranelift version or commit: main
Operating system: Linux
Architecture: AArch64

p3achyjr:
I can work on this--seems like a reasonable follow-up to https://github.com/bytecodealliance/wasmtime/issues/5466 :)

p3achyjr:
qq--how do we run aarch64 tests if we only have an x86 machine? I don't see qemu in my directory anywhere, even though it seems like that's how CI is being run.

afonso360:
Have a look at the [cross-compiling document](https://github.com/bytecodealliance/wasmtime/blob/main/docs/contributing-cross-compiling.md) It describes how we configure QEMU to run as part of `cargo test` and `cargo run`. A super useful way of doing things!

If you'd like to run the full testsuite we have a separate script that you can use:
```
[...]

PR #7185, opened 2023-10-08
add i128 support for iabs on aarch64

p3achyjr:
As per https://github.com/bytecodealliance/wasmtime/issues/5467, we do not yet have support for iabs on aarch64. This PR adds it, along with the instruction `csinv`.


elliottt (approved):
This looks great! Thanks for enabling the runtest for aarch64!

PR #7196, opened 2023-10-09
Ensure exports are properly namespaced

rylev:
Continuation of #7172 where imports and export _definitions_ are properly namespaced by version by the usage of exports was not.


alexcrichton:
That flaky test, and the prior one which is the same, should be fixed by https://github.com/bytecodealliance/wasmtime/pull/7198. In the meantime gonna keep throwing this at the wall and hope it sticks.

PR #7200, opened 2023-10-09
riscv64: Optimize `gen_bmask` slightly

alexcrichton:
If the input operand is an `icmp` or an `fcmp` there's no need to use `snez` since the output value is already guaranteed to be zero or one.


afonso360 (approved):
Nice!

Issue #6778, opened 2023-07-26
Cranelift: cranelift-object produces AArch64 object files not supported by lld


bjorn3:
### Steps to Reproduce

Write an AArch64 object file with a function using a TLS variable and link it using lld.

### Expected Results

It links.

### Actual Results

```
  = note: ld.lld: error: /home/gh-bjorn3/rust/build/aarch64-unknown-linux-gnu/stage1-tools/cg_clif/build/example/mini_core.mini_core.427c9c3b2a3159b4-cgu.0.rcgu.o:(function get_tls: .text+0x6d4): unknown relocation (513) against symbol mini_core::get_tls::A
          ld.lld: error: /home/gh-bjorn3/rust/build/aarch64-unknown-linux-gnu/stage1-tools/cg_clif/build/example/mini_core.mini_core.427c9c3b2a3159b4-cgu.0.rcgu.o:(function get_tls: .text+0x6d8): unknown relocation (514) against symbol mini_core::get_tls::A
          collect2: error: ld returned 1 exit status
```

### Versions and Environment

Cranelift version or commit: 92ccdcb8d68b586715957c2335031bc057e4876d

Operating system: Linux

Architecture: AArch64

### Extra Info

Cranelift uses `R_AARCH64_TLSGD_ADR_PAGE21` and `R_AARCH64_TLSGD_ADD_LO12_NC` relocations for TLS, which seems to be unsupported by lld. Instead I think TLSDESC has to be used. This is an improvement on the traditional `__tls_get_addr` based TLS lookup: https://www.fsfla.org/~lxoliva/writeups/TLS/paper-lk2006.pdf

PR #7201, opened 2023-10-09
aarch64: Implement TLSDESC for TLS GD accesses

afonso360:
👋 Hey,

This PR replaces the current TLS GD sequence with Thead-Local Storage Descriptors (TLSDESC). TLSDESC is a improved TLS GD model that switches to a custom calling convention that preserves most registers, preventing clobbering and restoring around the regular `__tls_get_addr` call.

Furthermore, TLSDESC supports a number of relaxations that actually avoid the function call entirely in some situations. These relaxations are not allowed to change the code sequence size, so there is no danger of changing the label offsets after being resolved by cranelift.

I've tested this with cg_clif's testsuite. Using both lld and the gnu linker.

Closes: #6778

### DIFF
--- a/cranelift/codegen/src/binemit/mod.rs
+++ b/cranelift/codegen/src/binemit/mod.rs
@@ -64,15 +64,21 @@ pub enum Reloc {
     /// Offset within page of TLVP slot.
     MachOAarch64TlsAdrPageOff12,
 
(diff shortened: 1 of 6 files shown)


alexcrichton (approved):
👍

PR #7206, opened 2023-10-10
Update MSRV to 1.71.0

alexcrichton:
With the 1.73.0 release today update our MSRV to a one-higher version. This will cause 1.73.0 to be tested in CI.


abrown (approved):

alexcrichton:
The most recent failure was a MIRI failure which I've now addressed through refactoring how `Table` with static memory is represented.

PR #7208, opened 2023-10-10
riscv64: Fix encoding for `c.addi4spn`

afonso360:
👋 Hey,

I was running fuzzgen on riscv when it crashed with an illegal instruction. When generating large stack offsets, we can use the compressed instruction `c.addi4spn`, It has a range of `{4,1020}` bytes. However we were accidentally not encoding the MSB on the immediate field.

This meant that the instruction would be encoded with an offset of 0, which is illegal for this instruction.  The fix here is to ensure that the MSB bit is encoded correctly.

### DIFF
--- a/cranelift/codegen/src/isa/riscv64/inst/encode.rs
+++ b/cranelift/codegen/src/isa/riscv64/inst/encode.rs
@@ -468,7 +468,7 @@ pub fn encode_ciw_type(op: CiwOp, rd: WritableReg, imm: u8) -> u16 {
     let mut imm_field = 0;
     imm_field |= ((imm >> 1) & 1) << 0;
     imm_field |= ((imm >> 0) & 1) << 1;
-    imm_field |= ((imm >> 4) & 7) << 2;
+    imm_field |= ((imm >> 4) & 15) << 2;
     imm_field |= ((imm >> 2) & 3) << 6;
 
     let mut bits = 0;
--- a/cranelift/filetests/filetests/isa/riscv64/zca.clif
+++ b/cranelift/filetests/filetests/isa/riscv64/zca.clif
@@ -469,6 +469,43 @@ block0:
 ;   c.addi16sp sp, 0x10
 ;   c.jr ra
 
+function %c_addi4spn_512() -> i64 {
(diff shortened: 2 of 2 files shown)


cranelift/codegen/src/isa/riscv64/inst/encode.rs
```diff
@@ -468,7 +468,7 @@ pub fn encode_ciw_type(op: CiwOp, rd: WritableReg, imm: u8) -> u16 {
     let mut imm_field = 0;
     imm_field |= ((imm >> 1) & 1) << 0;
     imm_field |= ((imm >> 0) & 1) << 1;
-    imm_field |= ((imm >> 4) & 7) << 2;
+    imm_field |= ((imm >> 4) & 15) << 2;
```
fitzgen, line 471: Do we have `emit_tests.rs` for riscv64? I would kind of expect some updated tests in there that exercise this, rather than just the checked in fuzz test.

afonso360: I've been using the disassembled section in the compile tests as a check that the encodings are correct. However, we don't have many compile tests specifically for C instructions. Most of them test only with the Base ISA and don't enable the other extensions so they won't be emitting this instruction.

fitzgen: Right, I guess before merging this I would like to have either a test in `emit_tests.rs` that does some loop over min/max/etc values for this type of immediate and encodes each of them or something else like that in spirit (e.g. a series of `c.addi4spn` instructions with various min/max/etc values in a filetest).

afonso360: I've added a test for both min and max values as well as one that sets a bit in each subfield in the immediate of the instruction. I've also added tests for out of bound values.

I added them into `zca.clif` instead of `emit_tests.rs`, since its slightly awkward to add these there because they are encoding only changes we don't actually know the name of the instruction in cranelift. So the test would look something like `addi a1, sp, 4` but the result is a 2 byte `c.addisp4n`.

fitzgen (approved):
Thanks!

PR #7209, opened 2023-10-10
Introduce API for custom stack memory

rockwotj:
Introduces an API to allow embedders to allocate stacks for fibers using custom memory instead of mmap.

[Discussed on Zulip](https://bytecodealliance.zulipchat.com/#narrow/stream/217126-wasmtime/topic/.E2.9C.94.20Custom.20allocator.20for.20Fibers)

### DIFF
--- a/crates/c-api/src/config.rs
+++ b/crates/c-api/src/config.rs
@@ -3,9 +3,9 @@
 #![cfg_attr(not(feature = "cache"), allow(unused_imports))]
 
 use crate::{handle_result, wasm_memorytype_t, wasmtime_error_t};
-use std::mem::MaybeUninit;
 use std::ops::Range;
 use std::os::raw::c_char;
+use std::ptr;
 use std::{ffi::CStr, sync::Arc};
 use wasmtime::{
     Config, LinearMemory, MemoryCreator, OptLevel, ProfilingStrategy, Result, Strategy,
@@ -362,7 +362,25 @@ unsafe impl MemoryCreator for CHostMemoryCreator {
         reserved_size_in_bytes: Option<usize>,
         guard_size_in_bytes: usize,
     ) -> Result<Box<dyn wasmtime::LinearMemory>, String> {
-        let mut memory = MaybeUninit::uninit();
+        extern "C" fn panic_get_callback(
+            _env: *mut std::ffi::c_void,
+            _byte_size: &mut usize,
+            _maximum_byte_size: &mut usize,
+        ) -> *mut u8 {
+            panic!("a callback must be set");
+        }
+        extern "C" fn panic_grow_callback(
+            _env: *mut std::ffi::c_void,
+            _size: usize,
+        ) -> Option<Box<wasmtime_error_t>> {
+            panic!("a callback must be set");
+        }
+        let mut memory = wasmtime_linear_memory_t {
+            env: ptr::null_mut(),
+            get_memory: panic_get_callback,
+            grow_memory: panic_grow_callback,
+            finalizer: None,
+        };
         let cb = self.new_memory;
         let error = cb(
             self.foreign.data,
@@ -371,11 +389,10 @@ unsafe impl MemoryCreator for CHostMemoryCreator {
             maximum.unwrap_or(usize::MAX),
             reserved_size_in_bytes.unwrap_or(0),
             guard_size_in_bytes,
-            memory.as_mut_ptr(),
+            &mut memory,
         );
         match error {
             None => {
-                let memory = unsafe { memory.assume_init() };
                 let foreign = crate::ForeignData {
                     data: memory.env,
                     finalizer: memory.finalizer,
--- a/crates/fiber/src/unix.rs
+++ b/crates/fiber/src/unix.rs
@@ -38,19 +38,13 @@ use std::ops::Range;
 use std::ptr;
 
 pub enum FiberStack {
-    Mmap {
+    Default {
         // The top of the stack; for stacks allocated by the fiber implementation itself,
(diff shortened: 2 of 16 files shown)


crates/c-api/src/config.rs
```diff
@@ -371,11 +389,10 @@ unsafe impl MemoryCreator for CHostMemoryCreator {
             maximum.unwrap_or(usize::MAX),
             reserved_size_in_bytes.unwrap_or(0),
             guard_size_in_bytes,
-            memory.as_mut_ptr(),
+            &mut memory,
         );
         match error {
             None => {
-                let memory = unsafe { memory.assume_init() };
```
rockwotj, line 378: This is a much safer API, because we don't need to assume they will sent `env` and `finalizer` to nullptr if they don't need them.

crates/wasmtime/src/stack.rs
```diff
@@ -0,0 +1,73 @@
...
+
+use anyhow::Error;
+use wasmtime_fiber::{RuntimeFiberStack, RuntimeFiberStackCreator};
+
+/// A stack creator. Can be used to provide a stack creator to wasmtime
+/// which supplies stacks for async support.
+///
+/// # Safety
+///
+/// This trait is unsafe, as memory safety depends on a proper implemenation
```
fitzgen, line 11:
```suggestion
/// This trait is unsafe, as memory safety depends on a proper implementation
```

crates/wasmtime/src/trampoline.rs
```diff
@@ -41,15 +41,18 @@ fn create_handle(
...
-                host_state,
-                store: StorePtr::new(store.traitobj()),
-                runtime_info,
-                wmemcheck: false,
-                pkey: None,
-            })?;
+        #[cfg(feature = "async")]
+        let allocator = OnDemandInstanceAllocator::new(config.mem_creator.clone(), None, 0);
+        #[cfg(not(feature = "async"))]
+        let allocator = OnDemandInstanceAllocator::new(config.mem_creator.clone());
```
fitzgen, line 47: Instead of changing the constructor of the `OnDemandInstanceAllocator` depending on whether `#[cfg(feature = "async")]`, how about we always leave out the stack creator, but then have a method for configuring it after construction:

```rust
impl OnDemandInstanceAllocator {
    #[cfg(feature = "async")]
    pub fn set_stack_creator(&mut self, stack_creator: Option<Arc<dyn StackCreator>>) {
        self.stack_creator = stack_creator;
    }
}
```

This should make construction sites a little bit simpler.

tests/all/stack_creator.rs
```diff
@@ -0,0 +1,157 @@
+#[cfg(all(not(target_os = "windows"), not(miri)))]
+mod not_for_windows {
+    use anyhow::{bail, Context};
+    use std::{
+        alloc::{GlobalAlloc, Layout, System},
+        ops::Range,
+        ptr::NonNull,
+        sync::Arc,
+    };
+    use wasmtime::*;
```
fitzgen, line 10: Can remove a level of indentation via `#![...]` to put the attribute on the whole `stack_creator` module, rather than defining a submodule inside it:

```suggestion
#![cfg(all(not(target_os = "windows"), not(miri)))]

use anyhow::{bail, Context};
use std::{
    alloc::{GlobalAlloc, Layout, System},
    ops::Range,
    ptr::NonNull,
    sync::Arc,
};
use wasmtime::*;
```

crates/fiber/src/unix.rs
```diff
@@ -75,37 +82,52 @@ impl FiberStack {
...
+    pub fn from_custom(custom: Box<dyn RuntimeFiberStack>) -> io::Result<Self> {
+        Ok(Self::Custom(custom))
+    }
+
     pub fn top(&self) -> Option<*mut u8> {
-        Some(self.top)
+        Some(match self {
+            FiberStack::Mmap { top, len: _ } => *top,
+            FiberStack::Manual { top, len: _ } => *top,
+            FiberStack::Custom(r) => r.top(),
```
fitzgen, line 107: Can we assert page alignment here?

crates/fiber/src/unix.rs
```diff
@@ -75,37 +82,52 @@ impl FiberStack {
...
+        Some(match self {
+            FiberStack::Mmap { top, len } => {
+                let base = unsafe { top.sub(*len) as usize };
+                base..base + len
+            }
+            FiberStack::Manual { top, len } => {
+                let base = unsafe { top.sub(*len) as usize };
+                base..base + len
+            }
+            FiberStack::Custom(r) => r.range(),
```
fitzgen, line 121: And here for both start and end parts of the range

fitzgen (commented):
(Going to leave the C API bits to Alex to review)

crates/fiber/src/unix.rs
rockwotj: Done.

crates/fiber/src/unix.rs
rockwotj: Done.

crates/fiber/src/unix.rs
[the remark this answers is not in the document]
rockwotj: Reverted that back.

fitzgen (approved):
Nice! I looked through the C API changes in more detail as well, and they look good as well.

Thanks @rockwotj!

rockwotj:
Thanks @fitzgen! I messed up some lint warnings when the `async` config is disabled. I've fixed them now.

fitzgen (approved):

PR #7219, opened 2023-10-11
PCC: add basic "memory types".

cfallin:
This PR adds a notion of "memory types" to the PCC (proof-carrying code) framework. To borrow a doc-comment:

A memory type is a struct-like definition -- fields with offsets, each field having a type and possibly an attached fact -- that we can use in proof-carrying code to validate accesses to structs and propagate facts onto the loaded values as well.

Memory types are meant to be rich enough to describe the *layout* of values in memory, but do not necessarily need to embody higher-level features such as subtyping directly. Rather, they should encode an implementation of a type or object system.

Note also that it is a non-goal for now for this type system to be "complete" or fully orthogonal: we have some restrictions now (e.g., struct fields are only primitives) because this is all we need for existing PCC applications, and it keeps the implementation simpler.

While developing this, we originally sketched out a more complete type-system that had arrays, allowed struct fields to be memory types themselves, and was in general more flexible. However, the possibility of cycles, non-statically-sized types, etc., was getting overly complex and I decided to YAGNI-simplify a bit for now. The Doc-comment on the module does sketch out what other types will be needed, and I think those plus a GV-bounded `max` fact will be enough to express PCC facts for all static and dynamic memories and table accesses in Wasmtime. (A partial further development of the more complete system, with some more validator checks for cycles, etc., is on [this WIP branch](https://github.com/cfallin/wasmtime/tree/pcc-memcap-wip).)

Part of this change was

### DIFF
--- a/cranelift/codegen/src/ir/memtype.rs
+++ b/cranelift/codegen/src/ir/memtype.rs
@@ -62,6 +62,9 @@
 //! - A discriminated union is a union of several memory types
 //!   together with a tag field. This will be useful to model and
 //!   verify subtyping/downcasting for Wasm GC, among other uses.
+//!
+//! - Nullability on pointer fields: the fact will hold only if the
+//!   field is not null (all zero bits).
 
 use crate::ir::pcc::Fact;
 use crate::ir::Type;
--- a/cranelift/codegen/src/ir/memtype.rs
+++ b/cranelift/codegen/src/ir/memtype.rs
@@ -5,6 +5,7 @@
 //! can use in proof-carrying code to validate accesses to structs and
 //! propagate facts onto the loaded values as well.
 
+use crate::ir::entities::MemoryType;
 use crate::ir::pcc::Fact;
 use crate::ir::Type;
 use alloc::vec::Vec;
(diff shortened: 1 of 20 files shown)


cranelift/codegen/src/ir/memtype.rs
```diff
@@ -0,0 +1,171 @@
...
+//!   by a global value (GV). This is otherwise just like the "static
+//!   memory" variant described above.
+//!
+//! - A dynamic array is a sequence of struct memory types, with a
+//!   length given by a global value (GV). This is useful to model,
+//!   e.g., tables.
+//!
+//! - A discriminated union is a union of several memory types
+//!   together with a tag field. This will be useful to model and
+//!   verify subtyping/downcasting for Wasm GC, among other uses.
```
fitzgen, line 64: Note for the future that wasm gc will also require nullability, which almost but doesn't quite fit into the idea of discriminant tags.

cranelift/codegen/src/ir/pcc.rs
```diff
@@ -373,9 +354,20 @@ impl FactContext {
...
+                let end_offset: i64 = offset
+                    .checked_add(i64::from(size))
+                    .ok_or(PccError::Overflow)?;
+                let end_offset: u64 =
+                    u64::try_from(end_offset).map_err(|_| PccError::OutOfBounds)?;
+                match &self.memory_types[*ty] {
+                    ir::MemoryTypeData::Struct { size, .. }
+                    | ir::MemoryTypeData::Memory { size } => {
+                        ensure!(end_offset <= *size, OutOfBounds)
+                    }
```
fitzgen, line 367: This could be
```suggestion
                ensure!(
                    self.memory_types[*ty].size().map_or(false, |s| end_offset <= s),
                    OutOfBounds
                );
```

cranelift/filetests/filetests/pcc/fail/simple.clif
```diff
@@ -2,14 +2,25 @@ test compile expect-fail
...
     v2 ! max(64, 0xffff_ffff) = uextend.i64 v1
-    v3 ! points_to(0x1) = iadd.i64 v0, v2
+    v3 ! mem(mt0, 0xffff_ffff) = iadd.i64 v0, v2
+    v4 = load.i8 checked v3
+    return v4
+}
+
+;; Check that the offset in the `mem` is validated too.
+
+function %simple1(i64 vmctx, i32) -> i8 {
```
fitzgen, line 19: I don't think it matters if function names are duplicated but worth making sure

```suggestion
function %simple2(i64 vmctx, i32) -> i8 {
```

crates/wasi-nn/spec
fitzgen, line 1: I don't think this change to the wasi-nn spec submodule was intended was it?

fitzgen (approved):
r=me with nitpick-y suggestions that you can take or leave but also a more important point about a git submodule that should definitely be addressed before landing

cranelift/codegen/src/ir/memtype.rs
cfallin: Added another note for this in the doc-comment!

cranelift/codegen/src/ir/pcc.rs
cfallin: I thought a bit about this, but I think on balance I like the current form better, as we'll need to go back to it once we have dynamic limits.

cranelift/filetests/filetests/pcc/fail/simple.clif
cfallin: Fixed!

crates/wasi-nn/spec
cfallin: Ah, yes, unintended from a missed `git submodule update`; rebased out!

PR #7220, opened 2023-10-11
threads: log every `wait` and `notify`

abrown:
When troubleshooting deadlocks in WebAssembly modules, it is important to understand which `wait` instructions are still pending a `notify`. It would be nice to have some kind of `--warn-deadlock-after=1s` flag available that would poll the parking lot for `wait`s hanging past the time limit, but I realized the real value would be to tie the `wait` instruction (through CLIF) to the original source code, if debug information were available. This did not seem to be entirely feasible, since CLIF loses the original Wasm source context (is this true?) and I was not confident that we would be able to use `addr2line` to map from Wasm instructions to source (e.g., see @cfallin's [issue](https://github.com/gimli-rs/addr2line/issues/265)).

Instead, this change simply logs each valid `wait` and `notify` execution, leaving it to the user to figure out which one is hanging (should not be too difficult) and how to map this back to their source code (more difficult).

### DIFF
--- a/crates/runtime/src/memory.rs
+++ b/crates/runtime/src/memory.rs
@@ -565,6 +565,7 @@ impl SharedMemory {
     /// Implementation of `memory.atomic.notify` for this shared memory.
     pub fn atomic_notify(&self, addr_index: u64, count: u32) -> Result<u32, Trap> {
         validate_atomic_addr(&self.0.def.0, addr_index, 4, 4)?;
+        log::trace!("memory.atomic.notify(addr={addr_index}, count={count})");
         Ok(self.0.spot.unpark(addr_index, count))
     }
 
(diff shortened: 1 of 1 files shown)


crates/runtime/src/memory.rs
```diff
@@ -565,6 +565,7 @@ impl SharedMemory {
     /// Implementation of `memory.atomic.notify` for this shared memory.
     pub fn atomic_notify(&self, addr_index: u64, count: u32) -> Result<u32, Trap> {
         validate_atomic_addr(&self.0.def.0, addr_index, 4, 4)?;
+        log::trace!("memory.atomic.notify(addr={addr_index}, count={count})");
```
pchickey, line 568: instead of using the format string, use `log::trace!("memory.atomic.notify", addr=addr_index, count=count);` to provide structured values.

pchickey (commented):
lgtm, modulo a style change across all of these traces

crates/runtime/src/memory.rs
abrown: I tried that but it complains: I think to use the key-value syntax we need to set `target: ...,` up front and still provide a string at the end (see [`log::log!`](https://docs.rs/log/latest/log/macro.log.html))?

pchickey: oh, my bad. I just realized this is `log` and the key value syntax only works with the `tracing` crate, and I forgot we use `log` in wasmtime-runtime in order to keep the dependencies as small as possible, but `tracing` in wasmtime-wit-bindgen, wasi, etc.

So, your PR is good to go as is. I don't think we need should using `log` vs `tracing` down in the runtime now, unless someone really wants to.

pchickey (approved):

Issue #7215, opened 2023-10-11
cranelift: Allow allocating heaps in clif filetests


afonso360:
👋 Hey,

This was something that @jameysharp brought up during today's cranelift meeting.

#### Feature

We could add support to our runtests to allocate a heap memory region and pass it to the function under test.

#### Benefit

This allows us to do two things:

* We can start testing load fusion for vector instructions in the cranelift fuzzer for x64. We can't do this currently because they require the address to be aligned to 16bytes, and we can't guarantee that yet for stack addresses.

* We can test our alias analysis memflags using this. We could for example map different heaps to a different alias analysis bits, and ensure that our optimizations work.

#### Implementation

We used to have something similar to this in https://github.com/bytecodealliance/wasmtime/pull/3302 that was then removed in https://github.com/bytecodealliance/wasmtime/pull/5386

We had an annotation like `; heap: static, size=0x1000, ptr=vmctx+0, bound=vmctx+8` which would pass the address of the heap via the `vmctx` parameter.

That used to involve some special handling for `vmctx` parameters that I think is incompatible with what we do today (don't quote me on this).

But we could do something similar like:

```
function %test(i64, 10) -> i64 {
...
}
; heap: <heap0> size=16, align=8
; run: %test(<heap0>, 10) == 10
```

And the `<heap0>` annotation would be resolved with the address of the heap when calling the test.

#### Alternatives

 An alternative to testing load fusion for vectors, is to add stack alignment flags to our stack slots. But doing the alias analysis bits is slightly harder although there are probably workarounds there too.

jameysharp:
Your observations in #7225 reminded me that, in both fuzz tests and runtests, we can theoretically pass pointers from one function to another.

So one approach in runtests would be to get a stackslot pointer in one function and pass it to a callee, which can then treat the pointer as heap or vmctx or whatever. In fuzzgen, we could add a fake "address" type to the set of function parameter types we might generate; callers would fill such arguments in using the same selection logic as addresses for loads and stores, and callees would add those arguments to the set of addresses which loads and stores can choose from.

Since we don't do any inter-procedural optimization, that ensures that code generation can't do any hypothetical optimizations that would rely on knowing the pointer refers to the stack.

We still need the ability to specify alignment on stack slots though.

afonso360:
That's a great idea. We might not even have to do anything too hacky since we don't yet use reftypes for anything we could use those to signal, this is a valid address for loads and stores.

PR #7225, opened 2023-10-12
fuzzgen: Add Alias Analysis Memflags to Loads and Stores

afonso360:
👋 Hey,

This is a follow up to #7215 and @jameysharp's comments on yesterdays cranelift meeting. @jameysharp suggested that we could add heaps to clif to start testing our alias analysis pass.

But having thought about it, there is nothing special about the "Heap" flag on `MemFlags`. It is only used by our alias analysis pass to optimize some accesses. The same happens with the other flags ("Table" and "VmCtx"). Additionally we already have a way to partition memory in the fuzzer, we call it stack slots. And we already enforce that memory accesses can't cross stack slots, so we also don't run into the risk of doing a partial store/load on a different region.

This PR tags each stack slot with a different Alias Analysis category (Although region might be a better name). Once we have that we just have to ensure that all accesses are correctly tagged, and we should be able to test all 4 categories.

I'm opening this as a draft since I still want to get some fuzzing on this. I suspect that if there is a bug to be found here, It's going to take a while to get those conditions.

### DIFF
--- a/cranelift/fuzzgen/src/function_generator.rs
+++ b/cranelift/fuzzgen/src/function_generator.rs
@@ -131,7 +131,14 @@ fn insert_stack_load(
 ) -> Result<()> {
     let typevar = rets[0];
     let type_size = typevar.bytes();
-    let (slot, slot_size) = fgen.stack_slot_with_size(type_size)?;
+    let (slot, slot_size, category) = fgen.stack_slot_with_size(type_size)?;
+
+    // `stack_load` doesen't support setting MemFlags, and it does not set any
+    // alias analysis bits, so we can only emit it for `Other` slots.
+    if category != AACategory::Other {
+        return Err(arbitrary::Error::IncorrectFormat.into());
+    }
+
     let offset = fgen.u.int_in_range(0..=(slot_size - type_size))? as i32;
 
     let val = builder.ins().stack_load(typevar, slot, offset);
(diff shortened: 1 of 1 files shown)


afonso360:
This has been fuzzing today pretty much the entire day and it hasn't crashed yet, so looks all good there.

cranelift/fuzzgen/src/function_generator.rs
```diff
@@ -1208,11 +1262,11 @@ where
     }
 
     /// Finds a stack slot with size of at least n bytes
-    fn stack_slot_with_size(&mut self, n: u32) -> Result<(StackSlot, StackSize)> {
+    fn stack_slot_with_size(&mut self, n: u32) -> Result<(StackSlot, StackSize, AACategory)> {
         let first = self
             .resources
             .stack_slots
-            .partition_point(|&(_slot, size)| size < n);
+            .partition_point(|&(_slot, size, _category)| size < n);
```
fitzgen, line 1269: Are stack slots sorted by size? Could we document that in a comment on the field definition?

cranelift/fuzzgen/src/function_generator.rs
```diff
@@ -1533,12 +1591,16 @@ where
...
+            // Generate one Alias Analysis Category for each slot
+            let category = *self.u.choose(AACategory::all())?;
+
+            self.resources.stack_slots.push((slot, bytes, category));
         }
 
         self.resources
             .stack_slots
-            .sort_unstable_by_key(|&(_slot, bytes)| bytes);
+            .sort_unstable_by_key(|&(_slot, bytes, _category)| bytes);
```
fitzgen, line 1603: Okay, I see the sorting done here. Yeah, would be good to document this invariant at the field so that it is easier for other folks extending or reading this stuff to figure it out.

fitzgen (approved):
Nice!

PR #7226, opened 2023-10-12
c-api/cmake: Support custom build options

rockwotj:
This can be useful if you want change to set the features wasmtime
builds with in CMake


rockwotj:
Our use case we only want to turn on the async feature and disable others, so we can configure like so (semicolons are lists and prevent quoting amazingly):

```bash
cmake .. -DWASMTIME_USER_CARGO_BUILD_OPTIONS='--no-default-features;--features=async'
```

alexcrichton (approved):

PR #7227, opened 2023-10-12
c-api: expose memory_init_cow

rockwotj:
The main motivation for exposing this is if you're using a custom host allocator then you can't use these, so you might as well not generate them in the first place.


fitzgen (approved):
Thanks!

PR #7155, opened 2023-10-05
winch(x64): Add support for table instructions

saulecabrera:
This change adds support for the following table insructions: `elem.drop`, `table.copy`, `table.set`, `table.get`, `table.fill`, `table.grow`, `table.size`, `table.init`.

This change also introduces partial support for the `Ref` WebAssembly type, more conretely the `Func` heap type, which means that all the table instructions above, only work this WebAssembly type as of this change.

Finally, this change is also a small follow up to the primitives introduced in https://github.com/bytecodealliance/wasmtime/pull/7100, more concretely:

* `FnCall::with_lib`: tracks the presence of a libcall and ensures that any result registers are freed right when the call is emitted.
* `MacroAssembler::table_elem_addr` returns an address rather than the value of the address, making it convenient for other use cases like `table.set`.

--

prtest:full

PR #7035, opened 2023-09-14
winch: Add floating point rounding instructions for sse41

elliottt:
Add floating point rounding instructions using `roundss` and `roundsd`. The fall-back path is currently unimplemented, but once we have support for libcalls we can use the same fallback that the cranelift lowerings use.

PR #7228, opened 2023-10-12
winch: Add a subset of known libcalls and improve call emission

saulecabrera:
This change is a follow up to:
- https://github.com/bytecodealliance/wasmtime/pull/7155
- https://github.com/bytecodealliance/wasmtime/pull/7035

One of the objectives of this change is to make it easy to emit function calls at the MacroAssembler layer, for cases in which it's challenging to know ahead-of-time if a particular functionality can be achieved natively (e.g. rounding and SSE4.2).  The original implementation of function call emission, made this objective difficult to achieve and it was also difficult to reason about.
I decided to simplify the overall approach to function calls as part of this PR; in essence, the `call` module now exposes a single function `FnCall::emit` which is reponsible of gathtering the dependencies and orchestrating the emission of the call. This new approach deliberately avoids holding any state regarding the function call for simplicity.

This change also standardizes the usage of `Callee` as the main entrypoint for function call emission, as of this change 4 `Callee` types exist (`Local`, `Builtin`, `Import`, `FuncRef`), each callee kind is mappable to a `CalleeKind` which is the materialized version of a callee which Cranelift understands.

This change also moves the creation of the `BuiltinFunctions` to the `ISA` level given that they can be safely used accross multiple function compilations.

Finally, this change also introduces support for some of the "well-known" libcalls and hooks those libcalls at the `MacroAssembler::float_round` callsite.

--

prtest:full

### DIFF
--- a/winch/codegen/src/codegen/builtin.rs
+++ b/winch/codegen/src/codegen/builtin.rs
@@ -5,16 +5,57 @@ use crate::{
     codegen::env::ptr_type_from_ptr_size,
     CallingConvention,
 };
-use wasmtime_environ::{BuiltinFunctionIndex, PtrSize, WasmType};
+use cranelift_codegen::ir::LibCall;
+use std::sync::Arc;
+use wasmtime_environ::{BuiltinFunctionIndex, PtrSize, VMOffsets, WasmType};
+
+#[derive(Copy, Clone)]
+pub(crate) enum BuiltinType {
+    /// Dynamic built-in function, derived from the VMContext.
+    Dynamic {
+        /// The offset of the built-in function.
+        offset: u32,
+        /// The built-in function base, relative to the VMContext.
+        base: u32,
+    },
+    /// A known libcall.
+    /// See [`cranelift_codegen::ir::LibCall`] for more details.
+    Known(LibCall),
+}
+
+impl BuiltinType {
+    /// Create a new dynamic built-in function type.
+    pub fn dynamic(offset: u32, base: u32) -> Self {
+        Self::Dynamic { offset, base }
+    }
+
+    /// Create a new known built-in function type.
+    pub fn known(libcall: LibCall) -> Self {
+        Self::Known(libcall)
+    }
+}
+
+#[derive(Clone)]
+pub struct BuiltinFunction {
+    inner: Arc<BuiltinFunctionInner>,
+}
+
+impl BuiltinFunction {
+    pub(crate) fn sig(&self) -> &ABISig {
+        &self.inner.sig
+    }
+
+    pub(crate) fn ty(&self) -> BuiltinType {
+        self.inner.ty
+    }
+}
 
 /// Metadata about a builtin function.
-pub(crate) struct BuiltinFunction {
+pub struct BuiltinFunctionInner {
     /// The ABI specific signature of the function.
-    pub sig: ABISig,
-    /// The offset of the builtin function
-    pub offset: u32,
-    /// The builtin function base, relative to the VMContext.
-    pub base: u32,
+    sig: ABISig,
+    /// The built-in function type.
+    ty: BuiltinType,
 }
 
 macro_rules! declare_function_sig {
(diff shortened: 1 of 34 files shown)


saulecabrera:
@elliottt this is the change that I mentioned earlier this week during the Cranelift weekly meeting.

winch/codegen/src/codegen/env.rs
```diff
@@ -32,17 +29,46 @@ pub struct TableData {
...
+    /// A built-in function.
+    Builtin(BuiltinFunction),
+}
+
+impl Callee {
+    /// Get the built-in function metadata.
+    ///
+    /// # Panics
+    /// This function panics if the [`Callee`] is not a built-in function.
+    pub fn get_builtin(&self) -> &BuiltinFunction {
```
fitzgen, line 49: Maybe name this `unwrap_builtin`?

winch/codegen/src/codegen/env.rs
```diff
@@ -54,15 +80,13 @@ pub struct CalleeInfo {
 ///
 /// Contains all information about the module and runtime that is accessible to
 /// to a particular function during code generation.
-pub struct FuncEnv<'a, P: PtrSize> {
+pub struct FuncEnv<'a, 'b: 'a, 'c: 'b, P: PtrSize> {
```
fitzgen, line 83: In my experience, when we start to get many lifetime parameters like this, it helps to give them more descriptive names.

winch/codegen/src/codegen/mod.rs
```diff
@@ -22,21 +18,21 @@ pub(crate) use call::*;
 mod control;
 pub(crate) use control::*;
 mod builtin;
-pub(crate) use builtin::*;
+pub use builtin::*;
 
 /// The code generation abstraction.
-pub(crate) struct CodeGen<'a, M>
+pub(crate) struct CodeGen<'a, 'b: 'a, 'c: 'b, M>
```
fitzgen, line 24: ditto

fitzgen (approved):

winch/codegen/src/codegen/builtin.rs
```diff
@@ -5,16 +5,57 @@ use crate::{
...
+use cranelift_codegen::ir::LibCall;
+use std::sync::Arc;
+use wasmtime_environ::{BuiltinFunctionIndex, PtrSize, VMOffsets, WasmType};
+
+#[derive(Copy, Clone)]
+pub(crate) enum BuiltinType {
+    /// Dynamic built-in function, derived from the VMContext.
+    Dynamic {
+        /// The offset of the built-in function.
+        offset: u32,
```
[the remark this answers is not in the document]
saulecabrera: Yeah, but note that this index is the built-in function's index multiplied by the pointer size, but index works for me.

winch/codegen/src/codegen/builtin.rs
```diff
@@ -5,16 +5,57 @@ use crate::{
...
+        /// The built-in function base, relative to the VMContext.
+        base: u32,
```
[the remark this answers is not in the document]
saulecabrera: I don't think that this will work (unless I'm missing something). The reason why I'm storing both separately is that loading the builtin's address requires two loads:
- base = load(vmctx + base)
- builtin_addr = load(base + index)

winch/codegen/src/codegen/env.rs
saulecabrera: As I was fixing this I realized that this was actually dead code, so I ended up removing `get_builtin` and `get_info` entirely.

winch/codegen/src/codegen/env.rs
```diff
@@ -54,15 +55,13 @@ pub struct CalleeInfo {
 ///
 /// Contains all information about the module and runtime that is accessible to
 /// to a particular function during code generation.
-pub struct FuncEnv<'a, P: PtrSize> {
+pub struct FuncEnv<'a, 'translation: 'a, 'data: 'translation, P: PtrSize> {
```
elliottt, line 58: Similarly to the `CodeGenContext` parameters, can this be simplified down to using the `'a` lifetime for everything?

elliottt (commented):
This is looking great, I'm so happy to see the default cases for fp rounding operations now!

winch/codegen/src/codegen/context.rs
```diff
@@ -27,25 +28,37 @@ use std::ops::RangeBounds;
 /// generation process. The code generation context should
 /// be generally used as the single entry point to access
 /// the compound functionality provided by its elements.
-pub(crate) struct CodeGenContext<'a> {
+pub(crate) struct CodeGenContext<'a, 'b: 'a> {
```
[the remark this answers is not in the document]
saulecabrera: It actually outlives 'a; the changes in this PR make it so the `compiled_function` function at the ISA level gets a reference to `BuiltinFunctions`, which will outlive any of the other structs created in the body of the `compile_function` function (e.g. `FuncEnv` and `VMOffsets`) and on top of that, there's is a third lifetime ('`data`, previously `'c`) which represents the lifetime of any references to the original WebAssembly binary through `wasmtime_environ::ModuleTranslation`.

winch/codegen/src/codegen/env.rs
saulecabrera: Yeah, [the same comment](https://github.com/bytecodealliance/wasmtime/pull/7228#discussion_r1358321683) applies here too.

saulecabrera:
I've addressed all of the review comments (thanks @fitzgen and @elliottt). I'll go ahead and merge this, but happy to follow up on any of the comments if need be.

Issue #6528, opened 2023-06-06
Winch Core Wasm Opcode Support


saulecabrera:
During the [debugging meeting on June 1, 2023](https://github.com/bytecodealliance/meetings/blob/main/SIG-Debugging/2023/debugging-06-01.md), we discussed opening an issue to track the missing Core Wasm opcodes in Winch. The objective of this issue is to make it easier for potential contributors to know where to start. 

This list is intended to be a comprehensive compilation of the missing opcodes required to achieve parity with Core Wasm(*). Each entry also includes an owner, indicating that work on a particular opcode is currently in progress.

(*) Excluding SIMD and Reference Types to start, as drafted in the initial RFC. We can open dedicated issues as we get to those and other proposals.

cc/ @fitzgen @itsrainy 

- [x]     Unreachable (@saulecabrera)
- [x]     Block (@saulecabrera)
- [x]     Loop (@saulecabrera)
- [x]     If (@saulecabrera)
- [x]     Else, (@saulecabrera)
- [x]     Br (@saulecabrera)
- [x]     BrIf (@saulecabrera)
- [x]     Return (@saulecabrera)
- [x]     CallIndirect (@saulecabrera)
- [x]     BrTable (@saulecabrera)
- [x]     Drop (@saulecabrera)
- [x]     Select (@saulecabrera)
- [x]     LocalTee (@saulecabrera)
- [x]     GlobalGet (@saulecabrera)
- [x]     GlobalSet (@saulecabrera)
- [x]     I32Load (@saulecabrera)
- [x]     I64Load (@saulecabrera)
- [x]     F32Load (@saulecabrera)
- [x]     F64Load (@saulecabrera)
- [x]     I32Load8S (@saulecabrera)
- [x]     I32Load8U (@saulecabrera)
- [x]     I32Load16S (@saulecabrera)
- [x]     I32Load16U (@saulecabrera)
- [x]     I64Load8S (@saulecabrera)
- [x]     I64Load8U (@saulecabrera)
- [x]     I64Load16S (@saulecabrera)
- [x]     I64Load16U (@saulecabrera)
- [x]     I64Load32S (@saulecabrera)
- [x]     I64Load32U (@saulecabrera)
- [x]     I32Store (@saulecabrera)
- [x]     I64Store (@saulecabrera)
- [x]     F32Store (@saulecabrera)
- [x]     F64Store (@saulecabrera)
- [x]     I32Store8 (@saulecabrera)
[...]

PR #7230, opened 2023-10-12
Add support for float sqrt operators to Winch

jeffcharles:
Adding support for `f32.sqrt` and `f64.sqrt` Wasm instructions to Winch. Part of #6528.


saulecabrera (approved):
LGTM!

PR #7232, opened 2023-10-12
Minor preview2 related test cleanups

pchickey:
* there was a duplicate test for the behavior of panics
* the test for stream pollable children lifetimes was different than all of the others to invoke, so split it into two tests (in preview2 instead of cli) and assert that one of them traps, the other succceeds


elliottt (approved):
Nice!

PR #7029, opened 2023-09-12
Start to port Wasmtime to the new wasi-io API with resources.

sunfishcode:
This imports the new APIs posted in WebAssembly/wasi-io#46.

PR #7233, opened 2023-10-12
remove logging from wit deps

pchickey:
I believe this was missed in the code review for #7029, it was introduced as an intermediate step in that PR but wasnt fully removed.

Also, the command extended world can now use `include wasi:cli/command;` which is a really nice improvement.


elliottt (approved):
:fire:

PR #7234, opened 2023-10-12
Implement stream splice, delete forward

pchickey:
This PR implements `output-stream.splice` and `blocking-splice`.

The wit definitions for these functions has been changed to take a `borrow<input-stream>`, rather than the owned resource.

Splice performs the equivalent to:
* uses `check-write` to determine how much input is accepted by the `output-stream`.
* uses `read` to read up to that much (or the `len` bound provided by caller) from the `input-stream`
* uses `write` to write the read data to the `output-stream`.

By implementing this in the host using the `HostInputStream` and `HostOutputStream` traits, the read and write are done in terms of `Bytes` representation, so there are zero copies of the underlying data made as part of the splice.

A blocking splice waits for both read and write readiness before performing those operations.

The `output-stream.forward` method has been deleted from the wit. We believe that the fusion of an `output-stream` into an `input-stream` (which gives up ownership of both resources, and happens perpetually in the background) is a worthwhile concept to implement some day, but it raises various questions on how progress and errors can be reported, and whether the background work can outlive the calling component, so rather than open that can of worms now we are going to ask users to implement it with their own event loops using `splice` for now, and look at adding it again in the future.

### DIFF
--- a/crates/wasi/src/preview2/host/io.rs
+++ b/crates/wasi/src/preview2/host/io.rs
@@ -124,71 +124,53 @@ impl<T: WasiView> streams::HostOutputStream for T {
 
     async fn splice(
         &mut self,
-        _dst: Resource<OutputStream>,
-        _src: Resource<InputStream>,
-        _len: u64,
+        dest: Resource<OutputStream>,
+        src: Resource<InputStream>,
+        len: u64,
     ) -> StreamResult<u64> {
-        // TODO: We can't get two streams at the same time because they both
-        // carry the exclusive lifetime of `ctx`. When [`get_many_mut`] is
-        // stabilized, that could allow us to add a `get_many_stream_mut` or
-        // so which lets us do this.
-        //
-        // [`get_many_mut`]: https://doc.rust-lang.org/stable/std/collections/hash_map/struct.HashMap.html#method.get_many_mut
-        /*
-        let s: &mut Box<dyn crate::InputStream> = ctx
-            .table_mut()
-            .get_input_stream_mut(src)
-            ?;
-        let d: &mut Box<dyn crate::OutputStream> = ctx
-            .table_mut()
-            .get_mut(dst)
-            ?;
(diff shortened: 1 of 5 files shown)


crates/wasi/src/preview2/host/io.rs
```diff
@@ -124,71 +124,53 @@ impl<T: WasiView> streams::HostOutputStream for T {
...
-
-        let bytes_spliced: u64 = s.splice(&mut **d, len).await?;
-
-        Ok(bytes_spliced)
-        */
-        todo!("stream splice is not implemented")
+        let len = len.try_into().unwrap_or(usize::MAX);
+        let output = self.table_mut().get_mut(&dest)?;
+        let permit = output.check_write()?;
+        let _ = output;
```
alexcrichton, line 134: no longer necessary I think?

crates/wasi/src/preview2/host/io.rs
```diff
@@ -124,71 +124,53 @@ impl<T: WasiView> streams::HostOutputStream for T {
...
-
-        let bytes_spliced: u64 = s.splice(&mut **d, len).await?;
-
-        Ok(bytes_spliced)
-        */
-
-        todo!("stream forward is not implemented")
+        let output = self.table_mut().get_mut(&dest)?;
+        output.ready().await;
+        let _ = output;
```
alexcrichton, line 167: Same as above, this and below may no longer be necessary?

alexcrichton (approved):

crates/wasi/src/preview2/host/io.rs
alexcrichton: Oh? I would imagine that NLL fixes this because `let _: &mut T = val;` AFAIK does nothing in terms of lifetime inference

pchickey: oh. yes, you're right. im not sure what error message i was trying to get around by writing it that way, but it isnt necessary. at any rate i like the way its written now.

Issue #5109, opened 2022-10-24
More control over fuel consumption


coolreader18:
#### Feature

A set of new APIs that provide more control over fuel consumption:

```rust
impl Config {
    fn fuel_cost(&mut self, f: impl Fn(WasmOpcode) -> u64 + Send + Sync + 'static) -> &mut Self;
}
enum WasmOpcode {
    Unreachable,
    Nop,
    Block,
    Loop,
    // ...
}
impl Store {
    fn fuel_remaining(&self) -> Option<u64>;
    /// Returns an error if the store is not configured for fuel consumption
    fn set_fuel(&mut self, value: u64) -> Result<()>;
}
#[non_exhaustive]
struct OutOfFuelError;
```

#### Benefit

Allows more direct control over fuel consumption and filling, letting users set custom fuel costs for each operand if they deem some to be more costly/intensive in their environment, as well letting them set precisely how much fuel they might allow a single function call or set of function calls. `OutOfFuelError` being made public allows users to check when a Trap was caused by fuel running out.

#### Implementation

I have an implementation for the custom fuel cost configuration, and the methods on `Store` should be fairly trivial.

#### Alternatives

`fuel_remaining` and `set_fuel` can be approximated by `consume_fuel(0)` and `if new_fuel > fuel_remaining() { add_fuel(delta) } else { consume_fuel(delta) }` respectively. `fuel_cost` cannot be emulated, so the alternative is to just accept wasmtime's default cost function. As an alternative in designing the API, the `fuel_cost` closure could take a `&wasmparser::Operator<'_>` as it did in the original version of this issue. However, that would require exposing `wasmparser` in wasmtime's public api, which is undesirable.

coolreader18:
Oh, looks like wasmtime-fuzzing has a shim for `set_fuel` the same as I described:

https://github.com/bytecodealliance/wasmtime/blob/main/crates/fuzzing/src/oracles.rs#L790-L802

howjmay:
Hi I am curious why #5220 can help us set customized fuel for each opcode?

RReverser:
Looks like #7240 only adds `reset_fuel`, I don't think it should've auto-closed this issue?

In particular,

> `fn fuel_cost(&mut self, f: impl Fn(WasmOpcode) -> u64 + Send + Sync + 'static) -> &mut Self;`

this would be still desirable.

rockwotj:
[...]

PR #5220, opened 2022-11-07
Add Store::set_fuel and Store::fuel_remaining

coolreader18:
Resolves #5109. As discussed there, adds `Store::set_fuel` and `Store::fuel_remaining`, which sets the absolute amount of fuel in the store and returns the amount of fuel remaining in the store respectively. In updating the examples and tests, I was already able to change some stuff to be nicer/express intent more clearly. (`set_fuel` provides more of a use for `fuel_consumed`! e.g. see examples/fuel.rs).

This feature doesn't have a maintainer assigned.

PR #7240, opened 2023-10-13
Support reset_fuel in store APIs

rockwotj:
Add a simpler interface to reset the amount of fuel in a store.

Fixes: https://github.com/bytecodealliance/wasmtime/issues/5109

A continuation of https://github.com/bytecodealliance/wasmtime/pull/5220, which has stalled. To make it more obvious that `consumed_fuel` is reset, I've changed the name to this method to be `reset_fuel`.

### DIFF
--- a/crates/c-api/include/wasmtime/store.h
+++ b/crates/c-api/include/wasmtime/store.h
@@ -165,6 +165,23 @@ WASM_API_EXTERN void wasmtime_context_gc(wasmtime_context_t* context);
 WASM_API_EXTERN wasmtime_error_t *wasmtime_context_add_fuel(wasmtime_context_t *store, uint64_t fuel);
 
 /**
+ * \brief Set fuel to this context's store for wasm to consume while executing.
+ *
+ * For this method to work fuel consumption must be enabled via
+ * #wasmtime_config_consume_fuel_set. By default a store starts with 0 fuel
+ * for wasm to execute with (meaning it will immediately trap).
+ * This function must be called for the store to have
+ * some fuel to allow WebAssembly to execute.
+ *
+ * Note that at this time when fuel is entirely consumed it will cause
+ * wasm to trap. More usages of fuel are planned for the future.
+ *
+ * If fuel is not enabled within this store then an error is returned. If fuel
+ * is successfully added then NULL is returned.
+ */
+WASM_API_EXTERN wasmtime_error_t *wasmtime_context_set_fuel(wasmtime_context_t *store, uint64_t fuel);
+
+/**
  * \brief Returns the amount of fuel consumed by this context's store execution
  * so far.
  *
--- a/crates/c-api/include/wasmtime/store.h
+++ b/crates/c-api/include/wasmtime/store.h
@@ -156,8 +156,10 @@ WASM_API_EXTERN void wasmtime_context_gc(wasmtime_context_t* context);
  * This function must be called for the store to have
  * some fuel to allow WebAssembly to execute.
  *
- * Note that at this time when fuel is entirely consumed it will cause
- * wasm to trap. More usages of fuel are planned for the future.
+ * Note by default when fuel is entirely consumed it will cause
+ * wasm to trap. If async_support is enabled, you can use 
+ * #wasmtime_context_out_of_fuel_async_yield and the async APIs to yield
+ * execution instead.
  *
  * If fuel is not enabled within this store then an error is returned. If fuel
  * is successfully added then NULL is returned.
@@ -173,8 +175,10 @@ WASM_API_EXTERN wasmtime_error_t *wasmtime_context_add_fuel(wasmtime_context_t *
  * This function must be called for the store to have
  * some fuel to allow WebAssembly to execute.
  *
- * Note that at this time when fuel is entirely consumed it will cause
- * wasm to trap. More usages of fuel are planned for the future.
+ * Note by default when fuel is entirely consumed it will cause
+ * wasm to trap. If async_support is enabled, you can use 
+ * #wasmtime_context_out_of_fuel_async_yield and the async APIs to yield
+ * execution instead.
  *
  * If fuel is not enabled within this store then an error is returned. If fuel
  * is successfully added then NULL is returned.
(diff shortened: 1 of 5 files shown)


crates/c-api/include/wasmtime/store.h
```diff
@@ -164,6 +164,23 @@ WASM_API_EXTERN void wasmtime_context_gc(wasmtime_context_t* context);
...
+ * \brief Set fuel to this context's store for wasm to consume while executing.
+ *
+ * For this method to work fuel consumption must be enabled via
+ * #wasmtime_config_consume_fuel_set. By default a store starts with 0 fuel
+ * for wasm to execute with (meaning it will immediately trap).
+ * This function must be called for the store to have
+ * some fuel to allow WebAssembly to execute.
+ *
+ * Note that at this time when fuel is entirely consumed it will cause
+ * wasm to trap. More usages of fuel are planned for the future.
```
fitzgen, line 177: Is this just copy-pasted? Because with the Rust API you can yield on out-of-fuel and I thought that the C API supported async as well now, but I am not sure if the fuel bits for that are connected or not.

fitzgen: Ah yeah I see that it is copy-pasted. Mind removing this paragraph here and in the other place?

If you want to go above and beyond, you could replace it with a bit about how one might configure execution to trap or async yield on out of fuel via https://docs.rs/wasmtime/latest/wasmtime/struct.Store.html#method.out_of_fuel_async_yield and https://docs.rs/wasmtime/latest/wasmtime/struct.Store.html#method.out_of_fuel_trap

fitzgen (commented):
Thanks! Just one documentation nitpick below to be addressed and then we can merge this.

crates/c-api/include/wasmtime/store.h
rockwotj: Done! Thanks for catching :)

alexcrichton:
One thing that's perhaps worth noting is that we have quite a litany of fuel-related methods and we seem to keep adding more over time. This isn't necessarily a problem, nor do I think this is a blocker for this PR, but I think it indicates that we haven't done a great job of picking the right abstraction related to fuel. Or perhaps they're all independently motivated but given the ad-hoc nature of the additions I suspect we can do better. The fuel methods we have right now are:

* [`fuel_consumed`](https://docs.rs/wasmtime/latest/wasmtime/struct.Store.html#method.fuel_consumed)
* [`fuel_remaining`](https://docs.rs/wasmtime/latest/wasmtime/struct.Store.html#method.fuel_remaining)
* [`add_fuel`](https://docs.rs/wasmtime/latest/wasmtime/struct.Store.html#method.add_fuel)
* [`consume_fuel`](https://docs.rs/wasmtime/latest/wasmtime/struct.Store.html#method.consume_fuel)

To be clear this is mostly food-for-thought. I do not wish to block this PR, and if @fitzgen approves we should merge.

rockwotj:
I'm happy to come up with a better API before this PR (or make breaking changes, not sure what the policy is on that). 

TBH we use fuel today and I didn't realize that consumed fuel was tracked independently of remaining fuel. Personally I probably would have pushed fuel consumed to embedders and make them track it based on how much fuel was remaining after each call into wasmtime. But I know taking away functionality is hard, but most people I've seen use fuel in the issue tracker are just attempting to prevent the engine from hogging too much CPU or getting stuck in a loop. Also I think if that functionality didn't exist you could potentially only have a set_fuel method and get_fuel method and let embedders calculate the delta. But instead there are different methods to remove fuel vs adding fuel which complicates the API vs just 'tell us how much fuel exists'. 

Not sure if that makes since, but that would be the simplest primitive and API, as well as still supporting all the functionality (even if some would be pushed to embedders like tracking consumption).

Additionally, I personally find out_of_fuel_async_yield a complicated API in terms of the number of times to inject fuel. IMO the simpler API is either a method to auto reset fuel after yield to a specific value (and to trap you change out of gas behavior in the store then reset fuel to 0 and then resume the VM [as another note I'm not sure if it's valid to drop the future and still use the VM for yielding support]) or there is a yield every N fuel and you trap after fuel runs out (there is no adding fuel automatically from the API although it might do that under the hood). 

If its valid to make breaking changes, I'd be happy to take a stab at a simpler fuel API based on the above in a followup PR.

rockwotj:
RE breaking changes I found https://github.com/bytecodealliance/wasmtime/blob/main/docs/stability-release.md which my reading of that means we would need an RFC to simplify these APIs? Not sure if this would be major or not but what I'm proposing seems to be.

rockwotj:
The more I think about this the more I like the appeal of three fuel APIs in Store, I believe you could still achieve everything there is today:

get_fuel() -> Result<u64>
set_fuel(u64) -> Result<()>
fuel_async_yield_interval(u64) -> Result<()>

Basically you can set and get the amount of fuel and then to enable async support and yielding we say every N fuel consumed we will yield. This also means you could get rid of the out of gas trap APIs and instead have an API to disable the yielding interval

Internally there is a current active fuel, and a reserve "tank" of fuel for restoring after a yield. The get and set operate on the combined amount of active fuel and fuel in the tank, then the current enum for out of gas behavior just becomes an optional amount of fuel to take from the tank after a yield. If the yield interval is cleared we just take all the fuel from the tank and add it to the active amount.

That feels way simpler to explain and implement. With a tighter API surface to boot. TBH i don't fully understand how fuel_adj works in the current implementation.

alexcrichton:
I want to be sure to reiterate that this isn't something you're expected to take on @rockwotj, although if you're willing we won't stop you of course! Additionally breaking changes are ok to Wasmtime and happen relatively frequently. We only require major breaking changes to go through RFCs, although that hasn't happened in quite some time. This change to fuel would not constitue "major" and could be done through consensus of various stakeholders.

I'd have to set aside some time to think about this personally, so if you're interested, would you be up for filing an issue with your idea and we could discuss on there? If not that's totally ok and I can file an issue in the future as well.

rockwotj:
Haha I get [nerdsniped](https://xkcd.com/356/) easily :smile:

I created an issue here with my strawman proposal: https://github.com/bytecodealliance/wasmtime/issues/7255

fitzgen (approved):
LGTM! We can land this in the meantime as the details in that other issue are hashed out.

PR #7249, opened 2023-10-16
cranelift: Add egraph rules for `bswap`

alexcrichton:
WebAssembly doesn't currently have a byte-swapping instruction so this commit pattern matches what LLVM currently generates to simplify to a single `bswap` instruction in CLIF which can be lowered within each backend to respective native instructions.

### DIFF
--- a/cranelift/filetests/filetests/isa/riscv64/wasm/zbb.wat
+++ b/cranelift/filetests/filetests/isa/riscv64/wasm/zbb.wat
@@ -1,0 +1,88 @@
+;;! target = "riscv64"
+;;! compile = true
+;;! settings = ["has_zbb", "opt_level=speed"]
+
+(module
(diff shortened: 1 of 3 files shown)


cranelift/filetests/filetests/isa/riscv64/wasm/zbb.wat
```diff
@@ -0,0 +1,88 @@
+;;! target = "riscv64"
+;;! compile = true
+;;! settings = ["has_zbb", "opt_level=speed"]
```
fitzgen, line 3: I think we can also write a wasm -> clif version of this test by not using `compile = true`. That would be kinda nice since the rule itself is target independent.

fitzgen (approved):
r=me with suggestion below

PR #7250, opened 2023-10-16
riscv64: Optimize `select_spectre_guard`

alexcrichton:
This commit optimizes some `select_spectre_guard` patterns which are frequently generated with the wasm backend when dynamic bounds checks are enabled by pattern-matching when one of the values being selected is zero. This enables shaving off a few instructions which can be constant folded away.


fitzgen (approved):
Nice!

Issue #1067, opened 2020-02-28
Bitfield Extract/Create instructions


moonheart08:
This would add two new instructions to cranelift, `bextr` and `bmak`. They provide a easy to optimize, and easy to generate mechanism for bitfield manipulation.

## bextr

    a = bextr target, size, offset

bextr will extract a _n_ bit large field at the provided offset, and return it in `a`, shifted so that the LSB of the field is also the LSB of `a`.
`a`'s type is inferred from `target`

### Visual example

    offset = 4
    size = 4
    typeof target = i8
    
    xxxx **** 
    \__/----V
    0000 xxxx

## bmak

    a = bmak target, size, offset

bmak will fill a _n_ bit large field at the provided offset with 1s, the rest with 0. The generated field will then be applied to `target` using a bitwise AND, and the result will be returned in `a`.
`a`'s type is inferred from `target`
bmak is designed so that you can simply use a bitwise or to combine the fields.

### Visual Example

    offset = 2
    size = 4
    typeof target = i8
    
    xx0101xx
      \__/
      /  \
    00010100

cfallin:
We discussed this during old-issue triage today and agreed that it is probably better to have users implement this behavior with shifts/masks and then pattern-match that during isel if a particular target ISA has a nice instruction for it. This fits better with the general trend of making CLIF a simple core IR, especially as not every ISA has such instructions -- this would have to be re-expanded in backends otherwise, and only after the point that that expansion could be optimized. I'll close this issue but thank you for the discussion!

moonheart08:
> We discussed this during old-issue triage today and agreed that it is probably better to have users implement this behavior with shifts/masks and then pattern-match that during isel if a particular target ISA has a nice instruction for it. This fits better with the general trend of making CLIF a simple core IR, especially as not every ISA has such instructions -- this would have to be re-expanded in backends otherwise, and only after the point that that expansion could be optimized. I'll close this issue but thank you for the discussion!

7 years later with a stronger understanding of compilers and architectures, and I agree this is probably the right approach. Yet another issue I didn't remember creating resolved today!

Rafferty97:
Hi, I'm writing a compiler that does a lot of bit manipulation, and using cranelift for the backend.

I've noticed that quite a few instances where two instructions could be compressed into a single UBFM or SBFM instruction.

I'd be willing to open a PR that adds the appropriate ISLE rules, if it's likely to be merged?

cfallin:
[...]

PR #7252, opened 2023-10-16
riscv64: Implement some `Zbs` extension lowerings

alexcrichton:
This commit extends the riscv64 backend with instructions from the `Zbs` extension. I did this awhile back for the bmi{1,2} instructions for x64 and this is mostly for completeness's sake rather than performance or anything like that, and because I had a long plane ride!

### DIFF
--- a/cranelift/codegen/src/isa/riscv64/lower.isle
+++ b/cranelift/codegen/src/isa/riscv64/lower.isle
@@ -710,6 +710,8 @@
   (if-let x_imm (replicated_imm5 x))
   (rv_vand_vi y x_imm (unmasked) ty))
 
+;; `bclr{,i}` specializations from `zbs`
+
 (rule 13 (lower (has_type $I32 (band x (bnot (ishl (i64_from_iconst 1) y)))))
   (if-let $true (has_zbs))
   (rv_bclr x (rv_andi y (imm12_const 31))))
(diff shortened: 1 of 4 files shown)


cranelift/codegen/src/isa/riscv64/lower.isle
```diff
@@ -710,6 +710,64 @@
   (if-let x_imm (replicated_imm5 x))
   (rv_vand_vi y x_imm (unmasked) ty))
 
+;; `bclr{,i}` specializations from `zbs`
+
+(rule 13 (lower (has_type $I32 (band x (bnot (ishl (i64_from_iconst 1) y)))))
+  (if-let $true (has_zbs))
+  (rv_bclr x (rv_andi y (imm12_const 31))))
```
afonso360, line 717: A small suggestion here would be to change this to `(u64_sub (ty_bits ty) 1)` so that we could also apply these patterns to i16 and i8.

afonso360: Also, perhaps we should add equivalent rules for these with swapped arguments. i.e. `(band (bnot ...) y)` we have them for the other  instructions but not for `bclr`.

afonso360 (approved):
Nice! Some small suggestions if you'd like to expand on this.

With so many patterns here I think it might be worth eventually creating intermediate instructions in CLIF and moving some of these rules to the egraphs layer. But this looks good to me! (cc: #1067)

I've also been running the fuzzer on this branch and it hasn't found anything yet.

alexcrichton:
I'm not sure if these are really all that common enough in practice to warrant new CLIF instructions, especially given that the lowerings aren't too bad to have, but it's also true though that if every platform has a variation of this then it might be reasonable to add.

I do think though that the suite of instructions on x64 is different than on riscv64, for example riscv64 has more immediate-related instructions because on x64 you can fit an entire 32-bit immediate into an instruction but you can't do that on riscv64. That means that `bclri` is motivated on riscv64 (or at least that's my inference as to the assumption), whereas on x64 there's only the `bclr` equivalent (or roughly speaking I think)

PR #7260, opened 2023-10-16
Add an asynchronous `wasi:io/output-stream::forward` method

tschneidereit:
Also add language to `wasi:http/outgoing-body::finish` explaining how it interacts with pending/in-process forwarding operations.

Together, these changes enable use cases where an instance forwards an incoming request's or response's body to an outgoing request/response and then terminates without waiting for the entire body to be streamed.

### DIFF
--- a/crates/wasi-http/wit/deps/io/streams.wit
+++ b/crates/wasi-http/wit/deps/io/streams.wit
@@ -275,5 +275,50 @@ interface streams {
             /// The number of bytes to splice
             len: u64,
         ) -> result<u64, stream-error>;
+
+        /// Forward the entire contents of an `input-stream` to an `output-stream`
+        /// without blocking.
+        ///
+        /// Returns immediately, with the forwarding operation happening concurrently,
+        /// and only after any previously queued forwarding operations have completed.
+        /// Can be called multiple times, causing the input streams to be queued up
+        /// in order.
+        ///
+        /// `forward` takes ownership of `src`, ensuring that the forwarding operation
+        /// is unobservable by and cannot be influenced by the calling component.
+        ///
+        /// If forwarding any of the queued-up streams results in an error, the
+        /// `future-forward-result` will be resolved to a `result` containing that
+        /// error as well as a list containing the remaining, non-forwarded streams.
+        forward: func(
+            /// The stream to forward.
+            src: input-stream,
+        ) -> result<future-forward-result>;
+    }
+
+    /// The following block defines a special resource type used by the
+    /// `wasi:io/output-stream` interface to emulate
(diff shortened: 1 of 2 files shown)


crates/wasi-http/wit/deps/io/streams.wit
```diff
@@ -275,5 +275,50 @@ interface streams {
...
+        /// - The option indicates readiness.
+        /// - The outer result indicates you are allowed to get the inner result at
+        ///   most once. subsequent calls after ready will return an error here.
+        /// - The inner result indicates whether the forwarding operation succeeded
+        ///   and returns the number of bytes written if so.
+        /// - The tuple in the inner result's error case contains the error that
+        ///   occurred, as well as a list of any remaining streams that were not
+        ///   yet completely forwarded.
+        ///   The first stream in this list might have been partially forwarded.
+        get: func() -> option<result<result<u64, tuple<stream-error, list<input-stream>>>>>;
```
pchickey, line 319: I don't think we should provide multiple input-streams here, there is only one that has been taken ownership of by the `forward` function so lets only give one back here. The caller can hold onto handles for all of their future-forward-results and unwrap each of them into an `some(ok(err((_error, input-stream))))`

pchickey (commented):
* how is write readiness of the output-stream defined - I believe it should not become ready for writing until all forwarded input-streams have been fully sent.
* if `future-forwarded-result` is dropped, what happens? I believe that we should continue to forward no matter what, but we need to make that explicit.
* if execution traps, what happens? I think in the case of a trap we should define forwarding to stop "at some point afterwards" - I don't think we want to be on the hook to keep forwarding after a trap, nor give a guarantee that it terminates promptly.

pchickey:
More feedback, based on discussion with @elliottt:
* Do we need to express an output-stream that can potentially have an unbounded number of input-streams forwarded to it? It would perhaps be much simpler if `forward` consumed both the `input-stream` and `output-stream`, and `future-forward-result` gave back the `output-stream` as well as the `u64`  in the `some(ok(ok(...))` case.
* I am concerned this could be too powerful, and force us to make some difficult implementation choices, when applied to all streams. We have the use case of forwarding http bodies. Could we instead implement this as a static method on `outgoing-body` as an alternative to `finish` which is something like `finish-by-forwarding: static func(this: outgoing-body, forward: input-stream) -> some-future`. That still requires us to keep that `input-stream` alive beyond the lifetime of the wasm execution, which could be the same can of worms...

alexcrichton:
I believe the current conclusion is to defer this to after the 0.2.0 release of WASI, so I'm going to close this as it's probably best to pick back up discussion after that happens, but please reopen if you disagree

PR #7261, opened 2023-10-16
Add accessors and mutators for all state of `wasi:http/outgoing-request` and `outgoing-response`

tschneidereit:
This change has two purposes:
    
1. adding the accessors enables treating the resource's representation of all state as canonical, instead of requiring a component to retain a copy of all state contained in the resource.
    
2. Adding mutators for all state supports use cases where it's required to change a request's/response's state after writing to its body. This is in particular the case when implementing preexisting systems or specifications, such as the WHATWG fetch spec used in JavaScript environments.

(Stacked on #7260)

### DIFF
--- a/crates/wasi-http/wit/deps/http/types.wit
+++ b/crates/wasi-http/wit/deps/http/types.wit
@@ -96,8 +96,22 @@ interface types {
       path-with-query: option<string>,
       scheme: option<scheme>,
       authority: option<string>,
-      headers: borrow<headers>
+      headers: headers
     );
+
+    method: func() -> method;
+    set-method: func(method: method);
+
+    path-with-query: func() -> option<string>;
+    set-path-with-query: func(path-with-query: option<string>);
+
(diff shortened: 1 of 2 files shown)


tschneidereit:
I'll ensure that both copies of the `wit` files are in sync once I've worked through any feedback

pchickey:
This is much easier to accept than #7260 which I think we want to see a tested implementation for in order to land. b663ccf, on the other hand, is trivial. Can you rebase this PR to be just that commit?

pchickey:
I sent this upstream as https://github.com/WebAssembly/wasi-http/pull/65, I'll work on an implementation in another PR

elliottt:
I think this is redundant with main now, as we've implemented the upstreamed mutators. Do you mind if we close this PR @tschneidereit ?

alexcrichton:
I'm going to close this due to @elliottt's prior comment and the main changes having been incorporated.

PR #7262, opened 2023-10-16
PCC: rework default-facts somewhat.

cfallin:
This removes the need for the awkward "max-range fact is subsumed by anything" rule noted by @fitzgen in [this
comment](https://github.com/bytecodealliance/wasmtime/pull/7231#discussion_r1358573147). It also makes checking a little more efficient and logically clear, as only the facts that the frontend/producer added are verified, rather than all default facts as well.


fitzgen (approved):
Nice!

PR #7263, opened 2023-10-17
PCC: switch "max" facts to "range" facts with two-sided ranges.

cfallin:
This is needed for soundness when verifying accesses to memtype fields: it's not enough to know that we're accessing an offset in `0` up to `field_offset` inclusive, we need to know the access is actually to `field_offset`.

The simplest change that validates this turned out to be the most general one: making ranges two-sided rather than one-sided. The transform is *mostly* mechanical, but a few new tests verify that ranges are updated on both sides, and some fail-tests verify that "fuzzily imprecise" pointers to struct fields fail to validate.

### DIFF
--- a/cranelift/codegen/src/ir/pcc.rs
+++ b/cranelift/codegen/src/ir/pcc.rs
@@ -128,19 +128,25 @@ pub enum PccError {
 #[derive(Clone, Debug, Hash, PartialEq, Eq)]
 #[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
 pub enum Fact {
-    /// A bitslice of a value (up to a bitwidth) is less than or equal
-    /// to a given maximum value.
+    /// A bitslice of a value (up to a bitwidth) is within the given
+    /// integer range.
     ///
     /// The slicing behavior is needed because this fact can describe
     /// both an SSA `Value`, whose entire value is well-defined, and a
     /// `VReg` in VCode, whose bits beyond the type stored in that
     /// register are don't-care (undefined).
-    ValueMax {
+    Range {
         /// The bitwidth of bits we care about, from the LSB upward.
         bit_width: u16,
+        /// The minimum value that the bitslice can take
+        /// (inclusive). The range is unsigned: the specified bits of
+        /// the actual value will be greater than or equal to this
+        /// value, as evaluated by an unsigned integer comparison.
+        min: u64,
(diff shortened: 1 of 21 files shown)


cranelift/codegen/src/ir/pcc.rs
```diff
@@ -128,36 +128,52 @@ pub enum PccError {
...
-    /// to a given maximum value.
+    /// A bitslice of a value (up to a bitwidth) is within the given
+    /// integer range.
     ///
     /// The slicing behavior is needed because this fact can describe
     /// both an SSA `Value`, whose entire value is well-defined, and a
     /// `VReg` in VCode, whose bits beyond the type stored in that
     /// register are don't-care (undefined).
-    ValueMax {
+    ValueRange {
```
fitzgen, line 138: :bike: :paintbrush:  Maybe we should just call this `Range`? That matches the text format.

cranelift/codegen/src/ir/pcc.rs
```diff
@@ -353,23 +403,26 @@ impl<'a> FactContext<'a> {
...
+                max,
+            } if *bit_width == width => {
+                let min = match min.checked_mul(u64::from(factor)) {
+                    Some(min) => min,
+                    None => return None,
+                };
                 let max = match max.checked_mul(u64::from(factor)) {
                     Some(max) => max,
-                    None => return Some(minimal_fact),
+                    None => return None,
```
fitzgen, line 418: This can use `?` propagation now.

cranelift/codegen/src/ir/pcc.rs
```diff
@@ -388,36 +441,48 @@ impl<'a> FactContext<'a> {
...
+                bit_width,
+                min,
+                max,
+            } if *bit_width == width => {
+                let min = match min.checked_add(offset) {
+                    Some(min) => min,
+                    None => {
+                        return None;
+                    }
+                };
```
fitzgen, line 460: Can use `?` here as well.

fitzgen (approved):

cranelift/codegen/src/ir/pcc.rs
cfallin: Ah, yeah, that makes more sense!

cranelift/codegen/src/ir/pcc.rs

cranelift/codegen/src/ir/pcc.rs

PR #7272, opened 2023-10-17
Put versions in all WASI WIT files

### LINKED ISSUE #7171: wasi: Add version numbers to all WIT files and packages
This should be done, or decided not to be done, before finalizing preview2.

alexcrichton:
This commit starts exercising the versioning feature of WIT by ensuring that all WASI descriptions have a version associated with them. The version chosen is 0.2.0 which reflects the upcoming "preview 2" release where in theory 0.1.0 was claimed by preview1. This is intended to stay as 0.2.0 for now and we'll determine how best to update these numbers in the future once preview2 is released.

Closes #7171


pchickey (changes requested):
Until we have a vote on 0.2.0 I don't want to risk any binaries being out in the world with that version in them - can we please use `0.2.0-rc-2023-10-17` per https://docs.google.com/presentation/d/1t8rTwkJA28IvFvyqy3K49jy8Sm0xUsk-vZssFAruD8o/edit#slide=id.g22e9d051ccb_0_15

alexcrichton:
Sounds reasonable to me yeah, although I'm not a fan of typing unwieldy version strings so that's motivated me to write up https://github.com/bytecodealliance/wasm-tools/pull/1255, https://github.com/bytecodealliance/wit-bindgen/pull/703, and some further improvements for wasmtime's own bindgen. That should enable avoid forcing the need for versions in more places throughout the tooling. It'll still be required in the WIT syntax itself but I think that should be fixable with Peter's `wit` tool or similar in the future.

alexcrichton:
On second though I'm gonna decouple these. There's some wasmtime improvements here to write versions less but the rest in the Rust sources will go away once I update wit-bindgen and wasm-tools which I'll do separately. Otherwise though the versions should be updated to `0.2.0-2023-10-18` now

pchickey (approved):
Excellent, really happy to see this moving forward!

PR #7274, opened 2023-10-17
PCC: initial end-to-end integration with Wasmtime's static memories.

cfallin:
This PR wires up end-to-end support for validating the compilation of accesses to static memories in Wasmtime. It creates "memory types" for the vmcontext struct and the memories it points to, and it adds PCC annotations to the relevant memory pointers. On aarch64 (where instruction semantics for PCC are implemented), this validates a simple `i32.load` function body with default memory configuration.

Note that facts are not yet propagated through egraph rewrites, so optimization has to be disabled for this to work; that support will come next.

### DIFF
--- a/cranelift/codegen/src/legalizer/globalvalue.rs
+++ b/cranelift/codegen/src/legalizer/globalvalue.rs
@@ -31,8 +31,8 @@ pub fn expand_global_value(
             base,
             offset,
             global_type,
-            readonly,
-        } => load_addr(inst, func, base, offset, global_type, readonly, isa),
+            flags,
+        } => load_addr(inst, func, base, offset, global_type, flags, isa),
         ir::GlobalValueData::Symbol { tls, .. } => symbol(inst, func, global_value, isa, tls),
         ir::GlobalValueData::DynScaleTargetConst { vector_type } => {
             const_vector_scale(inst, func, vector_type, isa)
@@ -96,7 +96,7 @@ fn load_addr(
     base: ir::GlobalValue,
     offset: ir::immediates::Offset32,
     global_type: ir::Type,
-    readonly: bool,
+    flags: ir::MemFlags,
     isa: &dyn TargetIsa,
 ) {
     // We need to load a pointer from the `base` global value, so insert a new `global_value`
@@ -116,17 +116,11 @@ fn load_addr(
         pos.ins().global_value(ptr_ty, base)
     };
 
-    // Global-value loads are always notrap and aligned. They may be readonly.
-    let mut mflags = ir::MemFlags::trusted();
-    if readonly {
-        mflags.set_readonly();
-    }
-
     // Perform the load.
     pos.func
         .dfg
         .replace(inst)
-        .load(global_type, mflags, base_addr, offset);
+        .load(global_type, flags, base_addr, offset);
 }
 
 /// Expand a `global_value` instruction for a symbolic name global.
--- a/cranelift/codegen/src/legalizer/globalvalue.rs
+++ b/cranelift/codegen/src/legalizer/globalvalue.rs
@@ -82,7 +82,11 @@ fn iadd_imm_addr(
             .special_param(ir::ArgumentPurpose::VMContext)
             .expect("Missing vmctx parameter")
     } else {
-        pos.ins().global_value(global_type, base)
+        let gv = pos.ins().global_value(global_type, base);
+        if let Some(fact) = &pos.func.global_value_facts[base] {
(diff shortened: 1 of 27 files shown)


cranelift/codegen/src/legalizer/globalvalue.rs
```diff
@@ -63,6 +63,22 @@ fn vmctx_addr(inst: ir::Inst, func: &mut ir::Function) {
...
     func.dfg.change_to_alias(result, vmctx);
     func.layout.remove_inst(inst);
+
+    crate::trace!(
+        "expanding vmctx gv {}: fact = {:?}, vmctx arg = {}",
+        global_value,
+        func.global_value_facts[global_value],
+        vmctx
+    );
+
```
fitzgen, line 73: Not 100% sure we want to leave this trace in.

cranelift/filetests/filetests/pcc/succeed/gv_fact.clif
fitzgen, line 1: Can we also test when the `vmctx` gv has a fact but the block param doesn't, so we exercise the fact copying in tests?

cranelift/wasm/src/code_translator/bounds_checks.rs
```diff
@@ -364,19 +374,63 @@ fn compute_addr(
...
     let base_and_index = pos.ins().iadd(heap_base, index);
+
+    if let Some(ty) = pcc_memtype {
+        // TODO: handle memory64 as well (pass in the CLIF type of
+        // `index`).
+        pos.func.dfg.facts[base_and_index] = Some(Fact::Mem {
+            ty,
+            min_offset: 0,
+            max_offset: u64::from(u32::MAX),
+        });
```
fitzgen, line 409: We can actually use `heap.index_type` already, we don't need another parameter. I think we should at least assert on that value now, if we don't update the actual max now, so that it isn't silently incorrect.

cranelift/wasm/src/code_translator/bounds_checks.rs
```diff
@@ -364,19 +374,63 @@ fn compute_addr(
...
     if offset == 0 {
         base_and_index
     } else {
         // NB: The addition of the offset immediate must happen *before* the
         // `select_spectre_guard`, if any. If it happens after, then we
         // potentially are letting speculative execution read the whole first
         // 4GiB of memory.
-        pos.ins().iadd_imm(base_and_index, offset as i64)
+        //
+        // TODO: add a fact on the result value.
```
fitzgen, line 420: I think this is done? Or is this referring to something else we still haven't done yet?

crates/cranelift/src/func_environ.rs
```diff
@@ -212,6 +224,73 @@ impl<'module_environment> FuncEnvironment<'module_environment> {
...
+                    let data_mt = func.create_memory_type(ir::MemoryTypeData::Memory {
+                        size: static_bound
+                            .checked_add(plan.offset_guard_size)
+                            .expect("Memory plan has overflowing size plus guard"),
+                    });
+                    log::trace!(
+                        "data_mt: static bound is {:x}, offset guard size is {:?}",
+                        static_bound,
+                        plan.offset_guard_size
+                    );
```
fitzgen, line 258: Similarly, I don't think this trace will be useful long term.

fitzgen (approved):

crates/cranelift/src/func_environ.rs
cfallin: Removed!

cranelift/wasm/src/code_translator/bounds_checks.rs
cfallin: Yep, just stale and I missed it -- thanks!

cranelift/wasm/src/code_translator/bounds_checks.rs
cfallin: Ah, yes, good point -- added that assert.

cranelift/filetests/filetests/pcc/succeed/gv_fact.clif
cfallin: Added!

cranelift/codegen/src/legalizer/globalvalue.rs
cfallin: Removed!

cfallin:
Just a heads-up that the last commit has a few blessed expected-output changes on the Wasm memory access tests, because we had removed the special case `vmctx` GV lowering (so there are a few more aliases now). No functional change so I think it should be fine...

PR #7282, opened 2023-10-18
Gate many CLI/Wasmtime features behind Cargo features

alexcrichton:
This PR is borne out of discussions from last week's CG meetings about the suitability of Wasmtime in embedded scenarios. One of the primary concerns that arose was the binary size of the Wasmtime engine being too large for these scenarios. In my experience binary size is one aspect of Rust where it doesn't come for free so this aligns with my expectations. That being said I've yet to see a case in Rust where a particular binary footprint couldn't be achieved, hence this PR.

One thing I've noticed is that many folks evaluating Wasmtime seem to look at the size of the `wasmtime` executable itself when performing size comparisons. While this likely doesn't reflect the true size of what a custom embedding would be it's nevertheless an easy comparison that can be made. To that effect I've added a number of Cargo features to the `wasmtime-cli` crate to disable functionality up to and including `wasmtime compile` for example. This PR enables building a `wasmtime` exectuable that is stripped down to the bare bones - all it can do is run precompiled WebAssembly files. That being said it should still be a full-featured runtime at this time where only some optional features such as DWARF processing for `addr2line` are disabled.

Additionally in this PR I've added a new chapter to the Wasmtime book about building a minimal build of Wasmtime. This explains not only the existence of the `--no-default-features` build flag but additionally the consequences in terms of functionality and binary size. Furthermore this section additionally directly expands on tips and tricks for building a minimal binary using various Rust compiler flags and nightly features. This section shows how the `wasmtime` CLI is 130M in debug mode and the smallest build producable after this PR is 2.1M. 

The goal of this PR is to get the ball rolling on size-related optimizations, not necessarily be the end-all-be-all. Further optimizations to binary size will likely require more invasive changes about how we develop Wasmtime or similar. For example we may have to lose the relatively rich error information Wasmtime currently has to get the next level of significant win. That being said I'm at least personally lacking a bit in direction for where to go next to optimize. My hope is that by showcasing a much smaller `wasmtime` CLI executable it shows to folks that Wasmtime for these use cases is viable, even if it's not suitable just yet. For example the CLI uses crates such as `clap` which won't be present in custom embeddings. Having an actual custom embedding to optimize for will be useful. In lieu of this I plan on continuing to poke to see what we can do.

One option as a result of this PR is to build a `wasmtime-lite` executable on CI and upload it to GitHub Actions/Releases as well. I'm not totally convinced that'd be too useful so I haven't done that. Another option though is to add a CI builder which builds a minimal binary and verifies that it doesn't go above a particular threshold to catch regressions in binary size (e.g. if a new feature is added ensure that it can be disabled with `--no-default-features`). I've left this for a future PR as well.

I should also note that my hope is that when optimizing for binary size it would not hinder the development of Wasmtime anywhere else. In that sense it should ideally be easy for anyone to understand how the conditionally compiled code works and it shouldn't appear as a maze of cfgs to wade through to get things working. While it's definitely more `#[cfg]` than before, I'm particularly interested if folks feel this is too onerous to maintain or burdensome. If so I'd be interested in improving how this is all designed.


Cargo.toml
```diff
@@ -300,6 +302,7 @@ wmemcheck = ["wasmtime/wmemcheck"]
 wat = ["dep:wat"]
 cache = ["dep:wasmtime-cache", "wasmtime-cli-flags/cache"]
 parallel-compilation = ["wasmtime-cli-flags/parallel-compilation"]
+compile-out-logging = ["log/max_level_off", "tracing/max_level_off"]
```
fitzgen: Maybe name this `disable-logging`?

src/bin/wasmtime.rs
```diff
@@ -47,17 +46,19 @@ enum Subcommand {
     #[cfg(feature = "cache")]
     Config(wasmtime_cli::commands::ConfigCommand),
     /// Compiles a WebAssembly module.
-    Compile(CompileCommand),
+    #[cfg(feature = "cranelift")]
+    Compile(wasmtime_cli::commands::CompileCommand),
```
fitzgen, line 50: Would be nice to add some empty lines between these variant definitions now that they are getting a bit noisier with all the `cfg`s and all that.

docs/examples-minimal.md
```diff
@@ -0,0 +1,205 @@
+# Building a Minimal Wasmtime embedding
+
+Wasmtime embeddings may wish to optimize for binary size and runtime footprint
+to fit on a small system. This documentation is intended to guide some features
+of Wasmtime and how to best produce a minimal build of Wasmtime.
+
+## Building a minimal CLI
+
+> *Note*: the exact numbers in this section were last updated on 2023-10-18 on a
+> macOS host. They should provide a general ballpark estimate but should be
```
fitzgen, line 10: Should probably specify the architecture as well.

fitzgen (approved):
Looks great to me, the new page in the guide looks great as well.

docs/examples-minimal.md
```diff
@@ -0,0 +1,220 @@
...
+`CARGO_PROFILE_RELEASE_CODEGEN_UNITS=1`.
+
+```
+$ export CARGO_PROFILE_RELEASE_OPT_LEVEL=s
+$ export CARGO_PROFILE_RELEASE_PANIC=abort
+$ export CARGO_PROFILE_RELEASE_LTO=true
+$ export CARGO_PROFILE_RELEASE_CODEGEN_UNITS=1
+$ cargo build --release --no-default-features --features disable-logging
+$ ls -l ./target/release/wasmtime
+-rwxr-xr-x@ 1 root  root    3.3M Oct 18 08:43 target/release/wasmtime
```
bjorn3, line 131: Stripping debuginfo using `-Cstrip=debuginfo` should also help a fair bit, right?

alexcrichton: Oh excellent suggestion I forgot about that! Turns out the majority of the win from `-Zbuild-std` was stripping debuginfo

crates/cli-flags/src/lib.rs
```diff
@@ -89,7 +89,7 @@ wasmtime_option_group! {
         /// Whether or not to enable parallel compilation of modules.
         pub parallel_compilation: Option<bool>,
         /// Whether to enable proof-carrying code (PCC)-based validation.
-        pub enable_pcc: Option<bool>,
+        pub pcc: Option<bool>,
```
alexcrichton, line 92: I should highlight this as a drive-by change by me (cc @cfallin), but this'll rename `-C enable-pcc` to `-C pcc` (most other options omit the "enable" or "disable" to have everything be "if you specify it it's enabled unless you say `=n` in which case you're asking to disable it")

abrown (commented):
Nice work documenting the pruning! I've always sort of wondered what the minimum size would be.

kayabaNerve:
This merged `jitdump` and `vtune` into a new, single feature `profiling`, adding the `ittapi` dependency to anyone who solely wants `jitdump`. It also didn't properly update documentation as https://docs.rs/wasmtime/latest/wasmtime/ still details a `jitdump`/`vtune` feature (with no `profiling` feature).

While I'm not against the existence of a new `profiling` feature, could the original two features be restored with `profiling` activating both of them?

alexcrichton:
Ah I merged those since they were both on-by-default and there didn't at the time appear to be much benefit to sharding further. Out of curiosity is this something where you're concerned about binary size of the ittapi dependency? Or is it otherwise not building for you for one reason or another? 

One reason I unified them was I also placed the guest profiler behind the same `profiling` feature and by having a complete on/off switch it would enable conditionally compiling out even more code as well in theory too. I was hoping to avoid a feature-per-profiler, so if possible I'd prefer to solve the problem without splitting it back up, but that depends on the issue that you're having and if it can only be solved by adding a vtune feature back that seems ok too

kayabaNerve:
As someone who refers to an upstream which disables default features, then re-enables features as needed, this did introduce the ittapi dependency for me as I updated it in my downstream. To be clear, I agree with that policy.

I don't personally care about ittapi's effect on binary size. I care about the fact I now have ittapi in scope and will have to, eventually, read through and audit ittapi to ensure it's non-malicious (or import the Bytecode Alliance's vet statement for it, which I can assume exists). While I can recognize the library is hosted and maintained by Intel, the actual publishers of the crate are three individuals, any of who could be compromised.

TL;DR It increases my supply chain and complicates the package graph for no benefit to me, making this a downgrade in wasmtime's utility.

Again. I don't mind the profiling feature. I just wish I could still access the original jitdump feature. Accordingly, I believe this can be implemented *without* hurting the performance/effects/code-gen of the profiling feature.

alexcrichton:
That makes sense yeah, thanks for explaining! If it's ok with you I'd prefer to stay the current course, however. Wasmtime doesn't guarantee that we won't add new dependencies with new version of Wasmtime even with default features disabled. That being said we do guarantee that all dependencies used by Wasmtime have a vetted version by BA maintainers.

Otherwise I'd personally prefer to avoid sharding the `profiling` feature into a feature-per-strategy as well as that feels like a bit of an overload of Cargo features (we already have a lot). But again I don't feel super strongly about this, just a slight preference.

kayabaNerve:
I fully understand dependencies may increase. That does not affect my commentary on needless dependencies having increased.

As for staying the course, I can respect your preference of simplicity. If in the future, you change your mind and would accept a PR re-defining profiling as a superset of both features, please let me know.

PR #7284, opened 2023-10-18
riscv64: Improve scalar_to_vector for constants

### LINKED ISSUE #7189: riscv64: Improve `scalar_to_vector` codegen for constants
👋 Hey,

This is a fairly simple improvement. Cranelift's [`scalar_to_vector` instruction](https://docs.rs/cranelift-codegen/latest/cranelift_codegen/ir/trait.InstBuilder.html#method.scalar_to_vector) grabs a scalar value and places it at the bottom of a vector register.

We translate this into the [`vmerge.vxm` instruction](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#1115-vector-integer-merge-instructions) in the RISC-V backend.

This is all fairly ok, but we don't have a rule to match the `vmerge.vim` instruction that can place an immediate value instead of a register.

Implementing this should be fairly simple, we can create a rule similar to the [existing ones](https://github.com/bytecodealliance/wasmtime/blob/fef8a90f258483ac6c2e470d3de277a743d5d392/cranelift/codegen/src/isa/riscv64/lower.isle#L2340-L2353) but using a `imm5_from_value` extractror:
```
(rule 2 (lower (has_type (ty_vec_fits_in_register ty) (scalar_to_vector (imm5_from_value x))))
```

This `imm5_from_value` extractor doesn't exist yet, but we already have an [equivalent in the `replicated_imm5` extractor](https://github.com/bytecodealliance/wasmtime/blob/fef8a90f258483ac6c2e470d3de277a743d5d392/cranelift/codegen/src/isa/riscv64/inst.isle#L1837C33-L1837C33). It should be fairly easy to extract that into it's own separate extractor for use in this new rule.

See the existing `uimm5_from_value` extractor for help.

If anyone needs help working on this issue let me know!

BieVic:
This simple improvement adds a rule for the `vmerge.vim` instruction and an `imm5_from_value` extractor.
Closes #7189

### DIFF
--- a/cranelift/codegen/src/isa/riscv64/inst.isle
+++ b/cranelift/codegen/src/isa/riscv64/inst.isle
@@ -1888,6 +1888,10 @@
 (decl pure partial i8_to_imm5 (i8) Imm5)
 (extern constructor i8_to_imm5 i8_to_imm5)
 
+;; Helper to go directly from a `Value` to an `Imm5`.
+(decl imm5_from_value (Imm5) Value)
+(extractor (imm5_from_value n) (i64_from_iconst (imm5_from_i64 n)))
+
 ;; Constructor that matches a `Value` equivalent to a replicated Imm5 on all lanes.
 (decl pure partial replicated_imm5 (Value) Imm5)
 (rule (replicated_imm5 (splat (i64_from_iconst (imm5_from_i64 n)))) n)
--- a/cranelift/codegen/src/isa/riscv64/inst.isle
+++ b/cranelift/codegen/src/isa/riscv64/inst.isle
@@ -1894,7 +1894,7 @@
 
 ;; Constructor that matches a `Value` equivalent to a replicated Imm5 on all lanes.
 (decl pure partial replicated_imm5 (Value) Imm5)
-(rule (replicated_imm5 (splat (i64_from_iconst (imm5_from_i64 n)))) n)
+(rule (replicated_imm5 (splat (imm5_from_value n))) n)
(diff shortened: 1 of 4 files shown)


cranelift/codegen/src/isa/riscv64/inst.isle
```diff
@@ -1888,6 +1888,10 @@
 (decl pure partial i8_to_imm5 (i8) Imm5)
 (extern constructor i8_to_imm5 i8_to_imm5)
 
+;; Helper to go directly from a `Value` to an `Imm5`.
+(decl imm5_from_value (Imm5) Value)
+(extractor (imm5_from_value n) (i64_from_iconst (imm5_from_i64 n)))
+
 ;; Constructor that matches a `Value` equivalent to a replicated Imm5 on all lanes.
 (decl pure partial replicated_imm5 (Value) Imm5)
 (rule (replicated_imm5 (splat (i64_from_iconst (imm5_from_i64 n)))) n)
```
afonso360, line 1897:
```suggestion
(rule (replicated_imm5 (splat (imm5_from_value n))) n)
```

Now that we've extracted `imm5_from_value` into a separate rule, we should also use it here, so that we don't have duplicate definitions.

afonso360 (commented):
This looks great to me 🎉 . Thanks for working on this!

Could you also add a run test over at `cranelift/filetests/filetests/runtests/simd-scalartovector.clif` so that this also gets tested that way in CI?

afonso360 (approved):
Looks great, Thanks!

PR #7288, opened 2023-10-18
wasi-http: Allow embedder to manage outgoing connections

### LINKED ISSUE #7259: wasi-http: Manage outgoing connections with a callback on WasiHttpCtx
Currently all outgoing connections assume http 1, and a specific tls library. We need to refactor this so that the logic for constructing the `hyper::SendRequest<_>` and spawning the connection handling task are done by a callback defined on the `WasiHttpCtx` instead.

dicej:
This adds a new `send_request` method to `WasiHttpView`, allowing embedders to override the default implementation with their own if they desire.  The default implementation behaves exactly as before.

I've also added a few new `wasi-http` tests: one to test the above, and two others to test streaming and concurrency.  These tests are ports of the `test_wasi_http_echo` and `test_wasi_http_hash_all` tests in the [Spin](https://github.com/fermyon/spin) integration test suite.  The component they instantiate is likewise ported from the Spin
`wasi-http-rust-streaming-outgoing-body` component.

Fixes #7259

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2668,8 +2668,12 @@ name = "test-programs"
 version = "0.0.0"
 dependencies = [
  "anyhow",
+ "base64",
+ "futures",
  "getrandom",
  "libc",
+ "sha2",
+ "url",
  "wasi",
  "wit-bindgen",
 ]
@@ -3782,6 +3786,7 @@ version = "15.0.0"
 dependencies = [
  "anyhow",
  "async-trait",
+ "base64",
  "bytes",
  "futures",
  "http",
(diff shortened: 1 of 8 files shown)


dicej:
I was wondering if `cargo-vet` would complain... and it does.  I thought maybe adding dependencies to test-only code would be okay, but clearly not.  I'll see what I can do to address it.

elliottt (approved):
This looks great to me, thank you!

pchickey:
Yes, thank you Joel, I really like how the design of this worked out.

PR #7289, opened 2023-10-18
Implement support for `thread` in `*.wast` tests

alexcrichton:
This commit implements support for `thread` and `wait` in `*.wast` files and imports the upstream spec test suite from the `threads` proposal. This additionally and hopefully makes it a bit easier to write threaded tests in the future if necessary too.


alexcrichton:
I'll note that the imported tests are not literally as-is upstream and are technically pending https://github.com/WebAssembly/threads/pull/217, so we can see how that turns out too.

alexcrichton:
Also cc @abrown as you might be interested in this

abrown (approved):
Makes sense to me! So I guess we will get rid of all of these imported `*.wast` files once the upstream testsuite contains them?

PR #7290, opened 2023-10-19
Add support for binary float operators to Winch

jeffcharles:
Related to #6528. Adds support for `f32.add`, `f64.add`, `f32.sub`, `f64.sub`, `f32.mul`, `f64.mul`, `f32.div`, `f64.div`, `f32.min`, `f64.min`, `f32.max`, `f64.max`, `f32.copysign`, and `f64.copysign` instructions to Winch.


saulecabrera (approved):
Looks good to me!

PR #6899, opened 2023-08-24
Add timezone host functions.

cdmurph32:
Issue:
https://github.com/bytecodealliance/wasmtime/issues/4274

WASI Clocks proposal:
https://github.com/WebAssembly/wasi-clocks/pull/48

Relate cap-std PR:
https://github.com/bytecodealliance/cap-std/pull/327

PR #7292, opened 2023-10-19
Describe workflow for testing preview2 components

cdmurph32:
This is information that would have been very useful when I was implementing timezones. https://github.com/bytecodealliance/wasmtime/pull/6899


fitzgen (commented):
Thanks!

Can you add newlines between paragraphs and wrap to 80 columns as the rest of the document does?

fitzgen (approved):
Thanks!

PR #7295, opened 2023-10-19
[14.0.0] Add a whole lot of semicolons to WIT files

alexcrichton:
This is a backport of #7159 to ensure that those who want to require semicolons in CI for example work correctly with Wasmtime 14.


alexcrichton:
I'll note that there's one exception to this which is wasi-nn. That uses a submodule and the updated version of the submodule slightly updates the APIs used, so I've avoided updating that. I think the important part though is updating the main body of WITs used by WASI.

pchickey (approved):

PR #7301, opened 2023-10-19
Swap the order of directories in `--dir`

alexcrichton:
This commit changes the `--dir` argument on the `wasmtime` CLI to be `HOST::GUEST` rather than `GUEST::HOST`. This matches Docker for example and is a little more consistent with only `--dir path` where the first argument is always treated as a host directory.

In terms of breaking-ness the movement from `--mapdir` to `--dir` hasn't been released with Wasmtime 14 yet so my hope is that this can land on both `main` and Wasmtime 14.0.0 before it's released to avoid any breakage other than existing scripts migrating from `--mapdir` to `--dir`.

### DIFF
--- a/ci/run-wasi-nn-example.sh
+++ b/ci/run-wasi-nn-example.sh
@@ -39,7 +39,7 @@ pushd $WASMTIME_DIR/crates/wasi-nn/examples/classification-example
 cargo build --release --target=wasm32-wasi
 cp target/wasm32-wasi/release/wasi-nn-example.wasm $TMP_DIR
 popd
-cargo run -- run --dir fixture::$TMP_DIR -S nn $TMP_DIR/wasi-nn-example.wasm
+cargo run -- run --dir $TMP_DIR::fixture -S nn $TMP_DIR/wasi-nn-example.wasm
 
 # Build and run another example, this time using Wasmtime's graph flag to
 # preload the model.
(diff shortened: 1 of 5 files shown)


pchickey (approved):

sunfishcode:
Matching docker makes sense to me too. As a historical note, the original syntax here wasn't something anyone thought about deeply, it was just something simple to get basic use cases working, that ended up sticking around.

jeffparsons:
Is there a way this arg could be extended in future (in a backward-compatible way) to support an optional verbose form? E.g. `--dir host=a;guest=b`, or even `--dir host=a guest=b`, where you can put host and guest in whatever order you want?

I'm trying to think through how the delimiters/escaping would work, and I realised I don't understand how it works today — e.g. what happens if your path contains `::`. Is the answer "don't have silly paths"? 😅

I ask this because these kinds of order-dependent syntaxes (Docker volume mounts, SSH port forwarding, etc.) have caused me seemingly endless confusion and suffering, and I'm not sure if I'm uniquely vulnerable to them or if this is something that lots of other people would like to be able to avoid as well.

alexcrichton:
I personally agree with @jeffparsons, although it's a bit too late for Wasmtime 14/15 to change the syntax here that much. This at least makes things "no worse" on the understandability front hopefully.

I've opened https://github.com/bytecodealliance/wasmtime/issues/7309 to continue discussion here now that this is merged, since I think there are still improvements to be had!

PR #7302, opened 2023-10-19
Update WASI versions to `0.2.0-rc-2023-11-05`

alexcrichton:
This commit updates the version numbers on `main` to no longer clash with the 14.0.0 release after #7299. The version number is chosen as the branch point for the 15.0.0 release of Wasmtime, at which point we'll update the versions again.


pchickey (approved):
This seems like a good choice for now. We will shoot to get some last wasi-http changes in here and upstream in the wasi proposal repos, and tag a RC in those proposal repos sometime between Nov 1 and 5, and then we can update these numbers (& wits, where comment changes happened) to be perfectly in sync.

PR #7304, opened 2023-10-19
A couple of small refactorings to the egraph elaboration pass

elliottt:
@jameysharp and I noticed a couple of refactoring opportunities while reading through the elaboration pass:

* The elaboration loop doesn't need to match on the top of the stack as a reference, because each case pops it immediately. Instead we can pop and match on the popped value.
* Computing the cost of a pure `ValueDef::Result` was using the block that contains the instruction to determine the level, but since pure values will not be in the DFG yet, this would default to `LoopLevel::root()` unconditionally. This also meant that the `Cost::at_level` function turned into an identity function on the cost given, making it unnecessary.


cfallin (approved):
LGTM! Good find w.r.t. the loop-level cost factor being a no-op; I suspect this was left over from an earlier version of extraction. I'd be interested to hear any thoughts you or others have about using the loop level somehow, but I'm happy to see this merged for now!

PR #7305, opened 2023-10-19
Gate debug builtins behind a Cargo feature

alexcrichton:
This doesn't shave off much but can help tidy up some symbols from exported objects if they aren't needed.


elliottt (approved):
Looks good to me!

PR #7315, opened 2023-10-20
Build "min" artifacts on CI

alexcrichton:
This commit updates the binary artifacts produced by CI to include "min"
builds where all default features are disabled. Additionally all the
stops are pulled in terms of build flags, nightly versions, etc, to get
a build that is as small as possible without actual source code changes.
This effectively codifies the instructions in https://github.com/bytecodealliance/wasmtime/pull/7282 into an easily
downloadable artifact.

No new tarballs are created for github releases but instead tarballs
that previously had a `wasmtime` executable for example now have a
`wasmtime-min` executable. Furthermore the C API which previously had
`libwasmtime.so` for example now has `libwasmtime-min.so`. The intention
is that the minimum-size artifacts are handy for determining a rough
size of Wasmtime but they're not so important that it seems worthwhile
to dedicate entire release entries for.

CI is refactored to support these minimum builds with separate builders.
This means that a single tarball produced as a final result is actually
two separate tarballs merged together, one for the normal build we do
today plus a new "min" tarball produced on the new "min" builders.
Various scripts and CI organization has been adjusted accordingly.

While here I went ahead and enabled `panic=abort` and debuginfo
stripping in our current release artifacts. While this doesn't affect a
whole lot it's less to upload to GitHub Actions all the time.

### DIFF
--- a/.github/actions/binary-compatible-builds/main.js
+++ b/.github/actions/binary-compatible-builds/main.js
@@ -34,7 +34,7 @@ if (process.env.CENTOS !== undefined) {
   return;
 }
 
-const name = process.env.INPUT_NAME;
+const name = process.env.INPUT_NAME.replace(/-min$/, '');
 
 child_process.execFileSync('docker', [
   'build',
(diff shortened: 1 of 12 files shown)


alexcrichton:
Two notes on this:

* I've prepared an [example release](https://github.com/alexcrichton/wasmtime/releases/tag/dev) in my own repository which has sample artifacts produced through CI.
* This PR required the first commit which is a riscv64-specific fix to handle what would otherwise result in a relocation error.

pchickey (approved):

PR #7330, opened 2023-10-23
Allow `LinearMemory`ies to control initialization

ureeves:
This allows for `LinearMemory` implementations to inform the runtime of their initialization status.


ureeves:
This wasn't meant to target upstream, apologies.

PR #7335, opened 2023-10-23
Use wasi-http::Error in the wasi-http implementation

rylev:
Currently any error in `WasiHttpView::send_request` will result in a trap instead of an error being returned back from the call to `outgoing_handler::handle`. This moves a large amount of the `Result` types in the implementation from `anyhow::Error` to `types::Error`. `HostFutureIncomingResponse` now contains a `anyhow::Result<Result<IncomingResponseInternal, types::Error>>` to keep track of both trappable errors and errors that will be returned to the caller.


pchickey (approved):

pchickey (approved):

pchickey (approved):

pchickey:
Just needs a conflict resolved, then I'll stick it in the queue

Issue #5628, opened 2023-01-24
bug: running with `--vtune` causes a crash


abrown:
### Test Case

```
(module)
```

### Steps to Reproduce

```
wasmtime --vtune /tmp/empty.wat
```

### Expected Results

Adding the `--vtune` flag, which enables VTune profiling, should succeed regardless of the environment.

### Actual Results

A snippet of the backtrace:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: error when notifying event', crates/jit/src/profiling/vtune.rs:60:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: PoisonError { .. }', crates/jit/src/profiling/vtune.rs:47:27
...
 17:     0x564e6a13571c - <wasmtime_jit::profiling::vtune::VTuneAgent as core::ops::drop::Drop>::drop::ha421249d00e9da25
                               at /.../wasmtime/src/crates/jit/src/profiling/vtune.rs:47:9
  18:     0x564e6935d677 - core::ptr::drop_in_place<wasmtime_jit::profiling::vtune::VTuneAgent>::h4021027fefeb76a5
                               at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/core/src/ptr/mod.rs:490:1
  19:     0x564e6935f22d - core::ptr::drop_in_place<alloc::boxed::Box<dyn wasmtime_jit::profiling::ProfilingAgent>>::h3e6586f99c95bc40
                               at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/core/src/ptr/mod.rs:490:1
  20:     0x564e6935c24c - core::ptr::drop_in_place<wasmtime::engine::EngineInner>::hb7a0d51da9c6653e
                               at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/core/src/ptr/mod.rs:490:1
  21:     0x564e693b2b64 - alloc::sync::Arc<T>::drop_slow::h6c0d38bb7bee604c
                               at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/alloc/src/sync.rs:1114:18
  22:     0x564e6935f9fb - <alloc::sync::Arc<T> as core::ops::drop::Drop>::drop::haeb7e9ee9f329b2a
[...]

PR #7337, opened 2023-10-23
update ittapi to 0.4.0

rahulchaphalkar:
Fixes the JIT Code visibility issue for VTune. Tested wasmtime cli with a simple Fibonacci fib.wasm file, and can see the following with vtune hotspots -

pr_upload

Previously this was [Outside Any Module] along with trace file corruption log in vtune.

This change enables vtune profiler to be used for wasm code hotspots, microarchitecture analysis, disassembly viewing etc.

Fixes issues like this https://github.com/bytecodealliance/wasmtime/issues/5628

Ittapi change which fixes the issue is here https://github.com/intel/ittapi/pull/105

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1638,9 +1638,9 @@ checksum = "1aab8fc367588b89dcee83ab0fd66b72b50b72fa1904d7095045ace2b0c81c35"
 
 [[package]]
 name = "ittapi"
-version = "0.3.4"
+version = "0.4.0"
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks! Looks good to me, but I'd like to double-check with @abrown as well

abrown (approved):
Yup, @rahulchaphalkar and I worked together to get this fixed.

PR #6765, opened 2023-07-25
Make C-API usable from Rust

RubixDev:
This PR gives the `c-api` library a distinct name and adds `rlib` to the `crate-type` list, so that it can also be used from Rust. This is useful in situations like https://github.com/tree-sitter/tree-sitter/pull/1864, where a project (here tree-sitter) is mainly written in C and uses the wasmtime C library, but also provides Rust bindings. The easiest way to vendor the wasmtime C library for the tree-sitter C library during compilation of the tree-sitter Rust crate is then to simply depend on the `wasmtime-c-api` crate in the Rust bindings.

PR #7341, opened 2023-10-23
Restructure building the C API

alexcrichton:
This commit introduces a wrapper crate which is now the new "real" C API. The purpose of this change is to enable using LTO when building the C API. Currently LTO is disabled because one of the crate types of the C API is an "rlib" which means that it can't have LTO performed due to rustc limitations. The solution here is to remove the "cdylib" and "staticlib" crate types from the "wasmtime-c-api" crate and introduce a new crate 'wasmtime-c-api-artfact' which wraps the previous crate and reexports it.

This way LTO can be enabled when just building the artifacts and the use case from #6765 is still satisfied by having a crate that can be linked to from Rust. Locally this reduces the size of the C API artifact for me by nearly 1M.


pchickey (approved):
Looks good to me, my only question is the naming - would it make more sense to rename the current `wasmtime-c-api` crate something like `wasmtime-c-api-impl` so that users of `wasmtime-c-api`, which becomes this new trivial facade, don't need to make any changes in order to benefit from this?

alexcrichton:
Good idea! I've switched to that naming scheme instead

Issue #5869, opened 2023-02-23
Add special cases in the x64 `emit_cmp` helper


elliottt:
#### Feature

The x64 backend has a helper named `emit_cmp` in  `inst.isle` that is used when emitting comparisons. Currently it will always use `x64_cmp` to emit the comparison, but for cases of equality comparisons with `0`, `x64_test` would be a better choice. Additionally, the `select` lowering does not use `emit_cmp` right now, and could be refactored to do so with this change in place.

#### Benefit

This will generate better code for cases where `emit_cmp` is used to compile a use of `icmp` in the x64 backend.

jameysharp:
The first part of this issue (using `test` for comparisons with 0) was finished in #6086, but there's still work to do with simplifying the lowering for Cranelift's `select` instruction on x86.

Solo-steven:
Hi @jameysharp  I am interested in this issue, I am a beginner to both compiler backend and isle, so I would like to discuss my solution there before I open PR ~

I think the second part of the issue can be solved by adding a helper for lowering select inst. 
```
(decl lower_select_icmp (Type IcmpCondResult Value Value) InstOutput)
(rule (lower_select_icmp ty (IcmpCondResult.Condition flags cc) x y)
      (with_flags flags (cmove_from_values ty cc x y)))
```

I think we can change the existing rule to use the helper above, form existed code 

https://github.com/bytecodealliance/wasmtime/blob/ed68661976554b8ea967d7d3ca2f2e0741323aad/cranelift/codegen/src/isa/x64/lower.isle#L2084

to use helper above :
```
(rule (lower (has_type ty (select (maybe_uextend (icmp cc a @ (value_type (fits_in_64 a_ty)) b)) x y)))
      (lower_select_icmp ty (emit_cmp cc a b ) x y))
```

cfallin:
@Solo-steven your refactor looks reasonable -- we'd be happy to review a PR with this. Thanks!

sudoHackIn:
Hi, @cfallin is this issue still actual or should be closed after Solo-steven's pr is merged?

cfallin:
@sudoHackIn the PR above addresed the second bit (`select` using the helper) but not the first (using `test` rather than `cmp`), so this issue should remain open, I think.

erikrose:
Would anyone mind if I have a try at the `select` half of this?

cfallin:
[...]

PR #7344, opened 2023-10-24
refactor: using emit_cmp helper to refactor lowering of selection ins…

Solo-steven:
This PR implements the second part of issue #5869 .

As lowering the `select` instruction based on  [`fcmp`](https://github.com/bytecodealliance/wasmtime/blob/9d8ca828d1888013b45570a94267778962846ad6/cranelift/codegen/src/isa/x64/lower.isle#L2073). I declare a helper function for lowering `icmp`-based select instruction.

### DIFF
--- a/cranelift/codegen/src/isa/x64/lower.isle
+++ b/cranelift/codegen/src/isa/x64/lower.isle
@@ -2082,7 +2082,7 @@
 ;; than one instruction for certain types (e.g., XMM-held, I128).
 
 (rule (lower (has_type ty (select (maybe_uextend (icmp cc a @ (value_type (fits_in_64 a_ty)) b)) x y)))
-      (lower_select_icmp ty (emit_cmp cc a b ) x y))
+      (lower_select_icmp ty (emit_cmp cc a b) x y))
 
 ;; Finally, we lower `select` from a condition value `c`. These rules are meant
 ;; to be the final, default lowerings if no other patterns matched above.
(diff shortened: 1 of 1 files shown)


cranelift/codegen/src/isa/x64/lower.isle
```diff
@@ -2082,8 +2082,7 @@
 ;; than one instruction for certain types (e.g., XMM-held, I128).
 
 (rule (lower (has_type ty (select (maybe_uextend (icmp cc a @ (value_type (fits_in_64 a_ty)) b)) x y)))
-      (let ((size OperandSize (raw_operand_size_of_type a_ty)))
-           (with_flags (x64_cmp size b a) (cmove_from_values ty cc x y))))
+      (lower_select_icmp ty (emit_cmp cc a b ) x y))
```
cfallin, line 2085: formatting nit: remove the space between `b` and the right-paren?

cfallin (commented):
Thanks! One formatting nit below, otherwise this seems correct.

Do you expect codegen to improve for any combination of compare and select instructions? If so, could we add a test showing that?

Solo-steven:
Hi @cfallin thanks for your review, I think this change will behave the same as before, just a simple refactor to using `emit_cmp` to generate the same x64 code, so I do not expect any improvements.

Because I am new to both ISLE and clif IR, I might be missing some potential improvements we can make. if so please point it out to me , thanks ~

PR #7345, opened 2023-10-24
wasmtime: Fix Bench API Build

afonso360:
👋 Hey,

It looks like the bench-api package is broken and no longer builds. I think we just need to add the default features to get the necessary API's enabled.

I've also added a check in CI that runs `cargo check` on this package to ensure that this does not regress in the future.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -281,6 +281,9 @@ jobs:
     # Check that benchmarks of the cranelift project build
     - run: cargo check --benches -p cranelift-codegen
 
+    # Check that the bench-api compiles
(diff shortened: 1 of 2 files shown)


afonso360:
I've also had to enable the `cranelift` feature for `wasmtime-cli-flags` otherwise it would just fail to run the benchmarks. Now I'm even less confident that I have the correct set of flags.

alexcrichton (approved):

alexcrichton:
This should be sufficient for now and for CLI flags we can add to that as necessary over time, but this is at least a reasonable starting point!

PR #7346, opened 2023-10-24
Add release notes for 14.0.1

### LINKED ISSUE #7342: No release notes for 14.0.1
https://github.com/bytecodealliance/wasmtime/blob/main/RELEASES.md only refers to 14.0.0 and https://github.com/bytecodealliance/wasmtime/releases/tag/v14.0.1 has the typical boilerplate that just says who triggered the release.

alexcrichton:
Closes #7342

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -17,7 +17,8 @@ Released 2023-10-23
 ### Fixed
 
 * Cranelift: preserve uext and sext flags for parameters on x86\_64 and apple
-  aarch64.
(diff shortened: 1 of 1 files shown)


pchickey (approved):
If we believe this bug is not reachable by wasmtime, I believe we should say so in this changelog entry. I don't have the expertise in this area to know one way or the other, though.

alexcrichton:
Good point! I've added some clarificatoin

PR #7358, opened 2023-10-25
monotonic clock: introduce `duration` type, split `subscribe`

pchickey:
Implement the interface change submitted upstream as https://github.com/WebAssembly/wasi-clocks/pull/53

> We are introducing a `duration` type because it has a distinct meaning
from `instant`: an `instant` can only be compared to other `instant`s
from the exact same `monotonic-clock`, whereas a `duration` represents
a duration of time which can be compared to any other duration of time.

> The `duration` type is motivated, in part, by a desire to reuse it to
specify durations such as timeouts in other WASI proposals.

> Instead of taking a boolean specifying whether the u64 is an absolute or
relative time, `subscribe-instant` takes an `instant` type and
`subscribe-duration` takes a `duration` type.


crates/wasi/src/preview2/host/clocks.rs
```diff
@@ -52,30 +52,57 @@ impl<T: WasiView> monotonic_clock::Host for T {
...
-        let deadline = tokio::time::Instant::now()
-            .checked_add(duration)
-            .ok_or_else(|| anyhow::anyhow!("time overflow: duration {duration:?}"))?;
         // NB: this resource created here is not actually exposed to wasm, it's
         // only an internal implementation detail used to match the signature
         // expected by `subscribe`.
-        let sleep = self.table_mut().push(Sleep(deadline))?;
+        let sleep = self.table_mut().push(Deadline::Instant(deadline))?;
+        subscribe(self.table_mut(), sleep)
+    }
```
alexcrichton, line 75: Could this method perhaps delegate to `subscribe_duration` or a similar helper once it calculates the relative desired `Duration`? Or alternatively would it be possible to directly convert between `when` and a `tokio::time::Instant`?

It basically felt a bit odf that the `Never` case was handled below but not here

alexcrichton (approved):

crates/wasi/src/preview2/host/clocks.rs
pchickey: Thanks, I thought that over again and factored it out into a helper function.

alexcrichton (approved):

PR #6807, opened 2023-08-04
Configure Mach ports vs signals via `Config`

alexcrichton:
This commit adds a `Config::macos_use_mach_ports` configuration option to replace the old `posix-signals-on-macos` compile-time Cargo feature. This'll make Wasmtime a tad larger on macOS but likely negligibly so. Otherwise this is intended to provide a resolution to #6785 where embedders will be able to use any build of Wasmtime and configure at runtime how trap handling should happen.

Functionally this commit additionally registers a `pthread_atfork` handler to cause any usage of Wasmtime in the child to panic. This should help head off a known-invalid state in case anyone runs into it in the future.

PR #7360, opened 2023-10-25
Remove the leftover usage of `posix-signals-on-macos`

ulan:
This follows up #6807 and removes the last remaining reference to the removed `posix-signals-on-macos` feature flag.

Note that `lib.rs` now imports `mod unix` on MacOS. This change is similar to the change in `traphandlers.rs` in #6807. It is needed for hosts that use signals instead of Mach ports on MacOs.

### DIFF
--- a/crates/wasmtime/src/lib.rs
+++ b/crates/wasmtime/src/lib.rs
@@ -476,9 +476,7 @@ pub use anyhow::{Error, Result};
 pub mod component;
 
 cfg_if::cfg_if! {
-    if #[cfg(all(target_os = "macos", not(feature = "posix-signals-on-macos")))] {
-        // no extensions for macOS at this time
-    } else if #[cfg(unix)] {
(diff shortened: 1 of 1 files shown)


alexcrichton (approved):
Thanks for this!

ulan:
@alexcrichton: fyi, I apologise for not spotting this earlier. We will try to follow and test the `main` branch more proactively from now on.

alexcrichton:
No worries! Would y'all like this backported to a Wasmtime release?

ulan:
> No worries! Would y'all like this backported to a Wasmtime release?

Thanks for the offer! We are updating to version 13. If it is not a lot of work to backport to 13 and 14, that would be great. 

Otherwise, we can also float the patch on our end until version 15.

ulan:
Great, I'll prepare the PRs to the release branches.

PR #7361, opened 2023-10-25
[release-13.0] Backport removal of `posix-signals-on-macos`

ulan:
Remove the leftover usage of `posix-signals-on-macos` (#7360)

This follows up #6807 and removes the last remaining reference to the removed `posix-signals-on-macos` feature flag.

Note that `lib.rs` now imports `mod unix` on MacOS. This change is similar to the change in `traphandlers.rs` in #6807. It is needed for hosts that use signals instead of Mach ports on MacOs.


alexcrichton (approved):
I pushed up a cherry-pick which I think will fix the CI failure from prior

PR #7363, opened 2023-10-25
mpk: protect memory with `PROT_NONE`

abrown:
This change fixes a bug with `ProtectionKey::protect`: previously it initialized each stripe with read and write permissions (i.e., `pkey_mprotect(..., PROT_READ | PROT_WRITE)` under the mistaken assumption that these permissions were MPK-specific, "what MPK permissions will we be allowed to set in the PKRU for these regions in the future?". This assumption is incorrect: the regions were immediately made accessible for reading and writing. The fix is to initially protect the regions with `PROT_NONE` and allow Wasmtime's `memory.grow` implementation to mark pages with `mprotect(..., PROT_READ | PROT_WRITE)` as usual. Whether a store can access a slice is still determined by the CPU state set in `mpk::allow`.

### DIFF
--- a/crates/runtime/src/mpk/enabled.rs
+++ b/crates/runtime/src/mpk/enabled.rs
@@ -78,7 +78,7 @@ impl ProtectionKey {
     pub fn protect(&self, region: &mut [u8]) -> Result<()> {
         let addr = region.as_mut_ptr() as usize;
         let len = region.len();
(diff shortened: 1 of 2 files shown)


abrown:
Yes, I'm working on a test to push soon. The spec tests caught this accidental read/write once I was able to actually run them.

PR #7364, opened 2023-10-25
mpk: limit the number of protection keys

abrown:
This change stems from how slicing memory slots into MPK-protected regions limits the number of memories each store can access: e.g., with fifteen keys in use, a store only has access to a fifteenth of the available slots. If we simply multiple the number of memory slots needed to run the `*.wast` spec tests by fifteen, we run out of available memory. This limits the number of protection keys used to two, which still allows us to test the functionality without reserving too much memory.

Also, if Wasmtime is ever embedded in an application that also uses memory protection keys, it could be useful to limit how many Wasmtime allocates and uses.. This change not only limits the number of protection keys used at runtime, but takes that
further to attempt to limit the initial number of keys allocated. The unfortunate side effect of using a `OnceLock` is that the `max` setting is only applicable on the first invocation, the one that sets the `OnceLock`.

### DIFF
--- a/crates/runtime/src/instance/allocator/pooling/memory_pool.rs
+++ b/crates/runtime/src/instance/allocator/pooling/memory_pool.rs
@@ -156,20 +156,21 @@ impl MemoryPool {
         let pkeys = match config.memory_protection_keys {
             MpkEnabled::Auto => {
                 if mpk::is_supported() {
-                    mpk::keys()
+                    mpk::keys(config.max_memory_protection_keys)
                 } else {
                     &[]
                 }
             }
             MpkEnabled::Enable => {
                 if mpk::is_supported() {
-                    mpk::keys()
+                    mpk::keys(config.max_memory_protection_keys)
                 } else {
                     bail!("mpk is disabled on this system")
                 }
             }
             MpkEnabled::Disable => &[],
         };
+        let pkeys = &pkeys[..pkeys.len().min(config.max_memory_protection_keys)];
 
         // This is a tricky bit of global state: when creating a memory pool
         // that uses memory protection keys, we ensure here that any host code
--- a/crates/runtime/src/instance/allocator/pooling/memory_pool.rs
+++ b/crates/runtime/src/instance/allocator/pooling/memory_pool.rs
@@ -170,7 +170,6 @@ impl MemoryPool {
             }
             MpkEnabled::Disable => &[],
         };
-        let pkeys = &pkeys[..pkeys.len().min(config.max_memory_protection_keys)];
 
         // This is a tricky bit of global state: when creating a memory pool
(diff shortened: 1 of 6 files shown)


tests/all/wast.rs
```diff
@@ -180,7 +182,7 @@ fn feature_found_src(bytes: &[u8], name: &str) -> bool {
 // specified maximum we can put a cap on the virtual address space reservations
 // made.
 fn lock_pooling() -> impl Drop {
-    const MAX_CONCURRENT_POOLING: u32 = 8;
+    const MAX_CONCURRENT_POOLING: u32 = 4;
```
abrown, line 185: Not sure about this?

crates/runtime/src/mpk/enabled.rs
```diff
@@ -9,29 +9,34 @@ pub fn is_supported() -> bool {
...
+/// when multiple threads try to allocate keys at the same time (e.g., during
+/// testing). It also ensures that a single copy of the keys is reserved for the
+/// lifetime of the process. Because of this, `max` is only a hint: it only
+/// is effective on the first invocation of this function.
 ///
 /// TODO: this is not the best-possible design. This creates global state that
 /// would prevent any other code in the process from using protection keys; the
 /// `KEYS` are never deallocated from the system with `pkey_dealloc`.
-pub fn keys() -> &'static [ProtectionKey] {
+pub fn keys(max: usize) -> &'static [ProtectionKey] {
```
abrown, line 24: @alexcrichton, I'm not sure I like how the `max` parameter is really only a hint: it only applies the first time one calls `keys`.

crates/runtime/src/instance/allocator/pooling/memory_pool.rs
```diff
@@ -156,20 +156,21 @@ impl MemoryPool {
...
                 if mpk::is_supported() {
-                    mpk::keys()
+                    mpk::keys(config.max_memory_protection_keys)
                 } else {
                     bail!("mpk is disabled on this system")
                 }
             }
             MpkEnabled::Disable => &[],
         };
+        let pkeys = &pkeys[..pkeys.len().min(config.max_memory_protection_keys)];
```
abrown, line 173: I guess I should just move this logic into `keys` if we end up using this approach.

tests/all/wast.rs
alexcrichton: This makes some sense to me in that you doubled the number of memories which would halve the limit of concurrent tests because right now each test creates its own engine.

alexcrichton (approved):
Seems reasonable to me! I think this is fine to start out with at least and we can iterate on it over time. I do agree with what you mentioned though of pushing the handling of `max` directly into the `keys` function.

PR #7365, opened 2023-10-25
Enable the new preview1 implementation by default

alexcrichton:
Currently Wasmtime has two implementations of the `wasi_snapshot_preview1` set of APIs. The now-historical implementation lives in the `wasi-common` crate and the more recent implementation lives in the `wasmtime-wasi` crate. The main difference is that the `wasmtime-wasi` implementation is based on the implementation of preview2, meaning that the preview1 implementation is a shim in that direction. Additionally currently the preview2 implementation of preview1 is accessible via the `-Spreview2` flag on the CLI.

This commit updates the interpretation of the `-Spreview2` flag and the defaults around which implementation to choose. By default the preview1-built-on-preview2 implementation (the new `wasmtime-wasi` implementation) is selected now. This means that the `wasi-common` implementation is disabled by default. There are still two use cases to keep running the `wasi-common` implementation, however:

* When running modules that import from `wasi_unstable`, the "snapshot" before `wasi_snapshot_preview1`, currently `wasi-common` is required. The shims to implement `wasi_unstable` have not yet been implemented in the `wasmtime-wasi` crate.

* When running with WASI threads (`-Sthreads`) the preview2 implementation does not work. This is because the preview2 implementation expects mutable access to the table which is not granted when threads are enabled.

Tests using `wasi_unstable` now pass `-Spreview2=n` to explicitly request the old `wasi-common` implementation. Additionally the `wasi-common` implementation is still selected by default when `-Sthreads` is passed (enabling the WASI threads proposal).

### DIFF
--- a/crates/wasi/src/preview2/preview1.rs
+++ b/crates/wasi/src/preview2/preview1.rs
@@ -9,7 +9,7 @@ use crate::preview2::bindings::{
     io::{poll, streams},
 };
 use crate::preview2::{FsError, IsATTY, StreamError, StreamResult, TableError, WasiView};
-use anyhow::{anyhow, bail, Context};
+use anyhow::{bail, Context};
 use std::borrow::Borrow;
 use std::collections::{BTreeMap, HashSet};
 use std::mem::{self, size_of, size_of_val};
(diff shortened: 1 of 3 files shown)


alexcrichton:
I'll also note that I've included a few minor fixes to `wasmtime-wasi`'s implementation of preview1 to fix some failing tests when it was enabled.

pchickey (approved):
Thanks. This is an exciting move! If we survive a release of this, I'll start moving the old implementation to be exported by an off-by-default wasmtime_wasi::legacy mod, and move ::preview2 to the root.

PR #7366, opened 2023-10-25
Add logging and request id tracking to `wasmtime serve`

### LINKED ISSUE #7257: wasi-http: Add unique request ids in `wasmtime serve`
For logging purposes, we'd like to have unique request ids prefixed on each log line in the `wasmtime serve` command. This would include stdout/stderr from the incoming handler, to make it easier to debug behavior on individual requests.

elliottt:
Add some missing log handling to `wasmtime serve`:

* Log the listening address and port before starting to accept incoming connections
* Track individual requests with unique ids
* Prefix stdout and stderr from the guest with the request id

Fixes #7257

### DIFF
--- a/src/commands/serve.rs
+++ b/src/commands/serve.rs
@@ -1,5 +1,5 @@
 use crate::common::{Profile, RunCommon, RunTarget};
-use anyhow::{bail, Result};
+use anyhow::{anyhow, bail, Result};
 use clap::Parser;
 use std::{
     path::PathBuf,
@@ -11,7 +11,9 @@ use std::{
 };
 use wasmtime::component::{InstancePre, Linker};
 use wasmtime::{Engine, Store, StoreLimits};
-use wasmtime_wasi::preview2::{self, StreamResult, Table, WasiCtx, WasiCtxBuilder, WasiView};
+use wasmtime_wasi::preview2::{
+    self, StreamError, StreamResult, Table, WasiCtx, WasiCtxBuilder, WasiView,
+};
 use wasmtime_wasi_http::{body::HyperOutgoingBody, WasiHttpCtx, WasiHttpView};
 
 #[cfg(feature = "wasi-nn")]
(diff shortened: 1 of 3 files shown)


src/commands/serve.rs
```diff
@@ -357,3 +391,55 @@ impl hyper::service::Service<Request> for ProxyHandler {
...
+
+        Ok(())
+    }
+
+    fn flush(&mut self) -> StreamResult<()> {
+        Ok(())
+    }
+
+    fn check_write(&mut self) -> StreamResult<usize> {
+        Ok(1024 * 1024)
```
elliottt, line 438: This is arbitrary.

src/commands/serve.rs
```diff
@@ -357,3 +391,55 @@ impl hyper::service::Service<Request> for ProxyHandler {
...
+    }
+}
+
+#[async_trait::async_trait]
+impl preview2::Subscribe for LogStream {
+    async fn ready(&mut self) {}
```
elliottt, line 444: It's not clear that there's a reasonable implementation for this here. Perhaps using `tokio::io::Stdout` instead of a mutex guarded `std::io::Stdout` would make more sense?

src/commands/serve.rs
```diff
@@ -135,10 +135,20 @@ impl ServeCommand {
...
+        builder.envs(&[("REQUEST_ID", req_id.to_string())]);
+
+        builder.stdout(LogStream {
+            prefix: format!("stdout [{req_id}] :: "),
+            output: output.clone(),
+        });
+
+        builder.stderr(LogStream {
+            prefix: format!("stderr [{req_id}] :: "),
+            output,
```
elliottt, line 150: I think it would be reasonable to make a separate `Stderr` handle here. My thought was that since the output is already prefixed with `stdout`/`stderr`, using the same output stream would be okay.

src/commands/serve.rs
```diff
@@ -357,3 +391,55 @@ impl hyper::service::Service<Request> for ProxyHandler {
...
+                msg.push(b'\n');
+            }
+        }
+
+        let output = self.output.clone();
+        tokio::task::spawn(async move {
+            use std::io::Write;
+            let mut output = output.lock().await;
+            output.write_all(&msg).expect("writing to stdout");
+        });
```
alexcrichton, line 428: What would you think of going ahead and directly doing a blocking write to stdout/stderr here? That'd help applying a bit of backpressure (albeit not in a perfect way) if necessary and additionally would avoid the need to pass around streams much. That'd also perhaps make it a bit easier to write stdout to stdout and stderr to stderr.

alexcrichton (approved):
Looks good to me! One thought about whether or not this should be an async write, but otherwise seems fine to me

src/commands/serve.rs
elliottt: Great suggestion! I've implemented that, and substantially simplified the `LogStream::write` implementation :)

PR #7369, opened 2023-10-25
Allow disabling the host-arch feature of cranelift-codegen

bjorn3:
This is required to compile for a target which doesn't have a cranelift backend. Before this change using any of the cranelift crates that depend on cranelift-codegen would forcefully enable all default features and thus host-arch. With this change only the std and unwind features are still forcefully enabled as cranelift-codegen doesn't compile with either disabled.

Required for https://github.com/rust-lang/rust/pull/81746

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -166,7 +166,7 @@ wasmtime-wit-bindgen = { path = "crates/wit-bindgen", version = "=15.0.0" }
 test-programs-artifacts = { path = 'crates/test-programs/artifacts' }
 
 cranelift-wasm = { path = "cranelift/wasm", version = "0.102.0" }
-cranelift-codegen = { path = "cranelift/codegen", version = "0.102.0" }
+cranelift-codegen = { path = "cranelift/codegen", version = "0.102.0", default-features = false, features = ["std", "unwind"] }
 cranelift-frontend = { path = "cranelift/frontend", version = "0.102.0" }
 cranelift-entity = { path = "cranelift/entity", version = "0.102.0" }
 cranelift-native = { path = "cranelift/native", version = "0.102.0" }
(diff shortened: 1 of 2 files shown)


bjorn3:
Without the `unwind` feature enabled I get:

```
error[E0433]: failed to resolve: could not find `systemv` in `unwind`
    --> cranelift/codegen/src/machinst/vcode.rs:1181:41
     |
1181 |                     crate::isa::unwind::systemv::caller_sp_to_cfa_offset();
     |                                         ^^^^^^^ could not find `systemv` in `unwind`
```

bjorn3:
It would be great if this could get backported too. Without this rust's CI fails when trying to build cg_clif for distribution with rustup.

alexcrichton:
Thanks! I'm a little surprised that the dependency from `wasmtime-cranelift` on the `cranelift` crate, without `features = ['default']`, still works. I think it's because without any enabled ISAs it still builds the host ISA? Mind adding it there though to be safe?

Otherwise yeah looks like disabling the `unwind` feature and not being able to compile is indeed a Cranelift bug. If you'd like feel free to open an issue (or a subsequent PR) to fix that. Enabling it here by default is definitely ok.

Also a point release is fine by me, and there's one [slated tomorrow](https://github.com/bytecodealliance/wasmtime/pull/7368) for https://github.com/bytecodealliance/wasmtime/pull/7362 as well so good timing!

If you wouldn't mind preparing the PR for the backport that'd be appreciated!

alexcrichton (approved):

bjorn3:
> Mind adding it there though to be safe?

Done

> If you'd like feel free to open an issue (or a subsequent PR) to fix that.

Sure

> If you wouldn't mind preparing the PR for the backport that'd be appreciated!

Will do.

PR #7374, opened 2023-10-25
`cranelift-wasm`: Add a small optimization for dynamic memories with `min_size == max_size`

fitzgen:
In this case, we don't need to load the dynamic heap bound from the vmctx
because it actually has a constant size. Instead, we can use the constant
directly.

### DIFF
--- a/crates/cranelift/src/func_environ.rs
+++ b/crates/cranelift/src/func_environ.rs
@@ -1839,14 +1839,10 @@ impl<'module_environment> cranelift_wasm::FuncEnvironment for FuncEnvironment<'m
                 u64::MAX
             });
 
-        let max_size = self.module.memory_plans[index].memory.maximum.map(|max| {
-            max.checked_mul(u64::from(WASM_PAGE_SIZE))
-                .unwrap_or_else(|| {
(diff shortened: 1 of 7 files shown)


alexcrichton (approved):
Mind adding a test for a `(memory 1 2)` as well to ensure that a constant size isn't used?

crates/cranelift/src/func_environ.rs
```diff
@@ -1839,6 +1839,15 @@ impl<'module_environment> cranelift_wasm::FuncEnvironment for FuncEnvironment<'m
...
             });
 
+        let max_size = self.module.memory_plans[index].memory.maximum.map(|max| {
+            max.checked_mul(u64::from(WASM_PAGE_SIZE))
+                .unwrap_or_else(|| {
+                    // See comment in `min_size` computation above.
+                    debug_assert_eq!(max, 1 << 48);
+                    u64::MAX
+                })
+        });
```
fitzgen, line 1849: Might be better to `and_then` instead of `map` here actually and not swap out the max for `u64::MAX` on overflow?

fitzgen: Done.

PR #7378, opened 2023-10-26
dep: target-lexicon: 0.12.3 -> 0.12.12

nagisa:
Largely just additions of new targets and such, as well as support for serde.

I personally am motivated to make this change, as the downstream development we’re doing depends on the target added in 0.12.12.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2570,9 +2570,9 @@ dependencies = [
 
 [[package]]
 name = "target-lexicon"
-version = "0.12.3"
+version = "0.12.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


alexcrichton:
Thanks! For `cargo vet` I've send a PR with the audit for this at https://github.com/bytecodealliance/wasmtime/pull/7380 and once that's landed you can rebase this on that and we can merge

nagisa:
Hmph, `cargo vet` is still failing in the CI, even though locally it works perfectly fine now :confused:

alexcrichton:
Apologies that's my bad, I opened https://github.com/bytecodealliance/wasmtime/pull/7391 to re-add the vet after I accidentally deleted it.

Note that `cargo vet` locally is probably automatically re-adding the entry, I think you'll need `cargo vet --locked` to reproduce the failure on CI

alexcrichton (approved):

PR #7383, opened 2023-10-26
wasi-http: Fallible fields set and append

### LINKED ISSUE #7270: wasi-http: make header methods fallible

elliottt:
This PR makes the `fields.set` and `fields.append` methods fallible, and introduces a new error type for describing the ways in which those operations can fail. Additionally it adds a notion of forbidden headers, which will always include the following:

* Connection
* Keep-Alive
* Proxy-Authenticate
* Proxy-Authorization
* Proxy-Connection
* TE
* Transfer-Encoding
* Upgrade
* HTTP2-Settings

There's an embedder-defined hook for adding additional header name validation, but that underlying set of headers can't be altered.

Fixes #7270

### DIFF
--- a/crates/wasi-http/src/types.rs
+++ b/crates/wasi-http/src/types.rs
@@ -7,6 +7,7 @@ use crate::{
 };
 use anyhow::Context;
 use http_body_util::BodyExt;
+use hyper::header::HeaderName;
 use std::any::Any;
 use std::time::Duration;
 use tokio::net::TcpStream;
@@ -64,6 +65,14 @@ pub trait WasiHttpView: Send {
         Self: Sized,
     {
         default_send_request(self, request)
+    }
+
+    fn is_forbidden_request_header(&mut self, name: &HeaderName) -> bool {
+        default_is_forbidden_request_header(name)
+    }
+
+    fn is_forbidden_response_header(&mut self, name: &HeaderName) -> bool {
+        default_is_forbidden_response_header(name)
     }
 }
 
@@ -165,6 +174,22 @@ pub fn default_send_request(
     let fut = view.table().push(HostFutureIncomingResponse::new(handle))?;
 
     Ok(fut)
+}
+
+pub fn default_is_forbidden_request_header(name: &HeaderName) -> bool {
+    use hyper::header;
+
+    const FORBIDDEN: &'static [HeaderName] = &[header::TRANSFER_ENCODING];
+
+    FORBIDDEN.contains(name)
+}
+
+pub fn default_is_forbidden_response_header(name: &HeaderName) -> bool {
+    use hyper::header;
+
+    const FORBIDDEN: &'static [HeaderName] = &[header::TRANSFER_ENCODING];
+
+    FORBIDDEN.contains(name)
 }
 
 pub fn timeout_error(kind: &str) -> anyhow::Error {
--- a/crates/wasi-http/src/types.rs
+++ b/crates/wasi-http/src/types.rs
@@ -67,12 +67,8 @@ pub trait WasiHttpView: Send {
         default_send_request(self, request)
     }
 
-    fn is_forbidden_request_header(&mut self, name: &HeaderName) -> bool {
-        default_is_forbidden_request_header(name)
-    }
-
-    fn is_forbidden_response_header(&mut self, name: &HeaderName) -> bool {
-        default_is_forbidden_response_header(name)
+    fn is_forbidden_header(&mut self, _name: &HeaderName) -> bool {
+        false
     }
 }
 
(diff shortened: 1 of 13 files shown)


crates/wasi-http/src/types_impl.rs
```diff
@@ -258,7 +294,17 @@ impl<T: WasiHttpView> crate::bindings::http::types::HostOutgoingRequest for T {
...
     ) -> wasmtime::Result<Resource<HostOutgoingRequest>> {
-        let headers = get_fields_mut(self.table(), &headers)?.clone();
+        let (ty, fields) = get_fields_mut(self.table(), &headers)?;
+
+        if !matches!(ty, FieldsType::Request) {
+            return Err(Error::FieldsError(
+                "Response fields used when constructing a Request".to_owned(),
+            )
+            .into());
+        }
```
elliottt, line 304: I don't like this trap when the fields provided aren't meant for a request. This pushes me towards thinking that we shouldn't use a `constructor` here, but instead a static method for making a request that can fail.

crates/wasi-http/src/types.rs
```diff
@@ -167,6 +177,22 @@ pub fn default_send_request(
...
+}
+
+pub fn default_is_forbidden_response_header(name: &HeaderName) -> bool {
+    use hyper::header;
+
+    const FORBIDDEN: &'static [HeaderName] = &[header::TRANSFER_ENCODING];
+
+    FORBIDDEN.contains(name)
+}
+
```
elliottt, line 195: I'd appreciate some input on what a good default set of forbidden headers is for requests and responses.

crates/wasi-http/src/types.rs
```diff
@@ -65,6 +67,14 @@ pub trait WasiHttpView: Send {
...
         default_send_request(self, request)
     }
+
+    fn is_forbidden_request_header(&mut self, name: &HeaderName) -> bool {
+        default_is_forbidden_request_header(name)
+    }
+
+    fn is_forbidden_response_header(&mut self, name: &HeaderName) -> bool {
+        default_is_forbidden_response_header(name)
+    }
```
elliottt, line 77: Would it be useful to add the `HeaderMap` that's being added to as an argument here as well?

crates/wasi-http/wit/deps/http/types.wit
```diff
@@ -33,7 +33,8 @@ interface types {
     invalid-url(string),
     timeout-error(string),
     protocol-error(string),
-    unexpected-error(string)
+    unexpected-error(string),
+    header-name-error(string),
```
lukewagner, line 37: I think in the case of header errors, the set of failure reasons is disjoint from all the other places that produce `error`s, which I think makes it worth having a separate variant just for header errors such as:
```wit
variant header-error {
  invalid-syntax,
  forbidden
}
```
(I know we don't usually maximally-specialize error variants, but it seems like in cases where they're fully disjoint like this, it's worthwhile.)

crates/wasi-http/wit/deps/http/types.wit
```diff
@@ -175,11 +176,13 @@ interface types {
     ///
     /// This method consumes the `response-outparam` to ensure that it is
     /// called at most once. If it is never called, the implementation
-    /// will respond with an error.
+    /// will respond with an error. If there is an error sending the request,
+    /// the error result will contain both a description of the error and the
+    /// `response-outparam`, allowing the `set` operation to be retried.
```
lukewagner, line 181: If we reject invalid headers eagerly, then I think there shouldn't be any error possible here.  In particular, the `set` operation shouldn't be describing the actual sending of the response over the network (that kind of error would be reflecting in the streaming operations for writing to and finishing the response body); this operation should be the equivalent of `*outparam = result;`.  Or am I missing something else that can happen here?

crates/wasi-http/wit/deps/http/types.wit
lukewagner: Note also that if we expand the WASI HTTP spec to distinguish "definitely-forbidden" headers (listed by the WASI HTTP spec) and "host-forbidden" headers (that's purely up to the current running host), we'd probably split `forbidden` into two cases to reflect the difference.

elliottt: I like that specific error type more than extending the more general `error` variant, thanks! I'm not sure about differentiating between definitely forbidden and host forbidden though: what would a guest do differently if it knew the difference between definitely forbidden and host forbidden?

crates/wasi-http/wit/deps/http/types.wit
elliottt: Good point. Switching to having forbidden headers rejected at `fields.set` and `fields.append` removes the need to allow validation here.

crates/wasi-http/wit/deps/http/types.wit
```diff
@@ -33,7 +33,14 @@ interface types {
     invalid-url(string),
     timeout-error(string),
     protocol-error(string),
-    unexpected-error(string)
+    unexpected-error(string),
```
lukewagner, line 36:
```suggestion
    unexpected-error(string)
```

pchickey (approved):
Looks great overall - rather than trap in the fields constructor, lets make the constructor create an empty set of fields, and have a static function that takes the list<tuple<...>> repr and returns result<fields, header-error>

lukewagner:
@pchickey I think (but may be missing a case), that, with the eager-validation changes Trevor made most recently, there isn't a header-related error that can occur in the constructor?

Independently, the discussion made me think that perhaps (in a post-Preview-2 timeframe) we should relax the C-M `constructor` validation rules to allow constructor functions to return a `result<containing-resource-type, T>` (for any `T`).  In languages with EH, this would map to the usual throwing-constructor pattern.  Otherwise, bindings would produce their best approximation of a fallible constructor.

pchickey:
The function `set: func(name: field-key, value: list<field-value>) -> result<_, header-error>; ` allows rejecting input, but `constructor(entries: list<tuple<field-key,field-value>>); ` would not allow that same input to be rejected with an error, so currently the implementation traps.

Issue #7384, opened 2023-10-26
Tracking issue for migrating to the new CLI


alexcrichton:
> **Note**: if you'd rather not read up on this change and deal with it later then set `WASMTIME_NEW_CLI=0` to force usage of the old CLI or `WASMTIME_NEW_CLI=1` to force usage of the new CLI. This environment variable will go away in the future and the old CLI will be removed. This won't happen at least until Wasmtime 16 and at this time no exact time to make this change has been determined.

I'm opening this as a tracking issue to assist users in migrating from the old
Wasmtime 13-and-prior CLI to the new Wasmtime 14-and-later CLI. Wasmtime 14.0.x
is going to receive a patch update which restores support for the old CLI to
simultaneously support both the old and the new. The support, however, will
issue warnings that point to this issue. Specifically:

* If a CLI invocation parses different in Wasmtime 13 ("the old") and Wasmtime
  14 ("the new") then a warning is issued. For example:

  ```
  wasmtime run foo.wasm --invoke bar
  ```

  Was parsed in the old parser as invoking the `bar` function of `foo.wasm`. In
  the new parser this parses as passing the `--invoke` flag to the `foo.wasm`
  CLI program. In this situation a warning will be issued. This example could be
  fixed by changing the order of the arguments:

  ```
  wasmtime run --invoke bar foo.wasm
  ```

  which will no longer emit a warning.

* CLI invocations which fail to parse in the new parser but succeed in the old
  parser will continue to work but emit a warning. For example:

  ```
  wasmtime run --disable-cache foo.wasm
  ```

  no longer works because `--disable-cache` was renamed to `-C cache=n`. This
  will continue to work through but emit a warning.

This issue is being filed as I'm going to hardcode in the source code a print
of a warning which points here for users to find and discuss. My hope is to
incorporate feedback here into future iterations of the warning and/or the CLI.

# Changes in Wasmtime 14

In Wasmtime 14 the `wasmtime` executable CLI syntax changed. #6925 has a
summary of the changes as well as a detailed table to renamed flags, but a
summary is restated here as well.

## Parsing Changes

The first major change in Wasmtime 14 was a change to how arguments are parsed
to Wasmtime vs the WebAssembly program being run. Previously Wasmtime would
parse arguments anywhere before a `--` on the CLI as its own, and everything
else was passed to the wasm program. This is a future compatibility hazard
though because `--` is often omitted meaning that in the future it could be
[...]

Issue #7336, opened 2023-10-23
wasmtime 14 argument parsing change broke TinyGo and Go wasip1 support


dgryski:
Go uses `wasmtime` as the default when running wasip1 binaries: https://github.com/golang/go/blob/master/misc/wasm/go_wasip1_wasm_exec  . This script will break the default runner for users once they upgrade wasmtime.

Similarly, TinyGo constructs command lines to run wasm binaries.  The changes to the argument parsing breaks that:
https://github.com/tinygo-org/tinygo/issues/3970 .

alexcrichton:
Thanks for the report! If you're curious more discussion about this can be found on https://github.com/bytecodealliance/wasmtime/pull/6925 and related links.

Is it possible to have these scripts to get updated to require Wasmtime 14? Or is it desired that any Wasmtime version in the environment suffices? In the case that any Wasmtime version would suffice I think that will unfortunately require version detection to select the right CLI.

ydnar:
Regardless of the merits of the redesigned CLI, I think having a compatibility shim with deprecation warnings might have been warranted, for at least one major version.

`--max-wasm-stack: deprecated in wasmtime v14. Did you mean -W max-wasm-stack=N?`

alexcrichton:
I wanted to be sure to copy over some comments [from here](https://github.com/bytecodealliance/wasmtime/pull/6925#issuecomment-1775847262) to this dedicated issue as well as I suspect others will land here in the future too.

First off, if this is catching folks by surprise, I'm sorry about that! The intent was to gradually roll this out the best we could to prevent catching anyone by surprise. The places this was previously discussed prior to the Wasmtime 14 release were:

* [original issue](https://github.com/bytecodealliance/wasmtime/issues/6741)
* [Zulip discussion](https://bytecodealliance.zulipchat.com/#narrow/stream/217126-wasmtime/topic/Upcoming.20changes.20to.20Wasmtime's.20CLI/near/388395163)
* [Discussed in a bi-weekly meeting](https://github.com/bytecodealliance/meetings/blob/main/wasmtime/2023/wasmtime-08-31.md)

That being said if you're surprised by this change, that definitely wasn't sufficient communication! I would at least personally be interested in learning how to better communicate possible breaking changes like this to downstream consumers. Feel free to reach out to me (or other project members) either here, on email, or on Zulip, and we can continue discussion.

Now with that being there's two points that were discussed historically in the above discussions which I think are also important to reiterate here as well:

[...]

PR #7385, opened 2023-10-26
Add compatibility shims for Wasmtime 13 CLI

alexcrichton:
This commit introduces a compatibility shim for the Wasmtime 13 CLI and prior. The goal of this commit is to address concerns raised in #7336 and other locations as well. While the new CLI cannot be un-shipped at this point this PR attempts to ameliorate the situation somewhat through a few avenues:

* A complete copy of the old CLI parser is now included in `wasmtime` by default.
* The `WASMTIME_NEW_CLI=0` environment variable can force usage of the old CLI parser for the `run` and `compile` commands.
* The `WASMTIME_NEW_CLI=1` environment variable can force usage of the new CLI parser.
* Otherwise both the old and the new CLI parser are executed. Depending on the result one is selected to be executed, possibly with a warning printed.
* If both CLI parsers succeed but produce the same result, then no warning is emitted and execution continues as usual.
* If both CLI parsers succeed but produce different results then a warning is emitted indicating this. The warning points to #7384 which has further examples of how to squash the warning. The warning also mentions the above env var to turn the warning off. In this situation the old semantics are used at this time instead of the new semantics. It's intended that eventually this will change in the future.
* If the new CLI succeeds and the old CLI fails then the new semantics are executed without warning.
* If the old CLI succeeds and the new CLI fails then a warning is issued and the old semantics are used.
* If both the old and the new CLI fail to parse then the error message for the new CLI is emitted.

Note that this doesn't add up to a perfect transition. The hope is that most of the major cases of change at the very least have something printed. My plan is to land this on `main` and then backport it to the 14.0.0 branch as a 14.0.3 release.

### DIFF
--- a/src/bin/wasmtime.rs
+++ b/src/bin/wasmtime.rs
@@ -183,7 +183,7 @@ warning: this CLI invocation of Wasmtime will be parsed differently in future
          however this behavior can also be temporarily configured with an
          environment variable:
 
-         - WASMTIME_NEW_CLI=0 to indicate old semantics are desired and silences this warning, or
+         - WASMTIME_NEW_CLI=0 to indicate old semantics are desired and silence this warning, or
          - WASMTIME_NEW_CLI=1 to indicate new semantics are desired and use the latest behavior\
 "
                     );
@@ -236,7 +236,7 @@ warning: this CLI invocation of Wasmtime is going to break in the future -- for
          however this behavior can also be temporarily configured with an
          environment variable:
 
-         - WASMTIME_NEW_CLI=0 to indicate old semantics are desired and silences this warning, or
+         - WASMTIME_NEW_CLI=0 to indicate old semantics are desired and silence this warning, or
          - WASMTIME_NEW_CLI=1 to indicate new semantics are desired and see the error\
 "
                 );
--- a/src/bin/wasmtime.rs
+++ b/src/bin/wasmtime.rs
@@ -176,20 +176,34 @@ mod old_cli {
                     eprintln!(
                         "\
 warning: this CLI invocation of Wasmtime will be parsed differently in future
-         Wasmtime versions, see this online issue for more information:
+         Wasmtime versions, but its interpretation is not changing yet -- see
+         this online issue for more information:
          https://github.com/bytecodealliance/wasmtime/issues/7384
-         you can use `WASMTIME_NEW_CLI=0` to temporarily disable this warning\
+
+         CLI parsing can be temporarily configured with an environment variable:
+
+         - WASMTIME_NEW_CLI=0 to silence this warning and not change behavior, or
+         - WASMTIME_NEW_CLI=1 to force using the new behavior\
 "
                     );
                     old.execute()
                 } else {
+                    // this error message is not statically reachable due to
+                    // `DEFAULT_OLD_BEHAVIOR=true` at this time, but when that
+                    // changes this should be updated to have a more accurate
+                    // set of text.
+                    assert!(false);
                     eprintln!(
                         "\
 warning: this CLI invocation of Wasmtime is parsed differently now than it was
-         in Wasmtime 13 and prior, and the new semantics are going to be used
-         instead of the old semantics -- see this issue for more information:
+         prior, and the new semantics are going to be used instead of the old
+         semantics -- see this issue for more information:
          https://github.com/bytecodealliance/wasmtime/issues/7384
-         you can use `WASMTIME_NEW_CLI=1` to disable this warning\
+
+         CLI parsing can be temporarily configured with an environment variable:
+
+         - WASMTIME_NEW_CLI=0 to use the old CLI parsing behavior, or
+         - WASMTIME_NEW_CLI=1 to silence this warning and not change behavior\
 "
                     );
                     new.execute()
@@ -212,10 +226,14 @@ warning: this CLI invocation of Wasmtime is parsed differently now than it was
             (Err(_new), Ok(old)) => {
                 eprintln!(
                     "\
-warning: this CLI invocation of Wasmtime is going to break in the future, for
+warning: this CLI invocation of Wasmtime is going to break in the future -- for
          more information see this issue online:
          https://github.com/bytecodealliance/wasmtime/issues/7384
-         use `WASMTIME_NEW_CLI=0` to temporarily disable this warning\
+
+         CLI parsing can be temporarily configured with an environment variable:
+
+         - WASMTIME_NEW_CLI=0 to silence this warning and not change behavior, or
+         - WASMTIME_NEW_CLI=1 to see how this command breaks with new syntax\
 "
                 );
                 old.execute()
--- a/src/bin/wasmtime.rs
+++ b/src/bin/wasmtime.rs
@@ -176,14 +176,15 @@ mod old_cli {
                     eprintln!(
                         "\
 warning: this CLI invocation of Wasmtime will be parsed differently in future
-         Wasmtime versions, but its interpretation is not changing yet -- see
-         this online issue for more information:
+         Wasmtime versions -- see this online issue for more information:
          https://github.com/bytecodealliance/wasmtime/issues/7384
 
-         CLI parsing can be temporarily configured with an environment variable:
+         Wasmtime will now execute with the old (<= Wasmtime 13) CLI parsing,
+         however this behavior can also be temporarily configured with an
+         environment variable:
 
-         - WASMTIME_NEW_CLI=0 to silence this warning and not change behavior, or
-         - WASMTIME_NEW_CLI=1 to force using the new behavior\
+         - WASMTIME_NEW_CLI=0 to indicate old semantics are desired and silences this warning, or
+         - WASMTIME_NEW_CLI=1 to indicate new semantics are desired and use the latest behavior\
 "
                     );
                     old.execute()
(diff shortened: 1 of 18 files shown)


src/bin/wasmtime.rs
```diff
@@ -99,11 +99,176 @@ impl Wasmtime {
...
+            // dictate which wins and additionally print a warning message.
+            (Ok(new), Ok(old)) => {
+                if new == old {
+                    return new.execute();
+                }
+                if DEFAULT_OLD_BEHAVIOR {
+                    eprintln!(
+                        "\
+warning: this CLI invocation of Wasmtime will be parsed differently in future
+         Wasmtime versions, see this online issue for more information:
```
cfallin, line 179: s/,/--/ (comma splice; replace with hyphen or semicolon)

src/bin/wasmtime.rs
```diff
@@ -99,11 +99,176 @@ impl Wasmtime {
...
+                    );
+                    old.execute()
+                } else {
+                    eprintln!(
+                        "\
+warning: this CLI invocation of Wasmtime is parsed differently now than it was
+         in Wasmtime 13 and prior, and the new semantics are going to be used
+         instead of the old semantics -- see this issue for more information:
+         https://github.com/bytecodealliance/wasmtime/issues/7384
+         you can use `WASMTIME_NEW_CLI=1` to disable this warning\
```
cfallin, line 192: s/you/You/, period at end, and blank line to separate this note (and perhaps also a blank line before the URL)?

Also: we should note that both settings of the environment variable will work for now. I think a common use-case is that someone has usage of the old CLI, they upgrade, and they need to silence warnings without being forced to immediately do the conversion. We should note that they have that option too.

Perhaps something like:

```
You can set an environment variable to force either the old or new CLI option format
with no warnings emitted. Set

    - WASMTIME_NEW_CLI=0 to get behavior identical to Wasmtime 13 and earlier, or
    - WASMTIME_NEW_CLI=1 to opt into the new CLI syntax without warnings or fallback.
```

src/bin/wasmtime.rs
```diff
@@ -99,11 +99,176 @@ impl Wasmtime {
...
+            //
+            // Here assume that the old semantics are desired but emit a warning
+            // indicating that this will change in the future.
+            (Err(_new), Ok(old)) => {
+                eprintln!(
+                    "\
+warning: this CLI invocation of Wasmtime is going to break in the future, for
+         more information see this issue online:
+         https://github.com/bytecodealliance/wasmtime/issues/7384
+         use `WASMTIME_NEW_CLI=0` to temporarily disable this warning\
```
cfallin, line 218: Likewise here; perhaps we can have a specific message ("this invocation is going to break" vs. "this invocation is parsed differently by old and new") and then a general informational blob afterward?

cfallin (approved):
A few thoughts on the messages but otherwise LGTM; I jumped in here as I had some thoughts but others may want to take a look too to make sure we're not missing anything.

Thanks for putting this together so quickly!

src/bin/wasmtime.rs
cfallin: Ah, actually, I see the different `NEW_CLI=0` vs `NEW_CLI=1` in the two different messages now; I like the idea of giving exactly the setting that will give the same behavior but without the warning but still think we should give the full picture. Could we give the above info then append "Use `WASMTIME_NEW_CLI=x` to silence this warning while retaining the same behavior in this case."?

alexcrichton:
Mind re-reviewing this? I'm trying to be really careful about the wording here since this message is likely all many folks will see and/or read so I'm trying to make it not too long, not too short, actionable, and accurate.

src/bin/wasmtime.rs
```diff
@@ -99,11 +99,194 @@ impl Wasmtime {
...
+                } else {
+                    // this error message is not statically reachable due to
+                    // `DEFAULT_OLD_BEHAVIOR=true` at this time, but when that
+                    // changes this should be updated to have a more accurate
+                    // set of text.
+                    assert!(false);
+                    eprintln!(
+                        "\
+warning: this CLI invocation of Wasmtime is parsed differently now than it was
+         prior, and the new semantics are going to be used instead of the old
```
cfallin, line 199: "going to be used" reads somewhat ambiguously to me: that could mean "in a few months when we flip a switch", or it could mean "in a few nanoseconds when I go ahead and call this function". Perhaps "and the new semantics are now used instead of the old semantics" (i.e., present rather than future tense)?

src/bin/wasmtime.rs
```diff
@@ -99,11 +99,194 @@ impl Wasmtime {
...
+            //
+            // Here assume that the old semantics are desired but emit a warning
+            // indicating that this will change in the future.
+            (Err(_new), Ok(old)) => {
+                eprintln!(
+                    "\
+warning: this CLI invocation of Wasmtime is going to break in the future -- for
+         more information see this issue online:
+         https://github.com/bytecodealliance/wasmtime/issues/7384
+
```
cfallin, line 232: Perhaps add a new paragraph here, a line on its own separated by blank lines above and below, that says "Executing according to the old (Wasmtime version <= 13) CLI semantics.", and above, analogously, "according to the nwe (Wasmtime version >= 14) CLI semantics."? That would make crystal-clear which behavior was chosen.

cfallin (commented):
Some more thoughts on wording here; others' eyeballs would also be useful on this phrasing!

src/bin/wasmtime.rs
alexcrichton: If you're ok I'd prefer to leave precise wordsmithing of this message to later. This won't ever actually get executed due to the value of `DEFAULT_OLD_BEHAVIOR` and the `assert!(false)` prior.

I don't know how to best word this message because its hedging what's going to happen in the future which I think will shape what this should precisely say.

FWIW though "going to be used" is attempting to model what's actually about to happen, e.g. `new.execute()` instead of `old.execute()` below. This is trying to be a warning of "if you intended for the old behavior then things are gonna be broken now"

src/bin/wasmtime.rs
alexcrichton: Is the version number necessary though? I would imagine for most users they're not really watching the version number that closely so to me it seems like it's adding noise to an already fairly wordy message. In theory the only way to hit this case is additionally previously-crafted command line via a script or similar.

For example this is the error message that will be printed out to Go and TinyGo users who probably aren't interacting with the script that's running Wasmtime at all and may not even be aware that Wasmtime is used under the hood. In such a situation all they'll want to do is (a) disable the warning and (b) open an issue upstream about fixing the warning.

src/bin/wasmtime.rs
cfallin: Sure, we can come back to it later! My thoughts still apply; fwiw, data point of one person, "going to be used" confused me, so I think it should probably be reworded somehow by the time we use this error message if we're hoping to be accurate/unambiguous.

src/bin/wasmtime.rs
cfallin: I guess the version number is still relevant to these folks because there will be guidance around "Go 1.21 is broken with Wasmtime 14 before patch release"; and my experience with error messages at least is that I'd rather have more information than less, so I don't have to infer what's going on. Again, data point of one person, but this was a pretty confusing message to read without that detail, and you asked for feedback re: ambiguity and comprehensibility; it's always possible that my brain is atypical though :-)

cfallin (approved):
r+ with above comments; I don't want to hold up the patch release so I'll leave it to your best judgment what final adjustments to make.

src/bin/wasmtime.rs
alexcrichton: Hm I'd still like to dig in to what you're thinking because I don't think I fully understand. One difference though is that I approach this sort of design assuming users are reactionary. For example I would predict that users won't a-priori take a look at their system and ensure that all the versions are aligned based on a blog post we publish. Instead they'll execute their normal workflow or some workflow documented elsewhere and then all of a sudden a warning pops up. In that sense they may not be aware that Wasmtime was even being used let alone perhaps what vector Wasmtime was installed through. In that sense I'd assume the immediate questions are:

* How do I resolve this warning now? For this the warning explains `WASMTIME_NEW_CLI=0` is the way to go. Most users might just stop here.
* How do I resolve the warning for the future? The warning for this explains that this will break, links to an issue with more context, and provides the env var to see the breakage. Further action though in fixing this would require them knowing how Wasmtime was invoked/installed and how to update, which I don't think the warning can help with.

In that sense I'm not sure I understand what you mean by this message requiring inference as to what's going on. My hope was to have full context behind an issue for interested folks, so not everything is explained here, but I'm not sure how adding versioning information would be helpful.

cfallin: I guess there are two "user stories" here:
- the person who is unknowingly invoking Wasmtime indirectly, e.g. as part of `go run` or a CI shell script or ... -- and wants to fix the warning (short-term and long-term);
- the person who is actually developing scripts or tools that invoke Wasmtime, understands that there are different versions, and is trying to parse this error message closely to figure out what's going on.

For the first case (your target), I think the error message is right on as it is. "Set this env var and move on" etc. For the second case (I guess I see myself in these shoes), it seems like it might be useful to have a mental model of what the parser actually did. I might see that it succeeded but there's still a warning and be a bit confused; the immediate suggestion is to set an environment variable; what I really want is direct feedback of "I interpreted this the old way" or "I interpreted this the new way". Without that, it's a little bit of an opaque magic box and I can see it being confusing. (Basically, my intuition/instinct here is to make the mechanism as explicit and visible as possible: there are two ways to parse, here's what I did.)

Anyway, please do feel free to go either way on this -- it's pretty subjective, I'm happy to expound on what intuition is telling me, but I am just one data point and it's possible (likely even?) that as long as we (i) do the right thing by default and (ii) give a GitHub issue link, that'll cover most cases :-)

alexcrichton: No no thank you for digging in here! There's no huge urgency as I don't plan on doing a patch release until Monday (unless someone wants to do it today themselves), so I wanted to spend some time to dig in more. What you say makes sense and I think I understand better, lemme take another pass at this.

alexcrichton: Ok, how about the current wording?

cfallin: Looks good to me at least! (one grammar nit, s/silences/silence/, "to ... silence this warning", a parallel construction thing; now you can throw me out the window because grammar nerds are annoying)

alexcrichton: It's ok I'm the son of two english majors so it's really shame on me!

PR #7386, opened 2023-10-26
Fix a possible panic in clock subscription

alexcrichton:
Change an `a - b` expression to `a.saturating_sub(b)` to handle the case when a clock is subscribed to an instance/duration in the past.


alexcrichton:
This fixes a spurious test failure [found here](https://github.com/bytecodealliance/wasmtime/actions/runs/6657952744/job/18093822208)

fitzgen (approved):

PR #7388, opened 2023-10-27
preview2: Use blocking io for stdout and stderr

### LINKED ISSUE #7258: wasi-streams for stdio: Spawned worker threads should not abort when the write end is dropped
Currently the worker thread managing writes to stdout will drop when the guest-side write end is dropped. We need to fix this, as it's quite common to repeatedly call `std::io::get_stdout()` and drop the result in rust.

The general approach that we discussed is to have the stdout/stderr worker tasks alive for the duration of the guest, and a separate channel for reporting new write channels to those tasks. The tasks would then be responsible for selecting between the new writer channel and all the known writer channels when processing incoming jobs, removing known writers from the managed set when those ends drop instead of aborting.

elliottt:
Use blocking IO for stdout and stderr `HostOutputStream` implementations. The reasoning here is that these should never block unless something outside of our control is failing to consume the streams.

This doesn't resolve #7258, but it does take the immediate pressure off, as the only instance of it was `stdout` and `stderr`.

### DIFF
--- a/crates/wasi/src/preview2/stdio.rs
+++ b/crates/wasi/src/preview2/stdio.rs
@@ -3,8 +3,11 @@ use crate::preview2::bindings::cli::{
     terminal_stdout,
 };
 use crate::preview2::bindings::io::streams;
-use crate::preview2::pipe::{self, AsyncWriteStream};
-use crate::preview2::{HostInputStream, HostOutputStream, WasiView};
+use crate::preview2::pipe;
+use crate::preview2::{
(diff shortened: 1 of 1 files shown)


alexcrichton:
This is definitely what I was thinking with respect to [this comment](https://github.com/bytecodealliance/wasmtime/issues/7258#issuecomment-1780151465), thanks!

I'd definitely want @pchickey to weigh in though as I suspect I'm forgetting something.

pchickey (approved):
LGTM modulo adding some comments as we discussed! Thanks for figuring this mess out.

alexcrichton (approved):

PR #7389, opened 2023-10-27
PCC: verification primitives for dynamic range checks.

cfallin:
This PR adds new kinds of facts on value `dynamic_range`, `dynamic_mem`, and `compare`; a new kind of memory type `dynamic_memory`; and a notion of symbolic expressions (a sum of either an SSA value or global value and a static offset) for min and max values in those ranges that are just rich enough to express the expected CLIF sequence for Wasm dynamic memory bounds-checks with Spectre mitigations (cmp/select rather than control-flow-based). It is sufficient to validate the following

```
; Equivalent to a Wasm `i64.load` from a dynamic memory.                                                                                                                                     
function %f0(i64 vmctx, i32) -> i64 {                                                                                                                                                         
    gv0 = vmctx                                                                                                                                                                               
    gv1 = load.i64 notrap aligned checked gv0+0 ;; base                                                                                                                                       
    gv2 = load.i64 notrap aligned checked gv0+8 ;; size                                                                                                                                       
                                                                                                                                                                                              
    ;; mock vmctx struct:                                                                                                                                                                     
    mt0 = struct 16 {                                                                                                                                                                         
        0: i64 readonly ! dynamic_mem(mt1, 0, 0),                                                                                                                                             
        8: i64 readonly ! dynamic_range(64, gv2, gv2),                                                                                                                                        
    }                                                                                                                                                                                         
    ;; mock dynamic memory: dynamic range, plus 2GiB guard                                                                                                                                    
    mt1 = dynamic_memory gv2 + 0x8000_0000                                                                                                                                                    
                                                                                                                                                                                              
block0(v0 ! mem(mt0, 0, 0): i64, v1 ! dynamic_range(32, v1, v1): i32):                                                                                                                        
    v2 ! dynamic_range(64, v1, v1)            = uextend.i64 v1       ;; extended Wasm offset                                                                                                  
    v3 ! dynamic_mem(mt1, 0, 0)               = global_value.i64 gv1 ;; base                                                                                                                  
    v4 ! dynamic_range(64, gv2, gv2)          = global_value.i64 gv2 ;; size                                                                                                                  
    v5 ! compare(uge, v1, gv2)                = icmp.i64 uge v2, v4  ;; bounds-check compare of extended Wasm offset to size                                                                  
    v6 ! dynamic_mem(mt1, v1, v1)              = iadd.i64 v3, v2      ;; compute access address: memory base plus extended Wasm offset                                                        
    v7 ! dynamic_mem(mt1, 0, 0, nullable)     = iconst.i64 0         ;; null pointer for speculative path                                                                                     
    v8 ! dynamic_mem(mt1, 0, gv2-1, nullable) = select_spectre_guard v5, v7, v6  ;; if OOB, pick null, otherwise the real address                                                             
    v9                                        = load.i64 checked v8                                                                                                                           
    return v9                                                                                                                                                                                 
}  
```

on both x86-64 and aarch64.

The first half of this was:

### DIFF
--- a/cranelift/codegen/src/ir/pcc.rs
+++ b/cranelift/codegen/src/ir/pcc.rs
@@ -321,7 +321,10 @@ impl Expr {
                 offset: lhs.offset.checked_add(rhs.offset)?,
             })
         } else {
-            None
+            Some(Expr {
+                base: BaseExpr::Max,
+                offset: 0,
+            })
         }
     }
 
@@ -585,7 +588,7 @@ impl Fact {
                     ty: *ty_lhs,
                     min: Expr::max(min_lhs, min_rhs),
                     max: Expr::min(max_lhs, max_rhs),
-                    nullable: *null_lhs || *null_rhs,
+                    nullable: *null_lhs && *null_rhs,
                 }
             }
 
@@ -623,7 +626,7 @@ impl Fact {
     }
 
     /// Is this fact a single-value range with a symbolic Expr?
-    fn is_single_expr(&self) -> Option<&Expr> {
+    fn as_symbol(&self) -> Option<&Expr> {
         match self {
             Fact::DynamicRange { min, max, .. } if min == max => Some(min),
             _ => None,
@@ -643,6 +646,16 @@ macro_rules! bail {
     ( $err:tt ) => {{
         return Err(PccError::$err);
     }};
+}
+
+/// The two kinds of inequalities: "strict" (`<`, `>`) and "loose"
+/// (`<=`, `>=`), the latter of which admit equality.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum InequalityKind {
+    /// Strict inequality: {less,greater}-than.
+    Strict,
+    /// Loose inequality: {less,greater}-than-or-equal.
+    Loose,
 }
 
 /// A "context" in which we can evaluate and derive facts. This
@@ -703,8 +716,12 @@ impl<'a> FactContext<'a> {
                     max: max_rhs,
                 },
             ) => {
-                // Same as above, but with dynamic-expression
-                // comparisons.
+                // Nearly same as above, but with dynamic-expression
+                // comparisons. Note that we require equal bitwidths
+                // here: unlike in the static case, we don't have
+                // fixed values for min and max, so we can't lean on
+                // the well-formedness requirements of the static
+                // ranges fitting within the bit-width max.
                 bw_lhs == bw_rhs && Expr::le(max_lhs, max_rhs) && Expr::le(min_rhs, min_lhs)
             }
 
@@ -749,11 +766,11 @@ impl<'a> FactContext<'a> {
             (
                 Fact::Range {
                     bit_width,
-                    min,
-                    max,
+                    min: 0,
+                    max: 0,
                 },
-                Fact::DynamicMem { nullable, .. },
-            ) if *bit_width == self.pointer_width && *min == 0 && *max == 0 && *nullable => true,
+                Fact::DynamicMem { nullable: true, .. },
+            ) if *bit_width == self.pointer_width => true,
 
             _ => false,
         }
@@ -1203,11 +1220,9 @@ impl<'a> FactContext<'a> {
                     let end_offset = max_offset
                         .checked_add(i64::from(size))
                         .ok_or(PccError::Overflow)?;
-                    ensure!(
-                        end_offset
-                            <= i64::try_from(*mem_static_size).map_err(|_| PccError::Overflow)?,
-                        OutOfBounds
-                    );
+                    let mem_static_size =
+                        i64::try_from(*mem_static_size).map_err(|_| PccError::Overflow)?;
+                    ensure!(end_offset <= mem_static_size, OutOfBounds);
                     Ok(None)
                 }
                 _ => bail!(OutOfBounds),
@@ -1272,15 +1287,16 @@ impl<'a> FactContext<'a> {
 
     /// Apply a known inequality to rewrite dynamic bounds using transitivity, if possible.
     ///
-    /// Given that `lhs >= rhs` (if not `strict`) or `lhs < rhs` (if
+    /// Given that `lhs >= rhs` (if not `strict`) or `lhs > rhs` (if
     /// `strict`), update `fact.
-    pub fn apply_inequality(&self, fact: &Fact, lhs: &Fact, rhs: &Fact, strict: bool) -> Fact {
-        trace!(
-            "lhs single expr = {:?} rhs single expr = {:?}",
-            lhs.is_single_expr(),
-            rhs.is_single_expr()
-        );
-        match (lhs.is_single_expr(), rhs.is_single_expr(), fact) {
+    pub fn apply_inequality(
+        &self,
+        fact: &Fact,
+        lhs: &Fact,
+        rhs: &Fact,
+        kind: InequalityKind,
+    ) -> Fact {
+        match (lhs.as_symbol(), rhs.as_symbol(), fact) {
             (
                 Some(lhs),
                 Some(rhs),
@@ -1291,12 +1307,15 @@ impl<'a> FactContext<'a> {
                     nullable,
                 },
             ) if rhs.base == max.base => {
-                trace!("bases match");
+                let strict_offset = match kind {
+                    InequalityKind::Strict => 1,
+                    InequalityKind::Loose => 0,
+                };
                 if let Some(offset) = max
                     .offset
                     .checked_add(lhs.offset)
                     .and_then(|x| x.checked_sub(rhs.offset))
-                    .and_then(|x| x.checked_sub(if strict { 1 } else { 0 }))
+                    .and_then(|x| x.checked_sub(strict_offset))
                 {
                     let new_max = Expr {
                         base: lhs.base.clone(),
--- a/cranelift/codegen/src/ir/pcc.rs
+++ b/cranelift/codegen/src/ir/pcc.rs
@@ -220,12 +220,126 @@ pub enum BaseExpr {
     /// facts even after we've lowered out of SSA: it becomes simply
     /// some symbolic value.
     Value(ir::Value),
+    /// Top of the address space. This is "saturating": the offset
+    /// doesn't matter.
+    Max,
+}
+
+impl BaseExpr {
+    /// Is one base less than or equal to another? (We can't always
+    /// know; in such cases, returns `false`.)
+    fn le(lhs: &BaseExpr, rhs: &BaseExpr) -> bool {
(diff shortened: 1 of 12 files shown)


cranelift/codegen/src/ir/pcc.rs
```diff
@@ -806,6 +1186,32 @@ impl<'a> FactContext<'a> {
...
+                ir::MemoryTypeData::DynamicMemory {
+                    gv,
+                    size: mem_static_size,
+                } if gv == max_gv => {
+                    let end_offset = max_offset
+                        .checked_add(i64::from(size))
+                        .ok_or(PccError::Overflow)?;
+                    ensure!(
+                        end_offset
+                            <= i64::try_from(*mem_static_size).map_err(|_| PccError::Overflow)?,
```
fitzgen, line 1208: Style nitpick: might be easier to read if the `try_from` expression was pulled out to its own variable, similar to how the `end_offset` is.

cranelift/codegen/src/ir/pcc.rs
```diff
@@ -863,6 +1269,53 @@ impl<'a> FactContext<'a> {
         }
         Ok(())
     }
+
+    /// Apply a known inequality to rewrite dynamic bounds using transitivity, if possible.
+    ///
+    /// Given that `lhs >= rhs` (if not `strict`) or `lhs < rhs` (if
+    /// `strict`), update `fact.
+    pub fn apply_inequality(&self, fact: &Fact, lhs: &Fact, rhs: &Fact, strict: bool) -> Fact {
```
fitzgen, line 1277: Can we define an `enum` to replace the `strict: bool` parameter? This should both make this code easier to follow and make it more obvious what callers are doing vs just passing a random bool parameter or needing to do stuff like `foo(/* strict = */ false)`.

Something like

```rust
enum InequalityKind {
    GreaterThanOrEqual,
    LessThan,
}
```

fitzgen: Also, it seems like maybe the comment is incorrect and this is `<=` or `<`, not `>=` or `<`?

cranelift/codegen/src/ir/pcc.rs
```diff
@@ -434,9 +567,68 @@ impl Fact {
...
+        }
+    }
+
+    /// Is this fact a single-value range with a symbolic Expr?
+    fn is_single_expr(&self) -> Option<&Expr> {
+        match self {
+            Fact::DynamicRange { min, max, .. } if min == max => Some(min),
+            _ => None,
+        }
+    }
```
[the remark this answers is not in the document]
fitzgen: Oh also, maybe we can call this `as_symbol` since it returns something. Just got confused for a second when reading callers.

cranelift/codegen/src/ir/pcc.rs
```diff
@@ -863,6 +1269,53 @@ impl<'a> FactContext<'a> {
...
+                Some(lhs),
+                Some(rhs),
+                Fact::DynamicMem {
+                    ty,
+                    min,
+                    max,
+                    nullable,
+                },
+            ) if rhs.base == max.base => {
+                trace!("bases match");
```
fitzgen, line 1294: Seems like maybe not the most useful trace long term.

cranelift/codegen/src/isa/aarch64/pcc.rs
```diff
@@ -254,6 +279,44 @@ pub(crate) fn check(
...
+                let rn = ctx.apply_inequality(&rn, &cmp_lhs, &cmp_rhs, false);
+                trace!(" -> rn = {rn:?}");
+                // false side: rhs < lhs, strict.
+                let rm = get_fact_or_default(vcode, rm, 64);
+                trace!("rm = {rm:?}");
+                let rm = ctx.apply_inequality(&rm, &cmp_rhs, &cmp_lhs, true);
+                trace!(" -> rm = {rm:?}");
+
+                let union = Fact::union(&rn, &rm);
+                trace!("union = {union:?}");
```
fitzgen, line 313: Let's either clean up these traces and give them more context to make them readable by general wasmtime/cranelift devs or remove them.

cranelift/codegen/src/isa/x64/pcc.rs
```diff
@@ -411,16 +435,35 @@ pub(crate) fn check(
...
+                    let in_true = get_fact_or_default(vcode, reg, 64);
+                    trace!("in_true = {:?}", in_true);
+                    let in_true = ctx.apply_inequality(&in_true, &cmp_lhs, &cmp_rhs, false);
+                    trace!("in_true = {:?}", in_true);
+                    let in_false = get_fact_or_default(vcode, alternative.to_reg(), 64);
+                    trace!("in_false = {:?}", in_false);
+                    let in_false = ctx.apply_inequality(&in_false, &cmp_rhs, &cmp_lhs, true);
+                    trace!("in_false = {:?}", in_false);
+                    let union = Fact::union(&in_true, &in_false);
+                    trace!("union = {:?}", union);
```
fitzgen, line 461: Ditto regarding traces.

cranelift/filetests/filetests/pcc/succeed/dynamic.clif
```diff
@@ -17,21 +17,21 @@ function %f0(i64 vmctx, i32) -> i64 {
...
-block0(v0 ! mem(mt0, 0, 0): i64, v1: i32):
-    v2 ! range(64, 0, 0xffff_ffff)            = uextend.i64 v1       ;; extended Wasm offset
+block0(v0 ! mem(mt0, 0, 0): i64, v1 ! dynamic_range(32, v1, v1): i32):
+    v2 ! dynamic_range(64, v1, v1)            = uextend.i64 v1       ;; extended Wasm offset
     v3 ! dynamic_mem(mt1, 0, 0)               = global_value.i64 gv1 ;; base
     v4 ! dynamic_range(64, gv2, gv2)          = global_value.i64 gv2 ;; size
-    v5 ! compare(uge, v2, gv2)                = icmp.i64 uge v2, v4  ;; bounds-check compare of extended Wasm offset to size
-    v6 ! dynamic_mem(mt1, 0, v2)              = iadd.i64 v3, v2      ;; compute access address: memory base plus extended Wasm offset
+    v5 ! compare(uge, v1, gv2)                = icmp.i64 uge v2, v4  ;; bounds-check compare of extended Wasm offset to size
+    v6 ! dynamic_mem(mt1, v1, v1)              = iadd.i64 v3, v2      ;; compute access address: memory base plus extended Wasm offset
```
fitzgen, line 25:
```suggestion
    v6 ! dynamic_mem(mt1, v1, v1)             = iadd.i64 v3, v2      ;; compute access address: memory base plus extended Wasm offset
```

fitzgen (approved):
r=me with some comments below

cranelift/codegen/src/ir/pcc.rs
```diff
@@ -220,20 +220,134 @@ pub enum BaseExpr {
...
+                base: rhs.base.clone(),
+                offset: lhs.offset.checked_add(rhs.offset)?,
+            })
+        } else if rhs.base == BaseExpr::None {
+            Some(Expr {
+                base: lhs.base.clone(),
+                offset: lhs.offset.checked_add(rhs.offset)?,
+            })
+        } else {
+            None
```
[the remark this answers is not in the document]
cfallin: Ah, good point! It arose out of a need for the other side of the lattice when defining `min` and `max`, but now that we have it we can saturate to it here too.

cranelift/codegen/src/ir/pcc.rs
```diff
@@ -434,9 +567,68 @@ impl Fact {
...
+                    ty: ty_rhs,
+                    min: min_rhs,
+                    max: max_rhs,
+                    nullable: nullable_rhs,
+                },
+            ) if ty_lhs == ty_rhs => Some(Fact::DynamicMem {
+                ty: *ty_lhs,
+                min: Expr::min(min_lhs, min_rhs),
+                max: Expr::max(max_lhs, max_rhs),
+                nullable: *nullable_lhs || *nullable_rhs,
```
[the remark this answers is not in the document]
cfallin: Ah, yeah, intersect was wrong -- it should have been an AND. Fixed, thanks!

cranelift/codegen/src/ir/pcc.rs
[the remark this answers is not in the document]

cranelift/codegen/src/ir/pcc.rs
```diff
@@ -499,6 +691,23 @@ impl<'a> FactContext<'a> {
...
+                },
+                Fact::DynamicRange {
+                    bit_width: bw_rhs,
+                    min: min_rhs,
+                    max: max_rhs,
+                },
+            ) => {
+                // Same as above, but with dynamic-expression
+                // comparisons.
+                bw_lhs == bw_rhs && Expr::le(max_lhs, max_rhs) && Expr::le(min_rhs, min_lhs)
```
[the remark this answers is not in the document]
cfallin: I was having some trouble convincing myself it works here... in the static case, we know that the concrete limit values are within the range provided by the bitwidth, so if the range gets bigger on the RHS (the subsuming case), and the bitwidth gets smaller, that implies that the claimed range values also fit within the bitwidth on the left, and so no claims are being made about upper bits. Whereas we don't know that about symbolic values. (It's still a little handwavy but that's the gist) I've added a comment to this effect.

cranelift/codegen/src/ir/pcc.rs
```diff
@@ -516,6 +725,36 @@ impl<'a> FactContext<'a> {
...
+
+            // Constant zero subsumes nullable DynamicMem pointers.
+            (
+                Fact::Range {
+                    bit_width,
+                    min,
+                    max,
+                },
+                Fact::DynamicMem { nullable, .. },
+            ) if *bit_width == self.pointer_width && *min == 0 && *max == 0 && *nullable => true,
```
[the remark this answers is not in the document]

cranelift/codegen/src/ir/pcc.rs

cranelift/codegen/src/ir/pcc.rs
cfallin: Done on both -- indeed, the comment was incorrect in the second case, thanks!

cranelift/codegen/src/ir/pcc.rs
cfallin: Removed; I neglected to go through and clean these up after getting things working, sorry!

cranelift/codegen/src/isa/aarch64/pcc.rs
cfallin: Done

cranelift/codegen/src/isa/x64/pcc.rs
cfallin: Done!

cranelift/filetests/filetests/pcc/succeed/dynamic.clif
cfallin: Done!

PR #7392, opened 2023-10-27
[14.0.0] Enable the `wasmtime serve` command by default

alexcrichton:
This enables the `serve` subcommand by default on the CLI instead of having it behind an off-by-default feature. In general our default build is the "most featureful" while we still retain the ability to turn off everything if desired for custom builds. This additionally matches the current `main` branch where through other refactorings the `serve` command is now enabled by default.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -282,6 +282,7 @@ default = [
   "wasi-threads",
   "wasi-http",
   "pooling-allocator",
+  "serve",
 ]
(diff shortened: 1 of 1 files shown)


alexcrichton:
For the `main` branch this happened in https://github.com/bytecodealliance/wasmtime/pull/7282 through other refactorings but I don't want to backport that whole PR, so this is a customized feature for the 14.0.0 branch. This'll get included with the upcoming point release next week.

PR #7393, opened 2023-10-27
mpk: turn on `memory_protection_keys` during fuzzing

abrown:
This also twists the `max_memory_protection_keys` knob.

### DIFF
--- a/crates/fuzzing/src/generators/pooling_config.rs
+++ b/crates/fuzzing/src/generators/pooling_config.rs
@@ -1,6 +1,7 @@
 //! Generate instance limits for the pooling allocation strategy.
 
 use arbitrary::{Arbitrary, Unstructured};
+use wasmtime::MpkEnabled;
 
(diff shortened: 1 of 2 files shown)


abrown:
I've been running some fuzzing locally with this and it seems OK:

- currently at `#227829` of `cargo +nighly fuzz run instantiate`
- currently at `#160974` of `cargo +nighly fuzz run spectests`
- I had trouble running the `differential` target, though: I ran into build issues with the `wasm-spec-interpreter` and, with that disabled, I observed a segfault from V8. Has anyone else run this target recently?

This PR never really force-enables the MPK feature: it is only ever set to `auto`, so if the machine doing the fuzzing does not have MPK enabled, this feature won't really get tested.

alexcrichton (approved):
Unsure what's going on with v8, but are you sure it's not related to mpk?

Otherwise looks good to me and let's see what oss-fuzz says

PR #7130, opened 2023-10-02
Refactor how `subscribe` works in WASI

alexcrichton:
This commit refactors how the `Pollable` resource is created in WASI. Previously this was created manually via two variants of a `Pollable` enum but this was somewhat error prone for a few reasons:

* The more common representation, `TableEntry`, had boilerplate associated with it that had dynamic downcasts and such. This is now all hidden behind a more typed API.

* Callers had to remember to use `push_child_resource` which is easy to forget.

* The previous signature of the readiness check returned a `Result<()>` which made it accidentally easy for errors to be propagated into traps rather than being returned as a "last operation failed" error.

This commit replaces the previous API with a single `subscribe` function which takes a `Resource<T>` and returns a `Resource<Pollable>`. The `T` type must implement a new trait, `Subscribe`, which indicates the asynchronous readiness check. This readiness check additionally returns `()` so it's no longer possible to accidentally return errors (or trap).

This namely required refactoring the `HostOutputStream` trait and implementations. The trait itself now has a `check_write` method corresponding to to the same WASI function. The old `write_ready` function is now a convenience helper combo between `ready` and `check_write`. Concrete implementations are refactored to store errors discovered during `ready` to get communicated through `check-write` later on.

PR #7394, opened 2023-10-27
Fix file writes in preview 2 implementation

### LINKED ISSUE #7390: file truncation with preview2 adapter?
i have a rust program which uses the [image](https://docs.rs/image/latest/image/) crate (0.24.7 if it matters)
to write png/jpg files to filesystem.
i build it for the `wasm32-wasi` target, and adapt it with the preview2 adapter,
and run it on my wasmtime-based embedder.

while small image files are generated fine, it seems larger files (> 4096 bytes?) are somehow corrupted.

i'm using wasmtime 14.0.2.

### Test Case

unfortunately i can't share my programs.

### Steps to Reproduce

see above.

### Expected Results

see above.

### Actual Results

see above.

### Versions and Environment

Wasmtime version or commit: 14.0.2

Operating system: macOS

Architecture: amd64

pchickey:
In https://github.com/bytecodealliance/wasmtime/pull/7130, there was a regression introduced in FileOutputStream where the position cursor was not updated after each write. This affects all file writes except for when the file is opened for appending.

This PR also adds a test which performs a larger (64k) write to a file than any prior tests.

Fixes #7390

### DIFF
--- a/crates/test-programs/src/bin/preview1_file_write.rs
+++ b/crates/test-programs/src/bin/preview1_file_write.rs
@@ -1,0 +1,106 @@
+use std::{env, process};
+use test_programs::preview1::open_scratch_directory;
+
+unsafe fn test_file_long_write(dir_fd: wasi::Fd, filename: &str) {
+    // Open a file for writing
+    let file_fd = wasi::path_open(
+        dir_fd,
+        0,
+        filename,
+        wasi::OFLAGS_CREAT,
+        wasi::RIGHTS_FD_WRITE,
+        0,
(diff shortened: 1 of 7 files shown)


alexcrichton:
Should this be backported to 14.0.x as well?

pchickey:
Yes. I was going to write more tests and I was chasing down some other potential improvements as well, but the bug here is obvious now, so I think we should just land this patch in main and in the next 14 release to go out.

alexcrichton (approved):

PR #7402, opened 2023-10-30
Don't panic in `wasmtime serve` when `set` isn't called

alexcrichton:
This commit fixes a panic in the `wasmtime serve` CLI command when the `response-outparam::set` method was never invoked by the guest. In this situation return a more first-class error than a panic explaining what the guest should be doing.

### DIFF
--- a/src/commands/serve.rs
+++ b/src/commands/serve.rs
@@ -380,8 +380,11 @@ impl hyper::service::Service<Request> for ProxyHandler {
         });
 
         Box::pin(async move {
-            let resp = receiver.await.unwrap()?;
-            Ok(resp)
+            match receiver.await {
+                Ok(Ok(resp)) => Ok(resp),
+                Ok(Err(e)) => Err(e.into()),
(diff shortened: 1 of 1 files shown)


fitzgen (approved):
CLI test?

alexcrichton:
Given the lack of preexisting tests for the CLI I'm hesitant to add all the infrastructure necessary to get it implemented just for this. Would you be ok deferring that for a future PR?

fitzgen:
I thought we already had some infra? But if the `serve` command has extra needs that we don't currently support then whatever. Although we should maybe have an issue on file.

alexcrichton:
We do for [individual tests](https://github.com/bytecodealliance/wasmtime/tree/main/crates/wasi-http/tests/all) yeah but nothing through `wasmtime serve` itself, instead it's through the embedding API.

alexcrichton:
I filed https://github.com/bytecodealliance/wasmtime/issues/7403

pchickey (approved):

Issue #7068, opened 2023-09-20
Provide release details in GitHub Releases


brettcannon:
When you look at https://github.com/bytecodealliance/wasmtime/releases/tag/v13.0.0 there's no information about what is in the release (and subscribing to GitHub release notifications means releases are, at least for me, the primary way to find out about new releases). You have to navigate to https://github.com/bytecodealliance/wasmtime/blob/main/RELEASES.md#1300 in order to find out what's contained in the release.

Any way to either inline the release notes into the GitHub release or at least provide a link to the document?

alexcrichton:
I agree! The best thing to do here would probably be to copy the release notes that you've linked to the release itself, ideally automatically. Fiddling with the release process is pretty difficult though since it's difficult to test, and I suspect this'll require some well-crafted shell commands and such to get right which I'm not myself 100% sure of.

rockwotj:
Maybe in the meantime having a single link to release file is a good middle ground?

bjorn3:
Rust has a bot which copies the release notes from RELEASES.md into github releases I believe. Maybe the same could be done for Wasmtime?

alexcrichton:
You wouldn't happen to have a link on-hand to how this is integrated in to rust-lang/rust, would you? I'm not sure how to configure this all myself.

bjorn3:
Seems to be part of triagebot: https://github.com/rust-lang/triagebot/tree/master/src/changelogs Should be possible to port it to github actions though.

alexcrichton:
Ok I think this will be fixed by https://github.com/bytecodealliance/wasmtime/pull/7405

PR #7405, opened 2023-10-30
Attempt to automatically configure release notes

alexcrichton:
This commit is an attempt to tackle #7068 by configuring the release notes in Github Releases with the handwritten release notes from `RELEASES.md`. The basic idea here is to split the markdown file on `-----` delimiters and then find the one which matches the version being released. Once one is found the `body` field of the API call to create the release is configured.

Closes https://github.com/bytecodealliance/wasmtime/issues/7068

### DIFF
--- a/.github/actions/github-release/main.js
+++ b/.github/actions/github-release/main.js
@@ -91,13 +91,25 @@ async function runOnce() {
   } catch (e) {
     console.log("ERROR: ", JSON.stringify(e, null, 2));
     core.info(`creating a release`);
+
+    const releaseNotes = fs.readFileSync('RELEASES.md').toString();
+    let notes = null;
+    const opts = {
+      owner,
+      repo,
+      tag_name: name,
+      prerelease: name === 'dev',
+    };
+    if (name !== 'dev') {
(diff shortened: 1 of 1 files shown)


alexcrichton:
An example release is [here](https://github.com/alexcrichton/wasmtime/releases/tag/v15.0.0) which is where I tested this.

.github/actions/github-release/main.js
```diff
@@ -91,13 +91,25 @@ async function runOnce() {
...
+    const releaseNotes = fs.readFileSync('RELEASES.md').toString();
+    let notes = null;
+    const opts = {
+      owner,
+      repo,
+      tag_name: name,
+      prerelease: name === 'dev',
+    };
+    if (name !== 'dev') {
+      for (let x of releaseNotes.split(/^-+$/m)) {
```
abrown, line 104: Do we need to be more defensive? What about lists that start a line with `-` or a `--` that somehow gets formatted to the beginning of a line?

```suggestion
      for (let x of releaseNotes.split(/^---+$/m)) {
```

abrown (approved):

PR #7406, opened 2023-10-30
Sync with upstream wasi-http

elliottt:
Add implementations of all the new getter/setter methods for the request and response resources, and handle the signature change to outgoing request/response constructors.


pchickey (approved):
lemon squeezy

PR #7407, opened 2023-10-30
Update wasm-tools crates

alexcrichton:
This commit updates the wasm-tools family of crate for a number of notable updates:

* bytecodealliance/wasm-tools#1257 - wasmparser's ID-based infrastructure has been refactored to have more precise types for each ID rather than one all-purpose `TypeId`.
* bytecodealliance/wasm-tools#1262 - the implementation of "implementation imports" for the component model which both updates the binary format in addition to adding more syntactic forms of imports.
* bytecodealliance/wasm-tools#1260 - a new encoding scheme for component information for `wit-component` in objects (not used by Wasmtime but used by bindings generators).

Translation for components needed to be updated to account for the first change, but otherwise this was a straightforward update.

### DIFF
--- a/crates/environ/src/component/types.rs
+++ b/crates/environ/src/component/types.rs
@@ -230,7 +230,7 @@ pub enum ComponentItem {
     Module(ModuleIndex),
     Component(ComponentIndex),
     ComponentInstance(ComponentInstanceIndex),
-    Type(wasmparser::types::TypeId),
+    Type(types::ComponentAnyTypeId),
 }
 
 /// Runtime information about the type information contained within a component.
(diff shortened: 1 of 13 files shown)


crates/environ/src/component/types.rs
```diff
@@ -493,31 +497,37 @@ impl ComponentTypesBuilder {
...
-            types::Type::Component(_) => TypeDef::Component(self.convert_component(types, id)?),
-            types::Type::ComponentInstance(_) => {
+        Ok(match id {
+            types::ComponentAnyTypeId::Defined(id) => {
+                TypeDef::Interface(self.defined_type(types, id)?)
+            }
+            // TODO
+            // types::ComponentAnyTypeId::Module(_) => {
+            //     TypeDef::Module(self.convert_module(types, id)?)
+            // }
```
pchickey, line 509: Never got back to this TODO?

alexcrichton: Oops yes, just needed to delete the code

alexcrichton: For the curious, this TODO led to https://github.com/WebAssembly/component-model/issues/269

pchickey (approved):

PR #7417, opened 2023-10-30
Turn `request-options` into a resource

### LINKED ISSUE #7409: wasi-http: make `request-options` into a resource

elliottt:
As we'd like to be able to modify `request-options` in the future without breaking compatibility with older guests, turn `request-options` into a resource. This allows us to expose getters and setters instead of fields on a record, which gives the embedder flexibility in what additional options are allowed.

Fixes #7409

### DIFF
--- a/crates/wasi-http/src/types.rs
+++ b/crates/wasi-http/src/types.rs
@@ -264,9 +264,9 @@ pub struct HostOutgoingRequest {
 
 #[derive(Default)]
 pub struct HostRequestOptions {
-    pub connect_timeout: Option<u32>,
-    pub first_byte_timeout: Option<u32>,
-    pub between_bytes_timeout: Option<u32>,
+    pub connect_timeout: Option<std::time::Duration>,
+    pub first_byte_timeout: Option<std::time::Duration>,
+    pub between_bytes_timeout: Option<std::time::Duration>,
 }
 
 pub struct HostIncomingResponse {
--- a/crates/wasi-http/src/types.rs
+++ b/crates/wasi-http/src/types.rs
@@ -262,6 +262,13 @@ pub struct HostOutgoingRequest {
     pub body: Option<HyperOutgoingBody>,
 }
 
+#[derive(Default)]
+pub struct HostRequestOptions {
+    pub connect_timeout: Option<u32>,
+    pub first_byte_timeout: Option<u32>,
+    pub between_bytes_timeout: Option<u32>,
+}
+
 pub struct HostIncomingResponse {
     pub status: u16,
     pub headers: FieldMap,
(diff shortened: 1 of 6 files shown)


crates/wasi-http/wit/deps/http/types.wit
```diff
@@ -202,21 +202,29 @@ interface types {
...
+  resource request-options {
+    /// Construct a default `request-options` value.
+    constructor();
 
     /// The timeout for the initial connect to the HTTP Server.
-    connect-timeout-ms: option<u32>,
+    connect-timeout-ms: func() -> option<u32>;
+
+    /// Set the timeout for the initial connect to the HTTP Server.
+    set-connect-timeout-ms: func(ms: option<u32>);
```
pchickey, line 213: Each of these setters needs to return result, where the error means that the setting is not supported.

crates/wasi-http/src/types.rs
```diff
@@ -257,6 +257,13 @@ pub struct HostOutgoingRequest {
     pub body: Option<HyperOutgoingBody>,
 }
 
+#[derive(Default)]
+pub struct HostRequestOptions {
+    pub connect_timeout: Option<u32>,
```
pchickey, line 262: Lets use either `std` or `tokio::time::Duration` to represent these values in the host as early as we can.

alexcrichton (approved):
Looks good to me.

Random drive-by thought unrelated to this change specifically, could the `*-ms` bits be updated to take the `duration` type in https://github.com/bytecodealliance/wasmtime/pull/7358 which would give more precision here as well?

pchickey:
> Random drive-by thought unrelated to this change specifically, could the *-ms bits be updated to take the duration type in https://github.com/bytecodealliance/wasmtime/pull/7358 which would give more precision here as well? 

Yes, we can make that change as soon as 7358 lands.

crates/wasi-http/src/types.rs
elliottt: The downside is that the getters become quite fallible as they're turning a `std::time::Duration` back into a `u32`.

pchickey: Isnt that resolved when we switch to using a monotonic duration, which is a u64?

elliottt: `std::time::Duration::as_millis` returns a u128, so we won't avoid it completely.

elliottt: I implemented this by treating conversion failure from `u128` as an `Ok(None)` return value. Since we default all these timeouts to `None` anyway, I think a conversion failure would mean that the guest somehow managed to set a larger than u32 value to begin with, then tried to get one back. The host setting an invalid default would be the other possibility, but we don't do this anywhere in wasmtime currently.

pchickey:
You can merge with `main` and adopt `duration` now!

PR #7423, opened 2023-10-31
wasi-sockets: Sync up UDP implementation & tests with TCP

badeend:
No major functional changes:
- Moved the validation utility methods out of tcp.rs and into network.rs so that the UDP implementation can use them too.
- Extracted most platform compatibility fixes out into separate functions and moved them to a place where they can be used by UDP too.
- Applied those validations and compatibility fixes to the UDP implementation as well.
- Copied the TCP tests and adjusted them to work with UDP

### DIFF
--- a/crates/test-programs/src/bin/preview2_udp_bind.rs
+++ b/crates/test-programs/src/bin/preview2_udp_bind.rs
@@ -1,0 +1,126 @@
+use test_programs::wasi::sockets::network::{
+    ErrorCode, IpAddress, IpAddressFamily, IpSocketAddress, Network,
+};
+use test_programs::wasi::sockets::udp::UdpSocket;
+
+/// Bind a socket and let the system determine a port.
+fn test_udp_bind_ephemeral_port(net: &Network, ip: IpAddress) {
+    let bind_addr = IpSocketAddress::new(ip, 0);
+
+    let sock = UdpSocket::new(ip.family()).unwrap();
+    sock.blocking_bind(net, bind_addr).unwrap();
+
+    let bound_addr = sock.local_address().unwrap();
+
+    assert_eq!(bind_addr.ip(), bound_addr.ip());
+    assert_ne!(bind_addr.port(), bound_addr.port());
+}
+
+/// Bind a socket on a specified port.
+fn test_udp_bind_specific_port(net: &Network, ip: IpAddress) {
+    const PORT: u16 = 54321;
+
+    let bind_addr = IpSocketAddress::new(ip, PORT);
+
+    let sock = UdpSocket::new(ip.family()).unwrap();
+    sock.blocking_bind(net, bind_addr).unwrap();
+
+    let bound_addr = sock.local_address().unwrap();
+
+    assert_eq!(bind_addr.ip(), bound_addr.ip());
+    assert_eq!(bind_addr.port(), bound_addr.port());
+}
+
+/// Two sockets may not be actively bound to the same address at the same time.
+fn test_udp_bind_addrinuse(net: &Network, ip: IpAddress) {
+    let bind_addr = IpSocketAddress::new(ip, 0);
+
+    let sock1 = UdpSocket::new(ip.family()).unwrap();
+    sock1.blocking_bind(net, bind_addr).unwrap();
+
+    let bound_addr = sock1.local_address().unwrap();
+
+    let sock2 = UdpSocket::new(ip.family()).unwrap();
+    assert!(matches!(
+        sock2.blocking_bind(net, bound_addr),
+        Err(ErrorCode::AddressInUse)
+    ));
+}
+
+// Try binding to an address that is not configured on the system.
+fn test_udp_bind_addrnotavail(net: &Network, ip: IpAddress) {
+    let bind_addr = IpSocketAddress::new(ip, 0);
+
+    let sock = UdpSocket::new(ip.family()).unwrap();
+
+    assert!(matches!(
+        sock.blocking_bind(net, bind_addr),
+        Err(ErrorCode::AddressNotBindable)
+    ));
+}
+
+/// Bind should validate the address family.
+fn test_udp_bind_wrong_family(net: &Network, family: IpAddressFamily) {
+    let wrong_ip = match family {
+        IpAddressFamily::Ipv4 => IpAddress::IPV6_LOOPBACK,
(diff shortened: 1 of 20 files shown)


crates/test-programs/src/bin/preview2_udp_bind.rs
```diff
@@ -0,0 +1,126 @@
...
+    let bind_addr = IpSocketAddress::new(ip, PORT);
+
+    let sock = UdpSocket::new(ip.family()).unwrap();
+    sock.blocking_bind(net, bind_addr).unwrap();
+
+    let bound_addr = sock.local_address().unwrap();
+
+    assert_eq!(bind_addr.ip(), bound_addr.ip());
+    assert_eq!(bind_addr.port(), bound_addr.port());
+}
```
alexcrichton, line 32: Can this fail if the port is already bound? We've had issues with tests running concurrently in the past and I think the TCP test has a case where it handles failure to bind here.

alexcrichton (approved):
Thanks!

Sort of a side-question as it's not directly related to this PR itself, where's all this error-handling code coming from? This seems extensive enough that the tests probably aren't exercising all the various cases here and I'm curious if you're porting from another library for example or something like that.

crates/test-programs/src/bin/preview2_udp_bind.rs
badeend: Ah right. Hadn't noticed the TCP test changed recently. Yes, this UDP test needs the same update.

badeend:
> where's all this error-handling code coming from? (...) I'm curious if you're porting from another library for example or something like that.

Not any single place but a combination of: "official" documentations, stack overflow, OS source code, trial-and-error, ..

> This seems extensive enough that the tests probably aren't exercising all the various cases

Are you referring to the error mappings in `network.rs -> tcp/udp_bind` and `network.rs -> tcp_accept`? Those are indeed not being tested, because I don't know how to without hogging up system resources or them being unreliable.

alexcrichton:
Ah ok makes sense. It's not really anything in particular I'm worried about or looking at, but it's lots of little things like:

* https://github.com/bytecodealliance/wasmtime/blob/f63350e06bf032e4c60bc460e4acbbfa6ce5bf48/crates/wasi/src/preview2/host/tcp.rs#L49
* https://github.com/bytecodealliance/wasmtime/blob/f63350e06bf032e4c60bc460e4acbbfa6ce5bf48/crates/wasi/src/preview2/host/tcp.rs#L203
* https://github.com/bytecodealliance/wasmtime/blob/f63350e06bf032e4c60bc460e4acbbfa6ce5bf48/crates/wasi/src/preview2/host/tcp.rs#L254-L268
* https://github.com/bytecodealliance/wasmtime/blob/f63350e06bf032e4c60bc460e4acbbfa6ce5bf48/crates/wasi/src/preview2/host/tcp.rs#L517-L525
* https://github.com/bytecodealliance/wasmtime/blob/f63350e06bf032e4c60bc460e4acbbfa6ce5bf48/crates/wasi/src/preview2/host/tcp.rs#L699-L702

I'm not worried about any of these, mostly curious instead how they all came into existence and if you hit issues locally or if you're copying. Either way is fine, and what you say makes sense, so sounds good to me!

badeend:
> * https://github.com/bytecodealliance/wasmtime/blob/f63350e06bf032e4c60bc460e4acbbfa6ce5bf48/crates/wasi/src/preview2/host/tcp.rs#L49

This is a standard POSIX error code for the bind syscall:
`EAFNOSUPPORT: The specified address is not a valid address for the address family of the specified socket.`
and complements our own validation slightly higher up in the function. The mapping is there just in case there is an edge case our validation logic didn't catch.

---

> * https://github.com/bytecodealliance/wasmtime/blob/f63350e06bf032e4c60bc460e4acbbfa6ce5bf48/crates/wasi/src/preview2/host/tcp.rs#L203

From the [docs](https://learn.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-listen#:~:text=is%20already%20connected.-,WSAEMFILE,-No%20more%20socket)

---

> * https://github.com/bytecodealliance/wasmtime/blob/f63350e06bf032e4c60bc460e4acbbfa6ce5bf48/crates/wasi/src/preview2/host/tcp.rs#L254-L268

From the [docs](https://man7.org/linux/man-pages/man2/accept.2.html#:~:text=Linux%20accept()%20(and%20accept4())%20passes%20already%2Dpending%20network%0A%20%20%20%20%20%20%20errors%20on%20the%20new%20socket%20as%20an%20error%20code%20from%20accept().%20%20This%0A%20%20%20%20%20%20%20behavior%20differs%20from%20other%20BSD%20socket%20implementations.)

---

> * https://github.com/bytecodealliance/wasmtime/blob/f63350e06bf032e4c60bc460e4acbbfa6ce5bf48/crates/wasi/src/preview2/host/tcp.rs#L517-L525

CI discovered that for me. This is actively tested by:

https://github.com/bytecodealliance/wasmtime/blob/f63350e06bf032e4c60bc460e4acbbfa6ce5bf48/crates/test-programs/src/bin/preview2_tcp_sockopts.rs#L46-L49

---

> * https://github.com/bytecodealliance/wasmtime/blob/f63350e06bf032e4c60bc460e4acbbfa6ce5bf48/crates/wasi/src/preview2/host/tcp.rs#L699-L702

From the [docs](https://man7.org/linux/man-pages/man7/socket.7.html#:~:text=The%20kernel%20doubles%20this%20value%20(to%20allow%20space%20for%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20bookkeeping%20overhead)%20when%20it%20is%20set%20using%20setsockopt(2)%2C%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20and%20this%20doubled%20value%20is%20returned%20by%20getsockopt(2).). This is actively tested by:

https://github.com/bytecodealliance/wasmtime/blob/f63350e06bf032e4c60bc460e4acbbfa6ce5bf48/crates/test-programs/src/bin/preview2_tcp_sockopts.rs#L75-L79

crates/test-programs/src/bin/preview2_udp_bind.rs
badeend: I've updated the test with the same workaround that's in the TCP test.

alexcrichton:
Thanks for the explanations! When you get a chance it might be goood to have some of that commentary in the code as well.

PR #7412, opened 2023-10-30
Add `wasi_http_double_echo` test

dicej:
This tests streaming outgoing request bodies, which we didn't previously have coverage for.

PR #7426, opened 2023-10-31
wasi-http: Hold the incoming body worker handle in an Arc

### LINKED ISSUE #7413: wasi-http: dropping `incoming-response` silently closes request connection
While implementing a test for outbound request body streaming (https://github.com/bytecodealliance/wasmtime/pull/7412), I discovered that dropping the `incoming-response` object would cause the request connection to be closed, even if I still had open handles to the request body and stream.

For example, if I insert `drop(response);` to https://github.com/dicej/wasmtime/blob/1782a165754a2dd62f45cbc670603cc85477323b/crates/test-programs/src/bin/api_proxy_streaming.rs#L129, the test will fail with no (or few) bytes copied.  The behavior I expected was that the test would pass (i.e. all bytes would be copied) regardless of whether the response handle was dropped or not.

elliottt:
Hold the handle to the worker task in an `Arc` to allow us to keep the worker alive through either the `HostIncomingResponse` or the `HostIncomingBody` values that manage it through its lifetime.

I tested this by adding the `drop` call that @dicej mentioned in #7413 on #7412, and verifying that the test now passes.

Fixes #7413

### DIFF
--- a/crates/wasi-http/src/body.rs
+++ b/crates/wasi-http/src/body.rs
@@ -20,6 +20,7 @@ pub type HyperIncomingBody = BoxBody<Bytes, anyhow::Error>;
 /// want to do that unless the user asks to consume the body.
 pub struct HostIncomingBodyBuilder {
     pub body: HyperIncomingBody,
+    pub worker: Option<Arc<AbortOnDropJoinHandle<anyhow::Result<()>>>>,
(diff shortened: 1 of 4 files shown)


dicej (approved):
Thanks, @elliottt!

pchickey:
Thanks for finding the bug @dicej!

PR #7427, opened 2023-10-31
Rename poll-list to poll, poll-one to pollable.block, and introduce pollable.ready

pchickey:
Downstreaming changes to wasi:io/poll proposed here:
https://github.com/WebAssembly/wasi-io/pull/54

poll-list becomes poll and poll-one becomes a method directly on pollable - these are NFCs, just renamings for ergonomics, and because poll-one was easily confused with poll-oneoff.

pollable.ready is new functionality that allows you to check the readiness of a pollable without blocking. I took the most trivial path to implement it via futures::poll_immediate. I had to add a minor optimization to the clock deadline future so that we can test it by asserting `subscribe_duration(0).ready()` and `subscribe_instant(now() - 1).ready()` are true.

### DIFF
--- a/crates/wasi/src/preview2/mod.rs
+++ b/crates/wasi/src/preview2/mod.rs
@@ -148,8 +148,9 @@ pub mod bindings {
                 "[method]output-stream.blocking-write-and-flush",
                 "[method]output-stream.blocking-write-zeroes-and-flush",
                 "[method]directory-entry-stream.read-directory-entry",
-                "poll-list",
-                "poll-one",
+                "poll",
+                "[method]pollable.block",
+                "[method]pollable.ready",
             ],
         },
         trappable_error_type: {
--- a/crates/wasi/src/preview2/poll.rs
+++ b/crates/wasi/src/preview2/poll.rs
@@ -132,7 +132,10 @@ impl<T: WasiView> crate::preview2::bindings::io::poll::HostPollable for T {
         let pollable = table.get(&pollable)?;
         let ready = (pollable.make_future)(table.get_any_mut(pollable.index)?);
         futures::pin_mut!(ready);
-        Ok(futures::future::poll_immediate(ready).await.is_some())
+        Ok(matches!(
+            futures::future::poll_immediate(ready).await,
+            Some(())
+        ))
     }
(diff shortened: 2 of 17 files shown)


crates/wasi/src/preview2/mod.rs
```diff
@@ -148,8 +148,9 @@ pub mod bindings {
                 "[method]output-stream.blocking-write-and-flush",
                 "[method]output-stream.blocking-write-zeroes-and-flush",
                 "[method]directory-entry-stream.read-directory-entry",
-                "poll-list",
-                "poll-one",
+                "poll",
+                "[method]pollable.block",
+                "[method]pollable.ready",
```
alexcrichton, line 153: Ideally this method wouldn't be on this list since it shouldn't block, but I can also see how that complicates the implementation, so I'm fine either way

crates/wasi/src/preview2/poll.rs
```diff
@@ -116,19 +116,24 @@ impl<T: WasiView> poll::Host for T {
...
-}
-
-#[async_trait::async_trait]
-impl<T: WasiView> crate::preview2::bindings::io::poll::HostPollable for T {
+    async fn ready(&mut self, pollable: Resource<Pollable>) -> Result<bool> {
+        let table = self.table_mut();
+        let pollable = table.get(&pollable)?;
+        let ready = (pollable.make_future)(table.get_any_mut(pollable.index)?);
+        futures::pin_mut!(ready);
+        Ok(futures::future::poll_immediate(ready).await.is_some())
```
alexcrichton, line 135: Could this perhaps be `matches!(..., Some(()))` to emphasize this isn't accidentally throwing away a "real" result from the poll?

alexcrichton (approved):

crates/wasi/src/preview2/poll.rs
pchickey: Good idea!

crates/wasi/src/preview2/mod.rs
pchickey: Agreed. The implementation, as it stands, shouldn't ever actually be Pending, and I think we will be able to keep an eye on that invariant going forwards.

PR #7433, opened 2023-10-31
fuzzgen: Allow restricting generated opcodes with env var

afonso360:
👋 Hey,

This PR allows users to pass a environment variable to fuzzgen that restricts the instructions that it can generate. Passing `FUZZGEN_ALLOWED_OPS=ineg,ushr` restricts fuzzgen to generate only those two instructions.

Using this allows me to get better fuzzing throughput when testing a specific rule or lowering.

### DIFF
--- a/cranelift/fuzzgen/src/function_generator.rs
+++ b/cranelift/fuzzgen/src/function_generator.rs
@@ -22,6 +22,7 @@ use cranelift::prelude::{
 use once_cell::sync::Lazy;
 use std::collections::HashMap;
 use std::ops::RangeInclusive;
+use std::str::FromStr;
 use target_lexicon::{Architecture, Triple};
 
 type BlockSignature = Vec<Type>;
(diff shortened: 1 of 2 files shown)


elliottt (approved):
This looks good to me! It would be great to document this, but the most relevant doc I could find is `fuzz/README.md`. Does that seem like a reasonable place?

abrown:
Drive-by: I'd agree with the documentation piece @elliottt mentioned and would go a bit further to note that this environment variable approach (which I also used for the `differential` target) means the corpus is invalid if it was generated in a different environment &mdash; I have forgotten this fact before (e.g., when sending in a reproducer).

afonso360:
👍 Added some docs in the README.md file.

> note that this environment variable approach (which I also used for the differential target) means the corpus is invalid if it was generated in a different environment 

IIRC @elliottt made it so that having different instructions available wouldn't automatically invalidate the whole corpus (i.e. previous inputs still work). We used to have this issue a lot when first adding instructions, where each PR would invalidate the entire OSS-Fuzz corpus.

PR #7439, opened 2023-11-01
egraphs: Remove extends and reduces from the shift amount

afonso360:
👋 Hey,

This PR adds a few egraph rules that remove extends from the inputs of shift and rotate instructions. These instructions support a different shift amount type than the instruction type, so we should be able to remove them.

Additionally it also removes ireduce and iconcat for the same reasons. This works for all types except for `i128` so we restrict those.


fitzgen (approved):
Fantastic! Thanks!

PR #7442, opened 2023-11-01
Support env & dirs with serve cli

fibonacci1729:
Currently `wasmtime serve` doesn't have support for accepting environment variables or directory placements. This PR adds them.

### DIFF
--- a/src/commands/run.rs
+++ b/src/commands/run.rs
@@ -5,7 +5,7 @@
     allow(irrefutable_let_patterns, unreachable_patterns)
 )]
 
-use crate::common::{Profile, RunCommon, RunTarget};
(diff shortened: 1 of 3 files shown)


pchickey:
So this is actually a design question that I think we need to figure out - right now we are exposing the filesystem and the various cli interfaces beyond stdio to the `wasmtime serve` component, not just strictly the `wasi:http/proxy` world, because the component adapter doesn't (yet!) have support for not importing those interfaces.

I had kind of just assumed `wasmtime serve` would converge on just exposing the proxy world once we fixed the component adapter, but also I did not mention that to anyone or track it as a bug anywhere. But, before we add support for actually using the non-canonical filesystem and cli interfaces, lets figure out what the design is here and how we'll get to the promised land of proxies while leaving it open to extensions as well.

PR #7443, opened 2023-11-01
Fix Winch bug for funcs with params and locals

jeffcharles:
I noticed a differential fuzzing failure in Winch when running the following module:

```wat
(module
  (type (;0;) (func (param i64)))
  (func (;0;) (type 0) (param i64)
    (local i32)
    global.get 1
    i32.eqz
    if ;; label = @1
      unreachable
    end
    global.get 1
    i32.const 1
    i32.sub
    global.set 1
    local.get 1
    local.tee 1
    global.get 0
    i32.xor
    global.set 0
  )
  (global (;0;) (mut i32) i32.const 0)
  (global (;1;) (mut i32) i32.const 1000)
  (export "\00\01\00" (func 0))
  (export "" (global 0))
)
```

I was able to narrow a repro case down to:

```wat
(module
    (local i32)
    local.get 1
  )
  (export "t" (func 0))
)
```

This strongly hinted that there was a potential issue with locals when there is a param present.

In the area of code I have changed, `defined_locals.stack_size` may be an equal or lower value than `defined_locals_start` since `defined_locals_start` is set based on the number and type of parameters and `defined_locals.stack_size` is set based on the number and type of locals but not parameters. The range should be `defined_locals_start` to `defined_locals_start` plus `defined_locals.stack_size`.

### DIFF
--- a/winch/codegen/src/frame/mod.rs
+++ b/winch/codegen/src/frame/mod.rs
@@ -108,7 +108,7 @@ impl Frame {
             locals_size,
             vmctx_slot: LocalSlot::i64(vmctx_offset),
             defined_locals_range: DefinedLocalsRange(
-                defined_locals_start..defined_locals.stack_size,
+                defined_locals_start..(defined_locals_start + defined_locals.stack_size),
             ),
         })
     }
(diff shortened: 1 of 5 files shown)


saulecabrera (approved):
Thanks!

PR #7444, opened 2023-11-01
add `reset_adapter_state` export to adapter

dicej:
This is useful as the last step of component pre-initialization with e.g. [component-init](https://github.com/dicej/component-init), in which case we want the adapter to forget about any open handles it has (and force it to re-request the stdio handles next time they're needed) since it will be talking to a brand new host at runtime.

### DIFF
--- a/crates/wasi-preview1-component-adapter/src/lib.rs
+++ b/crates/wasi-preview1-component-adapter/src/lib.rs
@@ -95,6 +95,14 @@ impl<T, E> TrappingUnwrap<T> for Result<T, E> {
 }
 
 #[no_mangle]
+pub unsafe extern "C" fn reset_adapter_state() {
+    let state = get_state_ptr();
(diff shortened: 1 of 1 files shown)


pchickey (approved):

crates/wasi-preview1-component-adapter/src/lib.rs
```diff
@@ -94,6 +94,11 @@ impl<T, E> TrappingUnwrap<T> for Result<T, E> {
     }
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn reset_adapter_state() {
+    State::init(get_state_ptr())
```
alexcrichton, line 99: Could this handle the case where `get_state_ptr` returns null as well?

dicej: Ah, good catch!

Issue #7445, opened 2023-11-01
mpk: some GitHub CI runners have MPK enabled but fail to run MPK code


abrown:
GitHub CI runners are showing some strange behavior: on certain runners (unknown which ones), the CPUID bits claim that MPK is supported, but running any MPK code (e.g., `RDPKRU`) causes a `SIGILL` crash. #7446 disables MPK until this is resolved.

Some instances of this failure:
- https://github.com/bytecodealliance/wasmtime/actions/runs/6721387888/job/18267026371#step:15:3283
- https://github.com/bytecodealliance/wasmtime/actions/runs/6714337681/job/18247418366?pr=7335#step:15:3280

alexcrichton:
Nice! Seems like this should be easy enough to manage by updating the supported check to require Intel CPUs and filter out AMD ones then?

abrown:
Yeah, it appears that Intel and AMD differ on what is presented in that `CPUID` leaf:
- https://github.com/torvalds/linux/blob/8bc9e6515183935fa0cccaf67455c439afe4982b/arch/x86/include/asm/cpufeatures.h#L380-L383
- https://github.com/torvalds/linux/blob/8bc9e6515183935fa0cccaf67455c439afe4982b/tools/testing/selftests/mm/pkey-x86.h#L65C1-L66
- https://github.com/torvalds/linux/blob/8bc9e6515183935fa0cccaf67455c439afe4982b/tools/testing/selftests/kvm/include/x86_64/processor.h#L122-L155

nagisa:
I can reproduce this issue locally reliably on my 2nd generation EPYC and can test things out it helps anyway.

PR #7448, opened 2023-11-01
mpk: re-enable support

abrown:
We discovered that certain CPUs claim that MPK is available but then
fail when running MPK instructions; MPK support was temporarily disabled
in https://github.com/bytecodealliance/wasmtime/pull/7446; this change re-enables it.


fitzgen (commented):
This seems to delete most of our CI?

abrown:
Oh, that last commit is just there to find the problem. It will get dropped and replaced with a fix. Looks like the check will need to add an Intel-only clause.

abrown:
This is superceded by #7513; this PR was useful to diagnose the problem (recorded in #7445) but I forgot to commit to this branch with the fix.

PR #7449, opened 2023-11-01
popcnt should check for sse4.2 support in Winch

jeffcharles:
I noticed a failure in Winch's fuzzing with the following:

```wat
(module
  (type (;0;) (func (param i64) (result i64)))
  (func (;0;) (type 0) (param i64) (result i64)
    local.get 0
    i64.popcnt
  )
  (export "test" (func 0))
)
```

with a configuration where `has_popcnt` was set to `true` but `has_sse42` was set to `false` with an error message saying:

```
Cannot emit inst 'popcntq %rax, %rax' for target; failed to match ISA requirements: [Popcnt]
```

I was able to reproduce the error message with the two new file tests in this PR.

Adding a check in the macroassember to use the fallback if either flag is missing appears to resolve the issue.

### DIFF
--- a/winch/codegen/src/isa/x64/asm.rs
+++ b/winch/codegen/src/isa/x64/asm.rs
@@ -773,7 +773,10 @@ impl Assembler {
     }
 
     pub fn popcnt(&mut self, src: Reg, size: OperandSize) {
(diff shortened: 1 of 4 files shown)


fitzgen (approved):
Thanks!

PR #7450, opened 2023-11-01
Cranelift: Reassociate constants out of nested shifts

fitzgen:
This allows for more constant propagation.

### DIFF
--- a/cranelift/codegen/src/opts/cprop.isle
+++ b/cranelift/codegen/src/opts/cprop.isle
@@ -198,3 +198,13 @@
 (rule (splat32 n) (splat64 (u64_or n (u64_shl n 32))))
 (decl splat64 (u64) Constant)
 (extern constructor splat64 splat64)
+
+;; Reassociate nested shifts of constants to put constants together for cprop.
+;;
+;; ((A shift b) shift C) ==> ((A shift C) shift b)
+(rule (simplify (ishl ty (ishl ty a@(iconst _ _) b) c@(iconst _ _)))
+      (ishl ty (ishl ty a c) b))
+(rule (simplify (ushr ty (ushr ty a@(iconst _ _) b) c@(iconst _ _)))
+      (ushr ty (ushr ty a c) b))
+(rule (simplify (sshr ty (sshr ty a@(iconst _ _) b) c@(iconst _ _)))
+      (sshr ty (sshr ty a c) b))
--- a/cranelift/filetests/filetests/egraph/reassociate-constants-in-shifts.clif
+++ b/cranelift/filetests/filetests/egraph/reassociate-constants-in-shifts.clif
@@ -1,0 +1,41 @@
+test optimize
+set opt_level=speed
+target x86_64
+
+;; Test egraph rewrite rules that reassociate constants out of nested shifts.
+
+function %a(i32) -> i32 {
+block0(v0: i32):
+    v1 = iconst.i32 1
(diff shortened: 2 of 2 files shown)


cranelift/codegen/src/opts/cprop.isle
```diff
@@ -198,3 +198,43 @@
 (rule (splat32 n) (splat64 (u64_or n (u64_shl n 32))))
 (decl splat64 (u64) Constant)
 (extern constructor splat64 splat64)
+
+;; Move constants out of nested shifts to expose them for other cprop rules.
+;;
+;; ((a shift b) shift c) ==> (a shift (c + b))
```
afonso360, line 204: I think this doesn't hold if `(c + b) >= ty_bits` right? Either way I'm going to run fuzzgen on this for a while to doublecheck.

fitzgen: Ah yeah unfortunately I think you are right:

```smt
(declare-const a (_ BitVec 32))
(declare-const b (_ BitVec 32))
(declare-const c (_ BitVec 32))

(define-fun shl ((x (_ BitVec 32)) (y (_ BitVec 32))) (_ BitVec 32)
  (let ((z (bvurem y #x00000020)))
    (bvshl x z)))


(assert (not (= (shl (shl a b) c)
                (shl a (bvadd b c)))))

(check-sat)
(get-model)
```

Produces

```smt
sat
(
  (define-fun b () (_ BitVec 32)
    #x00000010)
  (define-fun a () (_ BitVec 32)
    #xf879c486)
  (define-fun c () (_ BitVec 32)
    #x00000010)
)
```

and then

```rust
fn shl(a: u32, b: u32) -> u32 {
    a << (b % 32)
}

fn foo(a: u32, b: u32, c: u32) -> u32 {
    shl(shl(a, b), c)
}

fn bar(a: u32, b: u32, c: u32) -> u32 {
    shl(a, b.wrapping_add(c))
}

fn main() {
    let a = 0xf879c486;
    dbg!(a);
    
    let b = 0x00000010;
    dbg!(b);

    let c = 0x00000010;
    dbg!(c);
    
    dbg!(foo(a, b, c));
    dbg!(bar(a, b, c));
}
```

produces

```
[src/main.rs:15] a = 4168729734
[src/main.rs:18] b = 16
[src/main.rs:21] c = 16
[src/main.rs:23] foo(a, b, c) = 0
[src/main.rs:24] bar(a, b, c) = 4168729734
```

fitzgen: Removed the rules that depended on this buggy identity.

fitzgen:
The remaining rule is correct according to Z3:

```smt
(declare-const a (_ BitVec 32))
(declare-const b (_ BitVec 32))
(declare-const c (_ BitVec 32))

(define-fun shl ((x (_ BitVec 32)) (y (_ BitVec 32))) (_ BitVec 32)
  (let ((z (bvurem y #x00000020)))
    (bvshl x z)))


(assert (not (= (shl (shl a b) c)
                (shl (shl a c) b))))

(check-sat)
(get-model)
```

produces `unsat`, meaning there is no counterexample.

afonso360 (approved):
👍 LGTM!

Coincidentally I've been working on a similar set of rules at [egraphs-shifts-2](https://github.com/afonso360/wasmtime/tree/egraphs-shifts-2). I'll submit those as a PR tomorrow!

PR #7451, opened 2023-11-02
wasi-http: Make all setters fallible

elliottt:
Rename the `header-error` variant to `validation-error`, and use that variant as the error value for all setters of `outbound-request` and `outbound-response`. This ensures that we have the opportunity to validate outbound request and response values before they're given to their respective handlers.

Validation for uri components is done by using `http:uri::Builder` with only that value, ensuring that we don't raise an error. Validation for method and scheme is only done for the `Other` cases, and is accomplished by using the fallible parsers for each type from the `http` crate.

### DIFF
--- a/crates/wasi-http/wit/deps/http/types.wit
+++ b/crates/wasi-http/wit/deps/http/types.wit
@@ -37,10 +37,14 @@ interface types {
     unexpected-error(string)
   }
 
-  /// This tyep enumerates the different kinds of errors that may occur when
-  /// setting or appending to a `fields` resource.
-  variant header-error {
+  /// This type enumerates the different kinds of errors that may occur when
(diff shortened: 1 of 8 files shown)


crates/wasi-http/wit/deps/http/types.wit
```diff
@@ -185,21 +173,21 @@ interface types {
     /// Get the Method for the Request.
     method: func() -> method;
     /// Set the Method for the Request.
-    set-method: func(method: method);
+    set-method: func(method: method) -> result;
```
pchickey, line 176: Only feedback on this PR is that the wit docs here should describe that the set fails iff the value given is syntactically invalid.

pchickey (approved):

PR #7452, opened 2023-11-02
aarch64: Add support for `tbz` and `tbnz`

alexcrichton:
I noticed these instructions when glancing at some disassembly for other code and also noticed that Cranelift didn't have support for them. This adds a few new lowerings for conditional branches in the aarch64 backend which are for testing a bit and branching if it's zero or not zero.

### DIFF
--- a/cranelift/codegen/src/isa/aarch64/inst.isle
+++ b/cranelift/codegen/src/isa/aarch64/inst.isle
@@ -835,8 +835,8 @@
 
        ;; A conditional branch which tests the `bit` of `rn` and branches
        ;; depending on `kind`.
-       (TestBranch
-        (kind TestBranchKind)
+       (TestBitAndBranch
+        (kind TestBitAndBranchKind)
         (taken BranchTarget)
         (not_taken BranchTarget)
         (rn Reg)
@@ -1212,7 +1212,7 @@
       (enum Size32
             Size64))
 
-(type TestBranchKind (enum (Z) (NZ)))
+(type TestBitAndBranchKind (enum (Z) (NZ)))
 
 ;; Helper for calculating the `OperandSize` corresponding to a type
 (decl operand_size (Type) OperandSize)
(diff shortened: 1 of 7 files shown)


cranelift/codegen/src/isa/aarch64/inst.isle
```diff
@@ -833,6 +833,15 @@
         (not_taken BranchTarget)
         (kind CondBrKind))
 
+       ;; A conditional branch which tests the `bit` of `rn` and branches
+       ;; depending on `kind`.
+       (TestBranch
```
fitzgen, line 838: Maybe name this `TestBitAndBranch`? Because `cbz` is already "test (compare) and branch" and the unique thing about `tbz` is that it is testing a bit.

cranelift/codegen/src/isa/aarch64/inst.isle
```diff
@@ -4079,6 +4090,21 @@
       (ConsumesFlags.ConsumesFlagsSideEffect
        (MInst.CondBr taken not_taken kind)))
 
+;; Helper for emitting `MInst.Tbnz` instructions.
```
fitzgen, line 4093:
```suggestion
;; Helper for emitting `MInst.TestBitAndBranch` instructions.
```

cranelift/codegen/src/isa/aarch64/inst/args.rs
```diff
@@ -764,3 +769,15 @@ impl APIKey {
         0xd503201f | (crm << 8) | (op2 << 5)
     }
 }
+
+pub use crate::isa::aarch64::lower::isle::generated_code::TestBranchKind;
+
+impl TestBranchKind {
+    /// Inverts this branch condition.
+    pub fn invert(&self) -> TestBranchKind {
```
fitzgen, line 777: Can we call this `complement`? That's what we call the `IntCC` version after a bunch of bike shedding names.

cranelift/codegen/src/isa/aarch64/inst/emit.rs
```diff
@@ -172,6 +172,22 @@ fn enc_conditional_br(
     }
 }
 
+fn enc_test_branch(kind: TestBranchKind, taken: BranchTarget, reg: Reg, bit: u8) -> u32 {
```
fitzgen, line 175: Similar regarding naming

fitzgen: And can you add a test to `emit_tests.rs` for this new instruction format?

fitzgen (approved):
r=me with comments addressed

cranelift/codegen/src/isa/aarch64/inst/emit.rs
alexcrichton: Would you be ok omitting that? I personally find the `emit_tests.rs` files onerous and unwieldy to update when the golden tests serve basically the same purpose and are much easier to work with.

fitzgen: Sure

PR #7455, opened 2023-11-02
Print the error from the wasi-http incoming-handler to stderr

JakeChampion:
Prior to this change, the error was never printed out, making it a challenge to figure out what caused the error

### DIFF
--- a/src/commands/serve.rs
+++ b/src/commands/serve.rs
@@ -382,7 +382,7 @@ impl hyper::service::Service<Request> for ProxyHandler {
                 return Err(e);
             }
 
-            Ok::<_, anyhow::Error>(())
+            Ok(())
         });
 
         Box::pin(async move {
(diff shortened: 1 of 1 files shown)


src/commands/serve.rs
```diff
@@ -373,10 +373,14 @@ impl hyper::service::Service<Request> for ProxyHandler {
...
                 .wasi_http_incoming_handler()
                 .call_handle(store, req, out)
-                .await?;
+                .await
+            {
+                eprint!("{:?}\n", e);
+                return Err(e);
+            }
 
             Ok::<_, anyhow::Error>(())
```
elliottt, line 385: You can probably remove the annotation on `Ok`, now that there's an explicit `Err` use.

elliottt (approved):
I like this change! I had one comment about using `log::error!` instead of `eprintln!`.

src/commands/serve.rs
JakeChampion:
```suggestion
            Ok(())
```

PR #7456, opened 2023-11-02
Cranelift: Break op cost ties with expression depth in egraphs

fitzgen:
This means that, when the opcode cost is the same, we prefer shallow and wide
expressions to narrow and deep. For example, `(a + b) + (c + d)` is preferred to
`((a + b) + c) + d`. This is beneficial because it exposes more
instruction-level parallelism and shortens live ranges.

Follow up PRs will add rules that take advantage of this.

Depends on https://github.com/bytecodealliance/wasmtime/pull/7465

### DIFF
--- a/cranelift/codegen/src/egraph/cost.rs
+++ b/cranelift/codegen/src/egraph/cost.rs
@@ -74,7 +74,7 @@ impl Cost {
     const DEPTH_BITS: u8 = 8;
     const DEPTH_MASK: u32 = (1 << Self::DEPTH_BITS) - 1;
     const OP_COST_MASK: u32 = !Self::DEPTH_MASK;
-    const MAX_OP_COST: u32 = Self::OP_COST_MASK >> Self::DEPTH_BITS;
+    const MAX_OP_COST: u32 = (Self::OP_COST_MASK >> Self::DEPTH_BITS) - 1;
 
     pub(crate) fn infinity() -> Cost {
         // 2^32 - 1 is, uh, pretty close to infinite... (we use `Cost`
@@ -86,13 +86,14 @@ impl Cost {
         Cost(0)
     }
 
-    fn new(opcode_cost: u32, depth: u8) -> Cost {
-        debug_assert!(
-            opcode_cost <= Self::MAX_OP_COST,
-            "Cost::new: given opcode cost of {opcode_cost} is larger than max of {}",
-            Self::MAX_OP_COST,
-        );
-        Cost((opcode_cost << Self::DEPTH_BITS) | u32::from(depth))
+    /// Construct a new finite cost from the given parts.
+    ///
+    /// The opcode cost is clamped to the maximum value representable.
+    fn new_finite(opcode_cost: u32, depth: u8) -> Cost {
+        let opcode_cost = std::cmp::min(opcode_cost, Self::MAX_OP_COST);
+        let cost = Cost((opcode_cost << Self::DEPTH_BITS) | u32::from(depth));
+        debug_assert_ne!(cost, Cost::infinity());
+        cost
     }
 
     fn depth(&self) -> u8 {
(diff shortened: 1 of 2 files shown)


cranelift/codegen/src/egraph/cost.rs
```diff
@@ -30,24 +30,86 @@ use crate::ir::Opcode;
 /// `finite()` method.) An infinite cost is used to represent a value
 /// that cannot be computed, or otherwise serve as a sentinel when
 /// performing search for the lowest-cost representation of a value.
-#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
-pub(crate) struct Cost(u32);
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) struct Cost {
+    opcode_cost: u32,
+    depth: u32,
+}
```
cfallin, line 37: Drive-by comment here (sorry!): it might be nice to pack this into a single `u32`; CLIF data structures in general are pretty sensitive to size in memory, or at least that's been my experience. Maybe 8 bits for depth and 24 for cost (since cost has a bit more range with deep trees, potential loop nest multipliers at some point, etc)?

cfallin (approved):
LGTM to this as well, thanks! Will wait for #7465 to merge and this to rebase on `main` then we can merge.

cranelift/codegen/src/egraph/cost.rs
```diff
@@ -43,12 +86,49 @@ impl Cost {
...
+    }
+}
+
+impl std::iter::Sum<Cost> for Cost {
+    fn sum<I: Iterator<Item = Cost>>(iter: I) -> Self {
+        let mut c = Self::zero();
+        for x in iter {
+            c = c + x;
+        }
+        c
```
a1phyr, line 130:
```suggestion
        iter.fold(Self::zero(), |c, x| c + x)
```

cranelift/codegen/src/egraph/cost.rs
```diff
@@ -43,12 +86,49 @@ impl Cost {
         Cost(0)
     }
 
+    fn new(opcode_cost: u32, depth: u8) -> Cost {
+        debug_assert!(
+            opcode_cost <= Self::MAX_OP_COST,
```
elliottt, line 91: Should this be strictly less-than? If the value is equal to `MAX_OP_COST`, and `depth` is `255`, we'll allow constructing the special `infinity` value.

cranelift/codegen/src/egraph/cost.rs
```diff
@@ -43,12 +86,49 @@ impl Cost {
...
+        Cost((opcode_cost << Self::DEPTH_BITS) | u32::from(depth))
+    }
+
+    fn depth(&self) -> u8 {
+        let depth = self.0 & Self::DEPTH_MASK;
+        u8::try_from(depth).unwrap()
+    }
+
+    fn op_cost(&self) -> u32 {
+        (self.0 & Self::OP_COST_MASK) >> Self::DEPTH_BITS
```
elliottt, line 104: We could skip the mask here, as we're shifting off all the bits that get zeroed by the mask.
```suggestion
        self.0 >> Self::DEPTH_BITS
```

cranelift/codegen/src/egraph/cost.rs
```diff
@@ -43,12 +86,49 @@ impl Cost {
...
         Cost(std::cmp::min(u32::MAX - 1, self.0))
     }
+
+    /// Compute the cost of the operation and its given operands.
+    ///
+    /// Caller is responsible for checking that the opcode came from an instruction
+    /// that satisfies `inst_predicates::is_pure_for_egraph()`.
+    pub(crate) fn of_pure_op(op: Opcode, operand_costs: impl IntoIterator<Item = Self>) -> Self {
+        let c = pure_op_cost(op) + operand_costs.into_iter().sum();
+        Cost::new(c.op_cost(), c.depth().saturating_add(1)).finite()
```
elliottt, line 120: Should we only add `1` to depth when the `operand_costs` was non-empty?

elliottt (approved):
:tada:

cranelift/codegen/src/egraph/cost.rs
fitzgen: Oh we could probably have an ISLE rewrite for this too!

```
(ishl (band a k1) k2) ==> (ishl k2) when shift_mask(k1) == k2 - 1
```

cc @afonso360: maybe I can nerd snipe you :)

cranelift/codegen/src/egraph/cost.rs
fitzgen: I think it is fine to have `depth == 1` for ops without operands, and only adding `1` when we have non-empty operands requires avoiding `sum()` and makes the method look less elegant so I'm not gonna do it.

cranelift/codegen/src/egraph/cost.rs
fitzgen: Reworked things a bit to make constructing infinity by accident impossible.

PR #7457, opened 2023-11-02
Add #[inline] to some trivial methods

alexcrichton:
I was seeing these show up in profiles and while inlining them didn't help it helped reduce the noise a bit.


pchickey (approved):

bjorn3:
Recent rustc nightlies can do cross-crate inlining without explicit `#[inline]`.

alexcrichton:
Oh nice! It'll be convenient to not have to `#[inline]` methods like this in the future

PR #7459, opened 2023-11-02
Enable copy-on-write memories for components

alexcrichton:
This commit fixes a mistake in component compilation where copy-on-write and lazy initialization of function tables was accidentally not performed. I believe this is an accidental regression from a previous refactor, and this commit now ensures that the shared infrastructure between components and core modules accounts for copy-on-write and lazy table initialization.


fitzgen (approved):
Any way we can write a regression test here?

alexcrichton:
Sure!

PR #7462, opened 2023-11-02
winch: Add spec tests for `local_{get,set}`

saulecabrera:
This change is a follow up to https://github.com/bytecodealliance/wasmtime/pull/7443; after it landed I realized that Winch doesn't include spec tests for local.get and local.set.

Those tests uncovered a bug on the handling of the constant pool: given Winch's singlepass nature, there's very little room know all the constants ahead of time and to register them all at once at emission time; instead they are emitted when they are needed by an instruction. Even though Cranelift's machinery is capable of deuplicated constants in the pool, `register_constant` assumes and checks that each constat should only be pushed once. In Winch's case, since we emit as we go, we need to carefully check if the constant is one was not emitted before, and if that's the case, register it. Else we break the invariant that each constant should only be registered once.

### DIFF
--- a/cranelift/codegen/src/machinst/vcode.rs
+++ b/cranelift/codegen/src/machinst/vcode.rs
@@ -1738,6 +1738,15 @@ impl VCodeConstants {
     pub fn get(&self, c: VCodeConstant) -> &VCodeConstantData {
         &self.constants[c]
     }
+
+    /// Checks if the given [VCodeConstantData] is registered as
+    /// used by the pool.
+    pub fn pool_uses(&self, constant: &VCodeConstantData) -> bool {
+        match constant {
+            VCodeConstantData::Pool(c, _) => self.pool_uses.contains_key(c),
+            _ => false,
+        }
+    }
 }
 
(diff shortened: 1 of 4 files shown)


cranelift/codegen/src/machinst/vcode.rs
```diff
@@ -1738,6 +1738,15 @@ impl VCodeConstants {
     pub fn get(&self, c: VCodeConstant) -> &VCodeConstantData {
         &self.constants[c]
     }
+
+    /// Checks if the given [VCodeConstantData] is registered as
+    /// used by the pool.
+    pub fn pool_uses(&self, constant: &VCodeConstantData) -> bool {
```
saulecabrera, line 1744: I decided to add this function here, but if there's a reason why this shouldn't be done, I could find a different mechanism guarantee the invariant that I'm after.

fitzgen: Seems like a reasonable method to me!

fitzgen (approved):

Issue #7460, opened 2023-11-02
CLI command `wast` always produces an instantiation error


dhil:
It seems no matter what file (valid Wast, invalid Wast, existing, non-existing) you feed to the `wast` CLI command, it always fails with an instantiation error, e.g.

```shell
$ ./target/debug/wasmtime wast foo.wast
thread 'main' panicked at 'error instantiating "spectest": failed to parse WebAssembly module

Caused by:
    Invalid input WebAssembly code at offset 0: magic header not detected: bad magic number', src/commands/wast.rs:33:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
The above snippet uses the debug build from commit e0bfa7336de20f76048edbdc0157ee637a2c5fea.

dhil:
This issue seems to have been resolved by #7464.

PR #7464, opened 2023-11-02
 Fix some text-format related features 

alexcrichton:
* Enable `wasmtime/wat` when the `wat` feature is enabled on the CLI
* Additionally always enable the `wasmtime/wat` feature for the
  `wasmtime-wast` crate since the text format is used by some tests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -343,7 +343,7 @@ component-model = [
   "wasmtime-wast/component-model",
   "wasmtime-cli-flags/component-model"
 ]
-wat = ["dep:wat"]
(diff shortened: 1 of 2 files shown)


abrown (approved):

abrown:
I was just going to merge this under the assumption it fixed #7460 but I see @alexcrichton requested some reviewers directly so I'll wait to see if there is more to this.

alexcrichton:
Oh that's just how the auto-assignment looks, it seems to misattribute it to an explicit action taken on my part.

PR #7466, opened 2023-11-02
Cranelift: Reassociate long and narrow chains of operations into shallow and wide trees

fitzgen:
This increases instruction-level parallelism and shrinks live ranges. It also
canonicalizes into the shallow-and-wide form for reassociating constants
together for cprop.

Depends on https://github.com/bytecodealliance/wasmtime/pull/7456


elliottt (approved):
Looks good to me!

Issue #7454, opened 2023-11-02
Cranelift: inline probestack makes valgrind unhappy


narpfel:
I tried `rustc_codegen_cranelift` on some of my projects, and found that even though the binaries appeared to run normally, they produced errors and segfaults in `valgrind`. Looking at the disassembly, it appeared that valgrind doesn’t like the way Cranelift performs stack probing.

### `.clif` Test Case
This is the most minimal Rust code that I came up with:
```rust
fn main() {
    let _xs: [u32; 5_000];
}
```

which generates the following `.clif` file:

<details>

<summary>
output file 

`main.clif/_ZN4main4main17hf30ba8656d3abcbbE.unopt.clif`

</summary>

generated by `rustc -Z codegen-backend=cranelift src/main.rs --emit=llvm-ir`

```clif
set opt_level=none
set tls_model=elf_gd
set libcall_call_conv=isa_default
set probestack_size_log2=12
set probestack_strategy=inline
set bb_padding_log2_minus_one=0
set regalloc_checker=0
set regalloc_verbose_logs=0
set enable_alias_analysis=1
set enable_verifier=0
set is_pic=1
set use_colocated_libcalls=0
set enable_float=1
set enable_nan_canonicalization=0
set enable_pinned_reg=0
set enable_atomics=1
set enable_safepoints=0
set enable_llvm_abi_extensions=1
set unwind_info=1
set preserve_frame_pointers=0
set machine_code_cfg_info=0
set enable_probestack=1
set probestack_func_adjusts_sp=0
set enable_jump_tables=1
set enable_heap_access_spectre_mitigation=1
set enable_table_access_spectre_mitigation=1
set enable_incremental_compilation_cache_checks=0
target x86_64 has_sse3=1 has_ssse3=1 has_sse41=1 has_sse42=1 has_avx=0 has_avx2=0 has_fma=0 has_avx512bitalg=0 has_avx512dq=0 has_avx512vl=0 has_avx512vbmi=0 has_avx512f=0 has_popcnt=1 has_bmi1=0 has_bmi2=0 has_lzcnt=0


function u0:8() system_v {
; symbol _ZN4main4main17hf30ba8656d3abcbbE
; instance Instance { def: Item(DefId(0:3 ~ main[b61b]::main)), args: [] }
[...]

PR #7470, opened 2023-11-03
cranelift: Make inline stackprobe unroll sequence valgrind compliant

afonso360:
👋 Hey,

This PR alters our stackprobe unroll sequences to move the stackpointer before writing to the stack. I don't think there was anything wrong with what we were doing before but it makes valgrind really unhappy. (See #7454)

I've tested this with cg_clif on x86, and it now cleanly passes valgrind for the original reported testcase.

Fixes: #7454
prtest:full

### DIFF
--- a/cranelift/codegen/src/isa/aarch64/abi.rs
+++ b/cranelift/codegen/src/isa/aarch64/abi.rs
@@ -1178,23 +1178,25 @@ impl ABIMachineSpec for AArch64MachineDeps {
 
 impl AArch64MachineDeps {
     fn gen_probestack_unroll(insts: &mut SmallInstVec<Inst>, guard_size: u32, probe_count: u32) {
-        // When manually unrolling stick an instruction that stores 0 at a
-        // constant offset relative to the stack pointer. This will
-        // turn into something like `movn tmp, #n ; stur xzr [sp, tmp]`.
+        // When manually unrolling adjust the stack pointer and then write a zero
+        // to the stack at that offset. This generates something like
+        // `sub sp, sp, #1, lsl #12` followed by `stur wzr, [sp]`.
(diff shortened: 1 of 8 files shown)


cranelift/codegen/src/isa/aarch64/abi.rs
```diff
@@ -1210,6 +1176,52 @@ impl ABIMachineSpec for AArch64MachineDeps {
...
+
+            insts.push(Self::gen_store_stack(
+                StackAMode::SPOffset(0, I8),
+                zero_reg(),
+                I32,
+            ));
+        }
+
+        // Restore the stack pointer to its original value
+        insts.extend(Self::gen_sp_reg_adjust((guard_size * probe_count) as i32));
```
fitzgen, line 1199: Out of curiosity / unrelated to this PR: do you know why it is necessary to probe via storing to the stack rather than loading from the stack? It seems like either would accomplish the task AFAICT, and loads should be cheaper than stores.

fitzgen (approved):
Thanks!

cranelift/codegen/src/isa/aarch64/abi.rs
cfallin: From the point of view of global cost, stores might actually be better, I suspect: if it faults in a new page, a load would cause a read-only mapping to the global zero page to be created, and the first write would then cause another page fault (at least if the stack is an ordinary `MAP_ANON` region). Or at least, we've seen elsewhere that it's better for first touch to be write when TLB contention is high...

fitzgen: Ah okay that makes sense. Thanks!

PR #7471, opened 2023-11-03
riscv64: Find more oportunities to encode a compressed add

afonso360:
👋 Hey,

This is a small RISC-V specific follow up to #7470. I noticed that we don't always compress the adjust sp instruction when we could do so.

In the cases where we have build a constant in a temporary register and then `add` it to the `sp`, we encode the registers in a way that was preventing the compressed instruction from being emitted. Currently we only compress an `add` if the destination register is the same as the first source register. This is the canonical expansion of `c.add`.

This PR does two things. 
* I've switched the add instruction to have the stack pointer as the first register. i.e. `add sp, t?, sp` -> `add sp, sp, t?`. This allows us to compress that instruction with the current emit code.
* I've updated the emit code to recognize this pattern and flip the source register when emitting the compressed add. This in theory should allow us to emit this instruction more frequently.

Theoretically we only need the second step to solve this, but I've decided to also change `adjust_sp` since its the canonical representation, and at least to me is visually neater having sp as the first source register.

This is built on top of #7470, but only the last two commits belong to this PR.

### DIFF
--- a/cranelift/codegen/src/isa/riscv64/abi.rs
+++ b/cranelift/codegen/src/isa/riscv64/abi.rs
@@ -336,8 +336,8 @@ impl ABIMachineSpec for Riscv64MachineDeps {
             insts.push(Inst::AluRRR {
                 alu_op: AluOPRRR::Add,
                 rd: writable_stack_reg(),
-                rs1: tmp.to_reg(),
-                rs2: stack_reg(),
+                rs1: stack_reg(),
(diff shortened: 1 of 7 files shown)


alexcrichton (approved):

alexcrichton:
(approved for the last two commits that is, I'll defer the prior commits to the other PR)

PR #7476, opened 2023-11-03
c-api: support yielding in wasmtime_store_epoch_deadline_callback

rockwotj:
Allow supporting yielding in the epoch callback in addition to continuing, additionally support a finalizer, as that is the "standard" approach to the C API. Happy to take suggestion on API alternatives.


rockwotj:
Fixed `cargo fmt`

rockwotj:
[Force push](https://github.com/bytecodealliance/wasmtime/compare/35d808925831d8ee1891dd88adf028b564b9f592..c1d60b71292db9f1e57154626894c985c7a40838): Fixed linter warning for dead code and moved the CallbackDataPtr struct now that it's not used in `store.rs`

rockwotj:
[Force push](https://github.com/bytecodealliance/wasmtime/compare/c1d60b71292db9f1e57154626894c985c7a40838..54486be0e2aedbfcba04f9b1f36bbf97831f3ee0): Fix typedef (it's not a C++ using def...)

PR #7478, opened 2023-11-03
Reset sp_offset when resetting stack in Winch

jeffcharles:
This is a fix for a fuzzer testcase triggering an assertion of `sp_offset` not equaling `locals_size` where there was a function which had spilled a local and ended with `unreachable`. I've added a minimal reproduction of the test case as the filetest `with_spilled_local.wat`. While investigating, it looks like something similar happens when `unreachable` is the last instruction in an `if` followed by an `else` (covered by the other filetest added). Having `reset_stack` also reset the `sp_offset` clears up both assertion failures.

### DIFF
--- a/winch/codegen/src/codegen/mod.rs
+++ b/winch/codegen/src/codegen/mod.rs
@@ -104,6 +104,7 @@ where
             ControlStackFrame::If {
                 reachable,
                 original_stack_len,
+                original_sp_offset,
                 ..
             } => {
                 if *reachable {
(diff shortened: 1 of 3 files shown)


winch/codegen/src/codegen/mod.rs
```diff
@@ -166,6 +174,9 @@ where
                 _ => {}
             },
         );
+        if masm.sp_offset() > target_sp {
+            masm.free_stack(masm.sp_offset() - target_sp);
+        }
```
saulecabrera, line 179: Can we also assert that `masm.sp_offset() - target_sp` equals the size of all the `Memory` entries in the value stack that got dropped?

saulecabrera (commented):
I left a minor comment the regarding an invariant that we should verify. After fixing it, we can land this.

saulecabrera (approved):

Issue #7013, opened 2023-09-12
Wasmtime wasi-sockets implementation: Finish tweaks to spec


pchickey:
@badeend has outlined these tasks remaining before we declare the wasi-sockets spec stable and ready for Preview 2:

1. [x] Remove `(set-)no-delay`
1. [x] Remove the non-essential `ip-name-lookup::resolve-addresses` parameters.
1. [x] Lift the `ip-name-lookup::resolve-addresses` hostname restriction against parsing IP addresses. Implementations will still have to parse IP addresses to decide whether or not to return an error

alexcrichton:
I believe this is now all done on `main`

PR #7483, opened 2023-11-05
wasi-sockets: Simplify ip name lookup interface

badeend:
Simplify ip name lookup interface:
- Remove the non-essential parameters for now. Post-preview2 these can be reevaluated again.
- Lift the restriction against parsing IP addresses. Before, implementations still needed to parse IP addresses to decide whether or not to return an error.

Related: https://github.com/bytecodealliance/wasmtime/issues/7013

---

FYI, the test includes a lookup to `münchen.de` to test for internationalized domain names support. But it might be a bit weird to query this random, non-affiliated DNS server in the wasmtime CI. I tried looking for a standard testdomain (like example.com), but couldn't find any.


alexcrichton (approved):
Thanks!

We can add some code to the test later on if the resolution fails to try to cover up anything spurious that might show up

badeend:
I've resolved the conflicts

PR #7488, opened 2023-11-06
Update wasm-tools dependencies

alexcrichton:
Pull in bytecodealliance/wasm-tools#1277 to switch the defaults for some WIT behaviors:

* Semicolons are now required by default in WIT files
* Wasm-encoded WIT packages now use the "new" format by default.

WIT files without semicolons can still be parsed with `WIT_REQUIRE_SEMICOLONS=0`. The "old" format for wasm-encoded WIT packages can be emitted via `WIT_COMPONENT_ENCODING_V2=0`. Note that for wasm-encoded WIT packages both the old and the new format can be decoded irregardless of env vars.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2638,7 +2638,7 @@ version = "0.0.0"
 dependencies = [
  "cargo_metadata",
  "heck",
- "wit-component",
+ "wit-component 0.18.0",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


pchickey (approved):
LGTM. Do you intend to backport these to 15?

alexcrichton:
Indeed!

PR #7494, opened 2023-11-07
Use scratch XMM register for spilling floats in Winch

jeffcharles:
I observed an assertion failure of `assertion failed: self.sp_offset == locals_size` in the `epilogue` function when there's a float on the value stack and a spill is triggered when using Winch. When we spill, we use the scratch register which results in 8 bytes being allocated on the stack and when we pop the float to return that value, we only free 4 bytes because it's a float. This changes the spilling behaviour to use a floating point scratch register instead so we increment the stack by the same amount we will decrement it by when popping the function result.

### DIFF
--- a/winch/codegen/src/abi/mod.rs
+++ b/winch/codegen/src/abi/mod.rs
@@ -89,8 +89,11 @@ pub(crate) trait ABI {
         Self::word_bits() / 8
     }
 
-    /// Returns the designated scratch register.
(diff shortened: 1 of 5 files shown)


saulecabrera (approved):
Looks good to me, thanks! FWIW (as discussed offline) as part of my work on multi-value I'm changing the push/pop behaviour to be type-sized. But this change is good enough to address the bug.

PR #7497, opened 2023-11-08
crates/cache: print warning information only when the renaming truly fails

### LINKED ISSUE #7496: Question about creating module cache files
#### Feature

When I'm using the cache, I always encounter the following warning. Initially, I thought it was a failure to write the cache, but upon closer inspection, I found that the cache file was actually created successfully.

```
Failed to write file with rename,
lock path: /Users/xxx/Library/Caches/BytecodeAlliance.wasmtime/modules/wasmtime--1699432190543/KhAxrGdzhkuy9f-lsCzlq73dYFuGl3yOqWQyOpL8cBY.wip-atomic-write-mod,
target path: /Users/xxx/Library/Caches/BytecodeAlliance.wasmtime/modules/wasmtime--1699432190543/KhAxrGdzhkuy9f-lsCzlq73dYFuGl3yOqWQyOpL8cBY,
err: No such file or directory (os error 2)
```

So, I delved into the code and have some questions about the handling of writing cache files. According to the current implementation, every time a new module is written to the cache will inevitably trigger this warning because the directory doesn't exist. I looked at the comments in the code and found that the intention here is to optimize system calls.

```rust
// Optimize syscalls: first, try writing to disk. It should succeed in most cases.
// Otherwise, try creating the cache directory and retry writing to the file.
```

However, facing this warning every time we run a new module is quite confusing, I'm wondering if we can modify the implementation here.

#### Implementation

Check if the parent directory exists before writing the file. If it doesn't exist, create the directory first and then proceed with writing the file.

I'm not entirely sure about the performance impact of checking the path's existence every time. If it's feasible, I can go ahead and implement it this way.

#### Benefit

We won't encounter the warning anymore, and the code will be a bit more concise.

iawia002:
fix #7496

cc @bjorn3

### DIFF
--- a/crates/cache/src/lib.rs
+++ b/crates/cache/src/lib.rs
@@ -2,11 +2,11 @@ use base64::Engine;
 use log::{debug, trace, warn};
 use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
-use std::fs;
 use std::hash::Hash;
 use std::hash::Hasher;
 use std::io::Write;
 use std::path::{Path, PathBuf};
+use std::{fs, io};
 
 #[macro_use] // for tests
 mod config;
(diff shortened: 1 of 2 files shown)


alexcrichton (approved):
Thanks!

PR #7499, opened 2023-11-08
winch: Properly handle unconditional jumps

saulecabrera:
This commit improves unconditional jumps by balancing the stack pointer as well as the value stack when the current stack pointer and value stack are greater than the target stack pointer and value stack. The invariant that this changes maintains is that the the value stack should always reflect the the state of the machine stack. The value stack might have excess stack values in a presence of a fallthrough (`br_if` or `br_table`) in which the target branch is not known at compile time; in this situation instructions like `return` or `br` discard any excess values.

The fuzzer discovered the following test case, which triggered several assertions related to unconditional jump handling:

```wat
(module
  (func (;0;) (result i32)
    (local i32)
    local.get 0
    loop ;; label = @1
      local.get 0
      block ;; label = @2
      end
      br 0 (;@1;)
    end
  )
  (export "" (func 0))
)
```

### DIFF
--- a/winch/codegen/src/codegen/context.rs
+++ b/winch/codegen/src/codegen/context.rs
@@ -1,6 +1,6 @@
 use wasmtime_environ::{VMOffsets, WasmHeapType, WasmType};
 
-use super::ControlStackFrame;
+use super::{CodeGen, ControlStackFrame};
 use crate::{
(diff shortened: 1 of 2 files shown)


elliottt (approved):
Looks good to me!

PR #7501, opened 2023-11-08
winch: Properly derive a scratch register for arg assignment

saulecabrera:
This commit properly derives a scratch register for a particular WebAssembly type. The included spec test uncovered that the previous implementation used a int scratch register to assign float stack arguments, which resulted in a panic.


fitzgen (approved):
Nice!

Issue #7500, opened 2023-11-08
ISLE: make multi-constructors more efficient


fitzgen:
* [ ] Instead of having the generated code construct and return an iterator on every call, callers should pass in a reusable vec that gets pushed onto. We use smallvecs right now, which is better than nothing, but once we exceed the inline capacity, we start making lots of temporary allocations.

* [ ] We limit the number of rewritten values added to the egraph, and only consider the first 5 values returned by ISLE code:

  https://github.com/bytecodealliance/wasmtime/blob/b9f2a3060b95a479a5e52ad410f77d43bf1a2a59/cranelift/codegen/src/egraph.rs#L249-L260

  But the ISLE code still builds up an unbounded number of rewritten values to return. Ideally our rule set would never exhibit this behavior, but it does happen. Since we only take the first N anyways, the generated ISLE code should check if we've reached our max limit (configured at ISLE compile time? or maybe via the context?) and early return instead of spend time matching rewrites whose results will be ignored.

PR #7503, opened 2023-11-08
ISLE: Allow callers to pass in storage for multi returns

fitzgen:
This allows callers to reuse storage and avoid allocations.

Internal constructors calling internal constructors won't reuse storage (at
least for now) but it is an option for the user code that is calling into ISLE.

Part of #7500

### DIFF
--- a/cranelift/isle/isle/src/codegen.rs
+++ b/cranelift/isle/isle/src/codegen.rs
@@ -154,7 +154,7 @@ impl<'a> Codegen<'a> {
         if sig.ret_kind == ReturnKind::Iterator {
             writeln!(
                 code,
-                "{indent}type {name}_iter: ContextIter<Context = Self, Output = {output}>;",
+                "{indent}type {name}_returns: Default + IntoContextIter<Context = Self, Output = {output}>;",
                 indent = indent,
                 name = sig.func_name,
                 output = ret_tuple,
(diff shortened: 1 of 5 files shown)


fitzgen:
FWIW, no statistically significant changes to performance measured in sightglass. Still good to do this though, IMO.

cranelift/isle/isle/src/codegen.rs
```diff
@@ -534,15 +626,16 @@ impl<'a> Codegen<'a> {
                     match ret_kind {
                         ReturnKind::Plain => write!(ctx.out, "return ")?,
                         ReturnKind::Option => write!(ctx.out, "return Some(")?,
-                        ReturnKind::Iterator => write!(ctx.out, "returns.push(")?,
+                        ReturnKind::Iterator => write!(ctx.out, "returns.extend(Some(")?,
```
elliottt, line 629: Is the choice of `Option` for the iterator here significant, or would `std::iter::once` also work?

elliottt (approved):
Nice!

cranelift/isle/isle/src/codegen.rs
fitzgen: `std::iter::once` would also work, no strong feelings either way.

PR #7506, opened 2023-11-08
iOS target support

jmp-0x7C0:
As discussed with @alexcrichton on zulip, there appears to be some interest in supporting iOS as a platform. This PR makes the necessary changes for `wasmtime` to build when targeting `aarch64-apple-ios`. 

A few notes about this PR:
* Although it is now possible to compile `wasmtime` for the iOS target, being able to run `*.cwasm` modules is not yet possible. From what @alexcrichton explained to me it appears there would be some additional work would be needed
to postprocess a `*.cwasm` file into a linkable object files before this can be used in an iOS application. I would be interested in taking on this future work however my understanding of what this would involve is still very limited and I would need help sketching this out in detail.
* I haven't yet tested anything beyond making sure that the cargo check passes. If there are any suggestions of what might be useful to test at this stage before merging this I'd be happy to add in CI tests or do any manual testing deemed appropriate.
* With limited knowledge about the differences between iOS and macOS I am assuming that macOS is identical to iOS for the purposes of this PR, e.g. I have no idea if the whole mach ports vs signals makes sense for iOS.  
* I made the minimal set of changes to get the code compiling, before merging the PR it might make sense to rename certain things e.g. it might make sense to rename the `macos` module in the runtime crate to `apple` or similar. 

As this is the first time contributing, please let me know if this PR makes sense or would need any changes to align with the contributing guidelines for this project.

### DIFF
--- a/cranelift/native/src/lib.rs
+++ b/cranelift/native/src/lib.rs
@@ -115,7 +115,7 @@ pub fn infer_native_flags(isa_builder: &mut dyn Configurable) -> Result<(), &'st
             isa_builder.enable("has_pauth").unwrap();
         }
 
-        if cfg!(target_os = "macos") {
+        if cfg!(any(target_os = "macos", target_os = "ios")) {
             // Pointer authentication is always available on Apple Silicon.
             isa_builder.enable("sign_return_address").unwrap();
             // macOS enforces the use of the B key for return addresses.
@@ -167,7 +167,10 @@ mod tests {
                 .finish(settings::Flags::new(flag_builder))
                 .unwrap();
 
-            if cfg!(all(target_os = "macos", target_arch = "aarch64")) {
+            if cfg!(all(
+                any(target_os = "macos", target_os = "ios"),
+                target_arch = "aarch64"
+            )) {
                 assert_eq!(isa.default_call_conv(), CallConv::AppleAarch64);
             } else if cfg!(any(unix, target_os = "nebulet")) {
                 assert_eq!(isa.default_call_conv(), CallConv::SystemV);
--- a/crates/fiber/src/unix/aarch64.rs
+++ b/crates/fiber/src/unix/aarch64.rs
@@ -22,7 +22,7 @@ use super::wasmtime_fiber_start;
 use wasmtime_asm_macros::asm_func;
 
 cfg_if::cfg_if! {
-    if #[cfg(target_os = "macos")] {
+    if #[cfg(any(target_os = "macos", target_os = "ios"))] {
         macro_rules! paci1716 { () => ("pacib1716\n"); }
         macro_rules! pacisp { () => ("pacibsp\n"); }
         macro_rules! autisp { () => ("autibsp\n"); }
--- a/crates/runtime/src/traphandlers/unix.rs
+++ b/crates/runtime/src/traphandlers/unix.rs
@@ -56,7 +56,7 @@ pub unsafe fn platform_init() {
 
     // Sometimes we need to handle SIGBUS too:
     // - On Darwin, guard page accesses are raised as SIGBUS.
-    if cfg!(target_os = "macos") || cfg!(target_os = "freebsd") {
+    if cfg!(any(target_os = "macos", target_os = "ios")) || cfg!(target_os = "freebsd") {
         register(&mut PREV_SIGBUS, libc::SIGBUS);
     }
 
@@ -138,7 +138,7 @@ unsafe extern "C" fn trap_handler(
         // done running" which will clear the sigaltstack flag and allow
         // reusing it for the next signal. Then upon resuming in our custom
         // code we blow away the stack anyway with a longjmp.
-        if cfg!(target_os = "macos") {
+        if cfg!(any(target_os = "macos", target_os = "ios")) {
             unsafe extern "C" fn wasmtime_longjmp_shim(jmp_buf: *const u8) {
                 wasmtime_longjmp(jmp_buf)
             }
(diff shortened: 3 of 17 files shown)


cranelift/native/src/lib.rs
```diff
@@ -115,7 +115,7 @@ pub fn infer_native_flags(isa_builder: &mut dyn Configurable) -> Result<(), &'st
             isa_builder.enable("has_pauth").unwrap();
         }
 
-        if cfg!(target_os = "macos") {
+        if cfg!(any(target_os = "macos", target_os = "ios")) {
```
alexcrichton, line 118: This is something that I think may need to be confirmed. The comment below for example is incorrect since not all iOS devices will run on Apple Silicon, but I'm sure all recent ones do. That being said I don't know the story for return-address-signing on iOS and whether or not it's different than macOS.

cranelift/native/src/lib.rs
```diff
@@ -167,7 +167,10 @@ mod tests {
                 .finish(settings::Flags::new(flag_builder))
                 .unwrap();
 
-            if cfg!(all(target_os = "macos", target_arch = "aarch64")) {
+            if cfg!(all(
+                any(target_os = "macos", target_os = "ios"),
+                target_arch = "aarch64"
+            )) {
```
alexcrichton, line 173: This is also something to watch out for once you're at the point you can run some code. This block is true for macOS but iOS may very well be different in its calling convention. I'm not sure either way (or where to look it up), so it's mostly something to watch out for to see if apps crash when you load them.

crates/fiber/src/unix/aarch64.rs
```diff
@@ -22,7 +22,7 @@ use super::wasmtime_fiber_start;
 use wasmtime_asm_macros::asm_func;
 
 cfg_if::cfg_if! {
-    if #[cfg(target_os = "macos")] {
+    if #[cfg(any(target_os = "macos", target_os = "ios"))] {
```
alexcrichton, line 25: This relates to my comment about pointer authentication above, because if iOS uses the A key instead of the B key then this block would need to change.

crates/runtime/src/traphandlers/unix.rs
```diff
@@ -178,7 +178,7 @@ unsafe extern "C" fn trap_handler(
 // more accurate definition. When that PR and/or issue is resolved then this
 // should no longer be necessary.
 #[repr(C)]
-#[cfg(all(target_arch = "aarch64", target_os = "macos"))]
+#[cfg(all(target_arch = "aarch64", any(target_os = "macos", target_os = "ios")))]
```
alexcrichton, line 181: If possible you'll want to use `libc::ucontext_t` instead of this. This structure is one that has a high risk of being different between iOS and macOS. I posted https://github.com/bytecodealliance/wasmtime/pull/7507 to remove this structure definition entirely as it's not longer necessary (the PR referenced in the above comment has landed and has been published). I think you should be able to use `libc::ucontext_t` for iOS, but that depends on the `libc` crate which I can't say with certainty.

tests/all/custom_signal_handler.rs
```diff
@@ -1,6 +1,9 @@
 #[cfg(any(
     target_os = "linux",
-    all(target_os = "macos", feature = "posix-signals-on-macos")
+    all(
+        any(target_os = "macos", target_os = "ios"),
+        feature = "posix-signals-on-macos"
+    )
 ))]
 #[cfg(not(miri))]
```
alexcrichton, line 8: As a heads up I've posted https://github.com/bytecodealliance/wasmtime/pull/7509 which will conflict with this, although what you're doing here of "where macos is needed also allow ios" would also apply to that PR.

tests/all/debug/lldb.rs
```diff
@@ -12,7 +12,7 @@ fn lldb_with_script(args: &[&str], script: &str) -> Result<String> {
     let mut cmd = Command::new(&lldb_path);
 
     cmd.arg("--batch");
-    if cfg!(target_os = "macos") {
+    if cfg!(any(target_os = "macos", target_os = "ios")) {
```
alexcrichton, line 15: I think it's ok to skip iOS in this file, it's unlikely these tests will ever complete successfully on iOS (and they're disabled by default too)

alexcrichton: (ditto for all other `tests/all/debug/*.rs` files)

alexcrichton (commented):
Thanks for this! If you're ok with it though because this is a pretty small patch I'd personally recommend not merging this just yet and moving on to the next steps of trying to get code running (if you're interested in doing those next steps). I'm a bit worried about how there's a number of open questions and I find it easier to remember the questions if they're highlighted in a diff here rather than hunting around a preexisting codebase. 

If you'd prefer though I also think it'd be reasonable to land ahead of time.

One suggestion I'd have is that you can probably void updating any tests at this time. I think we're probably still aways out from running tests on iOS so I think it'd be ok to cross that bridge when we get there.

cranelift/native/src/lib.rs
turbolent: From what I can see from https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms, it doesn't look like macOS and iOS have different calling conventions on arm64.

crates/fiber/src/unix/aarch64.rs
turbolent: Looking at https://support.apple.com/en-ng/guide/security/sec8b776536b/1/web/1#sec0167b469d, there does not seem to be any indication there are any differences across iOS and macOS regarding which key is used, it always seems to be IB for return addresses.

turbolent:
@alexcrichton @jmp-0x7C0 I just wanted to report that making similar changes on `main`, i.e., extending cases of `target_os = "macos"` with `target_os = "ios"`, let me run the GCD example in an iPhone 15 simulator and on a real iPhone 11 Pro.

Even if some details, like trap and signal handling, might still need to be fully tested, it would be great to allow building wasmtime for iOS upstream. I'm happy to open a PR on top of `main`

alexcrichton:
@turbolent oh that's awesome, thanks for testing! If you'd like I think it would be reasonable to open an PR against `main`, likely starting from this PR to retain the original authorship and then adding bits and pieces on top as you've tested them.

PR #7507, opened 2023-11-08
Remove workaround of upstream libc issue

alexcrichton:
The referenced PR has long since landed and been published, so remove the FIXME.


pchickey (approved):
Nice

PR #7510, opened 2023-11-08
wasi-filesystem: delete access modes

pchickey:
Implements the change for WebAssembly/wasi-filesystem#139: Remove the definition of access modes from the wasi-filesystem interface, the three functions for reading and modifying those modes, and the modes argument from `open_at`.

We never made time to implement these in Wasmtime, so the implementations deleted are all just `todo!()`s. We will add these functions back to wasi-filesystem in a future spec revision and accompany them with an implementation.

### DIFF
--- a/crates/wasi-http/wit/deps/filesystem/types.wit
+++ b/crates/wasi-http/wit/deps/filesystem/types.wit
@@ -142,29 +142,6 @@ interface types {
         truncate,
     }
 
-    /// Permissions mode used by `open-at`, `change-file-permissions-at`, and
-    /// similar.
-    flags modes {
-        /// True if the resource is considered readable by the containing
-        /// filesystem.
-        readable,
-        /// True if the resource is considered writable by the containing
-        /// filesystem.
-        writable,
-        /// True if the resource is considered executable by the containing
-        /// filesystem. This does not apply to directories.
-        executable,
-    }
-
-    /// Access type used by `access-at`.
-    variant access-type {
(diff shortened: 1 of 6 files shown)


crates/wasi-http/wit/deps/filesystem/types.wit
```diff
@@ -538,8 +515,6 @@ interface types {
             open-flags: open-flags,
             /// Flags to use for the resulting descriptor.
             %flags: descriptor-flags,
-            /// Permissions to use when creating a new file.
-            modes: modes
```
alexcrichton, line 542: Question on this, if we know we want to add this in the future should we be trying to future-proof this function? For example we could add an `options: option<open-options>` with a currently empty `resource open-options` or something like that.

Alternatively we could add a feature to the component model and WIT for something like keyword arguments with defaults. That's a much longer- term vision though.

alexcrichton (approved):

crates/wasi-http/wit/deps/filesystem/types.wit
pchickey: I believe that we can eliminate the `modes` parameter here and let open-at continue to do the reasonable default behavior, and then when users of the re-introduced modes stuff want to opt into a non-default behavior for modes, we can expose a new `open-at-modes` function that has the additional argument. Then when it comes time to make a breaking change in preview 3 we can consolidate down to just one `open-at` with mode arguments.

We could get fancier than that, but I expect the above will work fine without introducing much complexity.

PR #7511, opened 2023-11-09
cranelift: Enable fuzzing with PCC enabled

elliottt:
- Configure PCC when fuzzing
- Ensure that we panic for pcc errors in the instantiate fuzz target


cfallin:
hi from parental leave -- quick note while I have a second, and then I'll re-disappear:

Just after getting the individual dynamic memories tests working and merging my last PR, I realized that letting optimization GVN together accesses that hit different bounds-check cases will cause some `conflict` facts, and I think the right solution is to merge range/dynamic-range and mem/dynamic-mem into one thing that can have either static or dynamic or both bounds on each end. You want to be able to say "`x` is less than `0xffff_fffc`, and also `gv2`" basically, and let `Fact::intersect` produce that. Given that, it might make sense to fuzz only static-memory cases with a guard region, which are the default in Wasmtime and are way simpler, and should not hit this case.

Sorry, was hacking on this fix then the kid intervened -- I can get back to it in February :-)

PR #7517, opened 2023-11-09
Get addr of local after popping from reg

jeffcharles:
I observed in Winch that the offset from the stack pointer for locals written to with `local.set` and `local.tee` are slightly wrong when they are writing to a local that had been spilled by a prior operation. The cause appears to be that we get the address of the local (that is, the offset from the stack pointer) before invoking `self.context.pop_to_reg` and then use that address to perform the store operation. `self.context.pop_to_reg` may change the value of the stack pointer if the value on top of the stack is in memory. So the previously calculated stack pointer offset of the local may not be correct when storing a value into that address. This change has us calculate the address of the local (that is, the offset from the stack pointer) after invoking `self.context.pop_to_reg` to avoid that problem.

### DIFF
--- a/winch/codegen/src/codegen/mod.rs
+++ b/winch/codegen/src/codegen/mod.rs
@@ -345,12 +345,14 @@ where
     }
 
     /// Pops the value at the stack top and assigns it to the local at
-    ///
     /// the given index, returning the typed register holding the
     /// source value.
-    pub fn emit_set_local(&mut self, addr: M::Address, size: OperandSize) -> TypedReg {
+    pub fn emit_set_local(&mut self, index: u32) -> TypedReg {
         let src = self.context.pop_to_reg(self.masm, None);
-        self.masm.store(RegImm::reg(src.reg), addr, size);
+        // Need to get address of local after `pop_to_reg` since `pop_to_reg`
+        // may change the local's offset from the SP.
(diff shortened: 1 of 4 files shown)


winch/codegen/src/codegen/mod.rs
```diff
@@ -345,12 +345,14 @@ where
...
     /// Pops the value at the stack top and assigns it to the local at
-    ///
     /// the given index, returning the typed register holding the
     /// source value.
-    pub fn emit_set_local(&mut self, addr: M::Address, size: OperandSize) -> TypedReg {
+    pub fn emit_set_local(&mut self, index: u32) -> TypedReg {
         let src = self.context.pop_to_reg(self.masm, None);
-        self.masm.store(RegImm::reg(src.reg), addr, size);
+        // Need to get address of local after `pop_to_reg` since `pop_to_reg`
+        // may change the local's offset from the SP.
```
saulecabrera, line 353: One minor nit on this comment: it won't change the local's offset, it will change the calculation of that offset from the current position of the stack pointer. I'd suggest rewording this a bit to something like: _Need to get the local address before popping the value given that `pop_to_reg` will pop the machine stack, causing a wrong address to be calculated_

saulecabrera (approved):
Thanks for finding this, I left one minor nit on wording, but else this looks good to me.

Issue #7481, opened 2023-11-04
riscv64: Use `vadd.vi` when subtracting an immediate vector


afonso360:
👋 Hey,

The RISC-V vector backend has no `vsub.vi` instruction. That instruction would subtract a vector register with an immediate. It does have a `vrsub.vi` that computes `imm - vector`.

Instead of emitting a `vsub.vi` we can negate the immediate and emit a `vadd.vi` that does exist.

#### Example testcase

Here's an example test case that currently does not compile to an optimal instruction sequence:
```
function %isub_imm(i64x2) -> i64x2 {
block0(v0: i64x2):
    v1 = vconst.i64x2 [1 1]
    v2 = isub v0, v1
    return v2
}
```

This emits:
```
  vle64.v v13,[const(0)] #avl=2, #vtype=(e64, m1, ta, ma)
  vsub.vv v13,v9,v13 #avl=2, #vtype=(e64, m1, ta, ma)
```

It loads the constant and then emits the `.vv` version of the instruction. But we could just emit `vadd.vi v13, v9, -1`.

#### Implementation

Recognizing constants is currently done using the `replicated_imm5` constructor. We probably need to build an equivalent constructor that matches with an equivalent negated Imm5

Afterwards we should only need to add a rule that matches an isub with an imm5 on the right hand side. We should only need one since for `sub`s with constants in the left hand side, we already have a `vrsub` instruction.

I've recently implemented a similar set of rules for scalar `sub` in https://github.com/bytecodealliance/wasmtime/pull/7480/commits/6bc69c095a1ada4b4ee7190baaf45aaf7d85c5f2.

BieVic:
I would like to do it, but I can wait in case someone else wants to get started with Cranelift :)

pyroMechanical:
I'd like to take this on as a first project! I have prior rust and compiler experience, but haven't worked with cranelift before.

BieVic:
Awesome! Let me or @afonso360 know if you need any help

BieVic:
For testing you can use clif-util (e.g. wasmtime/target/debug/clif-util), with which you can also target different ISAs. I'm also fairly new and this [doc](https://github.com/bytecodealliance/wasmtime/blob/main/cranelift/docs/testing.md) about testing helped me wrap my head around

PR #7522, opened 2023-11-10
riscv64: use `vadd.vi` for small immediate vector subtraction

pyroMechanical:
this PR implements the change requested in #7481. Since riscv64 doesn't have a `vsub.vi` instruction, this optimization removes a load when the immediate fits in an Imm5.

### DIFF
--- a/cranelift/codegen/src/isa/riscv64/inst.isle
+++ b/cranelift/codegen/src/isa/riscv64/inst.isle
@@ -1937,8 +1937,8 @@
 ;; Like imm5_from_value, but first negates the `Value`.
 (decl pure partial imm5_from_negated_value (Value) Imm5)
 (rule (imm5_from_negated_value (has_type ty (iconst n)))
-(if-let (imm5_from_i64 imm) (i64_neg (i64_sextend_imm64 ty n)))
-imm)
+  (if-let (imm5_from_i64 imm) (i64_neg (i64_sextend_imm64 ty n)))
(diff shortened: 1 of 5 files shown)


cranelift/codegen/src/isa/riscv64/inst.isle
```diff
@@ -1934,6 +1934,12 @@
 (decl imm5_from_value (Imm5) Value)
 (extractor (imm5_from_value n) (i64_from_iconst (imm5_from_i64 n)))
 
+;; Like imm5_from_value, but first negates the `Value`.
+(decl pure partial imm5_from_negated_value (Value) Imm5)
+(rule (imm5_from_negated_value (has_type ty (iconst n)))
+(if-let (imm5_from_i64 imm) (i64_neg (i64_sextend_imm64 ty n)))
+imm)
```
elliottt, line 1941:
```suggestion
;; Like imm5_from_value, but first negates the `Value`.
(decl pure partial imm5_from_negated_value (Value) Imm5)
(rule (imm5_from_negated_value (has_type ty (iconst n)))
  (if-let (imm5_from_i64 imm) (i64_neg (i64_sextend_imm64 ty n)))
  imm)
```

elliottt (commented):
This looks good to me! @afonso360, would you mind taking a look as well?

afonso360 (approved):
This looks great! Thanks! 🎉

PR #7523, opened 2023-11-10
cranelift-entity: Add `get_many_mut` method

simvux:
`PrimaryMap` does not expose an `as_slice_mut` method since that bypasses the need for the key. However; this also prevents you from being able to access multiple indices at once safely via `slice::split_mut()` or the unstable `slice::get_many_mut`. 

This PR adds a version of `get_many_mut` which takes `[K; N]` instead of slice's `[usize; N]`


fitzgen (approved):
LGTM, thanks!

PR #7526, opened 2023-11-10
ISLE: Build in maximum limit for multi constructor returns

### LINKED ISSUE #7500: ISLE: make multi-constructors more efficient

fitzgen:
Allows users to avoid runaway rules that match too much stuff.

No statistically significant speed up to compilation on sightglass, but good to have as a safeguard anyways.

Fixes #7500

### DIFF
--- a/cranelift/codegen/src/opts.rs
+++ b/cranelift/codegen/src/opts.rs
@@ -23,7 +23,16 @@ pub type Range = (usize, usize);
 pub type ValueArray2 = [Value; 2];
 pub type ValueArray3 = [Value; 3];
 
-pub type ConstructorVec<T> = SmallVec<[T; 8]>;
+const MAX_ISLE_RETURNS: usize = 8;
+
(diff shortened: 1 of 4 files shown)


elliottt (approved):
Looks good to me!

PR #7527, opened 2023-11-10
wasi-http: Validate Content-Length when present

elliottt:
Make use of immutable header state to track how much has been written to an output-body stream, ensuring that we write exactly as much as was mentioned in the `Content-Length` header.

### DIFF
--- a/crates/wasi-http/src/body.rs
+++ b/crates/wasi-http/src/body.rs
@@ -482,8 +482,6 @@ impl HostOutgoingBody {
             .take()
             .expect("outgoing-body trailer_sender consumed by a non-owning function");
 
-        // TODO: does sending guarantee that the message was received, or could it be queued to
-        // receive?
         if let Some(w) = self.written {
             use std::cmp::Ordering;
             let res = w.finish();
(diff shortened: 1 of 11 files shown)


pchickey (approved):

crates/wasi-http/src/body.rs
```diff
@@ -423,42 +454,115 @@ impl HostOutgoingBody {
...
+        // will immediately pick up the finish sender.
+        drop(self.body_output_stream);
+
+        let sender = self
+            .finish_sender
+            .take()
+            .expect("outgoing-body trailer_sender consumed by a non-owning function");
+
+        // TODO: does sending guarantee that the message was received, or could it be queued to
+        // receive?
```
alexcrichton, line 482: I'm not 100% sure as to the context of this comment but sending on a oneshot does not guarantee the message is received, it's only queued up at that time to possible be received in the future.

elliottt: I was wondering if we would potentially block on the oneshot send if there was nothing already waiting to receive it. It sounds like that's not the case, so I'll remove the TODO. Thanks!

PR #7531, opened 2023-11-13
winch: Materialize latent locals when setting them

jeffcharles:
Winch has a bug where when setting or teeing a local, and that local is in the value stack but not at the top of the value stack, the local on the stack will pop with an incorrect value because the value of the local is determined lazily (that is, when it's needed). To address that, we can detect if the value stack contains a reference to the local that is not the top-most element, and spill registers and locals into memory if it does, which will materialize the value in the local before the value of the local is changed.

### DIFF
--- a/winch/codegen/src/codegen/mod.rs
+++ b/winch/codegen/src/codegen/mod.rs
@@ -348,6 +348,17 @@ where
     /// the given index, returning the typed register holding the
     /// source value.
     pub fn emit_set_local(&mut self, index: u32) -> TypedReg {
+        // Materialize any references to the same local index that are in the
+        // value stack by spilling.
+        let stack_contains_latent_local = self
+            .context
+            .stack
+            .iter()
+            .skip(1) // top-most element is popped below.
+            .any(|v| v.is_local_at_index(index));
+        if stack_contains_latent_local {
+            self.context.spill(self.masm);
+        }
         let src = self.context.pop_to_reg(self.masm, None);
         // Need to get address of local after `pop_to_reg` since `pop_to_reg`
         // will pop the machine stack causing an incorrect address to be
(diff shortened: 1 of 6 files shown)


winch/codegen/src/masm.rs
```diff
@@ -292,7 +292,7 @@ pub enum RoundingMode {
 
 pub(crate) trait MacroAssembler {
     /// The addressing mode.
-    type Address: Copy;
+    type Address: Copy + Debug;
```
jeffcharles, line 295: I added this because it made some of my debugging a little easier and all the address implementations implement `Debug`.

winch/codegen/src/stack.rs
```diff
@@ -275,6 +283,11 @@ impl Stack {
         self.inner.range(partition..)
     }
 
+    /// Returns an iterator of the stack in top-most to bottom-most order.
+    pub fn iter(&self) -> impl Iterator<Item = &Val> + '_ {
```
jeffcharles, line 287: I looked at using `peekn` but the iterator it returns is in bottom-most to top-most order and the return type is not a trait that allows me to reverse the order of the entries so I can ignore the top-most element of the stack.

winch/codegen/src/codegen/mod.rs
```diff
@@ -348,6 +348,17 @@ where
...
     /// source value.
     pub fn emit_set_local(&mut self, index: u32) -> TypedReg {
+        // Materialize any references to the same local index that are in the
+        // value stack by spilling.
+        let stack_contains_latent_local = self
+            .context
+            .stack
+            .iter()
+            .skip(1) // top-most element is popped below.
+            .any(|v| v.is_local_at_index(index));
```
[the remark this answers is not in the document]
jeffcharles: We talked offline.

We can use an `iter().rev().take_while(|v| /* v is not mem or a const */)` approach to avoid iterating over the entire stack. And the stack method I'm adding will be named in such a way that it indicates it does not check the top element. This is because we can safely avoid spilling if it's only the top element that references the local since we're going to pop the top element immediately after performing the check to see if the stack contains the local.

saulecabrera (approved):

PR #7434, opened 2023-11-01
wasi-http: Migrate to more descriptive error variant

elliottt:
Integrate the more descriptive error variant present in https://github.com/WebAssembly/wasi-http/pull/52. This PR attempts to port the existing errors we raise directly to the new variant.

Fixes #7248

PR #7534, opened 2023-11-14
wasi-http: Implement http-error-code, and centralize error conversions

elliottt:
Implement the body of `http-error-code`, and add a test that shows its use when writing too much. This functionality was missing from #7434.

Additionally, add some functions in wasi-http for converting from hyper::Error and http::Error to the new `ErrorCode` variant. These functions log the error received as a warning, and return the translated error to the guest.

### DIFF
--- a/crates/wasi-http/src/lib.rs
+++ b/crates/wasi-http/src/lib.rs
@@ -65,18 +65,16 @@ pub fn http_request_error(err: http::Error) -> bindings::http::types::ErrorCode 
 /// Translate a [`hyper::Error`] to a wasi-http `ErrorCode` in the context of a request.
 pub fn hyper_request_error(err: hyper::Error) -> bindings::http::types::ErrorCode {
     use bindings::http::types::ErrorCode;
+    use std::error::Error;
 
-    // If it's a user error, we might be able to extract a wasi-http error from it.
-    if err.is_user() {
-        if let Some(cause) = err.into_cause() {
-            match cause.downcast::<ErrorCode>() {
-                Ok(err) => return *err.clone(),
-                Err(cause) => tracing::warn!("hyper request user error: {cause}"),
-            }
+    // If there's a source, we might be able to extract a wasi-http error from it.
+    if let Some(cause) = err.source() {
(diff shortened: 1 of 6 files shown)


crates/wasi-http/src/lib.rs
```diff
@@ -47,3 +47,52 @@ pub(crate) fn dns_error(rcode: String, info_code: u16) -> bindings::http::types:
 pub(crate) fn internal_error(msg: String) -> bindings::http::types::ErrorCode {
     bindings::http::types::ErrorCode::InternalError(Some(msg))
 }
+
+/// Translate a [`http::Error`] to a wasi-http `ErrorCode` in the context of a request.
+pub fn http_request_error(err: http::Error) -> bindings::http::types::ErrorCode {
+    use bindings::http::types::ErrorCode;
+
+    tracing::warn!("http request error: {err:?}");
```
alexcrichton, line 55: I realize that this matches what I was talking about yesterday, where if an error is dropped on the floor it's always at least logged, but I will also say that there's quite a lot of logging going on here. In theory almost all error context should be made available to end users, but it seems like almost all errors are being dropped and canonicalized in bland error codes. 

There's not really anything that can be done about that in this PR specifically, but it might be something good to keep in mind for future iterations of wasi-http. One possible example might be something like a "last error as debug string" API or something like that. Or alternatively using `wasi:io/error.{error}` for all http APIs and then having an optional "pull out the http-specific error kind from each error" API sort of like filesystems. Too late to do this for preview2 but perhaps later.

alexcrichton (approved):

crates/wasi-http/src/lib.rs
elliottt: This was going to be a bit noisy, so I've reworked it so that we only log on a path where there's not clear translation back to a wasi-http error (the fall-through cases). Looking more towards the future, I wonder if it would be better to return a `wasi:io/error/error` instead of a `wasi:http/types/error-code`. @pchickey and I started working on this last week, but held off on making this change as it was not as easy as swapping out uses of `error-code`.

alexcrichton:
@elliottt is this something that should be backported to Wasmtime 15?

PR #7535, opened 2023-11-14
winch: Multi-Value Part 1 

saulecabrera:
This change adds initial support for the Multi-Value proposal to Winch, focusing on function calls, support for Multi-Value for blocks will be added on top of this change in a different pull request.

This change is divided into two main parts:

* Infrastructure changes needed to support Multi-Value
* Handling Multi-Value returns for function calls

#### Infrastructure changes needed to support Multi-Value

Some notable changes were needed in order to support this proposal, namely:

* Introducing the notion of `ABIOperands` and rewriting the `ABIParams` and `ABIResults` in terms of `ABIOperands`.
* In Winch's default calling convention, results are stored in reverse order: if a function produces multiple results, the first result is stored in a memory location and the last result is stored in a register; given that function results are captured in the value stack after a function call, this approach makes it trivial to maintain the following value stack invariants:
  * Spilled memory values always precede register values
  * Spilled values are stored from oldest to newest, matching their respective locations on the machine stack.
* Introducing type-sized spills: Prior to this change, only floats produced type-sized spills, this created an inconsistency on how slots for results are handled and made it harder to manage the stack for multiple-results; this change introduces type-sized spills for integer and floats, making the spilling strategy uniform across the compiler. This might introduce some overhead, since this change is now, in the 32-bit case, introducing two instructions were previously there only used to be one (e.g. `push` vs `sub + mov`). My intention is to profile, and try to improve this strategy if needed, later on. 

#### Handling Multi-Value returns for function calls

The approach for handling multiple results for function calls involves:

Adding an implicit extra parameter to the `ABIParams` definition, when a function returns multiple values; from the callee's perspective, this parameter is treated like any other "special" parameter (e.g `VMContext`), to which, we give it a well know `LocalSlot` for later referencing when storing the return values. From the caller's perspective, extra stack space is created before the call, and the address to that space is passed in to the callee, via the return pointer parameter and once the function returns, the values in the created space are captured as values in the value stack.


winch/codegen/src/abi/mod.rs
```diff
@@ -118,178 +125,310 @@ pub(crate) trait ABI {
...
+
+    /// Returns the size in bytes of the given [`WasmType`].
+    fn sizeof(ty: &WasmType) -> u32 {
+        match ty {
+            WasmType::Ref(rt) => match rt.heap_type {
+                WasmHeapType::Func => Self::word_bytes(),
+                ht => unimplemented!("Support for WasmHeapType: {ht}"),
+            },
+            WasmType::F64 | WasmType::I64 => Self::word_bytes(),
+            WasmType::F32 | WasmType::I32 => Self::word_bytes() / 2,
```
fitzgen, line 140: This seems like it is unnecessarily setting us up for porting issues in the future when we have a 32-bit target. Can we just put the size directly or use `mem::size_of::<u32>()` or something instead?

winch/codegen/src/abi/mod.rs
```diff
@@ -118,178 +125,310 @@ pub(crate) trait ABI {
...
+pub(crate) struct ABIResults {
+    /// The result operands.
+    operands: ABIOperands,
+}
+
+impl ABIResults {
+    /// Creates [`ABIResults`] from a slice of `WasmType`.
+    pub fn from<F>(returns: &[WasmType], mut map: F) -> Self
+    where
+        F: FnMut(&WasmType, u32) -> (ABIOperand, u32),
```
fitzgen: Could we document what the `map` closure is supposed to do here? Was not immediately obvious to me when reading the code.

winch/codegen/src/abi/mod.rs
```diff
@@ -118,178 +125,310 @@ pub(crate) trait ABI {
...
     }
 
-    /// Get the result reg.
-    pub fn result_reg(&self) -> Option<Reg> {
-        match self {
-            Self::Reg { reg, .. } => Some(*reg),
-            _ => None,
-        }
+    /// Returns an iterator over the result registers.
+    pub fn regs(&self) -> &HashSet<Reg> {
```
fitzgen: Documentation is out of sync with the code: it returns a set not an iterator.

winch/codegen/src/abi/mod.rs
```diff
@@ -118,178 +125,310 @@ pub(crate) trait ABI {
...
+    /// have a single return.
     ///
-    /// NOTE: Currently only one or zero registers
-    /// will be returned until suport for multi-value is introduced.
-    pub fn regs(&self) -> impl Iterator<Item = Reg> + '_ {
-        std::iter::once(self.result_reg()).filter_map(|v| v)
+    /// # Panics
+    /// This function panics if the function signature contains more
+    pub fn unwrap_singleton(&self) -> &ABIOperand {
+        assert!(self.len() == 1);
```
fitzgen, line 390: `debug_assert_eq!` will give better diagnostics on panic.

winch/codegen/src/abi/mod.rs
```diff
@@ -118,178 +125,310 @@ pub(crate) trait ABI {
...
+pub(crate) struct ABIParams {
+    /// The param operands.
+    operands: ABIOperands,
+}
+
+impl ABIParams {
+    /// Creates [`ABIParams`] from a slice of `WasmType`.
+    pub fn from<F>(params: &[WasmType], initial_bytes: u32, mut map: F) -> Self
+    where
+        F: FnMut(&WasmType, u32) -> (ABIOperand, u32),
```
fitzgen: Ditto regarding `map`

winch/filetests/filetests/x64/br/as_call_all.wat
```diff
@@ -11,11 +11,11 @@
...
-;;   14:	 4c89742404           	mov	qword ptr [rsp + 4], r14
-;;   19:	 b8ffffffff           	mov	eax, 0xffffffff
-;;   1e:	 4883c418             	add	rsp, 0x18
-;;   22:	 5d                   	pop	rbp
-;;   23:	 c3                   	ret	
+;;   14:	 4c893424             	mov	qword ptr [rsp], r14
+;;   18:	 b8ffffffff           	mov	eax, 0xffffffff
+;;   1d:	 4883c418             	add	rsp, 0x18
+;;   21:	 5d                   	pop	rbp
+;;   22:	 c3                   	ret
```
fitzgen, line 18: Is this an expected/intended change in this first commit? I thought this was purely a refactoring to prep for multi-value, and I wouldn't expect any functional changes. Certainly not ones where we are changing from where we are pushing values to the stack when I don't see any corresponding changes in how we pop values from the stack.

But maybe this is addressed in follow up commits that I haven't gotten to yet...

winch/codegen/src/abi/local.rs
```diff
@@ -15,7 +15,14 @@ enum Base {
...
 /// Represents the type, location and addressing mode of a local
 /// in the stack's local and argument area.
-#[derive(Clone)]
+// TODO: Add more documentation here around:
+// - Relationship between ABIOperand and LocalSlot
+// - Types of LocalSlots
+// - Invariants around local slots
+// - How do they are different from StackSlots
+// - The perspective they are meant to be for (callee, more than caller),
+//   well actually caller too, but we'll see once we get there.
```
fitzgen: :eyes:

winch/codegen/src/abi/local.rs
```diff
@@ -15,13 +15,17 @@ enum Base {
...
+/// reference by the stack pointer register (SP) or the base pointer register (FP).
+/// * Local slots that are referenced by the stack pointer register are the
+///   function defined locals and the param locals.
+/// * Local slots that represent arguments in the stack, are referenced through the
+///   base pointer register.
+/// A [crate::masm::StackSlot] is a generalized form of a [LocalSlot]: they
+/// represent dynamic chunks of memory that get created throughout the function
+/// compilation lifetime when spilling values (register and locals) into the
+/// machine stack. A [LocalSlot] on the other hand gets created at the beginning
+/// of a function compilation and gets cleaned up at the end.
```
fitzgen, line 28: Ah, great!

fitzgen (approved):
LGTM but I'm still a little sketched out by those disassembly changes where I feel like I'm only seeing one half of a change or something like that. Maybe you can explain what is going on?

winch/codegen/src/abi/mod.rs
saulecabrera: I ended up moving the definition to the implementation of each ABI, that way once it'll be easier to integrate once we support 32-bit targets.

winch/codegen/src/abi/mod.rs
saulecabrera: Fixed!

winch/codegen/src/abi/mod.rs
saulecabrera: Fixed!

winch/codegen/src/abi/mod.rs
saulecabrera: Fixed!

winch/filetests/filetests/x64/br/as_call_all.wat
saulecabrera: It's indeed intended, I just forgot to mention it in the commit message: I realized that the slot for the `VMContext` was not aligned, so I introduced alignment for it, similar to what I did for the return pointer. 

In the first commit, that is here https://github.com/bytecodealliance/wasmtime/pull/7535/commits/5dcefe38b9dc4cac3f24bf9f4d0eb20f43f9b6a1#diff-b396908c6cc30abb2ad12e93c9443d5b8875dd320980cde6896068f5e5cbf4c1R101-R106; hopefully this clarifies things!

winch/codegen/src/abi/mod.rs
saulecabrera: Fixed

winch/filetests/filetests/x64/br/as_call_all.wat
fitzgen: That makes sense -- thanks for the explanation!

PR #7545, opened 2023-11-15
wasi-http: Populate the authority and scheme in `wasmtime serve`

elliottt:
Ensure that the scheme and authority are populated in the requests given to the incoming handler with `wasmtime serve`. We prefer the `Host` header if it's available, falling back on the existing value of the `authority` in the request. Additionally, we fall back on `http` as the scheme when none was present. This behavior mirrors what https://github.com/fastly/vicroy does for request handling, and ensures that runtimes that require an incoming authority to be set (like the javascript runtime) have one available.

cc @tschneidereit @JakeChampion

### DIFF
--- a/src/commands/serve.rs
+++ b/src/commands/serve.rs
@@ -15,7 +15,8 @@ use wasmtime_wasi::preview2::{
     self, StreamError, StreamResult, Table, WasiCtx, WasiCtxBuilder, WasiView,
 };
 use wasmtime_wasi_http::{
-    body::HyperOutgoingBody, hyper_response_error, WasiHttpCtx, WasiHttpView,
+    bindings::http::types as http_types, body::HyperOutgoingBody, hyper_response_error,
+    WasiHttpCtx, WasiHttpView,
 };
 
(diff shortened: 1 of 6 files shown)


src/commands/serve.rs
```diff
@@ -356,6 +357,37 @@ impl hyper::service::Service<Request> for ProxyHandler {
         // TODO: need to track the join handle, but don't want to block the response on it
         tokio::task::spawn(async move {
             let req_id = inner.next_req_id();
+            let (mut parts, body) = req.into_parts();
+
+            parts.uri = {
+                let uri_parts = parts.uri.into_parts();
+
+                let scheme = uri_parts.scheme.unwrap_or(http::uri::Scheme::HTTP);
```
guybedford, line 365: Why not default to `HTTPS`? Is this also to align with Viceroy?

elliottt: `wasmtime serve` doesn't currently support listening on https, but if that changes in the future we could make it default to what the server's listening on.

pchickey (approved):

PR #7547, opened 2023-11-15
winch: Solidify unreachable code handling

saulecabrera:
This commit solidifies the approach for unreachable code handling in control flow.

Prior to this change, at unconditional jump sites, the compiler would reset the machine stack as well as the value stack. Even though this appoach might seem natural at first, it actually broke several of the invariants that must be met at the end of each contol block, this was specially noticeable with programs that conditionally entered in an unreachable state, like for example

```wat
(module
    local.get 0
    local.get 0
    if (result i32)
      i32.const 1
      return
    else
      i32.const 2
    end
    i32.sub
  )
  (export "main" (func 0))
)
```

The approach followed in this commit ensures that all the invariants are met and introduces more guardrails around those invariants. In short, instead of resetting the value stack at unconditional jump sites, the value stack handling is deferred until the reachability analysis restores the reachability of the code generation process, ensuring that the value stack contains the exact amount of values expected by the frame where reachability is restored. Given that unconditional jumps reset the machine stack, when the reachability of the code generation process is restored, the SP offset is also restored which should match the size of the value stack.

### DIFF
--- a/winch/codegen/src/codegen/context.rs
+++ b/winch/codegen/src/codegen/context.rs
@@ -1,6 +1,6 @@
 use wasmtime_environ::{VMOffsets, WasmHeapType, WasmType};
 
-use super::{CodeGen, ControlStackFrame};
+use super::ControlStackFrame;
 use crate::{
     abi::{ABIOperand, ABIResultsData, RetArea, ABI},
     codegen::BuiltinFunctions,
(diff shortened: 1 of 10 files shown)


winch/codegen/src/codegen/context.rs
```diff
@@ -501,6 +508,43 @@ impl<'a, 'builtins> CodeGenContext<'a, 'builtins> {
         }
     }
 
+    /// Truncates the value stack to the specified target, returning the amount
+    /// of memory entries truncated.
+    /// This function is intended to only be used when restoring the code
+    /// generation's reachability state, when handling an unreachable end or
+    /// else.
+    pub fn truncate_stack_to(&mut self, target: usize) {
```
elliottt, line 516: This is only called from `ensure_stack_state`, is it worth making it not `pub`?

winch/filetests/filetests/x64/unreachable/with_spilled_local.wat
```diff
@@ -19,7 +19,6 @@
 ;;   1a:	 4883ec04             	sub	rsp, 4
 ;;   1e:	 44891c24             	mov	dword ptr [rsp], r11d
 ;;   22:	 0f0b                 	ud2	
-;;   24:	 4883c404             	add	rsp, 4
-;;   28:	 4883c410             	add	rsp, 0x10
-;;   2c:	 5d                   	pop	rbp
-;;   2d:	 c3                   	ret	
+;;   24:	 4883c410             	add	rsp, 0x10
+;;   28:	 5d                   	pop	rbp
+;;   29:	 c3                   	ret
```
elliottt, line 24: I wonder if it would be worth omitting the offsets (except for jump targets) so that the diff would have shown only the `add rsp, 4` instruction going away?

elliottt (approved):
This looks good to me, thank you for all the comments!

winch/codegen/src/codegen/context.rs
saulecabrera: Fixed, thanks!

winch/filetests/filetests/x64/unreachable/with_spilled_local.wat
saulecabrera: Can you clarify a bit what do you mean with this? More concretely, are you referring  to how the disassembly is formatted?

elliottt: The instructions in the original program from  offset 28 on are the same as those in the new program from 24 on, and the only thing that changed was the offsets. If we didn't emit the offsets, or selectively emitted them only for jump targets, the diff would show that the only thing missing was `add rsp, 4`.

saulecabrera: Oh right -- sorry for the confusion, I see it now. I think that's a good idea, it it has the potential to reduce verbosity. Are you good if I create a follow up for this? Because If I'm not wrong, this will change a good chunk of the filetests.

elliottt: Definitely not necessary to fix that in this PR, just something I was thinking would help future reviews :)

PR #7548, opened 2023-11-15
Implement WASI "preview0" in `wasmtime-wasi`

alexcrichton:
This commit implements the `wasi_unstable` module, sometimes referred to as "preview0", in the `wasmtime-wasi` crate. Previously this was only implemented by the `wasi-common` crate but now this is implemented for both meaning that the switch to preview2 won't lose this functionality.

The preview0 WITX files are vendored like the preview1 files and the implementation of preview0 is exclusively implemented by delegating to the preview1 implementation.


pchickey (approved):
`foo.diff` should be deleted, otherwise lgtm

PR #7562, opened 2023-11-20
Update the wasm-tools family of crates

alexcrichton:
Keeping up-to-date, no other pressing reason to update.


pchickey (approved):

alexcrichton:
This is actually going to be significantly more involved than I originally predicted, so will reopen later.

PR #7566, opened 2023-11-21
docs: clarify definition in Introduction

vwkd:
Clarify the definition of the project in the Introduction. A direct "X is a Z" is clearer than a "X is a Y that is a Z" where Y is some superset of Z.


alexcrichton (approved):
Thanks!

Issue #6741, opened 2023-07-17
Redesign Wasmtime's CLI flags


alexcrichton:
Currently the set of CLI flags added to Wasmtime has been ad-hoc and generally "throw another option in there" style historically, which is great for easily adding new options (which happens a fair bit), but I  personally feel like there have been some growing pains that may be worthwhile to address sooner rather than later. 

Some concerns I have with the current CLI are:

* Discoverability is hard right now as `--help` or ` -h` prints out quite a lot of information.
* Not a lot of single-letter flags are in use so CLI commands tend to be quite long
* It's tough to review "at a glance" what Wasmtime's capabilities are on the CLI through the help page since there's so much to sift through
* Some options have little papercuts like trying to remember `--dir` vs `--mapdir`, there's no way to inherit the caller's environment or a single env var from the environment easily, etc.
* Options can be randomly `--disable-foo` or `--enable-foo` and it's not clear why which is which when you're approaching the CLI for the first time.

To be clear none of these are dealbreakers by any means. While I think the Wasmtime CLI is important the embedding API are where things "get serious" so it's ok for the CLI to not be perfect. That being said folks often interact with Wasmtime through the CLI to start off with and it's also mega-useful during development and debugging and such.

So given all the above I'd like to propose a (hopefully) brief bikeshed about the Wasmtime CLI. I'd like to ideally address the above concerns and provide guidelines of how to add new options into the future. This will hopefully clean up our CLI, make it a bit more ergonomic to use, all while not hindering our ability to easily add various knobs here and there as they're implemented.

At a high level my proposal is "let's do what `rustc` does". I'm very biased in that regard as I helped design rustc's CLI as well. Some things I like about rustc though are:

* The output of `rustc -h` fits more-or-less on one terminal. This serves as a good index from which you can learn more.
* There are "option groups" that are behind other flags, such as `rustc -C foo` or `rustc -Z foo`. These single-letter capital flags are easy to type and easy to remember and serve as a good place to house sub-options.
* More information can be read through `rustc -C help` which provides an window into all the codegen capabilities of rustc
* Instead of `--enable-foo` or `--disable-foo` most rustc options are `-C foo` for `--enable-foo` or `-C foo=no` for `--disable-foo`. This makes it easy for callers to forcibly enable/disable something and you don't have to worry about what the defaults are since as a user you typically want to turn something on or off and you don't care too much about whatever the default happens to be.

My proposal here for Wasmtime's option groups are:

[...]

PR #7568, opened 2023-11-21
Add profile flags to cli-flags crate

rahulchaphalkar:
This change adds profile options like `vtune`, `jitdump` and `perfmap` to cli-flags crate, so that they can be used by `Sightglass` and other benchmarks (via `--engine-flags` for sightglass).
This will enable wasm code visibility via profilers like vtune in sightglass, which was lacking.

I have added these flags to the `Debug` group of options - I can create a separate `Profile` group if needed instead.

### DIFF
--- a/crates/cli-flags/src/lib.rs
+++ b/crates/cli-flags/src/lib.rs
@@ -126,6 +126,8 @@ wasmtime_option_group! {
         pub log_to_files: Option<bool>,
         /// Enable coredump generation to this file after a WebAssembly trap.
         pub coredump: Option<String>,
+        /// configure profiler as VTune, JitDump, or PerfMap
+        pub profile: Option<wasmtime::ProfilingStrategy>,
     }
 
     enum Debug {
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! Given though that there's already a `--profile` flag for `wasmtime run`, I think it would be best to reuse that rather than adding a separate version as well. I think that some code in `src/common.rs` might be good to extract to the cli-flags crate perhaps as I think that should have handling of the `--profile` flag (or it can be moved there)

rahulchaphalkar:
Thanks, Alex. Let me take a look at that.

jlb6740:
@rahulchaphalkar +1, this is a needed patch. I think this was even putting profiling under the debug option was even referenced in the CLI refactoring issue https://github.com/bytecodealliance/wasmtime/issues/6741#issuecomment-1639955767. For this patch isn't what you want to do is move the --profile flag to under DebugOptions option_group here https://github.com/bytecodealliance/wasmtime/pull/7568/files#diff-c971b96993f492c9102c58afb35a6cb1ce72351e8c32ffe9ffed91c656dc8484R129-R130
and also add the supporting code for mapping the profile strategy. It looks like you've done that but don't you also need to delete the code in common.rs, such as here: https://github.com/bytecodealliance/wasmtime/blob/main/src/common.rs#L52-L72 
and the implementation code so that the flag is not duplicated? Either that or crates/bench-api/src/lib.rs also parse the none CommonOptions https://github.com/bytecodealliance/wasmtime/blob/main/crates/bench-api/src/lib.rs#L252

@alexcrichton are you suggesting the former, or something different all together?

rahulchaphalkar:
Closing this stale PR, will open a new one if required.

PR #7569, opened 2023-11-21
wasi: test for mtime accuracy to 1ms only

guybedford:
Node.js mtime setting only works up to 1ms accuracy.

This updates the test to work at this level at minimum.

We could make this a configurable test option, I just wasn't sure how to set up such a conditional build feature for the test program - suggestions welcome if not reducing overall accuracy is preferred.


crates/test-programs/src/bin/preview1_fd_filestat_set.rs
```diff
@@ -30,14 +30,16 @@ unsafe fn test_fd_filestat_set(dir_fd: wasi::Fd) {
 
     // Check fd_filestat_set_times
     let old_atim = stat.atim;
-    let new_mtim = stat.mtim - 100;
+    let new_mtim = stat.mtim - 2000;
```
alexcrichton, line 33: With 100us accuracy in Node this gets rounded to no change, right? Could this -2000 be increased so the mtim is changed in node?

crates/test-programs/src/bin/preview1_fd_filestat_set.rs
```diff
@@ -30,14 +30,16 @@ unsafe fn test_fd_filestat_set(dir_fd: wasi::Fd) {
...
         .expect("fd_filestat_set_times");
 
     let stat = wasi::fd_filestat_get(file_fd).expect("failed filestat 3");
     assert_eq!(stat.size, 100, "file size should remain unchanged at 100");
-    assert_eq!(stat.mtim, new_mtim, "mtim should change");
-    assert_eq!(stat.atim, old_atim, "atim should not change");
+
+    // Support accuracy up to at least 1ms
+    assert_eq!(stat.mtim / 1000, new_mtim / 1000, "mtim should change");
+    assert_eq!(stat.atim / 1000, old_atim / 1000, "atim should not change");
```
alexcrichton, line 42: Could these assertions be left as-is with a larger adjustment above that isn't arounded?

pchickey:
The right way to set up a configurable test option is to add an element to `TestConfig`, initialize it from environment variables (so, set by the test runner and resolved at runtime) in `from_env`, and add an accessor method: https://github.com/bytecodealliance/wasmtime/blob/main/crates/test-programs/src/preview1.rs#L117

crates/test-programs/src/bin/preview1_fd_filestat_set.rs
guybedford: 2000 for u64 is 2000 * 100 nanoseconds is 200 microseconds. So I think this is the right accuracy. Reducing the order by one leads to the times being equal in the check. So I think this is right.

crates/test-programs/src/bin/preview1_fd_filestat_set.rs
guybedford: Do you mean does changing `2000` to say `7243` lead to the test still passing? If so, I've verified that is true, yes.

guybedford:
I've added a new `NO_ACCURATE_TIME` configuration option for this now.

guybedford:
Note that the underlying reason for this is the use of IEEE-754 floating point to represent these times in Node.js (some APIs have upgraded to BigInt, but not all).

alexcrichton:
Sorry I'm still pretty confused as to what's going on. 

> 2000 for u64 is 2000 * 100 nanoseconds is 200 microseconds

I don't understand what this means in the context of the structures here. All timestamps in preview1 are in units of nanoseconds, not in units of 100 nanoseconds. That means that 2000 nanoseconds is 2 microseconds, so I don't understand where your 200 number is coming from. Is this perhaps a bug in the Node bits where Node is reporting timestamps in units of 100 nanoseconds? If that's the case then numbers should be multiplied by 100 when making their way into WASI.

> Could these assertions be left as-is with a larger adjustment above that isn't arounded?

There's two axes here sort of I think. On one axis it should probably be mandated that some granularity is required to pass the test suite. The previous granularity was too small for Node so testing a higher granularity seems fine. That would be my suggestion of updating time in an increment of 200us for example as in theory that granularity should be supported. This could go as high as 2 seconds perhaps though.

The other axis is a fine-grained update of time. That part of the test could be disabled on Node. I think that if the test is updated though then it shouldn't test one thing in Node and another everywhere else. Ideally only parts of it are disabled in Node if Node doesn't support finer-grained timestamps.


---

I also don't understand why Node is testing equality using division. Shouldn't the exact values be equal in Node? For example if Node reports a timestamp is N and it's updated to N+200us, then shouldn't the resulting timestamp afterwards be exactly N+200us? I'm not sure why the division would be needed because N should already be at the maximal precision (or so I'm assuming) and the 200us would also be at the maximal precision too, neither is more precise than what the system supports.

alexcrichton:
Also sorry I don't mean to be pedantic about minor test changes. It's ok to land something that works since the cases here are relatively minor, but nevertheless I'd also like to understand this change more.

crates/test-programs/src/bin/preview1_fd_filestat_set.rs
guybedford: Ahh I misread the documentation here that the units were 100 nanoseconds.

The actual units are therefore microsecond accuracy and this is 2 microseconds.

guybedford:
@alexcrichton apologies for the confusion, I made an invalid assumption that the u64 mtime value was in units of 100 nanoseconds due to a misreading of the docs. I've updated the PR to now reference 1us accuracy.

I've changed the configuration option name to `FS_TIME_PRECISION`.

> I also don't understand why Node is testing equality using division. Shouldn't the exact values be equal in Node? For example if Node reports a timestamp is N and it's updated to N+200us, then shouldn't the resulting timestamp afterwards be exactly N+200us?

Because Node.js rounds to microseconds, if we set a timestamp N, the timestamp that actually gets set is a Node.js-rounded N.

> The other axis is a fine-grained update of time.

For now I've treated both of these axes the same - the fs time precision. Where the unit of time we update is twice the precision, and the unit of time we check for is equality under that same precision. This makes sense I think because we are effectively observing this implementation-defined rounding function.

Ready for review, further feedback welcome.

alexcrichton (approved):
Ok this looks reasonable to me, thanks!

I still don't fully understand why `fs_duration_to_precision` is required in these tests, and I'll write out here why, but in the meantime I'm going to approve this an queue it for merging since it's just my own curiousity.

All the tests from what I can tell are taking a timestamp reported by the system and then modifying it by some multiple of the configured precision. Assuming that the system is only reporting timestamps with rounded precision already, this means that all timestamps in play should all be appropriately rounded, right? If that's the case that's why I don't understand why this function exists, although I understand that if arbitrary durations were added to timestamps it would be required.

guybedford:
> Assuming that the system is only reporting timestamps with rounded precision already, this means that all timestamps in play should all be appropriately rounded, right?

Good point, yes this is the issue - Node.js provides the ability to read times as `BigInt` values, but not to set them as `BigInt` values. I've posted a Node.js feature request for this in https://github.com/nodejs/node/issues/50859.

alexcrichton:
Aha! Ok that makes a lot more sense now. So Node is reporting nanosecond-granularity but only supports setting at the microsecond-granularity?

guybedford:
> So Node is reporting nanosecond-granularity but only supports setting at the microsecond-granularity?

Yes, that's what I'm seeing. They only support IEEE-754 values in seconds, which would only store 24 bits of precision, so I think this is actually due to floating point arithmetic rounding, but haven't verified that's definitely the reason.

alexcrichton:
Ah I could definitely imagine that once floats are involved precision flies out the window. Thanks for putting up with my questions!

guybedford:
CI was failing due to the use of an experimental feature. I've simplified the check which should pass CI now.

PR #7573, opened 2023-11-22
Winch: fix bug by spilling when emitting func call

jeffcharles:
The differential fuzzer triggered a `mem.slot.offset.as_u32() == masm.sp_offset().as_u32()` assertion while compiling the following module:

```wat
(module
    local.get 0
    i32.const 1
    call 0
    i32.const 1
    call 0
    br_if 0 (;@0;)
    unreachable
  )
  (export "" (func 0))
)
```

It looks like cause of the issue is that using `save_live_registers_and_calculate_sizeof` can violate a value stack invariant by placing `Memory` entries after `Local` and `Reg` entries.

To remediate this, we can trigger a spill instead of only saving live registers when emitting the machine code for a function call and we can calculate the amount of stack space consumed by the call in the `cleanup` method by summing the size of each memory entry used as an argument for the call.

As a result of spilling, I needed to add specify which scratch register to use for `funcref`s. I also needed to add a spill at the start of visiting a `call_indirect` operator. This is because the machine code associated with a spill that occurs when calling a function to initialize the table element may be jumped over, resulting in the codegen context state being out of sync with the machine state, which potentially causes the machine stack to become unbalanced.

### DIFF
--- a/winch/codegen/src/abi/mod.rs
+++ b/winch/codegen/src/abi/mod.rs
@@ -106,7 +106,7 @@ pub(crate) trait ABI {
     /// Returns the designated scratch register for the given [WasmType].
     fn scratch_for(ty: &WasmType) -> Reg {
         match ty {
-            WasmType::I32 | WasmType::I64 => Self::scratch_reg(),
+            WasmType::I32 | WasmType::I64 | WasmType::Ref(_) => Self::scratch_reg(),
             WasmType::F32 | WasmType::F64 => Self::float_scratch_reg(),
             _ => unimplemented!(),
         }
--- a/winch/codegen/src/abi/mod.rs
+++ b/winch/codegen/src/abi/mod.rs
@@ -49,7 +49,7 @@ use crate::masm::{OperandSize, SPOffset};
 use smallvec::SmallVec;
 use std::collections::HashSet;
 use std::ops::{Add, BitAnd, Not, Sub};
-use wasmtime_environ::{WasmFuncType, WasmType};
+use wasmtime_environ::{WasmFuncType, WasmHeapType, WasmRefType, WasmType};
 
 pub(crate) mod local;
 pub(crate) use local::*;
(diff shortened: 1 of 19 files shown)


winch/codegen/src/codegen/call.rs
```diff
@@ -364,6 +313,15 @@ impl FnCall {
         // which was allocated last.
         masm.free_stack(reserved_space);
 
+        // Drop params from value stack and calculate amount of machine stack
+        // space they consumed.
+        let mut stack_consumed = 0;
+        context.drop_last(sig.params.len_without_retptr(), |_regalloc, v| {
```
jeffcharles, line 319: My understanding is `v` should never be a register when we reach this code because we spilled the value stack earlier. I've tried to confirm that by adding a panic if `v` is a register and running the test suite and filetests and both passed.

winch/codegen/src/abi/mod.rs
```diff
@@ -106,7 +106,7 @@ pub(crate) trait ABI {
     /// Returns the designated scratch register for the given [WasmType].
     fn scratch_for(ty: &WasmType) -> Reg {
         match ty {
-            WasmType::I32 | WasmType::I64 => Self::scratch_reg(),
+            WasmType::I32 | WasmType::I64 | WasmType::Ref(_) => Self::scratch_reg(),
```
saulecabrera, line 109: Only `WasmHeapType::Func` is supported at the moment. Can we update this snippet to reflect that? We'll be able to lift such restriction once externrefs and typed funcs are supported.

winch/codegen/src/codegen/call.rs
saulecabrera: Your understanding is correct, aside from matching against `Val::Memory`, I think it'd be useful to assert for such invariant in the code (e.g. `debug_assert!(v.is_mem() || v.is_const())`, that way we'd at least have a way of identifying any issues early on if this invariant ever breaks. 

NB: `is_const()` currently doesn't exist, so if we go with this approach, it needs  to be added.

saulecabrera (commented):
Thanks for digging into this. Left a few comments in the code, and also: can we update the comments in `call.rs` to reflect the new changes? The module documentation and the documentation for `FnCall::emit` both state that the implementation only saves live registers, which is no longer true.

winch/codegen/src/visitor.rs
```diff
@@ -925,6 +925,13 @@ where
     }
 
     fn visit_call_indirect(&mut self, type_index: u32, table_index: u32, _: u8) {
+        // Spill now because `emit_lazy_init_funcref` and the `FnCall::emit`
+        // invocations will both trigger spills since they both call functions.
+        // However, the machine instructions for the spill emitted by
+        // `emit_lazy_funcref` may be jumped over which may result in the
```
saulecabrera, line 931: Can you add a bit more context here on when these instructions get jumped over? e.g. if the funcref was previously initialized.

saulecabrera (approved):

PR #7576, opened 2023-11-22
mpk: allow forcing MPK during tests

abrown:
For testing on machines on which we know MPK is enabled, we want to be
able to force-enable MPK, ensuring we get coverage of MPK-related code.
This change adds a `WASMTIME_TEST_FORCE_MPK` environment variable which,
when set, sets the pooling allocator configuration to force-enable MPK.
This variable, like the `WASMTIME_TEST_NO_HOG_MEMORY` variable it is
styled from, could be used in CI workflows on which we know MPK should
be available.

### DIFF
--- a/crates/runtime/src/instance/allocator/pooling.rs
+++ b/crates/runtime/src/instance/allocator/pooling.rs
@@ -225,13 +225,6 @@ pub struct PoolingInstanceAllocatorConfig {
 
 impl Default for PoolingInstanceAllocatorConfig {
     fn default() -> PoolingInstanceAllocatorConfig {
-        // When testing, we may choose to start with MPK force-enabled to ensure
-        // we use that functionality.
-        let memory_protection_keys = if std::env::var("WASMTIME_TEST_FORCE_MPK").is_ok() {
-            MpkEnabled::Enable
-        } else {
-            MpkEnabled::Disable
-        };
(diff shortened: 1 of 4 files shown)


alexcrichton:
My first reaction to this is that this isn't where I'd like to see an env var being tested. I think currently for example `WASMTIME_TEST_NO_HOG_MEMORY` is only checked in test binaries, never in Wasmtime runtime code. Ideally we'd do the same here and only change test-related code to check this env var. That being said I can see how that's not easy to do since there isn't one location where all the `Engine`s are created.

Perhaps just the `wast` tests and maybe a few others could get mpk force-enabled with this env var though? Get some broad coverage but not necessarily 100%. Another option would be RUSTFLAGS with `--cfg force_mpk` or something like that

abrown:
> My first reaction to this is that this isn't where I'd like to see an env var being tested.

Makes sense. The commit I just pushed, 3d2102a, limits the use of this variable to the `wast` tests and the 30 uses of `small_pool_config`. Forcing MPK in tests that use `small_pool_config` is not _too_ useful: they only really have one memory slot so we don't check the multiple slices. But it may exercise a few more of the MPK code paths. In any case, the `wast` tests _are_ useful: any time in the past I've tested an incorrect implementation of MPK, the `wast` tests have had some failures.

PR #5550, opened 2023-01-08
aarch64: Support GOT Relative relocations in PIC mode

afonso360:
👋 Hey,

This PR adds support for using the GOT in PIC mode on AArch64. It implements the relocations for both ELF and MachO.

In order to support the MachO relocations we need to update the object crate. Unfortunately that also brings a few other dependencies which also need to be vetted. 

[@cfallin has already validated the object crate update up to `0.30.0` in #5434](https://github.com/bytecodealliance/wasmtime/pull/5434#issuecomment-1350337637) so we just need to do `0.30.0 -> 0.30.1` and the rest of the new dependencies.

Tested this using [the awesome example](https://github.com/acw/cranelift-hmmm/) provided by @acw in [zulip](https://bytecodealliance.zulipchat.com/#narrow/stream/217117-cranelift/topic/Tricks.20with.20linkage.20internal.20.2F.20module.20external.20functions.3F).
On aarch64-linux I was able to fully test this, on aarch64-darwin the decompiled file seems correct but I don't have a M1/2 machine to link and run it. Maybe @acw can test and report back if anything is wrong?

Fixes #2907
Fixes #5544

Issue #2735, opened 2021-03-17
JIT for arm


benmkw:
#### Feature

JIT for arm64

#### Benefit

x86 works and it would be nice to be able to use it on M1 mac as well

#### Implementation

it currently panics here: https://github.com/bytecodealliance/wasmtime/blob/5fecdfa49150e3304c1b949aab73bd4a0a02dbac/cranelift/jit/src/backend.rs#L183-L195 I'm unsure what needs to change/ whats blocking it

#### Alternatives

only make JIT available on x86

cfallin:
@bnjbvr any thoughts on this? I'm not sure how PLTs ordinarily work on macOS/aarch64 but perhaps it's a simple addition?

bjorn3:
As workaround you could disable the `is_pic` flag when creating the `TargetIsa`.

bjorn3:
This is one of the plt entries of an executable I compiled on my phone:

```asm
90 01 00 B0    adrp x16, #0x31000
11 F2 46 F9    ldr  x17, [x16, #0xde0]
10 82 37 91    add  x16, x16, #0xde0
20 02 1F D6    br   x17
```

benmkw:
> As workaround you could disable the is_pic flag when creating the TargetIsa.

Tried with https://github.com/bytecodealliance/cranelift-jit-demo
```diff
diff --git a/Cargo.toml b/Cargo.toml
index 756983f..f6dd93f 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,8 @@ description = "Toy language implemented using cranelift-jit"
 edition = "2018"
 
 [dependencies]
-cranelift = "0.69.0"
-cranelift-module = "0.69.0"
-cranelift-jit = "0.69.0"
+cranelift = "0.72.0"
+cranelift-module = "0.72.0"
+cranelift-jit = "0.72.0"
+cranelift-native = "0.72.0"
 peg = "0.6"
diff --git a/src/jit.rs b/src/jit.rs
index 160ac36..ffcff04 100644
--- a/src/jit.rs
+++ b/src/jit.rs
@@ -26,7 +26,17 @@ pub struct JIT {
 
 impl Default for JIT {
     fn default() -> Self {
-        let builder = JITBuilder::new(cranelift_module::default_libcall_names());
+        let mut flag_builder = settings::builder();
+        // On at least AArch64, "colocated" calls use shorter-range relocations,
[...]

PR #7584, opened 2023-11-27
cranelift-jit: implement GOT relocations for aarch64

Mrmaxmeier:
This is a prerequisite for #2735, but doesn't implement the PLT entries yet.
(I'm using hot function replace, which happens in the GOT but doesn't strictly require the PLT.)

The instruction encoding for `adrp` is kind of complicated, but re-using existing lowering infrastructure doesn't work well since it's not designed to do partial updates.  Xref: #5550, which introduced these relocation types.


cranelift/jit/src/compiled_blob.rs
```diff
@@ -93,6 +93,31 @@ impl CompiledBlob {
                     let imm26 = (diff as u32) << chop >> chop;
                     unsafe { modify_inst32(iptr, |inst| inst | imm26) };
                 }
+                Reloc::Aarch64AdrGotPage21 => {
+                    // Set the immediate value of an ADRP to bits [32:12] of X; check that –2^32 <= X < 2^32
+                    let base = get_got_entry(name);
+                    let what = unsafe { base.offset(isize::try_from(addend).unwrap()) };
```
elliottt, line 99: We only emit `Aarch64AdrGotPage21` relocations with an `addend` of `0`. Would you mind removing the offset, and instead asserting that it's `0`?

Additionally, to fully implement the operation `Page(G(GDAT(S+A)))-Page(P)` we would need to modify the function `get_got_entry` that's passed in, as we would need to make the offset calculation before looking up in the GOT. Asserting that the `addend` is zero means that we can avoid that refactoring.

cranelift/jit/src/compiled_blob.rs
```diff
@@ -93,6 +93,31 @@ impl CompiledBlob {
...
+
+                    let immlo = ((val as u32) & 0b11) << 29;
+                    let immhi = (((val as u32) >> 2) & &0x7ffff) << 5;
+                    let mask = !((0x7ffff << 5) | (0b11 << 29));
+                    unsafe { modify_inst32(at as *mut u32, |adrp| (adrp & mask) | immlo | immhi) };
+                }
+                Reloc::Aarch64Ld64GotLo12Nc => {
+                    // Set the LD/ST immediate field to bits 11:3 of X. No overflow check; check that X&7 = 0
+                    let base = get_got_entry(name);
+                    let what = unsafe { base.offset(isize::try_from(addend).unwrap()) };
```
elliottt, line 113: We only emit `Aarch64Ld64GotLo12Nc` relocations with an `addend` of `0`. Would you mind removing the offset, and instead asserting that it's `0`?

elliottt (commented):
Happy to approve after the offsets are removed, thank you!

cranelift/jit/src/compiled_blob.rs
Mrmaxmeier: Good catch! :+1:

elliottt (approved):
Thank you!

PR #7586, opened 2023-11-27
preview1: Update semantics of seeking on appending files

### LINKED ISSUE #7583: `fd_tell` not working when file is opened with `fdflags::append`
### Test Case

Compile the below C code with wasi-sdk.

```
#include <fcntl.h>
#include <string.h>
#include <stdio.h>
#include <unistd.h>

int main() {
	int f = open("tmp/a", O_CREAT | O_WRONLY | O_APPEND);
	if (f == -1) perror("open");

	int offset = lseek(f, 0, SEEK_CUR);
	printf("offset before write %d\n", offset);

	int written = write(f, "abc", 4);
	printf("written %d bytes\n", written);

	offset = lseek(f, 0, SEEK_CUR);
	printf("offset after write %d\n", offset);
}
```

### Steps to Reproduce

1. Compile the C code with wasi-sdk to a WASM file say: a.wasm
2. Run the code mounting an empty directory: `wasmtime run --dir tmp a.wasm`

### Expected Results

fd_tell should report the actual fd offset equal to bytes written.

```
offset before write 0
written 4 bytes
offset after write 4
```

### Actual Results

fd_tell always reports 0.

```
offset before write 0
written 4 bytes
offset after write 0
```

### Versions and Environment

Wasmtime v15.0.0

Operating system: Fedora 39

Architecture: amd64

alexcrichton:
This commit updates the semantics of `fd_{seek,tell}` on preview1 to match native Unix when used with appending files. On Unix `write` claims to always update the file position pointer to the end of the file, so this commit implements that instead of the previous logic of ignoring the position update for appending files. This currently requires an extra roundtrip via `stat` to figure out the size of the file, but for now that seems to be the best that can be done.

Closes #7583

### DIFF
--- a/crates/test-programs/src/bin/preview1_file_seek_tell.rs
+++ b/crates/test-programs/src/bin/preview1_file_seek_tell.rs
@@ -79,6 +79,48 @@ unsafe fn test_file_seek_tell(dir_fd: wasi::Fd) {
     wasi::fd_close(file_fd).expect("closing a file");
     wasi::path_unlink_file(dir_fd, "file").expect("deleting a file");
 }
+
+// Test that when a file is opened with `O_APPEND` that acquiring the current
+// position indicates the end of the file.
+unsafe fn seek_and_o_append(dir_fd: wasi::Fd) {
+    let path = "file2";
+    let file_fd = wasi::path_open(
+        dir_fd,
+        0,
+        path,
+        wasi::OFLAGS_CREAT,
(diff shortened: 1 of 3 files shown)


alexcrichton:
For the Unix behavior I'm reading [this site](https://man7.org/linux/man-pages/man2/lseek.2.html), specifically the "NOTES" section

pchickey (approved):
Thanks!

PR #7589, opened 2023-11-27
wasi-http spec tweaks

elliottt:
This is a draft of some proposed tweaks to the wasi-http spec:

* Removing the `-ms` suffix on the timout setters, and references to the value being in milliseconds
* Making the `future-trailers.get` method return the trailers only once, and as a result also returning an owned `HostFields`

### DIFF
--- a/crates/wasi-http/src/types_impl.rs
+++ b/crates/wasi-http/src/types_impl.rs
@@ -889,14 +889,14 @@ impl<T: WasiHttpView> crate::bindings::http::types::HostRequestOptions for T {
         &mut self,
         opts: Resource<types::RequestOptions>,
     ) -> wasmtime::Result<Option<types::Duration>> {
-        let millis = self
+        let nanos = self
             .table()
             .get(&opts)?
             .connect_timeout
-            .map(|d| d.as_millis());
+            .map(|d| d.as_nanos());
 
-        if let Some(millis) = millis {
-            Ok(Some(millis.try_into()?))
+        if let Some(nanos) = nanos {
+            Ok(Some(nanos.try_into()?))
         } else {
             Ok(None)
         }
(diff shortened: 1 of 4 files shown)


crates/wasi-http/src/types_impl.rs
```diff
@@ -886,7 +885,7 @@ impl<T: WasiHttpView> crate::bindings::http::types::HostRequestOptions for T {
         Ok(id)
     }
 
-    fn connect_timeout_ms(
+    fn connect_timeout(
```
alexcrichton, line 888: I think the internals of all these methods also need to change since right now they're using `*_millis`-related `Duration` methods

crates/wasi-http/wit/deps/http/types.wit
```diff
@@ -437,16 +436,20 @@ interface types {
...
+    /// body, as well as any trailers, were received successfully, or that an
+    /// error occured receiving them. The optional `trailers` indicates whether
+    /// or not trailers were present in the body.
     ///
     /// When some `trailers` are returned by this method, the `trailers`
     /// resource is immutable, and a child. Use of the `set`, `append`, or
     /// `delete` methods will return an error, and the resource must be
     /// dropped before the parent `future-trailers` is dropped.
-    get: func() -> option<result<option<trailers>, error-code>>;
+    get: func() -> option<result<result<option<trailers>, error-code>>>;
```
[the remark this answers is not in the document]
pchickey: Yes, we are sticking to the same convention for the sake of bindings generators knowing how to translate the "futures pattern" into some native future. its not a fully baked thing yet but we at least want to try.

crates/wasi-http/src/types_impl.rs
elliottt: Good catch, thanks!

pchickey (approved):

PR #7591, opened 2023-11-28
Raise Http{Request,Response}BodyError when Content-Length doesn't match

elliottt:
Instead of `InternalError`, raise a more specific request or response body size error when the number of bytes written to the outgoing body doesn't match the expectation set by the `Content-Length` header.

This change does overload the meaning of the `http-{request,response}-body-size` errors, as they were previously defined as meaning specifically that the body was too large, while this PR uses them to mean that the body length didn't match the expected length.

Addresses https://github.com/WebAssembly/wasi-http/issues/83

### DIFF
--- a/crates/test-programs/src/bin/http_outbound_request_content_length.rs
+++ b/crates/test-programs/src/bin/http_outbound_request_content_length.rs
@@ -54,11 +54,7 @@ fn main() {
             http_types::OutgoingBody::finish(outgoing_body, None).expect_err("finish should fail");
 
         assert!(
-            matches!(
-                &e,
-                http_types::ErrorCode::InternalError(Some(s))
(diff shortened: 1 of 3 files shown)


pchickey:
We never documented what the `option<u64>` payload in those error variants means - I guess I expected it to mean the expected content-length (none given, or some set value) rather than the number of bytes provided for the body. I guess either way is OK, I have a slight preference for the expected value rather than the got value, but we need to document what it should mean in the wit.

pchickey (approved):
LGTM, We discussed the expected vs got payload in chat,

@elliottt :
| the original rfc says that the size should be the received size, not the expected size: https://www.rfc-editor.org/rfc/rfc9209#section-2.3.21.

PR #7593, opened 2023-11-28
PCC: Fix several aarch64 check_constant failures

feilongjiang:
Hi team, please consider:
This patch fixes several aarch64 check_constant failures:

1. `check_subsume` for `AluRRImmLogic` failed due to the mismatch of fact width

  isle rule for `orr_imm` always generates fact with bit_width 64 regardless of
  immediate type. So when Type is I32, `check_subsume` will be failed.
  This could be reproduced with the simple wasm module [1].

2. `MovN` generates incorrect fact range value when `first_is_inverted` is true

  `running_value` should be calculated as `!(((!imm16) & 0xffff) << shift)` or
  `!(u64::from(imm.bits) << shift)`
  This could be reproduced with the simple wasm module [2].

Added two test cases in cranelift/filetests/filetests/pcc/succeed/const.clif.

Additional fix for `get_fact_or_default`:

  `trace!` treats `reg` as VirtualReg and it will panic when `reg` is RealReg and
  `trace-log` feature is enabled.

[1]:
```wasm
;; test_logic.wat
(module
  (func (export "test_logic") (result i32) (i32.const 0x100010))
)
```
[2]:
```wasm
;; test_const.wat
(module
  (func (export "test_constant") (result i64) (i64.const 0x0009ffffffff))
)
```

### DIFF
--- a/cranelift/codegen/src/machinst/pcc.rs
+++ b/cranelift/codegen/src/machinst/pcc.rs
@@ -6,12 +6,7 @@ use crate::trace;
 
 pub(crate) fn get_fact_or_default<I: VCodeInst>(vcode: &VCode<I>, reg: Reg, width: u16) -> Fact {
     trace!(
-        "get_fact_or_default: reg {} -> {:?}",
-        if reg.is_virtual() {
-            format!("v{}", reg.to_virtual_reg().unwrap().index())
-        } else {
-            format!("r{}", reg.to_real_reg().unwrap().hw_enc())
-        },
+        "get_fact_or_default: reg {reg:?} -> {:?}"
         vcode.vreg_fact(reg.into())
     );
     vcode
--- a/cranelift/codegen/src/machinst/pcc.rs
+++ b/cranelift/codegen/src/machinst/pcc.rs
@@ -6,7 +6,7 @@ use crate::trace;
 
 pub(crate) fn get_fact_or_default<I: VCodeInst>(vcode: &VCode<I>, reg: Reg, width: u16) -> Fact {
     trace!(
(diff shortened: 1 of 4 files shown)


alexcrichton:
Thanks for the PR!

 The two folks who are best to review this PR, @cfallin and @fitzgen, are both out on leave right now. They'll get to this PR when they get back though (@fitzgen's coming back mid-December). Thanks for your patience

cranelift/codegen/src/machinst/pcc.rs
```diff
@@ -6,8 +6,12 @@ use crate::trace;
...
 pub(crate) fn get_fact_or_default<I: VCodeInst>(vcode: &VCode<I>, reg: Reg, width: u16) -> Fact {
     trace!(
-        "get_fact_or_default: reg v{} -> {:?}",
-        reg.to_virtual_reg().unwrap().index(),
+        "get_fact_or_default: reg {} -> {:?}",
+        if reg.is_virtual() {
+            format!("v{}", reg.to_virtual_reg().unwrap().index())
+        } else {
+            format!("r{}", reg.to_real_reg().unwrap().hw_enc())
+        },
```
fitzgen, line 14: I think this can be simplified to

```suggestion
        "get_fact_or_default: reg {reg:?} -> {:?}"
```

fitzgen (commented):
LGTM with one suggestion to try before we merge this. Thanks!

fitzgen:
And thanks for your patience!

feilongjiang:
@fitzgen Thanks for the review and suggestions!

fitzgen (approved):
Thanks!

Issue #6785, opened 2023-07-30
macos: "failed to set thread exception port" in forked process


thibaultcha:
Hello,

Avid Wasmtime users in our [Nginx embedding](https://github.com/Kong/ngx_wasm_module), we are facing an assertion failure when running it with Wasmtime on macos (x86 and arm64):

```
thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `268435466`,
 right: `0`: failed to set thread exception port', crates/runtime/src/traphandlers/macos.rs:474:9
```

This assertion systematically fails with Wasmtime 8.0.1 (and probably earlier) to 11.0.1 when Nginx is configured with [daemon on](https://nginx.org/en/docs/ngx_core_module.html#daemon) (which ends [here](https://github.com/nginx/nginx/blob/master/src/os/unix/ngx_daemon.c#L13) in Nginx forking into a background process). It seems it occurs in our call to `wasmtime_linker_instantiate` after having initialized an engine and a single store.

Not forking the master process (`daemon off`) seems to be working fine, even in forked worker processes (managed by the master process in foreground); instances are created and behave as expected in the Nginx worker processes.
The assertion failure specifically occurs once the master process has forked itself into a background process with `daemon on`.

In summary:
- Works with `daemon off`: master process (foreground) -> `wasmtime_linker_instantiate` (temporary instance) -> fork() -> worker processes -> `wasmtime_linker_instantiate` (worker instances processing requests).
- Fails with `daemon on`: master process (foreground) -> fork() (background daemon) -> `wasmtime_linker_instantiate` (temporary instance) -> assertion failure, no instance created, and no chance to fork() into worker processes.

In the past we used to have a CI/CD pipeline with macos targets and Wasmtime used to work fine; but this CI/CD pipeline was taken down, and this bit us in the last few days. It seems like [older macos work](https://github.com/bytecodealliance/wasmtime/commit/5fecdfa49150e3304c1b949aab73bd4a0a02dbac#diff-af72c3bc843c88220e004d0703c683dee8306b270e8309e4bc26d4c7b59b2e52R436-R476) on Wasmtime has something to do with it, but I know nothing of the macos system interface...

So far the root cause of this assertion failure has eluded us; probably state that aren't being carried over to the forked process or something like this. Could someone shed some light on what may be at cause here?

Thanks!

thibaultcha:
@casimiro has a good example and complete stacktrace of the assertion in #6788. 

[...]

PR #7595, opened 2023-11-28
c-api: support disabling mach ports use on macos

casimiro:
This PR proposes exposing [macos_use_mach_ports](https://docs.wasmtime.dev/api/wasmtime/struct.Config.html#method.macos_use_mach_ports) engine configuration in the c-api.

As discussed [here](https://github.com/bytecodealliance/wasmtime/issues/6785) host applications that employ fork strategies might benefit from disabling `macos_use_mach_ports` when running on macOS.


pchickey (approved):
Thanks!

PR #7596, opened 2023-11-28
Replace the preview2 table's HashMap storage with a Vec

elliottt:
Fix the TODO in preview2's table implementation about performance of key generation once the index space has wrapped; instead of incrementing a u32, store resources in a vector and keep a free queue in the empty slots.

### DIFF
--- a/crates/wasi/src/preview2/table.rs
+++ b/crates/wasi/src/preview2/table.rs
@@ -336,3 +336,43 @@ impl Default for Table {
         Table::new()
     }
 }
+
+#[test]
+pub fn test_free_queue() {
+    let mut table = Table::with_capacity(0);
+
+    let x = table.push(()).unwrap();
+    assert_eq!(x.rep(), 0);
+
+    let y = table.push(()).unwrap();
+    assert_eq!(y.rep(), 1);
+
+    let start = y.rep() + 1;
+    let end = table.entries.capacity() as u32;
+
+    for i in start..end {
+        let x = table.push(()).unwrap();
+        assert_eq!(x.rep(), i);
+        assert_eq!(table.entries.capacity() as u32, end);
+    }
+
+    // Deleting x should put it on the free queue, so the next entry should have the same rep.
+    table.delete(x).unwrap();
+    let x = table.push(()).unwrap();
+    assert_eq!(x.rep(), 0);
+
+    // Deleting y and then x should yield indices 1 and then 0.
+    table.delete(y).unwrap();
+    table.delete(x).unwrap();
+
+    let y = table.push(()).unwrap();
+    assert_eq!(y.rep(), 1);
+
+    let x = table.push(()).unwrap();
+    assert_eq!(x.rep(), 0);
+
+    // As the vector is at capacity, this new entry should force it to grow, making the next entry
+    // the previous capacity.
+    let x = table.push(()).unwrap();
+    assert_eq!(x.rep(), end);
+}
--- a/crates/wasi/src/preview2/table.rs
+++ b/crates/wasi/src/preview2/table.rs
@@ -276,13 +276,10 @@ impl Table {
     ) -> impl Iterator<Item = (Result<&'a mut dyn Any, TableError>, T)> {
         map.into_iter().map(move |(k, v)| {
             let item = self
-                .entries
-                .get_mut(k as usize)
-                .and_then(Entry::occupied_mut)
+                .occupied_mut(k)
                 .map(|e| Box::as_mut(&mut e.entry))
                 // Safety: extending the lifetime of the mutable reference.
-                .map(|item| unsafe { &mut *(item as *mut dyn Any) })
-                .ok_or(TableError::NotPresent);
+                .map(|item| unsafe { &mut *(item as *mut dyn Any) });
             (item, v)
         })
     }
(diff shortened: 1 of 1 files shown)


crates/wasi/src/preview2/table.rs
```diff
@@ -67,13 +90,15 @@ impl TableEntry {
 impl Table {
     /// Create an empty table
     pub fn new() -> Self {
+        // TODO: what's a good default here?
+        Table::with_capacity(100)
```
pchickey, line 94: I don't have an intuition for what a good magic number for capacity is. @alexcrichton any guesses?

pchickey (approved):
Looks good to me. Thank you

crates/wasi/src/preview2/table.rs
alexcrichton: Could `with_capacity` be removed and this start as an empty vec? I'm not sure if there's much to gain from `with_capacity` really other than letting default heuristics take over.

crates/wasi/src/preview2/table.rs
```diff
@@ -23,8 +23,31 @@ pub enum TableError {
 /// up. Right now it is just an approximation.
 #[derive(Debug)]
 pub struct Table {
-    map: HashMap<u32, TableEntry>,
-    next_key: u32,
+    entries: Vec<Entry>,
+    free_head: Option<usize>,
+    free_tail: Option<usize>,
```
alexcrichton, line 28: Could this perhaps become just `free_head`? Currently this looks like a FIFO list but changing it to LIFO I think is ok as well and removes the amount of state to manage here.

crates/wasi/src/preview2/table.rs
```diff
@@ -87,23 +112,89 @@ impl Table {
...
+        self.free_tail = Some(ix);
+
+        entry
+    }
+
+    /// Push a new entry into the table, returning its handle. The following strategy is used:
+    /// 1. if there's capacity in the vector, push a new entry at the end,
+    /// 2. if the free queue is non-empty, use the entry at the head of the queue,
+    /// 3. finally, grow the vector and push on the end.
+    fn push_(&mut self, e: TableEntry) -> Result<u32, TableError> {
```
alexcrichton, line 168: Lots of methods here have `_` as a suffix, and while that was used for some prior that was only done when there's a `pub` method with the same name and the `_`-version erases generics effectively. Could the `_` be removed when that's not necessary?

crates/wasi/src/preview2/table.rs
```diff
@@ -87,23 +112,89 @@ impl Table {
...
+        let len = self.entries.len();
+        if len < self.entries.capacity() {
+            let ix = len.try_into().map_err(|_| TableError::Full)?;
+            self.entries.push(Entry::Occupied { entry: e });
+            Ok(ix)
```
alexcrichton, line 173: I was naively expecting a slab-like allocator when taking a look at this, which wouldn't have this branch. Could this perhaps be removed to help simplify this a bit further?

alexcrichton (commented):
I've got some thoughts on possibly simplifying this a bit, but overall looks good!

crates/wasi/src/preview2/table.rs
elliottt: My thought here was that it would be good to allow a bit of control so that embedders could make that decision depending on their workloads. Perhaps that's over-estimating the cost of reallocating the vector though.

crates/wasi/src/preview2/table.rs
elliottt: Absolutely, I was just trying to follow the conventions already present.

elliottt: What do you think about removing `push_child_`? it's called only by `push_child` and would inline without issues.

alexcrichton: Sounds good to me! Many of these are historical and the artifact of a number of refactorings, so it's definitely ripe for taking the liberty to restructure things while you're here.

crates/wasi/src/preview2/table.rs
alexcrichton: Ok keeping `with_capacity` makes sense to me, but I think that `new` should create with the default 0 capacity of `Vec` and then `with_capacity` sets an initial capacity.

crates/wasi/src/preview2/table.rs
alexcrichton: Trevor and I talked a bit on Zulip, and I agree that this would be useful for that form of debugging but I'd personally prefer to stick to a simpler FIFO queue since I think it's not clear whether this form of debugging would be relied on all that much. As an isolated implementation detail it's easy to swap this out in the future, and I'd prefer personally to see the need for such debugging come up first before we go with the slightly fancier LIFO implementation.

It's sort of 6-to-one-half-dozen-or-the-other though so I don't feel too strongly either way, but I do have a preference for the FIFO version. I'm probably biased though b/c that's what I was assuming this was before I read it.

elliottt: I mixed up my acronyms. Just for the record, I'm going to remove the tail pointer and make this a LIFO, removing the existing FIFO behavior. Sorry about the confusion there :)

alexcrichton: Oops, I also am appparently forgetting what I was just saying yesterday. Anyway sounds good!

alexcrichton (approved):

PR #7597, opened 2023-11-28
Trim `wasmtime serve`'s default world

alexcrichton:
This commit aims to address a discrepancy in Wasmtime where the world supported by `wasmtime serve` is too large today. This includes WIT interfaces which are not specified in `wasi:http/proxy` such as `wasi:filesystem/types`, aka access to a filesystem.

This commit slims down `wasmtime serve` to, by default, only supporting the `wasi:http/proxy` world. Like with `wasmtime run` various CLI flags can be passed to enable more interfaces, however:

* `-Scommon` - this enables "common" interfaces such as `wasi:filesystem`, `wasi:sockets`, and `wasi:cli/*`.
* `-Snn` - this enables wasi-nn

It's expected that more will get extended here over time too.

This change is enabled by a third build of the adapter, a "proxy" mode. In this mode most functions are cfg'd to return `ERRNO_NOTSUP` to indicate that the runtime does not support it. Notably this includes the filesystem, arguments, and environment variables.

This change is tested by updating all `api_proxy*` tests to use this new adapter which is now required that many previous interfaces are no longer present by default in the proxy world.

### DIFF
--- a/crates/wasi-http/src/proxy.rs
+++ b/crates/wasi-http/src/proxy.rs
@@ -25,10 +25,24 @@ pub fn add_to_linker<T>(l: &mut wasmtime::component::Linker<T>) -> anyhow::Resul
 where
     T: WasiHttpView + preview2::WasiView + bindings::http::types::Host,
 {
-    // TODO: this shouldn't be required, but the adapter unconditionally pulls in all of these
-    // dependencies.
-    preview2::command::add_to_linker(l)?;
+    preview2::bindings::clocks::wall_clock::add_to_linker(l, |t| t)?;
+    preview2::bindings::clocks::monotonic_clock::add_to_linker(l, |t| t)?;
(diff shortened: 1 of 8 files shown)


crates/wasi-http/src/proxy.rs
```diff
@@ -25,10 +25,24 @@ pub fn add_to_linker<T>(l: &mut wasmtime::component::Linker<T>) -> anyhow::Resul
 where
     T: WasiHttpView + preview2::WasiView + bindings::http::types::Host,
 {
-    // TODO: this shouldn't be required, but the adapter unconditionally pulls in all of these
-    // dependencies.
-    preview2::command::add_to_linker(l)?;
```
elliottt, line 30: :tada:

elliottt (approved):
This looks great!

PR #7599, opened 2023-11-29
Cranelift: Remove some unneeded `arithmetic.isle` patterns

scottmcm:
I noticed that, thanks to const prop, a few of the patterns in `arithmetic.isle` aren't actually needed to keep the same behaviour.

This seems to have mostly lowered the register numbers in the affected tests, so it seems at least plausible.  But if it's expected that patterns don't don't rely on canonicalization and it's better to keep these duplicates, then please just close this PR.

(I needed to learn how to edit patterns and write+run tests against them for what I actually want to do, so this was a useful learning exercise for me whether it lands or not.)


alexcrichton (approved):
No worries about the renumbering, that is indeed expected. Thanks for the PR!

PR #7602, opened 2023-11-29
winch: Ensure stack pointer for br_table

saulecabrera:
Follow up to:

https://github.com/bytecodealliance/wasmtime/pull/7547

In which I overlooked this change and the fuzzer found an issue with the following program:

```wat
(module
  (func (export "") (result i32)
    block (result i32)
       i32.const 0
    end
    i32.const 0
    i32.const 0
    br_table 0
  )
)
```

This commit ensures that the stack pointer is correctly positioned when emitting br_table.

We can't know for sure which branch will be taken, but since all branches must share the same type information, we can be certain that the expectations regarding the stack pointer are the same and thus can we use the default target in order to ensure the correct placement.

This change also adds some missing spec test cases for `br_table`.


elliottt (approved):
Looks good to me!

As an aside, is there a good way that we can be confident that we've flushed out any remaining stack balance issues? Perhaps fuzzing is enough?

saulecabrera:
> As an aside, is there a good way that we can be confident that we've flushed out any remaining stack balance issues? Perhaps fuzzing is enough?

We've been locally fuzzing after each new change; I think that's probably the best way to ensure that we've addressed all the stack balancing issues (or any other issues really), given that many of the edge cases that we've found are not covered through the spec tests.

saulecabrera:
Perhaps, we could also add additional run tests with these edge cases, that way they'll break, at runtime, even if they successfully compile. I'll remove this one from the queue momentarily and add an additional run test.

PR #7603, opened 2023-11-29
mpk: optimize layout of protected stripes

abrown:
While experimenting with the limits of MPK-protected memory pools, @alexcrichton and I discovered that the current slab layout calculations were too conservative. This meant that the memory pool could not pack in as many memories as it should have been able: we were expecting, but not seeing, ~15x more memory slots over non-MPK memory pools.

The fix ends up being simpler than the original: we must maintain the codegen constraints that expect a static memory to be inaccessible for OOB access within a `static_memory_maximum_size +
static_memory_guard_size` region (called `expected_slot_bytes + guard_bytes` in `memory_pool.rs`). By dividing up that region between the stripes, we still guarantee that the region is inaccessible by packing in other MPK-protected stripes. And we still need to make sure that the `post_slab_guard_bytes` add up to that region. These changes fix the memory inefficiency issues we were seeing.

### DIFF
--- a/crates/runtime/proptest-regressions/instance/allocator/pooling/memory_pool.txt
+++ b/crates/runtime/proptest-regressions/instance/allocator/pooling/memory_pool.txt
@@ -6,3 +6,4 @@
 # everyone who runs the test benefits from these saved cases.
 cc 696808084287d5d58b85c60c4720227ab4dd83ada7be6841a67162023aaf4914 # shrinks to c = SlabConstraints { max_memory_bytes: 0, num_memory_slots: 1, num_pkeys_available: 0, guard_bytes: 9223372036854775808 }
 cc cf9f6c36659f7f56ed8ea646e8c699cbf46708cef6911cdd376418ad69ea1388 # shrinks to c = SlabConstraints { max_memory_bytes: 14161452635954640438, num_memory_slots: 0, num_pkeys_available: 0, guard_bytes: 4285291437754911178 }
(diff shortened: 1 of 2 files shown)


alexcrichton:
Looks good to me!

I'd be happy to approve and land this as-is, but we talked about some other somewhat-unrelated things to this PR which I think you're going to include here as well.

I'll also note that I have one point of note which I'll emphasize shouldn't be addressed in this PR because I think it's not worth it. That being said I think that this strategy is a tiny bit wasteful in terms of pages because the slot size is unlikely to be a multiple of the number of stripes meaning something will be rounded up to a page boundary. For example with a 4G memory, a 2G guard, and a 512M max memory size that'll support up to 12 stripes (6G/512M == 12). If however only 11 stripes were available then what should in theory happen is that one 6G slot would have 11 memories of max size 512M packed together with some leftover guard space at the end. That would mean 11 memories would be packed into a 6G region.

With this PR, however, what would happen is that the 6G region would be divided into 11 chunks of size 0x22e8ba2e. That'd then be rounded up to 0x22e8c000 for a single slot, which if that's multiplied back by 11 yields 0x180004000 bytes. This means that with "ideal logic" a 6G region could hold 11 memories but this calculation instead requires a 6G+16K region to hold the 11 memories.

This is a very minor loss, however (6G vs four pages) however and accounting for some guard regions interspersed with all the slots would require deeper refactoring. That's why I want to note this, but I don't think it's worth changing.

PR #7609, opened 2023-11-29
mpk: add an example testing the memory limits

abrown:
This change adds an example program showing:
- how to build a pool-allocated engine that uses MPK
- the difference in the number of memory slots when creating an engine with MPK versus not

The output could look like:

```console
$ cargo run --example mpk
without MPK:    14591 memory slots      85.5 TiB reserved
with MPK:       218861 memory slots     85.5 TiB reserved
                14.999x more slots per reserved memory
```

The example program probes exponentially larger memory pools to find the maximum available on the current system. This search is highly dependent on the memory available, the OS limits on VMAs, and various engine configuration parameters. See the example docs for more details.

### DIFF
--- a/examples/mpk.rs
+++ b/examples/mpk.rs
@@ -1,0 +1,229 @@
+//! This example demonstrates:
+//! - how to enable memory protection keys (MPK) in a Wasmtime embedding (see
+//!   [`build_engine`])
+//! - the expected memory compression from using MPK: it will probe the system
+//!   by creating larger and larger memory pools until system memory is
+//!   exhausted (see [`probe_engine_size`]). Then, it prints a comparison of the
+//!   memory used in both the MPK enabled and MPK disabled configurations.
+//!
+//! You can execute this example with:
+//!
+//! ```console
+//! $ cargo run --example mpk
+//! ```
+//!
+//! Append `-- --help` for details about the configuring the memory size of the
+//! pool. Also, to inspect interesting configuration values used for
+//! constructing the pool, turn on logging:
+//!
+//! ```console
+//! $ RUST_LOG=debug cargo run --example mpk -- --memory-size 512MiB
+//! ```
+//!
+//! Note that OS limits on the number of virtual memory areas (VMAs) can
+//! significantly restrict the total number MPK-striped memory slots; each
+//! MPK-protected slot ends up using a new VMA entry. On Linux, one can raise
+//! this limit:
+//!
+//! ```console
+//! $ sysctl vm.max_map_count
+//! 65530
+//! $ sysctl vm.max_map_count=$LARGER_LIMIT
+//! ```
+
+use anyhow::{anyhow, Result};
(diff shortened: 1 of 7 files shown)


abrown:
`cargo vet --locked` should fail here with output like:

```
$ cargo vet --locked
Vetting Failed!

14 unvetted dependencies:
  bindgen:0.68.1 missing ["safe-to-run"]
  bytesize:1.3.0 missing ["safe-to-run"]
  cexpr:0.6.0 missing ["safe-to-run"]
  clang-sys:1.6.1 missing ["safe-to-run"]
  lazycell:1.3.0 missing ["safe-to-run"]
  libproc:0.14.2 missing ["safe-to-run"]
  mach2:0.4.1 missing ["safe-to-run"]
  minimal-lexical:0.2.1 missing ["safe-to-run"]
  nom:7.1.3 missing ["safe-to-run"]
  peeking_take_while:0.1.2 missing ["safe-to-run"]
  prettyplease:0.2.15 missing ["safe-to-run"]
  proc-maps:0.3.2 missing ["safe-to-run"]
  shlex:1.2.0 missing ["safe-to-run"]
  syn:2.0.32 missing ["safe-to-deploy"]
```

How do we want to handle this? The two added dependencies are `dev-dependencies`; should they require a vet? And if so, which kind?

alexcrichton:
The `bytesize` crate looks vet-able relatively easily but looks like `proc-maps` is a pretty heavyweight dependency pulling in quite a lot. Could that dependency be dropped in favor of parsing `/proc/self/maps` manually?

examples/mpk.rs
```diff
@@ -0,0 +1,229 @@
...
+//! constructing the pool, turn on logging:
+//!
+//! ```console
+//! $ RUST_LOG=debug cargo run --example mpk -- --memory-size 512MiB
+//! ```
+//!
+//! Note that OS limits on the number of virtual memory areas (VMAs) can
+//! significantly restrict the total number MPK-striped memory slots; each
+//! MPK-protected slot ends up using a new VMA entry. On Linux, one can raise
+//! this limit:
```
alexcrichton, line 26: You may want to note here that this documentation only applies to Linux. You may also want to note that MPK only works on Linux too here (in case anyone lands here first)

examples/mpk.rs
```diff
@@ -0,0 +1,229 @@
...
+
+use anyhow::{anyhow, Result};
+use bytesize::ByteSize;
+use clap::Parser;
+use log::{info, warn};
+use std::str::FromStr;
+use wasmtime::*;
+
+fn main() -> Result<()> {
+    drop(env_logger::try_init());
```
alexcrichton, line 42: Since this is the start of the program this can actually be `env_logger::init()`

alexcrichton (approved):
Neat! I like having little example programs like this.

alexcrichton (approved):

PR #7621, opened 2023-11-30
winch(fuzzing): Tighten fuzzing criteria

saulecabrera:
This commit tightens the fuzzing criteria for Winch. The previous implementation only accounted for unsupported instructions. However, unsupported types can also cause the fuzzer to crash.

Winch currently doesn't support `v128` and most of the `Ref` types.

### DIFF
--- a/fuzz/fuzz_targets/differential.rs
+++ b/fuzz/fuzz_targets/differential.rs
@@ -5,6 +5,7 @@ use libfuzzer_sys::fuzz_target;
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering::SeqCst;
 use std::sync::Once;
+use wasmparser::ValType;
 use wasmtime_fuzzing::generators::CompilerStrategy;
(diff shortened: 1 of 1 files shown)


alexcrichton:
Could this perhaps be done by disabling more wasm-smith features [around here](https://github.com/bytecodealliance/wasmtime/blob/08669427b5835282b23fdb96210a9b82aecabbb3/fuzz/fuzz_targets/differential.rs#L87) such as `simd`, `relaxed_simd`, and `reference_types`?

saulecabrera:
Indeed, I initially wanted to do something along those lines, but Winch partially supports reference types (`funcref`, `table.size`, `table.grow`, `table.fill`, etc), for which the fuzzer already found bugs (!) so I wouldn't want to disable that proposal entirely, with that in mind even if we disabled some proposals like simd via wasm-smith I'd still need some check like the one included here. So that's why I thought including a check for all types made sense. But I'm good disabling simd and relaxed simd via wasm-smith and keeping the check for reference types. What do you think?

PR #7622, opened 2023-11-30
mpk: optimize layout of protected stripes, again

abrown:
This is another attempt at #7603, attempting reduce the slab layout sizes of MPK-protected stripes. While experimenting with the limits of MPK-protected memory pools, @alexcrichton and I discovered that the current slab layout calculations were too conservative. This meant that the memory pool could not pack in as many memories as it should have been able: we were expecting, but not seeing, ~15x more memory slots over non-MPK memory pools.

This change brings together several fixes:
- it more aggressively divides up the stripes (as in b212152)
- it eliminates an extra stripe (as in 8813a30)
- it replaces some uses of `checked_*` with `saturating_*` (as in fb22a20)
- it improves some documentation
- and, crucially, it reports back a larger value for `memory_and_guard_size`

The failures observed with #7603 when run with MPK (`WASMTIME_TEST_FORCE_MPK=1 cargo test`) were due to `Store::wasm_fault` not being able to identify which memory an OOB address belonged to. This is because the `MemoryPool` was underreporting the size of the region in which OOB accesses would fault. The correct value is provided by the new `SlabLayout::bytes_to_next_stripe_slot`: any OOB access within that larger region must fault because (1) the other stripes have different protection keys and (2) a `Store` can only use one protection key. We also use (2) to guarantee that `Store::wasm_fault` will be able to calculate the Wasm address from the raw address.

This change also provides a new `traps` test that will reproduce the failures from #7603; if we observe `SIGABRT` from that test, it will be a regression.

### DIFF
--- a/crates/runtime/proptest-regressions/instance/allocator/pooling/memory_pool.txt
+++ b/crates/runtime/proptest-regressions/instance/allocator/pooling/memory_pool.txt
@@ -6,3 +6,4 @@
 # everyone who runs the test benefits from these saved cases.
 cc 696808084287d5d58b85c60c4720227ab4dd83ada7be6841a67162023aaf4914 # shrinks to c = SlabConstraints { max_memory_bytes: 0, num_memory_slots: 1, num_pkeys_available: 0, guard_bytes: 9223372036854775808 }
 cc cf9f6c36659f7f56ed8ea646e8c699cbf46708cef6911cdd376418ad69ea1388 # shrinks to c = SlabConstraints { max_memory_bytes: 14161452635954640438, num_memory_slots: 0, num_pkeys_available: 0, guard_bytes: 4285291437754911178 }
+cc 58f42405c4fbfb4b464950f372995d4d08a77d6884335e38c2e68d590cacf7d8 # shrinks to c = SlabConstraints { expected_slot_bytes: 0, max_memory_bytes: 8483846582232735745, num_slots: 0, num_pkeys_available: 0, guard_bytes: 0, guard_before_slots: false }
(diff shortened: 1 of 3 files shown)


abrown:
I manually checked that the following pass:
- `cargo test --package wasmtime-runtime --features pooling-allocator` passes, which means the proptested invariants still hold
- `WASMTIME_TEST_FORCE_MPK=1 cargo test` passes, which means that the subset of tests (e.g., spec tests) that we force to use MPK work on my MPK-enabled machine

Running the example program from #7609, I see expected improvements in the number of slots:

```console
$ cargo run --example mpk
without MPK:    14713 memory slots      86.2 TiB reserved
with MPK:       220685 memory slots     86.2 TiB reserved
                14.999x more slots per reserved memory
```

alexcrichton (approved):

abrown:
[This](https://github.com/bytecodealliance/wasmtime/actions/runs/7062145323/job/19225342835#step:18:2446) is kind of interesting: arm64 is miscomputing the `WasmFault` address by 8.

```
memory fault at wasm address 0xdeadbee8 in linear memory of size 0x10000
```

It should be `0xdeadbeef`, right?

alexcrichton:
Ah that's expected that some platforms report a different faulting address rounded to some boundary. s390x will round it to a page boundary I think.

There's a [similar test to what you're doing here](https://github.com/bytecodealliance/wasmtime/blob/338653878d89afb5fab99d2e1aa8575aac21feea/tests/all/traps.rs#L1331-L1365) which should be ok to copy

PR #7625, opened 2023-12-02
Add `fields.has` to query if a header is present in a `fields`

elliottt:
As discussed in https://github.com/WebAssembly/wasi-http/issues/82, it would be nice to be able to know when a header value exists, vs when it's empty.

### DIFF
--- a/crates/wasi-http/wit/deps/http/types.wit
+++ b/crates/wasi-http/wit/deps/http/types.wit
@@ -330,14 +330,14 @@ interface types {
 
     /// Set the timeout for the initial connect to the HTTP Server. An error
     /// return value indicates that this timeout is not supported.
-    set-connect-timeout: func(ms: option<duration>) -> result;
+    set-connect-timeout: func(duration: option<duration>) -> result;
 
     /// The timeout for receiving the first byte of the Response body.
     first-byte-timeout: func() -> option<duration>;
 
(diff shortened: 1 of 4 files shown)


alexcrichton (approved):
Looks reasonable to me, but @pchickey should sign-off on this as well

alexcrichton:
Also, should this also be sent upstream to WebAssembly/wasi-http?

elliottt:
Upstream pr: https://github.com/WebAssembly/wasi-http/pull/91

pchickey (approved):

crates/wasi-http/wit/deps/http/types.wit
```diff
@@ -322,14 +330,14 @@ interface types {
 
     /// Set the timeout for the initial connect to the HTTP Server. An error
     /// return value indicates that this timeout is not supported.
-    set-connect-timeout: func(duration: option<duration>) -> result;
+    set-connect-timeout: func(ms: option<duration>) -> result;
```
pchickey, line 333: lets not pull these `ms` names from upstream and instead fix those upstream?

pchickey (approved):

crates/wasi-http/wit/deps/http/types.wit
elliottt: Whoops, thanks for catching this!

PR #7626, opened 2023-12-03
Consolidate platform-specific definitions in Wasmtime

alexcrichton:
Prior to this commit Wasmtime did not not have a style or system for
containing platform-specific logic in files. The goal of this commit is
to consolidate all platform-specific functionality into one location to
make it easier to port Wasmtime to new systems. This commit creates a
`sys` module within the `wasmtime-runtime` crate which conditionally
defines all of the platform support that Wasmtime requires, namely
things related to virtual memory management and trap handling. Many of
the previous `unix.rs` files interspersed throughout the tree are now
all located in a single `unix` directory. This additionally helps split
out `miri`-specific functionality by pretending `miri` is its own
platform.

This change additionally goes through `#[cfg]` directives throughout
`wasmtime-runtime`, `wasmtime-jit`, and `wasmtime` itself to place all
of this target-specific functionality within this `sys` directory. The
end state is that there are two new top-level modules in the
`wasmtime-runtime` crate:

* `arch` - this conditionally defines architecture-specific logic such
  as the state used by backtraces, libcalls, etc.
* `sys` - this conditionally defines OS or platform-specific logic such
  as virtual memory management.

One secondary goal of this commit is to enable the ability to easily
add new platforms to Wasmtime, even if it's in a fork of Wasmtime.
Previously patches might have to touch a good number of locations where
now they'd ideally only have to touch `sys/mod.rs` to declare a new
platform and `sys/$platform/*.rs` to define all the functionality.


crates/runtime/src/cow.rs
```diff
@@ -37,7 +35,7 @@ pub struct MemoryImage {
     /// This might be an mmaped `*.cwasm` file or on Linux it could also be a
     /// `Memfd` as an anonymous file in memory. In either case this is used as
     /// the backing-source for the CoW image.
-    fd: FdSource,
+    fd: MemoryImageSource,
```
abrown, line 38:
```suggestion
    source: MemoryImageSource,
```
Maybe? Since it seems you're doing clarity refactors in this PR...

crates/runtime/src/sys/miri/vm.rs
```diff
@@ -0,0 +1,63 @@
+use std::fs::File;
+use std::io;
+use std::sync::Arc;
+
+pub unsafe fn expose_exisiting_mapping(ptr: *mut u8, len: usize) -> io::Result<()> {
```
abrown, line 5:
```suggestion
pub unsafe fn expose_existing_mapping(ptr: *mut u8, len: usize) -> io::Result<()> {
```

crates/runtime/src/sys/miri/traphandlers.rs
```diff
@@ -0,0 +1,42 @@
+// With MIRI set up just enough of a setjmp/longjmp with catching panics
```
abrown, line 1:
```suggestion
// For MIRI, set up just enough of a setjmp/longjmp with catching panics
```

crates/runtime/src/sys/mod.rs
```diff
@@ -0,0 +1,30 @@
+//! OS-related abstractions required by Wasmtime.
+//!
+//! This module is intended to house all logic that's specific to either Unix
+//! for Windows, for example. The goal of this module is to be the "single
```
abrown, line 4:
```suggestion
//! or Windows, for example. The goal of this module is to be the "single
```

crates/runtime/src/sys/unix/signals.rs
```diff
@@ -15,10 +31,11 @@ static mut PREV_SIGBUS: MaybeUninit<libc::sigaction> = MaybeUninit::uninit();
...
 static mut PREV_SIGFPE: MaybeUninit<libc::sigaction> = MaybeUninit::uninit();
 
-pub unsafe fn platform_init() {
-    if cfg!(miri) {
-        return;
-    }
+pub unsafe fn platform_init(macos_use_mach_ports: bool) {
+    // Either mach ports shouldn't be in use or we shouldn't be on macOS,
+    // otherwise the `machports.rs` module should be used instead.
+    assert!(!macos_use_mach_ports || !cfg!(target_os = "macos"));
```
abrown, line 37: Why not make this a compile-time check instead, then?

crates/runtime/src/arch/x86_64.rs
```diff
@@ -47,6 +88,7 @@ macro_rules! wasm_to_libcall_trampoline {
         );
     };
 }
+pub(crate) use wasm_to_libcall_trampoline;
 
 #[cfg(test)]
 mod wasm_to_libcall_trampoline_offsets_tests {
```
abrown, line 94: I guess a future refactor could move the MPK stuff here as well?

crates/runtime/src/sys/unix/signals.rs
alexcrichton: Could you elaborate? Do you mean asserting that this file isn't used on macos? (it's used on macos intentionally so wanted to clarify)

alexcrichton:
> my take is that none of this refactoring should substantially change any logic, at least observably for users; is that right?

Correct!

crates/runtime/src/sys/unix/signals.rs
abrown: Oh, sorry... I was just looking at the `cfg!` part but clearly `machos_use_mach_ports` is runtime-defined. Never mind!

PR #7627, opened 2023-12-04
Update tutorial to use wasi_snapshot_preview1

matthewp:
- Was going through the tutorial and noticed it used the older (I think?) `wasi_unstable` which no longer updated. So updating here.


alexcrichton (approved):

alexcrichton:
Thanks!

PR #7628, opened 2023-12-04
winch: Rework `br_table` jumps

saulecabrera:
Follow up to: https://github.com/bytecodealliance/wasmtime/pull/7616

The fuzzer has been running for a while now (~2 days) without reporting new issues. 

This commit reworks the `br_table` logic so that it correctly handles all the jumps involved to each of the targets.

Even though it is safe to use the default branch for type information, it is not safe to use it to derive the base stack pointer and base value stack length. Each block might and will probably have, different values as base stack pointer and base value stack length. This change ensures that each target offset is taken into account to balance the value stack prior to each jump.


elliottt (approved):
I think this makes sense, thank you for all the additional fuzzing!

PR #7636, opened 2023-12-05
Add opt patterns for 3-way comparison (`Ord::cmp` or `<=>`)

scottmcm:
Previous zulip conversation: <https://bytecodealliance.zulipchat.com/#narrow/stream/217117-cranelift/topic/3-way.20compare.20.28Rust.20.60cmp.60.2C.20C.2B.2B20.20.60.3C.3D.3E.60.29/near/404519375>

This PR adds a bunch of ISLE opt patterns for three-way comparison on primitives, as exposed by `Ord::cmp` in Rust or as `<=>` "spaceship" in C++20.  It's split in two parts:
1. Detecting nested `select`s that are actually doing this, of which there are sadly many possible forms and no well-known-best way.  By doing this cranelift can emit a much simpler MachInst pattern for it than if it needs to use cmovs and constants.
2. Adding simplifications when common comparisons are done on the *result* of the spaceship.  This is particularly relevant for any code written against a 3-way-comparison interface but that only needs a 2-way result.  For example, if you use `.sort_by(|a, b| b.cmp(a))` in Rust to reverse-sort something, [it immediately just checks for `Ordering::Less`](https://doc.rust-lang.org/1.74.0/src/alloc/slice.rs.html#263-268), and if you have `#[derive(PartialOrd)] struct Foo(u32);` then the generated `<` operator is doing the equivalent of `(a <=> b) < 0`, which this will improve to `a < b`.

### DIFF
--- a/cranelift/codegen/src/opts/spaceship.isle
+++ b/cranelift/codegen/src/opts/spaceship.isle
@@ -14,24 +14,24 @@
                            (iconst ty neg_1)
                            (ne rty x y)))
       (if-let -1 (i64_sextend_imm64 ty neg_1))
-      (sextend_from_i8 ty (isub $I8 (ugt rty x y) (ult rty x y))))
+      (sextend_from_i8 ty (spaceship_u rty x y)))
 (rule (simplify (select ty (ult rty x y)
                            (iconst ty neg_1)
                            (uextend ty (ne rty x y))))
       (if-let -1 (i64_sextend_imm64 ty neg_1))
-      (sextend_from_i8 ty (isub $I8 (ugt rty x y) (ult rty x y))))
+      (sextend_from_i8 ty (spaceship_u rty x y)))
 ;; x < y ? -1 : x <= y ? 0 : +1
 ;; x < y ? -1 : x > y ? +1 : 0
 (rule (simplify (select ty (ult rty x y)
                            (iconst ty neg_1)
                            (ugt rty x y)))
       (if-let -1 (i64_sextend_imm64 ty neg_1))
-      (sextend_from_i8 ty (isub $I8 (ugt rty x y) (ult rty x y))))
+      (sextend_from_i8 ty (spaceship_u rty x y)))
 (rule (simplify (select ty (ult rty x y)
                            (iconst ty neg_1)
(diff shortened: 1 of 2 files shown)


cranelift/codegen/src/opts/spaceship.isle
```diff
@@ -0,0 +1,237 @@
...
+      (sextend_from_i8 ty (isub $I8 (sgt rty x y) (slt rty x y))))
+
+;; Then once we have it normalized, we can apply some basic simplifications.
+;; For example, a derived `PartialOrd::lt` on a newtype in Rust will essentially
+;; emit `(a <=> b) < 0`, and replacing that with `a < b` can really help.
+;; `icmp.isle` prefers comparing against zero so we don't need to worry about
+;; also matching things like `(a <=> b) < 1` or `(a <=> b) <= -1`.
+
+;; (a <=> b) == 0 --> a == b
+(rule (simplify (eq ty (isub ty (sgt rty x y) (slt rty x y)) (iconst ty (u64_from_imm64 0))))
```
alexcrichton, line 202: One possible way you could simplify this is something like:

```
(decl spaceship_s (Value Value) Value)
(extractor (spaceship_s v) (isub _ (sgt rty x y) (slt rty x y)))
```

I forget the precise syntax for `(extractor ...)` but I think it can be used along the lines of a macro-of-sorts to reduce the repetition here. That being said I've found them historically a bit cumbersome to use and I'm not sure you can pattern-match out the `rty` still as well (and maybe `ty` too?) so this may not amount to much after all.

alexcrichton (commented):
This looks great to me, thanks for adding these!

I've left a few comments, although they're mostly just cosmetic. The one about `(extractor)` might be interesting to explore if you're curious but don't take it as a requirement for this PR.

alexcrichton (approved):
Nice I agree it worked out well! Thanks again!

alexcrichton:
Failure looks unrelated to this PR, I'll debug that separately.

alexcrichton:
Ok once https://github.com/bytecodealliance/wasmtime/pull/7651 lands I'll retry landing this

PR #7638, opened 2023-12-05
Discard 0-sized writes to files

### LINKED ISSUE #7633: A difference between windows and linux?
### Test Case
The test.c file is uploaded as 
[test-c.txt](https://github.com/bytecodealliance/wasmtime/files/13560062/test-c.txt)

The test.wasm file is uploaded as 
[test-wasm.txt](https://github.com/bytecodealliance/wasmtime/files/13560063/test-wasm.txt)

### Steps to Reproduce
We create a normal file Data/hello.txt and write some initial content into the file.
using command `wasmtime run --dir=./Data test.wasm` to execute the wasm file.

### Actual Results
In linux, wasmtime do not write any bytes into the file and prints:
```
Successfully get the file fd.
Current offset: 0
Wrote 0 bytes to the file.
File Size: 30 bytes
Leave snapshot.
```

In windows, wasmtime writes null characters (\u0000) into the file and prints:
```
Successfully get the file fd.
Current offset: 0
Wrote 0 bytes to the file.
File Size: 50 bytes
Leave snapshot.
```

There is a difference between executing the wasm file in windows and linux.
Maybe in linux, this could be a bug?

### Versions and Environment
Ubuntu 20.04 (amd64)
Windows 10 (x86_64)
wasmtime 13.0.0

alexcrichton:
This commit comes from #7633 where Windows and Unix would behave differently when writing at a particular file offset. Notably Unix semantics [indicate]:

> Before any action described below is taken, and if nbyte is zero
> and the file is a regular file, the write() function may detect
> and return errors as described below. In the absence of errors,
> or if error detection is not performed, the write() function
> shall return zero and have no other results. If nbyte is zero and
> the file is not a regular file, the results are unspecified.

These semantics are a bit easier to emulate on Windows so the host implementation now discards any attempt to perform I/O if a zero-sized write is detected.

[indicate]: https://man7.org/linux/man-pages/man3/write.3p.html

Closes #7633

### DIFF
--- a/crates/test-programs/src/bin/preview1_file_pread_pwrite.rs
+++ b/crates/test-programs/src/bin/preview1_file_pread_pwrite.rs
@@ -128,6 +128,55 @@ unsafe fn test_file_pread_pwrite(dir_fd: wasi::Fd) {
     wasi::path_unlink_file(dir_fd, "file").expect("removing a file");
 }
 
+unsafe fn test_file_pwrite_and_file_pos(dir_fd: wasi::Fd) {
+    let path = "file2";
+    let file_fd = wasi::path_open(
+        dir_fd,
+        0,
+        path,
+        wasi::OFLAGS_CREAT,
+        wasi::RIGHTS_FD_READ | wasi::RIGHTS_FD_WRITE,
+        0,
+        0,
+    )
+    .expect("opening a file");
+    assert!(
+        file_fd > libc::STDERR_FILENO as wasi::Fd,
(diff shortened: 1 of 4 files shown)


pchickey (approved):
Seems like a good fix to me. Thanks

alexcrichton:
Took me quite some time to set aside time to debug this failure on Windows, and it turned up https://github.com/bytecodealliance/system-interface/pull/43

PR #7645, opened 2023-12-06
Better error message for wrong import implementation

rylev:
This changes improve the error message when a component is importing/exporting something that the linker is either missing or has the wrong type.

Previously the error message for a missing function implementation would be:

```
Error: import `my-import` has the wrong type

Caused by:
    expected func found nothing
```

Now, this error will be:

```
Error: component imports `my-import`, but a correct implementation of `my-import` could not be found

Caused by:
    function implementation is missing
```

I went back and forth on whether the error message should reference the linker. For direct users of `Linker` this would surely help, but this error message also gets displayed to indirect users of `Linker` so talking at a higher level of abstraction seemed appropriate.

### DIFF
--- a/crates/environ/src/component/types.rs
+++ b/crates/environ/src/component/types.rs
@@ -1003,7 +1003,7 @@ pub enum TypeDef {
     ComponentInstance(TypeComponentInstanceIndex),
     /// A component function, not to be confused with a core wasm function.
     ComponentFunc(TypeFuncIndex),
-    /// An interface type.
+    /// An type in an interface.
     Interface(InterfaceType),
     /// A core wasm module and its type.
     Module(TypeModuleIndex),
@@ -1014,6 +1014,21 @@ pub enum TypeDef {
     /// Note that different resource tables may point to the same underlying
     /// actual resource type, but that's a private detail.
     Resource(TypeResourceTableIndex),
+}
+
+impl TypeDef {
+    /// A human readable description of what kind of type definition this is.
+    pub fn desc(&self) -> &str {
+        match self {
+            TypeDef::Component(_) => "component",
+            TypeDef::ComponentInstance(_) => "instance",
+            TypeDef::ComponentFunc(_) => "function",
+            TypeDef::Interface(_) => "type",
+            TypeDef::Module(_) => "core module",
+            TypeDef::CoreFunc(_) => "core function",
+            TypeDef::Resource(_) => "resource",
+        }
+    }
 }
 
 // NB: Note that maps below are stored as an `IndexMap` now but the order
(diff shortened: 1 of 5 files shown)


crates/wasmtime/src/component/linker.rs
```diff
@@ -161,8 +161,9 @@ impl<T> Linker<T> {
                 .strings
                 .lookup(name)
                 .and_then(|name| self.map.get(&name));
-            cx.definition(ty, import)
-                .with_context(|| format!("import `{name}` has the wrong type"))?;
+            cx.definition(ty, import).with_context(|| {
+                format!("component imports `{name}`, but a correct implementation of `{name}` could not be found")
```
peterhuene, line 165:
```suggestion
                format!("component imports {desc} `{name}`, but a matching implementation was not found", desc = ty.desc())
```

With an additional implementation of `TypeDef::desc` to say what it is, e.g. "instance", "type", "function", etc.

crates/wasmtime/src/component/matching.rs
```diff
@@ -36,19 +36,27 @@ impl TypeChecker<'_> {
         match *expected {
             TypeDef::Module(t) => match actual {
                 Some(Definition::Module(actual)) => self.module(&self.types[t], actual),
-                _ => bail!("expected module found {}", desc(actual)),
+                Some(actual) => bail!("expected module found {}", actual.desc()),
+                None => bail!("component implementation is missing"),
```
peterhuene, line 40:
```suggestion
                None => bail!("module implementation is missing"),
```

crates/wasmtime/src/component/matching.rs
```diff
@@ -36,19 +36,27 @@ impl TypeChecker<'_> {
...
+                Some(actual) => bail!("expected function found {}", actual.desc()),
+                None => bail!("function implementation is missing"),
+            },
+            TypeDef::Component(_) => match actual {
+                Some(actual) => bail!("expected component found {}", actual.desc()),
+                None => bail!("component implementation is missing"),
+            },
+            TypeDef::Interface(_) => match actual {
+                Some(actual) => bail!("expected interface found {}", actual.desc()),
+                None => bail!("interface implementation is missing"),
```
peterhuene, line 58: Unfortunately `TypeDef::Interface` is misnamed, it's actually representing a type so the message should indicate a type.

If it's not too invasive, I think we should probably change `Interface` to `Type` too.

peterhuene (commented):
I think what you have is definitely trending in the right direction for an easier to understand error message.

Just a few feedback comments.

Additionally, "implementation" in some of the errors rubs me the wrong way for some reason I can't seem to articulate, nor can I think of a better word to use.

crates/wasmtime/src/component/matching.rs
peterhuene: Actually, probably not worth taking a breaking change to rename the enum variant (I think the only external consumer of it is `jco`, but there could be others); let's leave it `Interface` and just change the error message here back to indicate a "type".

rylev:
@peterhuene I believe I've addressed your feedback. I agree that "implementation" isn't always the right word, but I think we'll need to let this sit for a while before the right word makes itself apparent.

peterhuene (approved):

PR #7648, opened 2023-12-06
Ensure `remote-address` is allowed when creating UDP stream

rylev:
Until now, the `remote-address` argument to `wasi:sockets/udp-socket#stream` was not checked against the `Pool` of allowed IP addresses. Now, we store the `Pool` from `network` into the `socket` object in `wasi:sockets/udp-socket#bind` and later use that `Pool` in `wasi:sockets/udp-socket#stream` to check that `remote-address` is allowed.

The check itself is a bit hacky since `Pool` doesn't seem to expose a way to just check an address so we use `PoolExt` to create a `UdpConnecter` which we will fail if we don't have permission to connect to the remote address.

I can add a test once I figure out the testing story in https://github.com/bytecodealliance/wasmtime/pull/7647

### DIFF
--- a/crates/wasi/src/preview2/host/udp.rs
+++ b/crates/wasi/src/preview2/host/udp.rs
@@ -43,7 +43,8 @@ impl<T: WasiView> udp::HostUdpSocket for T {
         }
 
         // Set the pool on the socket so later functions have access to it through the socket handle
-        table.get_mut(&this)?.pool = table.get(&network)?.pool.clone();
+        let pool = table.get(&network)?.pool.clone();
+        table.get_mut(&this)?.pool.replace(pool.clone());
 
         let socket = table.get(&this)?;
         let local_address: SocketAddr = local_address.into();
@@ -51,7 +52,7 @@ impl<T: WasiView> udp::HostUdpSocket for T {
         util::validate_address_family(&local_address, &socket.family)?;
 
         {
-            let binder = socket.pool.udp_binder(local_address)?;
+            let binder = pool.udp_binder(local_address)?;
             let udp_socket = &*socket
                 .udp_socket()
                 .as_socketlike_view::<cap_std::net::UdpSocket>();
@@ -131,10 +132,13 @@ impl<T: WasiView> udp::HostUdpSocket for T {
 
         // Step #2: (Re)connect
         if let Some(connect_addr) = remote_address {
+            let Some(pool) = socket.pool.as_ref() else {
+                return Err(ErrorCode::InvalidState.into());
+            };
             util::validate_remote_address(&connect_addr)?;
             util::validate_address_family(&connect_addr, &socket.family)?;
             // We don't actually use the connecter, we just use it to verify that `connect_addr` is allowed
-            let _ = socket.pool.udp_connecter(connect_addr)?;
+            let _ = pool.udp_connecter(connect_addr)?;
 
             rustix::net::connect(socket.udp_socket(), &connect_addr).map_err(
                 |error| match error {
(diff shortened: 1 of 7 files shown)


crates/wasi/src/preview2/host/udp.rs
```diff
@@ -34,8 +34,10 @@ impl<T: WasiView> udp::HostUdpSocket for T {
         local_address: IpSocketAddress,
     ) -> SocketResult<()> {
         let table = self.table_mut();
+        // Set the pool on the socket so later functions have access to it through the socket handle
+        table.get_mut(&this)?.pool = table.get(&network)?.pool.clone();
```
alexcrichton, line 38: Could this be done after the state validation below? That'd help prevent odd situations such as calling this multiple times just to switch the network while having the method return an error.

badeend: @rylev  Don't know if this was your intention, but this creates a full copy of the entire pool for each socket.

crates/wasi/src/preview2/host/udp.rs
```diff
@@ -129,6 +131,8 @@ impl<T: WasiView> udp::HostUdpSocket for T {
         if let Some(connect_addr) = remote_address {
             util::validate_remote_address(&connect_addr)?;
             util::validate_address_family(&connect_addr, &socket.family)?;
+            // We don't actually use the connecter, we just use it to verify that `connect_addr` is allowed
+            let _ = socket.pool.udp_connecter(connect_addr)?;
```
badeend, line 135: Might not be the prettiest solution, but: could we use this same workaround to validate the remote address in unconnected outgoing datagram streams?

rylev:
I've addressed the PR feedback:
* Pool is now wrapped in an `Arc` to make cloning cheaper.
* `HostOutgoingDatagramStream` will check the provided addr against the `Pool` before sending a datagram.

@alexcrichton I'm not sure what the testing story should look like here. I don't see where the wasi implementations are being tested against non-default `WasiCtx`s. I believe all tests are running with full network. I suppose whatever solution we decide on here should also work for #7647 as well.

crates/wasi/src/preview2/udp.rs
```diff
@@ -42,6 +43,9 @@ pub struct UdpSocket {
 
     /// Socket address family.
     pub(crate) family: SocketAddressFamily,
+
+    /// The pool of allowed address
+    pub(crate) pool: Arc<Pool>,
```
alexcrichton, line 48: Could this start as `None` to represent that it's bound-late? That could also help overwriting it by accident by only writing to it if it's `None`

alexcrichton: (same below too)

alexcrichton (approved):

crates/wasi/src/preview2/udp.rs
rylev: Pushed a change that does this, but I have a question about the exact desired semantics of `start-bind`. Take the following_example:

```rust
// `bad_addr` is an address that is not permitted
assert!(sock.start_bind(bad_addr).is_err());
// `good_addr` is permitted
sock.start_bind(good_addr);
```

As currently written the second call to `start-bind` will fail with `InvalidState` since the pool has already been set on the socket. Is this right? Previously, this would have succeeded. I would imagine that `start_bind` should be allowed to be called multiple times as long as the socket has not successfully been bound before.

rylev: I removed the check that the pool was not previously set as this broke some tests. Once we decide on what the semantics should be, I can adjust.

alexcrichton: cc @badeend on this you are probably best-equipped to answer this

alexcrichton:
I'm gonna go ahead and queue this for merge, and @rylev says they're happy to follow-up on semantic changes in a follow-up.

crates/wasi/src/preview2/udp.rs
badeend: @rylev 

From the API's perspective, unbound sockets (`Default`) never need a pool, and all other states (`BindStarted`, `Bound`, `Connected`) could've only be reached by calling `bind` at some point, so they always have access to a pool. A way to codify this could be:

```rs
pub(crate) enum UdpState {
    Default,
    BindStarted(Arc<Pool>),
    Bound(Arc<Pool>),
    Connected(Arc<Pool>),
}
```

Also, OutgoingDatagramStream's `pool` field doesn't have to be optional, because you can never call `stream` on an unbound socket.

---

> As currently written the second call to start-bind will fail with InvalidState since the pool has already been set on the socket. Is this right?

From the caller's perspective, the `start-bind` call should preferably succeed or fail atomically. Using the enum layout above would _automagically_ fix that.

PR #7649, opened 2023-12-06
Use the compilation cache for components

elliottt:
Cache whole components in the compilation cache. This doesn't address the situation where we could cache individual modules used within the component separately, and reuse those results across multiple components. However, it does improve the runtime of command-line uses of components today, and anecdotally people are benchmarking `wasmtime run` with components.


alexcrichton (approved):
Thanks!

PR #7651, opened 2023-12-06
Increase stack size in `examples/threads.c`

alexcrichton:
This commit is an attempt to address the CI failure popping up in #7636. I can reproduce the failure locally and it appears to be due to the fact that macOS is giving spawned threads via `pthread_create` a 512K stack by default. Cranelift when compiled in debug mode is taking more stack than this (but working in release mode). I was talking with Trevor and Jamey about possible ways to reduce the stack size here but for now I'm going with Jamey's suggestion of increasing the stack size as the best course of action for now.

### DIFF
--- a/examples/threads.c
+++ b/examples/threads.c
@@ -178,7 +178,14 @@ int main(int argc, const char *argv[]) {
     args->engine = engine;
     args->module = shared;
     printf("Initializing thread %d...\n", i);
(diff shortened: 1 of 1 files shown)


elliottt (approved):
Looks good to me! It's also nice that we'll use 2MB across all platforms for this test now, instead of allowing for varying defaults.

PR #7654, opened 2023-12-07
Cranelift: More simplification opts for `iadd`/`isub`

scottmcm:
This PR came out of me seeing that the opt patterns were written out to handle both
```
(!x) + 1 == !(x) - (-1) == ineg(x)
```
and thus being sad that every pattern including `(iadd _ 1)` needed to be written again for `(isub _ -1)`.

So this takes a look at more opt patterns around `iadd` and `isub`:
- Various combinations with `ineg`, like `x + -y == -y + x == -(y - x) == x - y`
- Avoiding subtract-a-"negative"-constant, so `(isub x 255_u8)` becomes `(iadd x 1_u8)` (but `(isub x 1)` is left alone)
- And removing now-unnecessary patterns in arithmetic.isle and bitops.isle, like I did in #7599 

I originally had `x + -1` → `x - 1` as well, but that regressed codegen with addressing modes -- something that had been in one x64 LEA ended up being a LEA and a SUB.  So there's an interesting question about whether to turn *all* `(isub x CONST)` into `(iadd x CONST)`, but that felt like something that would need a zulip conversation, and thus it's not included in this hopefully-simpler PR.

### DIFF
--- a/cranelift/codegen/src/opts/arithmetic.isle
+++ b/cranelift/codegen/src/opts/arithmetic.isle
@@ -20,6 +20,17 @@
                       x))
       (ineg ty x))
 
+;; x + -y == -y + x == -(y - x) == x - y
+(rule (simplify (iadd ty x (ineg ty y)))
+      (isub ty x y))
+(rule (simplify (iadd ty (ineg ty y) x))
+      (isub ty x y))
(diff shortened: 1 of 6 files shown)


alexcrichton (approved):
Thanks!

PR #7656, opened 2023-12-08
[bindgen] Include Version in the export name if needed

jsturtevant:
When exporting multiple versions of the same package the version name needs to be included.

Before this change the generator would produce:

```
pub fn test_dep_test(&self) -> &exports::test::dep0_1_0::test::Test { &self.interface1 } 
pub fn test_dep_test(&self) -> &exports::test::dep0_2_0::test::Test { &self.interface2 }
```

now it will produce: 

```
pub fn test_dep_0_1_0_test(&self) -> &exports::test::dep0_1_0::test::Test { &self.interface1 } 
pub fn test_dep_0_2_0_test(&self) -> &exports::test::dep0_2_0::test::Test { &self.interface2 }
```

Found in https://github.com/bytecodealliance/wit-bindgen/pull/787/files#r1416724258
Discussed in https://bytecodealliance.zulipchat.com/#narrow/stream/327223-wit-bindgen/topic/Host.20Exports.20across.20versions

### DIFF
--- a/crates/wit-bindgen/src/lib.rs
+++ b/crates/wit-bindgen/src/lib.rs
@@ -474,7 +474,7 @@ impl Wasmtime {
                         format!(
                             "{}_{}_{snake}",
                             pkgname.namespace.to_snake_case(),
-                            pkgname.name.to_snake_case()
+                            self.name_package_module(resolve, iface.package.unwrap())
                         ),
                     ),
                     None => (format!("exports::{snake}::{camel}"), snake.clone()),
--- a/crates/component-macro/tests/codegen/multiversion/root.wit
+++ b/crates/component-macro/tests/codegen/multiversion/root.wit
@@ -3,5 +3,6 @@ package foo:bar;
 world foo {
   import my:dep/a@0.1.0;
   import my:dep/a@0.2.0;
+  export my:dep/a@0.1.0;
   export my:dep/a@0.2.0;
 }


crates/wit-bindgen/src/lib.rs
```diff
@@ -465,18 +465,40 @@ impl Wasmtime {
...
+                            "{}_{}_",
+                            pkgname.namespace.to_snake_case(),
+                            pkgname.name.to_snake_case()
+                        );
+
+                        let version_count = resolve
+                            .package_names
+                            .keys()
+                            .filter(|x| x.name == pkgname.name && x.namespace == pkgname.namespace)
+                            .count();
```
jsturtevant, line 479: I don't think this is exactly right. Was having a hard time mapping interfaces to packages in an effective way but this proved to be a way to get it to build.

looking for betters ways...

alexcrichton: Coudl this perhaps go through [`name_package_module`](https://github.com/bytecodealliance/wasmtime/blob/62c143e1859945a888fe28d8534ffe0b3d022cc7/crates/wit-bindgen/src/lib.rs#L209-L252) to leverage the preexisting logic to drop the version?

alexcrichton (commented):
Thanks! Could you clarify though? You mention that CI is failing but it looks passing here. This additionally implements what I would expect which is that versions are dropped unless they're required, so this looks good to go to me modulo a suggestion below to use a preexisting helper, but I wanted to confirm I wasn't missing anything.

alexcrichton:
Personally at least, yes, I think all bindings generators should drop versions unless there's an ambiguity. Helps promotes the use of versions I think without causing all the generated bindings to look terrible ideally. I haven't gotten around to implementing this much elsewhere other than the Wasmtime host and Rust guest generators though.

crates/wit-bindgen/src/lib.rs
jsturtevant: That was exactly what I needed and it was right in front of me 😆

PR #7658, opened 2023-12-08
Replace clap attributes with command and arg

iawia002:
While reading the code, I noticed that we were mixing clap's `clap`, `command`, `arg`, and `structopt` attributes, and it seemed quite messy, and even though `#[clap()]` still works, `#[arg()]` and `#[command()]` are now recommended since clap v4, ref https://github.com/clap-rs/clap/pull/4180.

So this patch standardizes them, eliminating the use of outdated `clap` attributes and removing some meaningless name configurations.

### DIFF
--- a/src/commands/run.rs
+++ b/src/commands/run.rs
@@ -55,9 +55,8 @@ fn parse_preloads(s: &str) -> Result<(String, PathBuf)> {
 
 /// Runs a WebAssembly module
 #[derive(Parser, PartialEq)]
-#[structopt(name = "run")]
 pub struct RunCommand {
-    #[clap(flatten)]
+    #[command(flatten)]
     #[allow(missing_docs)]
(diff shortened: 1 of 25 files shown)


src/commands/run.rs
```diff
@@ -55,9 +55,8 @@ fn parse_preloads(s: &str) -> Result<(String, PathBuf)> {
 
 /// Runs a WebAssembly module
 #[derive(Parser, PartialEq)]
-#[structopt(name = "run")]
```
iawia002, line 58: This actually won't work, if we want to change the name, we need to config it in https://github.com/bytecodealliance/wasmtime/blob/3b61c511a0b152d2bb40343b87b5fb1e70a87564/src/bin/wasmtime.rs#L49C1-L51

PR #7661, opened 2023-12-08
Reserve handle index 0 in the component model

alexcrichton:
This commit updates the allocation scheme for resources in the component model to start at 1 instead of 0 when communicating with components. This is an implementation of WebAssembly/component-model#284.

While this broke a number of tests we have this shouldn't actually break any components in practice. The broken tests were all overly-precise in their assertions and error messages and this shouldn't idiomatically come up in any guest language, so this should not be a practically breaking change.

### DIFF
--- a/crates/runtime/src/component/resources.rs
+++ b/crates/runtime/src/component/resources.rs
@@ -28,6 +28,11 @@ use std::mem;
 use wasmtime_environ::component::TypeResourceTableIndex;
 use wasmtime_environ::PrimaryMap;
 
+/// The maximum handle value is specified in
+/// <https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md>
+/// currently and keeps the upper bit free for use in the component.
+const MAX_RESOURCE_HANDLE: u32 = 1 << 30;
+
 /// Contextual state necessary to perform resource-related operations.
 ///
 /// This state a bit odd since it has a few optional bits, but the idea is that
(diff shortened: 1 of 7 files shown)


crates/runtime/src/component/resources.rs
```diff
@@ -294,36 +298,42 @@ impl ResourceTable {
...
             _ => unreachable!(),
         };
-        u32::try_from(ret).unwrap()
+
+        // The component model reserves index 0 as never allocatable so add one
+        // to the table index to start the numbering at 1 instead. Also note
+        // that the component model places an upper-limit per-table on the
+        // maximum allowed index.
+        let ret = ret + 1;
+        if ret >= 1 << 30 {
```
rylev, line 307: Nit: can the upper limit be moved to a constant and perhaps link to the spec where it is specified?

crates/runtime/src/component/resources.rs
```diff
@@ -294,36 +298,42 @@ impl ResourceTable {
...
+        }
+        Ok(ret)
     }
 
     fn rep(&self, idx: u32) -> Result<u32> {
-        match usize::try_from(idx).ok().and_then(|i| self.slots.get(i)) {
+        // NB: `idx` is decremented by one to account for the `+1` above during
+        // allocation.
+        let table_index = idx.checked_sub(1).and_then(|i| usize::try_from(i).ok());
+        match table_index.and_then(|i| self.slots.get(i)) {
```
rylev, line 317: The fetching of a slot is finicky enough that I wonder if it makes sense to move it to its own function so if there is ever another need for that logic the chances of bad copy/paste can be minimized.

alexcrichton:
Oops it appears that I forgot about this, so I've now rebased this and this should be ready again after addressing @rylev's comments.

pchickey (approved):
Thanks, I also forgot all about this

PR #7670, opened 2023-12-09
Cranelift: Add iconst shorthand to simplify ISLE opts

scottmcm:
Extracting an `iconst` with value `1` today is easy enough, albeit needing more type-conversions than one might like

https://github.com/bytecodealliance/wasmtime/blob/4f2d634ca4291a09003eaba26f989cd544c1a289/cranelift/codegen/src/opts/arithmetic.isle#L52-L56

However there's no way to *extract* against a type-aware `-1`, meaning those patterns have an extra step

https://github.com/bytecodealliance/wasmtime/blob/4f2d634ca4291a09003eaba26f989cd544c1a289/cranelift/codegen/src/opts/arithmetic.isle#L64-L67

This PR adds `iconst_s ty c` that can directly extract a sign-extended value as well as construct a properly masked constant (so `iconst_s ty -1` will give `iconst.i8 255` or `iconst.i16 0xFFFF` or …).  And for symmetry, it also adds `iconst_u`, which also works as both constructor and extractor.

Thus the pattern looking for `-1` is simplified to just
```lisp
;; x*-1 == ineg(x).
(rule (simplify (imul ty x (iconst_s ty -1)))
      (ineg ty x))
```

And even unsigned cases are shortened and simplified by no longer needing to remember to get `u64_from_imm64` vs `imm64` correct depending whether something is extracting or constructing, since `iconst_u` works for both
```diff
 ;; uge(x, 0) == true.
-(rule (simplify (uge (fits_in_64 (ty_int bty)) x zero @ (iconst _ (u64_from_imm64 0))))
-      (subsume (iconst bty (imm64 1))))
+(rule (simplify (uge (fits_in_64 (ty_int bty)) x zero @ (iconst_u _ 0)))
+      (subsume (iconst_u bty 1)))
```

This makes no functional changes to the patterns, just makes them shorter using the new etor+ctors, and thus there are no test changes.  (There's way too many changes in the ISLE for it to be a good idea to try to also change any functionality here.  Similarly, to attempt to ease review, I did no formatting changes in the ISLE, even where I might have otherwise removed some line-breaking.)

---

This also moves `simm32` and `uimm8` to lowering ISLE only, as they appear not actually useful for *opts* ISLE.

In particular, it's very tempting to write
```lisp
;; x*-1 == ineg(x).
(rule (simplify (imul ty x (iconst _ (simm32 -1))))
      (ineg ty x))
```
but despite its name, **that doesn't even work for 32-bit numbers** because of how `simm32` is implemented.


cranelift/codegen/src/opts.rs
```diff
@@ -149,6 +149,27 @@ impl<'a, 'b, 'c> generated_code::Context for IsleContext<'a, 'b, 'c> {
...
+        let inst_def = &self.ctx.func.dfg.insts[inst];
+        let InstructionData::UnaryImm {
+            opcode: Opcode::Iconst,
+            imm,
+        } = inst_def
+        else {
+            return None;
+        };
+        Some((ty, self.i64_sextend_imm64(ty, *imm)))
+    }
```
alexcrichton, line 171: I think this might actually be able to get replicated into ISLE with `(extractor iconst_s (iconst ty (i64_sextend_imm64 n)))`, although I forget the precise syntax. Effectively though it should be possible to build up this new extractor based on previous extractors I think?

alexcrichton (commented):
Thanks for pushing on that! You're reaching my own personal limit of knowledge of extractors and ISLE as well as I've run into the same things myself. 

@fitzgen would you be ok taking a look at this too? You're more familiar with ISLE than I and you might spot something I'm missing.

fitzgen (approved):
Fantastic! We've wanted to do this kind of thing for a long time (couple constant values to their type and signedness) but no one ever had time or stepped up. Thanks for doing this!

PR #7674, opened 2023-12-12
ci: Update to `actions/checkout@v4` from `v3`.

waywardmonkeys:
This mainly updates to use Node 20 rather than Node 16 internally.


alexcrichton (approved):
Thanks!

PR #7677, opened 2023-12-12
Promote wasmtime-wasi preview2 implementation to root of crate

pchickey:
and cut the dependency between wasi-common and wasmtime-wasi, by moving all of the "glue" in wasmtime-wasi specific to wasi-common into the wasi-common crate.

Users of WASI Preview 1 should transition to using `wasmtime_wasi::WasiPreview1Ctx`, which can be constructed from a `wasmtime_wasi::WasiCtx` using a `From` impl.

TODO: docs!


pchickey:
I'm redoing this as broken up into many PRs

PR #6895, opened 2023-08-23
wasi-nn: switch testing infrastructure to Rust

abrown:
Previously, all wasi-nn testing was performed by a `ci/run-wasi-nn-example.sh` Bash script. This was inadequate for testing the `wasmtime-wasi-nn` crate. This change attempts to remedy that by (1) switching the "check the environment" logic of that script over to Rust and (2) writing some more tests. The tests all perform the same MobileNet inference but at different levels of abstraction; I would prefer to test other scenarios but this involves even more infrastructure and seems to be more the responsibility of the ML backend in question. At least this change makes it easier to test more ML backends as they are added and for developers to run tests directly with `cargo test`. Note that part of this infrastructure code is a way to skip tests when the environment is not ready for wasi-nn; developers should not see `cargo test` failures if they are not interested in the `wasmtime-wasi-nn` crate.

PR #7679, opened 2023-12-13
wasi-nn: adapt to new test infrastructure

abrown:
This is a modern refactoring of #6895 to make wasi-nn use the test infrastructure that other WASI proposals are using (i.e., `crates/test-programs`). This is only the beginning: in follow up PRs, wasi-nn could gain unit tests, the existing examples could be removed, more kinds of tests could be added, etc. This will be easier in the future, though, once this infrastructure is in place for the `wasmtime-wasi-nn` crate.


alexcrichton (approved):
Thanks! This is replacing the preexisting wasi-nn test, right? Would it make sense to delete that from CI while here?

Issue #7676, opened 2023-12-12
Runtime host resource types


rvolosatovs:
#### Feature

Given the existing component model API, the only way to invoke an exported component function for which type is not known at compile time is to use [`call`](https://docs.rs/wasmtime/latest/wasmtime/component/struct.Func.html#method.call) on the instantiated function, which provides sort of "runtime reflection" for Wasm components - this works great for all functions, but the ones taking resources as parameters, as the generic [`call`](https://docs.rs/wasmtime/latest/wasmtime/component/struct.Func.html#method.call) takes [`Val`](https://docs.rs/wasmtime/latest/wasmtime/component/enum.Val.html) as parameter, [`Resource`](https://docs.rs/wasmtime/latest/wasmtime/component/enum.Val.html#variant.Resource) variant of which, [`ResourceAny`](https://docs.rs/wasmtime/latest/wasmtime/component/struct.ResourceAny.html), cannot be constructed at runtime by the host directly:

>  This type cannot be directly constructed and is only available if the guest returns it to the host (e.g. a function returning a guest-defined resource)

The proposal is to allow providing an implementation of `ResourceAny` by the hosts.

#### Benefit

This would allow for hosts to run generic Wasm components utilizing resources, for which type may not be known at compile-time by means of runtime reflection of sorts.

#### Implementation

I would imagine there to be a way to construct a dynamically-typed resource at runtime by utilizing functionality similar to [`LinkerInstance::func_new`](https://docs.rs/wasmtime/latest/wasmtime/component/struct.LinkerInstance.html#method.func_new)

Something like this:

```rust
impl DynamicResource {
    pub fn func_new<
        F: Fn(StoreContextMut<'_, T>, &[Val], &mut [Val]) -> Result<()> + Send + Sync + 'static,
    >(
        &mut self,
        name: &str, // resource method name
        func: F,
    ) -> Result<()> {
        // TODO
    }

    // And the other `LinkerInstance` counterparts
}

impl From<DynamicResource> for ResourceAny {
   // TODO
}
```

`DynamicResource` most probably would have to store a reference to https://docs.rs/wit-parser/latest/wit_parser/struct.TypeDef.html or similar, uniquely identifying the resource type expected by the component, but I have not spent too much time studying this particular implementation and I am assuming resources are not duck-typed.

#### Alternatives

[...]

PR #7680, opened 2023-12-13
feat(wasmtime): allow definition of dynamic resource types in the linker

rvolosatovs:
This addresses the first part of https://github.com/bytecodealliance/wasmtime/issues/7676#issuecomment-1852531249, I still need to study the internals a bit more to figure out how to take `Resource<T>` into a `ResourceAny`, which will be addressed in a follow-up


alexcrichton (approved):
Thanks!

PR #7682, opened 2023-12-13
Winch: Change emits to return TypedReg to put on stack

jeffcharles:
There is a small bug where some instructions are putting values containing `TypedReg`s with the incorrect type on the stack. Specifically `i64.eqz` and the i64 comparison instructions. This change has the `emit` functions in the `unop` and `binop` helpers return the the typed registers to put on the stack instead of assigning a type that is the same as the input type.

The filetests for the comparison operators should now be moving the `e` variant of registers instead of the `r` variant of registers. I also needed to add an additional test for `i64.eqz` since the existing tests did not change as a result of fixing this issue.


saulecabrera (approved):
Looks reasonable to me, thanks!

PR #7688, opened 2023-12-14
feat: implement `Resource<T>` -> `ResourceAny` conversion

rvolosatovs:
Closes https://github.com/bytecodealliance/wasmtime/issues/7676
Follow-up on #7680 

~This feature requires changes in the resource table to allow specifying the `ResourceType` associated with the resource stored (since assumption that it's always equal to `ResourceType::host::<T>()` is no longer valid)~

Add support for `Resource<T>` -> `ResourceAny` conversion to allow invoking component functions without bindings generated at compile-time.

### DIFF
--- a/crates/wasmtime/src/component/instance.rs
+++ b/crates/wasmtime/src/component/instance.rs
@@ -38,7 +38,7 @@ pub(crate) struct InstanceData {
     // Otherwise the full guts of this component should only ever be used during
     // the instantiation of this instance, meaning that after instantiation much
     // of the component can be thrown away (theoretically).
-    component: Component,
+    pub(crate) component: Component,
 
     state: OwnedComponentInstance,
 
--- a/crates/wasmtime/src/component/instance.rs
+++ b/crates/wasmtime/src/component/instance.rs
@@ -1,7 +1,7 @@
 use crate::component::func::HostFunc;
 use crate::component::matching::InstanceType;
 use crate::component::{
-    Component, ComponentNamedList, Func, Lift, Lower, PathIndex, ResourceType, TypedFunc,
+    Component, ComponentNamedList, Func, Lift, Lower, ResourceImportIndex, ResourceType, TypedFunc,
 };
 use crate::instance::OwnedImports;
 use crate::linker::DefinitionType;
@@ -522,7 +522,7 @@ impl<'a> Instantiator<'a> {
 pub struct InstancePre<T> {
     component: Component,
     imports: Arc<PrimaryMap<RuntimeImportIndex, RuntimeImport>>,
-    paths: Arc<Vec<Option<RuntimeImportIndex>>>,
+    resource_imports: Arc<Vec<Option<RuntimeImportIndex>>>,
     _marker: marker::PhantomData<fn() -> T>,
 }
 
@@ -532,7 +532,7 @@ impl<T> Clone for InstancePre<T> {
         Self {
             component: self.component.clone(),
             imports: self.imports.clone(),
-            paths: self.paths.clone(),
+            resource_imports: self.resource_imports.clone(),
             _marker: self._marker,
         }
     }
@@ -548,22 +548,25 @@ impl<T> InstancePre<T> {
     pub(crate) unsafe fn new_unchecked(
         component: Component,
         imports: PrimaryMap<RuntimeImportIndex, RuntimeImport>,
-        paths: Vec<Option<RuntimeImportIndex>>,
+        resource_imports: Vec<Option<RuntimeImportIndex>>,
     ) -> InstancePre<T> {
         InstancePre {
             component,
             imports: Arc::new(imports),
-            paths: Arc::new(paths),
+            resource_imports: Arc::new(resource_imports),
             _marker: marker::PhantomData,
         }
     }
 
-    pub(crate) fn runtime_import_index(&self, path: PathIndex) -> Option<RuntimeImportIndex> {
-        *self.paths.get(*path)?
+    pub(crate) fn resource_import_index(
+        &self,
+        path: ResourceImportIndex,
+    ) -> Option<RuntimeImportIndex> {
+        *self.resource_imports.get(*path)?
     }
 
-    pub(crate) fn import_by_path(&self, path: PathIndex) -> Option<&RuntimeImport> {
-        let idx = self.runtime_import_index(path)?;
+    pub(crate) fn resource_import(&self, path: ResourceImportIndex) -> Option<&RuntimeImport> {
+        let idx = self.resource_import_index(path)?;
         self.imports.get(idx)
     }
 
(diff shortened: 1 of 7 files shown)


alexcrichton:
I think I might have missed something in this by accident perhaps. I don't think that this is the direction things should go in, for example `Resource<T>` is designed to have a statically known type as `ResourceType::host::<T>()` as opposed to a runtime-defined type. Having a runtime-defined type sort of defeats the purpose of the type parameter which at that point require a redesign of the resource-related APIs.

The title of this PR though, allowing host creation of guest resources, is not something Wasmtime can enable. Just as guests can't create host-defined resources without asking the host to do so the other way around is the same. The host can't conjure a guest resource out of thin air, it must originate in the guest. 

Given that I fear I might have missed some motivation/use case when talking about this, could you perhaps clarify?

rvolosatovs:
> Given that I fear I might have missed some motivation/use case when talking about this, could you perhaps clarify?

Here's a more concrete example of the use case:

Assume the following WIT:
```
interface types {
    use wasi:io/poll@0.2.0-rc-2023-11-10.{pollable};

    resource future-string {
        subscribe: func() -> pollable;
        get: func() -> option<string>;
    }
}

interface example {
    use types.{future-string};

    foo: func(future-string);
}

world opaque-world {
    export example
}
```
And a component exporting the `example` interface.

The intention is to call `example.foo` on the component without having the WIT at compilation time, i.e. parsing the WIT from the Wasm itself at runtime (which works already) and then invoking `example.foo` within the component.
The only blocker for this now is the fact that the host cannot construct `future-string` at runtime if it does not have the `future-string` WIT definition available at compilation time (and associated bindings)

The host, however, can already push a `Resource` to the table and link the associated methods on the linker, the only bit left is being able to turn the `Resource` constructed at runtime into a `ResourceAny` to be able to invoke the function via `call` or `call_async`

rvolosatovs:
Another way to look at this would be the following:
- If one could pass a `wasmtime::component::Val::U32`, which stores the `Resource::rep` value as a function parameter directly where a resource is expected, that would cover the use case.

Perhaps a an even more generic view on this could be allowing arbitrary `impl Lower` values to be passed as arguments to `call`/`call_async` (or yet another alternative)

This all really boils down to being able to pass the resource index to a component in cases where the resource type is not known at compilation time

rvolosatovs:
Small update: I found a way to obtain the `TypeResourceTableIndex` from the instance, which simplifies getting the destructor and flags, however I keep running into

```
    type mismatch for own, possibly due to mixing distinct composite types
```

I will reach-out offline

rvolosatovs:
@alexcrichton , as discussed, I introduced a `PathIndex` mapping `usize`s to `Vec<usize>` import paths.
I did not attempt to work on optimizing this, but rather focused on getting something simple over the line.
I added the `PathIndex` to the `NameMap` value, since in order to use these indexes a reverse mapping from `Vec<usize>` import path to `PathIndex` is required

crates/wasmtime/src/component/linker.rs
```diff
@@ -27,6 +27,7 @@ pub struct Linker<T> {
     strings: Strings,
     map: NameMap,
     path: Vec<usize>,
+    paths: Vec<Vec<usize>>, // all paths defined in this linker
```
alexcrichton, line 30: I'm a little confused about the purpose of this field, it seems like this is written/pushed to and the length is read, but otherewise the contents of this array aren't ever read. I was expecting just an integer for this field which increments on each defined resource within the linker, which tracks with how this is used I think.

crates/wasmtime/src/component/linker.rs
```diff
@@ -289,7 +326,7 @@ impl<T> LinkerInstance<'_, T> {
     /// argument which can be provided to the `func` given here.
     //
     // TODO: needs more words and examples
-    pub fn func_wrap<F, Params, Return>(&mut self, name: &str, func: F) -> Result<()>
+    pub fn func_wrap<F, Params, Return>(&mut self, name: &str, func: F) -> Result<PathIndex>
```
alexcrichton, line 329: I was thinking that the `PathIndex` here would only be returned for `fn resource` since it's not really applicable to any of the other types of imports. That'd also enable pushing `PathIndex` into the `Definition::Resource` variant instead of having it for all variants.

crates/wasmtime/src/component/resources.rs
```diff
@@ -501,6 +513,55 @@ struct OwnState {
 }
 
 impl ResourceAny {
+    /// Attempts to convert an imported [`Resource`] into [`ResourceAny`].
+    /// `root` is the import root name and `path` is the import path within `root`.
```
alexcrichton, line 517: I think this doc comment will need an update

alexcrichton (approved):
Thanks again for this! A few comments which I'm happy to defer to later, but for the implementation of `LinkerInstance::resource` I think it'd be good to continue to rely on the `insert` helper. I wouldn't worry too much about the interaction of bumping the resource counter if an error is returned.

crates/wasmtime/src/component/instance.rs
```diff
@@ -520,6 +522,7 @@ impl<'a> Instantiator<'a> {
 pub struct InstancePre<T> {
     component: Component,
     imports: Arc<PrimaryMap<RuntimeImportIndex, RuntimeImport>>,
+    resource_imports: Arc<Vec<Option<RuntimeImportIndex>>>,
```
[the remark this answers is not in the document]
rvolosatovs: This was a small enough change, that I just did it in this PR, but it did require a small addition to `PrimaryMap` impl to avoid iteration for constructing it

crates/wasmtime/src/component/linker.rs
```diff
@@ -61,18 +63,35 @@ pub struct LinkerInstance<'a, T> {
...
     strings: &'a mut Strings,
     map: &'a mut NameMap,
+    resource_imports: &'a mut usize,
     allow_shadowing: bool,
     _marker: marker::PhantomData<fn() -> T>,
 }
 
+/// Index correlating a resource definition to the import path
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub struct ResourceImportIndex(usize);
```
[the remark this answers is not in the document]
rvolosatovs: Small enough that I did it here

sehz:
Any progress?  This is blocker for us

rvolosatovs:
> Any progress? This is blocker for us

Hey @sehz,

There was a doc CI failure:

> error: unresolved link to `Linker::resource`
   --> crates/wasmtime/src/component/resources.rs:519:60
    |
519 |     /// `idx` is the [`ResourceImportIndex`] returned by [`Linker::resource`].
    |                                                            ^^^^^^^^^^^^^^^^ no item named `Linker` in scope

> warning: `cranelift-jit` (lib doc) generated 1 warning (1 duplicate)
 Documenting wasmtime-cranelift-shared v17.0.0 (/home/runner/work/wasmtime/wasmtime/crates/cranelift-shared)
warning: `wasmtime` (lib doc) generated 1 warning (1 duplicate)
error: could not document `wasmtime`

Which, I believe, is now fixed. I've rebased and added `prtest:full`(not sure if a scoped option exists) to run all CI checks on this PR, once that succeeds it should be ready for merge.

Please note that I'm generally out of office until Monday, the 8th of January, but I may be able to address small things like this doc CI failure in the meantime

PR #7690, opened 2023-12-14
wasi-sockets: Add SO_REUSEADDR back in

badeend:
In #7662 the call to `bind_existing_tcp_listener` was replaced with `rustix::net::bind`. Unfortunately, besides calling `bind`, `bind_existing_tcp_listener` also set `SO_REUSEADDR`. This PR restores that behavior. I've added an integration test and a note to the WITs too. 

Unrelated: I also removed the workarounds in the Bind tests:
```rs
        // Concurrent invocations of this test can yield `AddressInUse` and that
        // same error can show up on Windows as `AccessDenied`.
        Err(ErrorCode::AddressInUse | ErrorCode::AccessDenied) => return,
```
by using a random generated port instead of a single constant.


alexcrichton (approved):

crates/wasi/wit/deps/sockets/tcp.wit
```diff
@@ -42,6 +42,11 @@ interface tcp {
         /// - `address-not-bindable`:      `local-address` is not an address that the `network` can bind to. (EADDRNOTAVAIL)
         /// - `not-in-progress`:           A `bind` operation is not in progress.
         /// - `would-block`:               Can't finish the operation, it is still in progress. (EWOULDBLOCK, EAGAIN)
+        /// 
+        /// # Implementors note
+        /// When binding to a non-zero port, this bind operation shouldn't be affected the TIME_WAIT
+        /// state of a recently closed socket on the same local address. Ie. the SO_REUSEADDR socket
+        /// option should be set implicitly on platforms that require it.
```
rylev, line 49:
```suggestion
        /// When binding to a non-zero port, this bind operation shouldn't be affected by the TIME_WAIT
        /// state of a recently closed socket on the same local address (i.e., the SO_REUSEADDR socket
        /// option should be set implicitly on platforms that require it).
```

badeend:
This should be good to go again.

In ran into EADDRINUSE issues on my local machine when binding to a specific (but random) port. So I've modified the tests to give it a few attempts.

alexcrichton (approved):

PR #7691, opened 2023-12-14
[WASI-NN] Add support for a ONNXruntime backend using ort

devigned:
This change adds an ONNXruntime backend for WASI-NN. Since there is only one backend implemented, this will help to move the standardization of this proposal forward.

Also, the example usage of the ONNXruntime backend shows how to build a component using WASI-NN and ONNXruntime.


devigned:
@abrown please take a look when you have a minute. Thank you!!

abrown:
@devigned, I'll take a closer look next week. I will say I've been been changing the ground under your feet here (sorry 😁) with https://github.com/bytecodealliance/wasmtime/pull/7679; that PR will mean we have to move some of your tests over to `test-programs` and write Rust tests instead of bash scripts. An overall improvement, just inconvenient timing.

As for the CI failures here: I think we can add a line to `skip-tree` in `deny.toml` to ignore the version mismatch of the `half` dependency: { name = "half", depth = 1 }`. As for the `cargo vet` failures, I think we'll have to audit those, maybe even as a separate PR... anything you can do to reduce the number of dependencies coming in will make that easier (drop features?).

ci/run-wasi-nn-example.sh
```diff
@@ -19,13 +19,37 @@ else
...
+# └── lib
+#    ├── cmake
+#    │   ├── inferenceengine2023.2.0
+#    │   ├── ngraph2023.2.0
+#    │   └── openvino2023.2.0
+#    ├── openvino-2023.2.0
+#    ├── pkgconfig
+#    └── python3.11
+#        └── ...
+export DYLD_LIBRARY_PATH="${OSX_OPENVINO_LIB_PATH}"
```
abrown, line 43: I opened https://github.com/abrown/install-openvino-action/issues/29 to track this.

ci/run-wasi-nn-example.sh
```diff
@@ -50,6 +74,15 @@ popd
...
     $TMP_DIR/wasi-nn-example-named.wasm
 
+# Build and run an ONNX component example doing classification.
+cargo build -p wasi-preview1-component-adapter --target wasm32-unknown-unknown --release
+pushd $WASMTIME_DIR/crates/wasi-nn/examples/classification-component-onnx
+cargo component build --release
+cp target/wasm32-wasi/release/classification_component_onnx.wasm $TMP_DIR
+cp -rf ./fixture $TMP_DIR/
+popd
+cargo run --features wasi-nn,component-model -- run --wasm component-model --dir $TMP_DIR/fixture::fixture -S nn $TMP_DIR/classification_component_onnx.wasm
```
abrown, line 84: I don't think we'll need this script anymore now that #7679 is merged.

crates/wasi-nn/examples/classification-component-onnx/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+"wasi:filesystem" = { path = "../../../wasi/wit/deps/filesystem"}
+"wasi:sockets" = { path = "../../../wasi/wit/deps/sockets"}
+"wasi:random" = { path = "../../../wasi/wit/deps/random"}
+"wasi:nn" = { path = "../../spec/wit/wasi-nn.wit" }
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+cargo-component-bindings = { git="https://github.com/bytecodealliance/cargo-component" }
+image = { version = "0.24.6", default-features = false, features = ["gif", "jpeg", "ico", "png", "pnm", "tga", "tiff", "webp", "bmp", "hdr", "dxt", "dds", "farbfeld"]  }
+ndarray = "0.15.3"
```
abrown, line 26: I don't know how easy it will be to add these dependencies to `test-programs`. If it is difficult for some reason, you could always dump the decoded bytes to a file like I did with the OpenVINO test fixtures. If you do keep this, though, we probably only need the `jpeg` feature.

crates/wasi-nn/examples/classification-component-onnx/src/lib.rs
```diff
@@ -0,0 +1,131 @@
+cargo_component_bindings::generate!();
```
abrown, line 1: Good to see the auto-generated bindings work!

crates/wasi-nn/src/backend/onnxruntime.rs
```diff
@@ -0,0 +1,182 @@
...
+}
+
+struct ONNXGraph(Arc<Mutex<Session>>, ExecutionTarget);
+
+unsafe impl Send for ONNXGraph {}
+unsafe impl Sync for ONNXGraph {}
+
+impl BackendGraph for ONNXGraph {
+    fn init_execution_context(&self) -> Result<ExecutionContext, BackendError> {
+        let inputs = self
```
abrown, line 59: You could probably `self.0.lock()` once at the top of this function?

crates/wasi-nn/src/backend/onnxruntime.rs
```diff
@@ -0,0 +1,182 @@
...
+                        // Array::from_shape_vec(dims, input.data.to_vec()).unwrap()
+                    }
+                    TensorType::I64 => {
+                        unimplemented!("I64 not supported by ONNX");
+                        // Array::from_shape_vec(dims, input.data.to_vec()).unwrap()
+                    }
+                    TensorType::Fp64 => {
+                        unimplemented!("Fp64 not supported by ONNX");
+                        // Array::from_shape_vec(dims, input.data.to_vec()).unwrap()
+                    }
```
abrown, line 138: Maybe something like the following would be less verbose?

```
_ => {
  unimplemented!("{:?} not supported by ONNX", input.tensor_type);
}
```

crates/wasi-nn/src/backend/onnxruntime.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+}
+
+impl From<ort::Error> for BackendError {
+    fn from(e: ort::Error) -> Self {
+        BackendError::BackendAccess(e.into())
+    }
+}
+
+pub fn f32_vec_to_bytes(data: Vec<f32>) -> Vec<u8> {
```
abrown, line 168: Maybe we should explain somewhere why we are forced to copy the tensor twice &mdash; at input and output.

abrown (commented):
Thanks for all the work to make this happen! This is going to be a great addition to Wasmtime. Let me know on Zulip if you need help working through the test refactoring, CI issues, etc.

mtobin-tdab:
This is great, I have been testing different ONNX models and encountered a data type mismatch error for certain models, specifically at the wasi-nn compute function in bindings.rs. The error indicates a failure to parse ctx: GraphExecutionContext as i32:
 
           0: 0x245b42a3 - wit-component:shim!indirect-wasi:nn/inference-compute
           1: 0x698ad - octaioxide::bindings::wasi::nn::inference::compute::h1bf34b4c8ec1bf77
                           at ....................bindings.rs:11137:15
    2: Failed while accessing backend
    3: Data type mismatch: was Int64, tried to convert to Float32
As far as I can tell the ctx value is 0 when I pass it to compute, I've tried to cast it to f32, but the compiler expects a u32.
 
Very possible I'm missing something obvious as I am fairly new to both WASM and Rust. But thought I would share.
 
For some extra context, both models which have produced this error have been classification models (specifcially an xgboost and sklearn's SGDClassifier model) - Are there model types in ONNX format which we would never expect this to be compatible with?

ci/run-wasi-nn-example.sh
devigned: I fixed this and added a test program.

crates/wasi-nn/examples/classification-component-onnx/Cargo.toml
devigned: added to test-programs and reduced to jpeg.

crates/wasi-nn/examples/classification-component-onnx/src/lib.rs
devigned: I updated this to reflect the latest changes in `cargo component` and preview 2.

crates/wasi-nn/src/backend/onnxruntime.rs
devigned: I was verbose to be super clear about what is not implemented. I've since changed it to your suggestion.

devigned:
> This is great, I have been testing different ONNX models and encountered a data type mismatch error for certain models, specifically at the wasi-nn compute function in bindings.rs. The error indicates a failure to parse ctx: GraphExecutionContext as i32:
> 
> ```
> 1: error while executing at wasm backtrace:
>        0: 0x245b42a3 - wit-component:shim!indirect-wasi:nn/inference-compute
>        1: 0x698ad - octaioxide::bindings::wasi::nn::inference::compute::h1bf34b4c8ec1bf77
>                        at ....................bindings.rs:11137:15
> 2: Failed while accessing backend
> 3: Data type mismatch: was Int64, tried to convert to Float32
> ```
> 
> As far as I can tell the ctx value is 0 when I pass it to compute, I've tried to cast it to f32, but the compiler expects a u32.
> 
> Very possible I'm missing something obvious as I am fairly new to both WASM and Rust. But thought I would share.
> 
> For some extra context, both models which have produced this error have been classification models (specifcially an xgboost and sklearn's SGDClassifier model) - Are there model types in ONNX format which we would never expect this to be compatible with?

Is there any chance you could share code that reproduces this behavior, or create a branch with a failing test reproducing this behavior?

devigned:
@sunfishcode, I believe I need someone that is a trusted contributor to Wasmtime to approve the cargo vet dependencies added in this PR (per https://docs.wasmtime.dev/contributing-coding-guidelines.html#dependencies-of-wasmtime). Would you please consider allow listing the failing vet dependencies?

I'm not sure if the `cargo deny` result for `ittapi-sys` is spurious or if the license is going to be a problem. What would you advise?

abrown:
@devigned. I see a bunch of dependencies for the [failing vet check](https://github.com/bytecodealliance/wasmtime/actions/runs/7836371527/job/21383872144?pr=7691#step:5:15):

```
  bytemuck:1.14.2 missing ["safe-to-deploy"]
  bytes:1.5.0 missing ["safe-to-deploy"]
  castaway:0.2.2 missing ["safe-to-deploy"]
  color_quant:1.1.0 missing ["safe-to-deploy"]
  compact_str:0.7.1 missing ["safe-to-deploy"]
  crunchy:0.2.2 missing ["safe-to-deploy"]
  fdeflate:0.3.4 missing ["safe-to-deploy"]
  filetime:0.2.16 missing ["safe-to-deploy"]
  flate2:1.0.28 missing ["safe-to-deploy"]
  half:2.3.1 missing ["safe-to-deploy"]
  image:0.24.8 missing ["safe-to-deploy"]
  matrixmultiply:0.3.8 missing ["safe-to-deploy"]
  ndarray:0.15.6 missing ["safe-to-deploy"]
  num-complex:0.4.4 missing ["safe-to-deploy"]
  num-integer:0.1.45 missing ["safe-to-deploy"]
  ort:2.0.0-rc.0 missing ["safe-to-deploy"]
  ort-sys:2.0.0-rc.0 missing ["safe-to-deploy"]
  png:0.17.11 missing ["safe-to-deploy"]
  rawpointer:0.2.1 missing ["safe-to-deploy"]
  rustversion:1.0.14 missing ["safe-to-deploy"]
  simd-adler32:0.3.7 missing ["safe-to-deploy"]
  static_assertions:1.1.0 missing ["safe-to-deploy"]
  tar:0.4.40 missing ["safe-to-deploy"]
  ureq:2.9.1 missing ["safe-to-deploy"]
  xattr:1.1.1 missing ["safe-to-deploy"]
```
I would guess that at least some of these are related to testing? If that is the case, one thing that I suggested in #7807 is to create a test fixture file with all the tensor bytes in the right format. This could remove some of these dependencies, reducing the amount of crates that we will have to audit.

devigned:
@abrown, I've trimmed down to about as slim as I think I can get the dependencies. I've removed any dependency on `half` and `ndarray` which cut the list of dependencies by over half.

```
12 unvetted dependencies:
  bytes:1.5.0 missing ["safe-to-deploy"]
  castaway:0.2.2 missing ["safe-to-deploy"]
  compact_str:0.7.1 missing ["safe-to-deploy"]
  filetime:0.2.16 missing ["safe-to-deploy"]
  flate2:1.0.28 missing ["safe-to-deploy"]
  ort:2.0.0-rc.0 missing ["safe-to-deploy"]
  ort-sys:2.0.0-rc.0 missing ["safe-to-deploy"]
  rustversion:1.0.14 missing ["safe-to-deploy"]
  static_assertions:1.1.0 missing ["safe-to-deploy"]
  tar:0.4.40 missing ["safe-to-deploy"]
  ureq:2.9.1 missing ["safe-to-deploy"]
  xattr:1.2.0 missing ["safe-to-deploy"]
``` 

WDYT?

abrown:
> @abrown, I've trimmed down to about as slim as I think I can get the dependencies. I've removed any dependency on `half` and `ndarray` which cut the list of dependencies by over half.
> 
> WDYT?

Thanks! Let's talk about this in the Wasmtime meeting tomorrow.

abrown:
I audited the dependencies this PR would bring in and the only crate I didn't audit was `compact_str`: I do not feel comfortable putting a stamp of approval on this crate without some further discussion. The crate is well-designed and has many kinds of tests, including proptests, but the fundamental `unsafe` issue is that the crate manually implements what would otherwise be an `enum`: if a string is less than 24 bytes, it lives on the stack; otherwise, it lives on the heap. This causes all kinds of `unsafe` pointer copies for its special integer representation, `unsafe` ways to pass in unchecked UTF-8 strings, several `unsafe` implementations of `Sync`, `Send`, `LifetimeFree`, etc. I would note that all of these unsafe instances are carefully documented with "SAFETY:" comments which seemed reasonable, but the large amount of unsafety and how inherent it is to the crate gave me pause. Perhaps someone else can more confidently vouch for this?

alexcrichton:
Given your review and the fact that I'm seeing extensive fuzzing and miri testing in the repo I think it's safe to put all that into the vet entry. Vetting isn't so much guaranteeing that the crate is correct but moreso that it doesn't intentionally do bad things. You've done an initial review and the author is clearly covering all the bases they can, so I think it's reasonable to say it's well vetted.

devigned:
@abrown would you like me to add the vet audit entry for `compact_str`? Is that all we need to get this moving forward?

abrown:
@devigned: yeah, go for it!

mingqiusun (approved):

Cargo.toml
```diff
@@ -345,6 +345,8 @@ default = [
 all-arch = ["wasmtime/all-arch"]
 winch = ["wasmtime/winch"]
 wmemcheck = ["wasmtime/wmemcheck"]
+wasi-nn-onnx = ["wasmtime-wasi-nn/onnx"] # enable ONNX backend for WASI-NN
+wasi-nn-winml = ["wasmtime-wasi-nn/winml"] # enable WinML backend for WASI-NN
```
abrown, line 349: I don't think we need to add this top-level feature because we have workspaces. We should be able to reach down and control sub-crate features by doing: `cargo build --features wasmtime-wasi-nn/onnx`, e.g.

ci/run-tests.sh
```diff
@@ -5,6 +5,8 @@ cargo test \
     --features wasi-http \
     --features component-model \
     --features serve \
+    --features wasi-nn-onnx \
+    --features wasi-nn-winml \
```
abrown, line 9:
```suggestion
    --features wasmtime-wasi-nn/onnx \
```

See my previous comment about workspace features, plus I don't think we really want to enable the `wasmtime-wasi-nn/winml` feature unconditionally since (1) it is Windows-specific and (2) @jianjunz found that the WinML APIs aren't even available in the GitHub CI images.

crates/test-programs/src/bin/nn_image_classification_onnx.rs
```diff
@@ -0,0 +1,60 @@
...
+        &sort_results(&output_buffer)[..5]
+    );
+
+    Ok(())
+}
+
+// Sort the buffer of probabilities. The graph places the match probability for
+// each class at the index for that class (e.g. the probability of class 42 is
+// placed at buffer[42]). Here we convert to a wrapping InferenceResult and sort
+// the results. It is unclear why the MobileNet output indices are "off by one"
```
abrown, line 44: Might want to tweak these comments: "MobileNet" -> "SqueezeNet". BTW, are the SqueezeNet expectations here the same?

crates/wasi-nn/examples/classification-component-onnx/Cargo.toml
abrown: In case some Git mischief happened here: `image` looks to be pulling in all the features again, not just `jpeg`.

crates/wasi-nn/Cargo.toml
```diff
@@ -45,11 +47,14 @@ walkdir = { workspace = true }
 cap-std = { workspace = true }
 test-programs-artifacts = { workspace = true }
 wasi-common = { workspace = true, features = ["sync"] }
+wasmtime-wasi = { workspace = true }
```
abrown, line 50: Not sure this should be making a comeback &mdash; why is it back?

crates/wasi-nn/examples/classification-component-onnx/fixture/models/squeezenet1.1-7.onnx
abrown, line 1: Let's follow the precedent like with the other backends and just retrieve any models with a test helper function, e.g.:

https://github.com/bytecodealliance/wasmtime/blob/83417388c0aa03592600a7ed9a8e42316527be45/crates/wasi-nn/src/testing.rs#L84

crates/wasi-nn/src/testing.rs
```diff
@@ -81,7 +81,7 @@ static ARTIFACTS: Mutex<()> = Mutex::new(());
 /// Return `Ok` if we find the cached MobileNet test artifacts; this will
 /// download the artifacts if necessary.
 #[cfg(feature = "openvino")]
-fn check_openvino_artifacts_are_available() -> Result<()> {
+fn check_openvino_onnx_artifacts_are_available() -> Result<()> {
```
abrown, line 84: We can refactor all this stuff later, but for now it probably makes sense to retrieve any ONNX artifacts separately, e.g., with a `check_onnx_artifacts_are_available` function.

crates/wasi-nn/src/testing.rs
```diff
@@ -103,10 +103,19 @@ fn check_openvino_artifacts_are_available() -> Result<()> {
...
+    if !local_path.is_file() {
+        download(&remote_url, &local_path).with_context(|| "unable to retrieve test artifact")?;
+    } else {
+        println!("> using cached artifact: {}", local_path.display())
+    }
     Ok(())
 }
 
-#[cfg(feature = "winml")]
+#[cfg(all(feature = "winml", target_os = "windows"))]
```
abrown, line 118: Yeah, I think all these additions of `target_os = "windows"` are necessary; thanks for catching that.

crates/wasi-nn/src/witx.rs
```diff
@@ -44,7 +44,10 @@ mod gen {
             match e {
                 WasiNnError::BackendError(_) => unimplemented!(),
                 WasiNnError::GuestError(_) => unimplemented!(),
-                WasiNnError::UsageError(_) => unimplemented!(),
+                WasiNnError::UsageError(e) => {
+                    println!("[OUCH!!!] {:?}", e);
```
abrown, line 48: Let's just leave this unimplemented; we can follow up with better error handling in a separate PR.

crates/wasi-nn/tests/all.rs
```diff
@@ -103,8 +107,21 @@ fn nn_image_classification_named() {
...
 #[test]
 fn nn_image_classification_winml() {
     run(NN_IMAGE_CLASSIFICATION_WINML, true).unwrap()
 }
+
+#[cfg_attr(
+    not(all(
+        feature = "onnx",
+        target_arch = "x86_64",
+        any(target_os = "linux", target_os = "windows")
```
abrown, line 120: No `"macos"`?

tests/spec_testsuite
abrown, line 1: Friendly reminder that this Git submodule change is probably from a missing `git submodule update --init` at some point.

crates/test-programs/src/bin/nn_image_classification_onnx.rs
devigned: No, I don't think we needed to skip 1 there.

tests/spec_testsuite
devigned: Yep. Updated. Thank you.

crates/wasi-nn/tests/all.rs
```diff
@@ -13,41 +13,30 @@ const PREOPENED_DIR_NAME: &str = "fixture";
...
-    let mut backends = vec![];
-    #[cfg(feature = "openvino")]
-    {
-        backends.push(Backend::from(backend::openvino::OpenvinoBackend::default()));
-    }
-    #[cfg(feature = "winml")]
-    {
-        backends.push(Backend::from(backend::winml::WinMLBackend::default()));
-    }
-    for backend in backends {
```
devigned, line 34: @abrown this won't work with multiple features enabled. For example, if onnx and openvino are enabled, then each backend will try to be run for a given model. Unless the backends support the same model input, the caller is going to have a bad time.

Rather than running for every backend in a test, I have each test pass in their backend to run.

abrown: Yeah, that makes more sense! 👍

abrown (approved):

crates/wasi-nn/examples/classification-component-onnx/Cargo.toml
```diff
@@ -2,6 +2,9 @@
 name = "classification-component-onnx"
 version = "0.1.0"
 edition = "2021"
+authors = ["The Wasmtime Project Developers"]
+readme = "README.md"
+publish = false
```
devigned, line 7: @abrown I didn't realize that I needed to mark this as `publish = false`. Should be good now.

squillace:
outstanding. Thanks for your work, @devigned and to the reviewers!

squillace:
go figure: 

 thread 'main' panicked at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ort-sys-2.0.0-rc.0/build.rs:308:22:
  downloaded binaries not available for target riscv64gc-unknown-linux-gnu
  you may have to compile ONNX Runtime from source
  stack backtrace:
     0: rust_begin_unwind
               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/panicking.rs:645:5
     1: core::panicking::panic_fmt
               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/panicking.rs:72:14
     2: build_script_build::prepare_libort_dir
     3: build_script_build::real_main
     4: build_script_build::main
     5: core::ops::function::FnOnce::call_once
warning: build failed, waiting for other jobs to finish...
Error: Process completed with exit code 101.

abrown:
> downloaded binaries not available for target riscv64gc-unknown-linux-gnu

That's fine; this just needs a few more target checks.

devigned:
> > downloaded binaries not available for target riscv64gc-unknown-linux-gnu
> 
> That's fine; this just needs a few more target checks.

I'm on it.

squillace:
let's go baby

squillace:
GACK. MingGW!!! crazy amazing

squillace:
you having fun yet, @devigned ?

alexcrichton:
Since this isn't intended to work everywhere just yet, it might be best to add a new builder on CI specifically dedicated to testing ONNX rather than trying to get it to work across all the platforms.

If that works for you I'd recommend copying [this to start](https://github.com/bytecodealliance/wasmtime/blob/d29b863aaf6447f65f63de9596a2f8116d0b2447/.github/workflows/main.yml#L364-L384), updating all the bits there (e.g. run on ubuntu, not on windows), and then be sure to add an entry [down here](https://github.com/bytecodealliance/wasmtime/blob/d29b863aaf6447f65f63de9596a2f8116d0b2447/.github/workflows/main.yml#L828) to ensure we gate on it in CI.

Also, what I might also recommend, if you add `prtest:full` as a string in a commit message it'll run full CI on this PR before going to the merge queue, and that may help avoid the bouncing back and forth between the merge queue and back here.

squillace:
drinking booze all night now over here, best soap opera I've watched in years

devigned:
@alexcrichton, in https://github.com/bytecodealliance/wasmtime/pull/7691/commits/22d14d6e768d8bbc346b7ade3a99bed7b8571092 I'm reducing the set of triplets in which ONNX will run. I do not plan on pursuing riscv or s390 precompiled onnxruntime bins at this point. Are you good with the matrix-test approach or would you still prefer I break it out into another builder?

alexcrichton:
Nah if that works for you seems fine, just trying to save some CI headache.

PR #7699, opened 2023-12-18
PCC [AArch64]: load_constant64_full generates incorrect fact range for 32-bit MOVN

feilongjiang:
Hi team, find another PCC fact range issue on AArch64.
When the top 32 bits of the value are zero and the upper_halfword of the lower 32 bits is u16::MAX, load_constant64_full will emit a single `movn` instruction. In that case, add_range_fact only adds lower_halfword as the fact range, which misses the upper halfword value. We should add the hole 32-bit value as the fact range.

Here is a simple test to reproduce the issue:
```wasm
;; test_const.wat
(module
  (func (export "test_constant") (result i64) (i64.const 0xfffffff7))
)
```

Step to reproduce:

Compile the test with PCC on:
```
target/debug/wasmtime compile -C pcc=y test_const.wat
```

the output will be:
```
Error: Compilation error: Proof-carrying-code validation error: UnsupportedFact
```

### DIFF
--- a/cranelift/codegen/src/isa/aarch64/lower/isle.rs
+++ b/cranelift/codegen/src/isa/aarch64/lower/isle.rs
@@ -307,8 +307,8 @@ impl Context for IsleContext<'_, '_, MInst, AArch64Backend> {
                     self.lower_ctx.add_range_fact(
                         rd.to_reg(),
                         64,
-                        u64::from(lower_halfword),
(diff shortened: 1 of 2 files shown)


feilongjiang:
Hi @fitzgen, would you please take a look?

fitzgen (approved):
Thanks!

PR #7702, opened 2023-12-18
More spaceship patterns

scottmcm:
This is a continuation of #7636 that added the spaceship patterns.

In particular, I noticed that what Rust currently does in `sort_by`

https://github.com/rust-lang/rust/blob/bf9229a2e366b4c311f059014a4aa08af16de5d8/library/core/src/slice/mod.rs#L3008

is a simplification pattern that I'd missed, so this adds that, and a couple other related things I saw in the area.


fitzgen (approved):
Thanks!

Issue #7681, opened 2023-12-13
WASI: Fine-grained network policies


badeend:
#7662 is a good first step towards letting users of the wasmtime library customize network behavior. After that change, library users have two options: either use `inherit_network` which grants access to everything, or use `socket_addr_check` which allows the user to define arbitrarily complex rules. The upside of that last one is also its downside: the user _must_ define everything themself. On the command-line, the available choices are quite limiting: everything or nothing at all. Guess which one users will pick :P

I would like to add a middle ground option for both library & CLI users. This option should provide a default "good enough" option for the majority of users, while still allowing to fall back to fully customized control. My objective is to be able to declare network policies based on:
- Domain name
- Initiator (client vs server)
- Protocol (TCP, UDP, HTTP(S))

I tried to capture the gist of it in pseudo code:

```rs
// A single network permission
enum Grant {

    // Allow TCP sockets to connect to remote_host/port, optionally using a specific local_interface/port
    TcpOutbound {
        remote_host: RemoteHostPattern,
        remote_port: RemotePortPattern,
        local_interface: LocalInterfacePattern, // default = LocalInterfacePattern::Any
        local_port: LocalPortPattern, // default = LocalPortPattern::Ephemeral
    },

    // Allow TCP sockets to listen on a local port, optionally using a specific local_interface too
    TcpInbound {
        local_port: LocalPortPattern,
        local_interface: LocalInterfacePattern, // default = LocalInterfacePattern::Any
    },

    // Allow UDP sockets to initiate flows to remote_host/port, optionally using a specific local_interface/port
    UdpOutbound {
        remote_host: RemoteHostPattern,
        remote_port: RemotePortPattern,
        local_interface: LocalInterfacePattern, // default = LocalInterfacePattern::Any
        local_port: LocalPortPattern, // default = LocalPortPattern::Ephemeral
    },

    // Allow UDP sockets to handle incoming flows on a local port, optionally using a specific local_interface too
    UdpInbound {
        local_port: LocalPortPattern,
        local_interface: LocalInterfacePattern, // default = LocalInterfacePattern::Any
    },

    // Allows the component to make outgoing HTTP connections
    HttpOutbound {
        scheme: Http | Https,
        host: RemoteHostPattern,
        port: RemotePortPattern,
    },
}


// "*"             -> RemoteHostPattern::Any
// "localhost"     -> RemoteHostPattern::Loopback
[...]

Issue #7694, opened 2023-12-15
[`wasi:sockets/ip-name-lookup#resolve_addresses`] way to track name to `IpAddress` resolution


rylev:
My use case is essentially to allow any socket operation as long as the IP addressed used for that operation was resolved in the last call to `wasi:sockets/ip-name-lookup#resolve_addresses`. 

Given the changes in https://github.com/bytecodealliance/wasmtime/pull/7662 this would stand to reason that somehow I need the ability to have a list of resolved names to IP addresses in the closure passed to `socket_addr_check`. 

This could be accomplished in several ways:
1. Wasmtime passes such a list directly as an argument to the `SocketAddrCheck` closure.
2. Wasmtime gives a way for the embedder to query what the list is at any time.
3. Wasmtime allows the embedder to hook into DNS resolution and be called back with any (name, resolved IP address) pairs 
4. Other ways?

I think I might lean towards option 3. Perhaps this callback could be in the form `Fn(&str, IpAddress) -> Result<(), ErrorCode>` which would be called anytime `resolve_next_address` is about to return a `Ok(Some(addr))` to the guest. This would allow the embedder to do the following things:
* Observe DNS resolution as it happens (needed to unlock my use case)
* Change resolution so that a name resolves to a different set of IP addresses 
* Return an `ErrorCode` error from `resolve_next_address`

If we can agree on an approach, I'm happy to do the implementation.

alexcrichton:
I think this issue definitely makes sense, and this got me thinking a bit too. Reading over this issue and focusing solely on its text I'd also lean towards option 3 you've proposed but that quickly runs into another issue I think. The base problem is that Rust doesn't do well when building up a callback-style API with multiple callbacks.  For example the "this got resolved" callback is distinct from the "is this socket address valid" callback. These two callbacks don't share state meaning that you'd have to use `Rc`/`Arc`. With mutability you're forced to also use `RefCell`/`Mutex`. Given that these are used in a `Send`/`Sync` context, that's requiring `Arc<Mutex<...>>` which isn't great because there's no actual multithreading going on here.

Given that problem it's prompted me to step back a bit. We solved this in Wasmtime for host functions by passing `&mut T` (effectively) to all host functions. That's not possible here though because a `T` in `Store<T>` isn't availble from `WasiCtx`. This means that I don't think there's an easy solution to this problem.

So that leads me to a few possible alternatives:

[...]

PR #7705, opened 2023-12-19
Fine-grained network policies

rylev:
This follows the design outlined in bytecodealliance/wasmtime#7681 for IpNameLookup.

A new trait `IpNameLookup` is introduced that factors out the conversion from a `String` to a list of IP addresses. Another trait, `WasiIpNameLookupView` is used to configure which concrete type that implements `IpNameLookup` will be used in the actual IpNameLookup host implementation. Lastly. the `SystemIpNameLookup` is provided as a default implementation.

Some questions that might guide tweaks:
* Users are now forced to implement `WasiIpNameLookupView` - it would be nice if there was a to default to the built in `SystemIpNameLookup`. If there were default associated types for traits, we could get rid of `WasiIpNameLookupView` and make `ip_name_lookup` a function on `WasiView` that defaults to the the `SystemIpNameLookup`.  But alas that's not available to us.

r? @badeend


crates/wasi/src/preview2/ip_name_lookup.rs
```diff
@@ -5,35 +5,72 @@ use crate::preview2::poll::{subscribe, Pollable, Subscribe};
...
     fn resolve_addresses(
         &mut self,
         network: Resource<Network>,
         name: String,
     ) -> Result<Resource<ResolveAddressStream>, SocketError> {
         let network = self.table().get(&network)?;
 
-        let host = parse(&name)?;
-
         if !network.allow_ip_name_lookup {
```
badeend, line 65: With these changes, can `network.allow_ip_name_lookup` now be removed?

crates/wasi/src/preview2/ip_name_lookup.rs
```diff
@@ -5,35 +5,72 @@ use crate::preview2::poll::{subscribe, Pollable, Subscribe};
...
+
+    /// Get a new instance of the name lookup.
+    fn ip_name_lookup(&self) -> Self::IpNameLookup;
+}
+
+/// A trait for resolving IP addresses from a name.
+#[async_trait::async_trait]
+pub trait IpNameLookup {
+    /// Given a name, resolve to a list of IP addresses
+    async fn resolve_addresses(&mut self, name: String) -> std::io::Result<Vec<IpAddr>>;
```
badeend, line 32: I'm curious; why isn't `resolve_addresses` placed directly  on `WasiIpNameLookupView`? Is there something the extra indirection buys us?

crates/wasi/src/preview2/ip_name_lookup.rs
rylev: It could be potentially. Right now the default is to not allow ip name lookups, so if we remove this, the default will flip. Additionally, the new way to turn off name lookups would be through a trait implementation (i.e., `impl IpNameLookup for BlockedIpNameLookup` or similar) and not through `WasiCtx`/`WasiCtxBuilder`.

Though it's probably somewhat confusing to implement `WasiIpNameLookupView` and then not have name lookups work because you forgot to flip the bit on `WasiCtx`.

Perhaps we remove the `allow_ip_name_lookup` and provide an additional implementation of `IpNameLookup`: `BlockedIpNameLookup` so that users can easily turn off IP name lookup if they like. Thoughts?

crates/wasi/src/preview2/ip_name_lookup.rs
rylev: I'm not sure it gives a ton given that we initialize the type that implements `IpNameLookup` right before we call `resolve_addresses`. The reason I initially went this direction is that it mirrors how the UDP and TCP implementations will work. I can remove it though. Shall I?

crates/wasi/src/preview2/ip_name_lookup.rs
badeend: Hmm. Interesting.
I think the WASI translation layer (`preview2/ip_name_lookup.rs` in this case) should only depend on a single source of truth: `WasiIpNameLookupView`. It should not _also_ go peeking into `network.allow_ip_name_lookup` on its own.

It might be an idea to provide an additional implementation of `IpNameLookup` based on `WasiCtx`, that dynamically checks `allowed_network_uses`. This could supersede the `BlockedIpNameLookup` you suggested

crates/wasi/src/preview2/ip_name_lookup.rs
badeend: Alright. Was just curious.
No strong preference

crates/wasi/src/preview2/ip_name_lookup.rs
rylev: This does have the downside that the WasiCtx setting can be ignored by users. If their implementation of `IpNameLookup` doesn't check for `allowed_network_uses`, that option will simply be ignored. I guess it seems fine that the user can ignore that option (they're the ones setting it anyway), but it does seem somewhat error prone.

badeend:
All in all, looks good to me 🚀

rylev:
@badeend @alexcrichton 

I've moved the implementation towards having a single `WasiNetworkView` trait where all network based overrides will be implemented. I'm happier with this solution than the previous `IpNameLookupView` version, but I'm still not convinced we've arrived at a good solution. 

The current solution has the following charactertistics:
* Two new methods on `WasiView`: `network_view` and `network_view_mut` which return references to a `dyn WasiNetworkView`
  * This prevents all the methods necessarily being on `WasiView` which might get crowded over time. It also makes it easy for users who want to use an off-the-shelf solution to get implement 2 methods instead of the N methods `WasiNetworkView` will have.
  * The use of dynamic dispatch is new to WasiView but seems appropriate here.
* The various implementors in Wasmtime of `WasiView` gain a new field with `SystemNetwork` (the default implementation of  `WasiNetworkView`)

I would said, we move forward with this for now and we can adjust as we add TCP and UDP support (and of course continue to evolve over time). Thoughts?

crates/wasi/src/preview2/ip_name_lookup.rs
```diff
@@ -5,35 +5,73 @@ use crate::preview2::poll::{subscribe, Pollable, Subscribe};
...
+    Waiting(AbortOnDropJoinHandle<std::io::Result<Vec<IpAddr>>>),
+    Done(std::io::Result<vec::IntoIter<IpAddr>>),
 }
 
-#[async_trait::async_trait]
-impl<T: WasiView> Host for T {
+/// A trait for providing a `IpNameLookup` implementation.
+pub trait WasiNetworkView: Send {
+    /// Given a name, resolve to a list of IP addresses
     fn resolve_addresses(
```
alexcrichton, line 24: This might be best to model as an `async` function rather than returning an `AbortOnDropJoinHandle`?

crates/wasi/src/preview2/ip_name_lookup.rs
```diff
@@ -5,35 +5,73 @@ use crate::preview2::poll::{subscribe, Pollable, Subscribe};
...
-        if !network.allow_ip_name_lookup {
-            return Err(ErrorCode::PermanentResolverFailure.into());
+impl SystemNetwork {
+    /// Create a new `SystemIpNameLookup`
+    pub fn new(ctx: &WasiCtx) -> Self {
+        Self {
+            allowed: ctx.allowed_network_uses.ip_name_lookup,
         }
+    }
+}
```
alexcrichton, line 43: Could this perhaps manifest as a `impl WasiNetworkView for WasiCtx` to avoid a second type?

alexcrichton: (then it might be neat to have default implementations of `fn network_view` perhaps?)

crates/wasi/src/preview2/ip_name_lookup.rs
rylev: This is indeed very annoying... The value returned from `resolve_address` must be `Send + Sync` as it is used in `ResolveAddressStream` which impls `Subscribe` which requires the type implementing it to be `Send + Sync`.

The `#[async_trait]` macro puts a `Send` bound on the return type but not `Sync` and there is no way to instruct it to do so. We could move to using `Box<dyn Future>` but that means we loose the developer experience of `async_trait`. We could potentially move to using `impl Future` but that requires the latest version of Rust, and I'm not sure if that's ok.

crates/wasi/src/preview2/ip_name_lookup.rs
rylev: Great idea!

crates/wasi/src/preview2/ip_name_lookup.rs
```diff
@@ -5,35 +5,73 @@ use crate::preview2::poll::{subscribe, Pollable, Subscribe};
...
-        let host = parse(&name)?;
+/// The default implementation for `WasiIpNameLookupView`.
+#[derive(Debug, Clone, Default)]
+pub struct SystemNetwork {
+    allowed: bool,
+}
 
-        if !network.allow_ip_name_lookup {
-            return Err(ErrorCode::PermanentResolverFailure.into());
+impl SystemNetwork {
```
alexcrichton, line 36: Could this be removed in lieu of implementing `WasiNetworkView for WasiCtx` itself? (given that `WasiCtx` is needed to create this structure anyway it may not buy a whole lot)

crates/wasi/src/preview2/ip_name_lookup.rs
alexcrichton: Hm that's true, and yeah I see how there's no great way to manage this right now.

I suppose this can stay for now, but I'm still worried that this will inevitably cause some pain down the road. The best alternative I can think of is to return a `oneshot::Receiver<T>` instead but that's still pretty cumbersome to work with and wouldn't work great here.

rylev: We could take a page out of `WasiHttpView`'s playbook and have the function return a `wasmtime::Result<Resource<ResolveAddressStream>>` just like `WasiHttpView::send_request` returns a `wasmtime::Result<Resource<HostFutureIncomingResponse>>`. I don't really love this either, but at least it's consistent?

crates/wasi/src/preview2/ip_name_lookup.rs
rylev: If a user wants to have some logic where they conditionally switch between either their own network implementation and `SystemNetwork`, they need some way of constructing the `SystemNetwork`. Without this constructor, they would not be able to do so. We could do the following: 
* make all fields of `SystemNetwork` public
* Make the constructor just take a simple `bool` instead of a `WasiCtx`
* Leave it as is

I think I'd be most in favor of the last two options.

badeend:
Having thought about it some more I would like to offer one more suggestion. See code below.
This touches multiple discussions above, so tagging @alexcrichton.

The main difference is that this setup doesn't modify WasiView at all, but instead dispatches through WasiCtx::network

---

> @rylev: This does have the downside that the WasiCtx setting can be ignored by users.

In setup below that can't happen.

---

The strict separation between the "System*" and "Default*" structs below might be overkill for this PR because of the relative trivial "System*" and "Default*" implementations. However, for TCP/UDP sockets the implementation won't be so trivial anymore.

```rs
/// Trait for any WASI-compliant network implementation.
trait Network {
    fn resolve_addresses(&mut self, ) -> ...;

    // fn new_tcp_socket(family) -> ...
    // fn new_udp_socket(family) -> ...
}




/// A WASI-compliant "base" implementation.
/// Without any opinions on how to do filtering, permission checks or whatever.
struct SystemNetwork {}
impl SystemNetwork {
    fn new() -> Self {} // Probably no parameters
}
impl Network for SystemNetwork {
    fn resolve_addresses(&mut self, ) -> {
        // Perform the syscalls without any additional permission checks
    }
}




/// A "good enough" default implementation for Wasmtime users, _with_ permission checks.
struct DefaultNetwork {
    sys: SystemNetwork,
    allowed: bool
}
impl DefaultNetwork {
    /// Whatever parameters we want.
    /// Currently just a simple boolean, but could be extended to any 
    fn new(allowed: bool) -> Self {}
}
impl Network for DefaultNetwork {
    fn resolve_addresses(&mut self, ) -> {
        if self.allowed {
            self.sys.resolve_addresses()
        } else {
            Err(PermissionDenied)
        }
    }
}



pub struct WasiCtx {
    pub(crate) random: Box<dyn RngCore + Send + Sync>,
    pub(crate) insecure_random: Box<dyn RngCore + Send + Sync>,
    pub(crate) insecure_random_seed: u128,
    pub(crate) wall_clock: Box<dyn HostWallClock + Send + Sync>,
    pub(crate) monotonic_clock: Box<dyn HostMonotonicClock + Send + Sync>,
    pub(crate) env: Vec<(String, String)>,
    pub(crate) args: Vec<String>,
    pub(crate) preopens: Vec<(Dir, String)>,
    pub(crate) stdin: Box<dyn StdinStream>,
    pub(crate) stdout: Box<dyn StdoutStream>,
    pub(crate) stderr: Box<dyn StdoutStream>,

    // Remove `socket_addr_check`
    // Remove `allowed_network_uses`
    
    // Add:
    pub(crate) network: Box<dyn Network>,
}





pub struct WasiCtxBuilder {
    // (...)

    network: Box<dyn Network>,
}
impl WasiCtxBuilder {
    pub fn new() -> Self {
        Self {
            // (...)
            network: DefaultNetwork::new(/*allowed: */false),
        }
    }

    pub fn allow_network(&mut self, enable: bool) -> &mut Self {
        self.network = DefaultNetwork::new(enable);
        self
    }

    pub fn custom_network(&mut self, net: Box<dyn Network>) -> &mut Self {
        self.network = net;
        self
    }
}
```

crates/wasi/src/preview2/ip_name_lookup.rs
badeend: Of all the options mentioned so far, the `Box<dyn Future>` route seems the least worst to me.

The fact that we currently use a quick&dirty implementation using spawn_blocking should be an implementation detail of `SystemNetwork`

alexcrichton: Returning a `Resource<ResolveAddressStream>` would be nice yeah but it seems like that would have all the same design questions around "how do you actually create one of those". 

Although as I say that we could perhaps get around `async trait` by having something like `fn from_future<T: Future + Send + Sync + 'static> (...)` on the `ResolveAddressStream` type. That way you could pass an `async` block which would have bounds generated correctly and additioanlly the handle from `spawn_blocking` could also be passed in? That way we might be able to avoid async traits altogether and rely entirely on that.

crates/wasi/src/preview2/ip_name_lookup.rs
alexcrichton: Hm I'm not sure I understand? Given `impl WasiNetworkView for WasiCtx` then if users are themselves implementing `WasiNetworkView` for their own custom type then conditionally using the default behavior would be calling `self.ctx.the_method_name()` so I'm not sure where the need for having this separate `SystemNetwork` structure arises?

alexcrichton:
@badeend your sketch makes sense yeah, but my thoughts at https://github.com/bytecodealliance/wasmtime/issues/7694#issuecomment-1858351478 showcase how something like that breaks down unless all callbacks about configuration go through `trait Network`, so for example we'd also need to route IP address checks through that trait as well. (not a problem of course, just pointing out that once things are behind a trait everything needs to be behind that trait to share state between callbacks).

Otherwise though I think it'd be reasonable to configure a trait object on `WasiCtx` vs returning a `&mut dyn Trait` from `WasiView` (although it having a default impl returning `self.ctx_mut()` I think is pretty nifty and makes it almost hidden.

crates/wasi/src/preview2/ip_name_lookup.rs
rylev: Thinking about this more, returning `Resource<ResolveAddressStream>` would mean that the implementor would have to also implement `WasiCtx` or otherwise have multiple access to the `ResourceTable`. Right now we don't require that from the implementor of `WasiNetworkView`, and I'm not sure we want to.

alexcrichton: `ResolveAddressStream` itself could be returned though leaving the table management bits to the glue?

rylev:
I like @badeend's suggestion. I initially shared the same concern that @alexcrichton pointed out, but I think it should be relatively easy for users to implement the `Network` trait by delegating all the parts they don't care about to `SystemNetwork` or `DefaultNetwork` and only changing what they do care about. The only issue I can see with this is that if the `Network` trait grows large (which I think we anticipate it will), then there will be a large amount of boiler plate of delegating `Network` methods to the `SystemNetwork` or `DefaultNetwork`. This could be really annoying.

rylev:
@alexcrichton @badeend I've pushed a new change that moves more towards @badeend's design as well as changing the return type of `resolve_addresses` to be a `ResolveAddressStream`. 

I think the only way for us to be sure about this design is to keep going and implement UDP and TCP on top of it. With that in mind, I think we should merge this PR and continue the design exploration in a follow up. Thoughts?

badeend:
>  something like that breaks down unless all callbacks about configuration go through trait Network, so for example we'd also need to route IP address checks through that trait as well.

Correct. In my mind that was exactly the goal of this (and upcoming) PRs. `socket_addr_check` & `allowed_network_uses` would most likely be moved to `DefaultNetwork`

> I think the only way for us to be sure about this design is to keep going and implement UDP and TCP on top of it.

Agree.

> With that in mind, I think we should merge this PR and continue the design exploration in a follow up. Thoughts?

I'm not familiar with wasmtime's release process, but if a release gets cut partway through these changes it could be weird for consumers when e.g. IP lookup & TCP use the Network trait but UDP still uses an old & completely different customization mechanism. I'll leave this up for @alexcrichton

crates/wasi/src/preview2/ip_name_lookup.rs
```diff
@@ -2,39 +2,44 @@ use crate::preview2::bindings::sockets::ip_name_lookup::{Host, HostResolveAddres
...
+    }
+
+    pub fn done(result: std::io::Result<Vec<IpAddr>>) -> Self {
+        Self::Done(result.map(|v| v.into_iter()))
+    }
 }
 
 #[async_trait::async_trait]
-impl<T: WasiView> Host for T {
+impl<T: WasiView + Sized> Host for T {
```
alexcrichton, line 33: I think the `+ Sized` may not be necessary here?

alexcrichton (commented):
Looks reasonable to me!

Unless this is particularly urgent though I would personally agree with @badeend in that I'd prefer to see a more complete picture before landing if possible. If this is needed to unblock work I think it's ok to land, but if you're ok working with a fork for now though I think it'd be best to sketch out the other checks you're thinking of implementing to be able to see them altogether at once.

rylev:
I'm fine with not landing this on main. I just think we need to decide which fork/branch is the integration point so that if @badeend and I continue the work, we can bring that all together into one unified branch that will ultimately be merged into main.

badeend:
@alexcrichton One of the (many..) changes in this PR is that the actual socket code has been separated out from the component trait implementation to its own `SystemTcpSocket`/`SystemUdpSocket`. To reduce the size of this PR, do you mind if I split that refactor off from this PR and submit a new one with just those changes? These changes will only be internal reshuffling of code, without affecting the public interface.

alexcrichton:
Sounds good to me! And yeah refactorings are always good to land at any time

PR #7707, opened 2023-12-19
winch: Multi-Value Part 2: Blocks

saulecabrera:
Follow-up to: https://github.com/bytecodealliance/wasmtime/pull/7535

This commit adds support for the Multi-Value proposal for blocks.

In general, this change, introduces multiple building blocks to enable supporting arbitrary params and results in blocks:

* `BlockType`: Introduce a block type, to categorize the type of each block, this makes it easier to categorize blocks per type and also makes it possible to defer the calculation of the `ABIResults` until they are actually needed rather than calculating everyghing upfront even though they might not be needed (when in an unreachable state).
* Push/pop operations are now frame aware. Given that each `ControlStackFrame` contains all the information needed regarding params and results, this change moves the the implementation of the push and pop operations to the `ControlStackFrame` struct.
* `StackState`: this struct holds the entry and exit invariants of each block; these invariants are pre-computed when entering the block and used throughout the code generation, to handle params, results and assert the respective invariants.

In terms of the mechanics of the implementation: when entering each block, if there are results on the stack, the expected stack pointer offsets will be calculated via the `StackState`, and the `target_offset` will be used to create the block's `RetArea`. Note that when entering the block and calculating the `StackState` no space is actually reserved for the results, any space increase in the stack is deffered until the results are popped from the value stack via
`ControlStackFrame::pop_abi_results`.

The trickiest bit of the implementation is handling constant values that need to be placed on the right location on the machine stack. Given that constants are generally not spilled, this means that in order to keep the machine and value stack in sync (spilled-values-wise), values must be shuffled to ensure that constants are placed in the expected location results wise. See the comment in `ControlStackFrame::adjust_stack_results` for more details.

---

I've been continuously fuzzing this change for a bit now (~48hrs), and the fuzzer hasn't reported any new issues.


elliottt:
Sorry this has sat for so long! I'll get a review done by EOD today.

winch/codegen/src/codegen/call.rs
```diff
@@ -330,16 +333,28 @@ impl FnCall {
...
             }
         };
 
         // Free the bytes consumed by the call.
         masm.free_stack(stack_consumed);
 
-        let mut results_data = ABIResultsData::wrap(sig.results.clone());
-        results_data.ret_area = ret_area;
+        if let Some(area) = ret_area {
+            if stack_consumed > 0 {
```
elliottt, line 344: Is it worth adding an assert here that the area is not `Uninit`?

elliottt (approved):
This looks great, thank you for all the comments!

winch/codegen/src/isa/x64/masm.rs
```diff
@@ -672,7 +672,7 @@ impl Masm for MacroAssembler {
     }
 
     fn epilogue(&mut self, locals_size: u32) {
-        assert!(self.sp_offset == locals_size);
+        debug_assert_eq!(self.sp_offset, locals_size);
```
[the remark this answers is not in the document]
saulecabrera: I meant to convert it to an `assert_eq!` instead. I pushed a fixed that changes it to `assert_eq` instead.

winch/codegen/src/codegen/call.rs
saulecabrera: Good point, I added an assert!

PR #7710, opened 2023-12-20
Add `sextend_maybe` and `uextend_maybe` to opt ISLE

scottmcm:
Working on something else, I noticed that the `(x < y) != 0` opt was written in a way that only works for C

https://github.com/bytecodealliance/wasmtime/blob/f8c9f6711f6316001801480d3883888734835faf/cranelift/codegen/src/opts/icmp.isle#L16-L25

whereas in Rust, which doesn't have the "usual arithmetic conversions", the result of the comparison stays as `I8` and thus without the extension it didn't match that pattern.

This picks up a suggestion that alex had made in a previous PR (https://github.com/bytecodealliance/wasmtime/pull/7636#discussion_r1416208525) to make something like lowering ISLE's `maybe_uextend` extractor for optimization ISLE.

They're then used to improve the mentioned icmp-of-icmp pattern

https://github.com/bytecodealliance/wasmtime/commit/8ef0b8181ca23b9133a464da6ae10fa32ffdbf15#diff-fa420f878e321dc6579ac911f9ba8820a9381e4c49615584f6e6909f0e329dfcL124-R125

as well as simplify some of the duplication in the `<=>` patterns.

### DIFF
--- a/cranelift/codegen/src/opts.rs
+++ b/cranelift/codegen/src/opts.rs
@@ -5,6 +5,7 @@ use crate::ir::condcodes;
 pub use crate::ir::condcodes::{FloatCC, IntCC};
 use crate::ir::dfg::ValueDef;
 pub use crate::ir::immediates::{Ieee32, Ieee64, Imm64, Offset32, Uimm8, V128Imm};
+use crate::ir::instructions::InstructionFormat;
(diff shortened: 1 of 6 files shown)


fitzgen (approved):
Nice!

PR #7711, opened 2023-12-20
Optimize more reduction-of-an-extend cases

scottmcm:
I was exploring `ireduce` patterns more after I removed them from https://github.com/bytecodealliance/wasmtime/pull/7693#discussion_r1430497487 at **fitzgen**'s request.

In doing so, I noticed that there were some simpler cases missing first, like `ireduce.i8 sextend.i32 my_i16` → `ireduce.i8 my_i16` and `ireduce.i16 sextend.i32 my_i8` → `sextend.i16 my_i8`.

So this PR adds those (and the unsigned equivalents).

### DIFF
--- a/cranelift/codegen/src/opts/extends.isle
+++ b/cranelift/codegen/src/opts/extends.isle
@@ -47,8 +47,24 @@
 
 ;; A reduction-of-an-extend back to the same original type is the same as not
 ;; actually doing the extend in the first place.
-(rule (simplify (ireduce ty (sextend _ x @ (value_type ty)))) x)
-(rule (simplify (ireduce ty (uextend _ x @ (value_type ty)))) x)
+(rule (simplify (ireduce ty (sextend _ x @ (value_type ty)))) (subsume x))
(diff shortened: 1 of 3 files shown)


fitzgen:
Thanks for splitting this out!

Would you mind enabling the `trace-log` feature and looking at [this trace log](https://github.com/bytecodealliance/wasmtime/blob/61e1cdff5a90461ab8de471e100de5b824e079d1/cranelift/codegen/src/egraph.rs#L435) when running `wasmtime compile sightglass/benchmarks/spidermonkey/benchmark.wasm`[^0] and comparing the numbers before and after this PR? I want to make sure we aren't accidentally blowing up the size of the egraph.

[^0]: github.com/bytecodealliance/sightglass

fitzgen (approved):
Stats look good -- thanks for getting those!

alexcrichton:
The first failure above was spurious, but the [second failure](https://github.com/bytecodealliance/wasmtime/actions/runs/7389175948/job/20101482909) looks like it might be related to this PR

PR #7712, opened 2023-12-20
feat: implement `ResourceAny` -> `Resource<T>` conversion

### LINKED ISSUE #7704: Implement `ResourceAny` -> `Resource<T>` conversion
#### Feature

Refs #7688

Currently, developers can define resources via `Linker::resource`, but they can not get a handle to the `Resource<T>` from a `ResourceAny`, which is passed to e.g. `Linker::func_new` functions.

#### Benefit

Most importantly, it allows usage of resources for dynamically-typed functions, e.g. such as when WIT definition is not known at compilation time

#### Implementation

This should quite similar to #7688, we should make sure to account for the own/borrow state and possibly call the destructor of on `Drop`

#### Alternatives

The only alternative seems to be not supporting dynamically-typed functions utilizing resources, developers may use macros/templating to generate all possible type bounds of imported functions at compile time and choose the right one at runtime, but that is very complex and hard to maintain

A reasonable alternative could be using the "partially-typed" bound with https://github.com/bytecodealliance/wasmtime/issues/7701, but it still seems preferable to support fully-dynamic use cases

rvolosatovs:
Closes #7704 
Follow-up on #7688  (based on that PR for simplicity) 

Once #7688 is merged, I'll rebase such that only https://github.com/bytecodealliance/wasmtime/pull/7712/commits/5b20daf013bf064a32c53aefc93e7d4d52473c86 would be in this PR. Also happy to move the commit to #7688 if you prefer

### DIFF
--- a/crates/wasmtime/src/component/resources.rs
+++ b/crates/wasmtime/src/component/resources.rs
@@ -397,6 +397,32 @@ where
         })
     }
 
+    /// Attempts to convert a [`ResourceAny`] into [`Resource`].
+    pub fn try_from_resource_any(
+        ResourceAny { idx, ty, own_state }: ResourceAny,
+        mut store: impl AsContextMut,
+    ) -> Result<Self> {
+        let store = store.as_context_mut();
+        let store_id = store.0.id();
+        let mut tables = host_resource_tables(store.0);
+        assert_eq!(ty, ResourceType::host::<T>(), "resource type mismatch");
+        let (state, rep) = if let Some(OwnState { store, dtor, flags }) = own_state {
+            assert_eq!(store_id, store, "wrong store used to convert resource");
+            assert!(dtor.is_some(), "destructor must be set");
+            assert!(flags.is_none(), "flags must not be set");
+            let rep = tables.resource_lift_own(None, idx)?;
+            (AtomicU32::new(NOT_IN_TABLE), rep)
+        } else {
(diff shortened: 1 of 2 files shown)


rvolosatovs:
cc @alexcrichton

crates/wasmtime/src/component/resources.rs
```diff
@@ -397,6 +397,32 @@ where
...
+        mut store: impl AsContextMut,
+    ) -> Result<Self> {
+        let store = store.as_context_mut();
+        let store_id = store.0.id();
+        let mut tables = host_resource_tables(store.0);
+        ensure!(ty == ResourceType::host::<T>(), "resource type mismatch");
+        let (state, rep) = if let Some(OwnState { store, dtor, flags }) = own_state {
+            ensure!(store_id == store, "wrong store used to convert resource");
+            ensure!(dtor.is_some(), "destructor must be set");
+            ensure!(flags.is_none(), "flags must not be set");
```
alexcrichton, line 412: Could these be changed to asserts to match `Resource::lift_from_index`?

alexcrichton (commented):
Looks reasonable to me, thanks!

There's one minor API-design question where it's not possible to recover from this method if it fails due to consumption of the `ResourceAny` input. That being said the main failure condition, the wrong type, can be tested manually ahead of time, so I think that's ok.

alexcrichton (approved):

PR #7718, opened 2023-12-21
feat: support `RuntimeImportIndex` lookups by string paths

### LINKED ISSUE #7714: Support `Resource<T> -> ResourceAny` for WASI resources
#### Feature

With #7688 implemented it is now possible to take `Resource<T>` to `ResourceAny`, however the conversion does not work for resources provided by `wasmtime_wasi` crate, since the `ResourceImportIndex` is not exposed to the developer for those resources. 

This is particularly important for dynamically-defined functions defined via `Linker::func_new`, which operate on `wasmtime::Val` directly. For example, it is currently not possible to return a `wasi:io/poll/pollable` resource from such a function without defining a custom `wasi:io/poll/pollable` implementation in the linker to acquire the import index to pass to `try_into_resource_any` in order to convert it to `ResourceAny`

#### Benefit

This allows dymanically-typed function definitions to operate on resources provided by WASI, i.e. it removes the need for developers using dynamically-typed functions utilizing WASI resources from reimplementing (parts of) WASI

#### Implementation

Since lookups by string import paths were considered too slow for #7688, it appears that the only alternative would be adapting `wit-bindgen` (e.g https://github.com/bytecodealliance/wasmtime/blob/61e1cdff5a90461ab8de471e100de5b824e079d1/crates/wit-bindgen/src/lib.rs#L810) to capture the indexes returned by `LinkerInstance::resource` calls in a table of some sorts returned by `add_to_linker`, if we still want to avoid the string lookups, that could be a nested `struct` type, where an index lookup could look something like this:

```
let idx = resource_definitions.wasi.io.poll.pollable
```

#### Alternatives

Allow string lookups for the `RuntimeImportIndex` as an alternative to the `ResourceImportIndex` computed at `LinkerInstance::resource`

rvolosatovs:
Follow-up on #7688
Closes #7714 (a more performant solution should probably exist, but this addresses the immediate issue)

### DIFF
--- a/crates/wasmtime/src/component/instance.rs
+++ b/crates/wasmtime/src/component/instance.rs
@@ -558,15 +558,27 @@ impl<T> InstancePre<T> {
         }
     }
 
-    pub(crate) fn resource_import_index(
-        &self,
-        path: ResourceImportIndex,
-    ) -> Option<RuntimeImportIndex> {
-        *self.resource_imports.get(path)?
+    /// Returns [`RuntimeImportIndex`] associated with this resource.
+    /// `idx` is the [`ResourceImportIndex`] returned by [`Linker::resource`].
+    ///
+    /// [`Linker::resource`]: crate::component::LinkerInstance::resource
+    pub fn resource_import_index(&self, idx: ResourceImportIndex) -> Option<RuntimeImportIndex> {
+        *self.resource_imports.get(idx)?
     }
 
-    pub(crate) fn resource_import(&self, path: ResourceImportIndex) -> Option<&RuntimeImport> {
-        let idx = self.resource_import_index(path)?;
+    /// Returns [`RuntimeImportIndex`] associated with this `path` within a `root`.
+    pub fn path_import_index(&self, root: &str, path: &[&str]) -> Option<RuntimeImportIndex> {
+        let env_component = self.component().env_component();
+        env_component
+            .imports
+            .iter()
+            .find_map(|(idx, (import, import_path))| {
+                let (root_name, _) = env_component.import_types.get(*import)?;
+                (root_name == root && import_path == path).then_some(idx)
+            })
+    }
+
+    pub(crate) fn runtime_import(&self, idx: RuntimeImportIndex) -> Option<&RuntimeImport> {
         self.imports.get(idx)
     }
 
--- a/crates/wasmtime/src/component/resources.rs
+++ b/crates/wasmtime/src/component/resources.rs
@@ -1,6 +1,5 @@
 use crate::component::func::{bad_type_info, desc, LiftContext, LowerContext};
 use crate::component::instance::RuntimeImport;
-use crate::component::linker::ResourceImportIndex;
 use crate::component::matching::InstanceType;
 use crate::component::{ComponentType, InstancePre, Lift, Lower};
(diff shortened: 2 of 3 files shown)


alexcrichton:
I personally fall on the conservative side of API design, so my knee-jerk reaction is to not include something like this. Could you expand a bit more on the motivation, however?

rvolosatovs:
> I personally fall on the conservative side of API design, so my knee-jerk reaction is to not include something like this. Could you expand a bit more on the motivation, however?

Sure, this is primarily to address https://github.com/bytecodealliance/wasmtime/issues/7714 and potentially help with https://github.com/bytecodealliance/wasmtime/issues/7726.
Most important reason for this to exist is that the embedder may not necessarily be directly calling e.g. `Linker::resource` and therefore may not have the `ResourceImportIndex` associated with it - this is the case today, for example, for any resources added to the linker via generated `add_to_linker` from bindgen. To address this `add_to_linker` would need to return a mapping of resource type import paths (expressed as strings, generated statically-typed structs or otherwise) to their respective import indexes. The intention of this change is therefore to:
- Serve as a opt-in simple solution for use cases where initial linking performance is not a bottleneck/not critical and/or implementation simplicity and readability are more important - this is also relevant for prototyping.
- While a more performant solution (hooking into the bindgen, I suppose) is built, unblock WASI resource `ResourceAny` conversion.
- (in the general case) Provide for use cases where the `Linker::resource` calls are wrapped in an external API. (in case that Wasmtime WASI implementation is updated to expose the mapping somehow, third parties may still provide APIs wrapping `Linker::resource` without exposing the indexes and lack of path-based lookups, even if less performant than they could otherwise be, would block users from benefiting from those said APIs)

alexcrichton:
Thanks for expanding! What you're saying makes sense and give me a better understanding as to the pieces here.

I'd be inclined to go the route of #7714 first, however. This PR is exposing yet-more-details about the internal representation of components which otherwise needs documentation (e.g. the doc blocks here don't really explain anything about the deeper meaning of the constructs here). For example:

* Currently `RuntimeImportIndex` is a "private" type which is just an implementation detail of Wasmtime. This is exporting it to the public interface without explaining what it means or represents. Under the hood it's sort of an irrelevant implementation detail to embedders where it happens to be the specific order of arguments being passed to the component and it can be used to lookup the runtime type used for a particular import.
* Additionally the concept of a "path" for an import is introduced here in the public API. The `Linker` for example does not operate on this and instead works with instances.

I understand that this is an "easy" PR but that's typically how these things go. Wasmtime doesn't implement the real feature here (#7714) so it's easier to basically make more things `pub` than to implement the real feature. The cost of this though is maintenance over time. These APIs will need to be documented and maintained and kept in sync with all other component-related APIs. Additionally the runtime functionality would need to be preserved here across refactorings of Wasmtime's internals and all that.

More-or-less I understand that this is an easy PR and gets the job done, but if you're ok I'd prefer to go down the route of #7714 first to avoid adding two more concepts to the component embedding API (runtime import indices and import paths). If it's necessary to add them then it's necessary but if possible I'd prefer to avoid adding them since the embedding API is already quite complicated.

fitzgen:
Closing due to inactivity, feel free to reopen if this was a mistake.

PR #7719, opened 2023-12-21
Simplify ireduce of modular operators, when it mean strictly fewer instructions

scottmcm:
This is a follow-up to **fitzgen**'s comment in https://github.com/bytecodealliance/wasmtime/pull/7693#discussion_r1430497487

The version that was removed from that PR had some concerns because it added things to the egraph that might not be useful, since it added *more* instructions, potentially: going from `(small)(x OP y)` to `(small)x OP (small)y` isn't always a win since it's one more CLIF instruction.

This PR addresses that concern by restricting the pattern only to places where it's going to be *strictly fewer* total instructions.  It does that by only applying when all the arguments to the inner operator are either extended or constants.  In those cases, pushing the `ireduce` into the arguments will collapse with the instruction that's already there -- either because extend-then-reduce back is a nop, because cprop will simplify it, or thanks to #7711.

The particular motivating example for this -- which is why the `mulhi` pattern is in here too -- is Rust's [`u64::widening_mul`](https://doc.rust-lang.org/std/primitive.u64.html#method.widening_mul).  LLVM doesn't have an intrinsic for wide multiplication, so the way it's often written is by casting the arguments to a wider type, doing the multiplication, then reducing back again.  But that means the obvious translation in CLIF ends up with an expensive 128-bit multiplication.  Collapsing that down to a 64-bit `imul`+`umulhi` is much cheaper.

### DIFF
--- a/cranelift/codegen/src/opts/arithmetic.isle
+++ b/cranelift/codegen/src/opts/arithmetic.isle
@@ -151,3 +151,19 @@
       (subsume (bxor ty (bxor ty a b) (bxor ty c d))))
 (rule (simplify (bxor ty (bxor ty (bxor ty a b) c) d))
       (subsume (bxor ty (bxor ty a b) (bxor ty c d))))
+
+;; Detect people open-coding `mulhi`: (x as big * y as big) >> bits
+;; LLVM doesn't have an intrinsic for it, so you'll see it in code like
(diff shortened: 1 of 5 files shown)


fitzgen (approved):
Thanks!

alexcrichton:
A bisection I just ran points to this PR as the cause of https://github.com/bytecodealliance/wasmtime/issues/7874 (cc'ing here to close the loop)

PR #7721, opened 2023-12-21
winch: Add memory instructions

saulecabrera:
This commit adds support for the following memory instructions to winch:

* `data.drop`
* `memory.init`
* `memory.fill`
* `memory.copy`
* `memory.size`
* `memory.grow`

In general the implementation is similar to what other instructions via builtins are handled (e.g. table instructions), which involve stack manipulation prior to emitting a builtin function call, with the exception of `memory.size`, which involves loading the current length from the `VMContext` pointer.


winch/codegen/src/isa/x64/masm.rs
```diff
@@ -206,6 +206,30 @@ impl Masm for MacroAssembler {
...
+
+        let size_addr = Address::offset(base, heap_data.current_length_offset);
+        self.asm.mov_mr(&size_addr, size_reg, self.ptr_size);
+        // Prepare the stack to emit division and get the size in pages rather
+        // than in bytes.
+        context
+            .stack
+            .push(TypedReg::new(heap_data.ty, size_reg).into());
+        context.stack.push(Val::i64(WASM_PAGE_SIZE as i64));
+        self.div(context, DivKind::Unsigned, heap_data.ty.into());
```
alexcrichton, line 230: Given that the page size is a power-of-two could this be hardcoded as a right-shift instead of a division?

alexcrichton (approved):

winch/codegen/src/isa/x64/masm.rs
saulecabrera: Indeed, fixed! Thanks for the suggestion, using a shift should also help with register pressure (at least for x64).

PR #7723, opened 2023-12-22
Fix clif-util run fails directly when target is specified

penguin-wwy:
In testing.md doc, The `test run` command requires the `target`
https://github.com/bytecodealliance/wasmtime/blob/b583c54fda13b53dea362861125dd1e2ced1381d/cranelift/docs/testing.md?plain=1#L298-L300

However, when executing the `clif-util run` command (e.g `clif-util run cranelift/filetests/filetests/runtests/const.clif`), any file is found to fail. Because it is not allowed to specify a `target` in the `test run` command in the actual code.
https://github.com/bytecodealliance/wasmtime/blob/b583c54fda13b53dea362861125dd1e2ced1381d/cranelift/src/run.rs#L114-L117

In this PR I've modified it to filter only on architectures, as described in the documentation. The host platform's native targets will be used to actually compile the tests.


fitzgen (approved):
Thanks!

Issue #7716, opened 2023-12-21
Failed to instantiate the example world by component model


k82cn:
### Test Case

```
package component:matrix;

interface service {
    enum flame-error {
        internal,
    }
    type task-input = list<u8>;
    type task-output = list<u8>;

    record session-context {
        session-id: string,
        common-data: option<list<u8>>,
    }

    record task-context {
        task-id: string,
        session-id: string,
    }
    
    // handle request function
    on-session-enter: func(ctx: session-context) -> result<_, flame-error>;
    on-session-leave: func(ctx: session-context) -> result<_, flame-error>;
    on-task-invoke: func(ctx: task-context, input: option<task-input>) -> result<option<task-output>, flame-error>;
}

// exports the interface
world flame {
    export service;
}
```

### Steps to Reproduce

* build wasm `cargo component build -p matrix-server`
* build wasm main `cargo build -p matrix-local`
* run wasm main `./target/debug/matrix-local`

**References**:

* matrix-local: https://github.com/xflops/flame/pull/75/files#diff-448546749b6625cbca1a43c400f6d86637db5f014968c96ece09705e07f65c7c
* matrix-server (wasm): https://github.com/xflops/flame/pull/75/files#diff-e21ed78fa8d569631fed0fdddc6c99d585fc4d25806da0c7c00c4ebb2bacc8e9

### Expected Results

Execute the callbacks accordingly, e.g. on_session_enter.

### Actual Results

```
$ ./target/debug/matrix-local
Error: Internal("Failed to instantiate the flame world")
```

### Versions and Environment

Wasmtime version or commit: 
```
wasmtime = "16"
wasmtime-wasi = "16"
anyhow = "1"
```

```
$ rustc --version
rustc 1.74.1 (a28077b28 2023-12-04)
$ cargo --version
cargo 1.74.1 (ecb9851af 2023-10-18)
```

Operating system: `Ubuntu 22.04`

Architecture: x86

k82cn:
The same issue with the add example:

```
$ ./target/debug/host-add
Error: Failed to instantiate the example world

Caused by:
    0: import `wasi:cli/environment@0.2.0-rc-2023-11-10` has the wrong type
    1: instance export `get-environment` has the wrong type
    2: expected func found nothing
```

k82cn:
It works when rollback to wasmtime `15`.

alexcrichton:
[...]

PR #7728, opened 2023-12-27
Add more necessary linkers to wasi-http

iawia002:
When I attempt to run the [hello-wasi-http](https://github.com/sunfishcode/hello-wasi-http) example using wasmtime 16, I encounter the following error(which did not occur in previous versions):

```
$ wasmtime serve hello_wasi_http.wasm

Error: import `wasi:cli/environment@0.2.0-rc-2023-12-05` has the wrong type

Caused by:
    0: instance export `get-environment` has the wrong type
    1: expected func found nothing
```

 I realize that this may be related to PR #7597, where we might have trimmed too much "world" to the point where a simple HTTP hello world example cannot run. So, this patch adds some content, and now the example can run successfully:

```
$ ./target/debug/wasmtime serve hello_wasi_http.wasm

Serving HTTP on http://0.0.0.0:8080/
```

Related: #7716

### DIFF
--- a/crates/wasi-http/src/proxy.rs
+++ b/crates/wasi-http/src/proxy.rs
@@ -6,12 +6,16 @@ wasmtime::component::bindgen!({
     tracing: true,
     async: true,
     with: {
+        "wasi:cli/environment": preview2::bindings::cli::environment,
+        "wasi:cli/exit": preview2::bindings::cli::exit,
         "wasi:cli/stderr": preview2::bindings::cli::stderr,
         "wasi:cli/stdin": preview2::bindings::cli::stdin,
         "wasi:cli/stdout": preview2::bindings::cli::stdout,
         "wasi:clocks/monotonic-clock": preview2::bindings::clocks::monotonic_clock,
         "wasi:clocks/timezone": preview2::bindings::clocks::timezone,
         "wasi:clocks/wall-clock": preview2::bindings::clocks::wall_clock,
+        "wasi:filesystem/preopens": preview2::bindings::filesystem::preopens,
+        "wasi:filesystem/types": preview2::bindings::filesystem::types,
         "wasi:http/incoming-handler": bindings::http::incoming_handler,
         "wasi:http/outgoing-handler": bindings::http::outgoing_handler,
         "wasi:http/types": bindings::http::types,
@@ -27,9 +31,13 @@ where
 {
     preview2::bindings::clocks::wall_clock::add_to_linker(l, |t| t)?;
     preview2::bindings::clocks::monotonic_clock::add_to_linker(l, |t| t)?;
+    preview2::bindings::filesystem::preopens::add_to_linker(l, |t| t)?;
+    preview2::bindings::filesystem::types::add_to_linker(l, |t| t)?;
     preview2::bindings::io::poll::add_to_linker(l, |t| t)?;
     preview2::bindings::io::error::add_to_linker(l, |t| t)?;
     preview2::bindings::io::streams::add_to_linker(l, |t| t)?;
+    preview2::bindings::cli::environment::add_to_linker(l, |t| t)?;
+    preview2::bindings::cli::exit::add_to_linker(l, |t| t)?;
     preview2::bindings::cli::stdin::add_to_linker(l, |t| t)?;
     preview2::bindings::cli::stdout::add_to_linker(l, |t| t)?;
     preview2::bindings::cli::stderr::add_to_linker(l, |t| t)?;


alexcrichton:
Thanks for the PR, but these interfaces are all [explicitly missing from the standard `proxy` world](https://github.com/WebAssembly/wasi-http/blob/main/wit/proxy.wit). The `wasmtime serve` subcommand [has an `-Scommon` flag](https://github.com/bytecodealliance/wasmtime/blob/b583c54fda13b53dea362861125dd1e2ced1381d/src/commands/serve.rs#L202-L216) to enable the interfaces in an opt-in basis. Otherwise though this was an intentional behavior change from https://github.com/bytecodealliance/wasmtime/pull/7597

iawia002:
OK, but I'm getting a little confused, I know `-Scommon` flag will make it work because it adds all the linkers, in other words, can I say `wasmtime serve` won't work without the `-Scommon` flag? Since I'm just running the basic hello world example.

> but these interfaces are all [explicitly missing from the standard proxy world](https://github.com/WebAssembly/wasi-http/blob/main/wit/proxy.wit).

So do we count this as an issue? I'm not familiar with this, I can't figure out why the proxy world needs those.

pchickey:
The issue is that the https://github.com/sunfishcode/hello-wasi-http instructions aren't up to date with wasmtime 16, which requires passing in the `-Scommon` flag. Please file an issue or PR over there to add that command line flag.

iawia002:
> The issue is that the https://github.com/sunfishcode/hello-wasi-http instructions aren't up to date with wasmtime 16, which requires passing in the `-Scommon` flag. Please file an issue or PR over there to add that command line flag.

Actually, I made some updates based on hello-wasi-http to adapt to wasmtime 16. I updated [cargo-component](https://github.com/bytecodealliance/cargo-component) version to `0.6.0` and updated `wasi:http/proxy` to `0.2.0-rc-2023-12-05`, if you use the original hello-wasi-http directly, even if the `-Scommon` flag is added, it still cannot run (because the wit definition does not match):

```
wasmtime serve -Scommon hello_wasi_http.wasm
Error: import `wasi:cli/stdout@0.2.0-rc-2023-10-18` has the wrong type

Caused by:
    0: instance export `get-stdout` has the wrong type
    1: expected func found nothing
```

iawia002:
My code: https://github.com/iawia002/hello-wasi-http/tree/adapt-wasmtime-16
The diff (basically only version and dependency updates): https://github.com/iawia002/hello-wasi-http/commit/8fa967735f40988badf3a473459cb3e4c9cb8595
Compiled wasm file: [hello_wasi_http.wasm.zip](https://github.com/bytecodealliance/wasmtime/files/13826925/hello_wasi_http.wasm.zip)

alexcrichton:
It looks like you may be running into versioning issues, so I'd recommend double-checking that the adapter in use and the Wasmtime in use are from the same version. I'll also point out that there's an [adapter specific to a proxy](https://github.com/bytecodealliance/wasmtime/releases/download/v16.0.0/wasi_snapshot_preview1.proxy.wasm) which avoids the need for `-Scommon`

iawia002:
I'm using wasmtime 16 and [cargo-component 0.6.0](https://github.com/bytecodealliance/cargo-component/releases/tag/v0.6.0) which adapts wasmtime 16 and `wasi:http/proxy@0.2.0-rc-2023-12-05` which exactly the same with [wit definition in wasmtime 16](https://github.com/bytecodealliance/wasmtime/tree/release-16.0.0/crates/wasi-http/wit/deps).

```
$ wasmtime -V
wasmtime-cli 16.0.0 (6613acd1e 2023-12-20)
```

So I think I've taken care of versioning issues.

alexcrichton:
Are you still getting an error about `wasi:cli/stdout@0.2.0-rc-2023-10-18`? If so that means there's a versioning problem somewhere because 2023-10-18 is a more historical version than Wasmtime 16.

iawia002:
Oops, @alexcrichton I think you misunderstood, @pchickey says I'm encountering this issue due to the  https://github.com/sunfishcode/hello-wasi-http not adapting with wasmtime 16, I'm trying to explain to you that I'm not using the original hello-wasi-http example and I did some update to make it adapt with wasmtime 16, but even I make those updates, I'm still encountering the issue in the PR description.

Using https://github.com/sunfishcode/hello-wasi-http, with or without `-Scommon` flag, we'll encounter this error:

```
$ wasmtime serve -Scommon hello_wasi_http.wasm
Error: import `wasi:cli/stdout@0.2.0-rc-2023-10-18` has the wrong type

Caused by:
    0: instance export `get-stdout` has the wrong type
    1: expected func found nothing
```

Using https://github.com/iawia002/hello-wasi-http/tree/adapt-wasmtime-16 which adapts wasmtime 16, with `-Scommon` flag, it works, but without it, we'll encounter this error in the PR description:

```
$ wasmtime serve hello_wasi_http.wasm

Error: import `wasi:cli/environment@0.2.0-rc-2023-12-05` has the wrong type

Caused by:
    0: instance export `get-environment` has the wrong type
    1: expected func found nothing
```

Though it will work with `-Scommon` flag, I still want to ask the same question from my previous comments:

> I know `-Scommon` flag will make it work because it adds all the linkers, in other words, can I say `wasmtime serve` won't work without the `-Scommon` flag? Since I'm just running the basic hello world example.

alexcrichton:
This all boils down to the adapter you're using. If you use the `*.reactor.wasm` adapter then yes that won't work for `wasmtime serve` because it's not intended to work with `wasmtime serve`. If you use `*.proxy.wasm` then it will work with `wasmtime serve` because it's intended to work with the subcommand.

What you're basically running into here is difficulties and road bumps in how a component is built. This doesn't have all that much to do with the runtime per se. This is definitely something we need more documentation and guidance on, and we're doing our best on that.

PR #7740, opened 2024-01-02
Update several dependencies to windows-sys 0.52.

sunfishcode:
Update cap-std, rustix, io-extras, errno, fd-lock, fs-set-times, and winx to versions which use windows-sys 0.52.

I started out hoping that everything could be updated to windows-sys 0.52, however tokio and mio have not yet updated to windows-sys 0.52 due to supporting an older MSRV.


elliottt (approved):
Looks great!

PR #7747, opened 2024-01-04
remove authority and scheme from http header

### LINKED ISSUE #7737: wasmtime-wasi-http includes absolute URI in HTTP requests
Thanks for filing a bug report! Please fill out the TODOs below.

**Note: if you want to report a security issue, please read our [security policy](https://bytecodealliance.org/security)!**

### Steps to Reproduce

The default outgoing handler inserts the URI as-is into the request in absolute form which should only be used for proxies. It seems we can fix this fairly easily by stripping out the other parts from the uri in here https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasi-http/src/types.rs#L142

```diff
+    // remove the scheme and authority from the request
+    *request.uri_mut() = Uri::builder()
+        .path_and_query(
+            request
+                .uri()
+                .path_and_query()
+                .map(|p| p.as_str())
+                .unwrap_or("/"),
+        )
+        .build()
+        .expect("ok");
```

Otherwise, the absolute form uri will be used as-is, as described here https://github.com/hyperium/hyper/blob/master/src/client/conn/http1.rs#L175-L190.

### Expected Results

I expect the http handler _not_ to serialize requests using the absolute uri, since the support for this format is spotty (and potentially goes against the RFC)

### Actual Results

The absolute form of the uri ends up in the http packet.

### Versions and Environment

Version: Git master + hyper 1.1.0
Operating system: N/A
Architecture: N/A

### Extra Info

N/A

arlyon:
Closes #7737

### DIFF
--- a/crates/test-programs/src/bin/http_outbound_request_get.rs
+++ b/crates/test-programs/src/bin/http_outbound_request_get.rs
@@ -21,7 +21,7 @@ fn main() {
     let uri = res.header("x-wasmtime-test-uri").unwrap();
     assert_eq!(
         std::str::from_utf8(uri).unwrap(),
-        format!("http://{addr}/get?some=arg&goes=here")
+        format!("/get?some=arg&goes=here")
     );
     assert_eq!(res.body, b"");
 }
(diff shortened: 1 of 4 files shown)


alexcrichton:
Thanks! Would it be possible to add a test for this?

arlyon:
Got the toolchain running locally and updated with tests

alexcrichton (approved):
Thanks!

PR #7749, opened 2024-01-04
Make `shutdown` close the input/output-streams.

badeend:
Make `tcp-socket::shutdown` close the input/output-streams.
This was already specified in the WITs, but not implemented nor enforced by any tests.

### DIFF
--- a/crates/wasi/src/preview2/tcp.rs
+++ b/crates/wasi/src/preview2/tcp.rs
@@ -7,9 +7,11 @@ use crate::preview2::{
 use anyhow::{Error, Result};
 use cap_net_ext::{AddressFamily, Blocking};
 use io_lifetimes::raw::{FromRawSocketlike, IntoRawSocketlike};
+use io_lifetimes::AsSocketlike;
 use rustix::net::sockopt;
 use std::io;
 use std::mem;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 use tokio::io::Interest;
 
@@ -54,7 +56,7 @@ pub(crate) enum TcpState {
 pub struct TcpSocket {
     /// The part of a `TcpSocket` which is reference-counted so that we
     /// can pass it to async tasks.
-    pub(crate) inner: Arc<tokio::net::TcpStream>,
+    pub(crate) inner: Arc<TcpSocketInner>,
 
     /// The current state in the bind/listen/accept/connect progression.
     pub(crate) tcp_state: TcpState,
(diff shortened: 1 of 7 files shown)


crates/wasi/src/preview2/tcp.rs
```diff
@@ -208,21 +267,36 @@ impl HostOutputStream for TcpWriteStream {
...
         // `write_ready` will join the background write task if it's active, so following `flush`
         // with `write_ready` will have the desired effect.
         Ok(())
     }
 
     fn check_write(&mut self) -> Result<usize, StreamError> {
+        if self.inner.is_shut_down_for_sending() {
+            self.last_write = LastWrite::Done;
+            return Err(StreamError::Closed);
+        }
```
alexcrichton, line 298: Should this be checked after checking for `Waiting` below? It's a bit weird to have a pending write and then call shutdown, but I think the `shutdown` syscall will cause the background write to wake up and report its failure?

alexcrichton (approved):
Thanks! One question but otherwise looks good 👍

crates/wasi/src/preview2/tcp.rs
badeend: The user-facing behavior implemented in this PR seems correct to me. I.e. to return `StreamError::Closed` even if an background write was still in progress.

With the current solution it's the responsibility of the caller to properly wait for all data to be fully acknowledged/flushed _before_ calling `shutdown`.

This issue is not limited to just `shutdown`, but also `drop`.

I think the true question is: whether or not to abandon the background write (as currently current implemented) or detach the running background tasks and let the background write finish up even after the WASI input stream has been closed (on `drop` or `shutdown`).

alexcrichton: Oh I'm not advocating for letting this proceed in the background, I'm more thinking that it might be best to shuffle this to get the response from the background task before checking the shutdown status. I'm under the impression that a `shutdown` will cancel pending writes by marking the socket as writable and returning `Ok(0)` or `EPIPE` or something like that. That would mean that after the `shutdown` it's not an indefinite block for the background task to finish up and for us to get the result.

badeend: > it might be best to shuffle this to get the response from the background task before checking the shutdown status.

What would be the benefit? On line 296 I reset the last_write status which drops the `AbortOnDropJoinHandle`. So the background task will always be aborted, right?

alexcrichton: It's true yeah that largely the same thing will happen, but I'd at least personally prefer to explicitly clean up the task rather than let it get cleaned up in the background if possible. It's sort of six-to-one-half-dozen-or-the-other though.

I'm not certain of the portability of the behavior of marking the socket as writable and cancelling all writes once `shutdown` is issued though. If that's not the actual OS behavior then switching the order here wouldn't work.

Given the discussion [here](https://bytecodealliance.zulipchat.com/#narrow/stream/219900-wasi/topic/What.20happens.20to.20in-progress.20background.20writes.20on.20drop.3F/near/411819385) though I'm also a bit wary of cancelling active writes since that could show up as surprising for users, so maybe I should be advocating for letting this proceed in the background...

badeend:
Fixed in https://github.com/bytecodealliance/wasmtime/pull/9055 & https://github.com/bytecodealliance/wasmtime/pull/9225

PR #7756, opened 2024-01-05
Switch winch's internal stack from a VecDeque to a Vec

elliottt:
It doesn't look like which needs to modify the front of the `VecDeque` that it uses to model the value stack, and can instead use a `Vec`. This shouldn't change functionality at all, but seems nice for documentation purposes, as I initially assumed that there must be operations modifying the front of the stack given the use of `VecDeque`.


saulecabrera (approved):
This looks reasonable to me, thanks!

PR #7759, opened 2024-01-08
examples: add component model example

rockwotj:
As part of my journey on learning how the component model works, I realized there is no example in wasmtime, so I figured it'd be useful to add a simple example of guest and host calls.


examples/component/wasm/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "example-component-wasm"
+version = "0.0.0"
+authors = ["The Wasmtime Project Developers"]
+edition = "2021"
+publish = false
+
+[dependencies]
+wit-bindgen = { workspace = true, default-features = false, features = ["macros"] }
```
alexcrichton, line 9: For this the `default-features = false` isn't probably the best example to give since default features should be ok for 99% of embeddings. The reason it's disabled in this repository is for the unique use case of the adapter.

examples/component/wasm/guest.rs
```diff
@@ -0,0 +1,17 @@
+use wit_bindgen;
```
alexcrichton, line 1: I think this `use` can be omitted?

Also, can this file have a comment pointing to the `wit-bindgen` repository for more information about the generated bindings?

alexcrichton (commented):
Thanks for this!

I'm somewhat of two minds on this. On one hand this example showcases a lot of downsides to the unsupported nature of the component model in languages. For example embeddings may not always embed `wit-component` itself and turn core wasm into components on-the-fly. Additionally this probably won't work for many "real world" use cases in the sense that WASI isn't handled here. That would require tweaks to both the build process and the embedder. Finaly the `wasm32-unknown-unknown` target, rarely used off the web, is used here (for simplicity of course, which makes sense)

On the other hand this showcases a concrete working example which is something folks can base work on. In that sense it's great to have! Given the number of moving parts in the component model being able to start somewhere is definitely valuable.

I'd probably lean towards landing this, although perhaps with a few more comments with respect to WASI and usage. Could you add a comment indicating that `wasm32-unknown-unknown` is used for simplicity here and additionally indicate that the usage of `wit-component` in embeddings is done for the purposes of this example and otherwise may not be expected in all host embeddings of Wasmtime?

examples/component/wasm/guest.rs
rockwotj: Done.

examples/component/wasm/Cargo.toml
rockwotj: Fixed by using the default.

rockwotj:
Thanks for the review! I ended up submitting this because I wanted to:

1) Have a place where there was a single simple as possible example of the component model that was in code. Everywhere else I could find were documentation of the steps of how to do things and were not specific to wasmtime (so didn't include how to create custom worlds using wasmtime's binding macro).
2) Learn more about the generated bindings via `cargo expand` so I know what to expose via the C-API

I thought the first use case was something others wanted to do, so I ended up creating this. I whole heartedly agree with your comments and have updated the PR.

alexcrichton (approved):
Sounds good!

PR #7763, opened 2024-01-09
wasi: Test that pollables may be used multiple times

elliottt:
We didn't have a test that covered using a pollable multiple times in the test-programs, despite relying on that behavior in the preview1 adapter.


elliottt:
@pchickey and @guybedford, does this look like enough to have caused a test failure in the jco work?

guybedford:
Unfortunately I tried this in JCO and it is definitely passing, so I don't think this is capturing a case. JCO pollables will only ever resolve once as they are promise-based, after which they will be always-resolving. @elliottt let me know if I can help debug further at all.

crates/test-programs/src/bin/preview2_stream_pollable_correct.rs
```diff
@@ -1,11 +1,27 @@
...
+    // Pollables can be used many times over their lifetime
+    let stdout = stdout::get_stdout();
+    let stdout_pollable = stdout.subscribe();
+
+    let chunk = [b'a'; 50];
+    for _ in 1..10 {
+        stdout_pollable.block();
+        assert!(stdout_pollable.ready(), "after blocking, pollable is ready");
+
+        let n = stdout.check_write().unwrap() as usize;
```
guybedford, line 20: Perhaps this should write until check write doesn't allow anymore writing to ensure it's hitting backpressure?

The issue here in JCO is that we don't have backpressure for stdout since Node.js doesn't support it (https://nodejs.org/docs/latest/api/process.html#a-note-on-process-io).

elliottt: I'm working separately on a test that pipes stdout to stdin of another instantiation of the same component, and I think will give us a lot more control over how this test works. I'll ping you when it's ready :)

tests/all/piped_tests.rs
```diff
@@ -0,0 +1,72 @@
...
+pub fn run_wasmtime_piped(component_path: &str) -> Result<()> {
+    let mut producer = producer_command(component_path)?.spawn()?;
+    let stdout = producer.stdout.take().unwrap();
+    let mut consumer = consumer_command(component_path, stdout)?.spawn()?;
+
+    let producer = producer.wait()?;
+    if !producer.success() {
+        // make sure the consumer gets killed off
+        if consumer.try_wait().is_err() {
+            consumer.kill().expect("Failed to kill consumer");
```
pchickey, line 38: rather than kill the consumer, can we tell each the producer and consumer how many bytes or messages to send/expect total, so they can both exit gracefully once those conditions are met? this will also let us assert, in the consumer, that after everything expected is consumed, the pollable does not become ready "ever again" (poll with a 250ms timeout and call it good)

tests/all/piped_tests.rs
```diff
@@ -0,0 +1,72 @@
...
+            #[allow(unused_imports)]
+            use self::$name as _;
+        };
+    }
+    foreach_piped!(assert_test_exists);
+
+    // Below here is mechanical: there should be one test for every binary in
+    // wasi-tests.
+    #[test]
+    fn piped_simple() {
```
pchickey, line 64: Just curious why you chose to create a new prefix and test for these test programs and tests? or can we just make them part of the cli tests?

elliottt: These are slightly different, as they require two instances to be setup. I suppose they could be cli tests, but it seemed cleaner to separate them out, given that they have special run requirements.

tests/all/piped_tests.rs
elliottt: If the producer panics, we'll still want to kill the consumer. This was more about ensuring that we cleanup failures properly than anything else.

elliottt: Additionally, consumed amount is something that we could compute in the individual piped tests, and I think would be easier to debug as an assertion failure in the consumer side.

pchickey: got it! sounds good.

tests/all/piped_tests.rs
pchickey: works for me

pchickey (approved):

PR #7768, opened 2024-01-11
Document internals of DomTreeWithChildren

meithecatte:
I had a look at `egraph/domtree.rs` and spent a good few minutes confused by what exactly is happening. This PR adds some comments to explain how the data is represented, which should hopefully help the next person to look at this code grasp it much faster.

In particular, I clarified the comment in `egraph.rs` to avoid implying that `DomTreeWithChildren` duplicates the parent pointers.


fitzgen (approved):
Thanks!

PR #7769, opened 2024-01-11
Remove `wasmtime-jit`

adambratschikaye:
This refactoring has been discussed here: https://github.com/bytecodealliance/wasmtime/issues/7652

The `wasmtime-jit` crate is removed by moving the platform agnostic parts to `wasmtime-environ` and the remaining parts to `wasmtime`.

The goal of this is to simplify the crate organization and make it possible to add a `runtime` feature to the `wasmtime` crate such that `wasmtime` can be build without the `runtime` feature to enable Wasm compilation on targets that may not support actual Wasm execution (e.g. `wasm32` itself). A follow-up PR will add this feature.


adambratschikaye:
There are two main commits with the first one moving parts into `environ` and the second moving the remaining parts to `wasmtime`. If you'd prefer I could submit those as two separate PRs.

alexcrichton (approved):
Looks great to me, thanks again for working on this!

alexcrichton:
Ah so this sort of refatoring is likely to break various bits and pieces of CI, so if you'd like you can push a commit with the phrase "prtest:full" somewhere in the commit message and that'll force full CI to run on this PR instead of just the few builders. That way you should be able to weed out all the build failures before it goes through the merge queue.

adambratschikaye:
Looks like the full pipeline is passing now.

alexcrichton (approved):

PR #7770, opened 2024-01-11
Fix the ABI of Option<u128> on x86_64

bjorn3:
LLVM passes the tag in rax and the u128 val in rdx/rcx. Before this change Cranelift would pass the first half of the value in rdx and the second half using an implicit return value pointer.

Fixes https://github.com/rust-lang/rustc_codegen_cranelift/issues/1449


abrown (approved):
Makes sense to me.

abrown:
@bjorn3, I tried to look this up in the LLVM documentation but all I can find is a mention of RCX in the swiftcc calling convention [here](https://llvm.org/docs/LangRef.html#calling-conventions). Where do you see it?

bjorn3:
See eg https://rust.godbolt.org/z/fTM6o84xq Fair chance it isn't documented anywhere, this requires the enable_llvm_abi_extensions flag in Cranelift for a reason.

PR #7771, opened 2024-01-11
[17.0.0] Backport fix x64 option i128 abi

bjorn3:
LLVM passes the tag in rax and the u128 val in rdx/rcx. Before this change Cranelift would pass the first half of the value in rdx and the second half using an implicit return value pointer.

Backport of https://github.com/bytecodealliance/wasmtime/pull/7770 to 17.0.0


bjorn3:
This only affects programs that use enable_llvm_abi_extensions, which is not Wasmtime. As such I think this ok to backport.

alexcrichton (approved):

Issue #7552, opened 2023-11-16
Improve Winch's file tests disassembly


saulecabrera:
I wonder if it would be worth omitting the offsets (except for jump targets) so that the diff would have shown only the `add rsp, 4` instruction going away?

_Originally posted by @elliottt in https://github.com/bytecodealliance/wasmtime/pull/7547#discussion_r1395016578_

PR #7774, opened 2024-01-12
winch: Check for stack overflow

elliottt:
Add stack overflow checking to the x64 backend in winch, and enable the runtime test for overflow. The method for checking for overflow is to compare `sp` against the stack limit from `vmctx` after the stack pointer has been adjusted to hold the locals for the function.

### DIFF
--- a/tests/misc_testsuite/winch/call.wast
+++ b/tests/misc_testsuite/winch/call.wast
@@ -221,6 +221,10 @@
 (assert_return (invoke "odd" (i64.const 1)) (i32.const 44))
 (assert_return (invoke "odd" (i64.const 200)) (i32.const 99))
 (assert_return (invoke "odd" (i64.const 77)) (i32.const 44))
+
+(assert_exhaustion (invoke "runaway") "call stack exhausted")
+(assert_exhaustion (invoke "mutual-runaway") "call stack exhausted")
+
 (assert_return (invoke "as-if-condition") (i32.const 1))
 (assert_return (invoke "as-br_if-first") (i32.const 0x132))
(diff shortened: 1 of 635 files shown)


tests/misc_testsuite/winch/call.wast
saulecabrera, line 1: Can we also enable the stack exhaustion tests for `call_indirect`?

saulecabrera (approved):
LGTM, thanks! Just a comment regarding tests, and after that this should be ready to land. 

Regarding filetests and disassembly, -- as mentioned offline -- I think it's a very reasonable thing to do. I just couldn't come up with a straightforward and sound way to do it. I'd like to keep labels around for filetests that require jumps and such, if possible. Perhaps we can continue the discussion in https://github.com/bytecodealliance/wasmtime/issues/7552?

saulecabrera:
I _think_ this is due to incomplete unwind information encoding in Windows. All test that involve traps, currently fail on Windows for this reason. I think it's fine to ignore the tests introduced in this PR, temporarily in https://github.com/bytecodealliance/wasmtime/blob/main/build.rs#L210

elliottt:
> I _think_ this is due to incomplete unwind information encoding in Windows. All test that involve traps, currently fail on Windows for this reason. I think it's fine to ignore the tests introduced in this PR, temporarily in https://github.com/bytecodealliance/wasmtime/blob/main/build.rs#L210

Thank you for tracking this down! I was really dreading having to debug a windows failure :sweat_smile:

elliottt:
Also in an argument for merging #7782 first, removing instruction offsets cuts the diff of this PR in half :+1:

PR #7778, opened 2024-01-15
add linker to wasi preview2 that accepts custom closure

sehz:
Allow passing of closure to wasi command linking.  Current `add_to_linker` is re-implemented more general API.  Without this, it's hard to create reusable wasi context store

### DIFF
--- a/crates/wasi/src/preview2/command.rs
+++ b/crates/wasi/src/preview2/command.rs
@@ -90,33 +90,41 @@ pub mod sync {
     pub fn add_to_linker<T: WasiView>(
         l: &mut wasmtime::component::Linker<T>,
     ) -> anyhow::Result<()> {
-        crate::preview2::bindings::clocks::wall_clock::add_to_linker(l, |t| t)?;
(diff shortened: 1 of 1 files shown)


sehz:
Problem is that `T` is becoming very complex.   Ideally Store should be composable.    Key problem is that store is tied to linker which makes difficult separate host implementation

alexcrichton:
I understand that `T` can be complex, and I also understand that writing a closure is shorter than writing a trait impl, but I'm not sure why the complexity would make trait impls more difficult (as opposed to a simple `T`). The fact that the store and linker are tied together is indeed a bit limiting, but with trait bounds it should be possible to design roughly the sme APIs.

Is there anything preventing the trait impls on `T` in your case?

alexcrichton:
We chattec a bit more tonight about this and I think it's a reasonable enough feature to add given it's pretty small and solves a concrete use case. If you're ok fixing the failing test on CI (I think there's a new trait bound needed) I can r+

sehz:
Thanks.  Will fix CI

alexcrichton:
This PR is quite dated now at this point, so I'm going to close this, but happy to review if you'd like to resubmit.

PR #7781, opened 2024-01-16
wasi implementations: use rc-2024-01-16 for sockets, cli, http

pchickey:
https://github.com/WebAssembly/wasi-sockets/pull/93 eliminates the ipv6-only sockopt getter and setter, and the ability for cross-stack ipv6 to ipv4 connections to take place, so these are bubbled out through the test suites.

### DIFF
--- a/crates/test-programs/src/bin/preview2_tcp_connect.rs
+++ b/crates/test-programs/src/bin/preview2_tcp_connect.rs
@@ -76,14 +76,16 @@ fn test_tcp_connect_dual_stack(net: &Network) {
     v4_listener.blocking_listen().unwrap();
 
     let v4_listener_addr = v4_listener.local_address().unwrap();
+    let v6_listener_addr =
+        IpSocketAddress::new(IpAddress::IPV4_MAPPED_LOOPBACK, v4_listener_addr.port());
 
     let v6_client = TcpSocket::new(IpAddressFamily::Ipv6).unwrap();
 
     // Tests:
 
-    // Connecting to an IPv4 address on an IPv6 socket should fail:
+    // Connecting to an IPv4-mapped-IPv6 address on an IPv6 socket should fail:
     assert!(matches!(
-        v6_client.blocking_connect(net, v4_listener_addr),
+        v6_client.blocking_connect(net, v6_listener_addr),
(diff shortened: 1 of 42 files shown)


crates/test-programs/src/bin/preview2_tcp_connect.rs
```diff
@@ -76,30 +76,16 @@ fn test_tcp_connect_dual_stack(net: &Network) {
...
 
-    // Even on platforms that don't support dualstack sockets,
-    // setting ipv6_only to true (disabling dualstack mode) should work.
-    v6_client.set_ipv6_only(true).unwrap();
-
-    // Connecting to an IPv4-mapped-IPv6 address on an ipv6-only socket should fail:
+    // Connecting to an IPv4 address on an IPv6 socket should fail:
     assert!(matches!(
-        v6_client.blocking_connect(net, v6_listener_addr),
+        v6_client.blocking_connect(net, v4_listener_addr),
```
badeend, line 86: This should attempt to connect to `v6_listener_addr` which was a IPv4-mapped-IPv6 address. The way you rewrote it, this call doesn't actually test the change.

crates/test-programs/src/bin/preview2_udp_connect.rs
```diff
@@ -84,42 +84,17 @@ fn test_udp_connect_dual_stack(net: &Network) {
...
-            IpAddressFamily::Ipv6
-        );
-    }
+    let v6_client = UdpSocket::new(IpAddressFamily::Ipv6).unwrap();
+
+    v6_client.blocking_bind_unspecified(&net).unwrap();
+
+    // Connecting to an IPv4 address on an IPv6 socket should fail:
+    assert!(matches!(
+        v6_client.stream(Some(v4_server_addr)),
```
badeend, line 95: See previous comment

badeend (changes requested):
I have one remark. Other than that: LGMT!

sunfishcode (approved):

crates/test-programs/src/bin/preview2_tcp_connect.rs
pchickey: Got it - thanks!

PR #7782, opened 2024-01-16
winch: Reduce instruction offset printing in winch

### LINKED ISSUE #7552: Improve Winch's file tests disassembly

elliottt:
In order to make updates to winch filetess a bit less noisy, restrict isntruction offset printing to two cases:

1. The beginning of a basic block (omitting the first block)
2. All instructions after a `ret`

The reason behind `2` is that often instructions following a `ret` will be part of the trap table, and will thus be a common jump target from other areas of the function.

This approach isn't perfect, as jumping over single instructions won't end up forcing the destination to have a label. However, it does get most cases right, and makes changes to the disassembly much smaller.

Fixes #7552


saulecabrera (approved):
Thanks for working on this. I agree that the diff is now less noisy and that in most cases it's easy to verify by looking what's going on in the disassembly.

winch/filetests/src/disasm.rs
saulecabrera, line 1: Do you think it'll be worth conditionally ignoring the offsets only when the disassembly is invoked from the filetests? This disassembly utility is also used by [`winch-tools compile <file.wat>`](https://github.com/bytecodealliance/wasmtime/blob/main/winch/src/main.rs#L11), in which case it could be useful to print the offsets for a more in-depth analysis.

elliottt: That's a great idea, I'll plumb through a flag :+1:

PR #7783, opened 2024-01-17
fix: allow dynamic owned resources to be used as borrowed parameters

rvolosatovs:
This enables using owned resources in places where borrowed resources of the same type are expected (e.g. dynamically-typed functions/methods defined in `Linker::func_new`)
Note, that this is already possible in e.g. https://github.com/bytecodealliance/wasmtime/blob/4b2425dc3044ec57ddbfa94bcbe2d774df249d03/crates/wasmtime/src/component/resources.rs#L340-L352

cc @alexcrichton , we discussed this on a call, I believe this is the changeset we agreed upon

### DIFF
--- a/crates/wasmtime/src/component/func.rs
+++ b/crates/wasmtime/src/component/func.rs
@@ -362,7 +362,8 @@ impl Func {
         }
 
         for (param, ty) in params.iter().zip(param_tys.iter()) {
-            ty.check(param).context("type mismatch with parameters")?;
+            ty.is_supertype_of(param)
+                .context("type mismatch with parameters")?;
         }
 
         self.call_raw(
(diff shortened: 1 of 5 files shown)


alexcrichton (approved):
Looks great to me, thanks! Mind adding a test as well for this?

rvolosatovs:
Added a test case, but also uncovered a bug while doing that https://github.com/bytecodealliance/wasmtime/issues/7793

Honestly, I'm not sure what's the use case for using the resource borrows pointing to non-existent resources and it appears to me that neither of `Resources::new_borrow` cases should actually work (without panic though), but perhaps I don't understand the semantics here fully.

alexcrichton (approved):
Thanks for the test! I believe that issue is orthogonal to this, so I'm going to mark this for merge while that other issue is sorted out.

PR #7788, opened 2024-01-18
mpk: add `--memory-protection-keys` to the CLI options

abrown:
This only adds MPK configuration to the new Wasmtime flags.

### DIFF
--- a/crates/cli-flags/src/lib.rs
+++ b/crates/cli-flags/src/lib.rs
@@ -3,7 +3,7 @@
 use anyhow::Result;
 use clap::Parser;
 use std::time::Duration;
-use wasmtime::Config;
+use wasmtime::{Config, MpkEnabled};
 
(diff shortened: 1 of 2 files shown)


abrown:
@alexcrichton, I don't expect the current CI run to pass: when I run this locally, I think the following command should fail with `"memory protection keys require the pooling allocator"` but it does not:

```console
$ WASMTIME_NEW_CLI=1 cargo run -- run -O memory-protection-keys=y --invoke echo_f32 tests/all/cli_tests/simple.wat 1.1
```

alexcrichton (commented):
Oh I think that's because the block of code here is only hit when `-O pooling-allocator` is passed, otherwise it's skipped entirely. The `bail!` may need to move outside the `match_feature!` block

alexcrichton (approved):

PR #7789, opened 2024-01-18
mpk: restore PKRU state when a fiber resumes execution

abrown:
Previously, when a fiber was suspended, other computation could change the PKRU state on the current CPU. This means that the fiber could be resumed with a different PKRU state. This could be bad, resulting in situations in which the fiber can access more memory slots than it should or cannot even access its own memory slots.

This change saves the PKRU state prior to a fiber being suspended. When the fiber resumes execution, that PKRU state is restored.

### DIFF
--- a/crates/runtime/src/mpk/disabled.rs
+++ b/crates/runtime/src/mpk/disabled.rs
@@ -13,6 +13,10 @@ pub fn keys(_: usize) -> &'static [ProtectionKey] {
 }
 pub fn allow(_: ProtectionMask) {}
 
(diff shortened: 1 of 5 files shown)


abrown:
This probably needs an `async_functions.rs` test but I'm still struggling on how to craft a scenario that checks for this issue.

alexcrichton (commented):
Looks good to me 👍 

With a test that fails without this change I think it's good-to-go.

alexcrichton (approved):

PR #7791, opened 2024-01-18
Update the Wasmi fuzzing oracle to version `0.31.0`

Robbepop:
Currently Wasmtime uses the 2 years old Wasmi version `0.20.0`.

Since then Wasmi has improved substantially and added support for new Wasm proposals such as `bulk-memory`, `reference-types` and `tail-calls` which we can now enable.
Wasmi `v0.31.0` has recently been audited and is used by some large projects, thus is a lot more battle tested than the previous `v0.20.0`.

Besides that the most notable change are performance improvements which should make fuzzing with Wasmi a tiny bit faster.

Look into the future: Since roughly half a year I am working on the next major Wasmi version `v0.32.0` which is a complete rewrite of the Wasmi executor featuring a much more powerful register-machine execution model and optional lazy compilation & validation. I hope that it becomes stable enough for use soon to provide it as fuzzing oracle to Wasmtime. Due to the changes we refer to the new Wasmi version as Wasmi (register) and the old Wasmi as Wasmi (stack). It might even make sense to have both versions as oracles at the same time because they have very different strengths.

### DIFF
--- a/crates/fuzzing/src/oracles/diff_wasmi.rs
+++ b/crates/fuzzing/src/oracles/diff_wasmi.rs
@@ -20,8 +20,6 @@ impl WasmiEngine {
         config.exceptions_enabled = false;
         config.max_memories = config.max_memories.min(1);
         config.min_memories = config.min_memories.min(1);
-        config.max_tables = config.max_tables.min(1);
-        config.min_tables = config.min_tables.min(1);
 
         let mut wasmi_config = wasmi::Config::default();
         wasmi_config
--- a/crates/fuzzing/src/oracles/diff_wasmi.rs
+++ b/crates/fuzzing/src/oracles/diff_wasmi.rs
@@ -13,12 +13,18 @@ pub struct WasmiEngine {
 impl WasmiEngine {
     pub(crate) fn new(config: &mut Config) -> Self {
         let config = &mut config.module_config.config;
+        config.multi_value_enabled = true;
+        config.sign_extension_ops_enabled = true;
+        config.saturating_float_to_int_enabled = true;
         config.reference_types_enabled = true;
         config.simd_enabled = false;
+        config.relaxed_simd_enabled = false;
         config.memory64_enabled = false;
+        config.threads_enabled = false;
         config.bulk_memory_enabled = true;
         config.threads_enabled = false;
         config.tail_call_enabled = true;
+        config.exceptions_enabled = false;
         config.max_memories = config.max_memories.min(1);
         config.min_memories = config.min_memories.min(1);
         config.max_tables = config.max_tables.min(1);
--- a/crates/fuzzing/src/oracles/diff_wasmi.rs
+++ b/crates/fuzzing/src/oracles/diff_wasmi.rs
@@ -13,6 +13,7 @@ pub struct WasmiEngine {
 impl WasmiEngine {
     pub(crate) fn new(config: &mut Config) -> Self {
         let config = &mut config.module_config.config;
+        // Force generated Wasm modules to never have features that Wasmi doesn't support.
         config.simd_enabled = false;
         config.relaxed_simd_enabled = false;
         config.memory64_enabled = false;
--- a/crates/fuzzing/src/oracles/diff_wasmi.rs
+++ b/crates/fuzzing/src/oracles/diff_wasmi.rs
@@ -13,25 +13,31 @@ pub struct WasmiEngine {
 impl WasmiEngine {
     pub(crate) fn new(config: &mut Config) -> Self {
         let config = &mut config.module_config.config;
-        config.multi_value_enabled = true;
-        config.sign_extension_ops_enabled = true;
-        config.saturating_float_to_int_enabled = true;
-        config.reference_types_enabled = true;
         config.simd_enabled = false;
         config.relaxed_simd_enabled = false;
         config.memory64_enabled = false;
         config.threads_enabled = false;
-        config.bulk_memory_enabled = true;
         config.threads_enabled = false;
-        config.tail_call_enabled = true;
         config.exceptions_enabled = false;
         config.max_memories = config.max_memories.min(1);
         config.min_memories = config.min_memories.min(1);
         config.max_tables = config.max_tables.min(1);
         config.min_tables = config.min_tables.min(1);
 
+        let mut wasmi_config = wasmi::Config::default();
+        wasmi_config
+            .consume_fuel(false)
+            .floats(true)
+            .wasm_mutable_global(true)
+            .wasm_sign_extension(config.sign_extension_ops_enabled)
+            .wasm_saturating_float_to_int(config.saturating_float_to_int_enabled)
+            .wasm_multi_value(config.multi_value_enabled)
+            .wasm_bulk_memory(config.bulk_memory_enabled)
+            .wasm_reference_types(config.reference_types_enabled)
+            .wasm_tail_call(config.tail_call_enabled)
+            .wasm_extended_const(true);
         Self {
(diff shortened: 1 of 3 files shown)


Robbepop:
With respect to `cargo vet`:

```
  wasmi:0.31.1 missing ["safe-to-run"]
  wasmi_arena:0.4.0 missing ["safe-to-run"]
  wasmi_core:0.13.0 missing ["safe-to-run"]
  wasmparser-nostd:0.100.1 missing ["safe-to-run"]
```

All those crates are maintained by Wasmi maintainers (me).

crates/fuzzing/src/oracles/diff_wasmi.rs
```diff
@@ -13,18 +13,31 @@ pub struct WasmiEngine {
 impl WasmiEngine {
     pub(crate) fn new(config: &mut Config) -> Self {
         let config = &mut config.module_config.config;
-        config.reference_types_enabled = false;
         config.simd_enabled = false;
+        config.relaxed_simd_enabled = false;
         config.memory64_enabled = false;
-        config.bulk_memory_enabled = false;
         config.threads_enabled = false;
+        config.threads_enabled = false;
```
alexcrichton, line 20: This all looks right to me, thanks! The intention here is that the differential engine (wasmi here) both configures itself based on `Config` but additionally updates `Config` with anything it can't support. So for example simd is disabled here unconditionally in `Config` because wasmi doesn't support it, but you're doing the right thing otherwise where features are conditionally enabled in wasmi depending on `Config` for those that wasmi supports.

Two minor comments here:

* This is setting `threads_enabled = false` twice
* The `min_tables` and `max_tables` handling can probably be removed if wasmi supports reference types which is where multiple tables were added.

If you'd like, it might be good to leave some comments here too along the lines of:

```rust
// force generated modules to never have features that wasmi doesn't support
```

or something like that

Robbepop: I have implemented your code review suggestions.

Robbepop:
> Thanks!
> 
> To avoid a new vet of `wasmparser-nostd` would it be possible to conditionally build wasmi with the official `wasmparser` crate instead? That way we could get that auto-vetted.

Wasmi `v0.31.0` is no longer in development so I am not going to support the standard `wasmparser` for it, however, I am going to think about the implications doing so for future Wasmi versions starting from `v0.32.0` once released. Since Wasmi already has a `std` crate feature it might be simple to just use `wasmparser` if it is enabled and use `wasmparser-nostd` otherwise.

alexcrichton:
Ah ok, in that case this is going to have to hold off until one of us gets a chance to vet the dependencies here. I'm stretched a bit thin at the moment so it may be a bit before I can personally get to this (but others can of course beat me to it)

Robbepop:
Please tell me if and how I can help with vetting. :)

alexcrichton:
Reading over some code I think that [this impl](https://github.com/paritytech/wasmi/blob/27def282b06613e770d0ab96de88b9909973a12b/crates/arena/src/component_vec.rs#L18-L19) is not sound as it's supposed to be `T: Sync`. Additionally though I don't think you should need the `unsafe impl` at all. You can try changing the `PhantomData` to either `fn() -> Idx` or `fn(Idx)` I think as one of them might require the impl and one might not.

Otherwise the `wasmparser-nostd` diff is quite large or larger than I remember from the `wasmparser` diff, so diffing those directories is requiring a good deal of time to go through to verify it's the same. Additionally I haven't even gotten to `wasmi` yet which glancing at it is quite large and additionally contains a good deal of `unsafe` code to check.

Currently though wasmi is only used for fuzzing so it doesn't necessarily need a thorough review or strict vetting. Do others have thoughts on whether we should add exemptions for this dependency as opposed to vetting it?

Robbepop:
> Reading over some code I think that [this impl](https://github.com/paritytech/wasmi/blob/27def282b06613e770d0ab96de88b9909973a12b/crates/arena/src/component_vec.rs#L18-L19) is not sound as it's supposed to be `T: Sync`. Additionally though I don't think you should need the `unsafe impl` at all. You can try changing the `PhantomData` to either `fn() -> Idx` or `fn(Idx)` I think as one of them might require the impl and one might not.

Thanks for catching this bug! The code is quite old and hasn't seen a lot of love lately.

> Otherwise the `wasmparser-nostd` diff is quite large or larger than I remember from the `wasmparser` diff, so diffing those directories is requiring a good deal of time to go through to verify it's the same. Additionally I haven't even gotten to `wasmi` yet which glancing at it is quite large and additionally contains a good deal of `unsafe` code to check.

The unfortunate truth is that especially with the component model a lot of very `no_std` unfriendly abstractions such as the `IndexMap` has been added which required me to come up with my own `no_std` friendly version of it. I never intended this to be a long lasting effort but here we are.

> Currently though wasmi is only used for fuzzing so it doesn't necessarily need a thorough review or strict vetting. Do others have thoughts on whether we should add exemptions for this dependency as opposed to vetting it?

To provide you with a bit of transparency here about the state of Wasmi:

- Cons:

    - I work on the project alone.
    - There are pretty much no code reviews due to above point.
    - There is a fair amount of `unsafe` code use, but mostly in the Wasmi bytecode executor.

- Pros:

    - There are a good deal of unit, integration, e2e and fuzz tests.
    - We do have a very extensive CI (incl. `miri`)
    - Wasmi `v0.31.0` is already used in production by different companies some many months.
    - Wasmi `v0.31.0` has received a security audit by SRLabs.
    - Wasmi `0.31.0` is much less complex than the Wasmi `v0.32.0-beta.n` (on `master`).

Robbepop:
I just made this use Wasmi `v0.31.1` instead of `v0.31.0` which is more explicit. Note that `v0.31.0` got yanked some time ago.

pchickey:
>  Do others have thoughts on whether we should add exemptions for this dependency as opposed to vetting it?

I'm happy with adding a cargo vet exemption for `wasmi 0.31.1` which notes that its only intended to be used in the context of testing. I don't think there is any real benefit to our project to have a higher bar than wasmi already meets for the sake of a fuzzing oracle.

fitzgen:
> I'm happy with adding a cargo vet exemption for `wasmi 0.31.1` which notes that its only intended to be used in the context of testing. I don't think there is any real benefit to our project to have a higher bar than wasmi already meets for the sake of a fuzzing oracle.

Agreed.

Robbepop:
> Reading over some code I think that [this impl](https://github.com/paritytech/wasmi/blob/27def282b06613e770d0ab96de88b9909973a12b/crates/arena/src/component_vec.rs#L18-L19) is not sound as it's supposed to be `T: Sync`. Additionally though I don't think you should need the `unsafe impl` at all. You can try changing the `PhantomData` to either `fn() -> Idx` or `fn(Idx)` I think as one of them might require the impl and one might not.

btw. I just published a new `wasmi_arena v0.4.1` and yanked `v0.4.0` with the unsoundness fix. (https://crates.io/crates/wasmi_arena/0.4.1)

alexcrichton:
Ok I've posted https://github.com/bytecodealliance/wasmtime/pull/7810 with new vet metadata. If you rebase this PR on top of that once it lands should be good to go

Robbepop:
Something went terribly wrong with the rebase ... let me investigate ...

alexcrichton (approved):
Thanks for the patience while we figure out the vetting, and thanks again for helping us out with the update!

Robbepop:
> Thanks for the patience while we figure out the vetting, and thanks again for helping us out with the update!

Happy to help and thanks for having Wasmi as fuzzing oracle! :)

Issue #7776, opened 2024-01-15
wasi-common / wiggle support for disjoint host/guest memory


juntyr:
I am running WASI code in Wasmtime and the browser and am trying to reuse my WasiCtx implementation for both. While the Wasmtime host can access the guest's memory, which wiggle exposes by requiring a [`GuestMemory::base`](https://docs.rs/wiggle/latest/wiggle/trait.GuestMemory.html#tymethod.base) method implementation. In my browser use case, however, the "host" runs inside WASM itself and launches new guests as separate WASM instances. The backing memory for each instance, an ArrayBuffer, is disjoint, so any memory transfer has to go through copying read and write calls.

#### Feature

Would it be (or is it already) possible to generalise the [`GuestMemory`](https://docs.rs/wiggle/latest/wiggle/trait.GuestMemory.html) trait to allow for guest memory implementations which are disjoint? If the [`GuestMemory::base`](https://docs.rs/wiggle/latest/wiggle/trait.GuestMemory.html#tymethod.base) method is required by current Wasmtime(-adjacen) code, it could be moved into an extension trait implemented by all current implementations.

#### Benefit

This more general approach would allow further code reuse even on platforms where the WASM host cannot directly access a guest's memory.

Thank you for your help!

juntyr:
It seems like `GuestSlice` also assumes that the host has direct access to the guest as it implements `Deref` (and `GuestSliceMut` implements `DerefMut`)

juntyr:
Thank you for your response! I tried to hack together a workaround using an intermediary copy buffer that translates device memory into short-lived (per each syscall) host copies that are reflected back to the guest (the same idea that the original WASI polyfill was based on). While https://github.com/rust-lang/rust/pull/119616 is being worked on, this should provide a useful workaround.

For the future, are there plans to provide some common WASI implementation primitives that can also be used with other runtimes? For instance, most host implementations for the preview-2 components could reside in a new wasi-common crate and be re-exported in wasmtime-wasi (which would also add Wasmtime-specific plumbing to link in these implementations). This would be extremely helpful in use cases such as mine where a different runtime is needed for different targets but I want to (1) reuse most components (2) control a few components (e.g. clock and RNG for full reproducibility) and (3) share as much of this implementation as possible.

Thank you again for your time and help and keep up your fantastic work!

alexcrichton:
[...]

PR #7792, opened 2024-01-18
Make wasmtime dependency in wiggle and wasi-common optional with a default feature

juntyr:
I am trying to use `wasi-common` (and thus also `wiggle`) to provide a common `WasiCtx` implementation that is reused between `wasmtime` and a custom runtime (as briefly mentioned in https://github.com/bytecodealliance/wasmtime/issues/7776). At the moment, compiling these packages also compiles `wasmtime`, which is not supported on my platform. I propose to add an enabled-by-default feature to `wasi-common` and `wiggle` that guards all `wasmtime`-specific code and allows both packages to be compiled without a dependency on `wasmtime` using `default-features = false`.


alexcrichton (approved):
Thanks!

PR #7797, opened 2024-01-19
wasmtime: add EngineWeak which has ::upgrade, created by Engine::weak

pchickey:
Engine is, internally, just an Arc<EngineInner>, so this is trivial to implement - EngineWeak is a Weak<EngineInner>.

This behavior is desirable because `Engine::increment_epoch` typically happens in a worker thread, which in turn requires additional machinery to discard the `Engine` once it is no longer needed. If instead the worker thread holds an `EngineWeak`, it can stop ticking when all consumers of the `Engine` have dropped it. This has been documented as a suggestion in `increment_epoch`.

For an example of additional machinery which is simplified by this change, see https://github.com/fastly/Viceroy/blob/25edee0700ec0b20b1b56db0a3a8d6f090397b3a/lib/src/execute.rs#L108-L116)

### DIFF
--- a/crates/wasmtime/src/engine.rs
+++ b/crates/wasmtime/src/engine.rs
@@ -813,4 +813,17 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn engine_weak_upgrades() {
+        let engine = Engine::default();
+        let weak = engine.weak();
+        weak.upgrade()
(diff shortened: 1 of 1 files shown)


fitzgen (approved):
Sanity test for when it is expected to be upgradable and expected to not be?

PR #7802, opened 2024-01-20
WASI: remove unnecessary Sync constraints

badeend:
Fundamentally, preview2 is fully synchronous. So at least in theory there should be no need for all the Sync constraints. In practice it took me a while to figure out how to actually make it work. In summary:
- Removed many `+ Sync` constraints
- Removed the immutable reference accessors on WasiView. I also dropped the _mut suffix for consistency with WasiHttpView.
- Had to do a bit of refactoring in `crates/wasi/src/preview2/host/filesystem.rs` [here](https://github.com/bytecodealliance/wasmtime/commit/6b4d634c8ea0de212229c0e9528e06598dacbd22#diff-dff18465028496004c5e0e50b9d35bc5303ea19e2c50d6f1284dd45c6fe7a2c4) and `crates/wasi/src/preview2/preview1.rs` [here](https://github.com/bytecodealliance/wasmtime/commit/1e766c240df9081fcd2b617e1bc76a63dc719363) to make them play along.

---

Chat: https://bytecodealliance.zulipchat.com/#narrow/stream/217126-wasmtime/topic/Wasmtime.20preview2.20.60Sync.60

### DIFF
--- a/src/commands/run.rs
+++ b/src/commands/run.rs
@@ -794,9 +794,9 @@ impl RunCommand {
 struct Host {
     preview1_ctx: Option<wasmtime_wasi::WasiCtx>,
 
-    // The Mutex is only needed to satisfy the Sync constraint, but we never
-    // actually perform any locking on it as use Mutex::get_mut for
-    // all accesses.
+    // The Mutex is only needed to satisfy the Sync constraint but we never
+    // actually perform any locking on it as we use Mutex::get_mut for every
+    // access.
     preview2_ctx: Option<Arc<Mutex<preview2::WasiCtx>>>,
 
     // Resource table for preview2 if the `preview2_ctx` is in use, otherwise
--- a/src/commands/run.rs
+++ b/src/commands/run.rs
@@ -11,7 +11,7 @@ use anyhow::{anyhow, bail, Context as _, Error, Result};
 use clap::Parser;
 use std::ffi::OsString;
 use std::path::{Path, PathBuf};
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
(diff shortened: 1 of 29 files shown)


src/commands/run.rs
```diff
@@ -784,19 +784,24 @@ impl RunCommand {
...
+
+    // The Mutex is only needed to satisfy the Sync constraint but we never
+    // actually perform any locking on it as we use Mutex::get_mut for every
+    // access.
+    preview2_ctx: Option<Arc<Mutex<preview2::WasiCtx>>>,
 
     // Resource table for preview2 if the `preview2_ctx` is in use, otherwise
     // "just" an empty table.
-    preview2_table: Arc<wasmtime::component::ResourceTable>,
+    preview2_table: Arc<Mutex<wasmtime::component::ResourceTable>>,
```
pchickey, line 804: Needing a mutex here is a minor bummer, but I think the next refactor is to make there be two different ctx structs:`Host` for all preview 2 compatible things, and `LegacyHost` for using wasi-common if, and only if, threads are required.

I am working on a PR that separates out wasi-common from wasmtime-wasi, so that we can ship just the preview 2 implementation without any of the old one, so I'll fix this up after that lands.

pchickey (approved):
Thanks! These are all nice improvements.

src/commands/run.rs
badeend: Yeah, but at least `.lock()` never gets called, so effectively it's mostly a Box

badeend:
There were some build errors specific to Rust version 1.73 (and lower probably).

This should be good to go again.

PR #7803, opened 2024-01-22
examples: Don't `fread()` within `assert(...)`.

waywardmonkeys:
This is fine in debug builds but release builds without assertions results in the file data not being read.


alexcrichton (approved):
Thanks!

PR #7804, opened 2024-01-22
feat: support component type introspection

### LINKED ISSUE #7726: Allow `component::LinkerInstance::func_new` functions to return complex types
#### Feature

At present, only simple types can be returned in `component::LinkerInstance::func_new`

The following variants of `component::Val` cannot be returned by dynamic functions defined via `component::LinkerInstance::func_new`:

- List([List](https://docs.rs/wasmtime/latest/wasmtime/component/struct.List.html))
- Record([Record](https://docs.rs/wasmtime/latest/wasmtime/component/struct.Record.html))
- Tuple([Tuple](https://docs.rs/wasmtime/latest/wasmtime/component/struct.Tuple.html))
- Variant([Variant](https://docs.rs/wasmtime/latest/wasmtime/component/struct.Variant.html))
- Enum([Enum](https://docs.rs/wasmtime/latest/wasmtime/component/struct.Enum.html))
- Option([OptionVal](https://docs.rs/wasmtime/latest/wasmtime/component/struct.OptionVal.html))
- Result([ResultVal](https://docs.rs/wasmtime/latest/wasmtime/component/struct.ResultVal.html))
- Flags([Flags](https://docs.rs/wasmtime/latest/wasmtime/component/struct.Flags.html))

The reason for this is the fact that constructors of these values require a type parameter, which cannot be constructed outside the crate. While such type parameter can be extracted from a function signature exported by a component, there is currently no means to access the type for functions *imported* by the component. At best, implementations may parse the type information using `wit_parser` crate, but there is currently no way to map from `wit_parser` type to `wasmtime` type.

For example:
[`Enum::new`](https://docs.rs/wasmtime/latest/wasmtime/component/struct.Enum.html#method.new) requires [`types::Enum`](https://docs.rs/wasmtime/latest/wasmtime/component/types/struct.Enum.html) value as a parameter, which does not export a constructor or any conversion mechanism. This type can only be returned by wasmtime crate, but functionality to extract this from a component import does not currently appear to exist.

#### Benefit

This would allow dynamically-typed function definitions in the linker returning more that the basic numeric/string types (and resources once https://github.com/bytecodealliance/wasmtime/pull/7688 lands)

#### Implementation

Perhaps `func_new` function parameter signatures could be updated to take in a `Vec<Type>` (https://docs.rs/wasmtime/latest/wasmtime/component/types/enum.Type.html) or, perhaps even better signatures could get `Vec<(Val, Type)>` as a parameter for the result parameter. Note, that `params` are free from this issue, because the type info is already embedded in `Val` passed to the function, which is not the case for results (and there is probably no way to change that)

#### Alternatives

The only alternative seems not supporting dynamically-typed function imports, or only supporting a subset of those, as currently done.

rvolosatovs:
This allows for embedders to introspect the component type to e.g. construct `Val`s of complex types expected by the components at runtime.
Closes #7726 
I'll work on adding a simple test once I verify this covers my own use case

cc @alexcrichton

### DIFF
--- a/crates/environ/src/component/translate.rs
+++ b/crates/environ/src/component/translate.rs
@@ -200,7 +200,7 @@ enum LocalInitializer<'data> {
         HashMap<&'data str, ComponentItem>,
         ComponentInstanceTypeId,
     ),
-    ComponentSynthetic(HashMap<&'data str, ComponentItem>),
+    ComponentSynthetic(HashMap<&'data str, ComponentItem>, ComponentInstanceTypeId),
 
     // alias section
     AliasExportFunc(ModuleInstanceIndex, &'data str),
@@ -581,18 +581,18 @@ impl<'a, 'data> Translator<'a, 'data> {
                 let mut index = self.validator.types(0).unwrap().component_instance_count();
                 self.validator.component_instance_section(&s)?;
                 for instance in s {
+                    let types = self.validator.types(0).unwrap();
+                    let ty = types.component_instance_at(index);
                     let init = match instance? {
                         wasmparser::ComponentInstance::Instantiate {
                             component_index,
                             args,
                         } => {
-                            let types = self.validator.types(0).unwrap();
-                            let ty = types.component_instance_at(index);
                             let index = ComponentIndex::from_u32(component_index);
                             self.instantiate_component(index, &args, ty)?
                         }
                         wasmparser::ComponentInstance::FromExports(exports) => {
-                            self.instantiate_component_from_exports(&exports)?
+                            self.instantiate_component_from_exports(&exports, ty)?
                         }
                     };
                     self.result.initializers.push(init);
@@ -746,6 +746,7 @@ impl<'a, 'data> Translator<'a, 'data> {
     fn instantiate_component_from_exports(
         &mut self,
         exports: &[wasmparser::ComponentExport<'data>],
+        ty: ComponentInstanceTypeId,
     ) -> Result<LocalInitializer<'data>> {
         let mut map = HashMap::with_capacity(exports.len());
         for export in exports {
@@ -753,7 +754,7 @@ impl<'a, 'data> Translator<'a, 'data> {
             map.insert(export.name.0, idx);
         }
 
-        Ok(LocalInitializer::ComponentSynthetic(map))
+        Ok(LocalInitializer::ComponentSynthetic(map, ty))
     }
 
     fn kind_to_item(
--- a/crates/wasmtime/src/component/component.rs
+++ b/crates/wasmtime/src/component/component.rs
@@ -10,8 +10,8 @@ use std::path::Path;
 use std::ptr::NonNull;
 use std::sync::Arc;
 use wasmtime_environ::component::{
-    AllCallFunc, ComponentTypes, GlobalInitializer, InstantiateModule, StaticModuleIndex,
-    TrampolineIndex, Translator, VMComponentOffsets,
+    AllCallFunc, ComponentTypes, Export, GlobalInitializer, InstantiateModule, StaticModuleIndex,
+    TrampolineIndex, Translator, TypeComponentIndex, TypeDef, VMComponentOffsets,
 };
 use wasmtime_environ::{
     CompiledModuleInfo, FunctionLoc, HostPtr, ObjectKind, PrimaryMap, ScopeVec,
@@ -31,6 +31,9 @@ pub struct Component {
 }
 
 struct ComponentInner {
+    /// Component type index
+    ty: TypeComponentIndex,
+
     /// Core wasm modules that the component defined internally, indexed by the
     /// compile-time-assigned `ModuleUpvarIndex`.
     static_modules: PrimaryMap<StaticModuleIndex, Module>,
@@ -77,6 +80,7 @@ pub(crate) struct AllCallFuncPointers {
 
 #[derive(Serialize, Deserialize)]
 pub(crate) struct ComponentArtifacts {
+    ty: TypeComponentIndex,
     info: CompiledComponentInfo,
     types: ComponentTypes,
     static_modules: PrimaryMap<StaticModuleIndex, CompiledModuleInfo>,
@@ -239,7 +243,24 @@ impl Component {
             }),
         );
         let unlinked_compile_outputs = compile_inputs.compile(&engine)?;
-        let types = types.finish();
+
+        let (types, ty) = types.finish(
+            component
+                .component
+                .import_types
+                .iter()
+                .map(|(_, (name, ty))| (name.clone(), *ty)),
+            component.component.exports.iter().filter_map(|(name, ty)| {
+                let ty = match ty {
+                    Export::LiftedFunction { ty, .. } => TypeDef::ComponentFunc(*ty),
+                    Export::ModuleImport { ty, .. } => TypeDef::Module(*ty),
+                    Export::Instance { ty: Some(ty), .. } => TypeDef::ComponentInstance(*ty),
+                    Export::Type(ty) => *ty,
+                    _ => return None,
+                };
+                Some((name.clone(), ty))
+            }),
+        );
         let (compiled_funcs, function_indices) = unlinked_compile_outputs.pre_link();
 
         let mut object = compiler.object(ObjectKind::Component)?;
@@ -261,6 +282,7 @@ impl Component {
         };
         let artifacts = ComponentArtifacts {
             info,
+            ty,
             types,
             static_modules: compilation_artifacts.modules,
         };
@@ -280,6 +302,7 @@ impl Component {
         artifacts: Option<ComponentArtifacts>,
     ) -> Result<Component> {
         let ComponentArtifacts {
+            ty,
             info,
             types,
             static_modules,
@@ -317,11 +340,16 @@ impl Component {
 
         Ok(Component {
             inner: Arc::new(ComponentInner {
+                ty,
                 static_modules,
                 code,
                 info,
             }),
         })
+    }
+
+    pub(crate) fn ty(&self) -> TypeComponentIndex {
+        self.inner.ty
     }
 
     pub(crate) fn env_component(&self) -> &wasmtime_environ::component::Component {
--- a/crates/wasmtime/src/component/component.rs
+++ b/crates/wasmtime/src/component/component.rs
@@ -10,8 +10,8 @@ use std::path::Path;
 use std::ptr::NonNull;
 use std::sync::Arc;
 use wasmtime_environ::component::{
-    AllCallFunc, ComponentTypes, Export, GlobalInitializer, InstantiateModule, StaticModuleIndex,
-    TrampolineIndex, Translator, TypeComponentIndex, TypeDef, VMComponentOffsets,
+    AllCallFunc, ComponentTypes, GlobalInitializer, InstantiateModule, StaticModuleIndex,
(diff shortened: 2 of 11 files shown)


alexcrichton (commented):
Thanks for this! At a high level this looks pretty good, although there's two parts of this I'd have a comment on. One is that from a component model point of view an instance does not have imports, only exports and their types. In that sense `Instance::import_types` is not necessarily a component-model-based primitive here, and that's where `Component::imports()` would show up instead. I suspect though that this is critical to your use case, though.

To handle this I might recommend a slightly updated approach. From a `Linker` it should be possible to acquire a `types::Component` from a `component::Component`. This would synthesize the "InstanceType" data internally required to get a `Handle` which is more-or-less a glorified resource mapping. That would be a built-in way of asking "If I instantiate this component using this linker what resource would correspond to this import?" (happy to jump on a call to explain this more too)

The second thing is mostly minor API adjustments here and there, but nothing requiring major refactoring, so I'll note those later.

rvolosatovs:
> Thanks for this! At a high level this looks pretty good, although there's two parts of this I'd have a comment on. One is that from a component model point of view an instance does not have imports, only exports and their types. In that sense `Instance::import_types` is not necessarily a component-model-based primitive here, and that's where `Component::imports()` would show up instead. I suspect though that this is critical to your use case, though.
> 
> To handle this I might recommend a slightly updated approach. From a `Linker` it should be possible to acquire a `types::Component` from a `component::Component`. This would synthesize the "InstanceType" data internally required to get a `Handle` which is more-or-less a glorified resource mapping. That would be a built-in way of asking "If I instantiate this component using this linker what resource would correspond to this import?" (happy to jump on a call to explain this more too)
> 
> The second thing is mostly minor API adjustments here and there, but nothing requiring major refactoring, so I'll note those later.

Thanks for the quick response, is this what you mean https://github.com/bytecodealliance/wasmtime/pull/7804/commits/56f68d578a74694eb31e2cbc67ec4a913fb0b113 ?
Otherwise, happy to jump on a call tomorrow

alexcrichton:
Yep exactly! 

I'll go over this with a more fine-toothed comb tonight so you can have a review for when you wake up

crates/wasmtime/src/component/linker.rs
```diff
@@ -212,11 +195,59 @@ impl<T> Linker<T> {
             cx.definition(ty, import)
                 .with_context(|| format!("import `{name}` has the wrong type"))?;
         }
+        Ok(cx)
+    }
+
+    /// Iterates over all types imported by this component
+    pub fn import_types<'a>(
```
alexcrichton, line 202: Could this perhaps return a `types::Component`? If so then let's call this method `substituted_component_type` where the documentation indicates that the resource types imported by the `component` specified as an argument are replaced with the imports present in `Linker` in the returned type.

crates/wasmtime/src/component/mod.rs
```diff
@@ -24,7 +24,11 @@ pub use self::instance::{ExportInstance, Exports, Instance, InstancePre};
 pub use self::linker::{Linker, LinkerInstance, ResourceImportIndex};
 pub use self::resource_table::{ResourceTable, ResourceTableError};
 pub use self::resources::{Resource, ResourceAny};
-pub use self::types::{ResourceType, Type};
+pub use self::types::{
+    ComponentFunc as ComponentFuncType, ComponentInstance as ComponentInstanceType,
+    ComponentItem as ComponentItemType, CoreFunc as CoreFuncType, Module as ModuleType,
+    ResourceType, Type,
```
alexcrichton, line 30: Since `types` is a public module it's ok to avoid these reexports to avoid the same type being known under two different names.

crates/wasmtime/src/component/types.rs
```diff
@@ -757,3 +759,195 @@ impl Type {
...
+impl CoreFunc {
+    pub(crate) fn from(index: SignatureIndex, ty: &InstanceType<'_>) -> Self {
+        Self(Handle::new(index, ty))
+    }
+
+    /// Returns [`FuncType`] associated with this function
+    pub fn ty(&self) -> FuncType {
+        FuncType::from_wasm_func_type(self.0.types[self.0.index].clone())
+    }
+}
```
alexcrichton, line 776: Could those returning a `CoreFunc` return a `FuncType` directly instead?

crates/wasmtime/src/component/types.rs
```diff
@@ -757,3 +759,195 @@ impl Type {
...
+#[derive(Debug)]
+pub struct ComponentFunc(Handle<TypeFuncIndex>);
+
+impl ComponentFunc {
+    pub(crate) fn from(index: TypeFuncIndex, ty: &InstanceType<'_>) -> Self {
+        Self(Handle::new(index, ty))
+    }
+
+    /// Iterates over types of function parameters
+    pub fn params(&self) -> impl Iterator<Item = Type> + '_ {
```
alexcrichton, line 788: Could this and `results` be `ExactSizeIterator`?

crates/wasmtime/src/component/types.rs
```diff
@@ -757,3 +759,195 @@ impl Type {
...
+#[derive(Debug)]
+pub struct Module(Handle<TypeModuleIndex>);
+
+impl Module {
+    pub(crate) fn from(index: TypeModuleIndex, ty: &InstanceType<'_>) -> Self {
+        Self(Handle::new(index, ty))
+    }
+
+    /// Iterates over imports of the module
+    pub fn imports(&self) -> impl Iterator<Item = (&(String, String), ExternType)> {
```
alexcrichton, line 816: Could `&(String, String), ..` here become `&str, &str, ...`

crates/wasmtime/src/component/types.rs
```diff
@@ -757,3 +759,195 @@ impl Type {
...
+        self.0.types[self.0.index].imports.iter().map(|(name, ty)| {
+            (
+                name,
+                ExternType::from_wasmtime(self.0.types.module_types(), ty),
+            )
+        })
+    }
+
+    /// Iterates over exports of the module
+    pub fn exports(&self) -> impl Iterator<Item = (&String, ExternType)> {
```
alexcrichton, line 826: Could this be `&str` instead of `&String`?

crates/wasmtime/src/component/types.rs
alexcrichton: also `ExactSizeIterator` for this and `exports`

crates/wasmtime/src/component/types.rs
```diff
@@ -757,3 +759,195 @@ impl Type {
...
+    /// Returns import associated with `name`, if such exists in the component
+    pub fn get_import(&self, name: &str) -> Option<ComponentItem> {
+        self.0.types[self.0.index]
+            .imports
+            .get(name)
+            .map(|ty| ComponentItem::from(ty, &self.0.instance()))
+    }
+
+    /// Iterates over imports of the component
+    pub fn imports(&self) -> impl Iterator<Item = (&String, ComponentItem)> {
```
alexcrichton, line 854: Could this get the `ExactSizeIterator` plus `&str` treatment? (and `exports` below too)

crates/wasmtime/src/component/types.rs
```diff
@@ -757,3 +759,195 @@ impl Type {
...
+    /// Returns export associated with `name`, if such exists in the component instance
+    pub fn get_export(&self, name: &str) -> Option<ComponentItem> {
+        self.0.types[self.0.index]
+            .exports
+            .get(name)
+            .map(|ty| ComponentItem::from(ty, &self.0.instance()))
+    }
+
+    /// Iterates over exports of the component instance
+    pub fn exports(&self) -> impl Iterator<Item = (&String, ComponentItem)> {
```
alexcrichton, line 896: Like above, `ExactSizeIterator` + `&str`

crates/wasmtime/src/component/types.rs
```diff
@@ -757,3 +759,195 @@ impl Type {
...
+    pub(crate) fn from(index: TypeResourceTableIndex, ty: &InstanceType<'_>) -> Self {
+        Self(Handle::new(index, ty))
+    }
+
+    /// Returns [ResourceType] associated with this resource
+    pub fn ty(&self) -> ResourceType {
+        let idx = self.0.types[self.0.index].ty;
+        self.0.resources[idx]
+    }
+}
```
alexcrichton, line 918: Could this be removed in favor of using `ResourceType` directly?

crates/wasmtime/src/component/linker.rs
alexcrichton: Implementing this would probably require adding a `ComponentTypeIndex` somewhere inside of a component since I don't think that exists today. Furthermore that may not be altogether trivial to add. 

If it's not easy to add, some suggestions on this method:

* `ExactSizeIterator` + `&str` return value in the iterator
* Calling this `substituted_import_types` and documenting that the `Linker`'s resources are substituted in the component types. 
* Adding a `substituted_export_types` here as well for symmetry

rvolosatovs:
@alexcrichton I've studied the `wasmtime_environ` implementation a little bit and looks like I came up with a solution that makes sense (after all types are resolved, "synthesize" a `TypeComponent` and store in `ComponentTypes` on `finish`).

I've added a test case reusing the `dynamic::everything` (it does not actually contain everything, but hopefully this is "good enough"). PTAL

crates/wasmtime/src/component/component.rs
```diff
@@ -239,7 +243,24 @@ impl Component {
...
+                .import_types
+                .iter()
+                .map(|(_, (name, ty))| (name.clone(), *ty)),
+            component.component.exports.iter().filter_map(|(name, ty)| {
+                let ty = match ty {
+                    Export::LiftedFunction { ty, .. } => TypeDef::ComponentFunc(*ty),
+                    Export::ModuleImport { ty, .. } => TypeDef::Module(*ty),
+                    Export::Instance { ty: Some(ty), .. } => TypeDef::ComponentInstance(*ty),
+                    Export::Type(ty) => *ty,
+                    _ => return None,
```
alexcrichton, line 259: This is the main thing that I'm worried about with this PR right now. This means that if a component exports a locally-defined core wasm module and/or a "bag of items" instance then wasmtime will lie and not actually report the export from the type. I think it should be possible to implement this though with some more type-threading-around, however. All the types of all intermediate values are present and accounted for during validation, we just need to pluck it out at the right time and thread it through the system to show up here.

rvolosatovs: It appears that we can pluck most of these out, but not for all cases, see updated implementation

crates/environ/src/component/translate.rs
```diff
@@ -581,18 +581,18 @@ impl<'a, 'data> Translator<'a, 'data> {
...
                             args,
                         } => {
-                            let types = self.validator.types(0).unwrap();
-                            let ty = types.component_instance_at(index);
                             let index = ComponentIndex::from_u32(component_index);
                             self.instantiate_component(index, &args, ty)?
                         }
                         wasmparser::ComponentInstance::FromExports(exports) => {
-                            self.instantiate_component_from_exports(&exports)?
+                            self.instantiate_component_from_exports(&exports, ty)?
```
rvolosatovs, line 595: @alexcrichton I'm getting a bit lost here, is `ty` here the correct type index or is it actually the parent instance type index?

alexcrichton (approved):
Looks great to me, thanks!

sehz:
Nice work!

PR #7807, opened 2024-01-24
WinML backend for wasi-nn

jianjunz:
This change adds Windows Machine Learning backend for wasi-nn. Since WinML is a built-in feature of Windows, this change makes it possible to run wasi-nn on Windows without a third-party machine learning framework.

WinML backend is only available on Windows.

WinML: https://learn.microsoft.com/en-us/windows/ai/windows-ml/


crates/test-programs/Cargo.toml
```diff
@@ -20,3 +20,4 @@ futures = { workspace = true, default-features = false, features = ['alloc'] }
 url = { workspace = true }
 sha2 = "0.10.2"
 base64 = "0.21.0"
+image2tensor = { git = "https://github.com/bytecodealliance/wasi-nn.git", rev = "dec0dd4" }
```
jianjunz, line 23: It looks like `cargo vet` doesn't like this git link. We'll need to release a new version for image2tensor and update the version number here.

abrown:
@jianjunz:now that #7846 is merged you should be able to rebase on that and get `cargo vet` to pass.

crates/test-programs/Cargo.toml
```diff
@@ -19,4 +19,4 @@ getrandom = "0.2.9"
 futures = { workspace = true, default-features = false, features = ['alloc'] }
 url = { workspace = true }
 sha2 = "0.10.2"
-base64 = "0.21.0"
+base64 = "0.21.0"
```
abrown, line 22: Probably an editor auto-format thing: following this project's convention, we can leave this trailing new line in.

crates/wasi-nn/Cargo.toml
```diff
@@ -24,8 +24,20 @@ wasmtime = { workspace = true, features = ["component-model"] }
...
+    "runtime-linking",
+], optional = true }
+
+[target.'cfg(windows)'.dependencies.windows]
+version = "0.52"
+features = [
+    "AI_MachineLearning",
+    "Storage_Streams",
+    "Foundation",
+    "Foundation_Collections",
```
abrown, line 38: Are all of these features needed? I would have assumed that only `"AI_MachineLearning"` is needed, which adds the right dependent features (see [here](https://github.com/microsoft/windows-rs/blob/a2fbd102010a20ac35ff0d950e44f193cd032751/crates/libs/windows/Cargo.toml#L36)). The reason I'm thinking about this is that the `windows` crate totals 11MB and I don't want to unnecessarily bloat Wasmtime binaries on Windows.

crates/wasi-nn/src/backend/winml.rs
```diff
@@ -0,0 +1,165 @@
...
+            session,
+            result: None,
+        }
+    }
+}
+
+impl BackendExecutionContext for WinMLExecutionContext {
+    fn set_input(&mut self, index: u32, tensor: &Tensor) -> Result<(), BackendError> {
+        let input = self.session.Model()?.InputFeatures()?.GetAt(index)?;
+        // TODO: Bind input.
```
abrown, line 100: Is this resolved? Looking at the binding code below, I believe we need to handle more than just the `f32` case &mdash; there are other possible models with other tensor item types.

crates/wasi-nn/src/backend/winml.rs
```diff
@@ -0,0 +1,165 @@
...
+    }
+}
+
+impl BackendFromDir for WinMLBackend {
+    fn load_from_dir(
+        &mut self,
+        path: &Path,
+        target: ExecutionTarget,
+    ) -> Result<Graph, BackendError> {
+        let model = read(&path.join("mobilenet.onnx"))?;
```
abrown, line 59: This should handle any kind of model, not just mobilenet. Perhaps just look for the `.onnx` suffix?

crates/wasi-nn/src/backend/winml.rs
```diff
@@ -0,0 +1,165 @@
...
+            .Outputs()?
+            .Lookup(&output_name_hstring)?
+            .cast::<TensorFloat>()?
+            .GetAsVectorView()?;
+        let output: Vec<f32> = vector_view.into_iter().collect();
+        unsafe {
+            destination.copy_from_slice(std::slice::from_raw_parts(
+                output.as_ptr() as *const u8,
+                output.len() * 4,
+            ));
```
abrown, line 146: I believe this is another place where the tensor item type is hard-coded: we need to be able to handle other types.

crates/wasi-nn/tests/all.rs
```diff
@@ -45,13 +45,17 @@ impl Ctx {
         let wasi = builder.build();
 
         // Create the wasi-nn context.
-        let mut openvino = backend::openvino::OpenvinoBackend::default();
+        #[cfg(feature = "openvino")]
+        let mut backend = backend::openvino::OpenvinoBackend::default();
+        #[cfg(feature = "winml")]
+        let mut backend = backend::winml::WinMLBackend::default();
```
abrown, line 51: This won't work when both features are enabled. Features are additive, so both backends can be present at the same time.

crates/wasi-nn/tests/all.rs
```diff
@@ -90,3 +94,9 @@ fn nn_image_classification() {
 fn nn_image_classification_named() {
     run(NN_IMAGE_CLASSIFICATION_NAMED, true).unwrap()
 }
+
+#[cfg_attr(not(all(target_arch = "x86_64", target_os = "windows")), ignore)]
+#[test]
+fn nn_image_classification_winml() {
+    run(NN_IMAGE_CLASSIFICATION_WINML, true).unwrap()
```
abrown, line 101: You will want to double-check that this test is actually run. Right now in Wasmtime's CI, only a subset of jobs run to preserve runner time. The Windows "test" job does not run by default but if you add the string `prtest:full` in any commit, then all jobs will run. Then we can check the logs for the Windows job to ensure this test actually ran.

crates/test-programs/Cargo.toml
jianjunz: This dependency has been removed. Test image was converted to tensor data offline.

crates/test-programs/Cargo.toml
jianjunz: Done. Restored the trailing new line.

crates/wasi-nn/Cargo.toml
jianjunz: "Storage_Streams" and "Foundation_Collections" are needed because some WinML APIs need data structures defined by these two features. "Foundation" is removed.

crates/wasi-nn/src/backend/winml.rs
jianjunz: 4 is the size of TensorFloat, which is the output data type for probability.

crates/wasi-nn/tests/all.rs
jianjunz: We run openvino and winml backend in two different test cases, with different inputs, so only one backend is specified here for testing purpose.

backend.list() is able to list all available backends.

https://github.com/bytecodealliance/wasmtime/blob/9732234291ffc9486b95be5309f920775b20d963/crates/wasi-nn/src/backend/mod.rs#L21-L31

crates/wasi-nn/src/backend/winml.rs
jianjunz: Only f32 is supported right now. A check is added for rejecting unsupported tensor types.

I'm planning to add support for other types in a separate PR since more test data is needed for converging these new tensor types. It may need additional changes for image2tensor to generate the input data.

crates/wasi-nn/src/backend/winml.rs
jianjunz: Renamed to model.onnx to align with openvino backend.

crates/wasi-nn/tests/all.rs
jianjunz: With prtest:full, this test failed on Windows Server. I guess GitHub Windows Server runner doesn't have Desktop Experience enabled.

jianjunz: I'm afraid we need to disable this test at this time. GitHub Actions Windows Server image doesn't have Desktop Experience available, so it cannot be installed.

List of installed or available features:
https://github.com/jianjunz/wasmtime/actions/runs/7752997447/job/21143496106#step:9:1

jianjunz:
Thanks for auditing windows-* crates, but `cargo vet` still fails after #7900. Only 1 unvetted dependency now.

```
1 unvetted dependencies:
  windows:0.52.0 missing ["safe-to-deploy"]
```

crates/wasi-nn/tests/all.rs
abrown: Yeah, that's unfortunate. Why don't we add a function like `check_openvino_is_installed` but for WinML: `check_winml_is_available`. There's probably a way to check with the Windows API if ML system calls are possible and that function could return false if they are not.

That extra runtime check would be in addition to the `#[cfg_attr(not(feature = "winml"), ignore)]`. That can stay there for the case where the `winml` feature isn't even enabled.

jianjunz:
The code is rebased. No conflict. However, your commit https://github.com/bytecodealliance/wasmtime/commit/f81c59d0495630fe82e16d352786aa10358aa281 was dropped because I forgot to pull the code first before rebasing. Could you please submit it again to fix the `cargo vet` failure? Thanks and sorry for the mistake.

jianjunz:
Thanks for your review and merging this PR.

Issue #7665, opened 2023-12-08
Demangle symbol names in guest-profiler output


jameysharp:
#### Feature

The `GuestProfiler`, as invoked by its Rust API or with the Wasmtime CLI's `--profile=guest` flag, should try to demangle C++ and Rust function names before emitting them in the generated profile.

#### Benefit

It'd be nicer to see names like `std::rt::lang_start::{{closure}}::ha2ad3bd729827662` than like `_ZN3std2rt10lang_start28_$u7b$$u7b$closure$u7d$$u7d$17ha2ad3bd729827662E.llvm.16041094552022133104`.

#### Implementation

There's existing code in Wasmtime for demangling names in stack traces, in `crates/wasmtime/src/trap.rs`. But that's bypassed by the low-level way that the guest profiler collects stack traces. Instead, the guest profiler gets all the function names up front, in the `module_symbols` function in `crates/wasmtime/src/profiling.rs`, and it should demangle the names there.

#### Alternatives

There might be support in the https://profiler.firefox.com/ frontend for demangling names.

jameysharp:
Awesome!

After I wrote this I've noticed that some people are already getting demangled names in guest-profiler output, which is confusing me. That makes me think that there may have been a Rust toolchain change to how symbol names are emitted in WebAssembly target builds. So it might not be necessary to do this if toolchain developers are doing it for us. But I haven't looked into that yet to find out if this is a trend or something.

So there are two ways you could help here:
- One is to go ahead and implement the change I described. At worst, it's convenient when dealing with wasm modules built by older toolchains, even if new ones make it unnecessary.
- The other is to look into what Rust, Clang, LLVM, etc are doing and see whether they're changing how they emit symbol names. It would be nice to know how important this change is for usability.

I'd appreciate your help with either or both of these things!

high-cloud:
Hi @jameysharp , I have finished the first thing you memtioned. Can you help review the pr?

casimiro:
Hi,

I'm seeing an unexpected behaviour that seems related to this issue.

After upgrading from `14.0.3` to `15.0.0` I've noticed that `wasm_trap_message` returns the backtrace with mangled symbols:
```
error while executing at wasm backtrace:
    0: 0x2dbfa - <unknown>!__rust_start_panic
    1: 0x2d9e5 - <unknown>!rust_panic
[...]

PR #7809, opened 2024-01-24
demange names in guest-profiler output

high-cloud:
issue #7665

### DIFF
--- a/crates/wasmtime/src/profiling.rs
+++ b/crates/wasmtime/src/profiling.rs
@@ -10,6 +10,7 @@ use fxprof_processed_profile::{
 use std::ops::Range;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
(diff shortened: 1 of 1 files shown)


jameysharp (approved):
This is great, thank you!

I hesitated about the use of `unwrap`, but `demangle_function_name` can only fail if the writer fails, and I don't think the `Write` implementation for `String` can fail, so I'll sign off that this use is fine. We have another `unwrap` on the use of this function in `CompiledModule::register_debug_and_profiling`, anyway.

While looking at that, I just noticed that there's a `demangle_function_name_or_index` function which is very similar to what this whole `match` statement does. I think it's harmless to duplicate that here, with a slightly different format for the fallback case when there's no function name.

I'm going to merge this as-is. But if you want to open a follow-up PR, I think it would be an improvement to use `demangle_function_name_or_index` here.

PR #7811, opened 2024-01-24
c-api: expose tail call configuration option in c api

womeier:
The tail call proposal configuration is not yet exposed by the C api.
This change intends to allow configuring the option in e.g. the python library bindings:
https://github.com/bytecodealliance/wasmtime-py/issues/204

I wanted to test this change doing the following:

Following the [docs](https://docs.wasmtime.dev/contributing-building.html), I ran
`cargo build --release --manifest-path crates/c-api/Cargo.toml` which should place `libwasmtime.{a,so}` in `target/release`.
The command succeeded, but the file is not generated (and I cannot find libwasmtime.so with `find`).

~I thus didn't test the change further than the fact that it compiles, in particular I didn't yet test the change
with the python library.~
Tested to work together with the PR: https://github.com/bytecodealliance/wasmtime-py/pull/205 adjusting the python bindings.

### DIFF
--- a/crates/c-api/include/wasmtime/config.h
+++ b/crates/c-api/include/wasmtime/config.h
@@ -150,6 +150,13 @@ WASMTIME_CONFIG_PROP(void, max_wasm_stack, size_t)
 WASMTIME_CONFIG_PROP(void, wasm_threads, bool)
 
 /**
+ * \brief Configures whether the WebAssembly tail call proposal is enabled.
+ *
+ * This setting is `false` by default.
+ */
+WASMTIME_CONFIG_PROP(void, wasm_tail_call, bool)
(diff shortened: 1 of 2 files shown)


alexcrichton:
Looks great to me, thanks! Apologies for the documentation but it's outdated. Nowadays the correct invocation is with `-p wasmtime-c-api` instead of `--manifest-path ...`. I'll send a separate PR to update those docs.

womeier:
Thanks for the quick response @alexcrichton!

I created a PR with the changes to the python bindings, they work as expected together, i.e. I can run my wasm module with tail calls :)

alexcrichton (approved):

PR #7812, opened 2024-01-24
wasi-io: Reimplement wasi-io/poll using a Pollable trait

badeend:
Prior discussion: https://bytecodealliance.zulipchat.com/#narrow/stream/217126-wasmtime/topic/Change.20Subscribe.20trait

- Renamed the existing `Pollable` struct to `PollableResource`
- Reimplemented wasi-io/poll. This introduces a new `Pollable` trait which is lower level, doesn't require heap allocations to poll, has mutable access to the WasiView, and can be used as a standalone resource without a parent. The Subscribe trait is kept intact, but this is now a utility interface, implemented in terms of Pollable.
- Eliminate the (now) unnecessary surrogate parent resource of clock pollables
- Added ResourceTable `take` & `restore` as a general purpose replacement for `iter_entries`. That one was used only by the old `poll` implementation.

---

Additionally:
- @pchickey  Forbid empty poll list. Fixes: WebAssembly/wasi-io#67


pchickey (commented):
Thanks, this is excellent work. The new internal interface is definitely superior to the old one, and I appreciate the new tests as well. The lease system for resource table is a much better interface than iter_entries.

I believe that all of the `Slot` and `SlotIdentity` implementation makes sense and is correct (especially because, surprisingly to me, the only unsafe is in `unsafe impl Send`), but I wanted to tag @alexcrichton to double-check that part because it is not code I feel super confident in my ability to assess that code. If he is happy with that, this can land.

crates/wasi/src/preview2/poll.rs
```diff
@@ -1,174 +1,257 @@
...
     async fn ready(&mut self);
 }
 
-/// Creates a `pollable` resource which is susbcribed to the provided
-/// `resource`.
+impl<T: Subscribe> Pollable for T {
+    fn poll_ready(&mut self, cx: &mut Context<'_>, _view: &mut dyn WasiView) -> Poll<()> {
+        self.ready().as_mut().poll(cx)
+    }
+}
```
alexcrichton, line 44: This I think is actually subtly incorrect because it drops the future after the call to `poll` which signals that the future should be cancelled rather than keeping it alive until the whole poll is done. I think that means that we may not be guaranteed to get wakeups from cancelled futures, although we might get those for now given how the code is currently constructed. I think that this'll need to be a bit fancier with the adapter instead of having a blanket impl

crates/wasi/tests/all/sync.rs
```diff
@@ -294,6 +294,14 @@ fn preview2_stream_pollable_traps() {
     )
 }
 #[test_log::test]
+fn preview2_pollable_correct() {
+    run(PREVIEW2_POLLABLE_CORRECT_COMPONENT, false).unwrap()
+}
+#[test_log::test]
+fn preview2_pollable_traps() {
+    run(PREVIEW2_POLLABLE_TRAPS_COMPONENT, false).unwrap_err();
```
alexcrichton, line 302: There's quite a lot of errors which go through the `Err` here so I'm a bit wary of using `unwrap_err` to assert that a particular trap happened, could this instead a particular property about the error, such as its message?

crates/wasi/src/preview2/poll.rs
```diff
@@ -1,174 +1,257 @@
...
+
+/// Create a pollable that is always ready.
+///
+/// Similar to [std::future::ready].
+pub fn ready() -> impl Pollable {
```
alexcrichton, line 49: Could this, and the functions below, be implemented in terms of `poll_ready_fn` perhaps?

alexcrichton (commented):
Thanks for this! I've left a few comments but it's getting a bit later here so I'm going to head out. I want to read more about the `Lease<T>` though as that looks quite subtle and I need to think more about it.

crates/wasi/tests/all/sync.rs
badeend: Certainly

crates/wasi/src/preview2/poll.rs
badeend: Check!

alexcrichton (commented):
Ok I've gotten a chance now to take a closer look at `Lease<T>` and the changes `ResourceTable`. Given this new API design for the pollable trait something along these lines is required (e.g. `iter_children` can't work any more). The specific implementation here I think has a drawback where it's very "panicky" if you get it wrong. Most other aspects of WASI are "error-y" in that they try to return traps if anything is gotten wrong. This I think is pretty important for not accidentally becoming a DoS vector for embeddings. For example a panicking `Drop` implementation means that an early-return in an embedder function might accidentally take down the whole process where a wasm trap would only take down a single instance.

Given that my main thought on this is that this should ideally not ever panic and instead should switch to returning errors where possible. I might also recommending going a little bit further perhaps with a scheme such as:

* Leave `TableEntry::entry` as `Box<dyn Any>`.
* Change `take` to returning `Box<T>`. This would replace `entry` with something like `Box::new(Tombstone)` which is a private type to this module.
* Change `restore` to taking `Box<T>` and `Resource<T>`.

That way it's largely up to embedders to "get everything right" but they'd already be required to do so with this current API. Additionally any failed downcasts can additionally add a check for `Tombstone` to perhaps return a more precise error other than `ResourceTableError::WrongType` with a new variant such as `TakenValue`.

badeend:
Thanks for the feedback.
I agree on the "panicky" point. I'll add an error type and remove the panics.

One thing that's not in this PR, but I assume will most likely be added at some point, are untyped `take_any` and `restore_any` variants. The drawback of reverting the Lease & SlotIdentity design, is that the `restore(_any)` API becomes (even) easier to misuse. Because then the consumer can restore any value at the index of a previously differently typed entry. I'm worried about the developer experience of this, as the place where they encounter the WrongType errors could be miles away from where the problem actually is.

Anyway, I'm fine with your suggestions. I just wanted to make sure the trade-offs are known.

alexcrichton:
Hm ok, for that I think you have a good point about accidentally messing up these APIs. I think this may still be surmountable perhaps with some trickery, but I'd also need to see the usage of `take_any` and restore to know better. Want to discuss on a future PR with that implemented or hash it out here? (I'm fine either way)

crates/wasi/src/preview2/poll.rs
```diff
@@ -1,174 +1,257 @@
...
+            .expect("parent to exist");
+        let poll = parent.poll_ready(cx, view);
+        view.table().restore(parent);
+        poll
+    })));
     Ok(table.push_child(pollable, &resource)?)
 }
 
+/// A host representation of the `wasi:io/poll.pollable` resource.
+pub struct PollableResource {
```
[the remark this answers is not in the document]
badeend: The design has changed in the meantime. Now it's not just a box anymore, so I went with `PollableHandle`.

crates/wasi/src/preview2/poll.rs
badeend: The tests were failing precisely because of that. To fix it, I ended up rougly back at the `Subscribe` design, but this time returning a custom WasiFuture type that has an additional `WasiView` parameter on its `poll` method. To prevent scope creep of this PR, I kept `Subscribe` alive for now (as `PollableAsync`). But in the long run, I don't think there's much value in havin them both, as `Subscribe` can be trivially converted to `Pollable` from:

```rs
#[async_trait::async_trait]
impl PollableAsync for HostFutureIncomingResponse {
    async fn ready(&mut self) {
        if let Self::Pending(handle) = self {
            *self = Self::Ready(handle.await);
        }
    }
}
```

to:

```rs
impl Pollable for HostFutureIncomingResponse {
    fn ready<'a>(&'a mut self) -> Pin<Box<dyn WasiFuture<Output = ()> + Send + 'a>> {
        Box::pin(async {
            if let Self::Pending(handle) = self {
                *self = Self::Ready(handle.await);
            }
        })
    }
}
```

One obvious downside is the added visual noise.

badeend:
I chose to go with a hybrid approach. For the public API, I changed it to what you suggested. Internally, I removed Lease & SlotIdentity. But I kept Slot to perform the resource type check. Also, as part of the updated design (see above) I needed `take_any` and `restore_any` so I included those as well.

badeend:
I think you mean changing

```
async fn ready(&mut self);
```
to 
```
async fn ready(&mut self, table: &ResourceTable);
```
right?

That doesn;t work because `&ResourceTable` is not Send as `ResourceTable` is not Sync.

alexcrichton:
Good point, yes, I'm more-or-less saying we should do that. (either that or use a task-local but I think that still captures `&T`). 

Mind trying make `ResourceTable` implement `Sync`? I think that's probably the addition of a few trait bounds in its internal trait objects. I think everything we put in there is already `Sync` although if it things aren't currently `Sync` that'll pose a larger problem.

badeend:
I understand your concerns, yet I'd rather not go for round three right now, which would include reverting https://github.com/bytecodealliance/wasmtime/pull/7802. So instead, I've changed the questionable types to be private to the `poll.rs` module. That way, all the iffy-ness is contained to just a single file that we can iterate on later. From the outside nothing significant has changed, except that now I can use `poll_ready_fn`, which is what I personally was after.

alexcrichton:
Wanted to say I have not forgotten about this, I have been looking for time to write up something longer-form, which I hope to get to by tomorrow. Is this blocking anything though that it would be prudent to land now rather than later? If so I think it's good to go as-is, but otherwise I'd like to take some more time to write up longer-form thoughts.

badeend:
There's no immediate rush from my side, so feel free to take your time.

alexcrichton:
Ok thanks again for your patience here, very much appreciated!

I've gotten some time to think and work on this. I was leaning towards merging this, but then I realized that I'd prefer to avoid a situation where we land this and then later revert most of it towards a different strategy. In that sense I wanted, time permitting, to take a moment and figure out if alternative strategies would work. I'm getting a growing sense of unease with this direction as it's more-or-less a custom `Future` trait and is something we'd ideally avoid.

So assuming that the main goal of this PR is to get access to `ResourceTable` during `async fn ready` I originally suggested the `borrow`/`borrow_mut` idea above using `RefCell`. I tried implementing that and turns out it doesn't work. That means that `ResourceTable` contains `RefCell` and `async fn ready` would close over `&ResourceTable` (e.g. it's a new function argument). In such a situation it means that the returned future, which must be `Send`, closes over `&ResourceTable`. That type is not `Send` because it requires `ResourceTable: Sync` which is not satisfied with `RefCell`. So that cans the idea of. using `RefCell`.

After talking a bit more with @pchickey, however, I'm growing more fond of the idea of using `RwLock<T>` here instead of `RefCell<T>`. Not for the actual blocking aspect but instead only for the "it's `Sync`" aspect. To that end I [implemented this on a branch](https://github.com/alexcrichton/wasmtime/commit/980c8b7b852b974bd7c73c8ff95108e5dbf77618) and got tests passing with it. The changes are:

* Replace `ResourceTable::get` with `ResourceTable::{borrow, borrow_mut}` that return `RwLock{Read,Write}Guard<T>`.
* Remove table methods returning `Any`
* Add `&ResourceTable` as an argument to the `ready` async function.
* Replace most usage of `table().get()` with `table().get_mut()` (avoids locks)
* Use `u32` indices in `Pollable`'s `make_future` internals instead of `Any`
* Rewrite headers in wasi-http to avoid needing `Any` by representing headers as `Resource<Resource<hyper::HeaderMap>>`

The major consequences of this decision, however, are:

* `ResourceTable::{borrow, borrow_mut}` require atomic manipulations. No blocking, but it's atomics for something that's not contended 99.9% of the time.
* The `std::sync::RwLock` type cannot be used because `std::sync::RwLock{Read,Write}Guard` is not `Send`. I temporarily added a `tokio` dependency to `wasmtime`-the-crate and used `tokio::sync::RwLock` instead. Long-term I would like to avoid a `tokio` dep in the `wasmtime` crate.
* There's a few minor cleanup still to be had in terms of threading a few more errors in a few more places.

Personally I'm inclined to take a route that looks like this, namely threading arguments through `async fn` rather than threading arguments through `fn poll`. This is a foundational change to how things work though, especially around a new footgun of not being able to `borrow_mut` twice. In that sense I'd like to get feedback along the lines of:

* @pchickey does this all sound reasonable enough to you?
* @badeend does this still solve your original use case, and if so what do you think about this approach vs the `poll` approach?

alexcrichton:
also cc @elliottt since you've touched a lot of WASI internals and you probably want to take a look too

PR #7815, opened 2024-01-24
mpk: enable MPK if available in CI

abrown:
After discussing several options for testing MPK in CI, this stopgap
measure at least provides some coverage. Other options include
maintaining a separate MPK-enabled CI runner (configuration is not
transparent, hard to maintain) or running the MPK-enabled tests in a
system-mode QEMU VM (tricky to get right, also hard to maintain). For
now, those of us at the Cranelift meeting agreed this at least gets some
CI testing for the MPK bits, which shouldn't be changing too often.
Since not all GitHub runners have MPK available, we only set the
`WASMTIME_TEST_FORCE_MPK` environment variable when it is. This change
also emits a warning to the GitHub logs in either case for easier
troubleshooting (e.g., to attempt to provide context if MPK logic breaks
and is only found in a later CI run).

prtest:full

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -520,6 +520,18 @@ jobs:
       shell: pwsh
       if: matrix.os == 'windows-latest'
 
+    # Since MPK (PKU) is not present on some GitHub runners, we check if it is
+    # available before force-enabling it. This occasional testing is better than
+    # none at all; ideally we would test in a system-mode QEMU VM.
+    - name: Force-run with MPK enabled, if available
+      run: |
+        if cargo run --example mpk-available; then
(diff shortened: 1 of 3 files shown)


alexcrichton:
Could this add a dedicated builder which is the only one enabling MPK? That one could perhaps run just a small subset of the test suite as well. I'm also a bit worried about the inverse where we always run with MPK by accident somehow and never see issues when it's turned off.

Otherwise, could the warning be tweaked a bit? Ideally what I'd imagine is that on failure a warning would be printed, but if everything succeeded no warning necessary as otherwise it can clutter up the view (e.g. [this link](https://github.com/bytecodealliance/wasmtime/actions/runs/7646831639?pr=7815) has a number of non-actionable warnings at the bottom)

abrown:
Take a look at [this run](https://github.com/bytecodealliance/wasmtime/actions/runs/7763050750); this change will now only cause a single warning when MPK is not present.

alexcrichton (approved):

Issue #7813, opened 2024-01-24
`ESPIPE` errors unless `--wasi threads=y` is specified in wasmtime 15 and 16


brettcannon:
### Test Case

Various CPython test cases have started failing when moving from wasmtime 13 to 16 due to `ESPIPE`. Adding `--wasi threads=y` fixes the issue which seems unrelated to file seeking.

Example failure run can be seen at https://buildbot.python.org/all/#/builders/1124/builds/640/steps/11/logs/stdio right after I switched to wasmtime 16 from 13 (wasmtime 14.0.4 does not seem to have the issue).

### Steps to Reproduce

* `git clone https://github.com/python/cpython.git`
* `cd cpython`
* `python3 Tools/wasm/wasi.py build`
* Edit `cross-build/wasm32-wasi/python.sh` and remove `--wasi threads=y`
* `./cross-build/wasm32-wasi/python.sh -m test test___all__`

You can also try  test___all__ test_bufio test_compileall test_dbm_dumb test_importlib test_io test_marshal test_pathlib test_posix test_runpy test_tarfile test_zipfile .

### Expected Results

The test should pass w/o issue (it was working fine under wasmtime 13 w/ e.g. `wasmtime run --env PYTHONPATH=/builddir/wasi/build/lib.wasi-wasm32-3.12:/Lib --mapdir /::/home/brettcannon/Repositories/python/cpython-3.12 -- python.wasm`.

### Actual Results

W/o `--wasi threads=y`, it causes a `ESPIPE`.

### Versions and Environment

Wasmtime version or commit: 16.0.0 and 15.0.1

Operating system: Linux

Architecture: x64

### Extra Info

WASI-SDK 20.

alexcrichton:
> Adding --wasi threads=y fixes the issue which seems unrelated to file seeking.

Under the hood this toggles the `-Spreview2=n` flag (sorry it's confusingly named). Basically we've got two implementations of preview1 in-tree. One is based on preview2 primitives and is the current default (`-Spreview2`, the `wasmtime_wasi::preview2` module). The other is the historical version based on preview1 (`-Spreview2=n`, the `wasi-common` crate).

What you're probably seeing here is an unintended difference in the implementation between the two and a bug we need to fix (thanks for filing!).

[...]

PR #7816, opened 2024-01-24
Close intermediate streams for reads/writes

alexcrichton:
Previously temporary streams created as part of the preview1 adapter in the wasmtime-wasi crate were left open which meant that they continued to occupy space in the resource table and the underlying file accidentally wasn't ever actually closed.

cc #7813


pchickey (approved):
Nice find! Glad this was easy to fix

Issue #7731, opened 2023-12-28
Performance issue related to floating-point (FP) arithmetic


ShuyaoJiang:
### Summary
Hi, I ran the attached case (C program, compiled to Wasm by `Emscripten`) in different Wasm runtimes, and found abnormal performance in `Wasmtime` compared with three other runtimes. The execution time of this case (time interval from the start to the end of the execution of Wasm bytecode running command) on different runtimes is as follows:

- Wasmtime: 3057ms
- Wasmer: 920ms
- WasmEdge (AOT): 759ms
- WAMR (AOT): 754ms

#### Emscripten

- emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.49 (04a0cad4d4c9f5d62876821274d78cd0a52427af)
clang version 18.0.0 (https://github.com/llvm/llvm-project 269685545e439ad050b67740533c59f965cae955)
Target: wasm32-unknown-emscripten
Thread model: posix

#### Wasm Runtime Version

- Wasmtime: cli 15.0.0
- Wasmer: 4.2.3
- WasmEdge (AOT): 0.13.5
- WAMR (AOT): 1.2.3

#### Hardware & OS

- CPU: Intel(R) Xeon(R) E5-2686 v4 CPU @ 2.30GHz
- Memory: 16GB
- OS: Ubuntu 20.04.6 LTS

### Additional details
The attached source program is synthesized by a [Csmith](https://github.com/csmith-project/csmith) seed and a code snippet of floating-point (FP) arithmetic. The inserted code snippet is on line `946-1039` of the source program. We found that this abnormal performance would not occur when running the seed program alone. So, we believe that this anomaly may be caused by some improper handling of FP arithmetic. Could you please check this situation? Thank you!

[1702437936230.zip](https://github.com/bytecodealliance/wasmtime/files/13787137/1702437936230.zip)

fitzgen:
Thanks for the bug report. Could you also separate timing of compilation vs execution for this bug as well? Thanks!

alexcrichton:
It's been awhile since this was last opened, but I wanted to be sure to drop in to say thanks for the detailed report, it's appreciated!

I was able to reproduce this and I think I fixed at least some part of the slowdown in https://github.com/bytecodealliance/wasmtime/pull/7818. I haven't dug too much further though.

alexcrichton:
[...]

PR #7098, opened 2023-09-27
x64: Fix false dependencies in int-to-float conversions

alexcrichton:
This commit is a result of the investigation on #7085. The int-to-float conversion instructions used right now on the x64 backend will implicitly source the upper bits of the result from a different register. This implicitly creates a dependency on further consumers using the conversion result on whatever previously defined the upper bits, even though they aren't used. This false dependency is the primary reason for the slowdown witnessed in #7085.

The fix chosen in this commit is to model the int-to-float instructions with a new shape of instruction instead of the previous `GprToXmm{,Vex}`. This previous shape was modeled as single-input and single-output, but this does not reflect the actual nature of the `cvtsi2s{s,d}` instructions. Instead these now use `CvtIntToFloat{,Vex}` which have two source operands and one destination operand, modeling how the upper bits of a different register are used. In lowerings using this instruction the upper bits to preserver are always sourced from a zero'd out register to force breaking dependencies between instructions.

Closes #7085

PR #7818, opened 2024-01-24
x64: Break more data dependencies in float-related instructions

alexcrichton:
This commit takes a stab at #7816 without diving a whole lot into it. I noticed that the loop started with `vcvtss2sd` which is along the same lines as previous false dependencies found earlier in PRs such as #7098. I had forgotten these instructions at the time and meant to go back and touch them up and #7731 has provided sufficient motivation to do so!

Locally this takes that test case from 1.6s to 0.4s for me.


fitzgen (approved):
Nice!! Out of curiosity, how did you end up root causing that perf bug to this false dependency?

alexcrichton:
Ah it was mostly from previous experience. I knew there were a set of instructions in the back of my mind which we still did the "fake the output register as the input" for AVX (e.g. the instructions modified here) and when I ran `perf` over the program the first very hot instruction in a loop was `vcvtss2sd` which I remembered was one of those. To test out I split the dependencies and then the performance improved so I assumed it was the cause.

PR #7826, opened 2024-01-26
wasmtime: Rename `SignatureFooBar` to `TypeFooBar`

fitzgen:
No functional changes, just the following mechanical renames:

* `VMSharedSignatureIndex` to `VMSharedTypeIndex`
* `SignatureIndex` to `TypeIndex`
* `SignatureRegistry` to `TypeRegistry`
* and more

This is intended to start paving the way for Wasm GC support, where there are more than just function signatures in a Wasm module's type section, and we are going to need to register non-function-signature types in the registry as well, for things like casting between reference types and passing reference types across Wasm modules.


fitzgen:
Flagging @alexcrichton in particular for this review because I realized that we actually already had `wasm_types::TypeIndex` and AFAICT it is actually the same as `wasm_types::SignatureIndex` and so, in the process of fixing compile errors, I deleted the duplicate definition of `TypeIndex` (that used to be `SignatureIndex`) and everything is using the original `TypeIndex` now. But that is a little surprising, and I want to make sure that this isn't going to introduce any subtle bugs. Thoughts?

fitzgen:
@alexcrichton I added a second commit that reintroduces separate index types for wasm-index-space types vs module-interned types.

Going to work on rebasing and resolving conflicts now.

fitzgen:
Rebased.

alexcrichton (approved):

PR #7827, opened 2024-01-26
refactor demangling part in profiling.rs

high-cloud:
…r_index

issue #7665
replace the match statement with a function already exist

### DIFF
--- a/crates/wasmtime/src/runtime/profiling.rs
+++ b/crates/wasmtime/src/runtime/profiling.rs
@@ -10,7 +10,7 @@ use fxprof_processed_profile::{
 use std::ops::Range;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
-use wasmtime_environ::demangle_function_name;
+use wasmtime_environ::demangle_function_name_or_index;
 use wasmtime_runtime::Backtrace;
 
 // TODO: collect more data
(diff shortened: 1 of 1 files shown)


crates/wasmtime/src/profiling.rs
```diff
@@ -188,10 +188,12 @@ fn module_symbols(name: String, compiled: &CompiledModule) -> Option<LibraryInfo
...
-        match compiled.func_name(func_idx) {
-            None => name = format!("wasm_function_{}", defined_idx.as_u32()),
-            Some(func_name) => demangle_function_name(&mut name, func_name).unwrap(),
-        };
+        demangle_function_name_or_index(
+            &mut name,
+            compiled.func_name(func_idx),
+            defined_idx.as_u32() as usize,
+        )
+        .unwrap();
```
fitzgen, line 196: Could we change the `unwrap` to an `expect` where there is a brief message explaining why we expect this never to fail? Something like

```suggestion
        .expect("only errors if the Writer errors, and string writers never error");
```

fitzgen (approved):
LGTM with minor nitpick below. Thanks!

elliottt:
Would you mind rebasing this PR? Once that's done, I'm happy to click the merge button :+1:

Issue #7666, opened 2023-12-08
One-click guest profiler results


jameysharp:
#### Feature

When invoking the guest profiler via either its `GuestProfiler` Rust API or the Wasmtime CLI, we should provide the option of running a local web server to serve profile data, then generating links to https://profiler.firefox.com/ that direct it to load the profile data directly from the local web server.

#### Benefit

Currently, we write the profile data to a file, then tell the user the path to that file. They then browse to https://profiler.firefox.com/ and use their browser's file-open dialog to find the path we told them. It would be more convenient to give them a single URL which automatically loads the profile data.

Running a local web server also allows us to symbolicate lazily and serve source code to the viewer on-demand, in principle.

#### Implementation

The Samply profiler illustrates how to do this, in https://github.com/mstange/samply/blob/main/samply/src/server.rs.

#### Alternatives

We could continue just writing the profile to a file and letting the user find it in their filesystem.

I was hesitant to suggest doing this in Wasmtime before because it means we need to depend on an HTTP server like `hyper`. I thought it would be better to implement the parts which depend on a web server in embedders such as Viceroy that have their own web server. But now that we've implemented wasi-http, Wasmtime depends on `hyper` anyway, so we might as well do it in Wasmtime where it can be shared across multiple embedders.

PR #7832, opened 2024-01-27
Serve guest-profiler results from a local web server

jameysharp:
This is a proof of concept for #7666, but needs quite a bit of polish before merging.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1227,6 +1227,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4fff74096e71ed47f8e023204cfd0aa1289cd54ae5430a9523be060cdb849964"
 
 [[package]]
+name = "futures-macro"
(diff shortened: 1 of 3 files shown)


jameysharp:
If anyone has suggestions on how to do this better I'd be interested to hear them. In particular, I'm not happy with how many crate dependencies I had to add to the Wasmtime CLI to make this work.

Currently this PR reflects just enough hacking together of pieces to prove that #7666 is possible, so I'm not looking for review on small details, just the big picture design questions.

tschneidereit:
An alternative approach that shouldn't introduce any additional runtime dependencies would be to self-host this. I.e., to embed a wasi-http/proxy component that can load the profile from a pre-opened directory and serve it from the [incoming-handler](https://github.com/WebAssembly/wasi-http/blob/main/wit/proxy.wit#L31).

Such a component should be straightforward to run, and, after stripping, be quite small.

Issue #7735, opened 2023-12-31
Publish C API crate to crates.io


maxbrunsfeld:
#### Feature

It would be useful to me to be able to download `wasmtime-c-api` from crates.io, instead of from GitHub.

#### Benefit

I maintain a C library ([Tree-sitter](https://github.com/tree-sitter/tree-sitter)) that uses Wasmtime via the C API. Tree-sitter also has Rust bindings. The Tree-sitter Rust crate has a Cargo dependency on both `wasmtime` and `wasmtime-c-api` ([previous discussion](https://github.com/bytecodealliance/wasmtime/pull/6765#issuecomment-1740115770)). Right now, those dependencies specify Git revisions, not version numbers, because `wasmtime-c-api` is not available on crates.io.

It would be useful if those two cargo dependencies could both be expressed in terms of *version numbers*, as opposed to `git`/`rev` dependencies. That way, consumers of `tree-sitter` would have more flexibility in terms of which versions of wasmtime were compiled into their binaries.

#### Implementation

It looks like there is automation for publishing various crates to crates.io, in the [`publish.rs` script](https://github.com/bytecodealliance/wasmtime/blob/main/scripts/publish.rs). I think that `wasmtime-c-api` could be added to the list of "public crates" which are published whenever a `v` tag is pushed.

#### Alternatives

* Don't support this use case of combining C and Rust, since it's uncommon. If a C library depends on wasmtime, and that C library has Rust bindings, that Rust crate should not try to use Cargo to build and link Wasmtime.
* Suggest that such crates should download and compile the wasmtime-c-api programmatically in a custom build script, as opposed to using a Cargo dependency.
* Some other person or organization publish the wasmtime C API to crates.io. Don't do it as part of a Wasmtime workflow.

alexcrichton:
I think this is reasonable to support but one tricky bit will be that the name of the crate is `wasmtime-c-api-impl` right now where [`wasmtime-c-api`](https://github.com/bytecodealliance/wasmtime/blob/main/crates/c-api/artifact/Cargo.toml) is taken by the "artifact" which is a `cdylib`. That's what enables `cargo build -p wasmtime-c-api` in this repository to build the dynamic/shared library. This would probably want to be published as `wasmtime-c-api` to crates.io.

Otherwise though for a git dependency you should be able to use `branch = 'release-16.0.0'` for versioned releases, albeit only major versions and not between minor versions.

maxbrunsfeld:
[...]

PR #7837, opened 2024-01-29
Add wasmtime-c-api-impl to the list of crates to publish

maxbrunsfeld:
Prior discussion: https://github.com/bytecodealliance/wasmtime/issues/7735

Having the C API available on crates.io is useful for crates that use C code that use Wasmtime via the C API, allowing those crates themselves to be published to crates.io.

I'm not sure if there are drawbacks to adding this; I just thought I'd open this PR as a concrete suggestion, in case it's uncontroversial.

Currently, the package name is `wasmtime-c-api-impl` (since `wasmtime-c-api` is used by the cdylib). That seems fine to me, but I'm also open to some name change.


amaanq:
Wouldn't this line here need to be removed? I could be totally wrong

https://github.com/bytecodealliance/wasmtime/blob/94b3e84e904b854465a8efff31d9e850f7d68b1e/crates/c-api/Cargo.toml#L10

alexcrichton:
Thanks! I don't have a great suggestion for how to fix the namig problem, and it's probably not worth holding this up for fixing that. 

In addition to @amaanq's suggestion would you be up for helping spruce up the documentation a bit? For example I think it'd be good to have some basic crate documentation explaining what things are (as nothing has rustdoc internally otherwise) and ideally we'd also have a `README.md` explaining a bit about how to use it (as that'll get rendered on crates.io). I also think it'd be good to have a written log about how the crate is used so if we change things in the future we know what to keep working.

maxbrunsfeld:
Ok, I updated the Cargo metadata, enabling publishing and documentation. In order to make it convenient for a downstream crate to include wasmtime's headers, I also added a small `build.rs` that provides [`links` metadata](https://doc.rust-lang.org/cargo/reference/manifest.html#the-links-field). The resulting environment variables worked as expected for me in my downstream crate ([`tree-sitter`](https://github.com/tree-sitter/tree-sitter/pull/2840)) when I pull the crate from GitHub. I think that means that they will work when the source is pulled from crates.io, but it's possible I've made a mistake there.

I added some content to the `c-api` README about how to consume the crate from Rust, in the case where you're binding to a C library that uses wasmtime.

I also added a bit to the module-level doc comment for `c-api`.

amaanq:
ping @alexcrichton @peterhuene, this would be immensely helpful for tree-sitter to have the C API published to crates.io :sweat_smile: since this is a blocker for us to publish the latest release on crates.io

maxbrunsfeld:
I just realized I could test this with `dry-run` publishing, and that the `wasmtime-c-api-macros` would need to be published as well. I think that the most straightforward way to do that would be to move it up to the `crates` folder and add it to the Cargo workspace. I'm going to proceed with doing that on this branch. Let me know if I'm missing something there, and there's a different solution to that macros thing.

alexcrichton (approved):
Looks good to me, thanks! I suspect that the `publish.rs` may need reordering though as well, yeah. If you add `prtest:full` to a commit message in this PR it'll run the full CI if you'd like (before it goes to the merge queue)

Note that most of the Wasmtime team has been at the BA plumbers summit this week and is traveling home today.

maxbrunsfeld:
Ah ok, I'll do that. Thanks for responding while you're all traveling.

alexcrichton:
Ah apologies now that I've gotten back to this it now has a merge conflict. If you don't mind rebasing again I'll add it to the merge queue. Otherwise though I can do the rebase/merge later this week.

maxbrunsfeld:
Thanks so much @alexcrichton.

Just curious - Now that this is landed, would it be easy to publish to crates.io the C-API corresponding to the latest published `wasmtime` version, or do we need to wait until the next Wasmtime release?

alexcrichton:
Doing this for the current release would require a backport and we typically try not to backport features and instead only backport bugfixes. Not impossible but if you're ok I think this would be best to wait.

That being said there's three active branches at the moment. `release-17.0.0` has what was released in January and just today `release-18.0.0` was created to be released in two weeks. This PR landed on `main` which will become `release-19.0.0` which will be released in March. If you'd like I think it's ok to backport this to the 18.0.0 release branch if you'd prefer to have this in a release sooner rather than later.

maxbrunsfeld:
Got it. I would love to backport to 18.0 if possible. Is that something I can help with, via a PR against that branch?

alexcrichton:
Indeed that's all that's required, and please feel free!

maxbrunsfeld:
Ok, PR up: https://github.com/bytecodealliance/wasmtime/pull/7872

PR #7839, opened 2024-01-29
Shuffle some items around in `wasmtime`

alexcrichton:
This is a follow-up to #7766 with some more changes and reorganization. These are some small items here and there which shouldn't have any actual impact on functionality but are intended to reorganize a bit. Changes here include:

* Move component artifact definitions to `wasmtime-environ` as core module ones already live there.
* Rename the module with module artifacts from `instantiate` to `module_artifacts`.
* Make `wasmtime-jit-icache-coherence` an optional dependency as only the `runtime` feature requires it.
* Reorganize serialized metadata for wasmtime ELF files to consolidate everything back into `wasmtime::engine::serialization`. This is to prevent the definition of the serialization format being spread across a few files.
* Touching up the `serialization` module to compile in all builds of the `wasmtime` crate.


crates/wasmtime/src/lib.rs
```diff
@@ -238,6 +238,10 @@
 //! crate. Be sure to check the API you're using to see if any crate features
 //! are enabled.
 //!
+//! * `runtime` - Enabled by default, this features enables executing
```
fitzgen, line 241:
```suggestion
//! * `runtime` - Enabled by default, this feature enables executing
```

fitzgen (approved):

PR #7841, opened 2024-01-29
Enable the `component-model` feature by default in Wasmtime

alexcrichton:
This commit updates the `wasmtime` crate itself to have the `component-model` feature enabled by default. This was also done for the CLI but only for clarity because the `component-model` feature was already eanbled by default transitively through the `serve` feature.


elliottt (approved):
:tada:

PR #7844, opened 2024-01-29
Fix handling of `Tunables` on cross-compiles

alexcrichton:
This commit fixes how Wasmtime handles `Tunables` when targetting non-host platforms (or namely platforms with different pointer widths). Previously the host's `Tunables` would always be used instead of the target's tunables which meant that modules couldn't be loaded on the other platform due to the host having differing tunables by default.

This commit updates tunables in `wasmtime::Config` to all be optional and loading the actual `Tunables` is deferred until the target is known during `Engine`-creation time.


fitzgen (approved):
Nice!

PR #7845, opened 2024-01-30
Replace an explicit implementation of `Service` with `service_fn`

elliottt:
Simplify the implementation of the serve command by removing the explicit Service trait impl, and using `service_fn` instead.


alexcrichton (approved):
Nice!

PR #7855, opened 2024-02-01
Enable some dependency features for wasi-nn.

jianjunz:
This is a follow up of #7766 and #7792. It fixes compile issues for wasmtime-wasi-nn standalone build.


abrown (approved):

abrown:
Thanks, @jianjunz!

PR #7858, opened 2024-02-01
Add cloning for DWARF String attributes

yowl:
This PR (my first here), adds cloning for DWARF `String` attributes, copying how `Block` was done.  This allows `wasmtime -D debug-info` to work with programs compiled with emscripten:

```
emcc 1.c -o 1.wasm -g -gdwarf-5 -sWASM_BIGINT=1 -sSTANDALONE_WASM=1
```
And consequently that they can be debugged in lldb:

image

### DIFF
--- a/crates/cranelift/src/debug/transform/attr.rs
+++ b/crates/cranelift/src/debug/transform/attr.rs
@@ -295,6 +295,7 @@ where
                 pending_di_refs.insert(current_scope_id, attr.name(), offset);
                 continue;
             }
+            AttributeValue::String(d) => write::AttributeValue::String(d.to_slice()?.to_vec()),
             a => bail!("Unexpected attribute: {:?}", a),
         };
         let current_scope = out_unit.get_mut(current_scope_id);


crates/cranelift/src/debug/transform/attr.rs
```diff
@@ -295,6 +295,7 @@ where
                 pending_di_refs.insert(current_scope_id, attr.name(), offset);
                 continue;
             }
+            AttributeValue::String(d) => write::AttributeValue::String(d.to_slice()?.to_vec()),
```
bjorn3, line 298: I think you can use `into_owned()` instead of `to_vec()`.

yowl: Thanks, shall  I change `AttributeValue::Block(d) => write::AttributeValue::Block(d.to_slice()?.to_vec()),` as well?

yowl: I went ahead and changed `Block`, I have limited experience of `Cow` and slicing, but it seemed logical being as how similar `Block` is to `String`

alexcrichton (approved):
I won't really pretend to know what's going on here but I don't think many of us know much about the native dwarf transform this is doing. That being said it seems not obviously wrong at least and the same as everything else in this area.

Ideally we'd have tests for changes like this but we don't have many preexisting tests here yet so I think it's ok to land this as-is and we can look to improve testing later.

Issue #7856, opened 2024-02-01
Cranelift/ISLE/e-graphs: Add better observability into the chain of rewrites that produced the expressions we actually extracted


fitzgen:
In traditional peephole passes, it is easy to add logging for each rewrite you perform on the IR.

In e-graphs, we perform all possible rewrites, producing many equivalent expressions, and then afterwards extract just the best version. This means that logging each rewrite we perform will emit a bunch of largely irrelevant logs for expressions that we didn't ultimately extract. And by the time we do extraction, we no longer know which chain of rewrites produced that expression. This makes debugging a series of rewrites quite difficult.

We should figure out some way to improve things here.

Straw person idea:

* when adding an e-node to an e-class, record in a secondary map:
  * the id of the rewrite that produced that e-node
  * the e-node that was the input to that rewrite
* when printing clif, optionally use this secondary map to add annotation comments like

  ```
  ; Rewrite chain:
  ;   v97 = simplify(v96) rule at remat.isle:12
  ;   v120 = simplify(v97) rule at bitops.isle:283
  ;   v123 = simplify(v120) rule at cprop.isle:84 
  v123 = iadd v42, v36
  ```

We would probably only want to do this when a compile-time cargo feature is enabled. We could enable this feature for the filetests crate, but not be default in `cranelift-codegen`, so we could even write filetests that assert that certain rules we expect to chain together, do in fact chain together.

I think there were plans to add names to rules (maybe support for this even already landed in ISLE? can't remember) and we could use these names in the comments if they exist. That would be something that we could test for in the filetest directives, since we wouldn't want to include the source locations, as those would be too noisy for filetests when adding, removing, and tweaking rules.

But anyways, the filetest support is ultimately a secondary priority. I really just want to have better debugging of the e-graph rewrites.

cc @elliottt

avanhatt:
FWIW, egg's ["explanation"](https://egraphs-good.github.io/egg/egg/tutorials/_03_explanations/index.html) construct (and [the paper they reference](https://www.cs.upc.edu/~roberto/papers/rta05.pdf)) might be a useful place to look for inspiration.

avanhatt:
We also have code to name ISLE rules in the Crocus codebase, I can try to split that off as a separate PR soon.

PR #7859, opened 2024-02-02
Cranelift: Use a fixpoint loop to compute the best value for each eclass

### LINKED ISSUE #7857: Block order and value number affects whether we get valid CLIF after optimizations
Last night we got a [fuzz bug](https://oss-fuzz.com/testcase-detail/4583302363152384). Everything below is relative to Wasmtime at 0d662c9d3dd1db8490e30124c9126bfa00a9a5c6.

This input:

```wasm
(module
  (func
    (local f32)
    f32.const 100
    f32.sqrt
    i32.const 0
    if
      f32.const 100
      f32.sqrt
      block
        i32.const 1
        br_if 0
        f32.const 0
        local.set 0
      end
      local.get 0
      i32.const 1
      select
      i32.reinterpret_f32
      global.set 0
    end
    i32.reinterpret_f32
    global.set 0
  )
  (global (;0;) (mut i32) i32.const 0)
)
```

will panic in regalloc

```
$ cargo run compile -C cache=n -W nan-canonicalization ./foo.wat
...
thread '<unnamed>' panicked at cranelift/codegen/src/machinst/compile.rs:76:14:
register allocation: SSA(VReg(vreg = 198, class = Float), Inst(33))
```

This surfaces a validation error in CLIF earlier with validation enabled

```
$ cargo run compile -C cache=n -C cranelift-debug-verifier -W nan-canonicalization ./foo.wat
...
                                    v19 = f32const +NaN
                                    v4 = select v18, v19, v17  ; v19 = +NaN
@0043                               v12 = select v8, v4, v10  ; v8 = 1
;~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
; error: inst12 (v12 = select.f32 v8, v4, v10  ; v8 = 1): uses value arg from non-dominating block4

@0044                               v13 = bitcast.i32 v12
@0049                               store notrap aligned table v13, v0+80
@004b                               jump block1

                                block1:
@004b                               return
}

; 1 verifier error detected (see above). Compilation aborted.
```

This has been further reduced to this CLIF test case:

```
test optimize
set enable_verifier=true
set opt_level=speed
target x86_64

function %foo(i64) {
block0(v0: i64):
    v3 = f32const 0x1.900000p6
    v17 = sqrt v3
    v18 = fcmp ne v17, v17
    v19 = f32const +NaN
    v4 = select v18, v19, v17
    v5 = iconst.i32 0
    brif v5, block2, block3

block2:
    v6 = f32const 0x1.900000p6
    v20 = sqrt v6
    v21 = fcmp ne v20, v20
    v22 = f32const +NaN
    v7 = select v21, v22, v20
    v8 = iconst.i32 1
    v2 = f32const 0.0
    brif v8, block4(v2), block5

block5:
    v9 = f32const 0.0
    jump block4(v9)

block4(v10: f32):
    v11 = iconst.i32 1
    v12 = select.f32 v11, v7, v10
    v13 = bitcast.i32 v12
    store notrap aligned table v13, v0+80
    return

block3:
    v15 = bitcast.i32 v4
    store notrap aligned table v15, v0+80
    return
}
```

which can be reproduced with:

```
$ cd cranelift && cargo run test ./foo.clif
```

I've been investigating this with @elliottt and @fitzgen in person for a bit now. So far we have concluded a few "fixes" can be applied:

* One fix is to renumber the original `v12` input to `v1` in CLIF.
* Another fix is to move `block3` to be beneath `block0`.
* The final fix is to change [this line](https://github.com/bytecodealliance/wasmtime/blob/b6a8abc6ce31cf6db7ad5edb9b2567620190ce35/cranelift/codegen/src/opts/cprop.isle#L172) to `(subsume x)`

Naturally none of these are actual fixes but are symptoms of the "real" issue. We're still figuring things out at this time but I wanted to open this up.

Trevor and Nick are telling me as well that this is possibly related to https://github.com/bytecodealliance/wasmtime/issues/6126.

fitzgen:
Fixes #7857

### DIFF
--- a/cranelift/codegen/src/egraph/cost.rs
+++ b/cranelift/codegen/src/egraph/cost.rs
@@ -178,6 +178,7 @@ mod tests {
         let a = Cost::new(5, 2);
         let b = Cost::new(37, 3);
         assert_eq!(a + b, Cost::new(42, 3));
+        assert_eq!(b + a, Cost::new(42, 3));
     }
 
     #[test]
@@ -185,6 +186,7 @@ mod tests {
         let a = Cost::new(5, 2);
         let b = Cost::infinity();
         assert_eq!(a + b, Cost::infinity());
+        assert_eq!(b + a, Cost::infinity());
     }
 
     #[test]
@@ -192,6 +194,7 @@ mod tests {
         let a = Cost::new(Cost::MAX_OP_COST - 10, 2);
         let b = Cost::new(11, 2);
         assert_eq!(a + b, Cost::infinity());
+        assert_eq!(b + a, Cost::infinity());
     }
 
     #[test]
@@ -200,6 +203,10 @@ mod tests {
         let b = Cost::new(10, 1);
         assert_eq!(
             Cost::of_pure_op(Opcode::Iconst, [a, b]),
+            Cost::new(21, u8::MAX)
+        );
+        assert_eq!(
+            Cost::of_pure_op(Opcode::Iconst, [b, a]),
             Cost::new(21, u8::MAX)
         );
     }
--- a/cranelift/codegen/src/egraph/cost.rs
+++ b/cranelift/codegen/src/egraph/cost.rs
@@ -86,14 +86,6 @@ impl Cost {
         Cost(0)
     }
 
-    pub(crate) fn is_infinite(&self) -> bool {
-        *self == Cost::infinity()
-    }
-
-    pub(crate) fn is_finite(&self) -> bool {
-        !self.is_infinite()
-    }
-
     /// Construct a new `Cost` from the given parts.
     ///
     /// If the opcode cost is greater than or equal to the maximum representable
--- a/cranelift/codegen/src/egraph/cost.rs
+++ b/cranelift/codegen/src/egraph/cost.rs
@@ -74,7 +74,7 @@ impl Cost {
     const DEPTH_BITS: u8 = 8;
     const DEPTH_MASK: u32 = (1 << Self::DEPTH_BITS) - 1;
     const OP_COST_MASK: u32 = !Self::DEPTH_MASK;
-    const MAX_OP_COST: u32 = (Self::OP_COST_MASK >> Self::DEPTH_BITS) - 1;
+    const MAX_OP_COST: u32 = Self::OP_COST_MASK >> Self::DEPTH_BITS;
 
     pub(crate) fn infinity() -> Cost {
         // 2^32 - 1 is, uh, pretty close to infinite... (we use `Cost`
@@ -86,14 +86,24 @@ impl Cost {
         Cost(0)
     }
 
-    /// Construct a new finite cost from the given parts.
+    pub(crate) fn is_infinite(&self) -> bool {
+        *self == Cost::infinity()
+    }
+
+    pub(crate) fn is_finite(&self) -> bool {
+        !self.is_infinite()
+    }
+
+    /// Construct a new `Cost` from the given parts.
     ///
-    /// The opcode cost is clamped to the maximum value representable.
-    fn new_finite(opcode_cost: u32, depth: u8) -> Cost {
-        let opcode_cost = std::cmp::min(opcode_cost, Self::MAX_OP_COST);
-        let cost = Cost((opcode_cost << Self::DEPTH_BITS) | u32::from(depth));
-        debug_assert_ne!(cost, Cost::infinity());
-        cost
+    /// If the opcode cost is greater than or equal to the maximum representable
+    /// opcode cost, then the resulting `Cost` saturates to infinity.
+    fn new(opcode_cost: u32, depth: u8) -> Cost {
+        if opcode_cost >= Self::MAX_OP_COST {
+            Self::infinity()
+        } else {
+            Cost(opcode_cost << Self::DEPTH_BITS | u32::from(depth))
+        }
     }
 
     fn depth(&self) -> u8 {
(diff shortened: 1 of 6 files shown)


cranelift/codegen/src/egraph/elaborate.rs
```diff
@@ -606,7 +653,13 @@ impl<'a> Elaborator<'a> {
                         }
                         inst
                     };
+
                     // Place the inst just before `before`.
+                    debug_assert!(
+                        is_pure_for_egraph(self.func, inst),
```
elliottt, line 659: Really glad you added this one!

elliottt (approved):
This looks great!

cranelift/codegen/src/egraph/elaborate.rs
```diff
@@ -216,46 +217,92 @@ impl<'a> Elaborator<'a> {
...
+                            );
+                            best[value] = BestEntry(cost, value);
+                            trace!(" -> cost of value {} = {:?}", value, cost);
+                        }
+                    }
+                };
+
+                // Keep on iterating the fixpoint loop while we are finding new
+                // best values.
+                keep_going |= orig_best_value != best[value];
```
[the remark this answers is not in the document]
elliottt: Though the change I suggested hinges on all values definitely becoming non-infinite, so I think looking for the change in best value is better. Never mind!

cranelift/codegen/src/egraph/elaborate.rs
```diff
@@ -216,46 +217,92 @@ impl<'a> Elaborator<'a> {
...
+            keep_going = false;
+            trace!(
+                "fixpoint iteration {}",
+                self.stats.elaborate_best_cost_fixpoint_iters
+            );
+            self.stats.elaborate_best_cost_fixpoint_iters += 1;
+
+            for (value, def) in self.func.dfg.values_and_defs() {
+                // If the cost of this value is finite, then we've already found
+                // its final cost.
```
cfallin, line 245: Sorry for the drive-by from the sidelines, just a possible clarification request here though after thinking about cost updates during a long drive today:

It's not immediately obvious to me why this (once finite, then final) property is the case; I'm curious what reasoning y'all have gone through on this and/or what you've observed? I think a node's cost can continue to decrease as we discover more finite costs (consider a union node: `min(20, infinity) == 20` in first pass, `min(20, 10) == 10` in second pass; then another node that uses that as an arg). Or is there an argument we can make why this shouldn't happen in practice?

elliottt: This is a great point Chris! When @fitzgen and I were discussing the fixpoint change yesterday, we reasoned that it was okay to skip finite values because we were assuming two things:

* We would remove the behavior where `Cost` addition would saturate to `MAX_COST`, not `infinity()`
* As we can't produce cycles, a fixpoint would cause everything to eventually settle out to finite cost

As you pointed out, the flaw with this reasoning is that the handling of `Union` values will not behave this way, instead preferring finite values to infinite.

Since addition now saturates to infinity which will ensure that `Result` nodes don't appear finite until all their dependencies have been processed, what do you think about only computing the min if both arguments to a `Union` are finite? I think that change would make more concrete our use of the `infinity()` cost: it's a marker for where all the arguments have not yet been processed.

elliottt: Good point Nick, sorry for muddying the waters there.

fitzgen: Okay actually I was wrong, thanks Trevor for asking very pointed questions in private chat :-p

The union's operand values are always defined before the union, but if one of *those* operand values is a funky one where *its* operands are out of order, then the operand could still be infinite by the time we get to the union, and then the union's `min` would drop the infinite. That would be a finite cost that is potentially not in its final form, depending on the cost we still need to compute for the still-infinite operand.

So this "optimization" of early-continuing was not correct! Bullet dodged.

This ae-graphs code is all very subtle, and we should spend some time thinking about what we can do to make things more obviously correct, even if it is just adding additional debug asserts and comments. It shouldn't take 3.5 engineers who are all intimately familiar with Cranelift a full day to diagnose and fix this kind of bug and *still* introduce subtle flaws in the fix.

fitzgen: And for clarity: since we are doing the "full" fixpoint now, even if we "drop" an operand's infinite cost via `min` in one iteration of the loop, we will consider that operand's value again on the next iteration of the fix point, and eventually, as the fixpoint is reached, we will have the correct costs for everything.

fitzgen:
(Re-adding to merge queue after misunderstanding regarding https://github.com/bytecodealliance/wasmtime/pull/7859#discussion_r1476296142)

cranelift/codegen/src/egraph/elaborate.rs
fitzgen: Agreed, and not trying to point fingers or anything, just trying to improve the situation for everyone. I think something like https://github.com/bytecodealliance/wasmtime/issues/7856 would help a lot too.

alexcrichton:
Given that the same riscv64 failure happened twice in a row my guess is that it's probably a deterministic failure rather than a spurious failure. That may mean that a preexisting riscv64 lowering rule is buggy and this is starting to expose that. I'll note though that I haven't attempted to reproduce locally yet.

fitzgen:
Thanks Afonso!

cranelift/codegen/src/egraph/cost.rs
```diff
@@ -163,9 +170,45 @@ fn pure_op_cost(op: Opcode) -> Cost {
...
+        let a = Cost::new(5, 2);
+        let b = Cost::new(37, 3);
+        assert_eq!(a + b, Cost::new(42, 3));
+    }
+
+    #[test]
+    fn add_infinity() {
+        let a = Cost::new(5, 2);
+        let b = Cost::infinity();
+        assert_eq!(a + b, Cost::infinity());
```
jameysharp, line 187: Can I suggest adding the same assertion with the operands swapped? `b+a` should also be infinity.

jameysharp (approved):
Looks good to me! Just one optional suggestion.

cranelift/codegen/src/egraph/cost.rs
fitzgen: Good idea! Done.

PR #7862, opened 2024-02-02
Update and add lots of docs, mostly for components

alexcrichton:
When I originally implemented the embedding API for the component model I left a lot of `// TODO: say more here` style comments assuming I'd come back around before stabilization. While I didn't quite make it in time for Wasmtime 17 this is my attempt to improve things for the future. This should add substantive documentation for all component-model related things and flesh out some more details here and there. Nowhere near comprehensive but, hey, Rome wasn't built in a day either.


crates/wasmtime/src/runtime/component/component.rs
```diff
@@ -20,8 +20,36 @@ use wasmtime_runtime::{
...
+/// across threads. The compilation model of a component is the same as that of
+/// [a module](crate::Module) which is to say:
+///
+/// * Compilation happens synchronously during [`Component::new`].
+/// * The result of compilation can be saved into storage with
+///   [`Component::serialize`].
+/// * A previously compiled artifact can be parsed with
+///   [`Component::deserialize`].
+/// * No compilation happens at runtime for a component, everything is done
+///   after [`Component::new`] returns.
```
jeffparsons, line 41: I initially read this as (what I assume is) the opposite of what it means. How about:

```suggestion
/// * No compilation happens at runtime for a component — everything is done
///   by the time [`Component::new`] returns.
```

crates/wasmtime/src/runtime/component/component.rs
```diff
@@ -53,10 +81,64 @@ pub(crate) struct AllCallFuncPointers {
...
-    // FIXME: need to write more docs here.
+    ///
+    /// The `bytes` provided can either be the binary or text format of a
+    /// [WebAssembly component]. Note that the text format requires the `wat`
+    /// feature of this crate to be enabled. This API does not support
+    /// streaming compilation.
+    ///
+    /// This function will synchronously validate the entire component,
+    /// including all core modules, and then compile all components, modules,
+    /// etc, found within the provided bytes.
```
jeffparsons, line 94: Nit:

```suggestion
    /// etc., found within the provided bytes.
```

crates/wasmtime/src/config.rs
```diff
@@ -28,7 +28,9 @@ use wasmtime_fiber::RuntimeFiberStackCreator;
 
 pub use wasmtime_environ::CacheStore;
 #[cfg(feature = "pooling-allocator")]
-pub use wasmtime_runtime::{mpk, MpkEnabled};
+use wasmtime_runtime::mpk;
+#[cfg(feature = "pooling-allocator")]
+pub use wasmtime_runtime::MpkEnabled;
```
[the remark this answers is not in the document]
alexcrichton: While I was browsing documentation I realized that the `mpk` module was reexported to the crate root (e.g. `wasmtime::mpk`) as opposed to just `MpkEnabled`. Only the `MpkEnabled` type should be exported and the `mpk` module is intended to be private to the crate, hence the split. (I found it a bit unfortunate as well...)

PR #7866, opened 2024-02-03
x64: Add `{u,s}mulhi.i8` instruction support

afonso360:
👋 Hey,

This PR adds support for `smulhi.i8` and `umulhi.i8` to the x64 backend.

Closes #5468
Closes #7865

Issue #7865, opened 2024-02-03
Cranelift: Simple multiplication + out of bounds shift panics with "should be implemented in ISLE: inst = v12 = smulhi.i8 v0, v0, type = Some(types::I8)"


bjorn3:
### `.clif` Test Case

```
test compile
set opt_level=speed
target x86_64

function u0:9(i8) -> i16 system_v {
block0(v0: i8):
    v8 = sextend.i16 v0
    v9 = imul v8, v8
    v10 = iconst.i64 1000
    v13 = sshr v9, v10  ; v10 = 1000
    return v13
}
```

### Steps to Reproduce

* Compile the above test case.

### Expected Results

Compiles fine and masks the shift amount as expected.

### Actual Results

Panics with "should be implemented in ISLE: inst = v12 = smulhi.i8 v0, v0, type = Some(types::I8)"

### Versions and Environment

Cranelift version or commit: Tested with 0.104 and b6a8abc6ce31cf6db7ad5edb9b2567620190ce35.

Operating system: N/A

Architecture: x86_64

### Extra Info

Reported to cg_clif by @cbeuw in https://github.com/rust-lang/rustc_codegen_cranelift/issues/1455.

PR #7871, opened 2024-02-05
x64: Refactor multiplication instructions

alexcrichton:
This commit is inspired after reading over some code from #7865 and #7866. The goal of this commit was to refactor scalar multiplication-related instructions in the x64 backend to more closely align with their native instructions. Changes include:

* The `MulHi` instruction is renamed to `Mul`. This represents either `mul` or `imul` producing a doublewide result.
* A `Mul8` instruction was added to correspond to `Mul` for the 8-bit variants that produce a doublewide result in the `AX` register rather than the other instructions which split between `RAX` and `RDX`.
* The `UMulLo` instruction was removed as now it's covered by `Mul`
* The `AluRmiROpcode::Mul` opcode was removed in favor of new `IMul` and `IMulImm` instructions. Register allocation and emission already had special cases for `Mul` which felt better as standalone instructions rather than putting in an existing variant.

Lowerings using `imul` are not affected in general but the `IMulImm` instruction has different register allocation behavior than before which allows the destination to have a different register than the first operand. The `umulhi` and `smulhi` instructions are also reimplemented with their 8-bit variants instead of extension-plus-16-bit variants.


fitzgen (approved):
Nice!

alexcrichton:
cc @abrown you mentioned that you were also taking a look at this, want me to hold off on merging for that?

abrown:
Nah, it's good to go now that @fitzgen looked at it.

PR #7873, opened 2024-02-05
Accept explicit CPU delta time in GuestProfiler

Milek7:
Currently `GuestProfiler::sample` always records zero CPU time delta. This makes resulting profiles hard to interpret, especially because sampling is usually performed from epoch callback, which necessarily means that no samples at all are captured when guest is not running, thus there is no distinguishable "idle" stacktrace of any sort.

While it is possible to read OS thread time counters for that, GuestProfiler is not aware whether current thread is used exclusively for running that guest. For example, I'm running multiple guests on single-threaded event loop and that would partially misattribute CPU time to neighboring guests on first/last sample. Thus I think that the embedder is better suited to measure and provide time delta to sampling function.

I left old signature intact not to break any existing users. For symmetry C bindings also got new signature, though maybe old one should be removed as it doesn't have any users yet?

### DIFF
--- a/crates/c-api/include/wasmtime/profiling.h
+++ b/crates/c-api/include/wasmtime/profiling.h
@@ -85,6 +85,23 @@ wasmtime_guestprofiler_sample(wasmtime_guestprofiler_t *guestprofiler,
                               const wasmtime_store_t *store);
 
 /**
+ * \brief Add a sample to the profile, additionally specifying used CPU time.
+ *
+ * \param guestprofiler the profiler the sample is being added to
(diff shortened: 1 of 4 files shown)


alexcrichton:
Thanks for the PR! It's ok to break the current signature and change it, so I think I would agree with you that it seems like it might be best to unconditionally take a `Duration` argument. Could you update the documentation to indicate that passing `Duration::ZERO` is ok but might have some consequences when viewing the results?

alexcrichton (approved):

alexcrichton (approved):

Issue #7829, opened 2024-01-26
`utime()` does not accept a file descriptor under preview 2 (as found by Python)


brettcannon:
### Test Case

TODO: upload Wasm file here

### Steps to Reproduce

* https://github.com/python/cpython/blob/07236f5b39a2e534cf190cd4f7c73300d209520b/Lib/test/test_posix.py#L452-L454

### Expected Results

Test should pass (it does under preview 1).

### Actual Results

Raises errno 63: [`ENOSR`](https://docs.python.org/3/library/errno.html#errno.ENOSR) (out of stream resources).

### Versions and Environment

Wasmtime version or commit: 17.0.0

Operating system: Linux/WSL

Architecture: x64

### Extra Info

Anything else you'd like to add?

Success logs under preview 1: 
[success-utime-fd.txt](https://github.com/bytecodealliance/wasmtime/files/14069110/success-utime-fd.txt)

Failure logs under preview 2:
[failure-utime-fd.txt](https://github.com/bytecodealliance/wasmtime/files/14069114/failure-utime-fd.txt)

alexcrichton:
Could you perhaps reupload the failure logs? It currently seems to be empty and/or says success with no extra logs

brettcannon:
[failure-utime-fd.txt](https://github.com/bytecodealliance/wasmtime/files/14070134/failure-utime-fd.txt)


Sorry about that!

pchickey:
@elliottt and I investigated this and will be posting a fix shortly.

brettcannon:
> I don't know how to run your tests locally to verify that this fixed it

FYI I'm working on better docs for this, but https://github.com/python/cpython/tree/main/Tools/wasm#wasi-wasm32-wasi are the current docs if you're interested.



> Please verify for us whether this is a fix when you have time

Will do! I'll plan to give it a try this week.



> if it changes the status of any other of your posix tests, which exercise a lot more than our test suite.

Yeah, Python's test suite is old and extensive for this sort of thing: we're great at exercising C compilers and OSs. 😁 If you're ever interested in trying to get the Python test suite set up as a 3rd-party test suite to run, just let me know.

brettcannon:
Commented at https://github.com/bytecodealliance/wasmtime/pull/7876#issuecomment-1936792352 about the PR passing CPython's test suite.

pchickey:
[...]

PR #7876, opened 2024-02-06
Preview 2 filesystem: track open_mode instead of reporting the permissions

pchickey:
First update the fd_filestat_set test to assert that the filestat_set operations on a file descriptor behaves the same whether opened readonly or readwrite. This test now reproduces the behavior reported in https://github.com/bytecodealliance/wasmtime/issues/7829. wasi-common passes this changed test, but wasmtime-wasi::preview2 does not, showing that this is a regression between the two implementations.

Then, in preview2 implementation, track open_mode for a File and Dir, and use that instead of reporting permissions in DescriptorFlags.

FilePerms/DirPerms is a way of having wasmtime-wasi, instead of the operating system, mediate permissions on a preview2 Descriptor. This was conflated with the open mode, which should determine whether DescriptorFlags contains READ or WRITE or MUTATE_DIRECTORY.

* don't mask FilePerms with the DescriptorFlags (oflags) in path_open
* File and Dir now track their open_mode (represented using FilePerms)
  and report them in DescriptorFlags

### DIFF
--- a/crates/test-programs/src/bin/preview1_fd_filestat_set.rs
+++ b/crates/test-programs/src/bin/preview1_fd_filestat_set.rs
@@ -132,14 +132,18 @@ fn main() {
     };
 
     // Run the tests.
-    //
+
     unsafe { test_fd_filestat_set_size_rw(dir_fd) }
     unsafe { test_fd_filestat_set_size_ro(dir_fd) }
-    //
+
     // The fd_filestat_set_times function should behave the same whether the file
     // descriptor is open for read-only or read-write, because the underlying
     // permissions of the file determine whether or not the filestat can be
     // set or not, not than the open mode.
-    unsafe { test_fd_filestat_set_times(dir_fd, wasi::RIGHTS_FD_READ) }
+    if test_programs::preview1::config().support_dangling_filesystem() {
+        // Guarding to run on non-windows filesystems. Windows rejects set-times on read-only
+        // files.
+        unsafe { test_fd_filestat_set_times(dir_fd, wasi::RIGHTS_FD_READ) }
+    }
     unsafe { test_fd_filestat_set_times(dir_fd, wasi::RIGHTS_FD_READ | wasi::RIGHTS_FD_WRITE) }
 }
--- a/crates/test-programs/src/bin/preview1_fd_filestat_set.rs
+++ b/crates/test-programs/src/bin/preview1_fd_filestat_set.rs
@@ -1,9 +1,10 @@
 use std::{env, process, time::Duration};
-use test_programs::preview1::{assert_fs_time_eq, open_scratch_directory, TestConfig};
+use test_programs::preview1::{
+    assert_errno, assert_fs_time_eq, open_scratch_directory, TestConfig,
+};
 
-unsafe fn test_fd_filestat_set(dir_fd: wasi::Fd) {
-    let cfg = TestConfig::from_env();
-    // Create a file in the scratch directory.
+unsafe fn test_fd_filestat_set_size_rw(dir_fd: wasi::Fd) {
+    // Create a file in the scratch directory, opened read/write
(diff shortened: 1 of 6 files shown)


crates/test-programs/src/bin/preview1_fd_filestat_set.rs
```diff
@@ -81,5 +73,11 @@ fn main() {
     };
 
     // Run the tests.
-    unsafe { test_fd_filestat_set(dir_fd) }
+    //
+    // The filestat_set functions should behave the same whether the file
+    // descriptor is open for read-only or read-write, because the underlying
+    // permissions of the file determine whether or not the filestat can be
+    // set or not, not than the open mode.
+    unsafe { test_fd_filestat_set(dir_fd, wasi::RIGHTS_FD_READ) }
```
sunfishcode, line 81: This doesn't seem right. If we open the file with just `RIGHTS_FD_READ`, we shuldn't be able to do `fd_filestat_set_size` on it. That would be analogous in POSIX to opening a file with `FD_READ` and calling `ftruncate`, which is defined to fail.

pchickey: That was what I thought too but I didnt chase it down in the ftruncate case. From the python test suite I found that futime is expected to succeed even when a file is read-only. So, I'll have to change this test, as well as set-len to check the open mode and reject size changes when not open for writing (fail with EINVAL, which I prefer here to EBADF) but permit set-times.

pchickey: This problem ended up being down to how the test was written. wasi::OFLAGS_CREAT implies that the WASI impl open the file with the O_RRONLY or O_RDWR in posix, so we were never actually opening a file just for reading in this test. I changed the test to create the file, then open it under the requested mode, and it behaves as expected.

pchickey:
I believe the test now shows that the implementations all behave as expected.

crates/wasi/src/preview2/ctx.rs
```diff
@@ -144,8 +144,17 @@ impl WasiCtxBuilder {
         file_perms: FilePerms,
         path: impl AsRef<str>,
     ) -> &mut Self {
-        self.preopens
-            .push((Dir::new(dir, perms, file_perms), path.as_ref().to_owned()));
+        let mut open_flags = FilePerms::empty();
```
alexcrichton, line 147: While you're here mind adding documentation to this method? I forget how `perms` relates to `file_perms` myself

crates/wasi/src/preview2/filesystem.rs
```diff
@@ -109,14 +111,21 @@ pub struct Dir {
     pub dir: Arc<cap_std::fs::Dir>,
     pub perms: DirPerms,
     pub file_perms: FilePerms,
+    pub open_mode: FilePerms,
```
alexcrichton, line 114: Mind adding some brief comments here about what these fields are? Three sets of permissions seems like a good threshold for needing disambiguation

alexcrichton (approved):
Seems reasonable to me but I'm finding it difficult to follow what each of the sets of permissions are so I'm not 100% sure. Could you expand a bit on how this fixes https://github.com/bytecodealliance/wasmtime/issues/7829? I'm having a tough time connecting permissions and how that'd fail something fd-based vs path-based.

Also https://github.com/bytecodealliance/wasmtime/issues/7829 mentions errno 63 and ENOSR but I suspect that was something Python-specific or something like that?

crates/wasi/src/preview2/ctx.rs
pchickey: Yep improved and documented this :)

brettcannon:
I just finished running CPython's test suite and this fixed the issue I found and didn't cause any new ones!

PR #7879, opened 2024-02-06
Add missing subsume uses in egraph rules

elliottt:
I took a quick pass over the egraph rules to look for cases that violate our new guidelines about when `subsume` is required. There were a few cases where `subsume` was needed, and though I believe those instances were benign I think they're worth adding to make sure we're following the spirit of our new guidelines.

I've also added some clarifying text about constants in the text of guideline 2, to make it more obvious that cases where we discard constants are okay without `subsume`.


fitzgen (approved):
Thanks for auditing our rules!

PR #7882, opened 2024-02-07
Guard recursion in `will_simplify_with_ireduce`

### LINKED ISSUE #7874: wasmtime 17 hangs while trying to load wasm file
### Test Case

https://drive.google.com/file/d/1HYlz78KTuag9bTpNwGP_blp5yWo95i_7/view?usp=sharing

This wasm file successfully loads by version 16, but wasmtime 17 hangs and eats a lot of CPU while trying to load it

### Steps to Reproduce

This is reproducible in python and ruby. Example ruby code:

```
      @engine = Wasmtime::Engine.new
      @mod = Wasmtime::Module.new(@engine, wasm)
``` 

### Expected Results

Should work

### Actual Results

It hangs eating ~170% of CPU

### Versions and Environment

Wasmtime version or commit: 17

Operating system: MacOS

Architecture: aarch64

elliottt:
Expose the underlying issue of #7874 by adding a test with a long chain of additions, and then fix the bug by bounding the recursion of `will_simplify_with_ireduce`.

Reading through the changes in #7719, it seemed like the recursive cases in `will_simplify_with_ireduce` were missing two things:
1. an additional guard of `reducible_modular_op` before the recursion, which showed up in the entry-point in `simplify`
2. a depth bound to ensure that the recursion doesn't get out of control.

This PR makes both of these changes, and resolves the compilation time issues reported in #7874. The depth bound was arbitrary, and it would be fine to increase it if necessary.

Fixes #7874

### DIFF
--- a/cranelift/codegen/src/opts/extends.isle
+++ b/cranelift/codegen/src/opts/extends.isle
@@ -77,17 +77,25 @@
 
 ;; Matches values where `ireducing` them will not actually introduce another
 ;; instruction, since other rules will collapse them with the reduction.
+(decl pure multi will_simplify_with_ireduce_rec (u8 Value) Value)
+(rule (will_simplify_with_ireduce_rec _ x@(uextend _ _)) x)
+(rule (will_simplify_with_ireduce_rec _ x@(sextend _ _)) x)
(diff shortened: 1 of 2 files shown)


fitzgen (approved):
Looks great!

PR #7884, opened 2024-02-07
Allocate thread IDs for wasi_thread_spawn sequentially

Milek7:
TID returned by `wasi_thread_spawn` must be non-zero and upper three bits must always be zero: https://github.com/WebAssembly/wasi-threads?tab=readme-ov-file#design-choice-thread-ids

Current implementation didn't fulfill this requirements, only clearing the top bit from random value. This breaks wasi-libc mutexes when random TID is larger or equal to 0x3FFFFFFF.

Additionally, from birthday paradox probability of collision in allowed TID range nears 50% after spawning 27000 threads. This is large but not totally impossible, thus replace random generation with sequential allocation.

### DIFF
--- a/crates/wasi-threads/src/lib.rs
+++ b/crates/wasi-threads/src/lib.rs
@@ -3,8 +3,8 @@
 //! [`wasi-threads`]: https://github.com/WebAssembly/wasi-threads
 
 use anyhow::{anyhow, Result};
-use rand::Rng;
 use std::panic::{catch_unwind, AssertUnwindSafe};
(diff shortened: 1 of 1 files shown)


Milek7:
This PR makes 0x1FFFFFFF the last TID that will be returned, but it is incorrect on wasm-libc because it reserves that for main thread ID. But wasi-threads proposal as written allows for this, so I'm not sure whether assume that's wasm-libc problem or limit it there 0x1FFFFFFE.

abrown (approved):

PR #7888, opened 2024-02-07
Support finishing component types without compilation artifacts

macovedj:
As referenced in [this PR](https://github.com/bytecodealliance/jco/pull/368), the [addition of component type reflection support](https://github.com/bytecodealliance/wasmtime/pull/7804) breaks jco's current dependence on `wasmtime-environ`, and since jco doesn't leverage an `Engine`, I'm not immediately aware of a good way to adapt jco.  So this PR just does the simple thing and reintroduces the old `finish` function with the new name `finish_sans_reflection`, right next to the one that was updated for reflection support, though if there's a way I'm missing to adapt jco, then this could be reconsidered.

### DIFF
--- a/crates/environ/src/component/types.rs
+++ b/crates/environ/src/component/types.rs
@@ -438,6 +438,13 @@ impl ComponentTypesBuilder {
     }
 
     /// Finishes this list of component types and returns the finished
+    /// structure.
+    pub fn finish_sans_reflection(mut self) -> ComponentTypes {
+        self.component_types.module_types = self.module_types.finish();
+        self.component_types
+    }
+
+    /// Finishes this list of component types and returns the finished
     /// structure and the [`TypeComponentIndex`] corresponding to top-level component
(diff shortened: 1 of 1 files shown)


guybedford (approved):

guybedford:
Can confirm this gets the tests in the JCO upgrade PR to pass on https://github.com/bytecodealliance/jco/pull/368.

alexcrichton:
Thanks! Would it be possible to call `finish` as-is though? I think passing an empty `PrimaryMap` and empty iterators should preserve the same behavior for jco?

macovedj:
> Thanks! Would it be possible to call `finish` as-is though? I think passing an empty `PrimaryMap` and empty iterators should preserve the same behavior for jco?

Yeah it seems you're right. Just pushed to the jco branch and have it pointing at `wasmtime` main with empty `PrimaryMap` and iterators.

PR #7889, opened 2024-02-07
Backport egraph fixes to 17.0.1

elliottt:
Backport recent fixes to the 17.0.0 branch for a 17.0.1 patch release.

- Guard recursion in `will_simplify_with_ireduce` (#7882)
- Cranelift: Use a fixpoint loop to compute the best value for each eclass (#7859)
- Add missing subsume uses in egraph rules (#7879)

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,5 +1,20 @@
 --------------------------------------------------------------------------------
 
+## 17.0.1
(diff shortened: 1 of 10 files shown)


fitzgen (approved):
Thanks!

PR #7895, opened 2024-02-08
wasi-sockets: (TCP) Use tokio's built in methods to perform the state changes

badeend:
Full discussion: https://bytecodealliance.zulipchat.com/#narrow/stream/217126-wasmtime/topic/tokio.20always.20reports.20readiness

TLDR: Our implementation was flawed. It called `tokio::net::TcpStream::try_from(...)` directly after creating the socket. That triggers Tokio to call `epoll_ctl` to add it to the poll list. Linux sees that the socket is not connected (yet) and emits EPOLLHUP. Tokio interprets that to be a "final" state and never recovers from it. All future attempts to await I/O readiness immediately succeed.

This PR moves away from rustix and uses Tokio's own methods for constructing, binding, connecting, listening & accepting sockets. It still uses rustix for socket options

### DIFF
--- a/crates/wasi/src/preview2/tcp.rs
+++ b/crates/wasi/src/preview2/tcp.rs
@@ -1,5 +1,7 @@
 use super::network::SocketAddressFamily;
-use super::{with_ambient_tokio_runtime, HostInputStream, HostOutputStream, SocketResult, StreamError};
+use super::{
+    with_ambient_tokio_runtime, HostInputStream, HostOutputStream, SocketResult, StreamError,
+};
 use crate::preview2::{AbortOnDropJoinHandle, Subscribe};
 use anyhow::{Error, Result};
 use cap_net_ext::AddressFamily;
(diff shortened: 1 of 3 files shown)


crates/wasi/src/preview2/tcp.rs
```diff
@@ -299,37 +293,65 @@ impl TcpSocket {
...
+            | TcpState::Connecting(..)
+            | TcpState::ConnectReady(..)
+            | TcpState::Closed => Err(ErrorCode::InvalidState.into()),
+        }
     }
 
-    pub fn tcp_socket(&self) -> &tokio::net::TcpStream {
-        &self.inner
+    pub(crate) fn connect(socket: tokio::net::TcpSocket, addr: std::net::SocketAddr) -> impl Future<Output = io::Result<tokio::net::TcpStream>> + Send {
+        crate::preview2::spawn(socket.connect(addr))
```
alexcrichton, line 317: How come this does a `spawn`?

alexcrichton (approved):
Nice! One very small comment but otherwise everything looks as I'd expect here 👍

alexcrichton:
Looking into the `spawn` bits now one orthogonal thing to highlight is that this doesn't compile currently on macOS and there's a few more bits and pieces that need an update (feel free to `prtest:full` here)

crates/wasi/src/preview2/tcp.rs
alexcrichton: Ah ok in that case you can handle that with some judicious calls to `with_ambient_tokio_runtime`. This got a panicking test to pass for me for example:

```diff
diff --git a/crates/wasi/src/preview2/host/tcp.rs b/crates/wasi/src/preview2/host/tcp.rs
index 1fce052cd..614495874 100644
--- a/crates/wasi/src/preview2/host/tcp.rs
+++ b/crates/wasi/src/preview2/host/tcp.rs
@@ -158,7 +160,7 @@ impl<T: WasiView> crate::preview2::host::tcp::tcp::HostTcpSocket for T {
             TcpState::ConnectReady(result) => result,
             TcpState::Connecting(mut future) => {
                 let mut cx = std::task::Context::from_waker(futures::task::noop_waker_ref());
-                match future.as_mut().poll(&mut cx) {
+                match with_ambient_tokio_runtime(|| future.as_mut().poll(&mut cx)) {
                     Poll::Ready(result) => result,
                     Poll::Pending => {
                         socket.tcp_state = TcpState::Connecting(future);
```

badeend:
Thanks! I've removed the task spawn and fixed the MacOS build.

PR #7897, opened 2024-02-08
Fix test pooling_allocator::instance_too_large when wmemcheck is enabled

cardoso:
I noticed this test was failing with the wmemcheck feature enabled. Instead of disabling the test, I special-cased the expectations, but I'm happy to change the approach.

I'm developing on aarch64 (M1 Max) in case other platforms exhibit different behavior.


fitzgen (approved):
Thanks!

PR #7902, opened 2024-02-09
Perform stronger typechecks of host-owned resources

### LINKED ISSUE #7883: Running method that takes resource as argument and returns resource multiple times returns same instance
### Test Case

https://github.com/Lifeblossom/wasmtime-resource-paremeter-bug-report

Wit file:

```
package component:bug-report;

world example {
    import resource2-interface;
    import resource1-interface;
}

interface resource1-interface {
    create: func() -> resource1;
    resource resource1;
}

interface resource2-interface {

    use resource1-interface.{resource1};
    create-resource2: func(resource1: resource1) -> resource2;

    do-stuff: func(resource1: resource1, resource2: resource2) -> resource2;

    resource resource2;
}

```

### Steps to Reproduce

* Clone repository
* Run cargo component build
* Run cargo test --test test

Test that compares returned resource instances will run

### Expected Results

Test passes

### Actual Results

Test failes on resource2 comparison - that means that `create-resource2` returned the same value two times.

### Versions and Environment

Wasmtime version or commit: 17.0.0

Operating system: Windows 11

Architecture: amd64

### Extra Info

Test also includes method that uses both resource1 and resource2. You can comment out `ensure!(res2_1 != res2_2);` which causes this method to run. Invocation results in following error:

```
Error: unknown handle index 0
```

Other findings
When using debugger both `res2_1` and `res_2_2` have idx `0`, but when `res_1_1` is replaced with `res_1_2` it is `1`. So I guess the resource id is taken from resource that was used as parameter.

alexcrichton:
Right now the abstraction for host-owned resources in Wasmtime is quite simple, it's "just an index". This can be error prone because the host can suffer both from use-after-free and ABA-style problems. While there's not much that can be done about use-after-free the previous implementation would accidentally enable "AB" style problems where if a host-owned resource index was deallocated and then reallocated with another type the original handle would still work. While not a major bug this can be confusing and additionally violate's our guarantees as a component runtime to guests to ensure that resources always have valid `rep`s going into components.

This commit adds a new layer of storage which keeps track of the `ResourceType` for all host-owned resources. This means that resources going into a host-owned table now record their type and coming out of a host-owned table a typecheck is always performed. Note that guests can continue to skip this logic as they already have per-type tables and so won't suffer the same issue.

This change is inspired by my taking a look at #7883. The crux of the issue was a typo where a resource was reused by accident which led to confusing behavior due to the reuse. This change instead makes it return an error more quickly and doesn't allow the component to see any invalid state.

Closes #7883

### DIFF
--- a/crates/wasmtime/src/runtime/component/resources.rs
+++ b/crates/wasmtime/src/runtime/component/resources.rs
@@ -11,7 +11,9 @@ use std::fmt;
 use std::marker;
 use std::mem::MaybeUninit;
 use std::sync::atomic::{AtomicU32, Ordering::Relaxed};
-use wasmtime_environ::component::{CanonicalAbiInfo, DefinedResourceIndex, InterfaceType};
+use wasmtime_environ::component::{
+    CanonicalAbiInfo, DefinedResourceIndex, InterfaceType, TypeResourceTableIndex,
+};
 use wasmtime_runtime::component::{ComponentInstance, InstanceFlags, ResourceTables};
 use wasmtime_runtime::{SendSyncPtr, VMFuncRef, ValRaw};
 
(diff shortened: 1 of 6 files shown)


fitzgen:
I haven't looked at the actual PR yet, just read your comment, but:

Is recording the type info enough to solve the ABA problem?

I don't think so: we could allocate a resource of type A, create an index handle to it, deallocate it, allocate a new resource that is also of type A in that place, and then use the old index handle to access the wrong resource.

I think adding a generation counter is the most straightforward way to definitely resolve the ABA problem here.

crates/wasmtime/src/runtime/component/resources.rs
```diff
@@ -238,15 +240,200 @@ const BORROW: u32 = u32::MAX;
 const NOT_IN_TABLE: u32 = u32::MAX - 1;
 const TAKEN: u32 = u32::MAX - 2;
 
-fn host_resource_tables(store: &mut StoreOpaque) -> ResourceTables<'_> {
-    let (calls, host_table) = store.component_calls_and_host_table();
-    ResourceTables {
-        calls,
-        host_table: Some(host_table),
-        tables: None,
+/// TODO
```
fitzgen, line 243: todo

crates/wasmtime/src/runtime/component/resources.rs
```diff
@@ -238,15 +240,200 @@ const BORROW: u32 = u32::MAX;
...
 }
 
+// fn host_resource_tables(store: &mut StoreOpaque) -> ResourceTables<'_> {
+//     let (calls, host_table, _) = store.component_resource_state();
+//     ResourceTables {
+//         calls,
+//         host_table: Some(host_table),
+//         tables: None,
+//     }
+// }
```
fitzgen, line 435: commented out code

fitzgen (approved):
The code itself LGTM, and this is an improvement over the current status quo, so I don't mind landing this.

But I'd like to hear your thoughts on how we can integrate a generation to fully prevent ABA issues. I'm not 100% sure on the constraints we are working within here. For example, are we restricted to `u32` indices and therefore we can't add a second `u32` for generation? If so, perhaps we can repurpose 8 bits into a generation? Would 24 bits be enough for all active resources? Do we have a limit on how many active resources can exist at a time? Such a small generation likely wouldn't be enough to fully prevent ABA problems, but I think it would still be a further improvement.

crates/wasmtime/src/runtime/component/resources.rs
```diff
@@ -203,71 +203,166 @@ pub struct Resource<T> {
...
+pub struct HostResourceIndex(u64);
+
+impl HostResourceIndex {
+    fn new(idx: u32, gen: u32) -> HostResourceIndex {
+        HostResourceIndex(u64::from(idx) | (u64::from(gen) << 32))
+    }
+
+    fn index(&self) -> u32 {
+        self.0 as u32
+    }
```
fitzgen, line 340: Mind explicitly masking off the generation and doing `u32::try_from(masked).unwrap()`? I think that would make things a lot more clear and avoid hiding subtle issues with `as` casts as we change representations and stuff like that.

fitzgen (approved):
Nice!

PR #7903, opened 2024-02-09
Make some MemFlags functions const

elliottt:
Switch the setter operations on `MemFlags` to use const functions internally, allowing the `with` interface to become `const`. This allows for `MemFlags` values to be defined as constant, and removes the need to cache trusted and untrusted values in the winch x64 macro assembler.

### DIFF
--- a/winch/codegen/src/isa/x64/masm.rs
+++ b/winch/codegen/src/isa/x64/masm.rs
@@ -47,10 +47,6 @@ pub(crate) struct MacroAssembler {
     shared_flags: settings::Flags,
     /// The target pointer size.
     ptr_size: OperandSize,
-    /// Flags for native loads/stores.
-    trusted_flags: MemFlags,
-    /// Flags for Wasm loads/stores.
-    untrusted_flags: MemFlags,
 }
 
 impl Masm for MacroAssembler {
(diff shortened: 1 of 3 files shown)


winch/codegen/src/isa/x64/masm.rs
```diff
@@ -1145,27 +1141,28 @@ impl Masm for MacroAssembler {
...
 
 impl MacroAssembler {
+    /// Memory flags for trusted loads/stores.
+    const TRUSTED_FLAGS: MemFlags = MemFlags::trusted();
+
+    /// Flags used for WebAssembly loads / stores.
+    /// Untrusted by default so we don't set `no_trap`.
+    /// We also ensure that the endianess is the right one for WebAssembly.
+    const UNTRUSTED_FLAGS: MemFlags = MemFlags::new().with_endianness(Endianness::Little);
+
```
saulecabrera, line 1151: When I was looking at this, I had thought of defining them where the `MacroAssesmbler` trait is defined, because they'll probably be consumed by all the other ISAs. Do you think it makes sense to move them over?

fitzgen (approved):

saulecabrera (approved):
Left one minor suggestion. Thanks for adding this!

winch/codegen/src/isa/x64/masm.rs
elliottt: That's a great idea! I'll move them over :+1:

PR #7909, opened 2024-02-09
winch: Add saturating conversion instructions

saulecabrera:
This commit adds support for the saturating conversions instructions.

I missed these instructions in https://github.com/bytecodealliance/wasmtime/issues/6528


elliottt (approved):
Good catch, looks good to me!

PR #7910, opened 2024-02-10
vet: prune lockfile

abrown:
When running `cargo vet` in #7900, it warned me that we should consider pruning some unused entries, etc. This is the result of running `cargo vet prune`.

### DIFF
--- a/supply-chain/imports.lock
+++ b/supply-chain/imports.lock
@@ -2,392 +2,200 @@
 # cargo-vet imports lock
 
 [[unpublished.cranelift]]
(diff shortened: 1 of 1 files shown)


elliottt (approved):
Looks good! 🧹

alexcrichton:
Oh I think this has unfortunately undone the work of https://github.com/bytecodealliance/wasmtime/pull/7900 so we'll need to reland that for https://github.com/bytecodealliance/wasmtime/pull/7807

elliottt:
Should we revert this one?

PR #7129, opened 2023-10-02
CI: switch to AArch64 macOS runners.

cfallin:
This should result in faster builds for macOS, and additionally will let us actually run tests on aarch64-macos now (previously we only had binary builds).

Fixes #4177.

prtest:full

PR #7914, opened 2024-02-10
Test with aarch64 macOS on CI

alexcrichton:
Rebase of #7129 with the `macos-14` string which GitHub indicates should now work for public repos.


alexcrichton:
Oh wow the arm64 builders for macos are 2x faster than the x86_64 builders... Maybe we should cross-compile the other direction!

fitzgen (approved):

PR #7917, opened 2024-02-12
Shrink the size of int-to-float conversions

alexcrichton:
This commit adds egraph optimization rules to shrink the size of int-to-float conversions where possible. This helps backends like the x64 backend where the 64-bit-unsigned-integer-to-float conversions does not have a single instruction lowering unlike other types.

### DIFF
--- a/cranelift/codegen/src/opts/arithmetic.isle
+++ b/cranelift/codegen/src/opts/arithmetic.isle
@@ -170,7 +170,7 @@
 
 ;; Cranelift's `fcvt_from_{u,s}int` instructions are polymorphic over the input
 ;; type so remove any unnecessary `uextend` or `sextend` to give backends
-;; the chance to convert from the smallest integral type to the a float. This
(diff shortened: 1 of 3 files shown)


cranelift/codegen/src/opts/arithmetic.isle
```diff
@@ -167,3 +167,13 @@
       (if-let $true (ty_equal half_ty (ty_half_width ty)))
       (if-let $true (u64_eq k (ty_bits_u64 half_ty)))
       (uextend ty (umulhi half_ty x y)))
+
+;; Cranelift's `fcvt_from_{u,s}int` instructions are polymorphic over the input
+;; type so remove any unnecessary `uextend` or `sextend` to give backends
+;; the chance to convert from the smallest integral type to the a float. This
```
abrown, line 173:
```suggestion
;; the chance to convert from the smallest integral type to the float. This
```

abrown (approved):

PR #7919, opened 2024-02-12
x64: Add `fcvt_from_uint` lowering for i64x2

alexcrichton:
This commit adds a general purpose lowering for the `fcvt_from_uint` instruction in addition to the previously specialized lowering for what the wasm frontend produces. This is unlikely to get triggered much from the wasm frontend except when intermediate optimizations change the shape of code. The goal of this commit is to address issues such as those identified in #7915 and historically by ensuring that there's a lowering for the instruction for all input types instead of trying to massage the input into the right form.

This instruction lowering was crafted by studying LLVM's output and I've put commentary to the best of my ability as to what's going on.

### DIFF
--- a/cranelift/codegen/src/isa/x64/lower.isle
+++ b/cranelift/codegen/src/isa/x64/lower.isle
@@ -3417,6 +3417,45 @@
 (rule (lower (has_type ty (fcvt_from_uint val @ (value_type $I64))))
       (cvt_u64_to_float_seq ty val))
 
+;; Base case of u64x2 being converted to f64x2. No native instruction for this
+;; is available so it's emulated through a series of instructions that exploit
+;; the binary representation of 64-bit floats. This sequence of instructions is
+;; copied from LLVM and my understanding of the general idea is to roughly:
+;;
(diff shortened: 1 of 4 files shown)


cranelift/codegen/src/isa/x64/lower.isle
```diff
@@ -3417,6 +3417,45 @@
...
+;; * Move the low 32 bits of the input into one register and the upper
+;;   32-bits into a different register, where both have all 0s for the upper
+;;   32-bits. (e.g. split the 64-bit input into two locations)
+;; * For the low bits, create `1.<twenty-zeros><low32>p52` via bit tricks.
+;; * For the high bits, create `1.<twenty-zeros><high32>p84` via bit tricks.
+;; * Create the constant `1.0p84 + 1.0p52`
+;; * Add the two high halves and subtract the constant.
+;;
+;; Apply some math and this should produce the same result as the native
+;; conversion.
```
fitzgen, line 3435: :magic_wand:

fitzgen (approved):

PR #7927, opened 2024-02-13
winch: Optimize calls

saulecabrera:
This commit introduces several optimizations to speed up the compilation of function calls:

* Keep track of previously resolved function signatures for local or imported callees to avoid computing the `ABISig` on every function call.
* Keep track of previously resolved type signatures for indirect calls to avoid computing the `ABISig` on every function call.
* Refactor `CallKnown` and `CallUnknown` instructions to make the `BoxCallInfo` field in the struct optional. Prior to this change, from Winch's perspective each call lowering involved a heap allocation, using the default values for `BoxCallInfo`, which in the end are not used by Winch.
* Switch Winch's internal `Stack` to use a `SmallVec` rather than a `Vec`. Many of the operations involving builtin function calls require inserting elements at arbitrary offsets in the stack and using a `SmallVec` makes this process more efficient.

With the changes mentioned above, I observed ~30% improvement in compilation times for modules that are call-heavy.

### DIFF
--- a/cranelift/codegen/src/isa/x64/inst/mod.rs
+++ b/cranelift/codegen/src/isa/x64/inst/mod.rs
@@ -2351,36 +2351,34 @@ fn x64_get_operands<F: Fn(VReg) -> VReg>(inst: &Inst, collector: &mut OperandCol
             // regalloc, so we do not need to represent its ABI to the
             // register allocator. Assert that we don't alter that
             // arrangement.
-            if let Some(info) = info {
-                debug_assert_ne!(*dest, ExternalName::LibCall(LibCall::Probestack));
-                for u in &info.uses {
-                    collector.reg_fixed_use(u.vreg, u.preg);
-                }
-                for d in &info.defs {
-                    collector.reg_fixed_def(d.vreg, d.preg);
-                }
-                collector.reg_clobbers(info.clobbers);
+            let info = info.as_ref().expect("CallInfo is expected in this path");
+            debug_assert_ne!(*dest, ExternalName::LibCall(LibCall::Probestack));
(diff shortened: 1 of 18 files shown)


cranelift/codegen/src/isa/x64/inst/mod.rs
```diff
@@ -2348,37 +2346,41 @@ fn x64_get_operands<F: Fn(VReg) -> VReg>(inst: &Inst, collector: &mut OperandCol
...
             // regalloc, so we do not need to represent its ABI to the
             // register allocator. Assert that we don't alter that
             // arrangement.
-            debug_assert_ne!(*dest, ExternalName::LibCall(LibCall::Probestack));
-            for u in &info.uses {
-                collector.reg_fixed_use(u.vreg, u.preg);
-            }
-            for d in &info.defs {
-                collector.reg_fixed_def(d.vreg, d.preg);
+            if let Some(info) = info {
```
saulecabrera, line 2354: Some thoughts that I had while changing this:

* In case there are no objections or downsides to making `BoxCallInfo` optional, would it be preferred to expect that the info is present in this path?
* Another alternative that I considered in case that making `BoxCallInfo` optional is not desired, would be to create another instruction without `BoxCallInfo`, which would be only consumed by Winch, at the cost of some code duplication.

elliottt (approved):
This seems good to me!

cranelift/codegen/src/isa/x64/inst/mod.rs
saulecabrera: Thanks for the suggestion! Added a comment.

PR #7932, opened 2024-02-14
Fully enable Winch in the `differential` fuzzer

alexcrichton:
This commit fully enables usage of Winch in the `differential` fuzzer against all other engines with no special cases. I attempted enabling winch for the other fuzzers as well but Winch doesn't currently implement all methods for generating various trampolines required so it's currently only limited to the `differential` fuzzer.

This adds Winch as an "engine" and additionally ensures that when configured various wasm proposals are disabled that Winch doesn't support (similar to how enabling `wasmi` disables proposals that `wasmi` doesn't support).

This does reduce fuzzing of Winch slightly in that the reference-types proposal is completely disabled for Winch rather than half-enabled where Winch doesn't implement `externref` operations yet but does implement `funcref` operations. This, however, enables integrating it more cleanly into the rest of the fuzzing infrastructure with fewer special cases.

### DIFF
--- a/crates/fuzzing/src/generators/config.rs
+++ b/crates/fuzzing/src/generators/config.rs
@@ -316,6 +316,21 @@ impl Config {
         std::fs::write(&file, module.serialize().unwrap()).unwrap();
         unsafe { Ok(Module::deserialize_file(engine, &file).unwrap()) }
     }
+
+    /// Winch doesn't support the same set of wasm proposal as Cranelift at
+    /// this time, so if winch is selected be sure to disable wasm proposals
+    /// in `Config` to ensure that Winch can compile the module that
+    /// wasm-smith generates.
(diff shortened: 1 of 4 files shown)


saulecabrera (approved):
Thanks for working on this! 

> This does reduce fuzzing of Winch slightly in that the reference-types proposal is completely disabled for Winch rather than half-enabled where Winch doesn't implement externref operations yet but does implement funcref operations. 

I think this is fine, we're starting the work on reference types, so we'd be able permanently enable fuzzing for `funcref` and `externref` soon-ish.

Given this change, I've taken the liberty to add myself to oss-fuzz (https://github.com/google/oss-fuzz/pull/11602), as I'd like to monitor more closely any Winch fuzz bugs. Let me know if you're ok with that change.

alexcrichton:
> Let me know if you're ok with that change.

Definitely!

PR #7934, opened 2024-02-14
Restructure the MacroAssembler interface to clobbers

elliottt:
To prepare for moving the stack check into the function prologue, rework the macro assembler's interface to clobbers to ensure that the unwind info defines the prologue correctly on windows. Though clobbers are only used in two trampolines, it's a nice interface to use the save/restore points to bracket the code between the prologue and epilogue.

I made an additional change that's not strictly necessary in this PR: reserving the space for clobbers with `reserve_stack`, and then storing values, rather than calling push for each value. I opted to maintain 16-byte stack alignment in this case, despite winch fixing stack alignment up around calls, and this paid off in trampolines that require clobbers as they do one fewer push/pop combination.

### DIFF
--- a/winch/codegen/src/isa/x64/masm.rs
+++ b/winch/codegen/src/isa/x64/masm.rs
@@ -89,13 +89,16 @@ impl Masm for MacroAssembler {
     }
 
     fn save_clobbers(&mut self, clobbers: &[(Reg, OperandSize)]) {
+        let int_bytes: u32 = Self::ABI::word_bytes().try_into().unwrap();
+        let float_bytes = int_bytes * 2;
+
         // Determine how much space we need for the clobbers
         let clobbered_size = align_to(
             clobbers
                 .iter()
                 .fold(0u32, |total, (reg, _)| match reg.class() {
-                    RegClass::Int => total + 8,
-                    RegClass::Float => align_to(total, 16) + 16,
+                    RegClass::Int => total + int_bytes,
+                    RegClass::Float => align_to(total, float_bytes) + float_bytes,
                     RegClass::Vector => unimplemented!(),
                 }),
             16,
@@ -132,7 +135,7 @@ impl Masm for MacroAssembler {
             off += size.bytes();
         }
 
-        // debug_assert_eq!(align_to(off, 16), clobbered_size);
+        debug_assert_eq!(align_to(off, float_bytes), clobbered_size);
     }
 
     fn restore_clobbers(&mut self, clobbers: &[(Reg, OperandSize)]) {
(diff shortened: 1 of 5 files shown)


winch/codegen/src/trampoline.rs
```diff
@@ -718,16 +715,15 @@ where
...
-        }
+        self.masm.restore_clobbers(&self.callee_saved_regs);
         self.masm.epilogue(0);
     }
 
     /// The trampoline's epilogue.
     fn epilogue(&mut self, arg_size: u32) {
         // Free the stack space allocated by pushing the trampoline arguments.
         self.masm.free_stack(arg_size);
+        self.masm.restore_clobbers(&[]);
```
abrown, line 726: Should we have something like:

```suggestion
        debug_assert!(self.callee_saved_regs.is_empty());
        self.masm.restore_clobbers(&[]);
```

abrown (approved):

winch/codegen/src/trampoline.rs
elliottt: That's a great idea, thanks Andrew!

winch/codegen/src/isa/x64/masm.rs
```diff
@@ -98,6 +88,70 @@ impl Masm for MacroAssembler {
...
     }
 
+    fn save_clobbers(&mut self, clobbers: &[(Reg, OperandSize)]) {
+        // Determine how much space we need for the clobbers
+        let clobbered_size = align_to(
+            clobbers
+                .iter()
+                .fold(0u32, |total, (reg, _)| match reg.class() {
+                    RegClass::Int => total + 8,
+                    RegClass::Float => align_to(total, 16) + 16,
```
saulecabrera, line 98: Would it make sense to use `ABI::word_bytes()` and `ABI::word_bytes() * 2` here?

winch/codegen/src/isa/x64/masm.rs
```diff
@@ -98,6 +88,70 @@ impl Masm for MacroAssembler {
...
+                    clobber_offset: off,
+                    reg: reg.into(),
+                });
+            }
+
+            // Increment the offset
+            off += size.bytes();
+        }
+
+        // debug_assert_eq!(align_to(off, 16), clobbered_size);
```
saulecabrera, line 135: Do we still want to keep this `debug_assert_eq`?

saulecabrera (approved):
Thanks for re-structuring this, and sorry it took me a bit to get to it. I left a couple of small suggestions.

winch/codegen/src/isa/x64/masm.rs
elliottt: I like that! I think it also gives good justification to uncommenting the debug assert you mentioned below, as now it's checking the connection between the `word_bytes()` and `OperandSize::bytes()` results.

winch/codegen/src/isa/x64/masm.rs
elliottt: I think so, after the change above. I'll uncomment it :)

PR #7947, opened 2024-02-15
winch: Move the stack overflow check to the function prologue

elliottt:
Move the stack check to the function prologue in winch generated code by adding the final stack maximum to the stack min bound before comparing against the current SP.

This new approach is more accurate than the previous implementation that would only check the size of `locals`, but does require that we patch the function during `MacroAssembler::finalize`, as we don't know the stack high water mark until then. Enabling the patching required the addition of the `start_patchable`/`end_patchable` api to `MachBuffer`, which gives a way to name sections of the code buffer (with some restrictions) that can be edited later. We use this then to modify an add-with-immediate instruction to add the used stack space to the stack lower bound from `vmctx`, when finalizing the function.

### DIFF
--- a/cranelift/codegen/src/machinst/buffer.rs
+++ b/cranelift/codegen/src/machinst/buffer.rs
@@ -414,10 +414,18 @@ pub enum StackMapExtent {
     StartedAtOffset(CodeOffset),
 }
 
-/// An open patch region represents the beginning of an editable region
+/// Represents the beginning of an editable region in the [`MachBuffer`], while code emission is
+/// still occurring. An [`OpenPatchRegion`] is closed by [`MachBuffer::end_patchable`], consuming
+/// the [`OpenPatchRegion`] token in the process.
 pub struct OpenPatchRegion(usize);
 
-/// A patchable region in the [`MachBuffer`].
+/// A region in the [`MachBuffer`] code buffer that can be edited prior to finalization. An example
+/// of where you might want to use this is for patching instructions that mention constants that
+/// won't be known until later: [`MachBuffer::start_patchable`] can be used to begin the patchable
+/// region, instructions can be emitted with placeholder constants, and the [`PatchRegion`] token
+/// can be produced by [`MachBuffer::end_patchable`]. Once the values of those constants are known,
+/// the [`PatchRegion::patch`] function can be used to get a mutable buffer to the instruction
+/// bytes, and the constants uses can be updated directly.
 pub struct PatchRegion {
     range: std::ops::Range<usize>,
 }
@@ -539,8 +547,6 @@ impl<I: VCodeInst> MachBuffer<I> {
     ///
     /// 1. It must not introduce any instructions that could be chomped
     ///    (branches are an example of this)
-    /// 2. Additional calls to `start_patchable` must not occur, as we require
-    ///    that patchable regions do not overlap.
     pub fn start_patchable(&mut self) -> OpenPatchRegion {
         assert!(!self.open_patchable, "Patchable regions may not be nested");
         self.open_patchable = true;
@@ -750,6 +756,10 @@ impl<I: VCodeInst> MachBuffer<I> {
     /// Additional requirement: no labels may be bound between `start` and `end`
     /// (exclusive on both ends).
     pub fn add_uncond_branch(&mut self, start: CodeOffset, end: CodeOffset, target: MachLabel) {
+        debug_assert!(
+            !self.open_patchable,
+            "Branch instruction inserted within a patchable region"
+        );
         assert!(self.cur_offset() == start);
         debug_assert!(end > start);
         assert!(!self.pending_fixup_records.is_empty());
@@ -781,6 +791,10 @@ impl<I: VCodeInst> MachBuffer<I> {
         target: MachLabel,
         inverted: &[u8],
     ) {
+        debug_assert!(
+            !self.open_patchable,
+            "Branch instruction inserted within a patchable region"
+        );
         assert!(self.cur_offset() == start);
         debug_assert!(end > start);
         assert!(!self.pending_fixup_records.is_empty());
--- a/cranelift/codegen/src/machinst/buffer.rs
+++ b/cranelift/codegen/src/machinst/buffer.rs
@@ -311,6 +311,9 @@ pub struct MachBuffer<I: VCodeInst> {
     /// constant may appear in this array multiple times if it was emitted
     /// multiple times.
     used_constants: SmallVec<[(VCodeConstant, CodeOffset); 4]>,
+    /// Indicates when a patchable region is currently open, to guard that it's
+    /// not possible to nest patchable regions.
+    open_patchable: bool,
 }
 
 impl MachBufferFinalized<Stencil> {
@@ -411,6 +414,21 @@ pub enum StackMapExtent {
     StartedAtOffset(CodeOffset),
 }
 
+/// An open patch region represents the beginning of an editable region
+pub struct OpenPatchRegion(usize);
+
+/// A patchable region in the [`MachBuffer`].
+pub struct PatchRegion {
+    range: std::ops::Range<usize>,
+}
+
+impl PatchRegion {
+    /// Consume the patch region to yield a mutable slice of the [`MachBuffer`] data buffer.
+    pub fn patch<I: VCodeInst>(self, buffer: &mut MachBuffer<I>) -> &mut [u8] {
+        &mut buffer.data[self.range]
+    }
+}
+
 impl<I: VCodeInst> MachBuffer<I> {
     /// Create a new section, known to start at `start_offset` and with a size limited to
     /// `length_limit`.
@@ -437,6 +455,7 @@ impl<I: VCodeInst> MachBuffer<I> {
             labels_at_tail_off: 0,
             constants: Default::default(),
             used_constants: Default::default(),
+            open_patchable: false,
         }
     }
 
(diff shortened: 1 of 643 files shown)


cranelift/codegen/src/machinst/buffer.rs
```diff
@@ -411,6 +414,21 @@ pub enum StackMapExtent {
     StartedAtOffset(CodeOffset),
 }
 
+/// An open patch region represents the beginning of an editable region
```
cfallin, line 417: Can we add a bit more doc here? "Represents an editable region that is currently being emitted in the `MachBuffer`. When closed, this becomes a `PatchRegion` and allows editing the machine code post-emission." or something like that?

cranelift/codegen/src/machinst/buffer.rs
```diff
@@ -411,6 +414,21 @@ pub enum StackMapExtent {
...
+pub struct OpenPatchRegion(usize);
+
+/// A patchable region in the [`MachBuffer`].
```
cfallin, line 420: Likewise here; note what it's for.

cranelift/codegen/src/machinst/buffer.rs
```diff
@@ -515,6 +534,30 @@ impl<I: VCodeInst> MachBuffer<I> {
         // Post-invariant: as for `put1()`.
     }
 
+    /// Begin a region of patchable code. There are some requirements for the
+    /// code that is emitted:
+    ///
+    /// 1. It must not introduce any instructions that could be chomped
+    ///    (branches are an example of this)
```
cfallin, line 541: I think branches are the only case actually; and at the `MachBuffer` layer all that matters are branches that we tell the `MachBuffer` about, via `add_cond_branch` and `add_uncond_branch`. So I think we can make this more concrete by saying "It must not introduce any MachBuffer-visible branches (i.e., you must not call `add_cond_branch` or `add_uncond_branch` between `start_patchable` and `end_patchable`)."

cranelift/codegen/src/machinst/buffer.rs
```diff
@@ -759,6 +802,11 @@ impl<I: VCodeInst> MachBuffer<I> {
     }
 
     fn truncate_last_branch(&mut self) {
+        debug_assert!(
```
cfallin, line 805: Could we add this assert in `add_cond_branch` and `add_uncond_branch` as well?

winch/codegen/src/isa/x64/masm.rs
```diff
@@ -1163,15 +1174,40 @@ impl MacroAssembler {
...
     }
 
+    /// Add the maximum stack used to a register, recording an obligation to update the
+    /// add-with-immediate instruction emitted to use the real stack max when the masm is being
+    /// finalized.
+    fn add_stack_max(&mut self, reg: Reg) {
+        let open = self.asm.buffer_mut().start_patchable();
+        // NOTE: we write out a value that is not encodable as an imm8, to ensure that we pick the
+        // variation of the x64 add instruction that supports imm32
+        self.asm.add_ir(-129, reg, OperandSize::S32);
```
cfallin, line 1193: This and the asserts on opcode bytes below feel a little fragile. For example if we change the way immediates are emitted in the future, this might break (we're tying masm's lowering strategy to a specific assumption about its output here). Also the REX byte is pinned by an assert, and might change if the regalloc picks a register in the other half of the 16-reg bank (!). Perhaps what we could do instead is something like:

- Have a `PatchableValueLoad` abstraction (either an `MInst` case or a little struct here that does its own encoding, to nail down the length-invariance); ensure that it always produces machine code of the same length. `::new(value: u32, reg: Reg)`, `::update_value(value: u32)` and `::emit_to_slice(&mut [u8])` or something like that.
- Use it here to initially put a placeholder in; we can avoid using a magic value and magic knowledge about immediate encoding then. Perhaps we can even write a wrapper `::emit_to_buffer(&mut MachBuffer) -> PatchRegion`.
- Later, patch using this; perhaps with `::emit_to_region(&mut MachBuffer, PatchRegion)`.

Thoughts?

cfallin (commented):
Overall approach seems fine; some thoughts on the x64 machine-code patching specifically:

elliottt:
The tests are failing due to the addition of stack checking to one of the trampolines (as the vmctx isn't setup yet in one case). I might roll those back so that this can be debugged separately.

winch/codegen/src/isa/x64/masm.rs
elliottt: I agree that it's fragile as-is, and something a bit more robust would be good here.

Adding a special `PatchableValueLoad` type sounds good to me, as long as you don't mind the more specific "add patched immediate to register" behavior. I wanted to avoid using an additional register here, so that the sequence of loading the stack lower bound and adding the immediate for the final stack value could be done with only the scratch register.

I'll experiment with adding a `PatchableAddRegImm` struct in the x64 backend in cranelift, as that would make it easy to reuse logic from the existing emit implemetation for add.

winch/codegen/src/isa/x64/masm.rs
```diff
@@ -1163,15 +1174,40 @@ impl MacroAssembler {
...
+        let bytes = region.patch(self.asm.buffer_mut());
+        debug_assert_eq!(bytes.len(), 7);
+        debug_assert_eq!(bytes[0], 0x49);
+        debug_assert_eq!(bytes[1], 0x81);
+        bytes[3..].copy_from_slice(self.sp_max.to_le_bytes().as_slice());
+    }
+
     fn increment_sp(&mut self, bytes: u32) {
         self.sp_offset += bytes;
+        self.sp_max = self.sp_max.max(self.sp_offset);
```
saulecabrera, line 1210: Could we add a comment here on why we're using `.max`? The reason why I bring this up is that we currently don't have a way to track the "canonical stack pointer offset" across the compiler, which is what we'd probably want to use here instead (that's what `.max` is doing here if I'm understanding correctly). We currently use the `sp_offset` as the single source of truth, but lately I've been thinking that it might be a good idea to change that (but not a blocker for this change!). This is mostly relevant in control flow, where we have to  soft-reset the stack pointer in some situations.

saulecabrera (commented):
The Winch pieces look good to me. Left a small question and a comment. I'm going to defer the final approval to Chris, given that I'm not as familiar with the `MachBuffer` implementation.

winch/codegen/src/isa/x64/masm.rs
```diff
@@ -30,15 +30,20 @@ use cranelift_codegen::{
...
+use smallvec::SmallVec;
 use wasmtime_environ::{PtrSize, WasmValType, WASM_PAGE_SIZE};
 
 /// x64 MacroAssembler.
 pub(crate) struct MacroAssembler {
     /// Stack pointer offset.
     sp_offset: u32,
+    /// Stack high-water mark.
+    sp_max: u32,
+    /// Regions where the addition of the used stack must be fixed up
```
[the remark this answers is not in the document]
elliottt: This could definitely be `1` or even an optional. Given that we could possibly call `check_stack` multiple times, I wanted to make sure that we would still generate valid code. Perhaps the right solution is an `Option<PatchRegion>` and an assert?

winch/codegen/src/isa/x64/masm.rs
elliottt: The intent was to track the largest stack offset throughout code generation, so that we had an upper bound for the stack space needed. As push and reserve_stack both use `increment_sp`, this felt like a pretty good proxy for tracking the maximum seen stack allocation, but maybe I'm missing something here?

saulecabrera: > this felt like a pretty good proxy for tracking the maximum seen stack allocation

I think it is! I was just thinking that maybe it makes sense to have a comment stating that we do this (`.max`) given that we currently don't explicitly track the maximum amount of stack generated anywhere in the compiler yet, which I think we could as an optimization.

winch/codegen/src/isa/x64/masm.rs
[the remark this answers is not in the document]
saulecabrera: Oh that makes sense, thanks for explaining. Yeah perhaps we could try going with `Option<PatchRegion>` + assert.

winch/codegen/src/isa/x64/masm.rs
elliottt: Ah! I'll add a comment to clarify the intent here. I can also put some more documentation into the `sp_max` field itself 👍

winch/codegen/src/isa/x64/masm.rs
elliottt: Would you mind taking a look at the changes in x64/asm.rs? I had to copy some details over from cranelift's emit module for rex encoding, but I believe that this approach should be much more resilient: it doesn't assume anything about how cranelift would encode an instruction; it records the offset into the patch region where the immediate starts, so it should also be resilient to local changes.

The downside is that it brought a lot of details into the asm module, but I think that's somewhat unavoidable at this  point.

winch/codegen/src/isa/x64/masm.rs
elliottt: I added some more comments to this point in `increment_sp`, and to the definition of `sp_max`, would you mind taking another look?

cranelift/codegen/src/machinst/buffer.rs
elliottt: I've updated the language on this comment, what do you think?

cfallin (approved):
Looks great, thanks! Just one code-reuse nit below, otherwise good to go.

winch/codegen/src/isa/x64/masm.rs
```diff
@@ -39,6 +39,13 @@ use wasmtime_environ::{PtrSize, WasmValType, WASM_PAGE_SIZE};
 pub(crate) struct MacroAssembler {
     /// Stack pointer offset.
     sp_offset: u32,
+    /// This value represents the maximum stack size seen while compiling the function. While the
+    /// function is still being compiled its value will not be valid (the stack will grow and
+    /// shrink as space is reserved and freed during compilstion), but once all instructions have
```
saulecabrera, line 44:
```suggestion
    /// shrink as space is reserved and freed during compilation), but once all instructions have
```

winch/codegen/src/isa/x64/masm.rs
saulecabrera: Looks great, thanks!

winch/codegen/src/isa/x64/asm.rs
```diff
@@ -1340,3 +1340,136 @@ impl Assembler {
         });
     }
 }
+
+/// A small bit field to record a REX prefix specification:
+/// - bit 0 set to 1 indicates REX.W must be 0 (cleared).
+/// - bit 1 set to 1 indicates the REX prefix must always be emitted.
+#[repr(transparent)]
+#[derive(Clone, Copy)]
+struct RexFlags(u8);
```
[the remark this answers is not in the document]
elliottt: Yep! My original worry was needing to compute the offset for what was written to the `MachBuffer`, but that's done easily enough with `MachBuffer::cur_offset`. It's much nicer not duplicating this!

winch/codegen/src/isa/x64/masm.rs
elliottt: Amazing spelling mistake, thank you for catching that!

PR #7948, opened 2024-02-15
cranelift: Remove redundant dominator-tree structure

jameysharp:
The egraph work introduced a helper for walking over the dominator tree in pre-order, but it turns out that helper already existed in the `dominator_tree` module itself, and the older version also supports constant-time dominance checks using the same trick that @cfallin suggested in #7891. So let's just use that version.

cc: @elliottt @lpereira


cranelift/codegen/src/egraph.rs
```diff
@@ -53,7 +51,7 @@ pub struct EgraphPass<'a> {
     /// "Domtree with children": like `domtree`, but with an explicit
     /// list of children, complementing the parent pointers stored
     /// in `domtree`.
-    domtree_children: DomTreeWithChildren,
+    domtree_children: DominatorTreePreorder,
```
cfallin, line 54: Can we update the name and comment here to match the type ("a preorder iterator over the domtree" or something similar)?

cfallin (approved):
Woah,nice find! I had no idea we already had this (which I guess is why we wrote it again).

While I hesitate to violate the law that "any sufficiently advanced compiler contains three domtree implementations" (regalloc2 has one too!), I think we can do some spring cleaning here :-)

cranelift/codegen/src/egraph.rs
jameysharp: When I looked again today at these modules to decide what to rename the fields to and what documentation to write, I discovered that the original `DominatorTree` isn't actually needed anywhere once we have a `DominatorTreePreorder` available. There was one `debug_assert!` using the original tree for dominance queries, but we can do those in O(1) time against the preorder instead. So I've renamed `domtree_children` just to `domtree`, replacing the other `domtree` field entirely.

cfallin: Nice!

lpereira:
LGTM!

PR #7949, opened 2024-02-15
winch: Ensure correct handling of libcalls

saulecabrera:
This commit fixes a fuzz bug where modules involving known libcalls would fail to compile given that they were unconditionally treated as colocated libcalls.

This bug is only reproducible in non sse41 environments, given that some operations like `floor` default to libcalls in this case. The `use_colocated_libcalls` setting is not configurable within Wasmtime and as such, they should be loaded into a register prior to emitting the call. This will also ensure that the right 8-byte absolute relocation is used.

### DIFF
--- a/cranelift/codegen/src/lib.rs
+++ b/cranelift/codegen/src/lib.rs
@@ -59,8 +59,8 @@ pub use crate::machinst::buffer::{
 };
 pub use crate::machinst::{
     CompiledCode, Final, MachBuffer, MachBufferFinalized, MachInst, MachInstEmit,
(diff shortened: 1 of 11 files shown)


saulecabrera:
I'm unsure if there's an additional way to test this one aside from the filetests.

elliottt (approved):
This looks good to me! As far as testing goes, this was caught by fuzzing so I think we're good. We could always look into running some of the spec tests with different codegen configurations, but that probably still won't catch as much as fuzzing will.

PR #7950, opened 2024-02-16
Allow `traps.rs` tests to pass even when `RUST_BACKTRACE=1` on nightly Rust

fitzgen:
Instead of asserting exact error messages, which `anyhow` auto adds Rust backtrace to on nightly, assert that the important bits are there.

Split out from #7943.

### DIFF
--- a/tests/all/traps.rs
+++ b/tests/all/traps.rs
@@ -314,9 +314,12 @@ error while executing at wasm backtrace:
     0:   0x23 - m!die
     1:   0x27 - m!<wasm function 1>
     2:   0x2c - m!foo
-    3:   0x31 - m!<wasm function 3>"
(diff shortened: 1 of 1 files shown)


alexcrichton (commented):
Personally I find it pretty helpful to have exhaustive assertions (ish) to get a better picture of what the error message looks like. It looks like `anyhow` appends `Stack backtrace: ...` at the end of the message, so could all of these assertions change from `assert_eq!` to `assert!(s.contains(...))` perhaps?

fitzgen:
> so could all of these assertions change from `assert_eq!` to `assert!(s.contains(...))` perhaps?

Done!

alexcrichton (approved):

PR #7968, opened 2024-02-20
cranelift: Enable "chaos mode" in egraph pass

jameysharp:
First of all, thread a "chaos mode" control-plane into Context::optimize and from there into EgraphPass, OptimizeCtx, and Elaborator.

In this commit we use the control-plane to change the following behaviors in ways which shouldn't cause incorrect results:

- Dominator-tree block traversal order for both the rule application and elaboration passes
- Order of evaluating optimization alternatives from `simplify`
- Choose worst values instead of best in each eclass

@lpereira and I wrote this together.

### DIFF
--- a/cranelift/codegen/src/egraph/elaborate.rs
+++ b/cranelift/codegen/src/egraph/elaborate.rs
@@ -12,6 +12,7 @@ use crate::loop_analysis::{Loop, LoopAnalysis};
 use crate::scoped_hash_map::ScopedHashMap;
 use crate::trace;
 use alloc::vec::Vec;
+use cranelift_control::ControlPlane;
 use cranelift_entity::{packed_option::ReservedValue, SecondaryMap};
 use smallvec::{smallvec, SmallVec};
 
@@ -64,6 +65,9 @@ pub(crate) struct Elaborator<'a> {
     /// Stats for various events during egraph processing, to help
     /// with optimization of this infrastructure.
     stats: &'a mut Stats,
+    /// Chaos-mode control-plane so we can test that we still get
+    /// correct results when our heuristics make bad decisions.
+    ctrl_plane: &'a mut ControlPlane,
 }
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
@@ -139,6 +143,7 @@ impl<'a> Elaborator<'a> {
         loop_analysis: &'a LoopAnalysis,
         remat_values: &'a FxHashSet<Value>,
         stats: &'a mut Stats,
+        ctrl_plane: &'a mut ControlPlane,
     ) -> Self {
         let num_values = func.dfg.num_values();
         let mut value_to_best_value =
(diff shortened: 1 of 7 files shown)


alexcrichton:
I don't think I'm qualified to review this (deferring to @cfallin tagged here as well), but nevertheless I wanted to ask a question about this as well:

> Choose worst values instead of best in each eclass

From the discussion [in the recent egraph issue](https://github.com/bytecodealliance/wasmtime/issues/7857) I thought we settled on the opposite conclusion, that the minimum value must be chosen. I suspect though you've already considered that and it's already accounted for here, leading me to my question. I suspect my understanding of the outcome of that discussion is probably no longer accurate, so is there a tl;dr; or why that issue won't resurface if we choose any value within an eclass? (if there's not a tl;dr; that's also totally ok, happy to go hunting elsewhere too!)

cranelift/codegen/src/egraph/elaborate.rs
```diff
@@ -248,7 +269,17 @@ impl<'a> Elaborator<'a> {
                         // is a `(cost, value)` tuple; `cost` comes first so
                         // the natural comparison works based on cost, and
                         // breaks ties based on value number.
-                        best[value] = std::cmp::min(best[x], best[y]);
+                        best[value] = if use_worst {
```
cfallin, line 272: I wonder if there' a way to encapsulte this? I realize it's a bit tricky because the order relation actually changes based on `use_worst` (reserved values come "last" when picking mins, and "first" when picking maxes). Maybe something like `pick_non_reserved_with_cmp(|x, y| x <= y)` or something like that, then `picked_non_reserved_with_cmp` takes two `(Cost, Value)` tuples and has the if-ladder?

Maybe that's just the same complexity as before, I dunno; just code-golfing a bit to see if we can make it prettier :-)

cfallin (approved):
Thanks a bunch for tackling this; hopefully it helps us validate things!

cranelift/codegen/src/egraph/elaborate.rs
lpereira: We discussed this briefly while writing, and decided on this as it's more explicit and it's used only here.

jameysharp:
Yeah, that's absolutely a fair question, @alexcrichton. In [ongoing discussion](https://github.com/bytecodealliance/wasmtime/pull/7891#issuecomment-1947685813) we're currently thinking that we need to establish an invariant that it is always correct to choose any value from an eclass. We should only use the cost function to indicate a preference for some values over others, not to encode correctness constraints.

So this PR is part of checking that we've done that correctly. I believe that with #7879 we currently have this invariant, but it's currently fragile because it relies on rewrite-rule authors strictly following new guidelines about using `subsume`. In #7891, we've been discussing how to better enforce the invariant.

jameysharp:
I've been running the cranelift-fuzzgen on this branch locally for over an hour without finding any failures, so that's a good start.

Also I forced `use_worst` to `true` in the cost function and ran the filetests. Several egraph tests and precise-output compile tests failed due to selecting longer sequences of instructions, which is what I would hope would happen. Other than that, everything passed, which is also reassuring.

PR #7974, opened 2024-02-21
winch: Overhaul the internal ABI

saulecabrera:
This change overhauls Winch's ABI. This means that as part of this change,  the default ABI now closely resembles Cranelift's ABI, particularly on the treatment of the VMContext. This change also fixes many wrong assumptions about trampolines, which are tied to how the previous ABI operated.

The main motivation behind this change is:

* To make it easier to integrate Winch-generated functions with Wasmtime
* Fix fuzz bugs related to imports
* Solidify the implementation regarding the usage of a pinned register to hold the VMContext value throughout the lifetime of a function.

The previous implementation had the following characteristics, and wrong assumptions:
* Assumed that internal functions don't receive a caller or callee VMContexts as parameters.
* Worked correctly in the following scenarios: 
    * `Wasm -> Native`: since we can explicitly load the caller and callee `VMContext`, because we're calling a native import. 
    * `(Native, Array) -> Wasm`: because the native signatures define a tuple of  `VMContext` as arguments.
* It didn't work in the following scenario:
   * `Wasm -> Wasm`: When calling imports from another WebAssembly instance (via direct call or `call_indirect`. The previous implementation wrongly assumed that there should be a trampoline in this case, but there isn't. The code was generated by the same compiler, so the same ABI should be used in both functions. 

This change introduces the following changes, which fix the previous assumptions and bugs:

* All internal functions declare a two extra pointer-sized parameters, which will hold the callee and caller `VMContext`s
* Use a pinned register that will be considered live through the lifetime of the function instead of pinning it at the trampoline level. The pinning explicitlly happens when entering the function body and no other assumptions are made from there on.
* Introduce the concept of special `ContextArgs` for function calls. This enum holds metadata about which context arguments are needed depending on the callee. The previous implementation of introducing register values at arbitrary locations in the value stack conflicts with the stack ordering principle which states that older values must *always* precede newer values. So we can't insert a register, because if a spill happens the order of the values will be wrong. 

Finally, given that this change also enables the `imports.wast` test suite, it also includes a fix to `global.{get, set}` instructions which didn't account entirely for imported globals.


winch/codegen/src/abi/mod.rs
```diff
@@ -60,6 +64,84 @@ pub(super) enum ParamsOrReturns {
     Returns,
 }
 
+/// Macro to get the pinned register holding the [VMContext].
+macro_rules! vmctx {
```
saulecabrera, line 68: This change is fairly large already, so I opted to just introduce this macro and use it in a couple of places that are updated by this change. My plan is to follow up with a general replace after landing this change (but I can do so here too if preferred).

winch/codegen/src/abi/mod.rs
```diff
@@ -60,6 +64,84 @@ pub(super) enum ParamsOrReturns {
...
+/// * More than one values are returned via a return pointer.
+/// These variations look like:
+///
+/// Single return value.
+///
+/// ```ignore
+/// unsafe extern "C" fn(
+///     callee_vmctx: *mut VMOpaqueContext,
+///     caller_vmctx: *mut VMOpaqueContext,
+///     // rest of paramters
```
cfallin, line 122: s/paramters/parameters/

winch/codegen/src/frame/mod.rs
```diff
@@ -101,49 +101,87 @@ impl Frame {
...
+        let stack_align = <A as ABI>::stack_align();
+        let defined_locals_end = align_to(
+            defined_locals_start + defined_locals.stack_size,
+            stack_align as u32,
+        );
 
+        // Handle the results base slot for multi value returns.
         let (results_base_slot, locals_size) = if sig.params.has_retptr() {
             match sig.params.unwrap_results_area_operand() {
+                // If the results operand is a stack argument, ensure the the
```
cfallin, line 113: s/the the/the/

winch/codegen/src/frame/mod.rs
```diff
@@ -101,49 +101,87 @@ impl Frame {
...
+
+    /// Get the [LocalSlot] for a frame local.
+    /// This method doesn't make any asumptions about the local index passed in,
+    /// and simply delegates the [LocalSlot] retrieval to the underlying locals
+    /// vector.
+    ///
+    /// # Panics
+    /// This method panics if the index is not associated to a valid WebAssembly
+    /// local.
+    pub fn get_local_unchecked(&self, index: usize) -> &LocalSlot {
```
cfallin, line 181: Rather than `unchecked`, which to me implies lack of a bounds check or similar, can we name this in a way that indicates the index-space is shifted relative to above? E.g. perhaps the above `get_local` becomes `get_wasm_local` and `get_local_unchecked` becomes `get_raw_local`?

Alternately, avoid exposing `get_local_unchecked` at all and have two separate methods `get_callee_local()` / `get_caller_local()`?

winch/codegen/src/trampoline.rs
```diff
@@ -601,6 +528,9 @@ where
         callee_sig
             .params_without_retptr()
             .iter()
+            // Skip the first two arguments, which are the callee and caller
+            // VMContext pointers.
+            .skip(2)
```
cfallin, line 533: here and perhaps elsewhere, can we have a constant (maybe an associated constant of `ContextArgs`) indicating how many args we expect? Basically it'd be nice to tie this to other definitions so that we don't miss it if we ever add, e.g., a third context arg.

cfallin (approved):
Overall looks great, thanks! Just a few nits below.

saulecabrera:
Oh bad rebase -- sorry about that. Fixing it now.

winch/codegen/src/abi/mod.rs
saulecabrera: Fixed, thanks!

winch/codegen/src/frame/mod.rs
saulecabrera: Fixed!

winch/codegen/src/frame/mod.rs
saulecabrera: Thanks for the suggestion -- I ended up with `get_frame_local` and `get_wasm_local` to be able to distinguish one from another.

PR #7975, opened 2024-02-22
Wasmtime: Add a `gc` cargo feature

fitzgen:
This controls whether support for `ExternRef` and its associated deferred, reference-counting garbage collector is enabled at compile time or not. It will also be used for similarly for Wasmtime's full Wasm GC support as that gets added.


fitzgen:
I didn't add a smoke test build without the `gc` cargo feature to CI, but I can do that if we decide we can afford the CI time.

crates/wasmtime/src/runtime/values.rs
```diff
@@ -689,13 +702,25 @@ impl Ref {
...
             },
+
+            #[cfg(not(feature = "gc"))]
+            (Ref::Extern(e), HeapType::Extern) => match e {
+                None => bail!(
+                    "support for externref tables disabled at compile time \
+                     because the `gc` cargo feature was not enabled"
+                ),
+                Some(e) => match e._inner {},
+            },
```
alexcrichton, line 722: I realize that this is what I recommended, having an uninhabited public type, but I'm second guessing myself now. This feels like it's pushing a lot of `#[cfg]` noise into the consumers of externref.

Could the methods on `ExternRef`, at least used by Wasmtime itself, be preserved? That way only `ExternRef` itself needs `#[cfg]` and consumers don't have to mind whether it's defined or not.

crates/wasmtime/src/runtime/store.rs
```diff
@@ -794,6 +796,9 @@ impl<T> Store<T> {
     /// Note that it is not required to actively call this function. GC will
     /// automatically happen when internal buffers fill up. This is provided if
     /// fine-grained control over the GC is desired.
+    ///
+    /// This method is only available when the `gc` Cargo feature is enabled.
```
alexcrichton, line 800: You can actually skip this documentation as rustdoc will auto-generate it with the right annotations. Elsewhere in the codebase we've got `#[cfg_attr(docsrs, doc(cfg(feature = "async")))]`, so could all public methods gated behind the `gc` feature have a similar annotation?

crates/wasmtime/src/runtime/store.rs
```diff
@@ -1523,7 +1535,8 @@ impl StoreOpaque {
         &self.runtime_limits as *const VMRuntimeLimits as *mut VMRuntimeLimits
     }
 
-    pub unsafe fn insert_vmexternref_without_gc(&mut self, r: VMExternRef) {
+    #[cfg(feature = "gc")]
```
alexcrichton, line 1538: In the interest of reducing `#[cfg]` as much as possible (or at least centralizing it) could this and other helper methods be preserved here? That way in `wasmtime-runtime` itself it would have the `match x {}` to assert unreachability (or some method on `VMExternRef` itself)

alexcrichton (commented):
Nice!

My main comments below are about perhaps shifting around `#[cfg]` to try to reduce it. That's my main worried with these sorts of features is that it can become difficult in the limit to ensure all the `#[cfg]` annotations are aligned so I'm hoping we can try to reduce them as much as possible. More-or-less I think that `ExternRef` and probably `VMExternRef` should have the same API no matter the feature set (modulo constructors) and that way consumers don't have to `#[cfg]` which should help centralize this.

We'll also definitely want this on CI and I'd recommend putting it [around here](https://github.com/bytecodealliance/wasmtime/blob/0e98a8d54ec558724481cf963f879be69ec01795/.github/workflows/main.yml#L277-L292)

crates/wasmtime/src/runtime/store.rs
fitzgen: It requires the `doc_cfg` nightly feature for rustdoc to add that annotation, which `docs.rs` enables but local `cargo doc` invocations probably don't. Therefore, I think it is worth leaving the comment there.

crates/wasmtime/src/runtime/store.rs
fitzgen: The problem with this specific case is that we store the `VMExternRefActivationsTable` directly in the store, so if it is uninhabited, then the whole `Store` becomes uninhabited.

crates/wasmtime/src/runtime/store.rs
alexcrichton: True yeah, but that's why it's gated by `docsrs` so the docs.rs docs are the only ones that show this. I'd recommend using the `#[cfg_attr]` approach since that's what everything else does in the API docs right now, or otherwise we should remove all those and add manual docs there too

crates/wasmtime/src/runtime/ref/gc_ref.rs
```diff
@@ -0,0 +1,107 @@
...
+    where
+        T: 'static + Any + Send + Sync,
+    {
+        let inner = VMExternRef::new(value);
+        ExternRef { inner }
+    }
+
+    /// Get the underlying data for this `ExternRef`.
+    pub fn data(&self) -> &dyn Any {
+        &*self.inner
```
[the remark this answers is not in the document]
fitzgen: I actually think we want to keep whole implementations of `ExternRef` separate for GC vs not GC. Looking into the future, there are going to be a bunch more types that are going to be doing this same split, likely whole submodules of things, and I don't think it will be very nice to have all these `cfg`s for the whole submodule tree. I think it will be better to have (a) the GC-enabled submodule tree and (b) a single GC-disabled module that defines all the same things from (a) but with uninhabited types.

crates/wasmtime/src/runtime/store.rs
fitzgen: We could make it an `Option`, but then we enable the potential for accidentally setting it to `None` when GC *is* enabled. I think I'd rather leave these handful of `cfg`s in comparison to that.

fitzgen:
Okay, I've updated this PR to stub out more `VMExternRef` things and methods when GC is disabled at compile time and removed a bunch of `cfg`s from the rest of the code base. Not all of them in `store.rs`, as mentioned inline above. I think this is ready for another round of review.

crates/wasmtime/src/runtime/store.rs
fitzgen: I guess we could have a `GcOnly<T>` type like

```rust
enum GcOnly<T> {
    #[cfg(feature = "gc")]
    inner: T,
}

impl<T> GcOnly<T> {
    pub fn with<U>(&self, f: impl FnMut(&T) -> U) -> Option<U> {
        #[cfg(feature = "gc")]
        return Some(f(&self.inner));
        #[cfg(not(feature = "gc"))]
        return None;
    }
}
```

But that seems like overkill maybe?

alexcrichton: I'd personally prefer to avoid `GcOnly<T>`, I feel like those sorts of wrappers haven't worked out well in the past or require otherwise just as much `#[cfg]` traffic. For `VMExternRefActivationsTable` I'd expect that'd become an empty struct rather than uninhabited because there are no references to track?

fitzgen: Ah yeah I could make that change, good idea.

crates/wasmtime/src/runtime/ref/gc_ref.rs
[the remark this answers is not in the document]
fitzgen: They do? The only ones left throughout the code base should be things that call constructors.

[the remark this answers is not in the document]
fitzgen: (It just turns out we call constructors frequently, unfortunately)

crates/wasmtime/src/runtime/store.rs
fitzgen: Done!

crates/runtime/src/table.rs
```diff
@@ -580,7 +583,11 @@ impl Drop for Table {
     fn drop(&mut self) {
         let ty = self.element_type();
 
-        // funcref tables can skip this
+        // `funcref` tables don't need drops.
+        //
+        // This is an irrefutable pattern when the `gc` cargo feature is not
+        // enabled.
+        #[allow(irrefutable_let_patterns)]
```
alexcrichton, line 590: This is fine as-is but one trick I've done in the past is at the root of the crate:

```rust
#![cfg_attr(not(feature = "gc"), allow(irrefutable_let_patterns))]
```

or more often `unused_imports` or similar.

crates/wasmtime/src/runtime/func.rs
```diff
@@ -2023,6 +2027,9 @@ impl<T> Caller<'_, T> {
     /// Perform garbage collection of `ExternRef`s.
     ///
     /// Same as [`Store::gc`](crate::Store::gc).
+    ///
+    /// This method is only available when the `gc` cargo feature is enabled.
+    #[cfg(feature = "gc")]
```
alexcrichton, line 2032: more bits for `#[cfg_attr(docsrs, ...)]`

crates/wasmtime/src/runtime/ref/no_gc_ref.rs
```diff
@@ -0,0 +1,60 @@
...
+use std::ffi::c_void;
+use wasmtime_runtime::VMExternRef;
+
+/// Represents an opaque reference to any data within WebAssembly.
+///
+/// Due to compilation configuration, this is an uninhabited type: enable the
+/// `gc` cargo feature to properly use this type.
+#[derive(Clone, Debug)]
+pub struct ExternRef {
+    pub(crate) _inner: Uninhabited,
```
alexcrichton, line 21: Could this drop the `pub(crate)`?

crates/wasmtime/src/runtime/ref/no_gc_ref.rs
```diff
@@ -0,0 +1,60 @@
...
+}
+
+impl ExternRef {
+    pub(crate) fn from_vm_extern_ref(_inner: VMExternRef) -> Self {
+        unreachable!()
```
alexcrichton, line 26: Could this perhaps to `_inner.assert_unreachable()` or something like that? (where `VMExternRef` only has that method when the `gc` feature is disabled)

alexcrichton (approved):

PR #7978, opened 2024-02-22
s390x: Fix TLS GD relocation order

uweigand:
When emitting a call to __tls_get_offset, the instruction needs to carry two relocations, a R_390_PLT32DBL targeting __tls_get_offset and a R_390_TLS_GDCALL targeting the TLS symbol.  Specifically, the system linker expects to see these two relocation in that order.

However, the cranelift backend currently emits the relocations in reverse order - this unfortunately causes the linker to corrupt the instruction sequence when performing a TLS relaxation.

To fix this in the backend, I need support in machinst common code to emit a relocation at some offset to the instruction about to be emitted (e.g. the relocation should target two bytes into the 6-byte instruction that will be emitted next).  I've added a new routine add_reloc_at_offset to that effect.  This also allowed to simplify some existing code in the backend.

### DIFF
--- a/cranelift/filetests/filetests/isa/s390x/tls_elf.clif
+++ b/cranelift/filetests/filetests/isa/s390x/tls_elf.clif
@@ -37,7 +37,7 @@ block0(v0: i32):
 ;   .byte 0x00, 0x00
 ;   .byte 0x00, 0x00
 ;   lg %r2, 0(%r1)
-;   brasl %r14, 0x22 ; reloc_external TlsGdCall u1:0 0
+;   brasl %r14, 0x22 ; reloc_external PLTRel32Dbl %ElfTlsGetOffset 2 ; reloc_external TlsGdCall u1:0 0
 ;   ear %r3, %a0
 ;   sllg %r5, %r3, 0x20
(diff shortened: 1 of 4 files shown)


cranelift/filetests/filetests/isa/s390x/tls_elf.clif
```diff
@@ -37,7 +37,7 @@ block0(v0: i32):
 ;   .byte 0x00, 0x00
 ;   .byte 0x00, 0x00
 ;   lg %r2, 0(%r1)
-;   brasl %r14, 0x22 ; reloc_external TlsGdCall u1:0 0
+;   brasl %r14, 0x22 ; reloc_external PLTRel32Dbl %ElfTlsGetOffset 2
```
bjorn3, line 40: Shouldn't both relocs be printed here?

uweigand: Good point.  I've fixed the disassembler to print multiple relocs on a single insn if present.

abrown (approved):
Makes sense to me.

PR #7979, opened 2024-02-22
winch: Enable stack overflow checking in trampolines

elliottt:
Stack overflow checking in trampolines is slightly different than when translating arbitrary wasm functions, as the vmctx isn't necessarily in the pinned register yet. To avoid bloating the function prologue, I opted to parameterize the `MacroAssembler::check_stack` function on the register that currently contains the `vmctx`. This change allows the trampolines to pass the current location of `vmctx` instead of forcing it to live in the pinned register.


saulecabrera (approved):
Looks great to me, thanks!

PR #7982, opened 2024-02-22
winch: Refactor the prologue and epilogue interface

elliottt:
Rework the winch prologue and epilogue definitions as default implementations in the `MacroAssembler` trait, naming the common sequence of frame setup, check stack, handle clobbers, and reserve space for locals. This change also allows us to remove the special prologue and epilogue functions from the trampoline module, and rely instead on the same function used by the normal function compilation path.

As I moved the pinning of the vmctx register to after the prologue, this change had a pretty noisy effect on the filetests.

### DIFF
--- a/winch/codegen/src/trampoline.rs
+++ b/winch/codegen/src/trampoline.rs
@@ -102,7 +102,7 @@ where
         let (vmctx, caller_vmctx) = Self::callee_and_caller_vmctx(&array_sig.params)?;
         let (dst_callee_vmctx, dst_caller_vmctx) = Self::callee_and_caller_vmctx(&wasm_sig.params)?;
 
-        self.masm.prologue(caller_vmctx, &self.callee_saved_regs, 0);
+        self.masm.prologue(caller_vmctx, &self.callee_saved_regs);
 
         self.masm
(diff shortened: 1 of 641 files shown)


winch/codegen/src/trampoline.rs
```diff
@@ -230,7 +231,9 @@ where
         if wasm_sig.has_stack_results() {
             self.masm.free_stack(wasm_sig.results.size());
         }
-        self.epilogue_with_callee_saved_restore(spill_size);
+
+        self.masm.free_stack(spill_size);
+        self.masm.epilogue(&self.callee_saved_regs, 0);
```
saulecabrera, line 236: Could we do `self.masm.epilogue(&self.callee_saved_regs, spill_size)` here?

saulecabrera (approved):
One minor question. Else looks reasonable to me, thanks!

winch/codegen/src/trampoline.rs
elliottt: We could, but at that point i wonder if we should move the stack allocation out of  the prologue and epilogue. What do you think about not including the reserve/free calls?

saulecabrera: Yeah, that makes sense to me as well.

PR #7984, opened 2024-02-22
Remove the `WASI` submodule

alexcrichton:
Historically the `WASI` submodule in this repository was used to pull the sources of the `*.witx` files for preview1. This has never been used by the `wasmtime-wasi` crate (which has its own copy in the `crates/wasi/witx` folder) and was only ever used by `wasi-common`. This submodule also served as a location for the `witx` crate itself.

Neither of these are really needed any more as the `*.witx` files are unlikely to ever really change again. This commit removes the submodule entirely, as well as the `path` dependency on `witx`, and copies the `*.witx` files in the same manner as the `wasmtime-wasi` crate.


sunfishcode (approved):
fewer submodules woo! :partying_face:

pchickey:
wasmtime-wasi does use the witxs and wiggle to generate out its preview 1 module interface, but I agree the submodule should go :)

PR #7986, opened 2024-02-23
Split out `TypeRegistry`'s open-coded slab arena into into a reusable type

fitzgen:
First of all, it is just a nice separation of concerns.

Second of all, as I design the GC rooting APIs for Wasmtime's upcoming Wasm GC support, I want this same thing and I'd rather not open code it multiple times.

### DIFF
--- a/crates/slab/src/lib.rs
+++ b/crates/slab/src/lib.rs
@@ -1,0 +1,450 @@
+//! A very simple, uniformly-typed slab arena that supports deallocation and
+//! reusing deallocated entries' space.
+//!
+//! The free list of vacant entries in the slab are stored inline in the slab's
+//! existing storage.
+//!
+//! # Example
+//!
+//! ```
+//! use wasmtime_slab::{Id, Slab};
+//!
+//! let mut slab = Slab::new();
+//!
+//! // Insert some values into the slab.
+//! let rza = slab.alloc("Robert Fitzgerald Diggs");
+//! let gza = slab.alloc("Gary Grice");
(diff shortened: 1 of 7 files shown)


crates/slab/src/lib.rs
```diff
@@ -0,0 +1,450 @@
...
+    /// A vacant entry.
+    Free {
+        /// A link in the slab's free list, pointing to the next free entry, if
+        /// any.
+        next_free: Option<EntryIndex>,
+    },
+}
+
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
+struct EntryIndex(NonZeroU32);
```
elliottt, line 191: Does this use `NonZeroU32` to ensure that `Option<Id>` takes up four bytes?

elliottt (approved):
This seems good to me! We could probably use this for the wasi table in wasmtime-wasi as well, as it looks pretty similar in its implementation :)

crates/slab/src/lib.rs
fitzgen: Yep, and this also aligns with how `VMSharedTypeIndex` reserves `u32::MAX` for its default/null value.

fitzgen:
> We could probably use this for the `ResourceTable` as well, as it looks pretty similar in its implementation :)

I believe Alex added a generation to that table, so it would need to do the layering described in the docs. Might make sense to add a generic implementation of that layering to this crate if we make `ResourceTable` use this as an implementation detail.

PR #7988, opened 2024-02-23
Update outdated testing instructions

douglasjose:
This is a minor update to the documentation. In includes the following changes:

- Add instructions to install `wasm32-wasi` and `wasm32-unknown-unknown` targets - the `test-programs-artifacts` dev-dependency won't compile without them.
- Remove references to the `test-programs/test_programs` feature that doesn't seem to exist.
- Replace the `--all` flag as the [cargo-test documentation](https://doc.rust-lang.org/cargo/commands/cargo-test.html#package-selection) marks it as a "deprecated alias for `--workspace`".


alexcrichton (approved):
Thanks for this!

PR #7993, opened 2024-02-23
Winch: Use Option<u8> for register indexes

jeffcharles:
Returning an `Option<u8>` to track the register index allows us to support parameter lists that are longer than 255 since we can return `None` instead of overflowing when we hit the the maximum value for `u8`. It looks like v8 supports 1000 params.

### DIFF
--- a/winch/codegen/src/isa/x64/abi.rs
+++ b/winch/codegen/src/isa/x64/abi.rs
@@ -16,9 +16,9 @@ use wasmtime_environ::{WasmHeapType, WasmValType};
 struct RegIndexEnv {
     /// General purpose register index or the field used for absolute
     /// counts.
-    gpr_or_absolute_count: u16,
+    gpr_or_absolute_count: u8,
     /// Floating point register index.
-    fpr: u16,
+    fpr: u8,
     /// Whether the count should be absolute rather than per register class.
     /// When this field is true, only the `gpr_or_absolute_count` field is
     /// incremented.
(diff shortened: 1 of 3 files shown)


winch/codegen/src/isa/x64/abi.rs
saulecabrera, line 1: Could we also update the `aarch64` abi given that we're here? It still uses `u8`.

jeffcharles: I noticed the `RegIndexEnv` implementation in `aarch64` returns `Option<u8>` with a value of `None` if the index is incremented past the `limit` field and the `limit` field defaults to `8`. There is one place where we call `set_limit` with a value of `1`. So as far as I can tell, if we had over 8 parameters, the increment function would no longer be called so, with more than 255 parameters, an overflow would not happen like was the case with the x86 implementation.

Does it still make sense to use a `u16` instead of a `u8`?

saulecabrera: Yeah good point -- thinking a bit more about this, I think that the root of the issue is the lack of limits for each register class. Perhaps instead of widening the type to `u16` we should add an upper-bound and make `increment` return `None`. The upper-bound could be either:

* A `limit` like in the case of `aarch64`, which could be derived from the calling convention.
* Using a checked add in `increment` (`u8::MAX` is good enough to represent the register arguments)

fitzgen:
Note that we follow the JS embedding limits in Wasmtime for maximum compatibility with Web engines: https://webassembly.github.io/spec/js-api/index.html#limits

saulecabrera:
Thanks for the heads up @fitzgen; Winch is already compliant with those limits (via `wasmparser`), the limits that we're discussing are for ISA register counts.

PR #7994, opened 2024-02-24
Make `component::Linker` semver-aware

### LINKED ISSUE #7860: Automatically handle semver-compatibility in `wasmtime::component::Linker`
This was discussed this past week at the BA plumbers summit and the conclusion was that part of the WASI release process will be the expectation that runtimes can update WASI and won't need to carry weighty "old" implementation of prior versions of WASI. For example updating from 0.2.1 to 0.2.0 should mean that only one copy of the 0.2.1 WITs are required and only one host implementation is required. Additionally it was concluded that it should be possible to load a module which imports 0.2.0 into a runtime that has updated to 0.2.1, and additionally it should be possible to be forwards compatible as well. Any APIs defined in 0.2.0 but imported at 0.2.1 should still be possible to run on a runtime that only supports 0.2.0. This means, for example, that guest languages can update to 0.2.1 and so long as no new functionality is used from 0.2.1 they'll still run on all 0.2.0 runtimes.

I believe the best way to implement this in Wasmtime will be more fancy logic in the `wasmtime::component::Linker` type. More-or-less this would involve updating the string used to register imports to chop off the semver-compatible parts. For example `foo:bar/baz@0.2.0` would be inserted as `foo:bar/baz@0.2`. Additionally `foo:bar/baz@1.2.0` would become `foo:bar/baz@1`. That lookup string should be enough for both forwards and backwards compatibility. We'll still be able to rely on normal type-checking as well so this in theory shouldn't cause runtime issues.

This issue is intended to track this implementation change to `wasmtime::component::Linker`.

alexcrichton:
This commit is an implementation of #7860 for Wasmtime where `wasmtime::component::Linker` is now "semver aware". This means that it assumes that hosts are always managing WIT interfaces in a semver-aare fashion meaning that semver-compatible upgrade exclusively adds functionality. This neatly fits into the idea of subtyping at the instance-level where if a binary built against 0.2.0 only requests a subset of functionality from a runtime that provides 0.2.1, that should work just fine.

Specifically what this change does is:

* For all names inserted into a `Linker` there might also be a "semver compatible name" which is registered as well. For example `..@1.0.0` is also registered as `..@1`.

* Semver-compatible names are only provided for versions without a prerelease and with either a nonzero major or minor version number.

* When looking up an item in the linker if no exact match is found then if a semver-compatible-name is available for the lookup key then that's consulted as well.

This semantically means that if a components imports WASI 0.2.0 then a runtime which only provides WASI 0.2.1 will be able to instantiate the component. Furthermore if a component imports WASI 0.2.1 but only imports the subset of WASI that was available in 0.2.0 then it will be instantiable in a runtime that only supports 0.2.0.

This implementation is intended to be a crucial part of the evolution to WASI to make it more seamless to upgrade WASI from both a host and guest perspective. This no longer requires everyone to upgrade to the same version all at the same time but instead decouples the upgrade schedules.

Closes #7860

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3224,6 +3224,7 @@ dependencies = [
  "paste",
  "rayon",
  "rustix",
+ "semver",
  "serde",
  "serde_derive",
  "serde_json",
--- a/crates/wasmtime/Cargo.toml
+++ b/crates/wasmtime/Cargo.toml
@@ -58,6 +58,7 @@ gimli = { workspace = true }
 # Support address-to-file/line information in traps when wasm files have DWARF
 # debugging information.
 addr2line = { version = "0.21.0", default-features = false, optional = true }
+semver = { version = "1.0.17", optional = true }
 
 [target.'cfg(target_os = "windows")'.dependencies.windows-sys]
 workspace = true
@@ -156,6 +157,7 @@ component-model = [
   "dep:wasmtime-component-macro",
   "dep:wasmtime-component-util",
   "dep:encoding_rs",
+  "dep:semver",
 ]
 
 wmemcheck = ["wasmtime-runtime?/wmemcheck", "wasmtime-cranelift?/wmemcheck"]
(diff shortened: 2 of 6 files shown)


pchickey (approved):
Thanks, this is very nicely commented code

PR #7995, opened 2024-02-24
Add a "custom" platform configuration for Wasmtime

alexcrichton:
This commit leverages adds a new "platform" to Wasmtime to be supported in the `crates/runtime/src/sys` folder. This joins preexisting platforms such as Unix and Windows. The goal of this platform is to be an opt-in way to build Wasmtime for targets that don't have a predefined way to run.

The new "custom" platform requires `--cfg wasmtime_custom_platform` to be passed to the Rust compiler, for example by using `RUSTFLAGS`. This new platform bottoms out in a C API that is intended to be small and Linux-like. The C API is effectively the interface to virtual memory that Wasmtime requires. This C API is also available as a header file at `examples/min-platform/embedding/wasmtime-platform.h` (generated by `cbindgen`).

The main purpose of this is to make it easier to experiment with porting Wasmtime to new platforms. By decoupling a platform implementation from Wasmtime itself it should be possible to run these experiments out-of-tree. An example of this I've been working on is getting Wasmtime running on bare-metal with a custom kernel. This support enables defining the platform interface of the custom kernel's syscalls outside of Wasmtime.

### DIFF
--- a/crates/runtime/src/sys/custom/capi.rs
+++ b/crates/runtime/src/sys/custom/capi.rs
@@ -50,11 +50,12 @@ extern "C" {
     ///
     /// Memory can be lazily committed.
     ///
-    /// Returns the base pointer of the new mapping. Aborts the process on
-    /// failure.
+    /// Stores the base pointer of the new mapping in `ret` on success.
+    ///
+    /// Returns 0 on success and an error code on failure.
     ///
     /// Similar to `mmap(0, size, prot_flags, MAP_PRIVATE, 0, -1)` on Linux.
-    pub fn wasmtime_mmap_new(size: usize, prot_flags: u32) -> *mut u8;
+    pub fn wasmtime_mmap_new(size: usize, prot_flags: u32, ret: &mut *mut u8) -> i32;
 
     /// Remaps the virtual memory starting at `addr` going for `size` bytes to
     /// the protections specified with a new blank mapping.
@@ -63,28 +64,28 @@ extern "C" {
     /// anonymous memory are used to replace these mappings and the new area
     /// should have the protection specified by `prot_flags`.
     ///
-    /// Aborts the process on failure.
+    /// Returns 0 on success and an error code on failure.
     ///
     /// Similar to `mmap(addr, size, prot_flags, MAP_PRIVATE | MAP_FIXED, 0, -1)` on Linux.
-    pub fn wasmtime_mmap_remap(addr: *mut u8, size: usize, prot_flags: u32);
+    pub fn wasmtime_mmap_remap(addr: *mut u8, size: usize, prot_flags: u32) -> i32;
 
     /// Unmaps memory at the specified `ptr` for `size` bytes.
     ///
     /// The memory should be discarded and decommitted and should generate a
     /// segfault if accessed after this function call.
     ///
-    /// Aborts the process on failure.
+    /// Returns 0 on success and an error code on failure.
     ///
     /// Similar to `munmap` on Linux.
-    pub fn wasmtime_munmap(ptr: *mut u8, size: usize);
+    pub fn wasmtime_munmap(ptr: *mut u8, size: usize) -> i32;
 
     /// Configures the protections associated with a region of virtual memory
     /// starting at `ptr` and going to `size`.
     ///
-    /// Aborts the process on failure.
+    /// Returns 0 on success and an error code on failure.
     ///
     /// Similar to `mprotect` on Linux.
-    pub fn wasmtime_mprotect(ptr: *mut u8, size: usize, prot_flags: u32);
+    pub fn wasmtime_mprotect(ptr: *mut u8, size: usize, prot_flags: u32) -> i32;
 
     /// Returns the page size, in bytes, of the current system.
     pub fn wasmtime_page_size() -> usize;
@@ -131,13 +132,16 @@ extern "C" {
     /// The `handler` provided is a function pointer to invoke whenever a trap
     /// is encountered. The `handler` is invoked whenever a trap is caught by
     /// the system.
-    pub fn wasmtime_init_traps(handler: wasmtime_trap_handler_t);
+    ///
+    /// Returns 0 on success and an error code on failure.
+    pub fn wasmtime_init_traps(handler: wasmtime_trap_handler_t) -> i32;
 
     /// Attempts to create a new in-memory image of the `ptr`/`len` combo which
     /// can be mapped to virtual addresses in the future.
     ///
-    /// The returned `wasmtime_memory_image` pointer can be `NULL` to indicate
-    /// that an image cannot be created. The structure otherwise will later be
+    /// On successed the returned `wasmtime_memory_image` pointer is stored into `ret`.
+    /// This value stored can be `NULL` to indicate that an image cannot be
+    /// created but no failure occurred. The structure otherwise will later be
     /// deallocated with `wasmtime_memory_image_free` and
     /// `wasmtime_memory_image_map_at` will be used to map the image into new
     /// regions of the address space.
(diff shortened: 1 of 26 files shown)


examples/min-platform/embedding/wasmtime-platform.h
```diff
@@ -0,0 +1,203 @@
...
+
+/**
+ * Attempts to create a new in-memory image of the `ptr`/`len` combo which
+ * can be mapped to virtual addresses in the future. The returned
+ * `wasmtime_memory_image` pointer can be `NULL` to indicate that an image
+ * cannot be created. The structure otherwise will later be deallocated
+ * with `wasmtime_memory_image_free` and `wasmtime_memory_image_map_at`
+ * will be used to map the image into new regions of the address space.
+ */
+extern struct wasmtime_memory_image *wasmtime_memory_image_new(const uint8_t *ptr, uintptr_t len);
```
bjorn3, line 168: Is `ptr` guaranteed to stay valid until the memory image is freed or does the implementation need to copy the data elsewhere?

examples/min-platform/embedding/wasmtime-platform.h
```diff
@@ -0,0 +1,203 @@
...
+ * same as the contents of what the memory image was first created with.
+ * The mappings of `addr` should be private and changes do not reflect back
+ * to `wasmtime_memory_image`.
+ *
+ * In effect this is to create a copy-on-write mapping at `addr`/`len`
+ * pointing back to the memory used by the image originally.
+ *
+ * Aborts the process on failure.
+ */
+extern void wasmtime_memory_image_map_at(struct wasmtime_memory_image *image,
```
bjorn3, line 183: I assume wasmtime_munmap is used to unmap this memory range again. Is that correct?

examples/min-platform/embedding/wasmtime-platform.h
```diff
@@ -0,0 +1,203 @@
...
+                                         uintptr_t len);
+
+/**
+ * Replaces the VM mappings at `addr` and `len` with zeros.
+ *
+ * Aborts the process on failure.
+ */
+extern void wasmtime_memory_image_remap_zeros(struct wasmtime_memory_image *image,
+                                              uint8_t *addr,
+                                              uintptr_t len);
```
bjorn3, line 194: Why does this need access to the memory image? It only modifies the VM mapping, right? And doesn't need to read the original content of the memory image.

examples/min-platform/embedding/wasmtime-platform.h
```diff
@@ -0,0 +1,203 @@
...
+
+/**
+ * Deallocates the provided `wasmtime_memory_image`.
+ */
+extern void wasmtime_memory_image_free(struct wasmtime_memory_image *image);
```
bjorn3, line 199: Are all derived VM mappings guaranteed to be unmapped before calling this?

examples/min-platform/embedding/wasmtime-platform.h
```diff
@@ -0,0 +1,203 @@
...
+ *
+ * This semantically should make it such that `addr` and `len` looks the
+ * same as the contents of what the memory image was first created with.
+ * The mappings of `addr` should be private and changes do not reflect back
+ * to `wasmtime_memory_image`.
+ *
+ * In effect this is to create a copy-on-write mapping at `addr`/`len`
+ * pointing back to the memory used by the image originally.
+ *
+ * Aborts the process on failure.
```
bjorn3, line 181: Why not return an error and have the rust side panic? That would allow recovering from the error. It would be pretty bad for a kernel to simply crash because of an allocation failure.

examples/min-platform/src/main.rs
```diff
@@ -0,0 +1,74 @@
...
+    // This is effectively a double-check of the above assesrtion and showing
+    // how running `libembedding.so` in this case requires only minimal
+    // dependencies.
+    for sym in object.symbols() {
+        if !sym.is_undefined() || sym.is_weak() {
+            continue;
+        }
+
+        match sym.name()? {
+            "" | "memmove" | "memset" | "memcmp" | "memcpy" | "bcmp" | "__tls_get_addr" => {}
```
bjorn3, line 34: What is the empty symbol used for? Is it for the `STT_FILE` symbols? If so skipping `SymbolKind::File` above makes more sense I think.

examples/min-platform/embedding/wasmtime-platform.h
alexcrichton: Correct, I'll expand the docs here

examples/min-platform/embedding/wasmtime-platform.h
alexcrichton: I was mostly just reflecting the existing Wasmtime API, which has not been meticulously thought out, onto the C API. You're right though that this entire function I think is not necessary, and I've replaced it with a call to `wasmtime_mmap_remap`.

examples/min-platform/embedding/wasmtime-platform.h
alexcrichton: Not currently, no, I'll expand the documentation.

examples/min-platform/src/main.rs
alexcrichton: Ah I didn't question it much, but turns out it's `SymbolKind::Null` so I filtered that out to remove this case.

crates/runtime/src/sys/custom/capi.rs
```diff
@@ -0,0 +1,168 @@
...
+/// * `ip` - the instruction pointer at the time of the trap.
+/// * `fp` - the frame pointer register's value at the time of the trap.
+/// * `has_faulting_addr` - whether this trap is associated with an access
+///   violation (e.g. a segfault) meaning memory was accessed when it shouldn't
+///   be. If this is `true` then the next parameter is filled in.
+/// * `faulting_addr` - if `has_faulting_addr` is true then this is the address
+///   that was attempted to be accessed. Otherwise this value is not used.
+///
+/// If this function returns then the trap was not handled. This probably means
+/// that a fatal exception happened and the process should be aborted.
```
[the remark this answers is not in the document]
alexcrichton: Sort of, but also sort of not. I've expanded the words here to describe that the meaning of an unhandled trap is embedder-specific. Basically if Wasmtime doesn't handle a `SIGSEGV`, for example, it's up to the application of what to do next. Wasmtime can't know whether that is surely an "abort the process" kind of exception.

crates/runtime/src/sys/custom/capi.rs
```diff
@@ -0,0 +1,168 @@
...
+///
+/// This function may not return as it may invoke `wasmtime_longjmp` if a wasm
+/// trap is detected.
+pub type wasmtime_trap_handler_t =
+    extern "C" fn(ip: usize, fp: usize, has_faulting_addr: bool, faulting_addr: usize);
+
+/// Abstract pointer type used in the `wasmtime_memory_image_*` APIs which
+/// is defined by the embedder.
+pub enum wasmtime_memory_image {}
```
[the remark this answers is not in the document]
alexcrichton: Unfortunately that recommendation doesn't generate the right C header bindings.

I'm also not aware of any compiler issues using `*mut Uninhabited`, which is what these bindings are doing. While I know `&mut Uninhabited` is bad, the former is not plagued with the same issues as far as I'm aware.

crates/runtime/src/sys/custom/capi.rs
```diff
@@ -0,0 +1,168 @@
...
+    /// the system.
+    pub fn wasmtime_init_traps(handler: wasmtime_trap_handler_t);
+
+    /// Attempts to create a new in-memory image of the `ptr`/`len` combo which
+    /// can be mapped to virtual addresses in the future. The returned
+    /// `wasmtime_memory_image` pointer can be `NULL` to indicate that an image
+    /// cannot be created. The structure otherwise will later be deallocated
+    /// with `wasmtime_memory_image_free` and `wasmtime_memory_image_map_at`
+    /// will be used to map the image into new regions of the address space.
+    pub fn wasmtime_memory_image_new(ptr: *const u8, len: usize) -> *mut wasmtime_memory_image;
```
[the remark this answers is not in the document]
alexcrichton: Yes they're both guaranteed to be page aligned, updated docs to match.

> Is this stuff visible at all to Wasmtime?

Could you elaborate on this? (e.g. what you mean by "stuff")

examples/min-platform/embedding/wasmtime-platform.h
alexcrichton: I'd be tempted to say the embedding should crash in an embedding-defined way if possible to get better stack frames/debugging/etc. This API isn't intended to be set in stone until the end of time so I'm imagining it can change in the future to returning a result. In the meantime though whether the embedder aborts or Wasmtime aborts seems six-to-one-half-dozen-or-the-other to me

bjorn3: The embedder could return an error bringing down just a single wasm instance without crashing the entire process.

alexcrichton: That's a good point yeah, and it's also true that `Result` is already in all these locations, so you've convinced me and I'll remove the assumed aborts here

alexcrichton:
Ok I've updated to make the C API fallible and Wasmtime propagates the error codes (thanks for pushing back on that @bjorn3!)

PR #6705, opened 2023-07-07
Update WebAssembly testsuite

abrown:
This PR updates the WebAssembly testsuite to use all of the latest tests. It highlights four failures, I think:
- `elem.wast`
- `function_references/data.wast`
- `function_references/elem.wast`
- `function_references/global.wast`

Not sure how critical these are to fix but I thought I would point these out in a draft PR.

PR #7996, opened 2024-02-24
Support `global.get` in more constant expressions

alexcrichton:
This commit updates Wasmtime to support `global.get` in constant expressions when located in table initializers and element segments. Pre-reference-types this never came up because there was no valid `global.get` that would typecheck. After the reference-types proposal landed however this became possible but Wasmtime did not support it. This was surfaced in #6705 when the spec test suite was updated and has a new test that exercises this functionality.

This commit both updates the spec test suite and additionally adds support for this new form of element segment and table initialization expression.

The fact that Wasmtime hasn't supported this until now also means that we have a gap in our fuzz-testing infrastructure. The `wasm-smith` generator is being updated in bytecodealliance/wasm-tools#1426 to generate modules with this particular feature and I've tested that with that PR fuzzing here eventually generates an error before this PR.

Closes #6705

### DIFF
--- a/crates/environ/src/module.rs
+++ b/crates/environ/src/module.rs
@@ -459,7 +459,7 @@ impl ModuleTranslation<'_> {
             // Get the end of this segment. If out-of-bounds, or too
             // large for our dense table representation, then skip the
             // segment.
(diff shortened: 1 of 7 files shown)


fitzgen (approved):
Thanks!

PR #8005, opened 2024-02-28
cranelift: Fix ireduce rules (fixes #7999)

### LINKED ISSUE #7999: Cranelift: Misoptimization of imul + ireduce
### `.clif` Test Case

```
set opt_level=speed
target x86_64

function u0:11(i8) -> i8 system_v {
block0(v0: i8):
    v1 = uextend.i64 v0
    v2 = imul_imm v1, 256
    v3 = ireduce.i8 v2
    return v3
}
```

### Steps to Reproduce

* Compile reproducing test case with `opt_level=speed`.

### Expected Results

The function unconditionally returns 0.

### Actual Results

The function gets optimized to returning the argument:

```
function u0:11(i8) -> i8 system_v {
block0(v0: i8):
    return v0
}
```

### Versions and Environment

Cranelift version or commit: 0.105 and 36fb62ca3bea7dd1456f12ead03084e2ceb48cda

Operating system: N/A

Architecture: x86_64

### Extra Info

Found by @cbeuw in https://github.com/rust-lang/rustc_codegen_cranelift/issues/1460. Replacing `i8` with `i32` returns the correct result.

jameysharp:
We had two optimization rules which started off like this:

(rule (simplify (ireduce smallty val@(binary_op _ op x y)))
      (if-let _ (reducible_modular_op val))
      ...)

This was intended to check that `x` and `y` came from an instruction which not only was a binary op but also matched `reducible_modular_op`.

Unfortunately, both `binary_op` and `reducible_modular_op` were multi-terms.
- So `binary_op` would search the eclass rooted at `val` to find each instruction that uses a binary operator.
- Then `reducible_modular_op` would search the entire eclass again to find an instruction which matched its criteria.

Nothing ensured that both searches would find the same instruction.

The reason these rules were written this way was because they had additional guards (`will_simplify_with_ireduce`) which made them fairly complex, and it seemed desirable to not have to copy those guards for every operator where we wanted to apply this optimization.

However, we've decided that checking whether the rule is actually an improvement is not desirable. In general, that should be the job of the cost function. Blindly adding equivalent expressions gives us more opportunities for other rules to fire, and we have global recursion and growth limits to keep the process from going too wild.

As a result, we can just delete those guards. That allows us to write the rules in a more straightforward way.

cc: @elliottt @cfallin @lpereira

### DIFF
--- a/cranelift/codegen/src/opts/extends.isle
+++ b/cranelift/codegen/src/opts/extends.isle
@@ -75,50 +75,17 @@
 (rule (simplify (bxor bigty (uextend _ x@(value_type smallty)) (uextend _ y@(value_type smallty))))
       (uextend bigty (bxor smallty x y)))
 
-;; Matches values where `ireducing` them will not actually introduce another
-;; instruction, since other rules will collapse them with the reduction.
-(decl pure multi will_simplify_with_ireduce_rec (u8 Value) Value)
-(rule (will_simplify_with_ireduce_rec _ x@(uextend _ _)) x)
-(rule (will_simplify_with_ireduce_rec _ x@(sextend _ _)) x)
-(rule (will_simplify_with_ireduce_rec _ x@(iconst _ _)) x)
-(rule (will_simplify_with_ireduce_rec depth x@(unary_op _ _ a))
-      (if-let _ (u8_lt 0 depth))
-      (if-let _ (reducible_modular_op x))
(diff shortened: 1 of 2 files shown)


elliottt (approved):

jameysharp:
Yeah, we're all still figuring out how best to use this egraph optimizer. The best we can do is keep trying things out, and I appreciate your efforts to do that!

PR #8006, opened 2024-02-28
egraph: Ensure eclass members are all available in the same block

jameysharp:
During elaboration, we want the property that it's safe to pick any member of an eclass to implement that eclass. To establish that property, this commit defines a notion of the "available block" for every value, and enforces that all values in an eclass have the same available block.

Closes #7891 where we devised this plan.

TODO: this needs more comments, especially documenting the `available_block` structure and the linear-time filtering loop in `optimize_pure_enode`. If somebody else wants to write those I would really appreciate it.

Future work: immediately after `optimized_values.push(orig_value)`, we should try `optimized_values.sort_unstable()` and then `optimized_values.dedup()`. That avoids creating union nodes for duplicate values, which should save a little time downstream in the compiler pipeline, but causes values to get renumbered in some filetests so they need more inspection. It also means we can delete the `if optimized_value == orig_value` test in the loop afterward. But it's an O(n log n) pass over a potentially large list so we should benchmark the change too.

cc: @elliottt @cfallin @lpereira


cranelift/codegen/src/egraph.rs
```diff
@@ -266,7 +312,7 @@ where
         // Create a union of all new values with the original (or
         // maybe just one new value marked as "subsuming" the
         // original, if present.)
-        let mut union_value = orig_value;
+        let mut union_value = optimized_values.pop().unwrap();
```
fitzgen, line 315: Rather than the indirect pushing and then popping, this could be

```suggestion
        let mut union_value = optimized_values.pop().unwrap_or(orig_value);
```

Then, as a reader, I don't have to wonder if the push is going to trigger a realloc or not (probably not because it is a smallvec and we limit the number of returns from ISLE in ISLE, but I think it is possible if we return the max ISLE returns and then push `orig_vlaue`?)

cranelift/codegen/src/egraph.rs
```diff
@@ -483,6 +534,9 @@ impl<'a> EgraphPass<'a> {
         let mut effectful_gvn_map: ScopedHashMap<(Type, InstructionData), Value> =
             ScopedHashMap::new();
 
+        let mut available_block: SecondaryMap<Value, Block> =
+            SecondaryMap::with_default(Block::reserved_value());
```
fitzgen, line 538: Might be worth pre-reserving space here, since IIUC we are going to end up filling this in for every single value.

cranelift/codegen/src/egraph.rs
```diff
@@ -255,6 +284,23 @@ where
             optimized_values.len()
         );
 
+        optimized_values.push(orig_value);
+
+        let mut best_block = self.available_block[*optimized_values.last().unwrap()];
```
cfallin, line 289: This loop looks correct to me (very slick in-place compaction!) but could we have a comment describing what it's doing? It took me a little bit, with knowledge of what it's supposed to be doing, to grok it. Perhaps something like: "Remove any values from `optimized_values` that do not have the highest possible available block in the domtree." at the top, and "This value's available block is higher; all higher indices we've already scanned have lower blocks and need to be removed, so truncate after this element." in the first case, and "This value's available block is lower, and we want to skip it; remove it from the list and keep going (order does not matter so `swap_remove` is OK)." in the second?

lpereira: One of the things I'm not sure here is that, when `swap_remove()` is called, the element that was at the end of `optimized_values` is put in the `idx-th` position, but then it's never used again to lookup `this_block` for the next iteration.  Is this the desirable behavior, or am I missing something here?

cfallin: (Re: your edit) yep, indeed; I think it might be good to add a comment with the loop invariant here, for clarity. Something like: "Loop scans in reverse; all values at indices >= idx have the same available block, which is the best (highest in domtree) available block seen so far."

cfallin (approved):
LGTM wrt my comments at least; thanks very much for working on this!

cranelift/codegen/src/egraph.rs
```diff
@@ -166,6 +169,11 @@ where
             if let NewOrExistingInst::Existing(inst) = inst {
                 debug_assert_eq!(self.func.dfg.inst_results(inst).len(), 1);
                 let result = self.func.dfg.first_result(inst);
+                debug_assert_eq!(
+                    self.get_available_block(inst),
+                    self.available_block[orig_result]
+                );
```
jameysharp, line 175: This debug-assert is failing in a few of the wasm spec tests and I don't understand why yet.

In every case that's failed, the GVN map says the instruction we looked up is equivalent to an existing instruction, but the available-block for the existing instruction is the entry block, while the available-block for the new instruction is somewhere deeper in the dominator tree.

I believed that GVN would always yield the same available-block for pure instructions so this is concerning.

jameysharp (commented):
I forgot about Chris' request for some new test cases, so I guess I haven't actually addressed all review comments yet.

And I think I don't want to address Nick's request, to avoid a possible realloc, in this PR. It's an easy change but it changes a bunch of filetests unnecessarily because it means that the optimized_values and orig_value are visited in a different order.

Also, and more importantly, there's something wrong with GVN.

cranelift/codegen/src/egraph.rs
jameysharp: I'm still reasoning through this. For GVN to match, both instructions must be identical, except that each operand is only required to be in the same set in the union-find rather than actually equal. The assertion failing means that for some operand in the instruction, the two instructions consume values which are in the same union-find set but have different available-blocks.

I had been concerned about this specific situation but Chris and I reasoned that it was okay (https://github.com/bytecodealliance/wasmtime/pull/7891#issuecomment-1960531054) and now I'm confused about whether we were right and the assertion is overly strict. If so, what is the correct available-block to use here?

cfallin: I suspect what is happening here is that the instruction gets rewritten the first time it is seen and the GVN-map entry points to the rewritten result (which could be e.g. an `iconst` available everywhere), even though the instruction per its arguments is only available in one block. So I think the assertion that we want is that the available block of `orig_result` (which we're going to use as our value here) dominates the available block of `inst`, rather than is exactly the same. I think that resolves the issue?

jameysharp: That does make the assert pass in all the tests we have. I think you're correct but the reason why was obscured by the surrounding code.

First off, the GVN-hit case did not always return `orig_result`. If we were looking up a `NewOrExistingInst::Existing` instruction, `insert_pure_enode` actually returned the result of that instruction. But that case only happens when this function was called from `remove_pure_and_optimize`, which ignores the return value. The return value only matters for `make_inst_ctor`, which always passes `NewOrExistingInst::New`, in which case we were returning `orig_result`.

Therefore it's safe to change this function to always return `orig_result`, which I think is what we probably intended in the first place, and based on your comment is what you thought we were already doing. :grin:

The next observation is that if the result of `inst` never appears in the values of the `value_to_opt_value` map, then it will never get looked up in `available_block` (or in the union-find, for that matter). So we don't actually need an answer to the question of what block this instruction should be available in. In fact it's better to leave it at the "reserved" default value so we'll get assertion failures if we accidentally do look it up.

I've made these changes; could you review again when you have a chance? Hopefully CI will pass now too.

cfallin (approved):
Changes LGTM. OK to merge with a test-case added that exercises the "auto-subsume" behavior (or if you've seen that it's already covered by one of the existing tests, note that here and that's fine too). Thanks!

cfallin:
(Or, I guess that could be slightly tricky with our current rules around subsume; OK to punt test to followup removal of the `subsume`s if you like.)

Issue #5032, opened 2022-10-07
Implement the WebAssembly GC Proposal


Srinivasa314:
#### Feature
Implement the [WebAssembly GC proposal](https://github.com/WebAssembly/gc)

#### Benefit
Most modern languages use GC. The current way to do it is on top of linear memory. However there are disadvantages to this approach

- The WebAssembly stack cannot be scanned for GC roots. As of now a shadow stack is usually used but it is not ideal.
- Webassembly modules of GC languages are huge as they must include the GC
- GCs implemented under WASM are less efficient and slower 

#### Implementation

Wasmtime already supports stack maps and stack scanning but the rest has to be implemnted

sdeleuze:
FYI we are now able to run a [Kotlin/Wasm HTTP server via KoWasm on Node.js](https://twitter.com/sdeleuze/status/1629930949944147968) and I would be happy to leverage Wasmtime to run it as soon as it will get GC support.

This would be a really good fit since we would like to leverage WASI Preview2+ and Wasm Component Model asap.

See also my blog post [The huge potential of Kotlin/Wasm](https://seb.deleuze.fr/the-huge-potential-of-kotlin-wasm/) for more context.

fitzgen:
Heads up: I am working on an RFC to work out the design of our Wasm GC support. I expect to start implementing Wasm GC once that RFC reaches consensus and merges and I've finished implementing Wasm tail calls.

fitzgen:
RFC is up: https://github.com/bytecodealliance/rfcs/pull/31

tw-atroehrsm:
hey there! any updates on this front? Really looking forward to use Kotlin/WASM with this runtime. 
For now only V8 is feasible, as it includes a GC and has implemented the proposal

fitzgen:
There is work on parsing and validation, and refactors setting the stage for these things, happening in the `wasm-tools` repository, since we need to be able to parse and validate Wasm before we can compile and run it.

fitzgen:
@smndtrl since https://github.com/bytecodealliance/wasm-tools/pull/1346 merged, all GC spec tests are passing in wasm-tools.

You can look at https://github.com/bytecodealliance/rfcs/blob/main/accepted/wasm-gc.md#incremental-implementation-plan and https://docs.wasmtime.dev/contributing-implementing-wasm-proposals.html to get an idea of what is still to be done.

tw-atroehrsm:
[...]

PR #8011, opened 2024-02-28
Define garbage collection rooting APIs

fitzgen:
Rooting prevents GC objects from being collected while they are actively being used.

We have a few sometimes-conflicting goals with our GC rooting APIs:

1. Safety: It should never be possible to get a use-after-free bug because the user misused the rooting APIs, the collector "mistakenly" determined an object was unreachable and collected it, and then the user tried to access the object. This is our highest priority.

2. Moving GC: Our rooting APIs should moving collectors (such as generational and compacting collectors) where an object might get relocated after a collection and we need to update the GC root's pointer to the moved object. This means we either need cooperation and internal mutability from individual GC roots as well as the ability to enumerate all GC roots on the native Rust stack, or we need a level of indirection.

3. Performance: Our rooting APIs should generally be as low-overhead as possible. They definitely shouldn't require synchronization and locking to create, access, and drop GC roots.

4. Ergonomics: Our rooting APIs should be, if not a pleasure, then at least not a burden for users. Additionally, the API's types should be `Sync` and `Send` so that they work well with async Rust.

For example, goals (3) and (4) are in conflict when we think about how to support (2). Ideally, for ergonomics, a root would automatically unroot itself when dropped. But in the general case that requires holding a reference to the store's root set, and that root set needs to be held simultaneously by all GC roots, and they each need to mutate the set to unroot themselves. That implies `Rc<RefCell<...>>` or `Arc<Mutex<...>>`! The former makes the store and GC root types not `Send` and not `Sync`. The latter imposes synchronization and locking overhead. So we instead make GC roots indirect and require passing in a store context explicitly to unroot in the general case. This trades worse ergonomics for better performance and support for moving GC and async Rust.

Okay, with that out of the way, this module provides two flavors of rooting API. One for the common, scoped lifetime case, and another for the rare case where we really need a GC root with an arbitrary, non-LIFO/non-scoped lifetime:

1. `RootScope` and `Rooted<T>`: These are used for temporarily rooting GC objects for the duration of a scope. Upon exiting the scope, they are automatically unrooted. The internal implementation takes advantage of the LIFO property inherent in scopes, making creating and dropping `Rooted<T>`s and `RootScope`s super fast and roughly equivalent to bump allocation.

   This type is vaguely similar to V8's [`HandleScope`].

   [`HandleScope`]: https://v8.github.io/api/head/classv8_1_1HandleScope.html

   Note that `Rooted<T>` can't be statically tied to its context scope via a lifetime parameter, unfortunately, as that would allow the creation and use of only one `Rooted<T>` at a time, since the `Rooted<T>` would take a borrow of the whole context.

   This supports the common use case for rooting and provides good ergonomics.

2. `ManuallyRooted<T>`: This is the fully general rooting API used for holding onto non-LIFO GC roots with arbitrary lifetimes. However, users must manually unroot them. Failure to manually unroot a `ManuallyRooted<T>` before it is dropped will result in the GC object (and everything it transitively references) leaking for the duration of the `Store`'s lifetime.

   This type is roughly similar to SpiderMonkey's [`PersistentRooted<T>`], although they avoid the manual-unrooting with internal mutation and shared references. (Our constraints mean we can't do those things, as mentioned explained above.)

   [`PersistentRooted<T>`]: http://devdoc.net/web/developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/JSAPI_reference/JS::PersistentRooted.html

At the end of the day, both `Rooted<T>` and `ManuallyRooted<T>` are just tagged indices into the store's `RootSet`. This indirection allows working with Rust's borrowing discipline (we use `&mut Store` to represent mutable access to the GC heap) while still allowing rooted references to be moved around without tying up the whole store in borrows. Additionally, and crucially, this indirection allows us to update the *actual* GC pointers in the `RootSet` and support moving GCs (again, as mentioned above).


fitzgen:
Part of https://github.com/bytecodealliance/wasmtime/issues/5032

crates/wasmtime/src/runtime/ref/gc_ref.rs
```diff
@@ -1,52 +1,318 @@
...
+unsafe fn extend_to_store_lifetime_mut<'a, 'b, T>(
+    _store: &'a mut impl AsContextMut,
+    reference: &'b mut T,
+) -> &'a mut T
+where
+    'a: 'b,
+    T: ?Sized,
+{
+    std::mem::transmute(reference)
+}
```
alexcrichton, line 62: Personally I'm very wary of these functions in addition to the `ref_from_raw` functions in the runtime crate. Can these be removed entirely? I would like to ideally rely on natural lifetime stuff in rustc to avoid any need to do this sort of transmute.

For example there's basically no way to locally reason that this `transmute` is correct, so I need to build a total understanding of the whole GC as a result to know when it's safe and when it's not safe to use this.

crates/wasmtime/src/runtime/ref/gc_ref.rs
```diff
@@ -1,52 +1,318 @@
...
+        // `ExternRef` methods take `impl AsContext[Mut]` methods.
+        let inner = unsafe { VMExternRef::new(value) };
+        let inner = extern_ref_into_gc_ref(inner);
+
+        // Safety: `inner` is a GC reference pointing to an `externref` GC
+        // object.
+        unsafe { ManuallyRooted::new(store.as_context_mut().0, inner) }
+    }
+
+    pub(crate) fn from_vm_extern_ref(store: &mut StoreOpaque, inner: VMExternRef) -> Rooted<Self> {
```
alexcrichton, line 229: I don't think this can be a safe function because there's no guarantee that `inner` belongs to `store`, right?

crates/wasmtime/src/runtime/ref/gc_ref.rs
```diff
@@ -1,52 +1,318 @@
...
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use wasmtime::*;
+    /// # fn _foo() -> Result<()> {
+    /// let mut store = Store::<()>::default();
+    ///
+    /// // Create an `externref` wrapping a `str`.
+    /// let externref = ExternRef::new(&mut store, "hello!");
```
alexcrichton, line 174: Technically doesn't this API example leak the `externref` for the entire lifetime of `store`? If so can this perhaps show using a `RootScope` to not do that?

crates/wasmtime/src/runtime/ref/gc_ref/rooting.rs
```diff
@@ -0,0 +1,1423 @@
...
+    pub trait RootedGcRefImpl<T: GcRef> {
+        /// Get this rooted GC reference's raw `VMGcRef` out of the store's GC
+        /// root set.
+        ///
+        /// Returns `None` for objects that have since been unrooted (eg because
+        /// its associated `RootedScope` was dropped).
+        ///
+        /// Panics if this root is not associated with the given store.
+        #[allow(private_interfaces)]
+        fn get_gc_ref(&self, store: &StoreOpaque) -> Option<VMGcRef>;
```
alexcrichton, line 151: While I'm not certain that this would work, one thing worth trying to solve some of the lifetime issues I mentioned above is to change this method signature to:

```rust
        fn get_gc_ref<'a>(&self, store: &'a StoreOpaque) -> Option<&'a VMGcRef>;
```

crates/wasmtime/src/runtime/ref/gc_ref/rooting.rs
```diff
@@ -0,0 +1,1423 @@
...
+    /// * Only have `&self` methods.
+    pub unsafe trait GcRefImpl: Sized {
+        /// Transmute a `&GcRootIndex` into an `&Self`.
+        #[allow(private_interfaces)]
+        fn transmute_ref(index: &GcRootIndex) -> &Self {
+            // This is valid and safe because all implementers must be newtypes
+            // over `GcRootIndex`, which is part of the contract for
+            // implementing this trait.
+            unsafe { std::mem::transmute(index) }
+        }
```
alexcrichton, line 137: Can this not be a defaulted method? `transmute` in general is "really scary" and we should avoid it at all times if we can, and if we can't I think that it should ideally fall into the category of "should be trivial to verify this is correct". In that sense this is not trivial as I would have to audit all implementations of `GcRefImpl`. If, however, this was located on every impl of `GcRefImpl` I could take a look at the type, see it's `#[repr(transparent)]`, and conclude that it's correct.

also in the local impls I think it would be neat to do something like:

`assert!(matches!(self, Self(GcRootIndex { ... })))` to trigger an error if the structure ever changes.

crates/wasmtime/src/runtime/ref/gc_ref/rooting.rs
```diff
@@ -0,0 +1,1423 @@
...
+}
+
+impl RootSet {
+    /// Enter a LIFO rooting scope.
+    ///
+    /// Returns an integer that should be passed unmodified to `exit_lifo_scope`
+    /// when the scope is finished.
+    ///
+    /// Calls to `{enter,exit}_lifo_scope` must happen in a strict LIFO order.
+    pub(crate) fn enter_lifo_scope(&self) -> usize {
```
alexcrichton, line 353: Since this is on the hot path of calls out of wasm, perhaps `#[inline]`?

crates/wasmtime/src/runtime/ref/gc_ref/rooting.rs
```diff
@@ -0,0 +1,1423 @@
...
+        len
+    }
+
+    /// Exit a LIFO rooting scope.
+    ///
+    /// The `scope` argument must be the result of the corresponding
+    /// `enter_lifo_scope` call.
+    ///
+    /// Calls to `{enter,exit}_lifo_scope` must happen in a strict LIFO order.
+    pub(crate) fn exit_lifo_scope(&mut self, scope: usize) {
```
alexcrichton, line 365: Since this is on the wasm call hot path, could this be `#[inline]` with a "fast path" of "ok nothing changed" and then have a slow path for "ok gotta `drain`"

crates/wasmtime/src/runtime/ref/gc_ref/rooting.rs
```diff
@@ -0,0 +1,1423 @@
...
+        store: &mut StoreOpaque,
+        f: impl FnOnce(&mut StoreOpaque) -> T,
+    ) -> T {
+        let scope = store.gc_roots().enter_lifo_scope();
+        let ret = f(store);
+        store.gc_roots_mut().exit_lifo_scope(scope);
+        ret
+    }
+
+    fn unroot_gc_ref(&mut self, gc_ref: VMGcRef) {
```
alexcrichton, line 404: I think this needs to be `unsafe` since there's no guarantee `gc_ref` is valid or belongs to this store?

crates/wasmtime/src/runtime/ref/gc_ref/rooting.rs
```diff
@@ -0,0 +1,1423 @@
...
+    }
+}
+
+impl<T: GcRef> PartialEq for Rooted<T> {
+    fn eq(&self, other: &Self) -> bool {
+        self.inner == other.inner
+    }
+}
+
+impl<T: GcRef> Eq for Rooted<T> {}
```
alexcrichton, line 645: Could these be omitted for now? Deferred instead to something like `ref_eq`?

crates/wasmtime/src/runtime/ref/gc_ref/rooting.rs
```diff
@@ -0,0 +1,1423 @@
...
+    fn hash_slice<H: std::hash::Hasher>(data: &[Self], state: &mut H)
+    where
+        Self: Sized,
+    {
+        // Safety: `Rooted<T>` is a `repr(transparent)` newtype over
+        // `GcRootIndex`.
+        let data: &[GcRootIndex] = unsafe { std::mem::transmute(data) };
+        GcRootIndex::hash_slice(data, state);
+    }
+}
```
alexcrichton, line 661: Similar to above, could these become freestanding methods on `Rooted`?

crates/wasmtime/src/runtime/ref/gc_ref/rooting.rs
```diff
@@ -0,0 +1,1423 @@
...
+    T: GcRef,
+{
+    /// Construct a new manually-rooted GC root.
+    ///
+    /// # Safety
+    ///
+    /// `gc_ref` must be a valid GC reference pointing to an instance of the GC
+    /// type that `T` represents.
+    pub(crate) unsafe fn new(store: &mut StoreOpaque, gc_ref: VMGcRef) -> Self {
+        let id = store.gc_roots_mut().manually_rooted.alloc(gc_ref);
```
alexcrichton, line 1194: I commented on this elsewhere a bit, but this would be, IMO, a good location for something like:

```rust
store.debug_assert_gc_not_allowed();
```

crates/wasmtime/src/runtime/ref/gc_ref/rooting.rs
```diff
@@ -0,0 +1,1423 @@
...
+    pub fn clone(&self, mut store: impl AsContextMut) -> Self {
+        self._clone(store.as_context_mut().0)
+    }
+
+    pub(crate) fn _clone(&self, store: &mut StoreOpaque) -> Self {
+        let gc_ref = self
+            .get_gc_ref(store)
+            .expect("ManuallyRooted always has a gc ref");
+        let gc_ref = clone_root(store, gc_ref);
+        unsafe { Self::new(store, gc_ref) }
```
alexcrichton, line 1248: Sort of along the lines of my comment above, it's sort of a "bug", not practically but morally, that there's no `AutoAssertNoGc` marker here?

crates/runtime/src/externref/gc.rs
```diff
@@ -374,29 +382,63 @@ impl VMExternRef {
...
+    ///
+    /// Wildly unsafe to use with anything other than the result of a previous
+    /// `{as,into}_raw` call!
+    ///
+    /// Additionally, it is your responsibility to ensure that this raw
+    /// `VMExternRef` is still valid, or else you could have use-after-free
+    /// bugs.
+    pub unsafe fn ref_from_raw<'a>(ptr: &'a NonNull<u8>) -> &'a Self {
+        std::mem::transmute(ptr)
+    }
```
alexcrichton, line 412: I commented below on this but I want to be sure to leave a comment on this as well before I forget. More-or-less I think we should try to remove this method, I'm not a huge fan of methods whose bodies are just `transmute`

crates/wasmtime/src/runtime/ref/gc_ref.rs
```diff
@@ -1,52 +1,318 @@
...
+    /// // Access the `externref`'s host data.
+    /// let data = externref.data(&store)?;
+    /// // Dowcast it to a `&str`.
+    /// let data = data.downcast_ref::<&str>().ok_or_else(|| Error::msg("not a str"))?;
+    /// // We should have got the data we created the `externref` with!
+    /// assert_eq!(*data, "hello");
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn data<'a>(&self, store: &'a impl AsContext) -> Result<&'a (dyn Any + Send + Sync)> {
```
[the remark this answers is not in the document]
alexcrichton: Also for an example of this there's [`Memory::data`](https://docs.rs/wasmtime/latest/wasmtime/struct.Memory.html#method.data) and `data_mut` for mutable versions.

alexcrichton (commented):
At a high level this all looks good to me. I think there's work we can do to build confidence in the internals, though. While I realize we can't remove all `unsafe` here I suspect we don't need quite so many just-a-typed-`transmute` functions. Those are really difficult to reason about the safety.

Additionally I think this is definitely an area where we're going to be leaning on miri pretty heavily. Can you ensure that there's tests that run in miri doing all the various bits and bobs with the API other than actually calling in to wasm?

crates/wasmtime/src/runtime/func/typed.rs
```diff
@@ -545,18 +545,166 @@ unsafe impl WasmTy for Option<ExternRef> {
...
+
+    #[inline]
+    unsafe fn from_abi(abi: Self::Abi, store: &mut StoreOpaque) -> Self {
+        let inner = wasmtime_runtime::VMExternRef::clone_from_raw(abi)?;
+        Some(ExternRef::from_vm_extern_ref(store, inner))
+    }
+}
+
+#[cfg(feature = "gc")]
+unsafe impl WasmTy for ManuallyRooted<ExternRef> {
```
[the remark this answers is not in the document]
fitzgen: That's a good idea.

crates/wasmtime/src/runtime/func/typed.rs
```diff
@@ -545,18 +545,166 @@ unsafe impl WasmTy for Option<ExternRef> {
...
+            //
+            // * A is reclaimed because the only reference keeping it alive
+            //   was the activation table's reference (it isn't inside any
+            //   Wasm frames on the stack yet, so stack scanning and stack
+            //   maps don't increment its reference count).
+            //
+            // * We transfer control to Wasm, giving it A and B. Wasm uses
+            //   A. That's a use after free.
+            //
+            // In conclusion, to prevent uses after free, we cannot GC
```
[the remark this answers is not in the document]
fitzgen: This scope is sufficient to catch the particular instance of this larger bug that we actually hit, and which caused the writing of this comment. It was specifically that inserting the reference into the activations table (happens just below this comment) that would trigger a GC and the above bug. So this scope would catch that particular instance of the bug from happening again, but you are correct that we need to prevent GC in a larger scope than what this actually guards against. I don't think this is really so terrible, it really is the nature of debug assertions: we often can't assert the precise invariant we want to hold, but can assert something a little looser that is implied by that invariant. That said, in this case I think we can thread an `AutoAssertNoGC` through the `into_abi` method.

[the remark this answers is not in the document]
fitzgen: Heh, and it turns out we had an `AutoAssertNoGc` for the larger scope already: didn't have to adjust callers when changing `into_abi` to take an `AutoAssertNoGc` instead of a store.

crates/runtime/src/externref/gc.rs
```diff
@@ -186,13 +176,25 @@ pub(crate) struct VMExternData {
     /// Note: this field's offset must be kept in sync with
     /// `wasmtime_environ::VMOffsets::vm_extern_data_ref_count()` which is
     /// currently always zero.
-    ref_count: AtomicUsize,
+    ref_count: Cell<usize>,
```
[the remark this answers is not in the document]
fitzgen: > Also, I was expecting to see changes to the codegen for manipulating this reference count, but I don't think that's in this PR. Is that something you'd like to defer to a future PR?

Nah, that was just an oversight.

fitzgen:
@alexcrichton I think I've addressed everything in your review.

Going to start rebasing on `main`, but I think you can take another look now.

crates/runtime/src/gc.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    /// Get this GC reference as a non-null pointer.
+    pub fn as_non_null(&self) -> NonNull<u8> {
+        self.0.as_non_null()
+    }
+
+    /// Get this GC reference as a `VMExternRef`.
+    pub fn as_extern_ref(&self) -> &VMExternRef {
+        assert!(Self::ONLY_EXTERN_REF_IMPLEMENTED_YET);
+        unsafe { std::mem::transmute(self) }
```
alexcrichton, line 86: Sorry I know I'm harping on this a lot, but I think we should avoid `std::mem::transmute` as much as possible if we can given that it's got so many footguns and I always forget the whole set of footguns. Some suggestions here:

* Could this (and the one below) move to the `VMExternRef` type? That way it'll be easy to see that `VMExternRef` is a `#[repr(transparent)]`
* Could this perhaps have a debug assert that the size/align are the same?
* Instead of `std::mem::transmute`, could this do `&*(self as *const _).cast()`

I realize that it can be sort of six-to-one-half-dozen or another but I'm at least personally wary of using `transmute` unless we absolutely have to (e.g. the only way to go from `usize` to `fn()` is `transmute`, but pointers can be casted/etc)

crates/runtime/src/gc/enabled.rs
```diff
@@ -283,14 +286,12 @@ impl VMExternData {
...
-        // libstd's implementation of `Arc`, but the general gist is that we're
-        // creating a new pointer for our own thread, so there's no need to have
-        // any synchronization with orderings. The synchronization with other
-        // threads with respect to orderings happens when the pointer is sent to
-        // another thread.
-        self.ref_count.fetch_add(1, Ordering::Relaxed);
+        self.ref_count.set(self.ref_count.get() + 1);
+        log::trace!(
+            "Incrementing ref count of externref @ 0x{:p} -> {}",
+            self as *const _,
```
alexcrichton, line 292: Could this debug print be `self.0.as_ptr()` since `&self` changes enough that its pointer value probably isn't too interesting?

crates/runtime/src/gc/enabled.rs
```diff
@@ -333,30 +344,92 @@ impl VMExternRef {
...
+                    ref_count: Cell::new(1),
                     // Cast from `*mut T` to `*mut dyn Any` here.
                     value_ptr: SendSyncPtr::new(NonNull::new_unchecked(value_ptr.as_ptr())),
                 },
             );
 
             log::trace!("New externref data @ {:p}", extern_data_ptr);
-            VMExternRef(NonNull::new_unchecked(extern_data_ptr).into())
+            VMExternRef(VMGcRef::from_non_null(NonNull::new_unchecked(
+                extern_data_ptr as *mut u8,
```
alexcrichton, line 355: Whiel you're here, mind removing `as` casts where possible and replacing them with `.cast()`?

crates/runtime/src/gc/enabled.rs
```diff
@@ -333,30 +344,92 @@ impl VMExternRef {
...
+    }
+
+    /// Consume this `VMExternRef` into a raw, untyped pointer.
+    ///
+    /// # Safety
+    ///
+    /// This method forgets self, so it is possible to create a leak of the
+    /// underlying reference counted data if not used carefully.
+    ///
+    /// Use `from_raw` to recreate the `VMExternRef`.
```
alexcrichton, line 381: s/`from_raw`/`from_gc_ref`/ I think?

crates/runtime/src/gc/enabled.rs
```diff
@@ -333,30 +344,92 @@ impl VMExternRef {
...
+    ///
+    /// # Safety
+    ///
+    /// The given `gc_ref` must point to a valid `VMExternData`.
+    ///
+    /// Unlike `clone_from_raw`, this does not increment the reference count of
+    /// the underlying data. It is not safe to continue to use the pointer
+    /// passed to this function.
+    #[inline]
+    pub unsafe fn from_gc_ref(gc_ref: VMGcRef) -> Self {
```
alexcrichton, line 399: Could this an `clone_from_gc_ref` assert that vmexternref is the only implemented type?

crates/runtime/src/gc/enabled.rs
```diff
@@ -206,17 +207,15 @@ impl Drop for VMExternRef {
     fn drop(&mut self) {
```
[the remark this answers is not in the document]
alexcrichton: Oh and now I also realize that these same concerns apply to `Clone` above, that seems like a footgun that it's safe since there's no easy way to audit callers are all happening in the same thread.

crates/wasmtime/src/runtime/gc/enabled.rs
```diff
@@ -0,0 +1,391 @@
...
+    ///
+    /// When in doubt try to not use this. Instead use the safe Rust APIs of
+    /// [`TypedFunc`] and friends.
+    ///
+    /// [`Func::call_unchecked`]: crate::Func::call_unchecked
+    /// [`Func::new_unchecked`]: crate::Func::new_unchecked
+    /// [`Store`]: crate::Store
+    /// [`TypedFunc`]: crate::TypedFunc
+    /// [`ValRaw`]: crate::ValRaw
+    pub unsafe fn from_raw(
```
alexcrichton, line 361: Is it actually even possible to safely call this method? 

This only exist for the C API, and there's no rooting in the C API, so thinking about this it seems like we should remove this because it basically can't be used correctly

crates/runtime/src/gc/enabled.rs
fitzgen: This is a method on `VMExternData`, not `VMExternRef`, so `&self` is the interesting pointer value.

crates/wasmtime/src/runtime/gc/enabled.rs
fitzgen: > This only exist for the C API

We use this to convert `ValRaw`s back to `ExternRef`s, and the untyped function call path internally uses `ValRaw`s, so we can't just remove this without affecting non-C-API things.

> there's no rooting in the C API

There isn't explicit control of rooting, no, I haven't added C-API stuff for `RootScope` and `Rooted` vs `ManuallyRooted`. But we do root things in the C-API implementation nonetheless, for example a `wasm_ref_t` is a `ManuallyRooted<Ref>` under the hood and you could use this method safely in combination with that type.

crates/runtime/src/gc/enabled.rs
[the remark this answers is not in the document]
fitzgen: Yeah, it is definitely a footgun, and this isn't how I would design this from scratch. It is just the intermediate stepping stone on the way to the next PR, which introduces a proper `GcHeap` that is passed in to GC hooks/barriers which are proof of safe access to the heap and its objects (similar to what store does at the embedder API level).

I don't really know how to avoid this intermediate state without fusing this PR together with the next.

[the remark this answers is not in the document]
alexcrichton: Ok if this is just a stepping stone I think that's ok. If you're ok with it though I think it'd be best at this point to land this PR tomorrow at the earliest since the 19.0.0 branch point will happen tonight and I think it would be best to not have this pair of PRs get straddled across a release

crates/wasmtime/src/runtime/gc/enabled.rs
alexcrichton: Oh right yeah, but the untyped/unchecked path basically only exists for the C API. I'm definitely not saying that we should add a bunch of rooting-related things to the C API, but I'm wondering if we should basically forbid externref in the C API as a result of this (or at least in the `ValRaw` parts bound in the C API). 

This seems like it might be too big of a footgun to expose in the C API where you can very easily get an unrooted value and there's no way to go from that unrooted value to a rooted value, so even if you want to "do the right thing" you can't

crates/runtime/src/gc/enabled.rs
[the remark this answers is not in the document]
fitzgen: That makes sense to me.

crates/wasmtime/src/runtime/gc/enabled.rs
fitzgen: I think it does make sense to expose the rooting stuff in the C API eventually, it's just not a priority at the moment. Given that, I think we can forbid it in the C API in the meantime, yeah. How would you like to do this? Propagating errors or asserting? I think the former could be difficult/annoying because of the amount of places that are currently infallible but which would need to become fallible...

alexcrichton: Perhaps removing the `externref` field from the header (and the tag for that enum)? And then using `unreachable!()` everywhere inside of the C bindings since they shouldn't be triggerable?

fitzgen: Done.

alexcrichton (approved):

alexcrichton:
Handling the fallout for this in the wasmtime-cpp repository I think that the requirement to take a `wasmtime_context_t` as an argument for value operations invalidates the preexisting [copy constructor](https://github.com/bytecodealliance/wasmtime-cpp/blob/da579e78f799aca0a472875b7e348f74b3a04145/include/wasmtime.hh#L1900-L1904) and [copy assignment operator](https://github.com/bytecodealliance/wasmtime-cpp/blob/da579e78f799aca0a472875b7e348f74b3a04145/include/wasmtime.hh#L1918-L1925). Not an issue per se, but it appears that removing those also invalidates [this pattern](https://github.com/bytecodealliance/wasmtime-cpp/blob/da579e78f799aca0a472875b7e348f74b3a04145/examples/gcd.cc#L27) where arguments can't be constructed as `{6, 27}` any more. I'll admit I'm not C++ expert and the wall of error messages I'm looking at don't exactly explain why it's related to the copy constructor/assignment.

Anyway that's a long way of asking, do you think we're going to indefinitely want to have a context argument on these functios into the future, even with the idea of an indexed heap?

fitzgen:
> Anyway that's a long way of asking, do you think we're going to indefinitely want to have a context argument on these functios into the future, even with the idea of an indexed heap?

Yes, unfortunately, I do. At least in Rust. I could see a C/C++ specific approach to alleviating this, however...

Indexed GC heaps are orthogonal to the rooting approach here.

In order to support moving GCs, we need to either:

1. Be able to precisely identify raw (whether pointer or index) GC roots *and mutate them* after an object moves.

2. Or support pinning GC references temporarily such that they do not move while rooted. This allows us to avoid the need to update GC roots after an object moves.

Option (2) is a pretty big constraint on GC implementations. We would have to *really* bend over backwards to support that in our planned copying collector.

So focusing on option (1), there are basically two approaches we can take:

1. Indirection. Store the raw GC references in some sort of table (aka our `GcRoots`) and have the user's root type index into that table (aka our `Rooted<T>`/`ManuallyRooted<T>`). This is nice, and works well with Rust and its move semantics, because we don't need to track down where all the `Rooted<T>`s and `ManuallyRooted<T>`s are in memory and we don't need to pin them or do any sort of updates on move so that we can update their internal GC reference after an object moves. Instead, the GC refs are all in the table and we can just do a pass over the table to update the GC refs, and all the `Rooted<T>`s index into the table and get the moved GC ref the next time they do some operation on their referent. The downside is that all constructing and dropping GC roots in this scheme requires access to the table, in the general case, as you've found.

2. Intrusive lists/some other intrusive data structure. This is the approach that SpiderMonkey takes, for example. All the user's roots hold raw GC references and are additionally members of an intrusive list that is associated with their GC heap. This allows the GC to walk the intrusive list and update all rooted GC references after moving objects. But this means that you need to pin the GC root type or have move constructors that keep the intrusive list valid as the GC root is moved in memory. This works well with C++ but very much not so well with Rust, which is why we went with the first option.

So, if we wanted to make the C/C++ API a little more ergonomic, we *could* support the intrusive list option. Ideally not in the Rust embedder API, but maybe with some `doc(hidden)` stuff that is feature gated for when we are building the C API crate. FWIW, this is also fairly unsafe not just in terms of the invariants around maintianing the intrusive list, but in that right now with the table/indirect approach we only ever use indices and we bounds check their accesses into the table so "use after free"-style bugs are still memory safe, which limits how bad things can go wrong. That is not the case with the intrusive list approach.

alexcrichton:
Ok nah that all sounds good, no need to go the intrusive list route just yet, I think it's ok if things are slightly less ergonomic in the C++ bindings API unless a C++ wizard more adept than I can figure out a better solution

rockwotj:
> Not an issue per se, but it appears that removing those also invalidates [this pattern](https://github.com/bytecodealliance/wasmtime-cpp/blob/da579e78f799aca0a472875b7e348f74b3a04145/examples/gcd.cc#L27) where arguments can't be constructed as {6, 27} any more

https://github.com/bytecodealliance/wasmtime-cpp/blob/da579e78f799aca0a472875b7e348f74b3a04145/include/wasmtime.hh#L2457C32-L2458

The function you're calling takes a `std::vector` and you're passing in an initializer_list, which copies things out of the vector: https://godbolt.org/z/h844YKxYq

You'll need to use a move iterator to force moving out of the container. One of the annoying bits about C++ iterators. You could use ranges/views to make this more seemless.

As for dropping the copy constructor/assignment I think that's fine, but it can be useful to have an explicit copy function that takes a context.

alexcrichton:
Makes sense! In https://github.com/bytecodealliance/wasmtime-cpp/pull/48 I ended up adding overloaded versions for `std::initializer_list` and `std::vector` with a "base" version that takes an iterator. That being said I don't know how to best generically take an iterator in C++

rockwotj:
Added a suggestion here: https://github.com/bytecodealliance/wasmtime-cpp/pull/48/files#r1525490495

PR #8012, opened 2024-02-28
Add regression test for issue #7999

fitzgen:
Follow up to #8005 and #7999

### DIFF
--- a/cranelift/filetests/filetests/egraph/issue-7999.clif
+++ b/cranelift/filetests/filetests/egraph/issue-7999.clif
@@ -1,0 +1,12 @@
+test optimize
+set opt_level=speed
+target x86_64
+
+function u0:11(i8) -> i8 system_v {
+block0(v0: i8):
+    v1 = uextend.i64 v0
+    v2 = imul_imm v1, 256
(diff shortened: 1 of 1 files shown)


cfallin (approved):
Thanks!

cranelift/filetests/filetests/egraph/issue-7999.clif
```diff
@@ -0,0 +1,12 @@
...
+set opt_level=speed
+target x86_64
+
+function u0:11(i8) -> i8 system_v {
+block0(v0: i8):
+    v1 = uextend.i64 v0
+    v2 = imul_imm v1, 256
+    v3 = ireduce.i8 v2
+    return v3
+    ; return v10  ; v10 = 0
```
bjorn3, line 11: Without the precise-output flag set you have to add explicit [filecheck](https://github.com/bytecodealliance/filecheck) match statement like `; check: return v10  ; v10 = 0`

fitzgen: Whoops, didn't realize I was missing a `check:` here, thanks!

PR #8014, opened 2024-02-28
update capstone dependency to 0.12.0

pchickey:
this is only used for benchmarking, so the cargo vet is just an exemption which I updated to the latest version.

Closes https://github.com/bytecodealliance/wasmtime/issues/8009

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -333,9 +333,9 @@ dependencies = [
 
 [[package]]
 name = "capstone"
-version = "0.9.0"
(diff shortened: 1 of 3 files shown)


alexcrichton (approved):

alexcrichton:
@pchickey this may have gotten lost in the email but looks like CI is failing here due to some bindings needing an update

pchickey:
Thanks this fell off my notifs but it was trivial to fix.

PR #8016, opened 2024-02-28
Fix a memory leak found in fuzzing

alexcrichton:
This fixes a mistake introduced in #7996 where a `global.get` used to initialize a table did not properly drop previous items in the table causing them to leak as the reference count was not decremented.


fitzgen (approved):
Regression test in `tests/all/gc.rs`? Can use the `SetFlagOnDrop` helper.

alexcrichton:
Actually, there's another issue I've found, so I've opted for a larger refactor in https://github.com/bytecodealliance/wasmtime/pull/8018 to fix this (test included over there)

PR #8018, opened 2024-02-28
Refactor runtime table management for more type-safety

alexcrichton:
This commit is an attempt to make the mistake fixed in #8016 harder to happen again. This removes the `set_raw` helper entirely from tables and instead forces all callers to view the table as either a table of funcrefs or a table of externefs. By forcing that choice outwards instead of handling it inwards it enables dealing with a typed table in more contexts which should help naturally do the right thing.

### DIFF
--- a/crates/runtime/src/instance/allocator.rs
+++ b/crates/runtime/src/instance/allocator.rs
@@ -2,7 +2,7 @@ use crate::imports::Imports;
 use crate::instance::{Instance, InstanceHandle};
 use crate::memory::Memory;
 use crate::mpk::ProtectionKey;
-use crate::table::Table;
+use crate::table::{Table, TableElementType};
 use crate::{CompiledModuleId, ModuleRuntimeInfo, Store};
 use anyhow::{anyhow, bail, Result};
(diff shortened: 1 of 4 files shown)


crates/runtime/src/instance/allocator.rs
```diff
@@ -531,14 +531,25 @@ fn initialize_tables(instance: &mut Instance, module: &Module) -> Result<()> {
...
-                table.init_func(funcref)?;
+                let init = (0..table.size()).map(|_| funcref);
+                table.init_func(0, init)?;
             }
 
             TableInitialValue::GlobalGet(idx) => unsafe {
                 let global = instance.defined_or_imported_global_ptr(*idx);
-                let funcref = (*global).as_func_ref();
                 let table = &mut *instance.get_defined_table(table);
-                table.init_func(funcref)?;
```
alexcrichton, line 541: Relative to https://github.com/bytecodealliance/wasmtime/pull/8016 this was the other issue I found, this forgot to check whether the table was an externref table or funcref table and triggered a panic by assuming a global was a funcref.

fitzgen (approved):

PR #8026, opened 2024-02-28
Show true error reason in `wasmtime serve`

abrown:
The `wasmtime serve` command prints failed HTTP handling to `stderr` but, when the reason was due to failure inside the HTTP handler itself, the printed error message is not descriptive enough. After looking at long dumps of `guest never invoked `response-outparam::set` method`, I realized that errors inside the Tokio `task` are never propagated outwards. This change captures those errors and appends them to the printed error message.

### DIFF
--- a/src/commands/serve.rs
+++ b/src/commands/serve.rs
@@ -386,7 +386,7 @@ async fn handle_request(
     let (sender, receiver) = tokio::sync::oneshot::channel();
 
     // TODO: need to track the join handle, but don't want to block the response on it
-    tokio::task::spawn(async move {
+    let task = tokio::task::spawn(async move {
         let req_id = inner.next_req_id();
         let (mut parts, body) = req.into_parts();
 
@@ -450,7 +450,17 @@ async fn handle_request(
     match receiver.await {
         Ok(Ok(resp)) => Ok(resp),
         Ok(Err(e)) => Err(e.into()),
-        Err(_) => bail!("guest never invoked `response-outparam::set` method"),
+        Err(_) => {
+            // If the receiver has an error (`RecvError`), it will not describe
+            // the real failure in any meaningful way. Instead we retrieve and
+            // propagate the error from inside the task which should more
+            // clearly tell the user what went wrong.
+            let e = match task.await {
+                Ok(r) => r.expect_err("if the receiver has an error, the task must have failed"),
+                Err(e) => e.into(),
+            };
+            bail!("guest never invoked `response-outparam::set` method: {e:?}")
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/commands/serve.rs
```diff
@@ -386,7 +386,7 @@ async fn handle_request(
     let (sender, receiver) = tokio::sync::oneshot::channel();
 
     // TODO: need to track the join handle, but don't want to block the response on it
```
abrown, line 388: @elliottt, I guess if we adopt this PR we can remove this TODO line?

pchickey: Yes, please remove this TODO and we can land this PR

pchickey (approved):
Thanks for this fix!

src/commands/serve.rs
```diff
@@ -450,7 +450,17 @@ async fn handle_request(
     match receiver.await {
         Ok(Ok(resp)) => Ok(resp),
         Ok(Err(e)) => Err(e.into()),
-        Err(_) => bail!("guest never invoked `response-outparam::set` method"),
+        Err(_) => {
+            // If the receiver has an error (`RecvError`), it will not describe
+            // the real failure in any meaningful way. Instead we retrieve and
```
elliottt, line 455:
```suggestion
            // If the receiver has an error (`RecvError`), it only indicates that the task
            // exited before the sender was used, that the sender was dropped
            // before a response was sent. Instead we retrieve and
```

src/commands/serve.rs
```diff
@@ -450,7 +450,17 @@ async fn handle_request(
...
+            // propagate the error from inside the task which should more
+            // clearly tell the user what went wrong.
+            let e = match task.await {
```
elliottt, line 458: Do we need to bound the amount of time spent here, or are we guaranteed that the result should be available immediately because the `receiver.await` failed?

elliottt (approved):
Looks great to me, thanks for making this change @abrown!

src/commands/serve.rs
abrown: I kind of merged our two versions into 5a958a5; hopefully that still captures what you meant in this suggestion!

src/commands/serve.rs
abrown: As I looked at it, I concluded that yes, if the receiver failed it is because the task failed. But maybe I missed something or perhaps future changes will invalidate that. But in that case, `wasmtime serve` will be dumping errors to stderr anyways so it should be apparent to users that something is wrong apart from the (hypothetical) additional slowdown. And they'll probably care more about whatever caused the task to fail.

abrown: Ok, added a comment about this assumption: f09760e

PR #8027, opened 2024-02-28
Add CLI flags for instance limits

abrown:
While experimenting with `wasmtime serve`, I found it helpful to be able to configure some of the maximum limits of the various objects stored in the pooling allocator. This change surfaces some new `-O total-*` flags for controlling these limits from the command line.

### DIFF
--- a/crates/cli-flags/src/lib.rs
+++ b/crates/cli-flags/src/lib.rs
@@ -82,20 +82,25 @@ wasmtime_option_group! {
         /// copy-on-write mapping (default: yes)
         pub memory_init_cow: Option<bool>,
 
-        /// The maximum number of WebAssembly instances which can be created.
-        pub total_core_instances: Option<u32>,
+        /// The maximum number of WebAssembly instances which can be created
+        /// with the pooling allocator.
+        pub pooling_total_core_instances: Option<u32>,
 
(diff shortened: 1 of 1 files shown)


crates/cli-flags/src/lib.rs
```diff
@@ -81,6 +81,21 @@ wasmtime_option_group! {
...
+        pub total_component_instances: Option<u32>,
+
+        /// The maximum number of WebAssembly memories which can be created.
+        pub total_memories: Option<u32>,
+
+        /// The maximum number of WebAssembly tables which can be created.
+        pub total_tables: Option<u32>,
+
+        /// The maximum number of WebAssembly stacks which can be created.
+        pub total_stacks: Option<u32>,
```
abrown, line 98: Another option here, since all of these default to the same number currently, is to have a single flag `--total-pool-slots` or the like that sets all of these together. But maybe someone wants to control these in a bit more fine-grained way, so I went down this route instead.

alexcrichton (commented):
Agreed having all the knobs here is fine, the goal is to reflect the options of `Config` on the CLI and this does that well.

Mind adding `pool` or `pooling` at the front of the options though to emphasize that they're only used as configuration to the pooling allocator?

alexcrichton (approved):

PR #8032, opened 2024-02-29
Update to wit-bindgen 0.20.0

alexcrichton:
Keepin' up to date.

I don't know what's going on with `cargo vet` and the entries here. I can't run `cargo vet` on main and I'm not sure why. I ran `cargo vet regenerate audit-as-crates-io` and it flipped some flags to `false` and all of a sudden I could run it, so I stopped asking questions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -653,7 +653,7 @@ dependencies = [
  "target-lexicon",
  "thiserror",
  "toml",
- "wasmparser 0.201.0",
+ "wasmparser",
  "wat",
 ]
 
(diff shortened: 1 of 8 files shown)


pchickey (approved):
I know @elliottt was having some `cargo vet` trouble yesterday as well. I have no idea what is going on with that.

elliottt:
That's wild, when I ran the exact same command yesterday, it flipped those to `true`.

alexcrichton:
Yeah I'm not sure what's going on either, its says that the crate doesn't exist on crates.io  but it does, so it may be some sort of odd scenario we're running into...

Issue #8034, opened 2024-02-29
Wasmtime throughput low on linux systems


thomastaylor312:
### Test Case

This is the wasm file, zipped up in order to upload to GH. It is from https://github.com/sunfishcode/hello-wasi-http.git
[hello_wasi_http.wasm.zip](https://github.com/bytecodealliance/wasmtime/files/14454123/hello_wasi_http.wasm.zip)

### Steps to Reproduce

Try these steps on a linux machine and on a macos machine (preferably close to the same size):

1. Run the component with `wasmtime serve` (no additional flags)
2. Run `hey -z 10s -c 100 http://localhost:8080/`

### Expected Results

I expect the number of requests/second to be the same or greater on linux than they are on Mac

### Actual Results

On my Mac (details on OS below) that was running a bunch of other applications, I get around 20k req/s
On linux (details on OS below), I get around 4.3k req/s

### Versions and Environment

Wasmtime version or commit: 18.0.2

**Mac**
Operating system: Sonoma 14.3.1

Architecture: M1 Max (2 performance cores, 8 normal cores) and 64 GB of memory

**Linux**
Operating system: Debian Bookworm (6.1.0-18-cloud-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.76-1 (2024-02-01) x86_64 GNU/Linux)

Architecture: AMD64 (16 cores) and 64 GB of memory

This was run on a cloud VM but I also tested this on a ubuntu 20.04 amd64 server running at my house with similar performance

### Extra Info

On the linux server, I did double check my file descriptor limit had been raised and also observed that the wasmtime processes were all getting to an uninterruptible sleep state almost constantly through the whole test (which could mean nothing). Also, I did a similar test with wasmCloud and Spin, which both use wasmtime and was getting a similar drop in numbers between mac and linux. For reference, I also did some smoke tests with normal server traffic (I did a test with Caddy and with NATS) and all of them were getting easily into the 100k+ range. So this definitely seems like something on the wasmtime side.

I did see #4637 and that does explain some of the horizontal scaling issues, but I didn't expect such a drastic difference between Mac and Linux

fitzgen:
First off: are you enabling the pooling allocator? Eg `-O pooling-allocator` in the CLI. Enabling or disabling the pooling allocator is going to greatly affect requests/second.

[...]

PR #8038, opened 2024-03-01
Add commentary on advantages/disadvantages of the pooling allocator

alexcrichton:
Inspired by discussion on #8034


pchickey (approved):

alexcrichton:
I also added a commit here for commentary on the disadvantages of `memory_init_cow` with respect to extra IPIs it may generate.

PR #8040, opened 2024-03-02
winch: Experiment with callee saves

elliottt:
Move winch closer to cranelift's abi by modifying the winch calling convention to respect callee-save registers.

One optimization this will enable is to avoid spilling values in callee-save regsiters before a function call, but that's better left to future work given how noisy this change is. Additionally if winch follows the cranelift abi, we can remove the winch implementations of all trampolines, replacing them with the ones that the wasmtime-cranelift crate already generates. While that would introduce a dependency on cranelift from winch, the trampolines are all small and would likely not affect compilation performance.


elliottt:
After discussions with @cfallin, I'm going to take an alternate approach to reusing the cranelift trampolines with winch: restoring the old "save everything" calling convention in cranelift. This would mean that we could reuse the trampolines while avoiding changing winch's abi.

Issue #7588, opened 2023-11-27
Eagerly cast pointers in aarch64 relocation code to `usize` to avoid `unsafe` blocks in cranelift-jit


elliottt:
When performing relocations in cranelift-jit, there are many introductions of a variable `what` that's defined as an unsafe pointer offset. These variables are almost always cast to `usize` or `isize` before they're used, so perhaps we should instead turn the base pointer into its target numeric type eagerly, to avoid needing the `unsafe` blocks for calling the `offset` method on the pointers.

https://github.com/bytecodealliance/wasmtime/blob/cd97c9f14713d81de5453fb9861319e78a761432/cranelift/jit/src/compiled_blob.rs#L55-L57

bjorn3:
Using wrapping_offset it also an option. On CHERI you would need to use the current code to preserve the pointer capability I think, but we don't support CHERI yet anyway.

sudoHackIn:
Hi, @elliottt @bjorn3,is this issue is still actual? Unfortunately from glaziermag's pr and this comment (https://github.com/bytecodealliance/wasmtime/pull/8041#discussion_r1509922915) i cannot understand, whether pr is not correct, or bjorn3 decided that any modification(about this issue) doesn't make sense at all

PR #8041, opened 2024-03-02
Optimize aarch64 relocation code by eagerly casting pointers

glaziermag:
Hello, beginner here, impassioned by the project.  This pull request addresses issue #7588, which highlights the prevalent use of unsafe pointer offsets in the aarch64 relocation code within the cranelift-jit component. The current approach frequently casts these unsafe offsets to usize or isize types, suggesting an opportunity to improve code safety and readability by performing these casts eagerly.  Please feel free to correct any noob mistakes. I'm eager to learn and contribute more effectively!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1292,7 +1292,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4271d37baee1b8c7e4b708028c57d816cf9d2434acb33a549475f78c181f6253"
 dependencies = [
  "fallible-iterator",
- "indexmap 2.2.3",
+ "indexmap 2.2.5",
  "stable_deref_trait",
 ]
 
@@ -1314,7 +1314,7 @@ dependencies = [
  "futures-sink",
  "futures-util",
  "http",
- "indexmap 2.2.3",
+ "indexmap 2.2.5",
  "slab",
  "tokio",
  "tokio-util",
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -1292,7 +1292,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4271d37baee1b8c7e4b708028c57d816cf9d2434acb33a549475f78c181f6253"
 dependencies = [
  "fallible-iterator",
- "indexmap 2.2.3",
+ "indexmap 2.2.5",
```
bjorn3, line 1295: Please revert this commit. Dependency updates require reviewing the changes in those dependencies by a trusted reviewer, which cargo-vet enforces. Because no such review has been done, CI would fail once it runs.

cranelift/jit/src/compiled_blob.rs
```diff
@@ -38,39 +38,39 @@ impl CompiledBlob {
             match kind {
                 Reloc::Abs4 => {
                     let base = get_address(name);
-                    let what = unsafe { base.offset(isize::try_from(addend).unwrap()) };
+                    let what = base.wrapping_offset(isize::try_from(addend).unwrap());
```
bjorn3, line 41: This would loose optimizations and isn't more correct. `.offset()` is unsafe as it is UB to go out of bounds, but if the calculation goes out of bounds, this crate would produce UB anyway, except now sanitizers no longer know that going out of bounds is invalid and thus wouldn't trigger at the root cause of the UB.

glaziermag:
Thank you for your insights, @bjorn3. Based on your feedback, I understand the concerns regarding optimization and safety. I'll proceed to close this pull request.

PR #8043, opened 2024-03-04
[fix] broken links in WASI-intro.md

BlackAsLight:
This change is needed as the current links lead to a 404 page.


elliottt (approved):
Thank you!

PR #8047, opened 2024-03-05
Update object to 0.33.0

philipc:
object 0.33.0 has a breaking change for relocations which needs a larger than usual fix here. I already did this fix for testing purposes, but I don't need the change myself.


alexcrichton:
Thank you for this! I've pushed up a commit where I'm attempting to add a vet and address the `cargo deny` CI failure.

alexcrichton (approved):

PR #8051, opened 2024-03-05
winch: Defer the usage of `format!`

saulecabrera:
This commit defers the usage of `format!` until it's actually needed when retrieving addresses for locals. This helps improve performance in modules that incur in a considerable number of local lookups.


elliottt (approved):
:+1:

PR #8052, opened 2024-03-05
winch: Introduce `TypeConverter` in the function environment

saulecabrera:
This commit introduces the `TypeConverter` struct, used to enable partial borrowing of a `ModuleTranslation` and a `ModuleTypesBuilder` from the function environment. This makes it easier to defer type conversions until they are actually needed, for example when resolving callee signatures.


elliottt (approved):
Looks great!

PR #8053, opened 2024-03-05
wasmtime-wasi: introduce WasiP1Ctx

pchickey:
This PR adds `wasmtime_wasi::WasiP1Ctx`, which is a `WasiCtx` for just using WASI P1 (and P0).

This simplifies user experience for users migrating from wasi-common's P1 support - it means they don't have to know about the `*View` traits or the `WasiPreview1Adapter`, which are just implementation details to a P1 embedder.

The p1/p0 `add_to_linker_{a?}sync` functions now take a closure that is the lens for the T in Store<T>, to be consistent with wasi-common and in general what `wasmtime::Linker` users expect.

WasiCtxBuilder now has a convenient function `build_p1(&mut self) -> WasiP1Ctx` that can be used in the place of `build(&mut self) -> WasiCtx` for users targeting a P1 embedding.

This PR also adds two trivial methods to WasiCtxBuilder to reflect ones that existed in wasi-common;s: `inherit_env` and `inherit_args`.

Finally, this PR adds `AsyncStdinStream` and `AsyncStdoutStream` as wrappers on `AsyncReadStream` and `AsyncWriteStream` to provide implementations of `StdinStream` and `StdoutStream`. (This could be separated into a totally different PR, but it doesnt seem worth the e-paperwork overhead to bother.) These provide parity with the wasi-common affordances for piping stdio to/from files, which happen to be used in the c-api.

The top-level tests and examples that were using `wasi-common` are updated to use `wasmtime-wasi::WasiP1Ctx`.

### DIFF
--- a/crates/wasi/Cargo.toml
+++ b/crates/wasi/Cargo.toml
@@ -37,7 +37,7 @@ url = { workspace = true }
 once_cell = { workspace = true }
 
 [dev-dependencies]
-tokio = { workspace = true, features = ["time", "sync", "io-std", "io-util", "rt", "rt-multi-thread", "net", "macros"] }
+tokio = { workspace = true, features = ["time", "sync", "io-std", "io-util", "rt", "rt-multi-thread", "net", "macros", "fs"] }
 test-log = { workspace = true }
 tracing-subscriber = { workspace = true }
 test-programs-artifacts = { workspace = true }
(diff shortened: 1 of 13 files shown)


lpereira (approved):
LGTM.  (Looks like one of those things that if it builds and passes CI, it's most likely OK.)

elliottt (approved):
:tada:

pchickey:
This PR should be backported to the release-19.0.0 branch

PR #8054, opened 2024-03-06
winch: Fix sp_offset usage when callee save registers are present

elliottt:
If non-empty callee-save registers are passed to the winch `MacroAssembler::prologue` function, the stack offset used when computing the value in `address_from_sp` is incorrect. To fix this, we now reset the internal stack offset to `0` after the prologue, to ensure that we're skipping the callee save space. The prologue stack offset is restored for the epilogue, ensuring that the prologue and epilogue have a consistent view of the state of the stack when they run.


lpereira (approved):
:shipit:

PR #8055, opened 2024-03-07
Add initial support for DWARF Fission

yowl:
Wasmtime supports debugging with DWARF, and this PR adds initial support for DWARF fission.  DWARF fission has advantages in linking and wasm size (https://developer.chrome.com/blog/faster-wasm-debugging) as it separates the debug_info section into a separate file leaving a `DW_TAG_skeleton_unit` in the main wasm file.   

This PR only supports DWARF pacakges, `.dwp` files.  It should also be possible to support`.dwo` files, but I've not done that here.


crates/wasmtime/src/runtime/module.rs
```diff
@@ -238,11 +249,15 @@ impl Module {
     /// ```
     #[cfg(any(feature = "cranelift", feature = "winch"))]
     #[cfg_attr(docsrs, doc(cfg(any(feature = "cranelift", feature = "winch"))))]
-    pub fn new(engine: &Engine, bytes: impl AsRef<[u8]>) -> Result<Module> {
+    pub fn new(
+        engine: &Engine,
+        bytes: impl AsRef<[u8]>,
+        wasm_path: Option<PathBuf>,
```
yowl, line 255: @peterhuene Hi, tagging you as the bot marked you as reviewer, hope that's ok?  You'll see I've added a parameter to the constructor here so that I can load the `.dwp` file at the same time as loading the `.wasm` (if loading from file).  This means there are a zillion tests to update, which is fine with me, but before I do it, do you have any comment about this aspect of this PR?

alexcrichton: If possible I'd recommend avoiding changing the default constructor of `Module` since, as you've seen, there's a lot of callers. Additionally dealing with split-dwarf-things feels somewhat niche-ish where embedders should all be able to handle it but it shouldn't necessarily be front-and-center of the embedding API.

If you're up for it what I might recommend for something like this is perhaps a `ModuleBuilder` type, for example:

```rust
impl Module {
    pub fn builder() -> ModuleBuilder;
}

impl ModuleBuilder {
    fn allow_wat(&mut self, allow: bool) -> &mut Self; // example configuration method
    fn dwarf_package(&mut self, bytes: &[u8]) -> &mut Self; // explicitly specify dwarf
    fn dwarf_package_file(&mut self, path: impl AsRef<Path>) -> Result<&mut Self>; // explicitly specify path
    
    // finish the compilation by reading wasm from `path`
    //
    // auto-load `path.dwp` if `dwarf_package` files weren't called
    fn compile_path(&mut self, path: impl AsRef<Path>) -> Result<Module>;

    // finish compilation by using the bytes provided as wasm
    fn compile(&mut self, wasm: &[u8]) -> Result<Module>;
}
```

or something along those lines. That'll keep `Module::new` working as-is. `Module::from_file` would probe for a dwarf package (perhaps with a knob in `Config` to turn that off), and users with in-memory wasm and dwarf packages can use `ModuleBuilder`

crates/wasmtime/src/runtime/module.rs
```diff
@@ -348,8 +414,8 @@ impl Module {
                     &state,
 
                     // Cache miss, compute the actual artifacts
-                    |(engine, wasm)| -> Result<_> {
-                        let (mmap, info) = build_artifacts::<MmapVecWrapper>(engine.0, wasm)?;
+                    |(engine, wasm), dwarf_package| -> Result<_> {
```
alexcrichton, line 417: Instead of modifying `wasmtime-cache`, which is pretty carefully crafted as-is to ensure we don't get cache mistakes, could `dwarf_package` be threaded through with `wasm`?

crates/wasmtime/src/engine.rs
```diff
@@ -516,10 +517,14 @@ impl Engine {
     ///
     /// [binary]: https://webassembly.github.io/spec/core/binary/index.html
     /// [text]: https://webassembly.github.io/spec/core/text/index.html
-    pub fn precompile_module(&self, bytes: &[u8]) -> Result<Vec<u8>> {
+    pub fn precompile_module(
+        &self,
+        bytes: &[u8],
+        dwarf_package: Option<DwarfPackage<EndianSlice<'_, gimli::LittleEndian>>>,
```
alexcrichton, line 523: Similar to my comment below on `Module`, I'd recommend avoiding changing this but perhaps have `ModuleBuilder::new(engine)` work and then have a "finish" method returning just a `Vec<u8>` rather than a `Module`, that way the `ModuleBuilder` could be used with this too.

crates/wasmtime/src/runtime/module.rs
```diff
@@ -381,6 +448,117 @@ impl Module {
...
+            // Use a non-zero capacity so that `ReaderOffsetId`s are unique.
+            None => Cow::Owned(Vec::with_capacity(1)),
+        };
+        let data_ref = arena_data.alloc(data);
+        let reader = gimli::EndianSlice::new(data_ref, endian);
+        let section = reader;
+        Ok(section)
+    }
+
+    fn load_dwp<'a>(
```
alexcrichton, line 537: Would it be possible to thread through the `*.dwp` bytes through to the debug processing, but otherwise leave them unmodified before then? That'd help keep all the debuginfo processing centralized in one location rather than spread out.

crates/environ/src/module_environ.rs
```diff
@@ -124,6 +125,7 @@ pub struct DebugInfoData<'a> {
     debug_loclists: gimli::DebugLocLists<Reader<'a>>,
     pub debug_ranges: gimli::DebugRanges<Reader<'a>>,
     pub debug_rnglists: gimli::DebugRngLists<Reader<'a>>,
+    pub dwarf_package: Option<DwarfPackage<Reader<'a>>>,
```
alexcrichton, line 128: Mind leaving a comment here indicating that this is not parsed from the wasm module and must be configured externally?

Either that or, alternatively, if it makes sense it might be best to thread the dwarf package as an auxiliary parameter through to the debuginfo processing. If there's too many layers to thread through though I think it's reasonable to leave it here.

crates/environ/src/module_artifacts.rs
```diff
@@ -275,6 +275,15 @@ impl<'a> ObjectBuilder<'a> {
...
             self.push_debug(&mut dwarf, &debuginfo.debug_rnglists);
+
+            if let Some(dwarf_package) = debuginfo.dwarf_package {
+                self.push_debug(&mut dwarf, &dwarf_package.debug_abbrev);
+                self.push_debug(&mut dwarf, &dwarf_package.debug_info);
+                self.push_debug(&mut dwarf, &dwarf_package.debug_line);
+                self.push_debug(&mut dwarf, &dwarf_package.debug_str);
+                self.push_debug(&mut dwarf, &dwarf_package.debug_str_offsets);
+                self.push_debug(&mut dwarf, &dwarf_package.debug_rnglists);
+            }
```
alexcrichton, line 286: If these sections are present in the dwarf package are they not present in the original wasm? If so could that be validated here because otherwise this might create dual sections which might confuse some dwarf readers.

crates/cranelift/src/debug/transform/mod.rs
```diff
@@ -124,3 +150,34 @@ pub fn transform_dwarf(
...
+
+fn replace_unit_from_split_dwarf<'a>(
+    unit: &'a Unit<gimli::EndianSlice<'a, gimli::LittleEndian>, usize>,
+    dwp: &DwarfPackage<gimli::EndianSlice<'a, gimli::LittleEndian>>,
+    parent: &Dwarf<gimli::EndianSlice<'a, gimli::LittleEndian>>,
+) -> Option<(
+    Unit<gimli::EndianSlice<'a, gimli::LittleEndian>, usize>,
+    Dwarf<gimli::EndianSlice<'a, gimli::LittleEndian>>,
+)> {
+    if let Some(dwo_id) = unit.dwo_id {
```
alexcrichton, line 162: Idiomatically I might recommend some use of `?` and early-return to avoid the "tower" here with lots of indentation. For example this line here can be replaced with:

```rust
let dwo_id = unit.dwo_id?;
```

(and no indentation needed afterwards).

The next line could be replaced with:

```rust
let cu = dwp.find_cu(dwo_id, parent).ok()??;
```

(the double `??` is a bit odd but it handles the two layers of `Result<Option<_>>`)

alexcrichton (commented):
Thanks for this! I've got some high level comments about the rough structure here, mostly to help centralize things bit more in the "dwarf processing area" and additionally API guidance for enabling usage of this in all case but not necessarily requiring it to be front-and-center in the embedding API.

crates/wasmtime/src/runtime/module.rs
yowl: Sounds good to me, thanks!   Will get on it.

crates/wasi-common/WASI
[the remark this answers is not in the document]
yowl: Thanks, removed

crates/environ/src/module_artifacts.rs
yowl: From Appendix F in the spec https://dwarfstd.org/doc/DWARF5.pdf:

image


So they are there, optionally, but with a `.dwo` suffix.  Perhaps I don't need to push any here, which would solve things.

crates/wasmtime/src/compile.rs
```diff
@@ -66,6 +69,7 @@ pub(crate) fn build_artifacts<T: FinishedObject>(
     let mut translation = ModuleEnvironment::new(tunables, &mut validator, &mut types)
         .translate(parser, wasm)
         .context("failed to parse WebAssembly module")?;
+    translation.debuginfo.dwarf_package = dwarf_package;
```
yowl, line 72: @alexcrichton  With the changes for the builder, are you anticipating this lines stays or goes?

yowl: I'm going to assume it stays as does the extra parameter  to `build_artifacts` except that it will be `Option<gimli::DwarfPackage<EndianSlice<'_, LittleEndian>>>`

crates/wasmtime/src/runtime/module.rs
yowl: Do you mean a tuple `(engine, wasm, dwarf_package)` or a new struct that has the `wasm` bytes and the dwarf_package `Option<>`?  Sorry I'm new to this so have a few questtions.

crates/environ/src/module_artifacts.rs
alexcrichton: Ah ok, sounds good! You could also update `push_debug` to take an optional suffix perhaps of what to use as the new section name and set that to `".dwo"` for these method calls

crates/wasmtime/src/compile.rs
alexcrichton: I think this line is good to stay. This is sort of the choke point for "everyone before has to figure out how to thread it here" and "everyone after has to figure out how to read it from here".

The `build_artifacts` function is internal to Wasmtime so we can do whatever we want to its signature.

crates/wasmtime/src/runtime/module.rs
yowl: I've taken your outline for `ModuleBuilder` and have used it as the sole mechanism for creating a module with split dwarf.  So I've left `Module::from_file` and `new` as is.  To be honest I did that, then re-read your comment, so if you think it would be better if `from_file` understood to read dwarf packages automatically, let me know.

crates/wasmtime/src/engine.rs
yowl: I've removed this from the signature, but as I have `build_component_artifacts` taking the dwarf package binary:
https://github.com/bytecodealliance/wasmtime/pull/8055/files#diff-5d6d850074b24b869c91c37ca7a77d9e7d057d0afa44d9ce9fd31ef5774ea1f9R60
I don't quite see how I should be doing this with the builder.

crates/wasmtime/src/runtime/module.rs
yowl: ok, let me see how that works out.

yowl: I've passed the bytes through as low as module_environ.cs which is where I'm setting up the `debuginfo.dwarf_package`.  Is that what you had in mine, or having `debuginfo.dwarf_packages` as  bytes also?

crates/environ/src/module_artifacts.rs
yowl: I didn't need these after all, so have removed this change.

crates/cranelift/src/debug/transform/mod.rs
yowl: right-o

yowl:
I think I've addressed the first set of feedback and I think it looks cleaner.  There are some `safe-to-deploy` and some feature looking errors to deal with still

crates/environ/src/module_environ.rs
```diff
@@ -8,6 +8,7 @@ use crate::{
     Tunables, TypeConvert, TypeIndex, Unsigned, WasmError, WasmHeapType, WasmResult, WasmValType,
     WasmparserTypeConverter,
 };
+use anyhow::Result;
```
alexcrichton, line 11: Mind doing a pass over the PR (before the end) to remove noop changes like the ones in this file? There's a few other scattered throughout, and they're all minor. No need to do that immediately though, just something that'd be good to do before merging.

crates/wasmtime/src/compile.rs
```diff
@@ -110,9 +116,10 @@ pub(crate) fn build_artifacts<T: FinishedObject>(
 /// The output artifact here is the serialized object file contained within
 /// an owned mmap along with metadata about the compilation itself.
 #[cfg(feature = "component-model")]
-pub(crate) fn build_component_artifacts<T: FinishedObject>(
+pub(crate) fn build_component_artifacts<'a, T: FinishedObject>(
     engine: &Engine,
     binary: &[u8],
+    dwarf_package_binary: Option<&'a [u8]>,
```
alexcrichton, line 122: This is probably best left as unconditionally `None` for components since components will need more support for loading multiple possible `*.dwp` files or something like that in the future.

crates/wasmtime/src/engine.rs
```diff
@@ -519,7 +520,7 @@ impl Engine {
     pub fn precompile_module(&self, bytes: &[u8]) -> Result<Vec<u8>> {
         #[cfg(feature = "wat")]
         let bytes = wat::parse_bytes(&bytes)?;
-        let (v, _) = crate::compile::build_artifacts::<Vec<u8>>(self, &bytes)?;
+        let (v, _) = crate::compile::build_artifacts::<Vec<u8>>(self, &bytes, None)?;
```
alexcrichton, line 523: Could this change to using the `ModuleBuilder` API and have this implementation move to within the `ModuleBuilder`?

crates/wasmtime/src/lib.rs
```diff
@@ -166,6 +166,9 @@
 //! [`Func`] point within the [`Store`] and require the [`Store`] to be provided
 //! to actually access the internals of the WebAssembly function, for instance.
 //!
+//! * [`ModuleBuilder`] - a builder module that can be used to fuse DWRF `.dwp`
+//!   packages with a WebAssembly module function for the purposes of debugging.
+//!
```
alexcrichton, line 171: I think it's probably best to leave this out of the top-level documentation and instead refer to it in the documentation of `Module` itself.

crates/wasmtime/src/runtime/module.rs
```diff
@@ -242,7 +243,7 @@ impl Module {
         let bytes = bytes.as_ref();
         #[cfg(feature = "wat")]
         let bytes = wat::parse_bytes(bytes)?;
-        Self::from_binary(engine, &bytes)
+        Self::from_binary(engine, &bytes, None)
```
alexcrichton, line 246: If you're ok changing these constructors of `Module` I think it'd be best to reframe them all in terms of `ModuleBuilder`, but I can also have a follow-up to do that refactoring.

crates/wasmtime/src/runtime/module.rs
```diff
@@ -348,37 +370,29 @@ impl Module {
                     &state,
 
                     // Cache miss, compute the actual artifacts
-                    |(engine, wasm)| -> Result<_> {
-                        let (mmap, info) = build_artifacts::<MmapVecWrapper>(engine.0, wasm)?;
-                        let code = publish_mmap(mmap.0)?;
-                        Ok((code, info))
+                    // Implementation of how to serialize artifacts
```
alexcrichton, line 373: This comment might be a copy/paste error?

crates/wasmtime/src/runtime/module.rs
```diff
@@ -1077,6 +1091,11 @@ impl Module {
     pub(crate) fn id(&self) -> CompiledModuleId {
         self.inner.module.unique_id()
     }
+
+    // TODO: added this here, but not sure what the purpose is, likely I've misunderstood https://github.com/bytecodealliance/wasmtime/pull/8055#discussion_r1517064689
+    // pub fn builder<'a>() -> ModuleBuilder<'a> {
+    //     ModuleBuilder::new(&Engine::new(config)?)
+    // }
```
alexcrichton, line 1098: Ah it's ok to leave this out, I can have a follow-up.

alexcrichton (commented):
Looking good! I wanted to offer, though, if you'd like I'm happy to help out with the refactoring required to make `ModuleBuilder` come into existence. That's a big chunk of this change and isn't related directly to DWARF fission, so I can help out land the refactorings required to get `ModuleBuilder` itself in place and that would help slim down this change to just DWARF fission. 

I say this because I've got a number of comments about specific details about the API, but I don't want to bog you down too much if you're mainly interested in the DWARF bits.

crates/wasmtime/src/runtime/module.rs
yowl: oops

yowl:
>I wanted to offer, though, if you'd like I'm happy to help out with the refactoring required to make `ModuleBuilder` come into existence. 

That would be very kind, thanks!  You are right my primary goal is DWARF fission, so I can land https://github.com/dotnet/runtimelab/pull/2264 which I can't really do properly without wasmtime support.

crates/wasmtime/src/compile.rs
yowl: Yes, I'd like to tackle that after this.

alexcrichton:
Ok sounds good, I'll take some time today and split out the changes here for `ModuleBuilder`, make a PR for that, and in theory this should be a relatively straightforward "rebase on that and add a few llines to have support"

alexcrichton:
Ok I've extracted the refactorings I was envisioning to https://github.com/bytecodealliance/wasmtime/pull/8181

yowl:
Hi, Ive got a few of these errors in the Test Linux job:
```
---- cli_tests::test_programs::run_wasi_http_component stdout ----
thread 'main' panicked at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/object-0.33.0/src/read/wasm.rs:276:64:
index out of bounds: the len is 44 but the index is 357
stack backtrace:
   0: rust_begin_unwind
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/panicking.rs:72:14
   2: core::panicking::panic_bounds_check
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/panicking.rs:208:5
   3: <usize as core::slice::index::SliceIndex<[T]>>::index_mut
   4: <alloc::vec::Vec<T,A> as core::ops::index::IndexMut<I>>::index_mut
   5: object::read::wasm::WasmFile<R>::parse
   6: object::read::any::File<R>::parse
   7: wasmtime::engine::serialization::detect_precompiled_bytes
   8: wasmtime::engine::Engine::detect_precompiled
   9: wasmtime_cli::common::RunCommon::load_module_contents
  10: wasmtime_cli::common::RunCommon::load_module
  11: wasmtime_cli::commands::run::RunCommand::execute
  12: wasmtime::Wasmtime::execute
  13: wasmtime::old_cli::main
  14: wasmtime::main
  15: core::ops::function::FnOnce::call_once

```
Does it ring any bells, I don't immediately see a connection with this PR?  If not, I'll start up a Linux system to debug it.

Thanks!

philipc:
It's a bug in the `object` crate in a code path that is now being executed because this PR enables the `wasm` feature for `object`.

philipc:
Workaround:
```diff
--- a/crates/wasmtime/src/engine/serialization.rs
+++ b/crates/wasmtime/src/engine/serialization.rs
@@ -24,7 +24,7 @@
 use crate::{Engine, ModuleVersionStrategy, Precompiled};
 use anyhow::{anyhow, bail, ensure, Context, Result};
 use object::write::{Object, StandardSegment};
-use object::{File, FileFlags, Object as _, ObjectSection, SectionKind};
+use object::{NativeFile as File, FileFlags, Object as _, ObjectSection, SectionKind};
 use serde_derive::{Deserialize, Serialize};
 use std::str::FromStr;
 use wasmtime_environ::obj;
```
Alternatively `ElfFile64` instead of `NativeFile`.

yowl:
@philipc Thanks, that gets this green.  I'll merge in from main when #8181 is in, give it a once over, see if there is a sensible test to add, then mark it ready for review.

alexcrichton:
With https://github.com/bytecodealliance/wasmtime/pull/8181 now landed if you want to rebase @yowl I can help review any remaining bits

yowl:
Having merged with main, looks like I have a problem with wasmparser versions, 0.202.0 is the version generally used and listed in the toml, but `object` (version 0.33 and 0.34) wants 0.201.0.  Is there are sensible way to resolver this?

philipc:
I'm not a `wasmtime` developer, but as the `object` maintainer I can say that I think `object` has a slower release cadence, and this is likely to be an ongoing problem. `wasmtime` already has code to read DWARF sections from Wasm files; have you considered using that instead of enabling the `wasm` feature for `object`?

alexcrichton:
Ah yeah no need to worry about that, we already have an exception for `wit-bindgen`'s transitive dependency on `wasmparser` and we can add another for `object` if needed. We definitely don't expect `object` to be released at the same pace as `wasmparser`.

As @philipc said though if the `wasm` feature isn't needed and/or could be disabled that'd probably be best, but whatever works best for you @yowl we can make it work.

yowl:
Seems like whatever reason I had for adding the `wasm` feature, has gone now, so will revert that change.

yowl:
I've been looking at the tests in `tests\all\debug\lldb.rs` which seems like a place where a test could go checking the result of `fr v` for a simple c program with some dwp.  I note that all the tests here are marked `ignore` which presumably means they dont run as part of the CI.  Why is that?

yowl:
@alexcrichton Hi, do you know why the lldb tests are `ignore`d ?  Is there a reason they dont run in the CI ?

alexcrichton:
Oh the setup is actually a little wonky, but they're ignored-by-default but run-by-default [here](https://github.com/bytecodealliance/wasmtime/blob/bfc9f3173c7e5463c8707869fc504bb6051366e7/.github/workflows/main.yml#L584-L590). The original intention was to avoid placing extra platform restrictions on those running tests since lldb-with-wasm-support I think required a patch or was otherwise somewhat nontrivial to acquire at the time.

yowl:
I added a test for lldb with split dwarf using a c program I compiled with emscripten which checks for the ability to look at frame variables.  This is passing so I'll mark this ready for review.  Thanks for the help!

crates/wasmtime/src/compile/code_builder.rs
```diff
@@ -139,6 +149,53 @@ impl<'a> CodeBuilder<'a> {
...
+            self.dwarf_package_path = Some(Cow::Owned(dwarf_package_path_buf.as_path().to_owned()));
+            let path = dwarf_package_path_buf.as_path();
+            return self.dwarf_package_file(path);
+        }
+
+        Ok(self)
+    }
+
+    /// Gets the DWARF package.
+    pub fn dwarf_package_binary(&self) -> &Option<Cow<'_, [u8]>> {
```
alexcrichton, line 181: Mind making this not-`pub`? (I think it's only intended for internal bits)

As a minor nit here this might most appropriately return `Option<&[u8]>` as well, but that's just a minor thing.

crates/wasmtime/src/compile/code_builder.rs
```diff
@@ -139,6 +149,53 @@ impl<'a> CodeBuilder<'a> {
...
+        return &self.dwarf_package;
+    }
+
+    /// Set the dwarf package binary and path.
```
alexcrichton, line 185: Could you expand the documentation here about what this method is doing under the hood in terms of when this is required and what the `*.dwp` is used for?

For example this method isn't required with `wasm_file` since it'll probe for the `*.dwp` automatically.

crates/wasmtime/src/compile/code_builder.rs
```diff
@@ -119,6 +127,8 @@ impl<'a> CodeBuilder<'a> {
             .with_context(|| format!("failed to read input file: {}", file.display()))?;
         self.wasm = Some(wasm.into());
         self.wasm_path = Some(file.into());
+        self.dwarf_package_from_wasm_path()?;
```
alexcrichton, line 130: Could you update the documentation to indicate that the `*.dwp` file is probed for?

crates/wasmtime/src/compile/code_builder.rs
```diff
@@ -76,6 +80,10 @@ impl<'a> CodeBuilder<'a> {
         }
         self.wasm = Some(wasm_bytes.into());
         self.wasm_path = wasm_path.map(|p| p.into());
+        if self.wasm_path.is_some() {
+            self.dwarf_package_from_wasm_path()?;
+        }
```
alexcrichton, line 85: If it works for you I'd prefer to leave out this in this method if we can. I think it's good to have some "pure in memory methods" for the builder that don't touch the filesystem, and this `wasm` method was intended for that. The `wasm_file` is a nice convenience for probing the filesystem for `*.dwp`, but here `wasm_path` is mostly an error message hint rather than a location on disk per se.

crates/wasmtime/src/compile/code_builder.rs
```diff
@@ -139,6 +149,53 @@ impl<'a> CodeBuilder<'a> {
...
+            bail!("cannot call `dwarf_package` or `dwarf_package_file` twice");
+        }
+
+        if !file.exists() {
+            bail!("DWARF package file does not exist");
+        }
+
+        let dwarf_package = std::fs::read(file)
+            .with_context(|| format!("failed to read dwarf input file: {}", file.display()))?;
+        self.dwarf_package = Some(dwarf_package.into());
```
alexcrichton, line 164: In addition to setting `self.dwarf_package`, could this file also set `self.dwarf_package_path`? For example the mutation happening below in `dwarf_package_from_wasm_path` would be pushed into this method. That way if this method is called independently then it'll set `self.dwarf_package_path` automatically.

crates/wasmtime/src/runtime/module.rs
```diff
@@ -319,6 +318,25 @@ impl Module {
...
+        engine: &Engine,
+        binary: &[u8],
+        dwp_binary: &[u8],
+    ) -> Result<Module> {
+        crate::CodeBuilder::new(engine)
+            .wasm(binary, None)?
+            .wat(false)?
+            .dwarf_package(dwp_binary, None)?
+            .compile_module()
+    }
```
alexcrichton, line 338: If you're ok with it, could this be omitted? This is a convenience for `CodeBuilder` methods, and I'd personally prefer to leave combining the method to embedders where we can. (in retrospect we can probably trim the constructors of `Module` by deferring more use cases to `CodeBuilder` as well)

crates/cranelift/src/debug/transform/unit.rs
```diff
@@ -275,20 +277,33 @@ where
     let mut pending_di_refs = PendingDebugInfoRefs::new();
     let mut stack = Vec::new();
 
+    let mut program_unit = unit;
+    let mut skeleton_die = None;
+
+    // Get entries in outer scope to avoid borrowing on short lived temporary.
+    let mut skeleton_entries = unit.entries();
+    if split_unit.is_some() {
+        program_unit = split_unit.unwrap();
```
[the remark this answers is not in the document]
yowl: Thanks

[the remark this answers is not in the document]
yowl: The thing is `program_unit` is reassigned here, it starts as `unit` from the main module and is reassigned if there is a split unit, so I can't use `let` as it is declared a few lines higher up.

crates/cranelift/src/debug/transform/unit.rs
```diff
@@ -275,20 +277,33 @@ where
...
+
+        // From the spec, a skeleton unit has no children so we can assume the first, and only, entry is the DW_TAG_skeleton_unit (https://dwarfstd.org/doc/DWARF5.pdf).
+        skeleton_die = Some(skeleton_entries.next_dfs()?.unwrap().1);
```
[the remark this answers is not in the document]
yowl: Used a `if let Some` and have ignored the `None` case.

crates/cranelift/src/debug/transform/mod.rs
```diff
@@ -39,28 +42,115 @@ where
...
+            .unwrap();
+
+    match load_dwp(translation, buffer) {
+        Ok(package) => Some(package),
+        Err(err) => {
+            eprintln!("Failed to load Dwarf package {}", err);
+            None
+        }
+    }
+}
```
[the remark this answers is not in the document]
yowl: Thanks, done.

crates/cranelift/Cargo.toml
```diff
@@ -30,6 +30,8 @@ object = { workspace = true, features = ['write'] }
 thiserror = { workspace = true }
 cfg-if = { workspace = true }
 wasmtime-versioned-export-macros = { workspace = true }
+fallible-iterator = "0.3.0"
+typed-arena = "2"
```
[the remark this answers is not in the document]
yowl: Removed.

crates/wasmtime/src/runtime/module.rs
yowl: sure, deleted

crates/wasmtime/src/compile/code_builder.rs
yowl: Thanks, done.

crates/wasmtime/src/compile/code_builder.rs
yowl: sure, removed.

crates/wasmtime/src/compile/code_builder.rs
yowl: Its used in https://github.com/bytecodealliance/wasmtime/pull/8055/files#diff-92db9b7b9b675f97d5c8d9d94e0680d63aab09fb1454f5441cb61ddbff8d3d81R17

Have changed the return type, thanks.

alexcrichton (approved):
Thanks again for this! I had some more minor nits but I went ahead and pushed up the changes directly. If anything there looks off to you though please let me know

alexcrichton:
@yowl [this failure](https://github.com/bytecodealliance/wasmtime/actions/runs/8789182763/job/24118321243) looks legitimate in the sense of probably-caused-by-this PR. Do you know if that perhaps means the LLDB we're using in CI is missing something and/or is too old?

yowl:
This looks relevant https://github.com/llvm/llvm-project/issues/55575  Seems like we have to add some missing python stuff.

alexcrichton:
I'm not sure myself, we don't have anything special beyond the standard github actions runners. If you'd like you can push up a commit with "prtest:full" in the commit message and it'll run the full test suite here on the PR to help debug.

alexcrichton (approved):

PR #8056, opened 2024-03-07
cranelift: Add a winch calling convention

elliottt:
Add a `Winch` calling convention to cranelift that has no callee saves, and clobbers everything. This will enable us to use the `wasmtime-cranelift` generated trampolines in the `wasmtime-winch` crate.

I've enabled fuzzing for this calling convention, as well as added clif tests for the `x64` and `aarch64` backends. I've explicitly disabled support in `riscv64` and `s390x`, as it's not currently possible to target either of those architectures with winch.

### DIFF
--- a/cranelift/codegen/src/isa/aarch64/abi.rs
+++ b/cranelift/codegen/src/isa/aarch64/abi.rs
@@ -1107,8 +1107,8 @@ impl ABIMachineSpec for AArch64MachineDeps {
 
     fn get_regs_clobbered_by_call(call_conv_of_callee: isa::CallConv) -> PRegSet {
         match call_conv_of_callee {
-            isa::CallConv::Tail => TAIL_CLOBBERS,
-            isa::CallConv::Winch => WINCH_CLOBBERS,
+            isa::CallConv::Tail => ALL_CLOBBERS,
+            isa::CallConv::Winch => ALL_CLOBBERS,
             _ => DEFAULT_AAPCS_CLOBBERS,
         }
     }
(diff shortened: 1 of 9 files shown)


cranelift/codegen/src/isa/x64/abi.rs
```diff
@@ -1234,6 +1241,40 @@ const fn tail_clobbers() -> PRegSet {
...
+        .with(regs::fpr_preg(6))
+        .with(regs::fpr_preg(7))
+        .with(regs::fpr_preg(8))
+        .with(regs::fpr_preg(9))
+        .with(regs::fpr_preg(10))
+        .with(regs::fpr_preg(11))
+        .with(regs::fpr_preg(12))
+        .with(regs::fpr_preg(13))
+        .with(regs::fpr_preg(14))
+        .with(regs::fpr_preg(15))
```
alexcrichton, line 1275: Could this be shared with `tail_clobbers` above as something like `preg_set_all()`?

cranelift/codegen/src/isa/aarch64/abi.rs
```diff
@@ -1608,6 +1608,71 @@ const TAIL_CLOBBERS: PRegSet = PRegSet::empty()
...
+    .with(vreg_preg(22))
+    .with(vreg_preg(23))
+    .with(vreg_preg(24))
+    .with(vreg_preg(25))
+    .with(vreg_preg(26))
+    .with(vreg_preg(27))
+    .with(vreg_preg(28))
+    .with(vreg_preg(29))
+    .with(vreg_preg(30))
+    .with(vreg_preg(31));
```
alexcrichton, line 1675: Like x64, could the shared bits with tail be factored to a helper constant/function?

alexcrichton (approved):

cranelift/codegen/src/isa/x64/abi.rs
elliottt: My worry was that we've been discussing adding callee saves back to the tail calling convention, and I didn't want that change to accidentally undo this one.

alexcrichton: Oh sure, so not literally sharing the name "tail" but something like a helper function that both tail/winch call, and one day tail will stop calling it

elliottt: That works!

PR #8059, opened 2024-03-07
winch: Fix the bounds check for dynamic heaps

saulecabrera:
This commit fixes the bounds check comparison for dynamic heaps. The previous implementation wasn't using the right value: for the general case of dynamic heaps, we want to compare if 

      index + offset + access_size > bounds

But it was only comparing
      
     index > bounds

This commit addresses the issue by adding a new temporary register into the equation which will be used for the bounds check comparison and for overflow checks. This approach is preferred over using the scratch register because it's harder to know when the scratch register might get clobbered (in the case of spectre checks, it could for example)

### DIFF
--- a/tests/all/winch.rs
+++ b/tests/all/winch.rs
@@ -340,3 +340,57 @@ fn static_oob() -> Result<()> {
 
     Ok(())
 }
+
+#[test]
+#[cfg_attr(miri, ignore)]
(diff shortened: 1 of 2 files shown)


fitzgen (approved):

fitzgen:
Thanks!

PR #8061, opened 2024-03-07
Move dtor information from `ResourceAny` into a `Store`

alexcrichton:
This commit moves the internals of the `ResourceAny` type related to destructors into the auxiliary data within a `Store`. This avoids the need to carry around pointers with `ResourceAny` and additionally makes it easier to work with.

As part of this commit this also updates the behavior of `ResourceAny::try_from_resource` to no longer need an `InstancePre` and type information. This was required originally to get `ResourceAny::resource_drop` working to drop the host resource. In retrospect I'm not sure if this was the best goal to achieve because `Resource<T>` already has no destructor support and one of the more common use cases for the conversion is to simply turn around and give it back to a component where a component already has enough destructor information.

In the end this should make both `ResourceAny` and `Resource<T>` pretty close to "just an index" which is easier to reason about when working with resources than carrying additional pointers.

Closes https://github.com/bytecodealliance/wasmtime/issues/7714

### DIFF
--- a/crates/wasmtime/src/runtime/component/func/options.rs
+++ b/crates/wasmtime/src/runtime/component/func/options.rs
@@ -338,8 +338,14 @@ impl<'a, T> LowerContext<'a, T> {
     ///
     /// Note that this is a special case for `Resource<T>`. Most of the time a
     /// host value shouldn't be lowered with a lowering context.
-    pub fn host_resource_lower_own(&mut self, rep: u32) -> Result<HostResourceIndex> {
-        self.resource_tables().host_resource_lower_own(rep)
+    pub fn host_resource_lower_own(
+        &mut self,
+        rep: u32,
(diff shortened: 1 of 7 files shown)


alexcrichton:
cc @rvolosatovs as this is touching on the work of https://github.com/bytecodealliance/wasmtime/pull/7688

pchickey (approved):

rvolosatovs (approved):
Looks great, thanks.
Should `ResourceIndex` be removed altogether and not be returned anymore by `LinkerInstance::resource`?

It appears that this PR closes https://github.com/bytecodealliance/wasmtime/issues/7714

alexcrichton:
Good points! And sorry again for not realizing this is probably the best way to go sooner, I realize this is a bit of a runaround with the API :(

PR #8062, opened 2024-03-07
Remove type information from `wasmtime::component::Val`

alexcrichton:
This commit is a large refactor of the `Val` type as used with components to remove inherent type information present currently. The `Val` type is now only an AST of what a component model value looks like and cannot fully describe the type that it is without further context. For example enums only store the case that's being used, not the full set of cases.

The motivation for this commit is to make it simpler to use and construct `Val`, especially in the face of resources. Some problems solved here are:

* With resources in play managing type information is not trivial and can often be surprising. For example if you learn the type of a function from a component and the instantiate the component twice the type information is not suitable to use with either function due to exported resources acquiring unique types on all instantiations.

* Functionally it's much easier to construct values when type information is not required as it no longer requires probing various pieces for type information here and there.

* API-wise there's far less for us to maintain as there's no need for a type-per-variant of component model types. Pieces now fit much more naturally into a `Val` shape without extra types.

* Functionally when working with `Val` there's now only one typecheck instead of two. Previously a typecheck was performed first when a `Val` was created and then again later when it was passed to wasm. Now the typecheck only happens when passed to wasm.

It's worth pointing out that `Val` as-is is a pretty inefficient representation of component model values, for example flags are stored as a list of strings. While semantically correct this is quite inefficient for most purposes other than "get something working". To that extent my goal is to, in the future, add traits that enable building a custom user-defined `Val` (of sorts), but still dynamically. This should enable embedders to opt-in to a more efficient representation that relies on contextual knowledge.

### DIFF
--- a/crates/component-macro/src/component.rs
+++ b/crates/component-macro/src/component.rs
@@ -447,7 +447,7 @@ impl Expander for LiftExpander {
             if let Some(ty) = ty {
                 let payload_ty = match style {
                     VariantStyle::Variant => {
-                        quote!(ty.cases[#index].ty.unwrap_or_else(#internal::bad_type_info))
+                        quote!(ty.cases[#index].unwrap_or_else(#internal::bad_type_info))
                     }
                     VariantStyle::Enum => unreachable!(),
                 };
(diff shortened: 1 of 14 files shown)


lann (approved):

fitzgen (approved):
LGTM!

PR #8065, opened 2024-03-08
Fix generating bindings for dead code

### LINKED ISSUE #8057: Component model: `bindgen!` doesn't generate some structures
I'm working on a new WASI proposal and I'm implementing an example implementation with `wasmtime`. I have a correct WIT definition AFAIK and [I'm generating the bindings with the bindgen macro](https://github.com/tareksander/WASI-Windowing/blob/14b072cde5542bc81922e15275229fa3295301e1/example-runtime/src/main.rs#L14-L20). For some reason, the `key-data` record and `key` variant didn't generate a corresponding struct/enum. There's an error inside the generated bindgen code "cannot find type "KeyData" in this scope".

alexcrichton:
This commit fixes a bindgen problem where dead types, which had no generated bindings, could be referred to from other dead types, which did have generated bindings. The fix is to not generate code for dead types at all by fixing how type information is analyzed.

Closes #8057

### DIFF
--- a/crates/component-macro/tests/codegen/dead-code.wit
+++ b/crates/component-macro/tests/codegen/dead-code.wit
@@ -1,0 +1,27 @@
+package a:b;
+
+world imports {
+  import interface-with-live-type;
+  import interface-with-dead-type;
+}
+
+interface interface-with-live-type {
+  record live-type {
+    a: u32,
+  }
+  f: func() -> live-type;
+}
+
+
+interface interface-with-dead-type {
+  use interface-with-live-type.{live-type};
+
+  record dead-type {
+    a: u32,
+  }
+
+  variant v {
+    a(live-type),
+    b(dead-type),
+  }
+}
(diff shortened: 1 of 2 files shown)


crates/component-macro/tests/codegen/dead-code.wit
```diff
@@ -0,0 +1,27 @@
...
+
+  record dead-type {
+    a: u32,
+  }
+
+  variant v {
+    a(live-type),
+    b(dead-type),
+  }
+}
```
pchickey, line 27: Not clear to me how this tests liveness since both interfaces are part of the imports?

crates/wit-bindgen/src/types.rs
```diff
@@ -40,6 +40,14 @@ impl std::ops::BitOrAssign for TypeInfo {
 
 impl Types {
     pub fn analyze(&mut self, resolve: &Resolve, world: WorldId) {
+        // Build up all type information first which is inherited through types,
+        // such as properties of borrows/lists/etc.
+        for (t, _) in resolve.types.iter() {
+            self.type_id_info(resolve, t);
+        }
```
elliottt, line 47: Does doing this step first mean that the calls to `self.type_info` in `type_info_func` should never traverse the whole type (due to the early exit if the entry exists already)?

elliottt (approved):

PR #8066, opened 2024-03-08
wasmtime-c-api: switch from wasi-common to wasmtime-wasi

pchickey:
Based on #8053

This PR switches the wasmtime-c-api from using wasi-common to wasmtime-wasi.

The c-api still only exposes WASI P1 support to the guests - this isn't attempting to solve the problem of exposing components and WASI P2 on the c-api. It is only changing the implementation used under the hood so that we can push wasi-common into maintenance mode.

This PR removes the `wasi_config_preopen_socket` symbol from the c-api. It should be deployed as part of release-20.0.0 (not backported to 19) so that users get 1 release of deprecation warning via https://github.com/bytecodealliance/wasmtime/pull/8064 . The idea of a preopen socket is a weird corner of P1 that never got any appropriate testing in the wasmtime tree, and doesnt have a translation in the P2 world. I sought out if there were any users of that part of the P1 / wasi-common code, because the company that contributed it is now defunct, and I never found any users. The warning in #8064 may alert us to some and the symbol going away in this PR may as well.

### DIFF
--- a/crates/c-api/src/wasi.rs
+++ b/crates/c-api/src/wasi.rs
@@ -109,11 +109,7 @@ impl wasi_config_t {
                 builder.inherit_stdout();
             }
             WasiConfigWritePipe::File(file) => {
-                let file = tokio::fs::File::from_std(file);
-                let stdout_stream = wasmtime_wasi::AsyncStdoutStream::new(
-                    wasmtime_wasi::pipe::AsyncWriteStream::new(1024 * 1024, file),
-                );
-                builder.stdout(stdout_stream);
+                builder.stdout(wasmtime_wasi::OutputFile::new(file));
             }
         };
         match self.stderr {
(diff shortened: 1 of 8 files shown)


crates/c-api/src/wasi.rs
```diff
@@ -114,9 +109,11 @@ impl wasi_config_t {
                 builder.inherit_stdout();
             }
             WasiConfigWritePipe::File(file) => {
-                let file = cap_std::fs::File::from_std(file);
-                let file = wasi_common::sync::file::File::from_cap_std(file);
-                builder.stdout(Box::new(file));
+                let file = tokio::fs::File::from_std(file);
+                let stdout_stream = wasmtime_wasi::AsyncStdoutStream::new(
+                    wasmtime_wasi::pipe::AsyncWriteStream::new(1024 * 1024, file),
```
alexcrichton, line 114: I think the previous version didn't buffer things here, so could this keep the unbuffered versions?

alexcrichton (approved):
Thanks!

alexcrichton (approved):

PR #8070, opened 2024-03-08
Remove type information from dynamic component funcs

alexcrichton:
This commit removes the `&Component` argument from the `component::Linker::func_new` API. This is inspired by #8062 where `Val` holds less type information as well in addition to the realization that type-checking happens at runtime rather than instantiation time.

This argument was originally added to mirror
`wasmtime::Linker::func_new` which takes a type argument of the core wasm function that's being defined. Unlike core wasm, though, component functions already have to carry along their type information as part of function calls to handle resources correctly. This means that when a host function is invoked the type is already known of all the parameters and results. Additionally values are already required to be type-checked going back into wasm, so there's less of a need to perform an additional type-check up front.

The main consequence of this commit is that it's a bit more difficult for embeddings to know what the expected types of results are. No type information is provided when a host function is defined, not even function arity. This means that when the host function is invoked it may not know how many results are expected to be produced and of what type. Typically though a bindings generator is used somewhere along the way so that's expected to alleviate this issue.

Finally my hope is to enhance this "dynamic" API in the future with a bit more information so the type information is more readily accessible at runtime. For now though hosts will have to "simply know what to do".

### DIFF
--- a/crates/wasmtime/src/runtime/component/linker.rs
+++ b/crates/wasmtime/src/runtime/component/linker.rs
@@ -6,16 +6,13 @@ use crate::component::{
     Component, ComponentNamedList, Instance, InstancePre, Lift, Lower, ResourceType, Val,
 };
 use crate::{AsContextMut, Engine, Module, StoreContextMut};
-use anyhow::{anyhow, bail, Context, Result};
-use indexmap::IndexMap;
+use anyhow::{bail, Context, Result};
 use semver::Version;
 use std::collections::hash_map::{Entry, HashMap};
 use std::future::Future;
 use std::marker;
-use std::ops::Deref;
 use std::pin::Pin;
 use std::sync::Arc;
-use wasmtime_environ::component::TypeDef;
 use wasmtime_environ::PrimaryMap;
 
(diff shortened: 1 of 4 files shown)


crates/wasmtime/src/runtime/component/linker.rs
```diff
@@ -413,53 +410,111 @@ impl<T> LinkerInstance<'_, T> {
...
-    /// type of the import.
-    pub fn func_new<
-        F: Fn(StoreContextMut<'_, T>, &[Val], &mut [Val]) -> Result<()> + Send + Sync + 'static,
-    >(
+    /// The `name` provided is the name of the function to define and the
+    /// `func` provided is the host-defined closure to invoke when this
+    /// function is called.
+    ///
+    /// This function is the "dynamic" version of defining a host function as
+    /// copared to [`Linker::func_wrap`]. With [`Linker::func_wrap`] a
```
dicej, line 420:
```suggestion
    /// compared to [`Linker::func_wrap`]. With [`Linker::func_wrap`] a
```

dicej (approved):

Issue #8037, opened 2024-03-01
fd_write ignores iovs_len > 1


ThisGuyCodes:
Am I crazy? I thought I might be misunderstanding something, but then I found someone else's example online that also doesn't work.

### Test Case

```wat
;; hello.wat
(module
    ;; define the expected type for fd_write
    (type $write_type (func (param i32 i32 i32 i32) (result i32)))
    (import "wasi_snapshot_preview1" "fd_write" (func $write (type $write_type)))
    
    ;; Define a memory that is one page in size (64KiB).
    (memory (export "memory") 1)
    
    ;; At offset 66 and 80 in the memory, we store the data.
    (data (i32.const 66) "Hello, World!\n")
    (data (i32.const 80) "Howdy!\n")
    
    
    (func $main (export "_start")
        ;; Store the iovs
        (i32.store (i32.const 0) (i32.const 66)) ;; Start of data (= *buf)
        (i32.store (i32.const 4) (i32.const 14)) ;; Length of data (= buf_len)

        (i32.store (i32.const 8) (i32.const 80)) ;; Start of data (= *buf)
        (i32.store (i32.const 12) (i32.const 8)) ;; Length of data (= buf_len)

        i32.const 1 ;; fd param. In Unix this means: 0 = stdin, 1 = stdout, 2 = stderr

        ;; here are different combinations possible:
        ;; 0,1 would read only "Hello, World!"
        ;; 8,1 would read only 'Howdy'
        i32.const 0 ;; iovs -> Offset. Points to first iovec
        i32.const 2 ;; iovs_len -> How many iovs should be read, set it to 2 to read also Howdy

        i32.const 92 ;; Pointer to the place in memory where the number of written bytes shall be placed.

        call $write ;; call fd_write and drop the result code
        drop
    )
)
```

### Steps to Reproduce

```shell
wasmtime hello.wat
```

### Expected Results

```
Hello, World!
Howdy!
```

### Actual Results

```
Hello, World!
```

### Versions and Environment

Wasmtime version or commit: `wasmtime-cli 18.0.2 (90db6e99f 2024-02-28)`

Operating system: MacOS 14.3.1

Architecture: arm64 (M1 Pro)

### Extra Info

Anything else you'd like to add?

bjorn3:
[...]

PR #8072, opened 2024-03-09
Write all iovs in fd_write of wasi-preview1-component-adapter

ospencer:
I noticed that the implementation of `fd_write` in the preview1 adapter only writes the first iov in the list passed. This change writes all of them. I'm happy to add a test for this, though I couldn't find any tests for the adapter. I tested this locally via wasm-tools.

### DIFF
--- a/crates/wasi-preview1-component-adapter/src/lib.rs
+++ b/crates/wasi-preview1-component-adapter/src/lib.rs
@@ -1436,45 +1436,53 @@ pub unsafe extern "C" fn fd_write(
         return ERRNO_SUCCESS;
     }
 
-    let ptr = (*iovs_ptr).buf;
-    let len = (*iovs_ptr).buf_len;
-    let bytes = slice::from_raw_parts(ptr, len);
-
     State::with(|state| {
         let ds = state.descriptors();
         match ds.get(fd)? {
             Descriptor::Streams(streams) => {
                 let wasi_stream = streams.get_write_stream()?;
 
-                #[cfg(not(feature = "proxy"))]
-                let nbytes = if let StreamType::File(file) = &streams.type_ {
-                    file.blocking_mode.write(wasi_stream, bytes)?
-                } else {
-                    // Use blocking writes on non-file streams (stdout, stderr, as sockets
-                    // aren't currently used).
-                    BlockingMode::Blocking.write(wasi_stream, bytes)?
-                };
-                #[cfg(feature = "proxy")]
-                let nbytes = BlockingMode::Blocking.write(wasi_stream, bytes)?;
+                *nwritten = 0;
 
-                // If this is a file, keep the current-position pointer up
-                // to date. Note that for files that perform appending
-                // writes this function will always update the current
-                // position to the end of the file.
-                //
-                // NB: this isn't "atomic" as it doesn't necessarily account
-                // for concurrent writes, but there's not much that can be
-                // done about that.
-                #[cfg(not(feature = "proxy"))]
-                if let StreamType::File(file) = &streams.type_ {
-                    if file.append {
-                        file.position.set(file.fd.stat()?.size);
+                while iovs_len != 0 {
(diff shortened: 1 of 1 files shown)


bjorn3:
Writing in a loop is incompatible with POSIX. POSIX mandates that each writev is an atomic operation. While it may only write the first n bytes, it has to write all n bytes at once without the possibility of any other write getting interleaved with it. If you want to write everything, you did have to copy all buffers into one large buffer and then write this large buffer, which may actually be slower than both the status quo and your PR. See also https://github.com/bytecodealliance/wasmtime/issues/8037#issuecomment-1973512565

alexcrichton:
To add another case I would be worried about in addition to what @bjorn3 already mentioned: in the PR as-is if you have 4 buffers and successfully write the first two but then fail writing the third then the fact that the first two were successfully written is lost. At that point you've got a number of successfully written bytes plus an error and it's not clear which should be returned since only one can be returned.

ospencer:
That makes sense. It looks like some implementations of `readv`/`writev` do allocate a temporary large buffer in the case that the number of provided iovs exceeds the limit of the system, so that probably makes sense here too.

I agree that the current implementation we have of `fd_write` is technically correct, but I think the (totally allowable) behavior change from older preview1 versions of wasmtime is a little surprising to less resilient code, shown by the couple of issues/PRs opened about this.

I attempted to implement this so we could at least test it out/consider this approach, but I forgot that the adapter can't really use `Vec`s because of the panics. I'm not quite a seasoned-enough Rustacean to know how to do this without vecs in safe Rust, but if someone could give me a pointer that'd be much appreciated.

alexcrichton:
I agree it's unfortunate that this is causing issues, but IMO our hands are unfortunately tied here to the point that there's not much we can do. I think there's one thing we can do, which is to have a statically allocated or somewhere slice of bytes which the adapter copies into and then calls out to perform a single write. That would respect the semantics desired for languages and such here, at the cost of copying data. It's not clear to me when the copy is worth it vs when not, so I'm not sure how to best make such a decision.

The adapter has various bits and pieces of scratch space in its `State` and we can put a fixed-size buffer in there for copying if necessary.

titzer:
> I agree it's unfortunate that this is causing issues, but IMO our hands are unfortunately tied here to the point that there's not much we can do. I think there's one thing we can do, which is to have a statically allocated or somewhere slice of bytes which the adapter copies into and then calls out to perform a single write. That would respect the semantics desired for languages and such here, at the cost of copying data. It's not clear to me when the copy is worth it vs when not, so I'm not sure how to best make such a decision.

I hit this recently and thought I had made a mistake using the API; multiple iovecs are supported in at least three other engines. Though they are rare, supporting that functionality for application migration, even if the writes are not atomic, would be an improvement in the status quo and make the ecosystem more uniform. Given that I think these are rare, if we would like to make the POSIX atomicity guarantee, it seems like a copy into a temporary buffer would be a reasonable implementation--especially if the I/O is large, it's likely to be a small relative overhead.

alexcrichton:
I definitely agree that this can be a footgun, so I don't mean to discount that. That being said I continue to think "just call write twice" is not an acceptable solution. While that solves the footgun it means that if anyone tries to robustly handle errors/atomicity it's impossible to do so. I'm not sure @titzer if your issue is derivative of the same file changed in this PR, but this PR changes an adapter which converts WASIp1 to WASIp2. The WASIp1 APIs "looked like POSIX" and had its semantics, and callers may not even realize they're being wrapped to work in a WASIp2 world. In that sense callers with it can't be known in this context if callers with multiple iovecs care about atomicity or not.

This is why I mention that one possible solution is to copy some data. Perhaps when it's below a certain threshold or something like that it's copied into a different space and then issued as a single write. That doesn't have any atomicity issues and it's just a minor perf loss. I think it'd be reasonable to prototype that out and see what it would look like.

PR #8080, opened 2024-03-11
Update MSRV to 1.74.0, update nightly in CI

alexcrichton:
This commit updates the minimum Rust version supported by Wasmtime to 1.74.0 which is two behind the current stable 1.76. At the same time this additionally updates nightly in CI to stay up-to-date there.


elliottt (approved):
:tada:

Issue #8084, opened 2024-03-12
Lowering i32x4.splat + f64x2.convert_low_i32x4 results in panic


ShinWonho:
### Test Case

```
;; unsigned.wat
(module
  (func (export "foo") (result v128)
    (i32.const 0)
    (i32x4.splat)
    (f64x2.convert_low_i32x4_u)
  )
)
```
```
;; signed.wat
(module
  (func (export "foo") (result v128)
    (i32.const 0)
    (i32x4.splat)
    (f64x2.convert_low_i32x4_s)
  )
)
```

### Steps to Reproduce

```
wasmtime unsigned.wat
wasmtime signed.wat
```

### Expected Results

Terminate without any errors

### Actual Results

```
thread '<unnamed>' panicked at cranelift/codegen/src/machinst/lower.rs:766:21:
should be implemented in ISLE: inst = `v6 = fcvt_from_uint.f64x2 v13  ; v13 = const0`, type = `Some(types::F64X2)`
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
zsh: abort      wasmtime unsigned.wat
```
```
thread '<unnamed>' panicked at cranelift/codegen/src/machinst/lower.rs:766:21:
should be implemented in ISLE: inst = `v6 = fcvt_from_sint.f64x2 v13  ; v13 = const0`, type = `Some(types::F64X2)`
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
zsh: abort      wasmtime signed.wat
```

### Versions and Environment

Wasmtime version or commit: wasmtime-cli 18.0.2 (90db6e99f 2024-02-28)

Operating system & Architecture:
1. Server
  * OS: Ubuntu 18.04.3 LTS (Bionic Beaver)
  * Arc: x86_64
2. IMac
  * OS: Monterey 12.6.2
  * Arc: x86_64 

### Extra Info

It is not reproduced in
1. M1 MacBook Air
  * OS: Sonoma 14.3.1
  * Arc: Arm64
2. MacBook Pro
  * Monterey 12.6.3
  * Arc: x86_64

fitzgen:
Thanks for the bug report!

alexcrichton:
[...]

PR #8099, opened 2024-03-12
x64: Implement fallback for fcvt_from_sint.f64x2

alexcrichton:
This instruction was previously unimplemented in Cranelift as pointed out in #8084. While a fallback for `fcvt_from_uint` was implemented in #7919 I forgot to do the same for the signed version. This commit adds a fallback that decomposes the input into two scalars and converts each individually, then reassembling the result.


elliottt (approved):
Looks good!

Issue #5532, opened 2023-01-05
Cranelift: Remove `table_addr` and tables from CLIF, move them into `cranelift-wasm`'s Wasm -> CLIF lowering


fitzgen:
Similar to what we did with heaps in https://github.com/bytecodealliance/wasmtime/issues/5283

bjorn3:
Got a commit which removes the table_addr instruction at https://github.com/bjorn3/wasmtime/tree/remove_table_addr. I'm not sure how to implement a replacement in cranelift-wasm, but if someone else wants to pick this issue up, feel free to use my code.

PR #8105, opened 2024-03-12
Stop threading Cranelift tables through cranelift-wasm callbacks

jameysharp:
Previously, every table-related method in the FuncEnvironment trait took two IDs to identify a table: the TableIndex from the WebAssembly module, and the Cranelift Table we generated associated with it.

This is redundant and most implementations only used one or the other. The TableIndex is a perfectly good identifier and we can use it in a SecondaryMap if we need to associate additional data with the table, such as a Cranelift Table ID.

This is a step toward #5532, where we don't want to use Cranelift's table support at all, so the associated data that an implementation needs will be different.

### DIFF
--- a/crates/cranelift/src/func_environ.rs
+++ b/crates/cranelift/src/func_environ.rs
@@ -9,7 +9,8 @@ use cranelift_codegen::ir::{
     AbiParam, ArgumentPurpose, Function, InstBuilder, MemFlags, Signature, UserFuncName, Value,
 };
 use cranelift_codegen::isa::{self, CallConv, TargetFrontendConfig, TargetIsa};
-use cranelift_entity::{EntityRef, PrimaryMap};
+use cranelift_entity::packed_option::ReservedValue;
+use cranelift_entity::{EntityRef, PrimaryMap, SecondaryMap};
 use cranelift_frontend::FunctionBuilder;
 use cranelift_frontend::Variable;
 use cranelift_wasm::{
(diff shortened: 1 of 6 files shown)


crates/cranelift/src/func_environ.rs
```diff
@@ -865,14 +870,82 @@ impl<'module_environment> FuncEnvironment<'module_environment> {
...
     fn get_or_init_func_ref_table_elem(
         &mut self,
         builder: &mut FunctionBuilder,
         table_index: TableIndex,
-        table: ir::Table,
         index: ir::Value,
     ) -> ir::Value {
         let pointer_type = self.pointer_type();
+        self.ensure_table_exists(builder.func, table_index);
+        let table = self.tables[table_index];
```
fitzgen, line 948: It seems like every time we call `ensure_table_exists` we then do `let table = self.tables[table_index];`

Might be worth returning that value from `ensure_table_exists` to clean up call sites a bit.

fitzgen (approved):
Nice!

crates/cranelift/src/func_environ.rs
jameysharp: I wish I could! In this PR it would be fine, since the values in `self.tables` are `Copy`, but later I'm going to have to extend it to a full struct that I'd want to return a borrow of. Unfortunately, if you have a `fn f(&'_ mut T) -> &'_ T`, the shared borrow that's returned acts like a mutable borrow, for reasons I don't entirely understand; the mutability is apparently attached to the lifetime. So I'll have to introduce this pattern later anyway.

I've also considered eagerly creating all the associated table data so that `ensure_table_exists` can go away, but that seemed like a larger change I didn't want to think through.

PR #8109, opened 2024-03-12
winch: Switch to using cranelift for all trampolines

### LINKED ISSUE #8098: winch: Use cranelift as the trampoline compiler
https://github.com/bytecodealliance/wasmtime/pull/8082 introduces Cranelift as the trampoline compiler for components in Winch. This allows reducing code duplication and allowing for better maintainability. 

In order to have uniform trampoline compilation across the board, we want to migrate the rest of Winch's trampolines to use Cranelift.

elliottt:
Remove the winch MacroAssembler-based trampolines, and reuse the trampoline-generation from the `wasmtime-cranlift` crate instead.

Fixes #8098


fitzgen (approved):
Nice!

crates/cranelift/src/lib.rs
```diff
@@ -154,7 +154,7 @@ fn wasm_call_signature(
         }
 
         // The winch calling convention is only implemented for x64 and aarch64
-        arch if tunables.tail_callable => {
+        arch if tunables.winch_callable => {
```
elliottt, line 157: If this PR gets reverted, please preserve this change.

saulecabrera (approved):
LGTM, thanks!

One thing that I wanted to note: Winch is still missing support for `emit_trampolines_for_array_call_host_func`, would it make sense to implement that function in this PR? I think a follow-up also works!

crates/cranelift/src/lib.rs
elliottt: I merged this separately, so we can ignore this comment now.

PR #8111, opened 2024-03-13
Add a script to vendor WIT files

alexcrichton:
This is a follow-up to #7929 where it was concluded that we want to retain textual WIT files in this repository. Currently today we have no automated checks for vendoring and re-vendoring requires hand-editing. To help incrementally improve this situation this commit adds a script to do all the vendoring for us. This way CI can verify that everything is generated with respect to the output of this script.

### DIFF
--- a/ci/vendor-wit.sh
+++ b/ci/vendor-wit.sh
@@ -1,0 +1,26 @@
+#!/bin/sh
+
+# Script to re-vendor the WIT files that Wasmtime uses as defined by a
+# particular tag in upstream repositories.
+#
+# This script is executed on CI to ensure that everything is up-to-date.
+
+set -ex
+
+dst=crates/wasi/wit/deps
+
+rm -rf $dst
+mkdir -p $dst
+
+repos="cli clocks filesystem http io random sockets"
+tag=0.2.0
+
+for repo in $repos; do
+  mkdir $dst/$repo
+  curl -L https://github.com/WebAssembly/wasi-$repo/archive/refs/tags/v$tag.tar.gz | \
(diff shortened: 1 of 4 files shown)


ci/vendor-wit.sh
```diff
@@ -0,0 +1,26 @@
...
+# Script to re-vendor the WIT files that Wasmtime uses as defined by a
+# particular tag in upstream repositories.
+#
+# This script is executed on CI to ensure that everything is up-to-date.
+
+set -ex
+
+dst=crates/wasi/wit/deps
+
+rm -rf $dst
```
fitzgen, line 12: Maybe check `git diff --exit-code` before this and exit the script before doing any permanent deletions if there are local, uncommitted changes? Just to be safe.

ci/vendor-wit.sh
```diff
@@ -0,0 +1,26 @@
...
+mkdir -p $dst
+
+repos="cli clocks filesystem http io random sockets"
+tag=0.2.0
```
fitzgen, line 16: And maybe the same for how to update tag/WASI version?

May make sense to put all the "config" variables at the top of the script so that folks can update them without having to read the whole script and learn the implementation just to eg add support for WASI-foo to the script.

fitzgen (approved):
:tada:

ci/vendor-wit.sh
alexcrichton: Personally I'd prefer to leave that to CI because it makes editing a script like this hard to work with and git is otherwise pretty good about making sure things aren't overwritten.

PR #8113, opened 2024-03-13
Cranelift: fix load width on select-of-load.f64 by disallowing load fusion.

### LINKED ISSUE #8112: Cranelift: different memory access behavior in x86_64 (none vs speed)
### Test Case

```rs
// main.rs
use wasmtime::*;

fn main() -> Result<()> {
    let mut config = Config::default();
    config.strategy(Strategy::Cranelift);
    config.cranelift_opt_level(OptLevel::None);

    let engine1 = Engine::new(&config)?;
    let mut new_config = config.clone();
    new_config.cranelift_opt_level(OptLevel::Speed);
    let engine2 = Engine::new(&new_config)?;
    let wat = r#"
        (module
            (type (;0;) (func (param i32 v128 v128) (result f64)))
            (import "mem" "mem" (memory (;0;) 1))
            (func (;0;) (type 0) (param i32 v128 v128) (result f64)
                (local f64)
                local.get 0
                local.get 1
                local.get 2
                i8x16.extract_lane_s 11
                i16x8.shl
                v128.store16_lane align=1 1
                local.get 0
                i32.load align=1
                f64.load align=1
                f64.const 0x1.fffffffep+31 (;=4.29497e+09;)
                i32.const 1
                select)
            (export "main" (func 0)))
    "#;
    let module1 = Module::new(&engine1, wat)?;
    let module2 = Module::new(&engine2, wat)?;
    let mut store1 = Store::new(&engine1, ());
    let mut store2 = Store::new(&engine2, ());

    let memory_ty = MemoryType::new(1, None);
    let memory1 = Memory::new(&mut store1, memory_ty.clone())?;
    let memory2 = Memory::new(&mut store2, memory_ty)?;

    let instance1 = Instance::new(&mut store1, &module1, &[memory1.into()])?;
    let instance2 = Instance::new(&mut store2, &module2, &[memory2.into()])?;
    let main1 = instance1.get_func(&mut store1, "main")
        .expect("`main` was not an exported function");
    let main2 = instance2.get_func(&mut store2, "main")
        .expect("`main` was not an exported function");
    let params = vec![
        Val::I32(4097), 
        Val::V128(0x10203cccdcecf807f7e7dfffefdfc.into()), 
        Val::V128(0xd7a950c126c2ce62f786d3c83f7914e.into()),
    ];
    let mut results1 = vec![Val::null()];
    let mut results2 = vec![Val::null()];
    println!("Opt level None: {:?}", main1.call(
        &mut store1,
        &params,
        &mut results1
    ));
    println!("--------------");
    println!("Opt level Speed: {:?}", main2.call(
        &mut store2,
        &params,
        &mut results2
    ));

    Ok(())
}
```

```toml
[package]
name = "wasmtime-wrapper"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
wasmtime = "18.0.3"
```

### Steps to Reproduce

```bash
cargo run --release
```

### Expected Results

Both results from none and speed optimizations should show the same result (both in error or successful execution)

### Actual Results

```
Opt level None: Err(error while executing at wasm backtrace:
    0:   0x59 - <unknown>!<wasm function 0>

Caused by:
    0: memory fault at wasm address 0x10000 in linear memory of size 0x10000
    1: wasm trap: out of bounds memory access)
--------------
Opt level Speed: Ok(())
```

### Versions and Environment

- wasmtime version: v18.0.3 (but also on v18.0.2)
- Operating system & architecture: Ubuntu 22.04.3 LTS, Arch: x86_64

### Extra Info

- The test case is not fully minimized - need to minimize more
- Does not work on other architectures (arm64, s390x, riscv64)

cfallin:
We have an instruction lowering rule on x86-64 that allows a `select` operation to perform load fusion: when presented with any xmm-register-typed values (`f32`, `f64`, or `v128`), an argument to the select can become a load. Unfortunately, this lowering behavior is incorrect in the case of narrower-than-128-bit values: the cmove is converted into an if-else diamond with two 128-bit moves and so the load becomes a full 128-bit-width load.

The fix is to disallow load fusion of selects of XMM-typed values. We could make the rules more fine-grained and keep `v128`-typed load fusion, but we're opting for the simpler and more conservative fix first here.

Fixes #8112.

### DIFF
--- a/cranelift/codegen/src/isa/x64/inst.isle
+++ b/cranelift/codegen/src/isa/x64/inst.isle
@@ -2880,7 +2880,14 @@
       (cmove ty cc consequent alternative))
 
 (rule (cmove_from_values (is_xmm_type ty) cc consequent alternative)
-      (cmove_xmm ty cc consequent alternative))
+      ;; force the `Value`s to `Xmm`s (in-register values) here:
+      ;; `cmove_xmm` can take an `XmmMemAligned`, but we don't want to
+      ;; allow load fusion here because `ty` may not be
+      ;; XMM-register-wide even if it is `is_xmm_type`. (E.g.,
+      ;; consider a cmove of `f64` values.)
+      (let ((cons Xmm consequent)
+            (alt Xmm alternative))
+        (cmove_xmm ty cc cons alt)))
 
 ;; Helper for creating `cmove` instructions with the logical OR of multiple
 ;; flags. Note that these instructions will always result in more than one
--- a/cranelift/filetests/filetests/isa/x64/xmm-select-load.clif
+++ b/cranelift/filetests/filetests/isa/x64/xmm-select-load.clif
@@ -1,0 +1,34 @@
+test compile precise-output
+target x86_64
+
+function %a(i32, f64, i64) -> f64 {
+block0(v0: i32, v1: f64, v2: i64):
+  v3 = load.f64 v2
+  v4 = select v0, v3, v1
+  return v4
+}
+
+; VCode:
+;   pushq   %rbp
+;   movq    %rsp, %rbp
+; block0:
+;   movsd   0(%rsi), %xmm5
+;   testl   %edi, %edi
+;   movsd %xmm0, %xmm0; jz $next; movsd %xmm5, %xmm0; $next:
+;   movq    %rbp, %rsp
+;   popq    %rbp
+;   ret
+;
+; Disassembled:
+; block0: ; offset 0x0
+;   pushq %rbp
+;   movq %rsp, %rbp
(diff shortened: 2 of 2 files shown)


cfallin:
> We could make the rules more fine-grained and keep v128-typed load fusion

And actually, for posterity, I just realized we *can't* do this because it makes the load conditional as well (which it can't be, per Wasm trap semantics). What fun! Given that, I think this conservative fix is the most reasonable one and there's no further refinement to do.

elliottt (approved):

PR #8117, opened 2024-03-13
x64: Remove load sinking from `XmmCmove`

alexcrichton:
This commit completely replaces the `XmmMemAligned` operand in `XmmCmove` with `Xmm` instead. Looking more into the fix #8113 I was looking to add some `*.wast` runtime tests to assert that the fix works at the wasm layer in addition to the instruction selection layer. I was poking around and there's a second user of `XmmCmove` which wasn't addressed in #8113.

In #8113 the only caller of `cmove_xmm` was updated to ensure that everything was always in memory. This was done to both prevent the upgrade-to-an-aligned-load from loading too much but additionally to ensure that the load always happened, regardless of the condition. There was a second constructor of `XmmCmove`, however, from `cmove_or_xmm`. This is triggered when the condition is a `f64.ne` instruction, for example, and ran into the same bug that #8112 was exposing.

To fix both of these and prevent any future issues about skipping a load by accident due to control flow this commit removes the `XmmMemAligned` argument entirely from `XmmCmove` and replaces it with `Xmm`. This prevents sinking loads entirely and sidesteps all these issues at the type level.

### DIFF
--- a/cranelift/codegen/src/isa/x64/inst.isle
+++ b/cranelift/codegen/src/isa/x64/inst.isle
@@ -243,7 +243,7 @@
        ;; XMM conditional move; overwrites the destination register.
        (XmmCmove (ty Type)
                  (cc CC)
-                 (consequent XmmMemAligned)
+                 (consequent Xmm)
                  (alternative Xmm)
                  (dst WritableXmm))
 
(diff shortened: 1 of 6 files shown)


cfallin (approved):
Thanks!

PR #8121, opened 2024-03-13
Run all `*.wast` tests in fuzzing

alexcrichton:
Currently we have a `spectest` fuzzer which uses fuzz input to generate an arbitrary configuration for Wasmtime and then executes the spec test. This ensures that no matter the configuration Wasmtime can pass spec tests. This commit expands this testing to include all `*.wast` tests we have in this repository. While we don't have a ton we still have some significant ones like in #8118 which will only reproduce when turning knobs on CPU features.


fitzgen (approved):
Nice!

PR #2685, opened 2021-02-24
Fix incomplete trap metadata due to multiple traps at one address.

cfallin:
If an instruction has more than one trap record associated with it (for
example: a divide instruction that has participated in load-op fusion,
so we have both a heap-out-of-bounds trap record due to its load and a
divide-by-zero trap record due to its divide op), the current MachBuffer
code would emit only one of the trap records to the sink.

Separately, divide instructions probably shouldn't merge loads, because
the two separate possible traps at one location might be confusing for
some embedders (certainly in Lucet). Divide seems to be the only case in
our current codegen where such merging might occur. This PR changes the
lowering to always force the divisor into a register.

Finally, while working out why trap records were not appearing, I had
noticed that `isa::x64::emit_std_enc_mem()` was only emitting heap-OOB
trap metadata for loads/stores when it had a srcloc. This PR ensures
that the metadata is emitted even when the srcloc is empty.

Note that none of the above presents a security or correctness problem;
trap metadata only affects the status that we return to the embedder
when a Wasm program terminates with a trap.

PR #8122, opened 2024-03-13
cranelift: Remove srcloc from emit state on all targets

jameysharp:
In #2426, @cfallin wrote:

> […] don't emit trap info unless an op can trap.
>
> This end result was previously enacted by carrying a SourceLoc on
> every load/store, which was somewhat cumbersome, and only indirectly
> encoded metadata about a memory reference (can it trap) by its
> presence or absence.

That PR changed both backends that existed at the time to check both the source location and the memory flags to determine whether a memory access could trap.

Then in #2685, @cfallin wrote:

> Finally, while working out why trap records were not appearing, I had
> noticed that isa::x64::emit_std_enc_mem() was only emitting heap-OOB
> trap metadata for loads/stores when it had a srcloc. This PR ensures
> that the metadata is emitted even when the srcloc is empty.

However that PR did not apply the same change to other backends. Since then, the pattern from #2426 has been copied to new backends.

I believe checking the source location has been unnecessary since #2426 and is now just a source of confusion at best, and possibly bugs at worst. So this PR makes all targets match the behavior of the x64 backend.

In addition, this pattern was the only reason why source locations were provided to any backend's emit state, so I'm removing that entirely. The `cur_srcloc` field has been unused on x64 since #2685.

This change is mostly straightforward, but there are two questionable changes in the riscv64 backend:

- The riscv64 backend had one use of this pattern for a BadConversionToInteger trap. All other uses on all backends were for HeapOutOfBounds traps. I suspect that was a copy-paste bug so I've removed it just like all the others.

- The riscv64 `Inst::Atomic` does not have a MemFlags field, so this means the HeapOutOfBounds trap metadata is added unconditionally for such instructions.


cfallin (approved):
LGTM -- thanks for finding this cleanup!

FWIW it should always be safe to emit extra trap records -- the "record exists and wasn't supposed to case" would matter only in that it could mask a separate bug, wherein the lowering generated a trap it wasn't supposed to, and turn it into a Wasm-level trap rather than a process crash. And that itself only if we use the instruction for internal logic separately from accesses that correspond to Wasm-level operators. So I think this is fine re: riscv64's `Atomic` MachInst.

PR #8124, opened 2024-03-13
cranelift: Remove support for WebAssembly tables

### LINKED ISSUE #5532: Cranelift: Remove `table_addr` and tables from CLIF, move them into `cranelift-wasm`'s Wasm -> CLIF lowering

jameysharp:
Wasmtime no longer needs any of this infrastructure and neither should anybody else.

This diff is nearly identical to @bjorn3's version of the same change, except I didn't remove Uimm64, which has started being used in other places. I forgot bjorn3 had already tackled this part until after I was already done, but it's reassuring that we both made the same changes.

https://github.com/bjorn3/wasmtime/commit/fb82ccb3948e949641a6d9581aa84472f68f97b8

Fixes #5532


cfallin (approved):
LGTM! This is good code motion that makes Cranelift maintenance easier while moving tables up a layer where they can benefit from other opts too. Thanks for pushing it forward.

PR #8125, opened 2024-03-13
cranelift-wasm: Emit constant bounds for fixed-size tables

jameysharp:
WebAssembly tables have a minimum size, but may optionally also have a maximum size. If the maximum is set to the same number of elements as the minimum, then we can emit an `iconst` instruction for bounds-checks on tables, rather than loading the bounds from memory.

That's a win in its own right, but when optimization is enabled, this also allows bounds-checks to constant-fold if the table index is provided as an integer constant.

### DIFF
--- a/cranelift/filetests/filetests/wasm/table-get-fixed-size.wat
+++ b/cranelift/filetests/filetests/wasm/table-get-fixed-size.wat
@@ -1,0 +1,74 @@
+;;! target = "x86_64"
+;;! optimize = true
+
+;; Test basic code generation for table WebAssembly instructions on
+;; non-resizeable tables. Use optimization but with opt-level "none" to
+;; legalize away macro instructions.
+
+(module
+  (table (export "table") 7 7 externref)
+  (func (export "table.get.const") (result externref)
+    i32.const 0
+    table.get 0)
+  (func (export "table.get.var") (param i32) (result externref)
+    local.get 0
+    table.get 0))
+
+;; function u0:0(i64 vmctx) -> r64 fast {
+;;     gv0 = vmctx
+;;     gv1 = load.i64 notrap aligned readonly gv0
+;;
+;;                                 block0(v0: i64):
+;;                                     v12 -> v0
+;; @0052                               v2 = iconst.i32 0
+;; @0054                               v3 = iconst.i32 7
+;; @0054                               v4 = icmp uge v2, v3  ; v2 = 0, v3 = 7
+;; @0054                               brif v4, block2, block3
+;;
+;;                                 block2 cold:
+;; @0054                               trap table_oob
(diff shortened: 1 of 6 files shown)


alexcrichton (approved):
Looks good to me!

Before I hit the big green button though, two questions/observations:

* As this gets more complicated I fear that keeping `func_environ.rs` in sync with `dummy.rs` is going to get a bit cumbersome. I know that's a longstanding issue that we want to test the clif of wasmtime's output, but I'd probably caution going too far down the route of mirroring all table stuff in the two files. Either that, or do you have plans to move some of this instead to `cranelift-wasm`? 
* One thing I'm a little sad about here is that we get to give Cranelift a bunch of constants but because they're involved in control flow it may not end up helping a whole lot. For example in the examples you added here the `brif` can become `br`, but I don't think we currently optimize that. In lieu of adding more optimizations, do you think that the frontend will be able to detect two constants and omit the control flow entirely? Basically doing that form of constant folding in the frontend rather than in the middle-end.

To be clear I don't think either of these need to be addressed before this PR lands, so you can feel free to hit the button as well, figured this'd be a good place to ask anyway!

jameysharp:
Chris' response accurately reflects my plans for how to address your second question, Alex. I'm currently working on getting rid of the `trapnz`, and it works on x64, but I need to do a bit more work for the rest of the backends. #8122 was the first step on that. Once that is out of the way, the existing constant-folding optimizations get us most of the way to where we want to be, and as Chris said, eliminating `select_spectre_guard` when its condition is constant does the rest.

To your first question, I've been trying to push as much as I can into cranelift-wasm shared code, but many details (like "how do I discover the current table bounds", lazy table initialization, and GC refcount management) are unfortunately specific to a particular runtime, so I haven't been able to find much opportunity to share more than this. I'd love to hear suggestions though!

alexcrichton:
Oh nice, yeah changing to select{,_spectre} sounds like a nice way to solve this without much work in the frontend. And agreed that it'd be great for `i32.const N` + `call_indirect` to be load/call in the most optimal case.

> I'd love to hear suggestions though!

Ah yeah if all the major bits go in cranelift-wasm then I think that's fine, I think I over-rotated on the `let bound = ...` bit that's basically the same in `func_environ.rs` and `dummy.rs`.

AFAIK there's no use case for keeping `dummy.rs` though other than "we haven't figure out a way to test Wasmtime". I might be just so inspired by this to take a stab at migrating all the tests to using Wasmtime's codegen...

jameysharp:
I would love to have CLIF and compiled-output tests that run through Wasmtime instead of using the dummy environment, personally! My current sense is there are plenty of reasons why that's difficult, but if you get it working I'm all for it.

PR #8130, opened 2024-03-14
Reorder book-based documentation

rikhuijzer:
This PR suggests to reorder the book-based documentation. The main change is moving the CLI chapter before the API (embedding) chapter like was proposed in #7987 by @alexcrichton. The rest is mostly cleaning up of duplicate or outdated information.

With these changes, this PR fixes some confusion that I had with the current layout. Currently, it's not so clear how the Subsection "2.3 Embedding in Rust" relates to Chapter "3. Using WebAssembly from your language" with Subsection "3.1 Rust". With these changes, these two examples are merged and now all fall under Chapter "2. Using the Wasmtime API".

More specifically, this is what this PR changed:

1. Move the CLI chapter before the API chapter (like proposed in #7987).
1. Move some Rust and C examples from the "Examples" chapter into the API chapter.
1. Remove the original main C and Rust chapter introduction files. They contained mostly outdated or duplicate information.
1. Rename the "Examples" chapter to "Further Examples". (For lack of a better name.)
1. Rename "Embedding" to "API" at some places since API is a more generally known term. The term "embedding" is also slightly more confusing, I think, because the term nowadays also used a lot in LLM-related communication.
1. Fix some outdated links and documentation by reading through the entire book.
1. Fix a missing subheading for Elixir (it was mentioned on the lang page, but not in the menu).


cfallin:
Reassigning review to @alexcrichton as it's not totally clear to me how this should relate with the still-pending #7987. Thanks @rikhuijzer for the careful thought around all of these changes though; they do seem like reasonable cleanups to me!

alexcrichton (approved):
Thanks for this! I like the organization better here as well 👍

Issue #8132, opened 2024-03-14
Cranelift: s390x wrong execution of i16x8.extmul_high_i8x16_s with spilled arguments


candymate:
### Test Case

```rs
// main.rs
use wasmtime::*;

fn main() -> Result<()> {
    let mut config = Config::default();
    config.strategy(Strategy::Cranelift);
    config.cranelift_opt_level(OptLevel::None);

    let engine = Engine::new(&config)?;
    let wat = r#"
    (module
        (type (;0;) (func (param v128 v128 v128 v128 v128 v128 v128 v128 v128) (result v128)))
        (import "mem" "mem" (memory (;0;) 1))
        (func (;0;) (type 0) (param v128 v128 v128 v128 v128 v128 v128 v128 v128) (result v128)
          local.get 5
          local.get 8
          i16x8.extmul_high_i8x16_s)
        (export "main" (func 0)))
    "#;
    let module = Module::new(&engine, wat)?;
    let mut store = Store::new(&engine, ());

    let memory_ty = MemoryType::new(1, None);
    let memory = Memory::new(&mut store, memory_ty.clone())?;

    let instance = Instance::new(&mut store, &module, &[memory.into()])?;
    let main = instance.get_func(&mut store, "main")
        .expect("`main` was not an exported function");
    let params = vec![
        Val::V128(0.into()), // this is unused, but needed
        Val::V128(0.into()), // this is unused, but needed
        Val::V128(0.into()), // this is unused, but needed
        Val::V128(0.into()), // this is unused, but needed
        Val::V128(0.into()), // this is unused, but needed
        Val::V128(0xfffefdfccccdcecf807f7e7d00010203.into()), 
        Val::V128(0.into()), // not sure why - this is unused, but needed
        Val::V128(0.into()), // not sure why - this is unused, but needed
        Val::V128(0x1e33ad29b5b0951d3c5c954764cd108d.into())
    ];
    let mut results = vec![Val::V128(0.into())];
    println!("Opt level None: {:?}", main.call(
        &mut store,
        &params,
        &mut results
    ));
    println!("{:?}", results);

    Ok(())
}
```

```
[package]
name = "wasmtime-wrapper"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
# wasmtime = "18.0.3"
[...]

PR #8136, opened 2024-03-14
s390x: Fix ABI vector-lane swap for overflow args

uweigand:
On s390x we support ABIs with little-endian or big-endian vector lane ordering.  When calling to function with a different lane order, vector arguments need to be lane-swapped.  This currently happens for arguments passed in registers, but not for those passed on the stack - that is a bug.

Fixed by swapping stack arguments as well.

Fixes https://github.com/bytecodealliance/wasmtime/issues/8132


cfallin (approved):
LGTM!

PR #8138, opened 2024-03-14
Enable `-Scommon` by default for `wasmtime serve`

### LINKED ISSUE #8086: Change `wasmtime serve` to apply `-S common` by default
People keep running into problems with `wasmtime serve` because by default it only exposes `wasi:proxy`, nothing else. I think we should change the behavior here to imply `-S common` to make default toolchains and in particular the normal Preview1 adapter work.

One could see it as a downside that `wasmtime serve` would be a bit less specifically about HTTP, but I can't really come up with compelling arguments for why that'd be a problem.

alexcrichton:
This commit enables more WASI interfaces by default with `wasmtime serve`, notably through the use of the `-Scommon` flag. This means that the interfaces supported by `wasmtime serve` are the same as `wasmtime run` by default.

To fully implement this there's a number of refactorings here:

* More flags like `--dir` and `--env` are moved into `RunCommon` to be shared between `wasmtime serve` and `wasmtime run`, meaning that the `serve` command can now configure environment variables.

* A small test has been added as well as infrastructure for running tests with `wasmtime serve` itself. Previously there were no tests that executed `wasmtime serve`.

* The `test_programs` crate had a small refactoring to avoid double-generation of http bindings.

Closes #8086

### DIFF
--- a/crates/test-programs/src/bin/api_proxy.rs
+++ b/crates/test-programs/src/bin/api_proxy.rs
@@ -1,18 +1,12 @@
-pub mod bindings {
-    wit_bindgen::generate!({
-        path: "../wasi-http/wit",
-        world: "wasi:http/proxy",
-        default_bindings_module: "bindings",
-    });
-}
-
-use bindings::wasi::http::types::{IncomingRequest, ResponseOutparam};
+use test_programs::wasi::http::types::{
+    Headers, IncomingRequest, OutgoingBody, OutgoingResponse, ResponseOutparam,
+};
 
 struct T;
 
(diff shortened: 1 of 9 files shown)


tschneidereit:
Thank you for doing this, and fully aligning it with `run` while you were at it! ❤️

alexcrichton:
I've extracted refactorings here to https://github.com/bytecodealliance/wasmtime/pull/8279 and will leave the question of defaults to https://github.com/bytecodealliance/wasmtime/issues/8086

PR #8142, opened 2024-03-14
PCC: fuzz static (no dynamic checks) case in Wasmtime fuzzers.

cfallin:
This PR updates the fuzzing configuration generation logic to turn on proof-carrying code (PCC) when randomly selected. When PCC is on, it forces the memory configuration to one that PCC can handle: a statically-sized 4GiB memory with 2GiB guard region, so no dynamic checks occur. I have fuzzed this for ~100 CPU-hours so far (on x86-64) and haven't found anything, with a local change that forces the PCC arbitrary-bool to true to concentrate fuzzing effort in just the new bit. (I've confirmed that manually introducing failures into the PCC code does properly bubble up.)

The commits start with one that updates the `pcc_memory` integration test to put all the various loads together in one body such that the optimizer will merge some of their address-computation logic. This was the root of failures I was seeing when last working on PCC in the fall, and that still exist for dynamic-check cases. A subsequent commit reduces the coverage of that test to only the 4GiB/2GiB static case for now so it passes; in this PR I am arguing that we should support this and push it forward first, and fix the dynamic case later (more below).

This PR includes rebases of two commits from @elliottt's #7511, and then a last change to force the memory configuration to the supported static one only when PCC is selected by fuzzing.

Why only the static case? I've made approximately three attempts to get the general dynamic case to work, be fully general, handle all edge-cases, and be scalable (not blow up quadratically): combined static+dynamic ranges in all facts, a more flexible set-of-lower/upper-bound-expressions engine (#7965), and just recently an implementation of the "order graph" I described in the last comment on that PR. All of these suffer from some combination of maddening complexity and a quadratic worst-case when N accesses merge together and we have to symbolically reason about all of them together -- to the point that I know when to cut my losses and would rather ship the static case without digging further. I do think there's maybe something more possible, but it needs more thought and time when not under pressure (and under demand by a few other projects too).

Assuming this sticks and fuzzbugs if any are manageable, I plan to eventually propose an "on but permissive by default" approach (with warnings but no failures), then eventually an "on and enforcing by default", but all of those things will come after more experience and discussions with folks. For now I just want the fuzzers to do their thing!

### DIFF
--- a/cranelift/codegen/src/result.rs
+++ b/cranelift/codegen/src/result.rs
@@ -77,6 +77,9 @@ impl std::fmt::Display for CodegenError {
             #[cfg(feature = "unwind")]
             CodegenError::RegisterMappingError(_0) => write!(f, "Register mapping error"),
             CodegenError::Regalloc(errors) => write!(f, "Regalloc validation errors: {:?}", errors),
+
+            // NOTE: if this is changed, please update the `is_pcc_error` function defined in
+            // `wasmtime/crates/fuzzing/src/oracles.rs`
             CodegenError::Pcc(e) => write!(f, "Proof-carrying-code validation error: {:?}", e),
         }
     }
--- a/crates/fuzzing/src/generators/config.rs
+++ b/crates/fuzzing/src/generators/config.rs
@@ -170,7 +170,8 @@ impl Config {
                 self.wasmtime.memory_guaranteed_dense_image_size,
             ))
             .allocation_strategy(self.wasmtime.strategy.to_wasmtime())
-            .generate_address_map(self.wasmtime.generate_address_map);
+            .generate_address_map(self.wasmtime.generate_address_map)
+            .cranelift_pcc(self.wasmtime.pcc);
 
         if !self.module_config.config.simd_enabled {
             cfg.wasm_relaxed_simd(false);
@@ -427,6 +428,9 @@ pub struct WasmtimeConfig {
     native_unwind_info: bool,
     /// Configuration for the compiler to use.
     pub compiler_strategy: CompilerStrategy,
+
+    /// Whether or not fuzzing should enable PCC.
+    pcc: bool,
 }
 
 impl WasmtimeConfig {
--- a/crates/fuzzing/src/generators/config.rs
+++ b/crates/fuzzing/src/generators/config.rs
@@ -1,7 +1,8 @@
 //! Generate a configuration for both Wasmtime and the Wasm module to execute.
 
 use super::{
-    CodegenSettings, InstanceAllocationStrategy, MemoryConfig, ModuleConfig, UnalignedMemoryCreator,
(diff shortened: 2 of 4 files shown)


jameysharp (approved):
This looks sensible to me! I'm excited to start having PCC in our testing rotation.

alexcrichton:
> I have fuzzed this for ~100 CPU-hours so far (on x86-64) and haven't found anything

I wanted to query about this, are you sure this was on x86-64? I'm pretty surprised how quickly so many fuzz bugs are being found related to pcc on oss-fuzz if this much fuzzing was done before this PR. That makes me worry about the effectiveness of local fuzzing vs oss-fuzz. 

Are you able to perhaps double-check to make sure that this PR was fuzzed as-is on x64?

cfallin:
> I wanted to query about this, are you sure this was on x86-64?

Yes.

> Are you able to perhaps double-check to make sure that this PR was fuzzed as-is on x64?

Yes, it was.

alexcrichton:
Which fuzzer were you using? Checking out the merge commit here it takes ~10 seconds locally from an empty corpus to find an error with the `instantiate` fuzzer for me.

cfallin:
`differential`; I just verified that, on an x86-64 machine (18-core/36-thread Xeon), with commit `acbbf342b704d88b87312beda1c1a3b45b0ba283` (the merge commit for this PR), the command `cargo +nightly fuzz run -s none differential -j32` (same as I tested before) runs for quite a while without finding anything.

so
* `-s none` is my usual thing to get better throughput, because a compiler logic bug shouldn't require ASan to find (tell me if this is wrong?), and
* `differential` maybe is slower because it actually runs the code? or is throwing away test cases?

As mentioned above I did verify it was hitting PCC codepaths because injected bugs were found.

I do see pretty low throughput, e.g. from early in the run

```
#87140: cov: 69340 ft: 55743 corp: 1144 exec/s 56 oom/timeout/crash: 0/0/0 time: 91s job: 55 dft_time: 0
```

or 56 executions/second. Maybe my mistake is not using `instantiate`; I use `differential` out of {my, my commandline history database}'s habit, mostly...

cfallin:
(For completeness, I ran `differential` for ~10 minutes before killing, on 32 threads; no fuzzbugs; reached 1.4M testcase executions)

To add to that, I too get a pretty immediate fuzzbug (after 8 seconds) when running the `instantiate` target; so indeed, I'll eschew `differential` entirely unless its comparison powers are actually needed.

alexcrichton:
Weird! I wonder if your preexisting corpus might be holding you back somehow?

I'm having weird issues running v8 locally where it causes a segfault and the reproduction case never reproduces. If I do `ALLOWED_ENGINES=-v8` then locally I find a fuzz bug on the merge commit pretty quickly.

I'm not sure how to explain the fuzzing behavior here... but regardless glad oss-fuzz is running all the fuzzers still and has proven worthwhile!

I was mostly worried that we had some bad mutations or ... something ... but I dunno if I can really draw any conclusion from this...

PR #8144, opened 2024-03-15
Turn most cranelift-wasm wasmtests into disas tests

jameysharp:
These tests mostly verify compiler behavior for specific WebAssembly constructs or small combinations.

Currently they're only tested to make sure that cranelift-wasm doesn't crash or report an error. But they'd be even more useful if we track what CLIF we actually produce from Wasmtime.

So I'm moving them to the new disas test framework and adding basic CLIF test expectations for them.

I've excluded the .wat files which were produced by Emscripten or Rust, because they're very large and their compiled CLIF is not very interesting to look at.

For review purposes, the first commit moves the files and the second commit adds the test expectations to them.


alexcrichton (approved):
👍

PR #3509, opened 2021-11-05
Fix a panic with an invalid name section

alexcrichton:
This commit fixes a panic which can happen on a module with an invalid
name section where one of the functions named has the index `u32::MAX`.
Previously Wasmtime would create a new `FuncIndex` with the indices
found in the name section but the sentinel `u32::MAX` causes a panic.

Cranelift otherwise limits the number of functions through `wasmparser`
which has a hard limit (lower than `u32::MAX`) so this commit applies a
fix of only recording function names for function indices that are
actually present in the module.

PR #8147, opened 2024-03-15
Move all tests out of `cranelift-wasm`

alexcrichton:
This commit moves all remaining tests out of the `cranelift-wasm` crate into other various test suites. My end goal is to eventually remove the `Dummy*` trait for wasm translation so we don't have to keep updating that over time, but for now these tests I think reside in a more appropriate location nowadays.


abrown (approved):

jameysharp:
I was too slow to make these changes and you got to them first :laughing: 

I had done a bit more historical research about the issue-1306 test case which I'll throw here in case anyone wonders later:

This test case was extracted from a fuzz bug in 2019, before Cranelift was merged into the Wasmtime repo:
https://github.com/bytecodealliance/cranelift/issues/1306

That test is interesting, but running it here only tests code in DummyEnvironment, not anything reachable from Wasmtime. There was a report of exactly the same bug in Wasmtime two years later (#3509), and the fix included tests in `tests/misc_testsuite/empty.wast` to cover this case.

The `empty.wast` test is better than this one because it has minimal and well-commented binary modules which I've verified still trigger the same bug if the fix is reverted. By contrast, since this older test was produced by a fuzzer, it is larger than necessary and the end of its name section is malformed and unparseable.

Also, I had concluded that the reachability tests were pretty well covered by `tests/disas/*reach*.wat`, but your addition of `tests/disas/dead-code.wat` certainly can't hurt.

Issue #8135, opened 2024-03-14
Investigate use of stack probes and removal of explicit stack-limit checks


cfallin:
Currently, Wasmtime uses explicit stack-limit checks to avoid overflowing the stack, and these checks are generated in the prologue of every Wasm function. This is largely a consequence of the design decision to run on the host's stack (when not doing async): the two subcases are either Wasm hits the guard page itself (and we don't currently have a mechanism to recover from that, with a separate signal stack) or Wasm uses almost all the stack then calls back into the host and the host then overflows (which looks like a host crash). So the advantage of explicit stack-limit checks is that they retain control while giving us a well-defined path to return a trap, but the cost is that they impose overhead on every function call. (And potentially nontrivial overhead: the limit is several chained loads deep in data structures.)

The alternative idea is to switch to reliance on a guard page, with stack probes in functions that have frames larger than a single guard page (just as with native code today), and handle the two cases above by:

- Installing a sigaltstack and handling a trap in the function prologue as a stack-overflow Wasm trap;
- Explicitly checking against the stack limit on exit from the Wasm code back to host code.

(Not coincidentally, I believe this is also the exact scheme that was used by Lucet!)

alexcrichton:
Some other bits I've thought of after more thinking:

* There's one more way to exit wasm which isn't via trampolines, and that's via libcalls. Here there's two kinds of libcalls:
  * One kind of libcall is cranelift-inserted and is for things like `fma` when the architecture doesn't have fma. (also things like `ceil` without sse4.2). These relocations are resolved [here](https://github.com/bytecodealliance/wasmtime/blob/c44783347b398019d67b61e9d360012387ce7906/crates/wasmtime/src/runtime/code_memory.rs#L295) and we'd have to install custom versions of all libcalls that do stack checks before calling the native function. AFAIK we don't have the option to rely on cranelift-generated trampolines here.
  * Another kind of libcall is Wasmtime's libcalls for `memory.grow` and the like. We'll need to [modify these macros](https://github.com/bytecodealliance/wasmtime/blob/c44783347b398019d67b61e9d360012387ce7906/crates/runtime/src/arch/x86_64.rs#L63-L84) to handle the stack check too. Ideally though we'd rely on cranelift-generated trampolines here somehow, but that would be a refactoring we don't currently have in place.
[...]

PR #8149, opened 2024-03-15
Enhance test around stack overflow

alexcrichton:
This commit enhances the `host_always_has_some_stack` a bit in light of some thinking around #8135. Notably this test ensures that the host itself never segfaults even if the wasm exhausts all of its stack. There are a number of ways that we can exit out to the host, though, and only one was tested previously. This commit updates to ensure more cases are covered.

### DIFF
--- a/tests/all/stack_overflow.rs
+++ b/tests/all/stack_overflow.rs
@@ -9,35 +9,98 @@ fn host_always_has_some_stack() -> Result<()> {
     // assume hosts always have at least 128k of stack
     const HOST_STACK: usize = 128 * 1024;
 
-    let mut store = Store::<()>::default();
+    let mut store = if cfg!(target_arch = "x86_64") {
+        let mut config = Config::new();
+        // Force cranelift-based libcalls to show up by ensuring that platform
+        // support is turned off.
+        unsafe {
+            config.cranelift_flag_set("has_avx", "false");
+            config.cranelift_flag_set("has_sse42", "false");
+            config.cranelift_flag_set("has_sse41", "false");
+            config.cranelift_flag_set("has_ssse3", "false");
+            config.cranelift_flag_set("has_sse3", "false");
+        }
+        Store::new(&Engine::new(&config)?, ())
+    } else {
+        Store::<()>::default()
+    };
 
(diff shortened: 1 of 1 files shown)


tests/all/stack_overflow.rs
```diff
@@ -9,35 +9,97 @@ fn host_always_has_some_stack() -> Result<()> {
...
+
+                ;; exit via wasm-to-array trampoline
+                (func $recursive2 (export "f2")
+                    call $host2
+                    call $recursive2)
+
+                ;; exit via a wasmtime-based libcall
+                (memory 1)
+                (func $recursive3 (export "f3")
+                    (drop (memory.grow (i32.const 00)))
```
fitzgen, line 54:
```suggestion
                    (drop (memory.grow (i32.const 0)))
```

tests/all/stack_overflow.rs
```diff
@@ -9,35 +9,97 @@ fn host_always_has_some_stack() -> Result<()> {
...
     // Additionally, however, and this is the crucial test, make sure that the
     // host function actually completed. If HITS is 1 then we entered but didn't
     // exit meaning we segfaulted while executing the host, yet still tried to
     // recover from it with longjmp.
-    assert_eq!(HITS.load(SeqCst), 0);
+    assert_eq!(hits1, 0);
+    assert_eq!(hits2, 0);
+    assert_eq!(hits3, 0);
+    assert_eq!(hits4, 0);
+    assert_eq!(hits5, 0);
```
fitzgen, line 102: We never store to `HITS`; I assume we did at one point in time, but not anymore. Therefore, this isn't really testing what the comment says it is.

fitzgen (approved):
LGTM modulo comment below.

tests/all/stack_overflow.rs
alexcrichton: Oh the store is [just below in `test_host_stack`](https://github.com/bytecodealliance/wasmtime/blob/9d1e2e0aa4e3d4153f8638085e0029cbc722c774/tests/all/stack_overflow.rs#L45)

PR #8150, opened 2024-03-15
Don't lookup trap codes twice on traps

alexcrichton:
Currently whenever a signal or trap is handled in Wasmtime we perform two lookups of the trap code. One during the trap handling itself to ensure we can handle the trap, and then a second once the trap is handled. There's not really any need to do two here, however, as the result of the first can be carried over to the second.

While I was here refactoring things I also changed how some return values are encoded, such as `take_jmp_buf_if_trap` now returns a more self-descriptive enum.

### DIFF
--- a/crates/environ/src/trap_encoding.rs
+++ b/crates/environ/src/trap_encoding.rs
@@ -98,6 +98,42 @@ pub enum Trap {
     // if adding a variant here be sure to update the `check!` macro below
 }
 
+impl Trap {
+    /// Converts a byte back into a `Trap` if its in-bounds
+    pub fn from_u8(byte: u8) -> Option<Trap> {
+        // FIXME: this could use some sort of derive-like thing to avoid having to
+        // deduplicate the names here.
+        //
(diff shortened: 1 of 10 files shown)


fitzgen (approved):

crates/environ/src/trap_encoding.rs
```diff
@@ -98,6 +98,42 @@ pub enum Trap {
...
+    pub fn from_u8(byte: u8) -> Option<Trap> {
+        // FIXME: this could use some sort of derive-like thing to avoid having to
+        // deduplicate the names here.
+        //
+        // This simply converts from the a `u8`, to the `Trap` enum.
+        macro_rules! check {
+        ($($name:ident)*) => ($(if byte == Trap::$name as u8 {
+            return Some(Trap::$name);
+        })*);
+    }
```
bjorn3, line 112:
```suggestion
            ($($name:ident)*) => ($(if byte == Trap::$name as u8 {
                return Some(Trap::$name);
            })*);
        }
```

PR #8151, opened 2024-03-15
Delay argument location use in `CallSite::gen_arg`

elliottt:
In preparation for handling callee-save registers in the tail calling convention, split the logic for handling argument locations out of `CallSite::gen_arg`. This allows us to process the arguments as normal, but delay the actual move to the stack until later.

I also took the opportunity to directly emit the instruction vector at the end of `CallSite::gen_arg` instead of returning it, as this is what's done at every use of `CallSite::gen_arg` already. Unfortunately the intermediate vector is still necessary as all calls to emit would be underneath an immutable borrow of `ctx`, but this does clean things up a little bit.


cfallin (approved):
Seems reasonable to me!

PR #8152, opened 2024-03-15
Exit through Cranelift-generated trampolines for builtins

alexcrichton:
This commit changes how builtin functions in Wasmtime (think `memory.grow`) are implemented. These functions are required to exit through some manner of trampoline to handle runtime requirements for backtracing right now. Currently this is done via inline assembly for each architecture (or external assembly for s390x). This is a bit unfortunate as it's a lot of hand-coding and making sure everything is right, and it's not easy to update as it's multiple platforms to update.

The change in this commit is to instead use Cranelift-generated trampolines for this purpose instead. The path for invoking a builtin function now looks like:

* Wasm code calls a statically known symbol for each builtin.
* The statically known symbol will perform exit trampoline duties (e.g. pc/fp/etc) and then load a function pointer to the host implementation.
* The host implementation is invoked and then proceeds as usual.

The main new piece for this PR is that all wasm modules and functions are compiled in parallel but an output of this compilation phase is what builtin functions are required. All builtin functions are then unioned together into one set and then anything required is generated just afterwards. That means that only one builtin-trampoline per-module is generated per-builtin.

This work is inspired by #8135 and my own personal desire to have as much about our ABI details flowing through Cranelift as we can. This in theory makes it more flexible to deal with future improvements to our ABI.


alexcrichton:
I'll note that I'm opening this up as a draft because Winch doesn't work yet. I'll need to update invocation of those libcalls in addition to updating how trampolines are required. I plan on waiting until https://github.com/bytecodealliance/wasmtime/pull/8109 lands to fully implement that though to avoid implementing this new trampoline in Winch.

crates/cranelift/src/func_environ.rs
```diff
@@ -19,109 +20,74 @@ use cranelift_wasm::{
...
-                // If that ever changes, we will have to add a second type
-                // marker here.
-                AbiParam::new(I32).uext()
-            }
+    fn load_builtin(&mut self, func: &mut Function, index: BuiltinFunctionIndex) -> ir::FuncRef {
+        let cache = &mut self.builtins[index.index() as usize];
+        if let Some(f) = cache {
+            return *f;
+        }
+        self.required.push(index);
```
fitzgen, line 59: Should required be a set of some sort?

crates/runtime/src/libcalls.rs
```diff
@@ -69,18 +69,25 @@ use wasmtime_wmemcheck::AccessError::{
...
+/// Raw functions which are actually called from compiled code.
 ///
-/// Note that the trampolines here are actually defined in inline assembly right
-/// now to ensure that the fp/sp on exit are recorded for backtraces to work
-/// properly.
-pub mod trampolines {
+/// Invocation of a builtin currently looks like:
+///
+/// * A wasm function calls a cranelift-compile tramopline that's generated
+///   once-per-builtin.
```
fitzgen, line 77:
```suggestion
/// * A wasm function calls a cranelift-compiled tramopline that's generated
///   once-per-builtin.
```

crates/wasmtime/src/compile.rs
```diff
@@ -493,7 +508,26 @@ impl<'a> CompileInputs<'a> {
...
+        // Now that all functions have been compiled see if any wasmtime-builtin
+        // functions are necessary. If so those need to be collected and then
+        // those trampolines additionally need to be compiled.
+        let mut builtins = BTreeSet::new();
+        for output in raw_outputs.iter_mut() {
+            builtins.extend(output.builtins.drain(..));
+        }
+        let builtins = builtins.into_iter().collect::<Vec<_>>();
+        raw_outputs.extend(engine.run_maybe_parallel(builtins, |i| -> Result<_> {
+            let symbol = format!("wasmtime_builtin{}", i.index());
```
fitzgen, line 522: We could probably define a `fn builtin_name(builtin: BuiltinFunctionIndex) -> &'static str` method via the macros to get better names here, which could be useful in debugging/profiling.

fitzgen (commented):
Looks great! Excited to remove the last vestiges of hand-written asm.

crates/cranelift/src/func_environ.rs
alexcrichton: I think I'll actually refactor and remove this since I realized it's also present via relocations, so I'll handle this when I rebase over the winch trampolines going through Cranelift.

crates/wasmtime/src/compile.rs
```diff
@@ -496,32 +504,62 @@ impl<'a> CompileInputs<'a> {
...
-            K: PartialOrd,
-        {
-            items
-                .windows(2)
-                .all(|window| f(&window[0]) <= f(&window[1]))
-        }
-        debug_assert!(outputs
-            .values()
-            .all(|funcs| is_sorted_by_key(funcs, |x| x.key)));
-
```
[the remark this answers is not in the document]
fitzgen: I think because we rely on the order for when we resolve relocs? So if they are out of order, then we could resolve relocs to the wrong function, I think?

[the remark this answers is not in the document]
fitzgen: https://github.com/bytecodealliance/wasmtime/blob/a79cf76fe0995e8798db606495145ab80cb5cd4e/crates/wasmtime/src/compile.rs#L535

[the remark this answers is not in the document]
alexcrichton: Ok if that was historically the case then that's no longer needed. Definitely makes sense the order maps to the layout in the object itself, but for relocations order no longer matters since all relocations are tracked by an ID that's not an index.

alexcrichton:
Ok various prerequisites are now landed on `main` and I've additionally implemented this for Winch too. The Winch changes are pretty nontrivial, so I'm going to request review from @saulecabrera and @elliottt, so if one of y'all could look at those and make sure I didn't do anything too weird it'd be much appreciated!

fitzgen (approved):

crates/runtime/src/libcalls.rs
```diff
@@ -68,18 +68,25 @@ use wasmtime_wmemcheck::AccessError::{
...
-/// for libcalls.
+/// Raw functions which are actually called from compiled code.
 ///
-/// Note that the trampolines here are actually defined in inline assembly right
-/// now to ensure that the fp/sp on exit are recorded for backtraces to work
-/// properly.
-pub mod trampolines {
+/// Invocation of a builtin currently looks like:
+///
+/// * A wasm function calls a cranelift-compiled tramopline that's generated
```
elliottt, line 75:
```suggestion
/// * A wasm function calls a cranelift-compiled trampoline that's generated
```

winch/codegen/src/codegen/builtin.rs
```diff
@@ -12,26 +12,21 @@ use wasmtime_environ::{BuiltinFunctionIndex, PtrSize, VMOffsets, WasmValType};
...
     /// See [`cranelift_codegen::ir::LibCall`] for more details.
-    Known(LibCall),
+    LibCall(LibCall),
 }
 
 impl BuiltinType {
     /// Create a new dynamic built-in function type.
-    pub fn dynamic(index: u32, base: u32) -> Self {
-        Self::Dynamic { index, base }
+    pub fn builtin(idx: BuiltinFunctionIndex) -> Self {
```
saulecabrera, line 23: Do you mind updating the comment here and in `libcall` below?

winch/codegen/src/codegen/env.rs
```diff
@@ -85,38 +87,17 @@ pub struct HeapData {
...
+pub(crate) enum Callee {
     /// Locally defined function.
-    Local(&'a CalleeInfo),
+    Local(FuncIndex),
     /// Imported function.
-    Import(&'a CalleeInfo),
+    Import(FuncIndex),
     /// Function reference.
-    FuncRef(&'a ABISig),
+    FuncRef(TypeIndex),
```
saulecabrera, line 96: Nice!

saulecabrera (approved):
Winch pieces look great to me, thanks!

elliottt (approved):
Awesome!

PR #8156, opened 2024-03-16
winch: Correctly handle the result of memory grow

saulecabrera:
This commit correctly handles the  result of the memory grow builtin function. Previously, it was assumed that the result of memory grow must be of the the target's pointer type, which doesn't accurately represent the address space covered by the memory type.


alexcrichton (approved):

alexcrichton:
Ok Mr GitHub I'll lay off for a bit

PR #8159, opened 2024-03-17
Don't check for null in `call_indirect` and `call_ref`

### LINKED ISSUE #5291: Remove null-check in generated code for `call_indirect` (and `call_ref` soon)

alexcrichton:
This PR is an implementation of #5291 to slightly optimize the lowering of `call_indirect` and `call_ref` in Wasmtime. Explicitly checks for null function pointers are no longer present and instead we let a segfault happen when loading from a null function pointer. This segfault is caught and the relevant instruction is annotated with the appropriate trap code.

This support first starts by refactoring the `MemFlags` API to no longer purely be flags but instead be a mixture of flags and "flag regions". The `vmctx`/`heap`/`table` alias regions are bundled into two bits now and the various trap-related bits are now bundled into four bits. This enables putting arbitrary trap codes in a `MemFlags` so long as they aren't `TrapCode::User(_)`.

Closes #5291

### DIFF
--- a/crates/cranelift/src/func_environ.rs
+++ b/crates/cranelift/src/func_environ.rs
@@ -1155,11 +1155,6 @@ impl<'a, 'func, 'module_env> Call<'a, 'func, 'module_env> {
             self.env
                 .get_or_init_func_ref_table_elem(self.builder, table_index, callee);
 
-        // Check for whether the table element is null, and trap if so.
-        self.builder
-            .ins()
-            .trapz(funcref_ptr, ir::TrapCode::IndirectCallToNull);
-
         // If necessary, check the signature.
         match self.env.module.table_plans[table_index].style {
             TableStyle::CallerChecksSignature => {
@@ -1188,10 +1183,13 @@ impl<'a, 'func, 'module_env> Call<'a, 'func, 'module_env> {
                         .load(sig_id_type, mem_flags, signatures, offset);
 
                 // Load the callee ID.
+                //
+                // Note that the callee may be null in which case this load may
+                // trap. If so use the `IndirectCallToNull` trap code.
                 let mem_flags = ir::MemFlags::trusted().with_readonly();
                 let callee_sig_id = self.builder.ins().load(
                     sig_id_type,
-                    mem_flags,
+                    mem_flags.with_trap_code(Some(ir::TrapCode::IndirectCallToNull)),
                     funcref_ptr,
                     i32::from(self.env.offsets.ptr.vm_func_ref_type_index()),
                 );
@@ -1205,7 +1203,11 @@ impl<'a, 'func, 'module_env> Call<'a, 'func, 'module_env> {
             }
         }
 
-        self.unchecked_call(sig_ref, funcref_ptr, call_args)
+        // Note that `funcref_ptr` at this point is guaranteed to not be null as
+        // we've loaded its type which would have otherwise faulted. That means
+        // that further loads no longer need trap metadata, so use `None` here
+        // for the next trap code.
+        self.unchecked_call(sig_ref, funcref_ptr, None, call_args)
     }
 
     /// Call a typed function reference.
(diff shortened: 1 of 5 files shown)


cranelift/codegen/src/ir/memflags.rs
```diff
@@ -75,6 +69,8 @@ pub enum Endianness {
 #[derive(Clone, Copy, Debug, Hash, PartialEq, Eq)]
 #[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
 pub struct MemFlags {
+    // Initialized to all zeros to have all flags have their default value.
+    // This is the interpreted through various methods below.
```
cfallin, line 73: s/the interpreted/interpreted/

cfallin (approved):
LGTM, nice!

cranelift/codegen/src/ir/memflags.rs
```diff
@@ -326,32 +300,106 @@ impl MemFlags {
...
-        if self.notrap() {
-            None
-        } else if self.tabletrap() {
-            Some(TrapCode::TableOutOfBounds)
-        } else {
-            Some(TrapCode::HeapOutOfBounds)
+        match (self.bits & MASK_TRAP_CODE) >> TRAP_CODE_OFFSET {
+            0b0000 => Some(TrapCode::HeapOutOfBounds), // default mode
+            0b0001 => Some(TrapCode::StackOverflow),
+            0b0010 => Some(TrapCode::HeapOutOfBounds),
```
jameysharp, line 313: I'm inclined to think that we don't need to track whether `heap_oob` was set explicitly or just used as a default, so my instinct is to use just one encoding for it. I could easily be talked into the approach you've taken here, but I think it at least needs a bit more of a comment explaining why it appears twice here, and why the zero case doesn't appear in `with_trap_code` below.

jameysharp (commented):
I would kind of prefer to land the `MemFlags` changes separately from the "don't explicitly check for null function pointers" changes, if that's not too much trouble to split up. The commit history here is almost right for doing that with a quick rebase but I think only part of the "Trim the MemFlags API" commit applies.

I'm interested in whether you or Chris or anyone else have opinions on the other comments I've left below too.

alexcrichton:
> prefer to land the MemFlags changes separately

[Certainly!](https://github.com/bytecodealliance/wasmtime/pull/8162)

crates/cranelift/src/func_environ.rs
```diff
@@ -1215,19 +1217,9 @@ impl<'a, 'func, 'module_env> Call<'a, 'func, 'module_env> {
...
         args: &[ir::Value],
     ) -> WasmResult<ir::Inst> {
-        // Check for whether the callee is null, and trap if so.
-        //
-        // FIXME: the wasm type system tracks enough information to know whether
-        // `callee` is a null reference or not. In some situations it can be
-        // statically known here that `callee` cannot be null in which case this
-        // null check can be elided. This requires feeding type information from
-        // wasmparser's validator into this function, however, which is not
-        // easily done at this time.
```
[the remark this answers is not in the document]
alexcrichton: Good point, I'll be sure to add this back in when rebasing

cranelift/codegen/src/ir/memflags.rs
alexcrichton: Nah I think that makes the most sense in retrospect too, no need to be needlessly complicated.

crates/wasmtime/src/runtime/store.rs
```diff
@@ -1595,10 +1595,19 @@ impl StoreOpaque {
...
-        if addr == 0 {
+        // There are a few instances where a "close to zero" pointer is loaded
+        // and we expect that to happen:
+        //
+        // * Explicitly bounds-checked memories with spectre-guards enabled will
+        //   cause out-of-bounds accesses to get routed to address 0, so allow
+        //   wasm instructions to fault on the null address.
+        // * `call_indirect` when invoking a null function pointer may load data
+        //   from the a `VMFuncRef` whose address is null, meaning any field of
+        //   `VMFuncRef` could be the address of the fault.
```
fitzgen, line 1606: We should have a test somewhere, maybe here, that `mem::size_of<VMFuncRef>() < PAGE_SIZE`, since we should only rely on the presence of a single null guard page.

alexcrichton:
Updated and rebased!

Issue #8160, opened 2024-03-17
Wasmtime: avoid table lazy-init runtime checks on nullable funcref tables with null default


cfallin:
From [this comment](https://github.com/bytecodealliance/wasmtime/pull/8158#issuecomment-2002601881): it would be great to have an optimization such that

> if a table's default funcref is null, don't do lazy-init, and make the 0 bit-pattern mean null rather than non-initialized-pointer-to-anyfunc.

This comes from a circumstance where we know we'll initialize a table slot manually before calling it (not doing so is a program error that should lead to a trap), possibly because we already have a branch for the case where it is null at the program-logic level (e.g. fallback IC). We still must trap on null somehow but #8159 can do that implicitly by detecting a segfault in a load from a `NULL` anyfunc pointer. The above optimization then means that we do not need the [lengthy lazy-init-check sequence](https://github.com/bytecodealliance/wasmtime/blob/7fa71a4f9e34bdb3ba85679d46e9a31a955fdd3a/crates/cranelift/src/func_environ.rs#L954-L1002) on loads from such tables, and we can do so without requiring any initialization at instantiation time either.

cc @alexcrichton @jameysharp

alexcrichton:
I like this idea! Some implementation points I can think of off the top of my head:

* This would only be applicable when there are no `elem` segments with non-null entries for a table.
* We'd need to store runtime state for each table to know whether it contains tagged functions or non-tagged functions.
* In the `Module`'s type information we'd need to store this in a `TablePlan` as well so the table access codegen could take it into account as well as runtime instantiation

fitzgen:
See also https://github.com/bytecodealliance/wasmtime/issues/8002#issuecomment-2008139279

jameysharp:
An additional requirement, I think, is that we can only do this for tables which are not exported. If another module imports a table, it has no way of knowing at compile time whether this optimization applies.

I think this technique is still broadly applicable in spite of that though.

Nick pointed out that we could build a static page full of pointers which are null except the low bit is set, and CoW-map that page repeatedly as the initial table contents, if we know that the table is initially all null. That would be compatible with codegen that can't optimize away the lazy-init check, which would make it safe to export such tables.

[...]

PR #8161, opened 2024-03-17
Wasm ICs / typed-funcrefs test: switch to nullable table.

cfallin:
This is based on discussion in #8158: as noted in #8160, if we use a nullable typed funcref table instead (and given that we know we'll initialize a particular slot before use on the application side, so we won't actually call a null ref), and if we have a null-ref default value, we should be able to avoid the lazy table-init mechanism entirely.

(Ignore the part where this module doesn't actually have any update logic that would set non-null refs anywhere; it's a compile-test, not a runtest!)

Once #8159 is merged and #8160 is implemented (and table-at-constant-index is fully optimized), we should see zero branches in this test.

### DIFF
--- a/tests/disas/typed-funcrefs.wat
+++ b/tests/disas/typed-funcrefs.wat
@@ -14,9 +14,9 @@
   (type $ic-stub (func (param i32 i32 i32 i32) (result i32)))
 
   ;; This syntax declares a table that is exactly 100 elements, whose
-  ;; elements are non-nullable function references, and whose default
-  ;; value (needed because non-nullable) is a pointer to `$ic1`.
-  (table $ic-sites 100 100 (ref $ic-stub) (ref.func $ic1))
+  ;; elements are nullable function references, and whose default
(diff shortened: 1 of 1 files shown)


cfallin:
(I switched back from `call_indirect` to `table.get` + `call_ref` because it seems the former doesn't actually omit callee signature checks even when the callee is a typed funcref)

jameysharp (approved):
Looks great. The collection of optimizations we're discussing should be easy and should show up very well on this test. I'm looking forward to seeing each one shrink this!

alexcrichton:
> it seems the former doesn't actually omit callee signature checks even when the callee is a typed funcref

Ah yes I didn't mean to imply that we already did it, only that it's not too hard to add. I've got a commit after https://github.com/bytecodealliance/wasmtime/pull/8159 which implements the optimization and the codegen for `call_indirect` vs `table.get` + `call_ref` is the same after that.

PR #8162, opened 2024-03-18
Refactor the `MemFlags` internal representation

alexcrichton:
This commit refactors the internal `MemFlags` bits to be not just flags.
Instead some bits are now grouped together and interpreted as a unit.
This enables two primary API changes:

* First the `heap`/`table`/`vmctx` split is now represented as an "alias
  region" which is set as an enum. This means that all `MemFlags` carry
  an `Option<AliasRegion>` internally.

* Second trapping state is now represented as an `Option<TrapCode>`.
  This means that `notrap` is no longer a flag and `tabletrap` is no
  longer a flag. This does enable storing arbitrary trap codes though so
  long as they aren't `TrapCode::User(_)`.

The main purpose of this commit is to enable using more trap codes with
`MemFlags` for when a segfault is detected. For example with https://github.com/bytecodealliance/wasmtime/issues/5291 we
want a segfault to indicate a call-to-null, which is not currently
covered by `MemFlags`.


cranelift/codegen/src/ir/memflags.rs
```diff
@@ -91,35 +102,101 @@ impl MemFlags {
...
-                if (bits & (1 << FlagBit::LittleEndian as usize)) != 0
-                    && (bits & (1 << FlagBit::BigEndian as usize)) != 0
-                {
-                    false
-                } else {
-                    self.bits = bits;
-                    true
+    /// Returns true if the flag was found and set, false for an unknown flag
+    /// name.  Will also return false when trying to set inconsistent endianness
+    /// flags.
```
jameysharp, line 150: I think this comment no longer applies; the "Errors" section below covers semantic errors, `set_by_name` doesn't return `false` for them any more.

```suggestion
    /// name.
```

jameysharp (approved):
This looks great, after one minor doc-comment fix I think. Very clear and reviewable, thank you!

Issue #7916, opened 2024-02-12
Cranelift: Huge stack slot results in panic rather than an error returned from codegen


bjorn3:
### `.clif` Test Case

```
test compile
target x86_64

function u0:12() system_v {
    ss0 = explicit_slot 1073741824
    ss1 = explicit_slot 16
    ss2 = explicit_slot 1073741824
    ss3 = explicit_slot 16
    ss4 = explicit_slot 48

block0:
    v21 = stack_load.i8 ss3
    trap user0
}
```

### Steps to Reproduce

* Compile above test.

### Expected Results

Returns an error from `compile_function`.

### Actual Results

Panics with `Offset in NominalSPOffset is greater than 2GB; should hit impl limit first: TryFromIntError(())`

### Versions and Environment

Cranelift version or commit: 0.104

Operating system: N/A

Architecture: x86_64

### Extra Info

Reported by @matthiaskrgr at https://github.com/rust-lang/rustc_codegen_cranelift/issues/1434

bjorn3:
It looks like the referenced impl limit only applies to arguments passed on the stack, not to local stack slots.

PR #8163, opened 2024-03-18
Cranelift: shrink ABIArgSlot

jameysharp:
ABIArgSlot is currently 16 bytes with 4 bytes of padding. Shrinking the offset field in its Stack variant from i64 to i32 therefore reduces the enum to 8 bytes. It also reduces its alignment to 4 bytes, which might reduce padding in containing structures.

Some targets already limit stack frame sizes to 2GB or less, and in practice they must be much smaller than that, so an i32 is plenty.

By making this type only one word in size, we can put two of them in ABIArgSlotVec for free, which guarantees that such SmallVecs will never spill to the heap. In principle we could use arrayvec or something that doesn't support spilling to the heap at all.

There are many other places where we use i64 for stack frame offsets which would probably benefit from switching to i32, but I didn't want to change everything at once.

### DIFF
--- a/cranelift/codegen/src/isa/aarch64/abi.rs
+++ b/cranelift/codegen/src/isa/aarch64/abi.rs
@@ -320,7 +320,7 @@ impl ABIMachineSpec for AArch64MachineDeps {
                     Some((ty, slot_offset))
                 })
                 .map(|(ty, offset)| ABIArgSlot::Stack {
-                    offset,
+                    offset: i32::try_from(offset).unwrap(),
                     ty,
                     extension: param.extension,
                 })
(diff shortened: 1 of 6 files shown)


cranelift/codegen/src/isa/aarch64/abi.rs
```diff
@@ -320,7 +320,7 @@ impl ABIMachineSpec for AArch64MachineDeps {
                     Some((ty, slot_offset))
                 })
                 .map(|(ty, offset)| ABIArgSlot::Stack {
-                    offset,
+                    offset: i32::try_from(offset).unwrap(),
```
bjorn3, line 323: We don't currently check that the stack frame size is less than 2GB, so this unwrap can panic: https://github.com/bytecodealliance/wasmtime/issues/7916

fitzgen (approved):
LGTM, assuming we unblock it.

PR #8164, opened 2024-03-18
Cranelift: clone ABIArg instead of allocating SmallInstVec

jameysharp:
This allows us to make a single pass over an argument's slots instead of needing to first pre-allocate temporaries, because we don't need to hold a borrow of `ctx`. For the same reason, we can also emit the new instructions directly instead of buffering them and copying them at the end.

This approach also moves less data around. A SmallInstVec is a usize plus four `M::Inst`s, which are 32 or 40 bytes each, while an ABIArg is only 40 bytes. Since the SmallVecs inside ABIArg almost never spill to the heap, cloning one uses less memory than allocating temporary space for a few instructions.

I was inspired to write this after working on #8151 with @elliottt, and working on this also inspired #8163.


elliottt (approved):
:shipit:

PR #8166, opened 2024-03-18
Rename `-S common` to `-S cli`.

sunfishcode:
The "common" in `-S common` came from "wasi-common" which came from the idea of having code in common between Wasmtime, Lucet, and others. It doesn't have a clear meaning for end users, and has a risk of being interpreted as "common" functionality that's generally available everywhere.

This PR renames `-S common` to `-S cli`, and documents it as including the WASI CLI APIs, to clarify its purpose `-S common` is still accepted, with a warning.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -8,9 +8,9 @@ Unreleased.
 
 ### Changed
 
-The `-S common` flag is renamed to `-S cli`, to better reflect that it provides
-the wasi-cli APIs. `-S common` is still accepted for now, and will be deprecated
-in the future.
+* The `-S common` flag is renamed to `-S cli`, to better reflect that it provides
+  the wasi-cli APIs. `-S common` is still accepted for now, and will be deprecated
+  in the future.
 
 --------------------------------------------------------------------------------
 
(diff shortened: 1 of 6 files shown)


RELEASES.md
```diff
@@ -8,6 +8,9 @@ Unreleased.
 
 ### Changed
 
+The `-S common` flag is renamed to `-S cli`, to better reflect that it provides
+the wasi-cli APIs. `-S common` is still accepted, with a warning.
```
alexcrichton, line 12: Mind updating this following the same format as below with bullets and a link to the PR?

src/commands/run.rs
```diff
@@ -607,7 +607,21 @@ impl RunCommand {
...
+        let mut cli = self.run.common.wasi.cli;
+
+        // Accept -Scommon as a deprecated alias for -Scli.
+        if let Some(common) = self.run.common.wasi.common {
+            if cli.is_some() {
+                bail!(
+                    "The -Scommon option should not be use with -Scli as it is a deprecated alias"
+                );
+            } else {
+                eprintln!("warning: The -Scommon flag has been renamed to -Scli");
```
alexcrichton, line 619: Could this be omitted for the time being? I figure it might be good to let the change percolate before warnings start showing up.

alexcrichton (approved):

PR #8170, opened 2024-03-18
Remove wasm-c-api submodule

alexcrichton:
This submodule hasn't been updated in ~3 years at this point and we additionally don't need most of the submodule. Instead add a script to copy the files we need and verify in CI that the files are up-to-date.

This also makes using the C API a bit nicer where you don't have to have two `include` directories with a Wasmtime source tree, just one suffices.


fitzgen (approved):
Nice

PR #8171, opened 2024-03-18
cranelift-wasm: Attach table OOB traps to loads/stores

jameysharp:
Currently, every access to a table element does a bounds-check with a conditional branch to a block that explicitly traps.

Instead, when SPECTRE mitigations are enabled, let's change the address computation to return a null pointer for out-of-bounds accesses, and then allow the subsequent load or store to trap.

This is less code in that case since we can reuse instructions we needed anyway.

In addition, when the table has constant size and the element index is a constant and mid-end optimization is enabled, this allows the bounds-check to be constant folded away. Later, #8139 will let us optimize away the select_spectre_guard instruction in this case too.

Once we also implement #8160, `tests/disas/typed-funcrefs.wat` should be almost as fast as native indirect function calls.

### DIFF
--- a/cranelift/wasm/src/table.rs
+++ b/cranelift/wasm/src/table.rs
@@ -50,7 +50,7 @@ impl TableData {
         mut index: ir::Value,
         addr_ty: ir::Type,
         enable_table_access_spectre_mitigation: bool,
-    ) -> ir::Value {
+    ) -> (ir::Value, ir::MemFlags) {
         let index_ty = pos.func.dfg.value_type(index);
 
         // Start with the bounds check. Trap if `index + 1 > bound`.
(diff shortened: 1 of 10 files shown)


cranelift/wasm/src/table.rs
```diff
@@ -60,16 +60,10 @@ impl TableData {
...
-        pos.ins().trapnz(oob, ir::TrapCode::TableOutOfBounds);
 
-        // If Spectre mitigations are enabled, we will use a comparison to
-        // short-circuit the computed table element address to the start
-        // of the table on the misspeculation path when out-of-bounds.
-        let spectre_oob_cmp = if enable_table_access_spectre_mitigation {
-            Some((index, bound))
-        } else {
-            None
-        };
```
alexcrichton, line 72: AFAIK spectre mitigations are typically a flag due to their performance impact, but in this case because the spectre-related code is straight-line and actually involves less branches, is this actually the more performant path? Would it be worth, for example, unconditionally using `select_spectre_guard`?

alexcrichton (approved):
Nice 👍 

In addition to the question I put below, could you update `prepare_table_addr` to thread out some information "the pointer is known to not be null" or something like that? For example if spectre mitigations are disabled it's known that the load is in-bounds and therefore none of the trap metadata would be required.

Or perhaps `prepare_table_addr` could return a `MemFlags` which has the trap code pre-configured?

cranelift/wasm/src/table.rs
jameysharp: I actually totally confused myself with the same question at first! But I ran it by @cfallin who confirmed the following: A bounds-check branch is almost always predicted correctly, and that means that if we don't care about Spectre, an out-of-order CPU can make more progress by speculating the load and continuing on to dependent operations past that, rather than stalling on the bounds-check computation.

So although I haven't measured it, there should generally be a performance advantage to sticking with the `trapnz` approach, for anyone who ain't afraid of no Spectre.

jameysharp:
Returning the memflags is an interesting option! Let me think about that a bit.

jameysharp:
Yeah, returning `MemFlags` from `prepare_table_addr` tidied things up nicely. Thanks for the suggestion, Alex!

PR #8172, opened 2024-03-18
 Skip type checks on tables that don't need it 

alexcrichton:
This commit implements an optimization to skip type checks in
`call_indirect` for tables that don't require it. With the
function-references proposal it's possible to have tables of a single
type of function as opposed to today's default `funcref` which is a
heterogenous set of functions. In this situation it's possible that a
`call_indirect`'s type tag matches the type tag of a
`table`-of-typed-`funcref`-values, meaning that it's impossible for the
type check to fail.

The type check of a function pointer in `call_indirect` is refactored
here to take the table's type into account. Various things are shuffled
around to ensure that the right traps still show up in the right places
but the important part is that, when possible, the type check is omitted
entirely.

### DIFF
--- a/crates/cranelift/src/func_environ.rs
+++ b/crates/cranelift/src/func_environ.rs
@@ -13,15 +13,15 @@ use cranelift_entity::{EntityRef, PrimaryMap, SecondaryMap};
 use cranelift_frontend::FunctionBuilder;
 use cranelift_frontend::Variable;
 use cranelift_wasm::{
-    FuncIndex, FuncTranslationState, GlobalIndex, GlobalVariable, Heap, HeapData, HeapStyle,
-    MemoryIndex, TableData, TableIndex, TableSize, TargetEnvironment, TypeIndex, WasmHeapType,
-    WasmRefType, WasmResult, WasmValType,
+    EngineOrModuleTypeIndex, FuncIndex, FuncTranslationState, GlobalIndex, GlobalVariable, Heap,
+    HeapData, HeapStyle, MemoryIndex, TableData, TableIndex, TableSize, TargetEnvironment,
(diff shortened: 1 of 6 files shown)


jameysharp (approved):
This makes sense to me! I really like the way you've restructured things with the `CheckIndirectCallTypeSignature` enum.

crates/cranelift/src/func_environ.rs
```diff
@@ -1147,67 +1157,160 @@ impl<'a, 'func, 'module_env> Call<'a, 'func, 'module_env> {
...
+                    let mem_flags = ir::MemFlags::trusted().with_readonly();
+                    self.builder.ins().load(
+                        sig_id_type,
+                        mem_flags.with_trap_code(Some(ir::TrapCode::IndirectCallToNull)),
+                        funcref_ptr,
+                        i32::from(self.env.offsets.ptr.vm_func_ref_type_index()),
+                    );
+                }
+                self.builder.ins().trap(ir::TrapCode::BadSignature);
+                return CheckIndirectCallTypeSignature::StaticTrap;
```
fitzgen, line 1255: FWIW, when the GC proposal is enabled, the static immediate type on the `call_indirect` will be allowed to be a subtype of the actual function's type.

We don't implement this yet, but will need to.

https://webassembly.github.io/gc/core/exec/instructions.html#exec-call-indirect

alexcrichton: Oh good point! I keep consistently forgetting that contra/covariance applies to params/results...

This makes me think we should remove the `PartialEq` implementation for types in wasmtime-types like we did in the embedder API as well

fitzgen: Yeah that wouldn't be a bad idea.

PR #8174, opened 2024-03-18
PCC: 32-bit multiplies produce 32 bits of result.

cfallin:
In fact, we can even use this fact to infer ranges of the results!

(Previously we had a "default" copy-and-paste across all instruction cases that 64-bit reg writes produced 64 bits of undefinedness and we hadn't tightened the spec here.)

Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=67436.

### DIFF
--- a/cranelift/codegen/src/isa/x64/pcc.rs
+++ b/cranelift/codegen/src/isa/x64/pcc.rs
@@ -259,8 +259,8 @@ pub(crate) fn check(
                 }
                 RegMem::Reg { .. } => {}
             }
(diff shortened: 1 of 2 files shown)


cfallin:
(Stacked on top of #8173 and will rebase when that merges.)

jameysharp (approved):
I think I've confirmed that the widths this commit associates with the outputs of these instructions match the widths in the Intel instruction set reference. I'm willing to believe you that passing those widths to `undefined_result` is the right thing to do.

PR #8176, opened 2024-03-18
PCC: support imported memories as well.

cfallin:
Exposed by a fuzzbug (https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=67429); rather than exclude from fuzzing, it seemed easier to just implement. We need  to define a new memory type to describe the memory definition struct pointed to by vmctx, and set up points-to facts appropriately.

### DIFF
--- a/crates/cranelift/src/func_environ.rs
+++ b/crates/cranelift/src/func_environ.rs
@@ -2028,6 +2028,50 @@ impl<'module_environment> cranelift_wasm::FuncEnvironment for FuncEnvironment<'m
             .maximum
             .and_then(|max| max.checked_mul(u64::from(WASM_PAGE_SIZE)));
 
+        let create_vm_def_mt = |pcc_vmctx_memtype: Option<ir::MemoryType>,
+                                func: &mut ir::Function,
+                                from_offset: u32,
+                                gv: ir::GlobalValue| {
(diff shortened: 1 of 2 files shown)


alexcrichton:
I don't know enough here to really review what's going on, but could the pcc-related pieces get factored into their own functions/methods? Right now they're suffering quite a bit from a huge indentation due to being so nested so breaking them out might help with readability

cfallin:
Perhaps; the original intent was to put the PCC logic alongside the CLIF generation because they "mirror" each other (both have largely the same case breakdown, so we'd mirror the control flow elsewhere, which seems more brittle). Happy to consider refactors, unfortunately don't have much time to do a big refactor now. @fitzgen and I cowrote the original; @fitzgen would you be willing to review the fuzzbug fix here?

cfallin:
@alexcrichton On second look, I did a small refactor on some of the logic so hopefully it's at least a little easier to digest now -- let me know what you think. Thanks!

alexcrichton:
Seems reasonable to me! Alas though I fear I know so little about how this is all supposed to work that I still can't really make heads or tails of it without diving much deeper into pcc stuff, so I'll leave @fitzgen to review

fitzgen (approved):

PR #8177, opened 2024-03-18
PCC: x64: 32- and 64-bit XMM loads/stores are 32 and 64 bits, respectively.

cfallin:
These are definitely not 128 bits wide; let's make the PCC model aware of that!

Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=67427.


cfallin:
Stacked on top of #8176, #8174, #8173; only last commit is new.

alexcrichton (approved):
I fear this may be a bit of a long battle recategorizing all the various opcodes as how wide their load/store is, but otherwise looks good 👍

PR #8181, opened 2024-03-19
Add a `CodeBuilder` type to the `wasmtime` crate

alexcrichton:
This commit is extracted from #8055 and adds a new `CodeBuilder` type which is intended to be able to further configure compilation beyond what the constructors of `Module` already provide. For example in #8055 knobs will be added to process `*.dwp` files for more debuginfo processing.

### DIFF
--- a/crates/wasmtime/src/compile/module_builder.rs
+++ b/crates/wasmtime/src/compile/module_builder.rs
@@ -1,0 +1,297 @@
+#[cfg(all(feature = "runtime", feature = "component-model"))]
+use crate::component::Component;
+use crate::Engine;
+#[cfg(feature = "runtime")]
+use crate::{instantiate::MmapVecWrapper, CodeMemory, Module};
+use anyhow::{anyhow, bail, Context, Result};
+use std::borrow::Cow;
+use std::path::Path;
+use std::sync::Arc;
+use wasmtime_environ::ObjectKind;
+#[cfg(feature = "runtime")]
+use wasmtime_runtime::MmapVec;
+
+/// Builder-style structure used to create a [`Module`](crate::Module) or
+/// pre-compile a module to a serialized list of bytes.
+///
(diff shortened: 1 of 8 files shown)


alexcrichton:
I'll note that while I've called this `ModuleBuilder` it actually supports components as well, so there's probably a better name for this.

crates/wasmtime/src/compile/module_builder.rs
```diff
@@ -0,0 +1,297 @@
...
+use wasmtime_environ::ObjectKind;
+#[cfg(feature = "runtime")]
+use wasmtime_runtime::MmapVec;
+
+/// Builder-style structure used to create a [`Module`](crate::Module) or
+/// pre-compile a module to a serialized list of bytes.
+///
+/// This structure can be used for more advanced configuration when compiling a
+/// WebAssembly module. Less advanced configuration can use constructors such
+/// as:
```
fitzgen, line 19:
```suggestion
/// WebAssembly module. Most configuration can use simple constructors such
/// as:
```

fitzgen: (Don't want to create a weird "advanced" vs not thing where people who consider themselves Serious Wasmtime Users will want to use this without actually having a good reason. Might be overthinking things here.)

fitzgen (commented):
LGTM!

As far as naming goes, the only alternatives I can think of are

* `ModuleOrComponentBuilder`
* `CodeBuilder`

Or maybe splitting this type in two in the public API (and sharing an underlying impl) so that we have both a `ModuleBuilder` and a `ComponentBuilder`.

fitzgen (approved):

alexcrichton:
It feels kind of nice being able to configure components/modules in one builder since it's all wasm anyway so I think I'll stick with one for now vs copying APIs, but I'll rename to `CodeBuilder` as I like that name more.

PR #8182, opened 2024-03-19
Canonicalize fpromote/fdemote operations

### LINKED ISSUE #8179: Cranelift: inconsistent NaN signedness behavior for f64.copysign + f32.demote_f64
### Test Case

```rs
// main.rs
use wasmtime::*;

fn main() -> Result<()> {
    let mut config = Config::default();
    config.strategy(Strategy::Cranelift);
    config.cranelift_opt_level(OptLevel::None);
    config.cranelift_nan_canonicalization(true); // doesn't matter: both options show the same result

    let engine = Engine::new(&config)?;
    let wat = r#"
    (module
        (type (;0;) (func (param f64) (result f32)))
        (import "mem" "mem" (memory (;0;) 1))
        (func (;0;) (type 0) (param f64) (result f32)
          local.get 0
          f64.const -0x1 ;; anything with negative sign
          f64.copysign
          f32.demote_f64)
        (export "main" (func 0)))
    "#;
    let module = Module::new(&engine, wat)?;
    let mut store = Store::new(&engine, ());

    let memory_ty = MemoryType::new(1, None);
    let memory = Memory::new(&mut store, memory_ty.clone())?;

    let instance = Instance::new(&mut store, &module, &[memory.into()])?;
    let main = instance.get_func(&mut store, "main")
        .expect("`main` was not an exported function");
    let params = vec![
        Val::F64(0x7ff8000000000000), // +nan (canonicalized)
    ];
    let mut results = vec![Val::F32(0)];
    println!("Opt level None: {:?}", main.call(
        &mut store,
        &params,
        &mut results
    ));
    println!("{:?}", results);

    Ok(())
}
```

```
[package]
name = "wasmtime-wrapper"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
# wasmtime = "18.0.3"
wasmtime = { path = "../wasmtime/crates/wasmtime" } # commit: 3deaa8780bbef21648f2d9eebabcbae909c4a090(current latest, Date: Mon Mar 18 18:03:37 2024 -0700)
```

### Steps to reproduce

Compare the following executions:

```
cargo run --release
cargo run --release --target=aarch64-unknown-linux-gnu
cargo run --release --target=s390x-unknown-linux-gnu
-----
cargo run --release --target=riscv64gc-unknown-linux-gnu
```

QEMU run options I'm currently using are the following:

```
qemu-aarch64 -L /usr/aarch64-linux-gnu -E LD_LIBRARY_PATH=/usr/aarch64-linux-gnu/lib -E WASMTIME_TEST_NO_HOG_MEMORY=1 target/aarch64-unknown-linux-gnu/release/wasmtime-wrapper
qemu-s390x -L /usr/s390x-linux-gnu -E LD_LIBRARY_PATH=/usr/s390x-linux-gnu/lib -E WASMTIME_TEST_NO_HOG_MEMORY=1 target/s390x-unknown-linux-gnu/release/wasmtime-wrapper
qemu-riscv64 -cpu rv64,v=true,vlen=128,vext_spec=v1.0,zba=true,zbb=true,zbs=true,zbc=true,zbkb=true,zcb=true,zicond=true -L /usr/riscv64-linux-gnu -E LD_LIBRARY_PATH=/usr/riscv64-linux-gnu/lib -E WASMTIME_TEST_NO_HOG_MEMORY=1 target/riscv64gc-unknown-linux-gnu/release/wasmtime-wrapper
```

### Expected Results

Results from all architecture should be the same value of +NaN or -NaN (with `cranelift_nan_canonicalization` option)

### Actual Results

Execution results in a **negative** canonicalized NaN value, `0xffc00000` (regardless of `cranelift_nan_canonicalization` option)

```
Opt level None: Ok(())
[F32(4290772992)]
```

However, RISC-V64 shows the result `0x7fc00000`, which is a **positive** canonicalized NaN value:

```
Opt level None: Ok(())
[F32(2143289344)]
```

### Versions and Environment

- wasmtime version
  - commit: 3deaa8780bbef21648f2d9eebabcbae909c4a090 (current latest, Date: Mon Mar 18 18:03:37 2024 -0700)
  - However, also checked on v18.0.3
- Operating system & architecture: Ubuntu 22.04.3 LTS, Arch: x86_64
- QEMU version: all architectures (aarch64, s390x, riscv64) in `version 8.2.1 (v8.2.1)`

### Extra Info

- **As both results are canonicalized versions of NaN (and conforms to spec), I think this might not be a bug - however, this is worth checking**
- Specification of `f64.copysign`: https://webassembly.github.io/spec/core/exec/numerics.html#xref-exec-numerics-op-fcopysign-mathrm-fcopysign-n-z-1-z-2
- `cranelift_nan_canonicalization` option doesn't change the result
- Uploading as public (not a security vuln.) since the bug only exhibits a semantic difference and I don't think this can cause a security impact. Also, this is related to `cranelift_nan_canonicalization` flag, which is `false` by default.

alexcrichton:
This commit changes the strategy implemented in #8146 to canonicalize promotes/demotes of floats to additionally handle #8179.

Closes #8179

### DIFF
--- a/cranelift/codegen/src/nan_canonicalization.rs
+++ b/cranelift/codegen/src/nan_canonicalization.rs
@@ -38,6 +38,8 @@ fn is_fp_arith(pos: &mut FuncCursor, inst: Inst) -> bool {
                 || opcode == Opcode::Nearest
                 || opcode == Opcode::Sqrt
                 || opcode == Opcode::Trunc
+                || opcode == Opcode::Fdemote
+                || opcode == Opcode::Fpromote
         }
         InstructionData::Binary { opcode, .. } => {
             opcode == Opcode::Fadd
@@ -48,24 +50,6 @@ fn is_fp_arith(pos: &mut FuncCursor, inst: Inst) -> bool {
                 || opcode == Opcode::Fsub
         }
         InstructionData::Ternary { opcode, .. } => opcode == Opcode::Fma,
-
-        // bitcasts coming from arbitrary integers need canonicalization as the
-        // integer isn't known to necessarily be a canonical NaN
-        InstructionData::LoadNoOffset {
-            opcode: Opcode::Bitcast,
-            arg,
-            flags: _,
-        } => {
-            let ret = pos.func.dfg.first_result(inst);
-            let ret_type = pos.func.dfg.value_type(ret);
-            let arg_type = pos.func.dfg.value_type(arg);
-            match (arg_type, ret_type) {
-                (types::I32, types::F32) => true,
-                (types::I64, types::F64) => true,
-                _ => false,
-            }
-        }
-
         _ => false,
     }
 }
--- a/tests/misc_testsuite/simd/canonicalize-nan.wast
+++ b/tests/misc_testsuite/simd/canonicalize-nan.wast
@@ -39,6 +39,24 @@
     f64.reinterpret_i64
     f32.demote_f64
     i32.reinterpret_f32)
+
+  (func (export "reinterpret-and-promote") (param i32) (result i64)
+    local.get 0
+    f32.reinterpret_i32
+    f64.promote_f32
+    i64.reinterpret_f64)
+
+  (func (export "copysign-and-demote") (param f64) (result f32)
+    local.get 0
+    f64.const -0x1
+    f64.copysign
+    f32.demote_f64)
+
+  (func (export "copysign-and-promote") (param f32) (result f64)
+    local.get 0
+    f32.const -0x1
+    f32.copysign
+    f64.promote_f32)
 )
 
 (assert_return (invoke "f32x4.floor" (v128.const f32x4 1 -2.2 3.4 nan))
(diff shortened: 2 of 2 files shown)


cranelift/codegen/src/nan_canonicalization.rs
```diff
@@ -38,6 +38,8 @@ fn is_fp_arith(pos: &mut FuncCursor, inst: Inst) -> bool {
                 || opcode == Opcode::Nearest
                 || opcode == Opcode::Sqrt
                 || opcode == Opcode::Trunc
+                || opcode == Opcode::Fdemote
+                || opcode == Opcode::Fpromote
```
afonso360, line 42: Should [`fvpromote_low`](https://docs.rs/cranelift-codegen/latest/cranelift_codegen/ir/trait.InstBuilder.html#method.fvpromote_low) also be included here?

alexcrichton: Good catch! That did indeed previously give different answers for x86 and riscv64 for the test checked in now.

afonso360 (approved):
LGTM! 👍

PR #8190, opened 2024-03-20
Implement opt-in for enabling WASI to block the current thread

alexcrichton:
Currently all of Wasmtime's implementation of WASI is built on Tokio, but some operations are currently not asynchronous such as opening a file or reading a directory. Internally these use `spawn_blocking` to satisfy the requirements of async users of WASI to avoid blocking the current thread. This use of `spawn_blocking`, however, ends up causing mostly just performance overhead in the use case of the CLI, for example, where async wasm is not used. That then leads to this commit, implementing an opt-in mechanism to be able to block the current thread.

A `WasiCtx` now has a flag indicating whether it's ok to block the current thread and that's carried to various filesystem operations that use `spawn_blocking`. The call to `spawn_blocking` is now conditional and skipped if this flag is set.

Additionally the invocation point in the CLI for wasm modules is wrapped in a Tokio runtime to avoid entering/exiting Tokio in the "leaves" when wasm calls the host, as happens today. This hits a better fast path in Tokio that appears to be more efficient.

Semantically this should not result in any change for CLI programs except in one case: file writes. By default writes on `output-stream` in WASI are asynchronous meaning that only one write can be in flight at a time. That being said all current users are immediately blocking waiting for this write to finish anyway, so this commit won't end up changing much. It's already the case that file reads are always blocking, for example. If necessary in the future though this can be further special-cased at the preview1 layer.

### DIFF
--- a/crates/wasi/src/filesystem.rs
+++ b/crates/wasi/src/filesystem.rs
@@ -92,14 +92,22 @@ pub struct File {
     /// doesn't presently provide a cross-platform equivelant of reading the
     /// oflags back out using fcntl.
     pub open_mode: OpenMode,
+
+    allow_blocking_current_thread: bool,
 }
 
 impl File {
-    pub fn new(file: cap_std::fs::File, perms: FilePerms, open_mode: OpenMode) -> Self {
+    pub fn new(
+        file: cap_std::fs::File,
+        perms: FilePerms,
+        open_mode: OpenMode,
+        allow_blocking_current_thread: bool,
+    ) -> Self {
         Self {
             file: Arc::new(file),
             perms,
             open_mode,
+            allow_blocking_current_thread,
         }
     }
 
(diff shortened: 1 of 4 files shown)


crates/wasi/src/filesystem.rs
```diff
@@ -170,30 +186,47 @@ impl Dir {
...
             (r, buf)
-        })
-        .await;
+        };
+        let (r, mut buf) = if self.allow_blocking_current_thread {
+            run(&self.file)
+        } else {
+            let f = Arc::clone(&self.file);
+            spawn_blocking(move || run(&f)).await
+        };
```
fitzgen, line 229: Might be kinda nice to wrap this dance into a helper method so that we don't have to check `self.allow_blocking_current_thread` everywhere and instead call `self.spawn_blocking` or something that internally either does a real `spawn_blocking` or runs the thing immediately on the current thread.

fitzgen (approved):

crates/wasi/src/filesystem.rs
alexcrichton: Sounds good yeah, I did some refactoring to thread this through `File` which has a similar helper already.

PR #8192, opened 2024-03-20
support compile wasmtime with the target wasm32-wasi (Not implemented.)

PiCoPress:
Planning compiling recusively may takes new vision of WebAssembly, but there are many issues when building. 

1. Rust seems not to support WASI on stable version fully.
2. Unable to decide platform - Unix, Windows - of the WASM runtime. WASM cannot be regarded as OS. According to target naiming, then I assumed corresponding platform is WASI.
3. WASM Architecture is hardly different the other existings(x86_64, aarch64, etc.), therefore that just re-using code is no achieves.


cranelift/codegen/src/isa/wasm32/settings.rs
```diff
@@ -0,0 +1,4 @@
+use crate::settings::{detail, Builder, self, Value};
+use core::fmt;
+
+include!(concat!(env!("OUT_DIR"), "/settings-wasm32.rs"));
```
bjorn3, line 4: nit: missing trailing newline

cranelift/codegen/src/isa/wasm32/mod.rs
```diff
@@ -0,0 +1,19 @@
...
+
+#[cfg(feature = "unwind")]
+use crate::isa::unwind::systemv::RegisterMappingError;
+use cranelift_control::ControlPlane;
+use alloc::vec::Vec;
+
+pub(crate) mod inst;
+mod lower;
+mod abi;
+mod settings;
```
bjorn3, line 19: nit: missing trailing newline

bjorn3: And in all other files.

crates/fiber/src/wasm32.rs
bjorn3, line 1: Why these empty files? You can omit the `mod wasm32;`, right?

cranelift/codegen/src/isa/wasm32/lower/isle/generated_code.rs
```diff
@@ -0,0 +1,10 @@
+#![allow(dead_code, unreachable_code, unreachable_patterns)]
+#![allow(unused_imports, unused_variables, non_snake_case, unused_mut)]
+#![allow(
+    irrefutable_let_patterns,
+    unused_assignments,
+    non_camel_case_types,
+    missing_docs
+)] // <-- Why are they allow the bad code?
```
bjorn3, line 8:
```suggestion
)]
```

The isle generated files may contain constructs that trigger these lints as avoiding them would make the emitting code more complex without much benefit.

bjorn3:
Is your intent here to make a wasmtime compiled for wasi able to cross-compile to other architectures without running anything? Or to run code that uses wasmtime on wasi? In the latter case there will need to be a new interface to allow jitted code to actually run, at which point it can probably pass the wasm module given by the user of the wasmtime api could likely directly be passed to the host wasm engine without any translation by wasmtime.

PiCoPress:
Wasmtime built by wasm32(I will call as Meta-WASM) could simplify development processes between implementing real hardware architecture and wasmtime architecture. However I recently found the problem that it makes irony, although Meta-WASM needs the Wasmtime, also Wasmtime may needs the Meta-WASM to execute completely.
And there are some other techincal issues with JIT.

Are you agree to proceed this PR?

PR #8198, opened 2024-03-20
cranelift: Fix return value handling with the winch calling convention

elliottt:
Cranelift's Winch calling convention implementation was incorrect in its handling of return values in three ways:

* Winch returns the last return value in a register instead of the first,
* Winch doesn't align entries returned through the stack to 8-byte boundaries,
* and finally, Winch places the first returned value at the largest offset from the return location, instead of the smallest.

This PR addresses these three issues, enabling progress towards removing the trampoline implementations that are defined in terms of Winch's `MacroAssembler` trait.

Additionally, this PR disables the Winch calling convention for Cranelift's `aarch64` backend, as Winch support for `aarch64` is currently minimal.


alexcrichton (approved):
Nice!

Would it perhaps also make sense to add a runtest exercising this to ensure at runtime the result works in Cranelift at least?

elliottt:
> Would it perhaps also make sense to add a runtest exercising this to ensure at runtime the result works in Cranelift at least?

This seems reasonable to me! I'll investigate adding one.

saulecabrera (approved):

saulecabrera:
Oops, I didn't realize that Alex had already approved.

PR #8203, opened 2024-03-21
Compile out wmemcheck-related libcalls when not enabled

alexcrichton:
Currently even when the `wmemcheck` Cargo feature is disabled the various related libcalls are still compiled into `wasmtime-runtime`. Additionally their signatures are translated when lowering functions, although the signatures are never used. This commit adds `#[cfg]` annotations to compile these all out when they're not enabled.

Applying this change, however, uncovered a subtle bug in our libcalls. Libcalls are numbered in-order as-listed in the macro ignoring `#[cfg]`, but they're assigned a runtime slot in a `VMBuiltinFunctionsArray` structure which does respect `#[cfg]`. This meant, for example, that if `gc` was enabled and `wmemcheck` was disabled, as is the default for our tests, then there was a hole in the numbering where libcall numbers were mismatched at runtime and compile time.

To fix this I've first added a const assertion that the runtime-number of libcalls equals the build-time number of libcalls. I then updated the macro a bit to plumb the `#[cfg]` differently and now libcalls are unconditionally defined regardless of cfgs but the implementation is `std::process::abort()` if it's compiled out.

This ended up having a large-ish impact on the `disas` test suite. Lots of functions have fewer signatures translation because wmemcheck, even when disabled, was translating a few signatures. This also had some assembly changes, too, because I believe functions are considered leaves based on whether they declare a signature or not, so declaring an unused signature was preventing all wasm functions from being considered leaves.

### DIFF
--- a/crates/environ/src/builtin.rs
+++ b/crates/environ/src/builtin.rs
@@ -3,80 +3,88 @@
 macro_rules! foreach_builtin_function {
     ($mac:ident) => {
         $mac! {
-            /// Returns an index for wasm's `memory.grow` builtin function.
+            // Returns an index for wasm's `memory.grow` builtin function.
             memory32_grow(vmctx: vmctx, delta: i64, index: i32) -> pointer;
-            /// Returns an index for wasm's `table.copy` when both tables are locally
-            /// defined.
+            // Returns an index for wasm's `table.copy` when both tables are locally
+            // defined.
             table_copy(vmctx: vmctx, dst_index: i32, src_index: i32, dst: i32, src: i32, len: i32);
-            /// Returns an index for wasm's `table.init`.
+            // Returns an index for wasm's `table.init`.
             table_init(vmctx: vmctx, table: i32, elem: i32, dst: i32, src: i32, len: i32);
-            /// Returns an index for wasm's `elem.drop`.
+            // Returns an index for wasm's `elem.drop`.
             elem_drop(vmctx: vmctx, elem: i32);
-            /// Returns an index for wasm's `memory.copy`
+            // Returns an index for wasm's `memory.copy`
             memory_copy(vmctx: vmctx, dst_index: i32, dst: i64, src_index: i32, src: i64, len: i64);
-            /// Returns an index for wasm's `memory.fill` instruction.
+            // Returns an index for wasm's `memory.fill` instruction.
             memory_fill(vmctx: vmctx, memory: i32, dst: i64, val: i32, len: i64);
-            /// Returns an index for wasm's `memory.init` instruction.
+            // Returns an index for wasm's `memory.init` instruction.
             memory_init(vmctx: vmctx, memory: i32, data: i32, dst: i64, src: i32, len: i32);
-            /// Returns a value for wasm's `ref.func` instruction.
+            // Returns a value for wasm's `ref.func` instruction.
             ref_func(vmctx: vmctx, func: i32) -> pointer;
-            /// Returns an index for wasm's `data.drop` instruction.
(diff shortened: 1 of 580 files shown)


crates/environ/src/builtin.rs
```diff
@@ -3,80 +3,88 @@
...
             #[cfg(feature = "gc")]
             externref_global_set(vmctx: vmctx, global: i32, val: reference);
 
-            /// Returns an index for Wasm's `table.grow` instruction for `externref`s.
+            // Returns an index for Wasm's `table.grow` instruction for `externref`s.
             #[cfg(feature = "gc")]
             table_grow_externref(vmctx: vmctx, table: i32, delta: i32, init: reference) -> i32;
 
-            /// Returns an index for Wasm's `table.fill` instruction for `externref`s.
+            // Returns an index for Wasm's `table.fill` instruction for `externref`s.
```
fitzgen, line 87: It seems nicer to `allow(unused_doc_comments)` in a few places than to lose the doc comments themselves, since they show up not only in rustdoc but also in editors using rust-analyzer/LSP and all that on hover and such.

crates/runtime/src/libcalls.rs
```diff
@@ -121,28 +118,36 @@ pub mod trampolines {
...
                     }));
                     match result {
                         Ok(ret) => LibcallResult::convert(ret),
                         Err(panic) => crate::traphandlers::resume_panic(panic),
                     }
+                    }
+                    $(
+                        #[cfg(not($attr))]
+                        std::process::abort();
+                    )?
```
fitzgen, line 143: Why continue to define the function, but just have it abort? Is it really easier than not defining the function?

crates/runtime/src/vmcontext.rs
```diff
@@ -739,6 +738,14 @@ macro_rules! define_builtin_array {
 
 wasmtime_environ::foreach_builtin_function!(define_builtin_array);
 
+const _: () = {
+    assert!(
+        mem::size_of::<VMBuiltinFunctionsArray>()
+            == mem::size_of::<usize>()
+                * (BuiltinFunctionIndex::builtin_functions_total_number() as usize)
+    )
+};
```
fitzgen, line 747: Nice.

fitzgen (approved):

crates/environ/src/builtin.rs
alexcrichton: Good question! The answer actually relates to your below comment of why to define panicking stubs rather than omitting the functions entirely. I actually first went down the route of entirely omitting the libcalls from both numbering and runtime, but I stopped after implementing that and realized that it could create a confusing situation:

* One machine, with wasmtime features enabled, compiles a wasm module.
* Another machine, with minimal features enabled, loads this module. 

These two machines have different libcall numbering schemes without knowing it. I ended up realizing this would be a pretty confusing issue to run into, so I wanted to instead have all the libcalls always have the same number whether they're actually needed or not at runtime.

That then gave rise to needing to have something deal with `#[cfg]` and `#[cfg(not)]`. This is below in the trampoline where it uses `#[cfg(not(..))]` and that changes the pattern of the macro because it no longer matches all attributes but only `#[cfg]` attributes on each of these items.

More-or-less I couldn't find a nice and easy way to make a macro pattern that would separate doc comments and cfg attributes, so I figured the loss of some minor rustdoc wasn't the end of the world and I demoted these to normal comments instead of doc comments.

fitzgen: Makes sense, reasonable tradeoff. Thanks for explaining!

PR #8205, opened 2024-03-21
Bump MSRV to 1.75.0

alexcrichton:
Coupled with today's release of 1.77.0. Today's release actually has some nice functions and such I think we'll want to use in Wasmtime but we'll need to wait 3 months to be able to use them.

### DIFF
--- a/crates/wasi-nn/src/backend/onnxruntime.rs
+++ b/crates/wasi-nn/src/backend/onnxruntime.rs
@@ -48,7 +48,7 @@ impl BackendFromDir for OnnxBackend {
     }
 }
 
-struct ONNXGraph(Arc<Mutex<Session>>, ExecutionTarget);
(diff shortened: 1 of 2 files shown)


fitzgen (approved):

crates/wasi-nn/src/backend/onnxruntime.rs
```diff
@@ -48,7 +48,7 @@ impl BackendFromDir for OnnxBackend {
     }
 }
 
-struct ONNXGraph(Arc<Mutex<Session>>, ExecutionTarget);
+struct ONNXGraph(Arc<Mutex<Session>>, #[allow(dead_code)] ExecutionTarget);
```
alexcrichton, line 51: cc @devigned I had to add this as it's a new warning in rustc, and I'm not sure if this indicates a bug where this should have been used or if it should be removed.

devigned: Thank you for the heads up. I'll take a look.

PR #8214, opened 2024-03-21
Cranelift: resolve value aliases when printing CLIF functions

fitzgen:
This makes the CLIF much easier to follow.


cfallin (approved):
A long-overdue ease of use fix -- thanks a bunch!

PR #8220, opened 2024-03-22
cranelift: Constant propagate `bswap`

encounter:
This adds constant propagation for 16-, 32- and 64-bit `bswap`s. I did not add `u128`, since it seems like ISLE doesn't have any existing rules for dealing with `uimm128`. (Though they could be added, I assume)

Please let me know if there's a better approach for the added rules!

### DIFF
--- a/cranelift/codegen/src/opts/cprop.isle
+++ b/cranelift/codegen/src/opts/cprop.isle
@@ -264,3 +264,18 @@
 (extern constructor f32_from_sint f32_from_sint)
 (decl f64_from_sint (i64) Ieee64)
 (extern constructor f64_from_sint f64_from_sint)
+
+;; Constant fold bswap of a constant.
+(rule (simplify (bswap $I16 (iconst ty (u64_from_imm64 n))))
+      (subsume (iconst $I16 (imm64 (u64_bswap16 n)))))
+(rule (simplify (bswap $I32 (iconst ty (u64_from_imm64 n))))
+      (subsume (iconst $I32 (imm64 (u64_bswap32 n)))))
+(rule (simplify (bswap $I64 (iconst ty (u64_from_imm64 n))))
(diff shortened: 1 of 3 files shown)


cranelift/codegen/src/opts/cprop.isle
```diff
@@ -264,3 +264,18 @@
...
 (decl f64_from_sint (i64) Ieee64)
 (extern constructor f64_from_sint f64_from_sint)
+
+;; Constant fold bswap of a constant.
+(rule (simplify (bswap (ty_16 ty) (iconst ty (u64_from_imm64 n))))
+      (subsume (iconst ty (imm64 (u64_bswap16 n)))))
+(rule (simplify (bswap (ty_32 ty) (iconst ty (u64_from_imm64 n))))
+      (subsume (iconst ty (imm64 (u64_bswap32 n)))))
+(rule (simplify (bswap (ty_64 ty) (iconst ty (u64_from_imm64 n))))
+      (subsume (iconst ty (imm64 (u64_bswap64 n)))))
```
alexcrichton, line 274: Instead of the `ty_NN` matchers I think you can match `$I16`, `$I32`, and `$I64` here directly which would also avoid the need for a new `ty_16` (not that it's wrong, just perhaps a nice small cleanup)

alexcrichton (approved):
Thanks for this!

alexcrichton:
For 128-bit byteswaps that would probably show up through `(bswap (iconcat (iconst ...) (iconst ...)))` but it's also ok to not handle that in this PR too

cranelift/codegen/src/opts/cprop.isle
encounter: Aha, thanks, I knew I was missing something!

alexcrichton (approved):

Issue #7237, opened 2023-10-13
Partially out-of-bounds writes on ARM


fitzgen:
Apparently on ARM when

* a store is unaligned
* and crosses a page boundary
* and one page faults but the other doesn't

then there is no guarantee that the part of the store to the page that didn't fault will not succeed, despite the other part raising a signal.

This means that partially out-of-bounds Wasm stores that trigger a trap can potentially mutate memory for the in-bounds portion of the write, which is not spec compliant.

Apparently it is implementation-defined behavior, so it may or may not be an issue on any given ARM machine.

Thus far, @cfallin tested on the ARM machines he has access to and none of the following have failed the attached test case:

* Apple M2
* RPi4
* the BA ARM server ("Neoverse N1 core, I think, in an Ampere Altra CPU")

### Test Case

```wat
(module
  (memory 1)
  (func (export "i64.store") (param i32 i64)
    local.get 0
    local.get 1
    i64.store)
  (func (export "i32.load8_u") (param i32) (result i32)
    local.get 0
    i32.load8_u))

(assert_trap (invoke "i64.store"
                     (i32.const 65529)
                     (i64.const 0xffffffffffffffff))
             "out of bounds memory access")

;; Partial bytes were not written.
(assert_return (invoke "i32.load8_u" (i32.const 65529))
               (i32.const 0))
(assert_return (invoke "i32.load8_u" (i32.const 65530))
               (i32.const 0))
(assert_return (invoke "i32.load8_u" (i32.const 65531))
               (i32.const 0))
(assert_return (invoke "i32.load8_u" (i32.const 65532))
               (i32.const 0))
(assert_return (invoke "i32.load8_u" (i32.const 65533))
               (i32.const 0))
(assert_return (invoke "i32.load8_u" (i32.const 65534))
               (i32.const 0))
(assert_return (invoke "i32.load8_u" (i32.const 65535))
               (i32.const 0))
```

### See Also

* https://github.com/WebAssembly/design/issues/1490
* https://bugzilla.mozilla.org/show_bug.cgi?id=1666747

fitzgen:
Modulo relaxing the spec, the only 100% sure way to fix this that I am aware of would be to use explicit bounds checks for all stores on ARM.

fitzgen:
Oh also, we should double check whether risc-v and s390x give us the stronger guarantees we need to match wasm semantics.

cc @uweigand: happen to know the answer for s390x?

cfallin:
[...]

PR #8221, opened 2024-03-22
Cranelift: implement "precise store traps" in presence of store-tearing hardware.

cfallin:
As discussed at in #7237 and WebAssembly/design#1490, some instruction-set architectures do not guarantee that a store that "partially traps" (overlaps multiple pages, only one of which disallows the store) does not also have some side-effects. In particular, the part of the store that *is* legal might succeed.

This has fascinating implications for virtual memory-based WebAssembly heap implementations: when a store is partially out-of-bounds, it should trap (there is no "partially" here: if the last byte is out of bounds, the store is out of bounds). A trapping store should not alter the final memory state, which is observable by the outside world after the trap. Yet, the simple lowering of a Wasm store to a machine store instruction could violate this expectation, in the presence of "store tearing" as described above.

Neither ARMv8 (aarch64) nor RISC-V guarantee lack of store-tearing, and we have observed it in tests on RISC-V.

This PR implements the idea first proposed [here], namely to prepend a load of the same size to every store. The idea is that if the store will trap, the load will as well; and precise exception handling, a well-established principle in all modern ISAs, guarantees that instructions beyond a trapping instruction have no effect.

This is off by default, and is mainly meant as an option to study the impact of this idea and to allow for precise trap execution semantics on affected machines unless/until the spec is clarified.

On an Apple M2 Pro machine (aarch64), this was measured to have a 2% performance impact when running `spidermonkey.wasm` with a simple recursive Fibonacci program. It can be used via the `-Ccranelift-ensure_precise_store_traps=true` flag to Wasmtime.

[here]:
https://github.com/WebAssembly/design/issues/1490#issuecomment-1762043224

### DIFF
--- a/cranelift/codegen/src/precise_store_traps.rs
+++ b/cranelift/codegen/src/precise_store_traps.rs
@@ -1,0 +1,109 @@
+//! Precise-store-traps pass.
+//!
+//! On some instruction-set architectures, a store that crosses a page
+//! boundary such that one of the pages would fault on a write can
+//! sometimes still perform part of its memory update on the other
+//! page. This becomes relevant, and problematic, when page
+//! protections are load-bearing for Wasm VM semantics: see [this
+//! issue] where a partially-out-of-bounds store in Wasm is currently
+//! defined to perform no side-effect, but with a common lowering on
+//! several ISAs and on some microarchitectures does actually perform
+//! a "torn write".
+//!
+//! [this issue]: https://github.com/WebAssembly/design/issues/1490
+//!
+//! This pass performs a transform on CLIF that should avoid "torn
+//! partially-faulting stores" by performing a throwaway *load* before
+//! every store, of the same size and to the same address. This
+//! throwaway load will fault if the store would have faulted due to
+//! not-present pages (this still does nothing for
+//! readonly-page-faults). Because the load happens before the store
+//! in program order, if it faults, any ISA that guarantees precise
+//! exceptions (all ISAs that we support) will ensure that the store
+//! has no side-effects. (Microarchitecturally, once the faulting
+//! instruction retires, the later not-yet-retired entries in the
+//! store buffer will be flushed.)
+//!
+//! This is not on by default and remains an "experimental" option
+//! while the Wasm spec resolves this issue, and serves for now to
+//! allow collecting data on overheads and experimenting on affected
+//! machines.
+
+use crate::cursor::{Cursor, FuncCursor};
+use crate::ir::types::*;
+use crate::ir::*;
+
+fn covering_type_for_value(func: &Function, value: Value) -> Type {
+    match func.dfg.value_type(value).bits() {
+        8 => I8,
+        16 => I16,
+        32 => I32,
+        64 => I64,
+        128 => I8X16,
+        _ => unreachable!(),
+    }
+}
+
+/// Perform the precise-store-traps transform on a function body.
+pub fn do_precise_store_traps(func: &mut Function) {
+    let mut pos = FuncCursor::new(func);
+    while let Some(_block) = pos.next_block() {
+        while let Some(inst) = pos.next_inst() {
+            match &pos.func.dfg.insts[inst] {
+                &InstructionData::StackStore {
+                    opcode: _,
+                    arg: data,
+                    stack_slot,
+                    offset,
+                } => {
+                    let ty = covering_type_for_value(&pos.func, data);
+                    let _ = pos.ins().stack_load(ty, stack_slot, offset);
+                }
+                &InstructionData::DynamicStackStore {
+                    opcode: _,
+                    arg: data,
+                    dynamic_stack_slot,
+                } => {
+                    let ty = covering_type_for_value(&pos.func, data);
+                    let _ = pos.ins().dynamic_stack_load(ty, dynamic_stack_slot);
+                }
+                &InstructionData::Store {
+                    opcode,
+                    args,
+                    flags,
+                    offset,
+                } => {
+                    let (data, addr) = (args[0], args[1]);
+                    let ty = match opcode {
+                        Opcode::Store => covering_type_for_value(&pos.func, data),
+                        Opcode::Istore8 => I8,
+                        Opcode::Istore16 => I16,
+                        Opcode::Istore32 => I32,
+                        _ => unreachable!(),
+                    };
+                    let _ = pos.ins().load(ty, flags, addr, offset);
+                }
+                &InstructionData::StoreNoOffset {
+                    opcode: Opcode::AtomicStore,
+                    args,
+                    flags,
+                } => {
+                    let (data, addr) = (args[0], args[1]);
+                    let ty = covering_type_for_value(&pos.func, data);
+                    let _ = pos.ins().atomic_load(ty, flags, addr);
+                }
+                &InstructionData::AtomicCas { .. } | &InstructionData::AtomicRmw { .. } => {
+                    // Nothing: already does a read before the write.
+                }
+                &InstructionData::NullAry {
+                    opcode: Opcode::Debugtrap,
+                } => {
+                    // Marked as `can_store`, but no concerns here.
+                }
+                inst => {
+                    assert!(!inst.opcode().can_store());
+                }
+            }
+        }
+    }
+}
--- a/cranelift/codegen/meta/src/shared/settings.rs
+++ b/cranelift/codegen/meta/src/shared/settings.rs
@@ -119,6 +119,24 @@ pub(crate) fn define() -> SettingGroup {
     );
 
     settings.add_bool(
+        "ensure_precise_store_traps",
+        "Enable transform to ensure precise store traps in the presence of store-tearing hardware.",
+        r#"
+            This inserts a load before every store so that hardware that "tears" stores when
+            part of the stored address range would fault due to a not-present page does not do so.
+            This can be useful to ensure precise post-trap memory state: for example, a Wasm guest
+            that traps on an unaligned partially-out-of-bounds store might otherwise produce
+            incorrect final memory state on hardware that performs the in-bounds part of the store
+            before trapping.
+
+            See also:
+            * https://github.com/WebAssembly/design/issues/1490
+            * https://github.com/bytecodealliance/wasmtime/issues/7237
+        "#,
+        false,
+    );
+
+    settings.add_bool(
         "enable_pinned_reg",
         "Enable the use of the pinned register.",
         r#"
--- a/cranelift/codegen/src/context.rs
+++ b/cranelift/codegen/src/context.rs
@@ -20,6 +20,7 @@ use crate::legalizer::simple_legalize;
 use crate::loop_analysis::LoopAnalysis;
 use crate::machinst::{CompiledCode, CompiledCodeStencil};
 use crate::nan_canonicalization::do_nan_canonicalization;
+use crate::precise_store_traps::do_precise_store_traps;
 use crate::remove_constant_phis::do_remove_constant_phis;
 use crate::result::{CodegenResult, CompileResult};
 use crate::settings::{FlagsOrIsa, OptLevel};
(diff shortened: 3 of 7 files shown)


cfallin:
@afonso360, if you still have access to your RISC-V hardware and a bit of time to test, it would be useful to confirm that this changes the result of the "torn stores" test case from earlier! @candymate, it would be interesting to confirm that this resolves your fuzzbug as well, and this could be useful for further fuzzing for you.

cranelift/codegen/src/precise_store_traps.rs
```diff
@@ -0,0 +1,109 @@
...
+        while let Some(inst) = pos.next_inst() {
+            match &pos.func.dfg.insts[inst] {
+                &InstructionData::StackStore {
+                    opcode: _,
+                    arg: data,
+                    stack_slot,
+                    offset,
+                } => {
+                    let ty = covering_type_for_value(&pos.func, data);
+                    let _ = pos.ins().stack_load(ty, stack_slot, offset);
```
bjorn3, line 60: Stack stores are guaranteed to succeed when stack probing (or some other kind of stack size check) is used.

bjorn3:
> this was measured to have a 2% performance impact

How does this only have a 2% perf impact? Is some optimization pass removing part of the loads or is it the M2 chip being able to exploit a lot of instruction level parallelism?

cfallin:
Probably the latter -- the loads are not removed, and make it all the way to machine code.

cfallin:
One other performance-related note: this will possibly have worse impact when the pooling allocator with copy-on-write is enabled in Wasmtime, as the "first touch" to a page matters: if demanded in by a load first, a read-only mapping is created, which might immediately fault again on the store (doing the actual CoW). We've seen such "mapping change" operations become bottlenecks in multithreaded processes due to IPIs. I don't expect we would turn this option on in any production setting though; rather we'd ensure (as is already the case on x86) any production hardware does not have the torn-write behavior. But that's something to consider if we ever discuss having this on by default in all cases.

afonso360:
I can confirm that this fixes the torn writes at least on QEMU. 🎉 

Interestingly QEMU RISC-V only shows this torn write test when using the vector extensions to perform the write (`v128`), but not using the normal `i64.store`.

I'm going to test this on real hardware over the weekend, and also see if I can run some benchmarks.

For completeness here's the WAST test that I used.
<details><summary>WAST Test</summary>

```
(module
  (memory 1)
  (func (export "i128.store") (param i32 i64)
    local.get 0
    v128.const i32x4 0xffffffff 0xffffffff 0xffffffff 0xffffffff
    v128.store align=1)
  (func (export "i32.load8_u") (param i32) (result i32)
    local.get 0
    i32.load8_u))

(assert_trap (invoke "i128.store"
                     (i32.const 65529)
                     (i64.const 0xffffffffffffffff))
             "out of bounds memory access")

;; Partial bytes were not written.
(assert_return (invoke "i32.load8_u" (i32.const 65529))
               (i32.const 0))
(assert_return (invoke "i32.load8_u" (i32.const 65530))
               (i32.const 0))
(assert_return (invoke "i32.load8_u" (i32.const 65531))
               (i32.const 0))
(assert_return (invoke "i32.load8_u" (i32.const 65532))
               (i32.const 0))
(assert_return (invoke "i32.load8_u" (i32.const 65533))
               (i32.const 0))
(assert_return (invoke "i32.load8_u" (i32.const 65534))
               (i32.const 0))
(assert_return (invoke "i32.load8_u" (i32.const 65535))
               (i32.const 0))

```
</details>

cranelift/codegen/src/precise_store_traps.rs
```diff
@@ -0,0 +1,109 @@
...
+                        Opcode::Store => covering_type_for_value(&pos.func, data),
+                        Opcode::Istore8 => I8,
+                        Opcode::Istore16 => I16,
+                        Opcode::Istore32 => I32,
+                        _ => unreachable!(),
+                    };
+                    let _ = pos.ins().load(ty, flags, addr, offset);
+                }
+                &InstructionData::StoreNoOffset {
+                    opcode: Opcode::AtomicStore,
```
alexcrichton, line 87: I think atomics can be exempted here because they are already required to be aligned? At least at the wasm layer that's validated, I'm not sure about the Cranelift layer. This isn't needed for the wasm-level problem though.

cranelift/codegen/src/precise_store_traps.rs
```diff
@@ -0,0 +1,109 @@
...
+                } => {
+                    let (data, addr) = (args[0], args[1]);
+                    let ty = match opcode {
+                        Opcode::Store => covering_type_for_value(&pos.func, data),
+                        Opcode::Istore8 => I8,
+                        Opcode::Istore16 => I16,
+                        Opcode::Istore32 => I32,
+                        _ => unreachable!(),
+                    };
+                    let _ = pos.ins().load(ty, flags, addr, offset);
```
alexcrichton, line 84: Can this be skipped if `flags.trap_code().is_none()` or if `flags.aligned()`? Only possibly-trapping non-aligned loads (which is all wasm loads, but not anything to a vmctx for example)

tschneidereit:
I fully assume that I'm missing something and the answer is no, but just in case: would it not also work to first attempt to do the part of the store operation touching the highest page, and then do the rest of the write? ISTM that if any writes would fail that'd be guaranteed to include the highest page, and if that one succeeds, a write to all lower pages for the rest of the store operation is certain to succeed as well?

cfallin:
The problem I see with that approach is that we only know that dynamically: consider if we have a 64-bit store (allowed to be unaligned) to address `X`, we would have to decompose it into bytes and perform them in reverse order (store `X+7`, then `X+6`, then ...) always, because we know nothing about `X`'s alignment. A static analysis could sometimes know `X`'s alignment, but usually not, e.g. if it is a value loaded from memory itself. So that is a store-amplification factor of 8 for 64-bit ops, 4 for 32-bit ops, etc, likely to have quite a big perf impact, I think.

tschneidereit:
Hmm, I see, I think. Could we then instead write a single byte (of whatever value) at the highest address, and do the actual store if that succeeded?

cfallin:
I think that would work, yeah. There are some other interesting performance tradeoffs here -- two that come to mind are "partial-store forwarding" (partially overlapping loads/stores can be problematic for uarch memory perf) and the fact that stores in general are more expensive than loads (store buffers typically fewer entries than load buffers so can stall the OoO window, stores take extra effort at retire) that, balanced against all the other considerations here (we want something we could turn on by default on tier-2 architecture(s) if needed, and aren't as concerned with interactions with IPIs in pooling allocator where we're more likely to be very careful to select the right configuration and avoid this mechanism), I think I'd still prefer the load+store over narrow-store+store.

tschneidereit:
That all makes sense to me. I was mainly concerned about the impact on the pooling allocator, and the missing piece was the ability to select a configuration that'd allow us to avoid any of this.

cfallin:
Ah yes, indeed. This is only a problem on aarch64 and riscv64, and AFAICT all aarch64 hardware one would *want* to run a high-performance server on (Apple implementations, AWS machines, Ampere CPUs, even RPi4) do not exhibit store-tearing, so on aarch64 it's mostly theoretical and/or on very small (low-power / embedded) cores I think.

afonso360:
I can confirm that this fix also works in real hardware. I ran a few benchmarks at random, and it does have a bit of impact on smaller cores (as expected). Most benchmarks are 5-10% slower, but there is one outlier at 40%

The results below are on a Dual-Issue In Order Core (Sifive u74).

<details><summary>Results</summary>

```
execution :: cycles :: benchmarks/spidermonkey/benchmark.wasm

  Δ = 1459784.00 ± 60785.14 (confidence = 99%)

  noload.so is 1.07x to 1.07x faster than load.so!

  [22663159 22821360.40 22862178] load.so
  [21326450 21361576.40 21405304] noload.so

compilation :: cycles :: benchmarks/spidermonkey/benchmark.wasm

  Δ = 2560349.00 ± 1538031.16 (confidence = 99%)

  noload.so is 1.01x to 1.04x faster than load.so!

  [110537308 111942491.90 113530980] load.so
  [108818829 109382142.90 111609052] noload.so

instantiation :: cycles :: benchmarks/spidermonkey/benchmark.wasm

  No difference in performance.

  [3465 3954.40 6169] load.so
  [3444 4122.00 6204] noload.so

  Δ = 92094.60 ± 29972.88 (confidence = 99%)

  noload.so is 1.06x to 1.12x faster than load.so!

  [1100973 1120327.60 1169522] load.so
  [1019878 1028233.00 1083605] noload.so

compilation :: cycles :: benchmarks/bz2/benchmark.wasm

  No difference in performance.

  [2281318 2391173.30 2505976] load.so
  [2212191 2312207.60 2376594] noload.so

instantiation :: cycles :: benchmarks/bz2/benchmark.wasm

  No difference in performance.

  [1028 1098.00 1448] load.so
  [1050 1103.10 1170] noload.so

  execution :: cycles :: benchmarks/pulldown-cmark/benchmark.wasm

  Δ = 66417.80 ± 543.64 (confidence = 99%)

  noload.so is 1.39x to 1.40x faster than load.so!

  [233706 234547.90 235270] load.so
  [167648 168130.10 168742] noload.so

  No difference in performance.

  [4541616 4965704.40 5992445] load.so
  [4423989 4751356.50 5875859] noload.so

instantiation :: cycles :: benchmarks/pulldown-cmark/benchmark.wasm

  No difference in performance.

  [1714 2113.50 2415] load.so
  [1798 2135.50 2299] noload.so

  
compilation :: cycles :: benchmarks/blake3-scalar/benchmark.wasm

  Δ = 44240.30 ± 40968.63 (confidence = 99%)

  noload.so is 1.00x to 1.03x faster than load.so!

  [2848804 2880309.90 2923861] load.so
  [2803879 2836069.60 2912357] noload.so

instantiation :: cycles :: benchmarks/blake3-scalar/benchmark.wasm

  No difference in performance.

  [1186 1391.10 1495] load.so
  [1185 1445.70 1604] noload.so

execution :: cycles :: benchmarks/blake3-scalar/benchmark.wasm

  No difference in performance.

  [9675 9782.10 10005] load.so
  [9572 9637.80 9814] noload.so

  instantiation :: cycles :: benchmarks/regex/benchmark.wasm

  Δ = 340.10 ± 243.11 (confidence = 99%)

  noload.so is 1.03x to 1.19x faster than load.so!

  [2939 3406.10 3681] load.so
  [2992 3066.00 3281] noload.so

execution :: cycles :: benchmarks/regex/benchmark.wasm

  Δ = 245010.10 ± 1279.07 (confidence = 99%)

  noload.so is 1.06x to 1.06x faster than load.so!

  [4076050 4077873.90 4079406] load.so
  [3831298 3832863.80 3834736] noload.so

compilation :: cycles :: benchmarks/regex/benchmark.wasm

  Δ = 369482.20 ± 93448.03 (confidence = 99%)

  noload.so is 1.03x to 1.04x faster than load.so!

  [10706461 10872794.20 10932079] load.so
  [10360994 10503312.00 10590922] noload.so
```
</details>

candymate:
> @afonso360, if you still have access to your RISC-V hardware and a bit of time to test, it would be useful to confirm that this changes the result of the "torn stores" test case from earlier! @candymate, it would be interesting to confirm that this resolves your fuzzbug as well, and this could be useful for further fuzzing for you.

I'm sorry that I saw this just now - I'll check this in a day or two from now and let you know.

fitzgen:
@cfallin are you still planning on pushing this forward? Not sure what the status is here. But if not, then we should probably close the PR.

cfallin:
@fitzgen I put it together as a proof-of-concept for the approach, but it's stalled on the upstream standards discussion. I don't really have the energy to drive a conversation there at the moment, but the continued relevance of this code here hangs on that.

candymate:
Hi, sorry for the really late response here (totally forgot about this since I was really busy these days)

If I remember correctly, this resolved the issue and led to no further discovery of more similar bugs. (Ran the fuzzer around mid of May, 2024)

Thank you.

PR #8224, opened 2024-03-22
Don't take a custom FuncWriter in pretty_verifier_error

jameysharp:
This flexibility is not used anywhere, so let's specialize it.

As far as I can tell, it's never been used since being introduced in https://github.com/bytecodealliance/cranelift/pull/379 in 2018.

@bjorn3, is this used in cg-clif?


fitzgen (approved):

bjorn3:
cg_clif uses it: https://github.com/rust-lang/rustc_codegen_cranelift/blob/593228373d175650bb1033814f515df2e15a489d/src/base.rs#L242

jameysharp:
Good to know! Then I won't merge this. Thanks for checking!

PR #8226, opened 2024-03-22
cranelift: Split return instruction emission out of gen_epilogue_frame_restore

elliottt:
The current approach we're discussing for handling callee-save registers with the tail calling convention involves restoring the frame using the existing backend logic before emitting the jump for the tail call. To prepare for this, this PR splits the `ret` emission out of `gen_epilogue_frame_restore` so that it can be used purely to setup the frame for a return **or** jump.


cfallin (approved):
Straightforward refactor, LGTM!

Issue #8188, opened 2024-03-20
Async flavor of `wasmtime_wasi_http::proxy::add_to_linker` differs from sync flavor


rylev:
The `wasi:http/proxy` world contains a subset of `wasi` imports found in the `wasi:cli/command` world. `wasmtime_wasi_http::proxy::add_to_linker` seems to do the right thing and only link the `wasi` interfaces that the world says it imports. However, `wasmtime_wasi_http::proxy::sync::add_to_linker` simply [wholesale imports the entire command world](https://github.com/bytecodealliance/wasmtime/blob/03dc78b192f1d645bff55041b801e632dd5a9344/crates/wasi-http/src/proxy.rs#L61). 

There is a [TODO](https://github.com/bytecodealliance/wasmtime/blob/03dc78b192f1d645bff55041b801e632dd5a9344/crates/wasi-http/src/proxy.rs#L59-L60) stating that this is due to the adapter bringing in all of these imports along with a hidden function that only brings in the required imports. I'm unsure if this is all still needed for if this can be corrected. However, at the very least, this can be very confusing for users, and we should at least have some sort of documentation pointing out that this will happen.

Issue #8187, opened 2024-03-20
[wasi] Confusing that async linker support is the default


rylev:
When adding `wasi` or `wasi-http` support to a linker through the various `add_to_linker` functions, I was unaware that there was a `sync` module that offered the same functionality without requiring async instantiation. It is not obvious to me why the async flavor would be the default especially for `wasi` (one could argue that for `wasi-http` it does make sense for `async` to be the default). While it's more typing perhaps putting the `async` variant in its own module and requiring the user to explicitly select `async` or `sync` would make the API more intuitive.

PR #6770, opened 2023-07-25
Change preview2 builder methods to use `&mut self`

alexcrichton:
This commit changes the `WasiCtxBuilder` for preview2 to use a builder pattern more similar to `std::process::Command` where methods take `&mut self` and return `&mut Self` instead of taking `self` and returning `Self`. This pattern enables more easily building up configuration over time throughout code where ownership transfer might otherwise be awkward.

A small caveat to this is that the ergonomics of this pattern only really work out well if the final "build" method takes `&mut self` as well. In this situation it's difficult to try to figure out what's supposed to happen if this method is called twice, so I left it to panic for now so we can more easily update it in the future possibly.

I'll also note that the original motivations for this come from the upgrade at https://github.com/fermyon/spin/pull/1678

PR #8228, opened 2024-03-24
Add documentation and refactor `wasmtime-wasi`

alexcrichton:
Currently the `wasmtime-wasi` crate is pretty sparse on documentation and additionally had some papercuts in the API. I've attempted to document all the major bits with words and examples here in addition to performing some quality-of-life refactors for API usage in the future. I'd like to refactor the WASIp2 bits a bit more to make them a bit more similar to the WASIp1 bits too, but that's a larger refactoring for a future PR.

Closes https://github.com/bytecodealliance/wasmtime/issues/8187 (they're now at the same level)
Closes https://github.com/bytecodealliance/wasmtime/issues/8188 (I've synchronized interfaces added)

I'll note that `wasmtime-wasi-http` could still use some love as well, which I haven't touched in this PR.


crates/wasi/src/ctx.rs
```diff
@@ -156,37 +232,105 @@ impl WasiCtxBuilder {
...
+    /// // Make `./readonly` available in the guest as `./ro`
+    /// wasi.preopened_dir("./readonly", DirPerms::READ, FilePerms::READ, "./ro");
+    /// # Ok(())
+    /// # }
+    /// ```
     pub fn preopened_dir(
         &mut self,
-        dir: cap_std::fs::Dir,
+        host_path: impl AsRef<Path>,
         perms: DirPerms,
```
rylev, line 329: This always bothered me that this is called `perms` instead of `dir_perms` to contrast with `file_perms`. What do we think about renaming that arg?

crates/wasi/src/ctx.rs
```diff
@@ -156,37 +232,105 @@ impl WasiCtxBuilder {
...
         file_perms: FilePerms,
-        path: impl AsRef<str>,
-    ) -> &mut Self {
+        guest_path: impl AsRef<str>,
```
rylev, line 331: Should `guest_path` be second so that file paths are first and permissions second?

crates/wasi/src/ctx.rs
```diff
@@ -325,18 +499,101 @@ impl WasiCtxBuilder {
...
+/// This structure is created through [`WasiCtxBuilder`] and is stored within
+/// the `T` of [`Store<T>`][`Store`]. Access to the structure is provided
+/// through the [`WasiView`] trait as an implementation on `T`.
+///
+/// Note that this structure itself does not have any accessors, it's here for
+/// internal use within the `wasmtime-wasi` crate's implementation of
+/// bindgen-generated traits.
+///
+/// [`Store`]: wasmtime::Store
 pub struct WasiCtx {
```
rylev, line 597: Not related to this PR exactly, but one pattern that can be very useful when a builder is required to build a type is to a `builder` function to that type. So in this case:

```rust
WasiCtx::builder();
```

fitzgen:
Redirecting review to @sunfishcode since I haven't ever really touched this crate.

alexcrichton:
I'll also tag @elliottt since he's worked on these crates as well

crates/wasi/src/bindings.rs
```diff
@@ -31,99 +32,252 @@ pub mod sync_io {
...
     }
-    pub use self::_internal::wasi::{filesystem, io};
+    pub use self::generated::exports;
+    pub use self::generated::wasi::{filesystem, io};
+
+    /// Synchronous bindings to execute and run a `wasi:cli/command`.
+    ///
+    /// This structure is automatically generated by `bindgen!` and is intended
+    /// to be used with [`Config::async_support(false)`][async]. For the
+    /// asynchronous version see [`binidngs::Command`](super::Command).
```
rylev, line 45:
```suggestion
    /// asynchronous version see [`bindings::Command`](super::Command).
```

crates/wasi/src/bindings.rs
```diff
@@ -2,26 +2,27 @@
 //
 // Note that this is only done for interfaces which can block, or those which
 // have some functions in `only_imports` below for being async.
-pub mod sync_io {
-    pub(crate) mod _internal {
+pub mod sync {
```
rylev, line 5: I'm unsure about `sync` vs. `async_io`. It seems like the previous `sync_io` contrasts more clearly with `async_io`

src/commands/run.rs
```diff
@@ -798,8 +789,10 @@ impl RunCommand {
             num_fd += 1;
         }
 
-        for (name, dir) in self.compute_preopen_dirs()? {
-            builder.preopened_dir(dir, name)?;
+        for (host, guest) in self.dirs.iter() {
+            let dir = Dir::open_ambient_dir(host, ambient_authority())
+                .with_context(|| format!("failed to open directory '{}'", host))?;
+            builder.preopened_dir(dir, guest)?;
```
rylev, line 795: I'm confused by this. Why does the `preopened_dir` only take two arguments?

crates/wasi/src/bindings.rs
alexcrichton: Ah the `async_io` name doesn't show up at all, that's just to contain everything and then reexport only the bits we need. The only modules here are `bindings` and `bindings::sync`, but these bindings are relatively rare to access since they're the raw traits/types that most embedders won't be working with.

crates/wasi-http/src/proxy.rs
```diff
@@ -47,18 +47,24 @@ pub mod sync {
...
-        wasmtime_wasi::command::sync::add_to_linker(l)?;
+        wasmtime_wasi::bindings::clocks::wall_clock::add_to_linker(l, |t| t)?;
+        wasmtime_wasi::bindings::clocks::monotonic_clock::add_to_linker(l, |t| t)?;
+        wasmtime_wasi::bindings::sync::io::poll::add_to_linker(l, |t| t)?;
+        wasmtime_wasi::bindings::sync::io::streams::add_to_linker(l, |t| t)?;
+        wasmtime_wasi::bindings::io::error::add_to_linker(l, |t| t)?;
+        wasmtime_wasi::bindings::cli::stdin::add_to_linker(l, |t| t)?;
+        wasmtime_wasi::bindings::cli::stdout::add_to_linker(l, |t| t)?;
+        wasmtime_wasi::bindings::cli::stderr::add_to_linker(l, |t| t)?;
+        wasmtime_wasi::bindings::random::random::add_to_linker(l, |t| t)?;
```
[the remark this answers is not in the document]
alexcrichton: The main reason is that the auto-generated functions all take a closure that's `fn(&mut T) -> &mut U` but that's switched here to be `T: WasiView`. Otherwise though, you're right, and it's something I've wanted to clean up for awhile.

elliottt (commented):
This looks great to me! Just a few questions/suggestions.

crates/wasi/src/ctx.rs
```diff
@@ -325,18 +499,101 @@ impl WasiCtxBuilder {
...
+    ///
+    /// [`WasiP1Ctx`]: crate::preview1::WasiP1Ctx
+    /// [`preview1`]: crate::preview1
+    ///
+    /// # Panics
+    ///
+    /// Panics if this method is called twice. Each [`WasiCtxBuilder`] can be
+    /// used to create only a single [`WasiCtx`] or [`WasiP1Ctx`]. Repeated
+    /// usage of this method is not allowed and should use a second builder
+    /// instead.
```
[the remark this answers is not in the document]
alexcrichton: You can find more discussion of that [here](https://github.com/bytecodealliance/wasmtime/pull/6770)

elliottt (approved):
:shipit:

PR #8230, opened 2024-03-24
PCC: turn off fuzzing for now.

cfallin:
There have been more fuzzbugs than expected and the onslaught of issues is something I definitely don't have time to deal with right now; let's try again later in the year (unless someone else wants to drive this!).

This puts the fuzzing logic under an off-by-default feature so it can still be tested and developed in-tree as desired.


jameysharp (approved):
Aw, that's too bad. I hope we can take the time to figure out more of this soon.

PR #8231, opened 2024-03-25
cranelift: New InstructionData::map_values helper

jameysharp:
This is like the DataFlowGraph::map_inst_values method, but that method mutably borrows `self`, so it goes to a fair bit of trouble to drop its borrows around callbacks so the mutable borrow can be used in the callback too.

This new helper only actually needs to borrow two public fields from the DFG: `value_lists` and `jump_tables`. Therefore it's possible to use other fields in the callback as long as the compiler can see all the fields being used in the same scope. That's good enough for everywhere we were using this pattern, so we can simplify this way.

The case which motivated this change isn't shown here: It isn't possible to call `dfg.map_inst_values` on every instruction in `dfg.insts`, because the borrow on `dfg.insts` prevents taking a mutable borrow on `dfg`. But we can call this new helper in that case.


elliottt (approved):
Awesome!

PR #8232, opened 2024-03-25
Remove cranelift-wasm's "DummyEnvironment"

jameysharp:
There were two remaining uses, both in cranelift-tools, in the `wasm` and `souper-harvest` subcommands.

These days it's better to use `wasmtime compile --emit-clif` rather than `clif-util wasm`, in order to see the CLIF that we actually generate rather than a rough approximation of it.

And `clif-util souper-harvest` will give more relevant results if you feed it CLIF from `wasmtime compile --emit-clif` as well. To make that easier to do, I've made souper-harvest accept multiple input files, as well as accepting directories which recursively contain input files.


abrown (approved):

alexcrichton:
🎊

PR #8234, opened 2024-03-25
cranelift: Only enable winch calling convention for x64

saulecabrera:
This commit is a follow up to https://github.com/bytecodealliance/wasmtime/pull/8198; it ensures ensures that the Winch calling convention is only allowed  when the architecture is x64.


elliottt:
Thanks for catching this!

PR #8237, opened 2024-03-25
 Disassemble `*.cwasm` for `compile` disas tests 

alexcrichton:
This commit changes how the `compile` mode of the `disas` test suite
works. Previously this would use `--emit-clif` and run the Cranelift
pipeline for each individual function and use the custom VCode-based
disassembly for instruction output. This commit instead uses the raw
binary coming out of Wasmtime. The ELF file itself is parsed and is
disassembled in a manner similar to Winch tests.

The goal of this commit is somewhat twofold:

* Lay the groundwork to migrate all Winch-based filetests to
  `tests/disas`.
* Test the raw output from Cranelift/Wasmtime which includes
  optimizations like branch chomping in the `MachBuffer`.

This commit doesn't itself move the Winch tests yet, that's left for a
future commit.


fitzgen (approved):
Nice!

PR #8238, opened 2024-03-26
cranelift: Add DFG helper for resolving all value aliases

jameysharp:
And demonstrate its use in bugpoint. That doesn't have much impact on anything, which I hope makes this easier to review than changing more important things right away.

There are other places we would probably be better off doing a whole-function rewrite once rather than calling `resolve_aliases` a bunch of times, such as the disas tests, the egraph pass, and lowering. I have not changed those in this PR.


fitzgen (approved):
Nice!

PR #8239, opened 2024-03-26
Resolve value aliases in disas tests

jameysharp:
This reverts commit 6c5184809db3a92de4ee0c718c403bedc9a9ff4f, "Cranelift: resolve value aliases when printing CLIF functions (#8214)", then applies the same effect a different way.

In discussion on #8223, we decided to handle this a different way. So I've introduced a method on DataFlowGraph which eliminates all value aliases, and we can call it when necessary. If that function is not called then the CLIF printer should print value aliases the same way it did before #8214.

In this PR, I've specifically called it in disas tests. The changes to write.rs and frontend.rs are from the revert, while the changes to disas.rs are new.

In these tests, value aliases are just clutter that distracts from understanding what code will actually be generated. They may change due to changes in legalization, but that doesn't signal anything about whatever the test was intended to check.

Because the new helper deletes aliases after it's done resolving them, those aliases now disappear from the test expectations. Aside from that, the test expectations are unchanged compared to what #8214 did.


elliottt (approved):
Awesome! :tada:

PR #8240, opened 2024-03-26
egraph: Resolve all aliases at once

jameysharp:
This way we can use the linear-time alias rewriting pass, and then avoid having to think about value aliases ever again.


cfallin (approved):
Nice!

jameysharp:
@cfallin, could you have a look at my additional changes in 70b082781c93a3a3359597e36a9f5ef5f8feeca7 and make sure they make sense?

CI informed me that values deleted due to this alias-rewrite pass were still referenced. While auditing everything in the `DataFlowGraph` for references to `Value`s, I realized I think it makes sense to merge PCC facts onto the targets of value aliases. So that may affect your work on PCC, hopefully in good ways.

cfallin (approved):
New changes look good!

PR #8243, opened 2024-03-26
Migrate all Winch filetests to `tests/disas`

alexcrichton:
This commit builds on https://github.com/bytecodealliance/wasmtime/pull/8237 to migrate all Winch filetests to the top-level `tests/disas/` test suite. All files are copied to a `tests/disas/winch` folder.

This PR has a number of commit splitting out the various stages of this migration. The main changes are:

* `test = "winch"` is now supported in `disas` tests
* Support for the Winch ABI in Cranelift is ungated (it no longer panics) to get `wasmtime compile` working to get the tests working
* Winch tests now use ATT syntax instead of Intel syntax (the Cranelift default)

The next follow-up I'd like to do is to remove the filetest infrastructure for Winch, which will probably also remove the Winch executable itself. I'll do that as a follow-up though.


elliottt:
One change that we made to the winch disasm tests that I liked was to only print offsets after branch instructions (and always after a return). This made updates to filetests easier to read, as the diff was usually more precise. Would that be easy to add back in before this migration, or should we do that as a separate update?

alexcrichton:
I saw that yeah from Winch but I ended up leaving it out in favor of printing all offsets. I was a bit confused as to why the address after a jump instruction was printed because that didn't account for fall-through-style jumps (e.g. it didn't reliably catch all jump targets). For after-ret I was also not sure that would handle the case of multiple `ret` instructions in a function. 

I can add it back in though if desired!

alexcrichton:
Ah ok that makes sense. I've updated with that added back in plus some comments

elliottt (approved):

elliottt:
Thanks for changing the output back!

Issue #6759, opened 2023-07-21
Cranelift: Get the `tail` calling convention as fast as `fast` on Sightglass


fitzgen:
That is, regular calls with the `tail` calling convention should be as fast as regular calls with the `fast` calling convention.

https://github.com/bytecodealliance/wasmtime/issues/1065#issuecomment-1624395771

> So @jameysharp and I did a little profiling/investigation of switching the internal Wasm calling convention over to tail on our sightglass benchmarks. I was really expecting this to have no measurable change, but unfortunately it looks like it has a ~7% overhead on bz2 and spidermonkey.wasm and ~1% overhead on pulldown-cmark. This is surprising! We think this means that we ~frequently call functions that don't have enough register pressure to clobber all callee-save registers, and since tail only has caller-save registers and zero callee-save registers, we are doing more spills than we used to. Enough more that it is really measurable.

Here are the histograms of number of clobbered callee-save registers in a function for some of our benchmarks:

<details>

### pulldown-cmark

```
# Number of samples = 757
# Min = 0
# Max = 5
#
# Mean = 1.9682959048877162
# Standard deviation = 2.4428038716280174
# Variance = 5.967290755240832
#
# Each ∎ is a count of 9
#
 0 ..  1 [ 459 ]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
 1 ..  2 [   0 ]: 
 2 ..  3 [   0 ]: 
 3 ..  4 [   0 ]: 
 4 ..  5 [   0 ]: 
 5 ..  6 [ 298 ]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
 6 ..  7 [   0 ]: 
 7 ..  8 [   0 ]: 
 8 ..  9 [   0 ]: 
 9 .. 10 [   0 ]: 
```

### spidermonkey

```
# Number of samples = 18279
# Min = 0
# Max = 5
#
# Mean = 1.8119153126538674
# Standard deviation = 2.4034432514706436
# Variance = 5.77653946303978
#
# Each ∎ is a count of 233
#
[...]

PR #8246, opened 2024-03-26
cranelift: Support callee-saved registers with tail calls on x64

elliottt:
Rework the tail calling convention on x64 to support tail calls by changing the compilation strategy in the following ways:

1. Instead of making a shadow frame that we copy over the old one, instead determine if we need to grow or shrink the stack argument area and emit a special-purpose instruction to handle that case.
2. As we're now growing and shrinking the frame in a separate instruction sequence, emit arguments as moving to FP offsets instead of SP offsets, placing them where they'll need to go for the tail call being setup.
3. Finally, when emitting a `Inst::ReturnCall` in the x64 backend, reuse the `gen_clobber_save` and `gen_epilogue_frame_restore` functions to setup the frame and any callee-saved registers to the state expected by the callee we're jumping to.

With these three changes in place, we modified the x64 abi for tail calls to include a list of callee-saved registers, and observed that we now save them in the prologue, and restore them right before jumping to the tail-callee.

## TODO

- [x] Account for the extra space used by `GrowFrame` in the stack check
- [x] Revisit the use of `r14` for the stack limit in the `Tail` calling convention

### DIFF
--- a/cranelift/codegen/src/isa/riscv64/lower/isle.rs
+++ b/cranelift/codegen/src/isa/riscv64/lower/isle.rs
@@ -18,7 +18,7 @@ use crate::machinst::{VCodeConstant, VCodeConstantData};
 use crate::{
     ir::{
         immediates::*, types::*, AtomicRmwOp, BlockCall, ExternalName, Inst, InstructionData,
-        MemFlags, StackSlot, TrapCode, Value, ValueList,
+        MemFlags, Opcode, StackSlot, TrapCode, Value, ValueList,
     },
     isa::riscv64::inst::*,
     machinst::{ArgPair, InstOutput},
@@ -82,6 +82,9 @@ impl generated_code::Context for RV64IsleContext<'_, '_, MInst, Riscv64Backend> 
             self.lower_ctx.sigs(),
             callee_sig,
             &callee,
+            // TODO: this should be Opcode::ReturnCall, once riscv64 has been ported to the new
+            // tail call strategy.
+            Opcode::Call,
             distance,
             caller_conv,
             self.backend.flags().clone(),
--- a/cranelift/codegen/src/isa/x64/inst.isle
+++ b/cranelift/codegen/src/isa/x64/inst.isle
@@ -541,6 +541,16 @@
        (ReturnCallUnknown (callee RegMem)
                           (info BoxReturnCallInfo))
 
+       ;; Move the frame down in memory to accommodate additional stack
+       ;; arguments, when setting up for a tail call.
+       (GrowFrame (amount u32)
+                  (tmp WritableGpr))
+
+       ;; Move the frame up in memory to accommodate fewer stack arguments, when
+       ;; setting up for a tail call.
+       (ShrinkFrame (amount u32)
+                    (tmp WritableGpr))
+
        ;; A pseudo-instruction that captures register arguments in vregs.
        (Args
         (args VecArgPair))
--- a/cranelift/codegen/src/isa/x64/inst.isle
+++ b/cranelift/codegen/src/isa/x64/inst.isle
@@ -541,15 +541,22 @@
        (ReturnCallUnknown (callee RegMem)
                           (info BoxReturnCallInfo))
 
-       ;; Move the frame down in memory to accommodate additional stack
-       ;; arguments, when setting up for a tail call.
-       (GrowFrame (amount u32)
-                  (tmp WritableGpr))
+       ;; GrowArgumentArea does a memmove of everything in the frame except for
+       ;; the argument area, to make room for more arguments. That includes all
+       ;; the stack slots, the callee-saved registers, and the saved FP and
+       ;; return address. To keep the stack pointers in sync with that change,
+       ;; it also subtracts the given amount from both the FP and SP registers.
+       (GrowArgumentArea (amount u32)
+                         (tmp WritableGpr))
 
-       ;; Move the frame up in memory to accommodate fewer stack arguments, when
-       ;; setting up for a tail call.
-       (ShrinkFrame (amount u32)
-                    (tmp WritableGpr))
+       ;; ShrinkArgumentArea does a memmove of everything in the frame except
+       ;; for the argument area, to trim space for fewer arguments. That
+       ;; includes all the stack slots, the callee-saved registers, and the
+       ;; saved FP and return address. To keep the stack pointers in sync with
+       ;; that change, it also adds the given amount to both the FP and SP
+       ;; registers.
+       (ShrinkArgumentArea (amount u32)
+                           (tmp WritableGpr))
 
        ;; A pseudo-instruction that captures register arguments in vregs.
        (Args
--- a/cranelift/codegen/src/isa/x64/inst/emit.rs
+++ b/cranelift/codegen/src/isa/x64/inst/emit.rs
@@ -1722,6 +1722,103 @@ pub(crate) fn emit(
             }
         }
 
+        Inst::GrowFrame { amount, tmp } => {
+            debug_assert!(*amount > 0);
+            debug_assert_eq!(*amount % 8, 0);
+
+            let tmp = allocs.next(tmp.to_reg().to_reg());
+            let src = Gpr::new(tmp).unwrap();
+            let dst = WritableGpr::from_reg(src);
+
+            // As we're increasing the number of stack arguments, we need to move the frame down in
+            // memory, by decrementing SP by `amount` and looping from lower addresses to higher
+            // ones, copying down.
+
+            // TODO: this needs to be accounted for by the stack check, before `GrowFrame` is
+            // emitted, so that the increment here is expected.
+            //
+            // Decrement SP by `amount`
+            Inst::alu_rmi_r(
+                OperandSize::Size64,
+                AluRmiROpcode::Sub,
+                RegMemImm::imm(*amount),
+                Writable::from_reg(regs::rsp()),
+            )
+            .emit(&[], sink, info, state);
+
+            // The total size that we're going to copy, including the return address and frame
+            // pointer that are pushed on the stack alreadcy.
+            let size = i32::try_from(state.nominal_sp_to_fp()).unwrap() + 16;
+
+            debug_assert_eq!(size % 8, 0);
+
+            // Iterate word offsets from `SP + amount`, copying them to SP.
+            for sp_word_offset in 0..(size / 8) {
+                let sp_byte_offset = sp_word_offset * 8;
+                Inst::Mov64MR {
+                    src: SyntheticAmode::nominal_sp_offset(
+                        sp_byte_offset + i32::try_from(*amount).unwrap(),
+                    ),
+                    dst,
+                }
+                .emit(&[], sink, info, state);
+
+                Inst::MovRM {
+                    size: OperandSize::Size64,
+                    src,
+                    dst: SyntheticAmode::nominal_sp_offset(sp_byte_offset),
+                }
+                .emit(&[], sink, info, state);
+            }
+        }
+
+        Inst::ShrinkFrame { amount, tmp } => {
+            debug_assert!(*amount > 0);
+            debug_assert_eq!(*amount % 8, 0);
+
+            let tmp = allocs.next(tmp.to_reg().to_reg());
+            let src = Gpr::new(tmp).unwrap();
+            let dst = WritableGpr::from_reg(src);
+
+            // As we're decreasing the number of stack arguments, we need to move the frame up in
+            // memory, looping from higher addresses to lower ones copying up, and finally
+            // incrementing `SP` by `amount`.
+
+            // The total size that we're going to copy, including the return address and frame
+            // pointer that are pushed on the stack alreadcy.
+            let size = i32::try_from(state.nominal_sp_to_fp()).unwrap() + 16;
+
+            debug_assert_eq!(size % 8, 0);
+
+            // Iterate word offsets from `SP`, copying them to `SP + amount`.
+            for sp_word_offset in (0..(size / 8)).rev() {
+                let sp_byte_offset = sp_word_offset * 8;
+                Inst::Mov64MR {
+                    src: SyntheticAmode::nominal_sp_offset(sp_byte_offset),
+                    dst,
+                }
+                .emit(&[], sink, info, state);
+
+                Inst::MovRM {
+                    size: OperandSize::Size64,
+                    src,
+                    dst: SyntheticAmode::nominal_sp_offset(
+                        sp_byte_offset + i32::try_from(*amount).unwrap(),
+                    ),
+                }
+                .emit(&[], sink, info, state);
+            }
+
+            // Increment SP by `amount`
+            Inst::alu_rmi_r(
+                OperandSize::Size64,
+                AluRmiROpcode::Add,
+                RegMemImm::imm(*amount),
+                Writable::from_reg(regs::rsp()),
+            )
+            .emit(&[], sink, info, state);
+        }
+
         Inst::Args { .. } => {}
         Inst::Rets { .. } => {}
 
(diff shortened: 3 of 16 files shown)


cranelift/codegen/src/isa/x64/inst/emit.rs
```diff
@@ -1722,6 +1700,119 @@ pub(crate) fn emit(
...
+                OperandSize::Size64,
+                AluRmiROpcode::Sub,
+                RegMemImm::imm(*amount),
+                Writable::from_reg(regs::rbp()),
+            )
+            .emit(&[], sink, info, state);
+
+            // The total size that we're going to copy, including the return address and frame
+            // pointer that are pushed on the stack alreadcy.
+            let size = i32::try_from(state.nominal_sp_to_fp()).unwrap() + 16;
```
jameysharp, line 1736: After we wrote this we discovered the `fp_to_arg_offset` method which I now realize we could also use here instead of hard-coding 16. That might also be handy for generalizing this in shared code when we extend this work to other targets.

jameysharp (commented):
I'm proud of the work Trevor and I did on this!

Here's a few quick comments edits after giving this another complete read-through.

jameysharp:
Also I see we have only one function in the precise-output compile filetests that exercises `grow_frame`, and none that exercise `shrink_frame`. I suppose we ought to extend the tests.

fitzgen:
Super excited for this! Going to start digging in.

Concurrently, would you mind running the default sightglass suite with vs without this change (when enabling the wasm tail calls proposal for both) to determine if this fully resolves https://github.com/bytecodealliance/wasmtime/issues/6759 ?

fitzgen:
> Concurrently, would you mind running the default sightglass suite with vs without this change (when enabling the wasm tail calls proposal for both) to determine if this fully resolves #6759 ?

Or I guess the comparison we really care about is

* `main` with default wasm features, versus
* this branch with wasm tail calls enabled

This gives us the "wasmtime" vs "tail" calling conventions comparison we want.

cranelift/codegen/src/isa/x64/inst/emit.rs
```diff
@@ -1722,6 +1722,103 @@ pub(crate) fn emit(
...
+            // memory, looping from higher addresses to lower ones copying up, and finally
+            // incrementing `SP` by `amount`.
+
+            // The total size that we're going to copy, including the return address and frame
+            // pointer that are pushed on the stack alreadcy.
+            let size = i32::try_from(state.nominal_sp_to_fp()).unwrap() + 16;
+
+            debug_assert_eq!(size % 8, 0);
+
+            // Iterate word offsets from `SP`, copying them to `SP + amount`.
```
fitzgen, line 1778: Same suggestion on rewording this comment the way I suggested for the corresponding comment in `GrowFrame`.

cranelift/codegen/src/isa/riscv64/lower/isle.rs
```diff
@@ -82,6 +82,9 @@ impl generated_code::Context for RV64IsleContext<'_, '_, MInst, Riscv64Backend>
             self.lower_ctx.sigs(),
             callee_sig,
             &callee,
+            // TODO: this should be Opcode::ReturnCall, once riscv64 has been ported to the new
+            // tail call strategy.
+            Opcode::Call,
```
fitzgen, line 87: Is this going to break return calls on non-x64 until we get around to poring these architectures to the new tail call strategy? I don't think this PR should regress any existing functionality.

cranelift/codegen/src/isa/x64/inst/emit.rs
```diff
@@ -1737,14 +1715,21 @@ pub(crate) fn emit(
...
                 Writable::from_reg(regs::rsp()),
             )
             .emit(&[], sink, info, state);
+            Inst::alu_rmi_r(
+                OperandSize::Size64,
+                AluRmiROpcode::Sub,
+                RegMemImm::imm(*amount),
+                Writable::from_reg(regs::rbp()),
+            )
+            .emit(&[], sink, info, state);
```
fitzgen, line 1732: Can we debug assert that frame pointers are enabled here? Or guard this on frame pointers being enabled?

cranelift/codegen/src/isa/x64/inst/emit.rs
```diff
@@ -1817,6 +1802,15 @@ pub(crate) fn emit(
...
             .emit(&[], sink, info, state);
+
+            // Increment FP by `amount`
+            Inst::alu_rmi_r(
+                OperandSize::Size64,
+                AluRmiROpcode::Add,
+                RegMemImm::imm(*amount),
+                Writable::from_reg(regs::rbp()),
+            )
+            .emit(&[], sink, info, state);
```
fitzgen, line 1813: Likewise about frame pointers being enabled.

cranelift/codegen/src/isa/x64/inst/emit.rs
```diff
@@ -4366,125 +4355,18 @@ fn emit_return_call_common_sequence(
...
-    //            +---------------------+
-    //            | ...                 |
-    //    new     | new stack arguments |
-    //    frame   | ...                 | <-- SP
-    //            +---------------------+
-    //
-    // We need to restore the old FP, copy the new stack arguments over the old
-    // stack arguments, write the return address into the correct slot just
-    // after the new stack arguments, adjust SP to point to the new return
-    // address, and then jump to the callee (which will push the old FP again).
```
fitzgen: FWIW, this is the kind of thing I'm referring to when I talk about diagrams being nice to have in `{Grow,Shrink}Frame`.

cranelift/codegen/src/isa/x64/inst/mod.rs
```diff
@@ -52,20 +52,6 @@ pub struct CallInfo {
...
-    /// The size of the current/old stack frame's stack arguments.
-    pub old_stack_arg_size: u32,
-    /// The return address. Needs to be written into the correct stack slot
-    /// after the new stack frame is copied into place.
-    pub ret_addr: Option<Gpr>,
-    /// A copy of the frame pointer, because we will overwrite the current
-    /// `rbp`.
-    pub fp: Gpr,
-    /// A temporary register.
-    pub tmp: WritableGpr,
```
fitzgen, line 68: Nice.

cranelift/codegen/src/machinst/abi.rs
```diff
@@ -2069,7 +2072,7 @@ impl<M: ABIMachineSpec> Callee<M> {
 #[derive(Clone, Debug)]
 pub enum ArgLoc {
     Reg(PReg),
-    Stack(StackAMode),
+    Stack { offset: i64, ty: ir::Type },
```
fitzgen, line 2075: Can we document what base this is an offset from? SP right?

fitzgen: (And presumably this change is because we don't need the full generality of `StackAMode` anymore?)

cranelift/codegen/src/machinst/abi.rs
```diff
@@ -2329,7 +2344,17 @@ impl<M: ABIMachineSpec> CallSite<M> {
                     vreg,
                     preg: preg.into(),
                 }),
-                ArgLoc::Stack(amode) => ctx.emit(M::gen_store_stack(amode, vreg, amode.get_type())),
+                ArgLoc::Stack { offset, ty } => {
+                    let amode = if self.is_tail_call() {
+                        StackAMode::FPOffset(
+                            offset + M::fp_to_arg_offset(self.caller_conv, &self.flags),
+                            ty,
+                        )
```
fitzgen, line 2352: Can we debug assert that we have frame pointers enabled here?

cranelift/codegen/src/machinst/abi.rs
fitzgen: I guess the offset is either SP or FP depending on whether the call is a tail call or not? That is pretty subtle and seems relatively easy to mix up. Is there a reason we can't keep using `StackAMode` here? That would avoid mix ups because we would always have to match on the amode to determine whether we are dealing with SP or FP offsets. I think it would be fine to have `unreachable!()` arms in these matches if we always use SP for regular calls and FP for tail calls, but avoiding the mix-up footgun is nice.

fitzgen (commented):
Thanks!! Lots of nitpicks, but generally really like this.

+1 to comment about more filetests and the lack of `shrink_frame` testing. Seems like it would be good to exercise the following cases, which it seems like we aren't already:

* tail call from a function with multiple stack args to a function with a single stack arg
* tail call from a function with multiple stack args to a function with zero stack args
* tail call from a function with a single stack arg to a function with multiple stack args
* tail call from a function with zero stack args to a function with multiple stack args

cranelift/codegen/src/isa/riscv64/lower/isle.rs
jameysharp: Nope, we're using `Call` instead of `ReturnCall` here specifically to avoid breaking aarch64 and riscv64 until we can get around to applying similar changes to them. `from_func` was always using `Call` for the opcode before so passing that here now doesn't change what data structures that function constructs, and then we rely on the distinction in the new `is_tail_call` helper to ensure that the generated code isn't changed.

cranelift/codegen/src/isa/x64/inst.isle
```diff
@@ -541,6 +541,16 @@
        (ReturnCallUnknown (callee RegMem)
                           (info BoxReturnCallInfo))
 
+       ;; Move the frame down in memory to accommodate additional stack
+       ;; arguments, when setting up for a tail call.
+       (GrowFrame (amount u32)
+                  (tmp WritableGpr))
```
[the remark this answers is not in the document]
jameysharp: `GrowFrame` does a memmove of everything in the frame except for the argument area, to make more room in the argument area. That includes all the stack slots, the callee-saved registers, and the saved FP and return address. To keep the stack pointers in sync with that change, it also subtracts the given amount from both the FP and SP registers. `ShrinkFrame` is the same but in the other direction.

Ulrich's suggestion in today's Cranelift meeting that we allocate the maximum necessary space during the prologue instead lets us get rid of this memmove and only maybe have to move the return address, but we want to land this version first.

cranelift/codegen/src/isa/x64/inst/emit.rs
```diff
@@ -1722,6 +1722,103 @@ pub(crate) fn emit(
...
+            let tmp = allocs.next(tmp.to_reg().to_reg());
+            let src = Gpr::new(tmp).unwrap();
+            let dst = WritableGpr::from_reg(src);
+
+            // As we're increasing the number of stack arguments, we need to move the frame down in
+            // memory, by decrementing SP by `amount` and looping from lower addresses to higher
+            // ones, copying down.
+
+            // TODO: this needs to be accounted for by the stack check, before `GrowFrame` is
+            // emitted, so that the increment here is expected.
```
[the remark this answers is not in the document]
jameysharp: No. I haven't traced through enough to understand how we should influence the stack check yet.

cranelift/codegen/src/isa/x64/inst/emit.rs
jameysharp: We haven't removed the assert in `emit_return_call_common_sequence` that frame pointers are enabled, but yeah, we should add that in both `ShrinkFrame` and `GrowFrame` as well.

cranelift/codegen/src/machinst/abi.rs
jameysharp: This offset always describes the number of bytes into the argument area. Whether we find the argument area relative to FP or SP depends on how `gen_arg`'s caller is set up, not on ABI details. That's why we switched away from encoding that decision here.

cranelift/codegen/src/isa/x64/inst/emit.rs
```diff
@@ -1722,6 +1722,103 @@ pub(crate) fn emit(
...
+                    dst,
+                }
+                .emit(&[], sink, info, state);
+
+                Inst::MovRM {
+                    size: OperandSize::Size64,
+                    src,
+                    dst: SyntheticAmode::nominal_sp_offset(sp_byte_offset),
+                }
+                .emit(&[], sink, info, state);
```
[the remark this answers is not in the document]
elliottt: I went with `tmp` and `tmp_w` like we had in the previous implementation :+1:

cranelift/codegen/src/isa/x64/inst.isle
```diff
@@ -541,6 +541,23 @@
...
+                         (tmp WritableGpr))
+
+       ;; ShrinkArgumentArea does a memmove of everything in the frame except
+       ;; for the argument area, to trim space for fewer arguments. That
+       ;; includes all the stack slots, the callee-saved registers, and the
+       ;; saved FP and return address. To keep the stack pointers in sync with
+       ;; that change, it also adds the given amount to both the FP and SP
+       ;; registers.
+       (ShrinkArgumentArea (amount u32)
+                           (tmp WritableGpr))
```
fitzgen, line 559: These comments are fantastic -- thanks!

cranelift/filetests/filetests/isa/x64/return-call.clif
```diff
@@ -191,20 +185,177 @@ block0(v0: i8):
...
+;   movq %rbp, %rsp
+;   popq %rbp
+;   retq $0x10
+
+function %call_one_stack_arg(i32, i32, i32, i32, i32, i32, i32, i32, i32) tail {
+    fn0 = colocated %one_stack_arg(i32, i32, i32, i32, i32, i32, i32) tail
+
+block0(v0: i32, v1: i32, v2: i32, v3: i32, v4: i32, v5: i32, v6: i32, v7: i32, v8: i32):
+    return_call fn0(v2, v3, v4, v5, v6, v7, v8)
+}
```
fitzgen, line 232: Nice, thanks for these tests.

fitzgen (approved):
Looks great! Thanks! :tada: 

Excited for the benchmark results!

elliottt:
Here are the execution benchmarks for `spidermonkey`, `bz2`, and `pulldown-cmark` relative to the current tail calls implementation on main:

```
execution :: instructions-retired :: benchmarks/spidermonkey/benchmark.wasm

  Δ = 204255214.50 ± 513219.80 (confidence = 99%)

  tail-calls.so is 1.08x to 1.08x faster than main.so!

  [2862185961 2862312218.75 2862598037] main.so
  [2657985419 2658057004.25 2658128352] tail-calls.so

execution :: instructions-retired :: benchmarks/bz2/benchmark.wasm

  Δ = 13264568.75 ± 1.43 (confidence = 99%)

  tail-calls.so is 1.06x to 1.06x faster than main.so!

  [239534828 239534828.25 239534829] main.so
  [226270259 226270259.50 226270260] tail-calls.so

execution :: instructions-retired :: benchmarks/pulldown-cmark/benchmark.wasm

  Δ = 391358.50 ± 894.30 (confidence = 99%)

  tail-calls.so is 1.02x to 1.02x faster than main.so!

  [20465937 20466114.00 20466584] main.so
  [20074442 20074755.50 20075106] tail-calls.so
```

Running the same benchmarks on main without tail calls and this branch with tail calls separately, yields the following results:

### main
```
execution
  benchmarks/bz2/benchmark.wasm
    instructions-retired
      [226286674 226286674.00 226286674] /home/trevor/src/main.so
  benchmarks/pulldown-cmark/benchmark.wasm
    instructions-retired
      [20080325 20080512.75 20080999] /home/trevor/src/main.so
  benchmarks/spidermonkey/benchmark.wasm
    instructions-retired
      [2659583881 2659624337.75 2659672231] /home/trevor/src/main.so
```

### this branch
```
execution
  benchmarks/bz2/benchmark.wasm
    instructions-retired
      [226270259 226270259.50 226270260] /home/trevor/src/tail-calls.so
  benchmarks/pulldown-cmark/benchmark.wasm
    instructions-retired
      [20074405 20074419.00 20074445] /home/trevor/src/tail-calls.so
  benchmarks/spidermonkey/benchmark.wasm
    instructions-retired
      [2657979471 2658099662.50 2658220755] /home/trevor/src/tail-calls.so
```

It's exciting to see that the max for the tail-calls branch is smaller than the min of main for spidermonkey, though even if that's spurious they're still in the same ballpark :tada:

fitzgen:
Ship it!

PR #8248, opened 2024-03-27
Remove the `winch-tools` crate and CLI

alexcrichton:
With all Winch tests moved to `tests/disas` in #8243 plus the support of `wasmtime compile -C compiler=winch` this tool should in theory be supplanted nowadays with other alternatives. This commit removes the executable and the `winch-filetests` support.


alexcrichton:
I'll note that I'm not wed to this at all, and @saulecabrera or @elliottt or others find this tool useful I'm happy to close this. This can mostly be seen as asking the question of "with today's integration of Winch, is this executable still needed?"

saulecabrera (approved):
Makes sense to me. I don't have any particular concerns with removing this crate. Personally I feel that this crate was mostly useful in the early days when Winch was not fully integrated with Wasmtime, then it was mostly kept around for filetests.

elliottt (approved):

PR #8250, opened 2024-03-28
Add a `compile` feature to `wasmtime-environ`

alexcrichton:
This commit adds a compile-time feature to remove some dependencies of the `wasmtime-environ` crate. This compiles out support for compiling modules/components and makes the crate slimmer in terms of amount of code compiled along with its dependencies. Much of this should already have been statically removed by native linkers so this likely won't have any compile-size impact, but it's a nice-to-have in terms of organization.

This has a fair bit of shuffling around of code, but apart from renamings and movement there are no major changes here.


fitzgen (approved):
Didn't look closely at the actual code blocks, assuming it is all just code motion.

PR #8256, opened 2024-03-28
Provide example usage in wasmtime::component mod doc

rylev:
This might be slightly controversial as I think we want to push users towards using `wasmtime::component::bindgen!`, but it seems appropriate at the module level to show some simple code that gives an overview of what the module provides.


alexcrichton (approved):
Looks great to me, thanks!

PR #8259, opened 2024-03-28
Update Wasmtime's policy on `cargo vet`

alexcrichton:
This was discussed at today's Wasmtime meeting out of some concerns around our current policies. Namely I felt the current state of affairs is not striking the right balance between cost and benefit with our usage of `cargo vet`. After discussion we've reached consensus around two changes to our `cargo vet` policy documented here in this PR:

* An exemption can be added for "popular crates" at any time with no review required. This should handle most big crates that are needed for various dependencies. The thinking behind this is that a supply-chain attack against these crates is highly likely to be detected in a short time due to their popularity. Coupled with the fact that changes to Wasmtime take a minimum of two weeks to get released means that it's an unlikely exploitation vector.

* Maintainers are recommended to push directly to contributor's PRs for `cargo vet` entries instead of making separate PRs. This avoids the need for contributor rebasing and additionally solves the problem where the `vet` entries land in a separate PR but then the contributor's PR takes much longer to land. In the interim some `vet` entries have been cleaned up by accident which requires re-landing the PR to add the entries.

### DIFF
--- a/docs/contributing-coding-guidelines.md
+++ b/docs/contributing-coding-guidelines.md
@@ -61,6 +61,18 @@ update existing dependencies or add new dependencies will not be mergeable by
 default (CI will fail). This is expected from our project's configuration and
 this situation will be handled one of a few ways:
 
+Note that this process is not in place to prevent new dependencies or prevent
+updates, but rather it ensures that development of Wasmtime is done with a
+trusted set of code that has been reviewed by trusted parties. We welcome
+dependency updates and new functionality, so please don't be too alarmed when
+contributing and seeing a failure of `cargo vet` on CI!
+
+### `cargo vet` for Contributors
+
+If you're a contributor to Wasmtime and you've landed on this documentation,
+hello and thanks for your contribution! Here's some guidelines for changing the
+set of dependencies in Wasmtime:
+
 * If a new dependency is being added it might be worth trying to slim down
   what's required or avoiding the dependency altogether. Avoiding new
   dependencies is best when reasonable, but it is not always reasonable to do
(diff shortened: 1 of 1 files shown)


alexcrichton:
Some points of note:

* I kept the 10,000 crate download metric since spot checking a few dependencies seemed to show that as a resonable threshold. For example from the ONNX PR the `compact_str` dependency clears this threshold but the `capstone` dependency does not. That makes sense to me since I don't expect disassembling to be all that popular on crates.io
* I briefly looked into what it might take to "comment on a PR if a github actions fails". I'd love to drop a comment on PRs from contributors pointing to this documentation when the `cargo vet` entry fails in CI. Turns out GitHub Actions has an [example of doing exactly that](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#using-data-from-the-triggering-workflow) (well, mostly), but it's pretty nasty so I didn't pursue it yet.

docs/contributing-coding-guidelines.md
```diff
@@ -72,17 +84,72 @@ this situation will be handled one of a few ways:
...
+
+* For contributors who should not add vet entries themselves maintainers should
+  review the PR and then when the PR is ready to be approved the maintainer
+  should push a commit to the contributor's PR directly. Maintainers should
+  check out the PR locally, apply `cargo vet` entries as needed, and then push
+  to the contributor's PR directly.
+
+Note for the last case it's important to ensure that after you push to the PR
+any future updates pushed by the contributor either contain or don't overwrite
+your vet entries. It's recommended to add vet entries just before approval to
```
cfallin, line 120: Let's add some explicit note heres too -- maybe something like "... or don't overwrite your vet entries. Also verify that if the PR branch is rebased or force-pushed, the details of your previously pushed vetting remain the same: e.g., versions were not bumped and descriptive reasons remain the same. If pushing a vetting commit to a contributor's PR and also asking for more changes, request that the contributor make the requested fixes in an additional commit rather than force-pushing a rewritten history, so your existing vetting commit remains untouched. These guidelines make it easier to verify no tampering has occurred."

cfallin (approved):
LGTM -- this accurately represents what we came out of the discussion with, I think. Thanks very much for writing up the details!

PR #8260, opened 2024-03-28
Gate support for the wasm `threads` proposal behind a Cargo feature

alexcrichton:
This commit moves support for the `threads` proposal behind a new on-by-default Cargo feature: `threads`. This is intended to support building Wasmtime with fewer runtime dependencies such as those required for the atomic operations on memories.

This additionally adds the `gc` feature in a few missing places too.


abrown (approved):
Looks good to me; I assumed all the shared memory and parking spot changes were just refactoring, no functional changes.

Issue #8257, opened 2024-03-28
FileOutputStream::write() modify the position repeatedly


liutao-liu:
Two modifications to the ```p``` code：
https://github.com/bytecodealliance/wasmtime/blob/9c92881d3b936b5d2e593ded370609a6e31717d8/crates/wasi/src/filesystem.rs#L359

https://github.com/bytecodealliance/wasmtime/blob/9c92881d3b936b5d2e593ded370609a6e31717d8/crates/wasi/src/preview1.rs#L1705

The former is an invalid modification, because ```self.mode``` is not a reference of postion in [fd_write](https://github.com/bytecodealliance/wasmtime/blob/9c92881d3b936b5d2e593ded370609a6e31717d8/crates/wasi/src/preview1.rs#L1670) , which is passed by value in [write_via_stream](https://github.com/bytecodealliance/wasmtime/blob/9c92881d3b936b5d2e593ded370609a6e31717d8/crates/wasi/src/preview1.rs#L1690).

Please help confirm if this is a bug. If it is, I'll submit a pr with my local changes.

alexcrichton:
Could you clarify with a test case what the expected behavior is? Reviewing these mutations I don't see anything amiss myself, but a test case could help understand if there's a bug here

liutao-liu:
> Could you clarify with a test case what the expected behavior is? Reviewing these mutations I don't see anything amiss myself, but a test case could help understand if there's a bug here

Delete the modification in [here](https://github.com/bytecodealliance/wasmtime/blob/9c92881d3b936b5d2e593ded370609a6e31717d8/crates/wasi/src/filesystem.rs#L359) All test case can run success. The reason is what I said before:```self.mode``` is not a reference of postion in [fd_write](https://github.com/bytecodealliance/wasmtime/blob/9c92881d3b936b5d2e593ded370609a6e31717d8/crates/wasi/src/preview1.rs#L1670) , which is passed by value in [write_via_stream](https://github.com/bytecodealliance/wasmtime/blob/9c92881d3b936b5d2e593ded370609a6e31717d8/crates/wasi/src/preview1.rs#L1690).

![image](https://github.com/bytecodealliance/wasmtime/assets/10509166/b2d5fd02-8e3c-4273-8333-c1fe88432862)

alexcrichton:
[...]

PR #8268, opened 2024-03-30
fix invalid modification of file position when write file

liutao-liu:
- from [issue](https://github.com/bytecodealliance/wasmtime/issues/8257)

- ```self.mode``` is not a reference of postion in [fd_write](https://github.com/bytecodealliance/wasmtime/blob/9c92881d3b936b5d2e593ded370609a6e31717d8/crates/wasi/src/preview1.rs#L1670) , which is passed by value in [write_via_stream](https://github.com/bytecodealliance/wasmtime/blob/9c92881d3b936b5d2e593ded370609a6e31717d8/crates/wasi/src/preview1.rs#L1690)

### DIFF
--- a/crates/wasi/src/filesystem.rs
+++ b/crates/wasi/src/filesystem.rs
@@ -354,12 +354,7 @@ impl HostOutputStream for FileOutputStream {
             }
         });
         self.state = match result {
-            SpawnBlocking::Done(Ok(nwritten)) => {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! I followed up [here](https://github.com/bytecodealliance/wasmtime/issues/8257#issuecomment-2028887297) and also [posted a test](https://github.com/bytecodealliance/wasmtime/pull/8271) which fails if this PR lands

PR #8270, opened 2024-03-31
Fix the SystemV abi of i128 arguments

bjorn3:
i128 arguments must always be passed either fully in registers or fully on the stack according to GCC. LLVM until recently allowed splitting it into a half passed in a register and a half passed on the stack. This was recently fixed. For cg_clif to remain ABI compatible with cg_llvm, it is necessary to apply the same fix to Cranelift.

See also https://blog.rust-lang.org/2024/03/30/i128-layout-update.html

### DIFF
--- a/cranelift/codegen/src/isa/x64/abi.rs
+++ b/cranelift/codegen/src/isa/x64/abi.rs
@@ -153,9 +153,10 @@ impl ABIMachineSpec for X64ABIMachineSpec {
             //
             // - If `enable_llvm_abi_extensions` is set in the flags, each
             //   `i128` is split into two `i64`s and assigned exactly as if it
-            //   were two consecutive 64-bit args. This is consistent with LLVM's
-            //   behavior, and is needed for some uses of Cranelift (e.g., the
-            //   rustc backend).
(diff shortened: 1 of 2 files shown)


cranelift/codegen/src/isa/x64/abi.rs
```diff
@@ -153,9 +153,10 @@ impl ABIMachineSpec for X64ABIMachineSpec {
             //
             // - If `enable_llvm_abi_extensions` is set in the flags, each
             //   `i128` is split into two `i64`s and assigned exactly as if it
-            //   were two consecutive 64-bit args. This is consistent with LLVM's
-            //   behavior, and is needed for some uses of Cranelift (e.g., the
-            //   rustc backend).
+            //   were two consecutive 64-bit args, except that if one of the
+            //   two halves is forced onto the stack, the other half is too.
+            //   This is consistent with LLVM's behavior, and is needed for
+            //   some uses of Cranelift (e.g., the rustc backend).
```
bjorn3, line 159: The godbolt link below is outdated, but until current rustc beta gets promoted to stable I can't make a new permalink for a specific version.

alexcrichton (approved):

PR #8275, opened 2024-04-01
cranelift: Fix indirect tail calls on aarch64/riscv64

jameysharp:
x64 now uses a dedicated codegen strategy for tail-calls which writes the callee's stack arguments directly into their final location, but aarch64 and riscv64 still use our previous strategy of setting up for a normal function call and then moving the new stack frame up to overwrite the old one. In order to emit correct code for a normal function call, we need to fake a normal Call/CallIndirect opcode rather than ReturnCall/ReturnCallIndirect.

The new tests are a combination of x64's compile-tests for return-call-indirect.clif, plus a large-stack-frame test from aarch64/riscv64's return-call.clif modified to do an indirect call.

### DIFF
--- a/cranelift/codegen/src/isa/aarch64/lower/isle.rs
+++ b/cranelift/codegen/src/isa/aarch64/lower/isle.rs
@@ -131,7 +131,9 @@ impl Context for IsleContext<'_, '_, MInst, AArch64Backend> {
             self.lower_ctx.sigs(),
             callee_sig,
             callee,
-            Opcode::ReturnCallIndirect,
+            // TODO: this should be Opcode::ReturnCallIndirect, once aarch64 has
+            // been ported to the new tail call strategy.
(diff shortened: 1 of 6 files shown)


fitzgen (approved):
Can we have some runtests exercising this stuff as well?

jameysharp:
> Can we have some runtests exercising this stuff as well?

We do have `cranelift/filetests/filetests/runtests/return-call-indirect.clif`, we just didn't have the indirect case tested in compile-tests for aarch64 and riscv64. Do you want me to just add a similar stack-arguments test there? It wouldn't have caught this bug because we have `preserve_frame_pointers=true` in that test, since x64 currently requires it, but there's no harm adding it. I could do the same in the x64 compile tests for that matter.

fitzgen:
I think it would be good to have, yeah

Issue #7973, opened 2024-02-21
Poor performance of wasmtime file I/O maybe because tokio 


liutao-liu:
### Test Case

test.c
```
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <fcntl.h>

int main()
{
    FILE *fp; 
    char str[14];
    for (int i = 1; i < 1000; i++)
    {

        int fd = open("test.txt", O_CREAT | O_WRONLY, S_IRUSR | S_IWUSR);
        for (int j = 0; j < 1000; j++)
        {

            lseek(fd, i * 10 + j * 13, SEEK_SET);
            write(fd, "hello world! ", 13);
        }
        fsync(fd);
        close(fd); 
    }

    return 0;
}
```

### Steps to Reproduce

* first，compile test.c in the preceding test case into WASM bytecode using the WASI SDK
```
wasi-sdk/bin/clang -O3 test.c -o test.wasm
```
* second, WASMTIME AOT compile ,got test.aot.
```
wasmtime compile  -W simd,relaxed-simd test.wasm -o test.aot
```
* third, Test Case Running Duration. It takes about 40 seconds. 
```
time wasmtime run --allow-precompiled --dir ./ test.aot
```

### Expected Results
wasmtime takes about the same time as native and wamr.

### Actual Results

Wasmtime takes about 23 seconds. 
The same test.c, native or wamr only takes about 2 seconds.

### Versions and Environment

Wasmtime version :16.0.0

Operating system: ubuntu 20.04

Architecture: aarch64 (same as x86 for this case)

#### Profile

```
#  profile for wasmtime
perf record -g -k mono wasmtime run --profile=jitdump --allow-precompiled --dir ./ test.aot
sudo perf inject --jit --input perf.data --output perf.jit.data
perf report -i perf.jit.data
```
As shown in the following figure, most performance hotspots are on Tokio. This is because wasmtime uses Tokio to implement the file I/O interface, involving:
```
__imported_wasi_snapshot_preview1_fd_read  
__imported_wasi_snapshot_preview1_fd_seek  
__imported_wasi_snapshot_preview1_fd_sync 
__imported_wasi_snapshot_preview1_fd_write
 ```

![image](https://github.com/bytecodealliance/wasmtime/assets/10509166/082e97b5-3a03-40bf-87fa-9041de881571)

```
#  profile for native
perf record -g -k mono ./test
perf report 
```
[...]

PR #8277, opened 2024-04-01
Update wiggle's `BorrowChecker` implementation

alexcrichton:
This commit is a refactoring and modernization of wiggle's `BorrowChecker` implementation. This type is quite old and predates everything related to the component model for example. This type additionally predates the implementation of WASI threads for Wasmtime as well. In general, this type is old and has not been updated in a long time.

Originally a `BorrowChecker` was intended to be a somewhat cheap method of enabling the host to have active safe shared and mutable borrows to guest memory. Over time though this hasn't really panned out. The WASI threads proposal, for example, doesn't allow safe shared or mutable borrows at all. Instead everything must be modeled as a copy in or copy out of data. This means that all of `wasmtime-wasi` and `wasi-common` have largely already been rewritten in such a way to minimize borrows into linear memory.

Nowadays the only types that represent safe borrows are the `GuestSlice` type and its equivalents (e.g. `GuestSliceMut`, `GuestStr`, etc). These are minimally used throughout `wasi-common` and `wasmtime-wasi` and when they are used they're typically isolated to a small region of memory.

This is all coupled with the facst that `BorrowChecker` never ended up being optimized. It's a `Mutex<HashMap<..>>` effectively and a pretty expensive one at that. The `Mutex` is required because `&BorrowChecker` must both allow mutations and be `Sync`. The `HashMap` is used to implement precise byte-level region checking to fulfill the original design requirements of what `wiggle` was envisioned to be.

Given all that, this commit guts `BorrowChecker`'s implementation and functionality. The type is now effectively a glorified `RefCell` for the entire span of linear memory. Regions are no longer considered when borrows are made and instead a shared borrow is considered as borrowing the entirety of shared memory. This means that it's not possible to simultaneously have a safe shared and mutable borrow, even if they're disjoint, at the same time.

The goal of this commit is to address performance issues seen in #7973 which I've seen locally as well. The heavyweight implementation of `BorrowChecker` isn't really buying us much nowadays, especially with much development having since moved on to the component model. The hope is that this much coarser way of implementing borrow checking, which should be much more easily optimizable, is sufficient for the needs of WASI and not a whole lot else.

### DIFF
--- a/crates/wiggle/src/borrow.rs
+++ b/crates/wiggle/src/borrow.rs
@@ -1,259 +1,113 @@
 use crate::{BorrowHandle, GuestError, Region};
-use std::collections::HashMap;
-use std::sync::Mutex;
+use std::sync::atomic::{AtomicU32, Ordering::Relaxed};
 
+/// A simple borrow checker to implement the API guarantees of Wiggle.
+///
+/// This is not a generalized borrow checker and is coarse-grained where it
+/// doesn't actually take any regions into account. Instead it only tracks
+/// whether there are temporally any shared/mutable borrows. This is
+/// more-or-less a poor-man's `RefCell<T>`.
+///
+/// Note that this uses `&AtomicU32` because this is passed around as
+/// `&BorrowChecker` in all `GuestPtr` structures. This needs to be mutated
+/// which might look like it needs `Cell<u32>`, but `&Cell<u32>` isn't `Sync`
+/// and we want futures with `&BorrowChecker` to be `Sync`, so this is an atomic
+/// instead. Only one of these is created per-invocation though and it's not
+/// actually shared across threads, so mutations here are not done with
+/// compare-and-swap but instead just loads/stores.
 pub struct BorrowChecker {
-    /// Unfortunately, since the terminology of std::cell and the problem domain of borrow checking
-    /// overlap, the method calls on this member will be confusing.
-    ///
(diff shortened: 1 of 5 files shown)


crates/wiggle/src/borrow.rs
```diff
@@ -1,259 +1,113 @@
...
-        !(self.shared_borrows.is_empty() && self.mut_borrows.is_empty())
+    pub fn shared_unborrow(&self, _: BorrowHandle) {
+        let prev = self.state.load(Relaxed);
+        debug_assert!(prev > 0);
+        self.state.store(prev - 1, Relaxed);
     }
-
-    fn is_shared_borrowed(&self, r: Region) -> bool {
-        self.shared_borrows.values().any(|b| b.overlaps(r))
+    pub fn mut_unborrow(&self, _: BorrowHandle) {
```
fitzgen, line 54: Should there be different handle types for mut vs shared borrows?

fitzgen (approved):

crates/wiggle/src/borrow.rs
alexcrichton: In theory yeah, but I'm not too keen to do too much refactoring here because it's all basically a dated library on life support now. These are all ignored anyway so I should probably refactor to remove them entirely, but I'll leave that as a future TODO if necessary

PR #8278, opened 2024-04-01
Record cargo-vet violations for older zstd versions

jameysharp:
When I tried to audit our previous exemption for zstd, I found two buffer overruns which were reachable from safe Rust, although not reachable from Wasmtime. I got them fixed upstream but didn't update our cargo-vet audits to reflect the issue with the older versions.

Alex updated our dependencies to pull in the fixed versions in #7870, and this PR notes for the benefit of anyone importing the Bytecode Alliance audit set that older versions should not be used.

See https://github.com/gyscos/zstd-rs/pull/231

### DIFF
--- a/supply-chain/audits.toml
+++ b/supply-chain/audits.toml
@@ -3223,13 +3223,9 @@ version = "1.2.0"
 notes = "This crate contains `unsafe` calls to libc `extattr_*` functions as one would expect from the crate's purpose."
 
 [[audits.zstd]]
-who = "Alex Crichton <alex@alexcrichton.com>"
+who = "Jamey Sharp <jsharp@fastly.com>"
 criteria = "safe-to-deploy"
-delta = "0.11.1+zstd.1.5.2 -> 0.13.0"
-notes = """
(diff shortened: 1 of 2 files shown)


jameysharp:
I'm confused by the CI failure here. I ran `cargo vet` locally and, although it complained about the new `wasmtime-slab` crate not having an `audit-as-crates-io` entry, it didn't complain about my actual changes. However I do get the same errors as in CI if I used `cargo vet --locked`; maybe the audit-as-crates-io thing masked the issue.

alexcrichton (approved):
I personally get pretty confused with the interfactions of our configuration and new crates we add. I think it's an unfortunate interaction by how when a new crate is added it's not published on crates.io but then later on once it gets published we need new configuration, but only after it's later been published. Or... something like that? I've never bottomed it out to fully understand what's happening here, I tend to just throw things at the wall and see what sticks.

alexcrichton:
You may need to remove `~/.cache/cargo-vet`, I just did that to resolve a difference I was seeing on CI as well

PR #8279, opened 2024-04-01
Support dirs/env vars in `wasmtime serve`

alexcrichton:
* More flags like `--dir` and `--env` are moved into `RunCommon` to be shared between `wasmtime serve` and `wasmtime run`, meaning that the `serve` command can now configure environment variables.

* A small test has been added as well as infrastructure for running tests with `wasmtime serve` itself. Previously there were no tests that executed `wasmtime serve`.

* The `test_programs` crate had a small refactoring to avoid double-generation of http bindings.

### DIFF
--- a/crates/test-programs/src/bin/api_proxy.rs
+++ b/crates/test-programs/src/bin/api_proxy.rs
@@ -1,18 +1,12 @@
-pub mod bindings {
-    wit_bindgen::generate!({
-        path: "../wasi-http/wit",
-        world: "wasi:http/proxy",
-        default_bindings_module: "bindings",
-    });
-}
-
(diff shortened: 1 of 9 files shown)


fitzgen:
Did we resolve the discussion in https://github.com/bytecodealliance/wasmtime/issues/8086 about whether we actually want to do this kind of thing or not?

alexcrichton:
As https://github.com/bytecodealliance/wasmtime/issues/8086 is just about the default. This PR doesn't change the defaults, it just enables `-Scli` to be somewhat useful on `wasmtime serve` as previously you couldn't configure environment variables, preopens, etc.

fitzgen:
Ah okay, gotcha. Taking a look now.

fitzgen (approved):
Nice!

PR #8283, opened 2024-04-02
Fix a panic using tables with the wrong type

### LINKED ISSUE #8281: Panic in `table.init`
### Test Case

```
;; table-init.wat
(module
  (table 0 0 externref)
  (func (export "table-init")
    (i32.const 0)
    (i32.const 0)
    (i32.const 0)
    (table.init 0 0)
  )
  (elem declare externref)
)
```

### Steps to Reproduce

```
wasmtime --invoke table-init table-init.wat
```

### Expected Results

terminate normally

### Actual Results

```
thread 'main' panicked at crates/runtime/src/table.rs:539:9:
assertion `left == right` failed
  left: Extern
 right: Func
zsh: abort      wasmtime --invoke table-init table-init.wat
```

### Versions and Environment

Wasmtime version or commit: 19.0.0

Operating system: macOS Ventura 13.6.6

Architecture: x86_64

alexcrichton:
This commit fixes an accidental issue introduced in #8018 where using an element segment which had been dropped with an `externref` table would cause a panic. The panic happened due to an assertion that tables are being used with the right type of item and that was being mismatched. The underlying issue was that dropped element segments are modeled as an empty element segment but the empty element segment was using the "functions" encoding as opposed to the "expressions" encoding. This meant that code later assumed that due to the use of functions the table must be a table-of-functions, but this was not correct for externref-based tables.

The fix in this commit is to instead model the encoding as an "expressions" list which means that the table type is dispatched on to call the appropriate initializer.

There is no memory safety issue with this mistake as the assertion was specifically targetted at preventing memory safety. This does, however, enable any WebAssembly module to panic a host.

Closes #8281

### DIFF
--- a/crates/runtime/src/instance.rs
+++ b/crates/runtime/src/instance.rs
@@ -801,7 +801,12 @@ impl Instance {
         // disconnected from the lifetime of `self`.
         let module = self.module().clone();
 
-        let empty = TableSegmentElements::Functions(Box::new([]));
+        // NB: fall back to an expressions-based list of elements which doesn't
+        // have static type information (as opposed to `Functions`) since we
+        // don't know just yet what type the table has. The type will be be
+        // inferred in the next step within `table_init_segment`.
+        let empty = TableSegmentElements::Expressions(Box::new([]));
+
         let elements = match module.passive_elements_map.get(&elem_index) {
             Some(index) if !self.dropped_elements.contains(elem_index) => {
                 &module.passive_elements[*index]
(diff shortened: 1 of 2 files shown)


alexcrichton:
As I noted [on the issue](https://github.com/bytecodealliance/wasmtime/issues/8281) and per our [policy](https://docs.wasmtime.dev/security-what-is-considered-a-security-vulnerability.html) this is actually a security vulnerability. I'll work on getting a 19.0.1 out today with this fix.

fitzgen (approved):

PR #8293, opened 2024-04-03
cranelift: Minimize ways to manipulate instruction results

jameysharp:
In particular, remove support for detaching/attaching/appending instruction results.

The AliasAnalysis pass used detach_results, but leaked the detached ValueList; using clear_results instead is better.

The verifier's `test_printing_contextual_errors` needed to get the verifier to produce an error containing a pretty-printed instruction, and did so by appending too many results. Instead, failing to append any results gets a similar error out of the verifier, without requiring that we expose the easy-to-misuse append_result method. However, `iconst` is not a suitable instruction for this version of the test because its result type is its controlling type, so failing to create any results caused assertion failures rather than the desired verifier error. I switched to `f64const` which has a non-polymorphic type.

The DFG's `aliases` test cleared both results of an instruction and then reattached one of them. Since we have access to DFG internals in these tests, it's easier to directly manipulate the relevant ValueList than to use these unsafe methods.

The only other use of attach/append was in `make_inst_results_reusing` which decided which to use based on whether a particular result was supposed to reuse an existing value. Inlining both methods there revealed that they were nearly identical and could have most of their code factored out.

While I was looking at uses of `DataFlowGraph::results`, I also simplified replace_with_aliases a little bit.

### DIFF
--- a/cranelift/codegen/src/ir/dfg.rs
+++ b/cranelift/codegen/src/ir/dfg.rs
@@ -536,27 +536,26 @@ impl DataFlowGraph {
             "Replacing {} with itself would create a loop",
             dest_inst
         );
+
+        let dest_results = self.results[dest_inst].as_slice(&self.value_lists);
+        let src_results = self.results[src_inst].as_slice(&self.value_lists);
+
         debug_assert_eq!(
-            self.results[dest_inst].len(&self.value_lists),
-            self.results[src_inst].len(&self.value_lists),
+            dest_results.len(),
+            src_results.len(),
             "Replacing {} with {} would produce a different number of results.",
             dest_inst,
             src_inst
         );
 
-        for (&dest, &src) in self.results[dest_inst]
-            .as_slice(&self.value_lists)
-            .iter()
-            .zip(self.results[src_inst].as_slice(&self.value_lists))
-        {
-            let original = src;
+        for (&dest, &original) in dest_results.iter().zip(src_results) {
             let ty = self.value_type(original);
             debug_assert_eq!(
                 self.value_type(dest),
                 ty,
                 "Aliasing {} to {} would change its type {} to {}",
                 dest,
-                src,
+                original,
                 self.value_type(dest),
                 ty
             );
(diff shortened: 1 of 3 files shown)


cranelift/codegen/src/ir/dfg.rs
```diff
@@ -536,27 +536,26 @@ impl DataFlowGraph {
...
             src_inst
         );
 
-        for (&dest, &src) in self.results[dest_inst]
-            .as_slice(&self.value_lists)
-            .iter()
-            .zip(self.results[src_inst].as_slice(&self.value_lists))
-        {
-            let original = src;
+        for (&dest, &original) in dest_results.iter().zip(src_results) {
```
cfallin, line 551: The bound `(dest, original)` names are slightly confusing given we're iterating over `zip(dest_results, src_results)` -- can we pick either `original` or `src` and use it in both places?

cranelift/codegen/src/ir/dfg.rs
```diff
@@ -931,19 +930,26 @@ impl DataFlowGraph {
...
 
         let mut reuse = reuse.fuse();
         let result_tys: SmallVec<[_; 16]> = self.inst_result_types(inst, ctrl_typevar).collect();
         let num_results = result_tys.len();
+        debug_assert!(u16::try_from(num_results).is_ok(), "Too many result values");
 
-        for ty in result_tys {
-            if let Some(Some(v)) = reuse.next() {
+        for (expected, ty) in result_tys.into_iter().enumerate() {
+            let num = expected as u16;
```
cfallin, line 941: (I realize this is inlined code, but thoughts still follow:)

The `as u16` cast here is indeed protected by the assert above, but it's still something that catches my eye and takes a bit to understand safety thereof -- could we instead do the `try_from` here (overhead is the same in the common case of one result)?

Also, it occurs to me that we want this check even in release builds (if other limits somehow let through a many-result function, we don't want to rely on fuzzing catching a wraparound-caused miscompile before users do); so either a `try_from().expect()` or a full `assert!` is probably what we want.

cfallin (approved):
Thanks for this cleanup! A few things below but in general I'm always happy when we can shrink an API :-)

jameysharp:
Good suggestions, thanks!

PR #6850, opened 2023-08-16
cranelift: Validate `iconst` ranges

timjrd:
Add the following checks:
    
* `iconst.i8`  immediate must be within 0 .. 2^8-1
* `iconst.i16` immediate must be within 0 .. 2^16-1
* `iconst.i32` immediate must be within 0 .. 2^32-1
    
Resolves #3059

### Explain why this change is needed:

As mentioned in #3059, `iconst` currently allows any immediate within the range of an `i64`, even for `iconst.i8`, `iconst.i16` or `iconst.i32`.

### This breaks some tests!

Running `cargo test` in `/cranelift/codegen` returns successfully. I also added a few tests concerning the new checks.

Running `cargo test` in `/cranelift` returns some failures. For example:

```
1: - inst1 (v1 = iconst.i8 -1): constant immediate is out of bounds
[...]
failures:
    bugpoint::tests::test_reduce
    run::test::nop
```

Which is expected if we forbid negative immediates, @jameysharp could you please confirm?

Running `cargo test` at the root of this repo returns a lot of failures, possibly related to negative immediates.

#### This is my first contribution so this patch is probably broken.

PR #8299, opened 2024-04-04
cranelift/x64: Rely on zero-extended constants

jameysharp:
Since #6850, we've been able to rely on `iconst` instructions having their immediate operands' high bits zeroed before lowering.

So a couple of places in `x64/lower.rs` can be expressed more simply now as a result.

Out of an abundance of caution, I added a debug-assertion when constants are looked up during lowering, to check that earlier phases really did ensure the high bits are zero.

I also got rid of an `expect` where a simple pattern-match will do.

### DIFF
--- a/cranelift/codegen/src/isa/x64/lower.rs
+++ b/cranelift/codegen/src/isa/x64/lower.rs
@@ -4,7 +4,7 @@
 pub(super) mod isle;
 
 use crate::ir::pcc::{FactContext, PccResult};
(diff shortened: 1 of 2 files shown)


jameysharp:
Turns out I can't use `Option::inspect` yet because it was stabilized in 1.76 and our MSRV is currently 1.75, whoops.

cfallin (approved):
LGTM, thanks!

PR #8308, opened 2024-04-05
cranelift/aarch64: Remove types from address-modes

jameysharp:
Previously, the precondition on every use of AMode::*Offset was that the type specified in the address mode must be the same size as the access type of the memory instruction where the address mode was used.

However, at the point where we replace these magic address modes with real modes that the target architecture understands, we know what instruction is using the address, so we can get the type from there.

This simplifies constructing addresses while also ensuring that the types are always correct.

The changes here center around threading the access type through `mem_finalize` and `AMode::pretty_print`, then deleting it from all `AMode` variants.

In the process I introduced a helper method on `Inst` to centralize the mapping from instruction to access type, and used that everywhere.

### DIFF
--- a/cranelift/codegen/src/isa/aarch64/abi.rs
+++ b/cranelift/codegen/src/isa/aarch64/abi.rs
@@ -36,9 +36,9 @@ impl Into<AMode> for StackAMode {
     fn into(self) -> AMode {
         match self {
             // Argument area begins after saved frame pointer + return address.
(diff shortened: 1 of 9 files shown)


cfallin (approved):
This seems quite sensible -- thanks! I think the original thought I had had was to keep the `AMode` self-describing as much as possible but this PR shows we have the type implicitly from context anywhere we need it and so it's nicer not to have a redundant encoding. So, very good cleanup.

PR #8310, opened 2024-04-05
Disable traps by default in `bindgen!` imports

alexcrichton:
By default previously all return types were wrapped in `wasmtime::Result<T>` to enable any import to return a trap to the wasm guest. This is a fair bit of boilerplate, however, and it's easy to accidentally turn a normal error into a trap. This is additionally somewhat of a power-user method as many imports probably don't end up wanting to trap.

This commit adds a new configuration option to `bindgen!`: `trappable_imports`, and switches the default to `false`. The previous behavior can be recovered with `trappable_imports: true`.

### DIFF
--- a/crates/component-macro/src/bindgen.rs
+++ b/crates/component-macro/src/bindgen.rs
@@ -7,7 +7,7 @@ use std::sync::atomic::{AtomicUsize, Ordering::Relaxed};
 use syn::parse::{Error, Parse, ParseStream, Result};
 use syn::punctuated::Punctuated;
 use syn::{braced, token, Token};
-use wasmtime_wit_bindgen::{AsyncConfig, Opts, Ownership, TrappableError};
+use wasmtime_wit_bindgen::{AsyncConfig, Opts, Ownership, TrappableError, TrappableImports};
 use wit_parser::{PackageId, Resolve, UnresolvedPackage, WorldId};
 
 pub struct Config {
@@ -107,6 +107,7 @@ impl Parse for Config {
                         opts.async_ = val;
                     }
                     Opt::TrappableErrorType(val) => opts.trappable_error_type = val,
+                    Opt::TrappableImports(val) => opts.trappable_imports = val,
                     Opt::Ownership(val) => opts.ownership = val,
                     Opt::Interfaces(s) => {
                         if inline.is_some() {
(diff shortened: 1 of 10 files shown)


elliottt (approved):
Nice!

dearfl:
Hey guys, I'm relatively new to wasm/wasmtime/wasi, can you provide some per function trappable_imports example? Also I'd like to ask how should one handle errors in a Host impl?

Suppose I have some wit definition like this:

```wit
resource example {
    hello: func() -> result<string, error-type>,
}
```

Host side impl in Rust with trappable-imports it's like this

```Rust
impl HostExample {
    fn hello(&mut self, self_: Resource<Example>) -> wasmtime::Result<Result<String, ErrorType>> {
        // should I return a trap when resource table operations failed in here like this?
        let ext: &mut Example = self.table.get_mut(&self_)?;
        let ret: String = match ext.hello() {
            Ok(s) => s,
            Err(err) => return Ok(Err(err)),
        };
        Ok(Ok(s))
    }

    fn drop(&mut self, self_: Resource<Example>) -> wasmtime::Result<()> {
        // same here
        self.table.delete(self_)?;
        Ok(())
    }
}
```

alexcrichton:
Looks like you might have already found your answer @dearfl but I can try to fill in any possible gaps and/or confirm:

* Whether something should be a trap or a WIT-level error depends on whether the guest should be able to recover from it. Guests can't recover from traps but can recover from WIT-level errors.
* Table-related errors are best represented as traps since that indicates a bug in the host or resource limits being hit from the guest if an error is returned.
* You might be interested in the `trappable_error_type` configuration key to have a single `Result<T,E>` instead of `Result<Result<T, E>>`, but that's not required.

In any case the code you have gisted above looks reasonable to me!

dearfl:
> Looks like you might have already found your answer @dearfl but I can try to fill in any possible gaps and/or confirm:
> 
>     * Whether something should be a trap or a WIT-level error depends on whether the guest should be able to recover from it. Guests can't recover from traps but can recover from WIT-level errors.
> 
>     * Table-related errors are best represented as traps since that indicates a bug in the host or resource limits being hit from the guest if an error is returned.
> 
>     * You might be interested in the `trappable_error_type` configuration key to have a single `Result<T,E>` instead of `Result<Result<T, E>>`, but that's not required.
> 
> 
> In any case the code you have gisted above looks reasonable to me!

Thank you for your detailed reply.

PR #8311, opened 2024-04-05
cranelift/riscv64: Remove unused type parameters from address modes

jameysharp:
This backend copied the general style of the aarch64 backend's address modes but has no use for a type field in address modes. Just delete them.

Between this and #8308, there are no types in target dependent address modes now. The next step is to delete them from the target independent `StackAMode` as well.


afonso360 (approved):
👍 LGTM! Nice cleanup.

PR #8313, opened 2024-04-08
NaN-canonicalization without branching on x64

adambratschikaye:
Modify the cranelift pass that performs NaN-canonicalization to avoid branches on x64. The current implementation uses two branches:
```
       8: be 00 00 c0 7f               	mov	esi, 0x7fc00000
       d: c5 f9 6e de                  	vmovd	xmm3, esi
      11: 0f 2e c0                     	ucomiss	xmm0, xmm0
      14: 0f 8b 04 00 00 00            	jnp	0x1e <wasm[0]::function[0]+0x1e>
      1a: f2 0f 10 c3                  	movsd	xmm0, xmm3              # xmm0 = xmm3[0],xmm0[1]
      1e: 0f 84 04 00 00 00            	je	0x28 <wasm[0]::function[0]+0x28>
      24: f2 0f 10 c3                  	movsd	xmm0, xmm3              # xmm0 = xmm3[0],xmm0[1]
```
With these changes, NaN-canonicalization becomes:
```
       8: c5 e8 c2 da 03               	vcmpunordps	xmm3, xmm2, xmm2
       d: be 00 00 c0 7f               	mov	esi, 0x7fc00000
      12: c5 f9 6e e6                  	vmovd	xmm4, esi
      16: c4 e3 69 4c c4 30            	vpblendvb	xmm0, xmm2, xmm4, xmm3
```

Running both versions against an small image classification inference benchmark [here](https://github.com/adambratschikaye/image-classification-wasmtime-bench) resulted in a ~50% improvement:
```
image_classification/opt-level=0
                        time:   [728.16 ms 730.00 ms 732.05 ms]
                        change: [-44.476% -44.251% -44.029%] (p = 0.00 < 0.05)
                        Performance has improved.

image_classification/opt-level=1
                        time:   [593.90 ms 595.51 ms 597.34 ms]
                        change: [-51.561% -51.396% -51.211%] (p = 0.00 < 0.05)
                        Performance has improved.
```

As a side note, I didn't notice any sightglass benchmark that was performing mainly float arithmetic to test against. I'd be happy to add this image classification case if there's interest.

### DIFF
--- a/cranelift/codegen/src/nan_canonicalization.rs
+++ b/cranelift/codegen/src/nan_canonicalization.rs
@@ -15,13 +15,13 @@ static CANON_32BIT_NAN: u32 = 0b01111111110000000000000000000000;
 static CANON_64BIT_NAN: u64 = 0b0111111111111000000000000000000000000000000000000000000000000000;
 
 /// Perform the NaN canonicalization pass.
-pub fn do_nan_canonicalization(func: &mut Function) {
+pub fn do_nan_canonicalization(func: &mut Function, has_vector_support: bool) {
     let _tt = timing::canonicalize_nans();
     let mut pos = FuncCursor::new(func);
     while let Some(_block) = pos.next_block() {
         while let Some(inst) = pos.next_inst() {
             if is_fp_arith(&mut pos, inst) {
-                add_nan_canon_seq(&mut pos, inst);
+                add_nan_canon_seq(&mut pos, inst, has_vector_support);
             }
         }
     }
@@ -57,7 +57,7 @@ fn is_fp_arith(pos: &mut FuncCursor, inst: Inst) -> bool {
 }
 
 /// Append a sequence of canonicalizing instructions after the given instruction.
-fn add_nan_canon_seq(pos: &mut FuncCursor, inst: Inst) {
+fn add_nan_canon_seq(pos: &mut FuncCursor, inst: Inst, has_vector_support: bool) {
     // Select the instruction result, result type. Replace the instruction
     // result and step forward before inserting the canonicalization sequence.
     let val = pos.func.dfg.first_result(inst);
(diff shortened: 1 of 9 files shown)


adambratschikaye:
Also is there a way to enable NaN-canonicalization in a `clif` test to add a test for this?

afonso360:
You should be able to add something along these lines, to test with nan canonicalization enabled:
```
test {run,compile,etc...}
set enable_nan_canonicalization=true
target x86_64
```

cranelift/codegen/src/nan_canonicalization.rs
```diff
@@ -65,16 +65,23 @@ fn add_nan_canon_seq(pos: &mut FuncCursor, inst: Inst) {
...
-    // the canonical NaN value if `val` is NaN, assign the result to `inst`.
-    let is_nan = pos.ins().fcmp(FloatCC::NotEqual, new_res, new_res);
-
-    let scalar_select = |pos: &mut FuncCursor, canon_nan: Value| {
-        pos.ins()
-            .with_result(val)
-            .select(is_nan, canon_nan, new_res);
+    let scalar_select = |pos: &mut FuncCursor, canon_nan: Value, ty: types::Type| {
+        let canon_nan = pos.ins().scalar_to_vector(ty, canon_nan);
+        let new_res = pos.ins().scalar_to_vector(ty, new_res);
```
afonso360, line 70: I'm not sure this works with the RISC-V backend, our vector support is gated on a feature flag which may not always be available.

One solution would be to keep the previous code around for RISC-V, but it feels wrong to select on targets at this level.

adambratschikaye: Ah I see that now. What about choosing based on the targets and features (that is specifically `riscv64` when `has_v=false`)? Maybe in the future there will be other target/feature combinations that don't support vectors?

afonso360: I think that is what we need right now. I'm not sure if eventually we should add something to `TargetIsa`, but It might be worth reviewing later on if this situation ends up happening more often.

adambratschikaye: Added the fall back to the old version for RiscV [here](https://github.com/bytecodealliance/wasmtime/pull/8313/files#diff-c7a66b91ac03843c5aafe984938022ccba235c80c3fad786772964dc7b9da152R286).

afonso360 (approved):
This LGTM! Thanks! I don't know if @abrown wants to review it as well.

abrown (approved):

bjorn3:
This does the same lowering on arm64 and s390x. Does it actually benefit performance there?

adambratschikaye:
I'm not sure. At the moment I don't have a setup to test on arm or s390x.

PR #8314, opened 2024-04-08
Support `.wat` files in the `explore` command

adambratschikaye:
Many on the `wasmtime` subcommands work with either `.wasm` or `.wat` files, but `explore` only works with `.wasm`. This change accepts `.wat` as well.

### DIFF
--- a/src/commands/explore.rs
+++ b/src/commands/explore.rs
@@ -32,8 +32,10 @@ impl ExploreCommand {
 
         let config = self.common.config(self.target.as_deref())?;
 
-        let wasm = std::fs::read(&self.module)
+        let bytes = std::fs::read(&self.module)
             .with_context(|| format!("failed to read Wasm module: {}", self.module.display()))?;
+        let wasm = wat::parse_bytes(&bytes)
+            .with_context(|| format!("failed to parse Wasm module: {}", self.module.display()))?;
 
         let output = self
             .output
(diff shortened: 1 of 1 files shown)


src/commands/explore.rs
```diff
@@ -32,8 +32,10 @@ impl ExploreCommand {
 
         let config = self.common.config(self.target.as_deref())?;
 
-        let wasm = std::fs::read(&self.module)
+        let bytes = std::fs::read(&self.module)
             .with_context(|| format!("failed to read Wasm module: {}", self.module.display()))?;
+        let wasm = wat::parse_bytes(&bytes)
+            .with_context(|| format!("failed to parse Wasm module: {}", self.module.display()))?;
```
fitzgen, line 38: Can you make sure that this also compiles and works for wasm binaries when the `wat` dependency isn't enabled? Something like the following:

```rust
                #[cfg(feature = "wat")]
                let bytes = wat::parse_bytes(bytes).map_err(|mut e| {
                    e.set_path(path);
                    e
                })?;
```

fitzgen (commented):
Thanks! One nitpick below before we merge this.

src/commands/explore.rs
adambratschikaye: Sorry I forgot `wat` was behind a feature. Fixed it now.

fitzgen (approved):
Thanks very much!

PR #8316, opened 2024-04-08
cranelift: Simplify StackAMode variants

jameysharp:
Instead of describing how these address modes might be implemented (using a stack or frame pointer, say), describe what they represent. Also, remove the unused type fields from all variants and from `gen_get_stack_addr`.

The three kinds of stack addresses that we currently need to generate in a target independent way refer to either caller or callee argument areas, or refer to the spill slots and explicit stack slots within the frame.

It is a target-specific implementation detail whether the incoming argument area is indexed relative to the frame pointer, or whether the stack slots are located by tracking a "nominal" stack pointer offset.

So let's not muddy up the target-independent code by using names that refer to these target-specific concepts.

### DIFF
--- a/cranelift/codegen/src/isa/aarch64/abi.rs
+++ b/cranelift/codegen/src/isa/aarch64/abi.rs
@@ -36,9 +36,9 @@ impl Into<AMode> for StackAMode {
     fn into(self) -> AMode {
         match self {
             // Argument area begins after saved frame pointer + return address.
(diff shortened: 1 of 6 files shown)


elliottt (approved):
Nice!

jameysharp:
Looks to me like maybe a flaky test in the preview2 test suite, so I'm going to try re-submitting it and see if it goes through this time.

PR #8320, opened 2024-04-10
winch: Refactor the MacroAssembler trait

saulecabrera:
No new functionality is introduced as part of this change. 

This commit introduces a small refactoring of Winch's `MacroAssembler` trait by removing multiple methods that can be treated in an ISA-indepedent way by relying on the base building blocks already provided by the `MacroAssembler`. This change simplifies the process of adding support for other backends to Winch.


elliottt (approved):
This is great!

PR #8326, opened 2024-04-10
wasmtime-c-api: Add support for GC references to the `wasm_*` C API functions

fitzgen:
Does not add support for the `wasmtime_*` functions yet.

### DIFF
--- a/crates/c-api/src/store.rs
+++ b/crates/c-api/src/store.rs
@@ -7,6 +7,11 @@ use wasmtime::{
     UpdateDeadline, Val,
 };
 
+pub type WasmStoreData = Option<StoreRef>;
+pub type WasmStore = Store<WasmStoreData>;
+pub type WasmStoreContext<'a> = StoreContext<'a, WasmStoreData>;
+pub type WasmStoreContextMut<'a> = StoreContextMut<'a, WasmStoreData>;
+
 /// This representation of a `Store` is used to implement the `wasm.h` API.
 ///
 /// This is stored alongside `Func` and such for `wasm.h` so each object is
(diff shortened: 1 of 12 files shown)


crates/c-api/src/store.rs
```diff
@@ -60,16 +76,17 @@ pub extern "C" fn wasm_store_new(engine: &wasm_engine_t) -> Box<wasm_store_t> {
...
-    pub(crate) store: Store<StoreData>,
+    pub(crate) store: Store<WasmtimeStoreData>,
 }
 
 wasmtime_c_api_macros::declare_own!(wasmtime_store_t);
 
-pub type CStoreContext<'a> = StoreContext<'a, StoreData>;
-pub type CStoreContextMut<'a> = StoreContextMut<'a, StoreData>;
+pub type WasmtimeStoreContext<'a> = StoreContext<'a, WasmtimeStoreData>;
+pub type WasmtimeStoreContextMut<'a> = StoreContextMut<'a, WasmtimeStoreData>;
```
alexcrichton, line 86: Could these renamings perhaps be split out either commit-wise or PR-wise? It's a bit difficult to disentangle what's going on here with simultaneous renamings.

alexcrichton (commented):
Personally I'm feeling pretty uneasy about this having a self-referencing store. This is exclusively for the `wasm_*` API which none of our `wasmtime-*` bindings are using as well. Given that my gut would actually be the opposite of what you're thinking which is to basically stop supporting reference types in the `wasm_*` APIs and only support them in `wasmtime_*` APIs. The `wasm.h` header feels "too incompatible" with our implementation, to the extent that I'm pretty worried about the self-references and `unsafe` and bugs here. Given the fact that we'll have a full implementation available with `wasmtime_*` I'd be tempted to tailor those to what we need and only have that for now.

crates/c-api/src/store.rs
fitzgen: Split out the renaming to https://github.com/bytecodealliance/wasmtime/pull/8344

fitzgen:
> Personally I'm feeling pretty uneasy about this having a self-referencing store. This is exclusively for the `wasm_*` API which none of our `wasmtime-*` bindings are using as well. Given that my gut would actually be the opposite of what you're thinking which is to basically stop supporting reference types in the `wasm_*` APIs and only support them in `wasmtime_*` APIs. The `wasm.h` header feels "too incompatible" with our implementation, to the extent that I'm pretty worried about the self-references and `unsafe` and bugs here. Given the fact that we'll have a full implementation available with `wasmtime_*` I'd be tempted to tailor those to what we need and only have that for now.

Yeah, that's fair. I have a branch with `wasmtime_*` working, and it is indeed a lot cleaner than `wasm_*` so I'll close this PR in favor of https://github.com/bytecodealliance/wasmtime/pull/8344 and then open another new PR on top of that one with the `wasmtime_*` APIs.

PR #8327, opened 2024-04-11
Eagerly reserve tail call stack arg space

elliottt:
`return_call` instructions reuse the incoming argument area of the caller's frame. As such if the caller's incoming argument area is  not exactly the right size for the callee, some resizing will need to take place to ensure that the return address, frame pointer, clobbers, and stack slots don't get overwritten. The current solution on the main branch for the x64 backend is to explicitly resize the frame via `ShrinkArgumentArea` or `GrowArgumentArea` right before the `return_call` arguments are written to the stack, ensuring that there is sufficient space. While this does work, it does make a `return_call` more expensive when the resizing is necessary.

To simplify this, we instead resize the incoming argument area in the function prologue to accommodate the largest possible argument area of any `return_call` instruction in the function. We then shrink back down when necessary before an individual `return_call`. This simplifies the implementation of tail calls on x86_64, as we no longer need to move the entire frame, just the return address before we jump to the tail-callee.

### DIFF
--- a/cranelift/codegen/src/isa/x64/abi.rs
+++ b/cranelift/codegen/src/isa/x64/abi.rs
@@ -583,7 +583,7 @@ impl ABIMachineSpec for X64ABIMachineSpec {
     ) -> SmallInstVec<Self::I> {
         // Emit return instruction.
         let stack_bytes_to_pop = if call_conv == isa::CallConv::Tail {
-            frame_layout.stack_args_size
+            frame_layout.tail_args_size
         } else {
             0
         };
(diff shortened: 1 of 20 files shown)


cranelift/codegen/src/isa/x64/inst/args.rs
```diff
@@ -544,6 +558,16 @@ impl SyntheticAmode {
...
+            SyntheticAmode::IncomingArg {
+                offset,
+                stack_args_size,
+            } => {
+                // NOTE: this could be made relatie to RSP by adding additional offsets from the
+                // frame_layout.
+                let diff = state.frame_layout().stack_args_size - stack_args_size
+                    + state.frame_layout().setup_area_size;
+                Amode::imm_reg(i32::try_from(offset + diff).unwrap(), regs::rbp())
+            }
```
jameysharp, line 570: If we switch `IncomingArg` to holding offsets from the highest address in the incoming argument area, instead of the lowest address, then we don't need to also store the size of the incoming argument area with it. Finalizing such an amode then amounts to finding the highest address in the stack frame (e.g. `rbp` plus the `setup_area_size` and `stack_args_size` from the frame layout), then subtracting the offset that's in `IncomingArg`.

cranelift/codegen/src/isa/x64/abi.rs
```diff
@@ -640,6 +640,50 @@ impl ABIMachineSpec for X64ABIMachineSpec {
     ) -> SmallVec<[Self::I; 16]> {
         let mut insts = SmallVec::new();
 
+        // When a return_call within this function required more stack arguments than we have
+        // present, resize the incoming argument area of the frame to accommodate those arguments.
+        let incoming_args_diff = frame_layout.tail_args_size - frame_layout.incoming_args_size;
```
elliottt, line 645: We may need to add this to `offset_upward_to_caller_sp` in the unwind info below.

cranelift/filetests/filetests/isa/x64/return-call-indirect.clif
```diff
@@ -248,6 +248,12 @@ block0:
 ;   pushq   %rbp
 ;   movq    %rsp, %rbp
 ;   subq    %rsp, $160, %rsp
+;   movq    %rsp, %rbp
+;   movq    160(%rsp), %r11
+;   movq    %r11, 0(%rsp)
+;   movq    168(%rsp), %r11
+;   movq    %r11, 8(%rsp)
+;   subq    %rsp, $160, %rsp
```
fitzgen, line 256: Definitely not a blocker for landing this PR, but it would be nice if we could get this to be a single `sub $rsp, M+N` followed by `mov temp, [$rsp + OLD_RA_OFFSET]; mov [$rsp + NEW_RA_OFFSET]` and similar for the saved FP, instead of an initial sub to establish the frame followed by a second sub for tail call arguments space.

fitzgen (approved):
Nice!

cranelift/filetests/filetests/isa/x64/return-call-indirect.clif
elliottt: We're experimenting with that on a different branch based on this PR, as a follow-up. Moving the stack check before the frame setup means that we can cut out even more instructions, as there's no need to move the frame pointer as well.

PR #8332, opened 2024-04-11
wasmtime-wasi-http collection of small refactors

rylev:
This is a collection of refactors of `wasmtime-wasi-http` in an attempt to make the API slightly easier to use and more documented.

The various refactorings are all small enough that it seems appropriate to batch them into one PR. Each individual commit should be reviewable as a standalone refactor and any of them should be able to be dropped from the PR should there not be consensus that the refactor is appropriate.


rylev:
CI is failing, but I'm going to wait for an initial review before proceeding with fixing it.

alexcrichton (approved):
Looks like nice cleanups to me 👍 

Thanks!

PR #8343, opened 2024-04-11
Don't gate on `cargo vet` for PRs

alexcrichton:
Gate the `cargo vet` step for the merge queue by default but for PRs don't fail the PR's CI and instead allow entering the queue with a failing `cargo vet` entry. This should help make it easier to land a `cargo vet` entry on `main` outside of a PR while not requiring the PR is rebased.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -118,7 +118,7 @@ jobs:
     - uses: ./.github/actions/install-cargo-vet
     - id: vet
       run: cargo vet --locked
-      continue-on-error: ${{ github.event_name != 'pull_request' }}
+      continue-on-error: ${{ github.event_name == 'pull_request' }}
 
     # common logic to cancel the entire run if this job fails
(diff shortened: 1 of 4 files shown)


alexcrichton:
Ok I think this is where it's now working. Summary of changes:

* Example failure: d7f3e533fc4a45cfa30bc4eac1fc6ec3ca98c9c3
* Example success: 81e1e8d2bdc36bd7aafbced42661c4f41d920c07

See the "Cargo vet failed on a Pull Request" run.

My thinking here is that with this new setup PRs will have a "Red X" on a job indicating failure but they won't be blocked from entering the merge queue. The job with the "Red X" just kinda sits there and the merge queue doesn't even listen for it. This is done by allowing `cargo vet` to fail, but only on PRs. The PR-specific dependency job is then what triggers the "Red X" to give a visual indication to reviewers and authors.

fitzgen (approved):
Thanks!

PR #8346, opened 2024-04-12
`wasmtime-c-api`: Add support for GC references in `wasmtime.h` APIs

fitzgen:
Builds on top of https://github.com/bytecodealliance/wasmtime/pull/8344

-------

Restores support for `externref` in `wasmtime_val_t`, methods for manipulating them and getting their wrapped host data, and examples/tests for these things.

Additionally adds support for `anyref` in `wasmtime_val_t`, clone/delete methods similar to those for `externref`, and a few `i31ref`-specific methods. Also adds C and Rust example / test for working with `anyref`.

### DIFF
--- a/crates/c-api/src/ref.rs
+++ b/crates/c-api/src/ref.rs
@@ -1,6 +1,6 @@
-use crate::abort;
-use std::os::raw::c_void;
-use wasmtime::{Ref, Val};
+use crate::{abort, WasmtimeStoreContextMut};
+use std::{mem::MaybeUninit, os::raw::c_void, ptr};
+use wasmtime::{AnyRef, ExternRef, ManuallyRooted, Ref, RootScope, Val, I31};
 
 /// `*mut wasm_ref_t` is a reference type (`externref` or `funcref`), as seen by
 /// the C API. Because we do not have a uniform representation for `funcref`s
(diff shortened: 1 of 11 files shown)


crates/c-api/src/ref.rs
```diff
@@ -181,3 +185,158 @@ wasmtime_c_api_macros::declare_ref!(wasm_foreign_t);
 pub extern "C" fn wasm_foreign_new(_store: &crate::wasm_store_t) -> Box<wasm_foreign_t> {
     abort("wasm_foreign_new")
 }
+
+pub type wasmtime_anyref_t = ManuallyDrop<ManuallyRooted<AnyRef>>;
+pub type wasmtime_externref_t = ManuallyDrop<ManuallyRooted<ExternRef>>;
```
alexcrichton, line 190: Mind adding a comment here as to what the `ManuallyDrop` is doing? It seems related to the destructor of `wasmtime_val_t` but I couldn't quite figure it out

alexcrichton (approved):

crates/c-api/src/ref.rs
fitzgen: This is a leftover from the old externref support, and I don't think we actually need it anymore. I'll remove it.

PR #8347, opened 2024-04-12
`wasmtime-wasi-http` docs

rylev:
This documents almost all of `wasmtime-wasi-http` (excluding raw bindings) and adds `#![deny(missing_docs)]` to the crate. Along with documentation changes there are a few small changes in the same vain as #8332. 

It's probably easiest to go commit by commit to see each individual change.


crates/wasi-http/src/body.rs
```diff
@@ -1,3 +1,5 @@
+//! Implmentation of the `wasi:http/types` interface's various body types.
```
FrankReh, line 1: Implementation

alexcrichton (approved):
Thank you very much for this!

PR #8350, opened 2024-04-12
Update the `wasm-tools` family of crates to 0.203.0

fitzgen:

### DIFF
--- a/crates/wasmtime/src/engine/serialization.rs
+++ b/crates/wasmtime/src/engine/serialization.rs
@@ -215,6 +215,8 @@ impl Metadata<'_> {
             memory_control,
             function_references,
             gc,
+            custom_page_sizes,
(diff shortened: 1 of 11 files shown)


crates/wasmtime/src/engine/serialization.rs
```diff
@@ -218,6 +218,10 @@ impl Metadata<'_> {
             component_model_values,
             component_model_nested_names,
 
+            // Not yet implemented; we don't have knobs for these.
+            custom_page_sizes: _,
+            shared_everything_threads: _,
```
alexcrichton, line 223: Mind switching these to `assert!(!custom_page_sizes)` like a few other unimplemented proposals below?

(that way this is caught once we add knobs in the future)

alexcrichton (approved):

PR #8352, opened 2024-04-12
Move c-api testing into a separate build job

elliottt:
We're less constrained for action runners these days, so let's build the c-api tests in parallel with other tests. This shortens the long-pole for tests that are required for a merge to main, as well as ci time for PRs that touch the c-api.

One effect of this change is that we're no longer building the c-api changes when testing on Qemu, but we are still testing on linux, windows, and macos.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -482,6 +482,38 @@ jobs:
       env:
         GH_TOKEN: ${{ github.token }}
 
+  # Perform all tests of the c-api
+  test_capi:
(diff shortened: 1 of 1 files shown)


alexcrichton:
I'd recommend adding `prtest:full` to make sure this passes CI (and/or a dummy commit to modify the C API to ensure that the filter still works).

Also can you add the new job as a dependency of the "join node" at the bottom of the workflow?

alexcrichton (approved):

PR #8354, opened 2024-04-12
Use an incremental cache when testing WASI

alexcrichton:
Currently we've got a good number of WASI tests and they're all relatively large. We also can run a single test in up to three configurations:

* As-is with a module
* As a component in "sync" mode
* As a component in "async" mode

In debug mode compilation of all these modules can take a significant chunk of time (20-30s in total for test suites) This commit updates these test suites to use an in-memory per-process incremental cache backed by a simple `Mutex<HashMap>`. This gives some good speedups in debug mode, locally the wasi-common, wasmtime-wasi, and wasmtime-wasi-http test suites were reduced from 32 to 17 seconds. I'd expect larger speedups on less-parallel machines such as our CI.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2755,6 +2755,7 @@ version = "0.0.0"
 dependencies = [
  "cargo_metadata",
  "heck",
+ "wasmtime",
  "wit-component 0.203.0",
 ]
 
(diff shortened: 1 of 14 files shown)


elliottt (approved):
Looks great!

PR #8355, opened 2024-04-13
Fix calculation of gc refs in functions

alexcrichton:
In addition to excluding i31 also exclude funcrefs.

### DIFF
--- a/crates/types/src/lib.rs
+++ b/crates/types/src/lib.rs
@@ -371,24 +371,8 @@ impl TypeTrace for WasmFuncType {
 impl WasmFuncType {
     #[inline]
     pub fn new(params: Box<[WasmValType]>, returns: Box<[WasmValType]>) -> Self {
-        let non_i31_gc_ref_params_count = params
-            .iter()
-            .filter(|p| match **p {
-                WasmValType::Ref(rt) => {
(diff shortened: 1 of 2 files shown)


alexcrichton:
This was found updating https://github.com/bytecodealliance/wasmtime-cpp to the latest Wasmtime where a test was failing in Rust erroneously for using funcref which caused an error to be returned from a `Store`.

crates/types/src/lib.rs
```diff
@@ -374,14 +374,18 @@ impl WasmFuncType {
...
             .count();
         let non_i31_gc_ref_returns_count = returns
             .iter()
             .filter(|r| match **r {
-                WasmValType::Ref(rt) => rt.heap_type != WasmHeapType::I31,
+                WasmValType::Ref(rt) => {
+                    rt.heap_type != WasmHeapType::I31 && rt.heap_type != WasmHeapType::Func
+                }
                 _ => false,
             })
```
fitzgen, line 390: We can reuse [this helper](https://github.com/bytecodealliance/wasmtime/blob/2a6b13295542fd00c1628880619509a6bc1c422f/crates/types/src/lib.rs#L123) here:

```suggestion
            .filter(|r| r.is_gc_heap_type())
```

(and above)

fitzgen (approved):
r=me with suggestion below

Issue #8322, opened 2024-04-10
Panic in `wasmtime::component::Component::from_binary` with a truncated WASM component file


maxbrunsfeld:
### Test Case

Wasm file: [truncated.wasm.zip](https://github.com/bytecodealliance/wasmtime/files/14935987/truncated.wasm.zip)

### Steps to Reproduce

```rust
const TRUNCATED_WASM: &[u8] = include_bytes!("truncated.wasm");

async fn test() {
    let mut config = wasmtime::Config::new();
    config.wasm_component_model(true);
    config.async_support(true);

    let engine = wasmtime::Engine::new(&config).unwrap();

    wasmtime::component::Component::from_binary(&engine, TRUNCATED_WASM).await;
}
```

### Expected Results

The `Component::from_binary` call should return an `Err` value.

### Actual Results

```
Thread "main" panicked with "range end index 267037 out of range for slice of length 253952" at ~.cargo/registry/src/index.crates.io-6f17d22bba15001f/wasmtime-environ-19.0.0/src/component/translate.rs:537:46

   5: core::panicking::panic_fmt
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/core/src/panicking.rs:72:14
   6: core::slice::index::slice_end_index_len_fail_rt
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/core/src/slice/index.rs:76:5
      core::slice::index::slice_end_index_len_fail
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/core/src/slice/index.rs:68:9
   7: <core::ops::range::Range<usize> as core::slice::index::SliceIndex<[T]>>::index
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/core/src/slice/index.rs:394:13
      core::slice::index::<impl core::ops::index::Index<I> for [T]>::index
             at /rustc/25ef9e3d85d934b27d9dada2f9dd52b1dc63bb04/library/core/src/slice/index.rs:18:9
      wasmtime_environ::component::translate::Translator::translate_payload
             at /home/luke/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wasmtime-environ-19.0.0/src/component/translate.rs:537:46
   8: wasmtime_environ::component::translate::Translator::translate
[...]

PR #8357, opened 2024-04-13
Forward-port some release notes to main

alexcrichton:
Keeping `RELEASES.md` up-to-date ideally

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -29,14 +29,14 @@ Unreleased.
 Released 2024-04-11.
 
 * Fix a panic when compiling invalid components.
-  [#8322](https://github.com/bytecodealliance/wasmtime/issues/8322)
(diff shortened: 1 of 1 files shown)


jameysharp (approved):
This does make `main` match the corresponding patch release notes. Those patch release notes were a bit off though and maybe we should at least fix them on main?

- #8323 was a bugfix, not #8322 which was the issue that it fixed, and it went out in 17.0.3 as well as the others but only made it into the release notes for v18 and v19.
- #8334 is cited as included in all three releases but that's the v19 backport PR for #8301, which I assume we should cite instead.

I don't feel strongly though and would be happy to see this merge as-is.

alexcrichton:
Thanks for the close review! Makes sense to me to apply all those changes so I've gone ahead and done so.

alexcrichton:
That spurious failure is fixed in https://github.com/bytecodealliance/wasmtime/pull/8410

PR #8360, opened 2024-04-13
fuzzgen: Use the correct ISA when running NaN Canonicalization pass

### LINKED ISSUE #8359: riscv64: Panic on partial gen_extractlane rule
This input:

```clif
;; out.clif
test compile
target riscv64

function u1:0(i128, f32, f64) -> f64 fast {
block0(v0: i128, v1: f32, v2: f64):
    v9 = fpromote.f64 v1
    v10 = f64const +NaN
    v11 = scalar_to_vector.f64x2 v10  ; v10 = +NaN
    v12 = scalar_to_vector.f64x2 v9
    v13 = fcmp uno v12, v12
    v14 = bitcast.f64x2 v13
    v15 = bitselect v14, v11, v12
    v8 = extractlane v15, 0
    return v8
}
```

yields:

```
$ cargo run -q test ../out.clif
thread 'worker #1' panicked at /Users/alex/code/wasmtime/target/debug/build/cranelift-codegen-41964176c4a44af8/out/isle_riscv64.rs:9760:5:
internal error: entered unreachable code: no rule matched for term gen_extractlane at src/isa/riscv64/inst_vector.isle line 1496; should it be partial?
 ERROR cranelift_filetests::concurrent > FAIL: panicked in worker #1: internal error: entered unreachable code: no rule matched for term gen_extractlane at src/isa/riscv64/inst_vector.isle line 1496; should it be partial?
FAIL ../out.clif: panicked in worker #1: internal error: entered unreachable code: no rule matched for term gen_extractlane at src/isa/riscv64/inst_vector.isle line 1496; should it be partial?
1 tests
Error: 1 failure
```

afonso360:
👋 Hey,

cranelift-fuzzgen unconditionally runs the NaN Canonicalization pass on all functions that it generates. This is so that we can ensure that when running in the interpreter vs natively we get the same bitpattern for all NaN's.

Until now we just picked a random ISA (the host ISA), disabled the verifier and ran the pass with that. This was because the ISA didn't really matter for the passes that we wanted to run.

In #8313 the ISA now drives some codegen decisions for the NaN Canonicalization pass. Namely, if the ISA supports Vectors, it tries to use that.

In #8359 a fuzz bug was reported where fuzzgen generated vector code for RISC-V without the `has_v` flag, something that should *never* happen, because we simply cannot compile that code.

It turns out that fuzzgen did not generate vector code itself. But since we were passing the host ISA to the NaN canonicalization pass, it assumed that it could use vectors and did so. But the actual target ISA did not support vectors.

To fix this, we now correctly pass the target ISA that we are building a function for.

Fixes #8359


elliottt (approved):
Looks good to me!

PR #8363, opened 2024-04-14
c-api: Fix alignment of `wasmtime_val_*`

alexcrichton:
This commit fixes an issue where `wasmtime_val_raw_t` had an incorrect
alignment. In Rust `ValRaw` contains a `u128` which has an alignment of
16 but in C the representation had a smaller alignment meaning that the
alignment of the two structures was different. This was seen to cause
alignment faults when structure were passed from C++ to Rust, for
example.

This commit changes the Rust representation of `ValRaw`'s `v128` field
to do the same as C which is to use `[u8; 16]`. This avoids the need to
raise the alignment in C which appears to be nontrivial. Cranelift is
appropriately adjusted to understand that loads/stores from `ValRaw` are
no longer aligned. Technically this only applies to the `v128` field but
it's not too bad to apply it to the other fields as well.

### DIFF
--- a/crates/c-api/include/wasmtime/val.h
+++ b/crates/c-api/include/wasmtime/val.h
@@ -7,6 +7,7 @@
 #ifndef WASMTIME_VAL_H
 #define WASMTIME_VAL_H
 
+#include <stdalign.h>
 #include <wasm.h>
 #include <wasmtime/extern.h>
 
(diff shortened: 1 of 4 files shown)


fitzgen (approved):

alexcrichton:
I ended up going the reverse direction here instead, lowering the alignment in Rust instead of raising the alignment in the C API. I don't know how to portably raise the alignment in C so I'd rather tweak the Rust side of things.

Issue #8321, opened 2024-04-10
winch(aarch64): MacroAssembler completeness


saulecabrera:
Winch's support for Aarch64 is minimal. The objective of this issue to track the missing items in order to bring the Aarch64 MacroAssembler to parity with the x64 implementation.

If you're interested in taking on any of the elements below, please add your name beside it to signal that you're actively working on it (if you don't have permission to edit this issue, feel free to reach out via Zulip).

- [x] Ensure that all spec tests are passing (@saulecabrera)
- [x] Stack checks ( `check_stack`) (@graydon)
- [x] `address_from_sp` (@vulc41n)
- [x] `address_at_sp`(@vulc41n)
- [x] `address_at_vmctx` (@MarinPostma)
- [x] `wasm_store` (@vulc41n)
- [x] `wasm_load` (@vulc41n)
- [x] `call` (@MarinPostma)
- [x] `load_ptr`
- [x] `load_addr`
- [x] `pop` (@vulc41n)
- [x] `cmov` (@vulc41n)
- [x] `checked_uadd` (@vulc41n)
- [x] `float_add` (@evacchi)
- [x] `float_sub`(@evacchi)
- [x] `float_mul`(@evacchi)
- [x] `float_div`(@evacchi)
- [x] `float_min`(@evacchi)
- [x] `float_max`(@evacchi)
- [x] `float_copysign` (@evacchi)
- [x] `float_neg` (@evacchi)
- [x] `float_abs` (@evacchi)
- [x] `float_round`(@evacchi)
- [x] `float_sqrt`(@evacchi)
- [x] `and`(@evacchi)
- [x] `or`(@evacchi)
- [x] `xor`(@evacchi)
- [x] `shift`(@evacchi)
- [x] `div` (@MarinPostma)
- [x] `rem` (@MarinPostma)
- [x] `zero` (@saulecabrera)
- [x] `popcnt` (@vulc41n)
- [x] `unsigned_truncate` (@MarinPostma)
- [x] `unsigned_truncate` (@MarinPostma)
- [x] `signed_convert` (@minirop)
- [x] `unsigned_convert` (@minirop)
- [x] `reinterpret_float_as_int` (@minirop)
- [x] `reinterpret_int_as_float` (@minirop)
- [x] `demote` (@vulc41n)
- [x] `promote` (@vulc41n)
- [x] `cmp` (@vulc41n) 
- [x] `cmp_with_set` (@vulc41n)
[...]

PR #8365, opened 2024-04-14
winch(arm64): fpu arithmetics (add, sub, mul, div, min, max) and regalloc

evacchi:
- Add some of the instructions in #8321
- Introduces preliminary support to FPU registers to regalloc
- Introduces tests for the instructions above (ported over from x64)
- Fixes a few typos.

This is a draft ~with a handful of tests for `float_add{32,64}` (LGTM but another pair of eyes would help), I will add the remaining later~. Added the remaining tests.

### DIFF
--- a/winch/codegen/src/isa/aarch64/masm.rs
+++ b/winch/codegen/src/isa/aarch64/masm.rs
@@ -14,6 +14,7 @@ use cranelift_codegen::{
     ir::{RelSourceLoc, SourceLoc},
     settings, Final, MachBufferFinalized, MachLabel,
 };
+use regalloc2::RegClass;
 use wasmtime_environ::PtrSize;
 
 /// Aarch64 MacroAssembler.
@@ -197,16 +198,24 @@ impl Masm for MacroAssembler {
                 let imm = match v {
                     I::I32(v) => v as u64,
                     I::I64(v) => v,
-                    _ => panic!(),
+                    I::F32(v) => v as u64,
+                    I::F64(v) => v,
                 };
 
                 let scratch = regs::scratch();
                 self.asm.load_constant(imm, scratch);
-                self.asm.mov_rr(scratch, rd, size);
+                match rd.class() {
+                    RegClass::Int => self.asm.mov_rr(scratch, rd, size),
+                    RegClass::Float => self.asm.mov_to_fpu(scratch, rd, size),
+                    _ => todo!(),
+                }
             }
-            (RegImm::Reg(rs), rd) => {
-                self.asm.mov_rr(rs, rd, size);
-            }
+            (RegImm::Reg(rs), rd) => match (rs.class(), rd.class()) {
+                (RegClass::Int, RegClass::Int) => self.asm.mov_rr(rs, rd, size),
+                (RegClass::Float, RegClass::Float) => self.asm.fmov64_rr(rs, rd),
+                (RegClass::Int, RegClass::Float) => self.asm.mov_to_fpu(rs, rd, size),
+                _ => todo!(),
+            },
         }
     }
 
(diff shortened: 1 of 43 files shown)


winch/codegen/src/isa/aarch64/masm.rs
```diff
@@ -197,16 +198,24 @@ impl Masm for MacroAssembler {
...
-                self.asm.mov_rr(rs, rd, size);
+                match rd.class() {
+                    RegClass::Int => self.asm.mov_rr(scratch, rd, size),
+                    RegClass::Float => self.asm.mov_to_fpu(scratch, rd, size),
+                    _ => todo!(),
+                }
             }
+            (RegImm::Reg(rs), rd) => match (rs.class(), rd.class()) {
+                (RegClass::Int, RegClass::Int) => self.asm.mov_rr(rs, rd, size),
+                (RegClass::Float, RegClass::Float) => self.asm.fmov64_rr(rs, rd),
```
saulecabrera, line 215: I was under the impression that we could emit `fmov sd, sn` (the non-64-bit case) according to https://developer.arm.com/documentation/ddi0602/2024-03/SIMD-FP-Instructions/FMOV--register---Floating-point-Move-register-without-conversion-?lang=en, but maybe we'd need to implement this in Cranelift to make it possible. Could you a TODO comment here? I'll follow-up on this one.

saulecabrera (commented):
Thanks for the PR; the general direction looks good to me. I left a couple of comments that I think need to be resolved before landing this work.

winch/codegen/src/isa/aarch64/regs.rs
```diff
@@ -147,3 +152,10 @@ pub(crate) const NON_ALLOCATABLE_GPR: u32 = (1 << ip0().hw_enc())
 
 /// Bitmask to represent the available general purpose registers.
 pub(crate) const ALL_GPR: u32 = u32::MAX & !NON_ALLOCATABLE_GPR;
+
+/// Bitmask for non-allocatable FPR.
+/// All FPRs are allocatable, v0..=v7 are generally used for params and results.
+pub(crate) const NON_ALLOCATABLE_FPR: u32 = 0;
+
+/// Bitmask to represent the available general purpose registers
+pub(crate) const ALL_FPR: u32 = u32::MAX;
```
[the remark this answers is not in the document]
evacchi: oh good point!

winch/codegen/src/isa/aarch64/masm.rs
evacchi: sure I'll add a TODO! 

If I understand correctly, the optimizing backend always uses F64 unless it needs V128:

ISLE:
```
(decl fpu_move (Type Reg) Reg)
(rule (fpu_move _ src)
      (let ((dst WritableReg (temp_writable_reg $I8X16))
            (_ Unit (emit (MInst.FpuMove128 dst src))))
        dst))
(rule 1 (fpu_move (fits_in_64 _) src)
      (let ((dst WritableReg (temp_writable_reg $F64))
            (_ Unit (emit (MInst.FpuMove64 dst src))))
        dst))
```

generated code:

```rust
// Generated as internal constructor for term fpu_move.
pub fn constructor_fpu_move<C: Context>(ctx: &mut C, arg0: Type, arg1: Reg) -> Reg {
    let v7 = C::fits_in_64(ctx, arg0);
    if let Some(v8) = v7 {
        let v10 = C::temp_writable_reg(ctx, F64);
        let v11 = MInst::FpuMove64 { rd: v10, rn: arg1 };
        let v12 = C::emit(ctx, &v11);
        let v13 = C::writable_reg_to_reg(ctx, v10);
        // Rule at src/isa/aarch64/inst.isle line 2251.
        return v13;
    }
    ... // v128
```

saulecabrera (approved):

PR #8369, opened 2024-04-15
Use `--locked` on all `cargo install` in CI, also remove non-locked example

alexcrichton:
Prevents any updates to rustc or crates from accidentally causing issues by ensuring that the same set of deps is used over time.


alexcrichton:
I first saw this [here](https://github.com/bytecodealliance/wasmtime/actions/runs/8682024349/job/23805761838?pr=8368) where a nightly install was failing due to proc-macro2 not compiling on the nightly we're pinned to.

alexcrichton:
I'm going to go ahead and land this with the extra commit removing the auxiliary markdown example. If others disagree about its removal and would like to keep it in then we can re-land it, but for now CI is broken and nothing can land.

PR #8368, opened 2024-04-15
[20.0.0] More backports of fixes, mostly c-api related

alexcrichton:
* https://github.com/bytecodealliance/wasmtime/pull/8344 - needed for later cherry-picks
* https://github.com/bytecodealliance/wasmtime/pull/8345 - needed for smoother externref c-api changes
* https://github.com/bytecodealliance/wasmtime/pull/8353 - bugfix
* https://github.com/bytecodealliance/wasmtime/pull/8358 - release notes
* https://github.com/bytecodealliance/wasmtime/pull/8356 - bugfix for c-api
* https://github.com/bytecodealliance/wasmtime/pull/8346 - preserve externref support in the c-api
* https://github.com/bytecodealliance/wasmtime/pull/8355 - bugfix
* https://github.com/bytecodealliance/wasmtime/pull/8364 - bugfix
* https://github.com/bytecodealliance/wasmtime/pull/8366 - bugfix for wasmtime-py
* https://github.com/bytecodealliance/wasmtime/pull/8363 - bugfix for wasmtime-cpp (and others)

PR #8370, opened 2024-04-15
Remove rust/WASI markdown parser example

alexcrichton:
The documentation referring to this example was removed in #6994 and that forgot to remove this as well. This example is building without a lock file which is causing issues in #8368.


alexcrichton:
I'm actually going to fold this into https://github.com/bytecodealliance/wasmtime/pull/8369 as otherwise CI is failing.

Issue #8304, opened 2024-04-05
`wasmtime::component::bindgen!` config 'with' option should error if provided interface does not exist in world


rylev:
Currently, the `with` config option in `bindgen!` silently ignores interfaces that do not exist within the targeted world. Instead of this, preferably the macro should error with a quality message saying that the interface does not exist.

alexcrichton:
I think this is a great idea, and support in Wasmtime can be modelled after https://github.com/bytecodealliance/wit-bindgen/pull/842 which added an equivalent feature to the guest Rust support.

PR #8371, opened 2024-04-15
Error when `wasmtime::component::bindgen` has unused with param.

rylev:
Instead of silently dropping unused with parameters on the floor, we instead error letting the user know that this option is being ignored.

I don't believe there's scaffolding for testing errors produced by the `bindgen!` macro, so I'm not sure if this can currently be tested.
 
Fixes https://github.com/bytecodealliance/wasmtime/issues/8304


alexcrichton (approved):
Thanks for this!

PR #8372, opened 2024-04-15
Update ir.md

Jakersnell:
Update C example function for correct formatting. This is useful as the function return type and function signature were on separate lines, which appeared to be an error as this is not how C functions are commonly formatted.


abrown (approved):
Thanks for the PR. I'm not sure what error you were seeing (I think both versions are valid C) but let's go with this becasue "return types on the same line" might be a bit more conventional.

PR #8373, opened 2024-04-15
Use the test harness for the cranelift-tools::filetests test

elliottt:
While working towards running tests with `nextest`, I found that the `filetests` test in `cranelift-tools` doesn't behave in the way that `nextest` expects. Wrapping it in `#[test]` and having it use the test harness fixes this issue.

### DIFF
--- a/cranelift/Cargo.toml
+++ b/cranelift/Cargo.toml
@@ -16,10 +16,6 @@ workspace = true
 name = "clif-util"
 path = "src/clif-util.rs"
 
-[[test]]
-name = "filetests"
-path = "tests/filetests.rs"
-
 [dependencies]
 cfg-if = { workspace = true }
 cranelift-codegen = { workspace = true, features = ["disas", "trace-log", "timing"] }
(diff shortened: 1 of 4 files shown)


alexcrichton (commented):
For this particular case could `#[test]` be used? Using libtest-mimic here is basically the same as a single function with `#[test]` I think

elliottt:
> For this particular case could `#[test]` be used? Using libtest-mimic here is basically the same as a single function with `#[test]` I think

Great point, that's totally possible!

cranelift/Cargo.toml
```diff
@@ -19,7 +19,6 @@ path = "src/clif-util.rs"
 [[test]]
 name = "filetests"
 path = "tests/filetests.rs"
-harness = false
```
alexcrichton, line 22: If you're feeling particularly intrepid, this entire block can be deleted now.

alexcrichton (approved):
I briefly looked into this awhile ago to see if we could actually hook into libtest-mimic and generate a single test per filetests since I think that'd be preferable here, but `cranelift_filetests` has a fair bit of infrastructure which wasn't trivial to disentangle so I abandoned the effort pretty quickly. This seems like a reasonable compromise though.

cranelift/Cargo.toml
elliottt: :fire:

PR #8374, opened 2024-04-15
c-api: Create `RootScope` where necessary

### LINKED ISSUE #8367: C API prevents GC'ing externref values
I'm in the process of trying to update wasmtime-py to the latest C API to weed out any issues ahead of the 20.0.0 release. One test is currently failing which asserts that a value is collected and unreferenced in Python, but it's failing. The general shape is:

* Create a `wasmtime_externref_t`
* Put that in `wasmtime_val_t`
* Pass that to `wasmtime_func_call`
* Run `store.gc()`
* Assert that the value has geen gc'd

The problem appears to be [here](https://github.com/bytecodealliance/wasmtime/blob/e1de17fb1a104a7b2aa75933044b9fecd4138dc9/crates/c-api/src/val.rs#L239) if I'm understanding this correctly. Notably in this location there's no `RootScope` so the `to_val`, which creates a `Rooted<T>`, attaches itself to the `Store` directly meaning that it's now un-collect-able.

cc @fitzgen

alexcrichton:
This commit changes the `wasmtime_val_t::{from_val, to_val}` methods to take a `RootScope` instead of any `AsContextMut`. This then required a number of refactorings in callers to ensure that a `RootScope` was created for any function that needed one. This is required to ensure that GC references in the C API aren't forced to live for the entire lifetime of the store.

This additionally added `*_unrooted` variants which do the same thing but don't require `RootScope`. This was needed for when the C API calls out to the embedder through a function call because a new `RootScope` wouldn't work for return values (they're bound to a scope within the closure when we want them to outlive the closure). In these situations though we know a `RootScope` is already present at the entrypoint.

Closes #8367

### DIFF
--- a/crates/c-api/src/val.rs
+++ b/crates/c-api/src/val.rs
@@ -5,7 +5,7 @@ use crate::{
 };
 use std::mem::{self, ManuallyDrop, MaybeUninit};
 use std::ptr;
-use wasmtime::{AsContextMut, Func, HeapType, Ref, Val, ValType};
+use wasmtime::{AsContextMut, Func, HeapType, Ref, RootScope, Val, ValType};
 
 #[repr(C)]
 pub struct wasm_val_t {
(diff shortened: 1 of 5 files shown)


crates/c-api/src/val.rs
```diff
@@ -177,7 +177,24 @@ pub struct wasmtime_func_t {
...
+    }
+
+    /// Equivalent of [`wasmtime_val_t::from_val`] except that a `RootScope`
+    /// is not required.
+    ///
+    /// This method should only be used when a `RootScope` is known to be
+    /// elsewhere on the stack. For example this is used when we call back out
+    /// to the embedder. In such a situation we know we previously entered with
+    /// some other call so the root scope is on the stack there.
+    pub fn from_val_unrooted(cx: impl AsContextMut, val: Val) -> wasmtime_val_t {
```
fitzgen, line 197: Maybe we can rename this `from_val_unscoped` or something? "unrooted" had me convinced that this was passing around `ExternRef::to_raw` results and stuff like that potentially across GCs which is not what's going on.

fitzgen: Also, the `wasmtime_val_t::from_val[_unrooted]` constructors never create `Rooted`s so I don't think we actually need two variants of these ones? Its just creating a `Val` (which contains a `Rooted` and means we might be creating a `Rooted` ourselves) that needs the scoping.

fitzgen (approved):

crates/c-api/src/val.rs
alexcrichton: Good point about the naming, I'll switch to your suggestion.

I thought the same for `from_val` but I ended up realizing that APIs like `wasmtime_global_get` also will want a `RootScope` on them. That's only calling `from_val` but if we forgot a `RootScope` there the any value coming out of a global would forever-after be attached to the store (since the `Rooted` coming out has no other scope). That convinced me to leave the `from_val{,_unscoped}` distinction to continue to encourage the use of `RootScope` anywhere in the C API that might use it.

PR #8376, opened 2024-04-16
cranelift: Simplify checking whether probestack is needed

jameysharp:
`Callee::probestack_min_frame` was always set to the guard-page size. That is the same as the `guard_size` local in `gen_prologue`, which was also the only place which used `probestack_min_frame`. So don't even bother storing it, just compute it from the flags as needed.

Also, remove the long-obsolete `probestack_func_adjusts_sp` setting, which hasn't been used since at least 2021. We may actually want to do something like what that setting described, but even if we do, it should be a choice the backend makes rather than a global setting.


elliottt (approved):
Yay! :tada:

PR #8377, opened 2024-04-16
Avoid copying the frame for tail calls on aarch64

elliottt:
To mirror the implementation in the x64 backend, switch to eagerly reserving enough space in the incoming argument area for any tail call present in the function being compiled.

NOTE: this doesn't make the change to enable callee-save registers with the tail calling convention on aarch64, but that will be a relatively small change following this PR.


cranelift/codegen/src/isa/aarch64/inst/emit.rs
```diff
@@ -3778,145 +3798,180 @@ fn emit_return_call_common_sequence(
...
+    //     rm: tmp1.to_reg(),
+    // }
+    // .emit(&[], sink, emit_info, state);
+    //
+    // state.virtual_sp_offset -= i64::from(info.new_stack_arg_size);
+    // trace!(
+    //     "return_call[_ind] adjusts virtual sp offset by {} -> {}",
+    //     info.new_stack_arg_size,
+    //     state.virtual_sp_offset
+    // );
```
fitzgen, line 3974: This was supposed to be removed, right?

fitzgen (approved):
Nice!

cranelift/codegen/src/isa/aarch64/inst/emit.rs
elliottt: Yep, thanks for catching that!

PR #8378, opened 2024-04-16
cranelift/PCC: Only process real address modes

jameysharp:
The `compute_addr` helper only works on real address modes, and one of its callers checks that it only passes a real address mode to it. On x64 we have different Rust types for address modes which are real versus synthetic, so we can simplify a little by reflecting this requirement in the types.

cc: @elliottt


cfallin (approved):
Nice simplification, thanks!

jameysharp:
If I'm reading CI logs correctly, I think the `async_::preview2_ip_name_lookup` test timed out, so I think the CI failure was a transient flake. Re-submitting without changes in hopes that it'll work this time.

PR #8381, opened 2024-04-16
Remove the `MmapVec::split_off` method

alexcrichton:
This method isn't actually correct if `split_off` is used on the returned values again as it's missing some extra indexing processing. That being said nothing in Wasmtime currently uses it, so remove it entirely instead of keeping it around.

found by @FrankReh, thanks!

### DIFF
--- a/crates/runtime/src/mmap_vec.rs
+++ b/crates/runtime/src/mmap_vec.rs
@@ -68,28 +68,6 @@ impl MmapVec {
         Ok(MmapVec::new(mmap, len))
     }
 
-    /// Splits the collection into two at the given index.
-    ///
-    /// Returns a separate `MmapVec` which shares the underlying mapping, but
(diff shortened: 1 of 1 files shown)


alexcrichton:
Notably @FrankReh recommended replacing with this:

```rust


pub fn split_off(&mut self, at: usize) -> MmapVec {
    assert!(at <= self.range.len());

    // Adjust the start of the range for the new part
    let new_start = self.range.start + at;

    // Create a new `MmapVec` which refers to the same underlying mmap, but
    // has a disjoint range from ours. Our own range is adjusted to be
    // disjoint just after `ret` is created.
    let ret = MmapVec {
        mmap: self.mmap.clone(),
        range: new_start..self.range.end,
    };
    self.range.end = new_start;
    return ret;
}
```

which I believe is correct, but also given that we don't need this it seems best to remove for now.

fitzgen (approved):

PR #8383, opened 2024-04-16
Avoid copying the frame for tail calls on riscv64

elliottt:
To mirror the implementation in the x64 backend, switch to eagerly reserving enough space in the incoming argument area for any tail call present in the function being compiled.

The diff here is large due to the change of representation for incoming arguments in the riscv64 specific `AMode` type. Now that we reference incoming arguments with a negative offset from the beginning of the incoming argument area, all references to incoming arguments changed in the riscv64 filetests. These would have ended up being no-op changes in the disassembly, though I took this opportunity to make them SP relative now, to move closer to frame pointers not being strictly required for tail calls.

As this is the last backend that was waiting to be migrated to the approach of resizing the incoming argument area in the function prologue, I've also deleted the shared code in `machinst/abi.rs` that's no longer needed.

NOTE: this doesn't make the change to enable callee-save registers with the tail calling convention on aarch64, but that will be a relatively small change following this PR.


cranelift/codegen/src/isa/riscv64/inst/args.rs
```diff
@@ -183,6 +202,9 @@ impl Display for AMode {
             &AMode::NominalSPOffset(offset, ..) => {
                 write!(f, "{}(nominal_sp)", offset)
             }
+            &AMode::IncomingArg(offset) => {
+                write!(f, "{}(incoming_arg)", offset)
```
elliottt, line 206: Maybe this would be better written as `fp(incoming_arg_area - {offset})`?

cranelift/codegen/src/isa/riscv64/inst/emit.rs
```diff
@@ -3331,136 +3325,49 @@ fn emit_return_call_common_sequence(
...
+    info: &ReturnCallInfo,
 ) {
-    // We are emitting a dynamic number of instructions and might need an
-    // island. We emit four instructions regardless of how many stack arguments
-    // we have, up to two instructions for the actual call, and then two
-    // instructions per word of stack argument space.
-    let new_stack_words = new_stack_arg_size / 8;
-    let insts = 4 + 2 + 2 * new_stack_words;
-    let space_needed = insts * u32::try_from(Inst::UNCOMPRESSED_INSTRUCTION_SIZE).unwrap();
-    if sink.island_needed(space_needed) {
```
afonso360, line 3344: Is this island check now done in some common code, or do we no longer need it? I remember having some issues with missing islands when we enabled compressed jumps since with the compressed extensions we have a very short jump range.

afonso360 (commented):
👋 Hey,

I don't think I can do a very good review here, I don't remember a lot about how our frames are set up and haven't been following the rest of the tail calls work very closely. So I'd prefer if someone else could review this.

cranelift/codegen/src/isa/riscv64/inst/args.rs
jameysharp: I'd also take `"-{}(incoming_arg)"` to signal that the offset is subtracted from the base. I don't think it's worth bikeshedding this very much but I think clarifying that it's a subtraction would help with readability.

cranelift/codegen/src/isa/riscv64/inst/emit.rs
elliottt: Thanks for catching this, that check isn't being done in any generic code. I'll add this back in!

jameysharp: On further inspection, there is generic code in `VCode::emit` that emits islands based on the backend's `Inst::worst_case_size()` estimate. Both `gen_epilogue` and these `ReturnCall`/`ReturnCallInd` pseudo-instructions need to fit within that worst-case size. These pseudo-instructions might be slightly longer than `gen_epilogue` right now but they're almost the same, so I think they don't need an explicit island check.

elliottt: I did some more refactoring, and the resulting code generated for a `return_call` or `return_call_indirect` should now be pretty close to the same size as `gen_epilogue`. Lots of chatting with @jameysharp has me convinced that we shouldn't need to check for island emission at this point anymore, and I've removed the manual resets at the call-sites for `emit_return_call_common_sequence` to ensure that we trigger an assert if this isn't true.

jameysharp (approved):
I'm convinced this is all correct. Ship it!

PR #8384, opened 2024-04-16
Trace outbound http request in wasi-http

calebschoepp:
This is to support tracing of outbound HTTP request in downstream libraries like [Spin](https://github.com/fermyon/spin/issues/2354).

- Adds tracing to `default_send_request` that follows the OTel HTTP trace [semantic conventions](https://opentelemetry.io/docs/specs/semconv/http/http-spans/).
- Makes sure that the trace context is propagated across the async boundary introduced by the spawn.


crates/wasi-http/src/types.rs
```diff
@@ -136,11 +138,34 @@ pub struct OutgoingRequestConfig {
 ///
 /// This implementation is used by the `wasi:http/outgoing-handler` interface
 /// default implementation.
+#[instrument(skip_all, fields(otel.kind = "client", url.full = Empty, http.request.method = Empty, http.response.status_code = Empty, otel.name = Empty, server.address = Empty, server.port = Empty))]
```
alexcrichton, line 141: I've not used `#[instrument]` myself before, but would it be possible to do something like `url.full = request.uri()` here in the macro instead of in the body below? I'm not sure if that has type restrictions on the return value of `.ur()`, but adding `.to_string()` may work well too.

Also, mind adding some comments as to what `otel.kind`, `otel.name`, and `server.{address,port}` are doing? The first two seem like sort of "standard variables" to set, but the second two aren't set here in the method either.

calebschoepp: Sure, I'll look into that @alexcrichton. Also sorry, I force pushed a linting fix without realizing you'd already started reviewing this.

calebschoepp:
@alexcrichton is this something worth backporting to `v18`? It seems like this code changes a reasonable amount between `v18` and `v19`.

calebschoepp:
Closing this b/c after a discussion with @alexcrichton we decided it is best to do this tracing in Spin. A subsequent PR will follow that will expose `handler` for Spin to use.

PR #8385, opened 2024-04-16
Ensure that indirect tail calls on x64 don't clobber the destination register

elliottt:
When we changed how `return_call` is handled during instruction emission, we assumed that `r11` would always be safe to use as it's a caller-save register on x64. However, this also means that it's available to RA2 for allocation, and could end up holding the destination of a `return_call_indirect`. To fix this, I've added a temporary writable register to the x64 `ReturnCallInfo` structure, and marked it with an early def constraint to ensure that it won't conflict with the destination of an indirect call.


fitzgen (approved):

jameysharp (approved):
Looks great, thank you!

PR #8387, opened 2024-04-16
cranelift: Represent RealReg using PReg, not VReg

jameysharp:
Semantically, a "real register" is supposed to be a physical register, so let's use the type dedicated for that purpose.

This has the advantage that `PReg` is only one byte while `VReg` is four bytes, so the arrays where we record collections of `RealReg` become smaller.

There was an implementation of `From<VReg> for RealReg` which was not a sensible conversion, because not all `VReg`s are valid `RealReg`s. I could have replaced it with a `TryFrom` implementation but it wasn't used anywhere important, so I'm just deleting it instead.

Winch was using that `VReg`->`RealReg` conversion, but only in the implementation of another conversion that was itself unused, so I'm deleting that conversion as well. It's easy to implement correctly (the Winch `Reg` type is identical to `RealReg`, so all conversions for the latter are readily available) but as far as I can tell Winch doesn't need to use Cranelift's register wrappers or RA2's virtual register type, so it's simpler to just delete those conversions.

The riscv64 backend was relying on quirks of the existing conversions between `RealReg` and `VReg` when emitting clobber saves and restores. Just using the generic conversions between `RealReg` and `Reg` is simpler and works correctly with the rest of these changes.

cc: @elliottt

### DIFF
--- a/cranelift/codegen/src/isa/riscv64/abi.rs
+++ b/cranelift/codegen/src/isa/riscv64/abi.rs
@@ -491,7 +491,7 @@ impl ABIMachineSpec for Riscv64MachineDeps {
                 }
                 insts.push(Inst::gen_store(
                     AMode::SPOffset(-(cur_offset as i64)),
-                    real_reg_to_reg(reg.to_reg()),
+                    Reg::from(reg.to_reg()),
                     ty,
                     MemFlags::trusted(),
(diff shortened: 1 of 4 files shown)


elliottt (approved):
Very nice cleanup, thank you!

PR #8388, opened 2024-04-16
Make wasi-http handler public

calebschoepp:
Downstream dependencies may want to own the work of `spawn(handler())` themselves. This enables that.


calebschoepp:
@alexcrichton here is the replacement for #8384.

alexcrichton (approved):
Thanks! Reading over this again, mind renaming this to perhaps `default_send_request_handler`? (in retrospect a public function `handler` is a bit bland)

PR #8389, opened 2024-04-16
Use callee-save registers in tail calls on aarch64

elliottt:
Switch back to mostly following the AArch64 ABI for callee-save registers with the tail calling convention, except that we still reserve x0 for the return area pointer and x1 for the destination of an indirect tail call.

### DIFF
--- a/cranelift/filetests/filetests/isa/aarch64/return-call-indirect.clif
+++ b/cranelift/filetests/filetests/isa/aarch64/return-call-indirect.clif
@@ -217,92 +217,143 @@ block0:
 ; VCode:
 ;   stp fp, lr, [sp, #-16]!
 ;   mov fp, sp
-;   sub sp, sp, #16
-;   ldr fp, [sp, #16]
+;   sub sp, sp, #160
+;   ldr fp, [sp, #160]
 ;   stp fp, lr, [sp]
 ;   mov fp, sp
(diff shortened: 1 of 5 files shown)


cranelift/filetests/filetests/isa/aarch64/return-call-indirect.clif
```diff
@@ -217,92 +217,143 @@ block0:
...
+;   str x26, [sp, #192]
+;   str x27, [sp, #200]
+;   str x28, [sp, #208]
+;   str x21, [sp, #216]
+;   str x19, [sp, #224]
+;   str x20, [sp, #232]
+;   str x22, [sp, #240]
+;   str x23, [sp, #248]
+;   str x24, [sp, #256]
+;   str x25, [sp, #264]
```
fitzgen, line 278: Potential follow up issue: it seems like we should use the store-pair instructions for restoring callee-save registers in the future, when possible.

elliottt: Yeah, that's a great point!

elliottt: Though in this case I think these are incoming args area stores for the tail call, but it would still be great to have the infrastructure available to merge those into a single instruction.

PR #8390, opened 2024-04-16
cranelift: Factor out splitting clobbered regs by class

jameysharp:
Since the vector of clobbered registers is sorted first by class, we can return slices of that vector instead of allocating a temporary vector for each class.

This relies on the `RealReg` sort order and on the frame layout's `clobbered_callee_saves` being sorted in natural order.

cc: @elliottt


elliottt (approved):
Very nice cleanup!

Issue #8380, opened 2024-04-16
PCC facts checking failed on some workloads


feilongjiang:
When I run PCC checks on [PloyBenchC](https://github.com/MatthiasJReisinger/PolyBenchC-4.2.1), I get the following error:
```
Error: Compilation error: Proof-carrying-code validation error: UnsupportedFact
```

After adding `WASMTIME_LOG=trace`, the failure was actually caused by the following facts checking:
```
TRACE cranelift_codegen::isa::x64::pcc: Checking facts on inst: movl    $59944, %v2906l   
TRACE cranelift_codegen::machinst::pcc: checking if derived fact Some(Range { bit_width: 64, min: 59944, max: 59944 }) subsumes stated fact Some(Range { bit_width: 64, min: 59944, max: 59944 })    
TRACE cranelift_codegen::isa::x64::pcc: Checking facts on inst: movl    $8, %v2905l    
TRACE cranelift_codegen::machinst::pcc: checking if derived fact Some(Range { bit_width: 64, min: 8, max: 8 }) subsumes stated fact Some(Range { bit_width: 64, min: 8, max: 8 })    
TRACE cranelift_codegen::isa::x64::pcc: Checking facts on inst: lea     71072(%v299), %v2904l    
TRACE cranelift_codegen::isa::x64::pcc: compute_addr: Real(ImmReg { simm32: 71072, base: v299, flags: MemFlags { bits: 1921 } }), bits: 32    
TRACE cranelift_codegen::machinst::pcc: get_fact_or_default: reg v299 -> Some(Range { bit_width: 64, min: 8, max: 8 })    
TRACE cranelift_codegen::isa::x64::pcc: base = Range { bit_width: 64, min: 8, max: 8 }    
TRACE cranelift_codegen::ir::pcc: add: Range { bit_width: 64, min: 8, max: 8 } + Range { bit_width: 32, min: 71072, max: 71072 } -> None, add_width: 32    
TRACE cranelift_codegen::machinst::pcc: clamp_range: fact None from 32 to 64    
TRACE cranelift_codegen::machinst::pcc:  -> clamping to Range { bit_width: 64, min: 0, max: 4294967295 }    
TRACE cranelift_codegen::machinst::pcc: checking if derived fact Some(Range { bit_width: 64, min: 0, max: 4294967295 }) subsumes stated fact Some(Range { bit_width: 64, min: 71080, max: 71080 })    
[...]

PR #8393, opened 2024-04-17
PPC: Fix validation error on PolyBenchC

feilongjiang:
Hi team, please consider.
As described in #8380, when running PCC validation on workloads like PolyBenchC, it will report an `UnsupportedFact` error.
Based on the analysis of trace logs, I think `attach_constant_fact` introduced by #8173 should also consider the type of constant when adding facts for iconsts. After #8173, we may generate the following facts for i32 constants:
```
block9:
  v4339 ! range(64, 0xea28, 0xea28) = iconst.i32 0xea28
  v2858 ! range(64, 0x8, 0x8) = iconst.i32 8
  v4406 ! range(64, 0x115a0, 0x115a0) = iconst.i32 0x0001_15a0
  v4407 ! range(64, 0x115a8, 0x115a8) = iadd v4406, v2858  ; v4406 = 0x0001_15a0, v2858 = 8
```
These i32 constants have their `bit_width` set to 64, causing the error.

#8173 also exposes another issue on AArch64 when generating facts for `MovK`, we should calculate the running value for `MovK` with the field mask.

### DIFF
--- a/cranelift/codegen/src/isa/aarch64/pcc.rs
+++ b/cranelift/codegen/src/isa/aarch64/pcc.rs
@@ -306,7 +306,7 @@ pub(crate) fn check(
             if let Some(input_constant) = input.as_const(64) {
                 let mask = 0xffff << (imm.shift * 16);
                 let constant = u64::from(imm.bits) << (imm.shift * 16);
-                let constant = input_constant & !mask | constant;
+                let constant = (input_constant & !mask) | constant;
                 check_constant(ctx, vcode, rd, 64, constant)
             } else {
                 check_output(ctx, vcode, rd, &[], |_vcode| {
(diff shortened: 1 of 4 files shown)


cranelift/codegen/src/isa/aarch64/pcc.rs
```diff
@@ -304,8 +304,9 @@ pub(crate) fn check(
         Inst::MovK { rd, rn, imm, .. } => {
             let input = get_fact_or_default(vcode, rn, 64);
             if let Some(input_constant) = input.as_const(64) {
+                let mask = 0xffff << (imm.shift * 16);
                 let constant = u64::from(imm.bits) << (imm.shift * 16);
-                let constant = input_constant | constant;
+                let constant = input_constant & !mask | constant;
```
jameysharp, line 309: I'd prefer parentheses here even if they're not strictly required:
```suggestion
                let constant = (input_constant & !mask) | constant;
```

jameysharp (commented):
Thank you! This looks good to me, though I'd like @cfallin to take a look before merging it, and he may not be able to get to it until next week.

feilongjiang:
@jameysharp Thanks for the review!
> and he may not be able to get to it until next week.

That’s okay, we can wait until @cfallin comes back!

cfallin (approved):
LGTM, thanks!

PR #8396, opened 2024-04-17
Move doctest out to a dedicated job

elliottt:
Move doctests out of individual tests by shadowing `rustdoc` when running `./ci/run-tests.sh`, and explicitly run doctests only in their own dedicated job on an `ubuntu-latest` runner.


elliottt:
> Comparing this with another run sampled historically I don't think there's any speedup? That matches my gut on this given the low core count of actions runners and how all our tests are internally parallel too.

I'm seeing 15 vs 18 minutes for the Linux x64 run, and that's with the non-cached build of `cargo-nextest`.

elliottt:
> > Comparing this with another run sampled historically I don't think there's any speedup? That matches my gut on this given the low core count of actions runners and how all our tests are internally parallel too.
> 
> I'm seeing 15 vs 18 minutes for the Linux x64 run, though that's with the cached build of `cargo-nextest`.

Never mind, @alexcrichton pointed out that this is due to not running the lldb tests on branches :)

elliottt:
Given that this isn't really speeding up testing, I don't think this is worth pursuing. If we decide to start partitioning tests onto different runners in the future it would be worth picking this back up, but without that the benefit just isn't there.

elliottt:
OK, I've also concluded that it's not worth merging just the change to move the doctests out to a separate job, as it would only speed up CI on branches by about 2 minutes, and would have no effect on total merge time on main. It has the additional downside of ignoring any `doctest = false` settings in any Cargo.toml file, as `cargo test --doc` unconditionally runs doctests.

PR #8397, opened 2024-04-17
cranelift: Round inline stack probes down, not up

jameysharp:
When we have `enable_probestack` turned on and set `probestack_strategy` to "inline", we have to compute how many pages of the stack we'll probe.

The current implementation rounds our stack frame size up to the nearest multiple of the page size, then probes each page once.

However, if our stack frame is not a multiple of the page size, that means there's a partial page at the end. It's not necessary to probe that partial page, just like it's unnecessary to probe at all if the frame is smaller than one page. Either way, any signal handler needs to be prepared for stack accesses on that last page to fault at any time during the function's execution.

### DIFF
--- a/cranelift/codegen/src/isa/aarch64/abi.rs
+++ b/cranelift/codegen/src/isa/aarch64/abi.rs
@@ -705,6 +705,8 @@ impl ABIMachineSpec for AArch64MachineDeps {
         // Set this to 3 to keep the max size of the probe to 6 instructions.
         const PROBE_MAX_UNROLL: u32 = 3;
 
+        // Calculate how many probes we need to perform. Round down, as we only
+        // need to probe whole guard_size regions we'd otherwise skip over.
         let probe_count = frame_size / guard_size;
         if probe_count == 0 {
             // No probe necessary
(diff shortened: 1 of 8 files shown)


cranelift/codegen/src/isa/aarch64/abi.rs
```diff
@@ -677,8 +677,10 @@ impl ABIMachineSpec for AArch64MachineDeps {
         // Set this to 3 to keep the max size of the probe to 6 instructions.
         const PROBE_MAX_UNROLL: u32 = 3;
 
-        let probe_count = align_to(frame_size, guard_size) / guard_size;
-        if probe_count <= PROBE_MAX_UNROLL {
+        let probe_count = frame_size / guard_size;
+        if probe_count == 0 {
+            // No probe necessary
+        } else if probe_count <= PROBE_MAX_UNROLL {
```
fitzgen, line 683: Mind leaving a comment about why this doesn't need aligning up here? Kind of unfortunate it would need to be copied to all the other inline probestack code paths. Maybe that is a sign that we should factor this little bit of logic out to a generic helper?

fitzgen (approved):
Good find!

cranelift/codegen/src/isa/aarch64/abi.rs
sunfishcode: I've now added comments explaining why we round down. I looked at factoring this code out into a helper, but it felt like it made the code a little less readable, so I didn't add it to the PR, however I'm open to suggestions.

sunfishcode:
@fitzgen I've now rebased this on main, added comments addressing your review feedback above, and ported this patch to the s390x backend. Would you mind taking another look?

fitzgen (approved):
Thanks for getting this over the line!

PR #8398, opened 2024-04-18
cranelift: Emit argument location uses eagerly in gen_arg

jameysharp:
This reverts the key parts of e3a08d4c40f64ff7a2113d23af73f40af0859a10 (#8151), because it turns out that we didn't need that abstraction.

Several changes in the last month have enabled this:

- #8292 and then #8316 allow us to refer to either incoming or outgoing argument areas in a (mostly) consistent way

- #8327, #8377, and #8383 demonstrate that we never need to delay writing stack arguments directly to their final location


elliottt (approved):
Awesome! Now that I'm thinking about it, I believe we added this specifically for `GrowArgumentArea` and `ShrinkArgumentArea`, both of which no longer exist :tada:

jameysharp:
Of the several variations on the theme that we've considered, this one was even earlier than the grow/shrink frame approach. We intended to use regalloc2's parallel move resolver to shuffle values between stack slots, registers, and the argument area, to capture all the work of restoring clobber-saves, putting arguments in the right places, and getting the return address in the right place too. That turned out to be complicated. Fortunately, the fact that Ulrich's suggestion (proactively resizing the argument area) works shows we will never have to do parallel moves.

PR #8401, opened 2024-04-18
Rework the riscv64 prologue to spill clobbers with positive offsets to SP

elliottt:
Instead of generating negative offsets to SP for clobbers and updating SP after clobbers have been spilled, eagerly bump SP and write the clobbers to positive offsets instead. This change should preserve the locations on the stack where the clobbers are written, so no changes were necessary for the epilogue.


jameysharp (approved):
I believe this change is correct and also fixes a bug: We're generally not supposed to write important information, like clobber saves, beyond wherever the stack pointer points. A signal handler could run on our stack at any time and is allowed to overwrite anything outside our reserved portion of the stack. Some architectures provide a small "red zone" where we could get away with this, but we shouldn't rely on that.

For the same reason, the epilogue should also be updated to not adjust the stack pointer until it has already restored all the clobbers.

Would you update the commit message and PR description to mention this bug fix?

PR #8402, opened 2024-04-18
Fix x64 clobber offsets with unwind info

elliottt:
The stack offsets provided to `UnwindInst::SaveReg` instructions are relative to the beginning of the clobber area of the stack, and on main the x64 backend computes those offsets by assuming a SP-relative offset and then adjusting it. This is problematic, as any changes to the frame layout might require updates in two locations: where we compute the initial SP offset, and where we compute the adjustment to make the offset relative to the clobber area again.

This PR fixes this issue by making the offset we track in the loop always relative to the beginning of the clobber area, and adding the offset from SP only when emitting the store. This relies on the assumption that the clobber area is sized according to the constraints of the registers stored within it (16-byte alignment is required for floating point registers) but this assumption is enforced by existing behavior in `compute_clobber_size`.

This also fixes a long-standing bug with floating point register spill offsets, as the offset was saved before the alignment to a 16-byte offset was made, leaving the resulting unwind info potentially pointing at the wrong offset.

### DIFF
--- a/cranelift/codegen/src/isa/x64/abi.rs
+++ b/cranelift/codegen/src/isa/x64/abi.rs
@@ -723,35 +723,32 @@ impl ABIMachineSpec for X64ABIMachineSpec {
 
         // Store each clobbered register in order at offsets from RSP,
         // placing them above the fixed frame slots.
-        let mut cur_offset =
+        let clobber_offset =
             frame_layout.fixed_frame_storage_size + frame_layout.outgoing_args_size;
+        let mut cur_offset = 0;
         for reg in &frame_layout.clobbered_callee_saves {
             let r_reg = reg.to_reg();
-            let off = cur_offset;
-            match r_reg.class() {
-                RegClass::Int => {
-                    insts.push(Inst::store(
-                        types::I64,
-                        r_reg.into(),
-                        Amode::imm_reg(cur_offset.try_into().unwrap(), regs::rsp()),
-                    ));
-                    cur_offset += 8;
-                }
-                RegClass::Float => {
-                    cur_offset = align_to(cur_offset, 16);
-                    insts.push(Inst::store(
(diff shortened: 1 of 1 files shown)


fitzgen (approved):
Didn't audit this super carefully, since I trust you and Jamey on this stuff more than myself right now since you've been digging into it very recently, but I wonder if we should (in a follow up at some point in the future) only adjust SP via a helper function that not only emits the SP adjustment, but also the unwind info metadata instruction at the same time.

elliottt:
> Didn't audit this super carefully, since I trust you and Jamey on this stuff more than myself right now since you've been digging into it very recently, but I wonder if we should (in a follow up at some point in the future) only adjust SP via a helper function that not only emits the SP adjustment, but also the unwind info metadata instruction at the same time.

Possibly, though in this case the offset is expected to be relative to part of the stack. Were you thinking that we'd have a helper that assumed all unwind offsets to be SP relative, and made the adjustment when emitting the underlying `UnwindInst`?

jameysharp (approved):
I love it, this is so much better. Thank you!

jameysharp (approved):
Ah, making the same cleanup in clobber-restore was a good idea.

fitzgen:
> Possibly, though in this case the offset is expected to be relative to part of the stack. Were you thinking that we'd have a helper that assumed all unwind offsets to be SP relative, and made the adjustment when emitting the underlying `UnwindInst`?

It wasn't a fully formed idea. I was thinking that the canonical frame address (CFA) is always at a constant offset from the SP for any given PC (I think this holds true for us, since we don't have any `alloca`s) and that whenever we increment the SP, we always want to increment the offset to the CFA (i.e. emit an unwind inst) by the same amount. Not sure that this actually aligns 100% with what we are trying to do in general, though.

PR #8403, opened 2024-04-18
cranelift/aarch64: Simplify generating constants

jameysharp:
While reviewing #8393 I found the existing `load_constant64_full` function nearly incomprehensible, so I rewrote it. Besides having many fewer cases now, this should also fix part of the bug that #8393 addresses, but that PR fixes other issues as well.

This PR changes the operand size that we emit for some cases from 32-bit to 64-bit. In those cases, the operand size doesn't matter because we want the upper 32 bits to be zero. We set them that way with the initial instruction: widening `movz` from 32-bit to 64-bit doesn't change anything. And we preserve the upper bits with subsequent `movk` instructions, instead of zeroing them, but they were already zero.

### DIFF
--- a/cranelift/codegen/src/isa/aarch64/lower/isle.rs
+++ b/cranelift/codegen/src/isa/aarch64/lower/isle.rs
@@ -3,7 +3,6 @@
 // Pull in the ISLE generated code.
 pub mod generated_code;
 use generated_code::Context;
-use smallvec::SmallVec;
 
 // Types that the generated ISLE code uses via `use super::*`.
 use super::{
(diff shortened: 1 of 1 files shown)


jameysharp:
#8393 landed so I've rebased this.

Also, I've switched to an implementation which I believe generates exactly the same instructions as the prior implementation in all cases. There are some choices we could make which would synthesize the same constants using a different sequence of instructions, but in almost all cases we'd produce the same number of instructions, so it doesn't make much difference. This algorithm is simple and produces readable disassembly listings so we might as well preserve the existing behavior.

elliottt (approved):
Awesome!

PR #8404, opened 2024-04-18
Add support for recursion groups in the types registry

fitzgen:
This adds support for recursion groups in Wasmtime's type registry, but still just for function types. Adding struct and array types should be relatively easy after this. At least, we won't have to be concerned with canonicalization, just updating heap type matching everywhere and all that.

Depends on https://github.com/bytecodealliance/wasm-tools/pull/1506 and updating the `wasm-tools` crates before this can land.

### DIFF
--- a/crates/wasmtime/src/runtime/component/instance.rs
+++ b/crates/wasmtime/src/runtime/component/instance.rs
@@ -319,11 +319,11 @@ impl<'a> Instantiator<'a> {
         // below, but the funcrefs can all be configured here.
         for (idx, sig) in env_component.trampolines.iter() {
             let ptrs = self.component.trampoline_ptrs(idx);
-            let signature = self
-                .component
-                .signatures()
-                .shared_type(*sig)
-                .expect("found unregistered signature");
+            let signature = match self.component.signatures().shared_type(*sig) {
+                Some(s) => s,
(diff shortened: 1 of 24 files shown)


crates/wasmtime/src/runtime/component/instance.rs
```diff
@@ -502,9 +504,18 @@ impl<'a> Instantiator<'a> {
...
                 EngineOrModuleTypeIndex::Engine(e) => Some(e),
                 EngineOrModuleTypeIndex::Module(m) => module.signatures().shared_type(m),
+                EngineOrModuleTypeIndex::RecGroup(_) => unreachable!(),
             };
             let actual = unsafe { f.func_ref.as_ref().type_index };
-            assert_eq!(expected, Some(actual));
+            assert_eq!(
+                expected,
+                Some(actual),
+                "type mismatch for import {imp_module:?} {imp_name:?}!!!\n\n\
```
alexcrichton, line 513: !!!!1!!11!!

alexcrichton (approved):

crates/wasmtime/src/runtime/component/instance.rs
fitzgen: :exclamation:

PR #8405, opened 2024-04-18
Use callee-save registers for the riscv64 tail calling convention

elliottt:
Switch to using the same set of callee saved registers as the default calling convention on riscv64. This also introduces the contraint that indirect tail calls store their destination in `t0`, which was an assumption of indirect calls before, but not indirect return calls.


jameysharp (approved):
I believe this is correct for the same reasons as similar changes to aarch64, as discussed in https://github.com/bytecodealliance/wasmtime/pull/8406#pullrequestreview-2010186633.

PR #8406, opened 2024-04-18
aarch64: Fixed destination register for return_call_indirect

elliottt:
After switching to using callee-saved registers with the tail calling convention on aarch64, we missed using a fixed constraint for the destination of indirect return calls. This omission introduces a situation where the destination could be placed in a callee-saved register, and then clobbered as part of the call to `emit_return_call_common_sequence`.

This PR fixes the problem by always using `x1` for return call destinations, and relaxes the same constraint on indirect calls of tail call functions, as callee-saved registers can be used for normal indirect calls now.

### DIFF
--- a/cranelift/codegen/src/isa/aarch64/inst/mod.rs
+++ b/cranelift/codegen/src/isa/aarch64/inst/mod.rs
@@ -891,14 +891,7 @@ fn aarch64_get_operands<F: Fn(VReg) -> VReg>(inst: &Inst, collector: &mut Operan
             collector.reg_clobbers(info.clobbers);
         }
         &Inst::CallInd { ref info, .. } => {
-            match info.callee_callconv {
-                CallConv::Tail => {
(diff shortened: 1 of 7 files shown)


jameysharp (approved):
I hope @cfallin can double-check this later, but I'm pretty confident this is all correct and you can go ahead and merge it.

We need to constrain `ReturnCallInd`, because it will overwrite any callee-saved registers before reading the value of `callee`, so we must not use any of those registers for `callee`.

But we don't need to constrain `CallInd` to avoid using callee-saved registers, because it doesn't overwrite any of them before reading the value of `callee`. In fact the whole point of callee-saved registers is that the actual call will preserve them, so we should _prefer_ to use them for normal calls if we still need the address afterward.

cfallin:
> I hope @cfallin can double-check this later

Yep, looks good to me as well! I double-checked that with the tailcall ABI we start assigning args to registers at `x2`, so `x1` is free to use for this purpose here and I can't think of anything else that would clobber it between setup and use.

PR #8407, opened 2024-04-18
x64: Use r10 for the indirect tail call destination

elliottt:
Use `r10` for the destination of indirect return calls, and indirect calls using the winch calling convention, as it is a caller-saved register. Additionally, always use `r11` for the temporary that's used for moving the return address in `emit_return_call_common_sequence`, as it's also a caller-saved register.

For tail calls, this ensures that we won't accidentally pick a callee-saved register for the destination or temporary, clobbering it when we restore callee-saves in the call to `emit_return_call_common_sequence`.

For winch calls, using `r10` instead of `r15` means that it's still possible to use the pinned register in combination with winch.

### DIFF
--- a/cranelift/codegen/src/isa/x64/inst/mod.rs
+++ b/cranelift/codegen/src/isa/x64/inst/mod.rs
@@ -1690,7 +1690,7 @@ impl PrettyPrint for Inst {
                     new_stack_arg_size,
                     tmp,
                 } = &**info;
-                let callee = callee.pretty_print(8, allocs);
+                let callee = pretty_print_reg(*callee, 8, allocs);
                 let tmp = pretty_print_reg(tmp.to_reg().to_reg(), 8, allocs);
                 let mut s =
                     format!("return_call_unknown {callee} ({new_stack_arg_size}) tmp={tmp}");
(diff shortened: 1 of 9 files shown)


cranelift/codegen/src/isa/x64/inst/mod.rs
```diff
@@ -2362,7 +2363,17 @@ fn x64_get_operands<F: Fn(VReg) -> VReg>(inst: &Inst, collector: &mut OperandCol
...
+                    // TODO(https://github.com/bytecodealliance/regalloc2/issues/145):
+                    // This shouldn't be a fixed register constraint, but it's not clear how to
+                    // pick a register that won't be clobbered by the callee-save restore code
+                    // emitted with a return_call_indirect. r10 is caller-saved, so this should be
+                    // safe to use.
+                    collector.reg_fixed_use(*reg, regs::r10())
+                }
+                RegMem::Mem { addr } => addr.get_operands(collector),
+            }
             collector.reg_early_def(tmp.to_writable_reg());
```
elliottt, line 2377: I think that we could switch back to always using `r11` for the temp, as this change means we'll no longer accidentally pick `r11` for the destination.

jameysharp: I don't think the `RegMem::Mem` case is safe here, because any registers in the address mode could themselves be callee-saved. We might need to require that `ReturnCallUnknown` is used only with a register, never a memory operand, unlike `CallUnknown`.

As for using `r11` for the temporary register, I think we actually must do that. In general I'd rather let RA2 choose the temporary, but it's used after callee-saved registers are restored, so as it stands now we could clobber something we're supposed to have preserved.

jameysharp (commented):
I think there are a couple of x86-specific challenges we still need to work through here.

cranelift/codegen/src/isa/x64/inst/mod.rs
elliottt: I was wondering about this, especially since it's not clear to me why this is a `RegMem` and not a plain old `Reg`.

elliottt: Actually, switching it to accepting a `Reg` was really easy, and the only place we construct this instruction we were always providing a register that we turned into a `RegMem` :+1:

jameysharp: In general it makes sense to me to allow an indirect call on x86 to sink a load, if that's where the callee came from.

We could even hit this case from Wasmtime, where an indirect call through a table gets the callee function pointer by loading it out of a struct in memory. (I think load sinking currently fails there because there's another load in between; we might be able to micro-optimize this for x86 by swapping the two loads.)

However, I think that's a code-size optimization we just can't safely support for tail calls.

jameysharp (approved):
I believe this is correct, though again I hope @cfallin will look at it later.

I'm also noticing that our tail-call sequence, when we move the return address, is not a valid epilogue according to the Windows unwinding rules. That is probably not an issue since nothing should trigger unwinding there, but it's worth keeping in mind.

cfallin:
Likewise to the aarch64 PR, this looks right to me, and for the same reasons -- as you note in the comments, r10 is caller-saved (volatile) and it's also not used for an arg, so we're good to go!

PR #8408, opened 2024-04-19
x64: Swap operand order of `cmp` in internal representation

alexcrichton:
This PR is #8362 but for general purpose registers instead of xmm registers. Upon reading over the instructions in `inst.isle` it appears that `x64_cmp` and its variants is the only remaining (after #8362) instruction where the operands are not in left-to-right form. The purpose of this PR is to refactor this situation to make sure all instructions have left-to-right argument orderings in their instruction helpers.

This PR is broken down into a number of small commits. Each commit swaps just one method and then handles all callers of that method. By the end all callers of all `cmp` instructions should have their operand order swapped. Note that special care was taken in Winch where there were a few swaps-of-swaps. For example the `branch` helper previously had a variable called `lhs` which actually was the right-hand-side of a comparison. This caused no issues for code today as all operations were reflexive or with the same value, but I've documented and ensured that lhs is now indeed the left-hand-side.

My intention as well is that `./ci/run-tests.sh` passes at each individual commit here as well. CI won't test that, but if anyone's interested to check it should work.

### DIFF
--- a/cranelift/codegen/src/isa/x64/abi.rs
+++ b/cranelift/codegen/src/isa/x64/abi.rs
@@ -469,7 +469,7 @@ impl ABIMachineSpec for X64ABIMachineSpec {
 
     fn gen_stack_lower_bound_trap(limit_reg: Reg) -> SmallInstVec<Self::I> {
         smallvec![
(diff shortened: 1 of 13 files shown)


elliottt (approved):
This looks great to me, thank you for making this all so much more consistent! I went through the individual commits, and really appreciate you breaking this change out like that!

PR #8409, opened 2024-04-19
x64: Add AVX-encoded versions of `ucomis{s,d}` and `ptest`

alexcrichton:
This PR finishes the support of AVX instructions in Cranelift (to my knowledge at least). These last few instrucitons were omitted historically because I never got around to sorting the story with comparisons. That's now sorted out, leading to this PR.

Note that this PR is broken into separate commits which lead up to and enable the final change to "drop in" AVX support.


fitzgen (approved):
LGTM!

PR #8410, opened 2024-04-19
Fix use-after-free in externref example

alexcrichton:
This fixes a typo in the `externref.c` example where a deallocated `wasmtime_val_t` was used by accident. Additionally this introduces scoping to prevent this from arising again.

### DIFF
--- a/examples/externref.c
+++ b/examples/externref.c
@@ -101,11 +101,13 @@ int main() {
   void *data = wasmtime_externref_data(context, externref);
   assert(strcmp((char *)data, "Hello, World!") == 0);
 
-  printf("Touching `externref` table...\n");
-
   wasmtime_extern_t item;
+  wasmtime_val_t externref_val;
+  externref_val.kind = WASMTIME_EXTERNREF;
+  externref_val.of.externref = externref;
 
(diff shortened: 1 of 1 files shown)


examples/externref.c
```diff
@@ -106,71 +106,78 @@ int main() {
...
-                "Hello, World!") == 0);
-  wasmtime_val_delete(context, &elem);
+  {
+    ok = wasmtime_instance_export_get(context, &instance, "table",
+                                      strlen("table"), &item);
+    assert(ok);
+    assert(item.kind == WASMTIME_EXTERN_TABLE);
+
+    // Set `table[3]` to our `externref`.
+    wasmtime_val_t externref_val;
```
fitzgen, line 116: This is now declared inside this block, but then reused in the other block and has `wasmtime_val_delete` called outside of the block. Can we move it up, out of this block then?

fitzgen (approved):

examples/externref.c
alexcrichton: Wait what how does this compile given this mistake...

fitzgen: lol yeah I was wondering myself

PR #8415, opened 2024-04-19
Restrict the preview1 implementation of fd_read to one iovec

elliottt:
In the wasi-common implementation of `fd_read` for preview1, switch to only reading the first iovec given in a vectored read. This avoids issues with Wiggle's runtime borrow-checker implemented in #8277, which would cause the reads to unconditionally fail when two mutable borrows were taken out on the same memory.

While it's not as efficient to perform a single read at a time, it is also valid, as read never promises to read all of the vectors provided.

### DIFF
--- a/crates/test-programs/src/bin/preview1_file_read_write.rs
+++ b/crates/test-programs/src/bin/preview1_file_read_write.rs
@@ -1,7 +1,7 @@
 use std::{env, process};
 use test_programs::preview1::open_scratch_directory;
 
-unsafe fn test_file_pread_pwrite(dir_fd: wasi::Fd) {
+unsafe fn test_file_read_write(dir_fd: wasi::Fd) {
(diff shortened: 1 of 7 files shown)


alexcrichton:
I was initially confused reading over this trying to piece together how this wasn't covered by https://github.com/bytecodealliance/wasmtime/pull/8353 but that was `fd_pread` rather than `fd_read`.

Mind adding a test for this? Also, can you update the filter here to be the same as https://github.com/bytecodealliance/wasmtime/pull/8353 where it searches for the first non-empty buffer rather than the first buffer?

elliottt:
I added a test, and verified that it failed without the fix to `fd_read`. (It's the same as the pread/pwrite test, but using `fd_seek` to polyfill their behavior.)

alexcrichton (approved):

PR #8421, opened 2024-04-20
cranelift/x64: Narrow `test` immediate operands

jameysharp:
The x86 `test` instruction does a bitwise-and operation, setting flags, but otherwise discarding the result. And when one operand is an immediate constant, any zero bits in the constant can't have an effect on the result. So we can emit shorter versions of this instruction by truncating the most significant zero bits and using the narrowest possible immediate form for the provided constant.

I think this has merge conflicts with #8408 so I'll rebase it next week.


jameysharp:
It's not clear to me whether I made the right choice to make this change in the code for emitting into the machbuffer, rather than doing it earlier when building vcode.

It's a little weird that the machbuffer disassembly doesn't match the vcode disassembly for an instruction as simple as `test`.

On the other hand this means that Winch gets this optimization too, and I think it's simpler to implement this in Rust than in ISLE right now.

So I'd be interested in other people's opinions on this: @cfallin @alexcrichton @elliottt?

alexcrichton:
Personally I think I'd prefer to do this in ISLE, but I don't feel too strongly really. Given that this is already done here it seems reasonable to land to me

cfallin:
One reason to try to lift optimizations like this into ISLE, and keep the VCode insts as close to 1-to-1 correspondence with machine code as possible, is that it makes verification more tractable: rather than writing a specification for the vcode inst that is "may do this, or that, depending on the constant", we have that logic exposed in ISLE left-hand sides where we can reason about it. For that reason I think I'd prefer for us to do that as well, unless it turns out that it's not really possible or practical at all. (There's definitely room for compromise here and some VCode insts definitely are smarter than they should be! This is more of an aspirational direction for current and future work.)

PR #8422, opened 2024-04-20
Fix handling of large branch values in cranelift_frontend::Switch

bjorn3:
Previously trying to branch on a i128 with a branch value larger than i64::MAX after truncating to u64 would result in the branch value incorrectly getting interpreted as a value with the upper 64 bits being 1 rather than 0.

Fixes https://github.com/rust-lang/rustc_codegen_cranelift/issues/1484

### DIFF
--- a/cranelift/frontend/src/switch.rs
+++ b/cranelift/frontend/src/switch.rs
@@ -289,8 +289,11 @@ impl Switch {
 }
 
 fn icmp_imm_u128(bx: &mut FunctionBuilder, cond: IntCC, x: Value, y: u128) -> Value {
-    if let Ok(index) = u64::try_from(y) {
-        bx.ins().icmp_imm(cond, x, index as i64)
+    if bx.func.dfg.value_type(x) != types::I128 {
+        assert!(u64::try_from(y).is_ok());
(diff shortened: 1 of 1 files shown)


cranelift/frontend/src/switch.rs
```diff
@@ -289,8 +289,11 @@ impl Switch {
 }
 
 fn icmp_imm_u128(bx: &mut FunctionBuilder, cond: IntCC, x: Value, y: u128) -> Value {
-    if let Ok(index) = u64::try_from(y) {
-        bx.ins().icmp_imm(cond, x, index as i64)
+    if bx.func.dfg.value_type(x) != types::I128 {
+        assert!(u64::try_from(y).is_ok());
+        bx.ins().icmp_imm(cond, x, y as i64)
```
elliottt, line 294: This assert/conversion combination is a little surprising. If `y` converts to `u64`, doesn't that mean that there's a possibility that it would produce a negative `i64` value?

bjorn3: A negative i64 here is fine as the value on which we are operating is at most a 64bit integer and thus no implicit sign extension of the immediate will be done by `icmp_imm`.

elliottt (approved):

PR #8423, opened 2024-04-20
cranelift/codegen: move `get_isle_compilations` to meta crate

mmcloughlin:
This PR moves the `get_isle_compilations` function from the codegen `build.rs` into
the `meta` crate.

The purpose of this change is to enable sharing the ISLE compilations configuration with
other ISLE tooling.

Note this also makes a minor change that `get_isle_compilations` does not query the
current directory itself, instead expecting the caller to do it and pass in appropriate
full paths.


cfallin (approved):
Seems reasonable; LGTM!

PR #8425, opened 2024-04-21
winch(aarch64): abs, ceil, floor, nearest, trunc, neg, sqrt

evacchi:
Adds a few more float instructions abs, neg, sqrt and all roundings (ceil, floor, nearest, trunc).

I skipped copysign for now, as it is less trivial and we probably want to port the same logic from cranelift :p

Temporarily submitting as draft because I want to double-check the test cases.


saulecabrera (commented):
The general direction looks good to me, thanks! I'll give it a more in depth review once it's marked as ready.

evacchi:
I think it should be good to go

PR #8427, opened 2024-04-22
cranelift/x64: Optimize i128 comparisons

jameysharp:
Inequality comparisons between i128 values were previously eight instructions and this reduces them to two, plus one move if one of the inputs is still live afterward.

Equality comparisons were six instructions and are now three, plus up to two moves if both inputs are still live afterward.

This removes 45 instructions from the test in x64/i128.clif that generates all possible i128 comparisons. In addition to using fewer instructions for each comparison, it also reduces register pressure enough that the function no longer spills.

Conditional branches on i128 values are a special case but similar optimizations shrink them from six instructions to two.

This brings Cranelift in line with what rustc+LLVM generates for equivalent 128-bit comparisons.

This PR probably conflicts with #8421 in the filetest output; I'll rebase whichever one doesn't land first.

### DIFF
--- a/cranelift/codegen/src/isa/x64/inst.isle
+++ b/cranelift/codegen/src/isa/x64/inst.isle
@@ -2514,6 +2514,15 @@
                         dst)
          dst)))
 
+(decl x64_alurmi_flags_side_effect (AluRmiROpcode Type Gpr GprMemImm) ProducesFlags)
+(rule (x64_alurmi_flags_side_effect opc (fits_in_64 ty) src1 src2)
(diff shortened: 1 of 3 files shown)


fitzgen (approved):
Do we cover this in run tests and/or the fuzzer? Seems like a good time to double check that we do.

jameysharp:
We do cover this in runtests, and I know this because I got it wrong several times in ways that the runtests caught. :grin: 

I think we also cover it in the Cranelift fuzz targets but I don't remember for sure.

Issue #8433, opened 2024-04-22
Inconsistent GC behavior calling a host function returning an externref


alexcrichton:
For this program:

```rust
use wasmtime::*;

fn main() {
    let engine = Engine::default();

    let module = Module::new(
        &engine,
        r#"
            (module
                (import "" "a" (func $a (result externref)))

                (func (export "a")
                    (loop $l
                        call $a
                        drop
                        br $l
                    )
                )
            )
        "#,
    )
    .unwrap();

    let mut linker = Linker::new(&engine);
    linker
        .func_wrap("", "a", |mut caller: Caller<'_, ()>| {
            Ok(Some(ExternRef::new(&mut caller, 100)?))
        })
        .unwrap();

    let ty = FuncType::new(&engine, [], [ValType::EXTERNREF]);
    linker
        .func_new("", "b", ty, |mut caller, _, results| {
            results[0] = Some(ExternRef::new(&mut caller, 100)?).into();
            Ok(())
        })
        .unwrap();

    let mut store = Store::new(&engine, ());
    let i = linker.instantiate(&mut store, &module).unwrap();
    let f = i.get_typed_func::<(), ()>(&mut store, "a").unwrap();
    f.call(&mut store, ()).unwrap();
}
```

it will, as-is, run with:

```
$ cargo run -q
thread 'main' panicked at src/main.rs:42:28:
called `Result::unwrap()` on an `Err` value: error while executing at wasm backtrace:
    0:   0x2c - <unknown>!<wasm function 1>

Caused by:
    0: failed to allocate `externref`
    1: GC heap out of memory
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

However if the `"a"` is switched to `"b"` to use the `func_new` definition instead of `func_wrap`, it will run infinitely and have steady memory usage.

The reason is that [this block](https://github.com/bytecodealliance/wasmtime/blob/4b9f53a928a8721e9b955da1c50090086bbef112/crates/wasmtime/src/runtime/func.rs#L1349-L1365) is only present in the `func_new` path, not the `func_wrap` path.

cc @fitzgen

alexcrichton:
[...]

PR #8434, opened 2024-04-22
Fix a panic calling host functions with refs in async mode

alexcrichton:
This commit fixes a panic when a host function defined with `Func::new` returned GC references and was called in async mode. The logic to auto-gc before the return values go to wasm asserted that a synchronous GC was possible but the context this function is called in could be either async or sync. The fix applied in this commit is to remove the auto-gc. This means that hosts will need to explicitly GC in these situations until auto-gc is re-added back to Wasmtime.

cc #8433 as this will make the behavior consistent, but we'll want to re-add the gc behavior.


alexcrichton:
I'll also note that I plan on backporting this to the 20.0.0 release to happen today too.

fitzgen (approved):

PR #4571, opened 2022-08-01
Cranellift: remove Baldrdash support and related features.

cfallin:
As noted in Mozilla's bugzilla bug 1781425 [1], the SpiderMonkey team
has recently determined that their current form of integration with
Cranelift is too hard to maintain, and they have chosen to remove it
from their codebase. If and when they decide to build updated support
for Cranelift, they will adopt different approaches to several details
of the integration.

In the meantime, after discussion with the SpiderMonkey folks, they
agree that it makes sense to remove the bits of Cranelift that exist
to support the integration ("Baldrdash"), as they will not need
them. Many of these bits are difficult-to-maintain special cases that
are not actually tested in Cranelift proper: for example, the
Baldrdash integration required Cranelift to emit function bodies
without prologues/epilogues, and instead communicate very precise
information about the expected frame size and layout, then stitched
together something post-facto. This was brittle and caused a lot of
incidental complexity ("fallthrough returns", the resulting special
logic in block-ordering); this is just one example. As another
example, one particular Baldrdash ABI variant processed stack args in
reverse order, so our ABI code had to support both traversal
orders. We had a number of other Baldrdash-specific settings as well
that did various special things.

This PR removes Baldrdash ABI support, the `fallthrough_return`
instruction, and pulls some threads to remove now-unused bits as a
result of those two, with the  understanding that the SpiderMonkey folks
will build new functionality as needed in the future and we can perhaps
find cleaner abstractions to make it all work.

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1781425

PR #8438, opened 2024-04-22
cranelift: Drop unused arguments before regalloc

jameysharp:
Before this, Cranelift ABI code would emit a stack-load instruction for every stack argument and add all register arguments to the `args` pseudo-instruction, whether those arguments were used or not.

However, we already know which arguments are used at that point because we need the analysis for load-sinking, so it's easy to filter the unused arguments out.

This avoids generating loads that are immediately dead, which is good for the generated code. It also slightly reduces the size of the register allocation problem, which is a small win in compile time.

This also changes which registers RA2 chooses in some cases because it no longer considers unused defs from the `args` pseudo-instruction.

There was an existing method named `arg_is_needed_in_body` which sounded like it should be the right place to implement this. However, that method was only used for Baldrdash integration and has been a stub since that integration was removed in #4571. Also it didn't have access to the `value_ir_uses` map needed here. But the place where that method was called does have access to that map and was perfect for this.

Thanks to @elliottt for doing the initial investigation of this change with me.


cfallin (approved):
Super-clean precision change for a nice result -- thanks very much for thinking through this! I'm excited to see the useless loads go away in a bunch of functions.

jameysharp:
So `cargo test -p wasmtime engine::serialization::test::cache_accounts_for_opt_level` fails with this PR. The panic message is:

```
thread '<unnamed>' panicked at cranelift/codegen/src/machinst/compile.rs:76:14:
register allocation: SSA(VReg(vreg = 192, class = Int), Inst(1))
```

This test simply compiles `(module (func))` several times with different optimization settings and with Wasmtime's cache enabled.

From experimentation, this problem only seems to occur if Wasmtime's debug-info is enabled. I guess this is the only test that runs in that mode.

I thought `emit_value_label_markers_for_block_args`, or the special case to call `emit_value_label_marks_for_value` on the vmctx parameter, seemed like likely candidates for telling RA2 to use an argument that's not otherwise used. But commenting out those calls doesn't fix this.

I'm not sure how else to investigate this; any suggestions?

cfallin:
Can you dump the vcode from the failure? The most pertinent bit is figuring out what is still using the argument, I guess (as you suggest)...

PR #8440, opened 2024-04-22
Add a fuzzer for async wasm

alexcrichton:
This commit revives a very old branch of mine to add a fuzzer for Wasmtime in async mode. This work was originally blocked on llvm/llvm-project#53891 and while that's still an issue it now contains a workaround for that issue. Support for async fuzzing required a good deal of refactorings and changes, and the highlights are:

* The main part is that new intrinsics, `__sanitizer_{start,finish}_fiber_switch` are now invoked around the stack-switching routines of fibers. This only works on Unix and is set to only compile when ASAN is enabled (otherwise everything is a noop). This required refactoring of things to get it all in just the right way for ASAN since it appears that these functions not only need to be called but more-or-less need to be adjacent to each other in the code. My guess is that while we're switching ASAN is in a "weird state" and it's not ready to run arbitrary code.

* Stacks are a problem. The above issue in LLVM outlines how stacks cannot be deallocated at this time because if the deallocated virtual memory is later used for the heap then ASAN will have a false positive about stack overflow. To handle this stacks are specially handled in asan mode by using a special allocation path that never deallocates stacks. This logic additionally applies to the pooling allocator which uses a different stack allocation strategy with ASAN.

With all of the above a new fuzzer is added. This fuzzer generates an arbitrary module, selects an arbitrary means of async (e.g. epochs/fuel), and then tries to execute the exports of the module with various values. In general the fuzzer is looking for crashes/panics as opposed to correct answers as there's no oracle here. This is also intended to stress the code used to switch on and off stacks.


crates/fiber/src/unix.rs
```diff
@@ -29,28 +29,97 @@
...
+        // allocation strategy.
+        if cfg!(asan) {
+            return Self::from_custom(asan::new_fiber_stack(size)?);
+        }
+        let page_size = rustix::param::page_size();
+        let stack = MmapFiberStack::new(size)?;
+
+        Ok(FiberStack {
+            base: stack.mapping_base.wrapping_byte_add(page_size),
+            len: stack.mapping_len - page_size,
```
fitzgen, line 65: This is to leave room for a guard page? Can we have a comment to that effect?

fitzgen (approved):

PR #8442, opened 2024-04-23
Enable unit test for wasi-nn WinML backend.

### LINKED ISSUE #8391: Inconsistent results for wasi-nn different backends
### Test Case

[nn_image_classification](https://github.com/bytecodealliance/wasmtime/blob/main/crates/test-programs/src/bin/nn_image_classification.rs), [nn_image_classification_named](https://github.com/bytecodealliance/wasmtime/blob/main/crates/test-programs/src/bin/nn_image_classification_named.rs) and [nn_image_classification_onnx](https://github.com/bytecodealliance/wasmtime/blob/main/crates/test-programs/src/bin/nn_image_classification_onnx.rs) have different inference results. The first two cases are based on openvino backend, while the third one is based on onnxruntime backend.

### Steps to Reproduce

Run the cases above locally or check the output of GitHub Actions.

An example output:
https://github.com/bytecodealliance/wasmtime/actions/runs/8716252474/job/23909494323#step:17:3914

### Expected Results

Since both of them use mobilenet v2 and the same input tensor data, they should have similar results. We don't expect them to be exactly the same because of different model formats.

### Actual Results

The result of wasi-nn openvino backend is `[InferenceResult(963, 0.7113049), InferenceResult(762, 0.07070768), InferenceResult(909, 0.036356032), InferenceResult(926, 0.015456118), InferenceResult(567, 0.015344023)]`.

The result for onnx backend is `[InferenceResult(470, 479.08182), InferenceResult(862, 378.7252), InferenceResult(626, 364.8759), InferenceResult(644, 334.28488), InferenceResult(556, 288.65884)]`.

CI for WinML backend is not enabled yet, but it has the same result as onnx backend.

### Versions and Environment

Wasmtime version or commit: 19.0.1

Operating system: Windows

Architecture: x86_64

### Extra Info

Although all these tests use mobilenet v2, openvino model requires input data to be BGR format, with mean values: [127.5, 127.5, 127.5]([openvino model description](https://github.com/openvinotoolkit/open_model_zoo/tree/master/models/public/mobilenet-v2-1.0-224)). ONNX model (used by both onnxruntime backend and winml backend) requires input data to be RGB format, in the range of [0, 1] and then normalized using mean = [0.485, 0.456, 0.406] and std = [0.229, 0.224, 0.225] ([onnx model description](https://github.com/onnx/models/tree/main/validated/vision/classification/mobilenet)).

I'm not sure if we missed input data preprocessing, or it was processed somewhere else.

The test case for WinML backend uses a different input at this time. I'm trying to unifying the inputs for all backends so we can double check the correctness.

jianjunz:
This test was disabled because GitHub Actions Windows Server image doesn't have desktop experience included. But it looks like we can have a standalone WinML binary downloaded from ONNX Runtime project.

Wasi-nn WinML backend and ONNX Runtime backend now share the same test code since they accept the same input, and they are expected to produce the same result. Pre-processing and post-processing are added to nn_image_classification_onnx for improving its accuracy.

This change also make wasi-nn WinML backend as a default feature as it's covered by test now.

Fixes #8391.


crates/test-programs/Cargo.toml
```diff
@@ -20,3 +20,6 @@ futures = { workspace = true, default-features = false, features = ['alloc'] }
 url = { workspace = true }
 sha2 = "0.10.2"
 base64 = "0.21.0"
+# image and ndarray are used by nn_image_classification_onnx for image preprocessing.
+image = { version = "0.24.6", default-features = false, features = ["jpeg"] }
```
devigned, line 24: We had avoided adding image and ndarray dependencies in favor of using the raw image used in the openvino test. What you did with the dog image is pretty close to what I had originally. @abrown, WDYT?

devigned (commented):
Before a detailed review, we need to determine if adding these dependencies are ok.

crates/test-programs/Cargo.toml
jianjunz: Image processing and post-processing are copied from your classification-component-onnx example. If new dependencies are not allowed, we can do preprocessing offline, just put a RGB file here.

devigned: Personally, I like showing the image prep / processing as it provides guidance to folks learning how to get started. In fact, it'd probably be helpful to add more comments / documentation to samples to explain what is happening. I think we, as practitioners, may be too accustomed to the transformations we regularly do to realize how odd they may appear to the naive observer.

jianjunz: My original idea was to use an image as input for all backends, so we can cross check the results for correctness.

To avoid introducing extra dependencies, the image is replaced with processed tensor data (000000062808.rgb). I believe this is the same image for openvino backend ([tensorization for openvino](https://github.com/intel/openvino-rs/blob/7efb30d207e874d1dbf727bb52949a343d726567/crates/openvino/tests/fixtures/mobilenet/build.sh#L42)).

devigned (approved):
LGTM! Tests results are the same across the impls. 

Thank you for documenting how the raw image was created. Great work! 

@abrown tag, you're it

abrown (approved):
LGTM!

Issue #8446, opened 2024-04-23
Winch runtime segmentation fault in wasmtime 20.0.0


TerrorJack:
Reproducer: https://files.catbox.moe/jbsi1b.wasm

When `wasmtime-cli-v20.0.0` is built with `winch` on x86_64-linux and the above file is run with `wasmtime run -C compiler=winch --env PWD=/ --dir "$PWD"::/ -W tail-call -- hello.wasm`, then it fails with a segmentation fault. Remove `-C compiler=winch` and it properly runs without any issue. The module above does not contain any opcodes unsupported by winch at this time.

saulecabrera:
Thanks for the report! 

A couple of questions that might help diagnose the root cause:
- The Wasm binary is ~25MB in size, have you tried reducing the failure with `wasm-tools shrink` as outlined here https://github.com/bytecodealliance/wasmtime/blob/main/docs/contributing-reducing-test-cases.md?
- Does the Wasm program need access to the current working directory? (e.g. the command in the report is making use of `--dir`) If so, could you speak a bit more about what are the expectations between the program and that directory? This might make it easier to reduce the program and reproduce the bug in isolation. FWIW, I tried on a Linux machine, without the `--env` and without the `--dir` arguments and I get the following result:
```
main = putStrLn "hello world"
```

saulecabrera:
I was able to reproduce, only when the `--env`, `--dir` and `-W tail-call` params are specified (i.e. when  `-W tail-call` is dropped, it doesn't segfault), which makes me think that it has to do with the tail call calling convention and its support in Winch.

saulecabrera:
FYI, I've opened https://github.com/bytecodealliance/wasmtime/pull/8447, to address this issue.

PR #8447, opened 2024-04-23
winch: Handle -W tail-call

saulecabrera:
Fixes https://github.com/bytecodealliance/wasmtime/issues/8446

The WebAssembly tail call proposal is currently not supported in Winch. This commit returns an error when trying to enable the tail call proposal while using Winch as the compiler. 

Even though the issue linked above doesn't make use of any of the tail instructions, the trampolines were generated using Cranelift's tail call calling convention.


alexcrichton (approved):

alexcrichton:
> The hosted runner encountered an error while running your job. (Error Type: Disconnect).

alexcrichton:
Not a failure I've seen before...

Issue #8382, opened 2024-04-16
Changing the return type of `add_to_linker` closures


lann:
[From this Zulip discussion](https://bytecodealliance.zulipchat.com/#narrow/stream/217126-wasmtime/topic/Return.20type.20of.20.60add_to_linker.60.20closures):

I've been experimenting with ways to provide reusable bits of host functionality and have run into a stumbling block with wasmtime's generated `add_to_linker` functions, which take `get`closures with types (roughly) like:
```rust
get: impl Fn(&mut T) -> &mut U + Send + Sync + Copy + 'static
    ...
    where U: Host,
```
These closures are usually trivial field accesses like `|data| &mut data.some_host_impl`, but this is somewhat limiting because you can only return `&mut`s borrowing from `T`. As a specific example, the wasmtime-wasi [`WasiView::table`](https://docs.rs/wasmtime-wasi/19.0.2/wasmtime_wasi/trait.WasiView.html#tymethod.table) needs to be shared with wasmtime-wasi-http's [`WasiHttpView::table`](https://docs.rs/wasmtime-wasi-http/latest/wasmtime_wasi_http/types/trait.WasiHttpView.html#tymethod.table), but this currently requires both `*View` implementations to use the same concrete type within `T`, making the two host implementations uncomposable.

There are some `unsafe` approaches to deal with this but I'm wondering if it would be possible to change the `add_to_linker` closure signature to permit returning non-ref types, e.g. something that `impl`s `AsMut<U>`. I've been trying to find a way to do this backward-compatibility but I'm running into the limits of my understanding of closure lifetime

lann:
> @alexcrichton:
> I can definitely say that the `&mut T -> &mut U` closures I've never felt great about. I've always preferred to do `&mut T -> U` with trait bounds on `U`, which I think is what you're getting at as well. I've also tried to stretch my Rust knowledge of closures to achieve this and while it's possible it gets really nasty really quickly.
> 
> The next-closest approximation I've thought of is to allow returning `&mut dyn Trait` but I haven't gotten around to pursuing that much.
>
> I was experimenting with this to clean up wasmtime-wasi APIs a bit ago but I seem to have misplace the branch
>
> I [asked about this on the Rust user forums](https://users.rust-lang.org/t/generic-closure-returns-that-can-capture-arguments/76513) a long time ago too

alexcrichton:
[...]

PR #8448, opened 2024-04-23
Add `GetHost` to generated bindings for more flexible linking

lann:
Work in progress for #8382

### DIFF
--- a/crates/component-macro/tests/expanded/char.rs
+++ b/crates/component-macro/tests/expanded/char.rs
@@ -77,12 +77,21 @@ pub mod foo {
                 /// A function that returns a character
                 fn return_char(&mut self) -> char;
             }
-            pub trait GetHost<T>: Send + Sync + Copy + 'static {
-                fn get_host<'a>(&self, data: &'a mut T) -> impl Host;
+            pub trait GetHost<
+                T,
+            >: Fn(T) -> <Self as GetHost<T>>::Output + Send + Sync + Copy + 'static {
+                type Output: Host;
+            }
+            impl<F, T, O> GetHost<T> for F
+            where
+                F: Fn(T) -> O + Send + Sync + Copy + 'static,
+                O: Host,
+            {
+                type Output = O;
             }
             pub fn add_to_linker_get_host<T>(
                 linker: &mut wasmtime::component::Linker<T>,
-                host_getter: impl GetHost<T>,
+                host_getter: impl for<'a> GetHost<&'a mut T>,
             ) -> wasmtime::Result<()> {
                 let mut inst = linker.instance("foo:foo/chars")?;
                 inst.func_wrap(
@@ -91,7 +100,7 @@ pub mod foo {
                         mut caller: wasmtime::StoreContextMut<'_, T>,
                         (arg0,): (char,)|
                     {
-                        let host = &mut host_getter.get_host(caller.data_mut());
+                        let host = &mut host_getter(caller.data_mut());
                         let r = Host::take_char(host, arg0);
                         Ok(r)
                     },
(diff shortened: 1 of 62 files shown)


lann:
Tracking for myself: next step is to update the top-level world linking: add a e.g. `add_root_to_linker_get_host`, revert the world `add_to_linker` impl to call child trait closure-form `add_to_linker`s directly (avoiding the need for `WorldGetHost` to somehow impl all imported interface/resource `GetHost`s as well).

lann:
@alexcrichton 
re: https://github.com/bytecodealliance/wasmtime/issues/8382#issuecomment-2072986390 (`?Sized` for `impl <T: Host> Host for &mut T`), this doesn't work for methods that return the type itself:

```
27 | impl<T: FooImports + ?Sized> FooImports for &mut T {
   |      - this type parameter needs to be `Sized`
28 |     fn x(&mut self) -> T {
   |                        ^ doesn't have a size known at compile-time
```

I'm just going to drop the `?Sized` unless you think its worth trying to special case somehow.

alexcrichton:
Is that actually the type parameter `T` shadowing a WIT type `T`? I don't think that any WIT api should return `Self`, and that might be fixable by renaming the type parameter to `_T` perhaps?

lann:
Ah, you are correct. That explains why I couldn't figure out where this would be happening...

alexcrichton:
This has run into a bit of a snag. https://github.com/rust-lang/rust/issues/124984 shows code this PR was going to generate which compiled with Rust 1.77.0 but does not compile with Rust 1.78.0.

alexcrichton:
Ok I think most tests should be passing now. A summary of changes now at its current state:

Previously `bindgen!` would generate:

```rust
            pub fn add_to_linker<T, U>(
                linker: &mut wasmtime::component::Linker<T>,
                get: impl Fn(&mut T) -> &mut U + Send + Sync + Copy + 'static,
            ) -> wasmtime::Result<()>
            where
                U: Host,
```

Now it generates:

```rust
            pub trait GetHost<
                T,
            >: Fn(T) -> <Self as GetHost<T>>::Output + Send + Sync + Copy + 'static {
                type Output: Host;
            }

            impl<F, T, O> GetHost<T> for F
            where
                F: Fn(T) -> O + Send + Sync + Copy + 'static,
                O: Host,
            {
                type Output = O;
            }

            pub fn add_to_linker_get_host<T>(
                linker: &mut wasmtime::component::Linker<T>,
                host_getter: impl for<'a> GetHost<&'a mut T>,
            ) -> wasmtime::Result<()> {
                // ...
            }

            pub fn add_to_linker<T, U>(
                linker: &mut wasmtime::component::Linker<T>,
                get: impl Fn(&mut T) -> &mut U + Send + Sync + Copy + 'static,
            ) -> wasmtime::Result<()>
            where
                U: Host,
            {
                add_to_linker_get_host(linker, get)
            }
```

Effectively the old `add_to_linker` is still there but everything is actually built on a more-general `GetHost` trait which enables not being requried to use literally `&mut U` but instead you can use something like `MyStruct<'a>`. Usage of `add_to_linker_get_host` will probably require type-annotating closures. For example wasmtime-wasi contains:


```rust
    fn type_annotate<T, F>(val: F) -> F
    where
        F: Fn(&mut T) -> &mut T,
    {
        val
    }
    let closure = type_annotate::<T, _>(|t| t);
```

Usage of `add_to_linker` will not require type annotations over what's already required today.

I would like to, in a future commit I want to refactor all of wasmtime-wasi{,-http} to remove `imp<T: WasiView> Host for T` and replace that with `impl Host for dyn WasiView` using this support. That'll enable removing the odd `skip_mut_forwarding_impls` I've added to `bindgen!` here which is just to get WASI bits as-is compiling.

alexcrichton:
I should also say there are other minor details here and there but the comment above is the main thrust of the change, everything else is "smith a few things until everything compiles again"

crates/component-macro/tests/expanded/char.rs
```diff
@@ -77,35 +77,63 @@ pub mod foo {
...
                 fn return_char(&mut self) -> char;
             }
-            pub fn add_to_linker<T, U>(
-                linker: &mut wasmtime::component::Linker<T>,
-                get: impl Fn(&mut T) -> &mut U + Send + Sync + Copy + 'static,
-            ) -> wasmtime::Result<()>
+            pub trait GetHost<
+                T,
+            >: Fn(T) -> <Self as GetHost<T>>::Output + Send + Sync + Copy + 'static {
+                type Output: Host;
```
lann, line 83: Slightly confusing in the definition but maybe more obvious for implementers?

```suggestion
                type Host: Host;
```

crates/component-macro/tests/expanded/char.rs
lann, line 1: Side note: we should pick a small subset of these tests with good feature coverage to expand. As-is these are useful for review but its hard to know how many/which files to review before stopping, and having so many changed files screws with Github's UI...

lann (commented):
Approach lgtm

crates/component-macro/tests/expanded/char.rs
alexcrichton: That's a good point yeah I've been a bit sad that the actual changes were all hidden by default as well...

alexcrichton (approved):

PR #5132, opened 2022-10-26
Generate more fixed_nonallocatable constraints, and add debug assertions

elliottt:
Add assertions to the `OperandCollector` that show we're not using pinned vregs, and use `reg_fixed_nonallocatable` constraints when a real register is used with other constraint generation functions like `reg_use` etc.

I experimented with adding debug assertions that the reused operand referenced in `reg_reuse_def` constraints was compatible with the new constraint that would be generated. This didn't end up being possible as some backends will recursively process instructions when collecting operands, making it impossible without more invasive changes to tell where that operand is in the operand vector. Given that we would need to plumb through somewhat substantial changes to support adding debug assertions, I decided that rephrasing the existing warning comment about real registers would be good enough.

PR #5121, opened 2022-10-25
s390x: Fix more regalloc checker errors

uweigand:
For VecInsertLane[Undef] and VecExtractLane, if lane_reg is zero_reg(), the instruction does not actually use any register value.

Fixes https://github.com/bytecodealliance/wasmtime/issues/5090

CC @cfallin

PR #4973, opened 2022-09-28
s390x: Fix regalloc checker error

uweigand:
For ShiftRR and VecShiftRR, if shift_reg is zero_reg(), the instruction does not actually use any register value.

Fixes https://github.com/bytecodealliance/wasmtime/issues/4969

PR #8449, opened 2024-04-23
cranelift/s390x: Use PRegs consistently

jameysharp:
When OperandCollector's reg_use/reg_late_use/reg_def/reg_early_def methods are handed a Reg that refers to a physical ("real") register, they all delegate to reg_fixed_nonallocatable, ignoring the constraint kinds and positions. This behavior was introduced in #5132.

In several cases, the s390x backend was calling those methods with the result of the `gpr` or `writable_gpr` functions, which return physical registers. In these cases we can be more explicit that this is a non-allocatable register.

In addition, this PR reverts #4973 and #5121 because they became unecessary due, again, to #5132.

This doesn't change any behavior but is a nice change to split out of a larger PR I'm working on.

cc: @uweigand


cfallin (approved):
LGTM, thanks!

PR #8450, opened 2024-04-23
Wasmtime(gc): Support `(ref.i31 (global.get $g))` const expressions

fitzgen:
We had something hacked together to support `(ref.i31 (i32.const N))`. It wasn't a long-term solution. This is the first time that we have to really deal with multi-instruction const expressions.

This commit introduces a tiny interpreter to evaluate const expressions.


crates/runtime/src/vmcontext.rs
```diff
@@ -434,6 +436,62 @@ impl VMGlobalDefinition {
...
+    ///
+    /// This raw value's type must match the given `WasmValType`.
+    pub unsafe fn from_val_raw(wasm_ty: WasmValType, raw: ValRaw) -> Self {
+        let mut global = Self::new();
+        match wasm_ty {
+            WasmValType::I32 => *global.as_i32_mut() = raw.get_i32(),
+            WasmValType::I64 => *global.as_i64_mut() = raw.get_i64(),
+            WasmValType::F32 => *global.as_f32_bits_mut() = raw.get_f32(),
+            WasmValType::F64 => *global.as_f64_bits_mut() = raw.get_f64(),
+            WasmValType::V128 => *global.as_u128_bits_mut() = raw.get_v128().to_ne_bytes(),
```
alexcrichton, line 451: If this `.to_ne_bytes()` is required I think that means we're doing something wrong in `ValRaw`, perhaps a missing `from_le_bytes` in `get_v128`?

No matter what if this is the only one that does `.to_ne_bytes()` is should have docs as to why it's omitted on the others.

crates/runtime/src/vmcontext.rs
```diff
@@ -434,6 +436,62 @@ impl VMGlobalDefinition {
...
+            WasmValType::Ref(r) => match r.heap_type {
+                WasmHeapType::Extern => ValRaw::externref(
+                    self.as_gc_ref()
+                        .map_or(0, |r| gc_store.clone_gc_ref(r).as_raw_u32()),
+                ),
+                WasmHeapType::Any | WasmHeapType::I31 | WasmHeapType::None => ValRaw::anyref(
+                    self.as_gc_ref()
+                        .map_or(0, |r| gc_store.clone_gc_ref(r).as_raw_u32()),
+                ),
+                WasmHeapType::Func | WasmHeapType::Concrete(_) | WasmHeapType::NoFunc => {
```
alexcrichton, line 488: Bit of a side comment, but I get a little uncomfortable with `Concrete(_)` here since it seems very easy to overlook this in the future when the concrete type may be a function or may be a struct. Is it possible to perhaps restructure this or add an assert? 

It seems especially relevant here and above given how unsafe the methods are

crates/types/src/lib.rs
```diff
@@ -247,6 +248,10 @@ impl WasmRefType {
     pub fn is_gc_heap_type(&self) -> bool {
         self.heap_type.is_gc_heap_type()
     }
+
+    pub fn is_vmgcref_type(&self) -> bool {
+        self.heap_type.is_vmgcref_type()
+    }
```
alexcrichton, line 254: Do you think we can structure this as an enum or similar? E.g. a `.classify()` returning an enum of some kind. The differences between these functions feels really subtle and easy to get wrong, and getting it wrong feels like it could lead to unsafety pretty easily

crates/types/src/lib.rs
```diff
@@ -856,29 +865,105 @@ impl TypeTrace for Global {
...
-/// Globals are initialized via the `const` operators or by referring to another import.
-#[derive(Debug, Clone, Copy, Hash, Eq, PartialEq, Serialize, Deserialize)]
-pub enum GlobalInit {
-    /// An `i32.const`.
+/// A constant expression.
+///
+/// These are used to initialize globals, table elements, etc...
+#[derive(Clone, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
+pub struct ConstExpr {
+    ops: SmallVec<[ConstOp; 8]>,
```
alexcrichton, line 873: 8 seems pretty big here given how often we'll be storing this, could this perhaps be tuned to be the native size of `SmallVec`?

alexcrichton (commented):
Looks good to me, I think the organization worked out well here.

My main comment is that the `unsafe `methods feel a little too unsafe and while I don't think there's a whole lot that can be done about that I still think it would be good to at least assert that `Concrete` heap types are functions for now to avoid accidental confusion in the future.

crates/types/src/lib.rs
fitzgen: WasmHeapType::Extern

crates/runtime/src/vmcontext.rs
fitzgen: I was just doing this to get a `[u8; 16]` but it turns out that isn't necessary: we have a method on `wasmtime_runtime::Global` to get the value as a `&mut u128`, which I didn't see before.

crates/runtime/src/vmcontext.rs
fitzgen: Unfortunately there isn't a good way to resolve this, given the `wasmtime` vs `wasmtime-runtime` crate split. We would need to take a `wasmtime::Engine` here and then look up the type index in the engine's types registry and match on the resulting value (which can only be a function type right now). But we can't get an engine inside the `wasmtime-runtime` crate.

alexcrichton: We have to solve this in the near future though right?

Originally I was hoping we could have `Concrete` renamed to `ConcreteFunc` and then also have `Concrete{Struct,Array}` or something like that, but I'm not sure if that's still possible.

Do you have other ideas though on what to do here once arrays/structs are implemented?

fitzgen: My plan was to start passing in a function to these sorts of methods that need to do the index-to-type resolution. Between that, fuzzing, and using LSP to find all uses of `Concrete(_)` I wasn't too too worried about missing a spot to update.

The other thing we could do is try moving the type registry into `wasmtime-runtime`.

crates/types/src/lib.rs
alexcrichton: Even without a const fn could this perhaps be smaller to start? A `ConstOp` is 32 bytes large and 8 of them inline here is 256 bytes for a single expression which seems pretty egregious especially if a table is initialized with a list of constant expressions.

I'm basically pretty fearful of the impact of using this everywhere when the representation is 8x larger than it is today

crates/runtime/src/vmcontext.rs
alexcrichton: I may be blowing things out of proportion here perhaps, but I'm personally pretty worried about this. I just ran across [another instance](https://github.com/bytecodealliance/wasmtime/blob/1f8c3df2d305c4e8e32529911bd97fe376f90bd7/crates/wasmtime/src/runtime/types.rs#L592-L597) of where type information will be needed as well.

I'm basically pretty worried that we've pretty deeply baked in the assumption that "`Concrete` is always a function" and I'd like to avoid baking it in any further, especially here where it's necessary for safety. 

I suppose for the sake of this PR could you add an assertion to the constructors of `Concrete` that it's always a function? (or double-check such an assertion already exists?) More-or-less we need to re-audit all usages of `Concrete` when the GC implementation gets further along.

crates/types/src/lib.rs
fitzgen: Yeah I dropped it to 2 locally, just haven't pushed the updates yet until I finish addressing the other comments.

crates/types/src/lib.rs
fitzgen: (We talked offline about this and decided to just rename the predicates).

crates/runtime/src/vmcontext.rs
fitzgen: (We talked offline about this and decided that renaming `Concrete(_)` to `ConcreteFunc(_)` actually is okay, and will do it in a future PR. It wasn't workable for `wasmparser`, since that crate needs to be able to parse opcodes without having a validated types section where we can look things up in and determine whether we are talking about a concrete func or e.g. struct, but that constraint doesn't apply here.)

alexcrichton (approved):

PR #8451, opened 2024-04-23
c-api: Remove allocations from `wasmtime_val_t`

alexcrichton:
This commit redesigns how GC references work in the C API. previously `wasmtime_{any,extern}ref_t` were both opaque pointers in the C API represented as a `Box`. Wasmtime did not, however, provide the ability to deallocate just the `Box` part. This was intended to be handled with unrooting APIs but unrooting requires a `wasmtime_context_t` parameter, meaning that destructors of types in other languages don't have a suitable means of managing the memory around the
`wasmtime_{any,extern}ref_t` which might lead to leaks.

This PR takes an alternate approach for the representation of these types in the C API. Instead of being an opaque pointer they're now actual structures with definitions in the header file. These structures mirror the internals in Rust and Rust is tagged to ensure that changes are reflected in the C API as well. This is similar to how `wasmtime_func_t` matches `wasmtime::Func`. This enables embedders to not need to worry about memory management of these values outside of the manual rooting otherwise required.

The hope is that this will reduce the likelihood of memory leaks and otherwise not make it any harder to manage references in the C API.

### DIFF
--- a/crates/c-api/src/ref.rs
+++ b/crates/c-api/src/ref.rs
@@ -1,5 +1,9 @@
 use crate::{abort, WasmtimeStoreContextMut};
-use std::{mem::MaybeUninit, os::raw::c_void, ptr};
+use std::{
+    mem::{ManuallyDrop, MaybeUninit},
+    os::raw::c_void,
+    ptr,
+};
 use wasmtime::{AnyRef, ExternRef, ManuallyRooted, Ref, RootScope, Val, I31};
 
 /// `*mut wasm_ref_t` is a reference type (`externref` or `funcref`), as seen by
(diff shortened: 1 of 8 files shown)


crates/c-api/src/ref.rs
```diff
@@ -182,25 +186,100 @@ pub extern "C" fn wasm_foreign_new(_store: &crate::wasm_store_t) -> Box<wasm_for
...
+pub union wasmtime_anyref_t {
+    store_id: u64,
+    rooted: ManuallyDrop<ManuallyRooted<AnyRef>>,
+}
+
+/// Same as `wasmtime_anyref_t`, but for extenref.
+pub union wasmtime_externref_t {
+    store_id: u64,
+    rooted: ManuallyDrop<ManuallyRooted<ExternRef>>,
+}
```
fitzgen, line 206: A `ManuallyRooted<T>` already contains a store ID, is there a way we can avoid duplicating it here?

fitzgen: Ah nevermind I see now that this is a union, not a struct, and that `store_id == 0` is used as `None`.

fitzgen: But, could we just make this `Option<ManuallyDrop<ManuallyRooted<ExternRef>>>` instead of a union?

crates/wasmtime/src/runtime/gc/enabled/rooting.rs
```diff
@@ -186,12 +186,19 @@ pub(crate) use sealed::*;
...
     store_id: StoreId,
     generation: u32,
     index: PackedIndex,
 }
 
+const _: () = {
+    // NB: these match the C API which should also be updated if this changes
+    assert!(std::mem::size_of::<GcRootIndex>() == 16);
+    assert!(std::mem::align_of::<GcRootIndex>() == 8);
+};
```
fitzgen, line 200: Maybe these asserts should be for `ManuallyRooted` rather than `GcRootIndex`, just so that if we add a field to the former, but leave the latter as-is, we get build errors because that mismatch would introduce unsafety.

fitzgen (approved):

crates/c-api/src/ref.rs
alexcrichton: That's a good point about `Option`, but one caveat is that embedders need to be able to make a null reference, test for one, and ideally without calling an API functions. That's why I opted for this representation where with `#[repr(C)]` we put the `store_id` first everywhere and can test it for zero here (and it's `NonZeroU64` in `StoreId`). 

So in that sense we could indeed use `Option` here, but it would rely on having a static assertion along the lines of:


```rust
    union T {
        a: u64,
        b: std::mem::ManuallyDrop<Option<ManuallyRooted<ExternRef>>>,
    }
    unsafe {
        assert!(
            T {
                b: std::mem::ManuallyDrop::new(None)
            }
            .a == 0
        );
    }
```

which, in testing, does in fact work. I'm sort of 50/50 though on whether this is worth it or "too clever by half"... What do you think, worth it to rely on that niche optimization?

fitzgen: I feel like this is worth doing, and keeping the union inside just the assertion instead of permeating the rest of the Rust code. I don't think this is too niche of an optimization because I was under the impression that it was "well known" that Rust would pack discriminants into non-zero and non-null things.

fitzgen: The other option here is to have a

```rust
impl ManuallyRooted<T> {
    #[doc(hidden)]
    pub fn into_raw_parts_for_c_api(&self) -> (u64, u32) { ... }
}
```

method and a dual `from_raw_parts_for_c_api` method.

fitzgen: That might be simpler and easier to maintain and ensure remains correct long term.

alexcrichton: Good point about the `#[doc(hidden)]` bits, I like that route too so I've implemented that

PR #8452, opened 2024-04-24
cranelift/x64: Stop cloning Imm8Reg wrappers

jameysharp:
A minor cleanup setting up for a larger PR later.


cfallin (approved):
Reasonable cleanup; would have also taken "make more things `Copy`" as a valid path to avoid the clones, probably it's all the same if the accessors are inlined though!

fitzgen (approved):

jameysharp:
> would have also taken "make more things `Copy`" as a valid path to avoid the clones

I actually want the borrows for later. :grin:

PR #8453, opened 2024-04-24
cranelift(aarch64): Add single precision fmov

saulecabrera:
This commit is a follow up to https://github.com/bytecodealliance/wasmtime/pull/8365/files#r1565962730, to enable emission of  32-bit fmov from Winch. 

I opted to introduce a new instruction over refactoring the existing `FpuMov64` to be more generic to keep things simple,  but I'm definitely open to exploring a refactoring if that's preferred. 

Encoding reference: https://developer.arm.com/documentation/ddi0602/2024-03/SIMD-FP-Instructions/FMOV--register---Floating-point-Move-register-without-conversion-?lang=en

### DIFF
--- a/cranelift/codegen/src/isa/aarch64/inst.isle
+++ b/cranelift/codegen/src/isa/aarch64/inst.isle
@@ -331,6 +331,11 @@
        ;; Consumption of speculative data barrier.
        (Csdb)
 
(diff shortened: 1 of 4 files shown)


cfallin (approved):
LGTM!

PR #8454, opened 2024-04-24
cranelift/riscv64: Idiomatic AMode operand collection

jameysharp:
Other backends have functions for passing any registers in an AMode to the operand collector. Use the same pattern in this backend too: it's simpler to use, and also sets up for more changes I'm planning.

### DIFF
--- a/cranelift/codegen/src/isa/riscv64/inst/args.rs
+++ b/cranelift/codegen/src/isa/riscv64/inst/args.rs
@@ -125,17 +125,21 @@ impl AMode {
         }
     }
 
-    /// Returns the registers that known to the register allocator.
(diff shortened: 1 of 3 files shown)


cranelift/codegen/src/isa/riscv64/inst/args.rs
```diff
@@ -127,15 +127,18 @@ impl AMode {
 
     /// Returns the registers that known to the register allocator.
```
cfallin, line 128: pre-existing typo but: s/that known/that are known/? or actually maybe refactor the sentence a bit e.g. "Provides operands within this AMode to the register allocator's operand collector."

cfallin (approved):

PR #8455, opened 2024-04-24
cranelift: Simplify recording vcode register aliases

jameysharp:
`Lower::set_vreg_alias` is only used in one place. Inlining it eliminates some borrow-checker challenges and allows the trace log to include more context.


cfallin (approved):
Nice cleanup!

PR #6957, opened 2023-09-02
Make MachineEnv a per-ABI property

uweigand:
The MachineEnv structure contains the allocatable and preferred register sets.  This is currently fixed per TargetIsa - however, conceptually these register sets can differ between ABIs on the same ISA.

To allow for this, replace the TargetIsa machine_env routine with an ABIMachineSpec get_machine_env routine.  To ensure the structure is still only allocated once, cache it via static OnceLock variables.

No functional change intended.

FYI @cfallin - this is what I suggested in last week's meeting.

PR #8457, opened 2024-04-24
cranelift: Don't recompute allocatable register set

jameysharp:
The implementation of `PRegSet::from(self.vcode.machine_env())` does a surprising amount of work: it lazily initializes a `OnceLock` in the backend, then loops over six vectors of registers and adds them all to the `PRegSet`.

We could likely implement that better, but in the meantime at least we can avoid repeating that work for every single machine instruction. The `PRegSet` itself only takes a few words to store so it's cheap to just keep it around.

I discovered this because when the call to `self.vcode.machine_env()` is in the middle of the loop, that prevents holding a mutable borrow on parts of `self.vcode`, which I want to be able to do in another PR.

### DIFF
--- a/cranelift/codegen/src/machinst/vcode.rs
+++ b/cranelift/codegen/src/machinst/vcode.rs
@@ -509,6 +509,7 @@ impl<I: VCodeInst> VCodeBuilder<I> {
     }
 
     fn collect_operands(&mut self) {
+        let allocatable = PRegSet::from(self.vcode.machine_env());
         for (i, insn) in self.vcode.insts.iter().enumerate() {
             // Push operands from the instruction onto the operand list.
             //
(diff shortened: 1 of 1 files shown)


cfallin:
This one really surprised me so I went looking at history to figure out if we ever measured it: it looks like first appeared in #6957; previously the `PRegSet` was computed once when the lowering backend was created; and we didn't benchmark that. The PR did note caching the `MachineEnv` but we missed caching the derived information.

Out of curiosity have you seen a perf improvement with this PR? (We should do it even if not, but it seems to me it should improve compile time)

cfallin (approved):
also I'll go ahead and r+ since I dug into this (sorry to preempt reviewers!).

jameysharp:
According to hyperfine, this commit is 1.02 ± 0.01 times faster when running `wasmtime compile -C cache=n ../sightglass/benchmarks/spidermonkey/benchmark.wasm`. That's actually a bigger effect than I expected since I thought operand collection is a pretty small part of the overall time.

PR #8458, opened 2024-04-24
cranelift: Simplify operand collector

jameysharp:
`OperandCollector` has a lot of very similar methods. This commit refactors them down to their bare essentials to make the similarities and differences obvious.


cfallin (approved):
Nice cleanup, thanks!

PR #8459, opened 2024-04-24
cranelift/riscv64: Inline multi-reg operand collector helpers

jameysharp:
The `reg_uses` and `reg_defs` helpers are not widely useful and have become quite rarely used over time. The last few remaining callers are all in the riscv64 backend. This commit replaces those calls with simple loops or unrolled calls to `reg_use` or `reg_def`, then deletes the helpers.


afonso360 (approved):
LGTM! 👍

PR #8460, opened 2024-04-24
winch(aarch64): Use single precision fmov

saulecabrera:
This commit is a follow-up to https://github.com/bytecodealliance/wasmtime/pull/8365/. In https://github.com/bytecodealliance/wasmtime/pull/8453 single precision fmov was introduced in Cranelift, so now we can make use of that instruction in Winch.


fitzgen (approved):
\o/

PR #8462, opened 2024-04-24
cranelift/codegen/meta: provide entry point to generate ISLE code only

mmcloughlin:
This PR reshuffles the code-generation functions in `cranelift::codegen::meta` to allow
generating the Rust and ISLE parts independently.

Following on from #8423, this is intended to support ISLE tooling that needs to inspect
the ISLE parts of the ISA lowering backends independently.


elliottt (approved):
Looks good to me!

Issue #8341, opened 2024-04-11
Add no_std support to Wasmtime


alexcrichton:
In this issue I'd like to propose officially adding support for Rust's `#![no_std]` mode and crate attribute to the `wasmtime` crate and runtime. Notably, I'm proposing that we refute [current documentation about "why not `no_std`"](https://docs.wasmtime.dev/stability-platform-support.html#what-about-no_std). This would additionally revert earlier work in https://github.com/bytecodealliance/wasmtime/pull/554 and https://github.com/bytecodealliance/wasmtime/pull/2024.

Before I go into more depth, this is not a new issue to Wasmtime. This has been discussed in a number of places such as https://github.com/bytecodealliance/wasmtime/issues/1158, https://github.com/bytecodealliance/wasmtime/issues/3451, https://github.com/bytecodealliance/wasmtime/issues/3495, https://github.com/bytecodealliance/wasmtime/issues/6424, https://github.com/bytecodealliance/wasmtime/issues/7700, and I'm sure I'm missing others as well. I'll also point out that I was personally one of the ones previously advocating for specifically not supporting `no_std`, and my opinion has changed in the intervening years.

### Why support `no_std`?

As far as I know the benefits of `no_std` haven't really ever been in question. After all, who doesn't want a project to be able to run in as many places as possible? To me I would personally rephrase this question as why to use `no_std` to support platform as opposed to alternative strategies. I'd answer this with the fact that `no_std` is the most idiomatic and well-supported solution in the Rust ecosystem. Rust developers in the embedded space are already used to `no_std` and what it entails. Additionally there are community idioms/expectations around the `no_std` feature to follow which set precedent.

### Why now?

For a number of years now Wasmtime has had a page in its documentation for "What about `#[no_std]`?", so I think a fair question is why would we revert this and reconsider this previous decision at this point in time. The embedded space has become more interested in WebAssembly over time and there are a fair number of users today. The general feedback is that Wasmtime is not suitable in these environments, and one primary reason is that it's difficult to get Wasmtime working in these environments. Work such as https://github.com/bytecodealliance/wasmtime/pull/7995 is to obscure for others to productively use as-is without being able to slot more idiomatically into the `no_std` ecosystem. More-or-less there's interest to use WebAssembly on platforms Wasmtime cannot easily target, and `no_std` support is intended to be a step towards making Wasmtime support on these platforms easier. 

### Was the previous rationale wrong?

[...]

PR #8463, opened 2024-04-24
Start migrating some Wasmtime crates to no_std

alexcrichton:
This commit is the first in what will be multiple PRs to migrate Wasmtime to being compatible with `#![no_std]`. This work is outlined in #8341 and the rough plan I have in mind is to go on a crate-by-crate basis and use CI as a "ratchet" to ensure that `no_std` compat is preserved. In that sense this PR is a bit of a template for future PRs.

This PR migrates a few small crates to `no_std`, basically those that need no changes beyond simply adding the attribute. The nontrivial parts introduced in this PR are:

* CI is introduced to verify that a subset of crates can indeed be built on a `no_std` target. The target selected is `x86_64-unknown-none` which is known to not have `std` and will result in a build error if it's attempted to be used.

* The `anyhow` crate, which `wasmtime-jit-icache-coherence` now depends on, has its `std` feature disabled by default in Wasmtime's workspace. This means that some crates which require `std` now need to explicitly enable the feature, but it means that by-default its usage is appropriate for `no_std`.

The first point should provide CI checks that compatibility with `no_std` indeed works, at least from an "it compiles" perspective. Note that it's not sufficient to test with a target like `x86_64-unknown-linux-gnu` because `extern crate std` will work on that target, even when `#![no_std]` is active.

The second point however is likely to increase maintenance burden in Wasmtime unfortunately. Namely we'll inevitably, either here or in the future, forget to turn on some feature for some crate that's not covered in CI checks. While I've tried to do my best here in covering it there's no guarantee that everything will work and the combinatorial explosion of what could be checked in CI can't all be added to CI. Instead we'll have to rely on bug fixes, users, and perhaps point releases to add more use cases to CI over time as we see fit.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -360,6 +360,19 @@ jobs:
     - run: cargo check -p wasmtime-c-api --no-default-features --features wat
     - run: cargo check -p wasmtime-c-api --no-default-features --features wasi
 
+    # Checks for no_std support, ensure that crates can build on a no_std
+    # target
+    - run: rustup target add x86_64-unknown-none
+    - run: cargo check -p wasmtime-jit-icache-coherence
+      env:
+        CARGO_BUILD_TARGET: x86_64-unknown-none
+    - run: cargo check -p wasmtime-component-util
+      env:
+        CARGO_BUILD_TARGET: x86_64-unknown-none
+    - run: cargo check -p wasmtime-asm-macros
+      env:
+        CARGO_BUILD_TARGET: x86_64-unknown-none
+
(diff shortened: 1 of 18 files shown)


alexcrichton:
I'll also note that, procedurally, my plan is to discuss https://github.com/bytecodealliance/wasmtime/issues/8341 in the Wasmtime meeting tomorrow and assuming no objections start landing PRs, but the goal here isn't to jump the gun by any means but rather just be prepared.

fitzgen (approved):

PR #8465, opened 2024-04-24
Rename `Concrete` to `ConcreteFunc`; introduce `WasmSubType` and `WasmCompositeType`

fitzgen:
This refactoring doesn't do anything on its own, but it sets the stage for the future addition of struct and array types.


alexcrichton (approved):
Thanks for doing this!

PR #8467, opened 2024-04-24
cranelift/x64: Allow collecting Gpr/Xmm operands

jameysharp:
Previously, the operand collector required its arguments to be of type Reg. But the x86-64 backend uses newtype wrappers around Reg to ensure that different register classes don't get mixed up, so it had to convert types everywhere to pass registers to the operand collector.

This commit allows the operand collector to accept any type which implements `Into<Reg>` instead. The newtype wrappers already implemented that, so we can mostly just delete the `to_reg` and `to_writable_reg` calls.

However, those calls were also automatically dereferencing borrows as needed, so to minimize the need for explicit dereferences I've also introduced implementations for converting from &Gpr/&Xmm to Reg. That doesn't help with the Writable-wrapped registers but it's good enough. I have a PR in progress which will clean it up the rest of the way.

### DIFF
--- a/cranelift/codegen/src/isa/x64/inst/args.rs
+++ b/cranelift/codegen/src/isa/x64/inst/args.rs
@@ -55,6 +55,12 @@ macro_rules! newtype_of_reg {
             }
         }
 
+        impl From<&$newtype_reg> for Reg {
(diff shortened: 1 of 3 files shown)


cfallin (approved):
So much cleaner; nice!

PR #8470, opened 2024-04-25
cranelift: Remove some unnecessary Callee state

jameysharp:
The set_num_spillslots and set_clobbered methods on Callee are only used to provide input to compute_frame_layout, and they're called immediately beforehand. Just provide the input directly.


fitzgen (approved):
Nice.

PR #8471, opened 2024-04-25
cranelift: Pass source location through directly

jameysharp:
Stashing the source location in a side location is unnecessary: we already have it where we need it.


cfallin (approved):
LGTM, sorry for missing this one earlier!

Issue #5908, opened 2023-03-02
cranelift/egraphs: allow simplifying trapping arithmetic


jameysharp:
#### Feature

We currently have simplification rules in `opts/algebraic.isle` to rewrite `x/1` to `x`, and in `opts/cprop.isle` to constant-fold division when both operands are constant. But these rules can't fire, because `simplify` is only called on instructions which never have side effects, and division may trap if the divisor is 0.

I'd like to be able to use simplification rules like these on trapping arithmetic.

#### Benefit

We could optimize away more instructions. There are comments in `algebraic.isle` suggesting more desired optimizations that we can't implement today either:
```
;; TODO: strength reduction: div to shifts
;; TODO: div/rem by constants -> magic multiplications
```

In addition to replacing expressions with simpler equivalents, we could sometimes remove them entirely if they're unused. We currently can't do that because if the instruction would trap then we need to ensure that trap occurs. But if we have a valid rewrite from a possibly-trapping instruction to an expression which can't trap, then it's actually safe to treat it as a pure instruction.

I suspect this might be particularly valuable for `uadd_overflow_trap`, which is used in bounds checks for dynamic heaps in cranelift-wasm.

#### Implementation

One idea is in #5796, but that approach was not selected in #5800 because it "is a little more complex than I would like". I still like the idea of a `force` instruction to pull otherwise-pure instructions into the side-effecting skeleton, but it's possible some other approach is better.

#### Alternatives

We could maybe call `simplify` from the `is_mergeable_for_egraph` branch of `optimize_skeleton_inst`, or introduce a new ISLE term for simplifying idempotent instructions.

cfallin:
Strong +1 to this -- this is the right way to represent idempotent side-effects IMHO. (Doing it the simpler way was a matter of expediency on my part as I wanted to get the optimization we disabled back in, but I fully support reworking it.)

Making side-effecting-but-idempotent ops part of the egraph has potential beyond arithmetic too, IMHO: the simplest case I can think of is actually just `trapnz vN` where we've rewritten `vN` to a constant 0. If a `trapnz` is really a `(force (trapnz vN))` and `(trapnz 0)` is rewritten to `0`, or `(nop)`, or something appropriately-typed, then we can elide traps lowered/legalized from other logic as well if it's safe to do so.

jameysharp:
[...]

PR #8474, opened 2024-04-25
div to shift for positive power of 2 divs

KGrewal1:
An attempt at adding isle opts for turning division by powers of two into ushr or sshr bit shifts

### DIFF
--- a/cranelift/codegen/src/opts/arithmetic.isle
+++ b/cranelift/codegen/src/opts/arithmetic.isle
@@ -85,7 +85,13 @@
                       (iconst_u ty 1)))
       (subsume x))
 
-;; TODO: strength reduction: div to shifts
+;; TODO: strength reduction: div to shifts when negative power of two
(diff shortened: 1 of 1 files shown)


fitzgen:
I don't think this is viable yet because the e-graph cannot rewrite side effecting instructions in general, see https://github.com/bytecodealliance/wasmtime/issues/5908 for more information.

I suspect that if you attempted to write a filetest exercising this rule, you'd find that it never fires because we never put division instructions into the egraph, and leave them in the "skeleton" outside of the graph.

KGrewal1:
Ah ok: looking through that issue seems like the current div simplifications in that file also currently do nothing for the same reason...

elliottt:
@KGrewal1 I'm going to close this PR for now, as it's not possible to implement these rules with the constraints of our current e-graph implementation. We can re-open this PR if that changes in the future.

PR #8475, opened 2024-04-25
cranelift: Simplify allocating ABI temp VRegs

jameysharp:
This merges several existing passes over the ABI signature that handle incoming arguments:

1. Callee::temps_needed counted how many and what type of temporaries would be needed in some later steps, and Lower::lower allocated them.
2. Callee::init duplicated the same logic to determine where to save those temporaries for later.
3. Various places used the saved temporaries.

There were two kinds of temporaries allocated. One kind was the return-area pointer, which must be allocated before any instructions are lowered so that the return instructions will all use the same VReg that the prologue defines. After this commit, that allocation still happens at the same point, but now it follows the pattern of other ABI glue and uses a VRegAllocator directly.

The other kind was a truly local kind of temporary used in only one uncommon path in gen_copy_arg_to_regs. That is unnecessary because we already allocate temporaries as-needed in all other paths in that method, so this commit just follows the same pattern.

This change avoids heap-allocating two temporary Vecs during init, as well as a third Vec that was live for the entire lifetime of the Callee. The latter almost always contained only a bunch of `None` values, and was only used at the end of lowering.

### DIFF
--- a/cranelift/codegen/src/machinst/abi.rs
+++ b/cranelift/codegen/src/machinst/abi.rs
@@ -1097,9 +1097,7 @@ pub struct Callee<M: ABIMachineSpec> {
     /// Finalized frame layout for this function.
     frame_layout: Option<FrameLayout>,
     /// The register holding the return-area pointer, if needed.
-    ret_area_ptr: Option<Writable<Reg>>,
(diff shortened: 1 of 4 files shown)


fitzgen (approved):
Nice!

PR #8479, opened 2024-04-25
cranelift: Only set VReg type once

jameysharp:
We know the type of each VReg at the moment when we allocate it, so we never need to set the type again. That also means we don't need to deduplicate reftyped VRegs as we go, although it's still possible to have duplicates after aliases are resolved.

This PR and #8469 remove the only two uses of `FxHashSet` in `vcode.rs`, so whichever one lands second needs to also delete the `use` item for that type or it'll introduce a new unused-import warning.


jameysharp:
Since #8469 landed I've rebased this PR.

elliottt (approved):
Awesome!

PR #8482, opened 2024-04-25
Fix missing additional derives

### LINKED ISSUE #8477: The additional_derives option on bindgen misses some types
#8441  recently added support for additional_derives in `wasmtime::component::bindgen` but the new version appears not to apply the derives to all types. In particular it seems to miss variants which are only used as fields in records, as in this definition:

```
package test:bindgen-test;

interface types {
    variant inner-variant {
        a,
        b,
        c
    }
    record outer-record{
        inner: inner-variant
    }
}

world bindgen-test {
    use types.{outer-record};
    export test-failure: func() -> outer-record;
}
```

In my testing the derives will be applied to OuterRecord but not to InnerVariant.

Attached is a complete test case which fails to compile.
[bindgen_case.tgz](https://github.com/bytecodealliance/wasmtime/files/15120002/bindgen_case.tgz)

fibonacci1729:
Fixes a mixing case when applying `additional_derives` and updates the test to verify the fix.

Fixes #8477

### DIFF
--- a/crates/component-macro/tests/codegen.rs
+++ b/crates/component-macro/tests/codegen.rs
@@ -531,9 +531,16 @@ mod custom_derives {
             package my:inline;
 
             interface blah {
+                variant abc {
+                    a,
+                    b,
+                    c
+                }
+
                 record foo {
                     field1: string,
-                    field2: list<u32>
+                    field2: list<u32>,
+                    field3: abc
                 }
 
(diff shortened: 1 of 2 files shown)


alexcrichton (approved):

alec-deason:
I can confirm this resolves #8477 for me. Thanks

PR #8485, opened 2024-04-26
Migrate the wasmtime-types crate to no_std

alexcrichton:
This commit is where no_std for Wasmtime starts to get a bit interesting. Specifically the `wasmtime-types` crate is the first crate that depends on some nontrivial crates that also need to be migrated to `no_std`. This PR disables the default feature of `wasmparser` by default and additionally does the same for `serde`. This enables them to compile in `no_std` contexts by default and default features will be enabled elsewhere in this repository as necessary.

This also opts to drop the `thiserror` dependency entirely in favor of a manual `Display` implementation with a cfg'd implementation of `Error`.

As before CI checks are added for `wasmtime-types` with a `no_std` target itself to ensure the crate and all dependencies all avoid `std`.

cc #8341

### DIFF
--- a/crates/types/src/lib.rs
+++ b/crates/types/src/lib.rs
@@ -1,12 +1,22 @@
 //! Internal dependency of Wasmtime and Cranelift that defines types for
 //! WebAssembly.
 
-use smallvec::SmallVec;
+#![no_std]
+
+extern crate alloc;
+#[cfg(feature = "std")]
+extern crate std;
+
 pub use wasmparser;
 
+#[doc(hidden)]
+pub use alloc::format as __format;
+
+use alloc::boxed::Box;
+use core::{fmt, ops::Range};
 use cranelift_entity::entity_impl;
 use serde_derive::{Deserialize, Serialize};
-use std::{fmt, ops::Range};
+use smallvec::SmallVec;
 
 mod error;
 pub use error::*;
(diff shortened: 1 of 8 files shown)


crates/types/src/lib.rs
```diff
@@ -1,12 +1,22 @@
 //! Internal dependency of Wasmtime and Cranelift that defines types for
 //! WebAssembly.
 
-use smallvec::SmallVec;
+#![no_std]
+
+extern crate alloc;
+#[cfg(feature = "std")]
+extern crate std;
```
abrown, line 8: Hoping to learn about `no_std` here: why include this line? The crate should compile and work without it, so is it just here to shadow what gets used in `alloc` with the `std` version instead?

alexcrichton: Ah this is here for the (one) explicit use of std in this crate, basically just the Error trait implementation. Without it there's no std to pull that trait from

abrown: Got it, thanks!

abrown (approved):

abrown:
@alexcrichton, looks like we need to enable `std` for the component adapter: https://github.com/bytecodealliance/wasmtime/actions/runs/8851805718/job/24309173297#step:6:130.

PR #8486, opened 2024-04-26
cranelift: Update operand aliases in-place

jameysharp:
Now all registers passed to the operand collector are mutably borrowed directly out of their original locations in the Inst, so it is possible to update them in place.

As an initial demonstration of the utility of this change, the results of the VReg renamer are applied directly to the instructions during operand collection, and then all VReg aliases are cleared after operand collection.

Most of this commit consists of deleting noise from the many `get_operands` implementations in all the backends: most ampersands and asterisks, and all uses of the `ref` keyword.


fitzgen (approved):
Very nice!

PR #8488, opened 2024-04-26
cargo-vet: Exclude fuzzing-only dependencies

jameysharp:
We can't meaningfully audit the other WebAssembly implementations that we use for differential fuzzing, such as wasmi and especially v8. Let's acknowledge that the effort to do so is not practical for us, and focus our vetting efforts on crates that developers and users are more likely to build.

This reduces our estimated audit backlog by over three million lines, according to `cargo vet suggest`.

Note that our crates which depend on those engines, such as wasmtime-fuzzing, are not published to crates.io, so if we fall victim to a supply chain attack against dependencies of these crates, the folks who might be impacted are limited.

Although there is value in also auditing code that might be run by people who clone our git repository, in this case I propose that anyone who is concerned about the risks of supply chain attacks against their development systems should be running fuzzers inside a sandbox. After all, it's a fuzzer: it's specifically designed to try to do anything.

I'd like to especially seek comment from folks who've expressed interest in our use of cargo-vet, like @alexcrichton, @bholley, @cfallin, and @tschneidereit. I'm open to being persuaded that we shouldn't make this change, but I can't currently see that we get any value from auditing these particular dependencies.

### DIFF
--- a/fuzz/README.md
+++ b/fuzz/README.md
@@ -9,6 +9,31 @@ oracles and pass libFuzzer-provided inputs to them. The test case generators and
 oracles themselves are independent from the fuzzing engine that is driving the
 fuzzing process and are defined in `wasmtime/crates/fuzzing`.
 
+## Safety warning
+
+Fuzzers exist to generate random garbage and then try running it. **You
+should not trust these fuzz targets**: they could in theory try to read
+or write files on your disk, send your private data to reporters, or do
+anything else. If they succeed at doing something malicious, they are
+doing a great job at identifying dangerous bugs and we're proud of them.
+
+In addition, some of these fuzz targets use other libraries, such as to
+test that our implementation matches other WebAssembly runtimes. **We
+have not reviewed those runtimes or libraries** for safety, security,
+correctness, supply-chain attacks, or any other properties. Software
+used only during fuzzing is not subject to [our usual `cargo vet`
+requirements][vet-docs].
+
+[vet-docs]: https://docs.wasmtime.dev/contributing-coding-guidelines.html#dependencies-of-wasmtime
+
+Paragraphs 7 and 8 of the license which this work is distributed to you
+under are especially important here: **We disclaim all warranties and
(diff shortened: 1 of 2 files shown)


alexcrichton (approved):
I think this is a reasonable change to make yeah, we already don't vet these dependencies much and reducing our backlog seems reasonable to me

fitzgen:
Makes sense to me too.

tschneidereit:
Same for me, yeah. It'd be good to have this mentioned somewhere in documentation people running fuzzing would be likely to see

jameysharp:
> It'd be good to have this mentioned somewhere in documentation people running fuzzing would be likely to see

Good point. I'll add some text to `fuzz/README.md` and `docs/contributing-fuzzing.md` to this PR before merging it. I'm not sure we have anywhere else that anyone is likely to notice.

Also, do we have text somewhere that I should update describing our cargo-vet policy?

jameysharp:
I've pushed some new text for `fuzz/README.md`. @tschneidereit, does this text work for you?

I considered adding text to `docs/contributing-fuzzing.md` but it's really about developing fuzz targets. For people who just want to run them it points to `fuzz/README.md`, so I think the notice above suffices.

I also considered adding text to `docs/contributing-coding-guidelines.md`, where we talk about `cargo vet`, but couldn't decide if there's anything that would be useful to say there.

Finally, I considered trying to make suggestions about sandboxes people could try, and got as far as verifying that `cargo fuzz run` works under `unshare --map-current-user --net --ipc --mount` on Linux (but not with `--pid` for some reason). But then I remembered that I don't want to be in the business of recommending security tools.

alexcrichton:
I think this is pretty reasonable so I'm gonna go ahead and flag for merge given the consensus here. @tschneidereit if you've got further thoughts though on the docs though we can always have follow-ups too

tschneidereit:
Thank you for doing this, @jameysharp! I agree that these doc changes are good, and also with your reasoning for not making other doc changes :)

PR #8490, opened 2024-04-26
build: make the c-api CMakeLists.txt easier to import

dundargoc:
Currently the CMakeLists.txt is designed to only be consumed by a parent project, which limits its flexibility. Adding a `project()` function allows it to also be built as its own project.

Also set `USES_TERMINAL_BUILD` to `TRUE` in `ExternalProject_Add()`, which allows getting the installation progress output from `cargo build`.


alexcrichton (approved):
Thanks!

dundargoc:
Would it be fine to backport this to the `release-20.0.0` branch? I don't mind sending another PR to that branch.

PR #8491, opened 2024-04-26
Encapsulate StableMap, StableSet and DisjointSets

KGrewal1:
As mentioned in https://github.com/bytecodealliance/wasmtime/pull/8473#issuecomment-2078271067
separate out StableMap, StableSet and DisjointSets into separate modules so that the compiler can ensure that the internal implementations stay private


fitzgen (approved):
Thanks!

PR #8493, opened 2024-04-27
cranelift: Move VReg aliases to VRegAllocator

jameysharp:
Now that #8486 landed, allowing us to resolve aliases in machine instructions, we have ensured that all VReg aliases are resolved by the time we're done building the VCode. Therefore we only need to keep track of the aliases map before that.

The VReg allocator is also dropped when we finish building the VCode, and it makes sense to track aliases there. This lets us maintain an invariant, that PCC facts are only stored on VRegs which are not aliased, while only reasoning locally within VRegAllocator.

I've changed the trace-log output to print the VCode immediately before it's finalized, along with key details in the VRegAllocator. This allows seeing the instructions before aliases are rewritten, although they're in reverse order at that point. There's another trace-log message somewhere else which logs the finalized VCode, so you can see both.

Previously, the initial capacity of the vreg_aliases map was set to ten times the number of basic blocks in the function. However we can make a better estimate based on the number of SSA values in the function, and use that to preallocate storage for other things in VRegAllocator too.

Keeping the aliases outside the VCode fixes previous borrow-checker challenges, which is a nice bonus.

### DIFF
--- a/cranelift/codegen/src/machinst/lower.rs
+++ b/cranelift/codegen/src/machinst/lower.rs
@@ -362,7 +362,9 @@ impl<'func, I: VCodeInst> Lower<'func, I> {
             VCodeBuildDirection::Backward,
         );
 
-        let mut vregs = VRegAllocator::new();
+        // We usually need two VRegs per instruction result, plus extras for
+        // various temporaries, but two per Value is a good starting point.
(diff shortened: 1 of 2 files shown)


cfallin (approved):
This is a really nice simplification -- thanks and LGTM!

PR #8494, opened 2024-04-27
cranelift/riscv64: Fix `mv` instruction pretty-printing

jameysharp:
The operand collector and the instruction emitter for Inst::Mov both placed the `rm` register before `rd`, so the emitted code was correct, but the pretty-printer used the opposite order and so printed the operands backwards. Note that the VCode disassembly disagreed with Capstone's disassembly of the emitted machine code.


afonso360 (approved):
Neat, thanks for catching this!

PR #8496, opened 2024-04-27
[20.0.0]: Backport fix from main to the release branch

dundargoc:
This is a backport of PR https://github.com/bytecodealliance/wasmtime/pull/8490

build: make the c-api CMakeLists.txt easier to import

Currently the CMakeLists.txt is designed to only be consumed by a parent
project, which limits its flexibility. Adding a `project()` function
allows it to also be built as its own project.

Also set `USES_TERMINAL_BUILD` to `TRUE` in `ExternalProject_Add()`,
which allows getting the installation progress output from `cargo
build`.


dundargoc:
Unsure about how backporting procedures are done in wasmtime, lemme know if this isn't correct and I can close it. Didn't find specific info about it in the contributing docs.

alexcrichton:
No worries this is indeed the procedure! The only other step is that a maintainer will need to trigger the patch release but that's just hitting a button and then hitting the PR merge button.

To confirm, is this something you'd want in a point-release of wasmtime? This, as far as I would know, would only change the source tarball in our release artifacts. If you're not using that then it may not affect too much. Or are you mostly interested in seeing this on the `release-20.0.0` branch in git?

dundargoc:
Oh, I see. We are not using the tarballs, just the commits pointed by the git tags. Yes, I mostly wanted to backport it to the release branch.

alexcrichton (approved):
Ok sounds good! I'll work on pushing the point release buttons tomorrow

alexcrichton:
Ok 20.0.1 should be released with https://github.com/bytecodealliance/wasmtime/pull/8538 which will have a tag for you to use

PR #8497, opened 2024-04-27
fix(c-api): give a cleaner error message if cargo isn't found

dundargoc:
Instead of

> Performing build step for
'wasmtime-crate''WASMTIME_CARGO_BINARY-NOTFOUND' is not recognized as an
internal or external command, operable program or batch file.

this will now instead output

> "cargo" was not found. Ensure "cargo" is in PATH. Aborting...

### DIFF
--- a/crates/c-api/CMakeLists.txt
+++ b/crates/c-api/CMakeLists.txt
@@ -78,6 +78,9 @@ if(ANDROID)
 endif()
 include(ExternalProject)
 find_program(WASMTIME_CARGO_BINARY cargo)
+if(NOT WASMTIME_CARGO_BINARY)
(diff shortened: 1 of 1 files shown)


crates/c-api/CMakeLists.txt
```diff
@@ -78,6 +78,9 @@ if(ANDROID)
 endif()
 include(ExternalProject)
 find_program(WASMTIME_CARGO_BINARY cargo)
+if(NOT WASMTIME_CARGO_BINARY)
+    message(FATAL_ERROR [["cargo" was not found. Aborting...]])
```
rockwotj, line 82: Might be worth also recommending a resolution such as adding cargo to PATH or to CMAKE_PROGRAM_PATH.

rockwotj (approved):

fitzgen (approved):

PR #8498, opened 2024-04-27
Fxhash to rustchash

KGrewal1:
This removes the internal code for fxhash and instead uses the fxhash crate  and subsequently migrates use of the fxhash crate (https://crates.io/crates/fxhash) to the rustc-hash crate (https://crates.io/crates/rustc-hash) implementing the same hash function: I believe the latter will need to be audited however: also manually implement the hash convenience function from fxhash --- @fitzgen


Cargo.toml
```diff
@@ -292,6 +292,7 @@ url = "2.3.1"
 humantime = "2.0.0"
 postcard = { version = "1.0.8", default-features = false, features = ['alloc'] }
 criterion = { version = "0.5.0", default-features = false, features = ["html_reports", "rayon"] }
+fxhash = {package = "rustc-hash", version = "1.1.0"}
```
bjorn3, line 295: There is no need to rename this package, right?

KGrewal1: I actually made a mistake: fxhash specifically is currently needed as souper-harvest needs this hash convenience function which is not present in rustc-hash https://docs.rs/fxhash/0.2.1/fxhash/fn.hash.html .

bjorn3: I didn't mean undoing the migration, I mean using `rustc-hash = "1.1.0"` instead of `fxhash = {package = "rustc-hash", version = "1.1.0"}`.

KGrewal1: I chose to rename though to minimise the size of the diff: appreciate if that is considered ugly and so not done

bjorn3: > I actually made a mistake: fxhash specifically is currently needed as souper-harvest needs this hash convenience function which is not present in rustc-hash

Inlining the convenience function is fine I think.

> I chose to rename though to minimise the size of the diff: appreciate if that is considered ugly and so not done

Doing the move in a separate commit without renaming would still be easy to review.

KGrewal1: Have now done this

fitzgen (approved):
Thanks! I'll do the audit and merge this as well in another PR. Will link to it when it is ready.

fitzgen:
Ah, looks like the Firefox folks kindly already audited `rustc-hash` so we can just go ahead and merge this!

cranelift/src/souper_harvest.rs
```diff
@@ -139,3 +139,11 @@ fn parse_input(path: PathBuf) -> Result<Vec<Function>> {
         .with_context(|| format!("parse error in {}", path.display()))?;
     Ok(funcs)
 }
+
+/// A convenience functon for a quick usize hash
+#[inline]
+pub fn hash<T: std::hash::Hash + ?Sized>(v: &T) -> usize {
```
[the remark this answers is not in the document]
KGrewal1: Looking at the code I think that would work (and MSRV is 1.75)? Can't see any reason that it'd need to be usize, if anyone else knows more about it than me?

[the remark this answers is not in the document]
fitzgen: `u64` would indeed be fine.

PR #8499, opened 2024-04-27
cranelift: Generalize OperandCollector into a trait

jameysharp:
This paves the way for more implementations of this OperandVisitor trait which can do different things with the operands.

Of particular note, this commit introduces a second implementation which is used only in the s390x backend and only to implement a debug assertion. Previously, s390x used an OperandCollector::no_reuse_def method to implement this assertion, but I didn't want to require that all implementors of the new trait have to provide that method, so this captures the same check but keeps it local to where it's needed.


fitzgen (approved):
Nice!

PR #8501, opened 2024-04-29
Turn the `wasmtime-runtime` crate into the `wasmtime::runtime::vm` module

fitzgen:
We want to merge them together because there is no abstraction or layering or boundary between them in practice, and our architectures and designs have, at times, been constrained by this artificial crate boundary and what is visible in which crate and this has forced us to use nasty workarounds like defining traits in `wasmtime-runtime` that get implemented in `wasmtime` and used as dynamic trait objects inside `wasmtime-runtime`. Merging them will allow us to remove these hacks.


fitzgen:
> The next step after this, before actually merging the crates together, is to factor out `wasmtime_runtime::{sys,mmap,mmap_vec}` into its own crate, since this functionality is used directly by the CLI in one place, and I don't want to add that stuff to the `wasmtime` crate's public API.

This is proving to be a bit hairier than anticipated; investigating alternatives. Shouldn't affect this PR.

kaimast:
How do I call `wasmtime_runtime::raise_user_trap` after this change?

fitzgen:
Adding to what Alex said: that function was never intended to be exposed to or used by Wasmtime embedders. The `wasmtime-runtime` crate was an implementation detail of Wasmtime itself. Generally, to raise a trap in Wasm from a host API, you want to return an `Err(..)` from the host call.

kaimast:
> Generally, to raise a trap in Wasm from a host API, you want to return an `Err(..)` from the host call.

Oh, I did not realize that is possible. I fairly recently ported my code over from Wasmer, so I am not super familiar with API yet. That is indeed a nicer way of doing it... 

Thanks for the quick responses!

fitzgen:
@kaimast no problem! Feel free to pop into our Zulip or file an issue if you have any more questions.

PR #5575, opened 2023-01-14
Update AMD and generic x86 CPU presets to match LLVM

MozarellaMan:
This is for #5444, previous PR: #5490

This adds the remaining AMD and generic CPU presets that are defined in [LLVM](https://github.com/llvm/llvm-project/blob/d4493dd1ed58ac3f1eab0c4ca6e363e2b15bfd1c/llvm/lib/Target/X86/X86.td#L1300-L1643).
## Changes
- added AMD x86 presets
- added Generic x86 presets

PR #8502, opened 2024-04-29
cranelift: Add `znver4` arch preset

kauppie:
Added `znver4` architecture preset to `cranelift` settings. This should sync with LLVM at least on the AMD side. This was previously updated in #5575.


cfallin (approved):
Thanks!

PR #8503, opened 2024-04-29
cranelift: Split predecessor/successor lists

jameysharp:
This establishes the property that the VCode's various lists of ranges each fully cover the index range of another list. Previously, the block_succ_range list covered the first half of block_succs_preds, and the block_pred_range list covered the second half.

While I was in the area, I replaced the O(n log n) sort in compute_preds_from_succs with a linear-time counting sort, which uses less temporary storage and directly computes the ranges we want as a byproduct.

### DIFF
--- a/cranelift/codegen/src/machinst/vcode.rs
+++ b/cranelift/codegen/src/machinst/vcode.rs
@@ -90,18 +90,23 @@ pub struct VCode<I: VCodeInst> {
     /// Block instruction indices.
     block_ranges: Vec<(InsnIndex, InsnIndex)>,
 
-    /// Block successors: index range in the `block_succs_preds` list.
+    /// Block successors: index range in the `block_succs` list.
     block_succ_range: Vec<(u32, u32)>,
 
-    /// Block predecessors: index range in the `block_succs_preds` list.
+    /// Block successor lists, concatenated into one vec. The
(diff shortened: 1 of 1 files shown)


cfallin (approved):
This is super-cool! IMHO this is the "Book" algorithm (in the [Paul Erdos sense](https://en.wikipedia.org/w/index.php?title=Paul_Erd%C5%91s&oldid=1220526564#:~:text=When%20he%20saw%20a%20particularly%20beautiful%20mathematical%20proof%20he%20would%20exclaim%2C%20%22This%20one%27s%20from%20The%20Book!%22%20This%20later%20inspired%20a%20book%20titled%20Proofs%20from%20the%20Book.)) for finding preds from succs. Clean and direct.

Out of curiosity have you measured the compile-time impact of this? I wonder if it'll show up in code with complex CFGs... we should take it in any case though.

jameysharp:
I hadn't measured, but according to Hyperfine, compiling the Spidermonkey benchmark from Sightglass is unchanged with this commit. I think that's unsurprising; the largest function has 2650 blocks and 6777 edges (before legalization and critical-edge splitting), and that's just not that many. But I'm still absurdly pleased with finding a situation where a linear-time sort is actually the best choice.

PR #8505, opened 2024-04-29
wasmtime(gc): Add support for struct types

fitzgen:
Similar to https://github.com/bytecodealliance/wasmtime/pull/8481 but for struct types instead of array types.

Note that this is support for only defining these types in Wasm or the host; we don't support allocating instances of these types yet. That will come in follow up PRs.


crates/wasmtime/src/runtime/types.rs
```diff
@@ -1085,6 +1131,179 @@ impl FieldType {
...
+    pub fn matches(&self, other: &StructType) -> bool {
+        assert!(self.comes_from_same_engine(other.engine()));
+
+        // Avoid matching on structure for subtyping checks when we have
+        // precisely the same type.
+        if self.type_index() == other.type_index() {
+            return true;
+        }
+
+        todo!("FITZGEN")
```
alexcrichton, line 1231: FITZGEN

alexcrichton (approved):

PR #8506, opened 2024-04-29
cranelift: Compress vcode range-lists

jameysharp:
These lists of ranges always cover contiguous ranges of an index space, meaning the start of one range is the same as the end of the previous range, so we can cut storage in half by only storing one endpoint of each range.

This in turn means we don't have to keep track of the other endpoint while building these lists, reducing the state we need to keep while building vcode and simplifying the various build steps.

### DIFF
--- a/cranelift/codegen/src/ranges.rs
+++ b/cranelift/codegen/src/ranges.rs
@@ -1,0 +1,131 @@
+//! The [`Ranges`] type stores a list of contiguous index ranges that
+//! span some other list's full length.
+
+use alloc::vec::Vec;
+use core::ops::Range;
+
+/// A list of contiguous index ranges.
+#[derive(Default)]
+pub struct Ranges {
+    ranges: Vec<u32>,
+    reverse: bool,
+}
+
+impl Ranges {
+    /// Constructs a new, empty, list of ranges with at least the
+    /// specified capacity.
+    pub fn with_capacity(capacity: usize) -> Self {
+        let mut new = Ranges::default();
+        new.reserve(capacity);
+        new
+    }
+
+    /// Add a new range which begins at the end of the previous range
+    /// and ends at the specified offset, exclusive.
+    pub fn push_end(&mut self, end: usize) {
+        debug_assert!(!self.reverse);
+        // To keep this implementation simple we explicitly store the
+        // starting index, which is always 0, so that all ranges are
+        // represented by adjacent pairs in the list. But we add it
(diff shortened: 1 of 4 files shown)


cranelift/codegen/src/ranges.rs
```diff
@@ -0,0 +1,131 @@
...
+        self.ranges
+            .windows(2)
+            .enumerate()
+            .map(|(index, range)| (self.map_index(index), range[0] as usize..range[1] as usize))
+    }
+
+    /// Reverse this list of ranges, so that the first range is at the
+    /// last index and the last range is at the first index.
+    ///
+    /// ```ignore
```
fitzgen, line 73: Why ignore?

cranelift/codegen/src/ranges.rs
```diff
@@ -0,0 +1,131 @@
...
+        self.ranges.reserve(additional);
+    }
+
+    /// Get the range at the specified index.
+    pub fn get(&self, index: usize) -> Range<usize> {
+        let len = self.len();
+        assert!(index < len, "index {index} is too big for length {len}");
+        let index = self.map_index(index);
+        self.ranges[index] as usize..self.ranges[index + 1] as usize
+    }
```
fitzgen, line 57: Can we add some typed indices for `Ranges`? Or reuse `cranelift_entity::Entity`?

I've always been slightly nervous about all of our raw indices in the backends.

cranelift/codegen/src/ranges.rs
jameysharp: Because doc-tests don't work in non-public types, since they're compiled in a separate crate. I couldn't figure out a better way to document what these methods do than with code, so doc-test style seemed good, but didn't quite work how I wanted. I did build this with the module made public and the doc-tests un-ignored, and verified that they pass.

cranelift/codegen/src/ranges.rs
jameysharp: I guess we could add some sort of type-system tagging, but what we have now is `usize` almost everywhere this is used, so it would be noisy.

That said, I had wanted to somehow associate each instance of `Ranges` with the `Vec` or `Ranges` that it indexes. I couldn't figure out a good way to do that, but putting a newtype wrapper around the indexes could work, with some helpers.

I'd prefer to land this without that change, since I believe it's easier to understand and no less safe than the _status quo_, but I do think this is worth thinking about more.

elliottt (approved):
This is a really nice change!

PR #8507, opened 2024-04-29
winch(arm64): float_copysign

evacchi:
I eyeballed the codegen in ISLE, but I must check the output to make sure I actually understood it right 😬 
continues #8321

### DIFF
--- a/winch/codegen/src/isa/aarch64/asm.rs
+++ b/winch/codegen/src/isa/aarch64/asm.rs
@@ -300,7 +300,8 @@ impl Assembler {
         self.emit_fpu_round(fpu_mode, rn, rd)
     }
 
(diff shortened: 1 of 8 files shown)


winch/codegen/src/isa/aarch64/asm.rs
```diff
@@ -296,6 +300,32 @@ impl Assembler {
...
+        };
+        let ushr = match size {
+            OperandSize::S32 => UShr32(imm),
+            OperandSize::S64 => UShr64(imm),
+            _ => unreachable!(),
+        };
+        self.emit_fpu_rri(ushr, rn, rd)
+    }
+
+    pub fn fsli_rrr(&mut self, ri: Reg, rn: Reg, rd: Reg, amount: u8, size: OperandSize) {
```
saulecabrera, line 316: Would you mind adding comments to these public methods?

saulecabrera (approved):
Thanks!

PR #8509, opened 2024-04-29
Wasmtime: add one-entry call-indirect caching.

cfallin:
In WebAssembly, an indirect call is somewhat slow, because of the indirection required by CFI (control-flow integrity) sandboxing. In particular, a "function pointer" in most source languages compiled to Wasm is represented by an index into a table of funcrefs. The `call_indirect` instruction then has to do the following steps to invoke a function pointer:

- Load the funcref table's base and length values from the vmctx.
- Bounds-check the invoked index against the actual table size; trap if out-of-bounds.
- Spectre mitigation (cmove) on that bounds-check.
- Load the `vmfuncref` from the table given base and index.
  - For lazy table init, check if this is a non-initialized funcref pointer, and initialize the entry.
- Load the signature from the funcref struct and compare it against the `call_indirect`'s expected signature; trap if wrong.
- Load the actual code pointer for the callee's Wasm-ABI entry point.
- Load the callee vmctx (which may be different for a cross-module call).
- Put that vmctx in arg 0, our vmctx in arg 1, and invoke the loaded code pointer with an indirect call instruction.

Compare and contrast to the process involved in invoking a native function pointer:

- Invoke the code pointer with an indirect call instruction.

This overhead buys us something -- it is part of the SFI sandbox boundary -- but it is very repetitive and unnecessary work in *most* cases when indirect function calls are performed repeatedly (such as within an inner loop).

This PR introduces the idea of *caching*: if we know that the result of all the above checks won't change, then if we use the same index as "the last time" (for some definition), we can skip straight to the "invoke the code pointer" step, with a cached code pointer from that last time.

Concretely, it introduces a two-word struct inlined into the vmctx for each `call_indirect` instruction in the module (up to a limit):

- The last invoked index;
- The code pointer that index corresponded to.

When compiling the module, we check whether the table could possibly be mutable at a given index once read: any instructions like `table.set`, or the whole table exported thus writable from the outside. We also check whether index 0 is a non-null funcref. If neither of these things are true, then we know we can cache an index-to-code-pointer mapping, and we know we can use index 0 as a sentinel for "no cached value".

We then make use of the struct for each indirect call site and generate code to check if the index matches; if so, call cached pointer; if not, load the vmfuncref, check the signature, check that the callee vmctx is the same as caller (intra-module case), and stash the code pointer and index away (fill the cache), then make the call.

On an in-development branch of SpiderMonkey-in-Wasm with ICs (using indirect calls), this is about a 20% speedup; I haven't yet measured on other benchmarks. It is expected that this might be an instantiation-time slowdown due to a larger vmctx (but we could use madvise to zero if needed).

This feature is off by default right now.


cfallin:
(I'll run some benchmarks to see what the compile-time impact is like, and how this benefits non-SpiderMonkey workloads; TBD.)

cfallin:
Benchmarks:

- Sightglass measures the following for `spidermonkey.wasm`:
```
compilation :: cycles :: benchmarks/spidermonkey/benchmark.wasm

  Δ = 1456249767.40 ± 1315409020.02 (confidence = 99%)

  original.so is 1.01x to 1.17x faster than cache-indirect-calls.so!

  [17934023164 18070879867.60 18215783322] cache-indirect-calls.so
  [15377276916 16614630100.20 17840489432] original.so

execution :: cycles :: benchmarks/spidermonkey/benchmark.wasm

  Δ = 27069455.80 ± 22219096.46 (confidence = 99%)

  original.so is 1.00x to 1.05x faster than cache-indirect-calls.so!

  [1079208474 1095097604.00 1142114396] cache-indirect-calls.so
  [1035034082 1068028148.20 1094035960] original.so

instantiation :: cycles :: benchmarks/spidermonkey/benchmark.wasm

  No difference in performance.

  [834480 856949.40 902120] cache-indirect-calls.so
  [831744 860156.60 978994] original.so
```

so compilation time does take a little hit but my system is noisy; somewhere between 1% and 17%, the mean looks to be around 8.7%. Instantiation apparently unaffected. And execution time... slightly slower? (But again, noisy system.)

I tried running the next benchmark in the list (pulldown-cmark) and Sightglass crashed mysteriously, so on to trusty `hyperfine`: no deltas in runtime on spidermonkey or bz2, my two usual suspects.

I suspect that mostly we'll see mostly in-the-noise results for most benchmarks because indirect calls are somewhat rare in C-ish code; it will matter mostly when compiling dynamic languages, or something like Java with a lot of virtual calls. The speedup on weval'd SpiderMonkey is real though so I'd like to at least have this in tree as an option :-)

fitzgen:
FWIW, I did a 100-iteration sightglass run of the default suite and got these results:

```
instantiation :: cycles :: benchmarks/spidermonkey/benchmark.wasm

  No difference in performance.

  [472878 584750.13 1435913] cache.so
  [471975 647599.28 1762837] main.so

execution :: cycles :: benchmarks/pulldown-cmark/benchmark.wasm

  No difference in performance.

  [7422842 8744920.24 28196476] cache.so
  [7439782 9281270.19 32569639] main.so

instantiation :: cycles :: benchmarks/bz2/benchmark.wasm

  No difference in performance.

  [165456 245749.33 728069] cache.so
  [161764 251740.72 723740] main.so

  No difference in performance.

  [95484875 104221743.11 196171026] cache.so
  [95454637 105918836.28 208434699] main.so

  No difference in performance.

  [358282063 404458268.52 568579026] cache.so
  [367557812 399158971.64 551388853] main.so

compilation :: cycles :: benchmarks/bz2/benchmark.wasm

  No difference in performance.

  [194765165 225581880.12 437412507] cache.so
  [197703277 222917937.46 281571195] main.so

execution :: cycles :: benchmarks/spidermonkey/benchmark.wasm

  No difference in performance.

  [895587497 923280792.22 972838288] cache.so
  [891917856 927697965.55 1121115203] main.so

compilation :: cycles :: benchmarks/spidermonkey/benchmark.wasm

  No difference in performance.

  [8325711261 8502522129.94 8732981273] cache.so
  [8370183347 8530524550.49 9984684780] main.so

instantiation :: cycles :: benchmarks/pulldown-cmark/benchmark.wasm

  No difference in performance.

  [214742 289658.22 800685] cache.so
  [204642 290351.57 854625] main.so
```

I think it is safe to say that this doesn't have an affect on our regular suite, and I expect that the JS speed ups you are reporting rely on weval'ing first.

crates/wasmtime/src/config.rs
```diff
@@ -949,6 +950,12 @@ impl Config {
         self
     }
 
+    /// Configures whether we enable the "indirect call cache" optimization.
```
fitzgen, line 953: This could really use some more detail. As a user, why shouldn't I just enable this all the time? What are the trade offs? What is the default? What does it do? When does it apply? How does it affect memory footprint and pooling allocator integration? Etc...

The automatic github actions comment has a good template in the first check box: https://github.com/bytecodealliance/wasmtime/pull/8509#issuecomment-2084268706

tests/disas/indirect-call-caching-exclude-0-index.wat
```diff
@@ -0,0 +1,102 @@
+;;! target = "x86_64"
+;;! flags = [ "-Wcache-call-indirects=y" ]
+
+(module
```
fitzgen, line 4: Care to add a brief comment to this test file about what we are aiming to exercise here, so that future changes that happen to tweak this disassembly know what property to check is preserved in their changes?

tests/disas/indirect-call-caching-exclude-table-export.wat
```diff
@@ -0,0 +1,102 @@
+;;! target = "x86_64"
+;;! flags = [ "-Wcache-call-indirects=y" ]
+
+(module
```
fitzgen, line 4: Ditto

tests/disas/indirect-call-caching-exclude-table-writes.wat
```diff
@@ -0,0 +1,134 @@
+;;! target = "x86_64"
+;;! flags = [ "-Wcache-call-indirects=y" ]
+
+(module
```
fitzgen, line 4: (ditto for all these tests)

crates/wasmtime/src/config.rs
fitzgen: Speaking of that comment, the second item doesn't seem to be satisfied in this PR either: adding support for fuzzing this option. It should be just a couple-line change in these places:

* https://github.com/bytecodealliance/wasmtime/blob/0f4c0d4a358187f99f23897d306c8b2a3c94287e/crates/fuzzing/src/generators/config.rs#L458
* https://github.com/bytecodealliance/wasmtime/blob/0f4c0d4a358187f99f23897d306c8b2a3c94287e/crates/fuzzing/src/generators/config.rs#L150

crates/cranelift/src/func_environ.rs
```diff
@@ -1115,10 +1127,146 @@ impl<'a, 'func, 'module_env> Call<'a, 'func, 'module_env> {
         callee: ir::Value,
         call_args: &[ir::Value],
     ) -> WasmResult<Option<ir::Inst>> {
+        // If we are performing call-indirect caching with this table, check the cache.
+        let caching = if self.env.tunables.cache_call_indirects {
+            let plan = &self.env.module.table_plans[table_index];
+            // We can do the indireaching optimization only if table
```
fitzgen, line 1133: indireaching? This is a typo and not an intentional contraction, right?

crates/cranelift/src/func_environ.rs
```diff
@@ -1115,10 +1127,146 @@ impl<'a, 'func, 'module_env> Call<'a, 'func, 'module_env> {
...
+            let hit = self.builder.ins().icmp(IntCC::Equal, cached_index, callee);
+
+            let current_block = self.builder.current_block().unwrap();
+            let hit_block = self.builder.create_block();
+            let miss_block = self.builder.create_block();
+            self.builder.insert_block_after(hit_block, current_block);
+            self.builder.insert_block_after(miss_block, hit_block);
+            self.builder.set_cold_block(miss_block);
+            let continue_block = self.builder.create_block();
+            self.builder.insert_block_after(continue_block, miss_block);
```
fitzgen, line 1172: Could we factor this one out into smaller helpers and add comments describing the code and blocks we are emitting as well? I'm finding it really hard to read this giant block of clif-building code directly. See for example the overview comment and comments for each block we generate here:

https://github.com/bytecodealliance/wasmtime/blob/0f4c0d4a358187f99f23897d306c8b2a3c94287e/crates/cranelift/src/gc/enabled.rs#L341-L449

fitzgen (commented):
Overall looks great, but I have a few nitpicks, mostly related to improving things for future code readers. Thanks!

crates/environ/src/compile/module_environ.rs
```diff
@@ -566,10 +577,32 @@ impl<'a, 'data> ModuleEnvironment<'a, 'data> {
...
+                                self.flag_written_table(TableIndex::from_u32(table));
+                            }
+                            Operator::CallIndirect { .. } => {
+                                self.result.module.num_call_indirect_caches += 1;
+                            }
+
+                            _ => {}
+                        }
+                    }
+                }
```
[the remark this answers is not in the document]
cfallin: Done!

crates/environ/src/vmoffsets.rs
```diff
@@ -888,6 +926,28 @@ impl<P: PtrSize> VMOffsets<P> {
...
+    }
+
+    /// Return the offset to the `wasm_call` field in `*const VMCallIndirectCache` index `index`.
+    #[inline]
+    pub fn vmctx_call_indirect_cache_wasm_call(&self, index: u32) -> u32 {
+        self.vmctx_call_indirect_cache(index)
+            + u32::from(self.ptr.vmcall_indirect_cache_wasm_call())
+    }
+
+    /// Return the offset to the `index` field in `*const VMCallIndirectCache` index `index`.
```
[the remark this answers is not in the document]
cfallin: Done!

crates/environ/src/vmoffsets.rs
```diff
@@ -888,6 +926,28 @@ impl<P: PtrSize> VMOffsets<P> {
...
+    #[inline]
+    pub fn vmctx_call_indirect_cache_index(&self, index: u32) -> u32 {
```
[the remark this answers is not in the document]
cfallin: Done!

tests/disas/indirect-call-caching-exclude-0-index.wat
cfallin: For sure, added explanations here and below giving the reasons for expecting/not expecting the opt, and the key bits of IR we're asserting.

tests/disas/indirect-call-caching-exclude-table-export.wat
cfallin: Updated!

tests/disas/indirect-call-caching-exclude-table-writes.wat
cfallin: Updated!

crates/cranelift/src/func_environ.rs
cfallin: Huh, should be "indirect call caching"; somehow my fingers or a misplaced editor command swallowed the middle bits of that, sorry!

crates/cranelift/src/func_environ.rs
cfallin: Added a similar comment and factored out a little bit of the logic (and reduced the CFG by one block while I was at it) -- hopefully clearer now!

cfallin:
Updated, let me know what you think!

cfallin:
(Oh, and forgot to mention: thanks for the thorough benchmarks on this! I've gotta dig into why Sightglass is so finicky for me...)

fitzgen:
Oh also, I mentioned this in the Cranelift meeting, but wanted to put it down here for posterity too: it seems like the limit on the number of caches isn't actually implemented.

cfallin:
Yep, I missed that entirely in this iteration of the patch -- sorry about that! Just updated. I ended up wiring the limit through to a new config option (with docs, CLI flag, fuzzing) so that I could easily test it without a huge test input.

crates/cranelift/src/func_environ.rs
```diff
@@ -1106,6 +1149,68 @@ impl<'a, 'func, 'module_env> Call<'a, 'func, 'module_env> {
...
+    ///
+    /// Generates IR like:
+    ///
+    /// ```ignore
+    /// v1 = load.i64 cache_ptr+0 ;; cached index (cache key)
+    /// v2 = load.i64 cache_ptr+8 ;; cached raw code pointer (cache value)
+    /// ```
+    ///
+    /// and returns `(index, code_ptr)` (e.g. from above, `(v1, v2)`).
+    fn load_cached_indirect_index_and_code_ptr(
```
fitzgen, line 1172: This is great, thank you. I really appreciate having this stuff factored out into bite-sized chunks.

fitzgen (approved):
:rocket:

cfallin:
The merge queue did its job (nice!) and found a conflict with #8515 which merged after this PR's CI ran. I pushed a fix which is fairly small but @fitzgen maybe PTAL at least commit just to be sure?

cfallin:
(or @jameysharp feel free to review last commit too as it interacts with your change in #8515!)

crates/types/src/lib.rs
```diff
@@ -924,6 +924,12 @@ entity_impl!(ModuleInternedTypeIndex);
 pub struct VMSharedTypeIndex(u32);
 entity_impl!(VMSharedTypeIndex);
 
+/// Index of a `call_indirect` instruction in a module, used for
+/// caching that callsite's target in the VMContext.
+#[derive(Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Debug, Serialize, Deserialize)]
+pub struct CallIndirectSiteIndex(u32);
```
fitzgen, line 930: Oh I just noticed this new index type is in between the definition of VMSharedTypeIndex and the impl block defining methods for it. Can we move it before or after so that those two are together?

crates/types/src/lib.rs
```diff
@@ -1194,6 +1200,19 @@ impl ConstExpr {
     pub fn ops(&self) -> &[ConstOp] {
         &self.ops
     }
+
+    /// Is this ConstExpr a possibly zero integer value?
```
fitzgen, line 1204: Is this allowed to return false even when the value is actually zero? Jamey is adding support for the extended const proposal and it extends const expressions with addition, subtraction, and multiplication, so there will soon be many ways to evaluate to zero. Doesn’t mean we can’t land this PR now but we should be clear about the semantics of this method

fitzgen (approved):

crates/types/src/lib.rs
cfallin: Clarified a bit -- renamed to `provably_nonzero_i32` (inverted from is-possibly-nonzero -- clearer this way IMHO) and commented more. It's always allowed to return `false` (can't prove anything), but if `true`, must actually be a case we know is always nonzero (currently only an i32.const with nonzero constant).

jameysharp: I would suggest naming this with the opposite sense, like: `is_definitely_nonzero`, which returns true for `I32Const(x) if x != 0`, and may also return true if there are any other circumstances where it can prove this, but otherwise returns false. Then, extended constants or any other strange things will safely return false, meaning that you can't be sure your optimization is valid, and you'll only apply the optimization when it's definitely okay.

Further, when we follow Nick's suggestion for making the `ConstExpr` evaluator try to run even before an `Instance` is available, it will be clear how to invoke that to implement this. If it fails because it needs inputs which aren't known until instantiation, or it succeeds but returns 0, then we should return that the expression is not "definitely nonzero".

jameysharp: Hah, raced; I'm glad we're on the same page.

crates/cli-flags/src/lib.rs
```diff
@@ -244,6 +244,15 @@ wasmtime_option_group! {
...
         pub gc: Option<bool>,
+        /// Whether to enable call-indirect caching.
+        pub cache_call_indirects: Option<bool>,
+        /// The maximum call-indirect cache slot count.
+        ///
+        /// One slot is allocated per indirect callsite; if the module
+        /// has more indirect callsites than this limit, then the
+        /// first callsites in linear order in the code section, up to
+        /// the limit, will receive a cache slot.
+        pub max_call_indirect_cache_slots: Option<usize>,
```
alexcrichton, line 255: Could these be moved to `-O` options as "tuning" or "optimizaion" options?

The `WasmOptions` group here is intended for semantic differences in the generated code and these options shouldn't affect any generated code.

(thanks for adding cli flags though!)

cfallin: Ah, yeah, for sure -- will do in a followup PR.

PR #8510, opened 2024-04-30
cranelift: add icmp-of-icmp rules for comparisons with 1

tertsdiepraam:
Suggested by @cfallin in this zulip thread: https://bytecodealliance.zulipchat.com/#narrow/stream/217117-cranelift/topic/disassemly.20seems.20.28relatively.29.20unoptimized

The icmp-of-icmp rules only included comparisons with 0. So this
```
(x == y) != 0
```
gets simplified to
```
x == y
```

But the equivalent
```
(x == y) == 1
```
did not get simplified, even though the rules to apply are just the inverse of the rules for 0.

I've added the rules for `icmp(..) == 1` and `icmp(..) != 1` and tried to add some more comments to the existing rules.

I need some help to figure out how I could write a test for this, because I haven't found the tests in the codebase :)

### DIFF
--- a/cranelift/codegen/src/opts/icmp.isle
+++ b/cranelift/codegen/src/opts/icmp.isle
@@ -14,15 +14,33 @@
 (rule (simplify (sle (ty_int ty) x x)) (subsume (iconst_u ty 1)))
 
 ;; Optimize icmp-of-icmp.
+;; ne(icmp(ty, cc, x, y), 0) == icmp(ty, cc, x, y)
+;; e.g. neq(ugt(x, y), 0) == ugt(x, y) 
 (rule (simplify (ne ty
                       (uextend_maybe _ inner @ (icmp ty _ _ _))
                       (iconst_u _ 0)))
       (subsume inner))
 
+;; eq(icmp(ty, cc, x, y), 0) == icmp(ty, cc_complement, x, y)
+;; e.g. eq(ugt(x, y), 0) == ule(x, y) 
 (rule (simplify (eq ty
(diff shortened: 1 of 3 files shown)


cranelift/codegen/src/opts/icmp.isle
```diff
@@ -14,16 +14,34 @@
...
       (subsume inner))
 
+;; eq(icmp(ty, cc, x, y), 0) == icmp(ty, cc_complement, x, y)
+;; e.g. eq(ugt(x, y), 0) == ule(x, y) 
 (rule (simplify (eq ty
                       (uextend_maybe _ (icmp ty cc x y))
                       (iconst_u _ 0)))
       (subsume (icmp ty (intcc_complement cc) x y)))
 
+;; eq(icmp(ty, cc, x, y), 1) == icmp(ty, cc_complement, x, y)
```
KGrewal1, line 31: are these comments the wrong way around (ie should this be 

`eq(icmp(ty, cc, x, y), 1) == icmp(ty, cc, x, y)`

 and similar for the documentation of the other command
?

cfallin:
@tertsdiepraam thanks for this! The tests are in `cranelift/filetests/filetests/`; you can find midend-optimizer-specific tests that assert on expected simplified IR in `egraph/`, and tests that check execution results in `runtests/`. Probably both would be good for this case?

tertsdiepraam:
I've added tests both in `egraph/` and `runtests/` and fixed up the comment. I hope it's good now.

cfallin (approved):
LGTM -- thanks very much for this contribution!

PR #8514, opened 2024-05-01
wasmtime: Use ConstExpr for element segment offsets

jameysharp:
This shouldn't change any behavior currently, but prepares us for supporting extended constant expressions.

I have a separate commit for doing the same to data segments (#8515), but it needs more work.


fitzgen (approved):
Nice!

PR #8515, opened 2024-05-01
wasmtime: Use ConstExpr for data segment offsets

jameysharp:
This shouldn't change any behavior currently, but prepares us for supporting extended constant expressions.

~I want to think about this a bit more before merging it but here's a draft for @fitzgen or anyone else who wants to take a look.~ This is ready for review. This is basically the same change as #8514 but for memories instead of tables, and the existing code was more complicated for this case.

### DIFF
--- a/crates/environ/src/compile/module_environ.rs
+++ b/crates/environ/src/compile/module_environ.rs
@@ -619,24 +619,11 @@ impl<'a, 'data> ModuleEnvironment<'a, 'data> {
                         } => {
                             let range = mk_range(&mut self.result.total_data)?;
                             let memory_index = MemoryIndex::from_u32(memory_index);
-                            let mut offset_expr_reader = offset_expr.get_binary_reader();
-                            let (base, offset) = match offset_expr_reader.read_operator()? {
-                                Operator::I32Const { value } => (None, value.unsigned().into()),
-                                Operator::I64Const { value } => (None, value.unsigned()),
-                                Operator::GlobalGet { global_index } => {
-                                    (Some(GlobalIndex::from_u32(global_index)), 0)
-                                }
-                                s => {
-                                    bail!(WasmError::Unsupported(format!(
-                                        "unsupported init expr in data section: {:?}",
-                                        s
-                                    )));
-                                }
-                            };
+                            let (offset, escaped) = ConstExpr::from_wasmparser(offset_expr)?;
+                            debug_assert!(escaped.is_empty());
 
                             initializers.push(MemoryInitializer {
                                 memory_index,
-                                base,
                                 offset,
                                 data: range,
                             });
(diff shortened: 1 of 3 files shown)


crates/environ/src/compile/module_environ.rs
```diff
@@ -949,30 +936,54 @@ impl ModuleTranslation<'_> {
...
+            }
+
+            fn eval_offset(&mut self, memory_index: MemoryIndex, expr: &ConstExpr) -> Option<u64> {
+                let mem64 = self.module.memory_plans[memory_index].memory.memory64;
+                match expr.ops() {
+                    &[ConstOp::I32Const(offset)] if !mem64 => Some(offset.unsigned().into()),
+                    &[ConstOp::I64Const(offset)] if mem64 => Some(offset.unsigned()),
+                    _ => None,
+                }
+            }
```
fitzgen, line 957: It might be better to turn `ConstEvalContext` into a trait -- I initially did that but then every use site was the same so I made it concrete. But if it were a trait, then we could use a `Result` for getting globals and funcrefs and return an `Err` when they aren't available at compile time. This would allow both reusing `eval` here so you don't have to do this matching (nor in the other PR, iirc) and also support const expression offsets like `(i32.add (i32.const 3) (i32.const 5))` for compile-time heap image computation.

crates/wasmtime/src/runtime/vm/instance/allocator.rs
```diff
@@ -695,35 +677,69 @@ fn initialize_memories(instance: &mut Instance, module: &Module) -> Result<()> {
...
     // a trap is returned.
-    let ok = module.memory_initialization.init_memory(
-        instance,
-        InitMemory::Runtime {
-            memory_size_in_pages,
-            get_global_as_u64,
-        },
-        |instance, memory_index, init| {
+
+    struct InitMemoryRuntime<'a> {
```
fitzgen, line 681: nitpicky suggestion, take it or leave it: I'd personally rename this `InitMemoryAtRunTime` and the other impl as `InitMemoryAtCompileTime`. I think it is a little more clear but maybe it is just me.

jameysharp: I wasn't entirely happy with the names I picked and considered doing something like you suggest. So it's not hard to convince me on this one!

crates/environ/src/compile/module_environ.rs
jameysharp: It turns out it's easy to turn `ConstEvalContext` into a trait, but it's not useful without bigger changes that I haven't been able to figure out yet.

`ConstExprEvaluator` can only be used within the wasmtime crate, because it relies on `ValRaw`. Everything that does const evaluation in that crate already has access to an `Instance`. All the places that need const evaluation before instantiation are in wasmtime-environ or wasmtime-types.

I tried factoring out a type-specific, integer-only const-evaluator that can live in wasmtime-types alongside `ConstExpr` and `ConstOp`. But the general-purpose const-evaluator is still necessary for initializing globals and I'm not real happy with duplicating this functionality.

Maybe we need a trait in wasmtime-types that `Val` or `ValRaw` can implement?

jameysharp:
I've decided it's more complicated than I want to tackle in this PR to make const-eval available at compile time, so I hope to come back to that later. I've renamed types as @fitzgen suggested (though I decided I liked `AtInstantiation` better than `AtRunTime`).

And I've spent a couple days thinking about this and decided I don't have particularly better ideas for anything in here, so I've marked it ready for review now.

crates/environ/src/compile/module_environ.rs
fitzgen: > Maybe we need a trait in wasmtime-types that `Val` or `ValRaw` can implement?

Yeah I think we'll have to do something roughly like this.

fitzgen (approved):

PR #8518, opened 2024-05-01
Reassociate mixture of add and sub to wide trees

KGrewal1:
As in https://github.com/bytecodealliance/wasmtime/pull/7466 , converts deep nesting of addition and subtraction into a wide tree eg:

$((a-b)-c)-d)=(a-b)-(c+d)$

again using subsume to prevent blow up due to reassociating large chains


abrown (approved):
Looks fine to me but does anyone with ISLE conventions paged in want to comment: @cfallin, @elliottt, @jameysharp?

PR #8519, opened 2024-05-01
wasi-nn: remove Git submodule

abrown:
To more closely align with the conventions in the `wasmtime-wasi` and `wasmtime-wasi-http` crates, this change removes the Git submodule that previously provided the WIT and WITX files for `wasmtime-wasi-nn`. Like those other crates, the syncing of wasi-nn WIT and WITX files will happen manually for the time being. This is the first PR towards upgrading the wasi-nn implementation to match recent spec changes and better preview2-ABI compatibility.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,6 @@
 [submodule "tests/spec_testsuite"]
 	path = tests/spec_testsuite
 	url = https://github.com/WebAssembly/testsuite
-[submodule "crates/wasi-nn/spec"]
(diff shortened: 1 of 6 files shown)


alexcrichton:
Could you update [this script](https://github.com/bytecodealliance/wasmtime/blob/main/ci/vendor-wit.sh) to additionally vendor this WIT? That way we ideally have a single knob to flip to update when necessary and CI checks that we're matching upstream.

alexcrichton (approved):

PR #8524, opened 2024-05-02
cranelift: Rewrite operands with allocations in-place

jameysharp:
This makes all other code that uses AllocationConsumer into no-ops because the list of allocations in all other places are now always empty.

There is a bunch of dead code still left over after this. This part is just the stuff I can delete without changing function signatures everywhere. The various functions which are now no-ops are called in a lot of places and cleaning those up will take more careful review.


jameysharp:
Instead of reusing `AllocationConsumer` as the new implementer of `OperandVisitor`, I wonder if I should provide an `OperandVistor` implementation for `FnMut` and just use a closure before `do_emit`. Then the future PR that deletes the remaining dead code here will completely delete `AllocationConsumer` as well.

elliottt (approved):
This is awesome! Really looking forward to seeing the follow up prs that remove all that dead code :D

PR #8525, opened 2024-05-02
Respect pooling allocation options in `wasmtime serve`

### LINKED ISSUE #8504: `-W max-table-elements` has no effect for `wasmtime serve`
While working on resource support for the C# `wit-bindgen` generator, I created [an example app](https://github.com/dicej/dotnet9-wasi-http-example) targeting `wasi:http/incoming-handler@0.2.0`.  However, `wasmtime serve` couldn't run it due to one of the tables containing more than 10000 entries:

```
 $ wasmtime serve target/stripped.wasm 
Error: table index 0 has a minimum element size of 10511 which exceeds the limit of 10000
```

Next, I tried `wasmtime serve -W max-table-elements=20000 target/stripped.wasm` instead, but got the same error.  Looking at [the code](https://github.com/bytecodealliance/wasmtime/blob/3aa3206484f6242a2b1d34decce2e4558e69aebc/src/commands/serve.rs#L251-L266), it seems that the pooling allocator is enabled by default for `wasmtime serve`, and since it has its own configuration knobs which are not controlled by e.g. `-W max-table-elements`, there's no way to change the limit from its default.  There's also no way I can see to disable the pooling allocator, given that the only reference to the `--pooling-allocator` option is in `old_cli.rs`, which is disabled in the default build.

I'd suggest that `-W` options such as `max-table-elements` should be applied to the pooling allocator when it is enabled.  ~~We might also consider raising the default considering that even simple C# apps tend to exceed it (although we could instead consider that a bug in the C# Native AOT compiler, e.g. insufficient dead code elimination?).~~ EDIT: I was building with `-c Debug`; the release build results in a table about half the size, which is well under the current default limit.

alexcrichton:
This commit updates the processing of pooling allocator options in `wasmtime serve`. Previously the pooling allocator was enabled by default but the options to configure it weren't processed due to how this default-enable was implemented. The option to enable it by default for `wasmtime serve`, but only `wasmtime serve`, is now processed differently in a way that handles various other
pooling-allocator-related options.

Closes #8504

### DIFF
--- a/crates/bench-api/src/lib.rs
+++ b/crates/bench-api/src/lib.rs
@@ -435,7 +435,7 @@ impl BenchState {
         execution_end: extern "C" fn(*mut u8),
         make_wasi_cx: impl FnMut() -> Result<WasiCtx> + 'static,
     ) -> Result<Self> {
-        let mut config = options.config(Some(&Triple::host().to_string()))?;
+        let mut config = options.config(Some(&Triple::host().to_string()), None)?;
         // NB: always disable the compilation cache.
         config.disable_cache();
         let engine = Engine::new(&config)?;
--- a/crates/cli-flags/src/lib.rs
+++ b/crates/cli-flags/src/lib.rs
@@ -421,7 +421,11 @@ impl CommonOptions {
         Ok(())
     }
 
-    pub fn config(&mut self, target: Option<&str>) -> Result<Config> {
+    pub fn config(
+        &mut self,
+        target: Option<&str>,
+        pooling_allocator_default: Option<bool>,
+    ) -> Result<Config> {
         self.configure();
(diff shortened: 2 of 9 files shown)


alexcrichton:
I've made the decision to ignore the test I've added here with `#[ignore]`. The CI failure shows a bug where we're racing between two events:

* The `wasmtime serve` process is printing out error information
* The test process is running `child.kill()` to reap status information as the test is over

There's no way to resolve this race with `wasmtime serve` right now as there is no graceful shutdown of `wasmtime serve`. I'd like to add that in the future (e.g. ctrl-c waits for all active connections), but that's a bit out of scope for this. In the meantime it still seems good to get a fix in at least.

PR #8527, opened 2024-05-02
Double the default allowed table elements

alexcrichton:
This commit doubles the default allowed table elements per table in the pooling allocator from 10k to 20k. This helps to, by default, run the module produced in #8504.


fitzgen (approved):
Do we document the defaults in some `config.rs` doc comment? Should double check and update the docs if necessary.

alexcrichton:
Good catch 👍

PR #8530, opened 2024-05-02
wasi-nn: only support components

abrown:
This is a relatively drastic change to remove all the preview1, WITX-based support from Wasmtime's wasi-nn implementation. This has ramifications:
- Wasmtime will not be able to run core modules that target wasi-nn, either through the CLI or the embedding API
- Sightglass will not be able to benchmark wasi-nn-using code until the `bench-api` learns how to handle components
- the [`wasi-nn`] bindings crate is no longer usable with Wasmtime (it only understands the preview1 ABI)
- all wasi-nn testing in Wasmtime will now use components (the `examples` are left for a later refactor)

Why this change, then? The WebAssembly [specification] has pushed ahead, defining new functionality that depends on component model features (e.g., resources). So far, the `wasmtime-wasi-nn` has had its preview1- and preview2-ABI implementations coexisting side by side, `wit.rs` and `witx.rs`. This was only possible because the WIT and WITX definitions were roughly similar. But I plan to update the Wasmtime implementation to support the new spec changes, which means the older preview1-ABI WITX code would immediately be out of date. And the differences between the updated WIT code and the old WITX code would only grow over time.

[`wasi-nn`]: https://crates.io/crates/wasi-nn
[specification]: https://github.com/WebAssembly/wasi-nn


alexcrichton:
From a technical perspective this all looks fine to me, thanks! For a "is this the time to do this or not" I'd defer to the wasi-nn subgroup, which I know you're intimately tied into as well and likely have a consensus before making this PR

abrown:
Superseded by #8873.

PR #8531, opened 2024-05-02
wasmtime: Make table lazy-init configurable

jameysharp:
Lazy initialization of tables has trade-offs that we haven't explored in a while. Making it configurable makes it easier to test the effects of these trade-offs on a variety of WebAssembly programs, and allows embedders to decide whether the trade-offs are worth-while for their use cases.

TODO:
- [x] Finish documentation comments
- [x] Add CLI flag in the `-O` group
- [x] Either implement support in Winch, or check in `Config::validate` that when Winch is enabled so is lazy-init
- [x] Add `disas` tests with lazy init disabled
- [x] Hook up this flag in fuzzing

I've tested this by temporarily setting the flag to disable lazy-init by default (in `crates/environ/src/tunables.rs`) and verifying that the test suite still passes, except for a bunch of Winch tests (as expected), as well as the amusing case of `code_too_large::code_too_large_without_panic`, which fails because disabling lazy-init deletes enough code that the code is no longer too large.

With lazy-init disabled, the `disas` tests show that this shaves off six or seven CLIF instructions and a cold block per `call_indirect` or `table_get`. The `typed-funcrefs.wat` test becomes branch-free, with two loads and an indirect call per `call_indirect`, making it equivalent to the version that uses globals instead of tables.

cc: @cfallin

### DIFF
--- a/crates/wasmtime/src/runtime/vm/table.rs
+++ b/crates/wasmtime/src/runtime/vm/table.rs
@@ -14,7 +14,7 @@ use core::ptr::{self, NonNull};
 use core::slice;
 use sptr::Strict;
 use wasmtime_environ::{
-    TablePlan, Trap, WasmHeapTopType, WasmRefType, FUNCREF_INIT_BIT, FUNCREF_MASK,
+    TablePlan, TableStyle, Trap, WasmHeapTopType, WasmRefType, FUNCREF_INIT_BIT, FUNCREF_MASK,
 };
 
 /// An element going into or coming out of a table.
@@ -113,18 +113,24 @@ impl TaggedFuncRef {
 
     /// Converts the given `ptr`, a valid funcref pointer, into a tagged pointer
     /// by adding in the `FUNCREF_INIT_BIT`.
-    fn from(ptr: *mut VMFuncRef) -> Self {
-        let masked = Strict::map_addr(ptr, |a| a | FUNCREF_INIT_BIT);
-        TaggedFuncRef(masked)
+    fn from(ptr: *mut VMFuncRef, lazy_init: bool) -> Self {
+        if lazy_init {
+            let masked = Strict::map_addr(ptr, |a| a | FUNCREF_INIT_BIT);
+            TaggedFuncRef(masked)
+        } else {
+            TaggedFuncRef(ptr)
+        }
     }
 
     /// Converts a tagged pointer into a `TableElement`, returning `UninitFunc`
     /// for null (not a tagged value) or `FuncRef` for otherwise tagged values.
-    fn into_table_element(self) -> TableElement {
+    fn into_table_element(self, lazy_init: bool) -> TableElement {
         let ptr = self.0;
-        if ptr.is_null() {
+        if lazy_init && ptr.is_null() {
             TableElement::UninitFunc
         } else {
+            // Masking off the tag bit is harmless whether the table uses lazy
+            // init or not.
             let unmasked = Strict::map_addr(ptr, |a| a & FUNCREF_MASK);
             TableElement::FuncRef(unmasked)
         }
@@ -156,6 +162,8 @@ pub struct StaticFuncTable {
     data: SendSyncPtr<[FuncTableElem]>,
     /// The current size of the table.
     size: u32,
+    /// Whether elements of this table are initialized lazily.
+    lazy_init: bool,
 }
 
 pub struct StaticGcRefTable {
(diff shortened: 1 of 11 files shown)


crates/wasmtime/src/runtime/vm/table.rs
```diff
@@ -701,29 +726,60 @@ impl Table {
         self.element_type().matches(val)
     }
 
-    fn funcrefs(&self) -> &[TaggedFuncRef] {
+    fn funcrefs(&self) -> (&[TaggedFuncRef], bool) {
```
alexcrichton, line 729: What do you think about perhaps returning `enum Funcrefs<'a> { Tagged(&'a [TaggedFuncRef]), Untagged(&'a [*mut VMFuncRef]) }` here? That feels a bit more "type safe" if a bit wordier and reflects how this is a boolean decision at runtime as well

winch/codegen/src/visitor.rs
```diff
@@ -1428,7 +1429,10 @@ where
 
         match heap_type {
             WasmHeapType::Func => match style {
-                TableStyle::CallerChecksSignature => self.emit_lazy_init_funcref(table_index),
+                TableStyle::CallerChecksSignature { lazy_init: true } => {
+                    self.emit_lazy_init_funcref(table_index)
+                }
+                _ => unimplemented!("Support for eager table init"),
```
alexcrichton, line 1435: If this remains (which I think is fine) mind adding a check in `Config::validate` that returns an error if winch is enabled but lazy init is disabled?

alexcrichton (commented):
I like the idea of adding this, good idea!

Some other bits I might recommend:

* Mind adding a CLI flag under `-O` for this as well?
* For testing it might also be good enough to "just" hook this up to fuzzing
* Want to add a `disas` test or two with this flag?

crates/wasmtime/src/runtime/vm/table.rs
jameysharp: I like this idea, but while testing it out, I found so many more things I want to change about how the lazy-tagging invariants are maintained in these modules that I don't think I want to change any of them in this PR. :sweat_smile: A few:
- Getting a table reference should return enough information that the caller can initialize any elements it needs to, instead of passing in a range of indexes to init.
- `Table::get` and table copies should take `&mut src` and init the elements they need in-place; the caller always has a mutable pointer.
- Therefore, there's no need for `funcrefs()`, only `funcrefs_mut()`.
- `StaticFuncTable` should maybe put the current size in the `data` slice and the capacity in a separate field; it's currently the other way around.
- `funcrefs_mut()` is shorter and easier to reason about if it first gets `elements.as_mut_slice()`/`unsafe { data.as_mut() }` depending on whether the table is static or dynamic, and then afterwards does the unsafe `slice::from_raw_parts_mut`. I mixed up `*mut [T]` with `*mut [*mut [T]]` at one point and was afraid the `.cast()` would just silently "fix" it for me
- I'm imagining something resembling `MaybeUninit` as an interface for tracking at the type level whether it's okay to read a range of the table

alexcrichton: That all sounds great to me yeah, and seems fine to leave this as-is returning a bool in the meantime

jameysharp:
I think I've addressed all of your suggestions, Alex; thank you for them!

I'm trying to coax the fuzzer to discover an artificial bug I added in the non-lazy-init configuration just to make sure it can, and it's been running for a while without finding it. So maybe you (or whoever reviews this) can double-check if I did something wrong in the fuzzer configuration?

jameysharp:
Okay, I've learned that the `table_ops` fuzzer can't test this because it only uses `externref` tables, but lazy-init only applies to `funcref` tables. And the `single-inst` module-builder for `differential` doesn't generate tables or table instructions at all.

But I'm still not sure why `wasm-smith` isn't finding my asserts with either `differential` or `instantiate`. I tried both asserting that lazy-init is enabled, and that it's disabled, during codegen in wasmtime-cranelift, and neither one is getting hit.

jameysharp:
Nevermind, `instantiate` found my assert a few minutes after I wrote that, and then found the opposite assert after another 20 minutes. And I've also once again run `cargo test --test all` with `table_lazy_init` set false, and the only tests that failed were Winch-related (now with the error message from `Config::validate`) and the `code_too_large_without_panic` test again.

So I think this is ready to land.

fitzgen:
We should really add a top-down generation mode to `wasm-smith`...

PR #8532, opened 2024-05-03
Use rustdoc's `doc_auto_cfg` feature instead of `doc_cfg`

alexcrichton:
This commit removes all our `#[cfg_attr(..., doc(cfg(...)))]` annotations throughout Wasmtime and `wasmtime-wasi`. These are all replaced with `feature(doc_auto_cfg)` which automatically infers the attribute to show rather than requiring us to duplicate it. Spot-checking the docs this looks just-as-readable while being much easier to maintain over time.


fitzgen (approved):
Nice!

PR #8533, opened 2024-05-03
Add support for `#![no_std]` to the `wasmtime` crate

alexcrichton:
This commit is the final technical piece of #8341 which makes the `wasmtime` crate itself compatible with `#![no_std]`. This means that the crate can now be compiled on targets that don't have `std` when the activated features are a subset of `{runtime, gc, component-model}`. Support for Wasmtime requires an implementation of `wasmtime-platform.h` which is something I'd like to also start moving to release artifacts after this as well.

This PR is split into a series of commits with the final one being the "big one" and the others are minor refactorings leading up to it. The major changes made here are:

* The "custom" platform of Wasmtime is now enabled by default if nothing else is applicable. This means that there's no opt-in required for no_std support.
* Embedders can now define a custom feature detection function, for example "does this host support AVX". This is because the default implementation uses the standard library while that's not available with `no_std`.
* Internal idioms in `wasmtime` have change to be similar to previous commits for `wasmtime-environ` and other crates, for example. For example imports come from `core` and `alloc` and there's a custom prelude for core + alloc bits and pieces.
* Wasmtime now unconditionally uses `hashbrown::HashMap` for some items such as `Linker` and GC sets.
* The `wasmtime` crate has a few "custom synchronization primitives" which have real implementations backed by other crates in `std` mode or "dummy" implementations in `no_std` mode that effectively assert for no contention. This might require updates to `wasmtime-platform.h` in the future instead.
* Libcall intrinsics for float-related functionality are implemented with the `libm` crate in `no_std` mode. In `std` mode they still use standard library methods.

The main remaining piece I'd like to tackle after this is an update of the `no_std` documentation for Wasmtime. The page is outdated now and could use some updates. I'll also note that I still have yet to test all these changes beyond what CI is already doing. We don't actually test `no_std` mode in CI, only that it builds. If anything comes up though I'll work through that in follow-ups.


crates/wasmtime/src/config.rs
```diff
@@ -2042,6 +2057,43 @@ impl Config {
         self.macos_use_mach_ports = mach_ports;
         self
     }
+
+    /// Configures an embedder-provided function, `detect`, which is used to
+    /// determine if an ISA-specific feature is available on the current host.
+    ///
+    /// This function is used to verify that any features enabled for a compiler
+    /// backend, such as AVX support on x86\_64, are also available on the host.
+    /// It is undefined behavior to execute an AVX instruction on a host that
```
abrown, line 2066: Undefined behavior... in Wasmtime? I thought we could be pretty sure we'd get a SIGILL or the like?

alexcrichton: I'll admit I don't necessarily fully understand this, but my reasoning here is derived from Rust's stance that executing intrinsics your CPU doesn't have support for is undefined behavior. There was a fair amount of discussion about this when first stabilizing SIMD in Rust and I believe the general rationale boils down to:

* There's not an ironclad guarantee that unsupported instructions raise a SIGILL for all unsupported instructions. (at least at the time what was considered was all possible extensions to all possible ISAs)
* There can be "compiler weirdness" where once a feature is enabled it has subtle changes in other parts like assemblers and other chosen instructions which can cause subtle incompatibilities on systems that don't support the instructions.

Overall I don't think that there's a super simple "this is why it's unsafe" thing I can point to to justify this. I am relatively confident in saying, though, that we're probably not guaranteed a SIGILL for all possible encodings of all instructions we could use on all architectures. Coupling that with this is a niche method most won't use is mostly why I figured it'd be ok to mark this as `unsafe`

fitzgen:
(Taking a look, might be a little while to get through everything here)

crates/wasmtime/src/config.rs
bjorn3: `lzcnt` is encoded as `rep bsr`, which has a redundant prefix. Redundant prefixes tend to be ignored, but are not guaranteed to be. As such `lzcnt` is interpreted by older cpu's as `bsr`. This bit me once in cg_clif: https://github.com/rust-lang/rustc_codegen_cranelift/pull/951#issuecomment-663855992 See also https://stackoverflow.com/questions/25683690/confusion-about-bsr-and-lzcnt/43443701#43443701

crates/wasmtime/src/runtime/module.rs
```diff
@@ -336,7 +339,11 @@ impl Module {
     /// This is because the file is mapped into memory and lazily loaded pages
     /// reflect the current state of the file, not necessarily the original
     /// state of the file.
+<<<<<<< HEAD
     #[cfg(any(feature = "cranelift", feature = "winch"))]
+=======
+    #[cfg(all(feature = "std", any(feature = "cranelift", feature = "winch")))]
+>>>>>>> 025923159 (Add `#![no_std]` support to the `wasmtime` crate)
```
fitzgen: conflict

crates/wasmtime/src/runtime/stack.rs
```diff
@@ -1,6 +1,6 @@
-use std::{ops::Range, sync::Arc};
-
+use crate::prelude::*;
 use anyhow::Error;
+use std::{ops::Range, sync::Arc};
```
fitzgen, line 3: `std`?

crates/wasmtime/src/runtime/store.rs
```diff
@@ -1952,8 +1956,22 @@ from going any further and to alert what's going on. If this is a security
...
+                // Without the standard library but with `panic=abort` then
+                // it's safe to panic as that's known to halt execution. For
+                // now avoid the above error message as well since without
+                // `std` it's probably best to be a bit more size-conscious.
+                let _ = pc;
+                panic!("invalid fault");
+            } else {
+                // Without `std` and with `panic = "unwind"` there's no way to
+                // abort the process portably, so flag a compile time error.
+                compile_error!("either `std` or `panic=abort` must be enabled");
```
fitzgen, line 1971: FWIW, we could add another hook to the system interface header for this, if it is ever requested. Might be worth adding a comment about.

crates/wasmtime/src/runtime/vm/sys/unix/signals.rs
```diff
@@ -308,7 +308,7 @@ pub fn lazy_per_thread_init() {
     // This thread local is purely used to register a `Stack` to get deallocated
     // when the thread exists. Otherwise this function is only ever called at
     // most once per-thread.
-    thread_local! {
+    std::thread_local! {
```
fitzgen, line 311: Why is `std` okay to use here? Shouldn't this use the `wasmtime_tls_{get,set}` stuff?

fitzgen: I guess if we are on `unix` then we know we have `std` available?

crates/wasmtime/src/sync_nostd.rs
```diff
@@ -0,0 +1,249 @@
...
+    pub fn get_or_try_init<E>(&self, f: impl FnOnce() -> Result<T, E>) -> Result<&T, E> {
+        if let Some(ret) = self.get() {
+            return Ok(ret);
+        }
+        self.try_init(f)
+    }
+
+    fn get(&self) -> Option<&T> {
+        match self.state.load(Ordering::Acquire) {
+            INITIALIZED => Some(unsafe { (*self.val.get()).assume_init_ref() }),
```
fitzgen, line 53: This isn't going to accidentally bind `INITIALIZED` as a fresh variable, right? I always forget the details of when that can happen or not... Might be better to use an `if`/`else` here since there is just one checked-for case and one default case.

crates/wasmtime/src/sync_nostd.rs
```diff
@@ -0,0 +1,249 @@
+//! Synchronization primitives for Wasmtime for `no_std`.
+//!
+//! These primitives are intended for use in `no_std` contexts are are not as
+//! full-featured as the `std` brethren. Namely these panic and/or return an
+//! error on contention. This serves to continue to be correct in the fact of
+//! actual multiple threads, but if a system actually has multiple threads then
+//! something will need to change in the Wasmtime crate to enable the external
+//! system to perform necessary synchronization.
+//!
+//! See a brief overview of this module in `sync_std.rs` as well.
```
fitzgen, line 10: We could also add hooks to the system interface header to help implement these and push the choice of panicking/aborting or actually doing synchronization out tot he embedder as well down the line. Again, might be worth a comment.

crates/wasmtime/src/runtime/module.rs
```diff
@@ -339,11 +339,7 @@ impl Module {
     /// This is because the file is mapped into memory and lazily loaded pages
     /// reflect the current state of the file, not necessarily the original
     /// state of the file.
-<<<<<<< HEAD
-    #[cfg(any(feature = "cranelift", feature = "winch"))]
-=======
     #[cfg(all(feature = "std", any(feature = "cranelift", feature = "winch")))]
->>>>>>> 025923159 (Add `#![no_std]` support to the `wasmtime` crate)
```
fitzgen: Ah okay, you got it in a later commit

fitzgen (approved):
ship it!

crates/wasmtime/src/sync_nostd.rs
alexcrichton: Yeah that was my thinking as well in terms of moving it into the header, good point on leaving a comment though

crates/wasmtime/src/sync_nostd.rs
alexcrichton: Nah yeah we're ok to have the `match` here, I generally rely on the unused variables lint to catch mistakes like this. (I'll go ahead and switch it anyway to an `if`/`else` here)

crates/wasmtime/src/runtime/store.rs
alexcrichton: Yeah I think that'd work out too, but my hope is that all std-using platforms can funnel into the first case with eprintln and most no_std platforms probably don't want unwinding anyway

crates/wasmtime/src/runtime/stack.rs
alexcrichton: Yeah any modules which are gated on the `std` feature still import from `std` and don't need to get switched over. Here this module is only used for `async` which requires the `std` feature at this time, so we'll need to change this in the future to an import from `core` if we want the `async` feature to work on no_std for example.

crates/wasmtime/src/runtime/vm/sys/unix/signals.rs
alexcrichton: Yeah that was my thinking, on platforms that are known to have the standard library we go ahead and just import from the standard library. Things will get a bit messier otherwise because it's not a great experience for Wasmtime to depend on `#[repr(C)]` symbols silently so I'm hoping to keep it to a minimum.

Issue #7954, opened 2024-02-16
cranelift: Use `DominatorTreePreorder` in more places


jameysharp:
#### Feature

We have two interfaces in Cranelift for navigating [dominator](https://en.wikipedia.org/wiki/Dominator_(graph_theory)) trees, both defined in the `cranelift_codegen::dominator_tree` module: `DominatorTree` and `DominatorTreePreorder`. But we weren't using the latter outside of tests after #3434 landed in 2021, until I switched the egraph pass over to using it in #7948 today.

1. We should audit all uses of `DominatorTree` to see if they would be better served by using `DominatorTreePreorder` instead.
2. If we end up using `DominatorTreePreorder` in more places than just the egraph pass, then we should compute it once and share it.
3. If it turns out that we always need a `DominatorTreePreorder` when compiling any function, then we should fold the two types into one and always compute the preorder when we're computing the dominator tree.

#### Benefit

These two interfaces both provide a method named `dominates` which checks whether one basic block dominates another. However, `DominatorTree` does this in time proportional to the length of the path from one block to the other within the dominator tree. Thanks to a linear-time preprocessing step performed once, `DominatorTreePreorder` can answer this question in constant time.

So if we're using the `DominatorTree::dominates` method anywhere that's performance-critical, switching to `DominatorTreePreorder` could provide an asymptotic-complexity improvement.

On top of that, sharing a precomputed preorder across multiple uses saves time redoing the preprocessing step and also may allow us to reuse a heap allocation for the temporary storage used during that preprocessing step.

#### Implementation

To start with, search for all uses of `DominatorTree::dominates`. For each one, see if we can just replace it with `DominatorTreePreorder::dominates`.

This is easy if both arguments are `Block` IDs, but either one is currently also allowed to be an instruction ID (`Inst`) or a `ProgramPoint`. If we're relying on that feature somewhere, it's only slightly more complicated: If two instructions are in the same block then the earlier instruction dominates the later instruction; otherwise we can go back to the easy case and compare the blocks they're in to see if one block dominates the other.

If some instances of `DominatorTree` are only being used to call `dominates`, then removing that structure from those instances in favor of `DominatorTreePreorder` is the next step. However, some cases may also be using other methods such as `cfg_postorder` or `idom`, which are not available on `DominatorTreePreorder`.

#### Alternatives

We can always leave this alone, but I think it's a good source of small changes that may give us performance improvements during compilation.

MuhtasimTanmoy:
[...]

PR #8535, opened 2024-05-03
converted DominatorTree to DominatorTreePreorder in alias_analysis file

VamshiReddy02:
ref: #7954 
Hey @jameysharp,
I made a few changes; this is my first time contributing to wasmtime. please let me know if there are any changes to make.

### DIFF
--- a/cranelift/codegen/src/dominator_tree.rs
+++ b/cranelift/codegen/src/dominator_tree.rs
@@ -464,6 +464,20 @@ impl DominatorTreePreorder {
         }
     }
 
+    /// Checks if one instruction dominates another.
+    pub fn dominates_inst(&self, a: Inst, b: Inst, layout: &Layout) -> bool {
+        match (layout.inst_block(a), layout.inst_block(b)) {
+            (Some(block_a), Some(block_b)) => {
+                if block_a == block_b {
+                    layout.pp_cmp(a, b) != Ordering::Greater
+                } else {
+                    self.dominates(block_a, block_b)
+                }
+            }
+            _ => false,
+        }
+    }
+
     /// Recompute this data structure to match `domtree`.
     pub fn compute(&mut self, domtree: &DominatorTree, layout: &Layout) {
         self.nodes.clear();
--- a/cranelift/codegen/src/alias_analysis.rs
+++ b/cranelift/codegen/src/alias_analysis.rs
@@ -275,12 +275,12 @@ impl<'a> AliasAnalysis<'a> {
             state,
             func.dfg.insts[inst],
         );
-    
+
         let replacing_value = if let Some((address, offset, ty)) = inst_addr_offset_type(func, inst)
         {
             let address = func.dfg.resolve_aliases(address);
             let opcode = func.dfg.insts[inst].opcode();
-    
+
             if opcode.can_store() {
                 let store_data = inst_store_data(func, inst).unwrap();
                 let store_data = func.dfg.resolve_aliases(store_data);
@@ -298,7 +298,7 @@ impl<'a> AliasAnalysis<'a> {
                     mem_loc
                 );
                 self.mem_values.insert(mem_loc, (inst, store_data));
-    
+
                 None
             } else if opcode.can_load() {
                 let last_store = state.get_last_store(func, inst);
@@ -316,7 +316,7 @@ impl<'a> AliasAnalysis<'a> {
                     last_store.map(|inst| inst.index()).unwrap_or(usize::MAX),
                     mem_loc
                 );
-    
+
                 // Is there a Value already known to be stored
                 // at this specific memory location?  If so,
                 // we can alias the load result to this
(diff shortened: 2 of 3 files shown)


abrown:
@jameysharp, sounds like you have some previous context for this change?

jameysharp:
I do, thanks Andrew! I proposed doing this in #7954.

@VamshiReddy02, this is a good start! I'm excited that you're working on it.

I labeled this issue "easy" but there are some things that are not so easy about it, and you've started with one of the more difficult cases: Comparing two instructions, instead of two blocks. That's okay! Let's figure out how to make it work.

Instruction `a` dominates instruction `b` if either of these conditions are true:
1. `a` and `b` are in the same block, and `a` is not after `b`. Here's an example showing how to check if `inst_a` is not after `last`: https://github.com/bytecodealliance/wasmtime/blob/71800cc1a1daf01e8c3cfee09e651cf8f46ff31b/cranelift/codegen/src/dominator_tree.rs#L137
2. Or the block containing `a` dominates the block containing `b`.

You've implemented the second condition, but not the first.

The first thing I think you should do is add a new method to `DominatorTreePreorder` which answers the question of whether one instruction dominates another. I suggest this method should look like `pub fn dominates_inst(&self, a: Inst, b: Inst, layout: &Layout) -> bool`, and it needs to implement the above condition.

Then you can call `dominates_inst` in `alias_analysis.rs`. I think this should make the changes in `alias_analysis.rs` very small. This method will also be useful for the calls to `dominates` that are in `cranelift/codegen/src/verifier/mod.rs`. The only other calls are in `cranelift/codegen/src/loop_analysis.rs`, which can just compare blocks, but it's a little complicated to explain why.

I hope that helps. One more request:

Could you run `cargo fmt` before committing changes? That will clean up changes to whitespace and make your pull requests easier to review.

Thank you for working on this!

VamshiReddy02:
Hey @jameysharp,
Thank you for the feedback. Correct me if I am wrong, First I need to add a new method called `pub fn dominates_inst(&self, a: Inst, b: Inst, layout: &Layout) -> bool` in the `DominatorTreePreorder` applying all the above conditions. For example like this: 
```console
impl DominatorTreePreorder {
 pub fn dominates_inst(&self, a: Inst, b: Inst, layout: &Layout) -> bool {
        match (layout.inst_block(a), layout.inst_block(b)) {
            (Some(block_a), Some(block_b)) => {
                if block_a == block_b {
                    layout.pp_cmp::<ProgramPoint, ProgramPoint>(a.into(), b.into()) != Ordering::Greater
                } else {
                    self.dominates(block_a, block_b)
                }
            }
            _ => false,
        }
    }
}
```
Then I need to call the `dominates_inst` in `alias_analysis.rs` file, for example like this:

```console
let aliased =
                    if let Some((def_inst, value)) = self.mem_values.get(&mem_loc).cloned() {
                        trace!(
                            " -> sees known value v{} from inst{}",
                            value.index(),
                            def_inst.index()
                        );
                        if let (Some(_def_block), Some(_inst_block)) = (
                            func.layout.inst_block(def_inst),
                            func.layout.inst_block(inst),
                        ) {
                            if self.domtree.`dominates_inst(def_inst, inst, &func.layout)` {
                                trace!(
                                    " -> dominates; value equiv from v{} to v{} inserted",
                                    load_result.index(),
                                    value.index()
                                );
                                Some(value)
                            } else {
                                None
                            }
                        } else {
                            None
                        }
                    } else {
                        None
                    };
```
Am I going in the correct direction?

jameysharp:
Yes, that is the correct direction! There are a few ways to make this easier though:

1. In `alias_analysis.rs`, you don't need to call `inst_block`, because now you don't need to know the blocks there. So you can delete that `if` in the middle.

2. Instead of explicitly specifying the type of `pp_cmp`, you can let Rust figure it out. `layout.pp_cmp(a, b)` should work. Just don't call `.into()` in this case, because that gives the compiler too many choices and it gets confused.

3. When `dominates_inst` is called, I happen to know that the instructions should be in the `Layout`, so `inst_block` should not fail. As a result, instead of checking whether `inst_block` returns `Some`, you can use `expect`, like this: https://github.com/bytecodealliance/wasmtime/blob/81a89169f556d2ba8c6ddc75ae0ac471f213f1c3/cranelift/codegen/src/dominator_tree.rs#L133-L135

But yes, you are most of the way there! I look forward to seeing your updated pull request.

Some other projects expect "perfect" git commits, so I just want to let you know that we don't care about that for Wasmtime and Cranelift. As you make further changes, just commit them and push them to your `DominatorTree` branch. You don't need to open a new PR or use `git rebase` or anything. We'll "squash" your changes together into a single commit once they're ready to merge.

VamshiReddy02:
Hey @jameysharp, I made some changes, let me know if any changes need to be made.

cranelift/codegen/src/dominator_tree.rs
```diff
@@ -464,6 +464,20 @@ impl DominatorTreePreorder {
         }
     }
 
+    /// Checks if one instruction dominates another.
+    pub fn dominates_inst(&self, a: Inst, b: Inst, layout: &Layout) -> bool {
```
jameysharp, line 468: This is a minor thing, but I would prefer to have `dominates_inst` next to the `dominates` method that it uses. Then someone looking at one of them can easily see the other one too.

jameysharp (commented):
This is almost exactly what I wanted! Let's just figure out why the test suite is failing in CI.

Based on the logs from CI, you should be able to reproduce the test failures by running `cargo test -p wasi-common --test all`. It looks like Cranelift's tests all passed, which I think means we need some more tests…

The key error messages I'm looking at say "cranelift/codegen/src/egraph/elaborate.rs:691:21: something has gone very wrong if we are elaborating effectful instructions, they should have remained in the skeleton". This is not very helpful if you don't already know how the egraph optimization pass works, but as you make changes you can at least check whether this message changes or goes away.

My guess is that what happened is that you've called `DominatorTreePreorder::new` but it's uninitialized and empty. I looked at the implementation of `DominatorTreePreorder::dominates` and I think it always returns `true` when it isn't initialized. This seems like a reasonable explanation for the test failures to me: if it falsely reports that things dominate each other when they actually don't, then we would "optimize" in places where we aren't actually allowed to.

The first thing I would like you to do is add this to the existing `DominatorTreePreorder::dominates` function, and commit this change because if other people have the same problem this will help them figure it out more quickly:

```rust
// Check that both blocks are reachable.
debug_assert!(na.pre_number != 0);
debug_assert!(nb.pre_number != 0);
```

When you do that, I predict that more tests will start failing at these asserts. In particular, I hope that `cargo test -p cranelift-tools` will have some failing tests.

If you can confirm that those asserts fail for some tests, then you can try fixing this and then make sure that the tests pass afterward. I think all you need to do is, in `cranelift/codegen/src/context.rs`, change `compute_domtree` so after it calls `self.domtree.compute` it also calls `self.domtree_preorder.compute`.

If any of that doesn't work, I'm happy to look at it more carefully.

Once the tests are all passing, I have a couple more comments below. But again, I think this is almost ready!

jameysharp:
I think you are very close to having this working, so I just want to check how you're doing with it. I'd love to merge this PR once it passes the test suite!

fbrv:
happy to take a look if @VamshiReddy02 does not enough bandwith.

VamshiReddy02:
Sure! Sorry for being inactive, caught up with some work. I'll close this PR. @fbrv you can create a new PR.

jameysharp:
Thank you for your effort on this, @VamshiReddy02! I think you got most of the way there and I appreciate the time that you put into it.

@fbrv, welcome! Let me know if you have any questions.

PR #8536, opened 2024-05-03
riscv64: Add minimal support for the Zfa Extension

afonso360:
👋 Hey,

This PR adds initial support for the Zfa extension and implements the `fminm`/`fmaxm` instructions and lowerings.

The Zfa extension provides additional floating point related instructions not included in either F/D extensions. Here's a [link to version 1 of the extension](https://github.com/riscv/riscv-isa-manual/blob/0902ff22aeb2cabbd41f9b06a4af0523fd5ecfb1/src/zfa.adoc) if anyone feels up to some light reading.

The `f{min,max}m` instructions implemented in this PR are pretty much the same as the base `fmin`/`fmax` present in the ISA, but they have different NaN behavior, that conveniently matches the semantics of Cranelift's `fmin`/`fmax` instructions, allowing for a more efficient lowering.

### DIFF
--- a/ci/build-test-matrix.js
+++ b/ci/build-test-matrix.js
@@ -107,7 +107,7 @@ const array = [
     "target": "riscv64gc-unknown-linux-gnu",
     "gcc_package": "gcc-riscv64-linux-gnu",
     "gcc": "riscv64-linux-gnu-gcc",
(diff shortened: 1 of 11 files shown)


alexcrichton (approved):
Nice!

PR #8542, opened 2024-05-03
wasmtime(gc): Add support for `noextern` and `eq` heap types

fitzgen:
With this commit, we now support all GC heap types.

### DIFF
--- a/crates/wasmtime/src/runtime/types.rs
+++ b/crates/wasmtime/src/runtime/types.rs
@@ -431,6 +431,132 @@ impl RefType {
 
 /// The heap types that can Wasm can have references to.
 ///
+/// # Subtyping Hierarchy
+///
+/// Wasm has three different heap type hierarchies:
+///
+/// 1. Function types
+/// 2. External types
+/// 3. Internal types
+///
+/// Each hierarchy has a top type (the common supertype of which everything else
+/// in its hierarchy is a subtype of) and a bottom type (the common subtype of
+/// which everything else in its hierarchy is supertype of).
+///
+/// ## Function Types Hierarchy
+///
+/// The top of the function types hierarchy is `func`; the bottom is
+/// `nofunc`. In between are all the concrete function types.
+///
+/// ```text
+///                          func
+///                       /  /  \  \
+///      ,----------------  /    \  -------------------------.
+///     /                  /      \                           \
+///    |              ,----        -----------.                |
+///    |              |                       |                |
+///    |              |                       |                |
+/// (func)    (func (param i32))    (func (param i32 i32))    ...
+///    |              |                       |                |
+///    |              |                       |                |
+///    |              `---.        ,----------'                |
+///     \                  \      /                           /
+///      `---------------.  \    /  ,------------------------'
+///                       \  \  /  /
+///                         nofunc
+/// ```
+///
+/// Additionally, some concrete function types are sub- or supertypes of other
+/// concrete function types. For simplicity, this isn't depicted in the diagram
+/// above. Specifically, this is based on their parameter and result types, and
+/// whether those types are in a subtyping relationship:
+///
+/// * Parameters are contravariant: `(func (param $a) (result $c))` is a subtype
+///   of `(func (param $b) (result $c))` when `$b` is a subtype of `$a`.
+///
+///   For example, we can substitute `(func (param $cat))` with `(func (param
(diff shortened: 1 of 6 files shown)


crates/wasmtime/src/runtime/types.rs
```diff
@@ -431,6 +431,132 @@ impl RefType {
 
 /// The heap types that can Wasm can have references to.
 ///
+/// # Subtyping Hierarchy
+///
+/// Wasm has three different heap type hierarchies:
+///
+/// 1. Function types
+/// 2. External types
+/// 3. Internal types
```
alexcrichton, line 440: Out of curiosity is the term "internal" used at the spec level as well? It makes sense to mirror "external" but I figured these'd be called "any types" or... something along those lines. Anyway I'm just musing here.

crates/wasmtime/src/runtime/types.rs
```diff
@@ -431,6 +431,132 @@ impl RefType {
...
+///    |              |                       |                |
+///    |              |                       |                |
+/// (func)    (func (param i32))    (func (param i32 i32))    ...
+///    |              |                       |                |
+///    |              |                       |                |
+///    |              `---.        ,----------'                |
+///     \                  \      /                           /
+///      `---------------.  \    /  ,------------------------'
+///                       \  \  /  /
+///                         nofunc
```
alexcrichton, line 466: Ascii Art Diagrams: The Return Of The King

crates/wasmtime/src/runtime/types.rs
```diff
@@ -431,6 +431,132 @@ impl RefType {
...
+///
+/// Additionally, some concrete function types are sub- or supertypes of other
+/// concrete function types. For simplicity, this isn't depicted in the diagram
+/// above. Specifically, this is based on their parameter and result types, and
+/// whether those types are in a subtyping relationship:
+///
+/// * Parameters are contravariant: `(func (param $a) (result $c))` is a subtype
+///   of `(func (param $b) (result $c))` when `$b` is a subtype of `$a`.
+///
+///   For example, we can substitute `(func (param $cat))` with `(func (param
```
alexcrichton, line 477: The word "with" here makes me think that cats are replaced with animals, but I think you're intending the opposite here, so perhaps something like "s/with/for/"?

crates/wasmtime/src/runtime/types.rs
```diff
@@ -431,6 +431,132 @@ impl RefType {
...
+///    |              |                       |                |
+///    |              `---.        ,----------'                |
+///     \                  \      /                           /
+///      `---------------.  \    /  ,------------------------'
+///                       \  \  /  /
+///                         nofunc
+/// ```
+///
+/// Additionally, some concrete function types are sub- or supertypes of other
+/// concrete function types. For simplicity, this isn't depicted in the diagram
```
alexcrichton, line 470: Wait sort of unrelated question on this, in the type section you can explicitly list the supertype and there's also `is_final` which say "no one else can subtype this". With the logic you're explaining below though that's not quite right is it? An `is_final` function type can still be subtyped so long as you align the parameters correctly?

(totally unrelated to this PR I'm just trying to reconcile two different ways I've understood how subtyping is done in wasm)

alexcrichton (approved):

crates/wasmtime/src/runtime/types.rs
fitzgen: Yeah I've seen it around, eg https://github.com/WebAssembly/gc/blob/main/proposals/gc/MVP.md#heap-types

Also turning an `externref` into an `anyref` is called "internalizing"

crates/wasmtime/src/runtime/types.rs
fitzgen: :crown:

crates/wasmtime/src/runtime/types.rs
fitzgen: Yeah honestly I don't fully understand how all the dots connect in the way that the spec is written/defined, I just know what the intended behavior is.

FWIW, we don't supertypes and finality yet, but this code (and I suppose these docs) will need to be updated at that point.

fitzgen: > With the logic you're explaining below though that's not quite right is it? An `is_final` function type can still be subtyped so long as you align the parameters correctly?

And to be clear, yeah you can't subtype something (including a func type) that is declared final.

PR #8543, opened 2024-05-03
Indirect call caching: put options under `-O` namespace.

cfallin:
As per [this comment], the call-indirect caching options should really be in the `-O` option namespace (for optimizations), not `-W` (for semantically-visible Wasm standards options); the original PR simply put them in the wrong place, and this PR moves them.

[this comment]: https://github.com/bytecodealliance/wasmtime/pull/8509#discussion_r1589594152


alexcrichton (approved):
Thanks!

PR #8544, opened 2024-05-03
Fix combining `cache_call_indirects` and memory64

alexcrichton:
This commit fixes a fuzz bug that popped up where the `cache_call_indirects` feature wasn't reading memory64-based offsets correctly. This is due to (not great) API design in `wasmparser` (introduced by me) where wasmparser by default won't read 64-bit offsets unless explicitly allowed to. This is to handle how spec tests assert that overlong 32-bit encodings are invalid as the memory64 proposal isn't merged into the spec yet.

The fix here is to call `allow_memarg64` with whether memory64 is enabled or not and then that'll enable reading these overlong and/or larger offsets correctly.

### DIFF
--- a/crates/environ/src/compile/module_environ.rs
+++ b/crates/environ/src/compile/module_environ.rs
@@ -16,7 +16,6 @@ use std::collections::HashMap;
 use std::mem;
 use std::path::PathBuf;
 use std::sync::Arc;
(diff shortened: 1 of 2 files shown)


cfallin (approved):
Thanks!

PR #8546, opened 2024-05-03
riscv64: Move Round Instructions to ISLE

afonso360:
👋 Hey,

This PR is a small cleanup to our lowerings for rounding instructions (`ceil`/`floor`/`trunc`/`nearest`). It moves the generated code from a single pseudo-instruction into an ISLE lowering. 

The rounding lowering remains mostly the same, except it removes an extraneous `fadd` that we ended up not really needing. It also improves the constant codegen since we have a more advanced algorithm in ISLE that does better pattern matching.

This PR also moves the ISA tests to separate files, but that is done on the first commit, so looking at the second commit individually gives the actual changes to the assembly output.

Also, FloatRound was the only instruction in the `riscv64_worst_case_instruction_size` test but the `worst_case_size` was way larger than what was actually needed for that instruction. I suspect there is some other instruction that may need that much size. I added a few cases, but none ended up at the 124 bytes that it is currently set to. Something to track down later...


alexcrichton (approved):
Very nice refactoring 👍

PR #8548, opened 2024-05-04
`cranelift-frontend`: add and fix various links in doc comments

tertsdiepraam:
While reading through the documentation for `cranelift-frontend`, I noticed that there were some things that could be converted into links. There were also some links that were spelled out, such as `struct.FunctionBuilderContext.html`, which is unnecessary and confuses rust-analyzer.

I've tried to fix up all the easy cases that I could find within this crate. Let me know if you want me to split this up over multiple commits or something like that.

### DIFF
--- a/cranelift/frontend/src/frontend.rs
+++ b/cranelift/frontend/src/frontend.rs
@@ -19,7 +19,7 @@ use cranelift_codegen::packed_option::PackedOption;
 /// Structure used for translating a series of functions into Cranelift IR.
 ///
 /// In order to reduce memory reallocations when compiling multiple functions,
-/// `FunctionBuilderContext` holds various data structures which are cleared between
+/// [`FunctionBuilderContext`] holds various data structures which are cleared between
 /// functions, rather than dropped, preserving the underlying allocations.
 #[derive(Default)]
 pub struct FunctionBuilderContext {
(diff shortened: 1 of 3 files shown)


cranelift/frontend/src/frontend.rs
```diff
@@ -223,7 +223,7 @@ pub enum DefVariableError {
     ///
     /// note: to obtain the type of the value, you can call
     /// [`cranelift_codegen::ir::dfg::DataFlowGraph::value_type`] (using the
-    /// [`FunctionBuilder.func.dfg`] field)
+    /// `FunctionBuilder.func.stencil.dfg` field)
```
bjorn3, line 226: `Function` derefs to `FunctionStencil`, so the extra `.stencil` is not necessary.

tertsdiepraam: Alright, I'll remove it!

abrown:
Looks good to me! Thanks for the changes. Can you amend your commit and add the following somewhere in the doc message so we can trigger the doc build?

```
protest:full
```

That way we can see if there's any dead link warnings or the like.

jameysharp (approved):
I agree with @abrown, this all looks good, and it passed the `prtest:full` CI run so let's merge it. Thank you!

PR #8550, opened 2024-05-04
wasmtime(gc): Add the `wasmtime::NoExtern` value type

fitzgen:

### DIFF
--- a/crates/wasmtime/src/runtime/gc/noextern.rs
+++ b/crates/wasmtime/src/runtime/gc/noextern.rs
@@ -1,0 +1,179 @@
+use crate::{
+    store::{AutoAssertNoGc, StoreOpaque},
+    HeapType, Ref, RefType, Result, Uninhabited, Val, ValRaw, ValType, WasmTy,
+};
+
+/// A reference to the abstract `noextern` heap value.
+///
+/// The are no instances of `(ref noextern)`: it is an uninhabited type.
+///
+/// There is precisely one instance of `(ref null noextern)`, aka `nullexternref`:
+/// the null reference.
+///
+/// This `NoExtern` Rust type's sole purpose is for use with
+/// [`Func::wrap`][crate::Func::wrap]- and
+/// [`Func::typed`][crate::Func::typed]-style APIs for statically typing a
+/// function as taking or returning a `(ref null noextern)` (aka
+/// `Option<NoExtern>`) which is always `None`.
+///
+/// # Example
+///
+/// ```
+/// # use wasmtime::*;
+/// # fn _foo() -> Result<()> {
+/// let mut config = Config::new();
+/// config.wasm_function_references(true);
+/// config.wasm_gc(true);
+/// let engine = Engine::new(&config)?;
+///
+/// let module = Module::new(
+///     &engine,
+///     r#"
+///         (module
+///             (func (export "f") (param (ref null noextern))
+///                 ;; If the reference is null, return.
+///                 local.get 0
+///                 ref.is_null noextern
+///                 br_if 0
+///
+///                 ;; If the reference was not null (which is impossible)
+///                 ;; then raise a trap.
+///                 unreachable
+///             )
+///         )
+///     "#,
+/// )?;
+///
+/// let mut store = Store::new(&engine, ());
+/// let instance = Instance::new(&mut store, &module, &[])?;
+/// let f = instance.get_func(&mut store, "f").unwrap();
+///
+/// // We can cast a `(ref null noextern)`-taking function into a typed function that
+/// // takes an `Option<NoExtern>` via the `Func::typed` method.
+/// let f = f.typed::<Option<NoExtern>, ()>(&store)?;
+///
+/// // We can call the typed function, passing the null `noextern` reference.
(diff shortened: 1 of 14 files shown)


crates/wasmtime/src/runtime/gc/noextern.rs
```diff
@@ -0,0 +1,188 @@
...
+
+    #[inline]
+    fn into_abi(self, _store: &mut AutoAssertNoGc<'_>) -> Result<Self::Abi> {
+        unreachable!("NoExtern is uninhabited")
+    }
+
+    #[inline]
+    unsafe fn from_abi(_abi: Self::Abi, _store: &mut AutoAssertNoGc<'_>) -> Self {
+        unreachable!("NoExtern is uninhabited")
+    }
```
alexcrichton, line 133: Mind matching on self/abi here to avoid the `unreachable!`?

crates/wasmtime/src/runtime/gc/noextern.rs
```diff
@@ -0,0 +1,188 @@
...
+}
+
+unsafe impl WasmTy for Option<NoExtern> {
+    type Abi = *mut NoExtern;
```
alexcrichton, line 137: Is this perhaps best modelled as `usize`? Because technically this is a pointer-sized integer that is a relative offset into the GC heap?

crates/wasmtime/src/runtime/gc/noextern.rs
```diff
@@ -0,0 +1,188 @@
...
+
+    #[inline]
+    fn dynamic_concrete_type_check(
+        &self,
+        _: &StoreOpaque,
+        nullable: bool,
+        ty: &HeapType,
+    ) -> Result<()> {
+        if nullable {
+            // `(ref null nofunc) <: (ref null $f)` for all function types `$f`.
```
alexcrichton, line 157: copy/paste comment?

crates/wasmtime/src/runtime/gc/noextern.rs
```diff
@@ -0,0 +1,188 @@
...
+            Ok(())
+        } else {
+            bail!("argument type mismatch: expected non-nullable (ref {ty}), found null reference")
+        }
+    }
+
+    #[inline]
+    fn is_non_i31_gc_ref(&self) -> bool {
+        false
```
alexcrichton, line 166: This is technically `true`, right? But can be `false` due to no GC interactions here?

(if so mind adding a comment here and/or updating the method name?)

crates/wasmtime/src/runtime/gc/noextern.rs
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    #[inline]
+    unsafe fn abi_from_raw(_raw: *mut ValRaw) -> Self::Abi {
+        ptr::null_mut()
+    }
+
+    #[inline]
+    unsafe fn abi_into_raw(_abi: Self::Abi, raw: *mut ValRaw) {
+        *raw = ValRaw::funcref(ptr::null_mut());
```
alexcrichton, line 176: copy/paste for funcref

crates/wasmtime/src/runtime/gc/noextern.rs
```diff
@@ -0,0 +1,188 @@
...
+
+    #[inline]
+    fn into_abi(self, _store: &mut AutoAssertNoGc<'_>) -> Result<Self::Abi> {
+        Ok(ptr::null_mut())
+    }
+
+    #[inline]
+    unsafe fn from_abi(_abi: Self::Abi, _store: &mut AutoAssertNoGc<'_>) -> Self {
+        None
+    }
```
alexcrichton, line 187: Mind adding debug asserts `self` is none and `abi` is null?

crates/wasmtime/src/runtime/gc/noextern.rs
alexcrichton: Also I'm a bit confused by the implementation here, shouldn't this be `Ok(())` in this case? If the `valtype` matches then we're guaranteed that the expected/given is the exact same, right?

alexcrichton (approved):
Apologies if I'm a bit pedantic below, I'm still getting up-to-speed on this I think

crates/wasmtime/src/runtime/gc/noextern.rs
fitzgen: Yeah this is a copy-paste. This method should be unreachable because there are no concrete heap types in the external type hierarchy that this can be a subtype of.

crates/wasmtime/src/runtime/gc/noextern.rs
fitzgen: Yeah it should be a `u32`, since if inhabited `noextern` would be a `VMGcRef` which is a `NonZeroU32` so if we make it optional/nullable we are left with a `u32`.

crates/wasmtime/src/runtime/gc/noextern.rs
fitzgen: Did some renaming here.

PR #8558, opened 2024-05-06
bindgen: Commit expanded bindgen output for tests

lann:
These outputs are checked in and verified to be fresh by a test so that they can be relied on for code review.

Adds a `stringify` option to `bindgen!` which causes the macro to evaluate to a string literal of the generated code instead of actual tokens.


alexcrichton:
Thanks! I think that `cargo vet` will fail on this so mind adding some vets here as well? PR CI isn't detecting it because it's failing since the diff is too large apparently, otherwise this should be showing up in the CI matrix above

alexcrichton (approved):

PR #8566, opened 2024-05-07
cranelift: Remove `next_fixed_nonallocatable` and `with_allocs` methods

jameysharp:
The `next_fixed_nonallocatable` method doesn't do anything any more and doesn't return anything so calls to it can just be deleted.

The `with_allocs`, `allocate`, and `to_string_with_alloc` methods are all trivial at this point, so inline them. The bulk of this change was performed this way:

`git grep -lF '.with_allocs(' | xargs sed -i 's/\.with_allocs([^)]*)/.clone()/g'`

In a couple cases, this makes the `AllocationConsumer` unused at these methods' call sites. Rather than changing function signatures in this PR, just mark those arguments as deliberately unused.

The number of structures being cloned here is unfortunate, and unnecessary now that we don't need to mutate any of them. But switching to borrowing them is a bigger change than I want to include here.

### DIFF
--- a/cranelift/codegen/src/isa/aarch64/inst/args.rs
+++ b/cranelift/codegen/src/isa/aarch64/inst/args.rs
@@ -145,19 +145,9 @@ impl AMode {
             extendop: op,
         }
     }
-
-    pub(crate) fn with_allocs(&self, _allocs: &mut AllocationConsumer) -> Self {
-        self.clone()
-    }
 }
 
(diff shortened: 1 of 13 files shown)


cfallin (approved):
LGTM, thanks!

Re: cloning, I agree that taking borrows instead is the next step, but (as I'm sure you've thought through but stating for the record here) the original `.with_allocs()` also returned a copy so this shouldn't be a regression at least. I'm very much in favor of the incremental approach here :-)

PR #8568, opened 2024-05-07
wasmtime: Implement Extended Constant Expressions proposal

jameysharp:
This implements https://github.com/webassembly/extended-const for Wasmtime.

Currently the documentation comment for the `wasm_extended_const` configuration option says it's false by default, but `Config::new` sets it to true. One of those should be fixed and I'm not sure which. Other config-related things are still to-do as well, such as adding the appropriate CLI flag.

Also, I could use a little help understanding why one of the funcrefs spec tests fails:

```
Caused by:
    0: failed directive on tests/spec_testsuite/proposals/function-references/elem.wast:539:1
    1: assert_invalid: expected "constant expression required", got "failed to parse WebAssembly module
       
       Caused by:
           type mismatch: expected funcref, found i32 (at offset 0x1b)"
```

cc: @fitzgen


alexcrichton:
For the failure here I think you'll need to set `wasm_extended_const(false)` in `Config` when running the tests. Spec proposals aren't always in-sync with one another and it takes quite some time to merge into the official spec, so tests often assert that a specific construct is invalid when another proposal specifically makes it valid, meaning that we have to be careful about enabled features when testing various proposals.

fitzgen:
Yeah that failure is just because it is using an `i32.add` in a constant expression which used to be disallowed but is now allowed with the extended constants proposal.

We can ignore that particular failure by adding a clause here: https://github.com/bytecodealliance/wasmtime/blob/main/crates/wast/src/wast.rs#L544

Would be nice to ideally send a PR upstream to fix the issue (switch the add to some operator that isn't allowed even with the extended constants proposal) or at least make a local copy for the [`misc_testsuite`](https://github.com/bytecodealliance/wasmtime/tree/main/tests/misc_testsuite) that does that same thing.

fitzgen:
> Currently the documentation comment for the `wasm_extended_const` configuration option says it's false by default, but `Config::new` sets it to true. One of those should be fixed and I'm not sure which.

I think it is fine to have it enabled by default since it is a phase 4 proposal (basically done) and will have more than 2 weeks fuzzing before it hits a release.

alexcrichton:
Oh I'll note that wasm-smith doesn't have support for extended consts right now I think, so AFAIK we won't fuzz this at all

fitzgen:
Ah yes, we should add that support before enabling this proposal by default.

alexcrichton:
This ended up landing as part of https://github.com/bytecodealliance/wasmtime/pull/9141, so I'm going to close this.

PR #8572, opened 2024-05-07
Use WASI builder directly in C API

### LINKED ISSUE #8552: Since Wasmtime 20, preopens of non-existent directories fail eagerly rather than on first access
### Test Case

The simplest way to reproduce this is to create a Python virtual environment on a Windows machine and install `yowasp-yosys` in it. This application uses WASI preopens in the following way to make all of the files on the host available:

```python
            for letter in "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ":
                wasi_cfg.preopen_dir(letter + ":\\", letter + ":")
```

### Steps to Reproduce

* Create a preopen with a non-existent drive letter as a target.

### Expected Results

* As in Wasmtime 19, using the preopen fails when it's actually accessed.

### Actual Results

* In Wasmtime 20, configuring WASI fails.

### Versions and Environment

Wasmtime version or commit: 20.0.0

Operating system: Windows

Architecture: x64

### Additional information

I can work around this easily enough but I'm filing this to make sure the change was intentional. I looked through release notes and PRs that seemed relevant but I'm not sure what has actually caused this change to happen. *Is* the behavior intentional?

alexcrichton:
This commit updates the C API to use the `WasiCtxBuilder` directly within `wasi_config_t` instead of buffering up the options separately. This keeps the behavior of the Rust-based API more similar to the C API and should also help resolve #8552 due to errors being returned more eagerly in the builder-based API.

This additionally makes some minor modifications to the C APIs here as appropriate.

Close #8552

### DIFF
--- a/crates/c-api/src/wasi.rs
+++ b/crates/c-api/src/wasi.rs
@@ -152,7 +152,6 @@ pub unsafe extern "C" fn wasi_config_set_stderr_file(
     config: &mut wasi_config_t,
     path: *const c_char,
 ) -> bool {
-    config.builder.inherit_stderr();
     let file = match create_file(path) {
         Some(f) => f,
         None => return false,
--- a/crates/c-api/src/wasi.rs
+++ b/crates/c-api/src/wasi.rs
@@ -2,10 +2,9 @@
 
 use crate::wasm_byte_vec_t;
 use anyhow::Result;
-use std::ffi::CStr;
+use std::ffi::{c_char, CStr};
 use std::fs::File;
-use std::os::raw::{c_char, c_int};
-use std::path::{Path, PathBuf};
+use std::path::Path;
 use std::slice;
 use wasmtime_wasi::{preview1::WasiP1Ctx, WasiCtxBuilder};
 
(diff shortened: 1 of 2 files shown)


alexcrichton:
Sure yeah! Should be simple enough, the functions just didn't exist when I first wrote wasmtime-py

crates/c-api/src/wasi.rs
```diff
@@ -225,34 +137,35 @@ pub unsafe extern "C" fn wasi_config_set_stdout_file(
...
-    config.stdout = WasiConfigWritePipe::Inherit;
+    config.builder.inherit_stdout();
 }
 
 #[no_mangle]
 pub unsafe extern "C" fn wasi_config_set_stderr_file(
     config: &mut wasi_config_t,
     path: *const c_char,
 ) -> bool {
+    config.builder.inherit_stderr();
```
pchickey, line 155: I don't think this belongs here

pchickey (approved):
Nice simplification. One minor bug

Issue #8432, opened 2024-04-22
Panic about missing trampolines with subtyping in signatures


alexcrichton:
This program:

```rust
use wasmtime::*;

fn main() {
    let engine = Engine::default();

    let module = Module::new(
        &engine,
        r#"
            (module
                (import "" "a" (func $a (result externref)))

                (func (export "a")
                    (drop (call $a))
                )
            )
        "#,
    )
    .unwrap();

    let mut linker = Linker::new(&engine);
    linker
        .func_wrap("", "a", |mut caller: Caller<'_, ()>| {
            ExternRef::new(&mut caller, 100)
        })
        .unwrap();

    let mut store = Store::new(&engine, ());
    let i = linker.instantiate(&mut store, &module).unwrap();
}
```

compiled against the `main` branch currently runs as:

```
$ cargo run -q
thread 'main' panicked at /home/alex/.cargo/git/checkouts/wasmtime-41807828cb3a7a7e/4b9f53a/crates/wasmtime/src/runtime/func.rs:1306:74:
must have a wasm-to-native trampoline for this signature if the Wasm module is importing a function of this signature
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

cc @fitzgen

fitzgen:
As discussed offline, we should look up the trampoline based on the import signature, not the wrapped function's type (which can be a subtype of the import signature).

shamiao:
also happens when `externref` appeared as a func param.

fitzgen:
I started poking at this. Things are a bit hairy because we update `FuncData` based on whether we have found a trampoline for a host function or not yet, but we could update that to hold onto a trampoline for some supertype of the function which is compatible with that particular use but "incompatible" with some other use that uses the function as its precise type rather than the previous super type. I put "incompatible" in scare quotes because it isn't *actually* incompatible because the trampolines aren't doing anything with these values other than shuffling them around to different ABI argument/return registers and stack slots. So any trampoline we find would be valid for all future uses, at least as the system currently exists, but this is a pretty subtle invariant to rely upon almost by accident.

I think the best way to move forward is to center that invariant, and not make it an accidental after thought, by

* keeping track of every function type's "trampoline signature" in `ModuleTypes` or somewhere around there
* a "trampoline signature" is where we map all params/results from `(ref null? <heapty>)` to `(ref null <top(heapty)>)`
[...]

PR #8579, opened 2024-05-07
wasmtime(gc): Fix wasm-to-native trampoline lookup for subtyping

fitzgen:
Previously, we would look up a wasm-to-native trampoline in the Wasm module based on the host function's type. With Wasm GC and subtyping, this becomes problematic because a Wasm module can import a function of type `T` but the host can define a function of type `U` where `U <: T`. And if the Wasm has never defined type `U` then it wouldn't have a trampoline for it. But our trampolines don't actually care, they treat all reference values within the same type hierarchy identically. So the trampoline for `T` would have worked in practice. But once we find a trampoline for a function, we cache it and reuse it every time that function is used in the same store again. Even if the function is imported with its precise type somewhere else. So then we would have a trampoline of the wrong type. But this happened to be okay in practice because the trampolines happen not to inspect their arguments or do anything with them other than forward them between calling convention locations. But relying on that accidental invariant seems fragile and like a gun aimed at the future's feet.

This commit makes that invariant non-accidental, centering it and hopefully making it less fragile by doing so, by making every function type have an associated "trampoline type". A trampoline type is the original function type but where all the reference types in its params and results are replaced with the nullable top versions, e.g. `(ref $my_struct)` is replaced with `(ref null any)`. Often a function type is its own associated trampoline type, as is the case for all functions that don't have take or return any references, for example. Then, all trampoline lookup begins by first getting the trampoline type of the actual function type, or actual import type, and then only afterwards finding for the pre-compiled trampoline in the Wasm module.

Fixes https://github.com/bytecodealliance/wasmtime/issues/8432

### DIFF
--- a/crates/environ/src/compile/module_types.rs
+++ b/crates/environ/src/compile/module_types.rs
@@ -1,9 +1,9 @@
 use crate::{EntityRef, Module, ModuleTypes, TypeConvert};
-use std::{collections::HashMap, ops::Index};
+use std::{borrow::Cow, collections::HashMap, ops::Index};
 use wasmparser::{UnpackedIndex, Validator, ValidatorId};
 use wasmtime_types::{
     EngineOrModuleTypeIndex, ModuleInternedRecGroupIndex, ModuleInternedTypeIndex, TypeIndex,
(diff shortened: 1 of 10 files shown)


alexcrichton (approved):
Nice!

PR #8582, opened 2024-05-08
riscv64: Add remaining Zfa Instructions

afonso360:
👋 Hey,

This PR adds support for the remaining instructions in the Zfa extension (introduced in #8536).

The first one is `fround.{s,d}` that is a direct match of all of our float rounding instructions (`ceil`, `floor`, `trunc`, `nearest`) depending on how the instruction is encoded.

The second one is `fli.{s,d}` that loads a floating point constant from a predefined [table of constants](https://github.com/riscv/riscv-isa-manual/blob/0902ff22aeb2cabbd41f9b06a4af0523fd5ecfb1/src/zfa.adoc#load-immediate-instructions). Additionally I've also added a rule that supports loading the negated constant and then subsequently inverting the sign bit with `fneg`.

There are additional instructions in Zfa, but none that can be easily matched with our IR.

### DIFF
--- a/cranelift/filetests/filetests/isa/riscv64/zfa.clif
+++ b/cranelift/filetests/filetests/isa/riscv64/zfa.clif
@@ -68,3 +68,139 @@ block0(v0: f64, v1: f64):
 ;   .byte 0x53, 0x35, 0xb5, 0x2a
 ;   ret
 
+
+
+function %ceil_f32(f32) -> f32 {
+block0(v0: f32):
+  v1 = ceil v0
+  return v1
+}
+
+; VCode:
+; block0:
+;   fround.s fa0,fa0,rup
+;   ret
+;
+; Disassembled:
+; block0: ; offset 0x0
+;   .byte 0x53, 0x35, 0x45, 0x40
+;   ret
+
+
+function %ceil_f64(f64) -> f64 {
+block0(v0: f64):
+  v1 = ceil v0
+  return v1
+}
+
+; VCode:
+; block0:
+;   fround.d fa0,fa0,rup
+;   ret
+;
+; Disassembled:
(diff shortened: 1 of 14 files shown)


cranelift/filetests/filetests/isa/riscv64/zfa.clif
```diff
@@ -68,3 +68,606 @@ block0(v0: f64, v1: f64):
...
+;   fsd fs4, 0x40(sp)
+;   fsd fs5, 0x38(sp)
+;   fsd fs6, 0x30(sp)
+;   fsd fs7, 0x28(sp)
+;   fsd fs8, 0x20(sp)
+;   fsd fs9, 0x18(sp)
+;   fsd fs10, 0x10(sp)
+;   fsd fs11, 8(sp)
+; block1: ; offset 0x40
+;   .byte 0x53, 0x05, 0x10, 0xf0
```
cfallin, line 363: interesting that Capstone doesn't yet support these `fli`s; does a newer version exist that does?

cfallin (approved):
Thanks!

One thought below about Capstone but otherwise good to go 👍

cranelift/filetests/filetests/isa/riscv64/zfa.clif
afonso360: Capstone in general doesn't support any RISC-V extension, last time I checked the RISC-V support seemed a bit abandoned, but I'll go check again.

afonso360: I had a look at the release notes for the capstone engine for the past year or so, and there aren't many RISC-V specific fixes, and none of them add any extensions 😞

Issue #4637, opened 2022-08-08
Wasmtime horizontal scaling results in poor performance


Duslia:
When I try to use tokio to scale wasmtime horizontally, I found that wasmtime performance drops significantly. It looks like there are some shared resources inside. And no matter how many threads I open, the total number of execute num is almost the same. This is my performance data:
| thread and task | data | total execute nums |
| --- | ----------- | -------- |
| thread 1, task 1      | 11w/task       | 11w
| thread 3, task 3      | 4w/task       | 12w
| thread 8, task 8      | 1w/task       | 8w

This is my test code: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=e9fc13a7d68b80afdc07076482a5b787 .

And the wasm code is just a function return 1. 
```
package main

//export echo
func echo(ctxLen, reqLen int32) int32 { //nolint
	return 1
}

func main() {
}

```

alexcrichton:
Thanks for the report! Could you share the Wasmtime version you're using as well as the specific `*.wasm` file that you're benchmarking?

alexcrichton:
That being said what you're almost certainly running into is limits on concurrently running `mmap` in multiple threads within one process. Allocation of fiber stacks for asynchronously executing WebAsembly primarily goes through `mmap` right now and unmapping a stack requires broadcasting to other threads with an IPI which slows down concurrent execution.

One improvement that you can do to your benchmark is to use the pooling allocator instead of the on-demand allocator. Otherwise though improvements need to come from the kernel itself (assuming you're running Linux, which would also be great to specify in the issue description). Historically the kernel has prototyped a `madvisev` syscall which allows batch unmappings that Wasmtime could use with the pooling allocator. We've tested this historically and it's shown to greatly increase the throughput in situations like this. This isn't landed in Linux, however.

Finally it depends on your use case of whether this comes up much in practice. Typically the actual work performed by a wasm module dwarfs the cost of the memory mappings since the wasm modules live for a longer time than "return 1". If that's the case then this microbenchmark isn't necessarily indicative of performance and it's recommended to use a more realistic workload for your use case. If this sort of short-running computation is common for your embedding, however, then multithreaded execution will currently have the impact that you're measuring here.

Duslia:
Thanks!! I got it.

Duslia:
I'm trying to comment out these two lines and it can scale horizontally with a high performance. 
[...]

PR #8587, opened 2024-05-08
Handle all `decommit`s/`madvise(DONTNEED)`s homogenously

fitzgen:
We previously had different functions for tables vs memories vs stacks vs etc...

Now we use the same system virtual memory interface function for all of them. This enables follow up PRs where we will start batching these calls together (and potentially merging regions together in the further future, although that isn't a clear win as we've seen larger `madvise`s take longer than small ones in the past). Those future batching PRs will allow us to in turn start prototyping new syscalls that can take advantage of that batching, like `madvisev`.


fitzgen:
cc https://github.com/bytecodealliance/wasmtime/issues/4637

alexcrichton (approved):
I like it 👍

PR #8591, opened 2024-05-09
Fuzz the tail call proposal on OSS-Fuzz

alexcrichton:
This was disabled historically before it was fully implemented, but Wasmtime now fully implements it, so enable it for fuzzing.


elliottt (approved):
Awesome!

PR #8593, opened 2024-05-10
Parse object files with `ElfFile64` rather than `object::File`

alexcrichton:
The latter is what Wasmtime uses today but it pulls in parsers for all object formats supported by `object`. In the context of Wasmtime, however, we know that all objects produced are 64-bit ELF files so there's no need to pull in, for example, a COFF parser as that'll always return an error anyway. This commit switches uses of the `object::File` convenience to `ElfFile64` instead.


fitzgen (commented):
I think we should be able to disable the default features of `object` and enable just the `"elf"` feature now.

fitzgen (approved):

alexcrichton:
Ah turns out we're already not-enabling coff/xe/etc, so not much else to do!

PR #8595, opened 2024-05-10
Wasmtime(gc): Add support for supertypes and finality

fitzgen:
This is the final type system change for Wasm GC: the ability to explicitly declare supertypes and finality. A final type may not be a supertype of another type. A concrete heap type matches another concrete heap type if its concrete type is a subtype (potentially transitively) of the other heap type's concrete type.

Next, I'll begin support for allocating GC structs and arrays at runtime.

I've also implemented `O(1)` subtype checking in the types registry:

In a type system with single inheritance, the subtyping relationships between all types form a set of trees. The root of each tree is a type that has no supertype; each node's immediate children are the types that directly subtype that node.

For example, consider these types:

    class Base {}
    class A subtypes Base {}
    class B subtypes Base {}
    class C subtypes A {}
    class D subtypes A {}
    class E subtypes C {}

These types produce the following tree:

               Base
              /    \
             A      B
            / \
           C   D
          /
         E

Note the following properties:

1. If `sub` is a subtype of `sup` (either directly or transitively) then `sup` *must* be on the path from `sub` up to the root of `sub`'s tree.

2. Additionally, `sup` *must* be the `i`th node down from the root in that path, where `i` is the length of the path from `sup` to its tree's root.

Therefore, if we maintain a vector containing the path to the root for each type, then we can simply check if `sup` is at index `supertypes(sup).len()` within `supertypes(sub)`.

### DIFF
--- a/crates/environ/src/compile/module_types.rs
+++ b/crates/environ/src/compile/module_types.rs
@@ -113,7 +113,7 @@ impl ModuleTypesBuilder {
             let ty = &validator_types[id];
             let wasm_ty = WasmparserTypeConverter::new(self, module)
                 .with_rec_group(validator_types, rec_group_id)
-                .convert_sub_type(ty)?;
+                .convert_sub_type(ty);
             self.wasm_sub_type_in_rec_group(id, wasm_ty);
         }
 
@@ -161,6 +161,8 @@ impl ModuleTypesBuilder {
                 // `trampoline_types` so we can reuse it in the future.
                 Cow::Owned(f) => {
                     let idx = self.types.push(WasmSubType {
+                        is_final: true,
+                        supertype: None,
                         composite_type: WasmCompositeType::Func(f.clone()),
                     });
 
@@ -432,6 +434,21 @@ impl TypeConvert for WasmparserTypeConverter<'_> {
                 }
             }
 
+            UnpackedIndex::RecGroup(_) => unreachable!(),
+        }
+    }
+
+    fn lookup_type_index(&self, index: wasmparser::UnpackedIndex) -> EngineOrModuleTypeIndex {
+        match index {
+            UnpackedIndex::Id(id) => {
+                let interned = self.types.wasmparser_to_wasmtime[&id];
+                EngineOrModuleTypeIndex::Module(interned)
+            }
+            UnpackedIndex::Module(module_index) => {
+                let module_index = TypeIndex::from_u32(module_index);
+                let interned = self.module.types[module_index];
+                EngineOrModuleTypeIndex::Module(interned)
+            }
             UnpackedIndex::RecGroup(_) => unreachable!(),
         }
     }
--- a/crates/types/src/lib.rs
+++ b/crates/types/src/lib.rs
@@ -326,8 +326,23 @@ pub enum EngineOrModuleTypeIndex {
 }
 
 impl From<ModuleInternedTypeIndex> for EngineOrModuleTypeIndex {
+    #[inline]
     fn from(i: ModuleInternedTypeIndex) -> Self {
         Self::Module(i)
+    }
+}
+
+impl From<VMSharedTypeIndex> for EngineOrModuleTypeIndex {
+    #[inline]
+    fn from(i: VMSharedTypeIndex) -> Self {
+        Self::Engine(i)
+    }
+}
+
+impl From<RecGroupRelativeTypeIndex> for EngineOrModuleTypeIndex {
+    #[inline]
+    fn from(i: RecGroupRelativeTypeIndex) -> Self {
+        Self::RecGroup(i)
     }
 }
 
@@ -868,8 +883,13 @@ impl TypeTrace for WasmCompositeType {
 /// A concrete, user-defined (or host-defined) Wasm type.
 #[derive(Debug, Clone, Eq, PartialEq, Hash, Serialize, Deserialize)]
 pub struct WasmSubType {
-    // TODO: is_final, supertype
-    //
+    /// Whether this type is forbidden from being the supertype of any other
+    /// type.
+    pub is_final: bool,
+
+    /// This type's supertype, if any.
+    pub supertype: Option<EngineOrModuleTypeIndex>,
+
     /// The array, function, or struct that is defined.
     pub composite_type: WasmCompositeType,
 }
@@ -926,6 +946,9 @@ impl TypeTrace for WasmSubType {
     where
         F: FnMut(EngineOrModuleTypeIndex) -> Result<(), E>,
     {
+        if let Some(sup) = self.supertype {
+            func(sup)?;
+        }
         self.composite_type.trace(func)
     }
 
@@ -933,6 +956,9 @@ impl TypeTrace for WasmSubType {
     where
         F: FnMut(&mut EngineOrModuleTypeIndex) -> Result<(), E>,
     {
+        if let Some(sup) = self.supertype.as_mut() {
+            func(sup)?;
+        }
         self.composite_type.trace_mut(func)
     }
 }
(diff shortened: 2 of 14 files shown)


crates/environ/src/compile/module_types.rs
```diff
@@ -435,4 +437,19 @@ impl TypeConvert for WasmparserTypeConverter<'_> {
             UnpackedIndex::RecGroup(_) => unreachable!(),
         }
     }
+
+    fn lookup_type_index(&self, index: wasmparser::UnpackedIndex) -> EngineOrModuleTypeIndex {
```
alexcrichton, line 441: Could the above method delegate to this new method to start off? (looks like they both start with all the same code)

crates/types/src/lib.rs
```diff
@@ -868,8 +883,12 @@ impl TypeTrace for WasmCompositeType {
 /// A concrete, user-defined (or host-defined) Wasm type.
 #[derive(Debug, Clone, Eq, PartialEq, Hash, Serialize, Deserialize)]
 pub struct WasmSubType {
-    // TODO: is_final, supertype
-    //
+    /// Whether or not this type can be the supertype of any other type.
+    pub is_final: bool,
```
alexcrichton, line 887: Interpreted literally I think these docs are backwards since `true` means it can't be a supertype of anything else. Mind rewording the other way around?

crates/wasmtime/src/runtime/type_registry.rs
```diff
@@ -812,6 +853,11 @@ impl TypeRegistryInner {
                     }
                 }
 
+                // Remove the type's supertypes list, if any.
+                if self.type_to_supertypes.get(ty).is_some_and(|s| s.is_some()) {
+                    self.type_to_supertypes[ty] = None;
+                }
```
alexcrichton, line 859: Could this always do `= None` regardless of whether there's an array there or not?

crates/wasmtime/src/runtime/type_registry.rs
```diff
@@ -907,4 +958,55 @@ impl TypeRegistry {
...
+        //    `sub`'s tree.
+        //
+        // 2. Additionally, `sup` *must* be the `i`th node down from the root in
+        //    that path, where `i` is the length of the path from `sup` to its
+        //    tree's root.
+        //
+        // Therefore, if we have the path to the root for each type (we do) then
+        // we can simply check if `sup` is at index `supertypes(sup).len()`
+        // within `supertypes(sub)`.
+        let inner = self.0.read();
```
alexcrichton, line 1007: IIRC you had a scheme where a lock wasn't required from within wasm, right? (outside of wasm seems totally fine, just wanted to confirm for in-wasm)

crates/wasmtime/src/runtime/types.rs
```diff
@@ -867,7 +875,12 @@ impl HeapType {
             (HeapType::NoFunc, _) => false,
 
             (HeapType::ConcreteFunc(_), HeapType::Func) => true,
-            (HeapType::ConcreteFunc(a), HeapType::ConcreteFunc(b)) => a.matches(b),
+            (HeapType::ConcreteFunc(a), HeapType::ConcreteFunc(b)) => {
+                assert!(a.comes_from_same_engine(b.engine()));
+                a.engine()
+                    .signatures()
+                    .is_subtype(a.type_index(), b.type_index())
+            }
```
alexcrichton, line 883: Aha this clears up some historical confusion on my part! This is the missing link in the "break the recursion" thinking for me.

alexcrichton: Also, for the arms modified here, should the `matches` method be removed? Or is it used elsewhere?

alexcrichton (approved):
Nice 👍

jeffparsons:
> Therefore, if we maintain a vector containing the path to the root for each type, then we can simply check if sup is at index supertypes(sup).len() within supertypes(sub).

This means that the storage for the vecs in each tree is, in total, in `O(N log N)`, right? 

I think there's a cheaper `O(N)` storage per tree way to do it, that still offers `O(1)` subtype checking: 

- Store nodes of the tree in a vec where all descendents of a node occur after that node and before any non-descendent node.
- Each node stores the index of its last descendent. If it has none, then the index will be its own index.
- To test whether `sub` is a subtype of `sup`:
    - Look up trees and indexes of both types.
    - If they live in different trees, then there is no subtype relationship. 
    - Otherwise, look up `sup`'s node. If `sup_index < sub_index && sub_index <= sup_last_descendent` then we have a subtype.

There's a good chance I've misunderstood something (e.g. how the trees are updated?), but if there's nothing else special about these trees then maybe this optimisation is applicable! 🙏

fitzgen:
@jeffparsons, thanks for digging in here! Always appreciate a second pair of eyes.

> This means that the storage for the vecs in each tree is, in total, in `O(N log N)`, right?

It is indeed a time-space tradeoff, but as-implemented it can actually be even worse than `O(n log n)`: consider a chain where each type is a subtype of its predecessor: `a; b <: a; c <: b; d <: c; ...`. The `i`th type's supertype array will be of length `i - 1`, and this forms the classic "triangle" shape that indicates `O(n * n)` behavior. This particular case could be resolved (although the implementation would be a bit awkward in the presence of type canonicalization across modules) by defining the supertype arrays in sub-to-supertype order and then using a subslice of a subtype's supertype's vector, when one is present, instead of allocating a new vector. Then I think we would be at `O(n log n)`.

However, implementations are allowed to define their own limits on the maximum depth a subtyping chain may have. And for JS embeddings, it is additionally tightened from "some implementation limit" to `63`; for maximum cross-engine compatibility, we've always matched the JS embedding limits.

https://webassembly.github.io/gc/js-api/index.html#limits

Anyways, the point is that this limit provides bounds on the total memory overhead associated with these supertype vectors, preventing worst-case scenarios.

Finally, my understanding is that it is pretty much expected that engines implement subtyping checks with this technique, FWIW. For example, here is SpiderMonkey's docs about doing this same thing: https://searchfox.org/mozilla-central/rev/ee2ad260c25310a9fbf96031de05bbc0e94394cc/js/src/wasm/WasmTypeDef.h#480-541

This is a good reminder to double check that we are actually enforcing the subtyping depth limits in validation, though :-p

> I think there's a cheaper O(N) storage per tree way to do it, that still offers `O(1)` subtype checking:

If I understand correctly, this is storing the entry and exit indices of a depth-first traversal of the tree and then subtype checking checks that (a) we are looking at the same tree and (b) that the subtype's index (doesn't matter if we check entry vs exit index) is within the supertype's range, correct?

This would indeed work very nicely in the context of a single Wasm module and a static set of types.

But, as you alluded to, things get problematic when we consider updates. GC objects can be passed between modules, and a non-final type from module A can be subtyped by another type in module B and then an instance of that subtype can be passed from B to A, and A might want to do subtyping checks on this instance even though it never defined the exact type of that instance. This necessitates canonicalizing types across modules. So, even though from a single Wasm module's perspective the type set is static, from an engine perspective we are constantly dynamically adding and removing types from the same deduplicated, canonicalized types registry as modules are loaded or unloaded (or when the host defines/drops types via the public Wasmtime API). And therefore our type hierarchies and their trees aren't static, and must support having subtrees grafted on and pruned away at any moment. That means we would need to recompute potentially all those depth-first traversal indices on every mutation, an `O(n)` operation for each new type definition, which is already too expensive on its own, but also leads to `O(n * n)` behavior when defining `n` new types.

So I don't think this approach is something we can use in this case, unfortunately.

crates/environ/src/compile/module_types.rs
fitzgen: This doesn't quite work out because while both arms of the above method start with producing an `EngineOrModuleIndex`, and that happens in the same way as this method, the arms continue by reusing intermediate steps from how the `EngineOrModuleIndex` was produced in the code that follows after the index has been constructed. This makes it subtly difficult to pull apart or reuse.

crates/wasmtime/src/runtime/type_registry.rs
fitzgen: Unfortunately always doing `self.types_to_supertypes = None` would force the secondary map to allocate even when we aren't ever using Wasm GC. I could remove the `is_some_and(..)` suffix tho; I'll do that.

alexcrichton: Ah ok makes sense, mind leaving a comment that the purpose here is to avoid touching the map entirely?

crates/wasmtime/src/runtime/type_registry.rs
fitzgen: For the common case of checking subtyping of instances of types defined in this module, yes.

I don't have a plan for how to avoid locks in the terrible case of new external subtypes created dynamically:

* create module `A` with non-final type `T`
* instantiate module `A`
* create module `B`
* instantiate module B with type `U <: T`
* instantiate module `B`
* instantiate object of type `U`
* pass object to instance `A` for subtyping check

(Note that module `B` could also be the host using the public API)

My thinking is that we can just fall back to locking the type registry in the cross-module-subtyping case. I think there are some more tricks we can play with `InstancePre`s to make cross-module-subtyping fast for all types that exist at `InstancePre` creation time, but for new subtypes created during runtime we will always need to fall back to locking (or equivalent) AFAICT.

crates/wasmtime/src/runtime/types.rs
fitzgen: They are used elsewhere (e.g. inside the `with_finality_and_supertype` constructors) and are also generally useful as part of the public API.

crates/wasmtime/src/runtime/type_registry.rs
alexcrichton: Ok makes sense, I mostly wanted to confirm there was a means of intra-module checks that didn't require locking.

PR #8596, opened 2024-05-11
riscv64: Improve pattern matching rules for FMA

afonso360:
👋 Hey,

This PR reworks our FMA pattern matching to be slightly less verbose using the suggestions provided by @jameysharp on https://github.com/bytecodealliance/wasmtime/pull/8588#issuecomment-2103773727.

It additionally adds the `(fma x (splat y) z)` pattern for vectors, which can be proven to be equivalent to `(fma y (splat x) z)`. See https://github.com/bytecodealliance/wasmtime/pull/8588#issuecomment-2105622411. I've only added ISA tests for these patterns since it turns out they are already present in the SIMD FMA runtests.

I think all of the patterns matched here are sound although I can't formally prove all of them.  This comment https://github.com/bytecodealliance/wasmtime/pull/8588#issuecomment-2105622411 describes most of the validations that I did.

### DIFF
--- a/cranelift/codegen/src/isa/riscv64/lower.isle
+++ b/cranelift/codegen/src/isa/riscv64/lower.isle
@@ -1509,56 +1509,55 @@
 
 ;;;; Rules for `fma` ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
-;; fmadd: rs1 * rs2 + rs3
-(rule 0 (lower (has_type (ty_scalar_float ty) (fma x y z)))
-  (rv_fmadd ty (FRM.RNE) x y z))
+;; RISC-V has 4 FMA instructions that do a slightly different computation.
+;;
+;; fmadd:   (rs1 * rs2) + rs3
+;; fmsub:   (rs1 * rs2) - rs3
+;; fnmadd: -(rs1 * rs2) - rs3
+;; fnmsub: -(rs1 * rs2) + rs3
+;;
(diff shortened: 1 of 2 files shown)


fitzgen:
Redirecting to @jameysharp since it seems like he has more context here.

cranelift/codegen/src/isa/riscv64/lower.isle
```diff
@@ -1509,56 +1509,55 @@
...
+;; vfnmacc: vd[i] = -(vs1[i] * vs2[i]) - vd[i]
+;; vfnmsac: vd[i] = -(vs1[i] * vs2[i]) + vd[i]
+
+(type IsFneg (enum (Result (negate u64) (value Value))))
+
+(decl pure is_fneg (Value) IsFneg)
+(rule 1 (is_fneg (fneg x)) (IsFneg.Result 1 x))
+(rule 0 (is_fneg x) (IsFneg.Result 0 x))
+
+(rule 1 (lower (has_type ty (fma x_src y_src z_src)))
```
jameysharp, line 1537: Interesting, why does this rule need a priority?

jameysharp (approved):
Fantastic, I love that this worked out, and I really appreciate the effort you put into using Alive2 to check which of my guesses were correct. Too bad about signed zeroes…

By the way, this motivated me to write up #8599, which would make this kind of pattern easier to write someday.

cranelift/codegen/src/isa/riscv64/lower.isle
afonso360: Oops, it shouldn't. I was debugging some stuff and forgot to remove it.

PR #8597, opened 2024-05-11
Remove support for Wasmtime 13-and-prior CLI

alexcrichton:
This commit removes the support in the `wasmtime` CLI for old CLI options which were present in Wasmtime 13 and prior. This compatibility was added in #7385 and backported to the Wasmtime 14 release #7395. Wasmtime 14.0.0, which did not have this compatibility shim, was released on 2023-10-20. Wasmtime 14.0.3, which restored compatibility with this shim, was released on 2023-10-30. This means that Wasmtime since 2023-10-30 has been warning users about differences in the old and new CLI.

This commit will be released with Wasmtime 22 which will means that users will have had an 8-month transition window for warnings to migrate. The hope is that this is sufficient but it's also not too too burdensome to carry for longer if necessary.


jameysharp:
I want to say this is reasonable, but I feel like we should discuss it at the next Wasmtime bi-weekly meeting, at least. I'd also like to find out whether the Go folks have updated yet, for example.

alexcrichton:
Good point! [Added to the agenda](https://github.com/bytecodealliance/meetings/pull/294)

alexcrichton:
@dgryski or @ydnar do either of y'all have a good sense of whether it's still too soon to remove this support or whether the Go integrations with Wasmtime have been migrated?

alexcrichton:
This was discussed at today's Wasmtime meeting and there were no objections.

pchickey (approved):

PR #6480, opened 2023-05-30
Android release binaries

martindevans:
As discussed [here](https://github.com/bytecodealliance/wasmtime/issues/3867#issuecomment-1563121598). This adds Android targets to the build matrix (Android aarch64 and x86_64) and adds Android to the docs in "tier3". This will make wasmtime more convenient to use on Android devices.

I'm not very familiar with Github actions, so it's very possible that this isn't correct. Unfortunately I wasn't able to work out how to run this action in my fork to test it. If anyone can offer guidance on how to do that it would be much appreciated!

PR #8601, opened 2024-05-12
Add Android release binaries to CI

alexcrichton:
This commit is inspired by #6480 and historical asks for Android binaries. This does the bare minimum necessary to configure C compilers such that we can produce binaries but I'll admit that I'm no Android developer myself so I have no idea if these are actually suitable for use anywhere. Otherwise though this build subsumes the preexisting check in CI that the build works for Android, so that part is removed too.

This additionally changes how the NDK is managed from before. Previously a GitHub Action was used to download Java and the NDK and additionally used the `cargo ndk` subcommand. That's all removed now in favor of configuring C compilers directly with a pre-installed version of the NDK which should help reduce the CI dependencies a bit.

### DIFF
--- a/.github/actions/android-ndk/action.yml
+++ b/.github/actions/android-ndk/action.yml
@@ -1,0 +1,31 @@
+# This is a small action which uses the pre-installed Android NDK on GitHub
+# Actions builders, configured with `$ANDROID_NDK`, to compile and link Android
+# code. For Rust we mostly need to configure the linker to Cargo and the C
+# compiler to the `cc` crate, so this sets various environment variables to the
+# appropriate tool within `$ANDROID_NDK`.
+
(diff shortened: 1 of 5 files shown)


jameysharp (approved):
I worry somebody might think we consider Android a tier 1 platform or something if we start having builds in the releases, but we can navigate that confusion if it happens. I have a couple small suggestions but this seems reasonable to me!

alexcrichton:
True! I've also updated the tier listing to explicitly say that this is tier 3, like riscv64.

Part of my own motivation here is that I get the impression folks will glance at the release binaries and determine our platform support without digging much further, so this might help give the impression that Wasmtime isn't ONLY windows/mac/linux, perhaps. There's only so much we can do with that level of review from a prospective user though.

PR #5253, opened 2022-11-11
Add fixed_nonallocatable constraints when appropriate

elliottt:
Plumb the set of allocatable registers through the `OperandCollector` and use it validate uses of fixed-nonallocatable registers, like `%rsp` on x86_64.

PR #8605, opened 2024-05-13
cranelift: Drop non-allocatable register operands

jameysharp:
Consumption of non-allocatable operands was added in #5253 and #5132, and removed in #8524 and following PRs. Now they are not only ignored by regalloc2, but the placeholders that it leaves in the allocation results are also ignored by Cranelift. So let's stop adding them to the operands list entirely.


elliottt (approved):
This was surprising to me! However, if we don't ever generate RA2 operands for VRegs that have already been given allocations and we use those existing allocations during instruction emission, we just don't need the fixed-nonallocatable constraints!

Very cool change :)

PR #8606, opened 2024-05-13
cranelift: Resolve CLIF aliases before lowering

jameysharp:
The egraph pass was already doing this, when it ran, and it never adds any aliases. So do it slightly earlier and unconditionally, and avoid needing to resolve any aliases during lowering.


elliottt (approved):
Makes sense to me! I interpreted the title of the PR as meaning we would remove aliases later in the pipeline, so I was slightly confused by both the removal from the egraphs pass, and the addition to `optimize`. However, the description cleared my questions up :)

PR #8608, opened 2024-05-13
Change `MemoryStyle::Static` to store bytes, not pages

alexcrichton:
This commit is inspired by me looking at some configuration in the pooling allocator and noticing that configuration of wasm pages vs bytes of linear memory is somewhat inconsistent in `Config`. In the end I'd like to remove or update the `memory_pages` configuration in the pooling allocator to being bytes of linear memory instead to be more consistent with `Config` (and additionally anticipate the custom-page-sizes wasm proposal where terms-of-pages will become ambiguous). The first step in this change is to update one of the lowest layered usages of pages, the `MemoryStyle::Static` configuration.

Note that this is not a trivial conversion because the purpose of carrying around pages instead of bytes is that bytes may overflow where overflow-with-pages typically happens during validation. This means that extra care is taken to handle errors related to overflow to ensure that everything is still reported at the same time.

### DIFF
--- a/crates/wasmtime/src/runtime/vm/instance/allocator/pooling/memory_pool.rs
+++ b/crates/wasmtime/src/runtime/vm/instance/allocator/pooling/memory_pool.rs
@@ -276,7 +276,7 @@ impl MemoryPool {
                     {
                         bail!(
                             "memory size allocated per-memory is too small to \
-                             satisfy static bound of {byte_reservation:#x} pages"
+                             satisfy static bound of {byte_reservation:#x} bytes"
                         );
                     }
                 }
--- a/crates/wasmtime/src/runtime/vm/instance/allocator/pooling/memory_pool.rs
+++ b/crates/wasmtime/src/runtime/vm/instance/allocator/pooling/memory_pool.rs
@@ -270,11 +270,13 @@ impl MemoryPool {
             .skip(module.num_imported_memories)
         {
             match plan.style {
-                MemoryStyle::Static { bound } => {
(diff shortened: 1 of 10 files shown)


crates/wasmtime/src/runtime/vm/instance/allocator/pooling/memory_pool.rs
```diff
@@ -270,11 +270,13 @@ impl MemoryPool {
...
-                MemoryStyle::Static { bound } => {
-                    if self.layout.pages_to_next_stripe_slot() < bound {
+                MemoryStyle::Static { byte_reservation } => {
+                    if u64::try_from(self.layout.bytes_to_next_stripe_slot()).unwrap()
+                        < byte_reservation
+                    {
                         bail!(
                             "memory size allocated per-memory is too small to \
-                             satisfy static bound of {bound:#x} pages"
+                             satisfy static bound of {byte_reservation:#x} pages"
```
fitzgen, line 279:
```suggestion
                             satisfy static bound of {byte_reservation:#x} bytes"
```

fitzgen (approved):
Nice! Thanks for doing this, it will def make implementing the custom-page-sizes proposal easier once I get around to that

crates/wasmtime/src/runtime/vm/instance/allocator/pooling/memory_pool.rs
fitzgen: (your lack of rust-analyzer didn't save you this time)

PR #8609, opened 2024-05-13
Change all WASI impls to being concrete, not blanket

alexcrichton:
This commit builds on the support from #8448 to remove all blanket impls from the WASI crates and instead replace them with concrete impls. This is slightly functionally different from before where impls are now on trait objects meaning dynamic dispatch is involved where previously dynamic dispatch was used. That being said the perf hit here is expected to be negligible-to-nonexistent since the implementations are large enough that the dynamic dispatch won't be the hot path.

The motivations for this commit are:

* Removes the need for an odd `skip_mut_forwarding_impls` option - but this'll be left for a bit in case others need it.
* Improves incremental compile time of these crates where the crates themselves now contain all object code for all of WASI instead of forcing the final consume to codegen everything (although there's still a significant amount monomorphized).
* Improves future compatibility with refactorings of bindgen-generated-traits and such because blanket impls are pretty hard to work around where concrete impls are easier to reason about (and document).

### DIFF
--- a/crates/wasi-http/src/http_impl.rs
+++ b/crates/wasi-http/src/http_impl.rs
@@ -15,7 +15,7 @@ use http_body_util::{BodyExt, Empty};
 use hyper::Method;
 use wasmtime::component::Resource;
 
-impl<T: WasiHttpView> outgoing_handler::Host for T {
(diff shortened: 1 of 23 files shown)


pchickey (approved):
Love it!

PR #8611, opened 2024-05-14
wasi: Handle read(0) with file streams

elliottt:
The `read` method on streams should support a size argument of zero, to test if the stream is still open. However, the implementation of `read` for open files was unconditionally treating a read that would return zero bytes as an indication that the stream was closed.

This PR fixes this bug by handling a `read(0)` as an explicit test to see if the current position is at the end of the file.

### DIFF
--- a/crates/test-programs/src/bin/preview2_file_read_write.rs
+++ b/crates/test-programs/src/bin/preview2_file_read_write.rs
@@ -31,6 +31,12 @@ fn main() {
     loop {
         ready.block();
 
+        match stream.read(0) {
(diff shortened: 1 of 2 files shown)


alexcrichton (commented):
I'm a little hesitant to provide such a strong guarantee that a read of size 0 is effectively a `stat` on all files. Would it work to instead pass the read size to `read_result` and don't return `Closed` if the read size is zero?

elliottt:
Fair enough. I've pushed a change to that effect.

alexcrichton (approved):

PR #8612, opened 2024-05-14
CI: shard testing crates across multiple jobs

fitzgen:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -299,12 +299,67 @@ jobs:
       env:
         GH_TOKEN: ${{ github.token }}
 
-  # Checks of various feature combinations and whether things
-  # compile. The goal here isn't to run tests, mostly just serve as a
-  # double-check that Rust code compiles and is likely to work everywhere else.
-  checks:
+  # Checks of various feature combinations and whether things compile. The goal
+  # here isn't to run tests, mostly just serve as a double-check that Rust code
+  # compiles and is likely to work everywhere else.
+  micro_checks:
     needs: determine
-    name: Check
+    name: Check ${{matrix.name}}
+    strategy:
+      fail-fast: true
+      matrix:
+        include:
+          - name: wasmtime
+            checks: |
+              -p wasmtime --no-default-features
+              -p wasmtime --no-default-features --features wat
+              -p wasmtime --no-default-features --features profiling
+              -p wasmtime --no-default-features --features cache
+              -p wasmtime --no-default-features --features async
+              -p wasmtime --no-default-features --features pooling-allocator
+              -p wasmtime --no-default-features --features cranelift
+              -p wasmtime --no-default-features --features component-model
+              -p wasmtime --no-default-features --features runtime,component-model
+              -p wasmtime --no-default-features --features cranelift,wat,async,cache
+              -p wasmtime --no-default-features --features winch
+              -p wasmtime --no-default-features --features wmemcheck
+              -p wasmtime --no-default-features --features wmemcheck,cranelift,runtime
+              -p wasmtime --no-default-features --features demangle
+              -p wasmtime --no-default-features --features addr2line
+              -p wasmtime --no-default-features --features gc
+              -p wasmtime --no-default-features --features runtime,gc
+              -p wasmtime --no-default-features --features cranelift,gc
+              -p wasmtime --no-default-features --features runtime
+              -p wasmtime --no-default-features --features threads
+              -p wasmtime --no-default-features --features runtime,threads
+              -p wasmtime --no-default-features --features cranelift,threads
+              -p wasmtime --features incremental-cache
+              -p wasmtime --all-features
+
+          - name: wasmtime-cli
+            checks: |
+              -p wasmtime-cli --no-default-features
+              -p wasmtime-cli --no-default-features --features run
+              -p wasmtime-cli --no-default-features --features run,component-model
+              -p wasmtime-cli --no-default-features --features compile
+              -p wasmtime-cli --no-default-features --features compile,cranelift
+              -p wasmtime-cli --no-default-features --features compile,cranelift,component-model
+              -p wasmtime-cli --all-features
+              -p wasmtime-cli --features component-model
+
+          - name: cranelift-codegen benches
+            checks: |
+              --benches -p cranelift-codegen
+
+          - name: wasmtime-bench-api
+            checks: |
+              -p wasmtime-bench-api
+
+          - name: wasmtime-c-api
+            checks: |
+              -p wasmtime-c-api --no-default-features
+              -p wasmtime-c-api --no-default-features --features wat
+              -p wasmtime-c-api --no-default-features --features wasi
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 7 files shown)


fitzgen:
Successful default PR run in ~8 minutes now: https://github.com/bytecodealliance/wasmtime/actions/runs/9073195225

Going to do a full run in a second.

fitzgen:
Unfortunately, it looks like `cargo nextest`'s CI partitioning isn't exposed/supported by `cargo miri`.

fitzgen:
Successful full test run in 14 minutes: https://github.com/bytecodealliance/wasmtime/actions/runs/9075621159?pr=8612

alexcrichton:
Thanks for working on this! I'm personally a bit worried about the complexity creep here. Our CI is already pretty nontrivial and adding this into the mix I feel like takes it up to another level of complexity which makes it harder for future contributors/modifications. I can't argue with the results of course, but the merge queue currently reports that the average time-to-merge is 19 minutes so this whould shave 5 minutes off merge time. Given that we generally have an empty queue and we don't have many time-pressured PRs I'm not sure whether the 25% reduction in CI will be worth the complexity increase here.

I'll admit though I'm pretty biased as I'm still sort of used to the 3-hour-cycle-time from rust-lang/rust so my perception of what's appropriate for CI is likely pretty skewed. I'm curious how others feel as well as to the tradeoff of CI complexity here vs the time reduction.

abrown (approved):
I think this is great though I didn't take enough time to understand the sharding in `ci/build-test-matrix.js`; walking through a short example in the doc comment might help because it all looks a bit complex. But overall speeding up CI is great; it looks like this would bring it down from 20+ minutes to ~14 minutes. @alexcrichton, any other comments?

fitzgen:
Yeah I can definitely add more docs/comments if we want to continue down this route.

I think the default PR check time dropping from ~15 minutes to ~8 minutes is the real benefit here; more so than the 20 to 14 for full CI, at least IMO.

.github/workflows/main.yml
```diff
@@ -627,26 +654,45 @@ jobs:
...
 
-    # Test debug (DWARF) related functionality.
+    # NB: the test job here is explicitly lacking in cancellation of this run if
+    # something goes wrong. These take the longest anyway and otherwise if
+    # Windows fails GitHub Actions will confusingly mark the failed Windows job
+    # as cancelled instead of failed.
+
+
+  # Test debug (DWARF) related functionality.
+  test_debug_dwarf:
```
fitzgen, line 668: I think this job is worth splitting out regardless if we want the rest of this PR. This is an extra 4-5 minutes at the end of a full `ci/run-tests.sh` on `main`.

ci/run-tests.sh
```diff
@@ -1,10 +1,38 @@
...
+                )
+                echo "info: disabling cargo feature '$feature_to_remove' and retrying..."
+                FEATURES=$(echo "$FEATURES" | sed -e "s/$feature_to_remove,//g")
+                continue
+            fi
+
+            exit "$code"
+            ;;
+    esac
+done
```
alexcrichton, line 38: I think it might be best to simplify this and replace it with `--all-features`. Most of the above features are on-by-default but it's also not an exhaustive list of off-by-default features that should be tested for crates. For example `wasmtime-environ` should have the `compile` feature enabled when testing, which it is today due to feature unioning but if it's split to its own job then it's probably best to slap `--all-features` and call it a day here.

ci/build-test-matrix.js
```diff
@@ -5,154 +5,241 @@
...
+    }
+    if (item.filter && commits.includes(`prtest:${item.filter}`)) {
+      return true;
+    }
+
+    // If any runtest was modified, re-run the whole test suite as those can
+    // target any backend.
+    if (names.includes(`cranelift/filetests/filetests/runtests`)) {
+      return true;
+    }
```
alexcrichton, line 222: This logic may not be quite so true any more given the more matrix entries we have, could the `item.isa` item be required here to basically only run the isa-based builders on runtest modifications? (e.g. not msrv)

ci/build-test-matrix.js
```diff
@@ -5,154 +5,241 @@
...
+            .join(" "),
+          // Filter the extra features to just those that are within this bucket,
+          // otherwise cargo complains.
+          extra_features: Array.from(
+            Object.keys(config.extra_features)
+              .filter(c => bucket.has(c))
+              .map(c => Array.from(
+                config.extra_features[c].map(f => `--features ${c}/${f}`)
+              ).join(" ")))
+            .join(" "),
```
alexcrichton, line 74: AFAIK the `extra_features` bits here are exclusively for wasi-nn and ONNX stuff so I feel like we're carrying quite a lot of stuff in CI configuration just for that. At that point it might be best to just split out that build/test entirely into its own job? That'd help remove all the special casing around `extra_features` here

ci/build-test-matrix.js
```diff
@@ -5,154 +5,241 @@
...
+  await new Promise((resolve, reject) => {
+    child.on("close", resolve);
+    child.on("error", reject);
+  });
+  const metadata = JSON.parse(data);
+  return metadata.workspace_members.map(m => {
+    const matches = m.match(/#(.*)@\d+\.\d+\.\d+$/)
+          ||  m.match(/([^/]*)#\d+\.\d+\.\d+$/);
+    if (!matches) throw new Error("failed to parse crate name from: " + m);
+    return matches[1];
```
alexcrichton, line 30: Technically-speaking rather than using a regex for this it's probably best to build a map based on the output of `cargo metadata` and lookup things bad on id to project out the package name from the package metadata in the json blob. (probably more robust than regex-matching)

.github/workflows/main.yml
```diff
@@ -299,12 +299,62 @@ jobs:
...
+          "-p wasmtime-bench-api",
+
+          # Check some feature combinations of the `wasmtime-c-api` crate
+          "-p wasmtime-c-api --no-default-features",
+          "-p wasmtime-c-api --no-default-features --features wat",
+          "-p wasmtime-c-api --no-default-features --features wasi",
+        ]
     runs-on: ubuntu-latest
     env:
       CARGO_NDK_VERSION: 2.12.2
```
alexcrichton, line 360: This I think can be removed as it's only used by the one below.

alexcrichton (commented):
Ok I've added some thoughts below but if you could add some more documentation specifically to sharding I think that'd be good.

I'll also point out that this is quite a lot of sharding, more than I've ever seen done before on github actions, so while I have no reason to think things will go wrong there's always the possibility we uncover something strange in the future. Most projects I don't think approach anywhere near this concurrency. The micro_checks job is also very heavily sharded where commands that might take ~5 seconds now take ~1 minute due to the overhead of a runner, downloading rustc, downloading deps, etc. That's not going to break the bank with our concurrency limit though so I don't have any reason to not do this.

fitzgen:
> I'll also point out that this is quite a lot of sharding, more than I've ever seen done before on github actions, so while I have no reason to think things will go wrong there's always the possibility we uncover something strange in the future. Most projects I don't think approach anywhere near this concurrency. The micro_checks job is also very heavily sharded where commands that might take ~5 seconds now take ~1 minute due to the overhead of a runner, downloading rustc, downloading deps, etc. That's not going to break the bank with our concurrency limit though so I don't have any reason to not do this.

Yeah I actually wanted to group that CI job into buckets of ~10 checks each or something but I couldn't figure out an easy way to do that ¯\\\_(ツ)\_/¯

fitzgen:
Okay I think I figured out how to avoid having a billion `check` jobs

.github/workflows/main.yml
```diff
@@ -617,6 +650,67 @@ jobs:
     # Windows fails GitHub Actions will confusingly mark the failed Windows job
     # as cancelled instead of failed.
 
+  # Test `wasmtime-wasi-nn` in its own job, as not all of its backends are
+  # compatible with all targets, and each must be tested separately anyways.
+  test_wasi_nn:
+    strategy:
+      matrix:
+        feature: ["openvino", "onnx"]
+        os: ["ubuntu-latest", "macos-13", "macos-14", "windows-latest"]
```
alexcrichton, line 659: Do you think it's worth decreasing the size of the matrix here? This feels pretty intensive for something that isn't that platform-specific per-se. Maybe only doing linux/windows here or just linux?

alexcrichton (approved):

.github/workflows/main.yml
fitzgen: I was just copying what existed before via tagging along on the main tests, but I'm fine cutting this down to linux and windows.

PR #2980, opened 2021-06-12
Port to OpenBSD!

cfallin:
Verified to work with several simple Wasm modules on an OpenBSD 6.9
x86_64 machine, built with the Rust 1.51.0 included in the packages
collection (`pkg_add rust`) because `rustup` doesn't support
OpenBSD.

This requires a number of PRs to upstream crates to be merged and
released on crates.io first; as a temporary measure, this commit
includes a number of patches in the toplevel `Cargo.toml` to refer
to git branches. Once those patches are merged and released,
I can update this PR (cc @sunfishcode).

The only change to wasmtime code itself is in the trap-handler runtime.
I had to hardcode the offset to RIP in the signal frame, because `libc`
doesn't have a struct definition for OpenBSD. If we think it would be better
to get that upstreamed, I can do that instead. Or, we could go back to a
C helper and just include `<sys/signal.h>`, but I suspect we don't want to do that.

This also disables `wasi-nn` by default, because the `openvino-sys`
crate fails to build on OpenBSD. Ideally we would either fix
upstream or include some more build configuration to exclude the
feature on our end on OpenBSD; unfortunately it seems (according
to rust-lang/cargo#1197) that target-specific features are not
yet supported by Cargo, so this is not very easy to do. I'd be interested
in @abrown's thoughts on this!

Also note that we won't be able to build or test this platform on CI,
unfortunately, for three reasons: (i) `rustup` doesn't carry toolchain builds
for OpenBSD, I think because of the OS's "no ABI compatibility across releases"
policy; (ii) GitHub CI Actions doesn't have any runners
for this platform; and (iii) distributing binaries is difficult, for the above
ABI-compatibility reasons. Nevertheless, it would be nice to be able to say
that it has been known to work at one time, and also maybe at some point
in the distant future someone will see fit to include us in the OpenBSD
ports/packages tree.

Low priority for review (this was just a fun side-adventure). Thanks!

PR #8613, opened 2024-05-14
Added OpenBSD support to wasmtime's signals.rs

CalebHeydon:
This pull request adds support for OpenBSD to the wasmtime-runtime crate. The change is very small. It just adds a case for OpenBSD on x86_64. I'm not trying to port the entire project to OpenBSD. I'm just trying to get this crate to work because I am trying to get the Zed editor to run on OpenBSD.

### DIFF
--- a/crates/wasmtime/src/runtime/vm/sys/unix/signals.rs
+++ b/crates/wasmtime/src/runtime/vm/sys/unix/signals.rs
@@ -255,6 +255,12 @@ unsafe fn get_pc_and_fp(cx: *mut libc::c_void, _signum: libc::c_int) -> (*const 
                 cx.mc_gpregs.gp_elr as *const u8,
                 cx.mc_gpregs.gp_x[29] as usize,
             )
+        } else if #[cfg(all(target_os = "openbsd", target_arch = "x86_64"))] {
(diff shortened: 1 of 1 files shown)


alexcrichton (approved):
Thanks!

fitzgen:
Is there not an official openbsd target in `rustup`? I was looking at adding a CI check to ensure that this will keep compiling in the future, and I didn't see anything. I find that a little surprising.

cfallin:
Unfortunately no, because OpenBSD is explicit about not guaranteeing kernel ABI stability between releases (sort of the opposite of the Linux kernel's never-break-userland rule).

In my limited explorations here, on a native OpenBSD system the package system does have Rust but it's quickly out of date; the only way to have a current Rust compiler is to run OpenBSD -CURRENT (the development branch) with its version of the package tree.

(To @CalebHeydon: at one point I got all of wasmtime building and working on OpenBSD in #2980, but abandoned the effort because keeping everything up-to-date and building given the above vs. MSRV policies was just too much work. Maybe things will change eventually!)

lcheylus:
@CalebHeydon Thanks for your fix to add support for OpenBSD. With it, I was able to build successfully Yara-X https://github.com/VirusTotal/yara-x :)

PR #8614, opened 2024-05-14
Fix the version of mach2 in Cargo.toml

ulan:
The toml file specifies version `0.4.1` instead of `0.4.2`.

Using version `0.4.1` produces a compile error:

```
error[E0432]: unresolved import `mach2::ndr`
  --> external/crate_index__wasmtime-runtime-20.0.2/src/sys/unix/machports.rs:44:12
   |
44 | use mach2::ndr::*;
   |            ^^^ could not find `ndr` in `mach2`
```

That's because `ndr` was added in version `0.4.2`.

Note that the lock file specifies version `0.4.2` which explains why this error doesn't happen normally.


cfallin (approved):
Thanks!

cfallin:
@ulan if you want to cherrypick this onto `release-21.0.0` in a separate PR too, I'm happy to approve that as well. The release branched last week and is going out next week so we'll want this fix there too, probably.

ulan:
> @ulan if you want to cherrypick this onto release-21.0.0 in a separate PR too, I'm happy to approve that as well. The release branched last week and is going out next week so we'll want this fix there too, probably.

Thanks for the offer! We solved this on our end, so we don't need the cherrypick. I am happy to prepare the cherrypick PRs if you think it would be worthwhile for other users (which seems unlikely considering that no one reported the issue).

cfallin:
For completeness we should probably do so; I'm happy to do it quickly otherwise.

ulan:
> For completeness we should probably do so; I'm happy to do it quickly otherwise.

Sounds good! I'll cherry pick to 21.

> The release branched last week and is going out next week so we'll want this fix there too, probably.

Do you mean that would be `release-22.0.0` that would also need this fix?

cfallin:
Oh, no, I meant that `release-21.0.0` branched last week, then waits for two weeks (to give time to find and cherrypick issues just like this :-) ) and releases next week. Thanks!

ulan:
The cherrypick: https://github.com/bytecodealliance/wasmtime/pull/8615

PR #8615, opened 2024-05-14
Fix the version of mach2 in Cargo.toml

ulan:
The toml file specifies version `0.4.1` instead of `0.4.2`.

Using version `0.4.1` produces a compile error:

```
error[E0432]: unresolved import `mach2::ndr`
  --> external/crate_index__wasmtime-runtime-20.0.2/src/sys/unix/machports.rs:44:12
   |
44 | use mach2::ndr::*;
   |            ^^^ could not find `ndr` in `mach2`
```

That's because `ndr` was added in version `0.4.2`.

Note that the lock file specifies version `0.4.2` which explains why this error doesn't happen normally.


ulan:
@cfallin: here is the cherrypick onto 21 ^ (I cannot edit the reviewer and the system autoassigned @abrown).

cfallin (approved):
Thanks very much!

PR #8616, opened 2024-05-14
Change `Tunables::static_memory_bound` to bytes

alexcrichton:
This commit changes the wasm-page-sized `static_memory_bound` field to instead being a byte-defined unit rather than a page-defined unit. To accomplish this the field is renamed to `static_memory_reservation` and all references are updated. This builds on the support from #8608 to remove another page-based variable from the internals of Wasmtime.

### DIFF
--- a/crates/environ/src/module.rs
+++ b/crates/environ/src/module.rs
@@ -28,20 +28,16 @@ pub enum MemoryStyle {
 impl MemoryStyle {
     /// Decide on an implementation style for the given `Memory`.
     pub fn for_memory(memory: Memory, tunables: &Tunables) -> (Self, u64) {
-        let static_memory_bound = tunables
-            .static_memory_bound
-            .checked_mul(u64::from(WASM_PAGE_SIZE))
-            .unwrap();
-
         let is_static = match memory.maximum_byte_size() {
             Ok(mut maximum) => {
                 if tunables.static_memory_bound_is_maximum {
-                    maximum = maximum.min(static_memory_bound);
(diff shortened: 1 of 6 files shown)


abrown (approved):
Looks good to me! I think consistently using bytes is just better; when I was mucking around in the pooling allocator it was always tricky to understand whether variables were page-based or byte-based. I remember having to handle a lot of overflow conditions which I found by `proptest`-ing... I assume that was handled somewhere else?

alexcrichton:
Yeah I think most of those were handled in https://github.com/bytecodealliance/wasmtime/pull/8608 although there's still `memory_pages` in the pooling allocator which I'd like to remove as well and is probably also a large source of overflow checks.

fitzgen:
> I think consistently using bytes is just better

Agreed, especially since it is effectively necessary to support the custom-page-sizes proposal.

> when I was mucking around in the pooling allocator it was always tricky to understand whether variables were page-based or byte-based

I think we should probably start newtyping our units. Way back when I wrote some stuff like this for `wee_alloc` and then someone pulled it out into a crate: https://docs.rs/memory_units

Don't think we want to use exactly that, but more that kind of thing.

abrown:
> Don't think we want to use exactly that, but more that kind of thing.

I agree. That's inside the code base, but even from a CLI perspective I found it useful to use `bytesize` in the `mpk.rs` example to more easily just pass in strings like "4KiB" or whatever instead of having to do mental math.

PR #8622, opened 2024-05-15
wasmtime: Introduce the `test-macros` crate

saulecabrera:
This commit introduces the `test-macros` crate. The whole idea behind this crate is to export a single or multiple macros to make it easier to configure Wasmtime for integration tests. The main use-case at this time, is running a subset of the integration tests with Cranelift and Winch. This crate could be extended to serve other use-cases, like testing pooling allocator and/or GC configurations.  

This commit introduces a single example of how this macro could be used. If there's agreement on merging this change in some shape or form, I'll follow up with migrating the current tests to use `#[wasmtime_test]` where applicable. 

Part of what's implemented in this PR was discussed in Cranelift's meeting on [April 24th, 2024](https://github.com/bytecodealliance/meetings/blob/main/cranelift/2024/cranelift-04-24.md), however there are several discussion points that are still "in the air", like for example, what's the best way to avoid the combinatorial explosion problem for the potential test matrix.

-- 

cc/ @alexcrichton @fitzgen I'm opening this a draft, but feel free to leave any feedback according the discussion in the Cranelift meeting. I'm a bit concerned because I don't really have a good answer to the combinatorial problem of the test matrix, at least initially I suspect this won't be an issue necessarily, but if this macro grows in functionality I think it might be a problem.


fitzgen:
Excited for this! Will take a look today

crates/test-macros/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+
+            if strategies.len() > 2 {
+                return Err(meta.error("Expected at most 2 strategies"));
+            }
+
+            if strategies.is_empty() {
+                return Err(meta.error("Expected at least 1 strategy"));
+            }
+
+            Ok(())
```
alexcrichton, line 42: It might be good to replace this with [`parse_nested_meta`](https://docs.rs/syn/latest/syn/meta/struct.ParseNestedMeta.html#method.parse_nested_meta) perhaps?

crates/test-macros/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+        } else {
+            Err(meta.error("Unsupported attributes"))
+        }
+    });
+
+    parse_macro_input!(attrs with parser);
+
+    match expand(&strategies, parse_macro_input!(item as ItemFn)) {
```
alexcrichton, line 50: One possible concern about `ItemFn` here is that in the limit the entire test crate will be tagged with `#[wasmtime_test]` and that means the entire crate is parsed through `syn` and emitted back out. Technically the only needed part here is the function signature. 

I'm not sure if it's worth it, but this could try to be a bit clever where it parses `syn::ItemFn` manually up to the `block` and then keeps the `block` as an opaque `TokenStream` (since it's known that it's brace-delimited anyway).

crates/test-macros/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+        Ok(tok) => tok,
+        Err(e) => e.into_compile_error().into(),
+    }
+}
+
+fn expand(strategies: &[(String, Ident)], func: ItemFn) -> Result<TokenStream> {
```
alexcrichton, line 56: I might recommend replacing `strategies` with something like `config: TestConfig` which internally has `Vec<(String,Ident)>` or similar to make this more easily growable in the future to multiple configuration options.

tests/all/func.rs
```diff
@@ -3,10 +3,11 @@ use std::sync::atomic::Ordering;
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering::SeqCst};
 use std::sync::Arc;
 use wasmtime::*;
+use wasmtime_test_macros::wasmtime_test;
 
-#[test]
+#[wasmtime_test(strategies(Cranelift, Winch))]
 #[cfg_attr(miri, ignore)]
-fn call_wasm_to_wasm() -> Result<()> {
+fn call_wasm_to_wasm(config: &mut Config) -> Result<()> {
```
fitzgen, line 10: Yep, this is exactly what I was imagining usage would look like :+1:

fitzgen (commented):
Definitely going in a good direction!

alexcrichton (approved):
Looks great to me, thanks!

I'm going to go ahead and merge, but some other assorted thoughts (which are fine as follow-ups or to ignore entirely):

* Should the default perhaps be the "full matrix" of tests? For example pooling, default settings (cranelift), and winch might be the three main configurations to execute in right now.
* Instead of opting-in to strategy should the attributes be "skip this test on winch" or similar? (e.g. eagerly running all tests instead of having to opt-in to testing)

In any case it might make more sense to expand the current strategy implemented and see how it looks after?

saulecabrera:
Those are good suggestions Alex, I like the idea of opting out instead of opting in, it has the potential to make the macro less verbose and easier to integrate. I'll start migrating the tests and see how things shape up and update the macro definition if needed.

PR #8625, opened 2024-05-15
egraphs: Const propagate bitwise float operations

afonso360:
👋 Hey,

This PR adds constant propagation for the few bitwise floating point operations that we have (`fneg` / `fabs` / `fcopysign`). Since these operations are guaranteed to be done in a bitwise manner it should be safe to constant propagate them.


alexcrichton (approved):

cranelift/codegen/src/opts/cprop.isle
```diff
@@ -279,3 +279,32 @@
...
+(decl pure f32_abs (Ieee32) Ieee32)
+(extern constructor f32_abs f32_abs)
+(decl pure f32_copysign (Ieee32 Ieee32) Ieee32)
+(extern constructor f32_copysign f32_copysign)
+(decl pure f64_neg (Ieee64) Ieee64)
+(extern constructor f64_neg f64_neg)
+(decl pure f64_abs (Ieee64) Ieee64)
+(extern constructor f64_abs f64_abs)
+(decl pure f64_copysign (Ieee64 Ieee64) Ieee64)
+(extern constructor f64_copysign f64_copysign)
```
fitzgen, line 310: Can we move these to `prelude.isle`?

fitzgen (approved):
Thanks! One nitpick below :)

PR #8627, opened 2024-05-15
Use WASM function names in compiled objects

lpereira:
Instead of generating symbol names in the format
"wasm[$MODULE_ID]::function[$FUNCTION_INDEX]", generate (if possible) something more readable, such as "wasm[$MODULE_ID]::$FUNCTION_NAME". This helps when debugging or profiling the generated code.

### DIFF
--- a/crates/wasmtime/src/compile.rs
+++ b/crates/wasmtime/src/compile.rs
@@ -428,7 +428,7 @@ impl<'a> CompileInputs<'a> {
         // like "perf" or "objdump", and filter the name.  Let only characters usually
         // used for function names, plus some characters that might be used in name
         // mangling.
-        let bad_char = |c: char| !c.is_alphanumeric() && !r"<>[]_-:@$".contains(c);
+        let bad_char = |c: char| !c.is_ascii_alphanumeric() && !r"<>[]_-:@$".contains(c);
         if name.chars().any(bad_char) {
             let mut last_char_seen = '\u{0000}';
             Cow::Owned(
--- a/crates/wasmtime/src/compile.rs
+++ b/crates/wasmtime/src/compile.rs
@@ -347,9 +347,6 @@ struct CompileInputs<'a> {
 }
 
 impl<'a> CompileInputs<'a> {
-    /// Maximum length of symbols generated in objects.
-    const MAX_SYMBOL_LEN: usize = 96;
-
     fn push_input(&mut self, f: impl FnOnce(&dyn Compiler) -> Result<CompileOutput> + Send + 'a) {
         self.inputs.push(Box::new(f));
     }
@@ -424,22 +421,29 @@ impl<'a> CompileInputs<'a> {
     }
 
     fn clean_symbol(name: &str) -> Cow<str> {
+        /// Maximum length of symbols generated in objects.
+        const MAX_SYMBOL_LEN: usize = 96;
+
         // Just to be on the safe side, avoid passing user-provided data to tools
         // like "perf" or "objdump", and filter the name.  Let only characters usually
         // used for function names, plus some characters that might be used in name
         // mangling.
         let bad_char = |c: char| !c.is_alphanumeric() && !r"<>[]_-:@$".contains(c);
         if name.chars().any(bad_char) {
-            let mut name = name
-                .chars()
-                .map(|c| if bad_char(c) { '?' } else { c })
-                .collect::<String>()
-                .trim_matches('?')
-                .to_string();
-            name.truncate(Self::MAX_SYMBOL_LEN);
-            Cow::Owned(name)
+            let mut last_char_seen = '\u{0000}';
+            Cow::Owned(
+                name.chars()
+                    .map(|c| if bad_char(c) { '?' } else { c })
+                    .filter(|c| {
+                        let skip = last_char_seen == '?' && *c == '?';
+                        last_char_seen = *c;
+                        !skip
+                    })
+                    .take(MAX_SYMBOL_LEN)
+                    .collect::<String>(),
+            )
         } else {
-            Cow::Borrowed(&name[..])
+            Cow::Borrowed(&name[..MAX_SYMBOL_LEN])
         }
     }
 
--- a/crates/wasmtime/src/compile.rs
+++ b/crates/wasmtime/src/compile.rs
@@ -471,12 +471,16 @@ impl<'a> CompileInputs<'a> {
                         .get(&func_index)
                     {
                         Some(name) => format!(
-                            "wasm[{}]::func[{}]::{}",
+                            "wasm[{}]::function[{}]::{}",
                             module.as_u32(),
                             func_index.as_u32(),
                             Self::clean_symbol(&name)
                         ),
-                        None => format!("wasm[{}]::func[{}]", module.as_u32(), func_index.as_u32()),
+                        None => format!(
+                            "wasm[{}]::function[{}]",
+                            module.as_u32(),
+                            func_index.as_u32()
+                        ),
                     };
 
                     Ok(CompileOutput {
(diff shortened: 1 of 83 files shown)


bjorn3:
Is it guaranteed that the names section only contains unique names?

alexcrichton:
Thanks! This makes sense to me and I've often thought this would be nice as well. A hesitation I've had in the past though is that I'm a bit wary about exposing user-provided data to native tools like `perf` and `objdump` and such. The name section has completely arbitrary strings and I while I can't say for certainty that native tooling behaves well in the face of arbitrary strings I suspect that there might be a security issue waiting to happen along the lines of "if you run `ldd` over a binary it can run arbitrary code" or something like that.

(a good example being @bjorn3's suggestion where the `name` section has no guarantees of uniqueness and I'm not sure what tooling would do with same-named-symbols)

cfallin:
I wonder if we could define suitable "cleanup" for names: (i) rewrite duplicates with numeric suffixes, (ii) replace or remove any characters not in the usual set (`[A-Za-z0-9-_]` -- maybe spaces too? can tools handle those?), (iii) cap the length at something reasonable like 32 or 64 characters?

lpereira:
Could also add the function_index for the cases where uniqueness isn't guaranteed.  Alright, I'll cook up a patch with the suggestions.

alexcrichton:
Yeah I think that'd work well (albeit bit a bit more involved in the implementation). While I can't say for certainty that spaces are not supported my guess is that tools would likely eventually choke on non-C-looking symbols so I'd be tempted to leave them out.

fitzgen:
How about

* `wasm[N]::func[M]` if no name section entry
* `wasm[N]::func[m]: <name>` where `<name>` is sanitized and has a capped length if there is an entry

?

lpereira:
Alright, pushed some changes that:

- Will filter symbols to only include alphanumeric characters and others that might be generated by name mangling and whatnot, and replace runs of them with a single '?'
- Always include the function index in the symbol name
- Truncate the symbol name to 96 characters

I think this should be sufficient, although I'm not really sold on how good the string filtering code is. I wish there was a `String::translate()` that would work similarly to `String::retain()`, but replace the character with what the callback function returns.  I'm a newbie in Rust, though, so maybe there's a way I couldn't figure out?

crates/wasmtime/src/compile.rs
```diff
@@ -419,6 +422,28 @@ impl<'a> CompileInputs<'a> {
...
+        let mut cleaned = String::with_capacity(name.len());
+        for c in name.chars().map(|c| {
+            if c.is_alphanumeric() || r"<>[]_-:@$".contains(c) {
+                c
+            } else {
+                '?'
+            }
+        }) {
+            cleaned.push(c);
+        }
```
cfallin, line 437: I think you can do something better with iterators, and also we'll want to avoid allocating a string in the common case -- perhaps something like:

- Define a predicate `fn bad_char(c: char) -> bool { ... }` inside this function
- Guard on no bad chars, to allow the common case without allocations; if any bad chars, do an iterator stream filter

```
let name = if name.chars().any(|c| bad_char(c)) {
  Cow::Owned(name.chars().map(|c| if bad_char(c) { '?' } else { c }).collect::<String>())
} else {
  Cow::Borrowed(&name[..])
}
```

Thoughts?

lpereira: Yup, right after I pushed this I found about the `map()` method.  Let me use `Cow` too, as this should reduce quite a bit of writes in the happy path.

lpereira: Done, thanks!

crates/wasmtime/src/compile.rs
```diff
@@ -419,6 +423,27 @@ impl<'a> CompileInputs<'a> {
...
+        let bad_char = |c: char| !c.is_alphanumeric() && !r"<>[]_-:@$".contains(c);
+        if name.chars().any(bad_char) {
+            let mut name = name
+                .chars()
+                .map(|c| if bad_char(c) { '?' } else { c })
+                .collect::<String>()
+                .trim_matches('?')
+                .trim_end_matches('?')
+                .to_string();
+            name.truncate(Self::MAX_SYMBOL_LEN);
```
cfallin, line 440: .take(N) in the iterator chain should allow the truncation to happen inline (and stop iteration early if so, since the whole chain is lazy)

cfallin (approved):
Looks reasonable to me, thanks for the iterations!

crates/wasmtime/src/compile.rs
lpereira: Actually, no, I'm not sure how to do this; I want to truncate the string after `trim_matches('?')` removed repeated occurrences of `'?'` in the cleaned string, not before.

cfallin:
Ah, force-pushed after my review; could you do subsequent updates as additional commits? Let me know when stable and I can re-review.

lpereira:
It's stable now.  (I'm not using the `take()` method for the reason I stated above, so I reverted to the previous version.)

crates/wasmtime/src/compile.rs
```diff
@@ -419,6 +423,26 @@ impl<'a> CompileInputs<'a> {
         ret
     }
 
+    fn clean_symbol(name: &str) -> Cow<str> {
+        // Just to be on the safe side, avoid passing user-provided data to tools
+        // like "perf" or "objdump", and filter the name.  Let only characters usually
+        // used for function names, plus some characters that might be used in name
+        // mangling.
+        let bad_char = |c: char| !c.is_alphanumeric() && !r"<>[]_-:@$".contains(c);
```
jameysharp, line 431: Maybe you wanted [is_ascii_alphanumeric](https://doc.rust-lang.org/std/primitive.char.html#method.is_ascii_alphanumeric) here instead, rather than allowing all Unicode alpha/numeric codepoints?

```suggestion
        let bad_char = |c: char| !c.is_ascii_alphanumeric() && !r"<>[]_-:@$".contains(c);
```

jameysharp (commented):
This is going to be such a big usability improvement for people using tools like perf! I'm excited.

jameysharp:
Ah, the CI test failures are because `tests/disas.rs` is specifically looking for symbol names which contain both `wasm` and `function` in their names, and you've shortened the latter to `func`. This is near the beginning of `disas_elf`.

I don't like that this test is relying on this implementation detail of Wasmtime, but I suppose it's fine. So you can either change this PR to continue to use `function`, or change `disas.rs` to search for `func`.

Either way, you'll also need to run `WASMTIME_TEST_BLESS=1 cargo test --test disas` to update the tests to reflect this change. Some of them do have function names, and those now show up in the expected test output, which is a fantastic bonus.

crates/wasmtime/src/compile.rs
lpereira: Done!

crates/wasmtime/src/compile.rs
```diff
@@ -419,6 +423,26 @@ impl<'a> CompileInputs<'a> {
...
+            let mut name = name
+                .chars()
+                .map(|c| if bad_char(c) { '?' } else { c })
+                .collect::<String>()
+                .trim_matches('?')
+                .to_string();
+            name.truncate(Self::MAX_SYMBOL_LEN);
+            Cow::Owned(name)
+        } else {
+            Cow::Borrowed(&name[..])
```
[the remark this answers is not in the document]
lpereira: Yup, had this change locally and forgot to push. Good catch!

jameysharp:
> (@jameysharp, that filter wouldn't work -- words like "keep", which are fine, would be transformed to "kep" -- but I tweaked it to consider that case.)

Oh, you're absolutely right. :sweat_smile:

Looks good to me, let's ship it!

PR #8628, opened 2024-05-15
Use bytes for maximum size of linear memory with pooling

alexcrichton:
This commit changes configuration of the pooling allocator to use a byte-based unit rather than a page based unit. The previous `PoolingAllocatorConfig::memory_pages` configuration option configures the maximum size that a linear memory may grow to at runtime. This is an important factor in calculation of stripes for MPK and is also a coarse-grained knob apart from `StoreLimiter` to limit memory consumption. This configuration option has been renamed to `max_memory_size` and documented that it's in terms of bytes rather than pages as before.

Additionally the documented constraint of `max_memory_size` must be smaller than `static_memory_bound` is now additionally enforced as a minor clean-up as part of this PR as well.

### DIFF
--- a/crates/cli-flags/src/lib.rs
+++ b/crates/cli-flags/src/lib.rs
@@ -112,6 +112,10 @@ wasmtime_option_group! {
         /// the pooling allocator.
         pub pooling_total_stacks: Option<u32>,
 
+        /// The maximum runtime size of linear memories in the pooling
+        /// allocator.
+        pub pooling_max_memory_size: Option<usize>,
+
         /// Whether to enable call-indirect caching.
         pub cache_call_indirects: Option<bool>,
 
@@ -604,6 +608,9 @@ impl CommonOptions {
                     }
                     if let Some(limit) = self.opts.pooling_total_stacks {
                         cfg.total_stacks(limit);
+                    }
+                    if let Some(limit) = self.opts.pooling_max_memory_size {
+                        cfg.max_memory_size(limit);
                     }
                     if let Some(enable) = self.opts.memory_protection_keys {
                         if enable {
--- a/crates/cli-flags/src/lib.rs
+++ b/crates/cli-flags/src/lib.rs
@@ -112,8 +112,8 @@ wasmtime_option_group! {
         /// the pooling allocator.
         pub pooling_total_stacks: Option<u32>,
 
-        /// The maximum runtime size of linear memories in the pooling
-        /// allocator.
+        /// The maximum runtime size of each linear memory in the pooling
+        /// allocator, in bytes.
         pub pooling_max_memory_size: Option<usize>,
 
         /// Whether to enable call-indirect caching.
--- a/crates/fuzzing/src/generators/config.rs
+++ b/crates/fuzzing/src/generators/config.rs
@@ -76,7 +76,7 @@ impl Config {
         if let InstanceAllocationStrategy::Pooling(pooling) = &mut self.wasmtime.strategy {
             // One single-page memory
             pooling.total_memories = config.max_memories as u32;
-            pooling.max_memory_size = 10 * 65535;
+            pooling.max_memory_size = 10 << 16;
             pooling.max_memories_per_module = config.max_memories as u32;
 
             pooling.total_tables = config.max_tables as u32;
@@ -135,7 +135,7 @@ impl Config {
             if pooling.total_memories < 1
                 || pooling.total_tables < 5
                 || pooling.table_elements < 1_000
-                || pooling.max_memory_size < (900 * 65536)
+                || pooling.max_memory_size < (900 << 16)
                 || pooling.total_core_instances < 500
                 || pooling.core_instance_size < 64 * 1024
             {
@@ -415,17 +415,17 @@ impl<'a> Arbitrary<'a> for Config {
             // Ensure the pooling allocator can support the maximal size of
             // memory, picking the smaller of the two to win.
             let min_pages = cfg.max_memory32_pages.min(cfg.max_memory64_pages);
-            let min = (min_pages * 65536).min(pooling.max_memory_size as u64);
+            let min = (min_pages << 16).min(pooling.max_memory_size as u64);
             pooling.max_memory_size = min as usize;
-            cfg.max_memory32_pages = min / 65536;
-            cfg.max_memory64_pages = min / 65535;
+            cfg.max_memory32_pages = min >> 16;
+            cfg.max_memory64_pages = min >> 16;
 
             // If traps are disallowed then memories must have at least one page
             // of memory so if we still are only allowing 0 pages of memory then
             // increase that to one here.
             if cfg.disallow_traps {
                 if pooling.max_memory_size == 0 {
-                    pooling.max_memory_size = 65535;
+                    pooling.max_memory_size = 1 << 16;
                     cfg.max_memory32_pages = 1;
                     cfg.max_memory64_pages = 1;
                 }
--- a/crates/fuzzing/src/generators/config.rs
+++ b/crates/fuzzing/src/generators/config.rs
@@ -76,7 +76,7 @@ impl Config {
         if let InstanceAllocationStrategy::Pooling(pooling) = &mut self.wasmtime.strategy {
             // One single-page memory
             pooling.total_memories = config.max_memories as u32;
-            pooling.memory_pages = 10;
+            pooling.max_memory_size = 10 * 65535;
             pooling.max_memories_per_module = config.max_memories as u32;
 
             pooling.total_tables = config.max_tables as u32;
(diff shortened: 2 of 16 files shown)


crates/cli-flags/src/lib.rs
```diff
@@ -112,6 +112,10 @@ wasmtime_option_group! {
         /// the pooling allocator.
         pub pooling_total_stacks: Option<u32>,
 
+        /// The maximum runtime size of linear memories in the pooling
+        /// allocator.
+        pub pooling_max_memory_size: Option<usize>,
```
fitzgen, line 117:
```suggestion
        /// The maximum runtime size of each linear memory in the pooling
        /// allocator, in bytes.
        pub pooling_max_memory_size: Option<usize>,
```

crates/fuzzing/src/generators/config.rs
```diff
@@ -76,7 +76,7 @@ impl Config {
         if let InstanceAllocationStrategy::Pooling(pooling) = &mut self.wasmtime.strategy {
             // One single-page memory
             pooling.total_memories = config.max_memories as u32;
-            pooling.memory_pages = 10;
+            pooling.max_memory_size = 10 * 65535;
```
fitzgen, line 79: This should be `65536` right? If you instead want the max address it should be `10 * 65536 - 1`.

Also lets use named constants for the Wasm page size so that it is easier to find uses and update in the future for the custom-page-sizes proposal. We have it defined in `wasmtime_types` already, FWIW, and this could use that here: https://github.com/bytecodealliance/wasmtime/blob/c477424f45871563be02eba14815ba3446158441/crates/types/src/lib.rs#L1493

crates/fuzzing/src/generators/config.rs
```diff
@@ -135,7 +135,7 @@ impl Config {
             if pooling.total_memories < 1
                 || pooling.total_tables < 5
                 || pooling.table_elements < 1_000
-                || pooling.memory_pages < 900
+                || pooling.max_memory_size < (900 * 65536)
```
fitzgen, line 138: ditto regarding named constant

crates/fuzzing/src/generators/config.rs
```diff
@@ -414,20 +414,18 @@ impl<'a> Arbitrary<'a> for Config {
...
+            cfg.max_memory64_pages = min / 65535;
 
             // If traps are disallowed then memories must have at least one page
             // of memory so if we still are only allowing 0 pages of memory then
             // increase that to one here.
             if cfg.disallow_traps {
-                if pooling.memory_pages == 0 {
-                    pooling.memory_pages = 1;
+                if pooling.max_memory_size == 0 {
+                    pooling.max_memory_size = 65535;
```
fitzgen, line 428: Also confused about `65535` here since that isn't a multiple of the page size, so this will disallow any non-zero-length memories, assuming we don't do rounding elsewhere in this PR?

crates/fuzzing/src/generators/config.rs
```diff
@@ -414,20 +414,18 @@ impl<'a> Arbitrary<'a> for Config {
...
-                .min(cfg.max_memory64_pages)
-                .min(pooling.memory_pages);
-            pooling.memory_pages = min;
-            cfg.max_memory32_pages = min;
-            cfg.max_memory64_pages = min;
+            let min_pages = cfg.max_memory32_pages.min(cfg.max_memory64_pages);
+            let min = (min_pages * 65536).min(pooling.max_memory_size as u64);
+            pooling.max_memory_size = min as usize;
+            cfg.max_memory32_pages = min / 65536;
+            cfg.max_memory64_pages = min / 65535;
```
fitzgen, line 421: Also confused about `65535` vs `65536` here.

fitzgen: And ditto about named constants on everything here.

crates/wasmtime/src/config.rs
```diff
@@ -2757,27 +2757,25 @@ impl PoolingAllocationConfig {
         self
     }
 
-    /// The maximum number of Wasm pages for any linear memory defined in a
-    /// module (default is `160`).
+    /// The maximum size that any WebAssembly linear memory may grow to.
```
fitzgen, line 2760:
```suggestion
    /// The maximum byte size that any WebAssembly linear memory may grow to.
```

crates/wasmtime/src/runtime/vm/instance/allocator/pooling.rs
```diff
@@ -703,7 +704,7 @@ mod test {
         let config = PoolingInstanceAllocatorConfig {
             limits: InstanceLimits {
                 total_memories: 1,
-                memory_pages: 0x10001,
+                max_memory_size: (1 << 32) + 65536,
```
fitzgen, line 707: named constant, even tho it's just a test. we want to make sure we are actually testing sizes that wasm could potentially grow to, and not accidentally just testing rounding behavior or something.

crates/wasmtime/src/runtime/vm/instance/allocator/pooling/memory_pool.rs
```diff
@@ -767,7 +763,7 @@ mod tests {
                     max_tables_per_module: 0,
                     max_memories_per_module: 3,
                     table_elements: 0,
-                    memory_pages: 1,
+                    max_memory_size: 65536,
```
fitzgen, line 766: ditto

tests/all/async_functions.rs
```diff
@@ -344,7 +344,9 @@ async fn fuel_eventually_finishes() {
 #[tokio::test]
 async fn async_with_pooling_stacks() {
     let mut pool = crate::small_pool_config();
-    pool.total_stacks(1).memory_pages(1).table_elements(0);
+    pool.total_stacks(1)
+        .max_memory_size(65536)
```
fitzgen, line 348: ditto in this file's changes

tests/all/limits.rs
```diff
@@ -354,7 +354,7 @@ fn test_pooling_allocator_initial_limits_exceeded() -> Result<()> {
     let mut pool = crate::small_pool_config();
     pool.total_memories(2)
         .max_memories_per_module(2)
-        .memory_pages(5)
+        .max_memory_size(5 * 65536)
```
fitzgen, line 357: okay I am going to stop saying "ditto about named constants" now and assume you will go over the diff yourself to catch all of them at this point

fitzgen (commented):
LGTM but I am confused about one thing and have Opinions and Concerns about writing the constant value directly all over the place

crates/fuzzing/src/generators/config.rs
alexcrichton: Ah yeah 65535 is a typo here, and everywhere else, but I'm hesitant to go through all the work to import a constant in all these places since it's all in tests anyway and all the values are basically random constants. In the future the wasm page size won't be constant as well so it feels weird to proliferate usage of the constant... I can change these to `>>16` and `<<16` though where appropriate perhaps to reduce the risk of typos.

crates/wasmtime/src/runtime/vm/instance/allocator/pooling.rs
alexcrichton: Could you expand on this? I could see a comment going here but I'm not sure otherwise what you're looking for in terms of a constant?

alexcrichton:
I've updated with `<<16` and `>>16` to reduce the risk of typos, but I'm curious what you think of my pushback on using a name constant everywhere. I'm basically hesitant to add a bunch of process to tests where all the constants are basically already random

crates/wasmtime/src/runtime/vm/instance/allocator/pooling.rs
fitzgen: If the goal is to test behavior with N wasm pages of capacity, then we should make sure we are actually doing that (as opposed to testing with unaligned capacity) and using a constant helps in this scenario. I feel like the typos are evidence that we can mess this up.

Additionally, having a constant that we can grep for all uses of will make it easier to support the custom-page-sizes proposal down the line: we have a list of uses that we can turn into method calls if necessary, and a list of tests that we might want to update or copy and tweak.

alexcrichton: The meaning of this test has sort of been updated and lost a bit over time as I've been tweaking settings. This isn't specifically testing for unalignment or anything like that it's just one size being bigger than the other and various bits and pieces have been lost to translation over time.

I'm gonna go ahead and merge this as-is but we can of course update tests afterwards. I'm basically just somewhat doubtful if it's worth the effort to try to document all tests for precisely what they're doing page-wise and such. I don't think any of the tests really need to be updated with custom-page-sizes since they're all using wasm memories of 64k pages and the tests for custom-page-sizes are going to be pretty orthogonal anyway. Much of the time the page limits are just to reduce VM memory reservations anyway and don't have much to do with page sizes.

PR #8631, opened 2024-05-15
cranelift: Remove the virtual sp offset from all backends

elliottt:
After the changes required to support tail calls, the general frame layout in cranelift now always includes a stack allocation that's large enough to hold all outgoing args/return values. An effect of this change is that the only place that the virtual sp offset was meaningfully changed was in the function prologue, where it was always updated by the constant amount required for the outgoing arguments area. As the only other place that the virtual sp offset was modified was in the handling for call instructions to functions that used the tail-call abi, the need for the virtual sp offset is no longer obvious.

This PR removes the virtual sp offset from all backends, and instead maintains the invariant that SP always points to the end of the stack frame, right after the outgoing arguments area. This does require the call pseudo-instruction to decrement SP after a call to a tail-call function, as those free their incoming argument area, but I believe this change to be a benefit, as we now no longer directly manipulate SP outside of the function prologue and epilogue.


cranelift/codegen/src/isa/x64/inst/emit.rs
```diff
@@ -1607,9 +1611,18 @@ pub(crate) fn emit(
                 sink.add_call_site(*opcode);
             }
 
-            if let Some(call_info) = info {
-                let callee_pop_size = i64::from(call_info.callee_pop_size);
-                state.adjust_virtual_sp_offset(-callee_pop_size);
+            // Reclaim the outgoing argumetn area that was released by the callee, to ensure that
```
cfallin, line 1614: s/argumetn/argument/ (and below)

cranelift/codegen/src/machinst/abi.rs
```diff
@@ -2453,14 +2400,15 @@ impl<M: ABIMachineSpec> CallSite<M> {
...
-
         let tmp = ctx.alloc_tmp(word_type).only_reg().unwrap();
+
+        // Any adjustment to SP to account for required outgoing arguments/stack return values must
+        // be done inside of the the call pseudo-op, to ensure that SP is always in a consistent
+        // state for all other instructions. For example, if a tail-call abi function is called
+        // here, the reclamation of the outgoing argument area must be done inside of the call
+        // pseudo-op's emission to ensure that SP is consisten at all other points in the lowered
+        // function. (Except the prologue and epilogue, but those are fairly special parts of the
+        // function that establish the SP invariants that are relied on elsewhere.)
```
cfallin, line 2411: (amend comment) ...and are generated after the register allocator has run and thus cannot have register allocator-inserted references to SP offsets.

cfallin (approved):
Excellent simplification, thanks!

cranelift/codegen/src/isa/x64/inst/emit.rs
elliottt: Hello `argumetn`, my old friend.

Issue #8573, opened 2024-05-07
Small difference makes suspicion performance decreasing


hungryzzz:
### Test Cases
[cases.zip](https://github.com/bytecodealliance/wasmtime/files/15237560/cases.zip)

### Steps to Reproduce

Hi, I run the attached two cases(`good.wasm`&`bad.wasm`) in `Wasmtime`, `WasmEdge(AOT)` and `Wamr(AOT)`, and collect their execution time respectively (measured by `time` tool).

### Expected Results & Actual Results

For `good.wasm`, the execution time in different runtimes are as follows:
- `Wasmtime`: 4.67s
- `WasmEdge`: 2.81s
- `Wamr`: 3.29s

For `bad.wasm`, the execution time in different runtimes are as follows:
- `Wasmtime`: 9.34s
- `WasmEdge`: 2.35s
- `Wamr`: 3.29s

The difference between the attached two cases is as follow, i.e., changing the address of `i32.load` from `local.get 0` to `i32.const 0`, the case with `i32.const 0` has the bad performance in `Wasmtime`, whose execution time is 2x slower than the former one. I think the decreasing is confused and suspicious, why is an immediate operation slower than a local variable operation? Moreover, the `i32.load` is out of the loop, I cannot understand why the operation could affect the overall performance so vastly.

![截屏2024-05-07 23 19 36](https://github.com/bytecodealliance/wasmtime/assets/32137313/be606e8b-7da9-4f99-9697-e23b3b53e376)

### Versions and Environment

Wasmtime version or commit: d0cf46a098d97bab9c4cf54a2f3aeaddea7d3831

Operating system: Linux ringzzz-OptiPlex-7070 5.15.0-97-generic

Architecture: Intel(R) Core(TM) i5-9500T CPU @ 2.20GHz

cfallin:
cc @fitzgen , perhaps this is something to do with our bounds-checking cases?

fitzgen:
@hungryzzz can you share how you are configuring and executing the Wasm in Wasmtime? I.e. the exact flags you are passing to the CLI or the exact Rust code that you are using when embedding Wasmtime if you aren't using the CLI.

> cc @fitzgen , perhaps this is something to do with our bounds-checking cases?

With the default configuration of Wasmtime we should be eliding basically all bounds checks, I'd be surprised if this was bounds-check related unless guard pages are being disabled or something.

hungryzzz:
[...]

PR #8633, opened 2024-05-16
Change the function alignment of x86 to 32-bytes

hungryzzz:
The current default 16-bytes function alignment for x86-64 would cause suboptimal execution performance under some cases which are reported in https://github.com/bytecodealliance/wasmtime/issues/8573.

Based on the discussion "the CPU frontend grabs an aligned 32B or 64B chunk at a time" in https://github.com/bytecodealliance/wasmtime/issues/8573, this PR changes the default alignment from 16-bytes to 32-bytes for better performance. 

Also rerun the cases reported in https://github.com/bytecodealliance/wasmtime/issues/8573 and the execution time will back to normal.
```shell
# After changes
➜  case ✗ wasmtime compile good.wasm -o good.cwasm
➜  case ✗ wasmtime compile bad.wasm -o bad.cwasm
➜  case ✗ time wasmtime run --allow-precompiled good.cwasm
~/wasmtime/target/release/wasmtime run  good.cwasm  4.68s user 0.00s system 100% cpu 4.680 total
➜  case ✗ time wasmtime run --allow-precompiled bad.cwasm
~/wasmtime/target/release/wasmtime run  bad.cwasm  4.67s user 0.01s system 99% cpu 4.681 total


# Before changes
➜  case ✗ wasmtime compile good.wasm -o good.cwasm
➜  case ✗ wasmtime compile bad.wasm -o bad.cwasm
➜  case ✗ time wasmtime run --allow-precompiled good.cwasm
~/wasmtime/target/release/wasmtime run  good.cwasm  4.67s user 0.00s system 100% cpu 4.674 total
➜  case ✗ time wasmtime run --allow-precompiled bad.cwasm
~/wasmtime/target/release/wasmtime run  bad.cwasm  9.36s user 0.01s system 100% cpu 9.365 total
```

### DIFF
--- a/cranelift/codegen/src/isa/x64/inst/mod.rs
+++ b/cranelift/codegen/src/isa/x64/inst/mod.rs
@@ -2702,9 +2702,9 @@ impl MachInst for Inst {
     fn function_alignment() -> FunctionAlignment {
         FunctionAlignment {
             minimum: 1,
-            // Prefer an alignment of 16-bytes to hypothetically get the whole
-            // function into a minimum number of lines.
-            preferred: 16,
+            // Change the alignment from 16-bytes to 32-bytes for better performance.
+            // fix-8573: https://github.com/bytecodealliance/wasmtime/issues/8573
+            preferred: 32,
         }
     }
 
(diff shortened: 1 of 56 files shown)


alexcrichton:
Thanks for this! It looks like this is causing changes in the disassembly of some existing tests. You can update the disassembly with `WASMTIME_TEST_BLESS=1 cargo test --test disas` locally and commit those changes to get pushed up here

alexcrichton (approved):
Thanks!

Issue #7772, opened 2024-01-11
Cranelift: add all Cranelift crates to `cranelift` umbrella crate


Rodrigodd:
#### Feature

Add all Cranelift crates (`cranelift-codegen`, `cranelift-frontend`, `cranelift-module`, `cranelift-native`, `cranelift-jit`, `cranelift-object`, etc.) to the `cranelift` umbrella crate, each one behind a feature preferably.

#### Benefit

The `cranelift` crate is an "umbrella for commonly-used Cranelift crates" and "provides a convenient one-line dependency," but it just contains `frontend` and `codegen`, which is limited to creating in-memory JIT compilers. That means that if I expect to create a static compiler or generate object files in some form, I need to replace this crate with its underlying ones, defeating its purpose.

It also helps keep all versions in sync and make all these crates a little more discoverable.

#### Alternatives

- Keep it the way it is: Cranelift has a focus on JIT compilation, so maybe these two crates are the only common ones, and the remaining ones are niche enough.

PR #8634, opened 2024-05-16
cranelift: expand umbrella crate with more crates

tertsdiepraam:
Closes https://github.com/bytecodealliance/wasmtime/issues/7772

This adds the `interpreter`, `jit`, `module`, `native` & `object` crates to the umbrella crate. All crates (including the existing `codegen` and `frontend` crates) are feature gated. However, `codegen` and `frontend` are default features, so backwards compatibility is retained if default features is not set to false. I wasn't quite sure which crates to include, but in my own project, I already depend on `jit`, `module` and `native`. Based on their descriptions, `interpreter` and `object` also seemed relevant as public API. I might have missed some other important crate.

I figured this would be an easy fix, although I understand if the umbrella crate should not have all these crates. In that case, feel free to close this PR :)

### DIFF
--- a/cranelift/umbrella/Cargo.toml
+++ b/cranelift/umbrella/Cargo.toml
@@ -13,9 +13,28 @@ edition.workspace = true
 
 [dependencies]
 cranelift-codegen = { workspace = true }
-cranelift-frontend = { workspace = true }
+cranelift-frontend = { workspace = true, optional = true }
+cranelift-interpreter = { workspace = true, optional = true }
+cranelift-jit = { workspace = true, optional = true }
+cranelift-module = { workspace = true, optional = true }
+cranelift-native = { workspace = true, optional = true }
+cranelift-object = { workspace = true, optional = true }
 
 [features]
-default = ["std"]
-std = ["cranelift-codegen/std", "cranelift-frontend/std"]
-core = ["cranelift-codegen/core", "cranelift-frontend/core"]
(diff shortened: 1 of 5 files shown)


cranelift/umbrella/Cargo.toml
```diff
@@ -12,10 +12,30 @@ keywords = ["compile", "compiler", "jit"]
 edition.workspace = true
 
 [dependencies]
-cranelift-codegen = { workspace = true }
-cranelift-frontend = { workspace = true }
+cranelift-codegen = { workspace = true, optional = true }
```
bjorn3, line 15: All other crates depend on cranelift-codegen, so I don't see any reason why you should be able to disable it.

tertsdiepraam: Makes sense! Updated!

elliottt (approved):
This all seems reasonable to me, thanks for fixing this!

elliottt:
The publish step is failing because `cranelift-jit` has a dev-dependency on `cranelift` for its examples. Can we move the jit examples out of `cranelift-jit` to break this cycle? Perhaps the umbrella crate would be a better place to centralize stuff like this now.

elliottt:
(@alexcrichton mentioned that the cycle could be broken by using a `{ path = ... }` version instead, so I went ahead and pushed that change. I'll merge this once it passes branch CI)

PR #8635, opened 2024-05-16
Cranelift: add alignment parameter to stack slots.

### LINKED ISSUE #6716: Support specifying stack slot alignments
#### Feature

See title.

#### Benefit

Correct alignment are necessary to avoid crashes on some architectures and code may depend on correct alignment. In addition rustc checks that the right alignment is used when dereferencing a raw pointer in debug mode. This breaks rayon which allocates a stack value with cacheline alignment and then takes a raw pointer which it later dereferences. See https://github.com/bjorn3/rustc_codegen_cranelift/issues/1381.

#### Implementation

Sort stackslots by alignment for better packing and then add padding as necessary and if the alignment of a stack slot exceeds the ABI stack alignment realign the stack at runtime.

#### Alternatives

Doing dynamic alignment at runtime in the cranelift ir producer. This is slower and has higher stack usage overhead.

cfallin:
Fixes #6716.

Currently, stack slots on the stack are aligned only to a machine-word boundary. This is insufficient for some use-cases: for example, storing SIMD data or structs that require a larger alignment.

This PR adds a parameter to the `StackSlotData` to specify alignment, and the associated logic to the CLIF parser and printer. It updates the shared ABI code to compute the stackslot layout taking the alignment into account. In order to ensure the alignment is always a power of two, it is stored as a shift amount (log2 of actual alignment) in the IR.

### DIFF
--- a/cranelift/codegen/src/ir/stackslot.rs
+++ b/cranelift/codegen/src/ir/stackslot.rs
@@ -70,32 +70,38 @@ pub struct StackSlotData {
 
     /// Size of stack slot in bytes.
     pub size: StackSize,
+
+    /// Alignment of stack slot as a power-of-two exponent (log2
+    /// value). The stack slot will be at least this aligned; it may
+    /// be aligned according to other considerations, such as minimum
+    /// stack slot size or machine word size, as well.
+    pub align_shift: u8,
 }
 
 impl StackSlotData {
-    /// Create a stack slot with the specified byte size.
-    pub fn new(kind: StackSlotKind, size: StackSize) -> Self {
-        Self { kind, size }
-    }
-
-    /// Get the alignment in bytes of this stack slot given the stack pointer alignment.
-    pub fn alignment(&self, max_align: StackSize) -> StackSize {
-        debug_assert!(max_align.is_power_of_two());
-        if self.kind == StackSlotKind::ExplicitDynamicSlot {
-            max_align
-        } else {
-            // We want to find the largest power of two that divides both `self.size` and `max_align`.
-            // That is the same as isolating the rightmost bit in `x`.
-            let x = self.size | max_align;
-            // C.f. Hacker's delight.
-            x & x.wrapping_neg()
+    /// Create a stack slot with the specified byte size and alignment.
+    pub fn new(kind: StackSlotKind, size: StackSize, align_shift: u8) -> Self {
+        Self {
+            kind,
+            size,
+            align_shift,
         }
     }
 }
 
 impl fmt::Display for StackSlotData {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{} {}", self.kind, self.size)
+        if self.align_shift != 0 {
+            write!(
+                f,
+                "{} {}, align = {}",
+                self.kind,
+                self.size,
+                1u32 << self.align_shift
+            )
+        } else {
+            write!(f, "{} {}", self.kind, self.size)
+        }
     }
 }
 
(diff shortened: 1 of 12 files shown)


cranelift/codegen/src/ir/stackslot.rs
```diff
@@ -70,32 +70,28 @@ pub struct StackSlotData {
...
-        } else {
-            // We want to find the largest power of two that divides both `self.size` and `max_align`.
-            // That is the same as isolating the rightmost bit in `x`.
-            let x = self.size | max_align;
-            // C.f. Hacker's delight.
-            x & x.wrapping_neg()
-        }
+    /// Create a stack slot with the specified byte size and alignment.
+    pub fn new(kind: StackSlotKind, size: StackSize, align_shift: u8) -> Self {
+        Self { kind, size, align_shift }
```
cfallin, line 84: (Just below this point) the old `alignment` function didn't seem to be used in the ABI code or anywhere else; it's likely left over from some earlier use-case. The "largest power of two that divides..." logic is a little weird (it can result in an alignment less than requested?) so I removed it and pushed the alignment computation to the `std::cmp::max` in `abi.rs`.

cranelift/filetests/filetests/isa/x64/stackslot.clif
```diff
@@ -0,0 +1,50 @@
+test compile precise-output
+target x86_64
+
+function %f0() -> i64, i64, i64, i64 {
+  ss0 = explicit_slot 8, align=16
+  ss1 = explicit_slot 8, align=16
+  ss2 = explicit_slot 4
+  ss3 = explicit_slot 4
```
elliottt, line 8: Do we have a way of enabling an explicit stack check in clif tests? That would be a nice way of verifying that the total stack space required is accounted for.

cranelift/reader/src/parser.rs
```diff
@@ -1498,7 +1502,29 @@ impl<'a> Parser<'a> {
...
+            self.consume();
+            self.match_token(
+                Token::Identifier("align"),
+                "expected a valid stack-slot flag (currently only `align`)",
+            )?;
+            self.match_token(Token::Equal, "expected `=` after flag")?;
+            let align: i64 = self
+                .match_imm64("expected alignment-size after `align` flag")?
+                .into();
+            u32::try_from(align).map_err(|_| self.error("invalid alignment"))?
```
elliottt, line 1517:
```suggestion
            u32::try_from(align).map_err(|_| self.error("alignment must be a 32-bit unsigned integer"))?
```

elliottt (approved):
Awesome!

crates/cranelift/src/compiler.rs
```diff
@@ -942,6 +942,7 @@ impl Compiler {
         let slot = builder.func.create_sized_stack_slot(ir::StackSlotData::new(
             ir::StackSlotKind::ExplicitSlot,
             values_vec_byte_size,
+            0,
```
alexcrichton, line 945: Can this be set to 4? The ValRaw bits in Wasmtime require a 16-byte alignment

cranelift/filetests/filetests/isa/x64/stackslot.clif
elliottt: Ah!
```suggestion
function %f0() -> i64, i64, i64, i64 {
  gv0 = vmctx
  ss0 = explicit_slot 8, align=16
  ss1 = explicit_slot 8, align=16
  ss2 = explicit_slot 4
  ss3 = explicit_slot 4
  ss4 = stack_limit gv0
```

crates/cranelift/src/compiler.rs
cfallin: Done!

cranelift/codegen/src/machinst/abi.rs
```diff
@@ -1178,7 +1178,11 @@ impl<M: ABIMachineSpec> Callee<M> {
             sized_stack_offset = sized_stack_offset
                 .checked_add(data.size)
                 .ok_or(CodegenError::ImplLimitExceeded)?;
-            let mask = M::word_bytes() - 1;
+            // Always at least machine-word-align slots, but also
+            // satisfy the user's requested alignment.
+            debug_assert!(data.align_shift < 32);
+            let align = std::cmp::max(M::word_bytes(), 1u32 << data.align_shift);
+            let mask = align - 1;
```
bjorn3, line 1185: If the alignment is bigger than the ABI stack alignment it would be necessary to adjust the stack pointer to correctly align it. This is missing here. Rust allows arbitrary power of two alignments for stack variables, even something crazy like a 1GB alignment. More realistically though, cache line alignment is used by crates like rayon for performance reasons. Without correctly aligning, rustc will trigger a debug assertion in some cases like for the aforementioned rayon crate.

cfallin: Ah, yes, that's a good point -- to summarize: alignment of offset within the stack frame to alignment A does not imply alignment to A overall, because SP itself may be aligned less (e.g. only to A/4).

I don't have time to work further on this at the moment, but I'd be happy to review a PR that addressed this.

PR #8638, opened 2024-05-16
Fix the change introduced in #8636, as the offset is to the CFA

elliottt:
The offset computed in `compute_value_labels_ranges` is relative to the CFA, not the current value of SP, which means that the change introduced in #8636 is incorrect. This PR fixes that misunderstanding, and renames the `sp_to_caller_sp_offset` function to `slot_base_to_caller_sp_offset` to make it really clear what the value being computed is used for.

cc @jameysharp


jameysharp (approved):
I believe this is correct, thank you! It's very confusing to reason about.

PR #8639, opened 2024-05-16
 Show function names in explore tool instead of only function indices

lpereira:
Another one in the "let's make debugging a bit easier": show function names in the "explore" tool.  (There's a bit of code repetition here to clean up the names, but it's slightly different when we're considering HTML output, so I think that's OK.)

Here's how it works on my machine:
image

### DIFF
--- a/crates/explorer/src/index.js
+++ b/crates/explorer/src/index.js
@@ -191,12 +191,12 @@ const renderInst = (mnemonic, operands) => {
 
 // Render the ASM.
 
-let nthFunc = 0;
 for (const func of state.asm.functions) {
   const funcElem = document.createElement("div");
 
   const funcHeader = document.createElement("h3");
-  funcHeader.textContent = `Defined Function ${nthFunc}`;
+  funcHeader.textContent = `Disassembly of function <${func.demangled_name}>:`;
+  funcHeader.title = func.name;
   funcElem.appendChild(funcHeader);
 
   const bodyElem = document.createElement("pre");
@@ -216,7 +216,6 @@ for (const func of state.asm.functions) {
   funcElem.appendChild(bodyElem);
 
   asmElem.appendChild(funcElem);
-  nthFunc++;
 }
 
 // Render the WAT.
--- a/crates/explorer/src/index.js
+++ b/crates/explorer/src/index.js
@@ -195,8 +195,11 @@ for (const func of state.asm.functions) {
   const funcElem = document.createElement("div");
 
   const funcHeader = document.createElement("h3");
-  funcHeader.textContent = `Disassembly of function <${func.demangled_name}>:`;
-  funcHeader.title = func.name;
+  let func_name = func.name === null ? `function[${func.index}]` : func.name;
+  let demangled_name =
+    func.demangled_name === null ? func.demangled_name : func_name;
+  funcHeader.textContent = `Disassembly of function <${demangled_name}>:`;
+  funcHeader.title = `Function ${func.index}: ${func_name}`;
   funcElem.appendChild(funcHeader);
 
   const bodyElem = document.createElement("pre");
--- a/crates/explorer/src/index.js
+++ b/crates/explorer/src/index.js
@@ -195,11 +195,12 @@ for (const func of state.asm.functions) {
   const funcElem = document.createElement("div");
 
   const funcHeader = document.createElement("h3");
-  let func_name = func.name === null ? `function[${func.index}]` : func.name;
+  let func_name =
+    func.name === null ? `function[${func.func_index}]` : func.name;
   let demangled_name =
-    func.demangled_name === null ? func.demangled_name : func_name;
+    func.demangled_name !== null ? func.demangled_name : func_name;
   funcHeader.textContent = `Disassembly of function <${demangled_name}>:`;
-  funcHeader.title = `Function ${func.index}: ${func_name}`;
+  funcHeader.title = `Function ${func.func_index}: ${func_name}`;
   funcElem.appendChild(funcHeader);
 
   const bodyElem = document.createElement("pre");
--- a/crates/explorer/src/lib.rs
+++ b/crates/explorer/src/lib.rs
@@ -133,9 +133,9 @@ fn annotate_asm(
     };
 
     let functions = module
-        .function_locations_with_names()
-        .map(|loc| {
-            let body = &text[loc.offset..][..loc.len];
+        .functions()
+        .map(|function| {
+            let body = &text[function.offset..][..function.len];
 
             let mut cs = match target.architecture {
                 target_lexicon::Architecture::Aarch64(_) => capstone::Capstone::new()
@@ -168,13 +168,13 @@ fn annotate_asm(
             cs.set_skipdata(true).unwrap();
 
             let instructions = cs
-                .disasm_all(body, loc.offset as u64)
+                .disasm_all(body, function.offset as u64)
                 .map_err(|e| anyhow::anyhow!("{e}"))?;
             let instructions = instructions
                 .iter()
                 .map(|inst| {
                     let address = u32::try_from(inst.address()).unwrap();
-                    let wasm_offset = wasm_offset_for_address(loc.offset, address);
+                    let wasm_offset = wasm_offset_for_address(function.offset, address);
                     Ok(AnnotatedInstruction {
                         wasm_offset,
                         address,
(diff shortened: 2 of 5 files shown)


jameysharp:
Very nice!

This can be further improved by demangling the function names first. For some example code that does that, see: https://github.com/bytecodealliance/wasmtime/blob/54e53cc77a29c8e4eb819333d81ef4158685690a/crates/wasmtime/src/runtime/profiling.rs#L194-L200

I don't think we need to be nearly so careful about sanitizing function names in this setting. These strings are first serialized to JSON, and that should ensure that any string we provide is valid JSON. The JSON string is then used in a `<script>` tag; we might need to escape the JSON for embedding in HTML, but `<script>` tags are implicitly treated as CDATA in HTML5, right? Finally, the strings embedded in the JSON are spliced into the DOM using `textContent`, so they don't need to be escaped at that point.

As a result I think we should just give the user whatever they provided, preferably after demangling.

crates/wasmtime/src/runtime/module.rs
```diff
@@ -1027,6 +1027,25 @@ impl Module {
...
+    /// Get the locations of functions in this module's `.text` section, with the
+    /// function names.
+    ///
+    /// Each function's location is a (name, `.text` section offset, length) pair.
+    pub fn function_locations_with_names<'a>(
+        &'a self,
+    ) -> impl ExactSizeIterator<Item = (String, usize, usize)> + 'a {
+        let module = self.compiled_module();
+        self.function_locations().map(|(offset, len)| {
+            let (idx, _) = module.func_by_text_offset(offset).unwrap();
```
fitzgen, line 1039: We can avoid looking up the function index via instruction offset since `function_locations` should already return the functions in order:

```suggestion
        self.function_locations().enumerate().map(|(idx, (offset, len))| {
            let idx = u32::try_from(idx).unwrap();
```

fitzgen (commented):
Thanks! One nitpick below before we merge this.

fitzgen:
Agreed demangling would be great -- also fine with deferring that to a follow up PR.

lpereira:
Pushed the requested changes. Current output here is now this one (notice the pop-up showing the unmangled name):

image

jameysharp:
Ooh, the pop-up is a nice touch. There's a lot more that could be done with `wasmtime explore` eventually.

I'd be happy to review this next week, or just as happy if you want to approve it @fitzgen.

crates/explorer/src/lib.rs
```diff
@@ -181,7 +183,21 @@ fn annotate_asm(
...
-            Ok(AnnotatedFunction { instructions })
+
+            let demangled_name = match name.splitn(2, "::").nth(1) {
+                Some(name) => {
+                    let mut demangled = String::new();
+                    demangle_function_name(&mut demangled, name)
+                        .map_or_else(|_| format!("demangle-error::{}", name), |_| demangled)
+                }
+                None => name.to_string(),
+            };
```
fitzgen, line 194: I think it would be better to let `demangled_name` be an `Option<String>` rather than making a clone of `name` and then have the JS choose the demangled name when available and fall back to the plain name.

crates/wasmtime/src/runtime/module.rs
```diff
@@ -1027,6 +1027,27 @@ impl Module {
...
+        self.function_locations()
+            .enumerate()
+            .map(|(idx, (offset, len))| {
+                let idx = DefinedFuncIndex::from_u32(u32::try_from(idx).unwrap());
+                let idx = module.module().func_index(idx);
+                if let Some(name) = module.func_name(idx) {
+                    (format!("function[{}]::{}", idx.as_u32(), name), offset, len)
+                } else {
+                    (format!("function[{}]", idx.as_u32()), offset, len)
+                }
```
fitzgen, line 1047: Instead of stringifying the name with the index here, and then forcing the consumer of this API in `crates/explore/src/lib.rs` to parse the name out from stringified name out, can we just give the caller the various components before stringification? That is, return an iterator that yields items of type

```rust
(
    u32,            // wasm function index
    Option<String>, // name, if available
    usize,          // text section offset
    usize,          // length of function in text section
)
```

I think we could also update the existing `function_locations` method, and perhaps rename it, rather than defining a whole new method, since that method exists to basically enable `wasmtime-explore`.

fitzgen (commented):
Thanks! Really excited to get this, and your other PR, in `wasmtime explore`.

I have a few comments and suggestions below that we should address before merging this however. I'll also hold off on reviewing the other PR until after this one merges.

Let me know if anything I said isn't clear or something like that. Thanks again!

crates/wasmtime/src/runtime/module.rs
lpereira: Alright.

crates/explorer/src/lib.rs
lpereira: OK.

crates/explorer/src/index.js
```diff
@@ -191,12 +184,12 @@ const renderInst = (mnemonic, operands) => {
...
 // Render the ASM.
 
-let nthFunc = 0;
 for (const func of state.asm.functions) {
   const funcElem = document.createElement("div");
 
   const funcHeader = document.createElement("h3");
-  funcHeader.textContent = `Defined Function ${nthFunc}`;
+  funcHeader.textContent = `Disassembly of function <${func.demangled_name}>:`;
+  funcHeader.title = func.name;
```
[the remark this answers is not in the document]
lpereira: Sure.  It was the auto-formatting tool that my editor uses.  I'll have to force-push to fix this easily.

crates/explorer/src/index.js
```diff
@@ -97,44 +82,52 @@ const addAsmElem = (offset, elem) => {
...
-}, { passive: true });
+    const firstAsmElem = asmByOffset.get(offset)[0];
+    firstAsmElem.scrollIntoView({
+      behavior: "smooth",
+      block: "center",
+      inline: "nearest",
+    });
+  },
+  { passive: true },
+);
```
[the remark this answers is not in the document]
lpereira: OK.

lpereira:
Alright, force-pushed (sorry!) due to the formatting changes, but it should be a whole lot clearer now.  I'll rebase the other PR once this one lands.

crates/wasmtime/src/runtime/module.rs
```diff
@@ -1027,11 +1027,42 @@ impl Module {
         })
     }
 
+    /// Get the locations of functions in this module's `.text` section, with the
+    /// function names.
+    ///
+    /// Each function's location is a (name, `.text` section offset, length) pair.
+    pub fn function_locations_with_names<'a>(
+        &'a self,
+    ) -> impl ExactSizeIterator<Item = FunctionLocationWithName> + 'a {
```
fitzgen, line 1036: I think `function_loactions` is only used by this method, so we can we remove it or make it non-`pub`, as I suggested previously?

Also I like the named struct with named fields, that's indeed better than a large-ish tuple, but the docs here should match that now.

And since this is really more than just function locations and names (it also has an index, for example) maybe we should call this method `function_metadatas` and the struct `FunctionMetadata`? Feel free to take-or-leave this last one...

fitzgen (commented):
Thanks! Looks just about ready to merge, couple tiny things still to go first.

crates/wasmtime/src/runtime/module.rs
lpereira: Good catch, let me clean this up.

crates/explorer/src/index.js
```diff
@@ -191,12 +191,15 @@ const renderInst = (mnemonic, operands) => {
...
 
-let nthFunc = 0;
 for (const func of state.asm.functions) {
   const funcElem = document.createElement("div");
 
   const funcHeader = document.createElement("h3");
-  funcHeader.textContent = `Defined Function ${nthFunc}`;
+  let func_name = func.name === null ? `function[${func.index}]` : func.name;
+  let demangled_name =
+    func.demangled_name === null ? func.demangled_name : func.name;
```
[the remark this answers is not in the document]
lpereira: Oops!

lpereira:
Alright, @fitzgen, pushed a new revision that should address your comments.

fitzgen (approved):
Looks great, thanks! And thanks for your patience with review :)

PR #8640, opened 2024-05-16
Cranelift: Split out dominator tree's depth-first traversal into a reusable iterator

fitzgen:
We intend to use this when computing liveness of GC references in `cranelift-frontend` to manually construct safepoints and ultimately remove `r{32,64}` reference types from CLIF, `cranelift-codegen`, and `regalloc2`.


cfallin (approved):
A nice generalization -- LGTM!

PR #8641, opened 2024-05-16
Fix some typos

FrankReh:
Corrected typos from various parts of the crate.

The commits are broken down into different parts in case there are different owners who would take on reviewing.

Most are in comments. A few are in asserts and error messages.

Before anyone gets too annoyed with me, consider the time saved for all who will be reading this code in the years to come. :pray:

### DIFF
--- a/crates/c-api/include/wasmtime/store.h
+++ b/crates/c-api/include/wasmtime/store.h
@@ -43,7 +43,7 @@ typedef struct wasmtime_store wasmtime_store_t;
  * \brief An interior pointer into a #wasmtime_store_t which is used as
  * "context" for many functions.
  *
- * This context pointer is used pervasively throught Wasmtime's API. This can be
+ * This context pointer is used pervasively throughout Wasmtime's API. This can be
  * acquired from #wasmtime_store_context or #wasmtime_caller_context. The
  * context pointer for a store is the same for the entire lifetime of a store,
  * so it can safely be stored adjacent to a #wasmtime_store_t itself.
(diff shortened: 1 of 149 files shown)


abrown (approved):

abrown:
Might need to run clang-format on one of those header files...

FrankReh:
> Might need to run clang-format on one of those header files...

I'm very much over my head using the c headers. It's been a minute since I was compiling c. I'm happy to take help in understanding what I should do get the changes to the headers accepted by ci, but I can also revert them if the changes are making things too complicated.

crates/c-api/include/wasmtime/store.h
```diff
@@ -43,7 +43,7 @@ typedef struct wasmtime_store wasmtime_store_t;
  * \brief An interior pointer into a #wasmtime_store_t which is used as
  * "context" for many functions.
  *
- * This context pointer is used pervasively throught Wasmtime's API. This can be
+ * This context pointer is used pervasively throughout Wasmtime's API. This can be
  * acquired from #wasmtime_store_context or #wasmtime_caller_context. The
```
abrown, line 47: I think it boils down to this:

```suggestion
 * This context pointer is used pervasively throughout Wasmtime's API. This can
 * be acquired from #wasmtime_store_context or #wasmtime_caller_context. The
```

`clang-format` wants the comments to all be a certain length.

abrown: I went ahead and merged this suggestion to see if there are any other problems.

PR #8642, opened 2024-05-17
Refactor installation of C API and features supported 

alexcrichton:
This commit overhauls and refactors the management of the building of
the C API. Instead of this being script-based it's now done entirely
through CMake and CMake is the primary focus for building the C API. For
example building the C API release artifacts is now done through CMake
instead of through a shell script's `cargo build` and manually moving
artifacts.

The benefits that this brings are:

* The C API now properly hides symbols in its header files that weren't
  enabled at build time. This is done through a new build-time generated
  `conf.h` templated on a `conf.h.in` file in the source tree.

* The C API's project now supports enabling/disabling Cargo features to
  have finer-grained support over what's included (plus auto-management
  of the header file).

* Building the C API and managing it is now exclusively done through
  CMake. For example invoking `doxygen` now lives in CMake, installation
  lives there, etc.

The `CMakeLists.txt` file for the C API is overhauled in the process of
doing this. The build now primarily matches on the Rust target being
built rather than the host target. Additionally installation will now
install both the static and shared libraries instead of just one.
Additionally during this refactoring various bits and pieces of
Android-specific code were all removed. Management of the C toolchain
feels best left in scope of the caller (e.g. configuring `CC_*` env vars
and such) rather than here.


crates/c-api/CMakeLists.txt
```diff
@@ -1,140 +1,174 @@
 cmake_minimum_required(VERSION 3.10)
 project(wasmtime C)
 
-set(WASMTIME_USER_CARGO_BUILD_OPTIONS "" CACHE STRING "Additional cargo flags (such as --features) to apply to the build command")
+option(WASMTIME_USER_CARGO_BUILD_OPTIONS "Additional cargo flags (such as --features) to apply to the build command" OFF)
```
dundargoc, line 4: Sus. `option` is for cache variables that are booleans, but it looks like `WASMTIME_USER_CARGO_BUILD_OPTIONS` is supposed to be a string. This doesn't look correct.

crates/c-api/CMakeLists.txt
```diff
@@ -1,140 +1,174 @@
...
 option(BUILD_SHARED_LIBS "Build using shared libraries" OFF)
 option(WASMTIME_ALWAYS_BUILD "If cmake should always invoke cargo to build wasmtime" ON)
 option(WASMTIME_FASTEST_RUNTIME "Set flags designed to optimize runtime performance" OFF)
+option(WASMTIME_TARGET "Rust target to build for" OFF)
```
dundargoc, line 8: Same thing, `option` is for booleans and this expects a string.

crates/c-api/CMakeLists.txt
```diff
@@ -1,140 +1,174 @@
...
+macro(feature cmake_name rust_name default)
+  if(${default})
+    if(${WASMTIME_DISABLE_ALL_FEATURES})
+      set(feature_default OFF)
+    else()
+      set(feature_default ON)
+    endif()
+  else()
+    set(feature_default OFF)
+  endif()
```
dundargoc, line 35: Nitpick: arguably cleaner

```suggestion
  set(feature_default "${default} AND NOT ${WASMTIME_DISABLE_ALL_FEATURES}")
```

crates/c-api/CMakeLists.txt
alexcrichton: Thanks! I didn't really realize that `set(...)` here was doing that. I'll switch back.

crates/c-api/CMakeLists.txt
alexcrichton: I don't think I fully understand cmake evaluation contexts and strings and what's a variable when, but when I use your suggestion it looks like it sets `feature_default` to the string `ON AND NOT OFF` and I'm not sure why but the default build has all features off.

I'm also not sure if I'm supposed to do something like `if(${${cmake_name}})` below instead of just `if(${cmake_name})` (sorry I'm very new to cmake)

dundargoc: Actually I retried it and my suggestion was wrong. My bad, I should've investigated better beforehand. Ignore my comment, your method is the way to go.

crates/c-api/CMakeLists.txt
```diff
@@ -5,136 +5,167 @@ set(WASMTIME_USER_CARGO_BUILD_OPTIONS "" CACHE STRING "Additional cargo flags (s
...
-	wasmtime-crate
-	DOWNLOAD_COMMAND ""
-	CONFIGURE_COMMAND ""
-	INSTALL_COMMAND "${WASMTIME_INSTALL_COMMAND}"
+    wasmtime-crate
+    DOWNLOAD_COMMAND ""
+    CONFIGURE_COMMAND ""
+    INSTALL_COMMAND "${WASMTIME_INSTALL_COMMAND}"
         BUILD_COMMAND
             ${CMAKE_COMMAND} -E env ${CARGO_PROFILE_PANIC}=abort
```
rockwotj, line 107: While I am thinking about it - the c compiler should be set as an environment variable for the cc crate. There are a number of ways to set it in cmake that don't involve the environment variables that the cc crate uses

rockwotj (approved):
Very nice! This is a great change. Just a few small things

crates/c-api/CMakeLists.txt
alexcrichton: True! I'm a bit wary to do that though since cmake's cross-compilation defaults are not always as good as the `cc` crate's auto-detection. Do you know if there's a way to only pass the C compiler env vars and such if explicitly configured in cmake?

elliottt (approved):

alexcrichton:
Thanks @dundargoc and @rockwotj!

PR #8643, opened 2024-05-17
cranelift: Remove nominal-sp

elliottt:
As we now keep SP at the same place throughout the function body, or only modify it within the emission for individual pseudo-ops like call, we no longer need the notion of the nominal-SP. This PR removes all mentions of the nominal SP, and reworks the docs to more accurately describe the frame layout the backends currently assume.

I'm not 100% sold on the change of representation for slot offsets in the riscv64 backend, and am happy to take suggestions for a better representation than `(slot)`.


abrown (approved):
Looks good to me. I didn't page in all the context try to completely grok the doc changes for correctness but they seemed reasonable otherwise.

PR #8650, opened 2024-05-17
fuzzgen: Add support for stack slot alignment

afonso360:
👋 Hey,

This PR is a follow up to #8635. It adds the bare minimum possible for the Cranelift Fuzzgen to start fuzzing with different stack slot alignments.

Now that we can specify alignments, we should be able to start fuzzing load sinking with stack addresses 🎉 . That isn't done yet, but this is the first step.

### DIFF
--- a/cranelift/fuzzgen/src/config.rs
+++ b/cranelift/fuzzgen/src/config.rs
@@ -37,6 +37,8 @@ pub struct Config {
     pub static_stack_slots_per_function: RangeInclusive<usize>,
     /// Size in bytes
     pub static_stack_slot_size: RangeInclusive<usize>,
+    /// Stack slot alignment as a power of 2
+    pub stack_slot_alignment_log2: RangeInclusive<usize>,
     /// Allowed stack probe sizes
     pub stack_probe_size_log2: RangeInclusive<usize>,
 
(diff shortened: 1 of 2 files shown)


cranelift/fuzzgen/src/config.rs
```diff
@@ -86,6 +88,7 @@ impl Default for Config {
             switch_max_range_size: 2..=32,
             static_stack_slots_per_function: 0..=8,
             static_stack_slot_size: 0..=128,
+            stack_slot_alignment_log2: 0..=10,
```
bjorn3, line 91: Currently the max working alignment is 2^4 (16 bytes): https://github.com/bytecodealliance/wasmtime/pull/8635#discussion_r1603954465

afonso360: 👍 I've been following that thread. We currently don't depend on any alignment larger than 8 bytes. And at most we'll need 16bytes (for sinking SIMD loads).

But I think it's still a good idea to test with larger values, even if we end up not having any guarantee past 16 bytes.

bjorn3: I missed that you are not actually checking that the stack slot is correctly aligned at runtime.

elliottt (approved):
This looks good to me! Do you have a plan for verifying alignment in the future?

afonso360:
Yes, but hopefully we shouldn't need to.

The plan is to rely on the ABI alignment and generate loads/stores with the aligned flag if both the slot and the ABI support that alignment. That should allow us to start testing load sinking.

Now, I do plan on validating if these memory accesses are actually aligned with the interpreter. The interpreter already checks if stack accesses are aligned, when they contain the `aligned` memflag, but it currently doesn't support aligning the stack slots. Or align the stack at the `call` boundary. So those are the two missing items.

This hopefully shouldn't be needed since, if we're failing fuzz runs in the interpreter its not great for fuzzing efficiency.

Issue #8599, opened 2024-05-12
ISLE: eliminate distinction between constructors and extractors


jameysharp:
#### Feature

ISLE terms should just be functions, from any number of arguments to any number of results. Internal terms should all be defined using `rule`. Terms with exactly one argument can be used in pattern matches like extractors are today; all terms can be used on the right-hand side of a rule or if-let, like constructors are today. Possibly a form of type-directed overloading should be added to the language to allow the current pattern of using the same term name for both constructor and extractor.

#### Benefit

- Simplify the trie-again and serialized representations of ISLE rules. The current distinction between constructor and extractor makes no difference to code generation.
- Using `rule` to define "extractors" allows a form of "or-patterns" (#6128), without requiring us to figure out how priorities should interact when these patterns are inlined as internal extractors are today.

#### Implementation

I'd start by unifying constructors and extractors in the trie-again representation and ensuring that the generated code doesn't change.

Then maybe introduce something like `(overload iadd extract_iadd construct_iadd)` to declare that when the term `iadd` is used, whichever of `extract_iadd` or `construct_iadd` has the right type for that context should be used instead. I'm less sure about this.

Next, extend constructor terms to allow multiple results. A pattern binding the result of such a term needs some extra syntactic care to handle tuples. Semantically, though, we can approximate this today by wrapping up multiple results in an enum with a single variant, so I think we should just make the language support it directly.

Finally, remove the restriction that patterns can only use extractors and expressions can only use constructors, and replace `(extractor ...)` with `(rule ...)`. Optionally, we could add an `inline` flag on terms and use it for all current extractors to ensure we still generate the same code, since internal extractors are currently inlined automatically. It is difficult to define inlining for all current combinations of ISLE features, though.

#### Alternatives

There are so many alternatives. Let's talk about it…

cfallin:
I definitely think this would be a welcome change! As you know I've been saying similar things for a while; my take is that extractors can be lowered somehow to a sugar on constructors and if-lets, now that we have them. This kind of fulfills the original dream (that I had had, anyway!) of a simple core Prolog-like semantics of matching steps in the backtracking phase (originally LHS) prior to committing.

My take on a core calculus would be:
- Rewrite any mention of a term in the LHS into a dual (that can be specified) that flips args and returns -- "apply" to "unapply".
- Turn this application into an if-let clause.
- Build surface syntax-level desugaring of the `extractor` definition form into a definition of an ordinary rule and the "dual" declaration.

[...]

Issue #6128, opened 2023-03-30
ISLE: Support commutative matches


jameysharp:
#### Feature

An ISLE pattern like `(iadd (pattern_a) (pattern_b))` should also match if `(iadd (pattern_b) (pattern_a))` would match, because `iadd` is a commutative operation.

#### Benefit

We currently have many rules, both in mid-end optimization and in backend lowering, where we have to write out all possible rules that are equivalent up to commutativity. If there are N commutative patterns in a rule's left-hand side, that requires writing up to 2^N rules. This is tedious and error-prone.

It's worthwhile having ISLE generate code for each distinct form of these rules because then the ISLE compiler can build an efficient pattern-matching tree that evaluates each part of the pattern as few times as possible.

In the mid-end optimization rules, we've considered rewriting commutative instructions to their flipped equivalents, so that all commutative alternatives are available in the same e-class (see #5546). But it's cheaper to have more ISLE rules than to double the number of instructions in e-classes, the GVN map, and the dataflow graph. And this doesn't help backends because the egraph cost function can pick either version, so there's no guarantee the backend will see the operands in the order it wants (see #5709).

If we can declare once that a term like `iadd` is commutative and have ISLE generate all the equivalent rules, then we can get the performance advantages of an optimized pattern-matching tree while writing a minimum set of rules.

#### Challenges

These commutative alternative rules often overlap with each other. If the sub-patterns overlap each other, then swapping them will produce a rule which overlaps with the original. The simplest example is that `(iadd x y)` overlaps with `(iadd y x)` since `x` and `y` are wildcard patterns which only bind variables.

In some cases, swapping the operands produces exactly the same rule. For example, `(iadd x x)` only matches if both operands are the same SSA value or e-class, and matches exactly the same inputs when flipped. In these cases, generating a duplicate lowering rule will lead to either an error during overlap checking or an assertion failure during code generation, depending on where we expand the rule. Generating a duplicate egraph rule will produce correct results but do more work than necessary.

In other cases, swapping the operands matches the same inputs when flipped but the right-hand side could evaluate to different results depending on which version of the rule is selected. A rule which lowers `(iadd x y)` to `(x64_add x y)` would, when flipped, lower to `(x64_add y x)` instead. The result should be equivalent but is not the same machine instruction. We need to ensure that we choose one of these alternatives deterministically.

#### Implementation

We've discussed several alternative implementation plans, and I'm probably missing some. None have yet emerged as the "right" answer.

[...]

PR #8653, opened 2024-05-19
cranelift: Optimize __multi3-style multiplications

### LINKED ISSUE #4077: __multi3 optimization
#### Feature

Implement an optimization pass that would eliminate the `__multi3` function from WASM binary during JIT by replacing it with ISA specific (mainly for `x86_64` and `arm64`) sequences, and then inline such sequences into callsites that would allow further optimizations

#### Benefit

A lot of code dealing with cryptography would benefit form faster full width `u64` multiplications where such `__multi3` arises

#### Implementation

If someone would give a few hints about where to start I'd try to implement it by myself

#### Alternatives

Not that I'm aware of. Patching into calling come native library function is a huge overhead for modern CPUs (4 cycles for `x86_64` for e.g. `mulx` or `mul`), and while it would be faster most likely, it's still far from optimal case on a hot path

As an example a simple multiply-add-carry function like `a*b + c + carry -> (high, low)` that accumulates into `u128` without overflows compiles down to the listing below, and it can be a good test subject (transformed from `wasm` into `wat`, may be not the best readable)

```
(module
  (type (;0;) (func (param i32 i64 i64 i64 i64)))
  (func $mac (type 0) (param i32 i64 i64 i64 i64)
    (local i32)
    global.get $__stack_pointer
    i32.const 16
    i32.sub
    local.tee 5
    global.set $__stack_pointer
    local.get 5
    local.get 2
    i64.const 0
    local.get 1
    i64.const 0
    call $__multi3
    local.get 0
    local.get 5
    i64.load
    local.tee 2
    local.get 3
    i64.add
    local.tee 3
    local.get 4
    i64.add
    local.tee 4
    i64.store
    local.get 0
    local.get 5
    i32.const 8
    i32.add
    i64.load
    local.get 3
    local.get 2
    i64.lt_u
    i64.extend_i32_u
    i64.add
    local.get 4
    local.get 3
    i64.lt_u
    i64.extend_i32_u
    i64.add
    i64.store offset=8
    local.get 5
    i32.const 16
    i32.add
    global.set $__stack_pointer
  )
  (func $__multi3 (type 0) (param i32 i64 i64 i64 i64)
    (local i64 i64 i64 i64 i64 i64)
    local.get 0
    local.get 3
    i64.const 4294967295
    i64.and
    local.tee 5
    local.get 1
    i64.const 4294967295
    i64.and
    local.tee 6
    i64.mul
    local.tee 7
    local.get 5
    local.get 1
    i64.const 32
    i64.shr_u
    local.tee 8
    i64.mul
    local.tee 9
    local.get 3
    i64.const 32
    i64.shr_u
    local.tee 10
    local.get 6
    i64.mul
    i64.add
    local.tee 5
    i64.const 32
    i64.shl
    i64.add
    local.tee 6
    i64.store
    local.get 0
    local.get 10
    local.get 8
    i64.mul
    local.get 5
    local.get 9
    i64.lt_u
    i64.extend_i32_u
    i64.const 32
    i64.shl
    local.get 5
    i64.const 32
    i64.shr_u
    i64.or
    i64.add
    local.get 6
    local.get 7
    i64.lt_u
    i64.extend_i32_u
    i64.add
    local.get 4
    local.get 1
    i64.mul
    local.get 3
    local.get 2
    i64.mul
    i64.add
    i64.add
    i64.store offset=8
  )
  (table (;0;) 1 1 funcref)
  (memory (;0;) 16)
  (global $__stack_pointer (mut i32) i32.const 1048576)
  (global (;1;) i32 i32.const 1048576)
  (global (;2;) i32 i32.const 1048576)
  (export "memory" (memory 0))
  (export "mac" (func $mac))
  (export "__data_end" (global 1))
  (export "__heap_base" (global 2))
)
```

jameysharp:
LLVM's `__multi3` function works by splitting a wide multiplication into several narrower ones. This optimization recognizes the algebraic identities involved and merges them back into the original wide multiply.

This is not yet done but illustrates how part of the optimization can work, at least.

Currently, the lower half of the result is optimized into a single `imul` instruction, but most of the intermediate values that are optimized away there are still used in computing the upper half, so elaboration brings them back later.

Fixes #4077

### DIFF
--- a/cranelift/codegen/src/opts/arithmetic.isle
+++ b/cranelift/codegen/src/opts/arithmetic.isle
@@ -229,6 +229,37 @@
       (if-let $true (u64_eq k (ty_bits_u64 half_ty)))
       (uextend ty (umulhi half_ty x y)))
 
+(rule (simplify (iadd ty (imul _ a0 b0) (ishl _ (iadd _ x y) (iconst_u _ half))))
+      (wide_mul ty a0 b0 x y half))
+(rule (simplify (iadd ty (ishl _ (iadd _ x y) (iconst_u _ half)) (imul _ a0 b0)))
+      (wide_mul ty a0 b0 x y half))
+
+(decl multi wide_mul (Type Value Value Value Value u64) Value)
+(rule (wide_mul ty a0 b0 x y half)
+      (if-let $true (u64_eq (ty_bits ty) (u64_shl half 1)))
+      (if-let lo_mask (u64_sub (u64_shl 1 half) 1))
+      (if-let (band _ a (iconst_u _ lo_mask)) a0)
+      (if-let (band _ b (iconst_u _ lo_mask)) b0)
+      (if-let (IsCross.Result a1 b1) (is_cross a0 b0 x y))
+      (if-let (ushr _ a (iconst_u _ half)) a1)
+      (if-let (ushr _ b (iconst_u _ half)) b1)
+      (imul ty a b))
+
+(type IsCross (enum (Result (a1 Value) (b1 Value))))
+(decl pure multi is_cross (Value Value Value Value) IsCross)
+(rule (is_cross a0 b0 x y)
+      (if-let a1 (is_mul_by b0 x))
+      (if-let b1 (is_mul_by a0 y))
+      (IsCross.Result a1 b1))
+(rule (is_cross a0 b0 x y)
+      (if-let a1 (is_mul_by b0 y))
+      (if-let b1 (is_mul_by a0 x))
+      (IsCross.Result a1 b1))
+
+(decl pure multi is_mul_by (Value Value) Value)
+(rule (is_mul_by x (imul _ x y)) y)
+(rule (is_mul_by x (imul _ y x)) y)
+
 ;; Cranelift's `fcvt_from_{u,s}int` instructions are polymorphic over the input
 ;; type so remove any unnecessary `uextend` or `sextend` to give backends
 ;; the chance to convert from the smallest integral type to the float. This
--- a/cranelift/filetests/filetests/egraph/multi3.clif
+++ b/cranelift/filetests/filetests/egraph/multi3.clif
@@ -1,0 +1,103 @@
+test optimize precise-output
+set opt_level=speed_and_size
+target x86_64
+
+; v3  = a_lo
+; v4  = a_hi
+; v5  = b_lo
+; v6  = b_hi
+; v11 = a0
+; v14 = a1
+; v9  = b0
+; v17 = b1
+function %multi3(i64 vmctx, i64, i32, i64, i64, i64, i64) fast {
+    gv0 = vmctx
+    gv1 = load.i64 notrap aligned readonly gv0+80
+
+block0(v0: i64, v1: i64, v2: i32, v3: i64, v4: i64, v5: i64, v6: i64):
+    v7 = iconst.i64 0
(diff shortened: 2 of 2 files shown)


jameysharp:
> I don't know if it'd help at all, but cc https://github.com/bytecodealliance/wasmtime/pull/7719/files#diff-2041f67049d5ac3d8f62ea91d3cb45cdb8608d5f5cdab988731ae2addf90ef01 which will convert larger `mul`s back to smaller `mul`s and `mulhi`s. So I think if you can recognize the open-coded multiplication and turn it into a larger cranelift multiplication, the rest might just work.

Yeah, this definitely felt similar to that, so I placed my new rules next to those `mulhi` rules. I don't see a way to simplify these rules in terms of those, though. The hard part is recognizing that four multiplies, two adds with overflow checks, and some other instructions, all together are equivalent to a single `mulhi`. I suspect that by the time we can recognize all those pieces, it's easier to just emit the `umulhi` instruction than it is to emit a couple `uextend`s, an `imul`, and an `ireduce`. But if you can see a way to do it more simply I'd love that!

Thinking about this comment, though, led me to realize that once we can do that, ISLE is capable enough today to transform a sequence like this:

```
i = umulhi a_lo, b_lo
j = imul a_lo, b_hi
k = imul a_hi, b_lo
l = imul a_hi, b_hi
iadd i, j, k, l
```

into something like this:

```
a = iconcat a_hi, a_lo
b = iconcat b_hi, b_lo
x = imul a, b
y = ushr x, width/2
ireduce y
```

If our mid-end rules supported producing instructions with multiple results, the `ushr`/`ireduce` pair would be better as `isplit`, since `isplit` and `iconcat` produce no instructions for 128-bit values.

On x86 at least, an `imul.i128` currently lowers to roughly the first sequence above so I'm not sure this is an improvement, but I hadn't previously been sure it was even possible.

afonso360:
> On x86 at least, an imul.i128 currently lowers to roughly the first sequence above so I'm not sure this is an improvement, but I hadn't previously been sure it was even possible.

It might not be an improvement by itself, but if we recognize some other i128 ops we can probably start applying some optimization rules at the i128 level which would be really neat. For example, we could now recognize `(imul.i128 x (iconst 2))` and turn it into a `(iadd.i128 x x)` which would be hard to do with the other sequences.

jameysharp:
Good point! Except we currently can't deal with 128-bit constants, so rules like `imul x, 2` won't actually fire. We should figure out how to fix that too :sweat_smile: 

The other thing we can't currently do is notice that this sequence has a redundant multiply:

```
a = iconcat a_hi, a_lo
b = iconcat b_hi, b_lo
x = imul a, b
y = imul a_lo, b_lo
```

jameysharp:
> Since there's no 128-bit `iconst` it's `imul.i128 x (uextend i128 (iconst 2))`, but with ... hopefully it'll already work 🤞

Ooh, I hadn't looked at those carefully. That's awesome that `iconst_u` special cases i128! However I see that it doesn't do that for the extractor, only the constructor. Right now it would need an external extractor to match an optional `uextend`, which is yet more motivation for something like or-patterns (#6128 or #8599).

alexcrichton:
This is a relatively old PR at this point and with the [wasm wide-arithmetic proposal](https://github.com/WebAssembly/wide-arithmetic/) in the works that should in theory supplant the need for this in the long-term. If folks are interested in pushing on this in the nearer-term though I'm happy to reopen.

PR #8663, opened 2024-05-20
Update wasm-tools crates

alexcrichton:
This brings a number of changes here and there for updates that have happened to the crates in the meantime.


alexcrichton:
Ok I think I've sussed out all the issues here with wasm-tools so this should be good to go

pchickey (approved):

PR #8664, opened 2024-05-20
cranelift: Always consider sret arguments used

### LINKED ISSUE #8659: Cranelift: `sret` argument without use causes bad SSA to be passed to regalloc
As reported by @bjorn3 just now, we get a compilation panic with the following:

```
test compile precise-output
target aarch64

function u0:11(i64 sret) system_v {
                                block0(v0: i64):
                                    nop
                                    jump block1

                                block1:
                                    nop
@0001                               return
}
```

This was bisected just now to #8438 (first failing in 3befbe52c9c50100300c3a00dac4c082ae5f36d9). @jameysharp, could you take a look? We should backport and do a point-release on the just-released Cranelift as well as @bjorn3 discovered this while trying to upgrade.

jameysharp:
In #8438 we stopped emitting register bindings for unused arguments, based on the use-counts from `compute_use_states`. However, that doesn't count the use of the struct-return argument that's automatically added after lowering when the `rets` instruction is generated in the epilogue. As a result, using a struct-return argument caused register allocation to panic due to the VReg not being defined anywhere.

This commit adds a use to the struct-return argument so that it's always available in the epilogue.

Fixes #8659

### DIFF
--- a/cranelift/codegen/src/machinst/lower.rs
+++ b/cranelift/codegen/src/machinst/lower.rs
@@ -398,7 +398,7 @@ impl<'func, I: VCodeInst> Lower<'func, I> {
         }
 
         // Find the sret register, if it's used.
-        let mut sret_reg = None;
+        let mut sret_param = None;
         for ret in vcode.abi().signature().returns.iter() {
             if ret.purpose == ArgumentPurpose::StructReturn {
                 let entry_bb = f.stencil.layout.entry_block().unwrap();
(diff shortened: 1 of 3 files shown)


cranelift/codegen/src/machinst/lower.rs
```diff
@@ -502,6 +508,10 @@ impl<'func, I: VCodeInst> Lower<'func, I> {
 
         let mut value_ir_uses = SecondaryMap::with_default(ValueUseState::Unused);
 
+        if let Some(sret_param) = sret_param {
+            value_ir_uses[sret_param] = ValueUseState::Once;
```
cfallin, line 512: Could we add a comment here to note it starts at `Once` (for the one use that is the implicit return), but can still be updated below?

cfallin (approved):
LGTM!

cranelift/codegen/src/machinst/lower.rs
jameysharp: Actually, should I just start it at `Multiple` in case there are multiple epilogues? The definition of the sret vreg can't participate in load-sinking or anything so I don't think the value really matters, but I guess it should be "correct" anyway?

cfallin: Oh, yes, that makes more sense actually, I agree. It's unlikely to matter as you say but might as well get the analysis right...

PR #8668, opened 2024-05-21
Add release binaries for x86_64-musl

alexcrichton:
This was requested in bytecodealliance/wasmtime-py#237 and shouldn't cost us too much in terms of CI resources and maintenance overhead.

### DIFF
--- a/ci/build-build-matrix.js
+++ b/ci/build-build-matrix.js
@@ -59,6 +59,11 @@ const array = [
     "os": "ubuntu-latest",
     "target": "x86_64-linux-android",
   },
+  {
+    "build": "x86_64-musl",
+    "os": "ubuntu-latest",
+    "target": "x86_64-unknown-linux-musl",
+  },
 ];
 
 const builds = [];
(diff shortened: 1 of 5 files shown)


elliottt (approved):

whitequark:
Thanks again =^__^=

This will improve the compatibility story for YoWASP and the Glasgow Interface Explorer to even more Linux distributions.

aambrose-sonatus:
Hi @alexcrichton, I'm curious about a couple things here regarding musl support:
1. What is the wasmtime dependency on libgcc? Is this something core, or just how the toolchain is structured right now?
2. What is the limitation that is preventing wasmtime from being fully static with musl libraries? Is it the above libgcc dep?

Thank you

bjorn3:
For usage in C we compile a wasmtime.so. Producing a dynamic library is incompatible with statically linking libc. The musl based binaries also seem to be meant for usage on musl based distros where statically linking libc only has disadvantages. (bigger binaries, not getting security updates to libc)

aambrose-sonatus:
Is there anything that precludes compiling the static library version of the embedding (IE `libwasmtime.a`) as fully static? Perhaps not as an official release artifact, but e.g. when compiling from source oneself or when using wasmtime as a crate dependency for some other binary?

bjorn3:
If you compile it yourself, I don't see why it shouldn't work.

aambrose-sonatus:
Yeah it seemed to be working on my end; I used zig as the C toolchain so I get full static outputs by default. My questions were more regarding the statement `they are not suitable for "run on any linux distribution" style use cases` and whether this applied to all source builds, or only the publicly released artifacts. It sounds like it's the latter.

Thank you both!

PR #8671, opened 2024-05-21
wasi-http: add the port to authority when opening a TCP connection

iawia002:
#8563 removes the port from the authority, it makes the target address cannot be correctly resolved when opening a TCP connection. This results in that if we want to send an HTTP request in the wasi component, an error `ErrorCode::ConnectionRefused` will be reported directly. The underlying error is `invalid socket address`:

Screenshot 2024-05-21 at 17 21 56

/cc @alexcrichton @elliottt

### DIFF
--- a/crates/test-programs/src/bin/api_proxy_forward_request.rs
+++ b/crates/test-programs/src/bin/api_proxy_forward_request.rs
@@ -1,0 +1,42 @@
+use test_programs::wasi::http::types::{
+    Headers, IncomingRequest, OutgoingBody, OutgoingResponse, ResponseOutparam,
+};
+
+struct T;
+
+test_programs::proxy::export!(T);
+
+impl test_programs::proxy::exports::wasi::http::incoming_handler::Guest for T {
+    fn handle(request: IncomingRequest, outparam: ResponseOutparam) {
+        let res = test_programs::http::request(
+            request.method(),
+            request.scheme().unwrap(),
+            request.authority().unwrap().as_str(),
+            request.path_with_query().unwrap().as_str(),
+            None,
+            None,
+            None,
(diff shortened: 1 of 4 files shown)


bjorn3:
What if the program explicitly specified a port different from 80 or 443? Wouldn't this change break that by creating a nonsensical address like `127.0.0.1:1234:80` when trying to request `http://127.0.0.1:1234/some_page`?

iawia002:
> What if the program explicitly specified a port different from 80 or 443? Wouldn't this change break that by creating a nonsensical address like `127.0.0.1:1234:80` when trying to request `http://127.0.0.1:1234/some_page`?

Indeed, I didn't consider that. I will update it again.

alexcrichton:
Does this mean that `default_send_request_handler` basically doesn't work by default? If so, that seems bad!

Would you be up for adding a test for this?

morenol:
> Does this mean that `default_send_request_handler` basically previously didn't work by default? If so, that seems bad!
> 
> Would you be up for adding a test for this?

I think that that is the case, can we have a fix like this one on a 21.0.1 version?

iawia002:
I have added a test for this case, following @elliottt's advice in https://github.com/bytecodealliance/wasmtime/pull/8676#discussion_r1609131895, PTAL @alexcrichton @elliottt

alexcrichton (approved):
Thanks!

elliottt:
Could you remove the use of ssl in your test program? CI failed because it's not supported on riscv64.

alexcrichton:
Ah I went ahead and just pushed up a commit to ignore s390x/riscv64, I don't think it's too critical to get all the architectures tested here

PR #8672, opened 2024-05-21
Implement `Component::define_unknown_imports_as_traps`

### LINKED ISSUE #8645: Give `wasmtime::component::Linker` ability to stub out implementations
#### Feature

Several times I've simply wanted the linker to stub out all implementations for a component that it has not yet seen. 

Essentially I've wanted to do roughly the following:

```rust
let mut linker = wasmtime::component::Linker::<MyData>::new(&engine);
wasmtime_wasi::add_to_linker_sync(&mut linker)?;
linker.stub_imports(&component)?;
```

With the above configuration, any imports configured by `wasmtime_wasi` would work correctly, but any other imports would simply trap when invoked.

#### Benefit

This would provide users a simple way to get the linker to be happy and turn instantiation errors into runtime errors. This can be useful in cases where you want to instantiate a component and call one of its exports, and you are fine with calls to certain imports to simply trap if they occur.

#### Implementation

The implementation should be fairly straight forward I believe through the use of `Component::env_component` to do the necessary reflection to know when to call `LinkerInstance::func_new` with a closure that calls `panic!("stubbed function called")`.

#### Future possibilities

In the future, we could expose an additional API that would allow the user to supply a closure that is invoked on each import invocation so that the user can customize the behavior of the stub.

rylev:
Fixes #8645

### DIFF
--- a/crates/wasmtime/src/runtime/component/linker.rs
+++ b/crates/wasmtime/src/runtime/component/linker.rs
@@ -336,6 +336,64 @@ impl<T> Linker<T> {
             .instantiate_async(store)
             .await
     }
+
+    /// Implement any imports of the given [`Component`] with a function which traps.
+    ///
+    /// By default a [`Linker`] will error when unknown imports are encountered when instantiating a [`Component`].
+    /// This changes this behavior from an instant error to a trap that will happen if the import is called.
+    pub fn define_unknown_imports_as_traps(&mut self, component: &Component) -> Result<()> {
+        let types = component.types();
+        let component = component.env_component();
+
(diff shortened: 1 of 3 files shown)


alexcrichton (commented):
Thanks! Could this be written with a recursive helper that operates on a `&ComponentType` (or similar) and a `LinkerInstance`? That'll help remove some minor duplication here and additionally handle the fully general case of all component imports.

Additionally could this return an error for unsupported imports? For example this can't fill in the import of a component or import of a module at this time so I think those should return an error rather than being silently ignored.

And finally, mind adding a test or two for this?

alexcrichton (approved):

PR #8676, opened 2024-05-21
chore: add test to http request without specifying port

morenol:


crates/wasi-http/tests/all/main.rs
```diff
@@ -527,6 +527,29 @@ async fn do_wasi_http_echo(uri: &str, url_header: Option<&str>) -> Result<()> {
...
 
+#[test_log::test(tokio::test)]
+async fn wasi_http_without_port() -> Result<()> {
+    do_http_request_without_port().await
+}
+
+async fn do_http_request_without_port() -> Result<()> {
+    let request = hyper::Request::builder()
+        .method(http::Method::GET)
+        .uri("http://example.com/test-path");
```
elliottt, line 538: Could you add an additional request for a path that will return a 200?

morenol: Sure, for some reason this is returning `404` in the CI, any idea why?

iawia002:
> Thanks! To land this though it'll need to pass CI, so mind folding this into #8671?

I am back online, @morenol do you mind if I cherry-pick this commit into #8671 and square them into one commit? I hope we can land this fix soon.

morenol:
I don't mind but I think that this test is not accurate, feel free to check and create a test that could reproduce the current issue

morenol:
Closed since a similar test was added in #8671

PR #8677, opened 2024-05-22
Use BuildHasher::hash_one

scottmcm:
I noticed that a bit of #8498 could use the relatively-new `hash_one` helper, so here's a PR.

Up to you whether you want the change; it does exactly the same thing.


alexcrichton (approved):
Thanks!

PR #8680, opened 2024-05-22
Refactor how release notes are managed

alexcrichton:
This commit updates how Wasmtime manages release notes across released versions of Wasmtime. One of the most onerous parts about releases right now is writing release notes in all the locations and making sure they're all up-to-date and in-sync. This is inevitably forgotten in some cases and various pieces will slip through the cracks. The basic idea of this PR is to change our release notes to only document the release branch that they're on. All historical release notes are relegated to historical branches.

With this change there's no longer any need to backport or forward-port release notes for any changes. Instead release notes are written once on one branch and that's it.

The major downside of this change is that there's no easy way to get a bird's eye view of all release notes for Wasmtime any more. If necessary that could theoretically be automated in the future (like https://releases.rs/), but for now this feels like an acceptable compromise to make releases much easier.

The contents of this PR are to update `RELEASES.md` with back-links to historical release notes as well as the various pieces of automation we have about managing release notes.


RELEASES.md
```diff
@@ -10,4060 +8,37 @@ Unreleased.
...
-  [#974](https://github.com/bytecodealliance/wasmtime/pull/974)
-
-* Automated releases for tags should be up and running again, along with
-  automatic publication of the `wasmtime` Python package.
-  [#971](https://github.com/bytecodealliance/wasmtime/pull/971)
+Release notes for previous releases of Wasmtime can be found on the respective
+release branches of the Wasmtime repository.
+
+<!-- ARCHIVE_START -->
+* [21.0.0](https://github.com/bytecodealliance/wasmtime/blob/release-21.0.0/RELEASES.md)
```
cfallin, line 15: I *think* we don't need to include point-releases, since there is no separate branch to link to (and the `release-X.0.0` branch will contain all updates from `X.Y.Z`); but should we note somewhere, e.g. list the point-releases here in the link text or after the link?

Otherwise I could imagine some confusion: I'm curious what was released in 21.0.1, I look at this list, I don't see the version number and get confused (or grep for it, same result).

.github/workflows/release-process.yml
```diff
@@ -63,7 +63,15 @@ jobs:
           ./publish bump
           num=$(./ci/print-current-version.sh)
 
-          # Add a new section to the release notes for the new version.
+          # Remove all release notes for the currnet version now that the
```
cfallin, line 66: s/currnet/current/

cfallin (approved):
This seems reasonable, with one question below about point-releases. Thanks!

RELEASES.md
cfallin: Alternately: name the link text something like "21.x releases" or "21.0.0 and following" or ...

alexcrichton: I actually originally had this written here as 21.0.x but I ended up renaming since it was nontrivial to create the string "21.0.x" from the string "21.0.0" available in the release scripts in github actions. I think you're right though, I'll try to switch this to "21.0.x" (along with everything below), to make it more clear that it's for all those releases.

alexcrichton: I also regret choosing the release branch name as `release-N.0.0` originally since that'd be more appropriate as `release-N.0.x` as well...

PR #8686, opened 2024-05-23
Use `[su]extend_maybe` in the `iconst_[su]` extractors

scottmcm:
This enables various things -- like `x * 2` → `x + x` and `x * -1` → `-x` -- for 128-bit numbers as well.

Thanks to @jameysharp over in https://github.com/bytecodealliance/wasmtime/pull/8653#issuecomment-2125435660 for noticing that these extractors weren't ever going to match 128-bit numbers as-written, even though the rules could produce 128-bit values.


jameysharp (approved):
Very nice, thank you!

PR #8689, opened 2024-05-24
wasmtime: `VMRuntimeLimits` pointer is already SSA

jameysharp:
We don't need to use cranelift-frontend to keep track of the location of the `VMRuntimeLimits` pointer, because it is loaded once on entry to the function and never changed afterward.

Instead, just track the correct SSA `Value`.

### DIFF
--- a/crates/cranelift/src/func_environ.rs
+++ b/crates/cranelift/src/func_environ.rs
@@ -7,6 +7,7 @@ use cranelift_codegen::ir::pcc::Fact;
 use cranelift_codegen::ir::types::*;
 use cranelift_codegen::ir::{ArgumentPurpose, Function, InstBuilder, MemFlags};
 use cranelift_codegen::isa::{TargetFrontendConfig, TargetIsa};
+use cranelift_entity::packed_option::ReservedValue;
 use cranelift_entity::{EntityRef, PrimaryMap, SecondaryMap};
 use cranelift_frontend::FunctionBuilder;
 use cranelift_frontend::Variable;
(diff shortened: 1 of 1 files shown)


crates/cranelift/src/func_environ.rs
```diff
@@ -284,14 +285,13 @@ impl<'module_environment> FuncEnvironment<'module_environment> {
...
         let pointer_type = self.pointer_type();
-        builder.declare_var(self.vmruntime_limits_ptr, pointer_type);
         let vmctx = self.vmctx(builder.func);
         let base = builder.ins().global_value(pointer_type, vmctx);
         let offset = i32::try_from(self.offsets.vmctx_runtime_limits()).unwrap();
         let interrupt_ptr = builder
             .ins()
             .load(pointer_type, ir::MemFlags::trusted(), base, offset);
-        builder.def_var(self.vmruntime_limits_ptr, interrupt_ptr);
+        self.vmruntime_limits_ptr = interrupt_ptr;
```
fitzgen, line 294: Can we debug assert that `self.vmruntime_limits_ptr` is the reserved value before we initialize it, to make sure that we aren't defining it multiple times? And/or use an `Option` or `PackedOption` for the field?

crates/cranelift/src/func_environ.rs
```diff
@@ -457,12 +457,9 @@ impl<'module_environment> FuncEnvironment<'module_environment> {
...
     /// Returns the `(address, offset)` of the fuel consumption within
     /// `VMRuntimeLimits`, used to perform loads/stores later.
-    fn fuel_addr_offset(
-        &mut self,
-        builder: &mut FunctionBuilder<'_>,
-    ) -> (ir::Value, ir::immediates::Offset32) {
+    fn fuel_addr_offset(&mut self) -> (ir::Value, ir::immediates::Offset32) {
         (
-            builder.use_var(self.vmruntime_limits_ptr),
+            self.vmruntime_limits_ptr,
```
fitzgen, line 462: Similarly, can we debug assert that it has been initialized here?

fitzgen (approved):

PR #8691, opened 2024-05-25
Fix runtime::vm::cow tests on hosts with a 16k page size

refi64:
A bunch of the hardcoded sizes were only multiples of 4k, so they'd crash on hosts with a different page size. Fix this by making them all based on the runtime page size instead.


alexcrichton (approved):
Thanks for this!

PR #8693, opened 2024-05-25
Add basic support for DWARF processing with components

alexcrichton:
This commit updates the native-DWARF processing (the `-D debug-info` CLI flag) to support components. Previously component support was not implemented and if there was more than one core wasm module within a component then dwarf would be ignored entirely.

This commit contains a number of refactorings to plumb a more full compilation context throughout the dwarf processing pipeline. Previously the data structures used only were able to support a single module. A new `Compilation` structure is used to represent the results of an entire compilation and is plumbed through the various locations. Most of the refactorings in this commit were then to extend loops to loop over more things and handle the case where there is more than one core wasm module.

I'll admit I'm not expert on DWARF but basic examples appear to work locally and most of the additions here seemed relatively straightforward in terms of "add another loop to iterate over more things" but I'm not 100% sure how well this will work. In theory this now supports concatenating DWARF sections across multiple core wasm modules, but that's not super well tested.

### DIFF
--- a/tests/all/debug/lldb.rs
+++ b/tests/all/debug/lldb.rs
@@ -356,7 +356,7 @@ check: exited with status = 0
     #[test]
     #[ignore]
     fn dwarf_simple() -> Result<()> {
-        for wasm in [DWARF_SIMPLE] {
+        for wasm in [DWARF_SIMPLE, DWARF_SIMPLE_COMPONENT] {
             test_dwarf_simple(wasm, &[])?;
         }
         Ok(())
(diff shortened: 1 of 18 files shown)


tests/all/debug/lldb.rs
```diff
@@ -295,3 +298,58 @@ check: exited with status = 0
...
+    foreach_dwarf!(assert_test_exists);
+
+    #[test]
+    #[ignore]
+    #[cfg(all(
+        any(target_os = "linux", target_os = "macos"),
+        target_pointer_width = "64"
+    ))]
+    fn dwarf_simple() -> Result<()> {
+        for wasm in [DWARF_SIMPLE, DWARF_SIMPLE_COMPONENT] {
```
elliottt, line 322: Nice!

elliottt (approved):
Awesome!

PR #8698, opened 2024-05-28
Allow empty block param lists in clif files

elliottt:
Allow block definitions to be parsed with empty parens, rather than requiring that there be a non-empty list of arguments if they're present.

```
function %f() {
block0():
    trap ud0
}
```

This is a pretty superficial change, but when iterating on clif files, it is nice to have to make fewer changes to get to a file that can make it through the pipeline.


cfallin (approved):
Very reasonable ergonomics tweak, thanks!

PR #8701, opened 2024-05-28
riscv64: Special-case `f32const 0` and `f64const 0`

alexcrichton:
This commit is inspired by discussion on #8695 which made me remember the discussion around #7162 historically. In lieu of a deeper fix for the issue of "why can't `iconst 0` use `(zero_reg)`" it's still possible to add special-cases to rules throughout the backend so this commit does that for generating zero-value floats.


afonso360 (approved):
Nice! Thanks!

alexcrichton:
> Received request to deprovision: The request was cancelled by the remote provider.

After 4 times in a row I'm going to deprovision myself for tonight.

PR #8702, opened 2024-05-29
Remove the borrow checking from `wiggle` entirely

alexcrichton:
This commit is a refactoring of the `wiggle` crate which powers the `*.witx`-based bindings generation of Wasmtime for wasip1 support. Originally `wiggle` had a full-blown runtime borrow checker which verified that borrows were disjoint when appropriate. In #8277 this was removed in favor of a more coarse "either all shared or all mutable" guarantee. It turns out that this exactly matches what the Rust type system guarantees at compile time as well.

This commit removes all runtime borrow checking in favor of compile-time borrow checking instead. This means that there is no longer the possibility of a runtime error arising from borrowing errors. Current bindings in Wasmtime needed no restructuring to work with this new API.

The source of the refactors here are all in the `wiggle` crate. Changes include:

* The `GuestPtr` type lost its type parameter. Additionally it only contains a `u32` pointer now instead.
* The `GuestMemory` trait is replaced with a simple `enum` of possibilities.
* Helper methods on `GuestPtr` are all moved to `GuestMemory`.
* A number of abstractions were simplified now that borrow checking is no longer necessary.
* Generated trait methods now all take `&mut GuestMemory<'_>` as an argument.

These changes were then propagated to the `wasmtime-wasi` and `wasi-common` crates in their preview0 and preview1 implementations of WASI. All changes are just general refactors, no functional change is intended here.


crates/wasi-common/src/snapshots/preview_0.rs
```diff
@@ -521,283 +555,149 @@ impl wasi_unstable::WasiUnstable for WasiCtx {
     // The implementations are identical, but the `types::` in scope locally is different.
     // The bodies of these functions is mostly about converting the GuestPtr and types::-based
     // representation to a std::io::IoSlice(Mut) representation.
+    //
+    // TODO: update this ^
```
pchickey, line 559: Docs TODO

crates/wiggle/macro/src/lib.rs
```diff
@@ -168,7 +168,30 @@ pub fn from_witx(args: TokenStream) -> TokenStream {
         quote!()
     };
 
-    TokenStream::from(quote! { #code #metadata })
+    let mut ret = quote! { #code #metadata };
+
+    if std::env::var("WIGGLE_DEBUG_BINDGEN").is_ok() {
```
pchickey, line 173: Wish I had done this years ago

pchickey (approved):
We sure did spend turn a lot of CPU cycles into heat by dynamically checking a property that was statically provable all these years, didnt we

crates/wiggle/macro/src/lib.rs
alexcrichton: I should wrap this up in a helper as well since I'm starting to litter this all over the macros I work on...

iximeow:
to be fair to ourselves, when we thought embedders might provide functions that involve multiple concurrent outstanding borrows it all made a bit more sense :)

acfoltzer:
I'm so glad we're able to do this now, thank you!!

PR #8703, opened 2024-05-29
riscv64: Improve base select lowering

afonso360:
👋 Hey,

This PR is an attempt at improving the base lowering for the `select` instruction in the RISC-V backend. It reduces one instruction from the current lowering.

Currently our selects have the following instructions:
```asm
b{cond} rc, 0xc
mv rd, rx
j 8
mv rd, ry
```

With this change we now produce the following:
```asm
mv rd, rx
b{cond} rc, 8
mv rd, ry
```

So, we lose the jump instruction, but one of the moves becomes unconditional. This to me feels like it's worth it, but I have no data to prove one way or the other, and I think if i try to measure it with sightglass it probably won't detect this change.

One of the benefits of this lowering is that now if one of the input registers is the same as the destination register we can do a two instruction select, like so (assuming `rx = rd`):
```asm
b{cond} rc, 8
mv rd, ry
```

I've had to do some regalloc surgery to make that possible which makes some other cases get worse regalloc.

This is built on top of #8695, but only the last commit is relevant.


cranelift/filetests/filetests/isa/riscv64/umax.clif
```diff
@@ -87,12 +84,15 @@ block0(v0: i64, v1: i64):
...
-;   select a0,a0,a1##condition=(a0 ugt a1)
+;   mv a4,a0
+;   select a0,a4,a1##condition=(a4 ugt a1)
 ;   ret
 ;
 ; Disassembled:
 ; block0: ; offset 0x0
-;   bltu a1, a0, 8
+;   mv a4, a0
+;   mv a0, a4
```
afonso360, line 94: We now sometimes get this funky output from regalloc. I think I understand why it's doing this, but maybe there's some way to avoid this?

---

As a side note, maybe we could force the destination register to always be the same as one of the inputs. We would always emit the 2 instruction select and let regalloc emit the move beforehand if it finds it beneficial.

I'm not sure if this is better or not.

abrown:
I'm positive toward this change but someone more familiar with RISC-V should probably review this?

cfallin (approved):
Last commit LGTM, thanks! Happy to land once #8695 does and this rebases.

PR #8711, opened 2024-05-29
Enable rustc's `unused-lifetimes` lint

alexcrichton:
This is allow-by-default doesn't seem to have any false positives in Wasmtime's codebase so enable it by default to help clean up vestiges of old refactorings.


jameysharp (approved):
Ooh, nice!

PR #8716, opened 2024-05-30
Allow graphic ASCII characters in symbol names

lpereira:
The original idea of the filtering here was to avoid passing characters such as the NUL byte, whitespaces, or other things that could cause issues to tools that will consume this data.  Since it's impractical to update the list of allowed characters every time a language that
compiles to Wasm generates symbols slightly differently, allow all graphic ASCII characters instead.

(Tests don't need updating.)

### DIFF
--- a/crates/wasmtime/src/compile.rs
+++ b/crates/wasmtime/src/compile.rs
@@ -413,11 +413,12 @@ impl<'a> CompileInputs<'a> {
         /// Maximum length of symbols generated in objects.
         const MAX_SYMBOL_LEN: usize = 96;
 
-        // Just to be on the safe side, avoid passing user-provided data to tools
-        // like "perf" or "objdump", and filter the name.  Let only characters usually
-        // used for function names, plus some characters that might be used in name
-        // mangling.
-        let bad_char = |c: char| !c.is_ascii_alphanumeric() && !r"<>[]_-:@$".contains(c);
+        // Just to be on the safe side, filter out characters that could
+        // pose issues to tools such as "perf" or "objdump".  To avoid
+        // having to update a list of allowed characters for each different
+        // language that compiles to Wasm, allows only graphic ASCII
+        // characters; replace runs of everything else with a "?".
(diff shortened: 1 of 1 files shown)


jameysharp:
The golang function names I've seen have both `.` and `/` so it would be nice to allow both. On the other hand the original concern in #8627 was whether any downstream tools might mishandle these names in security-sensitive ways. Since both of these characters are significant in paths I wonder if we should worry about them?

lpereira:
> The golang function names I've seen have both `.` and `/` so it would be nice to allow both. On the other hand the original concern in #8627 was whether any downstream tools might mishandle these names in security-sensitive ways. Since both of these characters are significant in paths I wonder if we should worry about them?

While that's true, `$` can be used for shell variable expansion, `<` and `>` for shell redirection, and `@` can be used in Bash arrays. And those are already accepted.

(I wasn't sure about `/`, though, but I can add it too.)

jameysharp:
Yeah, fair point. Maybe the main goal here is to prohibit characters outside the ASCII graphic set so nobody's dealing with nul bytes or whitespace or invalid UTF-8. If so, I'm tempted to use [`u8::is_ascii_graphic`](https://doc.rust-lang.org/std/primitive.u8.html#method.is_ascii_graphic) and call it a day. @alexcrichton, what do you think?

lpereira:
Looking at the docs: yeah, that seems like a good idea.

lpereira:
Alright, pushed a new version that only tests for a bad character if it's not a graphic ASCII character.

alexcrichton (approved):
Sorry yeah I don't want to cause too much trouble/headache with this. My gut is that something will go wrong if random emoji is fed to `objdump` and/or mixing in things like right-to-left and left-to-right characters and things like that, but I've got no concrete evidence and everything might handle it all just fine.

In any case I think adjustments along the lines of this definitely make sense and are fine to land without trying to investigate deeply how tooling reacts to these symbols (since it's probably all fine anyway)

PR #8718, opened 2024-05-31
impl Send for JITModule

MolotovCherry:
`JITModule` currently is `!Send`, which makes it less useful. According to bjorn, it makes sense to make it `Send`.

I'm fairly new to this, but please let me know any issues/concerns you have with this and I'll do my best to address them. Thanks!

Ref: https://bytecodealliance.zulipchat.com/#narrow/stream/206238-general/topic/Cross.20thread.20JITModule/near/441536899

cc @bjorn3

### DIFF
--- a/cranelift/jit/src/backend.rs
+++ b/cranelift/jit/src/backend.rs
@@ -170,14 +170,9 @@ unsafe impl Send for GotUpdate {}
 /// A wrapper that impls Send for the contents.
 ///
 /// SAFETY: This must not be used for any types where it would be UB for them to be Send
+#[derive(Copy, Clone)]
 struct SendWrapper<T>(T);
 unsafe impl<T> Send for SendWrapper<T> {}
-impl<T: Copy> Copy for SendWrapper<T> {}
-impl<T: Clone> Clone for SendWrapper<T> {
-    fn clone(&self) -> Self {
-        Self(self.0.clone())
-    }
-}
 
 /// A `JITModule` implements `Module` and emits code and data into memory where it can be
 /// directly called and accessed.
--- a/cranelift/jit/src/backend.rs
+++ b/cranelift/jit/src/backend.rs
@@ -27,7 +27,7 @@ const READONLY_DATA_ALIGNMENT: u64 = 0x1;
 /// A builder for `JITModule`.
 pub struct JITBuilder {
     isa: OwnedTargetIsa,
-    symbols: HashMap<String, *const u8>,
+    symbols: HashMap<String, SendWrapper<*const u8>>,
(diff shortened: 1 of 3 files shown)


cranelift/jit/src/backend.rs
```diff
@@ -140,7 +140,7 @@ impl JITBuilder {
     /// symbol table. Symbol lookup fn's are called in reverse of the order in which they were added.
     pub fn symbol_lookup_fn(
         &mut self,
-        symbol_lookup_fn: Box<dyn Fn(&str) -> Option<*const u8>>,
+        symbol_lookup_fn: Box<dyn Fn(&str) -> Option<*const u8> + Send>,
```
MolotovCherry, line 143: Seeing that `libcall_names` already originally required `Send + Sync` in `JITBuilder`, I wonder if it would be better to also add `Sync` to this.

cranelift/jit/src/backend.rs
```diff
@@ -191,6 +191,8 @@ pub struct JITModule {
     pending_got_updates: Vec<GotUpdate>,
 }
 
+unsafe impl Send for JITModule {}
```
bjorn3, line 194: Instead of implementing `Send` on `JITModule`, which may cause issues if we add a new non-`Send` field, maybe it would be possible to implement `Send` on each type that doesn't currently like `CompiledBlob` and `PtrLen`. For the `symbols` field you can for example wrap the `*const u8` in a newtype.

cranelift/jit/src/backend.rs
bjorn3: I don't think there is much point in making it `Sync` too as `JITModule` can't be `Sync` anyway. For `libcall_names` it makes sense for it to be `Sync` to be able to pass `cranelift_module::default_libcall_names()`.

cranelift/jit/src/backend.rs
MolotovCherry: Should be mostly good now I think. I made a `SendWrapper` for fields which can be wrapped in it in `JITModule`, and impl'd `Send` on any types that didn't like `CompiledBlob` and `PtrLen`. I also had to wrap all the `NonNull` fields as they were also offenders

cranelift/jit/src/backend.rs
```diff
@@ -165,23 +165,37 @@ struct GotUpdate {
     ptr: *const u8,
 }
 
+unsafe impl Send for GotUpdate {}
+
+/// A wrapper that impls Send for the contents.
+///
+/// SAFETY: This must not be used for any types where it would be UB for them to be Send
+struct SendWrapper<T>(T);
```
bjorn3, line 173: Deriving Copy and Clone should be fine.

alexcrichton (approved):

PR #8719, opened 2024-05-31
riscv64: Optimize storing zero to memory by using `zero_reg`

alexcrichton:
This commit is similar to #8701 in that it adds a special case to `store` operations to use the `zero` register when applicable.

### DIFF
--- a/cranelift/codegen/src/isa/riscv64/inst.isle
+++ b/cranelift/codegen/src/isa/riscv64/inst.isle
@@ -2438,8 +2438,23 @@
     dst))
 
 ;; helper function to store to memory.
-(decl gen_store (AMode StoreOP MemFlags Reg) InstOutput)
-(rule (gen_store amode op flags src)
+;;
+;; This helper contains a special-case for zero constants stored to memory to
+;; directly store the `zero` register to memory. See #7162 for some discussion
+;; on why this doesn't just fall out.
+(decl gen_store (AMode MemFlags Value) InstOutput)
+(rule 1 (gen_store amode flags val @ (value_type ty))
+  (if-let (i64_from_iconst 0) val)
+  (rv_store amode (store_op ty) flags (zero_reg)))
+(rule 0 (gen_store amode flags val @ (value_type ty))
+  (rv_store amode (store_op ty) flags val))
+
+;; Emit a raw instruction to store a register into memory.
+;;
+;; Note that the `src` operand must have the correct type for the `op`
+;; specified.
+(decl rv_store (AMode StoreOP MemFlags Reg) InstOutput)
+(rule (rv_store amode op flags src)
   (side_effect (SideEffectNoResult.Inst (MInst.Store amode op flags src))))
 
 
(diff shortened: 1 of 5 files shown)


afonso360 (approved):
Neat!

I also have a suspicion that the CI failures might be that we are now trying to encode a compressed store with the zero register. I don't remember if this is legal, but I'm going to investigate it a bit.

Edit: Yeah, the [RISC-V Manual (Page 70)](https://riscv.org/wp-content/uploads/2017/05/riscv-spec-v2.2.pdf) states:
> For many RVC instructions, zero-valued immediates are disallowed and x0 is not a valid 5-bit register specifier. These restrictions free up encoding space for other instructions requiring fewer operand bits.

There is nothing specific about the store instructions falling under these restrictions and It's probably why we don't have the check [here](https://github.com/bytecodealliance/wasmtime/blob/34a9ad98264a351ae72928c1fb77d85c1c750d12/cranelift/codegen/src/isa/riscv64/inst/emit.rs#L676-L677). Note the regular stores (non-stack based) should not need an update since they use a compressed register which automatically excludes the zero register.

alexcrichton:
At least one issue I tracked to my misunderstanding of how `istore{8,16,32}` worked. I was inferring the type of store from the operand but that's not correct for those instructions. I'll take a look at the compressed bits now as well.

afonso360:
> My bugfix around my misinterpretation of istore8 and friends fixed CI, so there was no "smoking gun" per se to the compressed loads.

Nice! Good to know.

> We are emitting compressed loads to the stack for x0 though but at least capstone disassembles it so it seems like it might be intended to work?

For loads we shouldn't be, at least we have that specific [check in the emit code](https://github.com/bytecodealliance/wasmtime/blob/d65ccd85cc111ef279c9eb6b02f3aef7a839fb78/cranelift/codegen/src/isa/riscv64/inst/emit.rs#L577-L590)

For stores I would agree, If capstone disassembles it it's probably intended to work. It also seems like a very useful instruction to have, so I wouldn't be surprised that it is intended.

> Do you think I should add a condition to avoid generating the instruction to be safe though to ensure that it doesn't get emitted?

I think this is fine as is. I'm fairly sure that if it wasn't intended the testsuites would fail in a bunch of places.

> By this do you mean that we could use a compressed sw instruction for a f32 store for example? 

No, just that the zero_reg in the emit for `Fsd (float store double)`, means `f0` and that is a perfectly normal register, but since we are not restricting the zero reg, it doesn't matter.

> Turns out i64_from_iconst will match f32const and f64const nodes

That's super weird, and I would not expect that at all.

alexcrichton:
> For loads we shouldn't be, at least we have that specific [check in the emit code](https://github.com/bytecodealliance/wasmtime/blob/d65ccd85cc111ef279c9eb6b02f3aef7a839fb78/cranelift/codegen/src/isa/riscv64/inst/emit.rs#L577-L590)

Oops sorry I misspoke, I meant stores!

PR #8720, opened 2024-05-31
Update `wit-bindgen`, add `features` to `bindgen!`

alexcrichton:
This commit updates the wit-bindgen family of crates and additionally adds a `features` key to the `bindgen!` macro. This brings it in line with `wit-bindgen`'s support which enables usage of the new `@since` and `@unstable` features of WIT.


elliottt (approved):
Looks good!

PR #8715, opened 2024-05-30
wasi: Make WasiCtxBuilder::socket_addr_check closure async

lann:
Fixes #8714

PR #8721, opened 2024-05-31
Always generate the same output structure with `bindgen!`

alexcrichton:
Currently with the `bindgen!` macro when the `with` key is used then the generated bindings are different than if the `with` key was not used. Not only for replacement purposes but the generated bindings are missing two key pieces:

* In the generated `add_to_linker` functions bounds and invocations of `with`-overridden interfaces are all missing. This means that the generated `add_to_linker` functions don't actually represent the full world.

* The generated module hierarchy has "holes" for all the modules that are overridden. While it's mostly a minor inconvenience it's also easy enough to generate everything via `pub use` to have everything hooked up correctly.

After this PR it means that each `bindgen!` macro should, in isolation, work for any other `bindgen!` macro invocation. It shouldn't be necessary to weave things together and remember how each macro was invoked along the way. This is primarily to unblock #8715 which is running into a case where tcp/udp are generated as sync but their dependency, `wasi:sockets/network`, is used from an upstream async version. The generated `add_to_linker` does not compile because tcp/udp depend on `wasi:sockets/network` isn't added to the linker. After fixing that it then required more modules to be generated, hence this PR.

### DIFF
--- a/crates/component-macro/src/bindgen.rs
+++ b/crates/component-macro/src/bindgen.rs
@@ -154,6 +154,7 @@ impl Parse for Config {
                     Opt::Features(f) => {
                         features.extend(f.into_iter().map(|f| f.value()));
                     }
+                    Opt::RequireStoreDataSend(val) => opts.require_store_data_send = val,
                 }
             }
         } else {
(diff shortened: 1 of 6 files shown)


lann (approved):

pchickey (approved):
cc @MendyBerger can you see if this improves any of the code we looked at together earlier this week?

MendyBerger:
@pchickey I'm unable to get the code to compile when I rebase on top of this git rev.
Let me know if it's important that I test it. If so, I'll spend some more time on this. It's probably something minor.

PR #8725, opened 2024-05-31
wasi-http: Add Send bound to WasiHttpView

lann:
This is practically always desired and its absence has caused some downstream integration headaches.


pchickey (approved):

pchickey:
Hopefully CI failure was a flake? trying again

PR #8726, opened 2024-05-31
Make an HTTP test more robust to failure

alexcrichton:
This handles flakiness [seen online] and makes the test more robust. It was known that this test was not "fully robust" when added in #8671 and this is the "work over time" to handle failures we see in the wild.

[seen online]: https://github.com/bytecodealliance/wasmtime/actions/runs/9324271722/job/25669233977


elliottt (approved):
Thanks for the note in the test!

PR #8733, opened 2024-06-02
Update object to 0.36

philipc:
This fixes handling of empty symbols in cranelift-object for Mach-O, which is needed for https://github.com/rust-lang/rustc_codegen_cranelift/issues/1456 (@bjorn3)


alexcrichton (approved):
Thanks for this!

PR #8738, opened 2024-06-03
Set `SO_REUSEADDR` again for `wasmtime serve` on unix

### LINKED ISSUE #7852: `wasmtime serve` stuck when the port is used
I'd expect `wasmtime serve` to error out if the listening addr is already in use.

alexcrichton:
This effectively reverts #7863 which was a misunderstanding on my part about `SO_REUSEADDR`. I think I mixed it up with `SO_REUSEPORT`. Without `SO_REUSEADDR` it's possible to have this happen:

* Start a `wasmtime serve` session
* Connect to the session with `nc`
* Kill the server
* Restarting the server no longer works

Due to the active TCP connection at the time the server was killed the socket/address is in the `TIME_WAIT` state which apparently prevents rebinding the port until the OS has a chance to clean up everything. Setting `SO_REUSEADDR` enables rebinding the address quickly.

Now in #7863 that was trying to fix #7852 which said that it was possible to have multiple `wasmtime serve` instances binding the same port. That can lead to confusion and is generally something we don't want. That being said #7863 didn't actually change anything on either macOS or Unix, although it did fix the issue on Windows. Turns out the fix for both these issues was to only turn off `SO_REUSEADDR` on Windows, but leave it on for Unix.

This PR additionally adds two new tests, both for rebinding an in-use port quickly and additionally ensuring the same port can't be listened to twice.

### DIFF
--- a/src/commands/serve.rs
+++ b/src/commands/serve.rs
@@ -281,15 +281,19 @@ impl ServeCommand {
 
         let instance = linker.instantiate_pre(&component)?;
 
-        // Tokio by default sets `SO_REUSEADDR` for listeners but that makes it
-        // a bit confusing if you run Wasmtime but forget to close a previous
-        // `serve` session. To avoid that we explicitly disable `SO_REUSEADDR`
-        // here.
         let socket = match &self.addr {
             SocketAddr::V4(_) => tokio::net::TcpSocket::new_v4()?,
             SocketAddr::V6(_) => tokio::net::TcpSocket::new_v6()?,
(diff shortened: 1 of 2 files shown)


elliottt (approved):

alexcrichton:
Ok turns out the issue was that we want `SO_REUSEADDR` on Unix, but not Windows. The option means different things on different platforms. The two new tests assert we have the desired behavior on all platforms now, though.

PR #8739, opened 2024-06-03
Update `i64_from_iconst` to only match `iconst`

### LINKED ISSUE #8723: ISLE `i64_from_iconst` extractor can match f32const/f64const
I ran into this bug on https://github.com/bytecodealliance/wasmtime/pull/8719 but the [current definition](https://github.com/bytecodealliance/wasmtime/blob/271a232cac12e72bb55a2cba4ee017a0b43e8198/cranelift/codegen/src/machinst/isle.rs#L224) of `i64_from_iconst` can match the `f32const` and `f64const` nodes in addition to `iconst`. This seems counterintuitive given the name of the extractor, but I'm not sure how to best fix myself.

alexcrichton:
Previously this extractor would additionally match `f32const` and `f64const` which while theoretically not the end of the world can be confusing. Nowadays switch it to instead only matching the `iconst` instruction, as the name implies, and if necessary matching float constants is probably best done through separate ISLE lowering rules.

Closes #8723

### DIFF
--- a/cranelift/codegen/src/machinst/isle.rs
+++ b/cranelift/codegen/src/machinst/isle.rs
@@ -223,7 +223,13 @@ macro_rules! isle_lower_prelude_methods {
         #[inline]
         fn i64_from_iconst(&mut self, val: Value) -> Option<i64> {
             let inst = self.def_inst(val)?;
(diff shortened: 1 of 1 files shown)


elliottt (approved):
Seems reasonable to me!

PR #8742, opened 2024-06-04
Add a fuzz target for exercising bounds checks with various memory configs

fitzgen:


crates/fuzzing/src/oracles/memory.rs
```diff
@@ -0,0 +1,316 @@
...
+            );
+        }
+    };
+
+    do_accesses(&mut store, "initial size");
+    let _ = memory.grow(&mut store, u64::from(growth));
+    do_accesses(&mut store, "after growing");
+}
+
+fn build_wasm(image: &HeapImage, offset: u32) -> Vec<u8> {
```
alexcrichton, line 186: Do you think it's worth doing this in the text format instead of expanding out all the wasm-encoder bits?

crates/fuzzing/src/generators/memory.rs
```diff
@@ -5,6 +5,75 @@ use arbitrary::{Arbitrary, Unstructured};
...
+/// A description of a memory config, image, etc... that can be used to test
+/// memory accesses.
+#[derive(Debug)]
+#[allow(missing_docs)]
+pub struct MemoryAccesses {
+    pub memory_config: NormalMemoryConfig,
+    pub image: HeapImage,
+    pub offset: u32,
+    pub growth: u32,
+}
```
alexcrichton, line 17: Mind adding a memory64 knob to this as well? Also mind adding some docs for what each field member is? For example at a glance I'm not sure what `growth` is.

crates/fuzzing/src/generators/memory.rs
```diff
@@ -5,6 +5,75 @@ use arbitrary::{Arbitrary, Unstructured};
...
+
+impl<'a> Arbitrary<'a> for MemoryAccesses {
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        Ok(MemoryAccesses {
+            memory_config: u.arbitrary()?,
+            image: u.arbitrary()?,
+            offset: u.arbitrary()?,
+            growth: u.arbitrary()?,
```
alexcrichton, line 25: Does this mean that memories are going to grow an arbitrary 32-bit amount of bytes? If so should that be capped here to stay within limits on oss-fuzz?

crates/fuzzing/src/generators/memory.rs
alexcrichton: Also could this grow a fully general `WasmtimeConfig`? It would be good to test various compilation settings, Winch, the pooling allocator, etc.

alexcrichton (commented):
Nice! Any particular motivation for this or just curious to see how it fares?

fitzgen:
> Any particular motivation for this or just curious to see how it fares?

I started writing this for my implementation of the custom-page-sizes proposal, but figured I'd split it out and land it first to keep PRs small.

crates/fuzzing/src/oracles/memory.rs
fitzgen: Ha, I originally had that and rewrote it into the text format to get a little bit better throughput for the fuzzer. I can add a comment giving an overview of the generated module or something to make readability a bit better at a glance though.

crates/fuzzing/src/generators/memory.rs
fitzgen: Good idea, will do.

crates/fuzzing/src/generators/memory.rs
fitzgen: An arbitrary 32-bit number of pages. I figured we could just allow and ignore growth failure, but will that run afoul of oss-fuzz?

alexcrichton: I think so yeah, but not always. Oss-fuzz/libfuzzer I think override malloc and mainly look at the heap size through that but IIRC it also occasionally looks through vm stats and looks for things like `mmap`, so we could allocate 1G here and blow the limits on oss-fuzz when it gets around to detecting that. Given that I think we'll only want growths which are guaranteed to either succeed for fail given our memory limits.

cfallin (approved):

crates/fuzzing/src/oracles/memory.rs
```diff
@@ -0,0 +1,396 @@
...
+            assert!(
+                e.contains("bytes which exceeds the configured maximum of")
+                    || e.contains("exceeds the limit of")
+            );
+            return;
+        }
+    };
+
+    let limits = super::StoreLimits::new();
+    let mut store = Store::new(&engine, limits);
```
alexcrichton, line 40: This can use `input.config.to_store()` which should also handle things like epochs/fuel configuration I believe.

fitzgen: Unfortunately, we can't use that since we are setting extra config settings based on the arbitrary heap image, but we don't have access to the config/engine if we use that method.

PR #8749, opened 2024-06-05
Handle non-utf8 hosts in the adapter

alexcrichton:
This commit is a fix to the WASIp1 adapter for components to better handle the case where the host does not use a utf-8 string encoding. This is never the case for `wasmtime`-the-crate since it's a Rust-based host but this adapter is used outside of Wasmtime in jco, for example, where JS is not utf-8-based. When transcoding from utf-16 to utf-8 hosts may make an overlarge allocation and then shrink to a smaller allocation. This shrinking step has never been supported by the adapter and it's always aborted in this case.

Aside: why is this only a problem now? This hasn't been an issue before now because jco bindings never actually shrank. In doing so however this violated the canonical ABI because allocations are guaranteed to be precisely sized. New debug assertions in newer versions of Rust caught this mistake. This means that when jco tried to add downsizing of the allocation it quickly hit this panic in the adapter.

The fix in this commit is to handle the specific case of shrinking memory. The specific fix is to simply ignore the shrinking of memory. This is pretty subtle though why it seems to work out well enough for now (and it's probably still buggy). For now though this is enough to get jco's test suite passing with a shrinking allocation.

Unfortunately I don't know of a way to test this in this repository. Wasmtime does not support multiple encodings of host strings, only guest strings. This means that there's no wasmtime-based way to pass a non-utf-8 string into a guest.

### DIFF
--- a/crates/wasi-preview1-component-adapter/src/descriptors.rs
+++ b/crates/wasi-preview1-component-adapter/src/descriptors.rs
@@ -1,6 +1,6 @@
 use crate::bindings::wasi::cli::{stderr, stdin, stdout};
 use crate::bindings::wasi::io::streams::{InputStream, OutputStream};
-use crate::{BlockingMode, BumpAlloc, ImportAlloc, State, TrappingUnwrap, WasmStr};
+use crate::{BlockingMode, ImportAlloc, State, TrappingUnwrap, WasmStr};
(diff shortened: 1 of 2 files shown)


guybedford:
//cc @kateinoigakukun

pchickey (approved):

PR #8763, opened 2024-06-10
Wasmtime: Implement the custom-page-sizes proposal

fitzgen:
This commit adds support for the custom-page-sizes proposal to Wasmtime: https://github.com/WebAssembly/custom-page-sizes

I've migrated, fixed some bugs within, and extended the `*.wast` tests for this proposal from the `wasm-tools` repository. I intend to upstream them into the proposal shortly.

There is a new `wasmtime::Config::wasm_custom_page_sizes_proposal` method to enable or disable the proposal. It is disabled by default.

Our fuzzing config has been updated to turn this feature on/off as dictated by the arbitrary input given to us from the fuzzer.

Additionally, there were getting to be so many constructors for `wasmtime::MemoryType` that I added a builder rather than add yet another constructor.

In general, we store the `log2(page_size)` rather than the page size directly. This helps cut down on invalid states and properties we need to assert.

I've also intentionally written this code such that supporting any power of two page size (rather than just the exact values `1` and `65536` that are currently valid) will essentially just involve updating `wasmparser`'s validation and removing some debug asserts in Wasmtime.


crates/wasmtime/src/runtime/vm/vmcontext.rs
```diff
@@ -266,16 +266,20 @@ pub struct VMMemoryDefinition {
     /// atomically. For relaxed access, see
     /// [`VMMemoryDefinition::current_length()`].
     pub current_length: AtomicUsize,
+
+    /// The log2 of this memory's page size, in bytes.
+    pub page_size_log2: u8,
```
alexcrichton, line 271: How come this is stored here? This should be statically known from ambient context elsewhere I think? (e.g. this just stores the runtime state of the base/current byte length)

crates/wasmtime/src/runtime/vm/memory.rs
```diff
@@ -208,10 +228,18 @@ impl MmapMemory {
...
+            crate::runtime::vm::round_usize_up_to_host_pages(offset_guard_bytes);
+        let pre_guard_bytes = crate::runtime::vm::round_usize_up_to_host_pages(pre_guard_bytes);
+
         let (alloc_bytes, extra_to_reserve_on_growth) = match plan.style {
             // Dynamic memories start with the minimum size plus the `reserve`
             // amount specified to grow into.
-            MemoryStyle::Dynamic { reserve } => (minimum, usize::try_from(reserve).unwrap()),
+            MemoryStyle::Dynamic { reserve } => (
+                crate::runtime::vm::round_usize_up_to_host_pages(minimum),
+                crate::runtime::vm::round_usize_up_to_host_pages(usize::try_from(reserve).unwrap()),
```
alexcrichton, line 241: Mind importing this function since it's called a few times in succession here?

crates/wasmtime/src/runtime/vm/memory.rs
```diff
@@ -259,46 +289,71 @@ impl MmapMemory {
...
 
     fn grow_to(&mut self, new_size: usize) -> Result<()> {
-        if new_size > self.mmap.len() - self.offset_guard_size - self.pre_guard_size {
+        assert!(crate::runtime::vm::usize_is_multiple_of_host_page_size(
+            self.offset_guard_size
+        ));
+        assert!(crate::runtime::vm::usize_is_multiple_of_host_page_size(
+            self.pre_guard_size
+        ));
+        assert!(crate::runtime::vm::usize_is_multiple_of_host_page_size(
```
alexcrichton, line 332: Mind importing this function as well since it's used a few times?

alexcrichton (commented):
This looks great to me, thanks!

crates/wasmtime/src/runtime/vm/vmcontext.rs
fitzgen: I think I was using it for an approach that I ultimately abandoned.

crates/wasmtime/src/runtime/vm.rs
```diff
@@ -243,6 +243,28 @@ pub fn page_size() -> usize {
...
+}
+
+/// Round the given byte size up to a multiple of the host OS page size.
+pub fn round_u64_up_to_host_pages(bytes: u64) -> u64 {
+    let page_size = u64::try_from(crate::runtime::vm::host_page_size()).unwrap();
+    debug_assert!(page_size.is_power_of_two());
+    bytes
+        .checked_add(page_size - 1)
+        .map(|val| val & !(page_size - 1))
+        .unwrap_or(u64::MAX / page_size + 1)
```
elliottt, line 258: Should this be an `unwrap_or_else` to avoid always computing the default?
```suggestion
        .unwrap_or_else(|| u64::MAX / page_size + 1)
```
Also, could you add a comment that explains the default?

crates/cranelift/src/func_environ.rs
```diff
@@ -698,18 +708,14 @@ impl<'module_environment> FuncEnvironment<'module_environment> {
...
-            // made along the way...
-            pos.ins().sextend(desired_type, val)
+            // We have a 64-bit memory on a 32-bit host -- this combo doesn't
+            // really make a whole lot of sense to do from a user perspective
+            // but that is neither here nor there. We want to unsigned extend
+            // unless `val` is `-1`, as described in the doc comment above.
+            let extended = pos.ins().uextend(desired_type, val);
+            let neg_one = pos.ins().iconst(desired_type, -1);
+            let is_failure = pos.ins().icmp_imm(IntCC::Equal, val, -1);
+            pos.ins().select(is_failure, neg_one, extended)
```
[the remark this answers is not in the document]
fitzgen: Yeah with single-byte pages `-1` is now a valid return from a `memory.{grow,size}` in certain cases and so in this scenario we can't blindly sign extend. I can factor this out to another case for only when the memory has a 1-byte page size.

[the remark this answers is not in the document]
alexcrichton: I think sign extension still works since you can't actually consume the entire address space for wasm though, right?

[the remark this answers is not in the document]
fitzgen: But you could consume ~half of it and therefore have the sign bit set, which if we sign extend will result in a negative value / wrong unsigned value.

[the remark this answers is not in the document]
alexcrichton: The only "memory growth failed" value though is -1, not any negative number?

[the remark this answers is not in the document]
fitzgen: Yes, but you could have a valid memory size that has the sign bit set now, where that was impossible before because the maximum number of wasm pages that could fit in the address space wasn't large enough. That is, `(u32::MAX / 2) + 1` is a valid memory size with single-byte page sizes, and it has its high bit set, so if we sign-extend that value then we end up giving an incorrect memory size to the Wasm.

crates/wasmtime/src/runtime/vm/vmcontext.rs
```diff
@@ -266,16 +266,20 @@ pub struct VMMemoryDefinition {
...
-    /// smaller than what some other thread observes. Since Wasm memory only
-    /// grows, this under-estimation may be acceptable in certain cases.
+    /// Return the current length (in bytes) of the [`VMMemoryDefinition`] by
+    /// performing a relaxed load; do not use this function for situations in
+    /// which a precise length is needed. Owned memories (i.e., non-shared) will
+    /// always return a precise result (since no concurrent modification is
+    /// possible) but shared memories may see an imprecise value--a
+    /// `current_length` potentially smaller than what some other thread
+    /// observes. Since Wasm memory only grows, this under-estimation may be
+    /// acceptable in certain cases.
```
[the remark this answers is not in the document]
fitzgen: Just wrapped to 80 characters.

crates/wasmtime/src/runtime/vm.rs
fitzgen: I think it probably doesn't matter much either way in this case, since there isn't anything here that could prevent LLVM from cleaning this up itself (like a function call) but happy to make the change so that future readers don't ahve to reason through that.

fitzgen:
I believe all review feedback has been addressed at this point.

crates/cranelift/src/func_environ.rs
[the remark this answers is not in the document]
alexcrichton: Right! I'm apparently forgetting how negative numbers work...

alexcrichton (approved):

tests/all/memory.rs
```diff
@@ -375,7 +375,7 @@ fn massive_64_bit_still_limited() -> Result<()> {
             _max: Option<usize>,
         ) -> Result<bool> {
             self.hit = true;
-            Ok(true)
+            Ok(false)
```
alexcrichton, line 378: I think this'll want to stay as `true` as it's trying to exercise the path where the OS says "no you can't have that much memory" as opposed to the store limiting memory, and if `false` is returned here we won't ever hit the OS. I'm not sure why this test previously passed on MIRI though and it's failing on this PR

fitzgen: > I'm not sure why this test previously passed on MIRI though and it's failing on this PR

I'm not sure either, AFAICT we should be doing the same thing as before.

fitzgen:
oh my god this CI failure is driving me crazy:

https://github.com/bytecodealliance/wasmtime/actions/runs/9489825551/job/26152011334?pr=8763#step:5:243

```
 error: unused import: `prelude::*`
  --> crates\wasmtime\src\runtime\vm.rs:14:5
   |
14 |     prelude::*, DefinedFuncIndex, DefinedMemoryIndex, HostPtr, ModuleInternedTypeIndex, VMOffsets,
   |     ^^^^^^^^^^
   |
   = note: `-D unused-imports` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(unused_imports)]`
```

If I remove that import, then all the `err2anyhow()` calls break. I've tried with all sorts of different `cfg`s and I can't reproduce locally, even when I copy the `./ci/run-tests.sh` invocation from CI.

I have no idea why it thinks this is unused, AFAICT, it is definitely used.

crates/wasmtime/src/runtime/vm/memory.rs
```diff
@@ -52,8 +51,20 @@ impl RuntimeMemoryCreator for DefaultMemoryCreator {
     }
 }
 
-/// A linear memory
+/// A linear memory's backing storage.
+///
+/// This does not a full Wasm linear memory, as it may
```
XinyuZeng, line 56: This comment seems to be incomplete

fitzgen: Thanks for catching this!

Issue #6056, opened 2023-03-17
Cranelift: Simplify all our various CLIF load and store instructions


fitzgen:
* `{u,s}load{8,16,32}` these are all equivalent to `load.i{8,16,32}` followed by `{u,s}extend`. We should just use that larger sequence and backends that have special lowerings for `uload8()` that are better than `uextend(load8())` should add rules to match the compound pattern.

* Ditto for `{u,s}store{8,16,32}`.

* Same for our SIMD loads like `uload16x4` et al. These can be `load.i16x4` followed by a lane-wise `uextend`, etc.

PR #8765, opened 2024-06-10
riscv64: Add support for `load+extend` patterns

afonso360:
👋 Hey,

This PR adds support for merging `{s,u}extend` instructions into a preceding `load`.

RISC-V doesn't have sinkable loads per se, but the regular load instructions sign / zero extend the loaded values by default. So here we model that by pretending that that is a sinkable load on an extend instruction.

This PR is also a part of #6056. I'm working on that, the first step is to support generating the same code with `load+extend` on all backends as we currently do with the specialized `{u,s}loadNN` instructions.

### DIFF
--- a/cranelift/codegen/src/isa/riscv64/inst.isle
+++ b/cranelift/codegen/src/isa/riscv64/inst.isle
@@ -2424,6 +2424,11 @@
 
 ;; Helpers for sinkable loads ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
+;; RISC-V doesen't really have sinkable loads. But the regular load instructions
+;; sign / zero extend their results to 64 bits. So we can pretend they are
+;; an extend instruction with a sinkable load. This allows us to have better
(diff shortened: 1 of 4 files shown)


cranelift/codegen/src/isa/riscv64/inst.isle
```diff
@@ -2422,22 +2422,43 @@
...
+;; Extract a sinkable instruction from a value operand.
+(decl sinkable_inst (Inst) Value)
+(extern extractor sinkable_inst sinkable_inst)
+
+;; Matches a sinkable load.
+(decl sinkable_load (Inst Type MemFlags Value Offset32) Value)
+(extractor (sinkable_load inst ty flags addr offset)
+           (and
+              (load flags addr offset)
+              (sinkable_inst (has_type ty inst))))
```
fitzgen, line 2436: Maybe add a comment here about the `extend(load())` use case and the asterisk around "sinkable" that you added in the PR description? That seems like good context to have when reading this code.

fitzgen (approved):
Very nice!

PR #8766, opened 2024-06-10
Change WASI trait impls back to being blanket impls

alexcrichton:
This commit is a partial revert of #8609 to return `wasmtime-wasi` and `wasmtime-wasi-http` back to using blanket impls. The main change from before is to change the blanket impls to be in terms of a local newtype wrapper to avoid trait coherence issues. This is done because otherwise using the traits before required `&mut dyn WasiView` to exist but sometimes only a `Foo<'a>` is held which is not easy to get a `&mut dyn ...` view of. By changing to a blanket impl in terms of a newtype wrapper, `WasiImpl`, it's possible to call `bindgen!`-generated `add_to_linker_get_host` functions with a return value of `WasiImpl<Foo<'a>>` which enables hooking into all the generated bindings.


elliottt (approved):
Seems reasonable to me!

Issue #8768, opened 2024-06-10
riscv32imac no-std support for wasmtime


theoparis:
```
error: Wasmtime is being compiled for an architecture that it does not support. If this architecture is one you would like to see supported you may file an issue on Wasmtime's issue tracker: https://github.com/byt
ecodealliance/wasmtime/issues/new
  --> C:\Users\theo\.cargo\registry\src\index.crates.io-6f17d22bba15001f\wasmtime-21.0.1\src\runtime\vm\arch\mod.rs:24:9
   |
24 | /         compile_error!(
25 | |             "Wasmtime is being compiled for an architecture \
26 | |              that it does not support. If this architecture is \
27 | |              one you would like to see supported you may file an \
28 | |              issue on Wasmtime's issue tracker: \
29 | |              https://github.com/bytecodealliance/wasmtime/issues/new\
30 | |         ");
   | |__________^

error[E0432]: unresolved import `core::sync::atomic::AtomicU64`
   --> C:\Users\theo\.cargo\registry\src\index.crates.io-6f17d22bba15001f\wasmtime-21.0.1\src\runtime\store.rs:104:5
    |
104 | use core::sync::atomic::AtomicU64;
    |     ^^^^^^^^^^^^^^^^^^^^---------
    |     |                   |
    |     |                   help: a similar name exists in the module: `AtomicU32`
    |     no `AtomicU64` in `sync::atomic`

error[E0432]: unresolved import `core::sync::atomic::AtomicU64`
 --> C:\Users\theo\.cargo\registry\src\index.crates.io-6f17d22bba15001f\wasmtime-21.0.1\src\runtime\store\data.rs:8:26
  |
8 | use core::sync::atomic::{AtomicU64, Ordering::Relaxed};
  |                          ^^^^^^^^^
  |                          |
  |                          no `AtomicU64` in `sync::atomic`
  |                          help: a similar name exists in the module: `AtomicU32`
  |
  = help: consider importing this unresolved item through its public re-export instead:
          crate::store::AtomicU64

error[E0432]: unresolved import `core::sync::atomic::AtomicU64`
  --> C:\Users\theo\.cargo\registry\src\index.crates.io-6f17d22bba15001f\wasmtime-21.0.1\src\runtime\vm.rs:10:26
   |
10 | use core::sync::atomic::{AtomicU64, AtomicUsize, Ordering};
   |                          ^^^^^^^^^

...

[...]

PR #8770, opened 2024-06-10
Fix riscv64 for no-std

theoparis:
See issue #8768


alexcrichton (approved):
Thanks for this!

stlankes:
The latest release (22.0) is 2 days old and doesn't include this PR. Is this a correct behavior?

stlankes:
Sounds like a reasonable strategy!

PR #8772, opened 2024-06-11
upgrade to wasm-tools 210

pchickey:
@alexcrichton I don't think that Resolve::select_world has an obvious way to work anymore - how do we select the default packageid when multiple packages can be defined? I used get(0) but this is an arbitrary choice. This comes up in the upgrade of wit-bindgen as well.

### DIFF
--- a/crates/component-macro/src/bindgen.rs
+++ b/crates/component-macro/src/bindgen.rs
@@ -9,7 +9,7 @@ use syn::parse::{Error, Parse, ParseStream, Result};
 use syn::punctuated::Punctuated;
 use syn::{braced, token, Token};
 use wasmtime_wit_bindgen::{AsyncConfig, Opts, Ownership, TrappableError, TrappableImports};
-use wit_parser::{PackageId, Resolve, UnresolvedPackage, WorldId};
+use wit_parser::{PackageId, Resolve, UnresolvedPackageGroup, WorldId};
 
 pub struct Config {
     opts: Opts,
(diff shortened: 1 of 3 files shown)


alexcrichton:
Ah this is something I should have dealt with before publishing. There's some follow-up work to https://github.com/bytecodealliance/wasm-tools/pull/1577 to improve world selection, so let me try to get that in today to update to that. Otherwise [`wit_component::resolve_world_from_name`](https://docs.rs/wit-component/latest/wit_component/fn.resolve_world_from_name.html) is the rough equivalent of what to do but I'd like to change that too.

crates/component-macro/src/bindgen.rs
```diff
@@ -209,7 +214,16 @@ fn parse_source(
...
+        let UnresolvedPackageGroup {
+            packages,
+            source_map,
+        } = UnresolvedPackageGroup::parse("macro-input".as_ref(), inline)?;
+        Some(
+            packages
+                .into_iter()
+                .map(|p| resolve.push(p, &source_map))
+                .collect::<anyhow::Result<Vec<_>>>()?,
+        )
```
alexcrichton, line 226: This can be simplified a bit by using `Resolve::append` instead of iterating over the packages and calling `push` manually, although that does the same thing too.

crates/component-macro/src/bindgen.rs
```diff
@@ -170,11 +170,16 @@ impl Parse for Config {
...
+        let (resolve, pkgs, files) = parse_source(&path, &inline, &features)
             .map_err(|err| Error::new(call_site, format!("{err:?}")))?;
 
         let world = resolve
-            .select_world(pkg, world.as_deref())
+            .select_world(
+                *pkgs
+                    .get(0)
+                    .ok_or_else(|| Error::new(call_site, "at least one package must be defined"))?,
+                world.as_deref(),
```
alexcrichton, line 181: With https://github.com/bytecodealliance/wasm-tools/pull/1611 this'll ferry along the `pkgs` argument to `select_world` which is what we want here.

pchickey:
Lets wait to upgrade until 211 is out.

PR #8774, opened 2024-06-11
aarch64: Add support for `load+extends` patterns

afonso360:
👋 Hey,

This PR adds support for merging a `load` with a `{u,s}extend` instruction. On AArch64 the load instructions already do this by default, so we can just emit the regular loads during the extend lowering and mark them as sunk.

CC: https://github.com/bytecodealliance/wasmtime/issues/6056


cfallin (approved):
Nice!

Issue #8748, opened 2024-06-05
wasi-http: Allow setting TLS root certs for `default_send_request`


lann:
Currently, the TLS roots are [hard-coded](https://github.com/bytecodealliance/wasmtime/blob/ca405bb023d7f5f201eed8cdf37be3a9f3141e5b/crates/wasi-http/src/types.rs#L232-L239) to the `webpki-roots` set. This is a good default, but in some scenarios private roots are required. We should be able to add options to `OutgoingRequestConfig` to extend and/or replace that default set of roots with custom root(s).

bacongobbler:
Do you want the ability to add a few TLS certs to the existing webpki-roots set, or add a list of system certificates? `RootCertStore` [has two functions](https://docs.rs/rustls/latest/rustls/struct.RootCertStore.html):

- `add()`: Add a single DER-encoded certificate to the store. This is suitable for a small set of root certificates that are expected to parse successfully.
- `add_parsable_certificates()`: Parse the given DER-encoded certificates and add all that can be parsed in a best-effort fashion.

The former would be good when adding a small number of certificates to the store, whereas the latter would be better when adding a large collection of root certificates.

I am assuming we'd probably want to use `add` to add only a few select certificates on top of the `webpki-roots` set, but thought I'd check to make sure.

bacongobbler:
> to extend and/or replace that default set of roots with custom root(s).

Assuming we're going the route of "replace", would it be better if the caller provided its own `RootCertStore`? If not present, then we initialize a `root_cert_store` with the `webpki-roots` set?

lann:
It looks like `add` is more general purpose so if default to that approach.

bacongobbler:
Okay great. That solves the "extend" part of the equation.

Do we want to solve for replacement of the webpki-roots set? Would that be valuable to most callers of this function? Or would we just assume the caller should pass their own custom `send_request_handler` at that point?

lann:
Maybe it can take an `Option<rustls::RootCertStore>` where `None` represents the current behavior. It isn't too hard to reconstruct the defaults and extend if that's what you need.

rajatjindal:
I have been exploring what it will take to implement this. one question I have is how do folks expect the additional root ca config to be passed. 

- should this be a reference to a file name that contains the additional custom root ca certs or
- should this be a string representation of the additional certs

[...]

PR #8776, opened 2024-06-11
Update rustls root certificate handling in default_send_request_handler

bacongobbler:
While researching #8748, I noticed that the link in the comment results in a 404 error. The link refers to `tokio-rs/tls`. `tokio-rustls` has since moved to `rustls/tokio-rustls`.

I also noticed that the examples in the rustls documentation provide a more elegant method to add `webpki-roots` to the `RootCertStore`.

This PR doesn't change any functionality; it just updates Wasmtime's usage of `webpki-roots` and the `RootCertStore`, improving code readability, while also updating the link to one of the newer examples in the rustls repository.


pchickey (approved):
Thanks!

PR #8777, opened 2024-06-11
x64: Enable load sinking for small types in extend instructions

afonso360:
👋 Hey,

This PR adds support for load sinking in extend instructions on the x64 backend. These were pretty much already supported, but by default the x64 backend doesn't load sink small types. (It'd be neat to fix that at some point)

The only change here is to allow it to do so for `movzx` and `movsx`. These instructions have the correct implementation for small types and shouldn't try to perform a load larger than the source type.

This should also be the last backend that needs this implementation, s390x already supports this, so the next PR in this series is removing the `{u,s}loadNN` instructions from cranelift.

CC: https://github.com/bytecodealliance/wasmtime/issues/6056


cfallin (approved):
LGTM, thanks!

PR #8778, opened 2024-06-12
Remove the `ModuleRuntimeInfo` trait

alexcrichton:
Replace it with an `enum` of the two possibilities that it can be. This removes the need to have a trait dispatch indirection in the `vm` module. Previously this was required as `wasmtime-runtime` was a separate crate, but now it's no longer required.


fitzgen (approved):
Nice to start cleaning up these unnecessary trait objects now that `wasmtime` and `wasmtime-runtime` merged.

PR #8782, opened 2024-06-12
Fix lost `Waker` instances with async stdio streams

### LINKED ISSUE #8781: `AsyncStdoutStream` causes hangs after first write
### Test Case

A full reproduction, including the test case, can be found here: https://github.com/Licenser/wasi-io-hang

### Steps to Reproduce

* Create WASI environment
* Create an AsyncStdoutStream and use it as stdout
* load a wasi program that writes more than once to stdout
* in the wasi program, write to stdout
* read once and observe a correct result
* in the wasi program, the write call never returns
* the whole program hangs
* 
### Expected Results

I would expect to be able to write multiple times from withtin the wasi program when using an `AsyncStdoutStream` as stdout.

### Actual Results

The first write to stdout never finishes, and the wasi program hangs.

### Versions and Environment

Wasmtime version or commit: 21.0.1

Operating system: OSX

Architecture: ARMv8 (M3)

### Extra Info

I have looked for other examples of using `AsyncStdoutStream`, the main repository doesn't seem to have any more is it used in tests, I might be doing something wrong but this could be a genuine bug.

alexcrichton:
This commit fixes a bug in the `Subscribe` trait implementation for `AsyncStd{in,out}Stream` structures in the `wasmtime-wasi` crate. Previously these implementations would create a future for the duration of a single `poll` but then the future was dropped which could lead to lost wakeups as the waker is gone after the future is dropped. The fix was to use a `tokio::sync::Mutex` here instead of a `std::sync::Mutex` and leave some comments about why contention isn't expected.

Closes #8781

### DIFF
--- a/crates/wasi/src/stdio.rs
+++ b/crates/wasi/src/stdio.rs
@@ -561,8 +561,8 @@ mod test {
 
     #[tokio::test]
     async fn async_stdout_stream_unblocks() {
-        let (mut read, write) = tokio::io::duplex(1024);
-        let stdout = AsyncStdoutStream::new(AsyncWriteStream::new(1024, write));
+        let (mut read, write) = tokio::io::duplex(32);
+        let stdout = AsyncStdoutStream::new(AsyncWriteStream::new(32, write));
 
         let task = tokio::task::spawn(async move {
             let mut stream = stdout.stream();
--- a/crates/wasi/src/stdio.rs
+++ b/crates/wasi/src/stdio.rs
@@ -8,11 +8,9 @@ use crate::{
     HostInputStream, HostOutputStream, StreamError, StreamResult, Subscribe, WasiImpl, WasiView,
 };
 use bytes::Bytes;
-use std::future::Future;
 use std::io::IsTerminal;
-use std::pin::Pin;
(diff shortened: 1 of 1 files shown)


crates/wasi/src/stdio.rs
```diff
@@ -530,4 +558,39 @@ mod test {
...
+
+        let task = tokio::task::spawn(async move {
+            let mut stream = stdout.stream();
+            blocking_write_and_flush(&mut *stream, "x".into())
+                .await
+                .unwrap();
+        });
+
+        let mut buf = [0; 100];
+        let n = read.read(&mut buf).await.unwrap();
```
pchickey, line 575: I'd be more comfortable with this test if the amount read was in excess of the buffer sizes in the io::duplex and the AsyncWriteStream. You could just lower those to 32 each.

pchickey (approved):
Great catch. Thank you!

coolreader18:
Why not just use `tokio::sync::Mutex::blocking_lock()`?

alexcrichton:
Semantically these are all non-blocking functions, not blocking ones, so blocking on contention isn't appropriate for them.

PR #8789, opened 2024-06-12
Initial migration to `wasmtime_test`

saulecabrera:
This commit introduces the initial migration to the `wasmtime_test` macro.

This change starts by migrating all the applicable `func.rs` integration tests and at the same time it removes all the duplicated integration tests in `winch.rs`.

Additionally, this change introduces a slight change to how the macro works. Inspired by https://github.com/bytecodealliance/wasmtime/pull/8622#pullrequestreview-2083046911 it defaults to including all the known configuration combinations and allows the macro user to opt-out where applicable. This makes the usage of the macro less verbose.

The intention is to follow-up with subsequent PRs to migrate the all the applicable tests.


tests/all/func.rs
```diff
@@ -2139,12 +2152,11 @@ fn call_func_with_funcref_both_typed_and_untyped() -> Result<()> {
     Ok(())
 }
 
-#[test]
+#[wasmtime_test(strategies(not(Winch)))]
```
alexcrichton, line 2155: Thought on this, the reason that this test isn't supported on Winch is because it requires wasm gc, right? If that's the case it might be reasonable to have something like `features(gc)` for this macro. The macro would have built-in knowledge that Winch doesn't support `gc` but it would also additionally pre-configure the input `Config` to have `wasm_gc(true)` for example.

That might make things easier to un-ignore for winch in the future as it would only involve changing one location for "ok now winch supports gc" as opposed to hunting down the various exceptions

tests/all/func.rs
```diff
@@ -1890,11 +1903,12 @@ fn call_wasm_getting_subtype_func_return() -> anyhow::Result<()> {
...
```
alexcrichton, line 1906: Similar to below adding `features(simd)` here might make sense to auto-exclude Winch

saulecabrera:
Thanks for the review @alexcrichton. Your suggestions make sense to me, I've pushed an update in https://github.com/bytecodealliance/wasmtime/pull/8789/commits/9638a94bda3cc6dfa23101b393ded52f4fd0bbdd. Let me know what you think. I also removed some redundant `config.wasm_<feature>` for features that are now enabled by default (e.g. `reference_types`).

alexcrichton (approved):
Looks great, and thanks!

PR #8790, opened 2024-06-12
wasi-nn: update openvino backend to v0.7.2, use API 2.0

rahulchaphalkar:
- Updates [openvino](https://crates.io/crates/openvino) to v0.7.1 in wasi-nn crate
- Enables API 2.0 in `openvino.rs`.

### DIFF
--- a/crates/wasi-nn/src/backend/openvino.rs
+++ b/crates/wasi-nn/src/backend/openvino.rs
@@ -34,9 +34,8 @@ impl BackendInner for OpenvinoBackend {
         let xml = builders[0];
         let weights = builders[1];
 
-        //Construct new tensor with data.
-        let dims: [i64; 2] = [1, weights.len() as i64];
-        let shape = Shape::new(&dims)?;
+        // Construct a new tensor for the model weights.
+        let shape = Shape::new(&[1, weights.len() as i64])?;
         let mut weights_tensor = OvTensor::new(ElementType::U8, &shape)?;
         let buffer = weights_tensor.get_raw_data_mut()?;
         buffer.copy_from_slice(&weights);
--- a/crates/wasi-nn/src/backend/openvino.rs
+++ b/crates/wasi-nn/src/backend/openvino.rs
@@ -94,7 +94,7 @@ impl BackendExecutionContext for OpenvinoExecutionContext {
             Id::Name(name) => name,
         };
 
-        // Construct the blob structure. TODO: there must be some good way to
+        // Construct the tensor. TODO: there must be some good way to
         // discover the layout here; `desc` should not have to default to NHWC.
         let precision = map_tensor_type_to_precision(tensor.ty);
         let dimensions = tensor
--- a/crates/wasi-nn/src/backend/openvino.rs
+++ b/crates/wasi-nn/src/backend/openvino.rs
@@ -37,11 +37,9 @@ impl BackendInner for OpenvinoBackend {
         //Construct new tensor with data.
         let dims: [i64; 2] = [1, weights.len() as i64];
         let shape = Shape::new(&dims)?;
-        let mut weights_tensor = OvTensor::new(ElementType::F32, &shape)?;
-        let buffer = weights_tensor.buffer_mut()?;
-        for (index, bytes) in weights.iter().enumerate() {
-            buffer[index] = *bytes;
-        }
+        let mut weights_tensor = OvTensor::new(ElementType::U8, &shape)?;
+        let buffer = weights_tensor.get_raw_data_mut()?;
+        buffer.copy_from_slice(&weights);
 
         // Construct OpenVINO graph structures: `model` contains the graph
         // structure, `compiled_model` can perform inference.
@@ -107,10 +105,8 @@ impl BackendExecutionContext for OpenvinoExecutionContext {
             .collect::<Vec<_>>();
         let shape = Shape::new(&dimensions)?;
         let mut new_tensor = OvTensor::new(precision, &shape)?;
-        let buffer = new_tensor.buffer_mut()?;
-        for (index, bytes) in tensor.data.iter().enumerate() {
-            buffer[index] = *bytes;
-        }
+        let buffer = new_tensor.get_raw_data_mut()?;
+        buffer.copy_from_slice(&tensor.data);
 
         // Assign the tensor to the request.
         self.0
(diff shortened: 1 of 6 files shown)


rahulchaphalkar:
Converting this to draft temporarily, plan to bump openvino crate to a newer version than 0.7.1

crates/wasi-nn/src/backend/openvino.rs
```diff
@@ -23,41 +23,34 @@ impl BackendInner for OpenvinoBackend {
...
         // Read the guest array.
-        let xml = &builders[0];
-        let weights = &builders[1];
-
-        // Construct OpenVINO graph structures: `cnn_network` contains the graph
-        // structure, `exec_network` can perform inference.
+        let xml = builders[0];
+        let weights = builders[1];
+
+        //Construct new tensor with data.
```
abrown, line 37:
```suggestion
        // Construct a new tensor for the model weights.
```

crates/wasi-nn/src/backend/openvino.rs
```diff
@@ -23,41 +23,34 @@ impl BackendInner for OpenvinoBackend {
...
+        let dims: [i64; 2] = [1, weights.len() as i64];
+        let shape = Shape::new(&dims)?;
```
abrown, line 39:
```suggestion
        let shape = Shape::new(&[1, weights.len() as i64])?;
```

abrown (commented):
I think this makes complete sense. We still need to rebase so that CI runs (remember that one of these PRs should have `prtest:full` in it if that is not already the case?).

rahulchaphalkar:
@abrown I pushed some changes to fix failures due to some incompatibilities with `Id` enum. Tests are now passing, but vet step is failing, maybe due to one of my rebases. Can you vet the dependency? Thanks.

abrown:
I already did... where _did_ that commit go?

crates/wasi-nn/src/backend/openvino.rs
```diff
@@ -157,24 +148,51 @@ impl From<SetupError> for BackendError {
 
 /// Return the execution target string expected by OpenVINO from the
 /// `ExecutionTarget` enum provided by wasi-nn.
-fn map_execution_target_to_string(target: ExecutionTarget) -> &'static str {
+fn map_execution_target_to_string(target: ExecutionTarget) -> DeviceType<'static> {
```
abrown, line 151: If possible, let's rewrite this as `impl From<ExecutionTarget> for DeviceType<'static>`. If not possible then at least change the function name and documentation.

crates/wasi-nn/src/backend/openvino.rs
```diff
@@ -157,24 +148,51 @@ impl From<SetupError> for BackendError {
...
+        ExecutionTarget::Cpu => DeviceType::CPU,
+        ExecutionTarget::Gpu => DeviceType::GPU,
         ExecutionTarget::Tpu => unimplemented!("OpenVINO does not support TPU execution targets"),
     }
 }
 
 /// Return OpenVINO's precision type for the `TensorType` enum provided by
 /// wasi-nn.
-fn map_tensor_type_to_precision(tensor_type: TensorType) -> openvino::Precision {
+fn map_tensor_type_to_precision(tensor_type: TensorType) -> openvino::ElementType {
```
abrown, line 161: Similarly, this could be `impl From<TensorType> for ElementType`.

crates/wasi-nn/src/backend/openvino.rs
```diff
@@ -157,24 +148,51 @@ impl From<SetupError> for BackendError {
...
+        TensorType::Fp64 => ElementType::F64,
+        TensorType::U8 => ElementType::U8,
+        TensorType::I32 => ElementType::I32,
+        TensorType::I64 => ElementType::I64,
+        TensorType::Bf16 => ElementType::Bf16,
+    }
+}
+
+/// Return the `TensorType` enum provided by wasi-nn for OpenVINO's precision type
+fn map_precision_to_tensor_type(element_type: openvino::ElementType) -> TensorType {
```
abrown, line 174: This can be a `TryFrom` since it is more useful to return an error to the user than crash Wasmtime with `todo!`.

crates/wasi-nn/src/backend/openvino.rs
```diff
@@ -157,24 +148,51 @@ impl From<SetupError> for BackendError {
...
+    match element_type {
+        ElementType::F16 => TensorType::Fp16,
+        ElementType::F32 => TensorType::Fp32,
+        ElementType::F64 => TensorType::Fp64,
+        ElementType::U8 => TensorType::U8,
+        ElementType::I32 => TensorType::I32,
+        ElementType::I64 => TensorType::I64,
+        ElementType::Bf16 => TensorType::Bf16,
+        ElementType::Undefined => todo!("not yet supported in wasi-nn `TensorType`"),
```
abrown, line 183: I think this could get cleaned up to something like: `ElementType::Undefined | ElementType::Dynamic | ... => Err(BackendError::SomeNewVariant("{:?} is not yet supported in the openvino backend", element_type))`

crates/wasi-nn/src/backend/openvino.rs
```diff
@@ -78,63 +70,62 @@ impl BackendFromDir for OpenvinoBackend {
...
+            .map(|&d| d as i64)
             .collect::<Vec<_>>();
-        let desc = TensorDesc::new(Layout::NHWC, &dimensions, precision);
-        let blob = openvino::Blob::new(&desc, &tensor.data)?;
-
-        // Actually assign the blob to the request.
-        self.1.set_blob(&input_name, &blob)?;
+        let shape = Shape::new(&dimensions)?;
+        let mut new_tensor = OvTensor::new(precision, &shape)?;
+        let buffer = new_tensor.get_raw_data_mut()?;
```
lum1n0us, line 101: If no need to specify layout (of input tensor), and there is no PrePostProcessing in the whole file, Does it mean wasm code has to make sure the input tensor matches the requirement of model input? (I assume openvino models requires "NHWC" input)

rahulchaphalkar: Hi Liang, yes this requires input tensor layout match the expected layout for model input.

abrown (approved):

PR #8791, opened 2024-06-13
wasmtime: Make Store context available to call hooks

jameysharp:
Rather than digging only the embedder's store-specific data out and passing that along, pass access to the entire Store. This is the same thing we do for epoch interruption hooks, including the annoyance of having to take the callback out of the store temporarily while calling it to avoid having multiple mutable borrows.


jameysharp:
I think the only other example of this pattern in `Store` is [`Store::limiter`](https://docs.rs/wasmtime/latest/wasmtime/struct.Store.html#method.limiter) and its async variant. I don't want to change that today but I think it's worth discussing whether we should do that eventually.

alexcrichton (approved):

PR #8795, opened 2024-06-13
Add a compile-time feature for call hooks

alexcrichton:
This commit moves the `Store::call_hook` API behind a Cargo feature named `call-hook`. This helps speed up the path from wasm into the host by avoiding branches at the start and the end of the execution. In a thread on [Zulip] this is locally leading to significant performance gains in this particular microbenchmark so having an option to disable it at the crate layer seems like a reasonable way to thread this needle for now. This definitely has a downside in that it requires a crate feature at all, but I'm not sure of a better solution as LLVM can't dynamically detect that `Store::call_hook` is never invoked and therefore the branch can be optimized away.

[Zulip]: https://bytecodealliance.zulipchat.com/#narrow/stream/217126-wasmtime/topic/Performance.20regression.20since.20rust.201.2E65/near/444505571

### DIFF
--- a/crates/wasmtime/Cargo.toml
+++ b/crates/wasmtime/Cargo.toml
@@ -126,6 +126,7 @@ default = [
   'component-model',
   'threads',
   'std',
+  'call-hook',
 ]
 
 # An on-by-default feature enabling runtime compilation of WebAssembly modules
@@ -261,3 +262,9 @@ std = [
   'object/std',
   'once_cell/std',
 ]
+
+# Enables support for the `Store::call_hook` API which enables injecting custom
+# logic around all entries/exits from WebAssembly. This has a slight performance
+# cost for all host functions so is provided as a compile-time feature if
+# embedders would like to disable it.
+call-hook = []
--- a/crates/wasmtime/src/lib.rs
+++ b/crates/wasmtime/src/lib.rs
@@ -254,6 +254,11 @@
 //! * `threads` - Enabled by default, this enables compile-time support for the
 //!   WebAssembly `threads` proposal, notably shared memories.
 //!
+//! * `call-hook` - Enabled by default, this enables support for the
+//!   [`Store::call_hook`] API. This incurs a small overhead on all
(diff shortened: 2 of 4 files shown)


fitzgen (approved):

pchickey:
I think this is the correct choice - needing call hooks at all is a pretty particular need, and I'd even be fine if this feature was not enabled by default. Have we surveyed if anyone besides Fastly's embedding even makes use of it?

alexcrichton:
I haven't surveyed myself but carrying a Cargo feature for it isn't too bad. Most of our Cargo features have been relatively low overhead to maintain so far. The main question to me is that this might be an appropriate feature to be off-by-default instead of on-by-default as it can affect benchmarks, but even then it's pretty rare anyone benchmarks host-to-wasm calls or vice versa, most benchmarking is just of the wasm itself.

tschneidereit:
I think regardless of who might encounter this in benchmarks, as long it's not essentially always used, it makes sense to disable by default: we should keep the call overhead low wherever we can, after all, and requiring people to disable very specific features for best performance doesn't seem great to me.

alexcrichton:
I can try to investigate this a bit. I believe the `gc` feature has a much larger impact than call-hooks in terms of performance. I haven't bottomed that out but disabling `gc`-by-default I think would be much more significant than disabling `call-hook` by default.

alexcrichton:
Measuring locally from the benchmark from Zulip, which is very heavy in wasm->host calls and the host call itself does nothing, enabling the `gc` feature of Wasmtime yields a slowdown of 3.5x and enabling the `call-hook` feature slows down only 30%. 

Put another way I don't think it's worth over-rotating too much on `call-hook`, `gc` is the main feature that's causing a slowdown.

alexcrichton:
With https://github.com/bytecodealliance/wasmtime/pull/8807 the performance is on-par now with the `gc` feature enabled, so I'll follow that up by disabling the `call-hook` feature by default.

PR #8796, opened 2024-06-13
Update MSRV to 1.77.0

alexcrichton:
This accompanies today's release of Rust 1.79.0. This means that CI will now use the latest stable of 1.79.0 for testing primarily. Additionally this updates the pinned nightly commit used for testing to today's nightly and fixes a few warnings that uncovered. I plan on having a follow-up that leverages some new APIs and features in 1.77.0.

### DIFF
--- a/cranelift/native/src/lib.rs
+++ b/cranelift/native/src/lib.rs
@@ -167,7 +167,7 @@ mod tests {
 
             if cfg!(all(target_os = "macos", target_arch = "aarch64")) {
                 assert_eq!(isa.default_call_conv(), CallConv::AppleAarch64);
-            } else if cfg!(any(unix, target_os = "nebulet")) {
+            } else if cfg!(unix) {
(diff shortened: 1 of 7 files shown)


cranelift/native/src/lib.rs
```diff
@@ -167,7 +167,7 @@ mod tests {
 
             if cfg!(all(target_os = "macos", target_arch = "aarch64")) {
                 assert_eq!(isa.default_call_conv(), CallConv::AppleAarch64);
-            } else if cfg!(any(unix, target_os = "nebulet")) {
+            } else if cfg!(unix) {
```
alexcrichton, line 170: This change is because rustc now warns that `target_os = "nebulet"` doesn't exist. I think that means the target was probably removed here. I don't know what nebulet is myself, though, so I'm just obeying rustc.

elliottt (approved):

PR #8800, opened 2024-06-13
Try migrating to a "more official" github pages provider

alexcrichton:
This commit is an attempt to migrate from
`JamesIves/github-pages-deploy-action`, a third party action, to what appears to be a more official action under the `github.com/actions` organization. This was first pioneered in
bytecodealliance/wasm-tools#1597 and it looked like it would work well over here too.


cfallin (approved):
Seems reasonable to me!

PR #8805, opened 2024-06-14
Update the spec test suite submodule

alexcrichton:
* Ignore the `memory64` tests since Wasmtime doesn't implement the table64 extension yet.
* Ignore `gc/i31.wast` as it's got new tests which Wasmtime doesn't currently pass.

### DIFF
--- a/tests/wast.rs
+++ b/tests/wast.rs
@@ -78,6 +78,10 @@ fn ignore(test: &Path, strategy: Strategy) -> bool {
         if part == "extended-const" {
             return true;
         }
+        // Wasmtime doesn't implement the table64 extension yet.
+        if part == "memory64" {
+            return true;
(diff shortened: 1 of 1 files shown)


tests/wast.rs
```diff
@@ -78,6 +78,10 @@ fn ignore(test: &Path, strategy: Strategy) -> bool {
         if part == "extended-const" {
             return true;
         }
+        // Wasmtime doesn't implement the table64 extension yet.
+        if part == "memory64" {
+            return true;
+        }
```
fitzgen, line 84: It seems a bit surprising that we have to ignore *all* memory64 tests now and not just the table related ones. Is the latter possible?

fitzgen (approved):

PR #8811, opened 2024-06-14
Skip GC LIFO scope creation on some hostcalls

alexcrichton:
This commit skips the creation of a LIFO scope for GC `Rooted<T>` values for hostcalls using statically typed arguments/results (e.g. `Func::wrap`) where none of the arguments/results have GC values. This removes the last bit of "GC business" on `Func::wrap` calls that don't do anything related to GC.

### DIFF
--- a/crates/wasmtime/src/runtime/func.rs
+++ b/crates/wasmtime/src/runtime/func.rs
@@ -2013,7 +2013,7 @@ pub struct Caller<'a, T> {
 }
 
 impl<T> Caller<'_, T> {
-    unsafe fn with<F, R>(caller: *mut VMContext, f: F) -> R
+    unsafe fn with<F, R>(caller: *mut VMContext, create_lifo_scope: bool, f: F) -> R
     where
         // The closure must be valid for any `Caller` it is given; it doesn't
         // get to choose the `Caller`'s lifetime.
@@ -2024,7 +2024,11 @@ impl<T> Caller<'_, T> {
         debug_assert!(!caller.is_null());
         crate::runtime::vm::Instance::from_vmctx(caller, |instance| {
             let store = StoreContextMut::from_raw(instance.store());
-            let gc_lifo_scope = store.0.gc_roots().enter_lifo_scope();
+            let gc_lifo_scope = if create_lifo_scope {
+                Some(store.0.gc_roots().enter_lifo_scope())
+            } else {
+                None
+            };
 
             let ret = f(Caller {
                 store,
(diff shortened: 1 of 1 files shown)


crates/wasmtime/src/runtime/func.rs
```diff
@@ -2275,7 +2281,8 @@ impl HostContext {
         // closure and then run it as part of `Caller::with`.
         let result = crate::runtime::vm::catch_unwind_and_longjmp(move || {
             let caller_vmctx = VMContext::from_opaque(caller_vmctx);
-            Caller::with(caller_vmctx, run)
+            let create_lifo_scope = P::may_gc() || R::may_gc();
+            Caller::with(caller_vmctx, create_lifo_scope, run)
```
fitzgen, line 2285: I think this could actually be bad: if I define a host call with only integer params and results, but then create an `externref` inside it[^0], then that `externref` will be rooted in the outer LIFO scope, rather than just being rooted in the `Caller`'s scope, which is probably the store's scope. That means the `externref` is going to leak for the whole extent of the program. So I don't think we can actually do this.

Is [reading a vec's length](https://github.com/bytecodealliance/wasmtime/blob/9ffc9e67f6c19b8a3a80a4bd31aac8a08e628d3f/crates/wasmtime/src/runtime/gc/enabled/rooting.rs#L409) and then later [a well-predicted compare-and-branch to out-of-line code](https://github.com/bytecodealliance/wasmtime/blob/9ffc9e67f6c19b8a3a80a4bd31aac8a08e628d3f/crates/wasmtime/src/runtime/gc/enabled/rooting.rs#L420-L427) really that expensive?

[^0]: Maybe I then re-entrantly call some other Wasm export, giving it that `externref`. Doesn't really matter *why* I did it, just that I did it.

alexcrichton: My thinking was that it's still safe in that case, right? It means it'll be rooted long but a manual rooting scope can always be created to avoid that.

Perf-wise this was locally showing a 40% perf improvement but I cannot reproduce it against the latest `main`. Turns out that I was testing before #8806 landed, and now that that PR has landed this PR is largely perf-neutral. The benchmark here is pretty micro-benchmark-y but at the same time I think it's a nice feature of Wasmtime that the overhead for calling the host is as low as possible, so cleaning up the assembly as much as we can I think is reasonable.

fitzgen: > it's a nice feature of Wasmtime that the overhead for calling the host is as low as possible, so cleaning up the assembly as much as we can I think is reasonable.

Agreed in general.

> My thinking was that it's still safe in that case, right? It means it'll be rooted longer but a manual rooting scope can always be created to avoid that.

It is memory safe, yes, but it is a fairly big footgun, IMO. At minimum, we would need to be very clear about when a `Caller` is vs. isn't a rooting scope in our existing docs around rooting and scopes (I think on `Rooted`? maybe on `RootScope`? I forget)

But if this isn't a perf win anymore, then is it really worth adding the footgun?

alexcrichton: Ok I think that's reasonable yeah. I'll close this and we can revisit if it ever becomes necessary but that seems unlikely too.

PR #8813, opened 2024-06-14
Disable memory protection keys by default at compile time

alexcrichton:
This commit gates memory protection keys behind a new Cargo feature which is disabled by default. Memory protection keys are already disabled by default on all platforms and are only configured to possibly work with Linux x64. When enabled, however, it unconditionally adds a small amount of overhead to WebAssembly entries/exits even if the feature is disabled at runtime for the same reason that the `call-hook` feature adds overhead. With `call-hook` being disabled by default in #8808 it seemed reasonable to additionally gate memory protection keys to avoid needing to disable features in Wasmtime to get the best performance wasm<->host calls.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,7 +82,7 @@ rustix = { workspace = true, features = ["mm", "param", "process"] }
 
 [dev-dependencies]
 # depend again on wasmtime to activate its default features for tests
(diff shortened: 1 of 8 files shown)


alexcrichton:
cc @abrown to give you a heads up

fitzgen (approved):

abrown (approved):

PR #8823, opened 2024-06-17
Inherit Linux semantics for `fd_pwrite` with `O_APPEND`

### LINKED ISSUE #8817: File fd_pwrite bug
### Test Case
The c test case is:
```

#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <fcntl.h>
#include <sys/uio.h>

int get_fd(const char *filename, int flags) {
    int fd = open(filename, flags);
    
    if (fd == -1) {
        printf("Get file descriptor of file %s failed!\n", filename);
        return -1;
    } else {
        printf("Get file descriptor of file %s succeed!\n", filename);
        return fd;
    }
}

void closebyfd(int fd) {
    if (close(fd) == -1) {
        printf("Close the file %d by descriptor failed!\n", fd);
    }
}

void fd_pwrite_00019_nEEKq(int fd) {
    printf("Enter function fd_pwrite_00019_nEEKq\n");
    off_t size = lseek(fd, 0, SEEK_END); 
    printf("Current file size before: %ld\n", size);

    
    struct iovec iov[1];
    iov[0].iov_base = "JBr3HJK";
    iov[0].iov_len = 7;
    
    off_t offset = lseek(fd, 0, SEEK_CUR);
    if (offset == -1) {
        printf("Failed to get current offset\n");
        return;
    }
    
    ssize_t bytes_written = pwritev(fd, iov, 1, 0);
    if (bytes_written == -1) {
        printf("pwritev failed\n");
    } else {
        printf("pwritev successful. %zd bytes written\n", bytes_written);
    }
    size = lseek(fd, 0, SEEK_END); 
    printf("Current file size after: %ld\n", size);

}

int main() {
    int fd = get_fd("subdir_2/subdir_2/subfile_3", O_WRONLY | O_APPEND);
    if (fd == -1) {
        return 1;
    }
    
    fd_pwrite_00019_nEEKq(fd);
    
    closebyfd(fd);
    
    return 0;
}
```

### Steps to Reproduce
(1)compile to wasm:`./wasi-sdk-21.0/bin/clang --target=wasm32-unkown-wasi --sysroot=./wasi-sdk-21.0/share/wasi-sysroot test.c -o test.wasm`

(2)Running wasm:
(Before run the Wasm file, file subdir_2/subdir_2/subfile_3 exists, and the file size is 20 bytes.)
`wasmtime run --dir=. test.wasm`

### Expected Results
Print:
```
Get file descriptor of file subdir_2/subdir_2/subfile_3 succeed!
Enter function fd_pwrite_00019_nEEKq
Current file size before: 20
pwritev successful. 7 bytes written
Current file size after: 27
```

This is what WAMR, WasmEdge  and Linux native code do.

### Actual Results
wasmtime prints:
```
Get file descriptor of file subdir_2/subdir_2/subfile_3 succeed!
Enter function fd_pwrite_00019_nEEKq
Current file size before: 20
pwritev successful. 7 bytes written
Current file size after: 20
```
wasmtime do not write the content into the file although the file is opened with O_APPEND.

### Versions and Environment
Wasmtime version or commit: 19.0.2
Operating system: Ubuntu 20.04
Architecture: x86_64

alexcrichton:
This commit updates the implementation of `fd_pwrite` in WASI to match Linux semantics for an under-specified corner of WASI. Specifically if `fd_pwrite` is used the offset specified is ignored if the file is opened in append mode and the bytes are instead appended.

This commit additionally refactors `fd_write` and `fd_pwrite` to have basically the same code with only a minor branch internally when the final write is being performed to help deduplicate more logic.

Closes #8817

### DIFF
--- a/crates/test-programs/src/bin/preview1_file_pread_pwrite.rs
+++ b/crates/test-programs/src/bin/preview1_file_pread_pwrite.rs
@@ -176,6 +176,42 @@ unsafe fn test_file_pwrite_and_file_pos(dir_fd: wasi::Fd) {
     wasi::path_unlink_file(dir_fd, path).expect("removing a file");
 }
 
+unsafe fn test_file_pwrite_and_append(dir_fd: wasi::Fd) {
+    let path = "file3";
+    let file_fd = wasi::path_open(
+        dir_fd,
+        0,
+        path,
+        wasi::OFLAGS_CREAT,
+        wasi::RIGHTS_FD_READ | wasi::RIGHTS_FD_WRITE,
+        0,
+        wasi::FDFLAGS_APPEND,
+    )
+    .expect("opening a file");
+
+    // Inherit linux semantics for `pwrite` where if the file is opened in
+    // append mode then the offset to `pwrite` is ignored.
+    let buf = [0];
+    let ciovec = wasi::Ciovec {
+        buf: buf.as_ptr(),
+        buf_len: buf.len(),
+    };
+    let n = wasi::fd_pwrite(file_fd, &[ciovec], 50).expect("writing bytes at offset 50");
+    assert_eq!(n, 1);
+
+    let stat = wasi::fd_filestat_get(file_fd).unwrap();
+    assert_eq!(stat.size, 1);
+
+    let n = wasi::fd_pwrite(file_fd, &[ciovec], 0).expect("writing bytes at offset 50");
+    assert_eq!(n, 1);
+
+    let stat = wasi::fd_filestat_get(file_fd).unwrap();
+    assert_eq!(stat.size, 2);
+
+    wasi::fd_close(file_fd).expect("closing a file");
+    wasi::path_unlink_file(dir_fd, path).expect("removing a file");
+}
+
 fn main() {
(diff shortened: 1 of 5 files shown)


fitzgen:
Redirecting to @pchickey

pchickey (approved):

alexcrichton:
CI turned up an interesting phenomena which makes sense in retrospect. With wasi-common the behavior is subject to whatever the platform OS provides which makes Linux the odd-one-out. Papering over OS differences is easy enough in `wasmtime-wasi` but is not trivial in `wasi-common`.

@pchickey how do you feel about sticking with these semantics and just ignoring the whole test in wasi-common? Alternatives could include not landing this in Wasmtime and taking it up in an issue with WASI and/or adding more stuff to wasi-common.

pchickey:
I think it is fine to just leave this bug in wasi-common and ignore the test - I don't think there is much value in bringing bug fixes for corner cases like this, if users need the fix the real question is why they havent upgraded to wasmtime-wasi.

PR #8825, opened 2024-06-17
wasi-nn: use `libtest-mimic`

abrown:
wasi-nn's testing story is complicated by different levels of support on different platforms (some backends work on certain architectures, others only work on certain OSes, etc.). This change migrates the `testing` module, which was included in `src`, to exist solely under `tests`. It also dynamically checks whether each test is runnable and then chooses whether to ignore it with a `libtest-mimic` flag. This ensures we can see all the tests all the time and whether they are running or not, which is helpful during development.

This adds a bit of complexity to wasi-nn testing but this is necessary for future improvements. In a future PR I will attempt to upgrade the WIT-ABI support for wasi-nn, which means testing some of these same test programs in their component form instead of core module form. This change lays that ground work.


alexcrichton (approved):
Looks reasonable to me 👍

PR #8826, opened 2024-06-17
Introduce the `cranelift-bitset` crate; use it for stack maps in both Cranelift and Wasmtime

fitzgen:


cranelift/bitset/src/scalar.rs
```diff
@@ -0,0 +1,575 @@
...
+    ///
+    /// let mut bitset = ScalarBitSet::<u8>::new();
+    ///
+    /// // A `ScalarBitSet<u8>` can only hold the elements `0..=7`, so `8` is
+    /// // out of bounds and will trigger a panic.
+    /// bitset.contains(8);
+    /// ```
+    #[inline]
+    pub fn contains(&self, i: u8) -> bool {
+        assert!(i < Self::capacity());
```
alexcrichton, line 237: Given the perf-sensitive nature of bitsets should this be `debug_assert!`?

cranelift/bitset/src/compound.rs
```diff
@@ -0,0 +1,471 @@
...
+    ///
+    /// let bitset = CompoundBitSet::with_capacity(4096);
+    ///
+    /// assert!(bitset.is_empty());
+    /// assert!(bitset.capacity() >= 4096);
+    /// ```
+    #[inline]
+    pub fn with_capacity(capacity: usize) -> Self {
+        // Divide `capacity` by `BITS_PER_WORD`, rounding up rather than down.
+        let elems_cap = capacity + BITS_PER_WORD - 1 / BITS_PER_WORD;
```
alexcrichton, line 107: I think this might need `(...)` to disambiguate?

cranelift/bitset/src/compound.rs
```diff
@@ -0,0 +1,471 @@
...
+    ///         // storage.
+    ///         assert!(bitset.capacity() > i);
+    ///         bitset.insert(i);
+    ///     }
+    /// }
+    /// ```
+    #[inline]
+    pub fn reserve(&mut self, n: usize) {
+        // Divide `n` by `BITS_PER_WORD` rounding up, rather than down.
+        let elem_len = n + (BITS_PER_WORD - 1) / BITS_PER_WORD;
```
alexcrichton, line 258: I think this might also need some parens to get the calculation you're shooting for

cranelift/bitset/src/compound.rs
```diff
@@ -0,0 +1,471 @@
...
+        self.elems.resize_with(elem_len, ScalarBitSet::new);
```
alexcrichton, line 259: This is notably quite different from `Vec::reserve` where it's initializing the reserved space, is that desired here?

cranelift/bitset/src/compound.rs
```diff
@@ -0,0 +1,471 @@
...
+            if last.is_empty() {
+                self.elems.pop();
+                continue;
+            }
+
+            let bit = last.pop()?;
+            if last.is_empty() {
+                self.elems.pop();
+            }
+            return Some(Self::elem(word, bit));
```
alexcrichton, line 377: I think this needs to update `self.len`?

cranelift/bitset/src/compound.rs
```diff
@@ -0,0 +1,471 @@
...
+    /// assert_eq!(bitset.pop(), Some(222));
+    /// assert_eq!(bitset.pop(), Some(111));
+    /// assert_eq!(bitset.pop(), None);
+    /// ```
+    #[inline]
+    pub fn pop(&mut self) -> Option<usize> {
+        loop {
+            if self.elems.is_empty() {
+                return None;
+            }
```
alexcrichton, line 363: You can probably skip this due to the `?` on `last_mut()`

cranelift/bitset/src/scalar.rs
fitzgen: I was thinking that originally, but figured we might as well do the more-correct thing by default and add `unchecked` versions later if profiling shows that these asserts actually matter.

I kind of expect LLVM to clean them all up, at least when they are called via `CompoundBitSet` methods, since it should be able to determine the ranges of values flowing into these methods based on the modulo operation in `CompoundBitSet::word_and_bit`.

cranelift/bitset/src/compound.rs
fitzgen: Whoops!

fitzgen:
@alexcrichton this should be ready for a second round of review now.

cranelift/bitset/src/compound.rs
```diff
@@ -0,0 +1,495 @@
...
+    ///     if i % 2 == 0 {
+    ///         // Inserting this value should not require growing the backing
+    ///         // storage.
+    ///         assert!(bitset.capacity() > i);
+    ///         bitset.insert(i);
+    ///     }
+    /// }
+    /// ```
+    #[inline]
+    pub fn reserve(&mut self, n: usize) {
```
alexcrichton, line 255: I might recommend renaming this method to `resize` or some variant thereof. The `Vec::reserve` method reserves space for extra capacity which isn't what this is doing so I'd be hesitant to create confusion between the two. Also `Vec::reserve` is mostly about preallocating uninitialized memory where this is specifically initializing the memory as well.

alexcrichton (approved):

PR #8830, opened 2024-06-18
Implement semver compatibility for exports

### LINKED ISSUE #8395: WASI semver compatibility should apply to exports
In https://github.com/bytecodealliance/wasmtime/pull/7994 the `component::Linker` type in Wasmtime was made aware of semver by enabling loading a component that requires 0.2.0 of an interface with a supplied 0.2.1 of an interface. This enables us to update WASI in Wasmtime without breaking the ability to load older components.

This support, however, does not work for exports. If a component exports 0.2.0 and the runtime is generating bindings for 0.2.1 then loading the export will fail currently, there's no logic anywhere to probe for older verions.

This represents a snag in our story of WASI compatibility in Wasmtime and is something we should address before updating WASI to 0.2.1

alexcrichton:
This commit is an implementation of component model semver compatibility for export lookups. Previously in #7994 component imports were made semver-aware to ensure that bumping version numbers would not be a breaking change. This commit implements the same feature for component exports. This required some refactoring to move the definition of semver compat around and the previous refactoring in #8786 enables frontloading this work to happen before instantiation.

Closes #8395

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3647,6 +3647,7 @@ dependencies = [
  "object 0.36.0",
  "postcard",
  "rustc-demangle",
+ "semver",
  "serde",
  "serde_derive",
  "target-lexicon",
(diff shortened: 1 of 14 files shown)


elliottt (approved):
Looks good to me! I had a few suggestions that you can take or leave :)

PR #8837, opened 2024-06-19
wasmtime: Consume fuel on all return paths

jameysharp:
As far as I can tell, when functions use a `return` instruction rather than falling off the end, fuel was not being tracked correctly. The `fuel_function_exit` method was only called from
`after_translate_function`, which is only called once all the instructions in the function have been translated, not at each return.

In this commit I switched to calling `fuel_function_exit` from `handle_before_return` instead, alongside some of the `wmemcheck` hooks. That should ensure that it happens on every function exit, regardless of what form that exit takes.

I think `after_translate_function` is fundamentally difficult to use correctly, and it wasn't used for anything else, so I've also removed it in this commit. And I did a minor cleanup at the site of one of the calls to `handle_before_return` while I was looking at it.

### DIFF
--- a/cranelift/wasm/src/code_translator.rs
+++ b/cranelift/wasm/src/code_translator.rs
@@ -577,13 +577,10 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
             state.reachable = false;
         }
         Operator::Return => {
-            let return_count = {
-                let frame = &mut state.control_stack[0];
-                frame.num_return_values()
-            };
+            let return_count = state.control_stack[0].num_return_values();
             {
                 let return_args = state.peekn_mut(return_count);
-                environ.handle_before_return(&return_args, builder);
+                environ.handle_before_return(return_args, builder);
                 bitcast_wasm_returns(environ, return_args, builder);
                 builder.ins().return_(return_args);
             }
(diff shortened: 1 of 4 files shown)


jameysharp:
I think Nick's out for a few days so I'm handing this off to you instead, Alex.

alexcrichton (approved):
Good catch! Mind adding a test or two [here as well](https://github.com/bytecodealliance/wasmtime/blob/main/tests/all/fuel.wast)?

jameysharp:
Thanks for pointing me to those tests because I immediately got usefully confused about why the tests passed both before and after my changes.

It turns out that `fuel_save_from_var` is called by way of `before_translate_operator` for any of `unreachable`, `return`, or various kinds of calls. So doing it again in `handle_before_return` is unnecessary—and calling `fuel_save_from_var` multiple times is harmless but a waste of time—except when the return is an implicit fall-through off the end of the function. And that case is currently correctly handled by `after_translate_function`.

So the current implementation is not actually buggy!

But I still think the `after_translate_function` interface is sketchy. So I want to think more about what interface actually makes sense for instrumenting wasm translation like this, maybe drawing inspiration from [whamm](https://github.com/titzer/wizard-engine/blob/6e52d99b8d326d42c1ed1050c0e509a2b5741da0/src/monitors/WhammMonitor.v3#L8-L29). Perhaps for instrumentation purposes we should behave as if there's a `return` instruction after the end of every function…

PR #8838, opened 2024-06-19
upgrade to wasm-tools 0.211.1

pchickey:
Only one change I wasn't certain of, for the `storage: &mut String` parameter in wasmprinter:
https://github.com/bytecodealliance/wasmtime/compare/pch/wasm_tools_211?expand=1#diff-4863961fda2fef6610662e23fb35b7d7c2a367032798634425f2f407e0ee932aR70

### DIFF
--- a/crates/explorer/src/lib.rs
+++ b/crates/explorer/src/lib.rs
@@ -66,9 +66,10 @@ struct AnnotatedWatChunk {
 }
 
 fn annotate_wat(wasm: &[u8]) -> Result<AnnotatedWat> {
-    let mut printer = wasmprinter::Printer::new();
+    let printer = wasmprinter::Config::new();
+    let mut storage = String::new();
     let chunks = printer
-        .offsets_and_lines(wasm)?
+        .offsets_and_lines(wasm, &mut storage)?
         .map(|(offset, wat)| AnnotatedWatChunk {
             wasm_offset: offset.map(|o| WasmOffset(u32::try_from(o).unwrap())),
             wat: wat.to_string(),
--- a/crates/types/src/lib.rs
+++ b/crates/types/src/lib.rs
@@ -1736,7 +1736,7 @@ pub trait TypeConvert {
     fn convert_heap_type(&self, ty: wasmparser::HeapType) -> WasmHeapType {
         match ty {
             wasmparser::HeapType::Concrete(i) => self.lookup_heap_type(i),
-            wasmparser::HeapType::Abstract { ty, shared: _ } => match ty {
+            wasmparser::HeapType::Abstract { ty, shared: false } => match ty {
                 wasmparser::AbstractHeapType::Extern => WasmHeapType::Extern,
                 wasmparser::AbstractHeapType::NoExtern => WasmHeapType::NoExtern,
                 wasmparser::AbstractHeapType::Func => WasmHeapType::Func,
@@ -1752,6 +1752,7 @@ pub trait TypeConvert {
                     unimplemented!("unsupported heap type {ty:?}");
                 }
             },
+            _ => unimplemented!("unsupported heap type {ty:?}"),
         }
     }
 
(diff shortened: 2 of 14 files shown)


crates/types/src/lib.rs
```diff
@@ -1735,21 +1735,23 @@ pub trait TypeConvert {
...
-            wasmparser::HeapType::Eq => WasmHeapType::Eq,
-            wasmparser::HeapType::I31 => WasmHeapType::I31,
-            wasmparser::HeapType::Array => WasmHeapType::Array,
-            wasmparser::HeapType::Struct => WasmHeapType::Struct,
-            wasmparser::HeapType::None => WasmHeapType::None,
-
-            wasmparser::HeapType::Exn | wasmparser::HeapType::NoExn => {
-                unimplemented!("unsupported heap type {ty:?}");
-            }
+            wasmparser::HeapType::Abstract { ty, shared: _ } => match ty {
```
alexcrichton, line 1739: Mind matching `shared: false` here and running `unimplemented!` on `shared: true`?

crates/explorer/src/lib.rs
```diff
@@ -66,9 +66,10 @@ struct AnnotatedWatChunk {
 }
 
 fn annotate_wat(wasm: &[u8]) -> Result<AnnotatedWat> {
-    let mut printer = wasmprinter::Printer::new();
+    let printer = wasmprinter::Config::new();
+    let mut storage = String::new();
     let chunks = printer
-        .offsets_and_lines(wasm)?
+        .offsets_and_lines(wasm, &mut storage)?
```
alexcrichton, line 72: Looks good to me 👍

crates/wasmtime/src/runtime/coredump.rs
```diff
@@ -183,7 +183,10 @@ impl WasmCoreDump {
 
                         HeapType::Any => wasm_encoder::ValType::Ref(wasm_encoder::RefType {
                             nullable: true,
-                            heap_type: wasm_encoder::HeapType::Any,
+                            heap_type: wasm_encoder::HeapType::Abstract {
+                                ty: wasm_encoder::AbstractHeapType::Any,
+                                shared: false,
+                            },
                         }),
```
alexcrichton, line 190: While you're here mind updating this to the [`RefType::ANYREF`](https://docs.rs/wasm-encoder/latest/wasm_encoder/struct.RefType.html#associatedconstant.ANYREF) shortcut?

alexcrichton: (same for the changes below)

crates/wast/src/core.rs
```diff
@@ -14,8 +14,14 @@ pub fn val<T>(store: &mut Store<T>, v: &WastArgCore<'_>) -> Result<Val> {
         F32(x) => Val::F32(x.bits),
         F64(x) => Val::F64(x.bits),
         V128(x) => Val::V128(u128::from_le_bytes(x.to_le_bytes()).into()),
-        RefNull(HeapType::Extern) => Val::ExternRef(None),
-        RefNull(HeapType::Func) => Val::FuncRef(None),
+        RefNull(HeapType::Abstract {
+            ty: AbstractHeapType::Extern,
+            ..
```
alexcrichton, line 19: Mind matching `shared: false` here explicitly and panicking/bailing on `shared: true`?

alexcrichton (approved):
Minor things here and there but otherwise 👍

alexcrichton:
Ok now that's something I wasn't expecting. Changing the crate dependency graph is changing `-Cmetadata` to the adapter which is causing the symbols to change. For now that can be fixed by rebuilding the adapters locally but I'll try to work up a fix tomorrow to frob the name section to avoid the hashes being there.

alexcrichton:
After poking at this a bit more I think we may want to instead revert the inclusion of binaries in-tree -- https://github.com/bytecodealliance/wasmtime/pull/8856 -- which if that lands then this would be unblocked.

PR #8839, opened 2024-06-19
x64: Implement some minor optimizations related to SIMD lowerings

alexcrichton:
These help a benchmark I was taking a look at by ~7%. Not major but seemed like nice and relatively easy cleanups.


abrown (approved):
Makes sense to me from the CLIF output tests though I'm a bit surprised this wasn't already the case (?). The nest of ISLE rules are getting harder to read (in general, not due to this PR)...

Issue #8847, opened 2024-06-20
riscv64: Panic using the Zicond extension


alexcrichton:
This input CLIF:

<details>

```clif


;; Compile test case

test compile
set bb_padding_log2_minus_one=4
set enable_alias_analysis=false
set enable_llvm_abi_extensions=true
set machine_code_cfg_info=true
set enable_jump_tables=false
set enable_heap_access_spectre_mitigation=false
target riscv64 has_zcd has_zbkb has_zbc has_zbs has_zicond has_zvl32b has_zvl64b has_zvl128b has_zvl1024b has_zvl2048b has_zvl4096b has_zvl8192b has_zvl16384b has_zvl32768b

function u1:0() tail {
    ss0 = explicit_slot 50, align = 512
    ss1 = explicit_slot 47, align = 4
    ss2 = explicit_slot 34, align = 32
    ss3 = explicit_slot 103, align = 1024
    ss4 = explicit_slot 110, align = 512
    ss5 = explicit_slot 126, align = 512
    sig0 = (f32, f64, f64, f32, i8 uext, i128, i8 uext, i32, i16 sext, i64, i64 sext, i128, i8 sext, i8, i64, i64 sext) -> i16 sext, i64 sext, f64, i32 sext, f64, i8 sext, i64 sext, f32
    sig1 = () system_v
    sig2 = (i128, i16 sext, i128 sext, i32 sext, i16, i64 uext, f32, i8 sext, f32, i8, i64, i64, i64, i64 uext, f64) -> f64, i8 uext, f32, i128, i64 uext, i8, i16 sext, i64 sext tail
    sig3 = () -> i8 sext, f32, i128, i32, f32, i128 uext, i8, i8 uext, f64, i8 sext, f32 system_v
    sig4 = (i8, i16, i64 sext, i64 sext, i128 uext, i8, i32, f64, i32, f32, i128 uext, i8, i8 uext, f64, f64) -> i8 uext system_v
    sig5 = (i8 sext, i64 uext, i16 sext, i64 sext, i128 uext, i128 sext, f32, i16 uext, i64 sext, i32 sext, i64, i64 uext, f64, f64, i16 sext) -> f32 tail
    sig6 = (i64 sext, i64 sext, i64 sext, i64 sext, i64 sext, i64 sext, i64 sext, i8 uext, i16 uext, i16, i64 sext, i64 sext, i128 uext, i8 sext, f32) tail
    sig7 = (f32) -> f32 system_v
[...]

PR #8850, opened 2024-06-20
riscv64: Update `Inst::worst_case_size`

afonso360:
👋 Hey,

This PR updates the `Inst::worst_case_size` size for the RISC-V backend. The panic that happens in #8847 is entirely due to the `return_call_indirect` generating too many bytes.

I found it difficult to add an automatic calculation of the worst possible size for that instruction to the test that we have, so I attempted to manually calculate the worst case size and used that.

The two test cases here are the original test case, and a minimized version without zicond. I'm not entirely sure why it bisects to the ZiCond PR (https://github.com/bytecodealliance/wasmtime/pull/8695), but having both test cases is not too much of a burden, so might as well.

The increased worst case size now causes an island to be emitted in the `return-call.clif` test, which are the changes for that test in this PR.

### DIFF
--- a/cranelift/codegen/src/isa/riscv64/inst/mod.rs
+++ b/cranelift/codegen/src/isa/riscv64/inst/mod.rs
@@ -844,8 +844,8 @@ impl MachInst for Inst {
     }
 
     fn worst_case_size() -> CodeOffset {
-        // calculate by test function riscv64_worst_case_instruction_size()
-        124
+        // Our worst case size is determined by the riscv64_worst_case_instruction_size test
+        168
     }
 
     fn ref_type_regclass(_settings: &settings::Flags) -> RegClass {
--- a/cranelift/codegen/src/isa/riscv64/abi.rs
+++ b/cranelift/codegen/src/isa/riscv64/abi.rs
@@ -855,7 +855,7 @@ fn compute_clobber_size(clobbers: &[Writable<RealReg>]) -> u32 {
     align_to(clobbered_size, 16)
 }
 
-const DEFAULT_CLOBBERS: PRegSet = PRegSet::empty()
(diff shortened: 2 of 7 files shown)


cranelift/codegen/src/isa/riscv64/inst/mod.rs
```diff
@@ -844,8 +844,12 @@ impl MachInst for Inst {
...
 
     fn worst_case_size() -> CodeOffset {
-        // calculate by test function riscv64_worst_case_instruction_size()
-        124
+        // Currently our largest possible instruction is return_call_indirect.
+        // Depending on the number of clobbers, it can generate up to 172 bytes.
+        //
+        // We also have a test case that tests some other instructions which have
+        // the potential to generate a lot of bytes
+        172
```
cfallin, line 852: I agree it's theoretically hard to enumerate all the different possible `Inst`s` and programmatically test the worst case -- but could we add a test at least that manually constructs what we think is the worst-case `Inst` (`return_call_indirect` with all clobbers set, etc) and asserts its size is 172? Size tests like this for large pseudoinsts plus a social norm of adding them for new large pseudoinsts will at least give partial coverage for this issue I think...

cfallin (approved):
Thanks! Suggestion for a test but otherwise LGTM.

alexcrichton:
> I'm not entirely sure why it bisects to the ZiCond PR (https://github.com/bytecodealliance/wasmtime/pull/8695)

In retrospect this is my fault. The test case in #8847 uses `has_zicond` so during bisection I marked "unknown feature `has_zicond`" as "good" which pointed to the ZiCond PR. I suspect if I had just removed the zicond feature itself I would have found a different point of bisection.

Sorry about that, should have dug in a bit further myself! I naively assumed that the bug was in the zicond instructions added but in retrospect I see how that doesn't make sense

PR #8851, opened 2024-06-20
riscv64: Add support for `bitcast.i128` with a `i128` argument 

afonso360:
👋 Hey,

This PR fixes an issue discovered by fuzzing. #8692 expanded our support for bitcasts, and also enabled them in the cranelift fuzzer for the RISC-V backend.

It turns out that `bitcast.i128` with a `i128` argument wasn't supported in the backend, but was technically legal CLIF.

This PR adds that lowering and a testcase to ensure this doesn't regress.


elliottt (approved):
Looks good!

PR #8853, opened 2024-06-20
wasmtime: Remove redundant epoch check on function entry

jameysharp:
The epoch interruption implementation caches the current deadline in a register, and avoids reloading that cache until the cached deadline has passed.

However, the first epoch check happens immediately after the cache has been populated on function entry, so there's never a reason to reload the cache at that point. It only needs to be reloaded in loops. So this commit eliminates the double-check on function entry.

When Cranelift optimizations are enabled, the alias analysis pass correctly detected that this load was redundant, and the egraph pass optimized away the `icmp` as well. However, since we don't do conditional constant propagation, the branch couldn't be optimized away. On x86 this lowered to a redundant `cmp`/`jae` pair of instructions in every function entry, which this commit eliminates.

To keep track of what code we're generating for epoch interruptions, I've also added disas tests with a trivial infinite loop.

### DIFF
--- a/crates/cranelift/src/func_environ.rs
+++ b/crates/cranelift/src/func_environ.rs
@@ -510,7 +510,8 @@ impl<'module_environment> FuncEnvironment<'module_environment> {
 
     fn epoch_function_entry(&mut self, builder: &mut FunctionBuilder<'_>) {
         builder.declare_var(self.epoch_deadline_var, ir::types::I64);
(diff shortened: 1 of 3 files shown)


alexcrichton (approved):

cfallin:
Ah, I've wondered where that stray branch came from but never tracked it down -- thanks for finding this!

PR #8858, opened 2024-06-21
Reduce the size of the wasp1 component adapter by 4.6k

bjorn3:
This reduces the size of wasi_snapshot_preview1.command.wasm from 79625 bytes to 75029 bytes for a total win of 4596 bytes. Of this reduction enabling LTO is responsible for 3103 bytes, while enabling bulk-memory is responsible for 1493 bytes.

### DIFF
--- a/ci/build-wasi-preview1-component-adapter.sh
+++ b/ci/build-wasi-preview1-component-adapter.sh
@@ -1,6 +1,10 @@
 #!/usr/bin/env bash
 set -ex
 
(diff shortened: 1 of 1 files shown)


alexcrichton (approved):

alexcrichton:
Thanks!

bjorn3:
For future reference: https://bytecodealliance.zulipchat.com/#narrow/stream/219900-wasi/topic/wasi-preview1-component-adapter.20size

Issue #8312, opened 2024-04-07
Support `f16` and `f128`


tgross35:
Thanks for filing a feature request! Please fill out the TODOs below.

#### Feature

Add support for 16-bit and 128-bit IEEE floating point numbers.

#### Benefit

This will allow using these types with `rustc_codegen_clif`, as described at https://github.com/rust-lang/rustc_codegen_cranelift/issues/1461

#### Implementation

Unsure

#### Alternatives

Unsure, maybe it would be possible to stub out these types if adding them is difficult.

jameysharp:
Adding the types is pretty easy, I think, and extending CLIF floating-point instructions to allow the new types is also pretty easy. Successfully compiling anything that uses the types is the hard part, I imagine.

I think the places to add the types are:
- https://github.com/bytecodealliance/wasmtime/blob/e6f9ca5efa0c4bf59c9cf133d9abeaab63ec07bb/cranelift/codegen/meta/src/cdsl/typevar.rs#L646
- https://github.com/bytecodealliance/wasmtime/blob/e6f9ca5efa0c4bf59c9cf133d9abeaab63ec07bb/cranelift/codegen/meta/src/shared/types.rs#L44-L73
- various places in https://github.com/bytecodealliance/wasmtime/blob/e6f9ca5efa0c4bf59c9cf133d9abeaab63ec07bb/cranelift/codegen/src/ir/types.rs

Once the tests are passing again after that, I think you'll have a compiler that will accept f16/f128 instructions. But if you use them, then it will fail later: during lowering on all backends; in the interpreter; in compiletests and runtests; and in fuzzing.

I think this is a fairly large project, but if rustc needs these types, I think it's reasonable to add the types in one PR and then work on adding implementations for them everywhere over time.

PR #8860, opened 2024-06-21
Initial `f16` and `f128` support

beetrees:
Issue #8312

This PR adds initial basic support for `f16` and `f128` to Cranelift. This is enough to allow the interpreter to bitcast to and from `f16`s and `f128`s, as shown in the added filetest.

### DIFF
--- a/cranelift/codegen/meta/src/cdsl/types.rs
+++ b/cranelift/codegen/meta/src/cdsl/types.rs
@@ -136,6 +136,12 @@ impl LaneType {
     /// Return a string containing the documentation comment for this lane type.
     pub fn doc(self) -> String {
         match self {
+            LaneType::Float(shared_types::Float::F16) => String::from(
+                "A 16-bit floating point type represented in the IEEE 754-2008
+                *binary16* interchange format. This corresponds to the :c:type:`_Float16`
+                type in most C implementations.
+                WARNING: f16 support is a work-in-progress and is incomplete",
+            ),
             LaneType::Float(shared_types::Float::F32) => String::from(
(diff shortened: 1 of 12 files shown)


cranelift/codegen/meta/src/cdsl/types.rs
```diff
@@ -168,13 +180,15 @@ impl LaneType {
...
-                LaneType::Float(shared_types::Float::F64) => 12,
+                LaneType::Int(shared_types::Int::I8) => 4,
+                LaneType::Int(shared_types::Int::I16) => 5,
+                LaneType::Int(shared_types::Int::I32) => 6,
+                LaneType::Int(shared_types::Int::I64) => 7,
+                LaneType::Int(shared_types::Int::I128) => 8,
+                LaneType::Float(shared_types::Float::F16) => 9,
+                LaneType::Float(shared_types::Float::F32) => 10,
+                LaneType::Float(shared_types::Float::F64) => 11,
+                LaneType::Float(shared_types::Float::F128) => 12,
```
jameysharp, line 191: I don't remember exactly how we defined the bit-packed type IDs. Why is renumbering these lane types okay? Do we not use `4`/`5` somewhere else?

jameysharp (approved):
This looks great!

I didn't check the `Ieee16`/`Ieee128` implementations super carefully and it would be nice to find ways to deduplicate the code for those four `Ieee*` types a bit, but based on the tests I'm confident enough to merge these.

I have a couple quick questions but overall I think this covers everything necessary to get started implementing these types. Thank you!

jameysharp:
Awesome, thank you!

PR #8867, opened 2024-06-24
Rename the `wasm32-wasi` target to `wasm32-wasip1`

alexcrichton:
This rename is happening in upstream Rust and should be in enough places now.

prtest:full


alexcrichton:
Ok a full CI run confirms a fear I had which is that our current MSRV of 1.77.0 doesn't support `wasm32-wasip1`. I'll put this PR on ice and try to re-land once Rust 1.80 is released.

alexcrichton:
Ok with our MSRV now bumped I think this should be ready to land

abrown:
This made me think maybe we want to update this naming in wasi-sdk: https://github.com/WebAssembly/wasi-sdk/pull/459?

pchickey (approved):

PR #8872, opened 2024-06-25
Upstream wasm-wave instances in wasmtime

pchickey:
This PR is upstreaming instances of the `wasm-wave` crate's traits for wasmtime values. However, we won't be able to land this PR in wasmtime until `wasm-wave` is written entirely in safe rust, which it presently is not, because it uses `logos` to derive its lexer here: https://github.com/bytecodealliance/wasm-tools/blob/main/crates/wasm-wave/src/lex.rs#L11

I hadn't looked into how `logos` works prior to making this PR, but when I got to the `cargo vet` for the `logos` crate, I discovered that logos derive macro generates significant amounts of `unsafe` rust. While I don't have any evidence `logos` is unsound, but it seems too risky to me to accept it into our audits, since I also have no reasonable way of ruling out that `logos` is unsound. In the particular case of `wasm-wave`, we should expect that strings accepted by the crate come from untrusted sources, so we want to be very rigorous on how they are parsed into values.

The particular bits of Lann's root wasm-wave repo this PR is upstreaming are found in https://github.com/lann/wasm-wave/tree/main/src/wasmtime. I was able to eliminate the `struct FuncType` definition in that code by defining `component::Func::ty` [5918a4b](https://github.com/bytecodealliance/wasmtime/pull/8872/commits/5918a4ba245bd1ce64834b9901e4edf7faf0069a) tso that the`ComponentFunc` impl of WasmType is usable.


Kmeakin:
I'd be interested in contributing to this, but I'll have to check with my employer first since it is under a separate repo. 

In the meantime, you can use the `forbid-unsafe` feature from Logos: https://logos.maciej.codes/unsafe.html

pchickey:
Thanks! Yes, the plan is to merge this by enabling forbid-unsafe logos over in wasm tools, however I just haven't taken the time to get back to this because it fell off my radar. I don't think we should write a new lexer.

pchickey:
Once https://github.com/bytecodealliance/wasm-tools/pull/1874 lands and is in a wasm-tools release, I'll update this PR to use it.

pchickey:
@alexcrichton Can you please double-check my safe-to-deploy audit of https://diff.rs/beef/0.5.2/0.5.2 as part of reviewing this PR?

alexcrichton (approved):
Looks good to me 👍 

To help with version alignment of the `wasm_wave` crate, how about:

* Reexporting the `wasm_wave` crate somewhere (e.g. `wasmtime::component::wasm_wave`)
* Adding some top-level conveniences (e.g. `Val::{from_wave,to_wave}`)

and the docs for those helper methods could point to the `wasm_wave` reexport and additionally to the wave docs?

pchickey:
Suggestions done. Thanks!

PR #8874, opened 2024-06-26
wasi-adapter: Implement provider crate that embeds the adapter binaries [v2]

juntyr:
Follow-up to #8792, which was reverted in #8856. The new design follows @alexcrichton's suggestions in https://github.com/bytecodealliance/wasmtime/pull/8856#issuecomment-2182798604.

The provider repo is materialised into the workspace to ensure that it is built with the same version, lints, etc as the other crates.

- [x] repo contains a template for the provider crate
- [x] template crate is materialised in CI to be tested
- [x] template crate is materialised just before publishing

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -826,16 +826,9 @@ jobs:
       with:
         submodules: true
     - uses: ./.github/actions/install-rust
-    - run: rustup component add rustfmt clippy
-
-    - uses: actions/download-artifact@v4
+    - uses: ./.github/actions/build-adapter-provider
       with:
-        name: bins-wasi-preview1-component-adapter
-        path: crates/wasi-preview1-component-adapter/provider/artefacts
-    
-    - run: ./ci/build-wasi-preview1-component-adapter-provider.sh
-      env:
-        CHECK: 1
+        run-id: ${{ github.run_id }}
     
 
     # common logic to cancel the entire run if this job fails
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -794,9 +794,7 @@ jobs:
       with:
         submodules: true
     - uses: ./.github/actions/install-rust
-    - run: |
-        rustup target add wasm32-wasi wasm32-unknown-unknown
-        rustup component add rustfmt clippy
+    - run: rustup target add wasm32-wasi wasm32-unknown-unknown
 
     - name: Install wasm-tools
       run: |
@@ -812,6 +810,33 @@ jobs:
         name: bins-wasi-preview1-component-adapter
         path: target/wasm32-unknown-unknown/release/wasi_snapshot_preview1.*.wasm
 
+
+    # common logic to cancel the entire run if this job fails
+    - run: gh run cancel ${{ github.run_id }}
+      if: failure() && github.event_name != 'pull_request'
+      env:
+        GH_TOKEN: ${{ github.token }}
+  
+  build-preview1-component-adapter-provider:
+    name: Build wasi-preview1-component-adapter-provider
+    needs: build-preview1-component-adapter
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        submodules: true
+    - uses: ./.github/actions/install-rust
+    - run: rustup component add rustfmt clippy
+
+    - uses: actions/download-artifact@v4
+      with:
+        name: bins-wasi-preview1-component-adapter
+        path: crates/wasi-preview1-component-adapter/provider/artefacts
+    
+    - run: ./ci/build-wasi-preview1-component-adapter-provider.sh
+      env:
+        CHECK: 1
+    
 
     # common logic to cancel the entire run if this job fails
     - run: gh run cancel ${{ github.run_id }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1105,6 +1105,7 @@ jobs:
       - determine
       - miri
       - build-preview1-component-adapter
+      - build-preview1-component-adapter-provider
       - build-wasmtime-target-wasm32
       - test-min-platform-example
       - check_js
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -794,7 +794,9 @@ jobs:
       with:
         submodules: true
     - uses: ./.github/actions/install-rust
-    - run: rustup target add wasm32-wasi wasm32-unknown-unknown
+    - run: |
+        rustup target add wasm32-wasi wasm32-unknown-unknown
+        rustup component add rustfmt clippy
 
     - name: Install wasm-tools
       run: |
--- a/.github/actions/build-adapter-provider/action.yml
+++ b/.github/actions/build-adapter-provider/action.yml
@@ -1,0 +1,23 @@
+name: 'Build wasi component adapter provider'
+description: 'Build the wasi-preview1-component-adapter provider using the adapter artefacts'
+
+inputs:
+  run-id:
+    description: 'run id of the main.yml action that produced the adapter artefacts'
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - uses: actions/download-artifact@v4
+      with:
+        name: bins-wasi-preview1-component-adapter
+        path: crates/wasi-preview1-component-adapter/provider/artefacts
+        github-token: ${{ github.token }}
+        run-id: ${{ inputs.run-id }}
+
+    - run: rustup component add rustfmt clippy
+
+    - name: Build and checl the adapter provider
+      shell: bash
+      run: ./ci/build-wasi-preview1-component-adapter-provider.sh
--- a/.github/actions/build-adapter-provider/action.yml
+++ b/.github/actions/build-adapter-provider/action.yml
@@ -16,7 +16,9 @@ runs:
         github-token: ${{ github.token }}
         run-id: ${{ inputs.run-id }}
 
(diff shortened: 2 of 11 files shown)


alexcrichton (commented):
Thanks for doing this! My biggest worry with this though is that there's nontrivial logic which will only get exercised once-a-month at a time and place where no one's really watching CI for failures and it's difficult to fix. Notably this is the `publish-to-cratesio.yml` workflow with logic there.

The viability of this approach of synthesizing a crate on-the-fly to publish in my mind hinges on that being as simple as possible. For example the `sed` command there is duplicated with the one in `build-wasi-preview1-component-adapter.sh`. I understand why it's duplicated, but it's increasing the risk of a publish-only failure that isn't otherwise caught during development.

One example is that `publish.rs` runs in "verify" mode on all PRs and only does publication errors later on. This has caught a significant number of errors where they would have only otherwise appeared later on during publication. 

I'll admit though that I'm at a bit of a loss of how best to do this. Without native support in Cargo for something like this I've never come across a great way to orchestrate this. Here's some loose thoughts though:

* Could `publish.rs` perhaps be leveraged for this? That way delta in logic-on-each PR and logic-on-publish is as small as possible.
* I like the idea of downloading the adapter binaries as artifacts rather than rebuilding them as it guarantees they're the same as the published ones. Downloading from the tag though feels like it might be brittle and run a risk of racing with the job that's uploading artifacts per-tag. Could the artifact instead be downloaded with the `actions/download-artifact` action? Also, could the logic to download this and package it be shared with the per-PR CI too?

Also ideally this could be done with a "dry run" of sorts to ensure that everything actually works once we hit the publication of all this. Would you be up for getting this running on your fork of Wasmtime for example? You'll probably have to comment/tweak some other things to not publish to crates.io for example, though.

Sorry I understand I'm asking for a fair bit here. Much of this to me stems from Cargo not having native support for a feature like this and otherwise I don't think there's a non-brittle way to orchestrate this.

juntyr:
I absolutely understand your concerns.

> I like the idea of downloading the adapter binaries as artifacts rather than rebuilding them as it guarantees they're the same as the published ones. Downloading from the tag though feels like it might be brittle and run a risk of racing with the job that's uploading artifacts per-tag. Could the artifact instead be downloaded with the actions/download-artifact action? Also, could the logic to download this and package it be shared with the per-PR CI too?

That's a good idea! I don't have experience with workflow artefacts yet but I'll see how they work.

> Could publish.rs perhaps be leveraged for this? That way delta in logic-on-each PR and logic-on-publish is as small as possible.

That would be one option. Alternatively, we could have a common CI script, which is invoked in the normal CI and during publish, which downloads the artefact and does rustfmt+check+clippy+package checks (skipped during publish, so this would be a codepath that's not exercised during publish)? Then the publish action would call the same script as CI, and only the cargo publish to crates.io would only happen in the publish action.

If `cargo publish --dry-run` can execute without credentials, we could also run that in CI as well.

> Also ideally this could be done with a "dry run" of sorts to ensure that everything actually works once we hit the publication of all this. Would you be up for getting this running on your fork of Wasmtime for example? You'll probably have to comment/tweak some other things to not publish to crates.io for example, though.

I hadn't thought of this, but yes of course! In theory, adding `--dry-run` into the publish.rs script (and to the provider crate publish step) should be sufficient? Once we have the scripts in a good state, I'll do a run-through.

I hope I can implement your ideas tomorrow and then do a test run on Friday, so there is enough time before the next publish cycle kicks off (if things take longer, we'll just take the next train)

juntyr:
I tried to run the full publish cycle in my own repo but have come across some issues (I'm yet at the stage where I can actually test the adapter):

- the publish artefact actions always fail: e.g. https://github.com/juntyr/wasmtime/actions/runs/9695441124/job/26755206220
- my latest release attempt failed without any logs at all: https://github.com/juntyr/wasmtime/actions/runs/9695801920/job/26756378300

The tests were run on top of https://github.com/juntyr/wasmtime/pull/10, which includes the extra commit https://github.com/juntyr/wasmtime/pull/10/commits/cbb0a0a074a54781bcccdb2d4cf4ba9e6397fe37 to run the full tests / publish cycle in my repo and without actually executing any cargo publish.

Do you have any ideas? Also, what do you think of the current implementation?

.github/workflows/main.yml
```diff
@@ -811,6 +811,33 @@ jobs:
...
+    - run: rustup component add rustfmt clippy
+
+    - uses: actions/download-artifact@v4
+      with:
+        name: bins-wasi-preview1-component-adapter
+        path: crates/wasi-preview1-component-adapter/provider/artefacts
+    
+    - run: ./ci/build-wasi-preview1-component-adapter-provider.sh
+      env:
+        CHECK: 1
```
alexcrichton, line 838: Could these steps be pulled into a composite action as well? That way this could be shared with the publish step, notably the configuration of the `download-artifact` step. Also I think it's ok to remove the `CHECK` variable and run the same checks on publication since it's a quick-to-build crate and we can throw some more tools like clippy on the publish step.

In theory that means that the publish step is:

1. Run the composite action to figure out the run-id 
2. Run the composite action to setup the artifacts and perform checks
3. Run a publish

That feels like a good state to me -- aka I like this refactoring 👍

alexcrichton (commented):
> my latest release attempt failed without any logs at all

For this I check the [summary page](https://github.com/juntyr/wasmtime/actions/runs/9695801920) which shows:

> GitHub Actions has encountered an internal error when running your job.

So I think that's just a transient error.

> Also, what do you think of the current implementation?

I'm liking how it's shaping up! I left a few comments related to composite actions how I think this can refactor out a bit more, but otherwise looks good 👍

.github/workflows/main.yml
juntyr: So this would be two new composite actions?

alexcrichton: Yeah, one for acquiring the run-id matching `github.sha` used during publish-artifacts and publish-to-cratesio. Another for verifying the adapter shared between main.yml and publish-to-cratesio

juntyr:
@alexcrichton I've slowly been getting further and further into the try runs and I now have one running that I hope will work.

The extra commit to enable the try is https://github.com/juntyr/wasmtime/pull/19/commits/f1ad3a341bca34957dc543119a8046048269ebe9 - the issue with the artifact action was the wasmtime repo seems to have the artifacts under the hash of the merge commit (maybe because of the merge queue), which I cannot reproduce in my repo, so I just pick *a* run since it doesn't matter for the try right now.

I'll report back once https://github.com/juntyr/wasmtime/actions/runs/9715657045 (and the follow-up publish action has been completed) - does the implementation look good already or is there anything else that should be changed?

juntyr:
Just using any run id for testing perhaps wasn't such a good idea (I think the workflow should work but right now it chooses a run id without artifacts ... I'll try to see if I can get lucky, otherwise I need a better way to filter the runs so that we can demonstrate that the publish action works)

juntyr:
I feel like GitHub has started to rate limit my repo as the CI action just no longer even starts, without any notice. I'll keep an eye on this tonight, but I'm not sure my test run can do much more tonight, I'm sorry

juntyr:
@alexcrichton I got a full release process run to succeed:

- https://github.com/juntyr/wasmtime/pull/22 is the PR which includes this PR and the extra commit https://github.com/juntyr/wasmtime/pull/22/commits/0f9c4516adabc9d65cef0da7d45699a86044f260 to allow CI to run in my repo (and to pick *any* run id instead of the sha of the merged commit one since that one doesn't work in my repo, my guess is because it doesn't have a merge queue, but we already know it works in the main repo and at the latest this will be caught immediately after merging the PR when the publish artifacts workflow runs)
- https://github.com/juntyr/wasmtime/pull/24 is the release cut PR (full CI success in the PR; publish artifact action after merge fails in https://github.com/juntyr/wasmtime/actions/runs/9721774907/job/26834892676 after successfully completing my code, since I don't have GitHub pages set up in my fork)
- https://github.com/juntyr/wasmtime/pull/25 is the release latest PR (full CI success in the PR)
- https://github.com/juntyr/wasmtime/actions/runs/9722005042 is the full release CI run on the to-be-released branch (all succeed, needed a 2nd run since I forgot to remove the release tag from a previous run and so it succeeded but didn't trigger the following steps in my first attempt)
- https://github.com/juntyr/wasmtime/actions/runs/9722304795 is the publish artifacts run on the released branch (all succeed)
- https://github.com/juntyr/wasmtime/actions/runs/9722304796 is the publish to crates.io run on the released branch (all succeed, of course this one is modified to only do dry runs)

I hope the implementation and full try-run of the release process are now good to go :)

juntyr:
Is there any chance this PR could still be reviewed before the v23 cutoff?

tschneidereit:
@juntyr Alex is on vacation this week, so the review will take a bit longer still. I can't really comment on what that'll do to the timing relative to the release, however

juntyr:
> @juntyr Alex is on vacation this week, so the review will take a bit longer still. I can't really comment on what that'll do to the timing relative to the release, however

Ok thanks - I hope Alex has a refreshing holiday :) I'll be on holiday myself for the two weeks after that, so perhaps we can finalize and merge the PR in the v24 cycle

alexcrichton (approved):
Ok thanks again for this! Additionally thanks for proving this out with some runs in your own repo, it's much appreciated! I'll backport this to 23.0.0 so we can see how the release goes while it's fresh in our heads

juntyr:
@alexcrichton Thank you so much for your guidance and I'm excited to (hopefully) be released with v23 soon :D

PR #8876, opened 2024-06-26
Cranelift: Take user stack maps through lowering and emission

fitzgen:
Previously, user stack maps were inserted by the frontend and preserved in the mid-end. This commit takes them from the mid-end CLIF into the backend vcode, and then from that vcode into the finalized mach buffer during emission.

During lowering, we compile the `UserStackMapEntry`s into packed `UserStackMap`s. This is the appropriate moment in time to do that coalescing, packing, and compiling because the stack map entries are immutable from this point on.

Additionally, we include user stack maps in the `Debug` and disassembly implementations for vcode, just after their associated safepoint instructions. This allows us to see the stack maps we are generating when debugging, as well as write filetests that check we are generating the expected stack maps for the correct instructions.

### DIFF
--- a/cranelift/codegen/src/isa/aarch64/inst/emit.rs
+++ b/cranelift/codegen/src/isa/aarch64/inst/emit.rs
@@ -2938,12 +2938,12 @@ impl MachInstEmit for Inst {
                 if let Some(s) = stack_map {
                     sink.add_stack_map(StackMapExtent::UpcomingBytes(4), s);
                 }
-                if let Some(s) = user_stack_map {
-                    let offset = sink.cur_offset() + 4;
-                    sink.push_user_stack_map(state, offset, s);
-                }
                 sink.add_reloc(Reloc::Arm64Call, &info.dest, 0);
                 sink.put4(enc_jump26(0b100101, 0));
+                if let Some(s) = user_stack_map {
+                    let offset = sink.cur_offset();
+                    sink.push_user_stack_map(state, offset, s);
+                }
                 if info.opcode.is_call() {
                     sink.add_call_site(info.opcode);
                 }
@@ -2961,12 +2961,12 @@ impl MachInstEmit for Inst {
                 if let Some(s) = stack_map {
                     sink.add_stack_map(StackMapExtent::UpcomingBytes(4), s);
                 }
-                if let Some(s) = user_stack_map {
-                    let offset = sink.cur_offset() + 4;
-                    sink.push_user_stack_map(state, offset, s);
-                }
                 let rn = info.rn;
                 sink.put4(0b1101011_0001_11111_000000_00000_00000 | (machreg_to_gpr(rn) << 5));
+                if let Some(s) = user_stack_map {
+                    let offset = sink.cur_offset();
+                    sink.push_user_stack_map(state, offset, s);
+                }
                 if info.opcode.is_call() {
                     sink.add_call_site(info.opcode);
                 }
--- a/cranelift/codegen/src/isa/aarch64/inst/emit.rs
+++ b/cranelift/codegen/src/isa/aarch64/inst/emit.rs
@@ -651,11 +651,18 @@ fn enc_asimd_mod_imm(rd: Writable<Reg>, q_op: u32, cmode: u32, imm: u8) -> u32 {
 /// State carried between emissions of a sequence of instructions.
 #[derive(Default, Clone, Debug)]
 pub struct EmitState {
-    /// Safepoint stack map for upcoming instruction, as provided to `pre_safepoint()`.
+    /// Safepoint stack map for upcoming instruction, as provided to
+    /// `pre_safepoint()`.
     stack_map: Option<StackMap>,
+
+    /// The user stack map for the upcoming instruction, as provided to
+    /// `pre_safepoint()`.
+    user_stack_map: Option<ir::UserStackMap>,
+
     /// Only used during fuzz-testing. Otherwise, it is a zero-sized struct and
     /// optimized away at compiletime. See [cranelift_control].
     ctrl_plane: ControlPlane,
+
     frame_layout: FrameLayout,
 }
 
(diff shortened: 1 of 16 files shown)


cranelift/codegen/src/isa/aarch64/inst/emit.rs
```diff
@@ -2921,9 +2934,14 @@ impl MachInstEmit for Inst {
                 }
             }
             &Inst::Call { ref info } => {
-                if let Some(s) = state.take_stack_map() {
+                let (stack_map, user_stack_map) = state.take_stack_map();
+                if let Some(s) = stack_map {
                     sink.add_stack_map(StackMapExtent::UpcomingBytes(4), s);
                 }
+                if let Some(s) = user_stack_map {
+                    let offset = sink.cur_offset() + 4;
```
cfallin, line 2942: A little thing but rather than the explicit `+ 4` here, could we put the `push_user_stack_map()` call after the `.put4()` below? That way the order of operations here matches the semantics of "stackmap at end of instruction" more directly.

cranelift/codegen/src/isa/aarch64/inst/emit.rs
```diff
@@ -2939,9 +2957,14 @@ impl MachInstEmit for Inst {
                 }
             }
             &Inst::CallInd { ref info } => {
-                if let Some(s) = state.take_stack_map() {
+                let (stack_map, user_stack_map) = state.take_stack_map();
+                if let Some(s) = stack_map {
                     sink.add_stack_map(StackMapExtent::UpcomingBytes(4), s);
                 }
+                if let Some(s) = user_stack_map {
+                    let offset = sink.cur_offset() + 4;
```
cfallin, line 2965: (likewise here as above)

cranelift/codegen/src/isa/riscv64/inst/emit.rs
```diff
@@ -1127,9 +1141,16 @@ impl Inst {
...
                 }
                 sink.add_reloc(Reloc::RiscvCallPlt, &info.dest, 0);
-                if let Some(s) = state.take_stack_map() {
+
+                let (stack_map, user_stack_map) = state.take_stack_map();
+                if let Some(s) = stack_map {
                     sink.add_stack_map(StackMapExtent::UpcomingBytes(8), s);
                 }
+                if let Some(s) = user_stack_map {
+                    let offset = sink.cur_offset() + 8;
```
cfallin, line 1150: (likewise here as above; coupling feels even a little more brittle since we're depending on the fact we emit two four-byte instructions below)

cranelift/codegen/src/machinst/lower.rs
```diff
@@ -803,9 +803,31 @@ impl<'func, I: VCodeInst> Lower<'func, I> {
...
             self.finish_ir_inst(loc);
 
+            // If the instruction had a user stack map, forward it from the CLIF
+            // to the vcode.
+            if let Some(entries) = self.f.dfg.user_stack_map_entries(inst) {
+                let end = self.vcode.vcode.num_insts();
+                debug_assert!(end > start);
+                for i in start..end {
+                    let iix = InsnIndex::new(i);
+                    if self.vcode.vcode[iix].is_safepoint() {
```
cfallin, line 817: Should we assert that no more than one VCode inst lowered from a given CLIF inst is a safepoint-instruction? (In practice this should be the case as long as safepoints are calls only?) Duplicating the same CLIF-level stackmap for multiple VCode insts does seem correct-ish to me in the presence of a "the safepoint could actually happen at these N points" condition but is still somewhat surprising...

cranelift/codegen/src/machinst/vcode.rs
```diff
@@ -67,6 +67,14 @@ pub struct VCode<I: VCodeInst> {
     /// Lowered machine instructions in order corresponding to the original IR.
     insts: Vec<I>,
 
+    /// A map from backwards instruction index to the user stack map for that
+    /// instruction.
+    ///
+    /// This is a sparse side table that only has entries for instructions that
+    /// are safepoints, and only for a subset of those that have an associated
+    /// user stack map.
+    user_stack_maps: FxHashMap<InsnIndex, ir::UserStackMap>,
```
cfallin, line 76: Would it be possible to define a newtype over InsnIndex (`BackwardInsnIndex`) to make this a little more explicit? I don't think we need to weave it through the rest of lowering and translate out of it when doing the finalize/reverse, necessarily; just want to document in the types that this doesn't go through the same translation as the other index data.

cranelift/codegen/src/machinst/vcode.rs
```diff
@@ -888,13 +904,36 @@ impl<I: VCodeInst> VCode<I> {
...
+                            };
+
+                            let (user_stack_map, user_stack_map_disasm) = {
+                                // The `user_stack_maps` is keyed by reverse
+                                // instruction index, so we must flip the
+                                // index. We can't put this into a helper method
+                                // due to borrowck issues because parts of
+                                // `self` are borrowed mutably elsewhere in this
+                                // function.
+                                let index = InsnIndex::new(self.num_insts() - iix.index() - 1);
```
cfallin, line 924: Given the newtype idea above, could we have a `.to_insn_index(num_insts: usize)` on the `BackwardInsnIndex` type?

cfallin (approved):
I really like this and look forward to seeing the original stackmap support deleted soon! (At that point I suppose we can rename "user stack maps" back to simply "stack maps" again?)

Some thoughts below but overall shape LGTM.

cranelift/codegen/src/machinst/lower.rs
fitzgen: Yeah I went back and forth on this and had that same reasoning about the stack maps being correct even if there are multiple safepoint mach insts, since nothing has moved those values in the middle of this CLIF instruction's lowering. But conservatively asserting there is only one is fine and something we can relax if we ever need to in the future (seems unlikely).

cranelift/codegen/src/machinst/vcode.rs
fitzgen: Done!

cranelift/codegen/src/machinst/vcode.rs
fitzgen: And done!

fitzgen:
Thanks for the review, Chris!

fitzgen:
> I really like this and look forward to seeing the original stackmap support deleted soon! (At that point I suppose we can rename "user stack maps" back to simply "stack maps" again?)

Yeah, next I will get Wasmtime using the new system, and then I will start removing the old system. Not 100% sure if I will spend the time to rename "user stack maps" to "stack maps", as the name kinda makes sense to me, given that these are really provided by the frontend "user" and the "user" is responsible for them (as opposed to being generated by Cranelift itself). But I don't feel strongly about it.

PR #8879, opened 2024-06-27
Wasi adapter preview v2 try

juntyr:


juntyr:
Sorry for the noise, I managed to accidentally file this PR against the main repo instead of my private one (where we want to test the updated release pipeline for the wasi adapter provider)

PR #8882, opened 2024-06-27
Update the wasm-tools family of crates

alexcrichton:
This notably brings in a limitation where component model flags types must have 32 or fewer flags in accordance with the transition plan of https://github.com/WebAssembly/component-model/issues/370. A feature flag is added to go back to the previous behavior to avoid breaking anyone too much.

This additionally brings in a fix for a panic when validating invalid modules with tail calls.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -841,7 +841,7 @@ dependencies = [
  "serde_derive",
  "smallvec",
  "target-lexicon",
- "wasmparser 0.211.1",
+ "wasmparser 0.212.0",
  "wasmtime-types",
  "wat",
 ]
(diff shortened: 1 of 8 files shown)


alexcrichton:
I'll note that this is a draft as I'd like to get in https://github.com/bytecodealliance/wit-bindgen/pull/990 and https://github.com/bytecodealliance/wit-bindgen/pull/989 and update wit-bindgen here too

alexcrichton:
Ok should be reading pending CI

alexcrichton:
On second thought let's not update wit-bindgen.

pchickey (approved):

PR #8883, opened 2024-06-28
closes #2827: remove srclocs from bugpoint

### LINKED ISSUE #2827: Bugpoint: Try removing SourceLoc's
#### Feature

When using bugpoint to reduce clif ir turn something like

```
function u0:0() system_v {
                                block0:
@000c                               v0 = iconst.i32 0
@000c                               trap user0
}
```

into

```
function u0:0() system_v {
block0:
    v0 = iconst.i32 0
    trap user0
}
```

#### Benefit

Makes the reduced clif ir a bit easier to read.

zleyyij:
Prior conversation:
https://github.com/bytecodealliance/wasmtime/issues/2827#issuecomment-2184904073

https://bytecodealliance.zulipchat.com/#narrow/stream/217117-cranelift/topic/Removing.20srclocs.20with.20bugpoint

This change removes sourcelocs from bugpoint output to make the reduced clif IR easier to read.

### DIFF
--- a/cranelift/src/bugpoint.rs
+++ b/cranelift/src/bugpoint.rs
@@ -839,6 +839,8 @@ fn reduce(isa: &dyn TargetIsa, mut func: Function, verbose: bool) -> Result<(Fun
     }
 
     try_resolve_aliases(&mut context, &mut func);
+    // Remove SourceLocs to make reduced clif IR easier to read
+    func.srclocs.clear();
 
     let progress_bar = ProgressBar::with_draw_target(0, ProgressDrawTarget::stdout());
     progress_bar.set_style(
--- a/cranelift/src/bugpoint.rs
+++ b/cranelift/src/bugpoint.rs
@@ -831,6 +831,16 @@ fn try_resolve_aliases(context: &mut CrashCheckContext, func: &mut Function) {
     }
 }
 
+/// Remove sourcelocs if the function still crashes after they are removed, to make the reduced clif IR easier to read.
+fn try_remove_srclocs(context: &mut CrashCheckContext, func: &mut Function) {
+    let mut func_with_removed_sourcelocs = func.clone();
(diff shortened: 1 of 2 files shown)


cranelift/src/bugpoint.rs
```diff
@@ -839,6 +839,8 @@ fn reduce(isa: &dyn TargetIsa, mut func: Function, verbose: bool) -> Result<(Fun
     }
 
     try_resolve_aliases(&mut context, &mut func);
+    // Remove SourceLocs to make reduced clif IR easier to read
+    func.srclocs.clear();
```
fitzgen, line 843: It isn't clear to me that we should remove them all at once. There might be some panic due to the way we are processing a particular series of srclocs. Can we try removing subset(s) of them incrementally? (I have never touched bugpoint fwiw, so I don't have  a sense of the difficulty of what I am asking and if it isn't practical then we can land this as-is).

fitzgen (commented):
Thanks! One suggestion/question below.

Also cc @bjorn3 who I think has a bunch of experience with bugpoint.

cranelift/src/bugpoint.rs
bjorn3: I think the approach used by `try_resolve_aliases` would work here: Try removing all srclocs and see if the bug reproduces. If not revert the change and keep them for the remainder of the run.

zleyyij:
I refactored so that it's done in a manner similar to `try_resolve_aliases`. 

`try_resolve_aliases` appears to be called twice, I don't know why this is, but I currently only have `try_remove_sourcelocs` called once.

bjorn3 (approved):

fitzgen (approved):
Thanks!

bjorn3:
I think something went wrong here. It still isn't merged after a week.

PR #8887, opened 2024-06-29
cranelift/riscv: refactor rv_fma pattern

HerrCai0907:
replace `if-let` with `let`


afonso360 (approved):
Sorry for the delay, this somehow slipped off my radar! LGTM, thanks!

Issue #8602, opened 2024-05-12
x64: Add support for `fmsub` and `fnmsub` instructions


afonso360:
👋 Hey,

#### Feature

Add support for the [`fmsub`](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=vfmsub&ig_expand=3421,3211) and [`fnmsub`](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=vfnmsub&ig_expand=3421,3211,3421) instructions in the x64 backend.

#### Benefit

This improves instruction selection quality for a few special fma cases.

#### Implementation

We already have support for the `fmadd` and `fnmadd` versions of these instructions [here](https://github.com/bytecodealliance/wasmtime/blob/e6f9ca5efa0c4bf59c9cf133d9abeaab63ec07bb/cranelift/codegen/src/isa/x64/lower.isle#L2840-L2864) and [here](https://github.com/bytecodealliance/wasmtime/blob/e6f9ca5efa0c4bf59c9cf133d9abeaab63ec07bb/cranelift/codegen/src/isa/x64/inst.isle#L4341-L4367). So we adding support for these instructions should be a matter of introducing the new opcodes and implementing lowering rules similar to what already exists today.

Something similar was done in #8588 for the RISC-V backend. If anyone needs help implementing this, let me know!

UnlimitedHummus:
Hey, I would like to try to implement this. I might need some help, as this is my first contribution.

If I have questions, should I ask them here or in Zulip?

afonso360:
Awesome! 🎉 

> If I have questions, should I ask them here or in Zulip?

Up to you, I'm available in both. I might be able to give you a faster reply on Zulip since that usually sends me a notification.

aidenfoxivey:
@afonso360 Seems the same here - also complete.

PR #8888, opened 2024-06-30
Add support for fmsub and fnmsub instructions on x86

UnlimitedHummus:
Issue #8602

Allows generating fmsub and fnmsub instructions as mentioned in the issue.


cranelift/codegen/src/isa/x64/lower.isle
```diff
@@ -2860,6 +2860,38 @@
...
+(rule 2 (lower (has_type ty (fma x y (fneg z))))
+      (if-let $true (use_fma))
+      (fmsub ty x y z))
+
+;; fmsub and fnmsub
+(decl fmsub (Type Value Value Value) Xmm)
+(decl fnmsub (Type Value Value Value) Xmm)
+
+;; Base case, will sink a load of `z` automatically.
+(rule (fmsub ty x y z) (x64_vfmsub213 ty x y z))
```
fitzgen, line 2873: I forget what the details of our lowering and ISLE conversions are these days, but does this need to guard on sinkable loads? Or will it force things to registers as written? (In which case the comment should be updated)

cc @jameysharp, who I think would know the answer to this question.

fitzgen (commented):
Thanks! Overall looks good to me, modulo the one comment below.

Can you also add some run tests in addition to the compile tests, to make sure that the runtime behavior is as-expected as well? Since FMA is not quite the same across architectures, it may make sense to put these run tests under `cranelift/filetests/filetests/isa/x64` as well.

cranelift/codegen/src/isa/x64/lower.isle
jameysharp: I'm not actually sure. But I see that this is identical to the existing lowering rules above aside from replacing `add` with `sub` and so on. So it's probably fine?

UnlimitedHummus: I did just work off of the rules for the `add` instructions. However, its probably related to the semantics of `z` matching into `XmmMem` in the rule for `x64vfmsub213`. I didn't look much into how that works, but the first two test cases actually test this rule both in the case of passing a value in a register or a sinkable load. There, the load is definitely sunk.

UnlimitedHummus:
Sure, I'll add the runtime tests later, when I have time.

afonso360:
We already have some run tests that run on all arch's here ([`1`](https://github.com/bytecodealliance/wasmtime/blob/main/cranelift/filetests/filetests/runtests/fma.clif), [`2`](https://github.com/bytecodealliance/wasmtime/blob/main/cranelift/filetests/filetests/runtests/simd-fma-neg.clif), [`3`](https://github.com/bytecodealliance/wasmtime/blob/main/cranelift/filetests/filetests/runtests/simd-fma.clif)). None of them test load sinking, so that would be a nice addition.

cranelift/codegen/src/isa/x64/lower.isle
afonso360: > I forget what the details of our lowering and ISLE conversions are these days, but does this need to guard on sinkable loads?

We do some checks when converting from a `Value` into `XmmMem` where we force the value into a register if it's not eligible for being sunk (i.e. has multiple uses), so this looks okay to me.

afonso360 (approved):
This LGTM! I double checked the new instruction encodings against the intel manual and they all check out. Thanks!

cranelift/codegen/src/isa/x64/lower.isle
fitzgen: Thanks for the clarification @afonso360 and @jameysharp!

UnlimitedHummus:
To add runtime tests for load sinking, I would just duplicate some existing testcases and add loads to them. This doesn't test that the load is actually sunk, but rather that the results are still correct with load sinking. Is that the intention? Or am I missing some different type of testcase I could write?

fitzgen:
That is a good thing to test, but generally I was just asking for runtests that check that the new lowerings that turn multiple clif instructions into a single machine instruction produce the same values that interpreting the individual clif instructions would.

See for example

https://github.com/bytecodealliance/wasmtime/blob/main/cranelift/filetests/filetests/runtests/iabs.clif#L1-L18

although we only care about the interpreter and x86-64 here.

cranelift/filetests/filetests/runtests/x64/fmsub.clif
UnlimitedHummus, line 1: Is this what you had in mind? I didn't find a better way to test this, other than to store the value on the stack and then load it from there again.

UnlimitedHummus: If that's good I will add some more tests of course.

afonso360: Yeah! Just something to ensure that the instruction gets emitted and runs properly. I wouldn't worry about trying to get all of the variations running, that might be too much (In my opinion).

cranelift/filetests/filetests/isa/x64/fmsub-inst.clif
```diff
@@ -478,3 +478,242 @@ block0(v0: f64, v1: f64, v2: f64):
...
+; block1: ; offset 0x8
+;   leaq (%rsp), %r8
+;   vmovss %xmm1, (%r8) ; trap: heap_oob
+;   vfnmsub132ss (%r8), %xmm2, %xmm0 ; trap: heap_oob
+;   addq $0x10, %rsp
+;   movq %rbp, %rsp
+;   popq %rbp
+;   retq
+
+;; negate first, then load
```
UnlimitedHummus, line 598: This could generate more efficient core by selecting vfnmsub132ss instead and skipping the negation.

UnlimitedHummus:
I added some tests, but also identified an optimization I missed. This only affects the case where the negation happens before the load. I think the order of the two shouldn't matter. I still need to look into how to write rules to handle that case.

jameysharp:
> I added some tests, but also identified an optimization I missed. This only affects the case where the negation happens before the load. I think the order of the two shouldn't matter. I still need to look into how to write rules to handle that case.

This PR is already a nice improvement even if it doesn't optimize every case, so: If you want to dig into that you're welcome to, but I'd also be happy to merge this as-is. If you decide not to fix it right now, then it would be great to open a new issue describing what's not working, and include a link to that issue in the comments on the test that illustrates the problem.

That said, this is failing in CI right now. The error looks weird to me and I'm not sure whether you just need to re-run `CRANELIFT_TEST_BLESS=1 cargo test -p cranelift-tools --test filetests` or if there's something stranger going on, so please let us know if you have any trouble.

UnlimitedHummus:
CI failing was due to me annotating a copy of the expected output. I moved my comment before the testcase, which fixed that. I tried adding some rules to improve code generation for this case yesterday, but that didn't work. So I'd have to take a closer look at ISLE to see why it doesn't generate the right code. Not sure when I'll get around to that so I opened an issue describing the missed optimization like you suggested.

PR #8892, opened 2024-07-01
Use AlmaLinux:8 for x86_64-linux binary-compatible-builds

elliottt:
CentOS 7 has hit EOL, and is beginning to cause build failures due to mirror lists being unavailable. This PR switches us over to using AlmaLinux 8 instead.

AlmaLinux 8 has cmake-3.26, so it's reasonable to continue explicitly installing cmake-3.29.

### DIFF
--- a/ci/docker/x86_64-linux/Dockerfile
+++ b/ci/docker/x86_64-linux/Dockerfile
@@ -1,8 +1,8 @@
-FROM centos:7
+FROM almalinux:8
 
-RUN yum install -y git gcc make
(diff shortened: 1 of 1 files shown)


jameysharp (approved):
This makes sense to me.

I think it's worth specifically calling out that AlmaLinux 8 is the current recommended base for Python's [manylinux](https://github.com/pypa/manylinux?tab=readme-ov-file#manylinux), which exists to solve essentially the same problem.

PR #8893, opened 2024-07-02
cranelift: Add `f16const` and `f128const` instructions

beetrees:
This PR adds `f16const` and `f128const` instructions, as well as basic neg/abs/copysign constant propagation. The `f128const` instruction uses a `Constant` (like `vconst`) instead of directly storing the value in the instruction as `InstructionData` has a size of only 16 bytes.

Issue: #8312


afonso360 (approved):
I don't know if @elliottt wants to take a look as well, but this all looks correct to me!

elliottt (approved):
Sorry for the delay, this looks good to me!

Issue #8894, opened 2024-07-02
`wasi-common`: Does `sync` crate feature really rely on `wasmtime` dependency?


Robbepop:
I am writing this issue since I think it is the better place to ask than [my question at Zulip](https://bytecodealliance.zulipchat.com/#narrow/stream/219900-wasi/topic/Does.20.60wasi-common.2Fsync.60.20really.20require.20.60wasmtime.60.3F).

I was looking at `wasi-common`'s `sync` crate feature and it seems to be depending on `wasmtime` as dependency.

I found this to be weird and checked out the repository locally and [just removed the feature dependency line](https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasi-common/Cargo.toml#L82) in the `Cargo.toml` after not really seeing any real dependency on Wasmtime in the `sync` specific code at first glace. Everything still compiled with `cargo build -p wasi-common --no-default-features --features sync`.

Is this simply outdated? It would be great if other WASI runtimes (such as Wasmi) could use `wasi-common` with its `sync` feature to enable WASI and drop the heavily outdated `wasi-cap-std-sync` crate entirely.

This is especially important for Wasmi since for technical reasons Wasmi has to use the super outdated `v2.0.0` of `wasi-cap-std-sync` crate and `dependabot` recently turned my attention to a security issue with `atty` that is used by `wasi-cap-std-sync` version `v2.0.0`.

PR #8900, opened 2024-07-04
`wasi-common`: remove `wasmtime` dependency from `sync` crate feature

Robbepop:
Closes https://github.com/bytecodealliance/wasmtime/issues/8894.

**Motivation:** This allows usage of `wasi-common` in non-Wasmtime Wasm runtimes.

This was probably an oversight in https://github.com/bytecodealliance/wasmtime/pull/7881.

The only optional usage of the `wasmtime` crate that I could find within the `wasi-common/sync` module is the following:
```rust
#[cfg(feature = "wasmtime")]
super::define_wasi!(block_on);
```
Link: https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasi-common/src/sync/mod.rs

Which defines the `add_to_linker` methods for Wasmtime which is entirely optional.
Non Wasmtime users of `wasi-common` might want to define their own `add_to_linker` functions.

Locally I did not see any compile errors with the removal of this dependency of this PR.

### DIFF
--- a/crates/wasi-common/Cargo.toml
+++ b/crates/wasi-common/Cargo.toml
@@ -79,7 +79,6 @@ wasmtime = [
 # This feature enables an implementation of the Wasi traits for a
 # synchronous wasmtime embedding.
 sync = [
-    "wasmtime",
     "dep:cap-fs-ext",
     "dep:cap-time-ext",
     "dep:fs-set-times",


alexcrichton (approved):
Thanks!

PR #8907, opened 2024-07-07
Expose Engine::clone to c-api

CGamesPlay:
This creates an API to allow cloning references to wasm_engine_t in the C API. This is simple in the Rust API which implements Clone, however it is not exposed to the C API.

Use case: in tree-sitter's wasm support, we create a store object and want to instantiate modules into it over the course of the object's lifetime. This requires a handle to the underlying engine, which means our store must be kept alongside a reference to the engine (since tree-sitter is a library, it doesn't want to assume a global wasm engine). This currently requires transferring ownership of the engine to the tree sitter store container, which means we can't reuse engine objects between multiple stores.

This pertains to https://github.com/tree-sitter/tree-sitter/issues/3454.

### DIFF
--- a/crates/c-api/README.md
+++ b/crates/c-api/README.md
@@ -8,7 +8,7 @@ For more information you can find the documentation for this library
 To use Wasmtime from a C or C++ project, you can use Cargo to build the Wasmtime C bindings. From the root of the Wasmtime repository, run the following command:
 
 ```
-cargo build --release wasmtime-c-api
+cargo build --release -p wasmtime-c-api
 ```
 
 This will create static and dynamic libraries called `libwasmtime` in the `target/release` directory.
--- a/crates/c-api/include/wasmtime/engine.h
+++ b/crates/c-api/include/wasmtime/engine.h
@@ -14,6 +14,14 @@ extern "C" {
 #endif
 
 /**
+ * \brief Create a new reference to the same underlying engine.
+ *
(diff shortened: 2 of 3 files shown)


crates/c-api/include/wasmtime/engine.h
```diff
@@ -13,6 +13,11 @@
 extern "C" {
 #endif
 
+/**
+ * \brief Create a new reference to the same underlying engine.
+ */
```
alexcrichton, line 18: Could the documentation be updated here to indicate that `wasm_engine_delete` must be called on the returned reference?

crates/c-api/README.md
```diff
@@ -8,7 +8,7 @@ For more information you can find the documentation for this library
 To use Wasmtime from a C or C++ project, you can use Cargo to build the Wasmtime C bindings. From the root of the Wasmtime repository, run the following command:
 
 ```
-cargo build --release wasmtime-c-api
+cargo build --release -p wasmtime-c-api
```
alexcrichton, line 11: Good catch! We'll need to update this documentation further nowadays because some headers are generated as part of the build now so CMake is primarily needed to get the headers right.

alexcrichton (approved):

alexcrichton (approved):

crates/c-api/README.md
av1934413: 274df2d52a8568f99e608f59046b20a948a15dea

Issue #8784, opened 2024-06-12
Proposal for Configuration File Support in Wasmtime


Mossaka:
Thanks for filing a feature request! Please fill out the TODOs below.

#### Feature

Wasmtime's configurability is flexible such as the pooling allocation strategy or enable/disable parallel compilation. However, this configurations can also be programmatically set using the `wasmtime::Config` struct. I am proposing to add support for a configuration file that Wasmtime can consume.

#### Benefit

Services (such as [containerd wasmtime-shim](https://github.com/spinkube/skips/pull/4)) that uses wasmtime APIs to compile and execute Wasm components today rely on environment variables to configure Wasmtime engine. Whether it is to disable parallel compilation or change the pooling allocation strategy. With the huge number of configuration options that wasmtime provides, it is difficult for users and service providers to maintain these dynamic configuration options. We suggest using a TOML or YAML file for the configuration file due to their readability.

#### Implementation

1. Parsing Configuration File:
Implement a parser to read and deserialize the configuration file into a `wasmtime::Config` struct.

2. Applying Configuration:
Extend the `wasmtime::Config` struct with methods to apply settings from the deserialized configuration.

#### Alternatives

Interestingly, wasmtime CLI has a `config` command in which you can use it to create a new `config`. 
```bash
$ wasmtime config new
Successfully created a new configuration file at '/home/mossaka/.config/wasmtime/config.toml'.
```

However, this config command is used ONLY for cache configuration, which is a tiny fraction of the total configuration options that wasmtime provides.

fitzgen:
Extending the existing cache-config TOML for general Wasmtime options makes sense to me.

We probably want to refactor our config and CLI and (now) TOML code such that there is one source of truth that automatically makes knobs for each configurable thing in all those various places.

pchickey:
This seems like a good idea to me - thanks @Mossaka. Agreed with Nick.

Mossaka:
Thanks, I might get someone to help with this or do it myself.

alexcrichton:
Implemented in https://github.com/bytecodealliance/wasmtime/pull/9811

PR #8917, opened 2024-07-08
Add custom-pages-sizes to CLI flags

alexcrichton:
Enables `-W custom-page-sizes` from the CLI  to run this proposal.

### DIFF
--- a/crates/cli-flags/src/lib.rs
+++ b/crates/cli-flags/src/lib.rs
@@ -260,6 +260,8 @@ wasmtime_option_group! {
         pub function_references: Option<bool>,
         /// Configure support for the GC proposal.
         pub gc: Option<bool>,
+        /// Configure support for the custom-page-sizes proposal.
+        pub custom_page_sizes: Option<bool>,
     }
 
     enum Wasm {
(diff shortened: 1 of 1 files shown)


fitzgen (approved):
We should maybe eventually have a `for_each_config_option` macro to define both the CLI flags and the `Config` methods or something...

alexcrichton:
Agreed yeah, that'd kind fall under https://github.com/bytecodealliance/wasmtime/issues/8784 I think

Mossaka:
> We should maybe eventually have a for_each_config_option macro to define both the CLI flags and the Config methods or something...

I was looking at how the CLI flags and `Config` are defined and it seems like the grouping rules are very different between the two. For CLI, flags are grouped under struct names such as "Optimize", "Codegen", "Debug", etc, while in the Config, fields are grouped to different struct names, such as "Tunables". The lack of the 1-1 mapping makes it hard to define CLI flags from the Config sturct.

Perhaps, adding a macro like `[cli(Optimize)]` or others to Config fields helps to generate these fields in their corresponding CLI structs?

In addition, the methods in the Config are not a 1-1 map to the fields as well... It might be a headache to write the macro that generates the methods according fields.

fitzgen:
I think even something that defines the public API of `Config` and keeps that in sync with the CLI flags, but doesn't define the implementation of those `Config` methods, would be helpful. And that should be possible regardless if the implementation depends on tunables or wasm features or whatever.

PR #8918, opened 2024-07-08
Fix segfault with custom page sizes on aarch64

alexcrichton:
This commit fixes an issue with static memory initialization and custom page sizes interacting together on aarch64 Linux. (is that specific enough?)

When static memory initialization is enabled chunks of memory to initialize the linear memory are made in host-page-size increments of memory. This is done to enable page-mapping via copy-on-write if customized. With the custom page sizes proposal, however, for the first time it's possible for a linear memory to be smaller than this chunk of memory. This means that a virtual memory allocation of a single host page can be made which is smaller than the initialization chunk.

This currently only happens on aarch64 Linux where we coarsely approximate that the host page size is 64k but many hosts run with 4k pages. This means that a 64k initializer is created but the host only allocates 4k for a linear memory. This means that memory initialization can crash when a 64k initializer is copied into a 4k memory.

This was not caught via fuzzing because fuzzing only runs on x86_64. This was not caught via CI because on CI guard pages are disabled entirely on QEMU and we got lucky in that a number of virtual memory allocations were all placed next to each other meaning that this copy was probably corrupting some other memory. Locally this was found by running tests on `main` as-is on AArch64 Linux (by bjorn3).

This commit implements a few safeguards and a fix for this issue:

* On CI with QEMU modestly-size guard pages are now enabled to catch this sooner in development should it happen again in the future.
* An `assert!` is added during memory initialization that the memory copy is indeed valid. This causes the tests to fail as-is on `main` even on x86_64.
* The issue itself is fixed by bailing out of static memory initialization should the host page size exceed the wasm page size which can now happen on aarch64 Linux with smaller page sizes.

### DIFF
--- a/crates/environ/src/compile/module_environ.rs
+++ b/crates/environ/src/compile/module_environ.rs
@@ -996,8 +996,19 @@ impl ModuleTranslation<'_> {
 
         // Validate that the memory information collected is indeed valid for
         // static memory initialization.
-        for info in info.values().filter(|i| i.data_size > 0) {
+        for (i, info) in info.iter().filter(|(_, info)| info.data_size > 0) {
             let image_size = info.max_addr - info.min_addr;
+
+            // Simplify things for now by bailing out entirely if any memory has
+            // a page size smaller than the host's page size. This fixes a case
+            // where currently initializers are created in host-page-size units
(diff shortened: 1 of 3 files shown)


fitzgen (approved):
Nice! Thank you!

PR #8925, opened 2024-07-09
A handful of little tweaks for GC types

fitzgen:
Splitting this out from a larger PR so that it is easier to review and everything can land quicker.

### DIFF
--- a/crates/wasmtime/src/runtime/types.rs
+++ b/crates/wasmtime/src/runtime/types.rs
@@ -1,5 +1,4 @@
 use crate::prelude::*;
-use alloc::borrow::Cow;
 use core::fmt::{self, Display, Write};
 use wasmtime_environ::{
     EngineOrModuleTypeIndex, EntityType, Global, Memory, ModuleTypes, Table, TypeTrace,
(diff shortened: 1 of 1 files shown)


crates/wasmtime/src/runtime/types.rs
```diff
@@ -1241,6 +1294,20 @@ impl StorageType {
...
+    /// returned as-is.
+    ///
+    /// If this is a packed `StorageType::I8` or `StorageType::I16, then a
+    /// `ValType::I32` is returned.
+    pub fn unpack(&self) -> Cow<ValType> {
+        match self {
+            StorageType::I8 | StorageType::I16 => Cow::Owned(ValType::I32),
+            StorageType::ValType(ty) => Cow::Borrowed(ty),
+        }
+    }
```
alexcrichton, line 1309: Could this return `&ValType` by using `const I32: ValType = ValType::I32;` and `&I32` for the i8/i16 case?

alexcrichton (approved):

crates/wasmtime/src/runtime/types.rs
fitzgen: Good call

PR #8933, opened 2024-07-10
Introduce `wasmtime::StructRef` and allocating Wasm GC structs

fitzgen:
This commit introduces the `wasmtime::StructRef` type and support for allocating Wasm GC structs from the host. This commit does *not* add support for the `struct.new` family of Wasm instructions; guests still cannot allocate Wasm GC objects yet, but initial support should be pretty straightforward after this commit lands.

The `StructRef` type has everything you expect from other value types in the `wasmtime` crate:

* A method to get its type or check whether it matches a given type

* An implementation of `WasmTy` so that it can be used with `Func::wrap`-style APIs

* The ability to upcast it into an `AnyRef` and to do checked downcasts in the opposite direction

There are, additionally, methods for getting, setting, and enumerating a `StructRef`'s fields.

To allocate a `StructRef`, we need proof that the struct type we are allocating is being kept alive for the duration that the allocation may live. This is required for many reasons, but a basic example is getting a struct instance's type from the embedder API: this does a type-index-to-`StructType` lookup and conversion and if the type wasn't kept alive, then the type-index lookup will result in what is logically a use-after-free bug. This won't be a problem for Wasm guests (when we get around to implementing allocation for them) since their module defines the type, the store holds onto its instances' modules, and the allocation cannot outlive the store. For the host, we need another method of keeping the object's type alive, since it might be that the host defined the type and there is no module that also defined it, let alone such a module that is being kept alive in the store.

The solution to the struct-type-lifetime problem that this commit implements for hosts is for the store to hold a hash set of `RegisteredType`s specifically for objects which were allocated via the embedder API. But we also don't want to do a hash lookup on every allocation, so we also implement a `StructRefPre` type. A `StructRefPre` is proof that the embedder has inserted a `StructType`'s inner `RegisteredType` into a store. Structurally, it is a pair of the struct type and a store id. All `StructRef` allocation methods require a `StructRefPre` argument, which does a fast store id check, rather than a whole hash table insertion.

I opted to require `StructRefPre` in all allocation cases -- even though this has the downside of always forcing callers to create one before they allocate, even if they are only allocating a single object -- because of two reasons. First, this avoids needing to define duplicate methods, with and without a `StructRefPre` argument. Second, this avoids a performance footgun in the API where users don't realize that they *can* avoid extra work by creating a single `StructRefPre` and then using it multiple times. Anecdotally, I've heard multiple people complain about instantiation being slower than advertised but it turns out they weren't using `InstancePre`, and I'd like to avoid that situation for allocation if we can.

### DIFF
--- a/crates/wasmtime/src/runtime/gc/disabled/structref.rs
+++ b/crates/wasmtime/src/runtime/gc/disabled/structref.rs
@@ -14,7 +14,6 @@ pub enum StructRef {}
 
 impl GcRefImpl for StructRef {}
 
-#[allow(missing_docs, unreachable_code)]
 impl StructRef {
     pub(crate) fn from_cloned_gc_ref(
         _store: &mut AutoAssertNoGc<'_>,
--- a/crates/wasmtime/src/runtime/gc/disabled/structref.rs
+++ b/crates/wasmtime/src/runtime/gc/disabled/structref.rs
@@ -1,0 +1,53 @@
+use crate::runtime::vm::VMGcRef;
+use crate::{
+    store::{AutoAssertNoGc, StoreContextMut, StoreOpaque},
+    AsContext, AsContextMut, GcRefImpl, Result, Rooted, StructType, Val, I31,
+};
+
+/// Support for `StructRefPre` disabled at compile time because the `gc` cargo
+/// feature was not enabled.
+pub enum StructRefPre {}
+
+/// Support for `structref` disabled at compile time because the `gc` cargo feature
+/// was not enabled.
+pub enum StructRef {}
+
+impl GcRefImpl for StructRef {}
+
+#[allow(missing_docs, unreachable_code)]
+impl StructRef {
+    pub(crate) fn from_cloned_gc_ref(
+        _store: &mut AutoAssertNoGc<'_>,
+        _gc_ref: VMGcRef,
+    ) -> Rooted<Self> {
+        unreachable!()
+    }
+
+    pub fn ty(&self, _store: impl AsContext) -> Result<StructType> {
+        match *self {}
+    }
+
+    pub fn matches_ty(&self, _store: impl AsContext, _ty: &StructType) -> Result<bool> {
+        match *self {}
+    }
+
+    pub(crate) fn _matches_ty(&self, _store: &StoreOpaque, _ty: &StructType) -> Result<bool> {
+        match *self {}
+    }
+
+    pub fn fields<'a, T: 'a>(
+        &self,
+        _store: impl Into<StoreContextMut<'a, T>>,
+    ) -> Result<impl ExactSizeIterator<Item = Val> + 'a> {
+        match *self {}
+        Ok([].into_iter())
+    }
+
+    pub fn field(&self, _store: impl AsContextMut, _index: usize) -> Result<Val> {
+        match *self {}
+    }
+
+    pub fn set_field(&self, _store: impl AsContextMut, _index: usize, _value: Val) -> Result<()> {
+        match *self {}
+    }
+}
--- a/crates/wasmtime/src/runtime/gc/enabled/anyref.rs
+++ b/crates/wasmtime/src/runtime/gc/enabled/anyref.rs
@@ -98,14 +98,14 @@ pub struct AnyRef {
 impl From<Rooted<StructRef>> for Rooted<AnyRef> {
     #[inline]
     fn from(s: Rooted<StructRef>) -> Self {
-        s.cast()
+        s.unchecked_cast()
     }
 }
 
 impl From<ManuallyRooted<StructRef>> for ManuallyRooted<AnyRef> {
     #[inline]
     fn from(s: ManuallyRooted<StructRef>) -> Self {
-        s.cast()
+        s.unchecked_cast()
     }
 }
 
--- a/crates/wasmtime/src/runtime/gc/enabled/anyref.rs
+++ b/crates/wasmtime/src/runtime/gc/enabled/anyref.rs
@@ -1,16 +1,15 @@
 //! Implementation of `anyref` in Wasmtime.
 
-use wasmtime_environ::VMGcKind;
-
+use crate::prelude::*;
 use crate::runtime::vm::VMGcRef;
-use crate::{prelude::*, ArrayType, StructType};
 use crate::{
     store::{AutoAssertNoGc, StoreOpaque},
-    AsContext, AsContextMut, GcRefImpl, GcRootIndex, HeapType, ManuallyRooted, RefType, Result,
-    RootSet, Rooted, ValRaw, ValType, WasmTy, I31,
+    ArrayType, AsContext, AsContextMut, GcRefImpl, GcRootIndex, HeapType, ManuallyRooted, RefType,
+    Result, RootSet, Rooted, StructRef, StructType, ValRaw, ValType, WasmTy, I31,
 };
 use core::mem;
 use core::mem::MaybeUninit;
+use wasmtime_environ::VMGcKind;
 
 /// An `anyref` GC reference.
 ///
@@ -93,7 +92,21 @@ use core::mem::MaybeUninit;
 #[derive(Debug)]
 #[repr(transparent)]
 pub struct AnyRef {
-    inner: GcRootIndex,
+    pub(super) inner: GcRootIndex,
+}
+
+impl From<Rooted<StructRef>> for Rooted<AnyRef> {
+    #[inline]
+    fn from(s: Rooted<StructRef>) -> Self {
+        s.cast()
+    }
+}
+
+impl From<ManuallyRooted<StructRef>> for ManuallyRooted<AnyRef> {
+    #[inline]
+    fn from(s: ManuallyRooted<StructRef>) -> Self {
+        s.cast()
+    }
 }
 
 unsafe impl GcRefImpl for AnyRef {
@@ -190,8 +203,14 @@ impl AnyRef {
         store: &mut AutoAssertNoGc<'_>,
         gc_ref: VMGcRef,
     ) -> Rooted<Self> {
-        assert!(gc_ref.is_i31());
-        assert!(VMGcRef::ONLY_EXTERN_REF_AND_I31);
+        debug_assert!(
+            gc_ref.is_i31()
+                || store
+                    .unwrap_gc_store()
+                    .header(&gc_ref)
+                    .kind()
+                    .matches(VMGcKind::AnyRef)
+        );
         Rooted::new(store, gc_ref)
     }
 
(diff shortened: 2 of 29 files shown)


crates/wasmtime/src/runtime/vm/gc/enabled/drc.rs
```diff
@@ -218,20 +317,18 @@ impl DrcHeap {
                 host_data_table.dealloc(host_data_id);
             }
 
+            // TODO: `dec_ref_and_maybe_dealloc` each `VMGcRef` inside this
+            // object.
```
fitzgen, line 321: Note that this is intentionally left for follow up PRs. I want to focus on getting the spec tests passing and all that first, especially since the DRC collector is just the stand in until we eventually grow our semi-space collector.

crates/wasmtime/src/runtime/gc/disabled/structref.rs
```diff
@@ -0,0 +1,53 @@
...
+/// feature was not enabled.
+pub enum StructRefPre {}
+
+/// Support for `structref` disabled at compile time because the `gc` cargo feature
+/// was not enabled.
+pub enum StructRef {}
+
+impl GcRefImpl for StructRef {}
+
+#[allow(missing_docs, unreachable_code)]
```
alexcrichton, line 17: Should this perhaps be hoisted up to the `disabled` module layer?

crates/wasmtime/src/runtime/gc/enabled/anyref.rs
```diff
@@ -349,6 +368,72 @@ impl AnyRef {
...
+    /// # Panics
+    ///
+    /// Panics if this reference is associated with a different store.
+    pub fn is_struct(&self, store: impl AsContext) -> Result<bool> {
+        self._is_struct(store.as_context().0)
+    }
+
+    pub(crate) fn _is_struct(&self, store: &StoreOpaque) -> Result<bool> {
+        // NB: Can't use `AutoAssertNoGc` here because we only have a shared
+        // context, not a mutable context.
```
alexcrichton, line 387: Performing a GC requires `&mut` though, right? So given `&StoreOpaque` that's static proof that a GC won't happen? 

Musing though: because `unchecked_try_gc_ref` returns a lifetime bound to `&StoreOpaque` doesn't that mean that GC is already statically disallowed while we're working with the `gc_ref`? If that's the case could the `unchecked_` prefix be removed? (fine to defer this to a future PR of course)

crates/wasmtime/src/runtime/gc/enabled/rooting.rs
```diff
@@ -925,6 +943,15 @@ impl<T: GcRef> Rooted<T> {
         gc_ref.hash(state);
         Ok(())
     }
+
+    /// Cast `self` to a `Rooted<U>`.
+    ///
+    /// It is the caller's responsibility to ensure that `self` is actually a
+    /// `U`. Failure to uphold this invariant will be memory safe but will
+    /// result in general incorrectness such as panics and wrong results.
+    pub(crate) fn cast<U: GcRef>(self) -> Rooted<U> {
```
alexcrichton, line 952: Should this (and below) perhaps be called `unchecked_cast` to encourage more scrutiny at callsites?

crates/wasmtime/src/runtime/gc/enabled/structref.rs
```diff
@@ -0,0 +1,757 @@
...
+
+impl StructRef {
+    /// Allocate a new `struct` and get a reference to it.
+    ///
+    /// # Errors
+    ///
+    /// If the allocation cannot be satisfied because the GC heap is currently
+    /// out of memory, but performing a garbage collection might free up space
+    /// such that retrying the allocation afterwards might succeed, then a
+    /// [`GcHeapOutOfMemory<()>`][crate::GcHeapOutOfMemory] error is returned.
```
alexcrichton, line 193: Another source of errors is if `fields` have the wrong types for `allocator`, right?

crates/wasmtime/src/runtime/gc/enabled/structref.rs
```diff
@@ -0,0 +1,757 @@
...
+        );
+        for (ty, val) in allocator.ty.fields().zip(fields) {
+            assert!(
+                val.comes_from_same_store(store),
+                "field value comes from the wrong store",
+            );
+            let ty = match ty.element_type() {
+                StorageType::I8 | StorageType::I16 => &ValType::I32,
+                StorageType::ValType(ty) => ty,
+            };
```
alexcrichton, line 233: Oh I thought you made a helper that went from storage type to val type?

crates/wasmtime/src/runtime/gc/enabled/structref.rs
```diff
@@ -0,0 +1,757 @@
...
+            .err2anyhow()?;
+        let mut store = AutoAssertNoGc::new(store);
+        for (index, (ty, val)) in allocator.ty.fields().zip(fields).enumerate() {
+            structref.initialize_field(
+                &mut store,
+                allocator.layout(),
+                ty.element_type(),
+                index,
+                val.clone(),
+            )?;
```
alexcrichton, line 255: I haven't gotten far enough in the diff so this may be more apparent later, but the `?` here is a bit suspect while there's uninitialized memory above. What's the consequence of returning an error here with uninitialized memory? Is that ok insofar as the actual reference won't go anywhere and it'll get cleaned up on the next GC?

crates/wasmtime/src/runtime/gc/enabled/structref.rs
```diff
@@ -0,0 +1,757 @@
...
+
+    unsafe fn load(store: &mut AutoAssertNoGc<'_>, ptr: &ValRaw) -> Self {
+        let raw = ptr.get_anyref();
+        debug_assert_ne!(raw, 0);
+        let gc_ref = VMGcRef::from_r64(raw.into())
+            .expect("valid r64")
+            .expect("non-null");
+        let gc_ref = store.unwrap_gc_store_mut().clone_gc_ref(&gc_ref);
+        StructRef::from_cloned_gc_ref(store, gc_ref)
+    }
```
alexcrichton, line 577: I suspect that this looks the same for anyref/arrayref/etc, do you think this could be refactored into a shared method on `Rooted` perhaps or something like that?

alexcrichton: (same for the impls below, and of course ok to defer to a future PR if you'd like)

crates/wasmtime/src/runtime/gc/enabled/structref.rs
```diff
@@ -0,0 +1,757 @@
...
+
+        let layout = store
+            .engine()
+            .signatures()
+            .layout(index)
+            .expect("struct types should have GC layouts");
+        let layout = match layout {
+            crate::vm::GcLayout::Array(_) => unreachable!(),
+            crate::vm::GcLayout::Struct(s) => s,
+        };
```
alexcrichton, line 360: Much of this prefix is shared between here and `set`, would it be possible to refactor that to a shared helper?

Also it feels a little odd where reading fields is always done with the iterator here and `field` is defined in terms of the iterator while `set_field` is defined a field-at-a-time. Could the implementation here delegate to `field` instead of the other way around?

crates/wasmtime/src/runtime/values.rs
```diff
@@ -475,6 +482,20 @@ impl From<Option<Rooted<AnyRef>>> for Val {
     }
 }
 
+impl From<Rooted<StructRef>> for Val {
+    #[inline]
+    fn from(val: Rooted<StructRef>) -> Val {
+        Val::AnyRef(Some(val.cast()))
```
alexcrichton, line 488: Above I was thinking that there should be `unchecked_cast` to require more scrutiny, but now I'm wondering if it might be good to have that *and* specific methods per-type to cast "safely" to the parent types? Not memory-safety-related-safety but more correctness-related-safety. That way the upcast method could use `unchecked_cast` internally but `unchecked_cast` wouldn't be used externally much as the safer upcasts would be used instead.

crates/wasmtime/src/runtime/gc/enabled/structref.rs
```diff
@@ -0,0 +1,757 @@
...
+                ty.fields().len()
+            ),
+        };
+        ensure!(
+            field_ty.mutability().is_var(),
+            "cannot set field {index}: field is not mutable"
+        );
+        value
+            .ensure_matches_ty(&store, &field_ty.element_type().unpack())
+            .with_context(|| format!("cannot set field {index}: type mismatch"))?;
```
alexcrichton, line 496: I think that this is double-verifying the type because this type-check happens in `write_field` as well?

crates/wasmtime/src/runtime/vm/gc/enabled/structref.rs
```diff
@@ -0,0 +1,471 @@
...
+/// A plain-old-data type that can be stored in a `ValType` or a `StorageType`.
+///
+/// Safety: implementations must be POD and all bit patterns must be valid.
+pub unsafe trait PodValType: Copy {
+    /// Read an instance of `Self` from the given little-endian bytes.
+    fn read_le(le_bytes: &[u8]) -> Self;
+
+    /// Write `self` into the given memory location, as little-endian bytes.
+    unsafe fn write_le(&self, into: *mut u8);
+}
```
alexcrichton, line 325: Could this be made a safe trait with something like:

```rust
pub trait PodValueType: Copy {
    const SIZE: usize;
    fn from_le_bytes(bytes: &[u8; Self::SIZE]) -> Self;
    fn to_le_bytes(&self, dst: &mut [u8; Self::SIZE]);
}
```

crates/wasmtime/src/runtime/vm/gc/enabled/structref.rs
```diff
@@ -0,0 +1,471 @@
...
+/// indexed GC heaps. Providing incorrect offsets will result in general
+/// incorrectness, such as wrong answers or even panics, however.
+///
+/// Finally, these methods *will* panic on out-of-bounds accesses, either out of
+/// the GC heap's bounds or out of this struct's bounds. The former is necessary
+/// for preserving the memory safety of indexed GC heaps in the face of (for
+/// example) collector bugs, but the latter is just a defensive technique to
+/// catch bugs early and prevent action at a distance as much as possible.
+pub struct VMStructDataMut<'a> {
+    data: &'a mut [u8],
```
alexcrichton, line 443: To confirm, all data on the gc heap is always initialized, right? We don't have to use `MaybeUninit<u8>` here?

crates/wasmtime/src/runtime/vm/gc/enabled/structref.rs
```diff
@@ -0,0 +1,471 @@
...
+    /// Calling this method on a structref that has already had the associated
+    /// field initialized will result in GC bugs. These are memory safe but will
+    /// lead to generally incorrect behavior such as panics, leaks, and
+    /// incorrect results.
+    ///
+    /// Returns an error if `val` is a GC reference that has since been
+    /// unrooted.
+    ///
+    /// Panics on out-of-bounds accesses.
+    pub fn initialize_field(
```
alexcrichton, line 240: It took me a bit to work out why this exists given `write_field` above. I think I understand now in that it has to do with gc barriers and things like that. This leads me to a few questions:

* Could the documentation be updated to explicitly discuss the difference between this method and `write_field`?
* Would it be possible to change these to a single method? Perhaps with a boolean flag indicating if the previous data is "valid"?

alexcrichton (approved):
I'm liking how this is all shaping up, very nice! I've got cosmetic comments here and there but nothing major.

crates/wasmtime/src/runtime/gc/enabled/anyref.rs
fitzgen: Spun this out to https://github.com/bytecodealliance/wasmtime/issues/8940

crates/wasmtime/src/runtime/gc/enabled/structref.rs
fitzgen: heh, I wrote the helper only after I'd already open coded it a bunch and I guess I missed this spot

fitzgen: ah, actually I wrote the helper for `wasmtime_types::FieldType` not `wasmtime::FieldType`... we might want to actually do some trait nastiness to dedupe all this stuff at some point...

fitzgen: oh no nvm again: it is a method on `StorageType` not `FieldType`, d'oh

crates/wasmtime/src/runtime/values.rs
fitzgen: How are you imagining this? Because this `From` impl *is* that safe upcast in my imagination. We could have a trait or something tho too

```rust
pub trait Upcast<T> {}

impl Rooted<T> {
    pub fn upcast<U>(self) -> Rooted<U>
    where
        T: Upcast<U>
    {
        // ...
    }
}
```

fitzgen: (we would have to seal the trait as well)

crates/wasmtime/src/runtime/vm/gc/enabled/structref.rs
fitzgen: From the Rust perspective: yes. We get it zeroed from mmap. At worst it has stale data from a since-reclaimed GC allocation.

From the Wasm perspective, it could be uninitialized, and we have to initialize it before we expose the object to Wasm.

But to answer the question of `MaybeUninit`: no, it shouldn't be necessary.

crates/wasmtime/src/runtime/vm/gc/enabled/structref.rs
fitzgen: I can certainly add more docs clarifying their relationship.

I don't think I want to merge them into a single method, since if the method isn't inlined then that boolean argument will be materialized and branched on at runtime, which seems pretty silly since all call sites will be statically choosing one or the other. I guess I could have a `const bool` generic parameter... not sure that is actually any nicer than two different methods tho.

crates/wasmtime/src/runtime/values.rs
alexcrichton: Oh I'm thinking something like:

```rust
impl Rooted<StructRef> {
    fn to_anyref(&self) -> Rooted<AnyRef> { self.unchecked_cast() }
}
```

and repeating that for `Rooted<...>` as appropriate and having `to_foo` method for anything that's appropriate

crates/wasmtime/src/runtime/vm/gc/enabled/structref.rs
alexcrichton: None of this is perf-sensitive though, right? If perf matters there's tons of places we'd want to improve I think, chiefly being removing the usage of `Val`. Regardless though I think more docs is sufficient for now.

crates/wasmtime/src/runtime/values.rs
fitzgen: Ah, gotcha. Yeah I can do that.

crates/wasmtime/src/runtime/vm/gc/enabled/structref.rs
fitzgen: It isn't perf-sensitive yet, but it will likely eventually be, yes.

I'd certainly like to add a `TypedStructRefPre` or something to that effect, but I didn't want to do too much too eagerly, especially before I even have the spec tests passing.

crates/wasmtime/src/runtime/gc/enabled/structref.rs
fitzgen: > Is that ok insofar as the actual reference won't go anywhere and it'll get cleaned up on the next GC?

That would be the case for tracing GCs. For the DRC collector, I think we would actually permanently leak it, since it's refcount would be stuck at `1`.

So, good eye. I'll handle these errors locally and free the allocation before bailing out of the function.

crates/wasmtime/src/runtime/gc/enabled/structref.rs
fitzgen: Spun out into https://github.com/bytecodealliance/wasmtime/issues/8942 but yeah you're right, this is all basically the same.

Issue #8919, opened 2024-07-08
Cranelift: Doesn't compile anymore on latest Windows Nightly


CryZe:
```
error: failed to run custom build command for `cranelift-codegen v0.109.0`

Caused by:
  process didn't exit successfully: `D:\a\livesplit-core\livesplit-core\target\debug\build\cranelift-codegen-16ee0d76fd74ed56\build-script-build` (exit code: 0xc00000fd, STATUS_STACK_OVERFLOW)
  --- stdout
  cargo:rustc-cfg=feature="x86"
  cargo:rerun-if-changed=build.rs
  cargo:rerun-if-changed=src\prelude.isle
  cargo:rerun-if-changed=src\prelude_opt.isle
  cargo:rerun-if-changed=src\opts\arithmetic.isle
  cargo:rerun-if-changed=src\opts\bitops.isle
  cargo:rerun-if-changed=src\opts\cprop.isle
  cargo:rerun-if-changed=src\opts\extends.isle
  cargo:rerun-if-changed=src\opts\icmp.isle
  cargo:rerun-if-changed=src\opts\remat.isle
  cargo:rerun-if-changed=src\opts\selects.isle
  cargo:rerun-if-changed=src\opts\shifts.isle
  cargo:rerun-if-changed=src\opts\spaceship.isle
  cargo:rerun-if-changed=src\opts\spectre.isle
  cargo:rerun-if-changed=src\opts\vector.isle

  --- stderr
  Writing generated file: D:\a\livesplit-core\livesplit-core\target\x86_64-pc-windows-msvc\debug\build\cranelift-codegen-e0c1105143d63b20\out\settings.rs
  Writing generated file: D:\a\livesplit-core\livesplit-core\target\x86_64-pc-windows-msvc\debug\build\cranelift-codegen-e0c1105143d63b20\out\types.rs
  Writing generated file: D:\a\livesplit-core\livesplit-core\target\x86_64-pc-windows-msvc\debug\build\cranelift-codegen-e0c1105143d63b20\out\opcodes.rs
  Writing generated file: D:\a\livesplit-core\livesplit-core\target\x86_64-pc-windows-msvc\debug\build\cranelift-codegen-e0c1105143d63b20\out\inst_builder.rs
  Writing generated file: D:\a\livesplit-core\livesplit-core\target\x86_64-pc-windows-msvc\debug\build\cranelift-codegen-e0c1105143d63b20\out\settings-x86.rs
[...]

PR #8935, opened 2024-07-10
ISLE: Make block codegen non-recursive

jameysharp:
In #8919 we learned that `Codegen::emit_block` can overrun the stack when generating Rust source for deeply-nested blocks. Stack usage can be worse than one might expect because ISLE is called from a build script, and those are always built in debug mode.

This commit avoids the problem by using an explicit heap-allocated stack instead of recursion.

I recommend turning on the "ignore whitespace" option in your diff tool of choice when reviewing this commit.

The amount of stack space that this recursive function uses is affected by the version of rustc used to compile it, and the amount of stack space available depends on the host platform. As a result, this was only observed on Windows with a recent nightly version of rustc, but it could eventually affect other platforms or compiler versions, especially as ISLE rules grow in complexity.

Note that there are several other places in ISLE which recurse over the user-provided rules. This commit does not change those, but they could also become a problem in the future.

### DIFF
--- a/cranelift/isle/isle/src/codegen.rs
+++ b/cranelift/isle/isle/src/codegen.rs
@@ -5,6 +5,7 @@ use crate::serialize::{Block, ControlFlow, EvalStep, MatchArm};
 use crate::stablemapset::StableSet;
 use crate::trie_again::{Binding, BindingId, Constraint, RuleSet};
 use std::fmt::Write;
+use std::slice::Iter;
 
 /// Options for code generation.
 #[derive(Clone, Debug, Default)]
@@ -31,15 +32,21 @@ struct Codegen<'a> {
     terms: &'a [(TermId, RuleSet)],
 }
 
-struct BodyContext<'a, W> {
+enum Nested<'a> {
+    Cases(Iter<'a, EvalStep>),
+    Arms(BindingId, Iter<'a, MatchArm>),
+}
+
+struct BodyContext<'a, 'b, W> {
     out: &'a mut W,
     ruleset: &'a RuleSet,
     indent: String,
     is_ref: StableSet<BindingId>,
     is_bound: StableSet<BindingId>,
+    stack: Vec<(Nested<'b>, &'b str, StableSet<BindingId>)>,
 }
 
-impl<'a, W: Write> BodyContext<'a, W> {
+impl<'a, 'b, W: Write> BodyContext<'a, 'b, W> {
     fn new(out: &'a mut W, ruleset: &'a RuleSet) -> Self {
         Self {
             out,
@@ -47,6 +54,7 @@ impl<'a, W: Write> BodyContext<'a, W> {
             indent: Default::default(),
             is_ref: Default::default(),
             is_bound: Default::default(),
+            stack: Default::default(),
         }
     }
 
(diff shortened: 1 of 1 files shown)


cranelift/isle/isle/src/codegen.rs
```diff
@@ -60,7 +68,21 @@ impl<'a, W: Write> BodyContext<'a, W> {
...
+        nested: Nested<'b>,
+        last_line: &'b str,
+        scope: StableSet<BindingId>,
+    ) -> std::fmt::Result {
+        self.stack.push((nested, last_line, scope));
+        self.begin_block()
+    }
+
+    fn end_nested(&mut self) -> std::fmt::Result {
+        let (_, end, scope) = self.stack.pop().unwrap();
```
fitzgen, line 82: Can this method take these things as arguments, instead of popping them? IIUC, that would allow the `while let` loop to do `ctx.stack.pop()` instead of `ctx.stack.last_mut()` which would make reasoning about control flow and termination much more straightforward, IMO, since it would all be localized within the same function.

fitzgen: Similarly, soft preference for having `begin_nested` (or `begin_block`, and removing `begin_nested`) return the stack entry and having callers do something like

```rust
let entry = ctx.begin_nested()?;
ctx.stack.push(entry);
```

For the same reasons, I think this would make it a little easier to understand what's happening here.

cranelift/isle/isle/src/codegen.rs
```diff
@@ -499,167 +508,195 @@ impl<L: Length, C> Length for ContextIterWrapper<L, C> {{
...
+                                        "{}if returns.len() >= MAX_ISLE_RETURNS {{ return; }}",
+                                        ctx.indent
+                                    )?;
+                                }
+                            }
                         }
                     }
                 }
+
+                Nested::Arms(source, arms) => {
```
fitzgen, line 684: I can't seem to find anything that pops the `Nested::Arms` entries from the stack -- does that ever happen? If not, is that a bug? I would assume it is?

cranelift/isle/isle/src/codegen.rs
```diff
@@ -499,167 +508,195 @@ impl<L: Length, C> Length for ContextIterWrapper<L, C> {{
...
+                    let Some(arm) = arms.next() else {
+                        ctx.end_nested()?;
+                        continue;
+                    };
```
fitzgen, line 688: D'oh it is the `else { ctx.end_nested() }` that does the pop.

This kinda highlights what I was talking about w.r.t. following control and reasoning about termination, though.

I see that it is a little tricky though because of how we are pumping this iterator and only actually popping this stack entry once the iterator is exhausted. It is, of course, possible to pop the iterator and then push it back on if it is not empty (would require a peekable iterator) and this would be fine. But it might be more natural or "simpler" (depending on your taste) to instead push a separate stack entry for each item in the iterator. This latter approach doesn't require any extra is-it-done-or-not wrangling, which is kind of nice. It does use probably mean more entries pushed onto the stack and therefore potentially a larger heap allocation, but I don't think that matters at all here.

fitzgen (commented):
Broadly LGTM, but the details of the stack's management is a little obscured IMO, so I have some suggestions below. If you feel strongly that I am off base here, however, I can be convinced otherwise. Let me know!

jameysharp:
I've now made the stack local to `emit_block`, and also used `pop` instead of `last`. I don't think the result can be honestly said to be clear, but I agree that it's less unclear than it was before :sweat_smile: 

It's actually important to have some special treatment of the end of a list of cases or match-arms: we need to print some stuff and reset some state about what's still in scope. So if I take your suggestion of pushing each case/arm individually onto the stack, I also would need to push something like `Nested::EndBlock { last_line, scope }` first, which is certainly possible but is a fair bit more complexity each time stuff gets pushed onto the stack. Since that happens in more places (five) than calls to `end_block` (two), I feel like that's overall worse. But I'm still open to suggestions, and also may think differently about this after I get some sleep.

fitzgen (approved):
I like this version a lot better, thank you!

PR #8937, opened 2024-07-11
`cranelift-frontend`: Specify stack map behavior when defining variables

fitzgen:
This makes it so that all uses of the variable are automatically marked as needing stack maps, which is a big simplification and removal of potential footguns for users.

### DIFF
--- a/cranelift/frontend/src/frontend.rs
+++ b/cranelift/frontend/src/frontend.rs
@@ -261,20 +261,6 @@ impl fmt::Display for DefVariableError {
     }
 }
 
-/// Whether to include or exclude a variable inside stack maps.
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub enum StackMapBehavior {
-    /// Include the associated variable in stack maps.
-    ///
-    /// At every safepoint, all values that are uses of the associated variable
-    /// will be spilled to the stack and included in stack map metadata, so that
-    /// the collector can identify all on-stack roots.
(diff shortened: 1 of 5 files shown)


cranelift/frontend/src/frontend.rs
```diff
@@ -382,21 +397,56 @@ impl<'a> FunctionBuilder<'a> {
...
-    pub fn declare_var(&mut self, var: Variable, ty: Type) {
-        self.try_declare_var(var, ty)
+    /// Declares the type of a variable, panicking if it is already declared.
+    ///
+    /// # Panics
+    ///
+    /// Panics if either the variable has already been declared, or if
+    /// `stack_map_behavior` is `StackMapBehavior::Include` and the given type
+    /// is larger than 16 bytes.
+    pub fn declare_var(&mut self, var: Variable, ty: Type, stack_map_behavior: StackMapBehavior) {
```
bjorn3, line 448: Would it be possible to introduce a new method instead. Most users of cranelift-frontend will not use stackmaps and may not even know what they are and thus not know what they should pass.

fitzgen: I actually went back and forth on this, but after switching to a different method to declare that a variable needs inclusion in stack maps, rather than a parameter to `declare_var` (see most recent commit) I actually like the different method better. Thanks for the nudge!

alexcrichton (approved):

PR #8938, opened 2024-07-11
Fix set_epoch_deadline docs

pkoenig10:
The parameter is named `ticks_beyond_current` but the docs refer to it as `delta_beyond_current`.


alexcrichton (approved):
Thanks!

PR #8941, opened 2024-07-11
fuzzgen: Add stack map variables

afonso360:
👋 Hey,

This PR adds the new variable stack map apis (added in #8937) to cranelift fuzzgen.

I've tried to fuzz this for a bit, but it seems to generate invalid functions that don't pass the validator. I don't know enough about how stackmaps work to debug this. (cc @fitzgen)

I also don't know if this feature is ready for fuzzing yet, so if it isn't let me know!

Here's one of the fuzz bugs:

<details>
<summary>Panic</summary>

```
thread '<unnamed>' panicked at fuzz/fuzz_targets/cranelift-fuzzgen.rs:402:14:
called `Result::unwrap()` on an `Err` value: Compilation error: Verifier errors

Caused by:
    0: Verifier errors
    1: - inst10 (stack_store.f32 v24, ss1): uses value v24 from non-dominating inst6
```
</details>

<details>
<summary>base64</summary>

```
/34Vbf91NDQzW23Ny8vLNMvFNDQ3ycvKLjg4ODg4ODAwMHIwMTUxbTMwMDL/Gv8jygI0ODh4MDAwcjAxNTFtMzAwMv8CNDQ0eHg0NDQjygD///8+Pl0+Pj4+PgEBAQEBAQH3AQ==
```
</details>

<details>
<summary>CLIF</summary>

```
test interpret
test run
set probestack_size_log2=11
set probestack_strategy=inline
set enable_safepoints=true
set enable_llvm_abi_extensions=true
set preserve_frame_pointers=true
set machine_code_cfg_info=true
set enable_probestack=true
set enable_jump_tables=false
set enable_heap_access_spectre_mitigation=false
set enable_incremental_compilation_cache_checks=true
target x86_64 has_sse3 has_ssse3 has_sse41 has_sse42 has_avx has_avx2 has_fma has_popcnt has_bmi1 has_bmi2 has_lzcnt 

function u1:0(i64x2, i16 uext, f32, f64x2, i32x4, i32x4, f32x4, i16x8, i8 sext, i32x4, i8x16, i32 uext, i128, i64 sext) -> i32x4, i32x4, i64x2, i32x4, i64x2, i16 uext, f32, f64x2, i32x4, i32x4, f32x4, i16x8 fast {
    ss0 = explicit_slot 2, align = 2
    ss1 = explicit_slot 4, align = 4
    ss2 = explicit_slot 16, align = 16
    sig0 = (f32) -> f32 system_v
    sig1 = (f64) -> f64 system_v
[...]

PR #8945, opened 2024-07-12
Resolve aliases before inserting values into the live set

fitzgen:
This fixes a fuzz bug found in the development of https://github.com/bytecodealliance/wasmtime/pull/8941

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3711,6 +3711,7 @@ dependencies = [
  "cranelift-native",
  "cranelift-reader",
  "cranelift-wasm",
+ "env_logger",
  "libfuzzer-sys",
  "once_cell",
  "proc-macro2",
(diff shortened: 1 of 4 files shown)


alexcrichton (approved):

bjorn3:
Why does the alias resolving need to be done? cranelift-frontend can and does change aliases on the fly. It feels iffy to immediately resolve them inside cranelift-frontend as there is nothing preventing the alias that was resolved to be changed later.

fitzgen:
> Why does the alias resolving need to be done?

Because otherwise we would need to have a reverse-aliases map from a value to all of the values that alias it. This is necessary because when we find a definition, we kill it from the live set, so at that point we only have the non-alias value.

> It feels iffy to immediately resolve them inside cranelift-frontend as there is nothing preventing the alias that was resolved to be changed later.

Note that the safepoint spilling and its liveness analysis happens as part of finishing a `FunctionBuilder`, so we don't need to worry about anything changing under our feet here.

PR #8950, opened 2024-07-13
Implement wasi-runtime-config

iawia002:
Part of https://github.com/bytecodealliance/wasmtime/issues/8929, the integration of `wasmtime-cli` will be implemented in the next separate PR.

close https://github.com/bytecodealliance/wasmtime/pull/8939
close https://github.com/bytecodealliance/wasmtime/pull/8928

cc @alexcrichton

### DIFF
--- a/crates/wasi-runtime-config/src/lib.rs
+++ b/crates/wasi-runtime-config/src/lib.rs
@@ -18,7 +18,7 @@
 //!     Config, Engine, Result, Store,
 //! };
 //! use wasmtime_wasi::{WasiCtx, WasiCtxBuilder, WasiView};
-//! use wasmtime_wasi_runtime_config::WasiRuntimeConfig;
+//! use wasmtime_wasi_runtime_config::{WasiRuntimeConfig, WasiRuntimeConfigVariables};
 //!
 //! #[tokio::main]
 //! async fn main() -> Result<()> {
@@ -29,18 +29,17 @@
 //!     let mut store = Store::new(&engine, Ctx {
 //!         table: ResourceTable::new(),
 //!         wasi_ctx: WasiCtxBuilder::new().build(),
+//!         wasi_runtime_config_vars: WasiRuntimeConfigVariables::from_iter(vec![
+//!             ("config_key1", "value1"),
+//!             ("config_key2", "value2"),
+//!         ]),
 //!     });
 //!
 //!     let mut linker = Linker::<Ctx>::new(&engine);
 //!     wasmtime_wasi::add_to_linker_async(&mut linker)?;
 //!     // add `wasi-runtime-config` world's interfaces to the linker
-//!     wasmtime_wasi_runtime_config::add_to_linker(&mut linker, |_| {
-//!         // These configuration items can come from many places,
-//!         // and if necessary, `Ctx` can also be accessed in this closure
-//!         WasiRuntimeConfig::from_iter(vec![
-//!             ("config_key1", "value1"),
-//!             ("config_key2", "value2"),
-//!         ])
+//!     wasmtime_wasi_runtime_config::add_to_linker(&mut linker, |h: &mut Ctx| {
+//!         WasiRuntimeConfig::from(&h.wasi_runtime_config_vars)
 //!     })?;
 //!
 //!     // ... use `linker` to instantiate within `store` ...
@@ -51,6 +50,7 @@
 //! struct Ctx {
 //!     table: ResourceTable,
 //!     wasi_ctx: WasiCtx,
+//!     wasi_runtime_config_vars: WasiRuntimeConfigVariables,
 //! }
 //!
 //! impl WasiView for Ctx {
@@ -79,9 +79,9 @@ use self::gen_::wasi::config::runtime as generated;
 
 /// Capture the state necessary for use in the `wasi-runtime-config` API implementation.
 #[derive(Default)]
-pub struct WasiRuntimeConfig(HashMap<String, String>);
+pub struct WasiRuntimeConfigVariables(HashMap<String, String>);
 
-impl<S: Into<String>> FromIterator<(S, S)> for WasiRuntimeConfig {
+impl<S: Into<String>> FromIterator<(S, S)> for WasiRuntimeConfigVariables {
     fn from_iter<I: IntoIterator<Item = (S, S)>>(iter: I) -> Self {
         Self(
             iter.into_iter()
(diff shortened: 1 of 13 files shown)


crates/wasi-runtime-config/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+        S: Into<String>,
+        I: IntoIterator<Item = (S, S)>,
+    {
+        Self {
+            runtime_config: config
+                .into_iter()
+                .map(|(k, v)| (k.into(), v.into()))
+                .collect(),
+        }
+    }
```
fitzgen, line 31: I think it would be more ergonomic and idiomatic to turn this function into a [`FromIterator`](https://doc.rust-lang.org/nightly/std/iter/trait.FromIterator.html) trait implementation, derive the `Default` trait for this type, make `new` return the default, and also add a `fn define(&mut self, key: impl Into<String>, value: impl Into<String>) -> &mut Self` builder-style method.

crates/wasi-runtime-config/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+    wasmtime::component::bindgen!({
+        path: "wit",
+        world: "wasi:config/imports",
+        trappable_imports: true,
+    });
+}
+use self::gen_::wasi::config::runtime as generated;
+
+/// Capture the state necessary for use in the `wasi-runtime-config` API implementation.
+pub struct WasiRuntimeConfigCtx {
```
fitzgen, line 14: Soft preference to rename this to `WasiRuntimeConfigVariables`.

crates/wasi-runtime-config/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+            runtime_config: config
+                .into_iter()
+                .map(|(k, v)| (k.into(), v.into()))
+                .collect(),
+        }
+    }
+}
+
+/// A wrapper capturing the needed internal `wasi-runtime-config` state.
+pub struct WasiRuntimeConfigView<'a> {
```
fitzgen, line 35: And then also rename this to `WasiRuntimeConfig`.

crates/wasi-runtime-config/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+    ctx: &'a WasiRuntimeConfigCtx,
+}
+
+impl<'a> WasiRuntimeConfigView<'a> {
+    /// Create a new view into the `wasi-runtime-config` state.
+    pub fn new(ctx: &'a WasiRuntimeConfigCtx) -> Self {
```
fitzgen, line 41: And it would also be nice, for convenience, to

```rust
impl<'a> From<&'a WasiRuntimeConfigVariables> for WasiRuntimeConfig<'a> {
    // ...
}
```

You might also want to consider

```rust
impl WasiRuntimeConfig<'_> {
    /// Create a new, empty runtime configuration.
    pub fn empty() -> WasiRuntimeConfig<'static> {
        static EMPTY_VARS: WasiRuntimeConfigVariables { /* ... */ };
        Self::new(&EMPTY_VARS)
    }
}
```

crates/wasi-runtime-config/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+            .iter()
+            .map(|(k, v)| (k.to_string(), v.to_string()))
+            .collect()))
+    }
+}
+
+/// Add all the `wasi-runtime-config` world's interfaces to a [`wasmtime::component::Linker`].
+pub fn add_to_linker<T>(
+    l: &mut wasmtime::component::Linker<T>,
+    f: impl Fn(&mut T) -> WasiRuntimeConfigView + Send + Sync + Copy + 'static,
```
fitzgen, line 64: Isn't this `... + 'static` bound going to mean that you can effectively only use `WasiRuntimeConfigView<'static>`?

If so -- and if the `'static` bound is necessary, which I think it is -- then instead of using lifetimes and borrows between `WasiRuntimeConfigVariables` and `WasiRuntimeConfig`, we should make `WasiRuntimeConfig` a newtype of `Arc<WasiRuntimeConfigVariables>`.

fitzgen (commented):
Thanks! A bunch of stylistic/idiomatic/ergonomic suggestions and nitpicks below, but this looks good, and we should be able to land it shortly after those things are addressed.

crates/wasi-runtime-config/src/lib.rs
iawia002: The `'static` bound is necessary, but this made me think further, since the `wasi-runtime-config` API is very simple, so I reduced some excessive encapsulation by removing `WasiRuntimeConfigVariables`/`WasiRuntimeConfigCtx` and keeping only `WasiRuntimeConfig`. Now, this problem should no longer exists.

crates/wasi-runtime-config/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+pub struct WasiRuntimeConfigView<'a> {
+    ctx: &'a WasiRuntimeConfigCtx,
+}
+
+impl<'a> WasiRuntimeConfigView<'a> {
+    /// Create a new view into the `wasi-runtime-config` state.
+    pub fn new(ctx: &'a WasiRuntimeConfigCtx) -> Self {
+        Self { ctx }
+    }
+}
```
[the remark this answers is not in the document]
iawia002: > > Is the idea here to separate the configured variables from the per-`Store` state, so that the variables can be reused across many `Store`s?

The idea here is very simple: it's just passing some configuration variables from the runtime to the component. I did add some documentation, but it is not described this way. Please check if the current description is reasonable. Feel free to add more comments.

iawia002:
> Could this also get hooked up to the CLI as well? Under perhaps `-Sruntime-config` and/or some other `-S` flags for the actual keys themselves?

I made some changes based on all the comments except this one, because I plan to add support for the `wasi-runtime-config` API to `wasmtime-cli` in a separate PR. Otherwise, this PR would have too many changes, let's focus on the `wasi-runtime-config` implementation only in this PR.

And I think all comments have been addressed, PTAL.

crates/wasi-runtime-config/src/lib.rs
alexcrichton: The `+ 'static` doesn't require `<'static>` in this case since the `+ 'static` only applies to the lifetime of the captures of the closure itself. This I think was the desired signature which allows borrowing the config without cloning it on import calls.

crates/wasi-runtime-config/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+            .iter()
+            .map(|(k, v)| (k.to_string(), v.to_string()))
+            .collect()))
+    }
+}
+
+/// Add all the `wasi-runtime-config` world's interfaces to a [`wasmtime::component::Linker`].
+pub fn add_to_linker<T>(
+    l: &mut wasmtime::component::Linker<T>,
+    f: impl Fn(&mut T) -> WasiRuntimeConfig + Send + Sync + Copy + 'static,
```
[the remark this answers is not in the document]
iawia002: Yeah, you are right, I did not understand it correctly last time. I made some changes (https://github.com/bytecodealliance/wasmtime/pull/8950/commits/99b6a113c6a5e13ce3d1a016792b32948ae7d461) and returned to the original `WasiRuntimeConfigVariables` + `WasiRuntimeConfig` style. Hope I have understood it correctly this time.

fitzgen:
Thanks!

iawia002:
Thanks for the review!

PR #8952, opened 2024-07-13
Fix `clippy::doc_lazy_continuation` lints.

waywardmonkeys:
These make it easier to read ordered and unordered lists by having more consistent indentation.

### DIFF
--- a/crates/slab/src/lib.rs
+++ b/crates/slab/src/lib.rs
@@ -55,13 +55,13 @@
 //! * Value `A` is allocated into the slab, yielding id `i`.
 //!
 //! * `A` is deallocated, and so `i`'s associated entry is added to the slab's
-//! free list.
+//!   free list.
 //!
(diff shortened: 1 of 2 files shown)


alexcrichton:
Also, if you're interested, we [default allow all clippy lints](https://github.com/bytecodealliance/wasmtime/blob/225d20efe387bc3f4052ec3115125380f9c59774/Cargo.toml#L171-L175) in this workspace because it's found to be too noisy by default. These warnings showed up due to a [misconfiguration issue](https://github.com/bytecodealliance/wasmtime/pull/8958). If you'd like though selectively adding lints to that `[workspace.lints.clippy]` block is something that's reasonable to do on a one-off basis. For example adding this lint to that block sounds like a good idea, although I think it will need to wait a bit because this lint was added in Rust 1.80 and we're using 1.79 in CI.

PR #8954, opened 2024-07-13
Cranelift: Constant propagate floats

primoly:
Adds constant propagation of the following instructions for `$F32` and `$F64`:
`fadd`, `fsub`, `fmul`, `fdiv`, `sqrt`, `ceil`, `floor`, `trunc`, `nearest`.

`fmin` and `fmax` are still missing. Those are tricky due to their handling and propagation of NaNs: [cranelift_codegen::ir::InstBuilder::fmin](https://docs.rs/cranelift-codegen/latest/cranelift_codegen/ir/trait.InstBuilder.html#method.fmin)


cranelift/codegen/src/opts/cprop.isle
```diff
@@ -280,7 +278,53 @@
 (decl pure u64_bswap64 (u64) u64)
 (extern constructor u64_bswap64 u64_bswap64)
 
-;; Constant fold bitwise float operations (fneg/fabs/fcopysign)
+;; Constant fold float operations
+;; TODO: fmin, fmax, fcmp, fma, demote, promote, from and to ops
```
fitzgen, line 282: `fma`, and any other relaxed SIMD opcode that is non-deterministic, is a little bit tricky. We would want to double check the required Wasm semantics, and whether the non-determinism is allowed to "change" throughout the program or not. If it is spec'd to be non-deterministic, but always has the same behavior for a given evaluation of the N non-deterministic choices, then const folding could be problematic if we are compiling on an aarch64 machine but running on an x64 machine, for example. It would essentially be observable to the program whether the constant folding was happening or not, and I'm not sure if that is allowed by the spec.

So we should answer this question, and then document the answer here.

fitzgen (commented):
Thanks! Looks good to me, modulo one clarification below.

cranelift/codegen/src/opts/cprop.isle
bjorn3: > If it is spec'd to be non-deterministic, but always has the same behavior for a given evaluation of the N non-deterministic choices, then const folding could be problematic if we are compiling on an aarch64 machine but running on an x64 machine, for example.

IMHO it would be problematic either way. It would break reproducibility of the compiled executable/`.cwasm` file across compiler host architectures.

fitzgen: FWIW, I tried to answer this question for myself and didn't come away with anything firm.

Filed https://github.com/WebAssembly/relaxed-simd/issues/155 to ask for clarification.

I think this `TODO` comment can just add a note about the potential gotcha with non-deterministic instructions and link to that issue.

afonso360: I think we may already be able to observe const propagation with this PR already. At least on RISC-V any NaN producing operation is guaranteed to produce the canonical NaN, but I think some other platforms preserve the payload bits.

fitzgen: > At least on RISC-V any NaN producing operation is guaranteed to produce the canonical NaN, but I think some other platforms preserve the payload bits.

`NaN`s don't have any guarantee around having the same behavior across instructions, AFAIK. But the relaxed SIMD instructions might, or at least I'm not sure, which is why I filed that issue.

> It would break reproducibility of the compiled executable/`.cwasm` file across compiler host architectures.

I guess this is a philosophical choice. It isn't clear to me whether we want to provide the weaker (deterministic cross compilation given the same host architecture) or the stronger (deterministic cross compilation regardless of host architecture) guarantee.

If the latter, then we can't do compile-time evaluation of *any* floating point operation that could involve `NaN`s.

bjorn3: > If the latter, then we can't do compile-time evaluation of any floating point operation that could involve NaNs.

You could still attempt evaluation and bail out optimizing if the output turns out to be NaN.

fitzgen: Yeah that's true. We could make all these external constructors `partial` and have them return `None` if the operation is given or produces a `NaN`.

fitzgen:
I would be more comfortable with aborting the compile-time evaluation if either given or producing a `NaN`. It is more conservative, and we can always have follow up discussions/PRs focused on just that issue. I also suspect there isn't too much additional performance to gain w.r.t `NaN`s.

----------

FWIW, it seems like compile-time evaluating relaxed SIMD instructions like `fma` is probably a no-go: https://github.com/WebAssembly/relaxed-simd/issues/155#issuecomment-2232721287

fitzgen:
> Regarding `fma`: I still believe that it is a deterministic instruction when not involving NaNs, but since this PR doesn’t include it, this is something we can discuss in the future.

Yes, my b I was thinking of the `relaxed_madd` Wasm instruction.

cranelift/codegen/src/opts/cprop.isle
```diff
@@ -280,7 +278,88 @@
 (decl pure u64_bswap64 (u64) u64)
 (extern constructor u64_bswap64 u64_bswap64)
 
-;; Constant fold bitwise float operations (fneg/fabs/fcopysign)
+;; Constant fold float operations
+;; Note: With the exception of fabs, fneg and copysign,
+;; constant folding is only performed when
+;; the result of an instruction isn't NaN.
```
fitzgen, line 284: Can you follow this with a sentence saying *why* that is? Something like

> We want the NaN bit patterns produced by an instruction to be consistent, and compile-time evaluation in a cross-compilation scenario risks producing different NaN bit patterns than the target would have at run-time.

fitzgen (commented):
Looks good modulo a rebase to address conflicts with `main` and a couple tiny nitpicks. Thanks!!

fitzgen (approved):
Perfect! Thanks again!

fitzgen (approved):

PR #8960, opened 2024-07-15
Update the docs for building the C API

fitzgen:
Suggest `cmake` since that is what we test in CI and keep working.

### DIFF
--- a/crates/c-api/README.md
+++ b/crates/c-api/README.md
@@ -1,7 +1,9 @@
 # Wasmtime's C API
 
-For more information you can find the documentation for this library
-[online](https://bytecodealliance.github.io/wasmtime/c-api/).
+## API Documentation
+
+[The API documentation for the Wasmtime C library is hosted
+here.](https://bytecodealliance.github.io/wasmtime/c-api/).
 
 ## Using in a C Project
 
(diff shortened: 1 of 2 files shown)


crates/c-api/README.md
```diff
@@ -5,13 +5,45 @@ For more information you can find the documentation for this library
...
+
+To use Wasmtime from a C or C++ project, you must have
+[CMake](https://cmake.org/) and [a Rust
+toolchain](https://www.rust-lang.org/tools/install) installed.
+
+From the root of the Wasmtime repository, run the following commands:
 
 ```
-cargo build --release -p wasmtime-c-api
+$ cmake -S crates/c-api -B target/c-api
```
alexcrichton, line 24: Coul

alexcrichton (commented):
Thanks for this! Mind rewording this though based on `cmake --build target/c-api --target install`? Basically the "install" target should be run to produce the final output structure of the `include` dir and the `lib` dir which will have all the relevant artifacts. That'll avoid the need to document the internals of the build directory and asking users to weave together everything themselves.

For documentation if the link isn't already here we should also link to https://docs.wasmtime.dev/c-api/ to avoid users needing to install/configure doxygen themselves locally.

alexcrichton (approved):

PR #8964, opened 2024-07-16
Add FP16 and I64 support for wasi-nn WinML backend.

jianjunz:
Some devices may not support FP32.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -726,8 +726,13 @@ jobs:
     # Install Rust targets.
     - run: rustup target add wasm32-wasip1
 
-    # Run the tests!
+    # Run integration tests!
     - run: cargo test -p wasmtime-wasi-nn --features ${{ matrix.feature }}
+      env:
+        RUST_BACKTRACE: 1
+
+    # Run unit tests.
+    - run: cargo test -p wasmtime-wasi-nn --features ${{ matrix.feature }} --lib
       env:
         RUST_BACKTRACE: 1
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -726,13 +726,8 @@ jobs:
     # Install Rust targets.
     - run: rustup target add wasm32-wasip1
 
-    # Run integration tests!
+    # Run the tests!
     - run: cargo test -p wasmtime-wasi-nn --features ${{ matrix.feature }}
-      env:
-        RUST_BACKTRACE: 1
-
-    # Run unit tests.
-    - run: cargo test -p wasmtime-wasi-nn --features ${{ matrix.feature }} --lib
       env:
         RUST_BACKTRACE: 1
 
--- a/crates/wasi-nn/Cargo.toml
+++ b/crates/wasi-nn/Cargo.toml
@@ -44,7 +44,7 @@ features = [
     "AI_MachineLearning",
     "Storage_Streams",
     "Foundation_Collections",
-    # For Int64 input support.
+    # For getting IVectorView<i64> from tensor.dimensions.
     "implement",
 ]
 optional = true
--- a/crates/wasi-nn/Cargo.toml
+++ b/crates/wasi-nn/Cargo.toml
@@ -46,11 +46,6 @@ features = [
     "Foundation_Collections",
     # For Int64 input support.
     "implement",
-    # Following 4 features are needed for creating a LearningModelDevice from NPU.
-    "Win32_Graphics_Direct3D",
-    "Win32_Graphics_Direct3D12",
-    "Win32_Graphics_DXCore",
-    "Win32_System_WinRT_ML",
 ]
 optional = true
 
--- a/crates/wasi-nn/Cargo.toml
+++ b/crates/wasi-nn/Cargo.toml
@@ -40,7 +40,18 @@ ort = { version = "2.0.0-rc.2", default-features = false, features = [
 
 [target.'cfg(windows)'.dependencies.windows]
 version = "0.52"
-features = ["AI_MachineLearning", "Storage_Streams", "Foundation_Collections"]
+features = [
+    "AI_MachineLearning",
+    "Storage_Streams",
+    "Foundation_Collections",
+    # For Int64 input support.
+    "implement",
+    # Following 4 features are needed for creating a LearningModelDevice from NPU.
+    "Win32_Graphics_Direct3D",
+    "Win32_Graphics_Direct3D12",
+    "Win32_Graphics_DXCore",
+    "Win32_System_WinRT_ML",
+]
 optional = true
 
 [build-dependencies]
--- a/crates/wasi-nn/src/backend/winml.rs
+++ b/crates/wasi-nn/src/backend/winml.rs
@@ -220,22 +220,25 @@ fn to_inspectable(tensor: &Tensor) -> Result<IInspectable, Error> {
         TensorType::Fp16 => unsafe {
             let data = std::slice::from_raw_parts(
                 tensor.data.as_ptr().cast::<f32>(),
-                get_safe_len::<f32>(&tensor.data),
+                tensor.data.len() / size_of::<f32>(),
             );
+            check_alignment::<f32>(data);
             TensorFloat16Bit::CreateFromArray(&shape, data)?.cast::<IInspectable>()
         },
         TensorType::Fp32 => unsafe {
             let data = std::slice::from_raw_parts(
                 tensor.data.as_ptr().cast::<f32>(),
-                get_safe_len::<f32>(&tensor.data),
+                tensor.data.len() / size_of::<f32>(),
             );
+            check_alignment::<f32>(data);
             TensorFloat::CreateFromArray(&shape, data)?.cast::<IInspectable>()
         },
         TensorType::I64 => unsafe {
             let data = std::slice::from_raw_parts(
                 tensor.data.as_ptr().cast::<i64>(),
-                get_safe_len::<i64>(&tensor.data),
+                tensor.data.len() / size_of::<i64>(),
             );
+            check_alignment::<i64>(data);
             TensorInt64Bit::CreateFromArray(&shape, data)?.cast::<IInspectable>()
         },
         _ => unimplemented!(),
@@ -284,18 +287,13 @@ fn to_tensor(inspectable: IInspectable, tensor_kind: TensorKind) -> Result<Tenso
     Ok(tensor)
 }
 
-/// Convenience function for checking that we can cast `data` to a slice of `T`, returning the
-/// length of that slice.
-fn get_safe_len<T>(data: &[u8]) -> usize {
+fn check_alignment<T>(data: &[T]) {
+    let (prefix, _slice, suffix) = unsafe { data.align_to::<T>() };
     assert!(
-        data.len() % std::mem::size_of::<T>() == 0,
-        "data size is not a multiple of the size of `T`"
+        prefix.is_empty() && suffix.is_empty(),
+        "Data is not aligned to {:?}'s alignment",
+        std::any::type_name::<T>()
     );
-    assert!(
-        data.as_ptr() as usize % std::mem::align_of::<T>() == 0,
-        "raw data is not aligned to `T`'s alignment"
-    );
-    data.len() / std::mem::size_of::<T>()
 }
 
 #[cfg(test)]
--- a/crates/wasi-nn/src/backend/winml.rs
+++ b/crates/wasi-nn/src/backend/winml.rs
@@ -309,10 +309,11 @@ mod tests {
     #[test]
     fn fp16() {
         let data = vec![1.0_f32, 2.0, 3.0, 4.0, 5.0, 6.0];
-        let mut buffer = Vec::with_capacity(data.len() * size_of::<f32>());
-        for f in &data {
-            buffer.extend(f.to_ne_bytes());
-        }
+        let buffer = data
+            .iter()
+            .map(|f| f.to_ne_bytes())
+            .flatten()
+            .collect::<Vec<u8>>();
         let buffer_copy = buffer.clone();
         let tensor = Tensor {
             ty: TensorType::Fp16,
@@ -332,10 +333,7 @@ mod tests {
         // Convert back.
         let t = to_tensor(inspectable.unwrap(), TensorKind::Float16);
         assert!(t.as_ref().is_ok());
-        let t_ref = t.as_ref();
-        assert_eq!(t_ref.unwrap().data, buffer);
-        assert_eq!(t_ref.unwrap().dimensions, tensor.dimensions);
-        assert_eq!(t_ref.unwrap().ty, TensorType::Fp16);
+        assert_eq!(t.unwrap(), tensor);
     }
 
     #[test]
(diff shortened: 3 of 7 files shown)


crates/wasi-nn/src/backend/winml.rs
```diff
@@ -226,3 +205,180 @@ fn convert_i64(i: i64) -> Result<u32, BackendError> {
...
+        crate::wit::types::TensorType::Fp32 => unsafe {
+            let data = std::slice::from_raw_parts(
+                tensor.data.as_ptr() as *const f32,
+                tensor.data.len() / size_of::<f32>(),
+            );
+            TensorFloat::CreateFromArray(&shape, data)?.cast::<IInspectable>()
+        },
+        crate::wit::types::TensorType::I64 => unsafe {
+            let data = std::slice::from_raw_parts(
+                tensor.data.as_ptr() as *const i64,
```
abrown, line 228:
```suggestion
                tensor.data.as_ptr().cast<i64>,
```

These days clippy tells me this is the preferred way to do this.

crates/wasi-nn/src/backend/winml.rs
```diff
@@ -175,23 +158,19 @@ impl BackendExecutionContext for WinMLExecutionContext {
...
-                ty: TensorType::Fp32,
-                dimensions,
-                data,
-            })
+            let output_feature = output_features.GetAt(index)?;
+            let tensor_kind = match output_feature.Kind()? {
+                windows::AI::MachineLearning::LearningModelFeatureKind::Tensor => output_feature
+                    .cast::<TensorFeatureDescriptor>()?
+                    .TensorKind()?,
+                _ => unimplemented!(),
```
abrown, line 166:
```suggestion
                _ => unimplemented!("the winml backend only supports tensors, found: {}", kind),
```

crates/wasi-nn/src/backend/winml.rs
```diff
@@ -175,23 +158,19 @@ impl BackendExecutionContext for WinMLExecutionContext {
...
-            }
-            Ok(Tensor {
-                ty: TensorType::Fp32,
-                dimensions,
-                data,
-            })
+            let output_feature = output_features.GetAt(index)?;
+            let tensor_kind = match output_feature.Kind()? {
+                windows::AI::MachineLearning::LearningModelFeatureKind::Tensor => output_feature
+                    .cast::<TensorFeatureDescriptor>()?
```
abrown, line 164: Are we sure `cast` is right here? Poking around in `windows-rs`, I see that it has a `CanInto` implementation for `TensorFeatureDescriptor` [here](https://microsoft.github.io/windows-docs-rs/doc/windows/AI/MachineLearning/struct.ILearningModelFeatureDescriptor.html#impl-CanInto%3CILearningModelFeatureDescriptor%3E-for-TensorFeatureDescriptor) &mdash; doesn't that mean we should be using `into()` instead?

crates/wasi-nn/src/backend/winml.rs
```diff
@@ -175,23 +158,19 @@ impl BackendExecutionContext for WinMLExecutionContext {
...
+                    .TensorKind()?,
+                _ => unimplemented!(),
+            };
+            // TODO: this only handles FP16, FP32 and I64!
```
abrown, line 168: This kind of comment can be moved from here to the `unimplemented!()` at the end of `to_tensor` so that both users and code readers can take advantage of this information.

crates/wasi-nn/src/backend/winml.rs
```diff
@@ -226,3 +205,180 @@ fn convert_i64(i: i64) -> Result<u32, BackendError> {
...
     })
 }
+
+// Convert from wasi-nn tensor to WinML tensor.
+fn to_inspectable(tensor: &Tensor, shape: IVectorView<i64>) -> Result<IInspectable, Error> {
+    match tensor.ty {
+        crate::wit::types::TensorType::Fp16 => unsafe {
+            let data = std::slice::from_raw_parts(
+                tensor.data.as_ptr() as *const f32,
+                tensor.data.len() / size_of::<f32>(),
```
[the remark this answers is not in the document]
abrown: E.g., https://microsoft.github.io/windows-docs-rs/doc/windows/AI/MachineLearning/struct.TensorFloat16Bit.html#method.CreateFromArray.

crates/wasi-nn/src/backend/winml.rs
```diff
@@ -226,3 +205,180 @@ fn convert_i64(i: i64) -> Result<u32, BackendError> {
...
+        _ => unimplemented!(),
+    }
+}
+
+// Convert from WinML tensor to wasi-nn tensor.
+fn to_tensor(inspectable: IInspectable, tensor_kind: TensorKind) -> Result<Tensor, BackendError> {
+    let tensor = match tensor_kind {
+        TensorKind::Float16 => {
+            let output_tensor = inspectable.cast::<TensorFloat16Bit>()?;
+            let dimensions = dimensions_as_u32(&output_tensor.Shape()?)?;
```
abrown, line 242: We can do this once at the top of the function.

crates/wasi-nn/src/backend/winml.rs
```diff
@@ -226,3 +205,180 @@ fn convert_i64(i: i64) -> Result<u32, BackendError> {
...
+            }
+        }
+        TensorKind::Float => {
+            let output_tensor = inspectable.cast::<TensorFloat>()?;
+            let dimensions = dimensions_as_u32(&output_tensor.Shape()?)?;
+            let view = output_tensor.GetAsVectorView()?;
+            let mut data = Vec::with_capacity(view.Size()? as usize * size_of::<f32>());
+            for f in view.into_iter() {
+                data.extend(f.to_le_bytes());
+            }
```
abrown, line 262: Not sure if it's any faster, but this would avoid the size calculation:

```rust
let data = view.into_iter().map(f32::to_le_bytes).collect();
```

We might want to note somewhere that the LE ordering is here to match WebAssembly's ordering, not the platform we're running on (in which case always doing LE could be incorrect).

crates/wasi-nn/src/backend/winml.rs
```diff
@@ -226,3 +205,180 @@ fn convert_i64(i: i64) -> Result<u32, BackendError> {
...
+                tensor.data.as_ptr() as *const f32,
+                tensor.data.len() / size_of::<f32>(),
+            );
+            TensorFloat16Bit::CreateFromArray(&shape, data)?.cast::<IInspectable>()
+        },
+        crate::wit::types::TensorType::Fp32 => unsafe {
+            let data = std::slice::from_raw_parts(
+                tensor.data.as_ptr() as *const f32,
+                tensor.data.len() / size_of::<f32>(),
+            );
```
abrown, line 223: When doing this `unsafe` "cast" to a slice, we also need to make sure that the `u8` pointer to `tensor.data` is aligned in such a way that the `f32`s in `data` will also be aligned to their 4-byte alignment. Here's an [example](https://github.com/intel/openvino-rs/blob/main/crates/openvino/src/tensor.rs#L124-L151) of how to do the check from `openvino-rs`. And the [docs for `std::slice::from_raw_parts`](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html) have all the details. I think it would be fine here to fail if things aren't aligned correctly (instead of trying to shift everything!), at least for the time being.

crates/wasi-nn/src/backend/winml.rs
```diff
@@ -226,3 +205,180 @@ fn convert_i64(i: i64) -> Result<u32, BackendError> {
...
+            .unwrap();
+        let view = winml_tensor.GetAsVectorView().unwrap();
+        assert_eq!(view.into_iter().collect::<Vec<i64>>(), data);
+        // Convert back.
+        let t = to_tensor(inspectable.unwrap(), TensorKind::Int64);
+        assert!(t.as_ref().is_ok());
+        let t_ref = t.as_ref();
+        assert_eq!(t_ref.unwrap().data, buffer);
+        assert_eq!(t_ref.unwrap().dimensions, tensor.dimensions);
+        assert_eq!(t_ref.unwrap().ty, TensorType::I64);
```
abrown, line 382: It took me a second to understand that what we're trying to do is "roundtrip" a tensor through `to_inspectable` and back through `to_tensor`. To make this more clear, can you look at factoring out some of the duplication here?

.github/workflows/main.yml
```diff
@@ -718,11 +718,16 @@ jobs:
...
     - run: rustup target add wasm32-wasi
 
-    # Run the tests!
+    # Run integration tests!
     - run: cargo test -p wasmtime-wasi-nn --features ${{ matrix.feature }}
       env:
         RUST_BACKTRACE: 1
 
+    # Run unit tests.
+    - run: cargo test -p wasmtime-wasi-nn --features ${{ matrix.feature }} --lib
```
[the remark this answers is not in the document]
jianjunz: You're right. The `cargo test` above also runs unit tests. https://github.com/bytecodealliance/wasmtime/actions/runs/9969920948/job/27547820669#step:8:409

crates/wasi-nn/Cargo.toml
```diff
@@ -40,7 +40,18 @@ ort = { version = "2.0.0-rc.2", default-features = false, features = [
...
+    "AI_MachineLearning",
+    "Storage_Streams",
+    "Foundation_Collections",
+    # For Int64 input support.
+    "implement",
+    # Following 4 features are needed for creating a LearningModelDevice from NPU.
+    "Win32_Graphics_Direct3D",
+    "Win32_Graphics_Direct3D12",
+    "Win32_Graphics_DXCore",
+    "Win32_System_WinRT_ML",
```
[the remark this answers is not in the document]
jianjunz: Removed. They're not needed for this one.

crates/wasi-nn/src/backend/winml.rs
```diff
@@ -136,32 +136,15 @@ impl WinMLExecutionContext {
 
 impl BackendExecutionContext for WinMLExecutionContext {
     fn set_input(&mut self, id: Id, tensor: &Tensor) -> Result<(), BackendError> {
+        // TODO: Clear previous bindings when needed.
```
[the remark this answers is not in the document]
jianjunz: WinML may report an error
```
self.binding.Bind("input", tensor of shape [10]);
self.binding.Bind("input", tensor of shape [11]);  <-- error
```
But it works
```
self.binding.Bind("input", tensor of shape [10]);
self.binding.Clear();
self.binding.Bind("input", tensor of shape [11]);
```

crates/wasi-nn/src/backend/winml.rs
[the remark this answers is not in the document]
jianjunz: WinML may use 2 bytes internally because it's fp16, but f16 is not officially supported by Rust stable, so we use f32 here.

crates/wasi-nn/src/backend/winml.rs
jianjunz: `output_tensor`'s type is unknown at the top of the function.

crates/wasi-nn/src/backend/winml.rs
jianjunz: `view`'s type is `IVectorView`, not a rust collection, so the short version doesn't work here.

crates/wasi-nn/src/backend/winml.rs
jianjunz: My understanding is `cast` here is for COM interfaces, similar as [QueryInterface](https://learn.microsoft.com/en-us/cpp/atl/queryinterface), while `into` is for converting derived class into base class?

crates/wasi-nn/src/backend/winml.rs
jianjunz: Replaced with `cast`.

crates/wasi-nn/src/backend/winml.rs
jianjunz: Thanks for the info. Added assertion when getting length.

crates/wasi-nn/src/backend/winml.rs
jianjunz: Implemented `PartialEq` for `Tensor`, so we don't need a lot of `assert_eq!` here.

crates/wasi-nn/src/backend/winml.rs
```diff
@@ -226,3 +207,186 @@ fn convert_i64(i: i64) -> Result<u32, BackendError> {
         anyhow::anyhow!("unable to convert dimension to u32: {d}").into()
     })
 }
+
+// Convert from wasi-nn tensor to WinML tensor.
+fn to_inspectable(tensor: &Tensor, shape: IVectorView<i64>) -> Result<IInspectable, Error> {
+    match tensor.ty {
```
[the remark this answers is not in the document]
jianjunz: Thanks for pointing out the duplication here. I think we don't need to verify shape here, WinML will check it later, so `shape` is removed.

crates/wasi-nn/src/backend/winml.rs
jianjunz: `flat_map` works. Thanks.

crates/wasi-nn/src/backend/winml.rs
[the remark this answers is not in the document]
abrown: I think I understand what you're saying about `Clear`: it erases _all_ the bindings, even for other model inputs. We can't have that. But what happens in this conceivable sequence?
 - Wasm guest calls `set_input` on input N with shape A
 - Wasm guest again calls `set_input` on input N with shape B

This is valid, though silly. The user should not have to face an error from wasi-nn in this case, right? But, if WinML is going to raise an error, then should we protect this some other way, e.g., by checking that the tensor shape is what the model expects (either A or B)?

crates/wasi-nn/src/backend/winml.rs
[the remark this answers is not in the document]
abrown: Ok, please add a comment explaining that and link to the docs I mentioned.

crates/wasi-nn/src/backend/winml.rs
abrown: How about

```rust
let itensor = inspectable.cast<ITensor>()?;
let dimensions = dimensions_as_u32(&itensor.Shape()?)?;
```

or something like that?

crates/wasi-nn/src/lib.rs
```diff
@@ -84,6 +84,12 @@ impl fmt::Debug for Tensor {
     }
 }
 
+impl PartialEq for Tensor {
+    fn eq(&self, other: &Self) -> bool {
+        self.dimensions == other.dimensions && self.ty == other.ty && self.data == other.data
+    }
+}
+
```
abrown, line 92: Probably should just `#[derive(PartialEq)]` since it's equivalent to this I think.

crates/wasi-nn/src/backend/winml.rs
jianjunz: It works, but `GetAsVectorView` is a method of `TensorFloat16Bit`. Then we'll need to cast `itensor` again to specific tensor type. The change above makes code clean but will it be a performance issue to cast twice?

crates/wasi-nn/src/lib.rs
jianjunz: It works. Thanks.

PR #8968, opened 2024-07-17
Immediately return StreamError::Closed when the TCP stream is closed.

badeend:
This saves the caller from having to perform one last `poll`+`read` sequence.

Before this change, it would first return an empty list of bytes, which is the wasi-io equivalent of EWOULDBLOCK.


alexcrichton:
Thanks! Would it be possible to add a test for this too?

badeend:
Yes!

The test includes a `std::thread::sleep`. Under Linux this shouldn't be necessary as communication over loopback devices happens synchronously. But I've thrown it in anyway just in case Windows or MacOS works differently.

alexcrichton (approved):

PR #8970, opened 2024-07-18
Integrate wasi-runtime-config into wasmtime-cli

iawia002:
Part of #8929 

This patch adds two new flags:

```
  -S            runtime-config[=y|n] -- Enable support for WASI runtime config API (experimental)
  -S runtime-config-var=<name>=<val> -- Pass a wasi runtime config variable to the program.
```

Separating these two flags is considered to handle the scenario where the `wasi-runtime-config` API is enabled but no configuration needs to be passed.

To use the `wasi-runtime-config` API in `wasmtime-cli`:

1. Enable `wasi-runtime-config` API support by `-Sruntime-config`:

```
wasmtime run/serve -Sruntime-config ...
```

2. Pass custom runtime configs to the component:

```
wasmtime run/serve -Sruntime-config -Sruntime-config-var=key1=value1 -Sruntime-config-var=key2=value2 ...
```

cc @alexcrichton

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -350,6 +350,7 @@ default = [
   "wasi-nn",
   "wasi-threads",
   "wasi-http",
+  "wasi-runtime-config",
 
   # Most features of Wasmtime are enabled by default.
   "wat",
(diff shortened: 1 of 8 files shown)


tests/all/cli_tests.rs
```diff
@@ -1853,6 +1853,41 @@ stderr [1] :: after empty
         run_wasmtime(&["run", "--argv0=foo.wasm", CLI_ARGV0, "foo.wasm"])?;
         Ok(())
     }
+
+    #[cfg(feature = "wasi-runtime-config")]
```
alexcrichton, line 1857: Currently we assume that all features are enabled during testing, so it's ok to drop this and the marker below.

Cargo.toml
```diff
@@ -382,6 +382,7 @@ all-arch = ["wasmtime/all-arch"]
 winch = ["wasmtime/winch"]
 wmemcheck = ["wasmtime/wmemcheck"]
 memory-protection-keys = ["wasmtime-cli-flags/memory-protection-keys"]
+wasi-runtime-config = ["dep:wasmtime-wasi-runtime-config"]
```
alexcrichton, line 385: Could this be added to the on-by-default features list? That's generally what we're doing with WASI proposals right now so long as there's no extra build requirements, which this proposal doesn't have.

iawia002:
All comments addressed, PTAL

alexcrichton (approved):

Issue #8626, opened 2024-05-15
Add CLIF to `wasmtime explore` output


fitzgen:
(When using Cranelift.)

When I originally implemented `wasmtime explore` we didn't have a good way to get the CLIF. But since then we've added the `--emit-clif` CLI flag, so I suspect it is easier to get now than it was then. Would be very cool to have three panes, and the color-coded mapping between them synchronized: (1) wasm (2) clif (3) asm

kritzcreek:
Looks like the linked PR implemented this successfully (I just tried it), so I think you could close this issue.

alexcrichton:
👍

PR #8972, opened 2024-07-18
CLIF in explorer

vulc41n:
As discussed in #8626, I added a new pane to the explorer to see the CLIF emitted. 

It creates a temporary directory next to the html file, and discards it when it's done. 

Feel free to criticize as it is my first contribution on this project 🙂

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,6 +69,7 @@ wasmparser = { workspace = true }
 tracing = { workspace = true }
 log = { workspace = true }
 humantime = { workspace = true }
+tempfile = { workspace = true }
 
 async-trait = { workspace = true }
 bytes = { workspace = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,7 +69,7 @@ wasmparser = { workspace = true }
 tracing = { workspace = true }
 log = { workspace = true }
 humantime = { workspace = true }
-tempfile = { workspace = true }
+tempfile = { workspace = true, optional = true }
 
 async-trait = { workspace = true }
 bytes = { workspace = true }
@@ -418,7 +418,7 @@ gc = ["wasmtime-cli-flags/gc"]
 # CLI subcommands for the `wasmtime` executable. See `wasmtime $cmd --help`
 # for more information on each subcommand.
 serve = ["wasi-http", "component-model", "dep:http-body-util", "dep:http"]
-explore = ["dep:wasmtime-explorer"]
+explore = ["dep:wasmtime-explorer", "dep:tempfile"]
 wast = ["dep:wasmtime-wast"]
 config = ["cache"]
 compile = ["cranelift"]
--- a/crates/explorer/src/index.css
+++ b/crates/explorer/src/index.css
@@ -15,13 +15,19 @@ body {
 }
 
 #wat {
-  width: 50%;
+  width: 33%;
+  height: 100%;
+  overflow: scroll;
+}
+
+#clif {
+  width: 33%;
   height: 100%;
   overflow: scroll;
 }
 
 #asm {
-  width: 50%;
+  width: 33%;
   height: 100%;
   overflow: scroll;
 }
--- a/crates/explorer/src/index.css
+++ b/crates/explorer/src/index.css
@@ -15,19 +15,19 @@ body {
 }
 
 #wat {
-  width: 33%;
+  flex: 1;
   height: 100%;
   overflow: scroll;
 }
 
 #clif {
-  width: 33%;
+  flex: 1;
   height: 100%;
   overflow: scroll;
 }
 
 #asm {
-  width: 33%;
+  flex: 1;
   height: 100%;
   overflow: scroll;
 }
--- a/crates/explorer/src/index.js
+++ b/crates/explorer/src/index.js
@@ -3,13 +3,14 @@
 /*** State *********************************************************************/
 
 class State {
-  constructor(wat, asm) {
+  constructor(wat, clif, asm) {
     this.wat = wat;
+    this.clif = clif;
     this.asm = asm;
   }
 }
 
-const state = (window.STATE = new State(window.WAT, window.ASM));
+const state = (window.STATE = new State(window.WAT, window.CLIF, window.ASM));
 
 /*** Hues for Offsets **********************************************************/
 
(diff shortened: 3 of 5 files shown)


crates/explorer/src/index.css
```diff
@@ -15,13 +15,19 @@ body {
...
+
+#clif {
+  width: 33%;
   height: 100%;
   overflow: scroll;
 }
 
 #asm {
-  width: 50%;
+  width: 33%;
```
fitzgen, line 30: I don't think we want to apply this CSS all the time, only if we have CLIF to view. If we are doing `wasmtime explore` but using Winch instead of Cranelift, there won't be any CLIF to display, so it doesn't make sense to reserve a third of the screen for it.

I think we can resolve this by using flexbox instead of width percents. That way, if there are three things inside the container, they will each be a third of the width, and if there are two, they will each be half the width.

crates/explorer/src/index.js
```diff
@@ -128,6 +129,28 @@ asmElem.addEventListener(
...
+    if (!watByOffset.get(offset)) {
+      return;
+    }
+
+    const firstWatElem = watByOffset.get(offset)[0];
+    firstWatElem.scrollIntoView({
+      behavior: "smooth",
+      block: "center",
+      inline: "nearest",
+    });
```
fitzgen, line 150: I think we also want to scroll the relevant assembly into view as well, no?

src/commands/explore.rs
```diff
@@ -50,7 +55,35 @@ impl ExploreCommand {
...
-        wasmtime_explorer::generate(&config, self.target.as_deref(), &bytes, &mut output_file)?;
+        let clif_dir = output
+            .parent()
+            .map::<Result<PathBuf>, _>(|output_dir| {
+                let clif_dir = output_dir.join("clif");
+                if let Err(err) = create_dir(&clif_dir) {
+                    match err.kind() {
+                        io::ErrorKind::AlreadyExists => {
+                            return Err(anyhow!("clif directory exists"));
+                        }
```
fitzgen, line 66: I think we can just overwrite the existing files, otherwise doing `wasmtime explore` in the same location twice will be really annoying, since we'd have to remove a bunch of stuff before we could run the command again.

fitzgen: And for what it's worth, we overwrite the html file.

fitzgen: Oh, actually, let's just use a [temp directory](https://docs.rs/tempfile/latest/tempfile/fn.tempdir.html) for this. We already use the `tempfile` crate as a dev-dependency for the `wasmtime` crate, so I think we can also add that as a normal, but optional, dependency and then turn the dependency on when the `explore` cargo feature is enabled.

fitzgen (commented):
Thanks! This looks really good! Just a few nitpicks below before we merge this.

I'm excited to get a view of the CLIF alongside the Wasm and the asm!

src/commands/explore.rs
vulc41n: > I think we can just overwrite the existing files, otherwise doing `wasmtime explore` in the same location twice will be really annoying, since we'd have to remove a bunch of stuff before we could run the command again.

Actually, it removes the directory afterwards, that's why I thought it was unsafe to delete an existing directory. I will change to use the temp directory

vulc41n:
Thank you for taking the time to review my changes, I hope the code is fine now 🙂

Cargo.toml
```diff
@@ -69,6 +69,7 @@ wasmparser = { workspace = true }
 tracing = { workspace = true }
 log = { workspace = true }
 humantime = { workspace = true }
+tempfile = { workspace = true }
```
fitzgen, line 72: Let's make this an optional dependency and enable it when the `explore` cargo feature is enabled.

```suggestion
tempfile = { workspace = true }
```

```toml
[features]
# ...
explore = ["dep:wasmtime-explorer", "dep:tempfile"]
#...
```

fitzgen (commented):
This is looking really good! I tried running it locally, and when using Cranelift, this is super awesome. However, when using Winch, the explorer no longer works:

```shell-session
$ cargo run -- explore ~/scratch/hello.wasm -C compiler=winch
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.20s
     Running `target/debug/wasmtime explore /home/nick/scratch/hello.wasm -C compiler=winch`
Error: clif output not supported
```

Could you make sure this doesn't break Winch support in `wasmtime explore` before we land this? Thanks!!

vulc41n:
The dependency is now optional.

I'm not sure how to deal with winch. Should I display 2 panes when compiling with winch or add clif emission support to winch ?

fitzgen:
@vulc41n the latter: Winch doesn't have an intermediate IR like CLIF, it just goes directly from Wasm to assembly by design. So when using Winch, and we don't have CLIF available, we want to show the current view, and skip the CLIF pane in the explore UI. This is why we wanted to move to flex css, so that the panes would automatically fill the whole container width regardless whether we have CLIF or not.

Does that all make sense?

vulc41n:
Thanks for the clarification, I've made the change 🙂

In the future, I hope I can help with more advanced stuff but there are a lot of concepts to grasp

fitzgen (approved):
Fantastic, thanks! Appreciate your patience getting all the details sorted out here.

fitzgen:
> In the future, I hope I can help with more advanced stuff but there are a lot of concepts to grasp

Is there a specific area you are interested in getting more involved with? I might be able to find something that could be a good on-ramp for you.

vulc41n:
> > In the future, I hope I can help with more advanced stuff but there are a lot of concepts to grasp
> 
> Is there a specific area you are interested in getting more involved with? I might be able to find something that could be a good on-ramp for you.

Thanks, I appreciate. I was working on a source-to-source compiler, but right now I am unemployed and I just want to complete my experience with lower-level subjects. Anything about cranelift or winch would interest me 😀

fitzgen:
@vulc41n, helping add aarch64 support to Winch could be a good fit for you: https://github.com/bytecodealliance/wasmtime/issues/8321

Each instruction is a relatively small amount of work but gets you more familiar with Wasm, aarch64, and Winch. Winch is simpler and easier to get up and running on than Cranelift. And there are a lot of instructions available for filling out!

vulc41n:
Ok, I will do this 😀

fitzgen:
Feel free to ping me on github or zulip if you have any questions!

PR #8974, opened 2024-07-18
Cranelift: Introduce the `stack` alias region

fitzgen:
And do our usual alias analysis stuff on it.

### DIFF
--- a/cranelift/codegen/src/alias_analysis.rs
+++ b/cranelift/codegen/src/alias_analysis.rs
@@ -80,6 +80,7 @@ pub struct LastStores {
     heap: PackedOption<Inst>,
     table: PackedOption<Inst>,
     vmctx: PackedOption<Inst>,
+    stack: PackedOption<Inst>,
     other: PackedOption<Inst>,
 }
 
@@ -90,6 +91,7 @@ impl LastStores {
             self.heap = inst.into();
             self.table = inst.into();
             self.vmctx = inst.into();
+            self.stack = inst.into();
             self.other = inst.into();
         } else if opcode.can_store() {
             if let Some(memflags) = func.dfg.insts[inst].memflags() {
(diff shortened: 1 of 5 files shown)


bjorn3:
If you have a store with no alias flags and a load with the stack alias flag will they be considered as potentially aliasing? If not, this PR makes it impossible to combine loads and stores on the output of stack_addr with stack_load and stack_store.

Should the docs for stack_load and stack_store be updated to indicate that the stack alias region is used?

elliottt (approved):

fitzgen:
> If you have a store with no alias flags and a load with the stack alias flag will they be considered as potentially aliasing? If not, this PR makes it impossible to combine loads and stores on the output of stack_addr with stack_load and stack_store.
> 
> Should the docs for stack_load and stack_store be updated to indicate that the stack alias region is used?

You can still arrange for them to be combined, but you have to specify the `stack` alias region yourself on the manually-legalized loads and stores.

I'll update the docs tho, good point.

bjorn3:
> You can still arrange for them to be combined, but you have to specify the stack alias region yourself on the manually-legalized loads and stores.

That basically means I can no longer use stack_load and stack_store in cg_clif. A stack_addr result can be passed to arbitrary functions which don't know whether the pointer argument is on the stack or the heap. Maybe you could only apply the stack alias region if there is no stack_addr on the same stack slot anywhere inside the function?

bjorn3:
This would also be the first UB case where the generator of the IR doesn't explicitly opt in. UB for regular loads and stores is fully opt in. Even being able to access the pointer target is not UB if you don't set the right flag to indicate that the load/store must never fail.

cfallin:
Perhaps we could accept an alias region on the stack-slot declaration which is inherited by the `stack_addr`? That way, by default, we retain today's behavior, but producers that control aliasing more tightly (e.g. ensure the pointer doesn't escape) could get better optimization.

fitzgen:
Okay, on reflection let's skip this for the time being. Wasmtime actually has similar issues where we would need to start threading memflags through various places.

PR #8976, opened 2024-07-19
x64: Fix encoding of adc/sbb with immediate operands

alexcrichton:
This isn't reachable from Cranelift today but in some various testing this is one of the issues I encountered. This fixes apparent copy/paste bugs with the `Adc` and `Sbb` instructions with immediate operands.


fitzgen (approved):
Nice find!

PR #8978, opened 2024-07-19
`cranelift-frontend`: Support moving GCs with user stack maps

fitzgen:
This refactors the way that user stack maps work such that they are compatible with moving GCs. We now keep every GC value in a stack slot, spill to the stack slot on the value's definition, and reload from the stack slot on every use. This means that we can generate a bunch of redundant loads, but we let the mid-end's alias analysis clean that up. This has the added benefit of reducing the length of live ranges for GC values and also means we don't need to proactively spill every single live GC values at each safepoint, because we know they are already spilled. This latter bit actually helps us avoid an accidentally quadratic issue with many, long, overlapping live ranges where we would do `O(n^2)` spills for a series of safepoints that keep creating more GC refs.

We also implement two further optimizations:

1. We lazily allocate slots for live GC values, which means that if a GC value is not ever live across a safepoint, then we do never allocate a stack slot for it and never spill it to the stack slot. If we didn't do this, then frames would be larger and we'd have a dead store to the stack slot that would otherwise require the mid-end to grow a dead-store-elimination pass.

2. We keep a free list of available slots that will never be used again, and we reuse slots from here when possible. This means that a chain of non-overlapping live ranges, each of which still needs to appear in some safepoint's stack map, will all reuse the same stack slot, keeping frames from being bloated.

Finally, this commit also introduces some logs for the liveness analysis to ease future debugging.

### DIFF
--- a/cranelift/frontend/src/frontend.rs
+++ b/cranelift/frontend/src/frontend.rs
@@ -2,8 +2,9 @@
 use crate::ssa::{SSABuilder, SideEffects};
 use crate::variable::Variable;
 use alloc::collections::BTreeSet;
+use alloc::vec::Vec;
 use core::fmt::{self, Debug};
-use cranelift_codegen::cursor::{Cursor, FuncCursor};
+use cranelift_codegen::cursor::{Cursor, CursorPosition, FuncCursor};
 use cranelift_codegen::entity::{EntityRef, EntitySet, SecondaryMap};
 use cranelift_codegen::ir;
 use cranelift_codegen::ir::condcodes::IntCC;
@@ -272,8 +273,6 @@ impl fmt::Display for DefVariableError {
     }
 }
 
-const LOG2_SIZE_CAPACITY: usize = (16u8.ilog2() as usize) + 1;
-
 /// This module allows you to create a function in Cranelift IR in a straightforward way, hiding
 /// all the complexity of its internal representation.
 ///
@@ -504,6 +503,11 @@ impl<'a> FunctionBuilder<'a> {
             return Err(DefVariableError::TypeMismatch(var, val));
         }
 
+        // If `var` needs inclusion in stack maps, then `val` does too.
+        if self.func_ctx.stack_map_vars.contains(var) {
+            self.declare_value_needs_stack_map(val)
+        }
+
         self.func_ctx.ssa.def_var(var, val, self.position.unwrap());
         Ok(())
     }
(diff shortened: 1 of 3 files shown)


cranelift/frontend/src/frontend.rs
```diff
@@ -504,6 +503,11 @@ impl<'a> FunctionBuilder<'a> {
             return Err(DefVariableError::TypeMismatch(var, val));
         }
 
+        // If `var` needs inclusion in stack maps, then `val` does too.
+        if self.func_ctx.stack_map_vars.contains(var) {
+            self.declare_value_needs_stack_map(val)
```
cfallin, line 508: doesn't make a semantic difference here but style nit: semicolon after call?

cranelift/frontend/src/frontend.rs
```diff
@@ -704,45 +708,36 @@ impl<'a> FunctionBuilder<'a> {
...
+        // needs-stack-map value is never live across a safepoint, then we won't
+        // ever add it to this map, it can remain in a virtual register for the
+        // duration of its lifetime, and we won't replace all its uses with
+        // reloads and all that stuff.
+        let mut stack_slots: crate::HashMap<ir::Value, ir::StackSlot> = Default::default();
+
+        // A map from size/align to free stack slots that are not being used
+        // anymore. This allows us to reuse stack slots across multiple values
+        // helps cut down on the ultimate size of our stack frames.
+        let mut free_stack_slots: crate::HashMap<u32, SmallVec<[ir::StackSlot; 4]>> =
```
cfallin, line 739: If it's not too bad to do, could we have an enum for stack-slot size rather than a bare `u32`? Something like `enum StackSlotSize { Slot4, Slot8 }` and perhaps `.bytes()` and `.align()` methods as needed; and a `TryFrom<ir::Type>` impl maybe too...

(To be clear, I'm suggesting this as a refactor wherever slot-sizes appear, the hashmap key here just prompted the thought)

cranelift/frontend/src/frontend.rs
```diff
@@ -791,28 +804,65 @@ impl<'a> FunctionBuilder<'a> {
...
+                    log::trace!(
+                        "stack map liveness: found safepoint: {inst:?}: {}",
+                        self.func.dfg.display_inst(inst)
+                    );
+                    log::trace!("  live = {live:?}");
+
+                    for val in &live {
+                        let ty = self.func.dfg.value_type(*val);
+                        let slot = *stack_slots.entry(*val).or_insert_with(|| {
+                            let size = self.func.dfg.value_type(*val).bytes();
```
cfallin, line 843: perhaps abstract out to an analogous `process_use` lambda here to make things a bit clearer, as above for defs?

cranelift/frontend/src/frontend.rs
```diff
@@ -831,74 +886,79 @@ impl<'a> FunctionBuilder<'a> {
...
-        }
-        stack_slots
-    }
-
-    /// Insert spills to the given stack slots before each safepoint
-    /// instruction.
-    fn insert_safepoint_spills_to_stack_slots(
+    /// 2. Replace all uses of the needs-stack-map value with loads from that
+    ///    stack slot. This will introduce many redundant loads, but the alias
+    ///    analysis pass in the mid-end can clean all these up.
```
cfallin, line 909: s/can clean these all up/should eventually be able to clean most of these up, when not needed/ (aspirational description is aspirational, hopefully we'll get there! but don't want to confuse in the meantime)

cranelift/frontend/src/frontend.rs
```diff
@@ -831,74 +886,79 @@ impl<'a> FunctionBuilder<'a> {
...
+            pos.next_inst();
+            let mut spilled_any = false;
+            for val in vals.drain(..) {
+                if let Some(slot) = stack_slots.get(&val) {
+                    pos.ins().stack_store(val, *slot, 0);
+                    spilled_any = true;
+                }
+            }
+            if !spilled_any {
+                pos = pos.at_position(CursorPosition::Before(block));
```
cfallin, line 930: Could we have a comment here describing why we need to back up? Is it that we need to pass over the stack-stores we just inserted for blockparams above?

(And given that we do need this, is there any harm in making it unconditional, since it should be a no-op if we didn't insert anything?)

cfallin (approved):
I think this is correct, just some nits below -- I'm glad our idea to fall back to stackslots is working out!

I think we can land as-is if you prefer, but I do have some thoughts spawned from the handling of blockparams. My original thought when I suggested this was that (perhaps naively, but...) it would operate closer to the level of the frontend's "variables" that we use to represent Wasm locals, so that somehow we wouldn't need to handle blockparams -- since by construction, we insert blockparams only to carry the same variable from source block to dest block. I worry that inserting reloads and respills (if I've understood the above correctly) could lead to unexpected worst-case behavior still. (In the `vars_block_params_and_needs_stack_map` test example, `x` would have one stackslot, we would overwrite it in both `block1` and `block2`, and we wouldn't need loads at those blocks' tails and a re-store at `block3`.)

I think that if we have a closer correspondence between frontend variables (i.e. Wasm locals) and stackslots, we have a more understandable cost-model from the Wasm producer's point of view as well: at least in baseline compilers, locals generally become actual stack slots, so producers try to minimize local count and reuse them where possible. If we can carry over that consolidation to our IR, we have an understandable model of "one store per write to a Wasm local, one load per use of a Wasm local" which is great.

As things are now I think we can still see higher-than-linear resulting code size from linear number of reads/writes to refs: imagine a `br_table` that fans out to N blocks, the first (innermost) of which overwrites a reftyped Wasm local with an existing def. Then when we process `end` opcodes and generate N merge-points in sequence, each merge point will have to reload and re-store the slot. Ideally we'd instead directly translate the update to the variable to a `stack_store` and the uses to `stack_load`s.

I think that might point the way to an implementation solution that also avoids the need for the extra pass: branch in the variable-to-SSA-value translation to an alternate "value just lives on stack" path.

Thoughts?

cranelift/frontend/src/frontend.rs
fitzgen: We would actually need quite a few slot sizes, since we have types like `f64x8` in CLIF. I'm not fully convinced that we would actually benefit from such an `enum`, and even if we do want to go down that route, I'd want to do it in a separate PR.

I did waffle back and forth between a full hash map here vs an array indexed by `log2(size)` like the old approach did all over the place. I opted to just go for the simplest, most-obvious implementation approach and figured we can come back and optimize if it is ever a perf bottleneck, but I suspect it really isn't.

So given all that: are you okay with keeping this as-is for now at least?

cfallin: Sure, it doesn't need to happen as part of this PR. (The reason for `TryFrom<ir::Type>` rather than `From` was exactly to address this -- I don't think we need to define a slot size for every type, only for types that the `cranelift-frontend` user puts into stackmaps -- and in general I find bare integers pretty error-prone; but happy to discuss further!)

fitzgen:
> operate closer to the level of the frontend's "variables"

I also thought about creating a new entity for these things but I thought through how it would be used in `cranelift-wasm` and `wasmtime-cranelift` and my conclusion was that it would be a bit of a mess. First, we aren't *just* dealing with wasm locals, which are `Variable`s; we are also dealing with the results of calls and things like that, which produce `Value`s. So we have to have a solution for `Value`s either way. Second, `cranelift-wasm` doesn't make the decision whether certain types are managed by the GC or not, it is agnostic of runtime structures and implementation, `wasmtime-cranelift` makes those decisions based on its Wasmtime-specific knowledge. That means that `cranelift-wasm` would have to have some kind of `enum` of either a GC-managed `Variable` (or other new entity) or a plain `Value` and use that for its abstract operand stack, at which point the whole crate, essentially, is infected with this change and things are getting really hairy.

For this reason, we decided it was better, at least for now, to play nice with the existing system and its existing entities, rather than overhaul everything.

> I think that might point the way to an implementation solution that also avoids the need for the extra pass: branch in the variable-to-SSA-value translation to an alternate "value just lives on stack" path.

Unfortunately this won't work because the stack slots and liveness and all that are not created until we are finishing the function builder, long after calls to `builder.use_var`.

Trevor and I talked about another solution for this specific problem though: if we maintain a backwards map from `Value` to the `Variable` that it is associated with (if any) then we can force all `Value`s for the same `Variable` to use the same stack slot. For `Value`s that are not associated with a `Variable`, they would use the same approach we have in this PR. But this is work for the future.

fitzgen:
> I think we can land as-is if you prefer

I'm going to go ahead and enqueue this for merging, but I am still interested in your responses above and in the inline thread!

cranelift/frontend/src/frontend.rs
fitzgen: Did this in https://github.com/bytecodealliance/wasmtime/pull/8991

PR #8982, opened 2024-07-20
Refactor `Ieee{16,32,64,128}`

beetrees:
Deduplicate the `Ieee{16,32,64,128}` implementations using a macro.

The only methods/trait impls removed are:

- The inherent `neg()` method on `Ieee32` and `Ieee64`: This was redundant to the `Neg` trait implementation (whose method is also called `neg`). Anywhere that want to call `x.neg()` instead of `-x` can still do so by `use`ing `core::ops::Neg`.
- `impl From<u64> for Ieee64`: This only existed for `Ieee64`, was redundant to `Ieee64::with_bits`, and was only used in one place in a test. Using either `Ieee64::with_bits` or the `From<f64> for Ieee64` impl as required is much clearer about intent.

### DIFF
--- a/cranelift/codegen/src/ir/immediates.rs
+++ b/cranelift/codegen/src/ir/immediates.rs
@@ -468,65 +468,338 @@ impl FromStr for Offset32 {
     }
 }
 
-/// An IEEE binary16 immediate floating point value, represented as a u16
-/// containing the bit pattern.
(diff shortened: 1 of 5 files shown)


elliottt (approved):
This is a nice refactoring, thanks!

PR #8983, opened 2024-07-22
Implement wasi-keyvalue

iawia002:
The [wasi-keyvalue](https://github.com/WebAssembly/wasi-keyvalue) API has entered Phase 2. This patch implements the API, giving wasm components access to key-value storages.

Currently, only the Redis storage backend has been implemented, and support for more kv storage (e.g. etcd, etc.) will be added in the future.

cc @alexcrichton

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -197,6 +197,7 @@ jobs:
       audit: ${{ steps.calculate.outputs.audit }}
       preview1-adapter: ${{ steps.calculate.outputs.preview1-adapter }}
       run-dwarf: ${{ steps.calculate.outputs.run-dwarf }}
+      run-wasi-keyvalue: ${{ steps.calculate.outputs.run-wasi-keyvalue }}
     steps:
     - uses: actions/checkout@v4
     - id: calculate
@@ -241,6 +242,9 @@ jobs:
           if grep -q debug names.log; then
             echo run-dwarf=true >> $GITHUB_OUTPUT
           fi
+          if grep -q wasi-keyvalue names.log; then
+            echo run-wasi-keyvalue=true >> $GITHUB_OUTPUT
+          fi
         fi
         matrix="$(node ./ci/build-test-matrix.js ./commits.log ./names.log $run_full)"
         echo "test-matrix={\"include\":$(echo $matrix)}" >> $GITHUB_OUTPUT
@@ -256,6 +260,7 @@ jobs:
             echo audit=true >> $GITHUB_OUTPUT
             echo preview1-adapter=true >> $GITHUB_OUTPUT
             echo run-dwarf=true >> $GITHUB_OUTPUT
+            echo run-wasi-keyvalue=true >> $GITHUB_OUTPUT
         fi
 
   # Build all documentation of Wasmtime, including the C API documentation,
@@ -728,6 +733,44 @@ jobs:
 
     # Run the tests!
     - run: cargo test -p wasmtime-wasi-nn --features ${{ matrix.feature }}
+      env:
+        RUST_BACKTRACE: 1
+
+    # common logic to cancel the entire run if this job fails
+    - run: gh run cancel ${{ github.run_id }}
+      if: failure() && github.event_name != 'pull_request'
+      env:
+        GH_TOKEN: ${{ github.token }}
+
+  # Test the `wasmtime-wasi-keyvalue` crate. Split out from the main tests
+  # because it needs additional database service.
+  test_wasi_keyvalue:
+    name: Test wasi-keyvalue
+    runs-on: ubuntu-latest
+    needs: determine
+    if: needs.determine.outputs.run-wasi-keyvalue
+    # Setup redis server
+    services:
+      redis:
+        # Docker Hub image
+        image: redis
+        # Set health checks to wait until redis has started
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          # Maps port 6379 on service container to the host
+          - 6379:6379
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        submodules: true
+    - uses: ./.github/actions/install-rust
+    # Install Rust targets
+    - run: rustup target add wasm32-wasi
+    - run: cargo test --all-features -p wasmtime-wasi-keyvalue
       env:
         RUST_BACKTRACE: 1
 
--- a/crates/wasi-keyvalue/src/lib.rs
+++ b/crates/wasi-keyvalue/src/lib.rs
@@ -1,0 +1,416 @@
+//! # Wasmtime's [wasi-keyvalue] Implementation
+//!
+//! This crate provides a Wasmtime host implementation of the [wasi-keyvalue]
+//! API. With this crate, the runtime can run components that call APIs in
+//! [wasi-keyvalue] and provide components with access to key-value storages.
+//!
+//! Currently supported storage backends:
+//! * In-Memory (empty identifier)
+//! * Redis, supported identifier format:
+//!   * `redis://[<username>][:<password>@]<hostname>[:port][/<db>]`
+//!   * `redis+unix:///<path>[?db=<db>[&pass=<password>][&user=<username>]]`
+//!
+//! # Examples
+//!
+//! The usage of this crate is very similar to other WASI API implementations
+//! such as [wasi:cli] and [wasi:http].
+//!
+//! A common scenario is accessing redis in a [wasi:cli] component.
+//! A standalone example of doing all this looks like:
+//!
+//! ```
+//! use wasmtime::{
+//!     component::{Linker, ResourceTable},
+//!     Config, Engine, Result, Store,
+//! };
+//! use wasmtime_wasi::{WasiCtx, WasiCtxBuilder, WasiView};
+//! use wasmtime_wasi_keyvalue::{WasiKeyValue, WasiKeyValueCtx, WasiKeyValueCtxBuilder};
+//!
+//! #[tokio::main]
+//! async fn main() -> Result<()> {
+//!     let mut config = Config::new();
+//!     config.async_support(true);
+//!     let engine = Engine::new(&config)?;
+//!
+//!     let mut store = Store::new(&engine, Ctx {
+//!         table: ResourceTable::new(),
+//!         wasi_ctx: WasiCtxBuilder::new().build(),
+//!         wasi_keyvalue_ctx: WasiKeyValueCtxBuilder::new().build(),
+//!     });
+//!
+//!     let mut linker = Linker::<Ctx>::new(&engine);
+//!     wasmtime_wasi::add_to_linker_async(&mut linker)?;
+//!     // add `wasi-runtime-config` world's interfaces to the linker
+//!     wasmtime_wasi_keyvalue::add_to_linker(&mut linker, |h: &mut Ctx| {
+//!         WasiKeyValue::new(&h.wasi_keyvalue_ctx, &mut h.table)
+//!     })?;
+//!
+//!     // ... use `linker` to instantiate within `store` ...
+//!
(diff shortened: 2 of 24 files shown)


bjorn3:
Would it make sense to have a local storage method too? That would allow quick tests without requiring the setup of an entire database.

iawia002:
> Would it make sense to have a local storage method too? That would allow quick tests without requiring the setup of an entire database.

Sure, I'll implement an in-memory backend and make the other database backends optional.

iawia002:
I will fix the dead-code warning, but I have no idea why the [runtime_config_get test failed](https://github.com/bytecodealliance/wasmtime/actions/runs/10038599012/job/27740867443?pr=8983):

```
---- runtime_config_get stdout ----
thread 'runtime_config_get' panicked at crates/wasmtime/src/runtime/vm/libcalls.rs:147:5:
internal error: entered unreachable code
```

iawia002:
> I will fix the dead-code warning, but I have no idea why the [runtime_config_get test failed](https://github.com/bytecodealliance/wasmtime/actions/runs/10038599012/job/27740867443?pr=8983):
> 
> ```
> ---- runtime_config_get stdout ----
> thread 'runtime_config_get' panicked at crates/wasmtime/src/runtime/vm/libcalls.rs:147:5:
> internal error: entered unreachable code
> ```

I have figured it out (#8997) and made a temporary solution by adding the `wasmtime/wmemcheck` feature to `wasi-runtime-config`:

```yaml
[features]
# Do not use, this is only used for testing in CI,
# see https://github.com/bytecodealliance/wasmtime/issues/8997
test = ["wasmtime/wmemcheck"]
```

.github/workflows/main.yml
```diff
@@ -729,6 +734,35 @@ jobs:
...
+    if: needs.determine.outputs.run-wasi-keyvalue
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        submodules: true
+    - uses: ./.github/actions/install-rust
+    # Install Rust targets
+    - run: rustup target add wasm32-wasi
+    # Setup redis server
+    - uses: shogo82148/actions-setup-redis@v1
```
alexcrichton, line 755: Is there perhaps a more "official" action to use here to setup redis? We in theory try to avoid third-party dependencies on actions because of security concerns, but if it's a complicated and/or widely-enough-used action we can make an exception.

crates/wasi-keyvalue/src/lib.rs
```diff
@@ -0,0 +1,353 @@
...
+        Self { ctx, table }
+    }
+}
+
+#[async_trait]
+impl keyvalue::store::Host for WasiKeyValue<'_> {
+    async fn open(&mut self, identifier: String) -> Result<Resource<Bucket>, Error> {
+        if identifier == "" {
+            return Ok(self.table.push(Bucket {
+                inner: Box::new(provider::inmemory::InMemory::default()),
```
alexcrichton, line 211: Since `::default()` is used here it looks like there's no way to configure this in-memory configuration?

crates/wasi-keyvalue/src/lib.rs
```diff
@@ -0,0 +1,353 @@
...
+                    ));
+                }
+                #[cfg(feature = "redis")]
+                {
+                    let host = provider::redis::open(
+                        identifier,
+                        self.ctx.redis_response_timeout,
+                        self.ctx.redis_connection_timeout,
+                    )
+                    .await?;
```
alexcrichton, line 236: This exposes the ability for wasm to make a connection to any redis server it wants without the host being able to intervene. In general we try to model the implementations of WASI proposals to deny access to things by default with explicit opt-ins to provide access. 

Perhaps as an alternative design, could the configuration for `WasiKeyValueCtx` look along the lines of one of:

* Provides are added ahead-of-time to `WasiKeyValueCtx` which means the host has an allow-listed set of provides each component gets access to.
* The host has a callback or trait which receives the `open` call here and returns a `Box<dyn Host>`? That would enable the host to customize whether it wants the logic this method has implemneted or something more custom.

crates/wasi-keyvalue/src/provider/inmemory.rs
```diff
@@ -0,0 +1,84 @@
...
+        })
+    }
+
+    async fn increment(&mut self, key: String, delta: u64) -> Result<u64, Error> {
+        let mut store = self.store.lock().unwrap();
+        let value = store.entry(key.clone()).or_insert(vec![0]);
+        let current_value = String::from_utf8(value.clone())
+            .unwrap_or_else(|_| "0".to_string())
+            .parse::<u64>()
+            .unwrap_or(0);
```
alexcrichton, line 52: It loosk like this is silently ignoring non-integer previous values and converting them to 0?

alexcrichton (commented):
This is looking quite good to me, thanks for the PR and the work here! The biggest question below I have is about how buckets are opened.

crates/wasi-keyvalue/src/lib.rs
iawia002: Yes, the in-memory implementation is very simple. Currently, it doesn't have any configurable features, so configuration is not supported here.

crates/wasi-keyvalue/src/provider/inmemory.rs
iawia002: Fixed, now, non-integer data conversion errors will be returned directly.

.github/workflows/main.yml
iawia002: I am now using GitHub's [Redis service containers](https://docs.github.com/en/actions/using-containerized-services/creating-redis-service-containers) to start a Redis service, but the drawback is that it can only run on Linux. Therefore, the previous macOS testing has been removed, but I think this is an acceptable compromise.

crates/wasi-keyvalue/src/lib.rs
iawia002: I chose the whitelist strategy and added a general configuration method that applies to all providers. By default, servers at all addresses are blocked from connecting. Which addresses the components can connect to depends on the runtime configuration.

iawia002:
All comments addressed ec3373a, PTAL

.github/workflows/main.yml
alexcrichton: Yeah that sounds reasonable to me too, thanks!

crates/wasi-keyvalue/src/lib.rs
alexcrichton: I'm at least personally a bit hesitant to include this when it doesn't have any support/tests/etc. Would it be possible to hook this up to the configuration of `WasiKeyValueCtx`? Or somehow expose the ability to configure this? Otherwise I think it'd be better to remove the internals of `InMemory` and make it an empty struct which is a noop implementation of the trait.

crates/wasi-keyvalue/src/lib.rs
```diff
@@ -0,0 +1,416 @@
...
+    /// ```
+    /// use wasmtime_wasi_keyvalue::WasiKeyValueCtxBuilder;
+    ///
+    /// # fn main() {
+    /// let ctx = WasiKeyValueCtxBuilder::new()
+    ///     .allow_hosts(&["localhost:1234", "/var/run/redis.sock"])
+    ///     .build();
+    /// # }
+    /// ```
+    pub fn allow_hosts(mut self, hosts: &[impl AsRef<str>]) -> Self {
```
alexcrichton, line 174: Could this be renamed to `allow_redis_hosts`?

crates/wasi-keyvalue/src/lib.rs
iawia002: I added a configuration option to the In-Memory provider that allows presetting some data. I believe this provider is valuable, especially in testing scenarios, where users can use this provider to mock some data without relying on a real database. PTAL if this is reasonable.

iawia002:
All comments addressed b9d2bd2, PTAL.

The CI failed due to the feature combinations issue we encountered before. Since the changes are not related to this PR, I have submitted a separate PR #9016 to address that issue. Once that PR is merged, I will rebase this PR, and the CI should be happy then.

alexcrichton (approved):
Looks good to me, thanks! I'll queue for merge after a rebase makes CI happy

(thanks for helping to debug CI and fix the issue!)

iawia002:
Hi @alexcrichton, CI is happy now except `cargo vet`, I believe that should be handled by you.

thomastaylor312:
Hey there! wasi-keyvalue champion here. So I don't think any of the champions knew this was landing. I'm super excited someone implemented it here (thanks @iawia002)! However, we are in the process of incorporating a last round of feedback now that we've had people implementing against it: https://github.com/WebAssembly/wasi-keyvalue/pull/46. We might want to hold off on releasing this until we can make those changes and go to a full 0.2.0 release of wasi-keyvalue

iawia002:
Hi @thomastaylor312, basically, when a proposal enters Phase 2, we can start implementing it. If we wait for it to be completely ready before we begin, the timeline would become excessively long. Therefore, we also have corresponding implementations for draft versions of proposals, allowing users to start experimenting with the API.

For the `wasi-keyvalue`, the implementation targets a specific tag [219ea36](https://github.com/WebAssembly/wasi-keyvalue/commit/219ea3612a53f1bf5b2d137551b22d0268fd3c58). I will continue to track the latest changes in this proposal to keep the implementation up to date.

https://github.com/bytecodealliance/wasmtime/blob/ba864e987ef1ab87c439ca6b396264547d2425e1/ci/vendor-wit.sh#L61

Regarding the release issue, strictly speaking, this has not yet been landed. According to the [wasmtime release schedule](https://github.com/bytecodealliance/wasmtime/blob/main/docs/contributing-release-process.md#releasing-a-major-version), the code will be frozen on the 5th and will not be released until the 20th of this month. So we still have time to follow up on the latest changes in the `wasi-keyvalue` proposal.

Overall, I think we can go ahead and release this, allowing users of wasmtime v24 to try out the `wasi-keyvalue@0.2.0-draft` API.

> but I don't think we want to take hard deps on other KV stores in something like wasmtime. This was something that we had loosely talked about as being "host plugins" using something like wRPC

I agree with this, there are too many KV storage solutions. But I'm not sure if we should move all KV storage implementations out of wasmtime, because depending on Redis/etcd/Memcached, etc., does not affect `wasmtime` itself. It only affects the standalone crate `wasmtime-wasi-keyvalue` and the `wasmtime-cli`. I hope to implement some very popular providers in wasmtime and turn the others into "host plugins".

thomastaylor312:
Yeah we chatted about this today in the wasmtime meeting. So I think we are ok with the current pinned tag, especially since things are behind a feature flag.

However, I still think keeping Redis in tree is a bad idea. The reason why even the super popular ones should be a plugin type model is that now any time the redis client has a bug, you have to update the keyvalue crate and the CLI just to get the bug fix. I know this could happen with any crate, but with client APIs, these kinds of fixes happen quite often. If it happens to be security sensitive, it would necessitate a release of the crate and CLI for anyone depending on it. This wouldn't happen if this was done by loading a plugin separately

iawia002:
I understand your concerns. I am not sure what the "host plugins" will look like exactly. Since you mentioned wrpc, I assume it will involve a server and a client? If we turn all in-tree providers into plugins, would the user have to go through the following additional steps to access Redis?

1. Deploy the plugin server
2. Import the plugin

Compared to releasing patch versions for `wasmtime-wasi-keyvalue` and `wasmtime-cli`, I am more worried about the impact on usability. Will this be too difficult for users, especially the step of deploying a separate plugin server?

Anyway, before the 0.2 release of `wasi-keyvalue`, I don't plan to add more providers. I originally intended to consider these after the official release, but now it seems this will be on hold until we decide how to handle these providers.

iawia002:
I have now seen the [meeting notes](https://github.com/bytecodealliance/meetings/blob/main/wasmtime/2024/wasmtime-08-01.md). I apologize for any inconvenience caused by not notifying you during the implementation. I will submit a PR to revert the introduction of Redis.

PR #8990, opened 2024-07-22
Add v128.const support to Winch

jeffcharles:
Adding support for `v128.const` to Winch. Start of work to add SIMD support to Winch for the x64 architecture.


tests/all/func.rs
```diff
@@ -1966,7 +1966,7 @@ fn typed_v128(config: &mut Config) -> anyhow::Result<()> {
     Ok(())
 }
 
-#[wasmtime_test(wasm_features(simd))]
+#[wasmtime_test]
```
jeffcharles, line 1969: Not sure if we want to keep the `wasm_features(simd)` check on here since this is a SIMD test. I removed it so the test would run on Winch.

tests/disas/winch/x64/br_if/save_state_before_br_emission.wat
```diff
@@ -86,15 +86,13 @@
 ;;       movss   0x23(%rip), %xmm15
 ;;       movss   %xmm15, (%rsp)
 ;;       movq    4(%rsp), %rax
-;;       movl    (%rsp), %r11d
+;;       movss   (%rsp), %xmm15
 ;;       addq    $4, %rsp
-;;       movl    %r11d, (%rax)
+;;       movss   %xmm15, (%rax)
```
jeffcharles, line 91: This is a result of changing how memory values are popped from the stack. It now uses the type of the Wasm value to determine the scratch register to use so this now uses the XMM scratch register instead of a GPR.

tests/disas/winch/x64/load/grow_load.wat
```diff
@@ -74,29 +74,31 @@
...
-;;       popq    %r11
-;;       movq    %r11, (%rax)
-;;       movl    (%rsp), %r11d
+;;       movsd   (%rsp), %xmm15
+;;       addq    $8, %rsp
+;;       movsd   %xmm15, (%rax)
+;;       movss   (%rsp), %xmm15
 ;;       addq    $4, %rsp
-;;       movl    %r11d, 8(%rax)
+;;       movss   %xmm15, 8(%rax)
```
jeffcharles, line 89: Ditto for this change to the disass.

winch/codegen/src/codegen/context.rs
```diff
@@ -191,7 +193,7 @@ impl<'a> CodeGenContext<'a> {
                 masm.store(scratch.into(), addr, size);
             }
             Val::Memory(_) => {
-                let scratch = scratch!(M);
+                let scratch = scratch!(M, &ty);
```
jeffcharles, line 196: This now needs to handle 128 bit values so using the GPR scratch won't work.

winch/codegen/src/isa/x64/masm.rs
```diff
@@ -1043,6 +1047,7 @@ impl MacroAssembler {
     fn store_impl(&mut self, src: RegImm, dst: Address, size: OperandSize, flags: MemFlags) {
         let scratch = <Self as Masm>::ABI::scratch_reg();
         let float_scratch = <Self as Masm>::ABI::float_scratch_reg();
+        let vector_scratch = <Self as Masm>::ABI::vector_scratch_reg();
```
saulecabrera, line 1050: I've been giving a bit of thought to this change. If I'm not wrong the motivation is to future-proof, be explicit and correctly differentiate the scratch register for different types (GPR, FPR) and sizes (FPR vs v128). 

I'm wondering if there's an alternative way that would allow a similar outcome without need to introduce more `*_scratch_reg` variations. One option that I can think of is to remove `scratch_reg` and `float_scratch_reg` and rely exclusively on `scratch_for` and `sratch!(M, &ty)` method and macro respectively. This might make it easier to encapsulate the details of which register should be used depending on the type/size/architecture.

saulecabrera (commented):
It's going on a good direction!


One thing that we need to also update as part of this PR is the stack alignment details in the ABI.

Currently, the [stack alignment](https://github.com/bytecodealliance/wasmtime/blob/main/winch/codegen/src/isa/x64/abi.rs#L66) and [slot size](https://github.com/bytecodealliance/wasmtime/blob/main/winch/codegen/src/isa/x64/abi.rs#L156) are both set to 8 bytes, given that types > 8 bytes were not supported. We'd probably want to ensure that it's now set to 16.

winch/codegen/src/isa/x64/masm.rs
jeffcharles: Yeah I debated whether to just use the float scratch register directly but that seemed like it would be mixing different levels of abstraction. Changing everything to use `scratch_for` or `scratch!(M, &ty)` instead makes sense to me. I'll make those changes. Might try doing that in a separate PR to keep it easier to review.

saulecabrera: In my original comment I was leaning more towards the first option from your comment above: moving `scratch_for` to each ABI implementation, hoping to (i) get the most flexibility (architecture-wise) at the cost of some repetition (ii) reduce the number of scratch related methods from the ABI trait.

jeffcharles: Opened https://github.com/bytecodealliance/wasmtime/pull/8999 to do the refactoring.

winch/codegen/src/isa/x64/abi.rs
```diff
@@ -202,7 +207,10 @@ impl X64ABI {
                 ty,
             ),
 
-            ty => unimplemented!("Support for argument of WasmType: {ty}"),
+            ty @ WasmValType::V128 => (
+                Self::vector_reg_for(index_env.next_vr(), call_conv, params_or_returns),
```
jeffcharles, line 211: I could change this to just reuse the existing floating point register methods given this is inside the x64 ABI if that would be preferable.

jeffcharles:
The merge commit removed the new vector method on the ABI trait. Not sure if you would like me to do the same with the `next_vr` and associated methods I've added in the X64ABI file.

saulecabrera:
> The merge commit removed the new vector method on the ABI trait. Not sure if you would like me to do the same with the `next_vr` and associated methods I've added in the X64ABI file.

Yeah it sounds reasonable to me to get rid of `next_vr` and `vector_reg_for` 👍🏽

saulecabrera (approved):
This looks great to me, thanks!

PR #8992, opened 2024-07-22
Run `cargo publish --dry-run` instead of `cargo package`

### LINKED ISSUE #8987: Publish of 23.0.0 to crates.io failed
Logs at https://github.com/bytecodealliance/wasmtime/actions/runs/10043286629/job/27755712200

attempted fix at https://github.com/bytecodealliance/wasmtime/pull/8986

Filing this to track improvements to `./publish verify` to prevent this from happening in the future.

alexcrichton:
This commit updates the `./scripts/publish.rs` script used on CI to manage this workspace to run `cargo publish --dry-run`. That seems to have a few more checks than `cargo package` and would have caught one failure condition from #8987. The missing categories failure condition can't be caught by Cargo right now without actually publishing, so that'll have to be left for a future fix on a future day.

Closes #8987

### DIFF
--- a/scripts/publish.rs
+++ b/scripts/publish.rs
@@ -507,7 +507,7 @@ fn verify(crates: &[Crate]) {
         cmd.arg("publish")
             .arg("--manifest-path")
             .arg(&krate.manifest)
-            .arg("--dry-run");
+            .arg("--dry-run")
             .env("CARGO_TARGET_DIR", "./target");
         if krate.name.contains("wasi-nn") {
             cmd.arg("--no-verify");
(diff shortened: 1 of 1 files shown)


fitzgen (approved):
Do we still need to `rm ./.cargo/` and do all the vendoring and stuff after this change? I'm pretty sure I lost a `.cargo/config.toml` once because I was debugging the publish script...

alexcrichton:
All that should still be the same yeah, no difference from before. It wouldn't be too hard though to add some checks about warning of overwriting local changes to the script though.

alexcrichton:
Alas this won't work with Cargo I think, so we'll just have to eat the errors here and be more vigilant of new crates in the future. (aka we will all promptly forget about this until it bites us again)

PR #9000, opened 2024-07-23
Fix a safepoints bug involving block parameters and cursor positions

fitzgen:
We weren't setting the cursor position correctly after inserting spills for block parameters that need inclusion in stack maps, which caused us to accidentally skip over the first instruction in the block sometimes.

Also move some tests around and add logging for the rewriting phase of user stack maps (see each separate commit).


fitzgen:
heyyy #9000!!

elliottt (approved):
Makes sense to me!

PR #9003, opened 2024-07-24
Cranelift(x64): Fix lowering for `store(bitop(x, load(addr)), addr)` for bitops on floats

fitzgen:
x86-64 allows us to do these kinds of read-modify-write operations in one instruction in general, however we also need to ensure that the non-memory operand is in a GPR. Because Cranelift allows `b{and,or,xor}`s on floating point types, that means we might need to insert a move from an XMM to a GPR.

Found in https://github.com/bytecodealliance/wasmtime/pull/8941 but is actually unrelated to safepoints and stack maps, they just made this pattern more likely to occur. AFAICT, Wasm can't ever trigger this bug, and would only be something of concern for other CLIF producers.

### DIFF
--- a/cranelift/codegen/src/isa/x64/inst.isle
+++ b/cranelift/codegen/src/isa/x64/inst.isle
@@ -1629,10 +1629,22 @@
 
 ;; Put a value into a GPR.
 ;;
-;; Asserts that the value goes into a GPR.
+;; Moves the value into a GPR if it is a type that would naturally go into an
+;; XMM register.
 (decl put_in_gpr (Value) Gpr)
+
+;; Case for when the value naturally lives in a GPR.
 (rule (put_in_gpr val)
+      (if-let (value_type ty) val)
+      (if-let (type_register_class (RegisterClass.Gpr $true)) ty)
       (gpr_new (put_in_reg val)))
+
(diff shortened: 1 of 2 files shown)


cranelift/codegen/src/isa/x64/inst.isle
```diff
@@ -1629,11 +1629,23 @@
...
-;; Asserts that the value goes into a GPR.
+;; Moves the value into a GPR if it is a type that would naturally go into an
+;; XMM register.
 (decl put_in_gpr (Value) Gpr)
+
+;; Case for when the value naturally lives in a GPR.
 (rule (put_in_gpr val)
+      (if-let (value_type ty) val)
+      (if-let (type_register_class (RegisterClass.Gpr $true)) ty)
       (gpr_new (put_in_reg val)))
```
elliottt, line 1640: Should this match against `(RegisterClass.Gpr _)` instead, to avoid potentially backtracking here instead of forcing the assertion failure in `put_in_reg` like before?

fitzgen: Good idea.

elliottt (approved):

PR #9005, opened 2024-07-24
Cranelift(x64): provide better panic messages for `Gpr::new(reg).unwrap()` et al

fitzgen:
1. Add `Gpr::unwrap_new` and etc... constructors inside the macro that defines all the newtypes. These have better error messages on failure than `Gpr::new(r).unwrap()` would.

2. Use those new constructors everywhere we were otherwise doing `Gpr::new(r).unwrap()` and etc...

Note that (2) is purely mechanical, so whoever is assigned to review this can probably just focus on the changes inside the macro.


elliottt (approved):
Much nicer messages!

PR #9007, opened 2024-07-24
cranelift-codegen: no_std support

### LINKED ISSUE #1158: cranelift-codegen no longer builds with no_std
I've tracked down a few things that are preventing this from happening.

1. `packed_struct` requires std by default
2. There are a few uses of std in `cranelift-codegen` and `cranelift-codegen-shared` that can be replaced with core or alloc.
3. `thiserror` requires std, because the Error trait doesn't exist in core or alloc.

The first two are trivial to fix, and I'll create a PR with those changes in a second.

However, there are 11 uses of the `Display` and `Into<result::CodegenError>` traits that prevent me from placing `thiserror` behind a feature gate, and I'm not sure what to do about them.

<details>
<summary>Here are the 11 things that rely on those traits, as reported by rustc (click to expand)</summary>

    $ cargo build --no-default-features --features core
    Compiling cranelift-codegen v0.51.0 (/home/icydefiance/Documents/code/cranelift/cranelift-codegen)
    error[E0599]: no method named `to_string` found for type `verifier::VerifierError` in the current scope
    --> cranelift-codegen/src/print_errors.rs:218:36
        |
    218 |     writeln!(w, "; error: {}", err.to_string())?;
        |                                    ^^^^^^^^^ method not found in `verifier::VerifierError`
        | 
    ::: cranelift-codegen/src/verifier/mod.rs:138:1
        |
    138 | pub struct VerifierError {
        | ------------------------ method `to_string` not found for this
        |
        = help: items from traits can only be used if the trait is implemented and in scope
        = note: the following trait defines an item `to_string`, perhaps you need to implement it:
                candidate bytecodealliance/cranelift#1: `alloc::string::ToString`

    error[E0599]: no method named `to_string` found for type `result::CodegenError` in the current scope
    --> cranelift-codegen/src/print_errors.rs:227:13
        |
    227 |         err.to_string()
        |             ^^^^^^^^^ method not found in `result::CodegenError`
        | 
    ::: cranelift-codegen/src/result.rs:12:1
        |
    12  | pub enum CodegenError {
        | --------------------- method `to_string` not found for this
        |
        = help: items from traits can only be used if the trait is implemented and in scope
        = note: the following trait defines an item `to_string`, perhaps you need to implement it:
                candidate bytecodealliance/cranelift#1: `alloc::string::ToString`

    error[E0277]: `verifier::VerifierError` doesn't implement `core::fmt::Display`
    --> cranelift-codegen/src/verifier/mod.rs:235:33
        |
    235 |             writeln!(f, "- {}", err)?;
        |                                 ^^^ `verifier::VerifierError` cannot be formatted with the default formatter
        |
        = help: the trait `core::fmt::Display` is not implemented for `verifier::VerifierError`
        = note: in format strings you may be able to use `{:?}` (or {:#?} for pretty-print) instead
        = note: required because of the requirements on the impl of `core::fmt::Display` for `&verifier::VerifierError`
        = note: required by `core::fmt::Display::fmt`

    error[E0277]: the trait bound `result::CodegenError: core::convert::From<verifier::VerifierErrors>` is not satisfied
    --> cranelift-codegen/src/context.rs:224:30
        |
    224 |             self.verify(fisa)?;
        |                              ^ the trait `core::convert::From<verifier::VerifierErrors>` is not implemented for `result::CodegenError`
        |
        = note: required by `core::convert::From::from`

    error[E0277]: the trait bound `result::CodegenError: core::convert::From<verifier::VerifierErrors>` is not satisfied
    --> cranelift-codegen/src/context.rs:244:39
        |
    244 |             self.verify_locations(isa)?;
        |                                       ^ the trait `core::convert::From<verifier::VerifierErrors>` is not implemented for `result::CodegenError`
        |
        = note: required by `core::convert::From::from`

    error[E0277]: the trait bound `result::CodegenError: core::convert::From<verifier::VerifierErrors>` is not satisfied
    --> cranelift-codegen/src/regalloc/context.rs:111:28
        |
    111 |                 return Err(errors.into());
        |                            ^^^^^^^^^^^^^ the trait `core::convert::From<verifier::VerifierErrors>` is not implemented for `result::CodegenError`
        |
        = note: required because of the requirements on the impl of `core::convert::Into<result::CodegenError>` for `verifier::VerifierErrors`

    error[E0277]: the trait bound `result::CodegenError: core::convert::From<verifier::VerifierErrors>` is not satisfied
    --> cranelift-codegen/src/regalloc/context.rs:139:28
        |
    139 |                 return Err(errors.into());
        |                            ^^^^^^^^^^^^^ the trait `core::convert::From<verifier::VerifierErrors>` is not implemented for `result::CodegenError`
        |
        = note: required because of the requirements on the impl of `core::convert::Into<result::CodegenError>` for `verifier::VerifierErrors`

    error[E0277]: the trait bound `result::CodegenError: core::convert::From<verifier::VerifierErrors>` is not satisfied
    --> cranelift-codegen/src/regalloc/context.rs:168:28
        |
    168 |                 return Err(errors.into());
        |                            ^^^^^^^^^^^^^ the trait `core::convert::From<verifier::VerifierErrors>` is not implemented for `result::CodegenError`
        |
        = note: required because of the requirements on the impl of `core::convert::Into<result::CodegenError>` for `verifier::VerifierErrors`

    error[E0277]: the trait bound `result::CodegenError: core::convert::From<verifier::VerifierErrors>` is not satisfied
    --> cranelift-codegen/src/regalloc/context.rs:196:28
        |
    196 |                 return Err(errors.into());
        |                            ^^^^^^^^^^^^^ the trait `core::convert::From<verifier::VerifierErrors>` is not implemented for `result::CodegenError`
        |
        = note: required because of the requirements on the impl of `core::convert::Into<result::CodegenError>` for `verifier::VerifierErrors`

    error[E0277]: the trait bound `result::CodegenError: core::convert::From<verifier::VerifierErrors>` is not satisfied
    --> cranelift-codegen/src/regalloc/context.rs:239:28
        |
    239 |                 return Err(errors.into());
        |                            ^^^^^^^^^^^^^ the trait `core::convert::From<verifier::VerifierErrors>` is not implemented for `result::CodegenError`
        |
        = note: required because of the requirements on the impl of `core::convert::Into<result::CodegenError>` for `verifier::VerifierErrors`

    error[E0277]: the trait bound `result::CodegenError: core::convert::From<verifier::VerifierErrors>` is not satisfied
    --> cranelift-codegen/src/regalloc/context.rs:248:17
        |
    248 |             Err(errors.into())
        |                 ^^^^^^^^^^^^^ the trait `core::convert::From<verifier::VerifierErrors>` is not implemented for `result::CodegenError`
        |
        = note: required because of the requirements on the impl of `core::convert::Into<result::CodegenError>` for `verifier::VerifierErrors`

    error: aborting due to 11 previous errors

    Some errors have detailed explanations: E0277, E0599.
    For more information about an error, try `rustc --explain E0277`.
    error: could not compile `cranelift-codegen`.

    To learn more, run the command again with --verbose.
</details>

Additional info:
* The author of `thiserror` doesn't want to add a feature-gate for just the Error trait, but it's possible to provide another implementation of Display (and presumably Into). dtolnay/thiserror#43
* There was some discussion of moving Error to alloc, and a PR was even created to do so, but it was closed because [RFC 2504](https://rust-lang.github.io/rfcs/2504-fix-error.html) will add backtrace functionality that requires std.

marceline-cramer:
Closes #1158, previous context and discussion can be found there. This is a major milestone towards allowing Cranelift to compile code while in embedded or kernel environments.

I do still have a handful of pinch points to work through before this is ready to merge.

First of all, the most major overhaul I've made to the `codegen` crate is to replace `rustc-hash` with `hashbrown` and `ahash`. I'm not a cryptography expert but as far as I can tell, `ahash` does have a security improvement over `FxHash` in terms of improved DOS resistance. More importantly though, `hashbrown` has a feature to use `ahash` by default, which means that replacing `FxHashMap` and `FxHashSet` was as easy as a few invocations of `sed`. However, I believe that this switch in hashing function has caused some of the codegen expected output tests in CI to fail, and the semantics of the codegen have been slightly changed. I could experiment with swapping `rustc-hash` back into `hashbrown` as its hasher implementation locally but I wanted to know if preserving these semantics was important before I undertook another major type substitution throughout the code. Please take a look at the output of the failing CI for more info.

Second of all, on no_std, floating-point arithmetic is not available. To resolve this, I brought in the `core_math` crate, which implements floating-point arithmetic traits for core primitives using libm's implementation of softfloat. However, the function `rounding_ties_even()` is not available in `libm`, and I'm not sure how to go about replacing it. Input would be appreciated. I do know that if the failing codegen output unit tests I discussed are updated to pass with the new output, other tests catch the change in arithmetic semantics and fail.

Finally, there's this `else` block in `src/machinst/vcode.rs` that has some arithmetic with a hard dependency on the `unwind` feature, which is unsupported on `no_std`. I'm not sure what to do with that.

Thank you so much for the help with all of this, I appreciate your time immensely. I'm really happy to answer any questions and update my code as needed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -178,6 +178,7 @@ all = 'allow'
 
 [workspace.dependencies]
 arbitrary = { version = "1.3.1" }
+ahash = { version = "0.8", default-features = false }
 wasmtime-wmemcheck = { path = "crates/wmemcheck", version = "=24.0.0" }
 wasmtime = { path = "crates/wasmtime", version = "24.0.0", default-features = false }
 wasmtime-c-api-macros = { path = "crates/c-api-macros", version = "=24.0.0" }
@@ -226,7 +227,7 @@ cranelift-jit = { path = "cranelift/jit", version = "0.111.0" }
 cranelift-fuzzgen = { path = "cranelift/fuzzgen" }
 cranelift-bforest = { path = "cranelift/bforest", version = "0.111.0" }
 cranelift-bitset = { path = "cranelift/bitset", version = "0.111.0" }
-cranelift-control = { path = "cranelift/control", version = "0.111.0" }
+cranelift-control = { path = "cranelift/control", version = "0.111.0", default-features = false }
 cranelift = { path = "cranelift/umbrella", version = "0.111.0" }
 
 winch-codegen = { path = "winch/codegen", version = "=0.22.0" }
@@ -236,7 +237,7 @@ byte-array-literals = { path = "crates/wasi-preview1-component-adapter/byte-arra
 
 # Bytecode Alliance maintained dependencies:
 # ---------------------------
-regalloc2 = "0.9.3"
+regalloc2 = { version = "0.9.3", default-features = false }
 
 # cap-std family:
 target-lexicon = "0.12.13"
@@ -277,9 +278,9 @@ windows-sys = "0.52.0"
 env_logger = "0.10"
 log = { version = "0.4.8", default-features = false }
 clap = { version = "4.3.12", default-features = false, features = ["std", "derive"] }
-hashbrown = { version = "0.14", default-features = false }
+hashbrown = { version = "0.14" }
 capstone = "0.12.0"
-once_cell = { version = "1.12.0", default-features = false }
+once_cell = { version = "1.12.0", default-features = false, features = ["critical-section"] }
 smallvec = { version = "1.6.1", features = ["union"] }
 tracing = "0.1.26"
 bitflags = "2.0"
@@ -322,7 +323,7 @@ url = "2.3.1"
 humantime = "2.0.0"
 postcard = { version = "1.0.8", default-features = false, features = ['alloc'] }
 criterion = { version = "0.5.0", default-features = false, features = ["html_reports", "rayon"] }
-rustc-hash = "1.1.0"
+rustc-hash = { version = "1.1.0", default-features = false }
 libtest-mimic = "0.7.0"
 semver = { version = "1.0.17", default-features = false }
 
--- a/cranelift/codegen/src/ir/immediates.rs
+++ b/cranelift/codegen/src/ir/immediates.rs
@@ -13,6 +13,9 @@ use core::{i32, u32};
 #[cfg(feature = "enable-serde")]
 use serde_derive::{Deserialize, Serialize};
 
+#[cfg(not(feature = "std"))]
+use core_maths::CoreFloat;
+
 /// Convert a type into a vector of bytes; all implementors in this file must use little-endian
 /// orderings of bytes to match WebAssembly's little-endianness.
 pub trait IntoBytes {
@@ -972,7 +975,8 @@ impl Ieee32 {
     /// Returns the nearest integer to `self`. Rounds half-way cases to the number
     /// with an even least significant digit.
     pub fn round_ties_even(self) -> Self {
-        Self::with_float(self.as_f32().round_ties_even())
+        // TODO: find a suitable alternative to std round_ties_even() in no_std
+        Self::with_float(self.as_f32().round())
     }
 }
 
@@ -1189,7 +1193,8 @@ impl Ieee64 {
     /// Returns the nearest integer to `self`. Rounds half-way cases to the number
     /// with an even least significant digit.
     pub fn round_ties_even(self) -> Self {
-        Self::with_float(self.as_f64().round_ties_even())
+        // TODO: find a suitable alternative to std round_ties_even() in no_std
+        Self::with_float(self.as_f64().round())
     }
 }
 
@@ -1828,7 +1833,7 @@ mod tests {
     fn fcvt_to_sint_negative_overflow_ieee32() {
         for n in &[8, 16] {
             assert_eq!(-((1u32 << (n - 1)) as f32) - 1.0, unsafe {
-                mem::transmute(Ieee32::fcvt_to_sint_negative_overflow(*n))
+                mem::transmute::<_, f32>(Ieee32::fcvt_to_sint_negative_overflow(*n))
             });
         }
     }
@@ -1966,7 +1971,7 @@ mod tests {
     fn fcvt_to_sint_negative_overflow_ieee64() {
         for n in &[8, 16, 32] {
             assert_eq!(-((1u64 << (n - 1)) as f64) - 1.0, unsafe {
-                mem::transmute(Ieee64::fcvt_to_sint_negative_overflow(*n))
+                mem::transmute::<_, f64>(Ieee64::fcvt_to_sint_negative_overflow(*n))
             });
         }
     }
--- a/cranelift/codegen/src/lib.rs
+++ b/cranelift/codegen/src/lib.rs
@@ -8,6 +8,10 @@
 // built for one platform we don't have to worry too much about trimming
 // everything down.
 #![cfg_attr(not(feature = "all-arch"), allow(dead_code))]
+// Each architecture has several unwind-specific imports that are unused when
+// the unwind feature is not selected. This is much cleaner to work with than
+// having `use` statements guarded by `#[cfg(feature = "unwind")]`.
+#![cfg_attr(not(feature = "unwind"), allow(unused_imports))]
 
 #[allow(unused_imports)] // #[macro_use] is required for no_std
 #[macro_use]
@@ -17,10 +21,12 @@ extern crate alloc;
 #[macro_use]
 extern crate std;
 
+// Supports compiling ISLE.
 #[cfg(not(feature = "std"))]
+#[macro_use]
+extern crate core as std;
+
 use hashbrown::{hash_map, HashMap};
-#[cfg(feature = "std")]
-use std::collections::{hash_map, HashMap};
 
 pub use crate::context::Context;
 pub use crate::value_label::{LabelValueLoc, ValueLabelsRanges, ValueLocRange};
--- a/cranelift/codegen/src/machinst/vcode.rs
+++ b/cranelift/codegen/src/machinst/vcode.rs
@@ -25,17 +25,14 @@ use crate::timing;
 use crate::trace;
 use crate::CodegenError;
 use crate::{LabelValueLoc, ValueLocRange};
+use core::fmt;
+use core::mem::take;
+use cranelift_entity::{entity_impl, Keys};
+use hashbrown::{hash_map::Entry, HashMap};
 use regalloc2::{
     Edit, Function as RegallocFunction, InstOrEdit, InstRange, MachineEnv, Operand,
     OperandConstraint, OperandKind, PRegSet, RegClass,
 };
-use rustc_hash::FxHashMap;
-
-use core::mem::take;
-use cranelift_entity::{entity_impl, Keys};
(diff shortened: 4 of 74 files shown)


bjorn3:
rustc-hash is no_std compatible. You just need to disable the std feature.

cranelift/codegen/src/ir/immediates.rs
```diff
@@ -1189,7 +1193,8 @@ impl Ieee64 {
     /// Returns the nearest integer to `self`. Rounds half-way cases to the number
     /// with an even least significant digit.
     pub fn round_ties_even(self) -> Self {
-        Self::with_float(self.as_f64().round_ties_even())
+        // TODO: find a suitable alternative to std round_ties_even() in no_std
+        Self::with_float(self.as_f64().round())
```
afonso360, line 1197: I looked at the [source for `round_ties_even`](https://doc.rust-lang.org/src/std/f64.rs.html#127-129) and they use `rintf`. It looks like this is [due to portability between platforms](https://github.com/rust-lang/rust/pull/95317#issuecomment-1306215792), but it looks like it gives the same results if we don't change rounding modes. `libm` has support for `rint{f,}` so we can probably use that.

Cargo.toml
```diff
@@ -277,9 +278,9 @@ windows-sys = "0.52.0"
 env_logger = "0.10"
 log = { version = "0.4.8", default-features = false }
 clap = { version = "4.3.12", default-features = false, features = ["std", "derive"] }
-hashbrown = { version = "0.14", default-features = false }
+hashbrown = { version = "0.14" }
 capstone = "0.12.0"
-once_cell = { version = "1.12.0", default-features = false }
+once_cell = { version = "1.12.0", default-features = false, features = ["critical-section"] }
```
alexcrichton, line 283: We'll want to find alternative solutions to this since personally I'd consider the `critical-section` feature as inappropriate for our use cases. That enables spin locks which just keep spinning which are not suitable anywhere outside of a kernel-with-interrupts-disabled context, so the dependencies on `once_cell` will need to be reworked and/or have other solutions than "just" enabling this feature.

cranelift/codegen/src/ir/immediates.rs
alexcrichton: We'll want to be careful though such that where `std` is enabled we want the same implementation cranelift has today which is using presumably optimal routines in the standard library. Only when that feature is disabled should it fall back to the possibly-slower routines based in `libm`

cranelift/codegen/src/lib.rs
```diff
@@ -17,10 +21,12 @@ extern crate alloc;
 #[macro_use]
 extern crate std;
 
+// Supports compiling ISLE.
 #[cfg(not(feature = "std"))]
+#[macro_use]
+extern crate core as std;
```
alexcrichton, line 27: Oh this shouldn't be done in my opinion, it can create a lot of confusion by renaming core crates to each other.

I'd recommend following the pattern of Wasmtime crates which is to do:

```rust
#[cfg(feature = "std")]
#[macro_use]
extern crate std;
```

cranelift/codegen/src/machinst/vcode.rs
```diff
@@ -1216,15 +1213,21 @@ impl<I: VCodeInst> VCode<I> {
...
-                let cfa_to_sp_offset =
-                    -((slot_base_to_caller_sp_offset + caller_sp_to_cfa_offset) as i64);
-                LabelValueLoc::CFAOffset(cfa_to_sp_offset + slot_offset)
+                #[cfg(feature = "unwind")]
+                {
+                    let slot = alloc.as_stack().unwrap();
+                    let slot_offset = self.abi.get_spillslot_offset(slot);
+                    let slot_base_to_caller_sp_offset = self.abi.slot_base_to_caller_sp_offset();
+                    let caller_sp_to_cfa_offset =
+                        crate::isa::unwind::systemv::caller_sp_to_cfa_offset();
```
alexcrichton, line 1222: The definition of this function is "return 0" so I think it's fine to lift this function out of `unwind` and move it somewhere else.

cranelift/codegen/src/timing.rs
```diff
@@ -2,12 +2,13 @@
...
-use std::cell::RefCell;
-use std::mem;
-use std::time::Duration;
+use alloc::boxed::Box;
+use alloc::fmt;
+use core::any::Any;
+use core::time::Duration;
+
+#[cfg(feature = "timing")]
+use core::cell::RefCell;
```
alexcrichton, line 11: What I might recommend for this file is to have a `timing.rs` and a `timing_disabled.rs` as opposed to a single source file with lots of `#[cfg]`, that's worked pretty well for us in the past.

cranelift/isle/isle/src/codegen.rs
```diff
@@ -144,7 +147,13 @@ impl<'a> Codegen<'a> {
...
 
         writeln!(code, "\nuse super::*;  // Pulls in all external types.").unwrap();
-        writeln!(code, "use std::marker::PhantomData;").unwrap();
+
+        if options.no_std {
+            writeln!(code, "use alloc::vec::Vec;").unwrap();
+            writeln!(code, "use core::{{marker::PhantomData, ops}};").unwrap();
+        } else {
+            writeln!(code, "use std::{{marker::PhantomData, ops}};").unwrap();
+        }
```
alexcrichton, line 156: I might echo the desire to remove the `no_std` configuration here entirely perhaps. The generate code can itself contain `extern crate alloc` which can then be used to import `Vec`

alexcrichton (commented):
Thanks again for your work on this! I've left some additional comments to those above below. At a high level I might recommend splitting out the `s/std/core/` rename to a separate PR along with what's needed to handle hash maps (agreed would be best to keep using `rustc-hash`). 

The `once-cell` bits are going to require some careful thought since we can't enable spin locks for the entire workspace. One idea would be to shuffle around [the `sync_nostd.rs` file](https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasmtime/src/sync_nostd.rs) used by Wasmtime since Cranelift looks like it may have a similar use case.

tschneidereit:
@marceline-cramer is there anything we can to do help you get the review comments addressed? Getting this in would be excellent, and you've already put in the major part of the work, after all :)

marceline-cramer:
> @marceline-cramer is there anything we can to do help you get the review comments addressed? Getting this in would be excellent, and you've already put in the major part of the work, after all :)

Apologies for putting this off for so long. I've been capital-B Busy the last couple months. I'll try to provide an answer to that question over the weekend when I can refresh my memory on what precisely is happening here and process the comments.

DrewRidley:
Is there any update on this PR? I am very interested in using Cranelift in a ``no-std`` environment. Happy to help address the concerns or get this PR up to date if needed.

cfallin:
@DrewRidley I'm not sure about this PR in particular but one of the sticking points here was the `OnceCell` usage, which has another pending PR to refactor away in #9992. That one seems to have stalled as well (waiting for updates from the author) but if you want to contribute efforts, you might be able to help move that forward.

alexcrichton:
The cranelift-codegen crate nowadays supports no_std, so I'm going to close this

PR #9008, opened 2024-07-24
Introduce the `pulley-interpreter` crate

fitzgen:
This commit is the first step towards implementing https://github.com/bytecodealliance/rfcs/pull/35

This commit introduces the `pulley-interpreter` crate which contains the Pulley bytecode definition, encoder, decoder, disassembler, and interpreter.

This is still very much a work in progress! It is expected that we will tweak encodings and bytecode definitions, that we will overhaul the interpreter (to, for example, optionally support the unstable Rust `explicit_tail_calls` feature), and otherwise make large changes. This is just a starting point to get the ball rolling.

Subsequent commits and pull requests will do things like add the Cranelift backend to produce Pulley bytecode from Wasm as well as the runtime integration to run the Pulley interpreter inside Wasmtime.

Dedicated CI for Pulley, ensuring that it builds and runs on things like 32-bit `nostd` platforms, will also come in follow ups.


pulley/tests/all/main.rs
```diff
@@ -0,0 +1,7 @@
+#[cfg(feature = "disas")]
+mod disas;
+
+#[cfg(feature = "interp")]
+mod interp;
+
+fn main() {}
```
alexcrichton, line 7: Stray fn main?

pulley/src/regs.rs
```diff
@@ -0,0 +1,180 @@
...
+    /// Is this `x` register a special register?
+    pub fn is_special(&self) -> bool {
+        self.0 >= Self::SPECIAL_RANGE.start
+    }
+}
+
+impl fmt::Display for XReg {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            x if *x == Self::SP => write!(f, "sp"),
```
alexcrichton, line 160: Is it actually possible to do `Self::SP => write!(f, "sp")` here? I forget what we can do with constants-in-patterns...

pulley/src/regs.rs
```diff
@@ -0,0 +1,180 @@
...
+    pub const FP: Self = Self(34);
+
+    /// The special `spilltmp0` scratch register.
+    pub const SPILL_TMP_0: Self = Self(35);
+
+    /// The special `spilltmp1` scratch register.
+    pub const SPILL_TMP_1: Self = Self(36);
+
+    /// Is this `x` register a special register?
+    pub fn is_special(&self) -> bool {
```
alexcrichton, line 152: Mind adding a small test that the registers above all `assert!(thing.is_special())`?

pulley/src/op.rs
```diff
@@ -0,0 +1,258 @@
+//! Pulley bytecode operations with their operands.
+
+use crate::imms::*;
+use crate::regs::*;
+#[allow(unused_imports)] // Some `cfg`s don't use this.
```
alexcrichton, line 5: One thing I might recommend is `#[allow(unused_imports)]` at the crate root with a `cfg_attr` for the "unusual cfgs" or something like that? Along the lines of "default builds have no warnings" but less-than-default builds can have warnings suppressed for ease of use.

pulley/src/imms.rs
```diff
@@ -0,0 +1,28 @@
...
+/// This is relative to the start of this offset's containing instruction.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct PcRelOffset(i32);
+
+#[cfg(feature = "arbitrary")]
+impl<'a> arbitrary::Arbitrary<'a> for PcRelOffset {
+    fn arbitrary(u: &mut arbitrary::Unstructured<'a>) -> arbitrary::Result<Self> {
+        Ok(Self(u.arbitrary()?))
+    }
+}
```
alexcrichton, line 14: This seems a bit dangerous, should this perhaps be omitted? Or maybe it always generates 0 for the time being?

(seems like the "real fix" is a higher-level generator that generates semantically correct offsets)

pulley/src/decode.rs
```diff
@@ -0,0 +1,675 @@
...
+    fn get1(&mut self) -> Result<u8, Self::Error>;
+
+    /// Read two bytes from the stream.
+    fn get2(&mut self) -> Result<[u8; 2], Self::Error>;
+
+    /// Read four bytes from the stream.
+    fn get4(&mut self) -> Result<[u8; 4], Self::Error>;
+
+    /// Read eight bytes from the stream.
+    fn get8(&mut self) -> Result<[u8; 8], Self::Error>;
```
alexcrichton, line 122: Reading over this again, WDYT about something like:

```rust
fn advance<const N: usize>(&mut self) -> Result<[u8; N], Self::Error>;
```

It feels a little weird below that in `SafeBytecodeStream` the `advance` method looks like it can cause panics but I think it probably never does in practice. By folding everything into one method though it might be reasonable to make this more obvious that the safe stream never panics?

pulley/fuzz/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "pulley-interpreter-fuzz"
```
alexcrichton, line 2: This won't get executed on oss-fuzz, how do you feel about that?

If you're ok with it I might recommend a wasm-tools-style approach where the fuzz crate here isn't actually a `cargo fuzz` crate but has a single entrypoint that takes fuzz input and then uses the input to dictate which fuzzer is internally run. Then we'd add a single fuzzer at the crate root for pulley which would internally run all of pulley's fuzzers.

pulley/src/lib.rs
```diff
@@ -0,0 +1,183 @@
+//! The pulley bytecode for fast interpreters.
```
alexcrichton, line 1: Mind adding `#![cfg_attr(docsrs, feature(doc_auto_cfg))]` to the crate here too?

pulley/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+[features]
+std = []
+arbitrary = ["dep:arbitrary", "arbitrary/derive", "std"]
+encode = []
+decode = []
+disas = ["decode"]
+interp = ["decode"]
+
+[package.metadata.docs.rs]
+features = ["encode", "decode", "disas", "interp"]
```
alexcrichton, line 30: Perhaps `all-features = true` instead?

alexcrichton (approved):
Looks great to me, thanks again for pushing on this! 

For some background I've pre-reviewed much of this with Nick ahead of time hence the light comments here as many of my otherwise-comments have already been addressed.

alexcrichton:
> Dedicated CI for Pulley

One thing I'd also recommend here is to run all tests for pulley in miri as well (or otherwise just add it to our list of tests-in-miri)

pulley/src/regs.rs
fitzgen: Tried that first, unfortunately it isn't allowed :(

pulley/tests/all/main.rs
fitzgen: Good eye

pulley/src/regs.rs
fitzgen: Done.

pulley/src/op.rs
fitzgen: Turns out this import isn't used at all anymore...

pulley/src/imms.rs
fitzgen: Changed this to always choose zero.

Either way, whoever is working with the arbitrary ops is going to have to clean them up somehow if they are interpreting them. In the existing fuzz target for the interpreter we simply filter out instructions that we can't safely interpret, such as branches.

pulley/fuzz/Cargo.toml
fitzgen: Alright, I merged the fuzz targets together into one, like `wasm-tools`.

fitzgen:
> One thing I'd also recommend here is to run all tests for pulley in miri as well (or otherwise just add it to our list of tests-in-miri)

Will do as a follow up, with the rest of pulley-specific CI.

PR #9014, opened 2024-07-25
Get Pulley tests passing under MIRI; run Pulley tests under MIRI in CI

fitzgen:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -970,7 +970,7 @@ jobs:
   miri:
     strategy:
       matrix:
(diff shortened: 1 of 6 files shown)


cfallin (approved):
LGTM; I don't pretend to be an expert in Rust's pointer provenance model, but miri certainly is, and if miri is happy then so am I!

fitzgen:
> I don't pretend to be an expert in Rust's pointer provenance model, but miri certainly is, and if miri is happy then so am I!

same!

thanks

PR #9018, opened 2024-07-26
Winch aarch64 cmp

vulc41n:
This PR implements wasm comparison instructions for winch targeting aarch64.
#8321

### DIFF
--- a/tests/disas/winch/aarch64/f32_eq/const.wat
+++ b/tests/disas/winch/aarch64/f32_eq/const.wat
@@ -1,0 +1,30 @@
+;;! target = "aarch64"
+;;! test = "winch"
+
+(module
+    (func (result i32)
+        (f32.const 1)
+        (f32.const 1)
+        (f32.eq)
+    )
+)
+
+;; wasm[0]::function[0]:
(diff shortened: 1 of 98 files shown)


saulecabrera:
I can help reviewing this change.

saulecabrera (commented):
Sorry it took me a bit to review this change. In general it looks great to me, thanks. I left one nit below which I think it makes sense to address before merging.

saulecabrera (approved):
Thanks!

vulc41n:
No problem, thanks for reviewing my pull request 🙂
I've removed `rd` from `subs` methods, but I do see `Wd` in the doc you linked. Am I missing something ?
Please tell me if there is anything else I can do to improve this

saulecabrera:
> No problem, thanks for reviewing my pull request 🙂 I've removed `rd` from `subs` methods, but I do see `Wd` in the doc you linked. Am I missing something ? Please tell me if there is anything else I can do to improve this

Yeah, according to the docs there's a proper destination register, however, it's not currently used in the way we lower comparison operations in Winch, since we rely on the flags or in `cmp_with_set`, so my suggestion is mostly to make it easier to reflect the usage.

PR #9020, opened 2024-07-26
Update gimli to 0.31.0

bjorn3:
Gimli 0.29.0 and later includes a fix for corrupt .eh_frame tables being generated in certain cases. Wasmtime isn't affected by this as it never sets an LSDA, but cg_clif will need this for supporting unwinding from exceptions.


bjorn3:
Gimli 0.31.0 is the latest version, but the addr2line version used by the latest backtrace version depends on gimli 0.29.0, so I used gimli 0.29.0 here to avoid duplicate dependencies.

alexcrichton (approved):
Thanks!

PR #9022, opened 2024-07-26
Add an unsafe API to unload process trap handlers

alexcrichton:
This commit adds a new unsafe API `Engine::unload_process_handlers` which can be used to de-initialize trap-handling state in a process. In general this is an unsafe operation that we have no means of making safe, hence the `unsafe`. There are particular situations where this can be done safely though and this is provided for such situations. The safety conditions rely on invariants that Wasmtime itself cannot maintain so the burden is on callers to ensure that this is invoked in a safe manner.

This is inspired by discussion [on Zulip][zulip] where Wasmtime's trap handling infrastructure prevented unloading a DLL with Wasmtime in it. There's possibly other reasons that unloading DLLs are unsafe in Rust but it feels appropriate to at least provide this knob to embedders to be able to turn if they like.

[zulip]: https://bytecodealliance.zulipchat.com/#narrow/stream/206238-general/topic/How.20to.20safely.20drop.20wasmtime.3F/near/453366905


crates/wasmtime/src/engine.rs
```diff
@@ -699,6 +699,55 @@ impl Engine {
...
+    ///
+    /// This method will panic if this `Engine` handle is not the last remaining
+    /// engine handle.
+    ///
+    /// # Aborts
+    ///
+    /// This method will abort the process on some platforms in some situations
+    /// where unloading the handler cannot be performed and an unrecoverable
+    /// state is reached. For example on Unix platforms with signal handling
+    /// the process will be aborted if the current signal handlers are now
```
lann, line 720:
```suggestion
    /// the process will be aborted if the current signal handlers are not
```

crates/wasmtime/src/engine.rs
```diff
@@ -699,6 +699,55 @@ impl Engine {
...
+    /// Wasmtime's.
+    ///
+    /// # Unsafety
+    ///
+    /// This method is not generally safe to call and has a number of
+    /// preconditions that must be met to even possibly be safe. Even with these
+    /// known preconditions met there may be other unknown invariants to uphold
+    /// as well.
+    ///
+    /// * The must be no other instances of `Engine` anywhere for this Wasmtime
```
lann, line 730:
```suggestion
    /// * There must be no other instances of `Engine` anywhere for this Wasmtime
```

crates/wasmtime/src/engine.rs
```diff
@@ -699,6 +699,55 @@ impl Engine {
...
+    ///   DLL elsewhere in the process. Note that it's not just clones of this
```
lann, line 731: I'm not sure exactly what "for this Wasmtime DLL" means; maybe just drop it?

crates/wasmtime/src/engine.rs
```diff
@@ -699,6 +699,55 @@ impl Engine {
...
+    ///   `Engine`, but any other `Engine` instances at all.
+    ///
+    /// * On Unix platforms no other signal handlers could have been installed
+    ///   after Wasmtime's own signal handlers were installed.
```
lann, line 735: "no other" or just "conflicting"? I guess you don't want to commit to specific handlers being installed but this is already pretty unstable... :shrug:

crates/wasmtime/src/engine.rs
```diff
@@ -699,6 +699,55 @@ impl Engine {
...
+    ///
+    /// * All existing threads which have used this DLL or copy of Wasmtime may
+    ///   no longer use this copy of Wasmtime. Per-thread state is not iterated
+    ///   and destroyed. Only future threads may use future instances of this
+    ///   Wasmtime itself.
```
lann, line 740: Would it make sense to explicitly poison this copy of wasmtime with a global or is that already effectively happening?

lann (commented):
Given how specialized this you can consider all the functional comments just questions.

crates/wasmtime/src/engine.rs
alexcrichton: I'm hesitant to go out of our way to do that as it might slow down other parts for this niche unsafe part, so I'd lean towards making this part of the unsafe contract rather than trying to add more guard rails.

alexcrichton:
Good suggestions/questions! I've tried to wordsmith and update a bit.

tests/unload-engine.rs
```diff
@@ -0,0 +1,28 @@
+//! A single-test executable which only tests `Engine::unload_process_handlers`
+//! is possible.
+//!
+//! It's not safe for this test to contain any other tests.
```
fitzgen, line 4:
```suggestion
//! It's not safe for this binary to contain any other tests.
```

fitzgen (approved):

PR #9027, opened 2024-07-26
Explorer with clif fix: cache does not emit clif files

vulc41n:
Hey 👋

I found a bug introduced in [PR#8972](https://github.com/bytecodealliance/wasmtime/pull/8972). When cache is used, no clif files can be emited. I propose to disable the cache when exploring with cranelift

@fitzgen


elliottt (approved):
Nice find!

PR #9030, opened 2024-07-27
Fix `fmin`/`fmax` cprop miscompilation and add `f16`/`f128` `fmin`/`fmax` cprop support

beetrees:
Currently, constant propagation for `f32`/`f64` `fmin`/`fmax` always outputs the first input when both inputs are zeros. However, the semantics of the `fmin`/`fmax` instructions require that `-0.0` be treated as less than `0.0`. This PR fixes this and adds test cases to ensure the issue does not reoccur. Const prop support for `f16`/`f128` `fmin`/`fmax` is also added.

`f16`/`f128` issue: #8312


fitzgen (approved):
Looks great, thanks!

PR #9036, opened 2024-07-29
Add a more first-class error for Wasmtime on 32-bit

alexcrichton:
This commit adds a more first-class error in the build script for the `wasmtime` crate to indicate that 32-bit platforms are not supported. While there is a `compile_error!` with somewhat similar information it is not too helpful because the compiler keeps compiling things after the `compile_error!` is reached meaning that the absence of a `sys` module cascades to many other compile errors as well. This presents a more readable error with links to github issues in a way that halts the build and should prevent unsupported crates from being built.

### DIFF
--- a/crates/wasmtime/build.rs
+++ b/crates/wasmtime/build.rs
@@ -1,6 +1,61 @@
 fn main() {
     println!("cargo:rerun-if-changed=build.rs");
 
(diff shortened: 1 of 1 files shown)


fitzgen (approved):
:+1:

alexcrichton:
Given the error in CI I've decided on a different approach in https://github.com/bytecodealliance/wasmtime/pull/9039

PR #9042, opened 2024-07-29
Fix a pulley crash on OSS-Fuzz

alexcrichton:
This fixes a crash where the `GetSp` opcode was overwriting a special register, so apply a similar filter as to other instructions to ensure that the special registers are not clobbered.


fitzgen (approved):
Thanks! Sorry I forgot about this.

Issue #6104, opened 2023-03-27
Support bitcasting between scalar and vector types


bjorn3:
#### Feature

See title. For example `v0 = vconst.i64x2 ...; bitcast.i128 little v0`.

#### Benefit

This allows bitcasting between the two without having to go through memory. Cg_clif needs this bitcasting for implementing `mem::transmute` and float to 128bit int casts on Windows. Currently it has to go through the stack, which is slower and inhibits optimizations.

afonso360:
This seems like a good starter issue. It currently affects all of our 4 backends.

Here's a test demonstrating the issue:
```
test compile
target aarch64

function %bitcast_vec_to_int(i64x2) -> i128 {
block0(v0: i64x2):
    v1 = bitcast.i128 little v0
    return v1
}

function %bitcast_int_to_vec(i128) -> i64x2 {
block0(v0: i128):
    v1 = bitcast.i64x2 little v0
    return v1
}
```

When running this test it fails with: `should be implemented in ISLE: inst = v1 = bitcast.i128 little v0, type = Some(types::I128)`

Happy to help anyone who wants to try working on this!

aidenfoxivey:
@afonso360 It seems like the only remaining part to fix is s390x, right?

afonso360:
Yes, I think that's correct

aidenfoxivey:
> Yes, I think that's correct

I'll take a look at how to get a VM running on my ARM MacBook.

aidenfoxivey:
Maybe I can take on that remaining part.

aidenfoxivey:
Signed up for one of those LinuxONE servers from IBM, but no response from them so far lol. I dunno whether maybe QEMU is the preferred approach?

afonso360:
QEMU is usually easier for working on different arches. We have a great [doc on how to set it up](https://github.com/bytecodealliance/wasmtime/blob/main/docs/contributing-cross-compiling.md) if you're running in Linux

aidenfoxivey:
> QEMU is usually easier for working on different arches. We have a great [doc on how to set it up](https://github.com/bytecodealliance/wasmtime/blob/main/docs/contributing-cross-compiling.md) if you're running in Linux

Maybe I'll try to run those in a VM. Unfortunately I'm a Mac person for the time being.

PR #9045, opened 2024-07-30
Add initial `f16` and `f128` support to the x64 backend

beetrees:
This PR adds initial support for passing `f16` and `f128` values around to the x64 backend. Support is added for the `load`, `store`, `bitcast`, `f16const` and `f128const` CLIF instructions.

As a side effect of this PR generalising x64 `bitcast` support, this PR fixes (for x64) #6104.

`f16`/`f128` issue: #8312


fitzgen (approved):
Looks great, thanks!

PR #1687, opened 2020-05-12
cranelift: Sign extend `Imm64` immediates

fitzgen:
When an instruction has an `Imm64` immediate, but operates on values of a
narrower width, we need to sign extend the value.

Fixes #1095

cc @bjorn3

PR #9046, opened 2024-07-30
Cranelift: Stop sign-extending `Imm64` immediates in builder methods

fitzgen:
Originally, we sign-extended `Imm64` immediates from their controlling type var's width to `i64` in the builder methods for `BinaryImm64` and a couple other instruction formats: https://github.com/bytecodealliance/wasmtime/pull/1687

At some point, we decided that the unused bits in the `Imm64` when the controlling type var's width is less than 64 should be zero. And we started asserting that in various places, for example:
https://github.com/bytecodealliance/wasmtime/blob/87817f38a128caa76eaa6a3c3c8ceac81a329a3e/cranelift/codegen/src/machinst/lower.rs#L1237-L1243

However, we never removed or updated the sign-extension code in the builder methods. This commit switches the builder methods over to masking the `Imm64` to just the valid bits, effectively doing a zero extend from the controlling type var's width instead of a sign extend.

To avoid too much churn in tests, I made `display_inst` print the sign-extended-from-the-controlling-type-variable's-width value of `Imm64` immediates. I also made the `Display` implementation for `Imm64` always print in decimal form with a `-` for negative numbers, rather than the hex it would previously print for large negative numbers, so that `iconst.i32 0x8000_0000` doesn't display as `iconst.i32 0xffff_ffff_8000_0000`, which is otherwise pretty confusing.

### DIFF
--- a/cranelift/codegen/meta/src/gen_inst.rs
+++ b/cranelift/codegen/meta/src/gen_inst.rs
@@ -973,7 +973,7 @@ fn gen_format_constructor(format: &InstructionFormat, fmt: &mut Formatter) {
         args.join(", ")
     );
 
-    let imms_need_sign_extension = format
+    let imms_need_masking = format
         .imm_fields
         .iter()
         .any(|f| f.kind.rust_type == "ir::immediates::Imm64");
@@ -986,7 +986,7 @@ fn gen_format_constructor(format: &InstructionFormat, fmt: &mut Formatter) {
         fmtln!(
             fmt,
             "let{} data = ir::InstructionData::{} {{",
-            if imms_need_sign_extension { " mut" } else { "" },
+            if imms_need_masking { " mut" } else { "" },
             format.name
         );
         fmt.indent(|fmt| {
@@ -995,7 +995,7 @@ fn gen_format_constructor(format: &InstructionFormat, fmt: &mut Formatter) {
         });
         fmtln!(fmt, "};");
 
-        if imms_need_sign_extension {
+        if imms_need_masking {
             fmtln!(fmt, "data.mask_immediates(ctrl_typevar);");
         }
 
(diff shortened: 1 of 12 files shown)


cranelift/codegen/meta/src/gen_inst.rs
```diff
@@ -996,7 +996,7 @@ fn gen_format_constructor(format: &InstructionFormat, fmt: &mut Formatter) {
         fmtln!(fmt, "};");
 
         if imms_need_sign_extension {
-            fmtln!(fmt, "data.sign_extend_immediates(ctrl_typevar);");
+            fmtln!(fmt, "data.mask_immediates(ctrl_typevar);");
```
cfallin, line 999: Can we rename the boolean on this if-block (`imms_need_sign_extension`) as well to something that fits the new scheme?

cranelift/codegen/src/ir/immediates.rs
```diff
@@ -74,10 +74,27 @@ impl Imm64 {
         self.0
     }
 
+    /// Mask this immediate to the given power-of-two bit width.
+    pub(crate) fn mask_to_width(&mut self, bit_width: u32) {
```
cfallin, line 78: To fit with the other operator methods (e.g. `wrapping_neg`), can we do functional style here (`fn mask_to_width(self, bit_width: u32) -> Self`)?

cfallin (approved):
Thanks for finding this inconsistency!

A few minor nits below; my only other thought is that the rules around how we print seem a little convoluted, though I understand the goal in trying to preserve existing test expectations. The asymmetry (small positive numbers, or any negative number, are decimal) somehow doesn't sit right with me. I think I'd almost rather take a one-time diff and pick something more consistent -- always hex? Always decimal? Always decimal but with hex in a comment (like we do for constants now e.g. `; v0 = 0`)?

Feel free to disregard my bikeshed color preferences; just my thoughts as long as we're here :-)

fitzgen:
> The asymmetry (small positive numbers, or any negative number, are decimal) somehow doesn't sit right with me. I think I'd almost rather take a one-time diff and pick something more consistent -- always hex? Always decimal? Always decimal but with hex in a comment (like we do for constants now e.g. `; v0 = 0`)?

Agreed. I'll look into this as a follow up, rather than doing it in this PR.

PR #9051, opened 2024-07-31
Winch aarch64 jmp

vulc41n:
Hey 👋

This PR implements jmp, jmp_table and branch instructions for winch targeting aarch64.

#8321


winch/codegen/src/isa/aarch64/asm.rs
```diff
@@ -474,15 +474,54 @@ impl Assembler {
...
+    /// Emits a jump table sequence.
+    pub fn jmp_table(
+        &mut self,
+        targets: &[MachLabel],
+        default: MachLabel,
+        index: Reg,
+        tmp1: Reg,
+        tmp2: Reg,
+    ) {
+        self.emit(Inst::JTSequence {
```
cfallin, line 502: We'll need an `EmitIsland` instruction here too; see [here](https://github.com/bytecodealliance/wasmtime/blob/ba864e987ef1ab87c439ca6b396264547d2425e1/cranelift/codegen/src/isa/aarch64/lower.isle#L2967) and [here](https://github.com/bytecodealliance/wasmtime/blob/ba864e987ef1ab87c439ca6b396264547d2425e1/cranelift/codegen/src/isa/aarch64/inst.isle#L933-L952) for more details on how and why.

cfallin (commented):
Thanks! I think we need to add a little logic to handle large branch offsets and islands here (aarch64-specific issue, x64 doesn't need this) but otherwise LGTM.

vulc41n:
Thanks for reviewing this work 🙂

Should I push the tests ? They are obviously large (13 and 12Mb).

Please tell me if there is anything else that I can improve

cfallin:
> Should I push the tests ? They are obviously large (13 and 12Mb).

We have tests of the underlying parts (`MachBuffer` and its island-insertion logic); that is probably fine for coverage here, vs. checking in a very large test, IMHO. If we wanted to do better in the future we could add the equivalent of Cranelift's chaos mode choice to sometimes artificially lower the island threshold, to avoid the need for very large inputs to trigger this; but no need to do that in this PR.

cfallin (approved):

Issue #6530, opened 2023-06-06
Cranelift: get tail calls working on s390x


fitzgen:
> I'm having a hard time figuring out the details of s390x here, so I am going to disable support for the `tail` calling convention on s390x (with a loud assertion) to land this PR. We can get s390x working in follow ups.

_Originally posted by @fitzgen in https://github.com/bytecodealliance/wasmtime/issues/6500#issuecomment-1579427510_

fitzgen:
Hey @uweigand, would really appreciate your help here. There are two things that need to be done (or at least, I have done it with these two different steps for the other backends):

1. **Get regular, non-tail calls working with the `tail` calling convention.**

   This has involved modifying the calling convention such that callees (rather than callers) pop stack arguments as part of their function epilogue and before/at the same time as returning to their caller.

   There is [a runtest that use the `tail` calling convention without actually doing tail calls](https://github.com/bytecodealliance/wasmtime/blob/main/cranelift/filetests/filetests/runtests/tail-call-conv.clif), it is currently disabled on s390x but would presumably be enabled during this step. I also recommend copying that test to `filetests/isa/x390x` and turning it into a `compile precise-output` test as well, which the other backends have done too.

2. **Implement tail calls.**

   These are only valid CLIF when both the caller and callee use the `tail` calling convention, so that is something we can rely upon.

   None of the other backends have been doing parallel move solving during tail calls to make sure new arguments end up in the correct locations without overwriting old arguments that are on the stack but become new arguments elsewhere on the stack in the tail call. Instead, they bump SP and construct a temporary frame and then move the frame down on top of the current frame and then initialize SP for the callee. I recommend starting with this simple approach, and we can later do parallel move solving as an optimization, but whatever works for you.

   There are a few runtests for exercising tail calls: `filetests/runtests/return-call{.clif,-indirect.clif,-loop.clif}`. Each backend also has a copy of `return-call.clif` (which actually covers both colocated and not calls) to their ISA-specific tests as `compile precise-output` tests, and I recommend that s390x has a copy as well.

[...]

PR #9052, opened 2024-07-31
s390x: Support tail calls

uweigand:
This adds support for a newly defined tail-call ABI for s390x as well as supporting tail calls themselves.

We now use the tail-call ABI by default for Wasmtime, and enable tail calls by default.

This also allows getting rid of a number of special case and test exclusions for s390x.

Fixes: https://github.com/bytecodealliance/wasmtime/issues/6530


cranelift/codegen/src/isa/s390x/inst/args.rs
```diff
@@ -88,6 +91,21 @@ impl MemArg {
...
+        match base {
+            &MemArg::RegOffset { reg, off, flags } => MemArg::RegOffset {
+                reg,
+                off: off + offset,
+                flags,
+            },
+            &MemArg::InitialSPOffset { off } => MemArg::InitialSPOffset { off: off + offset },
+            &MemArg::NominalSPOffset { off } => MemArg::NominalSPOffset { off: off + offset },
+            &MemArg::SlotOffset { off } => MemArg::SlotOffset { off: off + offset },
+            _ => unreachable!(),
```
fitzgen, line 105: What cases is this matching? I think it might be better to explicitly list them, rather than using `_`, so that if more `MemArg` variants are added in the future we don't forget to consider/update this case.

cranelift/codegen/src/isa/s390x/inst/emit.rs
```diff
@@ -1305,7 +1314,7 @@ fn put_with_trap(sink: &mut MachBuffer<Inst>, enc: &[u8], trap_code: TrapCode) {
 /// State carried between emissions of a sequence of instructions.
 #[derive(Default, Clone, Debug)]
 pub struct EmitState {
-    pub(crate) initial_sp_offset: i64,
+    pub(crate) nominal_sp_offset: u32,
```
fitzgen, line 1317: Can we add some documentation for what the nominal SP is and what its relation to the actual SP is? This was something that I always forgot the nitty-gritty details of in the old system, so if we are bringing part of it back, I think it would be helpful to have comments spelling those things out.

cranelift/codegen/src/machinst/abi.rs
```diff
@@ -1335,6 +1335,11 @@ impl<M: ABIMachineSpec> Callee<M> {
         Ok(())
     }
 
+    /// Get the return area pointer register.
+    pub fn ret_area_ptr(&self) -> Reg {
+        self.ret_area_ptr.unwrap()
+    }
```
fitzgen, line 1341: We don't always have a return area pointer, so I think this should return an `Option<Reg>` and let callers do the unwrap if they have knowledge of why it definitely exists in their specific case.

```suggestion
    /// Get the return area pointer register, if any.
    pub fn ret_area_ptr(&self) -> Option<Reg> {
        self.ret_area_ptr
    }
```

cranelift/codegen/src/prelude_lower.isle
```diff
@@ -1001,6 +1001,10 @@
 (decl abi_sized_stack_ret_space (Sig) i64)
 (extern constructor abi_sized_stack_ret_space abi_sized_stack_ret_space)
 
+;; Incoming return area pointer.
+(decl abi_ret_area_ptr () Reg)
+(extern constructor abi_ret_area_ptr abi_ret_area_ptr)
```
fitzgen, line 1006: And then this external constructor could be called `unwrap_abi_ret_area_ptr` or some such.

tests/all/defaults.rs
```diff
@@ -23,7 +23,7 @@ fn test_tail_call_default() -> Result<()> {
                 .target("riscv64")?,
         ),
         (
-            false,
+            true,
```
fitzgen, line 26: \o/

fitzgen (approved):
Nice!

I didn't evaluate the details of the calling convention, as s390x isn't my area of expertise, but everything looks great in general, and I was very pleased to see the updates to the runtests and fuzzgen. Couple nitpick-y suggestions below, but I think this is basically good to go.

Thanks, Ulrich!!

uweigand:
Thanks for the review, @fitzgen !   The latest version should address all review comments.

fitzgen (approved):
Thanks!

PR #9057, opened 2024-08-01
Validate import names given to `trappable_imports`

elliottt:
It's easy to accidentally use the generated name when passing names to the `trappable_imports` field of the `bindgen!` macro, yielding generated bindings that don't use the expected `trappable_error_type`. This PR adds some validation, ensuring that arguments to `trappable_imports` are used during binding generation and raising an error if any are ignored.

It doesn't appear that we have negative tests for wasmtime-component-macro, but I'm happy to add them if I've missed something there.

### DIFF
--- a/crates/wit-bindgen/src/lib.rs
+++ b/crates/wit-bindgen/src/lib.rs
@@ -73,8 +73,7 @@ struct Wasmtime {
     // Track the with options that were used. Remapped interfaces provided via `with`
     // are required to be used.
     used_with_opts: HashSet<String>,
-    // Track the with options that were used. Remapped interfaces provided via `with`
-    // are required to be used.
+    // Track the imports that matched the `trappable_imports` spec.
     used_trappable_imports_opts: HashSet<String>,
 }
 
--- a/crates/wit-bindgen/src/lib.rs
+++ b/crates/wit-bindgen/src/lib.rs
@@ -73,6 +73,9 @@ struct Wasmtime {
     // Track the with options that were used. Remapped interfaces provided via `with`
     // are required to be used.
     used_with_opts: HashSet<String>,
+    // Track the with options that were used. Remapped interfaces provided via `with`
+    // are required to be used.
+    used_trappable_imports_opts: HashSet<String>,
 }
 
 struct ImportFunction {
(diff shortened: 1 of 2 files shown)


alexcrichton:
I realize it'd be more difficult, but would you be up for seeing how to validate that all the entries in this list are used in addition to them being valid? Or rather validation wouldn't be required if they're validated to be used. IIRC we already require that all `with` entries are used so it might be possible to reuse logic there perhaps?

I ask this because it feels like the most robust solution here in terms of ensuring we actually use all of the input and helping to catch typos/misconfiguration.

Also as for negative tests you're right in that we don't have any. They're unfortunately difficult to add because rustc's output changes over time, but we should probably look in to adding them at some point in the future.

elliottt:
> I realize it'd be more difficult, but would you be up for seeing how to validate that all the entries in this list are used in addition to them being valid? Or rather validation wouldn't be required if they're validated to be used. IIRC we already require that all `with` entries are used so it might be possible to reuse logic there perhaps?

Yep, I'll have a look at doing that instead :+1:

elliottt:
I've reworked it to track the import names that get used, and check that all names specified in the list initially get used. If none are used, a similar error to the handling of the `with` clause will be reported.

One wrinkle here is that I had to mark names as used if `can_trap` returned true, rather than only if it was a name that had a trappable return type. With the more strict interpretation the tests fail, as we test cases where names mentioned in the list don't have any return type. I'm not sure that requiring names in `trappable_imports` have a trappable error is useful, given that names can overlap between different interfaces.

crates/wit-bindgen/src/lib.rs
```diff
@@ -73,6 +73,9 @@ struct Wasmtime {
     // Track the with options that were used. Remapped interfaces provided via `with`
     // are required to be used.
     used_with_opts: HashSet<String>,
+    // Track the with options that were used. Remapped interfaces provided via `with`
+    // are required to be used.
```
alexcrichton, line 77: Mind updating these docs?

crates/wit-bindgen/src/lib.rs
```diff
@@ -2412,29 +2427,35 @@ impl<'a> InterfaceGenerator<'a> {
...
-                None => format!("Host"),
-            };
-            let convert = format!("{}::convert_{}", convert_trait, err_name.to_snake_case());
-            uwrite!(
-                self.src,
-                "Ok((match r {{
+        } else {
+            self.gen
+                .used_trappable_imports_opts
+                .insert(func.name.clone());
```
alexcrichton, line 2433: Could this be added to `special_case_trappable_error`? (perhaps by updating that to take `&Function`)?

alexcrichton (commented):
Looks reasonable to me!

alexcrichton (approved):

PR #9059, opened 2024-08-01
chore(adapters): add README

vados-cosmonic:
This commit adds a README to the
`wasi-preview1-component-adapter-provider` crate to make it clearer how to use it and what it's for.

While the README is mostly for those who are already familiar, a section giving a brief overview of context around the adapters was added.

---

This would be nice to have to explain the [existing crate](https://crates.io/crates/wasi-preview1-component-adapter-provider)


alexcrichton (approved):
Thanks for this!

Issue #8096, opened 2024-03-12
winch: Gracefully handle compilation errors


saulecabrera:
The default behaviour in Winch is to panic when an invariant is broken at compile time. This behavior is not friendly for all embedders, as it might be painful to recover from an erroneous compilation. Instead we could try to gracefully handle compilation errors.

PR #9061, opened 2024-08-01
Simplify ABI selection and tail call integration

alexcrichton:
This commit simplifies the selection of an ABI for wasm functions now that all Cranelift backends implement tail calls. All wasm functions use the `Tail` calling convention except when the `winch_callable` tunable is enabled meaning that Winch-generated functions are being called.

This then additionally simplifies the activation of the tail call proposal. It's not unconditionally active and the same across all compilers. The Winch compiler is updated to return an error for unsupported instructions rather than panicking so the embedder API is suitable for feeding unsupported modules to Winch. This means that tail calls in Winch behaves similarly to GC in Cranelift or other unsupported proposals like SIMD in Winch.

### DIFF
--- a/winch/codegen/src/visitor.rs
+++ b/winch/codegen/src/visitor.rs
@@ -36,7 +36,8 @@ macro_rules! def_unsupported {
 
                 fn $visit(&mut self $($(,$arg: $argty)*)?) -> Self::Output {
                     $($(let _ = $arg;)*)?
-                        todo!(stringify!($op))
+
+                    self.found_unsupported_instruction = Some(stringify!($op));
                 }
             );
         )*
--- a/crates/cranelift/src/lib.rs
+++ b/crates/cranelift/src/lib.rs
@@ -158,47 +158,26 @@ fn wasm_call_signature(
     wasm_func_ty: &WasmFuncType,
     tunables: &Tunables,
 ) -> ir::Signature {
-    // NB: this calling convention in the near future is expected to be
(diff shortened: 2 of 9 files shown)


winch/codegen/src/visitor.rs
```diff
@@ -36,7 +36,8 @@ macro_rules! def_unsupported {
 
                 fn $visit(&mut self $($(,$arg: $argty)*)?) -> Self::Output {
                     $($(let _ = $arg;)*)?
-                        todo!(stringify!($op))
+
+                    self.found_unsupported_instruction = Some(stringify!($op));
```
fitzgen, line 40: Seems like, longer term, it would be nice if the visitor's `Output` was a `Result`. Not something we need to address now though, unless it is a pretty easy change.

fitzgen (approved):

winch/codegen/src/visitor.rs
alexcrichton: yeah I'm a bit wary to refactor most of the visitor to return `Ok(())` at the end so I'll stick to this for now

saulecabrera: I got to this comment while looking at Alex's comment in https://github.com/bytecodealliance/wasmtime/pull/9158/files#r1723917083 just wanted to say that this is on my radar, I had opened [this issue](https://github.com/bytecodealliance/wasmtime/issues/8096) a while ago, in order to refactor the internals to return `Result` instead of panicking, where applicable

PR #9065, opened 2024-08-02
Enforce `uninlined_format_args` for the workspace

### LINKED ISSUE #9064: add `uninlined_format_args`
Oh nice! If you'd like feel free to add this lint [here](https://github.com/bytecodealliance/wasmtime/blob/ba864e987ef1ab87c439ca6b396264547d2425e1/Cargo.toml#L174-L182) which would be a ratchet for the whole workspace. Probably not all of `pedantic` but this specific lint I think would be a reasonable candidate to add.

_Originally posted by @alexcrichton in https://github.com/bytecodealliance/wasmtime/issues/9047#issuecomment-2263425790_

hamirmahal:
closes #9064.

### DIFF
--- a/tests/all/globals.rs
+++ b/tests/all/globals.rs
@@ -145,7 +145,7 @@ fn i31ref_global_get() -> Result<()> {
             let global = Global::new(
                 &mut store,
                 GlobalType::new(ValType::I31REF, mutability),
-                val.clone(),
+                val,
             )?;
 
             let got = global.get(&mut store);
@@ -187,7 +187,7 @@ fn i31ref_global_set() -> Result<()> {
             )?;
 
             let new_val = Val::from(new_val);
-            global.set(&mut store, new_val.clone())?;
+            global.set(&mut store, new_val)?;
             let got = global.get(&mut store);
 
             let new_val = new_val
@@ -269,7 +269,7 @@ fn i31ref_as_anyref_global_get() -> Result<()> {
             let global = Global::new(
                 &mut store,
                 GlobalType::new(ValType::ANYREF, mutability),
-                val.clone(),
+                val,
             )?;
 
             let got = global.get(&mut store);
@@ -311,7 +311,7 @@ fn i31ref_as_anyref_global_set() -> Result<()> {
             )?;
 
             let new_val = Val::from(new_val);
-            global.set(&mut store, new_val.clone())?;
+            global.set(&mut store, new_val)?;
             let got = global.get(&mut store);
 
             let new_val = new_val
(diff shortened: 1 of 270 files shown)


hamirmahal (commented):
I can fix [the `/home/runner/work/wasmtime/wasmtime/target/debug/build/cranelift-codegen-d924ecd4c940ea8b/out/inst_builder.rs` failures](https://github.com/bytecodealliance/wasmtime/actions/runs/10220802939/job/28281941147?pr=9065#step:15:365) locally, but they're not in source control.

elliottt:
> I can fix [the `/home/runner/work/wasmtime/wasmtime/target/debug/build/cranelift-codegen-d924ecd4c940ea8b/out/inst_builder.rs` failures](https://github.com/bytecodealliance/wasmtime/actions/runs/10220802939/job/28281941147?pr=9065#step:15:365) locally, but they're not in source control.

Those are generated during the build, so we should figure out a different approach for fixing the errors. Perhaps we can emit an allow annotation when generating them?

alexcrichton:
Similar `#[allow]` annotations look [like this](https://github.com/bytecodealliance/wasmtime/blob/dbc503f3fe1864e89f4f4a4e5739b3441fb4cbc0/cranelift/codegen/src/opts/generated_code.rs#L9) and there might be some other locations as well.

tests/all/globals.rs
```diff
@@ -145,7 +145,7 @@ fn i31ref_global_get() -> Result<()> {
             let global = Global::new(
                 &mut store,
                 GlobalType::new(ValType::I31REF, mutability),
-                val.clone(),
```
alexcrichton, line 148: Were these clones being removed part of a separate clippy lint that was firing? Perhaps another good candidate to enable at the workspace level? (find to do in a separate PR)

alexcrichton (approved):
Thank you again for your work on this!

hamirmahal:
You're welcome!

tests/all/globals.rs
hamirmahal: I believe so, yes. I might be able to get to that this weekend.

hamirmahal: I took a look, and it seems like `Cargo.toml` already warns about `clone_on_copy`.

```
[workspace.lints.clippy]
# The default set of lints in Clippy is viewed as "too noisy" right now so
# they're all turned off by default. Selective lints are then enabled below as
# necessary.
all = { level = 'allow', priority = -1 }
clone_on_copy = 'warn'
map_clone = 'warn'
uninlined_format_args = 'warn'
unnecessary_to_owned = 'warn'
manual_strip = 'warn'
```

PR #9071, opened 2024-08-02
`cranelift-frontend`: Fix stack maps and liveness for loops

fitzgen:
Previously, we were not properly handling back edges. This manifested in values incorrectly being considered not-live inside loop bodies where they definitely were live. Consider the following example:

    block0:
      v0 = needs stack map

    block1:
      call foo(v0)
      call foo(v0)
      jump block1

We were previously considering `v0` live only for the first `call foo(v0)` but not the second, because we mistakenly concluded that `v0` would not be used again after that second `call`. While it won't be used again in *this* iteration of the loop, it will be used again in the *next* iteration of the loop.

Trevor and I initially tried implementing a clever trick suggested by Chris where, if we know the minimum post-order index of all of a block's transitive predecessors, we can continue to compute liveness in a single pass over the IR. We believed we could compute the minimum predecessor post-order index via dynamic programming. It turns out, however, that approach doesn't provide correct answers out of the box for certain kinds of irreducible control flow, only nearly correct answers, and would require an additional clever fix-up pass afterwards. We deemed this cleverness on cleverness unacceptable.

Instead, Trevor and I opted to implement a worklist algorithm where we process blocks to a fixed-point. This has the advantages of being obviously correct and producing more-precise results. It has the disadvantage of requiring multiple passes over the IR in the face of loops and back edges. Because this analysis is only used when needs-stack-map values are present (i.e. when the function uses GC values) and is otherwise skipped, this additional compile-time overhead is tolerable.

### DIFF
--- a/cranelift/frontend/src/frontend/safepoints.rs
+++ b/cranelift/frontend/src/frontend/safepoints.rs
@@ -99,6 +99,9 @@ impl<T> SlotSizeMap<T> {
 }
 
 /// A set of live values.
+///
+/// Make sure to copy to a vec and sort, or something, before iterating over the
+/// values to ensure deterministic output.
 type LiveSet = HashSet<ir::Value>;
 
 /// A worklist of blocks' post-order indices that we need to process.
@@ -107,14 +110,16 @@ struct Worklist {
     /// Stack of blocks to process.
     stack: Vec<u32>,
 
-    /// The set of blocks that need to be updated. This is a subset of the
-    /// elements present in `self.stack` because `self.stack` is allowed to have
-    /// duplicates, and once we pop the first occurrence of a duplicate, we
-    /// remove it from this set, since it no longer needs updates at that
-    /// point. This potentially uses more stack space than necessary, but
-    /// prefers processing immediate predecessors, and therefore inner loop
-    /// bodies before continuing to process outer loop bodies. This ultimately
-    /// results in fewer iterations required to reach a fixed point.
+    /// The set of blocks that need to be updated.
+    ///
+    /// This is a subset of the elements present in `self.stack`, *not* the
+    /// exact same elements. `self.stack` is allowed to have duplicates, and
+    /// once we pop the first occurrence of a duplicate, we remove it from this
+    /// set, since it no longer needs updates at that point. This potentially
+    /// uses more stack space than necessary, but prefers processing immediate
+    /// predecessors, and therefore inner loop bodies before continuing to
+    /// process outer loop bodies. This ultimately results in fewer iterations
+    /// required to reach a fixed point.
     need_updates: HashSet<u32>,
 }
 
@@ -149,7 +154,10 @@ impl Worklist {
     }
 
     fn push(&mut self, block_index: u32) {
-        // Mark this block as needing an update, since we have (re?) pushed it.
+        // Mark this block as needing an update. If it wasn't in `self.stack`,
+        // now it is and it needs an update. If it was already in `self.stack`,
+        // then pushing this copy logically hoists it to the top of the
+        // stack. See the above note about processing inner-most loops first.
         self.need_updates.insert(block_index);
         self.stack.push(block_index);
     }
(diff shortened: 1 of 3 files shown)


cranelift/frontend/src/frontend/safepoints.rs
```diff
@@ -82,350 +98,686 @@ impl<T> SlotSizeMap<T> {
...
-    /// without any fixed-point loop. The result of this analysis is the mapping
-    /// from each needs-stack-map value that is live across a safepoint to its
-    /// associated stack slot.
-    ///
-    /// Finally, we spill each of the needs-stack-map values that are live
-    /// across a safepoint to their associated stack slot upon definition, and
-    /// insert reloads from that stack slot at each use of the value.
-    pub(super) fn insert_safepoint_spills(&mut self) {
+/// A set of live values.
+type LiveSet = HashSet<ir::Value>;
```
cfallin, line 102: Do we ever iterate over the live-set and produce code based on its order? It'd be good to confirm (with a comment noting here) that we don't, in order to retain deterministic compiler output.

cranelift/frontend/src/frontend/safepoints.rs
```diff
@@ -82,350 +98,686 @@ impl<T> SlotSizeMap<T> {
...
+/// We compute two live sets for each block:
+///
+/// 1. The live-in set, which is the set of values that are live when control
+///    enters the block.
+///
+/// 2. The live-out set, which is the set of values that are live when control
+///    exits the block.
+///
+/// A block's live-out set is the union of its successors' live-in sets
+/// successors. A block's live-in set is the set of values that are still live
```
cfallin, line 187: extra word (successors) at end of sentence?

cranelift/frontend/src/frontend/safepoints.rs
```diff
@@ -82,350 +98,686 @@ impl<T> SlotSizeMap<T> {
...
+    fn push(&mut self, block_index: u32) {
+        // Mark this block as needing an update, since we have (re?) pushed it.
+        self.need_updates.insert(block_index);
+        self.stack.push(block_index);
+    }
+
+    fn pop(&mut self) -> Option<u32> {
+        while let Some(block_index) = self.stack.pop() {
+            // If this block was pushed multiple times, we only need to update
+            // it once, so remove it from the need-updates set.
```
cfallin, line 160: This is pretty clever and it took me a second to understand why we don't simply avoid pushing duplicates; it's to do with the innermost-first heuristic mentioned above (so we want to process at the *last* duplicate pushed, not the first), is that right?

It might be worth a comment here explicitly noting that; we push multiple times because in effect we're "hoisting" the point at which we consider a block sooner into the future, and it would be too expensive to actually remove it from deeper in the stack.

(I usually write worklist algorithms with "if already in set, don't add to queue" logic and I've seen it that way more commonly so it's worth noting this trick here!)

cfallin (approved):
Looks great, thanks! Some minor requests for clarification in comments but otherwise good to go.

cranelift/frontend/src/frontend/safepoints.rs
fitzgen: The one place we do iterate over them, it is only after we've copied them into a smallvec and sorted it. But I added a note to the doc comment as well.

cranelift/frontend/src/frontend/safepoints.rs
fitzgen: > This is pretty clever and it took me a second to understand why we don't simply avoid pushing duplicates; it's to do with the innermost-first heuristic mentioned above (so we want to process at the _last_ duplicate pushed, not the first), is that right?

Correct, filtering on `push` means we process the least-recently-pushed duplicate entry. Filtering on `pop` means we process the most-recently-pushed duplicate entry, which gives us that inner-loops-first behavior that is desirable for minimizing the number of iterations until we reach the fixed-point.

Added some more comments; like your "hoisting" wording.

Issue #6055, opened 2023-03-17
Cranelift: Consider taking `trapnz` and `trapz` all the way to the backends rather than legalizing them


fitzgen:
Right now we legalize them into a conditional jump to a cold block with a trap intruction early on in the pipeline.

But every backend already has some variant of `MInst.TrapIf`, which is the same as `trapnz`, so it would be easy to support in lowering if we wanted to move it there.

Benefits:

* Not introducing a bunch of extra blocks makes compilation faster for everyone.

* It is also the only legalization that currently introduces new blocks, which is something ISLE can't do right now, and therefore is preventing us from porting legalization to ISLE.

* Finally it makes it so we could mark `trapnz` as having idempotent side effects which would allow us to GVN them. This would further improve bounds checks on multiple loads of the same dynamic index with differing static offsets for the non-Spectre case (e.g. what is talked about as next steps in https://github.com/bytecodealliance/wasmtime/pull/6031)

fitzgen:
Fixed in #0972

Issue #4781, opened 2022-08-25
cranelift: Allow testing trap codes in CLIF run tests


afonso360:
👋 Hey,

#### Feature

This was brought up by @dheaton-arm on [zulip](https://bytecodealliance.zulipchat.com/#narrow/stream/217117-cranelift/topic/testing.20traps/near/294881598). We should add a trap comparison operator to our CLIF test suite.

#### Benefit

As @cfallin [mentions in the thread](https://bytecodealliance.zulipchat.com/#narrow/stream/217117-cranelift/topic/testing.20traps/near/294893024) we already have reasonable coverage for traps via the wasm test suite.

However, this is still useful to improve the CLIF interpreter, to ensure that we are consistent in emitting the appropriate traps at the right time.

This could further be used in the clif differential fuzzer. The fuzzer currently stops executing any input, as soon as the interpreter traps.
This is overly restrictive since we could allow a `DivByZero` to execute in the backend and ensure that it comes up with the same trap. Other traps are also useful to check, but not all! We don't want to let a stack overflow keep executing since that can overwrite some memory and do terrible things.

I think this is also important if we want to start using the interpreter for const code execution, since we do need to ensure that we produce the same traps, and enabling it in the fuzzer is the best way I can think about it currently.

#### Implementation

We can add a trap handler either directly in `cranelift-jit` or the trampolines in the runtest suite. I'm not too familiar with how traps work in general to be able to expand on this.

#### Alternatives

So far I've been writing tests that trap in the interpreter test suite, and that is a viable option. But its not very scalable.

A very out of the box option would be to run wasmtime with the clif interpreter and use the wasm test suite and fuzzers to do this! I would love to do this, but right now its a bit too out there.

PR #9072, opened 2024-08-02
Lower conditional traps in backend

amartosch:
This PR moves conditional traps to backends from legalization, as described in  #6055.  Hope this is what was meant in the issue and is still needed. There are two things I'm not sure about:
* Sequences for s390x. I'm not familiar with the ISA, but conditional trap generation has already been implemented in the backend, so I've just blessed the tests.
* On RISCV traps are emitted inline; not trapping means jumping over it. On aarch64 and x86_64 they are emitted in islands. I wonder, if there is a deeper reason for this. If not, should the approach be unified with the other backends?

This is my first contribution, but I didn't ask any questions because this seemed to be straightforward. Please, feel free to simply discard it if the assumption was wrong.


fitzgen:
Excited to see this come in! Will take a closer look on Monday, thanks!

cranelift/codegen/src/isa/aarch64/lower.isle
```diff
@@ -1955,6 +1955,16 @@
...
 
+;;;;;  Rules for `trapz`;;;;;;;;;
+
+(rule (lower (trapz val trap_code))
+  (trap_if_zero $true val trap_code))
+
+;;;;;  Rules for `trapnz`;;;;;;;;;
+
+(rule (lower (trapnz val trap_code))
+  (trap_if_zero $false val trap_code))
```
fitzgen, line 1966: Can we rename `trap_if_zero` to something like `trap_if_reg`? Having `(trap_if_zero $false ...)` mean that we are trapping if something is *not* zero is a bit too easy to mix up. Even better would be to define a custom enum for zero and not zero because boolean arguments to functions almost always make things harder to read than otherwise.

cranelift/codegen/src/isa/x64/lower.isle
```diff
@@ -1812,6 +1812,15 @@
...
       (side_effect (x64_ud2 code)))
 
+;;;; Rules for `trapz` ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+(rule (lower (trapz val code))
+      (side_effect (trap_if_zero $true val code)))
+
+;;;; Rules for `trapnz` ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(rule (lower (trapnz val code))
+      (side_effect (trap_if_zero $false val code)))
```
fitzgen, line 1822: Ditto here.

fitzgen (commented):
Thanks! This is indeed exactly the kind of thing we were imagining in #6055. Overall the PR looks good, just a few nitpicks inline below.

> Sequences for s390x. I'm not familiar with the ISA, but conditional trap generation has already been implemented in the backend, so I've just blessed the tests.

This is generally fine. It would be good to double check whether we have run tests and fuzzgen support for conditional traps and add them if they are missing. Then we would be exercising actual execution behavior, which is going to give us a lot more confidence in the correctness of the implementation than blessing output will.

I don't see any `trap[n]z.clif` file in https://github.com/bytecodealliance/wasmtime/tree/main/cranelift/filetests/filetests/runtests nor does a `grep` show anything, so it would definitely be good to add as part of this PR.

I think CLIF interpreter and fuzzgen support could happen in follow ups, though.

> On RISCV traps are emitted inline; not trapping means jumping over it. On aarch64 and x86_64 they are emitted in islands. I wonder, if there is a deeper reason for this. If not, should the approach be unified with the other backends?

I believe that the answer is that we can implement them inline in risc-v without bloating code size (compared to branching out of line; s390x is inline as well, fwiw) but not on the other architectures. So, for those other architectures, we push the actual trapping instruction out of line to improve icache usage (since traps are extremely rare and effectively terminate the program).

afonso360:
> This is generally fine. It would be good to double check whether we have run tests and fuzzgen support for conditional traps and add them if they are missing. Then we would be exercising actual execution behavior, which is going to give us a lot more confidence in the correctness of the implementation than blessing output will.

We don't have great support for this in runtests (see https://github.com/bytecodealliance/wasmtime/issues/4781). So we might only be able to test the non trapping path.

cranelift/codegen/src/isa/x64/lower.isle
amartosch: Used enum

cranelift/codegen/src/isa/aarch64/lower.isle
amartosch: Renamed and used an enum

amartosch:
Also added very basic runtests that only check non-trapping path for now.

fitzgen:
Strange, for some reason CI hung and this never merged.

cranelift/codegen/src/context.rs
```diff
@@ -298,7 +298,7 @@ impl Context {
         self.loop_analysis.clear();
 
         // Run some specific legalizations only.
-        simple_legalize(&mut self.func, &mut self.cfg, isa);
+        simple_legalize(&mut self.func, isa);
```
fitzgen, line 301:
```suggestion
        simple_legalize(&mut self.func, isa);
```

cranelift/codegen/src/isa/aarch64/inst.isle
```diff
@@ -2907,6 +2907,35 @@
         (ConsumesFlags.ConsumesFlagsSideEffect
          (MInst.TrapIf (cond_br_cond cond) trap_code)))))
 
+;; Helpers for lowering `trapz`/`trapnz`
```
fitzgen, line 2910:
```suggestion
;; Helpers for lowering `trapz` and `trapnz`.
```

fitzgen:
Nice, looks like adding a commit triggered CI.

fitzgen:
Sorry for not catching this sooner!

alexcrichton (approved):

PR #9073, opened 2024-08-03
ci: check if we need to run cargo vet prune

iawia002:
In the context of #9062, I removed a dependency but forgot to remove the corresponding configuration in the `supply-chain`. However, the CI did not alert me about this. It is confusing to see a crate configuration in the `supply-chain` that we no longer depend on. Therefore, I added a step in the CI to check whether we need to execute the prune command.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -143,6 +143,11 @@ jobs:
         submodules: true
     - uses: ./.github/actions/install-rust
     - uses: ./.github/actions/install-cargo-vet
+    - name: Check if we need to prune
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! I don't think we'll want to do this though because one workflow for adding vets needed for contributors is for maintainers to send a PR which contains the vet entries which are then used by a contributor's later PR. That means that vet entries might be added without anything in-tree using them but they're required for a soon-to-land PR. Another option is to push commits to contributors' PRs but that's not currently required at this time.

PR #6219, opened 2023-04-16
ISLE: rewrite operation on `select` to `select` on operation

Kmeakin:
Rewrites operations on `select` to `select` on operations (eg `ineg(select(cond, x, y)) => select(cond, ineg(x), ineg(y))`. This transform is not obviously beneficial by itself, but it may expose more opportunities for other rewrite rules to trigger.

I'm not sure I put the extractor/constructor definitions for `unop`/`binop`/`ternop` in the correct places. Let me know if I should move them somewhere else.

PR #9074, opened 2024-08-03
egraphs: transpose select on operation to operation on select when it increases sharing

Kmeakin:
This is a second attempt of https://github.com/bytecodealliance/wasmtime/pull/6219, but this time only includes rewrites that are obviously beneficial by reducing number of instructions


abrown:
cc: @elliottt?

PR #9075, opened 2024-08-03
pulley: sign-extend `xconst` immediates

Kmeakin:
Change `xconst8`, `xconst16`, `xconst32` and `xconst64` instructions to sign-extend rather than zero-extend their immediates.

### DIFF
--- a/pulley/README.md
+++ b/pulley/README.md
@@ -47,26 +47,24 @@ to change, instructions to appear and disappear, and APIs to be overhauled.
 Here is the disassembly of `f(a, b) = a + b` in Pulley today:
 
 ```
-       0: 11 1f f0 ff ff ff ff ff ff ff   xconst64 x31, 18446744073709551600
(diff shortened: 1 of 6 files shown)


fitzgen (commented):
Would you be up for switching the `xconst*` instructions to sign extending instead of zero extending? I think it probably doesn't makes sense to burn the opcode space on both sign- and zero-extending constant instructions, and sign extending is a little more flexible in general, so I think it would be better to only have sign-extending constant instructions.

Kmeakin:
> Would you be up for switching the `xconst*` instructions to sign extending instead of zero extending? I think it probably doesn't makes sense to burn the opcode space on both sign- and zero-extending constant instructions, and sign extending is a little more flexible in general, so I think it would be better to only have sign-extending constant instructions.

Sure

fitzgen (approved):
Perfect, thanks!

PR #9079, opened 2024-08-05
riscv64: Stricter ISA flag checking when lowering instructions

afonso360:
👋 Hey,

The RISC-V ISA has a lot of extensions to support a lot of little things. One example of this, is that it does not support F32 types with the base ISA. To support that we need the F ISA extension.

The RISC-V backend mostly ignored these requirements since both F and D are part of the minimum ISA profile that we request (IMAFD). With this PR we now try to check for those flags when lowering F32 / F64 instructions. (This is fixed in the first commit)

Additionally on the vector side, we only ever checked the vector size, but not the supported types. This means that currently we can lower a F16x8 vector, but we never check that that type is supported by the target CPU. (This is fixed in the second commit)

This is a preparation commit for FP16 support, where we mostly just have to check for a ISA flag to support it. With this change we can now do that in a single place.

### DIFF
--- a/cranelift/codegen/src/isa/riscv64/inst.isle
+++ b/cranelift/codegen/src/isa/riscv64/inst.isle
@@ -1055,6 +1055,18 @@
 (extern constructor has_zicond has_zicond)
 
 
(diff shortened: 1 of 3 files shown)


cfallin (approved):
LGTM, thanks!

PR #9080, opened 2024-08-06
test-programs: reorganize wasi content generated by wit and remove unnecessary deps

iawia002:
A better version of #8939, since wit-bindgen now supports [parse WIT files from multiple paths](https://github.com/bytecodealliance/wit-bindgen/pull/1003).


pchickey (approved):
Thanks!

PR #9085, opened 2024-08-07
Pulley: Add memory access instructions with 64-bit offsets

fitzgen:
I had trimmed these instructions from the original upstreaming of the Pulley interpreter because I had mistakenly believed that they were unused. Turns out they are needed for Cranelift's Pulley backend to allow for lowering certain address modes to a single instruction. The alternative, lowering the address modes to a sequence of instructions, would be a bit annoying and these instructions seem generally useful.


pulley/src/interp.rs
```diff
@@ -951,6 +951,30 @@ impl OpVisitor for InterpreterVisitor<'_> {
...
+        let val = unsafe {
+            self.state
+                .x(ptr)
+                .get_ptr::<u32>()
+                .byte_offset(offset as isize)
+                .read_unaligned()
+        };
+        self.state.x_mut(dst).set_u64(u64::from(val));
+        Continuation::Continue
+    }
```
alexcrichton, line 964: Would it perhaps make sense to add a generlized helper `load32_u_offset` which takes `offset: usize` which no-offset/offset8/offset64 all call? (and would presumably get inlined and not actually matter at runtime)

alexcrichton (approved):

pulley/src/interp.rs
fitzgen: Mayyybe. Right now some of the `offset`-to-`isize` conversions are `.into()` and some are `as isize` which is a small hiccup, but probably inconsequential. All said, I think I'd prefer to just leave it as-is for now and see what the future brings.

PR #9088, opened 2024-08-07
pulley: pack `dst`, `src1` and `src2` registers into 2 bytes

Kmeakin:
Pack the `dst`, `src1` and `src2` register operands of binary operators into 2 bytes. Solves the `TODO` at the top of `lib.rs`.

~This required moving the special registers into their own register class, and adding `smov` and `spush`/`spop` to interact with SP/FP.~
I held off on splitting special registers into their own `SReg` class because I'm not sure how to integrate it with ISLE/Regalloc

### DIFF
--- a/pulley/README.md
+++ b/pulley/README.md
@@ -48,17 +48,17 @@ Here is the disassembly of `f(a, b) = a + b` in Pulley today:
 
 ```
        0: 0e 1a f0                        xconst8 x26, -16
-       3: 12 1b 1b 1a                     xadd32 sp, sp, x26
-       7: 2c 1b 08 1c                     store64_offset8 sp, 8, lr
-       b: 2a 1b 1d                        store64 sp, fp
-       e: 0b 1d 1b                        xmov fp, sp
-      11: 12 00 00 01                     xadd32 x0, x0, x1
-      15: 0b 1b 1d                        xmov sp, fp
-      18: 25 1c 1b 08                     load64_offset8 lr, sp, 8
-      1c: 22 1d 1b                        load64 fp, sp
-      1f: 0e 1a 10                        xconst8 x26, 16
-      22: 12 1b 1b 1a                     xadd32 sp, sp, x26
-      26: 00                              ret
+       3: 12 7b 6b                        xadd32 sp, sp, x26
+       6: 2c 1b 08 1c                     store64_offset8 sp, 8, lr
+       a: 2a 1b 1d                        store64 sp, fp
+       d: 0b 1d 1b                        xmov fp, sp
+      10: 12 00 04                        xadd32 x0, x0, x1
+      13: 0b 1b 1d                        xmov sp, fp
+      16: 25 1c 1b 08                     load64_offset8 lr, sp, 8
+      1a: 22 1d 1b                        load64 fp, sp
+      1d: 0e 1a 10                        xconst8 x26, 16
+      20: 12 7b 6b                        xadd32 sp, sp, x26
+      23: 00                              ret
 ```
 
 Note that there are a number of things that could be improved here:
--- a/pulley/README.md
+++ b/pulley/README.md
@@ -47,17 +47,17 @@ to change, instructions to appear and disappear, and APIs to be overhauled.
 Here is the disassembly of `f(a, b) = a + b` in Pulley today:
 
 ```
-       0: 0e 1f f0                        xconst8 x31, -16
-       3: 12 20 20 1f                     xadd32 sp, sp, x31
-       7: 29 20 08 21                     store64_offset8 sp, 8, lr
-       b: 27 20 22                        store64 sp, fp
-       e: 0b 22 20                        xmov fp, sp
+       0: 0e 1a f0                        xconst8 x26, -16
+       3: 12 1b 1b 1a                     xadd32 sp, sp, x26
+       7: 2c 1b 08 1c                     store64_offset8 sp, 8, lr
+       b: 2a 1b 1d                        store64 sp, fp
+       e: 0b 1d 1b                        xmov fp, sp
       11: 12 00 00 01                     xadd32 x0, x0, x1
-      15: 0b 20 22                        xmov sp, fp
-      18: 25 21 20 08                     load64_offset8 lr, sp, 8
-      1c: 22 22 20                        load64 fp, sp
-      1f: 0e 1f 10                        xconst8 x31, 16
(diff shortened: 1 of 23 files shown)


cfallin:
Reassigned @fitzgen to review, I'm not sure how this intersects with your other plans to add instructions with large offsets etc; happy to review this myself if you'd like otherwise.

pulley/README.md
```diff
@@ -47,18 +47,14 @@ to change, instructions to appear and disappear, and APIs to be overhauled.
...
-       e: 0b 22 20                        xmov fp, sp
-      11: 12 00 00 01                     xadd32 x0, x0, x1
-      15: 0b 20 22                        xmov sp, fp
-      18: 25 21 20 08                     load64_offset8 lr, sp, 8
-      1c: 22 22 20                        load64 fp, sp
-      1f: 0e 1f 10                        xconst8 x31, 16
-      22: 12 20 20 1f                     xadd32 sp, sp, x31
-      26: 00                              ret
+       0: 2d 01                           spush lr
+       2: 2d 02                           spush fp
```
fitzgen, line 51: I was indeed thinking that the special registers would only be accessed from special instructions, but I think we can do even more here by bundling the whole function prologue and epilogue up into instructions that just take the size to decrement the stack frame by:

```
enter_frame 16
...
exit_frame 16
ret
```

where

```
enter_frame N

==>

push lr
push fp
fp = sp
sp = sp - N
```

and the opposite for `exit_frame N`.

pulley/src/regs.rs
```diff
@@ -1,181 +1,187 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
+#[allow(non_camel_case_types, missing_docs)]
+#[rustfmt::skip]
+pub enum VReg {
+    v0,  v1,  v2,  v3,  v4,  v5,  v6,  v7,  v8,  v9,
+    v10, v11, v12, v13, v14, v15, v16, v17, v18, v19,
+    v20, v21, v22, v23, v24, v25, v26, v27, v28, v29,
+    v30, v31,
+}
```
fitzgen, line 119: Why all these changes? Seems unrelated?

pulley/src/lib.rs
```diff
@@ -50,13 +46,17 @@ macro_rules! for_each_op {
             /// Branch if unsigned `a <= b`.
             br_if_xulteq32 = BrIfXulteq32 { a: XReg, b: XReg, offset: PcRelOffset };
 
+
+            /// Move between `s` registers.
+            smov = Smov { dst: SReg, src: SReg };
```
fitzgen, line 51: We shouldn't need this, or any other general instructions for the special registers. They should only be manipulated by instructions that are doing other stuff as well, eg `call` will set `lr`. But these registers aren't allocated by our register allocator and are reserved for their special purpose, so we shouldn't need to have general instructions for them.

pulley/src/interp.rs
```diff
@@ -524,13 +532,37 @@ impl fmt::Debug for MachineState {
...
+                &mut self.$field[reg.index()]
+            }
+        }
+    };
+}
+
+index_reg!(SReg, XRegVal, s_regs);
+index_reg!(XReg, XRegVal, x_regs);
+index_reg!(FReg, FRegVal, f_regs);
+index_reg!(VReg, VRegVal, v_regs);
```
fitzgen, line 556: Could we split this out into its own PR?

fitzgen (commented):
Thanks! `BinaryOperands` is exactly what I was imagining!

I'd prefer holding off on merging this until I can land the Cranelift backend for Pulley though, just so that I don't have to chase instruction set updates, and we can start evolving the backend and the instruction set together and keep them in sync.

pulley/src/regs.rs
Kmeakin: Encodes the invariant that registers are always in the range `0..32`, so indexing into the registers array can never fail. The other option would be to use the `#[rustc_layout_scalar_valid_range_end(31)]` attribute, but that is nightly-only.

fitzgen: Are you saying we get better codegen with `enum`s here? That's neat, I wouldn't have expected the LLVM IR that rustc emits to change.

pulley/README.md
Kmeakin: I'm a bit confused about the purpose of the `N` to `enter_frame`/`exit_frame`. Each `push`/`pop` will itself automatically adjust the sp by `size_of<*u8>()`. So in the case of no other registers being stored on the stack, the argument will be zero. In the case where other registers need to be saved on the stack, `xpush{32,64}` and `xpop{32,64}` will also adjust the sp by 4/8 bytes. So in either case the argument is not needed? I don't think there is any scenario where you would need to adjust the stack pointer without also reading/writing from/to the stack

fitzgen: > I'm a bit confused about the purpose of the `N` to `enter_frame`/`exit_frame`.

To allocate the stack frame for this function, including all of its regalloc spills and all that.

That said, the way that Cranelift is organized and the abstraction boundaries it has might make doing the full function prologue in a single instruction difficult. We might have to settle for two instructions:

1. a macro instruction for `push lr; push fp; fp = sp` (and the opposite for exiting the function, maybe just fold this into `ret`)
2. an instruction for allocating and deallocating stack space: `sp = sp +/- imm32`

See `gen_prologue_frame_setup`, `gen_epilogue_frame_restore`, and `gen_sp_reg_adjust` in `../pulley_shared/abi.rs` in https://github.com/bytecodealliance/wasmtime/pull/9089 (or the equivalent methods in, e.g., `../aarch64/abi.rs`). I'm thinking that each of these should produce a single macro instruction because they otherwise always generate the exact same sequence of instructions modulo immediate constant values.

The benefit of a single macro instruction, compared to a sequence of composable and more general instructions, is performance. The expensive part of an interpreter is roundtripping through the opcode loop, by using macro instructions we are amortizing that cost.

Does that make sense?

Kmeakin: Decided the extra argument was not needed after all:
https://bytecodealliance.zulipchat.com/#narrow/stream/217126-wasmtime/topic/Pulley.20registers/near/459533622

fitzgen:
Heads up: the Pulley backend for Cranelift landed in #9089 and I think this can be rebased and landed now

fitzgen (approved):
Thanks!

Issue #9058, opened 2024-08-01
Remove special case for InputStream::File ?


badeend:
All input streams are implemented using the `HostInputStream` trait. Except files. They get special treatment. This can be seen in the definition of `InputStream`:

https://github.com/bytecodealliance/wasmtime/blob/ba864e987ef1ab87c439ca6b396264547d2425e1/crates/wasi/src/stream.rs#L165-L168

The special case was introduced to work around the fact that OS'es don't actually provide any _true_ async APIs for files. A more detailed explanation can be read in the PR that introduced this setup: https://github.com/bytecodealliance/wasmtime/pull/6556

---

Currently, `FileInputStream::read` spawns the read syscall on a background thread and then immediately `await`s the newly created task. This is done to prevent blocking the executor, but from the WASM guest's point-of-view this doesn't provide any asynchronicity.

My question: Instead of directly awaiting the background task, can we store the task handle in the `FileInputStream` instance and then use that handle to wait for readiness? Subsequent calls to `FileInputStream::read` then inspect the result of the previously started background task (if any). And then let `FileInputStream` implement `HostInputStream` directly and get rid of the special case.

From the surface this seems like an "obvious" solution, but the PR above also mentions:
> We spent a the better part of a week trying to square this circle and this is the best we could come up to make files work

which makes me suspect there's more to the story than meets the eye.

@pchickey, what do you think?

alexcrichton:
One major difficulty here is cancellation. All other `HostInputStream` primitives can be cancelled at any time and dropped at any time, but if a background task is spawned there's no way to cancel that if the write has actually reached the OS. Most other primitives use abort-on-drop handles around task handles but that only works because the async operations performed don't actually block so the abort process will actually abort whatever is pending.

badeend:
> One major difficulty here is cancellation

Do you mean "difficulty" from a correctness point of view? Or from a Technical implementation / resource usage POV?

> there's no way to cancel that if the write has actually reached the OS.

You mention "write", but this issue is specifically about input stream, i.e. _reads_ only. Don't know if that changes things.

Is there any harm in letting the read continue in the background?

alexcrichton:
Sorry yeah my reading comprehension isn't great recently... But yes without true support for cancellation the guest can create an unbounded number of cancelled reads and the host has no way to handle that.

[...]

PR #9091, opened 2024-08-08
Generate async drop methods for resources.

badeend:
Add the ability to generate async drop methods for resources.

In the component model, `resource.drop` is a canonical built-in without a proper name. So I invented a custom naming scheme for the component bindgen config. I went with: `"[drop]{resource-name}"` where `{resource-name}` is the name as defined in WIT. e.g. `"[drop]input-stream"`.

This shouldn't conflict with anything existing in the wild as WIT identifiers are not allowed to contain square brackets.

---

To the reviewer: please double check `resource_async`. I'm not familiar with this part of the codebase.

---

Spawned off from https://github.com/bytecodealliance/wasmtime/issues/9058. This change is a prerequisite for a potential solution of that issue, but seems useful enough on its own.


alexcrichton (approved):
Everything looks good to me, thanks! Would you be up for adding a test for this as well?

badeend:
If you're looking for something to test this feature in isolation I'd be glad to add a test for it, but could you point me towards where the component model tests live currently? I tried looking for them in the usual places but couldn't find them.

If all you're after is that this is tested _somehow_ and don't care about the test being in isolation or not, I've got a follow-up PR planned for https://github.com/bytecodealliance/wasmtime/issues/9058 that will be using this feature. So it will be tested indirectly through that.

alexcrichton:
Yes I think that this would be good to have a dedicated test for just this. Component-model API tests live at https://github.com/bytecodealliance/wasmtime/tree/main/tests/all/component_model and https://github.com/bytecodealliance/wasmtime/blob/main/tests/all/component_model/async.rs has some examples of doing async things

badeend:
Check!

alexcrichton (approved):

PR #9092, opened 2024-08-08
Winch aarch64 memory load & store

vulc41n:
Hey 👋

This PR implements `wasm_load`, `wasm_store` and `load_ptr` instructions for winch targeting aarch64.

#8321

### DIFF
--- a/winch/codegen/src/isa/aarch64/masm.rs
+++ b/winch/codegen/src/isa/aarch64/masm.rs
@@ -162,8 +162,8 @@ impl Masm for MacroAssembler {
         self.asm.ldr(src, dst, size, false);
     }
 
-    fn load_ptr(&mut self, _src: Self::Address, _dst: Reg) {
-        todo!()
+    fn load_ptr(&mut self, src: Self::Address, dst: Reg) {
+        self.load(src, dst, self.ptr_size);
     }
 
     fn wasm_load(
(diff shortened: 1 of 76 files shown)


fitzgen:
Redirecting review to @saulecabrera

winch/codegen/src/isa/aarch64/masm.rs
```diff
@@ -159,21 +159,27 @@ impl Masm for MacroAssembler {
...
+        dst: Reg,
+        size: OperandSize,
+        kind: Option<ExtendKind>,
     ) {
-        todo!()
+        self.asm.ldr(
+            src,
+            dst,
+            size,
+            kind.is_some(), // kind is some if the value is signed
```
saulecabrera, line 180: Could `ldr` instead of accepting a boolean accept `Option<ExtKind>`? Or instead -- and maybe this is what I'd prefer -- we could introduce two methods at the assembler level:

- `sload`
- `uload`

And we could dispatch from `wasm_load` to the corresponding method depending on the value of `kind`, this would make more explicit depending on the extension kind. Additionally this would make it less error prone than passing a boolean.

saulecabrera (commented):
Looking good. Left one minor comment before landing this one.

saulecabrera (approved):

PR #9096, opened 2024-08-09
add `rust-toolchain.toml`

Kmeakin:
I hit a confusing error trying to build wasmtime on the nightly toolchain (https://bytecodealliance.zulipchat.com/#narrow/stream/217126-wasmtime/topic/.22failed.20to.20reduce.20input.20adapter.20module.20to.20its.20minimal.20size.22/near/456716867). I thought it would be a good idea to include a `rust-toolchain.toml` file so `cargo build` just works out of the box without any setup required

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,4 @@
+[toolchain]
+channel = "stable"
+targets = ["wasm32-unknown-unknown", "wasm32-wasip1"]
+profile = "default"


alexcrichton:
Thanks! Personally though I'd prefer to not do this. I find `rust-toolchain.toml` difficult to deal with because it's difficult to override for example in CI to try out different versions (e.g. this is overriding on CI right now to `stable` instead of pinned versions). Possible to do of course but I also find it a bit jarring to run `cargo build` locally only to have a toolchain downloaded.

Not to say I disagree with the motivation of this PR though. Confusing errors are never great and pinning dependencies are a great way of weeding that out. My only defense against this is a hand-wavy argument that we're generally not affected by breakage in upstream rust-lang/rust. My experience is it's pretty common to get new lint warnings (which doesn't break builds) and otherwise full-on breakage is pretty rare. If we were a much larger project where this happened more frequently I think it might make sense to pin toolchains, but in lieu of that I'd lean towards the default of "fix the occasional bug report when this does happen" and otherwise stick to only having an msrv marker for warning the toolchain is too old

Kmeakin:
Thanks, I'll close this

PR #9097, opened 2024-08-09
pulley: use `Index{Mut}` instead of getters and setters

Kmeakin:
Split off from https://github.com/bytecodealliance/wasmtime/pull/9088


fitzgen (approved):
Fantastic, thanks!

fitzgen:
Looks like this might have been a flaky github upload-artifact action. Going to retry it.

PR #9099, opened 2024-08-09
pulley: superinstructions for pushing/popping list of registers

Kmeakin:
Adds superinstructions for pushing/popping any subset of XRegs in one go

### DIFF
--- a/pulley/fuzz/src/interp.rs
+++ b/pulley/fuzz/src/interp.rs
@@ -108,6 +108,8 @@ fn op_is_safe_for_fuzzing(op: &Op) -> bool {
         Op::PushFrame(_) | Op::PopFrame(_) => false,
         Op::XPush32(_) | Op::XPush64(_) => false,
         Op::XPop32(_) | Op::XPop64(_) => false,
+        Op::XPush32Many(_) | Op::XPush64Many(_) => false,
+        Op::XPop32Many(_) | Op::XPop64Many(_) => false,
     }
 }
 
--- a/pulley/fuzz/src/interp.rs
+++ b/pulley/fuzz/src/interp.rs
@@ -105,6 +105,9 @@ fn op_is_safe_for_fuzzing(op: &Op) -> bool {
         | Op::Xslteq32(Xslteq32 { operands, .. })
         | Op::Xult32(Xult32 { operands, .. })
         | Op::Xulteq32(Xulteq32 { operands, .. }) => !operands.dst.is_special(),
+        Op::PushFrame(_) | Op::PopFrame(_) => false,
+        Op::XPush32(_) | Op::XPush64(_) => false,
+        Op::XPop32(_) | Op::XPop64(_) => false,
     }
 }
 
--- a/pulley/src/interp.rs
+++ b/pulley/src/interp.rs
@@ -1008,8 +1008,22 @@ impl OpVisitor for InterpreterVisitor<'_> {
         Continuation::Continue
     }
 
+    fn xpush32_many(&mut self, srcs: RegSet<XReg>) -> Self::Return {
+        for src in srcs {
+            self.xpush32(src);
+        }
+        Continuation::Continue
+    }
+
     fn xpush64(&mut self, src: XReg) -> Self::Return {
         self.state.push(self.state[src].get_u64());
+        Continuation::Continue
+    }
+
+    fn xpush64_many(&mut self, srcs: RegSet<XReg>) -> Self::Return {
+        for src in srcs {
+            self.xpush64(src);
+        }
         Continuation::Continue
     }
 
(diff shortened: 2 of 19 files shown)


fitzgen:
And then once https://github.com/bytecodealliance/wasmtime/pull/9088 is rebased and lands, we can do the same with this one. Thanks again for making these improvements @Kmeakin!

fitzgen:
This needs a rebase after #9088

pulley/fuzz/src/interp.rs
```diff
@@ -105,6 +105,15 @@ fn op_is_safe_for_fuzzing(op: &Op) -> bool {
...
         | Op::Xulteq32(Xulteq32 { operands, .. }) => !operands.dst.is_special(),
+        Op::PushFrame(_) | Op::PopFrame(_) => false,
+        Op::XPush32(XPush32 { src }) | Op::XPush64(XPush64 { src }) => !src.is_special(),
+        Op::XPop32(XPop32 { dst }) | Op::XPop64(XPop64 { dst }) => !dst.is_special(),
+        Op::XPush32Many(XPush32Many { srcs }) | Op::XPush64Many(XPush64Many { srcs }) => {
+            !srcs.into_iter().any(|src| src.is_special())
+        }
+        Op::XPop32Many(XPop32Many { dsts }) | Op::XPop64Many(XPop64Many { dsts }) => {
+            !dsts.into_iter().any(|dst| dst.is_special())
+        }
```
fitzgen, line 116: These are only safe if we have enough stack space to push to or pop from, right?

So either we should make these not safe for fuzzing, or we should keep track of how much stack we've consumed and still have available at the point where this `op` is executed and gate `xpush`/`xpop`/`push_frame`/`pop_frame`'s fuzzing safety on whether there is enough stack to safely execute them.

fitzgen (commented):
Looks great! Just a couple things to address below before we merge this.

cranelift/bitset/src/compound.rs
```diff
@@ -346,7 +346,7 @@ impl CompoundBitSet {
             .enumerate()
             .rev()
             .filter_map(|(word, sub)| {
-                let bit = sub.max()?;
+                let bit = sub.last()?;
```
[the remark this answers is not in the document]
Kmeakin: `ScalarBitset` now implements `Ord` (because pulley `Instr` derives `Ord`). So calls to `bitset.min()` will resolve to the `min` method from `Ord`, rather than the inherent method attached to `ScalarBitset`

pulley/src/interp.rs
```diff
@@ -988,6 +1003,74 @@ impl OpVisitor for InterpreterVisitor<'_> {
...
+    }
+
+    /// `pop fp; pop lr; sp = fp`
+    fn pop_frame(&mut self) -> Self::Return {
+        let sp = self.state.pop();
+        let lr = self.state.pop();
+        self.state[XReg::sp].set_ptr::<u8>(sp);
+        self.state[XReg::lr].set_ptr::<u8>(lr);
+        self.state[XReg::sp] = self.state[XReg::fp];
+        Continuation::Continue
```
alexcrichton, line 1071: Are there perhaps some typos here? Otherwise `sp` is set twice. I also wouldn't expect `sp` to be set here at all other than through the side-effects of `pop()`?

cranelift/bitset/src/compound.rs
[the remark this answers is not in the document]
fitzgen: Can we avoid `Ord` for `ScalarBitSet`? And instead implement it for `RegSet`?

pulley/fuzz/src/interp.rs
Kmeakin: I'll return `false` for now, we can add a static analysis later

cranelift/bitset/src/compound.rs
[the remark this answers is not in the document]
Kmeakin: We can remove the `#[derive(PartialOrd, Ord)]` on `Op`, since it's not used anywhere.

pulley/src/interp.rs
Kmeakin: I didn't reverse the logic from `push_frame` completely. IIUC, sp should be restored from fp before popping fp and lr.

fitzgen (approved):
Looks great, thanks!

I hope to get the runtime integration soon, and then we can start enabling Wasm spec tests and such one by one until we have everything running under Pulley!

alexcrichton:
I believe the error in CI is that in the `scripts/publish.rs` file the `cranelift-bitset` crate needs to be reordered before the `pulley` family of crates since they now depend on `cranelift-bitset`

fitzgen:
Side note, it will be a little faster to iterate here to add an empty commit with the message `prtest:full` so that the full CI checks happen and we don't have to wait for the merge queue.

fitzgen:
I'm not totally sure what is going on here, new-ish dependencies for new-ish are always a little annoying to get set up correctly so that when we publish we don't run into any issues. I guess we are effectively front-loading the issues, rather than dealing with them at publish time, which is probably the right choice. Anyways, I think you should be able to repro locally with

```shell-session
$ rustc scripts/publish.rs
$ ./publish verify
```

Note that if you have a `.cargo` directory in your wasmtime checkout already, you'll want to move it somewhere else temporarily while running the above.

fitzgen:
> I believe the error in CI is that in the `scripts/publish.rs` file the `cranelift-bitset` crate needs to be reordered before the `pulley` family of crates since they now depend on `cranelift-bitset`

^ Did the fix for this perhaps get lost in a rebase?

PR #9100, opened 2024-08-09
Remove wasmtime link dep on libm

pchickey:
prtest:full


alexcrichton (approved):
> CI overrides the linker with a musl-gcc install. I'm not sure why, ...

This is because our build containers are all Ubuntu so the default linker uses glibc instead of musl. The `musl-gcc` package is sort of a few wrapper scripts to tell the linker to link musl instead of glibc. In your distro it might be musl-based so this all happens by default perhaps?

In any case looks good to me, now I'm not sure why I needed this but maybe I got something wrong while developing this and accidentally thought it was required.

PR #9105, opened 2024-08-11
Fix more `clippy::doc_lazy_continuation` lints

waywardmonkeys:
This fixes all remaining instances of `clippy::doc_lazy_continuation` except for in some generated code.

### DIFF
--- a/cranelift/codegen/meta/src/shared/mod.rs
+++ b/cranelift/codegen/meta/src/shared/mod.rs
@@ -43,9 +43,9 @@ pub(crate) fn define() -> Definitions {
 /// Verifies certain properties of formats.
 ///
 /// - Formats must be uniquely named: if two formats have the same name, they must refer to the
-/// same data. Otherwise, two format variants in the codegen crate would have the same name.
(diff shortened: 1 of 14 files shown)


alexcrichton (approved):
Thanks!

I'm curious though, how did you see this warning? We should in theory have all lints allowed for clippy by default except a few listed in the top-level `Cargo.toml`. Do you know how that configuration was ignored?

Orthogonally, if you're up for it, adding this lint to `Cargo.toml` to have this checked on CI I think would make sense.

waywardmonkeys:
I found them because I'd added it to the top level Cargo to enable it. :)

I'd wanted to enable it in the commit, but ran into an issue with a couple of remaining ones in generated code and didn't quite have the time at the moment to fix them, but also didn't want merge conflicts on these if I let it all sit until I had the time.

alexcrichton:
Oops I meant to flag this for merge yesterday. Doing so now.

Also makes sense! And yeah feel free to enable them at the top-level too. It's ok to add `#[allow]` to generated code as we've done that for ISLE code before (we can perhaps change that to `allow(clippy::all)` too or something like that)

PR #9108, opened 2024-08-11
ISLE: lexer simplifications

Kmeakin:
* Instead of pushing every byte in a lexed integer into a `Vec<u8>`, use a slice of the original buffer, and only reallocate if there are underscores that need to be removed
* Don't track line/column in `Pos`, so don't need to update line/column in `advance_pos()`
* Store text being lexed as `&str` rather than `&[u8]` to avoid checking substrings for UTF-8 validity
* Don't try to parse integers twice

### DIFF
--- a/cranelift/isle/isle/src/lexer.rs
+++ b/cranelift/isle/isle/src/lexer.rs
@@ -105,23 +105,22 @@ impl<'src> Lexer<'src> {
         }
 
         // Skip any whitespace and any comments.
-        while self.pos.offset < self.src.len() {
-            if self.src.as_bytes()[self.pos.offset].is_ascii_whitespace() {
-                self.advance_pos();
-                continue;
-            }
-            if self.src.as_bytes()[self.pos.offset] == b';' {
-                while self.pos.offset < self.src.len()
-                    && self.src.as_bytes()[self.pos.offset] != b'\n'
-                {
-                    self.advance_pos();
+        while let Some(c) = self.peek_byte() {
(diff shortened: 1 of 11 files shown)


pchickey:
Hi, thanks for these contributions! If I can nerd-snipe you into some unrelated lexer work, we have a `logos` lexer over in the `wasm-wave` crate that we need to be rewritten entirely in safe rust. If you are interested in learning more, ping me over on https://github.com/bytecodealliance/wasmtime/pull/8872

cranelift/isle/isle/src/lexer.rs
```diff
@@ -247,6 +244,10 @@ impl<'src> Lexer<'src> {
     pub fn eof(&self) -> bool {
         self.lookahead.is_none()
     }
+
+    fn peek_byte(&self) -> Option<u8> {
+        self.src.as_bytes().get(self.pos.offset).copied()
+    }
```
fitzgen, line 253: Why not `fn peek_char(&self) -> Option<char>`? I guess a byte is mildly more performant, in theory if LLVM doesn't clean things up, but that doesn't seem super compelling compared to the simplification of "we are dealing with strings, so we should also deal with chars".

fitzgen (approved):
This is fantastic, thanks!

Apologies for the slow review turn around, I've been having ISP issues.

One comment below that I'd like to get your take on, but shouldn't be a blocker for landing this. Can address it in a follow up, if necessary.

PR #9109, opened 2024-08-12
Fix file and line in generated DWARF

philipc:
We were using a string, but the DWARF standard specifies that: "The value of the DW_AT_decl_file attribute corresponds to a file number from the line number information table...".

Additionally, a typo meant we were overwriting the file attribute with the value that was meant to be used for a DW_AT_decl_line attribute.


alexcrichton (approved):
Thanks for this!

PR #9110, opened 2024-08-12
Support DWARF packages for `wasmtime compile`

philipc:
Previously we didn't pass the path to the Wasm file, so we didn't find the associated DWARF package.

I'm not sure if this is the right fix. Maybe `Engine` needs a new method instead?

I've also added a test for a similar problem in the tests. This doesn't actually test `wasmtime compile` though.


alexcrichton (approved):
Thanks!

PR #9111, opened 2024-08-12
test-programs: use the wasi:http/imports world

iawia002:
In version 0.2.1 of `wasi-http`, a new `imports` world was introduced, this allows us to manage our `test` world in `test-programs` with a simpler and more consistent style.


alexcrichton (approved):
Thanks!

PR #9113, opened 2024-08-12
Clean up dist build configuration

alexcrichton:
* Move updating `$PATH` to the `main.js` script which is the one that mounts `/rust/bin` so that knowledge isn't spread around.
* Remove some unused env vars in docker containers.
* Forward cargo/rust-specific env vars to the build from outside of containers to the build itself.

### DIFF
--- a/.github/actions/binary-compatible-builds/main.js
+++ b/.github/actions/binary-compatible-builds/main.js
@@ -32,7 +32,26 @@ if (process.env.INPUT_NAME && process.env.INPUT_NAME.indexOf("android") >= 0) {
 // commands in there with the `$CENTOS` env var.
 
 if (process.env.CENTOS !== undefined) {
-  const args = ['exec', '-w', process.cwd(), '-i', 'build-container'];
+  const args = ['exec', '--workdir', process.cwd(), '--interactive'];
+  // forward any rust-looking env vars from the environment into the container
+  // itself.
+  for (let key in process.env) {
+    if (key.startsWith('CARGO') || key.startsWith('RUST')) {
+      args.push('--env');
+      args.push(key);
+    }
+  }
+  args.push('build-container')
+
+  // Start the container by appending to `$PATH` with the `/rust/bin` path that
+  // is mounted below.
+  args.push('bash');
+  args.push('-c');
+  args.push('export PATH=$PATH:/rust/bin; exec "$@"');
+  args.push('bash');
+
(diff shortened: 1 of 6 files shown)


.github/actions/binary-compatible-builds/main.js
```diff
@@ -32,7 +32,26 @@ if (process.env.INPUT_NAME && process.env.INPUT_NAME.indexOf("android") >= 0) {
 // commands in there with the `$CENTOS` env var.
 
 if (process.env.CENTOS !== undefined) {
-  const args = ['exec', '-w', process.cwd(), '-i', 'build-container'];
+  const args = ['exec', '--workdir', process.cwd(), '--interactive'];
+  // forward any rust-looking env vars from the environment into the container
```
abrown, line 36:
```suggestion
  // Forward any rust-looking env vars from the environment into the container
```

pchickey (approved):

.github/actions/binary-compatible-builds/main.js
```diff
@@ -32,7 +32,26 @@ if (process.env.INPUT_NAME && process.env.INPUT_NAME.indexOf("android") >= 0) {
...
+    }
+  }
+  args.push('build-container')
+
+  // Start the container by appending to `$PATH` with the `/rust/bin` path that
+  // is mounted below.
+  args.push('bash');
+  args.push('-c');
+  args.push('export PATH=$PATH:/rust/bin; export RUSTFLAGS="$RUSTFLAGS $EXTRA_RUSTFLAGS"; exec "$@"');
+  args.push('bash');
```
abrown, line 51: Bash inside bash? Not sure I get why it's preferable to do this dance rather than setting things up in the Dockerfile.

abrown (approved):

.github/actions/binary-compatible-builds/main.js
alexcrichton: Apparently this particular argument is just informational and used for error messages, so I put "bash" here but any string could work. Otherwise though the main motivation here was to mroe tightly couple the addition of the environment's Rust compiler to the docker image itself. Previously that was spread across this `main.js` and each `Dockerfile`, but now it's entirely contained within this `main.js` alone. In theory that should make the `Dockerfile` images a bit more freestanding and self-describing. 

I realize though in the end this is all intertwined anyway so it probably doesn't help all that much really.

PR #9114, opened 2024-08-12
Winch aarch64 extend, wrap, popcnt, promote & demote

vulc41n:
Hey 👋

This PR implements `extend`, `wrap`, `popcnt`, `promote` & `demote` instructions for winch targeting aarch64.

#8321


saulecabrera:
I can help reviewing this one.

winch/codegen/src/isa/aarch64/masm.rs
```diff
@@ -429,8 +430,20 @@ impl Masm for MacroAssembler {
         self.asm.load_constant(0, reg);
     }
 
-    fn popcnt(&mut self, _context: &mut CodeGenContext, _size: OperandSize) {
-        todo!()
+    fn popcnt(&mut self, context: &mut CodeGenContext, size: OperandSize) {
+        let src = context.pop_to_reg(self, None);
+        let tmp = context.reg_for_class(RegClass::Float, self);
```
saulecabrera, line 435: To reduce register pressure, could we use the scratch float register here?

tests/disas/winch/aarch64/i64_popcnt/fallback.wat
saulecabrera, line 1: In the case of `aarch64` I believe that the `fallback` test is redundant i.e., since we are not generating an alternative sequence of instructions like in x64.

winch/codegen/src/isa/aarch64/masm.rs
```diff
@@ -429,8 +430,20 @@ impl Masm for MacroAssembler {
...
+        self.asm.mov_to_fpu(src.into(), tmp, size);
+        self.asm.cnt(tmp);
+        match size {
+            OperandSize::S8 => {}
+            OperandSize::S16 => self.asm.addp_rrr(tmp, tmp, tmp, VectorSize::Size8x8),
```
saulecabrera, line 440: If I'm not wrong, Wasm only defined `{i32, i64}_popcnt` so I believe we only need to handle the `S32 | S64` case and handle all the other operand sizes with `unreachable!()` instead of `unimplemented`, which also means that we can probably drop the `addp_rrr` implementation.

vulc41n: You're right, I used `lower.isle` as a reference but CLIF is different from WASM. 

> handle all the other operand sizes with `unreachable!()` instead of `unimplemented`

I just removed the match, please tell me if you think an assertion is needed

saulecabrera (commented):
One minor nit and this should be good to go.

saulecabrera (approved):
Thanks!

Issue #9090, opened 2024-08-08
s390x: no rule matched for term vec_elt_rev


alexcrichton:
This input:

<details>

<summary><code>input.clif</code></summary>

```
;;Output of `std::fmt::Debug`:
;; Compile test case

test compile
set opt_level=speed_and_size
set regalloc_checker=true
set preserve_frame_pointers=true
set machine_code_cfg_info=true
set enable_heap_access_spectre_mitigation=false
set enable_table_access_spectre_mitigation=false
target s390x has_vxrs_ext2

function u1:0(i64x2, i128 uext, i16x8, f64, i64, i16, i32x4, f32, f32x4, i8x16, f64x2, i32, i8 sext) -> i8x16, i8, i64x2, i64x2, i64x2, i64x2, i64x2, i64x2, i64x2, i64x2, i64x2, i64x2, i64x2, i64x2 system_v {
    ss0 = explicit_slot 17, align = 64
    ss1 = explicit_slot 17, align = 64
    ss2 = explicit_slot 17, align = 4
    ss3 = explicit_slot 1
    ss4 = explicit_slot 67, align = 512
    ss5 = explicit_slot 67, align = 512
    ss6 = explicit_slot 67, align = 512
    ss7 = explicit_slot 67, align = 512
    ss8 = explicit_slot 16, align = 16
    ss9 = explicit_slot 16, align = 16
    ss10 = explicit_slot 4, align = 4
    ss11 = explicit_slot 2, align = 2
    ss12 = explicit_slot 8, align = 8
    ss13 = explicit_slot 16, align = 16
    ss14 = explicit_slot 2, align = 2
    ss15 = explicit_slot 16, align = 16
    ss16 = explicit_slot 2, align = 2
    ss17 = explicit_slot 1
    ss18 = explicit_slot 16, align = 16
    ss19 = explicit_slot 16, align = 16
    ss20 = explicit_slot 2, align = 2
    sig0 = (i8 uext, i64x2, f32, i16 uext, i16, i8, i64x2, f64, f64, f64, f64x2, i64 uext, i16x8, i16, i64x2, i16x8) -> i8, f32x4, i128 uext, i16x8, i8 sext, i32x4, f64, f32, i32x4 tail
    sig1 = (i16 sext) -> i16x8, f32, f32 system_v
[...]

PR #9121, opened 2024-08-12
s390x: Fix vec_elt_rev for i128

uweigand:
Add missing support for $I128 in vec_elt_rev (which is a no-op).

Fixes: https://github.com/bytecodealliance/wasmtime/issues/9090


cfallin (approved):
Thanks!

PR #9128, opened 2024-08-14
Disable the trace-log feature of regalloc2 by default

bjorn3:
This makes register allocation a bit faster when these logs are not used.


cfallin (approved):
It looks like this was (probably accidentally) added as part of #8196 and has been on since March (cc @fitzgen, though I've made the same mistake in PRs before too; this one just slipped through I guess). Thanks for finding it!

Amanieu:
The `trace-log` feature on cranelift-codegen is also unconditionally enabled.

bjorn3:
Fixed.

cfallin (approved):
Thanks @Amanieu for catching the top-level change as well!

fitzgen:
Whoops yeah the intention was to enable it by default for `clif-util` but not for everything all the time!

PR #6556, opened 2023-06-09
WASI Preview 2: rewrite streams and pollable implementation

pchickey:
When we landed the WASI Preview 2 support in `wasmtime-wasi` we knew the streams/poll/scheduler implementation was going to get torn out and rewritten: see task 4 in https://github.com/bytecodealliance/wasmtime/issues/6370. This is that rewrite. Most of this was pair-programmed with @elliottt.

# Features 
`wasmtime-wasi` Preview 2 support is now built on top of `tokio`, and the host implementations for streams and pollable always use `async` Rust. It now passes all of the poll_oneoff tests, including on windows.

The new `HostInputStream`, `HostOutputStream`, and `HostPollable` traits are used to implement the `wasi:io/streams.{input-stream,output-stream}` and `wasi:poll/poll.pollable` resources.

## `Host{Input, Output}Stream` design

`input-stream` and `output-stream` resources have a significant number of methods. For the host traits, we managed to boil this down to two required methods, as well as some optional methods to cover vectored IO, splice, and write-zeroes. The default implementations provide an unoptimized but correct implementation - the user may override them if an optimization is desired.

`HostInputStream` has just two methods: `fn read(&mut self, ...)` and `async fn ready(&mut self)`. The most important invariant is that `read` must be non-blocking, and `ready` must block until the stream is ready for reading.

## Nonblocking IO is actually impossible on regular files, but we hide that detail from WASI programs

There's a huge lie in the description above: internally, wasmtime_wasi uses `InternalHost{Input,Output}Stream`, which are enums that contain either a `Host{Input,Output}Stream`, or a private `File{Input,Output}Stream`. These file streams always perform a blocking read/write, whether or not the WASI program has called the nonblocking `read` or the blocking `blocking_read` method. And, any pollable associated with file read or write readiness always returns ready immediately.

This is required because there is no way to implement streams on files with `epoll(7)` (or on macos or windows using mio, afaik) based readiness - `poll(2)` will lie and always tell you files are ready for reading of writing, but the actual syscall may end up blocking the thread due to disk IO. And, rather than stick the work on a background task, we must wait until the OS `read(2)` or `write(2)` for a file has completed before returning from wasi stream (nonblocking) read or write, because the userlands in WASI expect (reasonably!) that any errors in regular file IO are reported synchronously.

[...]

PR #9129, opened 2024-08-14
wasi-filesystem: Implement `HostInputStream` & `HostOutputStream`

### LINKED ISSUE #9058: Remove special case for InputStream::File ?

badeend:
Fixes #9058

---

All input streams are implemented using the `HostInputStream` trait. Except files. They got special treatment. This can be seen in the definition of `InputStream`:

https://github.com/bytecodealliance/wasmtime/blob/ba864e987ef1ab87c439ca6b396264547d2425e1/crates/wasi/src/stream.rs#L165-L168

The special case was introduced to work around the fact that OS'es don't actually provide any _true_ async APIs for files. A more detailed explanation can be read in the PR that introduced this setup: https://github.com/bytecodealliance/wasmtime/pull/6556

---

This PR properly implements the `HostInputStream` & `HostOutputStream` traits for files. And with "properly" I mean:
- methods that shouldn't block (e.g. `read` & `write`) now really don't block.
- methods that are allowed to block (e.g. `blocking_read` & `blocking_write_and_flush`) continue to block and still take advantage of `allow_blocking_current_thread`.

Primary changes:
- `InputStream` has been changed from an enum to a type alias, just like `Outputstream` already is.
- The Host(Input/Output)Stream traits have been extended to allow implementors to provide specialized `blocking_*` implementations. In this PR, only the File streams actually override these.
- Added an `async fn cancel(&mut self) {}` method to the Host(Input/Output)Stream traits. This is called right before the stream is dropped and can be used to wait for asynchronous cleanup.
  - On File streams, this is used to wait on any in flight read or write. Operating systems don't provide _real_ asynchronous APIs for files, so we end up spawning uncancellable blocking background tasks. In the new `cancel` method we wait for that read/write to run to completion. From the guest's point of view, `input/output-stream::drop` then appears to block. Certainly less than ideal, but arguably still better than letting the guest rack up an unbounded number of background tasks. Also, the guest is only blocked if the stream was dropped mid-read or mid-write, which is not expected to occur frequently.
  - I also implemented it for [various other stream types](https://github.com/bytecodealliance/wasmtime/commit/d1b891f30f389fd51e7c80a0297afff283e9b0fd). Unlike Files, the background tasks of these stream types are truly async. So aborting _without_ awaiting them was probably already good enough in practice. Nonetheless, waiting for the background task to actually shut down feels more "hygienic" to me in terms of resource management.

Slightly related changes:
- In Preview1 adapter: ignore BlockingMode and always perform blocking I/O. That's what preview1 did anyway.
- Remove `[method]output-stream.forward` from bindgen config. That method does not exist.
- Refactor `blocking_splice` to take advantage of specialized `blocking_read` & `blocking_write_and_flush` implementations
- Rename filesystem's `spawn_blocking` -> `run_blocking`


crates/wasi/src/filesystem.rs
```diff
@@ -236,45 +252,175 @@ impl Dir {
...
+                        Ok(b)
+                    }
+                    Err(e) => {
+                        self.state = ReadState::Closed;
+                        Err(StreamError::LastOperationFailed(e.into()))
+                    }
+                }
+            }
+            ReadState::DataAvailable(_) => Ok(self.read_available_data(size)),
+            ReadState::Waiting(_) => Ok(Bytes::new()),
```
alexcrichton, line 381: Should this perhaps `await` the result of the read here inline?

crates/wasi/src/filesystem.rs
```diff
@@ -236,45 +252,175 @@ impl Dir {
...
+        if let ReadState::Idle = self.state {
+            // The guest hasn't initiated any read, but is nonetheless waiting
+            // for data to be available. We'll start a read for them:
+
+            const DEFAULT_READ_SIZE: usize = 4096;
+            let p = self.position;
+            self.state = ReadState::Waiting(
+                self.file
+                    .spawn_blocking(move |f| Self::blocking_read(f, p, DEFAULT_READ_SIZE)),
+            );
```
alexcrichton, line 420: I'm a little confused by this, is it required? I would expect that reads are only initiated from the `read` method above (or blocking-read) and the `ready` bits here only wait for completion of an already pending read.

crates/wasi/src/filesystem.rs
```diff
@@ -236,45 +252,175 @@ impl Dir {
...
+                    }
+                    Ok(b) => {
+                        self.position += b.len() as u64;
+                        Ok(b)
+                    }
+                    Err(e) => {
+                        self.state = ReadState::Closed;
+                        Err(StreamError::LastOperationFailed(e.into()))
+                    }
+                }
```
alexcrichton, line 378: The implementation of this seems a little subtle and while I see no issue with it I think it would be best to share as much as possible with the above code. Could something like this work?

* In `blocking_read` first resolve the `Waiting` state if any with a `.await`
* Next handle the `Idle` state by performing a `run_blocking` but the result of that is handled in a shared way with what `wait_ready` does (setting `self.state` internall)
* Next handle the results of the read in the same way as `fn read` above (perhaps even just calling `self.read(size)`?)

That I think will keep the `blocking_read` bits here deduplicated with the rest and the only responsibility of `blocking_read` vs `read` would be to filter the states that `read` can see by doing blocking work up front.

alexcrichton (approved):
Thanks for taking on this refactoring, I really like how it's turned out! I think that this will work quite well.

One thing I can note which I was originally worried about but am no longer is that it's still possible for guests to queue up a lot of tasks. For example if a guest queues a write and then cancels it that'll block the guest waiting to cancel that write. The host may then cancel the host guest which will cancel-the-cancel in a way meaning that the write's just sitting in the ether on a blocking thread waiting to be completed. That initially sounded bad to me but I think this is ok because Tokio has a fixed size of blocking calls which can be "in the ether" and otherwise cancellation of a job in the queue that hasn't actually run yet is successful. This means that it's not possible to get an unbounded number of writes pending at any point with this cancellation.

I've got some minor stylistic thoughts below but otherwise this looks good-to-go to me 👍

crates/wasi/src/filesystem.rs
badeend: > is it required?

Not sure, but the current wasmtime-wasi implementation would push me towards a hesitant "Yes?"
I kinda stole the idea from the stdin implementation that also initiates a new read if there's no data available:

https://github.com/bytecodealliance/wasmtime/blob/c7756bd2654fe1ac795f9dd9d13de41a31cafbac/crates/wasi/src/stdio/worker_thread_stdin.rs#L156-L175

This, combined with the fact that there exist multiple places in the code that blindly assume that calling `.ready().await` will yield readable data:

https://github.com/bytecodealliance/wasmtime/blob/c7756bd2654fe1ac795f9dd9d13de41a31cafbac/crates/wasi/src/host/io.rs#L178-L182
https://github.com/bytecodealliance/wasmtime/blob/c7756bd2654fe1ac795f9dd9d13de41a31cafbac/crates/wasi/src/host/io.rs#L211-L214
https://github.com/bytecodealliance/wasmtime/blob/c7756bd2654fe1ac795f9dd9d13de41a31cafbac/crates/wasi/src/host/io.rs#L231-L234
https://github.com/bytecodealliance/wasmtime/blob/c7756bd2654fe1ac795f9dd9d13de41a31cafbac/crates/wasi/src/stream.rs#L153-L154

crates/wasi/src/filesystem.rs
badeend: Hmm, this should've actually been `unreachable!()` as this state is awaited higher up in the function.

crates/wasi/src/filesystem.rs
badeend: Something like [this](https://github.com/bytecodealliance/wasmtime/pull/9129/commits/7e2b7e9ec507313d13f877d6f7a1557577a2578a)?

alexcrichton: Perfect!

crates/wasi/src/filesystem.rs
alexcrichton: Hm ok, I'm not sure if this will survive the test of time but following suit with stdin doesn't seem unreasonable. This shouldn't get hit most of the time for the time being as WASIp1 shouldn't use this. Let's go ahead and land it 👍

pchickey:
@badeend sorry I didn't get time to review this before it landed, but thanks very much for working on it!

PR #9132, opened 2024-08-15
Use the correct DWARF sections and unit for parsing split DWARF

philipc:
The DWARF sections in `DebugInputContext` were always set to the DWARF sections from the .wasm file. However, when parsing split DWARF, most of the time we want to be using the sections from the .dwp file instead. We were already passing `gimli::Dwarf` to most places already so having the sections in `DebugInputContext` was no benefit. So a lot of this commit is replacing `context` with `dwarf`.

Next, sometimes we were using the wrong `gimli::Dwarf`. In general, we want to use the skeleton unit for parsing line info, and the split unit for everything else. `clone_unit` was wrongly using the DWARF associated with the skeleton unit in many places. Instead of changing all of those places, I've kept the variable names `dwarf` and `unit`, but changed which DWARF they refer to. This also fits better with the non-split operation. So some of the changes in this commit are updating places that were already correct to use the new variable meanings.

Finally, this commit adds a call to `copy_relocated_attributes`. This copies some info from the skeleton unit that is needed when parsing the split unit.

For testing, the existing `dwarf_fission.dwp` is too simple for this to make any difference. I've been testing with one of my own C projects, but that's not something I want to add to wasmtime. Additionally, further fixes are required. I'm open to suggestions for a suitable small but non-trivial project to use for testing purposes.

### DIFF
--- a/crates/cranelift/src/debug/transform/attr.rs
+++ b/crates/cranelift/src/debug/transform/attr.rs
@@ -2,7 +2,7 @@ use super::address_transform::AddressTransform;
 use super::expression::{compile_expression, CompiledExpression, FunctionFrameInfo};
 use super::range_info_builder::RangeInfoBuilder;
 use super::refs::{PendingDebugInfoRefs, PendingUnitRefs};
-use super::{DebugInputContext, Reader, TransformError};
+use super::{Reader, TransformError};
 use anyhow::{bail, Error};
 use cranelift_codegen::isa::TargetIsa;
 use gimli::{write, AttributeValue, DebugLineOffset, DebuggingInformationEntry, Unit};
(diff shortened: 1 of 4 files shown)


alexcrichton (approved):
Thank you for taking the time to review this, it's much appreciated!

> I've been testing with one of my own C projects, but that's not something I want to add to wasmtime. Additionally, further fixes are required. I'm open to suggestions for a suitable small but non-trivial project to use for testing purposes.

Would you happen to have a suggestion about how we can best organize tests here? Should we for example check in C files to get compiled in CI with various clang flags? Or is there a text format for DWARF which we could exploit? (or something else more clever you've done in gimli?)

philipc:
I'm not aware of any text format for DWARF. gimli's testing is nearly all artificial DWARF that the tests manually assemble while running. I think it would be tedious to extend that to manually generating the associated Wasm too.

Compiling a C file in CI would be an option for functionality testing (can I step a debugger to this line and see the correct variables), but for regression tests it becomes dependent on the compiler version installed on the test system (changing the compiler may give different DWARF, so it may no longer be testing the original bug).

Maybe I can come up with a contrived C file that manages to result in a variety of DWARF. Or more likely it would need to be C++.

PR #9134, opened 2024-08-16
Handle more split DWARF attributes and forms

philipc:
This includes GNU split support for DWARF version 4.


alexcrichton (approved):
Thanks!

PR #9136, opened 2024-08-16
Implement a few minor optimizations around 128-bit integers

alexcrichton:
This commit implements a few minor changes for `i128` in both the egraph optimizations and lowerings for x64. The optimization pass will now transform `iconcat` into a `uextend` or `sextend` where appropriate. The x64 backend then pattern-matches this to produce slightly more optimal machine code. Additionally the x64 backend now handles memory/immediate operands a bit better when the argument to a 128-bit operation is an `iconcat`.


elliottt (approved):
Nice!

PR #9137, opened 2024-08-16
 Don't force `Multiple` on multi-result instructions 

alexcrichton:
This commit is a result of [discussion on Zulip][Zulip] and is
attempting to fix an issue where some 128-bit instructions aren't fully
benefitting from load sinking on x64. At a high level 128-bit
addition isn't able to sink loads into instructions for halves of the
128-bit operation. At a lower level the reason for this is that
currently all operands of a multiple-result instruction are considered
multiply-used (as each result could be used) which prevents load
sinking.

Operations on 128-bit integers may be coupled with `isplit` afterwards
which is a multi-result instruction. This then means that the `Multiple`
state flows backwards to the 128-bit operation and all its operands,
including whatever is necessary to produce the individual components of
each 128-bit integer.

The fix in this commit is to introduce the concept of a "root"
instruction from the perspective of the calculation of `ValueUseState`.
In other words `ValueUseState` is no longer an accurate picture of the
function as a whole, but only the parts of the function rooted at a
"root" instruction. This is currently defined as multi-result
instructions meaning that `isplit` for example is a root instruction.
This is coupled with documentation/changes to
`get_value_as_source_or_const` to never allow looking through root
instructions (or considering them a `UniqueUse`).

This commit additionally updates some documentation in a few places and
refactors some usage of `get_value_as_source_or_const` to use other
helpers instead to reduce callers of `get_value_as_source_or_const` and
who to audit when modifying this function.

[Zulip]: https://bytecodealliance.zulipchat.com/#narrow/stream/217117-cranelift/topic/ValueUseState.3A.3AMultiple.20and.20multi-result.20instructions/near/462833578

### DIFF
--- a/cranelift/codegen/src/machinst/lower.rs
+++ b/cranelift/codegen/src/machinst/lower.rs
@@ -452,7 +452,7 @@ impl<'func, I: VCodeInst> Lower<'func, I> {
             block_end_colors[bb] = InstColor::new(cur_color);
         }
 
-        let value_ir_uses = Self::compute_use_states(f, sret_param);
+        let value_ir_uses = compute_use_states(f, sret_param);
 
         Ok(Lower {
             f,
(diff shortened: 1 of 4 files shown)


cranelift/codegen/src/machinst/lower.rs
```diff
@@ -295,15 +295,26 @@ pub struct Lower<'func, I: VCodeInst> {
 /// actually merged" point. Instead, we compute a
 /// transitive-uniqueness. That is what this enum represents.
 ///
+/// There is one final caveat as well to the result of this analysis. Notably
+/// some instructions are considered "root" instructions for this analysis whose
+/// operands don't accurately reflect the use of the results. Root instructions
+/// are currently defined as multi-result instructions. This means that even if
```
cfallin, line 301: I would maybe rephrase as:

"Notably, we define some instructions to be "root" instructions, which means that we assume they will always be codegen'd at the root of a matching tree, and not matched. (This comes with the caveat that we actually enforce this property by making them "opaque" to subtree matching in `get_value_as_source_or_const`.) Because they will always be codegen'd once, they in some sense "reset" multiplicity: these root instructions can be used many times, but because their result(s) are only computed once, they only use theirinputs once.

We currently define all multi-result instructions to be "root" instructions, because it is too complex to reason about matching through them, and they cause too-coarse-grained approximation of multiplicity otherwise: the analysis would have to assume (as it used to!) that they are *always* multiply-used, simply because they have multiple outputs even if those outputs are used only once.

In the future we could define other instructions to be "root" instructions as well, if we make the corresponding change to `get_value_as_source_or_const` as well."

cranelift/codegen/src/machinst/lower.rs
```diff
@@ -1338,6 +1402,7 @@ impl<'func, I: VCodeInst> Lower<'func, I> {
                             + 1
                             == self.cur_scan_entry_color.unwrap().get()
                     {
+                        debug_assert!(!is_value_use_root(self.f, src_inst));
```
cfallin, line 1405: This is asserting that (in effect) we don't have any multiple-result instructions that are side-effecting, I think. That may be true today but I wouldn't want to implicitly bake it in -- could we add `!is_value_use_root(...)` as an additional condition to this if-statement instead?

cfallin (approved):
Looks great; thanks for the careful thinking through this and all the additional documentation!

Some edits to my own taste on one of the comments, and a note on possibly generalizing a case, but nothing major below; LGTM.

cranelift/codegen/src/machinst/lower.rs
alexcrichton: I mostly wanted to make sure that the preexisting `self.num_outputs(src_inst) == 1` on this case didn't become accidentally load-bearing so I put this here to ensure that if `is_value_use_root` changed that it would alert that this needs to be changed as well. That being said I like your suggestion as well, so I've done a bit of refactoring to lift `is_value_use_root(...)` checking to the top of the conditional to handle both pure/impure cases at the same time.

cfallin (approved):

cranelift/codegen/src/machinst/lower.rs
cfallin: Nice!

PR #9141, opened 2024-08-19
Implement the extended-const proposal

alexcrichton:
This commit builds on the support added in #8450 to extend our simple interpreter with support for the `extended-const` proposal to WebAssembly. This is required when updating the spec-test-submodule since `extended-const` was merged into the mainline specification and some proposals are starting to rebase on that.


fitzgen (approved):
Nice!

Issue #6260, opened 2023-04-21
cranelift: Better heuristics for instruction scheduling


jameysharp:
#### Feature

When any optimization level is enabled, Cranelift runs e-graph equality saturation, which removes all pure instructions from the function layout and leaves only a data-flow graph representation. We then have to put those instructions back into a linear order before we can emit machine code.

Currently, we pick an arbitrary topo-sort of the instructions. Instead, I propose implementing list scheduling and, when there are multiple instructions whose operands are all ready, using a heuristic to decide between them.

#### Benefit

See #6159 for one in-depth analysis of a workload where our instruction scheduling has a measurable impact on performance.

#### Implementation

I suggest a combination of two heuristics, based on the results in this paper:

Shobaki, G., Sakka, L., Abu Rmaileh, N. E., and Al-Hamash, H. (2015) Experimental evaluation of various register-pressure-reduction heuristics. Softw. Pract. Exper., 45: 1497– 1517. doi: [10.1002/spe.2297](https://doi.org/10.1002/spe.2297)

The first heuristic is called "Last Use Count" (LUC). It counts how many live-ranges will end if we place this instruction now. Put another way, for each operand, check whether there are any other instructions that will use the same value but haven't been scheduled yet. If there aren't any, then this instruction is the last user of that value. Count how many operands that applies to. We want to choose an instruction with maximum LUC at each point, because that tends to reduce register pressure.

LUC should have frequent ties. Since most instructions have at most two operands, they can end zero, one, or two live ranges. So we need a tie-breaker.

The tie-breaker heuristic is "Critical Path" (CP) length. It counts the maximum length of any data-dependent chain of instructions that must execute after this instruction before the end of the block. By itself, CP optimizes for instruction-level parallelism (ILP) by starting long chains of operations as soon as possible, but it tends to increase register pressure. However, when used as a tie-breaker for LUC, it can reduce register pressure more than LUC alone. We want to choose an instruction with maximum CP.

If both LUC and CP are tied, we need a last-resort tie-breaker. Prefer the instruction with the smallest instruction number. That will tend to keep instructions in the same order that they appeared in the input.

[...]

PR #9150, opened 2024-08-20
Adding an --emit-opt-clif CLI option

dimitris-aspetakis:
The option `--emit-opt-clif` follows the `--emit-clif` implementation from wasmtime's CLI, but the output comes with the egraph optimizations applied. It has been helpful for work on #6260. From discussions in zulip, it seemed more generally desirable.

* I consider some small changes in `crates/cranelift/src/compiler.rs` to be a bit "hacky", probably due to my inexperience with the borrow checker...
* Maybe there are some CLI error messages you would like introduced (e.g. if the paths for `--emit-clif` and `--emit-opt-clif` are the same) — I haven't created any since I'm not even sure about the CLI interface you would prefer.
* The current state does not include support for the wasmtime explorer.

### DIFF
--- a/crates/cranelift/src/builder.rs
+++ b/crates/cranelift/src/builder.rs
@@ -21,6 +21,7 @@ struct Builder {
     linkopts: LinkOptions,
     cache_store: Option<Arc<dyn CacheStore>>,
     clif_dir: Option<path::PathBuf>,
+    opt_clif_dir: Option<path::PathBuf>,
     wmemcheck: bool,
 }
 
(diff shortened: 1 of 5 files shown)


alexcrichton (commented):
Thanks for this! As a possible alternative though, what do you think about not adding a new option but redefining `--emit-clif` as this option instead? That way optimized clif would be emitted by deafult and non-optimized clif could be emitted with `-C opt-level=0`

alexcrichton:
I neded up implementing my idea in https://github.com/bytecodealliance/wasmtime/pull/10011, so I'm going to close this, but thanks regardless for the PR!

PR #9151, opened 2024-08-20
Copy skeleton unit attributes for split DWARF

philipc:
The skeleton unit may contain attributes that don't appear in the split unit. In particular, this includes DW_AT_ranges and DW_AT_comp_dir.

Also, set the correct form for the default directory in the line program. Previously, the different form meant that we emitted the directory again when a file used it. Copying the DW_AT_comp_dir attribute is required for use of the default directory to work successfully.


alexcrichton (approved):

fitzgen:
Thanks!

PR #9158, opened 2024-08-20
Update handling of wasm features and `*.wast` testing

alexcrichton:
This commit is an attempt at making it easier to manage the set of variables we have in play with `*.wast` testing and wasm features. The Cranelift and Winch backends support different sets of wasm features at this time and historically Cranelift has also had architecture-specific support for wasm features. This is hoped to help simplify management of all of this in the future in addition to making `*.wast` testing more robust. Notable changes here are:

* A `Config` no longer tracks a `WasmFeatures` but instead only tracks explicitly disabled/enabled features. The final `WasmFeatures` is then computed given the result of compiler configuration.
* Default-enabled features now start from nothing instead of `wasmparser`'s defaults to avoid future breakage there.
* Each compiler configuration and/or backend now has a listed set of "this feature may panic" wasm features. These are all disabled by default and if explicitly enabled a `Config::validate` error is returned.
* All `*.wast` tests are now run through both Cranelift and Winch. Tests are now asserted to either fail or pass depending on configuration and the whole test will fail if the result doesn't match the expectation.
* Many `gc` proposal tests which are now passing are flagged as passing now. Failing `winch` tests are now explicitly listed instead of using a litany of patterns.

### DIFF
--- a/winch/codegen/src/codegen/context.rs
+++ b/winch/codegen/src/codegen/context.rs
@@ -73,7 +73,9 @@ impl<'a> CodeGenContext<'a> {
             // All of our supported architectures use the float registers for vector operations.
             V128 => self.reg_for_class(RegClass::Float, masm),
             Ref(rt) => match rt.heap_type {
-                WasmHeapType::Func => self.reg_for_class(RegClass::Int, masm),
+                WasmHeapType::Func | WasmHeapType::Extern => {
+                    self.reg_for_class(RegClass::Int, masm)
+                }
                 ht => unimplemented!("Support for WasmHeapType: {ht}"),
             },
         }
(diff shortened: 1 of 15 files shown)


winch/codegen/src/codegen/context.rs
```diff
@@ -73,7 +73,9 @@ impl<'a> CodeGenContext<'a> {
             // All of our supported architectures use the float registers for vector operations.
             V128 => self.reg_for_class(RegClass::Float, masm),
             Ref(rt) => match rt.heap_type {
-                WasmHeapType::Func => self.reg_for_class(RegClass::Int, masm),
+                WasmHeapType::Func | WasmHeapType::Extern => {
```
alexcrichton, line 76: cc @saulecabrera I added some bare-bones pieces for `Extern` in a few places in this PR to get winch as "supporting" reference-types. Flagging Winch as not supporting reference types ended up being particularly difficult because it ends up disabling nearly all tests one way or another. The goal here was to just not panic on the existence of `externref`, but if you prefer I can try to figure out how to detect this elsewhere and reject it while keeping these panics.

saulecabrera: This looks good to me, thanks!

elliottt (approved):
Awesome!

PR #9160, opened 2024-08-21
Use a smaller default GC heap under MIRI

fitzgen:


alexcrichton:
These tests aren't doing wasm compilation and in theory seem ripe for running in miri, could the store defaults perhaps be changed to make a smaller GC heap or something like that to enable them  to run faster?

fitzgen:
@alexcrichton I made the default GC heap 64KiB under MIRI, and this seems to speed up the slow tests.

alexcrichton (approved):

PR #9164, opened 2024-08-22
Remove reference types (`r32` and `r64`) from Cranelift

fitzgen:


cranelift/filetests/filetests/pcc/succeed/fuzz-float-loads.clif
```diff
@@ -1,86 +0,0 @@
...
-    sig2 = (i64 vmctx, i32 uext, i32 uext, i32 uext) -> i32 uext system_v
-    const0 = 0x02030077ff00ff04030077ff6cff5503
-    stack_limit = gv2
-
-block0(v0: i64, v1: i64):
-    v2 = iconst.i32 0
-    v25 -> v2
-    v45 -> v2
-    v24 -> v25
-    v3 = null.r64
```
cfallin, line 25: It looks like this whole test was deleted because it uses a `null.r64` instruction, but the instruction isn't critical to the test (result of random fuzz generation); could we replace it with an `iconst.i64 0` or somesuch and keep the test?

cfallin (approved):
(Cranelift) reftypes are dead; long life (Wasm) reftypes!

Really happy to see this simplification -- long time coming, thanks for all the work pushing it through.

cranelift/filetests/filetests/pcc/succeed/fuzz-float-loads.clif
fitzgen: Ah, I didn't look closely and assumed it relied on the `r64`; added it back and updated it!

PR #9165, opened 2024-08-22
Make `i31.get_{s.u}` instructions trap on null `i31ref`s

fitzgen:


fitzgen:
Hm, I thought this was independent of https://github.com/bytecodealliance/wasmtime/pull/9166 but I guess we can land that first and then I can rebase this.

alexcrichton (approved):

PR #9171, opened 2024-08-27
winch: Improve result handling for Aarch64

saulecabrera:
This commit introduces missing tests for Aarch64's ABI implementation. Compared to the x64 counterpart, the aarch64 implementation was missing tests for multi-value.

Additionally, this commit fixes an issue with how multi-value returns are handlded in Winch's default calling convention, which only allows 1 register result, independent of the register class.

This commit also refactors the register indexing environment so that it can be easily shared across the existing backends.


fitzgen (approved):
Nice catch!

PR #9173, opened 2024-08-27
Improve support for `Module` trait objects

Spartan2909:
Removes the implicit `Sized` requirement on the implementation for `&mut M where M: Module` and adds an implementation for `Box<dyn Module>`.

### DIFF
--- a/cranelift/module/src/module.rs
+++ b/cranelift/module/src/module.rs
@@ -18,6 +18,7 @@ use cranelift_codegen::{
 };
 use cranelift_control::ControlPlane;
 use std::borrow::{Cow, ToOwned};
+use std::boxed::Box;
 use std::string::String;
 
 /// A module relocation.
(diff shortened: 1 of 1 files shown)


cranelift/module/src/module.rs
```diff
@@ -988,7 +989,112 @@ pub trait Module {
...
+    ) -> ModuleResult<()> {
+        (**self).define_function_bytes(func_id, func, alignment, bytes, relocs)
+    }
+
+    fn define_data(&mut self, data_id: DataId, data: &DataDescription) -> ModuleResult<()> {
+        (**self).define_data(data_id, data)
+    }
+}
+
+impl Module for Box<dyn Module> {
```
alexcrichton, line 1097: Is it worth generalizing this to `impl<M: Module + ?Sized> Module for Box<M>` as well?

Spartan2909: Done

alexcrichton (approved):

PR #9175, opened 2024-08-27
Expand the set of constructors of `bindgen!`-generated bindings

alexcrichton:
This commit expands the set of supported constructors for `bindgen!` generated structures with the goal of bringing back `World::new(&mut store, instance)`. This method was removed previously in a refactoring to add `*Pre` structures but refactoring preexisting code to use `*Pre` isn't always easy, so these extra generated bindings provide a smoother migration path for code from before.


alexcrichton:
A memory lapse had me forget that I sent this PR... I opened a duplicate with https://github.com/bytecodealliance/wasmtime/pull/9177 which has a few more minor fixes so I'm going to close in favor of that.

PR #9176, opened 2024-08-28
Add ISLE rules to generate 128-bit comparisons

alexcrichton:
This commits adds pattern-matches to detect 128-bit comparisons encoded in 64-bit comparisons and promote them to 128-bit to enable backends to apply architecture-specific optimizations to these operations.

### DIFF
--- a/cranelift/codegen/src/opts/selects.isle
+++ b/cranelift/codegen/src/opts/selects.isle
@@ -75,3 +75,14 @@
 (rule (simplify (bor (ty_vec128 ty) (band ty (bnot ty c) y) (band ty x c))) (bitselect ty c x y))
 (rule (simplify (bor (ty_vec128 ty) (band ty y (bnot ty c)) (band ty c x))) (bitselect ty c x y))
 (rule (simplify (bor (ty_vec128 ty) (band ty y (bnot ty c)) (band ty x c))) (bitselect ty c x y))
+
+;; Lift an extend operation outside of a `select` if the extend is happening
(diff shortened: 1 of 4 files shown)


cranelift/codegen/src/opts/selects.isle
```diff
@@ -75,3 +75,10 @@
 (rule (simplify (bor (ty_vec128 ty) (band ty (bnot ty c) y) (band ty x c))) (bitselect ty c x y))
 (rule (simplify (bor (ty_vec128 ty) (band ty y (bnot ty c)) (band ty c x))) (bitselect ty c x y))
 (rule (simplify (bor (ty_vec128 ty) (band ty y (bnot ty c)) (band ty x c))) (bitselect ty c x y))
+
+;; Lift a `uextend` operation outside of a `select` if the extend is happening
+;; on both the consequent and the alternative.
+(rule (simplify (select ty cond
+                           (uextend ty a @ (value_type small))
+                           (uextend ty b @ (value_type small))))
+      (uextend ty (select small cond a b)))
```
fitzgen, line 84: Can we do the same with an `sextend`?

fitzgen (approved):

alexcrichton:
I've enabled `sextend` lifting as well now. I additionally limited this optimization to only happening for i64-to-i128-compares. While it should work for other widths current backends don't implement `iconcat` for two 32-bit halves to a 64-bit integer, so I was getting codegen errors on a module.

PR #9180, opened 2024-08-28
Do not deallocate disabled GC heaps via the instance allocator

fitzgen:
When the reference types proposal is disabled, we don't allocate a GC heap from the instance allocator, we instead create a `DisabledGcHeap`. Rather than returning this to the instance allocator when we're done with it, which won't even know what to do with it since the instance allocator didn't create the GC heap, we simply rely on the `DisabledGcHeap`'s `Drop` implementation.

### DIFF
--- a/crates/wasmtime/src/runtime/store.rs
+++ b/crates/wasmtime/src/runtime/store.rs
@@ -1519,10 +1519,7 @@ impl StoreOpaque {
 
         #[cfg(feature = "gc")]
         fn allocate_gc_store(engine: &Engine) -> Result<GcStore> {
-            let (index, heap) = if engine
-                .features()
-                .contains(wasmparser::WasmFeatures::REFERENCE_TYPES)
-            {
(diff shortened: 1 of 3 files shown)


crates/wasmtime/src/runtime/store.rs
```diff
@@ -1519,10 +1519,7 @@ impl StoreOpaque {
 
         #[cfg(feature = "gc")]
         fn allocate_gc_store(engine: &Engine) -> Result<GcStore> {
-            let (index, heap) = if engine
-                .features()
-                .contains(wasmparser::WasmFeatures::REFERENCE_TYPES)
-            {
+            let (index, heap) = if engine.features().reference_types() {
```
alexcrichton, line 1522: While you're here mind handling https://github.com/bytecodealliance/wasmtime/pull/9162 here as well and renaming `reference_types` in this PR to `gc_types` instead?

alexcrichton (approved):

PR #9182, opened 2024-08-29
Enable clap's `wrap_help` feature by default

alexcrichton:
This is similar to bytecodealliance/wasm-tools#1746 which makes the help text easier to read on various terminal widths.


abrown (approved):
Ah, the vetting...

PR #9191, opened 2024-08-30
Upgrade regalloc2 to 0.9.4

elliottt:
Upgrade to regalloc2 0.9.4, which brings with it some code generation improvements.

PR #9193, opened 2024-08-30
Revert "Upgrade regalloc2 to 0.9.4"

elliottt:
Reverts bytecodealliance/wasmtime#9191

We need to yank regalloc-0.9.4 and bump it to 0.10.0, as a function was removed in 0.9.4.


alexcrichton:
Oh I think this is fine to leave in wasmtime, yanks don't delete the crate it just prevents new dependencies being formed on the crate. In that sense if you yank regalloc2 now then wasmtime will keep building. In that sense this shouldn't be necessary

elliottt:
Ah, I misunderstood what the effect of yanking the crate was. I'll close the revert :+1:

bjorn3:
If anyone tries to depend on the wasmtime version on the main branch as library, they can't if they didn't have a lockfile from before 0.9.4 got yanked and there is no regalloc2 version semver compatible with the 0.9.4 dependency requirement. If a non-yanked 0.9.5 existed, this one would be picked, but alas it doesn't exist and thus cargo will return an error.

alexcrichton:
Ah that's a good point, if that's causing issues for you I can approve/merge a revert too

elliottt:
Should we upload a 0.9.5 that's just a re-release of 0.9.3?

elliottt:
Let's just revert the update and restore the changes when we release regalloc2 0.10.0 with safepoint support removed next week.

alexcrichton (approved):

PR #9196, opened 2024-09-03
Use the CallHook::CallingHost and ReturningFromHost with components

elliottt:
We never implemented calling the CallingHost and ReturningFromHost hooks for component host calls. This PR supports those hooks, and adds some tests.

co-authored-by: Nick Fitzgerald <fitzgen@gmail.com>


crates/wasmtime/src/runtime/component/func/host.rs
```diff
@@ -289,15 +289,40 @@ fn validate_inbounds<T: ComponentType>(memory: &[u8], ptr: &ValRaw) -> Result<us
...
+    let cx = VMComponentContext::from_opaque(cx);
+    let instance = (*cx).instance();
+    let types = (*instance).component_types();
+    let raw_store = (*instance).store();
+    let store = StoreContextMut::from_raw(raw_store);
+
+    let res = store
+        .0
+        .call_hook(CallHook::CallingHost)
+        .and_then(|_| crate::runtime::vm::catch_unwind_and_longjmp(|| func(instance, types, store)))
```
alexcrichton, line 309: Could you expand the `catch_unwind_and_longjmp` to encompass the invocation of the call hooks too?

alexcrichton (commented):
Thanks! Mind also adding some tests for this? There's some previous call-hook-related tests which might be good to copy/augment.

Additionally can you double-check that invoking the malloc function calls the hook as well? And also that we call the hooks on entry to wasm?

tests/all/component_model/call_hook.rs
```diff
@@ -0,0 +1,683 @@
...
+    assert_eq!(s.returns_from_host, 2);
+    assert_eq!(s.calls_into_wasm, 2);
+    assert_eq!(s.returns_from_wasm, 2);
+
+    let (s, e) = run(TRAP_IN_F, false);
+    assert!(format!("{:?}", e.unwrap()).contains("trapping in f"));
+    assert_eq!(s.calls_into_host, 1);
+    assert_eq!(s.returns_from_host, 0);
+    assert_eq!(s.calls_into_wasm, 1);
+    assert_eq!(s.returns_from_wasm, 0);
```
elliottt, line 239: This differs from the core wasm behavior, in that trapping in `f` does not register the return from host or return from wasm hooks. @alexcrichton, is this reasonable, or is there something missing from this PR?

crates/wasmtime/src/runtime/component/func/host.rs
```diff
@@ -289,15 +288,45 @@ fn validate_inbounds<T: ComponentType>(memory: &[u8], ptr: &ValRaw) -> Result<us
...
+    let store = StoreContextMut::from_raw(raw_store);
+
+    let res = crate::runtime::vm::catch_unwind_and_longjmp(|| {
+        if let Err(e) = store.0.call_hook(CallHook::CallingHost) {
+            return Err(e);
+        }
+
+        let res = func(instance, types, store);
+
+        let store = StoreContextMut::<T>::from_raw(raw_store);
```
alexcrichton, line 312: To avoid needing to call this twice could the above use `store.as_context_mut()` as an argument? That should effectively reborrow to allow `store` to be used here again

crates/wasmtime/src/runtime/component/func/host.rs
```diff
@@ -289,15 +288,45 @@ fn validate_inbounds<T: ComponentType>(memory: &[u8], ptr: &ValRaw) -> Result<us
...
+) {
+    let cx = VMComponentContext::from_opaque(cx);
+    let instance = (*cx).instance();
+    let types = (*instance).component_types();
+    let raw_store = (*instance).store();
+    let store = StoreContextMut::from_raw(raw_store);
+
+    let res = crate::runtime::vm::catch_unwind_and_longjmp(|| {
+        if let Err(e) = store.0.call_hook(CallHook::CallingHost) {
+            return Err(e);
```
alexcrichton, line 307: Could this be `call_hook(...)?`? (notably using `?` instead of `if let`)

tests/all/component_model/call_hook.rs
```diff
@@ -0,0 +1,681 @@
...
+    }
+}
+
+#[derive(Debug, PartialEq, Eq)]
+enum Context {
+    Host,
+    Wasm,
+}
+
+struct State {
```
alexcrichton, line 600: Could this perhaps be `use`'d from the other call hook test?

tests/all/component_model/call_hook.rs
```diff
@@ -0,0 +1,681 @@
...
+    assert_eq!(s.calls_into_host, 1);
+    assert_eq!(s.returns_from_host, 1);
+    assert_eq!(s.calls_into_wasm, 1);
+    assert_eq!(s.returns_from_wasm, 1);
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn basic_async_hook() -> Result<()> {
```
alexcrichton, line 294: I think this is the same as `call_wrapped_async_func`?

alexcrichton (commented):
Nice tests 👍 

I think this will be good to go with additional tests for using the `realloc` and `post-return` canonical options since I suspect those aren't covered by call hooks just yet

tests/all/component_model/call_hook.rs
```diff
@@ -70,6 +71,87 @@ fn call_wrapped_func() -> Result<()> {
     Ok(())
 }
 
+// Crate a synchronous Func, call it directly:
+#[test]
+fn call_wrapped_func_with_realloc() -> Result<()> {
+    let wat = format!(
+        r#"
+        (component
+            (import "f" (func $f (param "list" (list u8))))
```
alexcrichton, line 80: Could you shift this to being a result of a list? As-is this doesn't require a realloc, but as a return value it'll require a realloc

tests/all/component_model/call_hook.rs
```diff
@@ -70,6 +71,87 @@ fn call_wrapped_func() -> Result<()> {
...
+                (memory (export "memory") 1)
+
+                (func $export (param i32) (param i32)
+                    (call $f (local.get 0) (local.get 1))
+                )
+
+                (export "export" (func $export))
+
+                (func (export "realloc") (param i32 i32 i32 i32) (result i32)
+                    i32.const 65532)
```
alexcrichton, line 103: Instead of defining this here you can also use the `"realloc"` export of the `$libc` module

alexcrichton (approved):

tests/all/component_model/call_hook.rs
```diff
@@ -0,0 +1,638 @@
...
+
+    export.post_return(&mut store)?;
+
+    // There are no host calls in this example, but the post-return does increment the count of
+    // wasm calls by 1, putting the total number of wasm calls at 2.
+    let s = store.into_data();
+    assert_eq!(s.calls_into_host, 0);
+    assert_eq!(s.returns_from_host, 0);
+    assert_eq!(s.calls_into_wasm, 2);
+    assert_eq!(s.returns_from_wasm, 2);
```
elliottt, line 192: Here are the assertions about `post-return`, you  can see that the wasm calls increment after the `.post_return()` call.

tests/all/component_model/call_hook.rs
```diff
@@ -0,0 +1,635 @@
...
+
+    export.post_return(&mut store)?;
+
+    // There is one host call for the host-side realloc, and then two wasm calls for both the
+    // `list8-to-str` call and the guest realloc call for the list argument.
+    let s = store.into_data();
+    assert_eq!(s.calls_into_host, 1);
+    assert_eq!(s.returns_from_host, 1);
+    assert_eq!(s.calls_into_wasm, 2);
+    assert_eq!(s.returns_from_wasm, 2);
```
elliottt, line 138: Here's the handling of the allocated result, and the assertions about the number of host calls encountered. There is only one host call counted, as only `realloc` is used. However, there are two wasm calls for the `realloc` of the `list<u8>` argument, and the `list8-to-str` call itself.

crates/wasmtime/src/runtime/component/func/options.rs
```diff
@@ -248,7 +248,11 @@ impl<'a, T> LowerContext<'a, T> {
     ) -> Result<usize> {
         let realloc_func_ty = Arc::clone(unsafe { (*self.instance).realloc_func_ty() });
         let realloc_func_ty = realloc_func_ty.downcast_ref::<FuncType>().unwrap();
-        self.options
+
+        self.store.0.call_hook(CallHook::CallingHost)?;
```
alexcrichton, line 252: Here and below I think this is `CallingWasm` and `ReturningFromWasm`?

Also would it be possible to push this further down into the `realloc` call below? Or, I'm not sure if this works, even further into `TypedFunc::unchecked_call`? (or w/e API this bottoms out to)

alexcrichton (approved):

crates/wasmtime/src/runtime/component/func/options.rs
elliottt: It turns out this is already counted as a wasm call, and the host call hooks were just unnecessary and incorrect. I had an incorrect assumption about what the `Options::realloc` function was doing, and assumed it was handling host-side reallocs. Given that those aren't a thing and we're already accounting for wasm hooks, the changes to this module were unnecessary.

PR #9198, opened 2024-09-05
Explicitly require write permissions on release process

alexcrichton:
The default token for this repository is now read-only but this workflow needs writable permissions. Local testing seems to show that this should work so let's try it out here.


alexcrichton:
The scheduled release process [has failed](https://github.com/bytecodealliance/wasmtime/actions/runs/10711346434/job/29699854020) so this is an attempt to fix it.

cfallin (approved):

Issue #8674, opened 2024-05-21
Implement the table64 extension to the memory64 proposal


alexcrichton:
The memory64 proposal was [recently updated](https://github.com/WebAssembly/memory64/issues/51) to include 64-bit tables in addition to 64-bit memories. Support for this extension has been [implemented in wasm-tools](https://github.com/bytecodealliance/wasm-tools/pull/1531) and the `main` branch of Wasmtime is now using this version of wasm-tools. Wasmtime, however, does not yet implement the table64 extension and this is intended to track that.

Support for the table64 extension should be relatively simple in Wasmtime but it'll require a number of refactorings to switch indexing types and such. An (incomplete) list of implementation items to handle for this will be:

* Wasmtime's [representation of table types](https://github.com/bytecodealliance/wasmtime/blob/f99464727b69b3f7bfcdd334e0f1624526ed0b8c/crates/types/src/lib.rs#L1448-L1450) should be updated with 64-bit limits and a `table64` flag
* Methods taking indices in Wasmtime, [such as this](https://github.com/bytecodealliance/wasmtime/blob/f99464727b69b3f7bfcdd334e0f1624526ed0b8c/crates/wasmtime/src/runtime/vm/table.rs#L420), should switch to `u64` instead of `u32`.
* Libcalls involving tables, [like this](https://github.com/bytecodealliance/wasmtime/blob/f99464727b69b3f7bfcdd334e0f1624526ed0b8c/crates/wasmtime/src/runtime/vm/libcalls.rs#L210), should switch to `u64` values.
* Translation of invoking libcalls should switch to using 64-bit values instead of 32-bit values. 32-bit tables will zero-extend indices to 64-bit values.
* Embedder API methods, like [`Table::get`](https://docs.rs/wasmtime/latest/wasmtime/struct.Table.html#method.get), should switch to u64.

More-or-less what I'm thinking is that a few key locations need to switch to `u64` and then whack-a-mole is played with rustc to figure out where else needs to be switched to a `u64`. One point of note I'll say is that in general it's best to avoid `as` casts in this refactoring. Casting with `as` loses information in top bits if casting to a smaller size and that can accidentally cover up cases that should be out of bounds or trap. Instead `u32::try_from` or `usize::try_from` should be used where possible with appropriate handling of the error (e.g. bubbling it up or unwrapping with a comment as to why the unwrap is ok)

ssnover:
I'm up for a game of whack-a-mole if you could assign me.

[...]

PR #9206, opened 2024-09-06
Implement the table64 extension to the memory64 proposal

lwshang:
Implements #8674

### DIFF
--- a/crates/c-api/src/table.rs
+++ b/crates/c-api/src/table.rs
@@ -67,7 +67,7 @@ pub unsafe extern "C" fn wasm_table_get(
     index: wasm_table_size_t,
 ) -> Option<Box<wasm_ref_t>> {
     let table = t.table();
-    let r = table.get(t.ext.store.context_mut(), index)?;
+    let r = table.get(t.ext.store.context_mut(), u64::from(index))?;
     wasm_ref_t::new(r)
 }
 
@@ -79,14 +79,16 @@ pub unsafe extern "C" fn wasm_table_set(
 ) -> bool {
     let table = t.table();
     let val = option_wasm_ref_t_to_ref(r, &table.ty(t.ext.store.context()));
-    table.set(t.ext.store.context_mut(), index, val).is_ok()
+    table
+        .set(t.ext.store.context_mut(), u64::from(index), val)
+        .is_ok()
 }
 
 #[no_mangle]
 pub unsafe extern "C" fn wasm_table_size(t: &wasm_table_t) -> wasm_table_size_t {
     let table = t.table();
     let store = t.ext.store.context();
-    table.size(&store)
+    u32::try_from(table.size(&store)).unwrap()
 }
 
 #[no_mangle]
@@ -97,7 +99,9 @@ pub unsafe extern "C" fn wasm_table_grow(
 ) -> bool {
     let table = t.table();
     let init = option_wasm_ref_t_to_ref(init, &table.ty(t.ext.store.context()));
-    table.grow(t.ext.store.context_mut(), delta, init).is_ok()
+    table
+        .grow(t.ext.store.context_mut(), u64::from(delta), init)
+        .is_ok()
 }
 
 #[no_mangle]
@@ -139,7 +143,7 @@ pub unsafe extern "C" fn wasmtime_table_type(
 pub extern "C" fn wasmtime_table_get(
     store: WasmtimeStoreContextMut<'_>,
     table: &Table,
-    index: u32,
+    index: u64,
     ret: &mut MaybeUninit<wasmtime_val_t>,
 ) -> bool {
     let mut scope = RootScope::new(store);
@@ -156,7 +160,7 @@ pub extern "C" fn wasmtime_table_get(
 pub unsafe extern "C" fn wasmtime_table_set(
     mut store: WasmtimeStoreContextMut<'_>,
     table: &Table,
-    index: u32,
+    index: u64,
     val: &wasmtime_val_t,
 ) -> Option<Box<wasmtime_error_t>> {
     let mut scope = RootScope::new(&mut store);
@@ -170,7 +174,7 @@ pub unsafe extern "C" fn wasmtime_table_set(
 }
 
 #[no_mangle]
-pub extern "C" fn wasmtime_table_size(store: WasmtimeStoreContext<'_>, table: &Table) -> u32 {
+pub extern "C" fn wasmtime_table_size(store: WasmtimeStoreContext<'_>, table: &Table) -> u64 {
     table.size(store)
 }
 
@@ -178,9 +182,9 @@ pub extern "C" fn wasmtime_table_size(store: WasmtimeStoreContext<'_>, table: &T
 pub unsafe extern "C" fn wasmtime_table_grow(
     mut store: WasmtimeStoreContextMut<'_>,
     table: &Table,
-    delta: u32,
+    delta: u64,
     val: &wasmtime_val_t,
-    prev_size: &mut u32,
+    prev_size: &mut u64,
 ) -> Option<Box<wasmtime_error_t>> {
     let mut scope = RootScope::new(&mut store);
     handle_result(
--- a/crates/cranelift/src/func_environ.rs
+++ b/crates/cranelift/src/func_environ.rs
@@ -13,8 +13,9 @@ use cranelift_frontend::FunctionBuilder;
 use cranelift_frontend::Variable;
 use cranelift_wasm::{
     EngineOrModuleTypeIndex, FuncEnvironment as _, FuncIndex, FuncTranslationState, GlobalIndex,
-    GlobalVariable, Heap, HeapData, HeapStyle, MemoryIndex, TableData, TableIndex, TableSize,
-    TargetEnvironment, TypeIndex, WasmFuncType, WasmHeapTopType, WasmHeapType, WasmResult,
+    GlobalVariable, Heap, HeapData, HeapStyle, IndexType, Memory, MemoryIndex, Table, TableData,
+    TableIndex, TableSize, TargetEnvironment, TypeIndex, WasmFuncType, WasmHeapTopType,
+    WasmHeapType, WasmResult,
 };
 use smallvec::SmallVec;
 use std::mem;
@@ -672,27 +673,50 @@ impl<'module_environment> FuncEnvironment<'module_environment> {
         builder.switch_to_block(continuation_block);
     }
 
-    fn memory_index_type(&self, index: MemoryIndex) -> ir::Type {
-        if self.module.memory_plans[index].memory.memory64 {
-            I64
-        } else {
-            I32
+    /// Get the Memory for the given index.
+    fn memory(&self, index: MemoryIndex) -> Memory {
+        self.module.memory_plans[index].memory
+    }
+
+    /// Get the Table for the given index.
+    fn table(&self, index: TableIndex) -> Table {
+        self.module.table_plans[index].table
+    }
+
+    /// Cast the value to I64 and sign extend if necessary.
+    ///
+    /// Returns the value casted to I64.
+    fn cast_index_to_i64(
+        &self,
+        pos: &mut FuncCursor<'_>,
+        val: ir::Value,
+        index_type: IndexType,
+    ) -> ir::Value {
+        match index_type {
+            IndexType::I32 => pos.ins().uextend(I64, val),
+            IndexType::I64 => val,
         }
     }
 
-    /// Convert the target pointer-sized integer `val` that is holding a memory
-    /// length (or the `-1` `memory.grow`-failed sentinel) into the memory's
-    /// index type.
+    /// Convert the target pointer-sized integer `val` into the memory/table's index type.
     ///
-    /// This might involve extending or truncating it depending on the memory's
+    /// For memory, `val` is holding a memory length (or the `-1` `memory.grow`-failed sentinel).
+    /// For table, `val` is holding a table length.
+    ///
+    /// This might involve extending or truncating it depending on the memory/table's
     /// index type and the target's pointer type.
-    fn convert_memory_length_to_index_type(
+    fn convert_pointer_to_index_type(
         &self,
         mut pos: FuncCursor<'_>,
         val: ir::Value,
-        index: MemoryIndex,
+        index_type: IndexType,
+        // When it is a memory and the memory is using single-byte pages,
+        // we need to handle the tuncation differently. See comments below.
+        //
+        // When it is a table, this should be set to false.
+        single_byte_pages: bool,
     ) -> ir::Value {
-        let desired_type = self.memory_index_type(index);
+        let desired_type = index_type_to_ir_type(index_type);
         let pointer_type = self.pointer_type();
         assert_eq!(pos.func.dfg.value_type(val), pointer_type);
 
@@ -704,21 +728,24 @@ impl<'module_environment> FuncEnvironment<'module_environment> {
         } else if pointer_type.bits() > desired_type.bits() {
             pos.ins().ireduce(desired_type, val)
         } else {
-            // We have a 64-bit memory on a 32-bit host -- this combo doesn't
+            // We have a 64-bit memory/table on a 32-bit host -- this combo doesn't
             // really make a whole lot of sense to do from a user perspective
             // but that is neither here nor there. We want to logically do an
             // unsigned extend *except* when we are given the `-1` sentinel,
             // which we must preserve as `-1` in the wider type.
-            match self.module.memory_plans[index].memory.page_size_log2 {
-                16 => {
+            match single_byte_pages {
+                false => {
                     // In the case that we have default page sizes, we can
                     // always sign extend, since valid memory lengths (in pages)
                     // never have their sign bit set, and so if the sign bit is
                     // set then this must be the `-1` sentinel, which we want to
                     // preserve through the extension.
+                    //
+                    // When it comes to table, `single_byte_pages` should have always been set to false.
+                    // Then we simply do a signed extension.
                     pos.ins().sextend(desired_type, val)
                 }
-                0 => {
+                true => {
                     // For single-byte pages, we have to explicitly check for
                     // `-1` and choose whether to do an unsigned extension or
                     // return a larger `-1` because there are valid memory
@@ -728,21 +755,7 @@ impl<'module_environment> FuncEnvironment<'module_environment> {
                     let is_failure = pos.ins().icmp_imm(IntCC::Equal, val, -1);
                     pos.ins().select(is_failure, neg_one, extended)
                 }
-                _ => unreachable!("only page sizes 2**0 and 2**16 are currently valid"),
             }
-        }
-    }
-
-    fn cast_memory_index_to_i64(
-        &self,
-        pos: &mut FuncCursor<'_>,
-        val: ir::Value,
-        index: MemoryIndex,
-    ) -> ir::Value {
-        if self.memory_index_type(index) == I64 {
-            val
-        } else {
-            pos.ins().uextend(I64, val)
         }
     }
 
@@ -784,18 +797,18 @@ impl<'module_environment> FuncEnvironment<'module_environment> {
         };
 
         let table = &self.module.table_plans[index].table;
-        let element_size = if table.wasm_ty.is_vmgcref_type() {
+        let element_size = if table.ref_type.is_vmgcref_type() {
             // For GC-managed references, tables store `Option<VMGcRef>`s.
             ir::types::I32.bytes()
         } else {
-            self.reference_type(table.wasm_ty.heap_type).0.bytes()
+            self.reference_type(table.ref_type.heap_type).0.bytes()
         };
 
         let base_gv = func.create_global_value(ir::GlobalValueData::Load {
             base: ptr,
             offset: Offset32::new(base_offset),
             global_type: pointer_type,
-            flags: if Some(table.minimum) == table.maximum {
+            flags: if Some(table.limits.min) == table.limits.max {
                 // A fixed-size table can't be resized so its base address won't
                 // change.
                 MemFlags::trusted().with_readonly()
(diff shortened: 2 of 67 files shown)


lwshang:
There are still some [test failures](https://github.com/bytecodealliance/wasmtime/actions/runs/10741354477/job/29791554809?pr=9206) related to codegen.
I’m uncertain whether I should retain the existing code generation behavior or if it’s acceptable to simply update the tests by blessing them.

I think it would be beneficial to add unit tests to cover the table operations with a 64-bit table. However, I’m unsure about the best location to include them.

@alexcrichton I’d appreciate your guidance.

alexcrichton:
Blessing is fine to do at any time yeah. That'll show up in the diff of this PR so the changes to the output can be reviewed. If anything looks awry a reviewer can flag it for further investigation.

For tests the basic ones should already be running from the `memory64` proposal. If table64 is implemented and working then `cargo test --test wast` should in theory fail saying that some tests were expected to fail but they actually passed. That would be resolved by updating [these lines](https://github.com/bytecodealliance/wasmtime/blob/459378c7d7f19ceb66093ee2bfeca5a0eacd8fc4/tests/wast.rs#L238-L247).

Another way to add tests is to add `*.wast` files somewhere within `tests/misc_testsuite`. That is also run through `cargo test --test wast`.

Another way to add tests is in `tests/all/*.rs`, probably `tests/all/table.rs`. These are more focused on API tests though rather than wasm tests since that's best done through the `*.wast` frameworks if possible.

And finally one other thing this PR will want to do is to remove [this line](https://github.com/bytecodealliance/wasmtime/blob/459378c7d7f19ceb66093ee2bfeca5a0eacd8fc4/crates/fuzzing/src/generators/config.rs#L407-L410) which will re-enable fuzzing the memory64 proposal. Once that's done you can also run the fuzzers to get some more test coverage.

crates/c-api/src/table.rs
```diff
@@ -67,7 +67,7 @@ pub unsafe extern "C" fn wasm_table_get(
     index: wasm_table_size_t,
 ) -> Option<Box<wasm_ref_t>> {
     let table = t.table();
-    let r = table.get(t.ext.store.context_mut(), index)?;
+    let r = table.get(t.ext.store.context_mut(), index as u64)?;
```
alexcrichton, line 70: Where possible it's generally recommended to use `t::from(val)` instead of `val as t` since the former is guaranteed to be lossless where the latter can accidentally lose bits. Can this be changed to `u64::from(index)`? (here and some places below too)

crates/c-api/src/table.rs
```diff
@@ -170,17 +174,17 @@ pub unsafe extern "C" fn wasmtime_table_set(
...
 
 #[no_mangle]
 pub unsafe extern "C" fn wasmtime_table_grow(
     mut store: WasmtimeStoreContextMut<'_>,
     table: &Table,
-    delta: u32,
+    delta: u64,
     val: &wasmtime_val_t,
-    prev_size: &mut u32,
+    prev_size: &mut u64,
```
alexcrichton, line 187: Can you be sure to reflect these changes in the wasmtime C header files as well? (should be replacing a few `uint32_t` with `uint64_t` here and there)

crates/environ/src/compile/module_environ.rs
```diff
@@ -1194,7 +1194,8 @@ impl ModuleTranslation<'_> {
             // include it in the statically-built array of initial
             // contents.
             let offset = match segment.offset.ops() {
-                &[ConstOp::I32Const(offset)] => offset.unsigned(),
+                &[ConstOp::I32Const(offset)] => offset.unsigned() as u64,
```
alexcrichton, line 1197: Mind doing `u64::from(...)` here?

crates/types/src/lib.rs
```diff
@@ -1527,34 +1527,26 @@ pub struct Table {
...
     where
         F: FnMut(EngineOrModuleTypeIndex) -> Result<(), E>,
     {
-        let Table {
-            wasm_ty,
-            minimum: _,
-            maximum: _,
-        } = self;
-        wasm_ty.trace(func)
+        self.wasm_ty.trace(func)
```
alexcrichton, line 1542: Mind leaving this as it was prior? You'll be adding `table64: _` to the pattern but it helps us remember to consider updating these impls when fields are added or removed from `Table`

crates/wasmtime/src/runtime/externals/table.rs
```diff
@@ -111,7 +111,12 @@ impl Table {
             let table = Table::from_wasmtime_table(wasmtime_export, store);
             let wasmtime_table = table.wasmtime_table(store, iter::empty());
             (*wasmtime_table)
-                .fill(store.gc_store_mut()?, 0, init, ty.minimum())
+                .fill(
+                    store.gc_store_mut()?,
+                    0,
+                    init,
+                    ty.minimum().try_into().unwrap(),
```
alexcrichton, line 118: In general, especially in the runtime, we want to avoid `.unwrap()`. To that end could this failure case be turned into an error? 

This I think comes up if a 32-bit host tries to build a table with >= u32::MAX elements.

crates/wasmtime/src/runtime/externals/table.rs
```diff
@@ -203,14 +208,14 @@ impl Table {
...
-    pub fn set(&self, mut store: impl AsContextMut, index: u32, val: Ref) -> Result<()> {
+    pub fn set(&self, mut store: impl AsContextMut, index: u64, val: Ref) -> Result<()> {
         let store = store.as_context_mut().0;
         let ty = self.ty(&store);
         let val = val.into_table_element(store, ty.element())?;
         let table = self.wasmtime_table(store, iter::empty());
         unsafe {
             (*table)
-                .set(index, val)
+                .set(index.try_into().unwrap(), val)
```
alexcrichton, line 218: Similar to my comment above, could this cast's error be handled to return an error in this case? (it basically means out-of-bounds)

crates/wasmtime/src/runtime/externals/table.rs
```diff
@@ -220,12 +225,12 @@ impl Table {
...
     /// Panics if `store` does not own this table.
-    pub fn size(&self, store: impl AsContext) -> u32 {
+    pub fn size(&self, store: impl AsContext) -> u64 {
         self.internal_size(store.as_context().0)
     }
 
-    pub(crate) fn internal_size(&self, store: &StoreOpaque) -> u32 {
-        unsafe { (*store[self.0].definition).current_elements }
+    pub(crate) fn internal_size(&self, store: &StoreOpaque) -> u64 {
+        unsafe { (*store[self.0].definition).current_elements as u64 }
```
alexcrichton, line 233: Could this be `u64::try_from(...).unwrap()`? Here the `.unwrap()` should be ok because the runtiem should always guarantee that we can fit the number of elements in a 64-bit integer.

crates/wasmtime/src/runtime/externals/table.rs
```diff
@@ -259,7 +264,7 @@ impl Table {
                 Some(size) => {
                     let vm = (*table).vmtable();
                     *store[self.0].definition = vm;
-                    Ok(size)
+                    Ok(size as u64)
```
alexcrichton, line 267: Similar to above could this be `u64::try_from(...).unwrap()`?

crates/wasmtime/src/runtime/vm/instance.rs
```diff
@@ -855,7 +855,7 @@ impl Instance {
                     .and_then(|s| s.get(..len))
                     .ok_or(Trap::TableOutOfBounds)?;
                 table.init_func(
-                    dst,
+                    dst.try_into().unwrap(),
```
alexcrichton, line 858: Similar to some prior comments of mine, this `try_into()` will want to be checked and handled. In the context of these functions it'll want to return `Trap::TableOutOfBounds`.

I'll also clarify that this is mostly for hygiene. We can't easily write tests for this right now since it would require running on a 32-bit platform which Wasmtime doesn't support just yet. In the future though it'll help keep tests clean on a 32-bit platform.

crates/wasmtime/src/runtime/vm/instance.rs
```diff
@@ -1120,13 +1120,13 @@ impl Instance {
                     TableInitialValue::Null { precomputed } => precomputed,
                     TableInitialValue::Expr(_) => unreachable!(),
                 };
-                let func_index = precomputed.get(i as usize).cloned();
+                let func_index = precomputed.get(usize::try_from(i).unwrap()).cloned();
```
alexcrichton, line 1123: Also to clarify this `try_from(...).unwrap()` is ok (and the one below). Panicking here helps catch bugs rather than silently truncating by accident. Thanks for updating this!

crates/wasmtime/src/runtime/vm/table.rs
```diff
@@ -270,17 +270,22 @@ impl Table {
     pub fn new_dynamic(plan: &TablePlan, store: &mut dyn Store) -> Result<Self> {
         Self::limit_new(plan, store)?;
         let TableStyle::CallerChecksSignature { lazy_init } = plan.style;
+        let maximum = if !plan.table.table64 && plan.table.maximum.is_none() {
+            Some(usize::try_from(u32::MAX).unwrap())
+        } else {
+            plan.table.maximum.map(|i| usize::try_from(i).unwrap())
```
alexcrichton, line 276: Here this `.unwrap()` will want to be replaced with `.unwrap_or(usize::MAX)`. That'll mean that on a 32-bit system if you load a table with a maximum size of `u64::MAX` the effective maximum size of the table will end up being `usize::MAX`

crates/wasmtime/src/runtime/vm/table.rs
```diff
@@ -349,7 +354,11 @@ impl Table {
     }
 
     fn limit_new(plan: &TablePlan, store: &mut dyn Store) -> Result<()> {
-        if !store.table_growing(0, plan.table.minimum, plan.table.maximum)? {
+        if !store.table_growing(
+            0,
+            plan.table.minimum.try_into().unwrap(),
+            plan.table.maximum.map(|i| i.try_into().unwrap()),
+        )? {
```
alexcrichton, line 361: Here I might recommend duplicating (in spirit at least) what's happening in `vm/memory.rs`. Namely the `table_growing `callback` is always called no matter what but the min/max passed in are `usize::MAX` if they overflow. Notably this shouldn't `.unwrap()` here but you'll want to try to convert to a `usize`.

After the `table_growing` call is made though you'll convert the min/max into `usize` and return an error if something overflows.

Also like `Memory::limit_new` I'd recomend returning `(usize, Option<usize>)` from this method to avoid re-casting things in the callers and reusing the results of this function.

crates/wasmtime/src/runtime/vm/table.rs
```diff
@@ -465,16 +468,16 @@ impl Table {
     pub fn fill(
         &mut self,
         gc_store: &mut GcStore,
-        dst: u32,
+        dst: u64,
         val: TableElement,
-        len: u32,
+        len: u64,
     ) -> Result<(), Trap> {
         let start = dst as usize;
```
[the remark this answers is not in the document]
alexcrichton: (and `len` below)

crates/wasmtime/src/runtime/vm/table.rs
```diff
@@ -526,17 +529,18 @@ impl Table {
...
-    ) -> Result<Option<u32>, Error> {
+    ) -> Result<Option<usize>, Error> {
         let old_size = self.size();
 
         // Don't try to resize the table if its size isn't changing, just return
         // success.
         if delta == 0 {
             return Ok(Some(old_size));
         }
+        let delta: usize = delta.try_into().unwrap();
```
alexcrichton, line 543: Here this'll want to return an error if this returns `Err` rather than panicking

crates/wasmtime/src/runtime/vm/table.rs
```diff
@@ -605,8 +605,8 @@ impl Table {
     /// Get reference to the specified element.
     ///
     /// Returns `None` if the index is out of bounds.
-    pub fn get(&self, gc_store: &mut GcStore, index: u32) -> Option<TableElement> {
-        let index = usize::try_from(index).ok()?;
+    pub fn get(&self, gc_store: &mut GcStore, index: u64) -> Option<TableElement> {
+        let index: usize = index.try_into().unwrap();
```
alexcrichton, line 609: This'll want to return `None` if this cast fails (since it's out-of-bounds)

tests/all/externals.rs
```diff
@@ -29,7 +29,7 @@ fn bad_tables() {
     let ty = TableType::new(RefType::FUNCREF, 0, Some(1));
     let t = Table::new(&mut store, ty.clone(), Ref::Func(None)).unwrap();
     assert!(t.get(&mut store, 0).is_none());
-    assert!(t.get(&mut store, u32::max_value()).is_none());
+    assert!(t.get(&mut store, u64::MAX).is_none());
```
alexcrichton, line 32: Mind adding both of these as a test? Both growing by `u64::from(u32::MAX)` and `u64::MAX` should both fail

tests/all/cli_tests.rs
```diff
@@ -921,7 +921,7 @@ fn table_growth_failure2() -> Result<()> {
     assert!(!output.status.success());
     let stderr = String::from_utf8_lossy(&output.stderr);
     assert!(
-        stderr.contains("forcing a table growth failure to be a trap"),
+        stderr.contains("forcing trap when growing table to"),
```
alexcrichton, line 924: Do you know why this error changed? Is that perhaps a cast gone wrong somewhere?

crates/cranelift/src/func_environ.rs
```diff
@@ -746,6 +746,60 @@ impl<'module_environment> FuncEnvironment<'module_environment> {
...
+            // always cleared for "valid values". The one case we care about
+            // sextend would be when the return value of `memory.grow` is `-1`,
+            // in which case we want to copy the sign bit.
+            //
+            // This should only come up on 32-bit hosts running wasm64 modules,
+            // which at some point also makes you question various assumptions
+            // made along the way...
+            pos.ins().sextend(desired_type, val)
+        }
+    }
```
alexcrichton, line 788: I think this is probably copy/pasted from a memory-related function, so could the memory-related function be refactored so both tables/memories could use the same underlying logic?

crates/cranelift/src/func_environ.rs
```diff
@@ -1558,12 +1612,13 @@ impl<'module_environment> cranelift_wasm::FuncEnvironment for FuncEnvironment<'m
...
 
+        let delta = self.cast_table_index_to_i64(&mut pos, delta, table_index);
         let table_index_arg = pos.ins().iconst(I32, table_index.as_u32() as i64);
         let call_inst = pos
             .ins()
             .call(grow, &[vmctx, table_index_arg, delta, init_value]);
-
-        Ok(pos.func.dfg.first_result(call_inst))
+        let result = pos.func.dfg.first_result(call_inst);
+        Ok(self.cast_pointer_to_table_index(&mut pos, result, table_index))
```
alexcrichton, line 1621: I think that this might be causing issues because the table.grow builtin is returning `i64` rather than `pointer`, so that may need to be updated to return `pointer` instead?

alexcrichton (commented):
Thank you again so much for spearheading the work on this, it's very much appreciated! This looks like a great start and it looks like you've basically figured out how to touch all the bits here and there. I think you've made some great decisions about when to use `u64` vs `usize` as well, thanks for that!

I mostly left a bunch of comments below about things related to casts. I in general want to make sure we're really careful to minimize issues with the principles of (a) avoiding `as` in general and (b) avoiding unwraps unless we know that panicking is indicative of a fatal bug in the runtime. 

Let me know if I can clarify anything further though!

alexcrichton:
Also as a heads up I was inspired reading over this to send https://github.com/bytecodealliance/wasmtime/pull/9209 which may cause more errors in CI for this if that lands first. My hope though is that'll help reviewing and flagging locations to audit for casts though

crates/wasmtime/src/runtime/vm/table.rs
```diff
@@ -596,17 +618,26 @@ impl Table {
...
         Ok(Some(old_size))
     }
 
     /// Get reference to the specified element.
     ///
     /// Returns `None` if the index is out of bounds.
-    pub fn get(&self, gc_store: &mut GcStore, index: u32) -> Option<TableElement> {
-        let index = usize::try_from(index).ok()?;
+    pub fn get(&self, gc_store: &mut GcStore, index: u64) -> Option<TableElement> {
+        let index = match usize::try_from(index) {
```
alexcrichton, line 637: If you'd like this could also be `let index = usize::try_from(index).ok()?;`

crates/cranelift/src/func_environ.rs
```diff
@@ -1573,9 +1589,12 @@ impl<'module_environment> cranelift_wasm::FuncEnvironment for FuncEnvironment<'m
         index: ir::Value,
     ) -> WasmResult<ir::Value> {
         let plan = &self.module.table_plans[table_index];
+        let table = plan.table;
+        let index_type = table.idx_type;
+        let index = self.cast_index_to_i64(&mut builder.cursor(), index, index_type);
```
alexcrichton, line 1594: I think that this may be roughly where the verifier error is coming from. We only want to cast indices to `i64` when it's going out to a libcall, but otherwise the index will stay in its native type.

crates/cranelift/src/func_environ.rs
```diff
@@ -1627,11 +1646,13 @@ impl<'module_environment> cranelift_wasm::FuncEnvironment for FuncEnvironment<'m
         value: ir::Value,
         index: ir::Value,
     ) -> WasmResult<()> {
-        let pointer_type = self.pointer_type();
         let plan = &self.module.table_plans[table_index];
+        let table = plan.table;
+        let index_type = table.idx_type;
+        let index = self.cast_index_to_i64(&mut builder.cursor(), index, index_type);
```
alexcrichton, line 1652: Like `table_get` above this'll want to avoid casting to i64 since some table sets are handled inline. This may need to propagate the `index_type` into various calls below so they can handle the 32/64 split.

alexcrichton (commented):
No worries I'm happy to help! I've flagged a few locations in Wasm->CLIF translation around `table.get` and `table.set` but the `call_indirect` one didn't jump out to me (but it may still be there). In general though the high-level comment is that we won't want to extend a 32-bit index to a 64-bit index too eagerly and instead only do that when necessary. Various checks already handle the index type being smaller than the host pointer type, but you might need to add a few more checks in a few more places.

alexcrichton:
I poked around a bit at this locally and this fixed the call_indirect case I believe:

```diff
diff --git a/crates/cranelift/src/func_environ.rs b/crates/cranelift/src/func_environ.rs
index 568c8e29c7..d486cf082d 100644
--- a/crates/cranelift/src/func_environ.rs
+++ b/crates/cranelift/src/func_environ.rs
@@ -894,11 +894,13 @@ impl<'module_environment> FuncEnvironment<'module_environment> {
         builder.seal_block(null_block);

         builder.switch_to_block(null_block);
+        let index_type = self.table(table_index).idx_type;
         let table_index = builder.ins().iconst(I32, table_index.index() as i64);
         let lazy_init = self
             .builtin_functions
             .table_get_lazy_init_func_ref(builder.func);
         let vmctx = self.vmctx_val(&mut builder.cursor());
+        let index = self.cast_index_to_i64(&mut builder.cursor(), index, index_type);
         let call_inst = builder.ins().call(lazy_init, &[vmctx, table_index, index]);
         let returned_entry = builder.func.dfg.inst_results(call_inst)[0];
         builder.ins().jump(continuation_block, &[returned_entry]);
```

crates/cranelift/src/func_environ.rs
```diff
@@ -672,27 +673,50 @@ impl<'module_environment> FuncEnvironment<'module_environment> {
...
+    ///
+    /// Returns the value casted to I64.
+    fn cast_index_to_i64(
+        &self,
+        pos: &mut FuncCursor<'_>,
+        val: ir::Value,
+        index_type: IndexType,
+    ) -> ir::Value {
+        match index_type {
+            IndexType::I32 => pos.ins().sextend(I64, val),
```
alexcrichton, line 696: I think sextend may have snuck back in instead of a uextend?

alexcrichton (commented):
Looks great! I think the remaining test failure is just matching error messages. Feel free to update [this block](https://github.com/bytecodealliance/wasmtime/blob/2b29e4590ee7e3089dc4b5886c951e0c02920d91/crates/wast/src/wast.rs#L559-L572) if you'd like.

alexcrichton:
If you're curious I poked a bit at this and https://github.com/bytecodealliance/wasmtime/commit/6fff1193035cad2464871b8d53c29cb2087e3bd5 should fix the remaining `*.wast` tests, but there may still be other test failures. Just some minor things here and there.

lwshang:
@alexcrichton A big thank you for providing the fix that resolved the CI failure! I really appreciate your help.

IIUC the re-enabled spec and fuzzing tests provide fairly good coverage for table64.
As for the bot’s comment regarding `wasmtime:config`, it appears that no action is required.

Do you think the PR is now ready to be merged? If so, would you mind turning it from a draft into “Ready for review”?

alexcrichton:
Sounds good, and happy to help! I think you should have a "Ready for review" button on your end, but I'm happy to hit it too if it's hidden. Would you be ok squashing these commits as well? After that I agree that this looks good to go 👍, thanks so much for your work on this!

lwshang:
> Would you be ok squashing these commits as well?

Should I do anything here? Or if it's just you click "Squash and merge" button?

alexcrichton:
Oh that's something you'll do locally and the main purpose is to edit the message on the commit since the default GitHub would generate would otherwise concatenate all the commits so far. The basic idea is you can run `git rebase -i origin/main` and then that'll allow you to squash commits all into one. If you have trouble with that though let me know

alexcrichton (approved):
Thanks again for your work on this!

PR #9208, opened 2024-09-06
[wasi-http] Return `none` from `future-trailers.get` when trailers absent

dicej:
Previously, we would return `error-code::connection-terminated` in this case, whereas the `wasi-http` spec says we should simply return `none` (i.e. it's not an error if there are no trailers present).

This also adds code to `api_proxy_streaming.rs` to check for trailers after reading the body content.


alexcrichton (approved):

alexcrichton:
Thanks!

PR #9209, opened 2024-09-06
Warn against `clippy::cast_possible_truncation` in Wasmtime

alexcrichton:
This commit explicitly enables the `clippy::cast_possible_truncation` lint in Clippy for just the `wasmtime::runtime` module. This does not enable it for the entire workspace since it's a very noisy lint and in general has a low signal value. For the domain that `wasmtime::runtime` is working in, however, this is a much more useful lint. We in general want to be very careful about casting between `usize`, `u32`, and `u64` and the purpose of this module-targeted lint is to help with just that. I was inspired to do this after reading over #9206 where especially when refactoring code and changing types I think it would be useful to have locations flagged as "truncation may happen here" which previously weren't truncating.

The failure mode for this lint is that panics might be introduced where truncation is explicitly intended. Most of the time though this isn't actually desired so the more practical consequence of this lint is to probably slow down wasmtime ever so slightly and bloat it ever so slightly by having a few more checks in a few places. This is likely best addressed in a more comprehensive manner, however, rather than specifically for just this one case. This problem isn't unique to just casts, but to many other forms of `.unwrap()` for example.

### DIFF
--- a/crates/wasmtime/src/runtime/debug.rs
+++ b/crates/wasmtime/src/runtime/debug.rs
@@ -63,8 +63,11 @@ fn relocate_dwarf_sections(bytes: &mut [u8], code_region: (*const u8, usize)) ->
     }
 
     for (offset, value) in relocations {
-        let (loc, _) = object::from_bytes_mut::<U64Bytes<NE>>(&mut bytes[offset as usize..])
-            .map_err(|()| anyhow!("invalid dwarf relocations"))?;
+        let (loc, _) = offset
+            .try_into()
+            .ok()
+            .and_then(|offset| object::from_bytes_mut::<U64Bytes<NE>>(&mut bytes[offset..]).ok())
+            .ok_or_else(|| anyhow!("invalid dwarf relocations"))?;
         loc.set(NE, value);
     }
     Ok(())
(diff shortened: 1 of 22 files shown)


crates/wasmtime/src/runtime/debug.rs
```diff
@@ -63,8 +63,11 @@ fn relocate_dwarf_sections(bytes: &mut [u8], code_region: (*const u8, usize)) ->
     }
 
     for (offset, value) in relocations {
-        let (loc, _) = object::from_bytes_mut::<U64Bytes<NE>>(&mut bytes[offset as usize..])
-            .map_err(|()| anyhow!("invalid dwarf relocations"))?;
+        let (loc, _) = offset
+            .try_into()
+            .ok()
+            .and_then(|offset| object::from_bytes_mut::<U64Bytes<NE>>(&mut bytes[offset..]).ok())
+            .ok_or_else(|| anyhow!("invalid dwarf relocations"))?;
```
elliottt, line 70: Does this indirect through `Option` because the error types are different between line `67` and `69`? Otherwise, wouldn't a combination of `and_then` and `unwrap_or_else` work here?

elliottt (approved):
Looks great!

crates/wasmtime/src/runtime/debug.rs
alexcrichton: Yeah this had to do with different error types across the two functions and I figured options was the easiest way to go here. This is code is on the older side in Wasmtime and is probably ripe for some improvement.

PR #9214, opened 2024-09-08
riscv64/x390: add *_overflow

ghostway0:
currently a draft and only riscv64

### DIFF
--- a/cranelift/codegen/src/isa/riscv64/inst.isle
+++ b/cranelift/codegen/src/isa/riscv64/inst.isle
@@ -2871,6 +2871,19 @@
 ;; zero-extension for the `Value` input.
 (decl put_value_in_reg_for_icmp (IntCC Value) XReg)
 
+(decl madd_overflow64 (XReg XReg XReg) ValueRegs)
+
+(rule (madd_overflow64 x y z)
+  (let ((one XReg (imm $I8 1))
+        (hi XReg (rv_mulhu x y))
+        (m XReg (rv_mul x y))
+        (of_mul XReg (gen_select_xreg (cmp_eqz hi) (zero_reg) one))
+        (sum XReg (rv_add m z))
+        (of_add XReg (gen_select_xreg (int_compare (IntCC.UnsignedLessThan) sum m) one (zero_reg)))
+        (of XReg (rv_or of_mul of_add)))
+    (value_regs sum of)))
+
+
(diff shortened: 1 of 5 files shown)


ghostway0:
@bjorn3 can you take a look? also, I haven't found umul equivalent (or mul with zext*) in s390x. do you know what are their names?

uweigand:
Hi @ghostway0,   a few comments on the s390x part:

- All the new instruction rules you added seem to provide only a single return, the overflow bit.  However, my understanding is that `smul_overflow` and all the other overflow instructions are defined to have *two* returns, the low-part result and the overflow bit.  I think you'll need to use some form of `with_flags` to construct the pair of results (like x86 and aarch64 already do).
- You're simply re-using the same instructions used for the "normal" operation (add/sub/mul) also for the overflow operation.  That is correct for 32-bit and 64-bit operations, but *not* for 8-bit and 16-bit operations.  The reason is that the 390x ISA does not actually have any 8-bit or 16-bit arithmetic instructions, so we simply use the 32-bit version also for 8-bit and 16-bit operations.  That provides the correct (low-part) result, but any overflow indication would be incorrect.
- There is no unsigned-multiply instruction with overflow indication on our platform.  What other compilers do is to use the 32x32->64 or 64x64->128 bit wide multiply instruction, and check whether the high-part of the output is zero.

afonso360 (commented):
👋 Hey,

I don't know if this is ready for review yet, but It's a great start!

A few comments for the RISC-V part. I didn't check the lowerings in a lot of detail, mostly just spotting a few things that could be shorter.

Thanks for working on this!

Edit: I also ran the fuzzer (with these [changes](https://github.com/afonso360/wasmtime/commit/b6b15dac87c4854b8ab7c75bf887aeb78ed9b4f4)) and it pointed out this testcase.

<details>

<summary>Fuzzer testcase</summary>

Testcase:
```
test interpret
test run
target riscv64gc 
target x86_64

function %a(i8) -> i8 {
block0(v0: i8):
    v1, v2 = smul_overflow v0, v0
    return v2
}
; run: %a(-15) == 1
```

Result:
```
 ERROR cranelift_filetests::concurrent > FAIL: run
FAIL ./test.clif: run

Caused by:
    Failed test: run: %a(-15) == 1, actual: 0
1 tests
Error: 1 failure
```
</summary>

PR #9217, opened 2024-09-09
Use `tracing::Instrument` in generated bindings to avoid holding spans across await points

### LINKED ISSUE #9210: Tracing spans in code generated by `wasmtime-wit-bindgen` are entered incorrectly for async functions
**TLDR:** The generated code creates a span and enters it by calling `span.enter()`. This works fine for sync functions but for async functions it results in the span still being entered even if the inner function has yielded. The end result is that the trace span will be shown on other unrelated tokio tasks that happen to run on the same thread at the same time.

Now, in more detail. If we have a `component::bindgen!` call like this
```rust
wasmtime::component::bindgen!({
    world: "imports",
    inline: "
package test:example@1.0.0;

interface test {
  my-func: func();
}

world imports {
  import test;
}
",

    tracing: true,
    async: true,
    include_generated_code_from_file: true
});
```

Then the generated code in `add_to_linker_get_host` will look something like this
```rust
inst.func_wrap_async(
    "my-func",
    move |mut caller: wasmtime::StoreContextMut<'_, T>, (): ()| {
        wasmtime::component::__internal::Box::new(async move {
            let span = tracing::span!(
                tracing::Level::TRACE,
                "wit-bindgen import",
                module = "test",
                function = "my-func",
            );
            let _enter = span.enter();
            tracing::event!(tracing::Level::TRACE, "call");
            let host = &mut host_getter(caller.data_mut());
            let r = Host::my_func(host).await;
            tracing::event!(
                tracing::Level::TRACE,
                result = tracing::field::debug(&r),
                "return"
            );
            Ok(r)
        })
    },
)?;
```

Note that it holds the `_enter` guard across an await point. This is wrong, since the span remains entered even when the resulting future is suspended.

The correct thing to do here is to use tracing's [`Instrument`](https://docs.rs/tracing/latest/tracing/trait.Instrument.html) trait which takes care of entering/exiting the span when the future is polled.

swlynch99:
This PR fixes #9210.

Previously, if we had a WIT file with a function

```wit
async-fn: func();
```

then the generated code in `add_to_linker_get_host` would look like this (if tracing is enabled and the function is async)

```rust
inst.func_wrap_async(
    "my-func",
    move |mut caller: wasmtime::StoreContextMut<'_, T>, (): ()| {
        wasmtime::component::__internal::Box::new(async move {
            let span = tracing::span!(
                tracing::Level::TRACE,
                "wit-bindgen import",
                module = "test",
                function = "my-func",
            );
            let _enter = span.enter();
            tracing::event!(tracing::Level::TRACE, "call");
            let host = &mut host_getter(caller.data_mut());
            let r = Host::my_func(host).await;
            tracing::event!(
                tracing::Level::TRACE,
                result = tracing::field::debug(&r),
                "return"
            );
            Ok(r)
        })
    },
)?;
```

This keeps the tracing span active, even when the resulting future is suspended. The end result is that other unrelated tokio tasks running on the same thread in the meantime will be shown as executing within the `wit-bindgen import` span.

This commit changes the codegen to instead generate

```rust
inst.func_wrap_async(
    "async-fn",
    move |mut caller: wasmtime::StoreContextMut<'_, T>, (): ()| {
        use tracing::Instrument;

        let span = tracing::span!(
            tracing::Level::TRACE,
            "wit-bindgen import",
            module = "test",
            function = "async-fn",
        );
        wasmtime::component::__internal::Box::new(
            async move {
                tracing::event!(tracing::Level::TRACE, "call");
                let host = &mut host_getter(caller.data_mut());
                let r = Host::async_fn(host).await;
                tracing::event!(
                    tracing::Level::TRACE,
                    result = tracing::field::debug(&r),
                    "return"
                );
                Ok(r)
            }
            .instrument(span),
        )
    },
)?;
```

Here, `tracing::Instrument` takes care of entering the span when the future is polled and exiting it when it is suspended.

I have also manually verified that the codegen for sync functions remains exactly the same.

### DIFF
--- a/crates/component-macro/tests/expanded/char_async.rs
+++ b/crates/component-macro/tests/expanded/char_async.rs
@@ -225,19 +225,23 @@ pub mod foo {
                     move |
                         mut caller: wasmtime::StoreContextMut<'_, T>,
                         (arg0,): (char,)|
-                    wasmtime::component::__internal::Box::new(async move {
-                        let host = &mut host_getter(caller.data_mut());
-                        let r = Host::take_char(host, arg0).await;
-                        Ok(r)
-                    }),
+                    {
+                        wasmtime::component::__internal::Box::new(async move {
+                            let host = &mut host_getter(caller.data_mut());
+                            let r = Host::take_char(host, arg0).await;
+                            Ok(r)
+                        })
+                    },
                 )?;
                 inst.func_wrap_async(
                     "return-char",
-                    move |mut caller: wasmtime::StoreContextMut<'_, T>, (): ()| wasmtime::component::__internal::Box::new(async move {
-                        let host = &mut host_getter(caller.data_mut());
-                        let r = Host::return_char(host).await;
-                        Ok((r,))
-                    }),
+                    move |mut caller: wasmtime::StoreContextMut<'_, T>, (): ()| {
+                        wasmtime::component::__internal::Box::new(async move {
+                            let host = &mut host_getter(caller.data_mut());
+                            let r = Host::return_char(host).await;
+                            Ok((r,))
+                        })
+                    },
                 )?;
                 Ok(())
             }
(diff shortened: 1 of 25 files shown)


alexcrichton (approved):
Thanks for this!

swlynch99:
The macos wasmtime CI job seems to be stalled when running `cargo fetch`? I'm not sure what's up with that.

pchickey:
Sometimes runners just flake, lets try it again

PR #9218, opened 2024-09-10
Winch multi values results

vulc41n:
Hey 👋

This PR fixes how results are returned when using Winch. Cranelift uses x0-x7 and v0-v7 to store the first results, and then stores the remaining ones in the stack from sp to fp. Winch stores the last result in either x0 or v0, with the other results placed on the stack from fp to sp.

Thank you @saulecabrera for helping me figure this out ☺️

### DIFF
--- a/cranelift/codegen/src/isa/aarch64/abi.rs
+++ b/cranelift/codegen/src/isa/aarch64/abi.rs
@@ -281,8 +281,10 @@ impl ABIMachineSpec for AArch64MachineDeps {
 
                 let push_to_reg =
                     if call_conv == isa::CallConv::Winch && args_or_rets == ArgsOrRets::Rets {
+                        // Winch uses the first registry to return the last result
                         i == params.len() - 1
                     } else {
+                        // Use max_per_class_reg_vals & remaining_reg_vals otherwise
                         *next_reg < max_per_class_reg_vals && remaining_reg_vals > 0
                     };
 
@@ -318,7 +320,7 @@ impl ABIMachineSpec for AArch64MachineDeps {
             let size = if is_apple_cc
                 || (call_conv == isa::CallConv::Winch && args_or_rets == ArgsOrRets::Rets)
             {
-                // MacOS aarch64 allows stack slots with
+                // MacOS and Winch aarch64 allows stack slots with
                 // sizes less than 8 bytes. They still need to be
                 // properly aligned on their natural data alignment,
                 // though.
--- a/cranelift/codegen/src/isa/aarch64/abi.rs
+++ b/cranelift/codegen/src/isa/aarch64/abi.rs
@@ -109,6 +109,7 @@ impl ABIMachineSpec for AArch64MachineDeps {
         mut args: ArgsAccumulator,
     ) -> CodegenResult<(u32, Option<usize>)> {
         let is_apple_cc = call_conv.extends_apple_aarch64();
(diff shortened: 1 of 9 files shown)


cranelift/codegen/src/isa/aarch64/abi.rs
```diff
@@ -284,7 +279,16 @@ impl ABIMachineSpec for AArch64MachineDeps {
...
 
-                if *next_reg < max_per_class_reg_vals && remaining_reg_vals > 0 {
+                let push_to_reg =
+                    if call_conv == isa::CallConv::Winch && args_or_rets == ArgsOrRets::Rets {
+                        // Winch uses the first registry to return the last result
+                        i == params.len() - 1
+                    } else {
+                        // Use max_per_class_reg_vals & remaining_reg_vals otherwise
+                        *next_reg < max_per_class_reg_vals && remaining_reg_vals > 0
+                    };
```
fitzgen, line 289: Rather than checking for winch returns every iteration of the loop, can we initialize `remaining_reg_vals` such that this behavior happens automatically? I think that means initializing it to one when we are doing winch returns. That feels cleaner to me than adding edge case conditions to check in the middle of the body of the register assignment loop.

cranelift/codegen/src/isa/aarch64/abi.rs
```diff
@@ -325,9 +331,11 @@ impl ABIMachineSpec for AArch64MachineDeps {
                 std::cmp::max(size, 8)
             };
 
-            // Align the stack slot.
-            debug_assert!(size.is_power_of_two());
-            next_stack = align_to(next_stack, size);
+            if call_conv != isa::CallConv::Winch || args_or_rets != ArgsOrRets::Rets {
```
fitzgen, line 334: Since we are checking this condition so frequently, it may help to define an `is_winch_return` boolean at the start of this procedure that we can reuse throughout.

fitzgen (commented):
Looks good, just a couple nitpicks below. Thanks!

cranelift/codegen/src/isa/aarch64/abi.rs
```diff
@@ -386,6 +394,23 @@ impl ABIMachineSpec for AArch64MachineDeps {
...
+                            let size = i64::from(ty.bytes());
+                            *offset = i64::from(next_stack) - *offset - size;
+                        }
+                    }
+                } else {
+                    unreachable!("Winch cannot handle {arg:?}");
+                }
+            }
+        }
+
```
saulecabrera, line 413: We have this exact implementation in the x64 backend. Maybe it's worth factoring it out into a function shared between both backends?

saulecabrera (commented):
Left one small nit as well.

cranelift/codegen/src/isa/aarch64/abi.rs
vulc41n: If `remaining_reg_vals` is one, it will put the first result in the registry, not the last one. I can add another variable named `skip_reg_vals` which is zero except when processing winch returns, then it is `params.len() - 1`. What do you think ?

fitzgen: Ah I missed that detail. In that case, I think it is fine as-is.

fitzgen (approved):
Thanks!

PR #9223, opened 2024-09-10
Implement new `*.wast` directives in `wasmtime-wast`

alexcrichton:
This is a follow-up from #9219 to add support for a few more directives in `*.wast` tests to notably compile a module and possibly instantiate it multiple times.

This copies the upstream spec test using these new directives and edits it to remove the unsupported exception-handling bits.

### DIFF
--- a/crates/wast/src/wast.rs
+++ b/crates/wast/src/wast.rs
@@ -542,7 +542,8 @@ where
                 module,
                 message,
             } => {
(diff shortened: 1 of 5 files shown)


pchickey:
Looks good, but the component paths aren't exercised in the tests. Is there missing functionality where those can't be exercised yet, or can we add a test for it?

alexcrichton:
Good point! I've added a small test for components too

pchickey (approved):

PR #9224, opened 2024-09-11
Undefine FORTIFY_SOURCE for longjmp

rockwotj:
See the comment in the source file, we can't have FORTIFY_SOURCE enabled for longjmp code, so undefine it.

See the source for longjmp_chk in glibc, this assertion is fired: https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/____longjmp_chk.c;hb=HEAD#l54

Here is a similar commit in glibc tests: https://sourceware.org/git/?p=glibc.git;a=commit;h=46432be2f1d4de962b51ca6b9f80fc37744be9f7

We're only hitting this now because we're properly propagating our C/C++ compiler flags to the C code that Wasmtime's build scripts are running.


alexcrichton (approved):
Thanks for this!

PR #9230, opened 2024-09-12
Make host trap handlers optional

### LINKED ISSUE #6926: Support for not relying on signals for trap handling
Currently, on unix platforms, Wasmtime relies on a few signals to operate, here's a list of the ones I'm aware of:

SIGSEGV: Used for bounds checking when static memory is used. AFAIK using dynamic memory disables the need to handle this signal
SIGFPE: Only used on x86_64 for division by zero.
SIGILL: The implementation of `unreachable` wasm instructions or other explicit traps are implemented by using [`ud2`](https://mudongliang.github.io/x86/html/file_module_x86_id_318.html) on x86_64 and [`udf`](https://developer.arm.com/documentation/dui0801/h/A32-and-T32-Instructions/UDF) on aarch64, this triggers a SIGILL.

[Link to Zulip Conversation about this topic.](https://bytecodealliance.zulipchat.com/#narrow/stream/217126-wasmtime/topic/Option.20to.20disable.20signal.20handling/near/386672498)

#### Benefit

Some environments/frameworks also rely on signals (or block them) which can cause issues with Wasmtime. Wasmtime currently registers signals lazily for the whole process when the first engine is created and other library code etc, could also be registering/unregistering signal handlers and these don't always cleanup properly or yield signal control to wasmtime correctly. This happens for me with the seastar.io framework on Linux, but I'm told Macos has also had it's own set of woes with signal handling.

Being able to disable signal handling would allow wasmtime to be used in these sorts of environments.

An additional positive of this note would be that it allows more testing of Cranelift, as it's my understanding some of the [trap handling code is not run in tests/fuzzing](https://bytecodealliance.zulipchat.com/#narrow/stream/217126-wasmtime/topic/Option.20to.20disable.20signal.20handling/near/386688540).

#### Implementation

I believe nothing would need to change for removing the need to handle SIGSEGV, as for SIGFPE, we'd need to generate code for division by zero checks and explicitly trap.

That leaves SIGILL/`ud2`/`udf`, which seems to be the lion's share of the work here.

This is mostly just copying the notes from @cfallin 😄 

Three things to consider with the code to replace ud2: 

Essentially we will replace our current usages of `ud2` with a `jmp` to an offset of `vmctx` which holds the address of a small assembly function that would read the return address then call into the runtime with the return address as an argument. This return address would then be used to lookup the underlying trap code for that location.  Using a pointer in `vmctx` allows for portability when code is loaded directly from disk.

One thing to look out for when doing this work is to determine if the size blowup from `ud2` (2 bytes) to `jmp offset+reg` (8 bytes) is acceptable or not. Some small analysis of generated assembly should be able to easily determine that however before embarking on this task.

alexcrichton:
This commit introduces a new configuration option
`Config::host_trap_handlers` which enables disabling the reliance on host trap handlers (e.g. signal handlers on Unix) at runtime. This is intended to help increase portability in cases where signal handlers are otherwise difficult. This is achieved by plumbing the translation environment to more locations which now conditionally lowers to calls to a function to raise a trap instead of a trap instruction.

The caveats of this support are:

* This is not yet implemented for Winch
* This requires disabling spectre mitigations
* This does not yet support shared memories since it forces dynamic memories to be used.

These points are all possible to address but it might be best to see how this feature evolves over time. I'll also note that this is not an optimized implementation and likely has a fair bit of overhead. For example the solution in #6926 of a more optimized stub to jump-and-trap is not implemented yet.

Closes #6926

### DIFF
--- a/crates/environ/src/builtin.rs
+++ b/crates/environ/src/builtin.rs
@@ -98,7 +98,7 @@ macro_rules! foreach_builtin_function {
             table_fill_gc_ref(vmctx: vmctx, table: i32, dst: i64, val: reference, len: i64);
 
             // Raises an unconditional trap.
-            trap(vmctx: vmctx, code: i32);
+            trap(vmctx: vmctx, code: u8);
 
             // Implementation of `i{32,64}.trunc_f{32,64}_{u,s}` when host trap
             // handlers are disabled. These will raise a trap if necessary. Note
--- a/crates/environ/src/builtin.rs
+++ b/crates/environ/src/builtin.rs
@@ -96,6 +96,19 @@ macro_rules! foreach_builtin_function {
             // Returns an index for Wasm's `table.fill` instruction for GC references.
             #[cfg(feature = "gc")]
             table_fill_gc_ref(vmctx: vmctx, table: i32, dst: i64, val: reference, len: i64);
+
+            // Raises an unconditional trap.
+            trap(vmctx: vmctx, code: i32);
+
+            // Implementation of `i{32,64}.trunc_f{32,64}_{u,s}` when host trap
+            // handlers are disabled. These will raise a trap if necessary. Note
+            // that f32 inputs are always converted to f64 as the argument. Also
+            // note that the signed-ness of the result is not reflected in the
+            // type here.
+            f64_to_i64(vmctx: vmctx, float: f64) -> i64;
+            f64_to_u64(vmctx: vmctx, float: f64) -> i64;
+            f64_to_i32(vmctx: vmctx, float: f64) -> i32;
+            f64_to_u32(vmctx: vmctx, float: f64) -> i32;
         }
     };
 }
--- a/crates/environ/src/tunables.rs
+++ b/crates/environ/src/tunables.rs
@@ -62,6 +62,10 @@ pub struct Tunables {
 
     /// Whether or not Wasm functions target the winch abi.
     pub winch_callable: bool,
+
+    /// Whether or not the host will be using trap handlers such as signal
+    /// handlers.
+    pub host_trap_handlers: bool,
 }
 
 impl Tunables {
@@ -113,6 +117,7 @@ impl Tunables {
             debug_adapter_modules: false,
             relaxed_simd_deterministic: false,
             winch_callable: false,
+            host_trap_handlers: true,
         }
     }
 
--- a/crates/environ/src/tunables.rs
+++ b/crates/environ/src/tunables.rs
@@ -63,9 +63,9 @@ pub struct Tunables {
     /// Whether or not Wasm functions target the winch abi.
     pub winch_callable: bool,
 
-    /// Whether or not the host will be using trap handlers such as signal
-    /// handlers.
-    pub host_trap_handlers: bool,
+    /// Whether or not the host will be using native signals (e.g. SIGILL,
+    /// SIGSEGV, etc) to implement traps.
+    pub signals_based_traps: bool,
 }
 
 impl Tunables {
@@ -117,7 +117,7 @@ impl Tunables {
             debug_adapter_modules: false,
             relaxed_simd_deterministic: false,
             winch_callable: false,
-            host_trap_handlers: true,
+            signals_based_traps: true,
         }
     }
 
--- a/crates/fuzzing/src/generators/config.rs
+++ b/crates/fuzzing/src/generators/config.rs
@@ -181,7 +181,7 @@ impl Config {
             ))
             .allocation_strategy(self.wasmtime.strategy.to_wasmtime())
             .generate_address_map(self.wasmtime.generate_address_map)
-            .host_trap_handlers(self.wasmtime.host_trap_handlers);
+            .signals_based_traps(self.wasmtime.signals_based_traps);
 
         if !self.module_config.config.simd_enabled {
             cfg.wasm_relaxed_simd(false);
@@ -428,9 +428,9 @@ pub struct WasmtimeConfig {
     /// it's cooperatively time-sliced.
     pub async_config: AsyncConfig,
 
-    /// Whether or not host trap handlers are enabled for this configuration,
+    /// Whether or not host signal handlers are enabled for this configuration,
     /// aka whether signal handlers are supported.
-    host_trap_handlers: bool,
+    signals_based_traps: bool,
 }
 
 impl WasmtimeConfig {
@@ -478,7 +478,7 @@ impl WasmtimeConfig {
             config.reference_types_enabled = false;
 
             // Winch requires host trap handlers to be enabled at this time.
-            self.host_trap_handlers = true;
+            self.signals_based_traps = true;
         }
 
         // If using the pooling allocator, constrain the memory and module configurations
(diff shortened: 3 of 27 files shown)


crates/wasmtime/src/config.rs
```diff
@@ -2203,6 +2220,36 @@ impl Config {
...
+    /// disabled. This means that generated code must have spectre mitigations
+    /// disabled. This is because spectre mitigations rely on faults from
+    /// loading from the null address to implement bounds checks.
+    ///
+    /// This option defaults to `true` meaning that host trap handlers are
+    /// enabled by default.
+    pub fn host_trap_handlers(&mut self, enable: bool) -> &mut Self {
+        self.tunables.host_trap_handlers = Some(enable);
+        self
+    }
```
fitzgen, line 2252: Yeah, expanding on my previous naming bikeshed, I think phrasing/renaming this stuff as `virtual_memory_traps` or maybe `signals_based_traps` will clarify things for users a lot here.

The latter is slightly more precise in that a `ud2` isn't actually related to virtual memory, but opens the can of worms that is "windows has something else that is similar to signals but not actually called signals" I think.

The other thing that might be good to flush out a tiny bit more here is the portability hazard bit and say something about which environments might want to disable these handlers. Even just adding something like "for example, embedded operating systems that do not support virtual memory" in parens after "can be a portability hazard in some environments" is probably good enough.

crates/wasmtime/src/runtime/vm/component/libcalls.rs
```diff
@@ -569,3 +569,7 @@ unsafe fn resource_enter_call(vmctx: *mut VMComponentContext) -> Result<()> {
 unsafe fn resource_exit_call(vmctx: *mut VMComponentContext) -> Result<()> {
     ComponentInstance::from_vmctx(vmctx, |instance| instance.resource_exit_call())
 }
+
+unsafe fn trap(_vmctx: *mut VMComponentContext, code: u32) -> Result<()> {
+    Err(wasmtime_environ::Trap::from_u8(code.try_into().unwrap()).unwrap()).err2anyhow()
```
fitzgen, line 574: Can this take a `code: u8` instead of a `code: u32` if it is just going to `code.try_into().unwrap()` it? Is there a reason for the weird mismatch of bit widths?

crates/wasmtime/src/runtime/vm/libcalls.rs
```diff
@@ -655,6 +657,60 @@ fn update_mem_size(instance: &mut Instance, num_pages: u32) {
     }
 }
 
+fn trap(_instance: &mut Instance, code: u32) -> Result<(), TrapReason> {
+    Err(TrapReason::Wasm(
+        wasmtime_environ::Trap::from_u8(code.try_into().unwrap()).unwrap(),
+    ))
+}
```
fitzgen, line 664: This seems to be identical to the component trapping libcall -- is there a reason we need two copies and can't just use this one everywhere?

crates/wasmtime/src/runtime/vm/libcalls.rs
```diff
@@ -655,6 +657,60 @@ fn update_mem_size(instance: &mut Instance, num_pages: u32) {
...
+fn f64_to_u32(_instance: &mut Instance, val: f64) -> Result<u32, TrapReason> {
+    if val.is_nan() {
+        return Err(TrapReason::Wasm(Trap::BadConversionToInteger));
+    }
+    let val = relocs::truncf64(val);
+    if val <= -1.0 || val >= 4294967296.0 {
+        return Err(TrapReason::Wasm(Trap::IntegerOverflow));
+    }
+    Ok(val as u32)
+}
```
fitzgen, line 712: It would be mildly nice/tidy/future-proof to scope down the `allow`s to just the relevant statements.

tests/disable_host_trap_handlers.rs
fitzgen, line 1: Thanks for writing this test! Could you also add a comment at the top of the file detailing 

* what this test is checking, and
* why it needs to be isolated in its own test binary

?

tests/disable_host_trap_handlers.rs
```diff
@@ -0,0 +1,75 @@
...
+        .downcast::<Trap>()?;
+    assert_eq!(trap, Trap::StackOverflow);
+
+    assert_host_signal_handlers_are_unset();
+
+    Ok(())
+}
+
+fn assert_host_signal_handlers_are_unset() {
+    #[cfg(unix)]
```
fitzgen, line 64: And maybe just a quick one-line comment about why this is unix-only and (I presume) how we don't have a way of checking whether we installed exception handlers on windows or not.

fitzgen (approved):
Very nice, thanks!

Did we also want to run the full spec test suite under this configuration, or do you think the differential fuzzing and spec test fuzzer are good enough?

r=me with a bunch of nitpicks and such below

crates/environ/src/tunables.rs
```diff
@@ -62,6 +62,10 @@ pub struct Tunables {
 
     /// Whether or not Wasm functions target the winch abi.
     pub winch_callable: bool,
+
+    /// Whether or not the host will be using trap handlers such as signal
+    /// handlers.
+    pub host_trap_handlers: bool,
```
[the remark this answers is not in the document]
alexcrichton: Very good points yeah, and I like the `signals_based_traps` name you suggested below, so I've gone with that.

crates/fuzzing/src/generators/config.rs
```diff
@@ -527,6 +456,98 @@ impl WasmtimeConfig {
...
+            // Force this pooling allocator to always be able to accommodate the
+            // module that may be generated.
+            pooling.total_memories = config.max_memories as u32;
+            pooling.total_tables = config.max_tables as u32;
+        }
+
+        // TODO: should fix this
+        if !self.host_trap_handlers {
+            config.threads_enabled = false;
+        }
```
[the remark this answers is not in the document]
alexcrichton: Whoops I forgot to fill this out

crates/wasmtime/src/runtime/vm/component/libcalls.rs
alexcrichton: Nah just more effort to plumb u8 to more locations, but I've now done that.

crates/wasmtime/src/runtime/vm/libcalls.rs
alexcrichton: There's some subtle differences in the ABI like the exact return value and the first parameter between the component/core wasm side of things. The component side of things is a bit unfortunate because it's only needed for a degenerate case of always-trapping adapters, but I'll probably leave the minor duplication for now in the hopes this can be refactored away in the future

crates/environ/src/builtin.rs
```diff
@@ -96,6 +96,19 @@ macro_rules! foreach_builtin_function {
...
             table_fill_gc_ref(vmctx: vmctx, table: i32, dst: i64, val: reference, len: i64);
+
+            // Raises an unconditional trap.
+            trap(vmctx: vmctx, code: i32);
+
+            // Implementation of `i{32,64}.trunc_f{32,64}_{u,s}` when host trap
+            // handlers are disabled. These will raise a trap if necessary. Note
+            // that f32 inputs are always converted to f64 as the argument. Also
+            // note that the signed-ness of the result is not reflected in the
+            // type here.
```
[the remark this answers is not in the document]
alexcrichton: I believe that `f64` can losslessly represent a `f32`, yes, it's just the other way around that's bad. Additionally I think it's always valid to promote, perform an op, then demote. The bad thing is to promote, perform a few ops, then demote. Here though there's no "demotion" and it's just promote-and-convert which I believe in theory should.

The wasm spec tests are actually somewhat comprehensive in the boundary cases here and differential fuzzing is also enabled, so I believe that this is correct. (famous last words)

crates/wasmtime/src/runtime/vm/libcalls.rs
fitzgen: Ah gotcha. SGTM

PR #9234, opened 2024-09-12
[WASI-NN] Add support for a PyTorch backend for wasi-nn

rahulchaphalkar:
This change adds a PyTorch backend for wasi-nn.
[tch crate](https://crates.io/crates/tch) is used for Libtorch bindings. I have added an image classification example to demonstrate its usage, which uses a torchscript model.
This backend is currently gated behind a wasi-nn feature flag `--features pytorch` as due to dynamic linking, a Libtorch v2.4.0 installation on the system (specified by `LIBTORCH=/path/to/libtorch`) is needed for building.


crates/wasi-nn/src/backend/pytorch.rs
```diff
@@ -0,0 +1,174 @@
...
+            .iter()
+            .map(|&dim| dim as i64)
+            .collect::<Vec<_>>();
+        self.1 = TchTensor::from_data_size(&input_tensor.data, &dimensions, kind);
+        Ok(())
+    }
+
+    fn compute(&mut self) -> Result<(), BackendError> {
+        // Use forward method on the compiled module/model after locking the mutex, and pass the input tensor to it
+        self.1 = self.0.lock().unwrap().forward_ts(&[&self.1]).unwrap();
```
abrown, line 97: Storing the output tensor in the same location as the input tensor means that `set_input` followed immediately by `get_output` would return the input tensor... probably not what you want here. It looks like `forward_ts` only returns a single tensor so perhaps just create an `output` field for that and another `input: Vec<Tensor>` for the inputs.

crates/wasi-nn/src/backend/pytorch.rs
```diff
@@ -0,0 +1,174 @@
...
+        // Output index is not used in pytorch models. The forward method to a model returns the output tensor.
+        let numel = self.1.numel();
+        let dimensions = self.1.size();
+        let ty = self.1.kind().try_into()?;
+        let data = vec![0f32; numel];
+        let mut data_u8: Vec<u8> = data
+            .iter()
+            .flat_map(|&x| x.to_le_bytes().to_vec())
+            .collect();
+        self.1.copy_data_u8(&mut data_u8, numel);
```
abrown, line 111: This is incorrect: we need to retrieve the data regardless of the type, so we need to first figure out how many bytes each `Kind` is before constructing the receiving buffer, like:

```suggestion
        let data = vec![0u8; size_of(ty) * numel];
        self.1.copy_data_u8(&mut data, numel);
```

crates/wasi-nn/src/backend/pytorch.rs
```diff
@@ -0,0 +1,174 @@
...
+            data: data_u8,
+        })
+    }
+}
+
+fn map_execution_target_to_string(target: ExecutionTarget) -> Device {
+    match target {
+        ExecutionTarget::Cpu => Device::Cpu,
+        ExecutionTarget::Gpu => {
+            unimplemented!("Pytorch does not yet support GPU execution targets")
```
abrown, line 125:
```suggestion
            unimplemented!("the pytorch backend does not yet support GPU execution targets")
```

abrown: ...since PyTorch backend does indeed support GPU execution.

crates/wasi-nn/src/backend/pytorch.rs
```diff
@@ -0,0 +1,174 @@
...
+        }
+        ExecutionTarget::Tpu => {
+            unimplemented!("Pytorch does not yet support TPU execution targets")
```
abrown, line 128:
```suggestion
            unimplemented!("the pytorch backend does not yet support TPU execution targets")
```

abrown (commented):
This is a good start. The main thing to fix is the handling of the input and output tensors.

abrown:
The `cargo vet` situation is a bit much:

```
    cargo vet diff zstd-safe 5.0.1+zstd.1.5.2 5.0.2+zstd.1.5.2
                                                          gyscos         zstd                        2 files changed, 4 insertions(+), 4 deletions(-)
    cargo vet diff zstd 0.11.1+zstd.1.5.2 0.11.2+zstd.1.5.2
                                                          gyscos         zip                         3 files changed, 5 insertions(+), 5 deletions(-)
    cargo vet diff num-complex 0.4.2 0.4.6                cuviper        ndarray                     6 files changed, 188 insertions(+), 48 deletions(-)
      NOTE: this project trusts Josh Stone (cuviper) - consider cargo vet trust num-complex or cargo vet trust --all cuviper
    cargo vet inspect constant_time_eq 0.1.5              cesarb         zip                         311 lines
    cargo vet diff sha1 0.10.5 0.10.6                     newpavlov      zip                         7 files changed, 302 insertions(+), 20 deletions(-)
    cargo vet inspect rawpointer 0.2.1                    bluss          ndarray and matrixmultiply  559 lines
    cargo vet diff zip 0.6.4 0.6.6                        Plecra         tch and torch-sys           14 files changed, 604 insertions(+), 109 deletions(-)
    cargo vet inspect inout 0.1.3                         newpavlov      cipher                      1112 lines
      NOTE: cargo vet import zcash would eliminate this
    cargo vet inspect pbkdf2 0.9.0                        tarcieri       zip                         1120 lines
    cargo vet inspect bzip2 0.4.4                         alexcrichton   zip                         2094 lines
      NOTE: this project trusts Alex Crichton (alexcrichton) - consider cargo vet trust bzip2 or cargo vet trust --all alexcrichton
    cargo vet inspect safetensors 0.3.3                   Narsil         tch                         2200 lines
    cargo vet inspect cipher 0.4.4                        newpavlov      aes                         2635 lines
      NOTE: cargo vet import zcash would reduce this to a [130](https://github.com/bytecodealliance/wasmtime/actions/runs/10836457564/job/30070281197?pr=9234#step:6:131)0-line diff
    cargo vet inspect password-hash 0.3.2                 tarcieri       pbkdf2                      3139 lines
    cargo vet inspect base64ct 1.6.0                      tarcieri       password-hash               3381 lines
    cargo vet diff half 1.8.2 2.4.1                       starkat99      tch                         19 files changed, 2546 insertions(+), 958 deletions(-)
    cargo vet inspect time 0.1.44                         jhpratt        zip                         3915 lines
    cargo vet inspect aes 0.7.5                           tarcieri       zip                         6822 lines
    cargo vet inspect matrixmultiply 0.3.8                bluss          ndarray                     7934 lines
    cargo vet inspect ndarray 0.15.6                      jturner314     tch                         41996 lines
    cargo vet inspect torch-sys 0.17.0                    LaurentMazare  tch                         52119 lines
    cargo vet inspect bzip2-sys 0.1.11+1.0.8              alexcrichton   bzip2                       264[133](https://github.com/bytecodealliance/wasmtime/actions/runs/10836457564/job/30070281197?pr=9234#step:6:134) lines
      NOTE: this project trusts Alex Crichton (alexcrichton) - consider cargo vet trust bzip2-sys or cargo vet trust --all alexcrichton
    cargo vet inspect tch 0.17.0                          LaurentMazare  wasmtime-wasi-nn            2287297 lines
```

crates/wasi-nn/src/backend/pytorch.rs
```diff
@@ -0,0 +1,174 @@
...
+
+impl BackendFromDir for PytorchBackend {
+    fn load_from_dir(
+        &mut self,
+        path: &Path,
+        target: ExecutionTarget,
+    ) -> Result<Graph, BackendError> {
+        // Load the model from the file path
+        let compiled_module =
+            CModule::load_on_device(path, map_execution_target_to_string(target)).unwrap();
```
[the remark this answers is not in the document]
rahulchaphalkar: Changed to `path.join("model.pt")`

crates/wasi-nn/src/backend/pytorch.rs
```diff
@@ -0,0 +1,174 @@
...
+//!
+use super::{
+    BackendError, BackendExecutionContext, BackendFromDir, BackendGraph, BackendInner, Id,
+};
+use crate::wit::types::{ExecutionTarget, GraphEncoding, Tensor, TensorType};
+use crate::{ExecutionContext, Graph};
+use std::path::Path;
+use std::sync::{Arc, Mutex};
+use tch::CModule;
+use tch::{Device, Kind, TchError, Tensor as TchTensor};
```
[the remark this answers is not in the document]
rahulchaphalkar: Done

crates/wasi-nn/src/backend/pytorch.rs
rahulchaphalkar: Good point. I added `inputs` as vector of tensors. Also changed structs in general to use named fields to make code more readable.

crates/wasi-nn/src/backend/pytorch.rs
rahulchaphalkar: Done, added `kind_to_size()` for conversions.

crates/wasi-nn/src/backend/pytorch.rs
rahulchaphalkar: Done

crates/wasi-nn/src/backend/pytorch.rs
```diff
@@ -0,0 +1,174 @@
...
+    fn set_input(&mut self, _index: Id, input_tensor: &Tensor) -> Result<(), BackendError> {
+        // Input index is not used in pytorch models. The forward method to a model passes the tensor/data to
+        // the appropriate layer of the model.
+        let kind = input_tensor.ty.try_into()?;
+        let dimensions = input_tensor
+            .dimensions
+            .iter()
+            .map(|&dim| dim as i64)
+            .collect::<Vec<_>>();
+        self.1 = TchTensor::from_data_size(&input_tensor.data, &dimensions, kind);
```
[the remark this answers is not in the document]
rahulchaphalkar: This is one of the differences for this backend. The module's `forward` method should handle multiple inputs appropriately if it does support multiple inputs. The vector of input tensors being passed to `forward` should be sufficient, no index or name is needed.

crates/wasi-nn/src/backend/pytorch.rs
```diff
@@ -88,32 +101,36 @@ impl BackendExecutionContext for PytorchExecutionContext {
             .iter()
             .map(|&dim| dim as i64)
             .collect::<Vec<_>>();
-        self.1 = TchTensor::from_data_size(&input_tensor.data, &dimensions, kind);
+        self.inputs.push(TchTensor::from_data_size(
+            &input_tensor.data,
+            &dimensions,
+            kind,
+        ));
```
abrown, line 108: This is almost there: we're still ignoring the index, though. What might need to happen is that we set up inputs as a `Vec<Option<TchTensor>>` filled with `None` and then set the right item based on the index. We are able to retrieve the number of inputs, right?

rahulchaphalkar: Did you mean use the index as the position/index for the vector of tensors? That sounds good. I'll see if there's a way to determine the max number of inputs to the given model.

rahulchaphalkar: Alright, so here's what I've changed -

- `set_input()` now uses the `id` passed to it. If it is `u32`, it assigns the tensor to the vector `inputs[id]`, assigning `None` if there are any empty spots below the given `id`. This will give the user flexibility to set input tensors non sequentially.
- `compute()` will check the vector for any `None` values, and give an error if present, before calling `forward_ts()`.
- There is no reliable way at this time to get max inputs for a model. However, assigning more inputs than available returns an error message detailing the max inputs the model expects at run time, so this is helpful for the user. The expectation, similar to other backends, is for the user to be aware of input size/shape etc.
- If `id` is a string, I've currently permitted only single input to be set, effectively ignoring the index. If more inputs are assigned, or if `u32` and `String` indexes are used together, we give out an error.

Let me know if this looks fine. Thanks.

abrown: Yeah, looking at [727a4aa](https://github.com/bytecodealliance/wasmtime/pull/9234/commits/727a4aa31e54519371377a2796afe73719e745b2), I think that makes more sense. I was kind of hoping that [`CModule::named_parameters`](https://docs.rs/tch/latest/tch/struct.CModule.html#method.named_parameters) would return the list of inputs by name, but, if that's not the case, then let's just return an error for `Id::Name` side of things. In any case, we probably don't need `id_type`: if we get an `Id::Index` we know were to put it in the vector and if we get an `Id::Name` then we either (a) look up its index in `named_parameters` or (b) if that's not possible, return an error.

rahulchaphalkar: I had added the Id:Name case because the wit test cases (`nn_wit_image_classification_pytorch` in this case) need a Name instead of Index to pass - [nn.rs](https://github.com/bytecodealliance/wasmtime/blob/80ed395e185b7092bb0b520b67150b7b984e1148/crates/test-programs/src/nn.rs#L43) and [wasi-nn.wit](https://github.com/rahulchaphalkar/wasmtime/blob/786b073664950ebcf3b4fe801826bc17ad3207e3/crates/wasi-nn/wit/wasi-nn.wit#L121). Although it looks like `set-input` might be going away anyway.

crates/wasi-nn/examples/classification-example-pytorch/fixture/model.pt
abrown, line 1: Can we find a smaller model or download this instead? Not all Wasmtime users probably want to download this file... twice.

crates/wasi-nn/Cargo.toml
```diff
@@ -69,7 +70,9 @@ openvino = ["dep:openvino"]
 onnx = ["dep:ort"]
 # WinML is only available on Windows 10 1809 and later.
 winml = ["dep:windows"]
+# PyTorch is available on all platforms; requires Libtorch to be installed
+pytorch = ["dep:tch"]
 
 [[test]]
 name = "test-programs"
-harness = false
+harness = false
```
abrown, line 78: Can you re-add the new line?

crates/wasi-nn/tests/fixtures/resnet.pt
abrown, line 1: Here's the identical 44.7MB file checked in again.

crates/wasi-nn/Cargo.toml
rahulchaphalkar: Done

crates/wasi-nn/examples/classification-example-pytorch/fixture/model.pt
rahulchaphalkar: I've made the following changes -

- Created a repository to generate jit-compiled libtorch models, and download models from Release for tests and examples. 
https://github.com/rahulchaphalkar/libtorch-models
https://github.com/rahulchaphalkar/libtorch-models/releases/tag/v0.1
- Switched from Resnet18(45MB) to Squeezenet1.1(4.5MB) as the file to download from above repo. 
- This `libtorch-models` repo can issue new releases if we update the pytorch/libtorch version in this backend from current 2.4.0 to ensure model compatibility.

crates/wasi-nn/tests/fixtures/resnet.pt
rahulchaphalkar: Downloading from external repo, same as tests

abrown:
I think [51930fa](https://github.com/bytecodealliance/wasmtime/pull/9234/commits/51930fa31b8053b17e9a856fac4c89a98a96bb80) and [da1f467](https://github.com/bytecodealliance/wasmtime/pull/9234/commits/da1f467d805909464102a1134464134e147be21d) have unintentionally updated some crates. Can you remove those commits?

rahulchaphalkar:
I rolled back 2 commits, but there's an issue with the lock file. I had previously attempted to fix this by deleting my lockfile, rebasing off of latest main, and then doing a `cargo build` to generate any of my lock file changes on top of that.

rahulchaphalkar:
The failing tests fail due to
```
error: failed retrieving file 'mingw-w64-x86_64-headers-git-12.0.0.r329.g8f7b5ce36-1-any.pkg.tar.zst.sig' from mirror.clarkson.edu : Operation too slow. Less than 1 bytes/sec transferred the last 10 seconds
 mingw-w64-x86_64-gcc-14.2.0-1-any downloading...
error: failed retrieving file 'mingw-w64-x86_64-mpfr-4.2.1-2-any.pkg.tar.zst' from mirror.msys2.org : Operation too slow. Less than 1 bytes/sec transferred the last 10 seconds
```
I'm hoping a rerun of CI would help.

rahulchaphalkar:
@abrown can you take a look

abrown (approved):

PR #9235, opened 2024-09-12
Pulley: Add 64-bit variants of compare-and-branch super instructions

fitzgen:
Need these for a follow up PR.


pulley/src/interp.rs
```diff
@@ -718,6 +718,66 @@ impl OpVisitor for InterpreterVisitor<'_> {
         }
     }
 
+    fn br_if_xeq64(&mut self, a: XReg, b: XReg, offset: PcRelOffset) -> Self::Return {
+        let a = self.state[a].get_u64();
+        let b = self.state[b].get_u64();
+        if a == b {
+            self.pc_rel_jump(offset, 7)
```
alexcrichton, line 725: Not necessary for this PR, but I think it'd be good to avoid hardcoding 7 here eventually. This'll be hard to remember to update if we pack the `a/b` regs into a single byte for example (not that that's possible because there's 32-registers but similar-ish other encoding schemes might happen)

alexcrichton (approved):

pulley/src/lib.rs
```diff
@@ -45,6 +45,18 @@ macro_rules! for_each_op {
...
+            /// Branch if `a != `b.
+            br_if_xneq64 = BrIfXneq64 { a: XReg, b: XReg, offset: PcRelOffset };
+            /// Branch if signed `a < b`.
+            br_if_xslt64 = BrIfXslt64 { a: XReg, b: XReg, offset: PcRelOffset };
+            /// Branch if signed `a <= b`.
+            br_if_xslteq64 = BrIfXslteq64 { a: XReg, b: XReg, offset: PcRelOffset };
+            /// Branch if unsigned `a < b`.
+            br_if_xult64 = BrIfXult64 { a: XReg, b: XReg, offset: PcRelOffset };
+            /// Branch if unsigned `a <= b`.
+            br_if_xulteq64 = BrIfXulteq64 { a: XReg, b: XReg, offset: PcRelOffset };
```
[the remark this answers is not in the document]
fitzgen: These names match the 32-bit variants and the comparison-without-branching instructions. We can rename everything in a follow up, if you want, but I won't do it in this PR.

pulley/src/interp.rs
fitzgen: We can certainly define a `const OFFSET_OF_JUMP_IN_COND_BR_INST: usize = 7` constant and use that in place of this, but I don't know how we could get the macro to automatically keep the value up-to-date, so I'm not sure it would actually change things in practice. We'd still need to remember to update the constant definition when we bitpack stuff, for example.

pulley/src/lib.rs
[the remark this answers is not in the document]
alexcrichton: Sounds good 👍

pulley/src/interp.rs
alexcrichton: I forget if this was already tried, but could relocation offsets be done such that a static offset here isn't needed? I forget if it's relative to the start or the end of the instruction but if we flip it then this constant here would go away and it's just the producer, Cranelift, that would get a bit more complicated right?

fitzgen: Yeah, right now the jump is relative from the relative-offset's position, not the instruction's position. We could indeed make it relative to the instruction's position and then tweak emission in Cranelift a bit. That would be kind of nice, one less ALU op at runtime!

PR #9240, opened 2024-09-13
Wasmtime: Allow compiling Wasm modules with Pulley

fitzgen:
This does not yet add support for running Wasm modules compiled with Pulley, but it does allow compiling them via `wasmtime compile`. This is a first step towards integrating Pulley into Wasmtime itself.

This is also enough to get basic `tests/disas` tests working with Pulley.


tests/disas.rs
```diff
@@ -311,17 +314,122 @@ fn assert_output(
...
+        Item = (
+            // Address
+            u64,
+            // is jump?
+            bool,
+            // is return?
+            bool,
+            // disasembly
+            String,
+        ),
```
alexcrichton, line 430: Mind switching this to something like `struct Instr { /* named fields here */ }`?

tests/disas.rs
```diff
@@ -311,17 +314,122 @@ fn assert_output(
...
+            }
+        },
     }
     let actual = actual.trim();
     assert_or_bless_output(path, wat, actual)
 }
 
-fn disas_elf(disas: &capstone::Capstone, result: &mut String, elf: &[u8]) -> Result<()> {
-    use capstone::InsnGroupType::{CS_GRP_JUMP, CS_GRP_RET};
+fn disas_elf<I>(
```
alexcrichton, line 414: Perhaps `disas_insns` to drop the "elf" bit?

crates/wasmtime/src/compile.rs
```diff
@@ -330,13 +335,20 @@ impl<'a> CompileInputs<'a> {
...
         types: &'a ModuleTypesBuilder,
         translation: &'a ModuleTranslation<'a>,
         functions: PrimaryMap<DefinedFuncIndex, FunctionBodyData<'a>>,
     ) -> Self {
-        let mut ret = Self::default();
-        let module_index = StaticModuleIndex::from_u32(0);
+        let mut ret = CompileInputs {
+            need_wasm_to_array_trampolines: !matches!(
+                triple.architecture,
+                target_lexicon::Architecture::Pulley32 | target_lexicon::Architecture::Pulley64
```
alexcrichton, line 346: Do you think it's worth having for example an extension trait on `TargetTriple` for `.is_pulley()` to use here and in wasmtime-environ as well? Not that this set of architectures is going to change over time really

alexcrichton (approved):

crates/wasmtime/src/compile.rs
fitzgen: Good idea, will do in a follow up

PR #9242, opened 2024-09-13
CI: Split no-std and clippy out from monolith checks

### LINKED ISSUE #9236: CI: split up the "monolith checks" CI job
I'm seeing it be the bottleneck for entering the merge queue these days, and it shouldn't be hard to split into multiple jobs that can run in parallel. Don't have time at the moment, so just noting this down for posterity.

fitzgen:
Fixes #9236

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -419,6 +419,52 @@ jobs:
       env:
         GH_TOKEN: ${{ github.token }}
 
+  # Checks for no_std support, ensure that crates can build on a no_std target
+  no_std_checks:
+    needs: determine
+    name: no_std checks
+    runs-on: ubuntu-latest
+    env:
+      CARGO_NDK_VERSION: 2.12.2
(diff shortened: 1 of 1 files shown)


fitzgen:
I think clippy was the thing that put the monolith checks job over the edge, but it makes sense to split out the no-std bits as well, since they are expected to grow in the future.

alexcrichton (approved):
Also needs to update [this block](https://github.com/bytecodealliance/wasmtime/blob/3ba9f16c2fa35920553edac8703d17bd598da15e/.github/workflows/main.yml#L1139-L1164), but otherwise r=me

fitzgen:
> Also needs to update [this block](https://github.com/bytecodealliance/wasmtime/blob/3ba9f16c2fa35920553edac8703d17bd598da15e/.github/workflows/main.yml#L1139-L1164)

gah I always forget that

PR #9248, opened 2024-09-13
x64: Add support for the BMI2 `mulx` instruction

alexcrichton:
This commit adds backend support for the `mulx` instruction in the BMI2 instruction set which supports arbitrary destination registers. This instruction also doesn't read/clobber flags so it can help pipeline instructions in niche situations.

### DIFF
--- a/cranelift/codegen/src/isa/x64/inst.isle
+++ b/cranelift/codegen/src/isa/x64/inst.isle
@@ -99,6 +99,21 @@
             (dst_lo WritableGpr)
             (dst_hi WritableGpr))
 
+       ;; Same as `Mul`, but for the BMI2 `mulx` instruction. This is different
+       ;; where the two `dst_*` registers can be arbitrary registers and it
+       ;; is always unsigned multiplication. Note that this instruction does
+       ;; not modify or read flags.
+       ;;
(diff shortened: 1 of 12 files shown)


cranelift/codegen/src/isa/x64/inst.isle
```diff
@@ -2636,6 +2651,26 @@
             (_ Unit (emit (MInst.Mul size signed src1 src2 dst_lo dst_hi))))
         (value_gprs dst_lo dst_hi)))
 
+;; Special case the `mulx` pattern with the BMI2 instruction set.
+(rule 1 (x64_mul (ty_32_or_64 ty) $false src1 src2)
```
fitzgen, line 2655: Is there no `imulx` instruction for signed multiplication?

fitzgen (approved):

cranelift/codegen/src/isa/x64/inst.isle
alexcrichton: Alas not that I can find, just an unsigned variant. I think it's geared towards bignum libraries where most of the intermediate ops are unsigned ones.

fitzgen: Makes sense, thanks for checking

PR #9251, opened 2024-09-15
Pulley: tail calls

Kmeakin:
Sorry for the delay, I was on holiday.
The proof of concept works, but codegen for `become` is not actually implemented yet and so triggers an ICE:
```
error: internal compiler error: /rustc/9b72238eb813e9d06e9e9d270168512fbffd7ee7/compiler/rustc_codegen_ssa/src/mir/block.rs:1372:17: `TailCall` terminator is not yet supported by `rustc_codegen_ssa`
   --> pulley/src/interp/interp_loop.rs:85:25
    |
85  |                           become (next_handler.fun)($state, $pc);
    |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

So for now `return` is used instead of `become` and we must hope that LLVM will be smart enough to recongize the tail call


alexcrichton:
Thanks for this! I might jump in with a small recommendation on overall structure / CI here -- with a Cargo feature controlling this it unfortunately doesn't play well with our "test with all features enabled" in CI well because it enables the feature when a stable compiler is in use. To fix this I might recommend `#[cfg(pulley_tail_call)]` perhaps? That would require setting `RUSTFLAGS` in CI but that could be added to the "Monolith Checks" perhaps as a single `cargo check` of pulley on nightly? That would also enable this to land despite codegen support not being in upstream rust-lang/rust yet.

fitzgen:
Thanks! I'll take a look either later today or tomorrow.

Mildly surprised about `become` not working anymore, as it was working when I did a few experiments around the time of the RFC. I guess rustc is doing a new ssa thing in the midend and hasn't extended support to `become` yet?

fitzgen:
FWIW, it looks like https://github.com/rust-lang/rust/issues/127520 was filed about this regression, but closed as WONTFIX, so I also cross-posted the bug to the `explicit_tail_calls` tracking issue: https://github.com/rust-lang/rust/issues/112788#issuecomment-2353680606

WaffleLapkin:
@fitzgen I'm not sure how it could have worked for you -- `become` was never fully implemented. (maybe you tried to use it when it was just an alternative syntax for `return`?)

either way the reason why https://github.com/rust-lang/rust/issues/127520 was WONTFIXED is that tails calls are simply not implemented yet, they are an incomplete feature. The error even says as much:
> `TailCall` terminator is not yet supported by `rustc_codegen_ssa`

there is no reason to bring any more awareness to this issue -- it will be fixed naturally once the implementation is complete.

pulley/src/decode.rs
```diff
@@ -655,3 +682,43 @@ macro_rules! define_extended_decoder {
...
 for_each_extended_op!(define_extended_decoder);
+
+fn unwrap_uninhabited<T>(res: Result<T, Uninhabited>) -> T {
+    match res {
+        Ok(ok) => ok,
+    }
+}
+
+#[allow(missing_docs)]
+pub mod operands {
```
fitzgen, line 693: Maybe some docs for the module, even if we are skipping them for the individual functions?

pulley/src/decode.rs
```diff
@@ -655,3 +682,43 @@ macro_rules! define_extended_decoder {
...
+                        $field:ident : $field_ty:ty
+                    ),*
+                } )? ;
+            )*
+        ) => {
+            $(
+                #[allow(unused_variables)]
+                pub fn $snake_name(pc: &mut UnsafeBytecodeStream) -> ($($($field_ty,)*)?) {
+                    ($($(unwrap_uninhabited(<$field_ty>::decode(pc)),)*)?)
+                }
```
fitzgen, line 712: Can this take a generic `T: BytecodeStream` and return a `Result<(...), T::Error>`? No need to foist the unsafety upon everyone. We can make `unwrap_uninhabited` a method on an exported extension trait, if we want to make it generally available to users of the decoder.

```rust
pub trait UnwrapUninhabited<T> {
    fn unwrap_uninhabited(self) -> T;
}

impl<T> UnwrapUninhabited<T> for Result<T, Uninhabited> {
    ...
}

...

pub fn $snake_name<T>(pc: &mut T) -> Result<($($($field_ty,)*)?), T::Error> {
    ...
}
```

pulley/src/interp/interp_loop.rs
```diff
@@ -0,0 +1,656 @@
...
+    }
+
+    fn br_if_xeq32(state: &mut MachineState, pc: &mut UnsafeBytecodeStream, a: XReg, b: XReg, offset: PcRelOffset) {
+        let a = state[a].get_u32();
+        let b = state[b].get_u32();
+        if a == b {
+            pc_rel_jump(pc, offset, 7)
+        }
+        ControlFlow::Continue(())
+    }
```
fitzgen, line 166: Nice! I think this worked out really well, especially how each opcode handler doesn't need to explicitly decode immediates/operands, and everything Just Works

pulley/src/lib.rs
```diff
@@ -1,6 +1,8 @@
 //! The pulley bytecode for fast interpreters.
 
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
+#![cfg_attr(feature = "tail_calls", feature(explicit_tail_calls))]
+#![cfg_attr(feature = "tail_calls", allow(incomplete_features, unstable_features))]
```
fitzgen, line 5: As Alex mentioned, to better integrate with our CI infrastructure (which passes `--all-features` to tests), let's replace all `feature = "tail_calls"` in `cfg(...)`s with `cfg(pulley_tail_calls)` and `cfg(not(pulley_tail_calls))`.

Then we can turn it on and off via the `RUSTFLAGS` env var, for example

```
RUSTFLAGS="--cfg pulley_tail_calls" cargo test --all-features
```

And then by default CI will test the regular-calls version, but we can also [add a step to the "Monolith Checks" CI job](https://github.com/bytecodealliance/wasmtime/blob/e38ffa197cf951ed513896fc63bb8e5c7b23135a/.github/workflows/main.yml#L465-L515) that sets the above `RUSTFLAGS` to check that the tail-calls version compiles.

pulley/src/interp/interp_loop.rs
```diff
@@ -0,0 +1,656 @@
...
+                let ($($field,)*) = crate::decode::operands::$name($pc);
+                match $body {
+                    ControlFlow::Continue(()) => {
+                        // Decode the next handler and return it so that `run`
+                        // can call it.
+                        let next_opcode = Opcode::decode($pc).unwrap();
+                        let next_handler = OPCODE_HANDLER_TABLE[next_opcode as usize];
+
+                        // FIXME: ICE
+                        // become (next_handler.fun)($state, $pc);
```
fitzgen, line 84: And then we can actually use `become` here. I've tested locally and `cargo check` works with `become`, and avoids the ICE.

fitzgen: Although maybe we want another `cfg(pulley_become)` or something so that if people want to rely on LLVM to optimize calls in tail position into tail calls, then they can?

fitzgen (commented):
Looks great! Really happy with how this shaped up.

Feel free to split out subcommits if you want, but it isn't necessary. Not sure they will actually land first at this point, since the final commit doesn't have as much feedback. Probably good to do for future PRs tho, just to keep things nice and easy for reviewers :+1: 

Big thing needed to land this is integrating with our CI infrastructure and replacing the cargo feature, as Alex mentioned.

Thanks!

pulley/src/interp.rs
```diff
@@ -597,9 +596,14 @@ impl MachineState {
...
-    Continue,
+/// The reason the interpreter loop terminated.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+enum Done {
+    /// A `ret` instruction was executed and the call stack was empty. This is
+    /// how the loop normally ends.
     ReturnToHost,
+
+    /// a `trap` instruction was executed.
     Trap,
```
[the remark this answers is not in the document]
Kmeakin: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4536593481e0f790a4f2d31c969b7e19

`ControlFlow<Done, ()>` and `Continuation` are both 1 byte (it seems rustc is smart enough to use the "unused discriminants" from `Done` as niches for `ControlFlow`).

`ControlFlow<Done, OpcodeHandler>` is 2 words. The alternative would be to return `ControlFlow<Done, ()>` and set the next handler via an `&mut OpcodeHandler` parameter, but I expect that might be harder for LLVM to optimize. In the worst case, returning a large aggregate will be lowered to returning via an out-parameter anyway, so it shouldn't be any worse performance wise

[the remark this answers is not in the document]
Kmeakin: We could also squeeze `ControlFlow<Done, OpcodeHandler>` into a single word by using the lower 2 bits as the discriminant tag, since function pointers must be word aligned. I think there was a proposal for rustc to do this automatically but it has not yet been implemented.

pulley/src/interp/interp_loop.rs
Kmeakin: `cargo check` works because it doesn't actually generate LLVM/assembly, but if you try `cargo build` or `cargo test` it will

fitzgen: Yep, that's the point: to be using `become` and still have `cargo check` type check the Rust code that we `cfg` on.

Not sure if it is actually *implemented* yet, but that should check that we aren't relying on any `Drop` implementations running after the tail call (which would preclude it from being an actual tail call instead of a call in tail position).

Kmeakin: So run `cargo check`/`cargo clippy` with `RUSTFLAGS="--cfg pulley_tail_calls"`, but not `cargo test`?

pulley/src/interp.rs
```diff
@@ -624,15 +624,15 @@ impl OpVisitor for InterpreterVisitor<'_> {
             Continuation::ReturnToHost
         } else {
             let return_addr = self.state[XReg::lr].get_ptr();
-            self.pc = unsafe { UnsafeBytecodeStream::new(return_addr) };
+            self.pc = unsafe { UnsafeBytecodeStream::new(NonNull::new_unchecked(return_addr)) };
```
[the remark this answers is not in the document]
Kmeakin: > Also, can this debug_assert!(!return_addr.is_null())?

`NonNull::new_unchecked` already performs the check in debug builds: https://doc.rust-lang.org/1.81.0/src/core/ptr/non_null.rs.html#217

pulley/src/interp/interp_loop.rs
fitzgen: Yeah just add a line something like

```yaml
    # Make sure that the tail-calls version of Pulley checks okay.
    - run: cargo check -p pulley-interpreter
      env:
        RUSTFLAGS: "--cfg pulley_tail_calls"
```

around here: https://github.com/bytecodealliance/wasmtime/blob/e38ffa197cf951ed513896fc63bb8e5c7b23135a/.github/workflows/main.yml#L499

fitzgen:
> Once #9274 lands, is this ready to merge as well?

@Kmeakin, now that #9274 landed, is this ready to rebase and land as well, or are there still more things left to do first?

Kmeakin:
> > Once #9274 lands, is this ready to merge as well?
> 
> @Kmeakin, now that #9274 landed, is this ready to rebase and land as well, or are there still more things left to do first?

I think it is ready to land, I don't have anything else I want to add to it.

fitzgen (approved):
Thanks!

fitzgen:
I pushed a commit moving the new CI check to the nightly tests CI job, since it requires a nightly rustc.

But it looks like there are some failures running pulley tests under miri now as well, taking a quick look.

fitzgen:
Pushed a fix for the miri tests: `Done::Trap` needed to have the trapping PC as a payload so that we could pass that trapping PC to `self.trap(pc)` rather than the first PC passed to `run`.

fitzgen:
Thanks again @Kmeakin!

pulley/src/interp/interp_loop.rs
```diff
@@ -0,0 +1,669 @@
...
+            ControlFlow::Break(done) => return done,
+        }
+    }
+}
+
+/// The extra indirection through a macro is necessary to avoid a compiler error
+/// when compiling without `#![feature(explicit_tail_calls)]` enabled (via
+/// `--cfg pulley_tail_calls`).
+///
+/// It seems rustc first parses the the function, encounters `become` and emits
```
Rudxain, line 37: "the the" mistake (typo?)

fitzgen: @Rudxain good eye! Care to make a PR fixing it?

fitzgen: Nevermind, I see that you already did -- thanks!

PR #9253, opened 2024-09-16
Couple of cleanups to the ABI computation

bjorn3:
This also removes the `WasmtimeSystemV` call conv as it is no longer used anywhere.

### DIFF
--- a/cranelift/codegen/src/isa/s390x/lower/isle.rs
+++ b/cranelift/codegen/src/isa/s390x/lower/isle.rs
@@ -1060,7 +1060,6 @@ impl IsleContext<'_, '_, MInst, S390xBackend> {
 #[inline]
 fn lane_order_for_call_conv(call_conv: CallConv) -> LaneOrder {
     match call_conv {
-        CallConv::WasmtimeSystemV => LaneOrder::LittleEndian,
         CallConv::Tail => LaneOrder::LittleEndian,
         _ => LaneOrder::BigEndian,
     }
--- a/cranelift/codegen/src/isa/aarch64/abi.rs
+++ b/cranelift/codegen/src/isa/aarch64/abi.rs
@@ -109,7 +109,7 @@ impl ABIMachineSpec for AArch64MachineDeps {
         add_ret_area_ptr: bool,
         mut args: ArgsAccumulator,
     ) -> CodegenResult<(u32, Option<usize>)> {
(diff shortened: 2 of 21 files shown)


cranelift/codegen/src/isa/s390x/lower/isle.rs
```diff
@@ -1060,7 +1060,6 @@ impl IsleContext<'_, '_, MInst, S390xBackend> {
 #[inline]
 fn lane_order_for_call_conv(call_conv: CallConv) -> LaneOrder {
     match call_conv {
-        CallConv::WasmtimeSystemV => LaneOrder::LittleEndian,
         CallConv::Tail => LaneOrder::LittleEndian,
```
bjorn3, line 1063: @uweigand should CallConv::Winch be LittleEndian too?

uweigand: We don't support Winch at all on s390x right now, so this probably doesn't matter at this point.  If and when we ever add support for Winch, we can re-think this, based on what the actual use case would be.

alexcrichton (approved):
Thanks for this!

cranelift/codegen/src/isa/s390x/abi.rs
```diff
@@ -327,40 +327,44 @@ impl ABIMachineSpec for S390xMachineDeps {
...
             debug_assert!(intreg as i32 + fltreg as i32 + vecreg as i32 <= 1);
 
-            let (next_reg, candidate, implicit_ref) = if intreg {
+            let (candidate, implicit_ref) = if intreg {
                 let candidate = match args_or_rets {
                     ArgsOrRets::Args => get_intreg_for_arg(call_conv, next_gpr),
                     ArgsOrRets::Rets => get_intreg_for_ret(call_conv, next_gpr),
                 };
-                (&mut next_gpr, candidate, None)
+                next_gpr += 1;
```
uweigand, line 335: Not sure I understand the point of this change.  This now increments next_gpr *always*, even when get_intreg_for_arg returns None, i.e. no GPR is actually used.  This is only correct under the implicit assumption that if get_intreg_for_arg failed once, it will always fail in the future.  That may be true with the current ABI, but not necessarily with any possible future ABIs ...

alexcrichton:
(removed from the merge queue for Ulrich's comment)

PR #9258, opened 2024-09-16
Remove StructArgument support from the arm64, riscv64 and s390x backends

bjorn3:
Unlike x86_64 and some other architectures where large structs are passed at fixed stack offsets, necessitating special Cranelift support, arm64, riscv64 and s390x pass them using regular pointer arguments that can trivially be implemented by frontends.

The current implementation of StructArgument on arm64 and riscv64 uses fixed stack offsets just like on x86_64, which is incorrect as per the ABI documentation. In other words any correct frontend already had to avoid StructArgument on arm64 and riscv64 anyway. For s390x it is correctly implemented but unnecessary and taints even non-s390x specific code.


bjorn3:
All cg_clif tests pass on arm64.

cfallin (approved):
Looks reasonable to me -- thanks for this simplification!

PR #9260, opened 2024-09-16
Stop special-casing `i31ref` accesses in many places

fitzgen:
Consolidate all `i31ref` special casing into the `translate_{read,write}_gc_reference` methods, which already emit more or less inline GC barriers depending on the reference's heap type, so it makes sense to also handle `i31ref` there.


fitzgen:
Thought I landed this on Friday, but I guess not.

alexcrichton (approved):

PR #9262, opened 2024-09-16
wit-bindgen: Don't trace values containing lists by default.

sunfishcode:
`list` values in Wit interfaces can represent things like HTTP bodies which can be very large. To enable tracing without spamming logfiles with all this data, put printing of values containing `list`s behind a separate `verbose_tracing` option.

This is a coarse-grained approach; but it seems like a pretty good default for tracing, and enabling full tracing when one needs it is straightforward.

In the future, we may want to refine the option by implementing the `Valuable` trait and using `tracing::field::valuable`, which could allow us to do things like print non-`list` fields of records that otherwise contain `list`s.

### DIFF
--- a/crates/wit-bindgen/src/lib.rs
+++ b/crates/wit-bindgen/src/lib.rs
@@ -3022,26 +3022,10 @@ fn type_contains_lists(ty: Type, resolve: &Resolve) -> bool {
                 .iter()
                 .any(|case| option_type_contains_lists(case.ty, resolve)),
             TypeDefKind::Type(ty) => type_contains_lists(*ty, resolve),
-            TypeDefKind::Future(ty) => {
-                if let Some(ty) = ty {
-                    if type_contains_lists(*ty, resolve) {
-                        return true;
-                    }
-                }
-                false
-            }
+            TypeDefKind::Future(ty) => option_type_contains_lists(*ty, resolve),
             TypeDefKind::Stream(Stream { element, end }) => {
-                if let Some(element) = element {
-                    if type_contains_lists(*element, resolve) {
-                        return true;
-                    }
-                }
-                if let Some(end) = end {
-                    if type_contains_lists(*end, resolve) {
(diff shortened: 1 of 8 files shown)


crates/wit-bindgen/src/lib.rs
```diff
@@ -125,6 +125,10 @@ pub struct Opts {
     /// Whether or not to emit `tracing` macro calls on function entry/exit.
     pub tracing: bool,
 
+    /// Whether or not `tracing` macro calls should included argument and
+    /// return values which contain dynamically-sized `list` values.
+    pub verbose_tracing: bool,
+
```
alexcrichton, line 131: Mind hooking this up as well to the macro input?

crates/wit-bindgen/src/lib.rs
```diff
@@ -2933,6 +2938,105 @@ impl<'a> RustGenerator<'a> for InterfaceGenerator<'a> {
...
+                .iter()
+                .any(|case| option_type_contains_lists(case.ty, resolve)),
+            TypeDefKind::Type(ty) => type_contains_lists(*ty, resolve),
+            TypeDefKind::Future(ty) => {
+                if let Some(ty) = ty {
+                    if type_contains_lists(*ty, resolve) {
+                        return true;
+                    }
+                }
+                false
```
alexcrichton, line 3009: This might be a good use case for the `option_type_contains_lists` helper? (and perhaps `stream` below?)

alexcrichton (approved):

crates/wit-bindgen/src/lib.rs
sunfishcode: Done!

crates/wit-bindgen/src/lib.rs
sunfishcode: Done!

alexcrichton (approved):

PR #9263, opened 2024-09-17
Use `instrument` in generated `call_` functions

elliottt:
Following on from #9217, generate `call_*` functions that use `tracing::Instrument` instead of `Span::enter`.

To test this change I added an additional set of generated files that corresponds to the combination of `async: true` and `tracing: true`. This way we can verify that the changes in #9217 and this PR remove uses of `Span::enter` from the code generated by the `bindgen!` macro.

The additional exp files add an awful lot to this diff, and I think it would be reasonable to remove them before merging. The meat of the changes is in the middle commit, 4c1a4f0998b8f3ebca9e6873a781be69a9121a82, and the final commit shows the differences to the generated code that the middle commit introduces. I believe this catches all outstanding uses of `enter` for the combination of `async: true` and `tracing: true`.

### DIFF
--- a/crates/wit-bindgen/src/lib.rs
+++ b/crates/wit-bindgen/src/lib.rs
@@ -2782,6 +2782,10 @@ impl<'a> InterfaceGenerator<'a> {
         }
 
         if self.gen.opts.tracing {
+            if is_async {
+                self.src.push_str("use tracing::Instrument;\n");
+            }
+
             let ns = match ns {
                 Some(key) => resolve.name_world_key(key),
                 None => "default".to_string(),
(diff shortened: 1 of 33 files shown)


crates/wit-bindgen/src/lib.rs
```diff
@@ -2832,11 +2843,22 @@ impl<'a> InterfaceGenerator<'a> {
...
+
+        let instrument = if is_async && self.gen.opts.tracing {
+            ".instrument(span)"
+        } else {
+            ""
+        };
         uwriteln!(
             self.src,
-            "callee.post_return{async__}(store.as_context_mut()){await_}?;"
+            "callee.post_return{async__}(store.as_context_mut()){instrument}{await_}?;"
```
elliottt, line 2861: This all feels a bit verbose, but it seemed slightly easier to read than the version where I inlined the conditional into the args to the format string. Happy to add the inlined version if that's preferred.

alexcrichton (approved):

PR #9266, opened 2024-09-17
Implement unwind support for ARM64 Windows

### LINKED ISSUE #4992: Trap handling for Windows ARM64
After #4990 merged, we can build wasmtime for Windows ARM64, and it should work without trap handling.

We now need to add trap handling to make it totally work.

dpaoliello:
Implements unwind support for ARM64 Windows.

This also fixes an issue with the current AArch64 ABI code where it wasn't emitting unwind instructions when adjusting the stack pointer (i.e., doing a stack alloc).

With this change, `cargo test` passes all tests on my Windows ARM64 device.

Fixes #4992


alexcrichton:
At a high level this looks great, thanks for working on this! We'll probably be trusting you on the particulars of the table format here, and we unfortunately also probably won't be able to run this on CI (unless you know of way to run this and/or emulation on CI). That being said once you're able to run tests locally that's probably a good enough sign for now and should be good to land afterwards.

It may also be worth noting that the test failures may not be related to unwinding necessarily, I believe Wasmtime has seen very little testing on Windows ARM64 meaning that there could be little other things here and there to overcome as well.

cranelift/codegen/src/isa/unwind/winarm64.rs
```diff
@@ -0,0 +1,308 @@
...
+            }
+            &UnwindInst::StackAlloc { size } => {
+                last_stackalloc = Some(size);
+                assert!(size % 16 == 0, "Size must be a multiple of 16");
+                const SMALL_STACK_ALLOC_MAX: u32 = (1 << 5) * 16;
+                const MEDIUM_STACK_ALLOC_MAX: u32 = (1 << 11) * 16;
+                const LARGE_STACK_ALLOC_MAX: u32 = (1 << 24) * 16;
+                match size {
+                    0..SMALL_STACK_ALLOC_MAX => {
+                        unwind_codes.push(UnwindCode::SmallStackAlloc { size: size as u16 })
```
cfallin, line 229: At least in new code, we're trying to avoid `as`-casts; even though here it's infallibly correct, could we write `u16::try_from(size).unwrap()` (or `.expect("constant max size for this case exceeds u16")` or somesuch so that auditing for overflows is a little easier?

cranelift/codegen/src/isa/unwind/winarm64.rs
```diff
@@ -0,0 +1,308 @@
...
+    /// PAC sign the LR register.
+    PacSignLr,
+    /// Set the frame-pointer register to the stack-pointer register.
+    SetFp,
+    /// Set the frame-pointer register to the stack-pointer register with an
+    /// offset.
+    AddFp {
+        offset: u16,
+    },
+}
```
fitzgen, line 55: Can the names of each of these individual codes match the name in the link above? Specifically `SmallStackAlloc` is called `alloc_s` there so I'd expect it to be `AllocS` here.

Also, it would be nice to have doc comments for the variants that are missing them. For example, the difference between small vs medium stack alloc isn't immediately clear to the reader because they have the same `u16` data representation.

cfallin (approved):
I made a pass over this code as well and it's generally very high-quality, thanks!

To confirm specifically: does the whole Wasmtime suite (`cargo test` in the root) pass now on Windows/aarch64?

Echoing Alex's thoughts -- if we can test this somehow (WINE on qemu-aarch64? or does Windows/x64 support running aarch64 binaries with emulation somehow? or ...?) then we can add release binaries.

(There's precedent for adding release binaries before that, from the macOS/aarch64 case -- we just got GitHub runners for that a few months ago -- though in that case there was external CI run by a third party and those of us with new Macs also manually checked when able.)

fitzgen (approved):
LGTM modulo one nitpick below

dpaoliello:
> To confirm specifically: does the whole Wasmtime suite (cargo test in the root) pass now on Windows/aarch64?

Correct: `cargo test` in the root passes on an ARM64 Windows device.

> Echoing Alex's thoughts -- if we can test this somehow (WINE on qemu-aarch64? or does Windows/x64 support running aarch64 binaries with emulation somehow? or ...?) then we can add release binaries.

That's a tough one.

I've never tried WINE on AArch64 (emulated or physical), so I can't comment on that.

Windows x64 can't run AArch64 binaries.

GitHub does offer [Windows ARM64 runners](https://github.blog/news-insights/product-news/arm64-on-github-actions-powering-faster-more-efficient-build-systems/) however the only image available is a clean Windows 11 install (No Visual Studio, no Git, etc.). The lack of an image with pre-installed tools is also blocking making aarch64-pc-windows-msvc a tier 1 Rust target, so it's high on my priority list to get the correct team at GitHub/Microsoft to make and support those images.

cranelift/codegen/src/isa/unwind/winarm64.rs
dpaoliello: Done: removes the `as` casts.

cranelift/codegen/src/isa/unwind/winarm64.rs
dpaoliello: Done: Renamed the alloc items and added a comment to explain the max size of each one.

alexcrichton:
If you're interested @dpaoliello the "fully exhaustive test suite" can be run with `./ci/run-tests.sh` locally (and/or copying out the cargo invocation since you probably don't have bash on windows) which basically is `cargo test --workspace` with a few `--exclude` for things that aren't supposed to be tested. That will run all `wasmtime-wasi` tests for example which exercise some "larger modules" than the `cargo test` does at the root of the repo. (but for development most of us I think do just `cargo test` and let CI figure everything else out). I mention this if you want to perform an extra double-check that all the various bits are working, but I suspect that if `cargo test` works this won't turn up much else.

I'm also happy to leave this untested on CI until it's easier to set up CI with tools and such. The likelihood of this regressing is relatively low and we can always ping you @dpaoliello (if you're ok with that) if something looks like it's changing. Otherwise I'm also happy to investigate adding ARM64 binaries for Windows to our CI to get release artifacts as well.

dpaoliello:
> If you're interested @dpaoliello the "fully exhaustive test suite" can be run with `./ci/run-tests.sh` locally (and/or copying out the cargo invocation since you probably don't have bash on windows) which basically is `cargo test --workspace` with a few `--exclude` for things that aren't supposed to be tested. That will run all `wasmtime-wasi` tests for example which exercise some "larger modules" than the `cargo test` does at the root of the repo. (but for development most of us I think do just `cargo test` and let CI figure everything else out). I mention this if you want to perform an extra double-check that all the various bits are working, but I suspect that if `cargo test` works this won't turn up much else.

Sure, I'll give it a go. Otherwise, I was starting to look into getting `rustc_codegen_cranelift` working.

> I'm also happy to leave this untested on CI until it's easier to set up CI with tools and such.

I'll add it to my list of repos to setup CIs once the Windows ARM64 images are available...

> we can always ping you @dpaoliello (if you're ok with that) if something looks like it's changing.

Yes, please!
Feel free to summon me for any Rust `-msvc` target or Windows on ARM issue.

fitzgen (approved):
Thanks so much!

dpaoliello:
> Looks like this is failing on our MSRV of Rust 1.78

Fixed

PR #9273, opened 2024-09-18
Use `split_first_chunk` instead of splitting manually

Kmeakin:
Split off from https://github.com/bytecodealliance/wasmtime/pull/9251


fitzgen (approved):
Thanks!

PR #9274, opened 2024-09-18
Use `NonNull<u8>` for `pc` instead of `*mut u8`

Kmeakin:
Split off from https://github.com/bytecodealliance/wasmtime/pull/9251

### DIFF
--- a/pulley/fuzz/src/interp.rs
+++ b/pulley/fuzz/src/interp.rs
@@ -32,7 +32,7 @@ pub fn interp(ops: Vec<Op>) {
         match vm.call(NonNull::from(&encoded[0]), args, rets.into_iter().copied()) {
             Ok(rets) => assert_eq!(rets.count(), 0),
             Err(pc) => {
-                let pc = pc as usize;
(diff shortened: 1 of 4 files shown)


alexcrichton (approved):

alexcrichton:
We're unfortunately held back on Rust versions by OSS-Fuzz right now (see https://github.com/bytecodealliance/wasmtime/pull/9231 for a bit more info), so it might be best to go ahead and get this working on our current 1.78 MSRV. I think it may just require a few more accessor methods in a few places? (e.g. `.as_ptr().read()` instead of `.read()`)

Kmeakin:
> We're unfortunately held back on Rust versions by OSS-Fuzz right now (see #9231 for a bit more info), so it might be best to go ahead and get this working on our current 1.78 MSRV. I think it may just require a few more accessor methods in a few places? (e.g. `.as_ptr().read()` instead of `.read()`)

Should pass on 1.78 rust with the newest patch

alexcrichton:
Oh oops sorry I think I missed the update

PR #9275, opened 2024-09-18
Implement the `struct.*` GC instructions

fitzgen:
This commit implements the `struct.*` instructions for the GC proposal.  These instructions allow allocating new structs and getting and setting their fields.

The implemented instructions are:

* `struct.new`
* `struct.new_default`
* `struct.get`
* `struct.get_s`
* `struct.get_u`
* `struct.set`

The `struct.new*` instructions are also allowed in constant expressions, but support for that is not yet implemented.

### DIFF
--- a/crates/cranelift/src/func_environ.rs
+++ b/crates/cranelift/src/func_environ.rs
@@ -13,9 +13,9 @@ use cranelift_frontend::FunctionBuilder;
 use cranelift_frontend::Variable;
 use cranelift_wasm::{
     EngineOrModuleTypeIndex, FuncEnvironment as _, FuncIndex, FuncTranslationState, GlobalIndex,
-    GlobalVariable, Heap, HeapData, HeapStyle, IndexType, Memory, MemoryIndex, Table, TableData,
-    TableIndex, TableSize, TargetEnvironment, TypeIndex, WasmFuncType, WasmHeapTopType,
-    WasmHeapType, WasmResult,
+    GlobalVariable, Heap, HeapData, HeapStyle, IndexType, Memory, MemoryIndex, StructFieldsVec,
+    Table, TableData, TableIndex, TableSize, TargetEnvironment, TypeIndex, WasmCompositeType,
+    WasmFuncType, WasmHeapTopType, WasmHeapType, WasmResult, WasmValType,
 };
 use smallvec::SmallVec;
 use std::mem;
@@ -84,10 +84,16 @@ wasmtime_environ::foreach_builtin_function!(declare_function_signatures);
 /// The `FuncEnvironment` implementation for use by the `ModuleEnvironment`.
 pub struct FuncEnvironment<'module_environment> {
     isa: &'module_environment (dyn TargetIsa + 'module_environment),
-    module: &'module_environment Module,
-    types: &'module_environment ModuleTypesBuilder,
+    pub(crate) module: &'module_environment Module,
+    pub(crate) types: &'module_environment ModuleTypesBuilder,
     wasm_func_ty: &'module_environment WasmFuncType,
     sig_ref_to_ty: SecondaryMap<ir::SigRef, Option<&'module_environment WasmFuncType>>,
+
+    #[cfg(feature = "gc")]
+    pub(crate) ty_to_struct_layout: std::collections::HashMap<
+        wasmtime_environ::ModuleInternedTypeIndex,
+        wasmtime_environ::GcStructLayout,
+    >,
 
     #[cfg(feature = "wmemcheck")]
     translation: &'module_environment ModuleTranslation<'module_environment>,
@@ -174,6 +180,9 @@ impl<'module_environment> FuncEnvironment<'module_environment> {
             types,
             wasm_func_ty,
             sig_ref_to_ty: SecondaryMap::default(),
+
+            #[cfg(feature = "gc")]
+            ty_to_struct_layout: std::collections::HashMap::new(),
 
             heaps: PrimaryMap::default(),
             tables: SecondaryMap::default(),
@@ -1839,6 +1848,9 @@ impl<'module_environment> cranelift_wasm::FuncEnvironment for FuncEnvironment<'m
         builder: &mut FunctionBuilder,
         i31ref: ir::Value,
     ) -> WasmResult<ir::Value> {
+        // TODO: If we knew we have a `(ref i31)` here, instead of maybe a `(ref
+        // null i31)`, we could omit the `trapz`. But plumbing that type info
+        // from `wasmparser` and through to here is a bit funky.
         self.trapz(builder, i31ref, ir::TrapCode::NullI31Ref);
         Ok(builder.ins().sshr_imm(i31ref, 1))
     }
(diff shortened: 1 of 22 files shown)


crates/cranelift/src/gc/enabled.rs
```diff
@@ -56,7 +64,309 @@ pub fn gc_ref_table_fill_builtin(
...
+    let field_size = wasmtime_environ::byte_size_of_wasm_ty_in_gc_heap(&field_ty.element_type);
+
+    // TODO: See comment in `translate_struct_get` about the `prepare_gc_ref_access`.
+    let field_addr = func_env.prepare_gc_ref_access(builder, struct_ref, field_offset, field_size);
+
+    match &field_ty.element_type {
+        WasmStorageType::I8 => {
+            builder
+                .ins()
+                .istore8(ir::MemFlags::trusted(), new_val, field_addr, 0);
```
alexcrichton, line 321: Is `::trusted()` the right default here? I would have thought that this would be "can possibly trap" flags. We can stuff in a debug trap code which doesn't cause trap table information to get emitted but it should still hinder optimizations and such in Cranelift (which I think we want at least for now?)

crates/cranelift/src/gc/enabled.rs
```diff
@@ -79,6 +389,15 @@ impl FuncEnvironment<'_> {
...
+    ///
+    /// The given `gc_ref` must be a non-null, non-i31 GC reference.
+    ///
+    /// Returns the raw pointer to `gc_ref[offset]` -- not a raw pointer to the
+    /// GC object itself (unless `offset == 0`). This raw pointer may be used to
+    /// read or write up to `size` bytes. Do NOT attempt accesses larger than
+    /// `size` bytes; that may lead to unchecked out-of-bounds accesses.
+    ///
+    /// This method is collector-agnostic.
     fn prepare_gc_ref_access(
```
alexcrichton, line 401: This method may want to be refactored to return `(ir::Value, u32)` to enable optionally embedding the offset in the load/store instruction. For now it'd always return 0 but maybe a future refactoring to have to optimize this a bit in the future.

tests/disas/gc/struct-new.wat
```diff
@@ -0,0 +1,119 @@
...
+;;                                     v71 = stack_addr.i64 ss0
+;;                                     store notrap v4, v71
+;; @002a                               v7 = iconst.i32 -1476395008
+;; @002a                               v8 = iconst.i32 0
+;; @002a                               v9 = iconst.i32 32
+;; @002a                               v10 = iconst.i32 8
+;; @002a                               v11 = call fn0(v0, v7, v8, v9, v10), stack_map=[i32 @ ss0+0]  ; v7 = -1476395008, v8 = 0, v9 = 32, v10 = 8
+;; @002a                               v15 = uextend.i64 v11
+;; @002a                               v16 = iconst.i64 16
+;; @002a                               v17 = uadd_overflow_trap v15, v16, user65535  ; v16 = 16
```
alexcrichton, line 34: As a minor possible optimization (future PR) this should be pretty easy to optimize in the egraph pass to downgrade this to `iadd`. It's pretty obvious by construction that this can't overflow, along with many other patterns too.

Whether it's worth it I'm not sure. Optimizing bounds checks is naturally going to be much more involved than just this b/c we want to rely on guard pages exclusively in theory.

crates/cranelift/src/gc/enabled.rs
```diff
@@ -56,7 +64,309 @@ pub fn gc_ref_table_fill_builtin(
...
+    func_env: &mut FuncEnvironment<'_>,
+    builder: &mut FunctionBuilder<'_>,
+    struct_type_index: TypeIndex,
+    field_index: u32,
+    struct_ref: ir::Value,
+) -> WasmResult<ir::Value> {
+    // TODO: If we know we have a `(ref $my_struct)` here, instead of maybe a
+    // `(ref null $my_struct)`, we could omit the `trapz`. But plumbing that
+    // type info from `wasmparser` and through to here is a bit funky.
+    builder.ins().trapz(struct_ref, ir::TrapCode::NullReference);
```
alexcrichton, line 130: One idea for a future trick: we could unmap the lowest page of memory in the GC heap so address 0 in the gc heap translates to a fault. That way we could push this trap into an annotated trap code on the `MemFlags` on the load below (or store for `struct.set`) which means we could, by construction, completely eliminate this trap as well. 

(possibly all on top of a cranelift optimization pass to remove the checks entirely)

alexcrichton (approved):
Nice progress! 👍

crates/cranelift/src/func_environ.rs
```diff
@@ -1848,10 +1860,86 @@ impl<'module_environment> cranelift_wasm::FuncEnvironment for FuncEnvironment<'m
         builder: &mut FunctionBuilder,
         i31ref: ir::Value,
     ) -> WasmResult<ir::Value> {
+        // TODO: If we knew we have a `(ref i31)` here, instead of maybe a `(ref
+        // null i31)`, we could omit the `trapz`. But plumbing that type info
+        // from `wasmparser` and through to here is a bit funky.
```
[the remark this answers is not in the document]
fitzgen: If it is a GC ref that this function allocated, or has already accessed, then we should be able to clean it up. In theory at least; in practice it might be a little tricky without introducing some new powers to the mid-end. But for an argument that we haven't accessed yet, then we can't really do anything without the type info, I don't think.

crates/cranelift/src/gc/enabled.rs
fitzgen: I think we actually don't want to hinder optimizations. These aren't visible to Wasm semantics, or at least the whole point is they shouldn't be, so if the optimizer can prove that a load is redundant or whatever then I think we should let it.

crates/cranelift/src/gc/enabled.rs
fitzgen: Yeah, I was also thinking it may want to take a `object_size: Option<u32>` argument as well to make bounds checks of repeated accesses dedupe-able by the mid-end. (Has to be an `Option` because we don't statically know array lengths).

fitzgen: But: simple, naive things first and we can clean up the code we generate in follow ups.

tests/disas/gc/struct-new.wat
fitzgen: Ideally we'd have a mid-end ISLE rule something like

```lisp
(rule (simplify (uadd_overflow_trap $i64 x y))
    (if-let $true (add_cannot_overflow $I64 x y))
    (iadd x y))
```

where `add_cannot_overflow` handles constants and `(uextend ...)`s and such.

But because `uadd_overflow_trap` has side effects, we can't write such a rule in the mid-end today.

crates/cranelift/src/gc/enabled.rs
fitzgen: We discussed this a bit in today's cranelift meeting. In summary, this is neat but kind of conflicts with our plans of eventually using linear memories to implement the GC heap, since the pooling allocator would have to allocate a linear memory for the GC heap, then unmap the first page, give it to Wasm until the instance is done, then remap the first page before returning it to the memory pool. This is all doable of course, but the extra special-case code paths are annoying and adding more virtual memory calls in the pooling allocator is always fraught perf-wise.

PR #9277, opened 2024-09-18
Refactor the `wasmtime::runtime::vm::Store` trait

fitzgen:
This dates back from the `wasmtime` vs `wasmtime-runtime` crate split. Now that those crates merged, we can simplify some bits and remove unnecessary abstractions and indirections.

### DIFF
--- a/crates/wasmtime/src/runtime/vm/table.rs
+++ b/crates/wasmtime/src/runtime/vm/table.rs
@@ -618,9 +618,8 @@ impl Table {
             }
         }
 
-        // casting to u64 is ok to unwrap
         self.fill(
-            store.unwrap_gc_store_mut(),
+            store.store_opaque_mut().unwrap_gc_store_mut(),
             u64::try_from(old_size).unwrap(),
             init_value,
             u64::try_from(delta).unwrap(),
--- a/crates/wasmtime/src/runtime/vm/table.rs
+++ b/crates/wasmtime/src/runtime/vm/table.rs
@@ -620,7 +620,7 @@ impl Table {
 
         // casting to u64 is ok to unwrap
         self.fill(
-            store.gc_store(),
+            store.unwrap_gc_store_mut(),
             u64::try_from(old_size).unwrap(),
             init_value,
             u64::try_from(delta).unwrap(),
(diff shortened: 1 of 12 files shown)


crates/wasmtime/src/runtime/vm/table.rs
```diff
@@ -618,9 +618,8 @@ impl Table {
             }
         }
 
-        // casting to u64 is ok to unwrap
         self.fill(
-            store.gc_store(),
+            store.store_opaque_mut().unwrap_gc_store_mut(),
```
fitzgen, line 622: Note: This can't use the `DerefMut` impl, and must instead explicitly call `store_opaque_mut()`, because mutable borrows are invariant over type parameters.

alexcrichton: I'm not sure I understand this, can you expand on this a bit? Is this something where we're accidentally handing out `'static`? Or something is acidentally required to be `'static` when it shouldn't be? Should the deref impls become `impl Deref for dyn (Store + '_)` or something like that?

alexcrichton (approved):

crates/wasmtime/src/runtime/vm/table.rs
fitzgen: Here is the error, fwiw:

```
error: lifetime may not live long enough
   --> crates/wasmtime/src/runtime/vm/table.rs:622:13
    |
554 |         store: &mut dyn VMStore,
    |                - let's call the lifetime of this reference `'1`
...
622 |             store.unwrap_gc_store_mut(),
    |             ^^^^^ requires that `'1` must outlive `'static`
    |
    = note: requirement occurs because of a mutable reference to `dyn VMStore`
    = note: mutable references are invariant over their type parameter
    = help: see <https://doc.rust-lang.org/nomicon/subtyping.html> for more information about variance
```

My understanding is:

* We have a `store: &mut dyn VMStore`
* We are implicitly reborrowing it, in an attempt to get a shorter lifetime, by using the `DerefMut` impl at this call site
* However, because mutable references are invariant over their type parameter, we don't actually end up with a shorter lifetime, but the same lifetime
* And then for reasons I don't fully understand, that ends up forcing a `'static` constraint. I think because that is the only way that we could unify those two lifetimes? Not sure.

FWIW, making the reborrow explicit with `(&mut *store).unwrap_gc_store_mut()` results in the same error message, but pointing at the explicit reborrow instead of the method invocation.

I don't know if adding a `T: 'static` on the impl would solve the problem or not. But also we don't require that for `Store::new` so I don't think it is really an option.

Quite possible I am misunderstanding something here, but at the end of the day: without the `.store_opaque_mut()` we get a compiler error, with it we don't.

PR #9279, opened 2024-09-19
Align the start of stackslots instead of the end

tertsdiepraam:
Hi! I think that the stackslots weren't properly aligned, because the padding for the alignment is being added to the end of the slot instead of at the start, meaning that the slot itself isn't really being aligned.

This was creating issues for me where I requested a slot that was aligned at 16 bytes, but in reality it wasn't.

With this change this case from the tests:
```
ss0 = explicit_slot 8, align=16
ss1 = explicit_slot 8, align=16
ss2 = explicit_slot 4
ss3 = explicit_slot 4
```
gets these offsets:
```
ss0: 0
ss1: 16 (because size of ss0 + padding to get to 16)
ss2: 24 (because size of ss1)
ss3: 32 (because size of ss2 + padding to get to the word size of 8)
```
Note that the `align=16` of `ss0` is essentially a noop now.

And this case from a test I added:
```
ss0 = explicit_slot 8
ss1 = explicit_slot 32, align=16
```
gets these offsets:
```
ss0: 0
ss1: 16 (because size of ss0 + padding to get to 16)
```

I hope I didn't make any mistakes here or misinterpreted what stackslot alignment should do. Let me know if I need to change anything!

The issue was originally found here: https://bytecodealliance.zulipchat.com/#narrow/stream/217117-cranelift/topic/Retrieve.20alignment.20of.20architecture/near/471006324

### DIFF
--- a/cranelift/codegen/src/machinst/abi.rs
+++ b/cranelift/codegen/src/machinst/abi.rs
@@ -1152,42 +1152,59 @@ impl<M: ABIMachineSpec> Callee<M> {
         );
 
         // Compute sized stackslot locations and total stackslot size.
-        let mut sized_stack_offset: u32 = 0;
+        let mut end_offset: u32 = 0;
         let mut sized_stackslots = PrimaryMap::new();
+
         for (stackslot, data) in f.sized_stack_slots.iter() {
-            let off = sized_stack_offset;
-            sized_stack_offset = sized_stack_offset
-                .checked_add(data.size)
-                .ok_or(CodegenError::ImplLimitExceeded)?;
-            // Always at least machine-word-align slots, but also
+            // We start our computation possibly unaligned where the previous
+            // stackslot left off.
+            let unaligned_start_offset = end_offset;
+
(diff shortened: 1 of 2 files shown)


cranelift/codegen/src/machinst/abi.rs
```diff
@@ -1731,6 +1748,15 @@ impl<M: ABIMachineSpec> Callee<M> {
         let bytes = M::word_bytes();
         let total_stacksize = self.stackslots_size + bytes * spillslots as u32;
         let mask = M::stack_align(self.call_conv) - 1;
+        if total_stacksize.checked_add(mask).is_none() {
+            dbg!(
```
bal-e, line 1752: Should this cause a panic?

tertsdiepraam: Oops committed by accident thanks

cfallin (approved):
Thanks a bunch for fixing this!

tertsdiepraam:
Thanks!

PR #9281, opened 2024-09-19
Deprecate Context::compile_and_emit

bjorn3:
The only thing it does over Context::compile is writing the machine code to a Vec specified by the user. Every user of Context::compile_and_emit in Cranelift didn't actually need ownership over the machine code bytes and thus would have been able to use Context::compile instead to avoid an unnecessary allocation. This commit moves all in-tree users of compile_and_emit to compile and additionally to avoid other users adding an unnecessary allocation as they though it to be necessary, this commit additionally deprecates compile_and_emit.


fitzgen (approved):
Thanks!

PR #9282, opened 2024-09-19
Add constant evaluation support for the `struct.new[_default]` Wasm GC instructions

fitzgen:

### DIFF
--- a/crates/wasmtime/src/runtime/vm/const_expr.rs
+++ b/crates/wasmtime/src/runtime/vm/const_expr.rs
@@ -30,14 +30,14 @@ impl<'a, 'b> ConstEvalContext<'a, 'b> {
         Self { instance, module }
     }
 
-    fn global_get(&mut self, index: GlobalIndex) -> Result<ValRaw> {
+    fn global_get(&mut self, store: &mut AutoAssertNoGc<'_>, index: GlobalIndex) -> Result<ValRaw> {
         unsafe {
             let global = self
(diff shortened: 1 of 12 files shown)


crates/wasmtime/src/runtime/vm/const_expr.rs
```diff
@@ -43,6 +47,116 @@ impl<'a, 'b> ConstEvalContext<'a, 'b> {
...
+        // and we can't use `AsyncCx::block_on` here to fake it because we
+        // aren't guaranteed to be running on a fiber yet, so we would be forced
+        // to create sync and async versions of every function up to at least
+        // `ConstExprEvaluator::eval`.
+        //
+        // In the meantime, just give up early without trying a GC if allocation
+        // fails.
+        let struct_ref = StructRef::_new(&mut store, &allocator, &fields)?;
+
+        let mut store = AutoAssertNoGc::new(&mut store);
```
alexcrichton, line 108: This has me thinking, isn't rooting a problem for the entire duration of the const eval? For example we've just got these `ValRaw::anyref` values floating around internally on the const eval stack, so if a GC were to happen in `StructRef::_new` above, for example, it wouldn't be able to find those roots?

Should `AutoAssertNoGc` be wrapped around the entire const eval operation?

alexcrichton (approved):

alexcrichton (approved):

PR #9283, opened 2024-09-19
Add release artifacts for ARM64 Windows

alexcrichton:
With #9266 Wasmtime should have full support for this platform. While not tested in CI it can still be useful to build artifacts nonetheless in case users are interested.


elliottt (approved):
:window:

PR #9285, opened 2024-09-19
Introduce the `wasmtime::EqRef` type

fitzgen:
This commit introduces the `wasmtime::EqRef` type, which corresponds to Wasm's `(ref eq)` type, and statically represents Wasm references that can be tested for equality. This is primarily for use with `Func::typed`-style APIs.

### DIFF
--- a/crates/wasmtime/src/runtime/func.rs
+++ b/crates/wasmtime/src/runtime/func.rs
@@ -601,18 +601,23 @@ impl Func {
     /// | `Func`                            | `(ref func)`                              |
     /// | `Option<Nofunc>`                  | `nullfuncref` aka `(ref null nofunc)`     |
     /// | `NoFunc`                          | `(ref nofunc)`                            |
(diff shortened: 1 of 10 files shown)


crates/wasmtime/src/runtime/func.rs
```diff
@@ -1445,7 +1452,8 @@ impl Func {
     /// | `(ref nofunc)`                            | `NoFunc`                              |
     /// | `v128`                                    | `V128` on `x86-64` and `aarch64` only |
```
alexcrichton, line 1453: While you're here mind removing this clause about aarch64/x86_64? (it's available on all platforms now)

alexcrichton (approved):

crates/wasmtime/src/runtime/func.rs
alexcrichton: Ah I had no other comments so gonna go ahead and merge

PR #9287, opened 2024-09-19
Make the Tail call conv follow the system call conv for the return area ptr

bjorn3:


cranelift/codegen/src/isa/aarch64/abi.rs
```diff
@@ -151,6 +151,33 @@ impl ABIMachineSpec for AArch64MachineDeps {
         let max_per_class_reg_vals = 8; // x0-x7 and v0-v7
         let mut remaining_reg_vals = 16;
 
+        let ret_area_ptr = if add_ret_area_ptr {
+            debug_assert!(args_or_rets == ArgsOrRets::Args);
```
fitzgen, line 155:
```suggestion
            debug_assert_eq!(args_or_rets, ArgsOrRets::Args);
```

for slightly better panic messages

cranelift/codegen/src/isa/pulley_shared/abi.rs
```diff
@@ -70,6 +70,19 @@ where
         let mut next_v_reg = 0;
         let mut next_stack: u32 = 0;
 
+        let ret_area_ptr = if add_ret_area_ptr {
+            assert!(ArgsOrRets::Args == args_or_rets);
```
fitzgen, line 74:
```suggestion
            assert_eq!(ArgsOrRets::Args, args_or_rets);
```

cranelift/codegen/src/isa/pulley_shared/abi.rs
```diff
@@ -70,6 +70,19 @@ where
...
+            next_x_reg += 1;
+            Some(ABIArg::reg(
+                x_reg(0).to_real_reg().unwrap(),
+                I64,
+                ir::ArgumentExtension::None,
+                ir::ArgumentPurpose::Normal,
+            ))
```
fitzgen, line 81: To make this code a little future proof to shuffling it around at some point in the future, lets not hard code `0`:

```suggestion
            next_x_reg += 1;
            Some(ABIArg::reg(
                x_reg(next_x_reg - 1).to_real_reg().unwrap(),
                I64,
                ir::ArgumentExtension::None,
                ir::ArgumentPurpose::Normal,
            ))
```

It may be convenient to have a helper closure that does the increment of `next_x_reg` and returns `Option<Reg>` or whatever for you. I'll leave it up to you, and how much you are annoyed by `next_x_reg - 1` :-p

cranelift/codegen/src/isa/riscv64/abi.rs
```diff
@@ -111,11 +111,18 @@ impl ABIMachineSpec for Riscv64MachineDeps {
...
+        let ret_area_ptr = if add_ret_area_ptr {
+            assert!(ArgsOrRets::Args == args_or_rets);
             next_x_reg += 1;
-        }
+            Some(ABIArg::reg(
+                x_reg(x_start).to_real_reg().unwrap(),
+                I64,
+                ir::ArgumentExtension::None,
+                ir::ArgumentPurpose::Normal,
+            ))
```
fitzgen, line 122: Feel similarly here

fitzgen (approved):
Thanks! LGTM, although I have some minor nitpicks. If you feel strongly that some of them are unnecessary, I can be convinced otherwise.

cranelift/codegen/src/isa/aarch64/abi.rs
bjorn3: There are only two options in this enum and this matches the other debug assertions, but sure.

cranelift/codegen/src/isa/riscv64/abi.rs
bjorn3: This really needs to be the first argument register. If next_x_reg - 1 is not x_start at this point, that is a bug.

fitzgen (approved):

Issue #9249, opened 2024-09-15
Support list of paths in component bindgen macro


lwansbrough:
wit-bindgen includes the option to provide a list of paths for the wit sources. This makes it more practical to satisfy the requirements of imports from other packages. Enabling this behaviour for the component bindgen macro would be helpful.

wit-bindgen docs:

```
// Path to parse WIT and its dependencies from. Defaults to the `wit`
// folder adjacent to your `Cargo.toml`.
//
// This parameter also supports the form of a list, such as:
// ["../path/to/wit1", "../path/to/wit2"]
// Usually used in testing, our test suite may want to generate code
// from wit files located in multiple paths within a single mod, and we
// don't want to copy these files again.
path: "../path/to/wit",
```

alexcrichton:
Thanks for the report! To link things here this was implemented in wit-bindgen in https://github.com/bytecodealliance/wit-bindgen/pull/1003 and I think it'd be reasonable to copy over a similar structure into Wasmtime as well. @lwansbrough would you be interested in creating a PR perhaps?

lwansbrough:
@alexcrichton I'll see what I can do.

I was looking through that commit and this looks like a bug to me. https://github.com/bytecodealliance/wit-bindgen/blob/3e3877da85daf1ed98bd1aee7007e496461c3bd1/crates/guest-rust/macro/src/lib.rs#L215

Line 237 passes `root.join("wit")` which would mean `root` is joined to `root`.

alexcrichton:
Good catch! Semantically that'll work out ok since `"/a".join("/b")` yields `"/b"`, so the end result is the same. That being said it's unnecessarily complex, and just passing in `"wit"` instead of `root.join("wit")` should also work and be cleaner.

MarinPostma:
Will this enable importing transitive dependencies of the package? Say I am generating binding for package A, but package `A` includes package `B` in a different directory, should I just be able to add the path to `B` and have it all work?

alexcrichton:
@MarinPostma I believe so yeah, but it if it doesn't

MarinPostma:
I'll give it a try then :)

PR #9288, opened 2024-09-19
feat: component bindgen: add support for multiple wit paths

lwansbrough:
This change makes it possible to compile bindings from multiple wit source paths. It resolves https://github.com/bytecodealliance/wasmtime/issues/9249 and is heavily informed by the work done for wit-bindgen in https://github.com/bytecodealliance/wit-bindgen/pull/1003

I had to modify one existing test (`trappable_errors_with_versioned_and_unversioned_packages`) to support this behaviour as the existing test was created without the expectation that the world in `tests/codegen/unversioned-foo.wit` would be considered a candidate world for selection.

### DIFF
--- a/crates/component-macro/src/bindgen.rs
+++ b/crates/component-macro/src/bindgen.rs
@@ -171,8 +171,7 @@ impl Parse for Config {
         let (resolve, pkgs, files) = parse_source(&paths, &inline, &features)
             .map_err(|err| Error::new(call_site, format!("{err:?}")))?;
 
-        let world = 
-            select_world(&resolve, &pkgs, world.as_deref())
+        let world = select_world(&resolve, &pkgs, world.as_deref())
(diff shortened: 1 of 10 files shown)


lwansbrough:
might need some guidance on how to setup the expanded tests

alexcrichton:
Thanks for the work here! This all looks great to me, so all that's left is sorting out CI and tests and such. For the expanded tests you can run the tests with `BINDGEN_TEST_BLESS=1` set in your environment to automatically update the expectations (and then check them in as a commit). For testing this feature itself feel free to use either the codegen test suite (which may mean you'll need to set up some directories if you don't use preexisting ones) or set up some directories in `tests/all/component_model/*`, whichever you find easiest.

alexcrichton (approved):
Thanks again for this!

PR #9297, opened 2024-09-24
ci: use node20 for github release action

trivikr:
Blog post: https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/


alexcrichton (approved):
Thanks!

PR #9301, opened 2024-09-24
Add the `ArrayOutOfBounds` trap code

fitzgen:
This is needed in the implementation of Wasm GC's array instructions.


alexcrichton (approved):
Musing about this, we may actually want to one day not have the exact list of traps in Cranelift match the exact list of traps in Wasmtime. We might want to define Cranelift trap codes are "simply a u8" which might be much more flexible for Wasmtime as well since we wouldn't have to mirror everything in two locations. No need to do that now though, just a future possibility.

fitzgen:
We'd have to figure out what to do about how they are packed into `MemFlags`. Might be worth taking flags out into a secondary map or something.

alexcrichton:
Oh `MemFlags` should be able to easily fit 8 bits. Custom user traps in Cranelift right now though are 16 bits and that we can't fit the whole width into `MemFlags`

PR #9302, opened 2024-09-24
Add `-Wasync-stack-size`, automatically configure it too

### LINKED ISSUE #9298: Error: max_wasm_stack size cannot exceed the async_stack_size
### Test Case

Testing dotnet AOT with the new wasmtime v 25 in the dotnet CI.
The [full log](https://helixre107v0xdcypoyl9e7f.blob.core.windows.net/dotnet-runtime-refs-pull-108179-merge-a238d0a9053e45129a/System.Collections.Specialized.Tests.Attempt.3/1/console.c5dc63d9.log?helixlogtype=result)

### Steps to Reproduce

`wasmtime --dir=. --wasm max-wasm-stack=134217728 --wasi http --wasi inherit-network --wasi allow-ip-name-lookup dotnet.wasm`

### Expected Results

probably we need a way how to specify `--wasm async-stack-size=134217728` on cli.
But I'm not sure I understand the feature well enough.

### Actual Results

`Error: max_wasm_stack size cannot exceed the async_stack_size`

### Versions and Environment

Wasmtime version or commit: wasmtime 25.0.0 (0b195ef5d 2024-09-20)

Operating system: ubuntu-22.04

Architecture: x64

### Extra Info

alexcrichton:
Recently the `wasmtime` CLI switched from sync to async to handle interrupting host APIs in WASI. Previously the CLI provided no means to configure the async stack size, however, which now means that when increasing the max wasm stack the CLI prints an error that cannot be resolved about the wasm stack being larger than the async stack. This commit fixes this issue by both adding a configuration option for the async stack size and additionally automatically increasing the async stack size when only the wasm stack size is provided.

Closes #9298

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -437,12 +437,25 @@ gc = ["wasmtime-cli-flags/gc"]
 
 # CLI subcommands for the `wasmtime` executable. See `wasmtime $cmd --help`
 # for more information on each subcommand.
-serve = ["wasi-http", "component-model", "dep:http-body-util", "dep:http"]
+serve = [
+  "wasi-http",
+  "component-model",
+  "dep:http-body-util",
+  "dep:http",
+  "wasmtime-cli-flags/async",
+]
 explore = ["dep:wasmtime-explorer", "dep:tempfile"]
 wast = ["dep:wasmtime-wast"]
 config = ["cache"]
 compile = ["cranelift"]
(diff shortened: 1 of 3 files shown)


alexcrichton:
I also think this'd make a good candidate for a 25.0.1 release since otherwise `-Wmax-wasm-stack` is a little broken right now.

abrown (approved):

PR #9304, opened 2024-09-24
Print an error on async stack overflow

alexcrichton:
This commit updates Wasmtime's handling of traps on Unix platforms to print an error message on stack overflow when the guard page is hit. This is distinct from stack overflow in WebAssembly which raises a normal trap and can be caught. This is instead to be used on misconfigured hosts where the async stack is too small or wasm was allowed to take up too much of the async stack. Currently no error message is printed and the program simply aborts with a core dump which can be difficult to debug.

This instead registers the range of the async guard page with the trap handling infrastructure to test the faulting address and if it lies within this range. If so then a small message is printed and then the program is aborted with `libc::abort()`.

This does not impact the safety of any prior embedding or fix any issues. It's instead intended purely as a diagnostic tool to help users more quickly understand that stack size configuration settings are the likely culprit.

### DIFF
--- a/crates/c-api/src/async.rs
+++ b/crates/c-api/src/async.rs
@@ -389,6 +389,9 @@ unsafe impl StackMemory for CHostStackMemory {
         let base = unsafe { top.sub(len) as usize };
         base..base + len
     }
+    fn guard_range(&self) -> Range<*mut u8> {
+        std::ptr::null_mut()..std::ptr::null_mut()
+    }
 }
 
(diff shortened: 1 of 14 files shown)


alexcrichton:
This is inspired by https://github.com/bytecodealliance/wasmtime/issues/9298 where stack overflows on the CLI in async fibers might be more common now.

elliottt (approved):
This all seems reasonable to me!

PR #9307, opened 2024-09-24
Update to cap-std 3.3.0.

### LINKED ISSUE #9272: Improper resolution of .. components in symlinks
I believe I've run across a bug in wasmtime's handling of symlink targets which include `".."`. I've reproduced this using the Go program below which calls `path_open` on a path and reports the result. It's not the simplest test program, but it avoids dependencies and I have a Go compiler handy.

Simple case, no bug:

```
$ echo foo > file
$ ln -s file link
$ GOOS=wasip1 GOARCH=wasm go run main.go link
errno 0
```

The bug shows up when resolving a link with a target of `".."`:

```
$ echo foo > target
$ mkdir b
$ (cd b && ln -s .. up)
$ cat b/up/target
foo
$ GOOS=wasip1 GOARCH=wasm go run main.go b/up
errno 0
$ GOOS=wasip1 GOARCH=wasm go run main.go b/up/target
Not a directory
```

I'm not sure exactly what is going awry, but the "not a directory" error is clearly wrong. Note that we can open `b/up` (resolves to `.` as expected). The problem seems to be specific to links that end in `..`: if we change the link target to `../a`, everything works as expected:

```
$ mkdir a
$ echo bar > a/target
$ mkdir b
$ (cd b && ln -s ../a upa)
$ cat b/upa/target
bar
$ GOOS=wasip1 GOARCH=wasm go run main.go b/upa/target
errno 0
```

Test program:

```
package main

import (
        "fmt"
        "os"
        "syscall"
        "unsafe"
)

//go:wasmimport wasi_snapshot_preview1 path_open
//go:noescape
func path_open(rootFD int32, dirflags uint32, path unsafe.Pointer, pathLen uint32, oflags uint32, fsRightsBase uint64, fsRightsInheriting uint64, fsFlags uint32, fd unsafe.Pointer) uint32

const LOOKUP_SYMLINK_FOLLOW = 1

func main() {
        f, err := os.Open(".")
        if err != nil {
                panic(err)
        }

        filename := os.Args[1]
        var fd int32
        errno := path_open(
                int32(f.Fd()),
                syscall.LOOKUP_SYMLINK_FOLLOW,
                unsafe.Pointer(unsafe.StringData(filename)),
                uint32(len(filename)),
                0,
                syscall.RIGHT_FD_READ,
                syscall.RIGHT_FD_READ,
                0,
                unsafe.Pointer(&fd),
        )
        fmt.Println(syscall.Errno(errno))
}
```

sunfishcode:
cap-std 3.3.0 contains bytecodealliance/cap-std#366, which fixes #9272, which is about the handling of `..` in symlink destinations.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -259,9 +259,9 @@ dependencies = [
 
 [[package]]
 name = "cap-fs-ext"
-version = "3.0.0"
+version = "3.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "769f8cd02eb04d57f14e2e371ebb533f96817f9b2525d73a5c72b61ca7973747"
+checksum = "712695628f77a28acd7c9135b9f05f9c1563f8eb91b317f63876bac550032403"
 dependencies = [
  "cap-primitives",
  "cap-std",
@@ -271,9 +271,9 @@ dependencies = [
 
 [[package]]
 name = "cap-net-ext"
-version = "3.0.0"
+version = "3.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "59ff6d3fb274292a9af283417e383afe6ded1fe66f6472d2c781216d3d80c218"
+checksum = "7d609980992759cef960324ccece956ee87929cc05a75d6546168192063dd8b1"
 dependencies = [
  "cap-primitives",
  "cap-std",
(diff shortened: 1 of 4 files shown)


elliottt (approved):

sunfishcode:
This PR doesn't involve pulley-interpreter at all, however cargo-vet in CI appears to be failing with:

```

alexcrichton:
There's something I think none of us understand about cargo-vet and how it relates to (a) adding new crates in the workspace and (b) causing vets to fail after the crate was actually published. The weird part is we've even landed PRs today after the publication of Wasmtime 25, so I'm not sure why this PR ended up being the victim...

PR #9312, opened 2024-09-25
Add a `wasmtime completion` subcommand

alexcrichton:
This commit adds a new subcommand to the `wasmtime` CLI which can be used to generate shell completions for the `wasmtime` version that is installed. This is relatively easy to do with the `clap_complete` crate which is also used with various other clap-based CLIs in Rust to generate completions.


elliottt (approved):
Nice!

PR #9322, opened 2024-09-28
bug: bindgen erraneously parses extra comma

lwansbrough:
Fixes an issue introduced in #9288 where an additional comma token is erroneously consumed. This meant if path was not the last property in the bindgen options, you would see an error "expected ,"

The provided tests were simple enough that path happened to be the last property. I haven't included a new test here because I think testing for such an issue is outside the scope of the tests.


alexcrichton (approved):
Thanks!

PR #9323, opened 2024-09-28
Add contributors docs about lints/warnings/clippy

alexcrichton:
Inspired by #9318 I realized we haven't actually documented anything about compiler warnings or Clippy warnings in our contributor docs, so I have aspired to do so in this commit. My goal is to reflect the current state of the repository in this documentation, not add anything new.


fitzgen (approved):
Thanks!

PR #9325, opened 2024-09-28
Enable a few miscellaneous Clippy lints

alexcrichton:
I was browsing the [list of lints](https://rust-lang.github.io/rust-clippy/master/index.html) and enabled a few that seemed useful for cleaning up code after a refactoring.

### DIFF
--- a/crates/wasi/src/preview1.rs
+++ b/crates/wasi/src/preview1.rs
@@ -1290,12 +1290,8 @@ impl wasi_snapshot_preview1::WasiSnapshotPreview1 for WasiP1Ctx {
         let fd = self.get_file_fd(fd)?;
         self.as_wasi_impl()
             .advise(fd, offset, len, advice.into())
-            .await
-            .map_err(|e| {
-                e.try_into()
-                    .context("failed to call `advise`")
-                    .unwrap_or_else(types::Error::trap)
-            })
+            .await?;
+        Ok(())
     }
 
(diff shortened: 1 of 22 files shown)


crates/wasi/src/preview1.rs
```diff
@@ -1290,12 +1290,8 @@ impl wasi_snapshot_preview1::WasiSnapshotPreview1 for WasiP1Ctx {
...
         self.as_wasi_impl()
             .advise(fd, offset, len, advice.into())
-            .await
-            .map_err(|e| {
-                e.try_into()
-                    .context("failed to call `advise`")
-                    .unwrap_or_else(types::Error::trap)
-            })
+            .await?;
+        Ok(())
```
elliottt, line 1294: Was it intentional that the `.context` calls were all removed in this module?

elliottt (approved):
I'm curious about the changes to the preview1 module removing all the `.context` calls, but other than that these all seem like great improvements 👍

crates/wasi/src/preview1.rs
alexcrichton: Intentional yeah, turns out that the `try_into` in these cases was infallible and couldn't ever fail. That meant that the error case never happened and the `.context(...)` calls were basically all dead code. I switched them all over to using `?` which leverages the `into` conversion automatically which is known to never fail.

PR #9326, opened 2024-09-28
Implement most `array.*` instructions for the GC proposal

fitzgen:
This does not implement `array.copy` and `array.init_elem` yet, but implements all other `array.*` instructions:

* `array.new`
* `array.new_fixed`
* `array.new_default`
* `array.new_data`
* `array.new_elem`
* `array.fill`
* `array.init_data`
* `array.len`
* `array.get`
* `array.get_s`
* `array.get_u`
* `array.set`

Note that the initial plumbing for `array.{copy,init_elem}` is in place, but the instructions themselves are not implemented yet.


crates/cranelift/Cargo.toml
```diff
@@ -19,7 +19,7 @@ anyhow = { workspace = true }
 log = { workspace = true }
 wasmtime-environ = { workspace = true, features = ['compile'] }
 cranelift-wasm = { workspace = true }
-cranelift-codegen = { workspace = true, features = ["host-arch"] }
+cranelift-codegen = { workspace = true, features = ["host-arch", "trace-log"] }
```
alexcrichton, line 22: stray debug?

crates/cranelift/src/gc/enabled/drc.rs
```diff
@@ -207,11 +260,244 @@ impl DrcCompiler {
...
+
+        match init {
+            ArrayInit::Elems(elems) => {
+                for val in elems {
+                    self.init_field(func_env, builder, elem_addr, &elem_ty, *val)?;
+                    elem_addr = builder.ins().iadd(elem_addr, elem_size);
+                }
+            }
+            ArrayInit::Fill { elem, len: _ } => {
+                // To initialize the array by filling it with `elem`, emit the
```
alexcrichton, line 432: This looks pretty similar to `translate_array_fill` above, would it be possible to refactor the two to share more code?

crates/cranelift/src/gc/enabled/drc.rs
```diff
@@ -105,6 +117,47 @@ impl DrcCompiler {
...
+        let flags = ir::MemFlags::trusted().with_endianness(ir::Endianness::Little);
+
+        match ty {
+            WasmStorageType::Val(WasmValType::Ref(r))
+                if r.heap_type.top() == WasmHeapTopType::Func =>
+            {
+                return Err(wasm_unsupported!(
+                    "funcrefs inside the GC heap are not yet implemented"
+                ));
+            }
```
alexcrichton, line 139: Could `translate_init_gc_reference` below internally do this so it doesn't have to be done each time a translate helper is called? (I saw this above as well for writes I think)

crates/wasmtime/src/runtime/vm/instance.rs
```diff
@@ -469,6 +469,19 @@ impl Instance {
...
+    /// Serves a similar purpose as `OpaqueRootScope`, but for situations where
+    /// you need to enter a GC scope (which would normally require mutably
+    /// borrowing the instance's underlying store) but also access to `Instance`
+    /// methods that will internally mutably borrow that store as well.
+    pub(crate) fn with_gc_lifo_scope<T>(&mut self, f: impl FnOnce(&mut Self) -> T) -> T {
+        let scope = unsafe { (*self.store()).gc_roots().enter_lifo_scope() };
+        let result = f(self);
+        unsafe {
+            (*self.store()).exit_gc_lifo_scope(scope);
+        }
```
alexcrichton, line 481: Worth using a destructor here for panic-safety?

alexcrichton (approved):

crates/cranelift/Cargo.toml
fitzgen: thanks, yeah

crates/cranelift/src/gc/enabled/drc.rs
fitzgen: good idea

crates/cranelift/src/gc/enabled/drc.rs
fitzgen: I actually like having it repeated, so I know exactly all the places I need to fix up later (which I plan on doing soon, so that I can run some of the spec tests that use funcrefs)

PR #9330, opened 2024-09-29
Update region to 3.0.2

sunshowers:
Hi there -- I've been working on some code changes to wasmtime (primarily around illumos support) and I noticed that the region dependency was out of date. Here's a PR to update the dependency to version 3, which supports illumos.

The main API difference is that `region::page::ceil` now works on addresses rather than usizes, so there's a small adaptation for that in this PR. Thanks!


alexcrichton (approved):
Thanks for the PR!

PR #9332, opened 2024-09-30
feat(cli): allow underscores for numeric options used in wasmtime

vados-cosmonic:
This commit allows for underscores to be used in numeric arguments to the wasmtime CLI.

For example, a string like `1_200` will be interpreted as `1200`.

This change would make it easier to enter large values (ex. `--max-memory-size`) when using `wasmtime` via CLI


alexcrichton (approved):
Thanks!

PR #9336, opened 2024-09-30
Update wasm-tools crates

alexcrichton:
* WIT `float32` and `float64` are now hard errors (renamed to `f32` and `f64`)
* Stack switching has been implemented in wasm-tools, but not Wasmtime.
* New shared-everything-threads component model intrinsics are not implemented

"Holes" for these new features in wasmparser are now present in Wasmtime and panics shouldn't be hit because the new proposals aren't enabled in the validator at this time.


fitzgen (approved):
THanks!

PR #9353, opened 2024-10-01
Enable Rust's `unused-macro-rules` lint

alexcrichton:
This is allow-by-default in the compiler but looks like it's got a high fidelity for us where we don't use macros super heavily, so let's try having it enabled.


elliottt (approved):
Nice!

PR #9356, opened 2024-10-01
Allow alignments that eat the remaining allocation

elliottt:
The preview1 adapter has an implementation of cabi_realloc that facilitates calling component imports that return dynamically allocated results, like lists and strings. However, when returning a zero-length value to a buffer that's been allocated with zero bytes available, a trap will occur. This is because the alignment check requires that there is space left over after alignment has been added to the base pointer, which in the case of an empty buffer, will not be possible.

This PR fixes the issue by relaxing the alignment check to require that the amount required to align the base does not exceed the length of the buffer, but may consume the entire allocation.

### DIFF
--- a/crates/test-programs/src/bin/cli_stdin_empty.rs
+++ b/crates/test-programs/src/bin/cli_stdin_empty.rs
@@ -1,0 +1,16 @@
+use test_programs::preview1::STDIN_FD;
+
+fn main() {
+    let mut buffer = [0_u8; 0];
+
+    unsafe {
+        wasi::fd_read(
+            STDIN_FD,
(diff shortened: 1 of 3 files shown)


alexcrichton:
Would it be possible to add a test for this as well? Or is that out-of-scope in the WASI APIs we have bound?

alexcrichton (approved):

elliottt:
> Would it be possible to add a test for this as well? Or is that out-of-scope in the WASI APIs we have bound?

The test added fails when the change to the `align_to` function is reverted, triggering the original issue in the adapter.

PR #9358, opened 2024-10-02
Implement the `array.init_elem` Wasm GC instruction

fitzgen:

### DIFF
--- a/crates/wasmtime/src/runtime/vm/libcalls.rs
+++ b/crates/wasmtime/src/runtime/vm/libcalls.rs
@@ -727,6 +727,7 @@ unsafe fn array_new_elem(
     };
     use wasmtime_environ::{ModuleInternedTypeIndex, TableSegmentElements};
 
+    // Convert indices to their typed forms.
     let array_type_index = ModuleInternedTypeIndex::from_u32(array_type_index);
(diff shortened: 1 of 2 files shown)


crates/wasmtime/src/runtime/vm/libcalls.rs
```diff
@@ -808,15 +809,113 @@ unsafe fn array_new_elem(
...
+    };
+    use wasmtime_environ::{ModuleInternedTypeIndex, TableSegmentElements};
+
+    // NB: Don't use `OpaqueRootScope` here because we need to borrow the store
+    // through `instance` during const evaluation, which is within the same
+    // region that the `OpaqueRootScope` would otherwise span while borrowing
+    // the same store, resulting in double borrows.
+    instance.with_gc_lifo_scope(|instance| {
+        // Convert the indices into their typed forms.
+        let _array_type_index = ModuleInternedTypeIndex::from_u32(array_type_index);
```
elliottt, line 833: Should this be removed?

fitzgen: I left it in just so that if we ever refactor this implementation and need to use it, we don't mix up which kind of index it is.

elliottt (approved):

PR #5172, opened 2022-11-01
Generate SSA code from returns

elliottt:
Modify return pseudoinstructions to have pairs of registers: virtual and real. This allows us to constrain the virtual registers to the real ones specified by the abi, instead of directly emitting moves to those real regisers.

Additionally, this PR marks `r14` on s390x as non-allocatable, as it's used as the link register in the return instruction.

PR #9361, opened 2024-10-03
s390x: Hard code the link register for ElfTlsGetOffset

bjorn3:
The ElfTlsGetOffset pseudo-instruction already hard codes the rest of the abi anyway and if another abi is ever needed, adding another pseudo-instruction or an enum field to the existing one would likely be necessary anyway. Also remove a couple of dead_code allows.

### DIFF
--- a/cranelift/bforest/src/node.rs
+++ b/cranelift/bforest/src/node.rs
@@ -13,7 +13,6 @@ use core::fmt;
 ///
 /// An inner node contains at least M/2 node references unless it is the right-most node at its
 /// level. A leaf node contains at least N/2 keys unless it is the right-most leaf.
-#[allow(dead_code)] // workaround for https://github.com/rust-lang/rust/issues/64362
 pub(super) enum NodeData<F: Forest> {
     Inner {
         /// The number of keys in this node.
--- a/cranelift/codegen/src/isa/riscv64/inst/imms.rs
+++ b/cranelift/codegen/src/isa/riscv64/inst/imms.rs
@@ -2,7 +2,6 @@
 
 // Some variants are never constructed, but we still want them as options in the future.
 use super::Inst;
-#[allow(dead_code)]
 use std::fmt::{Debug, Display, Formatter, Result};
 
 #[derive(Copy, Clone, Debug, Default)]
--- a/cranelift/codegen/src/isa/s390x/inst.isle
+++ b/cranelift/codegen/src/isa/s390x/inst.isle
@@ -1020,7 +1020,6 @@
       (tls_offset WritableReg)
       (got Reg)
       (got_offset Reg)
-      (link WritableReg)
       (symbol BoxSymbolReloc))
 ))
 
--- a/cranelift/codegen/src/isa/s390x/inst/emit.rs
+++ b/cranelift/codegen/src/isa/s390x/inst/emit.rs
@@ -3250,9 +3250,7 @@ impl Inst {
                 put(sink, &enc_rr(opcode, gpr(15), rn));
                 sink.add_call_site();
             }
-            &Inst::ElfTlsGetOffset {
-                ref symbol, link, ..
-            } => {
+            &Inst::ElfTlsGetOffset { ref symbol, .. } => {
                 let opcode = 0xc05; // BRASL
 
                 // Add relocation for target function. This has to be done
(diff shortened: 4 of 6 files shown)


cfallin:
cc @uweigand -- this looks reasonable to me overall, and there's no functional change to the emitted code or the regalloc metadata (fields were only ever used with one value) -- but maybe there's a reason it was generic?

uweigand:
Well, as long as the ELF ABI is the only s390x ABI supported by wasmtime, I guess it doesn't make any difference, really.

That said, this seems a step in the wrong direction to me: there *are* other ABIs used by other OSes on the platforms, and they do use different registers than r14 as the link register, so I'm not sure why it's an improvement to hard-code that value?  Logically, it is an ABI property, that's why I placed that choice in ABI code rather than ISA code.

bjorn3:
The ret instruction is also hard coding r15 (stack pointer?) and it has a debug assert that r14 is used as link register. What other OSes use a different link register?

uweigand:
> The ret instruction is also hard coding r15 (stack pointer?) 

Ah, not really.  This instruction uses a condition mask field as "R1", not an actual register field, and the contents are 15 (all mask bits set) to indicate a "branch on any condition", i.e. an unconditional branch.  It would have been clearer to add a variant of the `enc_rr` helper instead of passing the condition mask as a "register".

> and it has a debug assert that r14 is used as link register.

Huh, I'm surprised to see that.  Looks like @elliottt added a bunch of asserts here: https://github.com/bytecodealliance/wasmtime/pull/5172 - I'm not sure if there's anything in that diff that makes those asserts necessary?

> What other OSes use a different link register?

Well, XPLINK (the current main calling convention on z/OS) uses %r4, as one major example.


In general, I'm pretty sure in case we ever wanted to support a significantly different calling convention like XPLINK that uses different registers for the link register, the stack register, and pretty much any ABI-defined register, we will find places where the current code isn't fully generalized.  I tried to keep things general where it seemed easily possible, but didn't attempt any full review ...   I just don't see why it's beneficial now to make the code *less* general.

elliottt:
> Huh, I'm surprised to see that. Looks like @elliottt added a bunch of asserts here: #5172 - I'm not sure if there's anything in that diff that makes those asserts necessary?

I think you're right, we should probably remove them. They're added in other cases to check that fixed constraints are respected, but as there's no fixed constraints in the constraint generation code in mod.rs, I don't think that assertion applies.

uweigand:
> > Huh, I'm surprised to see that. Looks like @elliottt added a bunch of asserts here: #5172 - I'm not sure if there's anything in that diff that makes those asserts necessary?
> 
> I think you're right, we should probably remove them. They're added in other cases to check that fixed constraints are respected, but as there's no fixed constraints in the constraint generation code in mod.rs, I don't think that assertion applies.

Thanks.  I've now submitted a PR: https://github.com/bytecodealliance/wasmtime/pull/9397

bjorn3:
I removed the call instruction changes. I think it still makes sense to keep them for the ElfTlsGetOffset pseudo-instruction as that one hard codes a single abi anyway.

cranelift/codegen/src/isa/s390x/inst/regs.rs
```diff
@@ -78,6 +78,10 @@ pub fn zero_reg() -> Reg {
     gpr(0)
 }
 
+pub fn link_reg() -> Reg {
```
uweigand, line 81: I don't really like this.  This function, with this name in this place, gives the wrong impression that this is a ISA-defined register, and encourages additional uses in the future.

If we need a function like this, it should be an *ABI* function.

In any case, if you really want to hard-code the link register in `ElfTlsGetOffset`, then I'd suggest to just hard-code 14 directly, like this code also hard-codes register numbers 12 and 2.  (Again, I'd prefer to go the other way and not hard-code any of them, and instead get them from ABI code.  But given that this opcode is ABI-specific, I don't really have a strong opinion here.)

bjorn3: Inlined in all callers.

uweigand: LGTM now, thanks!

cfallin (approved):

PR #9381, opened 2024-10-06
Configure WIT feature gates at runtime & implement wasi-cli `exit-with-code`

badeend:
For extra context, see: https://github.com/bytecodealliance/wasmtime/pull/9276

---

This PR adds the ability to enable/disable WIT APIs gated by `@unstable(feature = ...)` at link-time.

Previously, unstable features could only be controlled at compile-time through the `features` option of `wasmtime::component::bindgen!`. That option has been removed. Bindings for unstable APIs are now generated unconditionally. Instead, the generated `add_to_linker***` functions take an additional `LinkOptions` parameter if the world/interface has any unstable features. The availability of unstable features is now controlled through that.

The component bindgen generates a fresh `LinkOptions` type for each `interface` that needs it and a single `LinkOptions` type for the entire world, which is a union of all the options contained in that world. An example is included in the component-macro tests: [`unstable-features.wit`](https://github.com/badeend/wasmtime/blob/unstable-features/crates/component-macro/tests/codegen/unstable-features.wit) produces:
https://github.com/badeend/wasmtime/compare/245dcab34c2fcae69e0f6c3f2daa15837ea6ff61..4fcfe6f866f36ed663f24199e1744bed3cf0bc2c#diff-ad635965474c257cbafce510463aaf52c1ac813f22168045ae22546d3b3ebfc6

---

A happy coincidence is that WASI-0.2.1 already includes a tiny but practical first use case: `exit-with-code`.
I've implemented that function and is gated behind the `-Scli-exit-with-code` CLI flag.

@sunfishcode, if I interpret https://github.com/WebAssembly/wasi-cli/issues/11 correctly, you were OK with removing the Windows-specific restriction, right? See: https://github.com/badeend/wasmtime/commit/2f48c77867d0bd02c4c910510f491a71da22a0f0


crates/wit-bindgen/src/lib.rs
```diff
@@ -2960,6 +3046,160 @@ impl<'a> RustGenerator<'a> for InterfaceGenerator<'a> {
     }
 }
 
+#[derive(Default)]
+struct LinkOptionsBuilder {
+    unstable_features: HashSet<String>,
```
alexcrichton, line 3051: Could this use a `BTreeSet` or `IndexSet` to ensure it always generates a deterministic order of options?

alexcrichton (approved):
This looks great to me, thanks for this! 

In retrospect though given the large number of changes related to `add_to_linker` what do you think about:

* Interface/world-level `add_to_linker` retains the `LinkOptions` parameter 
* Top-level `add_to_linker` loses the `LinkOptions` parameter, and there's a new top-level `add_to_linker_with_options` function?

Or perhaps this would only affect the helpers we have in `wasmtime-wasi` and `wasmtime-wasi-http`? Basically it looks like there should be a convenience function for "default options" rather than requiring all callers to call it.

crates/wit-bindgen/src/lib.rs
badeend: Definitely

badeend:
> - Interface/world-level add_to_linker retains the LinkOptions parameter
> - Top-level add_to_linker loses the LinkOptions parameter, and there's a new top-level add_to_linker_with_options function?

Don't know what top-level function you're referring to. I think the Interface/world-level add_to_linker functions are the only ones generated by the bindgen?

I have just updated the wasmtime-wasi's functions back to their old signature and added an extra set of `_with_options` variants. I didn't touch the add_to_linker functions generated by bindgen.

alexcrichton (approved):
Ah sorry for my rambling, this is exactly what I was thinking 👍

badeend:
Try again please

badeend:
Sigh.

Try again, again?

PR #9389, opened 2024-10-07
Implement the `array.copy` Wasm GC instruction

fitzgen:
This also involved fixing some `VMGcRef` initialization and GC barrier code for globals. That in turn involved making global initialization fallible, which meant that it needed to be pulled out of `vmctx` initialization and put into instance initialization.

### DIFF
--- a/crates/wasmtime/src/runtime/vm/gc.rs
+++ b/crates/wasmtime/src/runtime/vm/gc.rs
@@ -23,6 +23,7 @@ pub use i31::*;
 use crate::prelude::*;
 use crate::runtime::vm::GcHeapAllocationIndex;
 use core::alloc::Layout;
+use core::mem::MaybeUninit;
 use core::ptr;
 use core::{any::Any, num::NonZeroUsize};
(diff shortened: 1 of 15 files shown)


crates/wasmtime/src/runtime/vm/gc.rs
```diff
@@ -96,6 +97,21 @@ impl GcStore {
...
+    /// slot, performing write barriers as necessary.
+    pub fn init_gc_ref(
+        &mut self,
+        destination: &mut MaybeUninit<Option<VMGcRef>>,
+        source: Option<&VMGcRef>,
+    ) {
+        // Initialize the destination to `None`, at which point the regular GC
+        // write barrier is safe to reuse.
+        destination.write(None);
+        let destination = unsafe { destination.assume_init_mut() };
```
alexcrichton, line 110: Looks like `MaybeUninit::write` returns `&mut T` so I think this can be `let destination = destination.write(None);`

alexcrichton (approved):

PR #9391, opened 2024-10-07
Run the nightly CI job when `pulley` files are touched in a PR

fitzgen:
After https://github.com/bytecodealliance/wasmtime/pull/9251, we will check that Pulley builds with rustc's experimental tail calls feature enabled in the nightly CI job. So if we touch any pulley source files in a PR, we should also run that job in the PR's CI.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -247,7 +247,7 @@ jobs:
           if grep -q debug names.log; then
             echo run-dwarf=true >> $GITHUB_OUTPUT
           fi
-          if grep -q pulley commits.log; then
+          if grep -q pulley names.log; then
             echo test-nightly=true >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -1027,7 +1033,7 @@ jobs:
           - "wasmtime-environ"
           - "pulley-interpreter --all-features"
     needs: determine
-    if: needs.determine.outputs.run-full && github.repository == 'bytecodealliance/wasmtime'
+    if: (needs.determine.outputs.run-full || needs.determine.outputs.test-miri) && github.repository == 'bytecodealliance/wasmtime'
```
alexcrichton, line 1036: This is ok to change to just `test-miri` since `run-full` always implies that `test-miri` is `true` above (most of the other jobs don't check both, they only check the one)

alexcrichton (approved):

PR #9395, opened 2024-10-08
Update WASI WITs to 0.2.2

alexcrichton:
This bumps a number of versions in a number of locations and will be slated for Wasmtime 27. This additionally requires implementing the `network-error-code` function which I've left as a stub for now to get filled out with tests at a later date.


elliottt (approved):

alexcrichton:
cc @badeend on `network-error-code` here, wanna make sure you're aware of it (and the stub impl to-be-filled-in later)

PR #9397, opened 2024-10-08
s390x cleanup: Do not hard-code ABI registers in ISA code

uweigand:
This cleanup removes some places in ISA code where particular register values were hard-coded that actually do not have any particular significance at the ISA level, only in the ABI.

Specifically, the link register for calls is fully determined by the ABI and can be any GPR.  In addition, the MovPReg pseudo instruction works correctly as long as the source is any non-allocatable register - this can be checked directly instead of hardcoding a list of registers.

No functional change intended.


elliottt (approved):
Looks good to me, thanks for cleaning this up Ulrich!

PR #9401, opened 2024-10-08
Implement the `ref.test` Wasm GC instruction

fitzgen:
This commit implements the `ref.test` instruction, which tests whether a reference is of a given type.

We implement inline fast paths for abstract types, but currently rely on an out-of-line libcall for concrete types in the general case. This is known to be suboptimal. (FWIW, we also emit a fast path in front of the libcall where we first check the actual type and expected type for equality and skip the libcall if they are equal.)

This implementation is expected to be improved in the future by exposing a module's types' supertypes arrays to Wasm, so that the Wasm can do the O(1) subtype checks inline. This will make the vast majority of all `ref.test`s inlinable. After that, the only remaining case that would require out-of-line libcalls would be when a module is given an instance of a type that it did not itself define (but which could be a subtype of a type it defined, for example, and which itself might not even have been defined until after this module's instance was created!)


crates/cranelift/src/gc/enabled.rs
```diff
@@ -826,6 +826,237 @@ pub fn translate_array_set(
...
+        } else {
+            is_i31
+        };
+        return Ok(result);
+    }
+
+    // Otherwise, in the general case, we need to inspect our given object's
+    // actual type, which also requires null-checking and i31-checking it.
+
+    let is_any_hierarchy = matches!(
```
alexcrichton, line 887: Could this be replaced with `.top_type() == Something`?

alexcrichton (approved):
Thanks for all the helpful comments!

crates/cranelift/src/gc/enabled.rs
fitzgen: ah yes, good call

PR #9403, opened 2024-10-08
Implement the wide-arithmetic proposal

alexcrichton:
This commit implements the [wide-arithmetic] proposal in Wasmtime. This is a pretty easy proposal to implement due to Cranelift already having support for all the various instructions. The features implemented here are:

* Cranelift support for the four new instructions.
* A new `Config::wasm_wide_arithmetic` option.
* A new `-Wwide-arithmetic` CLI flag.
* A new `wasmtime_config_wasm_wide_arithmetic_set` C API function.
* Support for fuzzing this proposal
  * Generation is implemented in `wasm-smith` .
  * While it's off-by-default in `wasm-smith` it's enabled-by-default here.
  * Differential execution is only possible against Wasmtime right now.
  * Single-instruction module support was added for these new instructions.
* Tests for some simple cases plus randomly-generated tests.
* Example disassemblies for new instructions on Cranelift architectures.

[wide-arithmetic]: https://github.com/WebAssembly/wide-arithmetic

### DIFF
--- a/crates/c-api/include/wasmtime/config.h
+++ b/crates/c-api/include/wasmtime/config.h
@@ -240,6 +240,14 @@ WASMTIME_CONFIG_PROP(void, wasm_multi_memory, bool)
  */
 WASMTIME_CONFIG_PROP(void, wasm_memory64, bool)
 
+/**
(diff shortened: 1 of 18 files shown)


fitzgen (approved):
Very nice!

Issue #9396, opened 2024-10-08
s390x: Crash on accessing backchain after stack overflow


uweigand:
When running `tests/host_segfault.rs` natively on s390x (these tests are skipped in the QEMU based CI), the following sub-tests fail with an unexpected SIGSEGV: 
```
            "overrun 8k with misconfigured host"
            "overrun 32k with misconfigured host"
```

These tests verify the correct behavior after a segmentation fault due to accessing the guard page after a stack overrun.  That actually all works correctly on s390x, but then the signal handler itself crashes.   This is because the `trap_handler` does
```
        let faulting_addr = match signum {
            libc::SIGSEGV | libc::SIGBUS => Some((*siginfo).si_addr() as usize),
            _ => None,
        };
        let regs = get_trap_registers(context, signum);
        let test = info.test_if_trap(regs, faulting_addr, |handler| {
            handler(signum, siginfo, context)
        });

        // Figure out what to do based on the result of this handling of
        // the trap. Note that our sentinel value of 1 means that the
        // exception was handled by a custom exception handler, so we
        // keep executing.
        let jmp_buf = match test {
            TrapTest::NotWasm => {
                if let Some(faulting_addr) = faulting_addr {
                    let start = info.async_guard_range.start;
                    let end = info.async_guard_range.end;
                    if start as usize <= faulting_addr && faulting_addr < end as usize {
                        abort_stack_overflow();
                    }
                }
                return false;
            }
            TrapTest::HandledByEmbedder => return true,
            TrapTest::Trap { jmp_buf } => jmp_buf,
        };
```

Note that it calls `get_trap_registers` *before* checking for the stack overflow case.  That works on other platforms, but on s390x this routine actually accesses the backchain value (our equivalent to the frame pointer used for stack unwinding) - and this is not stored in a register but in the bottom-most word on the stack:

```
        } else if #[cfg(all(target_os = "linux", target_arch = "s390x"))] {
            // On s390x, SIGILL and SIGFPE are delivered with the PSW address
            // pointing *after* the faulting instruction, while SIGSEGV and
            // SIGBUS are delivered with the PSW address pointing *to* the
            // faulting instruction.  To handle this, the code generator registers
            // any trap that results in one of "late" signals on the last byte
            // of the instruction, and any trap that results in one of the "early"
            // signals on the first byte of the instruction (as usual).  This
            // means we simply need to decrement the reported PSW address by
            // one in the case of a "late" signal here to ensure we always
[...]

PR #9423, opened 2024-10-09
s390x: Add support for inline probestack

uweigand:
This implements the gen_inline_probestack callback in line with the implementation on other platforms, which allows detection of stack overflows with misconfigured hosts.

Fixes: https://github.com/bytecodealliance/wasmtime/issues/9396

### DIFF
--- a/crates/wasmtime/src/config.rs
+++ b/crates/wasmtime/src/config.rs
@@ -8,7 +8,6 @@ use core::str::FromStr;
 use serde_derive::{Deserialize, Serialize};
 #[cfg(any(feature = "cache", feature = "cranelift", feature = "winch"))]
 use std::path::Path;
-use target_lexicon::Architecture;
 use wasmparser::WasmFeatures;
 #[cfg(feature = "cache")]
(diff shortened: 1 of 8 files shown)


crates/wasmtime/src/config.rs
```diff
@@ -3024,7 +3024,10 @@ impl PoolingAllocationConfig {
 pub(crate) fn probestack_supported(arch: Architecture) -> bool {
     matches!(
         arch,
-        Architecture::X86_64 | Architecture::Aarch64(_) | Architecture::Riscv64(_)
+        Architecture::X86_64
+            | Architecture::Aarch64(_)
+            | Architecture::Riscv64(_)
+            | Architecture::S390x
     )
 }
```
alexcrichton, line 3032: Mind going ahead and removing this function altogether?

alexcrichton (approved):
Thanks!

crates/wasmtime/src/config.rs
uweigand: Sure, will do.

PR #9434, opened 2024-10-09
Update docs/stabilization of wasm proposals

alexcrichton:
This commit updates our documentation and documented status of various WebAssembly proposals. The goal of this commit it to explicitly list all requirements for each WebAssembly proposal in a way such that enabling a feature requires actively checking this table and performing effort to fill in. This is intended to mitigate GHSA-q8hx-mm92-4wvg where it was found that we were accidentally not fuzzing tail calls but had enabled it by default.

This shuffles around some documentation, modernizes a bit, and notably tries to list out a green checkbox or red X for the status of various proposals. Notes for "holes" in the "on by default matrix" are also added.

### DIFF
--- a/crates/wasmtime/src/config.rs
+++ b/crates/wasmtime/src/config.rs
@@ -1823,10 +1823,16 @@ impl Config {
         // On-by-default features that wasmtime has. Note that these are all
         // subject to the criteria at
         // https://docs.wasmtime.dev/contributing-implementing-wasm-proposals.html
+        // and
+        // https://docs.wasmtime.dev/stability-wasm-proposals.html
         features |= WasmFeatures::MULTI_MEMORY;
         features |= WasmFeatures::RELAXED_SIMD;
         features |= WasmFeatures::TAIL_CALL;
         features |= WasmFeatures::EXTENDED_CONST;
+        // NB: if you add a feature above this line please double-check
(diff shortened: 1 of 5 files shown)


elliottt (approved):
This looks good to me!

I wonder if there's a good place for documentation about determining when fuzzing is good enough. Perhaps something as simple as manually injecting `todo!()` into new implementations so that running the fuzzer locally will panic when exercising the new behavior?

alexcrichton:
We had a bunch of [discussion today](https://github.com/bytecodealliance/meetings/blob/main/wasmtime/2024/wasmtime-10-10.md) about this as well, but at least by my read there were no firm conclusions. I'm going to try to document better The One Location in fuzzing where proposals are enabled/disabled but beyond that while there was also discussion in the meeting about automatically probing coverage for fuzzers we didn't reach a firm conclusion about how best to do this.

In the meantime I'm going to go ahead and merge this and I'll follow-up with more minor improvements too.

cfallin:
> We had a bunch of [discussion today](https://github.com/bytecodealliance/meetings/blob/main/wasmtime/2024/wasmtime-10-10.md) about this as well, but at least by my read there were no firm conclusions.

Thanks for bringing this up in discussion (and these changes!). To further the discussions around testing our fuzzing, I filed #9449 just now.

PR #9437, opened 2024-10-10
Implement the `ref.cast` Wasm GC instruction

fitzgen:
Just building on top of our existing `ref.test` support and trapping if the test fails.

### DIFF
--- a/crates/cranelift/src/translate/code_translator.rs
+++ b/crates/cranelift/src/translate/code_translator.rs
@@ -2755,6 +2755,34 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
             )?;
             state.push1(result);
         }
+        Operator::RefCastNonNull { hty } => {
+            let r = state.pop1();
+            let heap_type = environ.convert_heap_type(*hty);
+            let cast_okay = environ.translate_ref_test(
+                builder,
+                WasmRefType {
+                    heap_type,
+                    nullable: false,
+                },
+                r,
+            )?;
+            environ.trapz(builder, cast_okay, crate::TRAP_CAST_FAILURE);
+            state.push1(r);
+        }
+        Operator::RefCastNullable { hty } => {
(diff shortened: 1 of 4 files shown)


crates/cranelift/src/translate/code_translator.rs
```diff
@@ -2755,6 +2755,34 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
...
+                builder,
+                WasmRefType {
+                    heap_type,
+                    nullable: true,
+                },
+                r,
+            )?;
+            environ.trapz(builder, cast_okay, crate::TRAP_CAST_FAILURE);
+            state.push1(r);
+        }
```
alexcrichton, line 2785: I can sort of see how this is going to get used for `BrOnCast` and such as well. Would it make sense to update `translate_ref_test` with an argument along the lines of "what to do as the conclusion"? For example here the `trapz` could get propagated upwards to avoid creating so many basic blocks as part of `translate_ref_test`. For `BrOnCast` and friend it could be used for "I already created the blocks and I'm telling you where to jump".

Basically there's cleanups we can do here in the frontend which Cranelift will not do in the mid-end. I don't know whether the better answer is to invest in the mid-end, assume that `MachBuffer` cleans up everything, or try to produce better IR here in the first place.

alexcrichton (approved):

crates/cranelift/src/translate/code_translator.rs
alexcrichton: Yeah I just fear that we've talked for a very long time about theoretical mid-end cleanups to basic blocks but haven't gotten to them, so it seems like it's probably easier to do it here than the mid-end.

fitzgen: Yeah, quite possible it is easier in the front end for now. Either way, not doing it as part of this PR :)

PR #9446, opened 2024-10-10
Enable some miscellaneous Wasm GC spec tests; fix module import/export linking and subtyping

fitzgen:

### DIFF
--- a/tests/wast.rs
+++ b/tests/wast.rs
@@ -203,15 +203,8 @@ fn should_fail(test: &Path, strategy: Strategy) -> bool {
             }
         }
     }
-    let unsupported_gc_tests = [
-        "binary_gc.wast",
-        "table_sub.wast",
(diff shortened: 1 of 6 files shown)


alexcrichton (approved):

tests/wast.rs
```diff
@@ -203,15 +203,8 @@ fn should_fail(test: &Path, strategy: Strategy) -> bool {
             }
         }
     }
-    let unsupported_gc_tests = [
-        "binary_gc.wast",
```
alexcrichton, line 207: Looks like some of these were because there weren't any actual files of that name, so must have been from some historical version of the gc proposal's spec tests or something like that?

fitzgen: I think maybe some didn't have any assertions either? I don't know

Issue #9351, opened 2024-10-01
List of future optimizations for Wasm GC performance


fitzgen:
* [ ] Speed up getting and setting `funcref`s inside GC objects: https://github.com/bytecodealliance/wasmtime/issues/9347
* [ ] Const fold through `uadd_overflow_trap`/`trap{z,nz}` insts that show up frequently in GC heap accesses: https://github.com/bytecodealliance/wasmtime/issues/5908
* [ ] Type-based alias analysis for Wasm GC: https://github.com/bytecodealliance/wasmtime/issues/9348
* [ ] Escape analysis for Wasm GC: https://github.com/bytecodealliance/wasmtime/issues/9349
* [ ] Reuse linear memories and their optimized bounds checks for Wasm GC heaps: https://github.com/bytecodealliance/wasmtime/issues/9350

fitzgen:
See also https://github.com/WebAssembly/gc/issues/565 for potential benchmarks.

fitzgen:
Closing this meta issue since all work is already captured in sub-issues and they're all labeled appropriately.

PR #9448, opened 2024-10-10
Implement subtype checking for `[return_]call_indirect` instructions

fitzgen:
When Wasm GC is enabled, the `[return_]call_indirect` instructions must do full subtyping checks, rather than simple strict equality type checks.

This adds an additional branch and slow path to indirect calls, so we only emit code for this check when Wasm GC is enabled, even though it would otherwise be correct to always emit it (because the `is_subtype` check would always fail for non-equal types, since there is no subtyping before Wasm GC).

### DIFF
--- a/crates/cranelift/src/func_environ.rs
+++ b/crates/cranelift/src/func_environ.rs
@@ -1495,8 +1495,15 @@ impl<'a, 'func, 'module_env> Call<'a, 'func, 'module_env> {
         // Check that they match: in the case of Wasm GC, this means doing a
         // full subtype check. Otherwise, we do a simple equality check.
         let matches = if features.gc() {
-            self.env
-                .is_subtype(self.builder, callee_sig_id, caller_sig_id)
+            #[cfg(feature = "gc")]
+            {
+                self.env
+                    .is_subtype(self.builder, callee_sig_id, caller_sig_id)
+            }
+            #[cfg(not(feature = "gc"))]
+            {
+                unreachable!()
+            }
         } else {
             self.builder
                 .ins()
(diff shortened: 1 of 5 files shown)


crates/cranelift/src/func_environ.rs
```diff
@@ -1402,33 +1407,40 @@ impl<'a, 'func, 'module_env> Call<'a, 'func, 'module_env> {
...
-                            funcref_ptr,
-                            i32::from(self.env.offsets.ptr.vm_func_ref_type_index()),
-                        );
-                    } else {
-                        self.env.trapz(
-                            self.builder,
-                            funcref_ptr,
-                            crate::TRAP_INDIRECT_CALL_TO_NULL,
-                        );
+                if features.gc() {
```
alexcrichton, line 1410: Should this be `features.function_references() || features.gc()`? Or just `features.function_references()`?

alexcrichton (commented):
Mind filing an issue for this performance issue as well? This to me is going to be a hard blocker for enabling GC by default because a failed subtype check will involve acquiring a global lock even for non-GC modules which may not be appropriate for all embedders.

alexcrichton (approved):

crates/cranelift/src/func_environ.rs
fitzgen: I believe just `gc`

fitzgen:
> Mind filing an issue for this performance issue as well? This to me is going to be a hard blocker for enabling GC by default because a failed subtype check will involve acquiring a global lock even for non-GC modules which may not be appropriate for all embedders.

Added an item to https://github.com/bytecodealliance/wasmtime/issues/9351

I think much of this is equivalent to making `ref.test` fast (i.e. expose the supertypes arrays to wasm so that we can do the O(1) subtype check inline, rather than behind a lock) but even after we address that, it is another branch to another slow path to do a full subtype check when the types don't exactly match.

PR #9450, opened 2024-10-10
Centralize wasm-features-in-fuzzing a bit more

alexcrichton:
Move everything into the same location about what features are enabled, which are disabled, and such. Also statically enumerate all the features that wasm-smith enables for us and acknowledge that here too.

### DIFF
--- a/crates/fuzzing/src/generators/module.rs
+++ b/crates/fuzzing/src/generators/module.rs
@@ -16,27 +16,42 @@ impl<'a> Arbitrary<'a> for ModuleConfig {
     fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<ModuleConfig> {
         let mut config = wasm_smith::Config::arbitrary(u)?;
 
+        // This list is intended to be the definintive source of truth for
+        // what's at least possible to fuzz within Wasmtime. This is a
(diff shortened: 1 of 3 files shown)


crates/fuzzing/src/generators/module.rs
```diff
@@ -16,27 +16,42 @@ impl<'a> Arbitrary<'a> for ModuleConfig {
...
+        config.exceptions_enabled = false;
+        config.gc_enabled = false;
+        config.wide_arithmetic_enabled = u.arbitrary()?;
+        config.memory64_enabled = u.ratio(1, 20)?;
+        // Allow the threads proposal if memory64 is not already enabled. FIXME:
+        // to allow threads and memory64 to coexist, see
+        // https://github.com/bytecodealliance/wasmtime/issues/4267.
+        config.threads_enabled = !config.memory64_enabled && u.ratio(1, 20)?;
+        // FIXME: this may be safe to enable
+        config.custom_page_sizes_enabled = false;
```
alexcrichton, line 46: cc @fitzgen on this point, I left this as it was before but you might be interested in changing this to `u.arbitrary()?` to have it on-by-default in fuzzing. (will probably require disabling it in other differential fuzzers and maybe winch too)

fitzgen (approved):

Issue #9449, opened 2024-10-10
Wasmtime fuzzing: manually inject faults in a branch to verify fuzzing end-to-end


cfallin:
In the [Wasmtime meeting today](https://github.com/bytecodealliance/meetings/blob/main/wasmtime/2024/wasmtime-10-10.md), there was a lot of good discussion about our procedures and standards around fuzzing, and how to make process improvements to make sure fuzzing catches issues in on-by-default features. I thought I'd expand a bit on an idea I proposed there.

The basic problem is: the configuration space of Wasm features that get fuzzed is the result of a fairly nontrivial interaction between feature selection logic in multiple places and crates. There are plausible reasons why this isn't all centralized; even if we do eventually get to a point where there is One Main Function that determines what is fuzzed, there is still the question of whether the plumbing that flows outward from that definition faithfully preserves the feature.

I'm suggesting as a principle that the only way to *really* know whether fuzzing will catch bugs is to... make fuzzing catch bugs. An "integration test" wrapped around our whole setup, as it were. The idea (which I'll maybe call a "fire drill" approach) is:

- Have a branch in the repo (`fuzzing-firedrill`) that has `main` at some recent point plus one commit.
- Update our fuzzing configuration to run one or several targets on this branch as well as `main` in oss-fuzz.
- As a weekly/monthly/?? chore, recreate this branch by taking current `main` and injecting one fault. Perhaps we can keep a list/queue of ideas. They should be simple changes -- one-line commits that invert a bit or conditional, or delete a register update, or ... Ideally, this is ~10 minutes per week of work.
- Wait for oss-fuzz to find the issue. Revert the fault commit (or move to the next fault?) when it does. If it doesn't, find out why.

The main idea is that we want to test the whole pipeline all the way to oss-fuzz-email-in-inboxes. Nothing else tests this whole pipeline today. We simply don't know if it works, except when we get emails. The idea here is to turn the email rate up from "0 + actual errors" to "fault-injected stream + actual errors", so it's always nonzero, so we can keep verifying it works.

There were a few objections or alternative ideas in discussions today:

- Suggestions to do automated mutation testing (the meeting notes above describe the proposal as mutation testing even). I don't think this is quite the same thing: the idea is explicitly *not* to have automation (which could fail), but to simulate a human making a faulty commit. It's like a firedrill: do everything except actually set the fire (put the commit on `main`).
[...]

PR #9452, opened 2024-10-10
Add a test that fuzzers generate wasm proposals

alexcrichton:
This commit adds a test to the `wasmtime-fuzzing` crate as a sanity-check that eventually a module is generated requiring all of the features that wasmtime supports. This is intended to be another double-check in the process of enabling a proposal in Wasmtime by ensuring that the feature is added to this list which then transitively requires that fuzzing eventually generates a module needing that feature.

cc #9449

### DIFF
--- a/crates/fuzzing/src/oracles.rs
+++ b/crates/fuzzing/src/oracles.rs
@@ -1166,6 +1166,28 @@ pub fn call_async(wasm: &[u8], config: &generators::Config, mut poll_amts: &[u32
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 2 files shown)


alexcrichton:
One thing I hope this can help guarantee is that if a feature is not listed in the list here then we're basically guaranteed we're not fuzzing it in the general-purpose fuzzers. That's a quick-and-easy check of "when stabilizing make sure the feature is in this list" because if not it doesn't check the fuzzing checkbox.

fitzgen (approved):
Nice! Low effort but (hopefully) big long-term payoff.

PR #9457, opened 2024-10-11
Change how env vars work in `differential` fuzzer

alexcrichton:
This commit updates the processing of the `ALLOWED_*` environment variables to work differently than before. Previously the list of engines and module-generation-strategies were filtered based on `ALLOWED_*` environment variables but this meant that the meaning of a fuzz input changed depending on environment variables. This commit instead replaces the handling of these environment variables to ensure that the meaning of the fuzz input doesn't change depending on their values. Instead fuzz test cases are early-thrown-out if they request an engine that's disabled or a module-generation-strategy that's disabled.

### DIFF
--- a/fuzz/fuzz_targets/differential.rs
+++ b/fuzz/fuzz_targets/differential.rs
@@ -76,7 +76,7 @@ fn execute_one(data: &[u8]) -> Result<()> {
     let lhs = match *u.choose(&allowed_engines)? {
         Some(engine) => engine,
         None => {
-            log::debug!("test case uses a runtime-disabled engien");
+            log::debug!("test case uses a runtime-disabled engine");
             return Ok(());
         }
     };
--- a/fuzz/fuzz_targets/differential.rs
+++ b/fuzz/fuzz_targets/differential.rs
@@ -22,8 +22,8 @@ static SETUP: Once = Once::new();
 // - ALLOWED_ENGINES=wasmi,spec cargo +nightly fuzz run ...
 // - ALLOWED_ENGINES=-v8 cargo +nightly fuzz run ...
 // - ALLOWED_MODULES=single-inst cargo +nightly fuzz run ...
-static ALLOWED_ENGINES: Mutex<Vec<&str>> = Mutex::new(vec![]);
-static ALLOWED_MODULES: Mutex<Vec<&str>> = Mutex::new(vec![]);
+static ALLOWED_ENGINES: Mutex<Vec<Option<&str>>> = Mutex::new(vec![]);
+static ALLOWED_MODULES: Mutex<Vec<Option<&str>>> = Mutex::new(vec![]);
 
 // Statistics about what's actually getting executed during fuzzing
 static STATS: RuntimeStats = RuntimeStats::new();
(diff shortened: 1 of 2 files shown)


fuzz/fuzz_targets/differential.rs
```diff
@@ -73,7 +73,13 @@ fn execute_one(data: &[u8]) -> Result<()> {
     // Choose an engine that Wasmtime will be differentially executed against.
     // The chosen engine is then created, which might update `config`, and
     // returned as a trait object.
-    let lhs = u.choose(&allowed_engines)?;
+    let lhs = match *u.choose(&allowed_engines)? {
+        Some(engine) => engine,
+        None => {
+            log::debug!("test case uses a runtime-disabled engien");
```
elliottt, line 79:
```suggestion
            log::debug!("test case uses a runtime-disabled engine");
```

fuzz/fuzz_targets/differential.rs
```diff
@@ -100,8 +106,12 @@ fn execute_one(data: &[u8]) -> Result<()> {
         panic!("unable to generate a module to fuzz against; check `ALLOWED_MODULES`")
     }
     let wasm = match *u.choose(&allowed_modules)? {
-        "wasm-smith" => build_wasm_smith_module(&mut u, &config)?,
-        "single-inst" => build_single_inst_module(&mut u, &config)?,
+        Some("wasm-smith") => build_wasm_smith_module(&mut u, &config)?,
+        Some("single-inst") => build_single_inst_module(&mut u, &config)?,
+        None => {
+            log::debug!("test case uses a runtime-disabled module strategy");
+            return Ok(());
```
elliottt, line 113: Do we have a way to categorize the early return cases, or can we find out if the fuzzer is spending all its time in unproductive configurations?

fuzz/fuzz_targets/differential.rs
fitzgen, line 79:
```suggestion
            log::debug!("test case uses a runtime-disabled engine");
```

fitzgen (approved):

fuzz/fuzz_targets/differential.rs
cfallin: In the regalloc.rs fuzzer we used atomics to count stats (iirc similar thing, rejected cases) and the fuzz case that got a value 0 mod 1000 from the fetch-and-add printed a progress message with all stats -- this at least gave us a general sense watching interactively. I don't think there's a way to get such output from oss-fuzz though. IIRC there is a way to return a status to libFuzzer that tells it specifically the case is uninteresting?

abrown: It might not be clear in this diff but `RuntimeStats` is what does that here.

alexcrichton: Can confirm what @abrown mentioned should work and it's the same idea as regalloc's fuzzers @cfallin mentions. 

AFAIK though there's no way to tell libFuzzer something is uninteresting, but by not doing anything you're also saying "this introduces no new coverage" which means it should naturally get pruned and deduplicated from the corpus over time.

fitzgen: > AFAIK though there's no way to tell libFuzzer something is uninteresting, but by not doing anything you're also saying "this introduces no new coverage" which means it should naturally get pruned and deduplicated from the corpus over time.

There actually is: if you return this https://docs.rs/libfuzzer-sys/latest/libfuzzer_sys/enum.Corpus.html#variant.Reject from a fuzz target, then it tells libfuzzer not to add that testcase into the corpus or reuse it for future mutations or whatever.

alexcrichton: Whoa fancy!

PR #9470, opened 2024-10-15
Move JIT debug image registration to CodeMemory

### LINKED ISSUE #9461: Duplicate JIT debug module registrations with components
Reproduction:
1) Compile any component, e. g.
```cpp
int main()
{
    int a = 1;
    return a;
}
```
```ps
> & $env:WASI_SDK_PATH/bin/clang main.cpp -o main.wasm -g3 -target wasm32-unknown-wasip2
```
2) Debug with lldb+wasmtime:
```
> lldb wasmtime -- -D debug-info -O opt-level=0 main.wasm
> (lldb) b main
> (lldb) c ; Until you reach WASM's 'main'
```
3) Observe four `(JIT)` modules (for each core module in a component). All of them have the exact same content.
```
> (lldb) image list
```
4) Also observe that LLDB needed to index the same DWARF info four times. This means that "startup with debugging" is 4x slower if we exclude the time to compile code and transform DWARF.

The bug is somewhere around here:
https://github.com/bytecodealliance/wasmtime/blob/460a4c0ad5d3655b7dbaec193e726196de848898/crates/wasmtime/src/runtime/instantiate.rs#L72-L79

AFAICT, with components, the generated image is a shared (among core modules) resource, so each `Module::from_parts_raw` call here:
https://github.com/bytecodealliance/wasmtime/blob/460a4c0ad5d3655b7dbaec193e726196de848898/crates/wasmtime/src/runtime/component/component.rs#L408-L414

Registers the same image.

SingleAccretion:
JIT images correspond to ELF images, which may represent multiple modules within a single component.

Fixes #9461.

Tested manually on a moderately large component as well as with the existing LLDB tests.

### DIFF
--- a/crates/wasmtime/src/runtime/code_memory.rs
+++ b/crates/wasmtime/src/runtime/code_memory.rs
@@ -19,8 +19,12 @@ pub struct CodeMemory {
     // dropped first since it refers to memory owned by `mmap`.
     mmap: ManuallyDrop<MmapVec>,
     unwind_registration: ManuallyDrop<Option<UnwindRegistration>>,
+    #[cfg(feature = "debug-builtins")]
+    debug_registration: ManuallyDrop<Option<crate::runtime::vm::GdbJitImageRegistration>>,
     published: bool,
     enable_branch_protection: bool,
+    #[cfg(feature = "debug-builtins")]
+    has_native_debug_info: bool,
 
     relocations: Vec<(usize, obj::LibCall)>,
 
@@ -32,7 +36,7 @@ pub struct CodeMemory {
     address_map_data: Range<usize>,
     func_name_data: Range<usize>,
     info_data: Range<usize>,
-    dwarf: Range<usize>,
+    wasm_dwarf: Range<usize>,
 }
 
 impl Drop for CodeMemory {
@@ -40,6 +44,7 @@ impl Drop for CodeMemory {
         // Drop `unwind_registration` before `self.mmap`
         unsafe {
             ManuallyDrop::drop(&mut self.unwind_registration);
+            ManuallyDrop::drop(&mut self.debug_registration);
             ManuallyDrop::drop(&mut self.mmap);
         }
     }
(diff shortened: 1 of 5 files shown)


tests/all/debug/lldb.rs
```diff
@@ -63,10 +63,6 @@ fn check_lldb_output(output: &str, directives: &str) -> Result<()> {
 
 #[test]
 #[ignore]
-#[cfg(all(
-    any(target_os = "linux", target_os = "macos"),
```
SingleAccretion, line 67: Since these tests are `#[ignore]`d by default, and run fine on Windows when explicitly requested, it seems there is no reason to have these guards anymore.

alexcrichton (approved):
Thanks for the PR as well as for the issue! This looks like the right fix to me. I suspect that writing a test for this will be relatively hard, so otherwise having the preexisting tests pass I think is sufficient for this.

I also think that the bug you highlighted where we're double-registering with the native profiler is probably also an issue as well. If you'd like that's ok to defer to a second PR to fix, or if you'd like bundling it in here I think would be reasonable too

SingleAccretion:
> I also think that the bug you highlighted where we're double-registering with the native profiler is probably also an issue as well. 

Opened #9478 about this.

crates/wasmtime/src/runtime/code_memory.rs
```diff
@@ -19,8 +19,12 @@ pub struct CodeMemory {
     // dropped first since it refers to memory owned by `mmap`.
     mmap: ManuallyDrop<MmapVec>,
     unwind_registration: ManuallyDrop<Option<UnwindRegistration>>,
+    #[cfg(feature = "debug-builtins")]
+    debug_registration: ManuallyDrop<Option<crate::runtime::vm::GdbJitImageRegistration>>,
```
alexcrichton, line 23: Reading over this again, mind removing the `ManualllyDrop` wrapper here? This is already paying the cost of the `Option` overhead so I don't think the unsafe drop part is buying a whole lot here.

The destructor can still be early executed through `let _ = self.debug_registration.take();` in the `Drop` implementation though.

(I realize you were probably copying `unwind_registration` and I think we should apply this change there too, as a separate PR if you'd prefer to leave that as-is for now)

SingleAccretion: While we're on the topic - I don't quite get this manual `drop`ping. Wouldn't reordering `mmap` to be after the registrations work equivalently?

alexcrichton: I think so yeah, but the drop order of fields with respect to field orderings in structs is pretty subtle so we've occasionally used `ManuallyDrop` to make it more clear that the drop order changing is an intentional change. 

Once both `debug_registration` and `unwind_registration` are just `Option` though it would be fine to remove all the `ManuallyDrop` here and the `Drop` implementation would just be `let _ = self.field.take()` for the two registrations.

Issue #8090, opened 2024-03-12
winch: Fuel based interruption


saulecabrera:
Winch is missing support for some of Wasmtime's features. The objective of this issue is to track progress of adding fuel based interruption to Winch.

PR #9472, opened 2024-10-15
winch: Implement Fuel-Based Interruption

saulecabrera:
Closes: https://github.com/bytecodealliance/wasmtime/issues/8090

This commit introduces the initial implementation of fuel-based interruption in Winch.

To maintain consistency with existing fuel semantics, this implementation closely follows the Wasmtime/Cranelift approach, with the following exception:

* Local Fuel Cache: Given Winch's emphasis on compilation speed, this implementation does not optimize for minimizing loads and stores. As a result, checking and incrementing fuel currently requires explicit loads and stores. Future optimizations may be considered to improve this aspect.

This commit also includes a small refactoring in the visitor, which introduces more generic "visitor hook" which enable handling the invariants that need to happen before and after emitting machine code for each Wasm operator.

### DIFF
--- a/winch/codegen/src/codegen/env.rs
+++ b/winch/codegen/src/codegen/env.rs
@@ -56,13 +56,22 @@ pub enum HeapStyle {
     Dynamic,
 }
 
+/// Resolved offsets to the VM limits.
+pub struct VMRuntimeLimits {
+    /// The offset to the `VMRuntimeLimits` pointer, relative to the `VMContext`
+    /// pointer.
+    pub(crate) ptr_offset: u32,
+    /// The offset to the fuel consumed field in the `VMRuntimeLimits` pointer.
+    pub(crate) fuel_consumed_offset: u32,
+}
+
 /// Heap metadata.
 ///
 /// Heaps represent a WebAssembly linear memory.
 #[derive(Debug, Copy, Clone)]
 pub struct HeapData {
     /// The offset to the base of the heap.
-    /// Relative to the VMContext pointer if the WebAssembly memory is locally
+    /// Relative to the `VMContext` pointer if the WebAssembly memory is locally
     /// defined. Else this is relative to the location of the imported WebAssembly
     /// memory location.
     pub offset: u32,
@@ -136,6 +145,8 @@ pub struct FuncEnv<'a, 'translation: 'a, 'data: 'translation, P: PtrSize> {
     table_access_spectre_mitigation: bool,
     name_map: PrimaryMap<UserExternalNameRef, UserExternalName>,
     name_intern: HashMap<UserExternalName, UserExternalNameRef>,
+    /// Metadata about the offset to and fields of `VMRuntimeLimits`.
+    vmruntime_limits: Option<VMRuntimeLimits>,
 }
 
 pub fn ptr_type_from_ptr_size(size: u8) -> WasmValType {
@@ -169,6 +180,7 @@ impl<'a, 'translation, 'data, P: PtrSize> FuncEnv<'a, 'translation, 'data, P> {
             builtins,
             name_map: Default::default(),
             name_intern: Default::default(),
+            vmruntime_limits: None,
         }
     }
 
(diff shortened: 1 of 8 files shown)


winch/codegen/src/codegen/env.rs
```diff
@@ -386,6 +398,15 @@ impl<'a, 'translation, 'data, P: PtrSize> FuncEnv<'a, 'translation, 'data, P> {
         self.name_intern.clear();
         mem::take(&mut self.name_map)
     }
+
+    /// Lazily calculates and returns [`VMRuntimeLimits`].
+    pub fn resolve_vmruntime_limits(&mut self) -> &VMRuntimeLimits {
+        self.vmruntime_limits
+            .get_or_insert_with(|| VMRuntimeLimits {
+                ptr_offset: self.vmoffsets.ptr.vmctx_runtime_limits() as u32,
+                fuel_consumed_offset: self.vmoffsets.ptr.vmruntime_limits_fuel_consumed() as u32,
```
alexcrichton, line 407: Out of curiosity was the `vmruntime_limits` cache-of-sorts here added because the original methods were too slow? Or because the original `self.vmoffsets.ptr` wasn't accessible when the cache was needed?

winch/codegen/src/codegen/mod.rs
```diff
@@ -188,7 +203,7 @@ where
 
     pub fn handle_unreachable_end(&mut self) {
         let mut frame = self.control_frames.pop().unwrap();
-        // We just popped the outermost block.
+        // We j;ust popped the outermost block.
```
alexcrichton, line 206: stray `;`

alexcrichton (approved):
Looks reasonable to me, tahnks for this!

One thought I had reading over this was that this should probably update some documentation somewhere around enabling fuel that Winch is supported. I don't think that's quite right though since we don't exhaustively document Winch's support on all other `Config` options, so documenting this as a one-off doesn't feel right. That being said it led me to another thought though that there was no change here associated with "ok actually allow Winch to use fuel". This means that it looks like we were silently accepting fuel-enabled-`Config` even though Winch didn't support it.

To handle that, either here or as a follow-up, mind updating the `Compiler for ...` in winch-codegen? IIRC there's a method `set_tunables` and it might be good to have Winch-specific logic to reject configurations that Winch doesn't support. For example Winch doesn't support `generate_native_debuginfo` or `epoch_interruption`. This might help move some code out of `Config::validate` as well and keep it in a Winch-specific location.

winch/codegen/src/codegen/env.rs
saulecabrera: Not really, I didn't perform extensive benchmarking here, I think it might be fine to immediately resolve the limits when constructing `FuncEnv`.

saulecabrera:
Thanks for the review @alexcrichton.

> One thought I had reading over this was that this should probably update some documentation somewhere around enabling fuel that Winch is supported. I don't think that's quite right though since we don't exhaustively document Winch's support on all other Config options, so documenting this as a one-off doesn't feel right. That being said it led me to another thought though that there was no change here associated with "ok actually allow Winch to use fuel". This means that it looks like we were silently accepting fuel-enabled-Config even though Winch didn't support it.

I think it's reasonable to audit the `Config` options and be explicit in the documentation regarding those options that are not fully supported by Winch. I'll do that in a separate PR paired with the change to perform winch-specific configuration validation in `set_tunables` in Winch's compiler implementation.

winch/codegen/src/codegen/env.rs
alexcrichton: Ah ok makes sense, I ask because it felt a bit unidiomatic to copy a few fields from `VMOffsets` into a separate structure vs threading around the `VMOffsets` themselves, but it's also a pretty minor point

saulecabrera:
@alexcrichton I've opened https://github.com/bytecodealliance/wasmtime/pull/9490 -- once https://github.com/bytecodealliance/wasmtime/pull/9490 lands, I'll rebase and land this change.

PR #9476, opened 2024-10-16
A couple minor Wasm GC codegen improvements

fitzgen:
1. Don't bounds-check GC refs that are returned from trusted libcalls
2. Mark loading the DRC collector's activations table pointer from the `vmctx` as `readonly`

Also a minor refactor to emitting code to get the GC heap base and bound that resulted in a mechanical value renumbering in the GC disas tests. See each commit in isolation for better diffs.

### DIFF
--- a/crates/cranelift/src/gc/enabled.rs
+++ b/crates/cranelift/src/gc/enabled.rs
@@ -1119,8 +1119,8 @@ impl FuncEnvironment<'_> {
         self.gc_layout(type_index).unwrap_struct()
     }
 
-    /// Get the GC heap's base pointer and bound.
-    fn get_gc_heap_base_bound(&mut self, builder: &mut FunctionBuilder) -> (ir::Value, ir::Value) {
+    /// Get the GC heap's base pointer.
+    fn get_gc_heap_base(&mut self, builder: &mut FunctionBuilder) -> ir::Value {
         let ptr_ty = self.pointer_type();
         let flags = ir::MemFlags::trusted().with_readonly();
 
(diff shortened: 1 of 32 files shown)


crates/cranelift/src/gc/enabled.rs
```diff
@@ -1130,12 +1130,27 @@ impl FuncEnvironment<'_> {
         let base_offset = self.offsets.ptr.vmctx_gc_heap_base();
         let base_offset = i32::from(base_offset);
 
+        builder.ins().load(ptr_ty, flags, vmctx, base_offset)
+    }
+
+    /// Get the GC heap's bound.
+    fn get_gc_heap_bound(&mut self, builder: &mut FunctionBuilder) -> ir::Value {
+        let ptr_ty = self.pointer_type();
+        let flags = ir::MemFlags::trusted().with_readonly();
```
alexcrichton, line 1139: Is the `with_readonly` here going to be accurate in the future? If I understand this right it's akin to `memory.size` which is readonly if `memory.grow` can't be used, but if we support for example more dynamic configurations of GC heaps in the future will this still be true?

crates/cranelift/src/gc/enabled.rs
```diff
@@ -1119,8 +1119,8 @@ impl FuncEnvironment<'_> {
         self.gc_layout(type_index).unwrap_struct()
     }
 
-    /// Get the GC heap's base pointer and bound.
-    fn get_gc_heap_base_bound(&mut self, builder: &mut FunctionBuilder) -> (ir::Value, ir::Value) {
+    /// Get the GC heap's base pointer.
+    fn get_gc_heap_base(&mut self, builder: &mut FunctionBuilder) -> ir::Value {
         let ptr_ty = self.pointer_type();
         let flags = ir::MemFlags::trusted().with_readonly();
```
alexcrichton, line 1125: Similar to below, is this `with_readonly` going to be true in the face of a dynamic GC heap that can move in the future?

alexcrichton (approved):

crates/cranelift/src/gc/enabled.rs
fitzgen: This code will all be refactored quite a bit for when we support backing GC heaps with linear memories and the static vs dynamic knobs and all that. This is really just for in the meantime, and to split out from some other stuff I am working on.

PR #9477, opened 2024-10-16
Add permissions to `wasi_config_preopen_dir` C API

ajalt:
The current `wasi_config_preopen_dir` function does not expose the `dir_perms` and `file_perms` parameters that were added to `preopened_dir`. This commit adds them and updates the docs for those functions to reflect the current signature.

This is a prerequisite for https://github.com/bytecodealliance/wasmtime-py/issues/251

This change isn't backwards compatible, so please let me know if there's anything I need to do differently.

### DIFF
--- a/crates/c-api/include/wasi.h
+++ b/crates/c-api/include/wasi.h
@@ -171,12 +171,27 @@ WASI_API_EXTERN void wasi_config_inherit_stderr(wasi_config_t *config);
  * filesystem, but only that directory (its whole contents but nothing above
  * it).
  *
- * The `path` argument here is a path name on the host filesystem, and
+ * The `host_path` argument here is a path name on the host filesystem, and
  * `guest_path` is the name by which it will be known in wasm.
+ *
+ * The `dir_perms` argument is the permissions that wasm will have to operate on
+ * `guest_path`. This can be used, for example, to provide readonly access to a
+ * directory. This argument is a bitmask with the following flag values:
+ * - 0b01: This directory can be read, for example its entries can be iterated
+           over and files can be opened.
+ * - 0b10: This directory can be mutated, for example by creating new files within it.
+ *
+ * The `file_perms` argument is similar to `dir_perms` but corresponds to the maximum
+ * set of permissions that can be used for any file in this directory.
+ * This argument is a bitmask with the following flag values:
+ * - 0b01: This file can be read.
+ * - 0b10: This file can be written to.
  */
 WASI_API_EXTERN bool wasi_config_preopen_dir(wasi_config_t *config,
-                                             const char *path,
-                                             const char *guest_path);
+                                             const char *host_path,
+                                             const char *guest_path,
+                                             size_t dir_perms,
+                                             size_t file_perms);
 
 #undef own
 
--- a/crates/c-api/src/wasi.rs
+++ b/crates/c-api/src/wasi.rs
@@ -172,6 +172,8 @@ pub unsafe extern "C" fn wasi_config_preopen_dir(
     config: &mut wasi_config_t,
     path: *const c_char,
     guest_path: *const c_char,
+    dir_perms: usize,
+    file_perms: usize,
 ) -> bool {
     let guest_path = match cstr_to_str(guest_path) {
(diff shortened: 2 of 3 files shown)


alexcrichton (commented):
Thanks for this! Could you also add some `#define` entries for the various flags here? Perhaps a flag-per-bit and a `*_ALL` define which has everything OR'd together?

Also I think it might be best to check for unknown bits in the method itself. For example using `from_bits` and translating the optional return value into an error being returned.

alexcrichton:
Oh sorry, forgot to mention, but:

> This change isn't backwards compatible, so please let me know if there's anything I need to do differently.

That's ok, so no worries! We don't provide an ironclad guarantee of stability in the C API at this time, so it's ok to change it.

crates/c-api/include/wasi.h
```diff
@@ -163,6 +163,28 @@ WASI_API_EXTERN bool wasi_config_set_stderr_file(wasi_config_t *config,
  */
 WASI_API_EXTERN void wasi_config_inherit_stderr(wasi_config_t *config);
 
+/**
+ * \brief This directory can be read, for example its entries can be iterated
+ * over and files can be opened.
+ */
+#define WASMTIME_WASI_DIR_PERMS_READ 1
```
ajalt, line 170: I didn't see any other `#defines` for enums that I could use for a naming convention, so let me know if these should be moved or renamed.

crates/c-api/src/wasi.rs
```diff
@@ -183,13 +185,16 @@ pub unsafe extern "C" fn wasi_config_preopen_dir(
...
+    let file_perms = wasmtime_wasi::FilePerms::from_bits(file_perms);
+
+    dir_perms
+        .and_then(|dir_perms| {
+            file_perms.map(|file_perms| {
+                config
+                    .builder
+                    .preopened_dir(host_path, guest_path, dir_perms, file_perms)
+            })
+        })
```
ajalt, line 198: This feels clunky, but I'm not a rust expert so if you know of a more idiomatic way of handling the error cases, I'd be happy to clean this up.

alexcrichton: Oh for this I might recommend:

```rust
let dir_perms = wasmtime_wasi::DirPerms::from_bits(dir_perms)
    .ok_or_else(|| anyhow!("invalid directory permissions"))?;
```

That way you can avoid the nested closures

crates/c-api/include/wasi.h
alexcrichton: Looks good to me!

crates/c-api/src/wasi.rs
ajalt: That's certainly cleaner! But `wasi_config_preopen_dir` returns `bool` rather than a `Result`, so it doesn't appear that you can use the `?` operator. Is there a way to do an unwrap / early return like that in a function that returns `bool`?
 Or would it be better to change the function signature? Returning strings from a C API would require callers to `free` the returned result, but we could potentially use a numerical error code instead.

alexcrichton: Ah excellent point! I forgot this returns `bool`, so sorry about that.

For that I'd recommend a `match`-per-translation-of-config to return `false` early and that way it doesn't have to nest.

I do like the idea of changing the function signature too. It's overdue for the WASI APIs here. In doing so though it might be best to rename and shuffle things around. For example the original intention of `wasi.h` was to become a standard API like the wasm-c-api proposal (which hasn't finished being standardized yet). In lieu of that nowadays it's probably best to rename things to `wasmtime_wasi_*`, move the header to `wasmtime/wasi.h`, and update idioms to match the rest of the C API (e.g. `wasmtime_error_t*` instead of `bool`). That's a bit of a larger project though so how about we keep the `bool` for now and defer this refactoring for later?

ajalt: Thanks, that's much cleaner. That's also the pattern already being used for the other parameters of the function, which I somehow didn't recognize 🤦. Sorry for the churn.

alexcrichton (approved):
No worries!

PR #9479, opened 2024-10-16
Switch libunwind detection to using weak symbols

### LINKED ISSUE #8897: libunwind warning when compiling a wasmtime project against musl
### Test Case

Load _any_ wasm component (wasip2) with wasmtime 21.0.1, if you compile the rust host against musl, a warning is printed to the terminal:

```
libunwind: __unw_add_dynamic_fde: bad fde: FDE is really a CIE
```

The component and the whole system works correctly, but the log is confusing and there's no way to tell if something is actually broken.

### Expected Results

It should work as with builds against glibc, or builds on macos with no libunwind warning printed to the terminal.

### Actual Results

I can't seem to find anything being broken really...

### Versions and Environment

Wasmtime version or commit: 21.0.1

Operating system: nixos linux unstable

Architecture: x86_64-unknown-linux-musl for the host and wasm32-wasip1 for the guest.

### Extra Info

I found issues from the past, such as

https://github.com/wasmerio/wasmer/issues/2150
https://github.com/bytecodealliance/wasmtime/issues/1904

But nothing more recent...

alexcrichton:
This is an attempt to resolve #8897.

Closes https://github.com/bytecodealliance/wasmtime/issues/8897.

### DIFF
--- a/crates/wasmtime/src/runtime/vm/helpers.c
+++ b/crates/wasmtime/src/runtime/vm/helpers.c
@@ -6,6 +6,7 @@
 #include <setjmp.h>
 #include <stdint.h>
 #include <stdlib.h>
+#include <stdbool.h>
 
 #if (defined(__GNUC__) && !defined(__clang__))
 #define WASMTIME_GCC 1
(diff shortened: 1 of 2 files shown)


alexcrichton:
Posting this to see what CI says, I won't be able to test this for a week or so.

alexcrichton:
Ok I've done a bit of local testing and I think this should be good to land

pchickey (approved):

PR #9483, opened 2024-10-17
Don't bounds-check the struct-ref result of the trusted allocation libcall

fitzgen:
It is unnecessary for Wasm code to do this bounds check because the libcall is trusted. We already avoided it for arrays, but forgot to do structs as well.


elliottt (approved):
Nice!

PR #9484, opened 2024-10-17
Add the "null" garbage collector

fitzgen:
The null collector does not actually collect any garbage, it simply bump-allocates until the heap is exhausted, at which point further allocation attempts will fail. It does not require any GC barriers.

Users can configure which collector to use via the `wasmtime::Config::collector` method or the `-C collector=drc|null` CLI flag. The `wasmtime::Collector` enumeration, similar to the `wasmtime::Strategy` enumeration but for choosing a collector rather than a compiler, additionally has a table summarizing the properties and characteristics of our current set of collector implementations.

Finally, we also run `.wast` tests that exercise GC types under both the DRC and null collectors. I tried to avoid running tests that are not related to GC under both configurations just to avoid a combinatorial blow up.


crates/cli-flags/src/lib.rs
```diff
@@ -532,6 +542,23 @@ impl CommonOptions {
...
+                #[cfg(not(feature = "gc-drc"))]
+                wasmtime::Collector::DeferredReferenceCounting => anyhow::bail!(
+                    "support for the `drc` collector is unavailable because the \
+                     `gc-drc` feature was disabled at compile time",
+                ),
+                #[cfg(not(feature = "gc-null"))]
+                wasmtime::Collector::Null => anyhow::bail!(
+                    "support for the `null` collector is unavailable because the \
+                     `gc-null` feature was disabled at compile time",
+                ),
```
alexcrichton, line 557: Could this get pushed into a `Result`-return of `Config::collector` or within `Config::validate`?

crates/wasmtime/src/config.rs
```diff
@@ -1980,6 +1997,22 @@ impl Config {
             }
         }
 
+        tunables.collector = if self.features().gc_types() {
```
alexcrichton, line 2000: Instead of `self.features()` can you use the `features` that are in scope from above?

tests/disas/gc/drc/array-len.wat
```diff
@@ -1,5 +1,5 @@
 ;;! target = "x86_64"
-;;! flags = "-W function-references,gc"
+;;! flags = "-W function-references,gc -C collector=drc"
```
alexcrichton, line 2: To confirm this technically isn't necessary right? (to specify `-C collector=drc`)

I think it's good to have in the test for future-proofing, but I want to double-check that as-is in isolation this isn't needed if we make no more changes to wasmtime in the future

alexcrichton (approved):
Very nice! I like how this shaped up. I hope the `#[cfg]` wasn't too much to add, but it all looks relatively focused to be so it's hopefully just a one-time cost.

tests/disas/gc/drc/array-len.wat
fitzgen: Correct, this is just future-proofing for if/when we change the default collector.

fitzgen:
> Very nice! I like how this shaped up.

Thanks!

> I hope the `#[cfg]` wasn't too much to add, but it all looks relatively focused to be so it's hopefully just a one-time cost.

Indeed, it was mostly some shuffling to get rid of the disabled collector impl and then realizing that various things that were DRC-specific modules could actually be made a little more generic and reused with other collectors. I think adding the copying collector will be relatively small in comparison to this PR, which is kind of funny.

PR #9490, opened 2024-10-20
winch: Improve tunables/configuration validation

saulecabrera:
This commit introduces proper validation and documentation to handle the engine features not currently supported by Winch (e.g., `consume_fuel`, follow-up to the discussion in
https://github.com/bytecodealliance/wasmtime/pull/9472).

In this commit, the tunables validation is done at the `set_tunables` method in Winch's `wasmtime_environ::compile::CompilerBuilder` implemetation, which enables removing Winch specific logic from `Config::validate`

This change also introduces the question of how to consolidate the compiler specific flags with Winch (e.g, the user-specified `cranelift_*` options) given that not all of them are applicable to Winch (e.g., `cranelift_debug_verifier`,
`cranelift_nan_canonicalization`), this change doesn't introduce any functionality on this front, however, it's probably something worth considering/discussing.


alexcrichton (approved):

alexcrichton:
For the fuzz failure I might recommend updating [somewhere around here](https://github.com/bytecodealliance/wasmtime/blob/dc2c94dd66a64416515f32d8057f7a689b734f04/crates/fuzzing/src/generators/config.rs#L474) to switch back to Cranelift if epochs (or other Winch-unsupported configs) are enabled during fuzzing

PR #9492, opened 2024-10-21
test-macros: Handle dead code warnings

saulecabrera:
This commit handles dead code warnings produced by the compiler when the macro is invoked as:

    #[wasmtime_test(strategies(not(Cranelift))]

Which occur because Winch currently only offers support for x86_64.

The motivation behind this change comes from
https://github.com/bytecodealliance/wasmtime/pull/9490, which is where the warnings surfaced.

### DIFF
--- a/crates/test-macros/src/lib.rs
+++ b/crates/test-macros/src/lib.rs
@@ -205,7 +205,19 @@ pub fn wasmtime_test(attrs: TokenStream, item: TokenStream) -> TokenStream {
 }
 
 fn expand(test_config: &TestConfig, func: Fn) -> Result<TokenStream> {
-    let mut tests = vec![quote! { #func }];
+    let mut tests = if test_config.strategies.len() == 1
+        && test_config.strategies.get(0).map(|s| s.to_string()) == Some("Winch".to_string())
+    {
+        vec![quote! {
+            // This prevents dead code warning when the macro is invoked as:
+            //     #[wasmtime_test(strategies(not(Cranelift))]
+            // Given that Winch only fully supports x86_64.
+            #[allow(dead_code)]
+            #func
+        }]
+    } else {
+        vec![quote! { #func }]
+    };
     let attrs = &func.attrs;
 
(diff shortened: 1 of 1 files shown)


crates/test-macros/src/lib.rs
```diff
@@ -205,7 +205,19 @@ pub fn wasmtime_test(attrs: TokenStream, item: TokenStream) -> TokenStream {
...
 fn expand(test_config: &TestConfig, func: Fn) -> Result<TokenStream> {
-    let mut tests = vec![quote! { #func }];
+    let mut tests = if test_config.strategies.len() == 1
+        && test_config.strategies.get(0).map(|s| s.to_string()) == Some("Winch".to_string())
+    {
+        vec![quote! {
+            // This prevents dead code warning when the macro is invoked as:
+            //     #[wasmtime_test(strategies(not(Cranelift))]
+            // Given that Winch only fully supports x86_64.
+            #[allow(dead_code)]
```
alexcrichton, line 215: As an alternative instead of `#[cfg(target_arch = "x86_64")]` being emitted, how about `#[cfg_attr(target_arch = "x86_64", ignore)]`? That way we wouldn't have to worry about unused warnings or dead code I think

alexcrichton (approved):

crates/test-macros/src/lib.rs
saulecabrera: I briefly tried this approach, I didn't continue down this route because the compiler was complaining about `multiple cfg predicates are specified`, presumably because some tests have other extra attributes like `#[cfg_attr(miri, ignore)]` -- I think we could parse those attributes and modify them, however, since this is "temporary" (i.e., once we can fully test `aarch64`) I thought it might make sense to go with this approach.

saulecabrera: Had an offline discussion and concluded that `should_panic` probably won't play nicely with tests that return `Result`, so we're moving forward with the current approach.

PR #9493, opened 2024-10-22
Enable --all-features and display feature requirements in Cranelift docs on docs.rs

nihalpasham:
### Summary
- Enabled `--all-features` for Cranelift documentation via:
  ```toml
  [package.metadata.docs.rs]
  all-features = true
  ```
- Added `#![cfg_attr(docsrs, feature(doc_auto_cfg))]` to display feature requirements in documentation when building on `docs.rs`.
- Verified functionality using:
  ```bash
  RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc --open
  ```

### DIFF
--- a/cranelift/Cargo.toml
+++ b/cranelift/Cargo.toml
@@ -13,13 +13,21 @@ rust-version.workspace = true
 [lints]
 workspace = true
 
+[package.metadata.docs.rs]
+# Ask Cargo to build docs with `--all-features` set to true
+all-features = true
+
 [[bin]]
 name = "clif-util"
 path = "src/clif-util.rs"
 
 [dependencies]
 cfg-if = { workspace = true }
-cranelift-codegen = { workspace = true, features = ["disas", "trace-log", "timing"] }
+cranelift-codegen = { workspace = true, features = [
+    "disas",
+    "trace-log",
+    "timing",
+] }
 cranelift-entity = { workspace = true }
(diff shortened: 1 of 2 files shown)


cranelift/Cargo.toml
```diff
@@ -13,13 +13,21 @@ rust-version.workspace = true
 [lints]
 workspace = true
 
+[package.metadata.docs.rs]
+# Ask Cargo to build docs with `--all-features` set to true
+all-features = true
```
bjorn3, line 18: You need to add this to `cranelift/codegen/Cargo.toml`, not `cranelift/Cargo.toml`. `cranelift/Cargo.toml` is for the cranelift-tools crate which isn't published to crates.io anyway. Also instead of enabling all non-default features, I would suggest only enabling `all-arch`. Some like `souper-harvest` and `disas` don't affect the public api while still making building the documentation slower.

nihalpasham: - Updated `cranelift/codegen/Cargo.toml` to enable the `all-arch` feature:
  ```toml
  [package.metadata.docs.rs]
  features = ["all-arch"]
  ```
- Verified feature requirements are displayed in the docs using:
  ```bash
  RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc --open
  ```

alexcrichton (approved):
Thanks!

PR #9495, opened 2024-10-22
Improve `atomic_rmw` lowering on `x86`

### LINKED ISSUE #2153: x64/newBE: improve lowering of CLIF `AtomicRmw` in the case where the old value is not required
PR #2149 provides an initial implementation of wasm atomics with x64/newBE.  That includes an initial lowering of CLIF `AtomicRmw` into a multi-instruction sequence that is general, but inefficient in the case where the old value at the location is not required, which is probably a common occurrence.  In such cases we could do much better by producing a single LOCK-prefixed integer read-modify-write instruction, for example `lock xorq $27, (amode)`.

beetrees:
`xchg` and `lock xadd` can be used in all cases for `Xchg`, `Add` and `Sub`, while various `lock`-prefixed arithmetic instructions can be used for `Add`, `Sub`, `And`, `Or` and `Xor` if the result value of `atomic_rmw` (the old value) isn't used.

Closes #2153

### DIFF
--- a/cranelift/codegen/src/isa/x64/inst/emit.rs
+++ b/cranelift/codegen/src/isa/x64/inst/emit.rs
@@ -9,6 +9,7 @@ use crate::isa::x64::encoding::rex::{
 use crate::isa::x64::encoding::vex::{VexInstruction, VexVectorLength};
 use crate::isa::x64::inst::args::*;
 use crate::isa::x64::inst::*;
+use crate::isa::x64::lower::isle::generated_code::{Atomic128RmwSeqOp, AtomicRmwSeqOp};
 
 /// A small helper to generate a signed conversion instruction.
 fn emit_signed_cvt(
@@ -255,6 +256,7 @@ pub(crate) fn emit(
             src1_dst,
             src2,
             op,
+            lock,
         } => {
             let src2 = src2.to_reg();
             let src1_dst = src1_dst.finalize(state, sink).clone();
@@ -268,10 +270,11 @@ pub(crate) fn emit(
                 _ => panic!("Unsupported read-modify-write ALU opcode"),
             };
 
-            let prefix = if *size == OperandSize::Size16 {
-                LegacyPrefixes::_66
-            } else {
-                LegacyPrefixes::None
+            let prefix = match (size, lock) {
+                (OperandSize::Size16, false) => LegacyPrefixes::_66,
+                (OperandSize::Size16, true) => LegacyPrefixes::_66F0,
+                (_, false) => LegacyPrefixes::None,
+                (_, true) => LegacyPrefixes::_F0,
             };
             let opcode = if *size == OperandSize::Size8 {
                 opcode - 1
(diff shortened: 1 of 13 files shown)


cranelift/codegen/src/isa/x64/inst/mod.rs
```diff
@@ -2542,6 +2572,28 @@ fn x64_get_operands(inst: &mut Inst, collector: &mut impl OperandVisitor) {
...
+        Inst::Xchg {
+            operand,
+            mem,
+            dst_old,
+            ..
+        } => {
+            collector.reg_use(operand);
+            collector.reg_reuse_def(dst_old, 0);
+            mem.get_operands(collector);
+        }
```
abrown, line 2595: I believe this is correct but @cfallin do you want to double-check?

cranelift/codegen/src/isa/x64/lower.isle
```diff
@@ -3386,19 +3386,53 @@
...
+      (x64_lock_sub (raw_operand_size_of_type ty) (to_amode flags address (zero_offset)) input))
+(rule 2 (lower i @ (has_type (fits_in_64 (ty_int ty))
+                  (atomic_rmw flags (AtomicRmwOp.And) address input)))
+      (if-let (first_result res) i)
+      (if-let $true (value_is_unused res))
+      (x64_lock_and (raw_operand_size_of_type ty) (to_amode flags address (zero_offset)) input))
+(rule 2 (lower i @ (has_type (fits_in_64 (ty_int ty))
+                  (atomic_rmw flags (AtomicRmwOp.Or) address input)))
+      (if-let (first_result res) i)
+      (if-let $true (value_is_unused res))
```
abrown, line 3426: Another note for @cfallin to check: this does in fact do what we would expect it to?

abrown (approved):
Thanks for the PR! This looks fine to me but I'm going to ask @cfallin to double-check a couple spots below.

cranelift/codegen/src/isa/x64/inst/emit.rs
```diff
@@ -4120,10 +4152,8 @@ pub(crate) fn emit(
             let operand_rmi = RegMemImm::reg(operand);
             use inst_common::MachAtomicRmwOp as RmwOp;
             match op {
-                RmwOp::Xchg => {
-                    // movq %r_operand, %r_temp
-                    let i3 = Inst::mov_r_r(OperandSize::Size64, operand, temp);
-                    i3.emit(sink, info, state);
+                RmwOp::Add | RmwOp::Sub | RmwOp::Xchg => {
```
cfallin, line 4155: Now that we don't use all `MachAtomicRmwOp` options in lowerings anymore, could we define a separate enum for x64 and translate into it at the rule-match step, so we don't have this problem of representable but invalid combinations?

cranelift/codegen/src/isa/x64/inst/mod.rs
cfallin: Looks right to me!

cranelift/codegen/src/isa/x64/lower.isle
cfallin: Yes, I think so! We're doing a backward pass so we know for sure if a value is used or not, and this helper looks like it was added to be used for a similar purpose in a few other backends; perfectly valid to use it here to know if we can avoid producing the result.

cranelift/codegen/src/isa/x64/pcc.rs
```diff
@@ -163,6 +163,7 @@ pub(crate) fn check(
             op: _,
             ref src1_dst,
             src2: _,
+            ..
```
cfallin, line 166: We try to avoid any `..` here so we don't miss new fields being added. Could we instead do `lock: _`?

cranelift/codegen/src/isa/x64/pcc.rs
```diff
@@ -898,6 +899,28 @@ pub(crate) fn check(
             Ok(())
         }
 
+        Inst::LockXadd {
+            size,
+            ref mem,
+            dst_old,
+            ..
```
cfallin, line 906: Likewise here

cranelift/codegen/src/isa/x64/pcc.rs
```diff
@@ -898,6 +899,28 @@ pub(crate) fn check(
...
+            ensure_no_fact(vcode, dst_old.to_reg())?;
+            check_store(ctx, None, mem, vcode, size.to_type())?;
+            Ok(())
+        }
+
+        Inst::Xchg {
+            size,
+            ref mem,
+            dst_old,
+            ..
```
cfallin, line 917: Likewise here

cfallin (approved):
Thanks!

PR #9500, opened 2024-10-22
Implement wide-arithmetic for Winch

alexcrichton:
This commit implements the wide-arithmetic proposal for Winch on x64. This is mostly for me to get my feet wet doing things in Winch. The proposal itself is relatively modest with just four new instructions.


saulecabrera (approved):
Thanks!

PR #9496, opened 2024-10-22
Bump MSRV to 1.80

alexcrichton:
This commit updates the minimum Rust version required to compile Wasmtime to 1.80.0 from the previous 1.78.0.

PR #9501, opened 2024-10-23
Update nightly in CI

alexcrichton:
This commit updates nightly again in CI after failing to do so in #9496. This fixes an issue in our release CI where CMake was misconfigured when cross-compiling and creating `aarch64-pc-windows-msvc` artifacts.

prtest:full


alexcrichton:
There's a little more discussion [here](https://github.com/rust-lang/rust/pull/131070#issuecomment-2430252221) as to what's going on

PR #9504, opened 2024-10-23
x64: simplify `SyntheticAmode::finalize`

abrown:
There is no need for a `SyntheticAmode` to request mutable access to the entire `EmitState`; it only needs a read-only `FrameLayout`.

### DIFF
--- a/cranelift/codegen/src/isa/x64/inst/args.rs
+++ b/cranelift/codegen/src/isa/x64/inst/args.rs
@@ -1,7 +1,6 @@
 //! Instruction operand sub-components (aka "parts"): definitions and printing.
 
 use super::regs::{self};
-use super::EmitState;
 use crate::ir::condcodes::{FloatCC, IntCC};
 use crate::ir::types::*;
 use crate::ir::MemFlags;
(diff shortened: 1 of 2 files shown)


cfallin (approved):

abrown:
What I'm aiming at is allowing any new assembler code to be ignorant of as much of the internals of `cranelift-codegen` as possible. I think this will make it easier to test, etc. It really would be nice if the assembler didn't need to know about `SyntheticAmode` at all; if a `SyntheticAmode` could be finalized down into its equivalent `Amode` before the assembler ever sees it, that would be a good simplification. But all of the `GprMem` types and such point at `SyntheticAmodes`...

abrown:
Oh, sorry, had to fix some build issues so I force-pushed!

cfallin:
Interesting; in an old version of the MachInst backend, there was actually a separate pass to rewrite `SyntheticAmode`s, but we try to avoid separate passes and I don't think we have info to lower them when generating the instructions because they depend on details only known after regalloc, like frame layout. Perhaps one could design a way with parameterization of the enum (e.g. `Inst<A: Into<Amode>>`) and then a method to "lift" the into to `Inst` (Haskell brain: I want `fmap`)?

PR #9505, opened 2024-10-23
x64: use `ByteSink` more liberally in the `rex` module

abrown:
In looking at adding auto-generated assembler code to Cranelift, I've noticed that we pass `MachBuffer` down when it is not always necessary. The `ByteSink` trait (which `MachBuffer` implements) is a simplified view to `put*` bytes into the buffer and it is a bit simpler to use when testing since it is also implemented by `Vec<u8>`. This change propagates the trait to the `rex` module.

### DIFF
--- a/cranelift/codegen/src/isa/x64/encoding/rex.rs
+++ b/cranelift/codegen/src/isa/x64/encoding/rex.rs
@@ -1,21 +1,19 @@
-//! Encodes instructions in the standard x86 encoding mode. This is called IA-32E mode in the Intel
-//! manuals but corresponds to the addition of the REX-prefix format (hence the name of this module)
-//! that allowed encoding instructions in both compatibility mode (32-bit instructions running on a
+//! Encodes instructions in the standard x86 encoding mode. This is called
+//! IA-32E mode in the Intel manuals but corresponds to the addition of the
+//! REX-prefix format (hence the name of this module) that allowed encoding
+//! instructions in both compatibility mode (32-bit instructions running on a
 //! 64-bit OS) and in 64-bit mode (using the full 64-bit address space).
 //!
-//! For all of the routines that take both a memory-or-reg operand (sometimes called "E" in the
-//! Intel documentation, see the Intel Developer's manual, vol. 2, section A.2) and a reg-only
-//! operand ("G" in Intelese), the order is always G first, then E. The term "enc" in the following
-//! means "hardware register encoding number".
+//! For all of the routines that take both a memory-or-reg operand (sometimes
+//! called "E" in the Intel documentation, see the Intel Developer's manual,
+//! vol. 2, section A.2) and a reg-only operand ("G" in Intel-ese), the order is
+//! always G first, then E. The term "enc" in the following means "hardware
+//! register encoding number".
 
-use crate::machinst::{Reg, RegClass};
-use crate::{
-    isa::x64::inst::{
-        args::{Amode, OperandSize},
-        regs, Inst, LabelUse,
-    },
-    machinst::MachBuffer,
-};
+use super::ByteSink;
+use crate::isa::x64::inst::args::{Amode, OperandSize};
+use crate::isa::x64::inst::{regs, Inst, LabelUse};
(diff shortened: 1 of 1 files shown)


abrown:
Similar to #9504, this is just an attempt at trying to simplify what is needed to actually assemble an instruction. We still need `MachBuffer` for instructions that touch memory; thinking about it, it may just be unavoidable.

cranelift/codegen/src/isa/x64/encoding/rex.rs
```diff
@@ -157,6 +155,31 @@ impl RexFlags {
...
+                let enc_index = int_reg_enc(*reg_index);
+                self.emit_three_op(sink, enc_g, enc_index, enc_base);
+            }
+
+            Amode::RipRelative { .. } => {
+                // note REX.B = 0.
+                self.emit_two_op(sink, enc_g, 0);
+            }
+        }
+    }
```
abrown, line 182: Oh, actually let me remove this: this is for the new assembler but isn't used yet.

cranelift/codegen/src/isa/x64/encoding/rex.rs
```diff
@@ -81,13 +79,25 @@ impl RexFlags {
...
         self.0 = self.0 | 2;
         self
     }
 
+    /// True if the REX prefix must always be emitted.
+    #[inline(always)]
+    pub fn must_always_emit(&self) -> bool {
+        (self.0 & 2) != 0
+    }
+
```
abrown, line 100: ...and a little bit of code movement.

cfallin (approved):
This looks fine, thanks!

Re: handling `Amode`s and the helper that briefly appeared here: I think we'll need `MachBuffer` unavoidably once we have amodes that can refer to labels; that's the semantic thing that MachBuffer gives us above an arbitrary byte sequence. But totally fine to trim the close coupling at lower levels.

I'd love to talk more about the assembler design w.r.t. dependencies -- intuitively I would have thought that `MachBuffer` would be an integral part of it, i.e. it's the assembler's way of providing its output (given that assemblers need label references, and to flow through other metadata like trap info and relocs from insts to the code blob), but maybe there's another use-case or need for generalization I'm missing.

abrown:
> intuitively I would have thought that MachBuffer would be an integral part of it

I mean, that's how it has to be today. I'm kind of wondering if it should, though: it's nice to be able to "just emit" an instruction into a byte buffer during testing and that's what I've been thinking about. The setup in `emit_tests.rs` just to be able to emit an instruction is a lot, e.g. If we're proptest-ing or fuzzing we also want the "emit a single instruction" path to be relatively quick.

On the other hand, sometimes its unavoidable: I've notice that we register traps and labels using a `MachBuffer` in the encoding logic and that simply cannot be done with `ByteSink`. Maybe `ByteSink` can gain extra trait functions that do nothing for the `Vec<u8>` case or something like that...

cfallin:
> Maybe ByteSink can gain extra trait functions that do nothing for the Vec<u8> case or something like that...

Ah! Yeah, that's an interesting idea: lift the whole interface of MachBuffer into a trait, and provide an implementation of that trait for a `Vec<u8>`.

There are interesting semantic implementations of this though when label references are involved: what *should* the assembled bytes from a `mov rax, [rip+label]` instruction be? Zeroes in the RIP offset (i.e. labels not resolved)? One aspect of the `MachBuffer` that is nice is that its types essentially force one to be aware one is referencing addresses in the buffer: you have to ask for a label to refer to it, the various amodes require a label, and if you don't bind it, (IIRC?) it should panic when finalizing.

Maybe another way of reading this friction is that we need some convenience APIs? It should be possible to write an `assemble(inst: Inst) -> Vec<u8>` that is something like:

```
fn assemble(inst: Inst) -> Vec<u8> {
  let mut buffer = MachBuffer::new();
  inst.emit(&mut inst, ...);
  buffer.finalize(&VCodeConstants::default(), &ControlPlane::default()).apply_base_srcloc(SourceLoc::new(0)).data.into()
}
```

(whew, that is a bit of a mouthful, I see why you want it!)

PR #9507, opened 2024-10-24
Shared memory support

atilag:
I was trying to add Shared Memory support in the wasmtime-py library, but I found out that C-API support here was not finished, so this is my bold attempt to implement something that works.

As far as a I read, threads on WASI are still in discussion and Shared Memory is part of the proposal. In other hand there's some basic support for Shared memory already present in the Rust API interface, so I guess it's fine to try gluing the C-API with what is supported today, and make wasmtime-py works with Shared Memory (?).

### DIFF
--- a/crates/c-api/src/extern.rs
+++ b/crates/c-api/src/extern.rs
@@ -110,7 +110,7 @@ impl wasmtime_extern_t {
             WASMTIME_EXTERN_GLOBAL => Extern::Global(self.of.global),
             WASMTIME_EXTERN_TABLE => Extern::Table(self.of.table),
             WASMTIME_EXTERN_MEMORY => Extern::Memory(self.of.memory),
-            WASMTIME_EXTERN_SHAREDMEMORY => Extern::SharedMemory(self.of.sharedmemory),
+            WASMTIME_EXTERN_SHAREDMEMORY => Extern::SharedMemory((**self.of.sharedmemory).clone()),
             other => panic!("unknown wasmtime_extern_kind_t: {other}"),
         }
     }
@@ -138,7 +138,7 @@ impl From<Extern> for wasmtime_extern_t {
             Extern::SharedMemory(sharedmemory) => wasmtime_extern_t {
                 kind: WASMTIME_EXTERN_SHAREDMEMORY,
                 of: wasmtime_extern_union {
-                    sharedmemory: sharedmemory,
+                    sharedmemory: ManuallyDrop::new(Box::new(sharedmemory)),
                 },
             },
         }
--- a/crates/c-api/src/extern.rs
+++ b/crates/c-api/src/extern.rs
@@ -20,7 +20,7 @@ pub extern "C" fn wasm_extern_kind(e: &wasm_extern_t) -> wasm_externkind_t {
         Extern::Global(_) => crate::WASM_EXTERN_GLOBAL,
         Extern::Table(_) => crate::WASM_EXTERN_TABLE,
         Extern::Memory(_) => crate::WASM_EXTERN_MEMORY,
-        Extern::SharedMemory(_) => todo!(),
+        Extern::SharedMemory(_) => crate::WASM_EXTERN_SHAREDMEMORY,
     }
 }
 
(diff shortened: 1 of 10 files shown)


crates/fuzzing/src/oracles/dummy.rs
```diff
@@ -23,6 +23,7 @@ pub fn dummy_extern<T>(store: &mut Store<T>, ty: ExternType) -> Result<Extern> {
         ExternType::Global(global_ty) => Extern::Global(dummy_global(store, global_ty)?),
         ExternType::Table(table_ty) => Extern::Table(dummy_table(store, table_ty)?),
         ExternType::Memory(mem_ty) => Extern::Memory(dummy_memory(store, mem_ty)?),
+        ExternType::SharedMemory(mem_ty) => Extern::SharedMemory(dummy_memory(store, mem_ty)?),
```
atilag, line 26: Ok, this is wrong.

crates/c-api/src/extern.rs
```diff
@@ -20,7 +20,7 @@ pub extern "C" fn wasm_extern_kind(e: &wasm_extern_t) -> wasm_externkind_t {
         Extern::Global(_) => crate::WASM_EXTERN_GLOBAL,
         Extern::Table(_) => crate::WASM_EXTERN_TABLE,
         Extern::Memory(_) => crate::WASM_EXTERN_MEMORY,
-        Extern::SharedMemory(_) => todo!(),
+        Extern::SharedMemory(_) => crate::WASM_EXTERN_SHAREDMEMORY,
```
alexcrichton, line 23: There's a split between the `wasm_*` APIs and `wasmtime_*` APIs and we can't change the `wasm_*` ones so this unfortunately can't change. Perhaps rename the `todo!()` though to a `panic!()` with an error message along the lines of "please use the `wasmtime_*` method instead"?

I should also note that if the `wasmtime_*` method doesn't exist yet and you need this method specifically then it's fine to add a `wasmtime_*` method too.

crates/c-api/src/types/extern.rs
```diff
@@ -15,6 +15,7 @@ pub(crate) enum CExternType {
     Func(CFuncType),
     Global(CGlobalType),
     Memory(CMemoryType),
+    SharedMemory(CMemoryType),
```
alexcrichton, line 18: This I think we'll want to avoid where we decided to split the runtime representation of `Memory` and `SharedMemory` but at the type level they're the same type. In that sense the `Memory` variant above this should be sufficient.

crates/c-api/src/types/memory.rs
```diff
@@ -71,20 +73,28 @@ pub extern "C" fn wasmtime_memorytype_new(
...
     let maximum = if maximum_specified {
         Some(maximum)
     } else {
         None
     };
-    Box::new(wasm_memorytype_t::new(if memory64 {
-        MemoryType::new64(minimum, maximum)
-    } else {
-        MemoryType::new(
+    Box::new(if shared {
```
alexcrichton, line 83: Mind panicking here for now with `unimplemented!()` if both `shared` and `memory64` are set? (we need to add a constructor for that but haven't gotten around to it yet)

alexcrichton: Actually I take this back, this may want to use `MemoryTypeBuilder` to support shared 64-bit memories

atilag:
Suggestions applied!
Wanted to link this issue with [this other one] from wasmtime-py, as this is a required dependency. We will need to rebuild the `_bindings.py` after this is merged.

alexcrichton:
Thanks! I think we'll still want to remove the `SharedMemory` arm of `ExternType` which will help simplify this a bit more though.

atilag:
Done!

crates/fuzzing/src/oracles/dummy.rs
```diff
@@ -77,8 +77,14 @@ pub fn dummy_memory<T>(store: &mut Store<T>, ty: MemoryType) -> Result<Memory> {
     Memory::new(store, ty)
 }
 
+/// Construct a dummy shared memory for the given memory type.
+pub fn dummy_shared_memory(engine: &Engine, ty: MemoryType) -> Result<SharedMemory> {
+    SharedMemory::new(engine, ty)
+}
```
alexcrichton, line 83: Is this still needed? (or should it perhaps be integrated elsewhere?)

alexcrichton (commented):
Looking good! With `MemoryTypeBuilder` for the implementation I think this will be good to go 👍

crates/fuzzing/src/oracles/dummy.rs
atilag: Nah, I just removed it

atilag:
Oh, nice, the MemoryTypeBuilder made the code way more readable :)
Done!

alexcrichton (approved):
There's a minor issue with clang-format in CI, but otherwise I think this is good to go 👍

atilag:
Oh, no problem. I can run clang-format and see a green CI :)

alexcrichton:
I think there might be something else with clang-format in CI?

If you'd like also feel free to mark this as "ready for review" so it can be approved+merged once CI passes.

atilag:
Yeah, might be the version of clang format usedd in the CI (15?) has different default options than the one I use locally.
Anyway, ready!
Thanks for taking a look Alex!

alexcrichton:
I dunno what's going on with clang-format myself, I've attempted to resolve the latest issue with a new commit

Issue #9510, opened 2024-10-24
Cranelift: Remove support for implicitly adding a return area pointer


bjorn3:
#### Feature

 When there are more returns than available return registers, the return value has to be returned through the introduction of a return area pointer. This can be done both explicitly using a `ArgumentPurpose::StructReturn` parameter (as cg_clif does) or implicitly by Cranelift. The support for implicitly adding it adds a lot of complexity to the ABI handling code in Cranelift (which is already way too complex) and has two ABI incompatibilities:

* On x86_64 sysv the implicitly introduced return area pointer is not returned again in rax. I've tried to fix this, but the different code paths between explicit and implicit return area pointers are both too different and too similar to be able to handle this.
* We split the return values between registers and the return area when everything should be returned using the return area pointer when it doesn't fit in registers according to the abi.

Given that there is a lot of complexity in Cranelift caused by this and it is effectively broken anyway for native calling conventions, I propose removing support for implicit return area pointers, requiring the frontend to use explicit return area pointers instead.

#### Benefit

* Reduced complexity of Cranelift.
* Can't accidentally use a broken ABI.

cfallin:
@bjorn3 and I talked about this a bit offline -- I was initially skeptical for the reason that supporting arbitrary return values is ergonomic, fits well with the CLIF abstraction level in general, and we have it already. However, if we're facing some brokenness in our implementation already with respect to what the ABI says should happen, it may not be a bad idea to say that the Cranelift embedder (Wasmtime or cg\_clif or other) needs to build their own mechanism for multiple return values. After all, "N args, 1 return" is also a perfectly consistent abstraction level and one that we live with in many other languages.

Practically speaking in this repo, this means a good amount of code removal in the ABI code as bjorn3 mentions, and then building the equivalent in `cranelift-wasm`. I suspect we could do it in a single pass still by, lazily when needed, creating a stackslot for returns, generating a CLIF call with extra arg, and loads of values from stackslot to define the return values. (The stackslot can be reused for all callsites, and we could update its size as we discover larger return-value sets.)

What do others think? We can talk about this in the next CL meeting if needed, too.

bjorn3:
I've opened https://github.com/bytecodealliance/wasmtime/pull/9511 to gate the support behind an option. This should help with removing it in the future and reduces likelihood of accidentally hitting the ABI issue in the short term.

fitzgen:
[...]

PR #9511, opened 2024-10-24
Gate support for implicit return area pointers behind an option

bjorn3:
It implements an incorrect ABI and may be removed in the future due to complexity reasons. By requiring to enable an option to use it, it becomes harder to accidentally hit the ABI issue and allows a more deprecation and eventual removal.

### DIFF
--- a/cranelift/codegen/meta/src/shared/settings.rs
+++ b/cranelift/codegen/meta/src/shared/settings.rs
@@ -209,6 +209,25 @@ pub(crate) fn define() -> SettingGroup {
     );
 
     settings.add_bool(
+        "enable_multi_ret_implicit_sret",
+        "Enable support for sret arg introduction when there are too many ret vals.",
+        r#"
+            When there are more returns than available return registers, the
+            return value has to be returned through the introduction of a
+            return area pointer. Normally this return area pointer has to be
+            introduced as `ArgumentPurpose::StructReturn` parameter, but for
+            back compat reasons Cranelift also supports implicitly introducing
+            this parameter and writing the return values through it.
+
+            **This option violates the ABI of all targets and the used ABI should
+            not be assumed to remain the same between Cranelift versions.**
+
+            This option is **deprecated** and will be removed in the future.
+        "#,
+        false,
+    );
+
+    settings.add_bool(
         "unwind_info",
         "Generate unwind information.",
         r#"
--- a/cranelift/codegen/meta/src/shared/settings.rs
+++ b/cranelift/codegen/meta/src/shared/settings.rs
@@ -216,13 +216,23 @@ pub(crate) fn define() -> SettingGroup {
             return value has to be returned through the introduction of a
             return area pointer. Normally this return area pointer has to be
             introduced as `ArgumentPurpose::StructReturn` parameter, but for
-            back compat reasons Cranelift also supports implicitly introducing
-            this parameter and writing the return values through it.
+            backward compatibility reasons Cranelift also supports implicitly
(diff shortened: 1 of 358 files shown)


cranelift/codegen/meta/src/shared/settings.rs
```diff
@@ -208,6 +208,25 @@ pub(crate) fn define() -> SettingGroup {
...
 
+    settings.add_bool(
+        "enable_multi_ret_implicit_sret",
+        "Enable support for sret arg introduction when there are too many ret vals.",
+        r#"
+            When there are more returns than available return registers, the
+            return value has to be returned through the introduction of a
+            return area pointer. Normally this return area pointer has to be
+            introduced as `ArgumentPurpose::StructReturn` parameter, but for
+            back compat reasons Cranelift also supports implicitly introducing
```
cfallin, line 219: s/back compat/backward compatibility/

cranelift/codegen/meta/src/shared/settings.rs
```diff
@@ -208,6 +208,25 @@ pub(crate) fn define() -> SettingGroup {
...
+            this parameter and writing the return values through it.
+
+            **This option violates the ABI of all targets and the used ABI should
```
cfallin, line 222: s/violates the ABI of all targets/currently does not conform to platform ABIs/

cranelift/codegen/meta/src/shared/settings.rs
```diff
@@ -208,6 +208,25 @@ pub(crate) fn define() -> SettingGroup {
...
+            not be assumed to remain the same between Cranelift versions.**
+
+            This option is **deprecated** and will be removed in the future.
```
cfallin, line 225: Can we add a paragraph at the end here recommending that the embedder develop their own multi-return convention as the most stable option? Something like:

```
Because of the above issues, and complexities of native ABI support for the concept
in general, Cranelift's support for multiple return values may also be removed in the
future (#9510). For the most robust solution, it is recommended to build a convention
on top of Cranelift's primitives for passing multiple return values, for example by allocating
a stackslot in the caller, passing it as an explicit argument, storing return values in the callee,
and loading results in the caller.
```

cranelift/codegen/src/isa/aarch64/abi.rs
```diff
@@ -330,6 +331,14 @@ impl ABIMachineSpec for AArch64MachineDeps {
 
             // Spill to the stack
 
+            if args_or_rets == ArgsOrRets::Rets && !flags.enable_multi_ret_implicit_sret() {
+                return Err(crate::CodegenError::Unsupported(
+                    "Too many return values to fit in registers. \
+                    Use a StructReturn argument instead"
```
cfallin, line 337: In these errors can we mention #9510 as well?

cfallin (approved):
Looks good, thanks -- I agree it's a good idea to wall this off a little bit while we work out what to do. A few comments on the description and error messages to offer some more tips but otherwise LGTM.

cfallin:
@bjorn3 it looks like a few riscv64 tests need the new option as well. Let me know when green and I'm happy to take a final look and merge!

bjorn3:
CI is green.

cfallin:
A few more failures in [this](https://github.com/bytecodealliance/wasmtime/actions/runs/11536083350/job/32111925617) run when it tried the full set on the merge queue, it looks like. (Feel free to use `prtest:full` in a commit message to test this on the branch!)

bjorn3:
Fixed those tests.

PR #9516, opened 2024-10-29
bench-api: Allow using `winch` in the bench api

saulecabrera:
This commit allows building the `bench-api` crate via a `winch` cargo feature. This allows using `--engine-flags="-C compiler=winch"` for benchmarking.

### DIFF
--- a/crates/bench-api/Cargo.toml
+++ b/crates/bench-api/Cargo.toml
@@ -37,3 +37,4 @@ wat = { workspace = true }
 [features]
 default = ["shuffling-allocator", "wasi-nn"]
 wasi-nn = ["wasmtime-wasi-nn"]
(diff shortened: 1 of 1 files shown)


crates/bench-api/Cargo.toml
```diff
@@ -37,3 +37,4 @@ wat = { workspace = true }
 [features]
 default = ["shuffling-allocator", "wasi-nn"]
 wasi-nn = ["wasmtime-wasi-nn"]
+winch = ["wasmtime/winch"]
```
saulecabrera, line 40: I decided to go with a feature here, given that `winch` is still not a default wasmtime feature, however I wonder if it might make sense to unconditionally enable the `winch` feature in this crate.

alexcrichton (approved):

PR #9518, opened 2024-10-29
Enable fuzzing Winch + Fuel

alexcrichton:
This was implemented in #9472, so let's let the fuzzers have a go at it.


elliottt (approved):
🎉

PR #9520, opened 2024-10-29
 threads: add `WasmCompositeInnerType` layer

abrown:
As in `wasm-tools`, push the type down a layer so that `WasmCompositeType` can be marked shared. This leaves several holes to be filled in by future commits (all marked with `TODO: handle shared`); in effect, this change allows `WasmCompositeType` to be marked shared but mostly does not wire up the sharedness during translation.


crates/cranelift/src/gc/enabled/drc.rs
```diff
@@ -415,6 +415,7 @@ impl GcCompiler for DrcCompiler {
 
         let array_ty = func_env.types[interned_type_index]
             .composite_type
+            .inner
```
alexcrichton, line 418: Like `gc/enabled.rs`, this and below will want to panic/error on `shared`

crates/environ/src/types.rs
```diff
@@ -1020,47 +1043,47 @@ impl fmt::Display for WasmSubType {
 impl WasmSubType {
     #[inline]
     pub fn is_func(&self) -> bool {
-        self.composite_type.is_func()
+        self.composite_type.inner.is_func()
```
alexcrichton, line 1046: For this accessor and those below I think we'll want to implicitly redefine `is_func` as `is_unshared_func` (although without the renaming for now). Eventually if needed we can add `is_shared_func` but otherwise these should all get extra boolean conditions and such to check if `self.shared` and act appropriately (e.g. return `false`, `None`, or `panic!`)

crates/wasmtime/src/runtime/type_registry.rs
```diff
@@ -806,12 +809,12 @@ impl TypeRegistryInner {
             "type is not canonicalized for runtime usage: {ty:?}"
         );
 
-        let gc_layout = match &ty.composite_type {
```
alexcrichton, line 809: This I think we concluded might be able to be replaced with the trait helper method above.

crates/wasmtime/src/runtime/types.rs
```diff
@@ -1617,6 +1618,7 @@ impl StructType {
         Self::from_wasm_struct_type(
             engine,
             finality.is_final(),
+            false, // TODO: handle shared
```
alexcrichton, line 1621: In this file I think it's fine to just set `shared: false` at the end. There'll be more plumbing eventually to handle this if needed but for now it's an accurate reflection that only unshared types are created here.

crates/wasmtime/src/runtime/vm/const_expr.rs
```diff
@@ -118,7 +118,11 @@ impl<'a> ConstEvalContext<'a> {
             .borrow(shared_ty)
             .expect("should have a registered type for struct");
         let WasmSubType {
-            composite_type: WasmCompositeType::Struct(struct_ty),
+            composite_type:
+                WasmCompositeType {
+                    shared: false, // TODO: handle shared
```
alexcrichton, line 123: I think this is correct to have, this is basically pattern-matching to only unshared structures for `struct.new_default` which is desired behavior at this time. This panic can then get weeded out in the future once support for `struct.new_default` comes around.

crates/environ/src/compile/module_types.rs
```diff
@@ -163,7 +163,10 @@ impl ModuleTypesBuilder {
                     let idx = self.types.push(WasmSubType {
                         is_final: true,
                         supertype: None,
-                        composite_type: WasmCompositeType::Func(f.clone()),
+                        composite_type: WasmCompositeType {
+                            inner: WasmCompositeInnerType::Func(f.clone()),
+                            shared: false, // TODO: handle shared
```
alexcrichton, line 168: Here I think you'll want to propagate the shared-ness from `self.types[for_func_ty]` up above

alexcrichton (commented):
I talked about this a bit with @abrown yesterday but I wanted to paste the conclusion here as well. My general worry as we propagate this up further is that it's accidentally easy to forget to check `.shared` which means that further down the road we might have a bug of "oh forgot to check `.shared` there". I'd prefer to continue to panic/error in variuos places where `shared` isn't handled today and I've commented a few of them here and there.

crates/wasmtime/src/runtime/vm/gc/disabled.rs
```diff
@@ -23,6 +23,21 @@ unsafe impl GcRuntime for DisabledCollector {
...
+impl GcTypeLayouts for DisabledLayouts {
+    fn array_length_field_offset(&self) -> u32 {
+        unreachable!()
+    }
+
+    fn array_layout(&self, _: &WasmArrayType) -> GcArrayLayout {
+        unreachable!()
+    }
+
+    fn struct_layout(&self, _: &WasmStructType) -> GcStructLayout {
```
alexcrichton, line 40: Did this sneak in? Or was this to fix something else?

crates/environ/src/types.rs
```diff
@@ -1850,19 +1891,23 @@ pub trait TypeConvert {
 
     fn convert_composite_type(&self, ty: &wasmparser::CompositeType) -> WasmCompositeType {
         assert!(!ty.shared);
```
alexcrichton, line 1893: This line might be able to be deleted now?

crates/wasmtime/src/runtime/vm/gc/disabled.rs
abrown: This was due to the refactor in 8445481; in order to get rid of the duplication there we need to add the disabled version of this trait in bbce595.

crates/wasmtime/src/runtime/type_registry.rs
```diff
@@ -809,7 +809,24 @@ impl TypeRegistryInner {
...
+                    .into(),
+            ),
+            wasmtime_environ::WasmCompositeInnerType::Struct(s) => Some(
+                gc_runtime
+                    .expect("must have a GC runtime to register array types")
+                    .layouts()
+                    .struct_layout(s)
+                    .into(),
+            ),
+        };
```
abrown, line 829: I ended up removing the refactor we used in this PR. Though indeed we could have done `gc_runtime.map(...).flatten()` and ended up with the same result, I'm concerned that @fitzgen probably is using these `expect`s as an early warning sign for winch, e.g.

crates/wasmtime/src/runtime/vm/gc/disabled.rs
abrown: This is now gone as a result of fixing conflicts; some intervening commit started using an `Option` to solve this.

PR #9522, opened 2024-10-30
Remove Scheme booleans

Kmeakin:
The `#t` and `#f` syntax is misleading: they are actually expressions, not booleans, so `#t` represents the integer `1` but *not* the boolean `true`. They have been replaced by the named constants `$true` and `$false`, and not used anywhere except tests, so they can be removed without breaking any lowering/optimization code.

### DIFF
--- a/cranelift/isle/docs/language-reference.md
+++ b/cranelift/isle/docs/language-reference.md
@@ -495,10 +495,10 @@ operators:
 
 * Wildcards (`_`).
 * Integer constants (decimal/hex/binary/octal, positive/negative: `1`, `-1`,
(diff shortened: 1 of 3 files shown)


cfallin (approved):
LGTM for the change in the title; would you mind rebasing out the langref stuff from the other PR? Once that's separated, I'm happy to land this!

Kmeakin:
> would you mind rebasing out the langref stuff from the other PR?

Thanks, I must have messed up the rebase somehow. GitHub PRs are more confusing compared to Gerrit IMO

cfallin (approved):

Issue #6848, opened 2023-08-15
Get instance data in host defined functions with component model


bobogei81123:
#### Feature

Provide instance (`Instance` struct) or instance data to host defined functions in component model.

#### Benefit

In #2491, it is suggested to bind the instance data to the closure passed to `Func::wrap` to store per-instance data. However, this will not work for instances in component model because to bind a function (let host export a function to be called by guests), one need to configure [`LinkerInstance`](https://docs.wasmtime.dev/api/wasmtime/component/struct.LinkerInstance.html#) (e.g., use [`LinkerInstance::func_wrap`](https://docs.wasmtime.dev/api/wasmtime/component/struct.LinkerInstance.html#method.func_wrap) or similar methods). Yet, there is only a single linker set up for instantiating all the instances in a store, so there is no way to configure per-instance data.

Providing host function instance and instance data will let host function able to get data tied to the instance that calls the function, or let host perform a nested call to a guest function, etc. 

#### Implementation

I think a possible implementation is to let the function passed to [`LinkerInstance::func_wrap`](https://docs.wasmtime.dev/api/wasmtime/component/struct.LinkerInstance.html#method.func_wrap) to be able to take `Caller` instead of `StoreContextMut`. The non-component model counterpart [`Func::wrap`](https://docs.wasmtime.dev/api/wasmtime/struct.Func.html#method.wrap) takes `AsContextMut` to achieve this, and I'm not sure why `func_wrap` takes `StoreContextMut` instead. Then, we can expose the instance stored in the `Caller` struct. Of course, this all rely on the assumption that `Caller` can be retrieve in calls to component host.

alexcrichton:
You mention that a single linker is used for a store, but that's not quite accurate because you can use any number of linkers to instantiate within a `Store`. In that sense you can still provide per-function state within the `Linker`. Otherwise though I'll also note that the question there was in reference to the C API and the Rust API is a bit different. In Rust the closures in `Linker` are `Fn` meaning you don't get mutable access to closed-over-state. Instead that sort of information typically goes within the `T` of a `Store<T>` which you do get mutable access to.

So, if possible, the recommendation is to store per-instance information into a `T` in `Store<T>` and only if necessary store per-instance information within the closures passed to a `Linker`. Does that work for you?

bobogei81123:
> I'll also note that the question there was in reference to the C API and the Rust API is a bit different.

Sorry for not stating the question clearly. I'm using the rust API. Otherwise

[...]

PR #9525, opened 2024-10-30
Expose `get_module_export` from `Caller`

richardpringle:
Not necessarily trying to get this merged, just using it as an example for my comment [here](https://github.com/bytecodealliance/wasmtime/issues/6848#issuecomment-2447900164)

### DIFF
--- a/crates/wasmtime/src/runtime/func.rs
+++ b/crates/wasmtime/src/runtime/func.rs
@@ -7,8 +7,8 @@ use crate::runtime::Uninhabited;
 use crate::store::{AutoAssertNoGc, StoreData, StoreOpaque, Stored};
 use crate::type_registry::RegisteredType;
 use crate::{
-    AsContext, AsContextMut, CallHook, Engine, Extern, FuncType, Instance, Module, Ref,
-    StoreContext, StoreContextMut, Val, ValRaw, ValType,
+    AsContext, AsContextMut, CallHook, Engine, Extern, FuncType, Instance, Module, ModuleExport,
+    Ref, StoreContext, StoreContextMut, Val, ValRaw, ValType,
 };
 use alloc::sync::Arc;
 use core::ffi::c_void;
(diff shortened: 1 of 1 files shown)


alexcrichton:
What I might recommend for your use case is to store a `TypedFunc<P, R>` within the `T` of your `Store<T>`. That'd probably start as `Option<TypedFunc<...>>` and be `None` on store construction. After instantiation you'd load the function, set it in the store, and then all your host functions have access to that allocation function.

Alternatively if you want the absolute fastest route what I think you'd do is:

* Load a `Module`
* Load the [export and type-check it manually](https://docs.rs/wasmtime/latest/wasmtime/struct.Module.html#method.get_export)
* Load the [`ModuleExport`](https://docs.rs/wasmtime/latest/wasmtime/struct.Module.html#method.get_export_index)
* Use `InstancePre` to instantiate quickly
* Use the [`ModuleExport` to get the exported function](https://docs.rs/wasmtime/latest/wasmtime/struct.Instance.html#method.get_module_export)
* Use [`TypedFunc::new_unchecked`](https://docs.rs/wasmtime/latest/wasmtime/struct.TypedFunc.html#method.new_unchecked) with the resulting function

That front-loads as much as possible at the cost of safety, so I'd recommend using `.get_typed_func` if possible after the initial instantiation of the module.

crates/wasmtime/src/runtime/func.rs
```diff
@@ -2109,6 +2109,26 @@ impl<T> Caller<'_, T> {
...
+    /// string lookups where possible. [`ModuleExport`]s can be obtained by calling
+    /// [`Module::get_export_index`] on the [`Module`] that an instance was instantiated with.
+    ///
+    /// This method will search the module for an export with a matching entity index and return
+    /// the value, if found.
+    ///
+    /// Returns `None` if there was no export with a matching entity index.
+    /// # Panics
+    ///
+    /// Panics if `store` does not own this instance.
```
richardpringle, line 2124: Docs are effectively a copy/paste of `Instance::get_module_export`. Happy to take recommendations or if you want to make a suggestion, I will definitely accept it. I think `see [Instance::get_module_export] for details` would probably also be fine so long as it's clear that the `caller` has access to the instance (wasn't clear to me without looking at the code).

richardpringle:
> Looks good to me! Mind adding a small smoke test too?

@alexcrichton not sure if this is what you meant by a smoke test, but it works as one, haha 😅 .

alexcrichton (approved):
Thanks!

PR #9528, opened 2024-10-31
Merge static/dynamic guard size options

alexcrichton:
This commit is the first of what will likely be a few to refactor the memory-related configuration options in Wasmtime. The end goal of these refactorings is to fix some preexisting issues and additionally make the configuration easier to understand for both users and implementors alike. First on the chopping block here is to merge the `dynamic_memory_guard_size` and `static_memory_guard_size` options into one option. AFAIK there's not a strong reason to have separate configuration options for these so it's hopefully simpler to have a single `memory_guard_size` option which applies to all linear memories equally.

I'll note that the old CLI options are preserved but are documented as deprecated. We don't currently warn on using "deprecated options" so for now the old options are just documented as deprecated and are otherwise silently accepted.


pchickey (approved):
Thanks, this is a nice simplification

Issue #3542, opened 2021-11-17
ISLE: support a block-comment syntax


alexcrichton:
I've noticed that `;;` works for comments, but I think it would also be nifty to support block comments along the same lines as wasm with `(; ... ;)` to handle blocks of code during development as well.

Kmeakin:
Closed by https://github.com/bytecodealliance/wasmtime/commit/4ee35532ac23670f23a61f2a23e497d69ce2c531

PR #9529, opened 2024-10-31
ISLE: Add block comments

Kmeakin:
Solves https://github.com/bytecodealliance/wasmtime/issues/3542.

Also adjusts the lexical syntax to be closer to wasm while I was at it ("\f" is no longer considered whitespace, but no-one uses that character anyway)

We could also align the line-comment syntax with wasm (require two semicolons to start a line comment instead of one), but that would IMO create too much churn


cranelift/isle/isle/src/lexer.rs
```diff
@@ -282,7 +312,7 @@ mod test {
     #[test]
     fn lexer_basic() {
         assert_eq!(
-            lex(";; comment\n; another\r\n   \t(one two three 23 -568  )\n"),
+            lex(";; comment\n; another\r\n   \t(one two three (; block comment ;) 23 -568  )\n"),
```
cfallin, line 315: Can we add a test for the nested block comment case as well?

cranelift/isle/docs/language-reference.md
```diff
@@ -1430,10 +1430,30 @@ for more information regarding the trie construction algorithm.
...
+               | "\r"
+
+<comment> ::= <line-comment> | <block-comment>
+
+<line-comment> ::= ";" <line-char>* (<newline> | eof)
+<newline> ::= "\n" | "\r"
+<line-char> ::= <any character other than "\n" or "\r">
+
+<block-comment> ::= "(;" <block-char>* ";)"
+<block-char> ::= <any character other than ";" or "(">
```
cfallin, line 1453: It looks like the parser below supports nested block comments; can we represent this in the BNF?

(Maybe something like `<block-char> += <block-comment>`, i.e., a nested comment is one `<block-char>` in the parse tree)

cfallin (approved):
Thanks!

cranelift/isle/docs/language-reference.md
Kmeakin: The nesting is represented by the `| <block-comment>` alternative. I copied pretty much directly from the wasm spec: https://webassembly.github.io/spec/core/text/lexical.html#comments

cfallin: Ah, I missed the last option -- very sorry. Thanks, this looks right!

cfallin (approved):

PR #9530, opened 2024-10-31
Remove the `wasmtime_environ::TablePlan` type

alexcrichton:
In my quest to simplify memory configuration and how things work internally in Wasmtime one thing I've identified to accomplish is the removal of the `TablePlan` and `MemoryPlan` types. These introduce an abstraction layer between table/memory implementations and `Tunables` and personally I find it simpler to directly reference `Tunables` and use that instead. The goal of this commit is to plumb `Tunables` closer to where it's directly read by removing the "indirection" through the `*Plan` types.

The `TablePlan` and `MemoryPlan` types are pervasively used throughout Wasmtime so instead of having one large commit delete everything this is instead a piecemeal approach to incrementally get towards the goal of removal. Here just `TablePlan` is removed and `Tunables` is plumbed in a few more places. I plan to also in the future remove `TableStyle` and `MemoryStyle` in favor of directly reading `Tunables` but that's left for a future commit. For now `TableStyle` persists and its usage is a bit odd in isolation but I plan to follow this up with the removal of `TableStyle`.

### DIFF
--- a/crates/cranelift/src/func_environ.rs
+++ b/crates/cranelift/src/func_environ.rs
@@ -701,7 +701,7 @@ impl<'module_environment> FuncEnvironment<'module_environment> {
 
     /// Get the Table for the given index.
     fn table(&self, index: TableIndex) -> Table {
-        self.module.table_plans[index].table
+        self.module.tables[index]
     }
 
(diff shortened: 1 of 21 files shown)


cfallin (approved):
Looks great, thanks!

PR #9531, opened 2024-10-31
Remove `wasmtime_environ::TableStyle`

alexcrichton:
This is a follow-up to #9530 to remove the
`wasmtime_environ::TableStyle` type. This type was added quite a long time ago for future-proofing other styles of tables but at this point it's simpler to remove the type as adding another style of table hasn't been on the "table" (heh) for quite some time. This intends to further the goal of #9530 of plumbing `Tunables` closer to where it's ready to avoid extra layers of abstraction between the configuration options and what is processing them.


abrown (approved):
Makes sense to me!

PR #9533, opened 2024-10-31
Fix custom-page-sizes + pooling-allocator + CoW

alexcrichton:
This commit fixes a runtime error that happened with the pooling allocator when combined with the custom-page-sizes proposal. The bug that happened was that sizes flowing into the pooling allocator were no longer host-page-size-aligned which caused syscalls to return an error unexpectedly. This meant that situations which were supposed to work were in fact not working.

The fix in this commit is to page-align incoming sizes into the CoW-pieces of memory management. This is coupled with a few more assertions that the `accessible` field is always page-aligned, as expected.

### DIFF
--- a/tests/all/pooling_allocator.rs
+++ b/tests/all/pooling_allocator.rs
@@ -1259,7 +1259,7 @@ fn tricky_empty_table_with_empty_virtual_memory_alloc() -> Result<()> {
 }
 
 #[test]
-fn custom_page_sizes_resuing_same_slot() -> Result<()> {
+fn custom_page_sizes_reusing_same_slot() -> Result<()> {
     let mut config = Config::new();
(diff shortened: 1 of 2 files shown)


tests/all/pooling_allocator.rs
```diff
@@ -1257,3 +1257,48 @@ fn tricky_empty_table_with_empty_virtual_memory_alloc() -> Result<()> {
     Instance::new(&mut store, &module, &[])?;
     Ok(())
 }
+
+#[test]
+fn custom_page_sizes_resuing_same_slot() -> Result<()> {
```
fitzgen, line 1262:
```suggestion
fn custom_page_sizes_reusing_same_slot() -> Result<()> {
```

fitzgen (approved):
Thanks!

PR #9536, opened 2024-10-31
Enable some Clippy conversion lints for `wasmtime-cranelift`

alexcrichton:
Similar to how `wasmtime::runtime` has a few off-by-default lints turned on for it do the same for the compilation phase of `wasmtime-cranelift`. This is intended to help weed out lossy `as` casts and instead steer users to `from` or `try_from` conversions.

### DIFF
--- a/crates/cranelift/src/lib.rs
+++ b/crates/cranelift/src/lib.rs
@@ -3,6 +3,10 @@
 //! This crate provides an implementation of the `wasmtime_environ::Compiler`
 //! and `wasmtime_environ::CompilerBuilder` traits.
 
+// See documentation in crates/wasmtime/src/runtime.rs for why this is
+// selectively enabled here.
+#![warn(clippy::cast_possible_truncation, clippy::cast_sign_loss)]
+
 use cranelift_codegen::{
     binemit,
     cursor::FuncCursor,
--- a/cranelift/codegen/Cargo.toml
+++ b/cranelift/codegen/Cargo.toml
@@ -17,7 +17,7 @@ rust-version.workspace = true
 workspace = true
 
 [package.metadata.docs.rs]
-# Ask Cargo to build docs with the feature `all-arch` 
+# Ask Cargo to build docs with the feature `all-arch`
 features = ["all-arch"]
 
 [dependencies]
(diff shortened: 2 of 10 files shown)


pchickey (approved):

crates/cranelift/src/lib.rs
```diff
@@ -3,6 +3,10 @@
 //! This crate provides an implementation of the `wasmtime_environ::Compiler`
 //! and `wasmtime_environ::CompilerBuilder` traits.
 
+// See documentation in crates/wasmtime/src/runtime.rs for why this is
+// selectively enabled here.
+#![warn(clippy::cast_possible_truncation, clippy::cast_sign_loss)]
```
fitzgen, line 8: We can't do this in `Cargo.toml` right? You can't use workspace lints and then also add some more, correct?

alexcrichton: That's right yeah, I believe that this is the intended method of configuring lints if workspace lints are inherited and then per-crate configuration is desired

fitzgen: We might want to make these `deny` since CI is going to gate on them anyways, and since we are explicitly naming and opting into them, we don't have to worry about warnings on this rustc version vs that rustc version.

alexcrichton: Personally I prefer to leave lints like this as warn instead of deny because in the middle of development it can be useful to not have to handle everything and defer it to later. Could you elaborate though on the rustc version part? I don't think I fully understand that.

fitzgen: > Personally I prefer to leave lints like this as warn instead of deny because in the middle of development it can be useful to not have to handle everything and defer it to later.

Fair, I don't have strong opinions here.

> Could you elaborate though on the rustc version part? I don't think I fully understand that.

Oh I just mean how we do `-D warnings` in CI and how, if we didn't pin `rustc` versions, then updating to new `rustc`s could introduce new warnings and break CI despite no code changes to the repo itself. But that kind of update-the-version breakage isn't a concern for specific clippy lints that we enable, because we aren't opting into all default clippy lints, just specific ones that we know we are abiding by.

PR #9543, opened 2024-11-02
More refactoring to remove `MemoryStyle`

alexcrichton:
This commit has more refactoring to simplify handling of linear memories in Wasmtime. The end goal is to remove the static/dynamic distinction and simplify the internal implementation and external interface. This is along the same lines of #9528, #9532, #9531, and #9530. The goal of this PR is a series of standalone commits which work towards futher "pushing down" the usage of `MemoryStyle` down to the consumers of bounds check code. The goal here is to remove layers of abstraction and instead unify everything around the same abstraction of `Tunables` and `wasmtime_environ::Memory`. Many fields in Winch/Cranelift `HeapData` have been removed in favor of storing `Memory` directly.

This PR affects generated golden tests because previously the global value for the heap bound was not generated if the heap was a "static" heap. With the eventual removal of static/dynamic distinction it's easier to just always generate this and have it not get used most of the time. Thus while the clif output tests are changing none of the actual generated code itself should be changing.

### DIFF
--- a/crates/cranelift/src/func_environ.rs
+++ b/crates/cranelift/src/func_environ.rs
@@ -2373,7 +2373,8 @@ impl<'module_environment> crate::translate::FuncEnvironment
 
         // If we have a declared maximum, we can make this a "static" heap, which is
         // allocated up front and never moved.
-        let (style, offset_guard_size) = MemoryStyle::for_memory(memory, self.tunables);
+        let style = MemoryStyle::for_memory(memory, self.tunables);
+        let offset_guard_size = self.tunables.memory_guard_size;
         let (heap_style, readonly_base, base_fact, memory_type) = match style {
             MemoryStyle::Dynamic { .. } => {
                 let heap_bound = func.create_global_value(ir::GlobalValueData::Load {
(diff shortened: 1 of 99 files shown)


crates/cranelift/src/func_environ.rs
```diff
@@ -2369,25 +2361,23 @@ impl<'module_environment> crate::translate::FuncEnvironment
             }
         };
 
-        let page_size_log2 = memory.page_size_log2;
+        let heap_bound = func.create_global_value(ir::GlobalValueData::Load {
+            base: ptr,
+            offset: Offset32::new(current_length_offset),
+            global_type: pointer_type,
+            flags: MemFlags::trusted(),
+        });
```
abrown, line 2369: It's this extra global that is causing the changes in the disassembly tests, right? E.g., `gv5`?

alexcrichton: Right yeah this is the global being added. I [wrote up a bit above about this](https://github.com/bytecodealliance/wasmtime/pull/9543#issue-2630163393) but it's essentially easier given the current code shape to assume that the global value is always there rather than calculate if it'll be needed in a function body and create it and/or lazily creating it on-demand. My hope is that the "overhead" here is just a few bytes in a vector per-function which in theory doesn't have too much impact, but I'll admit I haven't measured.

abrown (approved):
Well, beyond my question about the additional global, which I'm not terribly opposed to anyways, I think this makes sense.

alexcrichton:
@fitzgen mind giving a once-over confirmation here as well?

fitzgen (approved):
LGTM!

PR #9545, opened 2024-11-02
Remove static/dynamic memories from public docs

alexcrichton:
This commit removes the terminology of "static" and "dynamic" memories from the public-facing documentation of Wasmtime, notably on the `Config` structure and its various configuration settings. The goal of this commit is in the same vein as #9543 which is to simplify the memory settings of Wasmtime for users in this case.

This change doesn't actually have any code changes beyond renames (and handling now-deprecated CLI options). The goal of this commit is to basically rewrite how we document the effect of various settings of Wasmtime. Notably:

* `Config::static_memory_maximum_size` is now `memory_reservation`.
* `Config::static_memory_forced` is now `memory_may_move`.
* `Config::dynamic_memory_reserved_for_growth` is now `memory_reservation_for_growth`.

Documentation for all of these options has been rewritten and updated to take into account the removal of "dynamic" and "static" terminology. Additionally more words have been written about the various effects of each setting and how things related to wasm features such as index type sizes and custom page sizes.

The rewritten documentation is intended to basically already match what Wasmtime does today. I believe that all of these settings are useful in one form or another so none have been dropped but the updated documentation is intended to help simplify the mental model for how they're processed internally and how they affect allocations and such.

### DIFF
--- a/crates/cli-flags/src/lib.rs
+++ b/crates/cli-flags/src/lib.rs
@@ -42,7 +42,8 @@ wasmtime_option_group! {
         /// Optimization level of generated code (0-2, s; default: 2)
         pub opt_level: Option<wasmtime::OptLevel>,
 
-        /// Do not allow memories to grow beyond `-O memory-reservation`
+        /// Do not allow Wasm linear memories to move in the host process's
+        /// address space.
         pub memory_may_move: Option<bool>,
 
         /// Initial virtual memory allocation size for memories.
--- a/crates/cli-flags/src/lib.rs
+++ b/crates/cli-flags/src/lib.rs
@@ -42,19 +42,17 @@ wasmtime_option_group! {
         /// Optimization level of generated code (0-2, s; default: 2)
         pub opt_level: Option<wasmtime::OptLevel>,
 
-        /// Force using a "static" style for all wasm memories
-        pub static_memory_forced: Option<bool>,
+        /// Do not allow memories to grow beyond `-O memory-reservation`
+        pub memory_may_move: Option<bool>,
 
-        /// Maximum size in bytes of wasm memory before it becomes dynamically
-        /// relocatable instead of up-front-reserved.
-        pub static_memory_maximum_size: Option<u64>,
+        /// Initial virtual memory allocation size for memories.
+        pub memory_reservation: Option<u64>,
+
+        /// Bytes to reserve at the end of linear memory for growth into.
+        pub memory_reservation_for_growth: Option<u64>,
 
         /// Size, in bytes, of guard pages for linear memories.
         pub memory_guard_size: Option<u64>,
-
-        /// Bytes to reserve at the end of linear memory for growth for dynamic
-        /// memories.
-        pub dynamic_memory_reserved_for_growth: Option<u64>,
 
         /// Indicates whether an unmapped region of memory is placed before all
         /// linear memories.
@@ -170,6 +168,15 @@ wasmtime_option_group! {
 
         /// DEPRECATED: Use `-Cmemory-guard-size=N` instead.
         pub static_memory_guard_size: Option<u64>,
+
+        /// DEPRECATED: Use `-Cmemory-may-move` instead.
+        pub static_memory_forced: Option<bool>,
+
+        /// DEPRECATED: Use `-Cmemory-reservation=N` instead.
+        pub static_memory_maximum_size: Option<u64>,
+
+        /// DEPRECATED: Use `-Cmemory-reservation-for-growth=N` instead.
+        pub dynamic_memory_reserved_for_growth: Option<u64>,
     }
 
     enum Optimize {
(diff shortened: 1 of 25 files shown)


crates/wasmtime/src/config.rs
```diff
@@ -1283,126 +1283,210 @@ impl Config {
...
+    /// after initial allocation a linear memory is never allowed to move in
+    /// memory. See more information on [`Config::memory_reservation`] for
+    /// various effects on performance this can have.
     ///
-    /// For the difference between static and dynamic memories, see the
-    /// [`Config::static_memory_maximum_size`].
-    pub fn static_memory_forced(&mut self, force: bool) -> &mut Self {
-        self.tunables.static_memory_bound_is_maximum = Some(force);
+    /// The default value for this option is `false`.
+    pub fn memory_reservation_is_maximum(&mut self, enable: bool) -> &mut Self {
```
fitzgen, line 1488: I don't love the name `memory_reservation_is_maximum`. Here are a few alternative bike sheds:

* `memory_reservation_is_final`
* `memory_reservation_is_immovable`
* `memory_reservation_moves`
* `memory_reservation_relocates`
* `memory_reservation_reallocs`
* `memory_is_final`
* `memory_is_immovable`
* `memory_moves`
* `memory_relocates`
* `memory_reallocs`

I don't love any of these. Soft preference for `memory_moves`, I guess? Thoughts?

fitzgen (approved):
Thanks! Some bike shedding inline below, interested in what you think.

crates/wasmtime/src/config.rs
alexcrichton: Ok I've settled on `memory_may_move`, mind doing another pass over the docs I have for the method?

crates/cli-flags/src/lib.rs
```diff
@@ -42,20 +42,18 @@ wasmtime_option_group! {
         /// Optimization level of generated code (0-2, s; default: 2)
         pub opt_level: Option<wasmtime::OptLevel>,
 
-        /// Force using a "static" style for all wasm memories
-        pub static_memory_forced: Option<bool>,
+        /// Do not allow memories to grow beyond `-O memory-reservation`
+        pub memory_may_move: Option<bool>,
```
fitzgen, line 46: This should be reworded, maybe

```suggestion
        /// Do not allow Wasm linear memories to move in the host
        /// process's address space.
        pub memory_may_move: Option<bool>,
```

crates/wasmtime/src/config.rs
```diff
@@ -1304,116 +1301,217 @@ impl Config {
...
-    /// exactly what's used by the wasm itself.
-    ///
-    /// For 32-bit memories a static memory must contain at least 4GB of
-    /// reserved address space plus a guard page to elide any bounds checks at
-    /// all. Smaller static memories will use similar bounds checks as dynamic
-    /// memories.
+    /// * Memory `$a` initially allocates 4 WebAssembly pages (256KiB) and can
+    ///   grow up to 4GiB, the limit of the 32-bit index space.
+    /// * Memory `$b` initially allocates 4096 WebAssembly pages, but in this
+    ///   case the page size is 1, so it's 4096 bytes. Memory can also grow no
```
fitzgen, line 1337: tiny edit to clarify that pagesize=1 is only for this memory

```suggestion
    ///   case its page size is 1, so it's 4096 bytes. Memory can also grow no
```

crates/wasmtime/src/config.rs
```diff
@@ -1304,116 +1301,217 @@ impl Config {
...
     ///
-    /// This configuration option enables selecting the "static" option for all
-    /// linear memories created within this `Config`. This means that all
-    /// memories will be allocated up-front and will never move. Additionally
-    /// this means that all memories are synthetically limited by the
-    /// [`Config::static_memory_maximum_size`] option, regardless of what the
-    /// actual maximum size is on the memory's original type.
+    /// * Modules can be compiled with static knowledge the base pointer of
+    ///   linear memory never changes to enable optimizations such as
+    ///   loop-invariant-code-motion (hosting the base pointer out of a loop).
```
fitzgen, line 1505:
```suggestion
    ///   loop-invariant code motion (hoisting the base pointer out of a loop).
```

crates/wasmtime/src/config.rs
```diff
@@ -1304,116 +1301,217 @@ impl Config {
...
     ///
-    /// For the difference between static and dynamic memories, see the
-    /// [`Config::static_memory_maximum_size`].
-    pub fn static_memory_forced(&mut self, force: bool) -> &mut Self {
-        self.tunables.static_memory_bound_is_maximum = Some(force);
+    /// * Memories cannot grow in excess of their original allocation. This
+    ///   means that [`Config::memory_reservation`] and
+    ///   [`Config::memory_reservation_for_growth`] may need tuning to ensure
+    ///   the memory work at runtime.
```
fitzgen, line 1510:
```suggestion
    ///   the memory configuration works at runtime.
```

fitzgen (approved):
nitpicks on docs wording below, but lgtm

PR #9546, opened 2024-11-02
Minor ergonomic improvements to pooling config

alexcrichton:
* Change `Config::allocation_strategy` to take `impl Into<...>` to avoid importing `InstanceAllocationStrategy`.
* Add `PoolingAllocationConfig::new` as an alias for `::default`

### DIFF
--- a/crates/wasmtime/src/config.rs
+++ b/crates/wasmtime/src/config.rs
@@ -2627,7 +2627,7 @@ impl PoolingAllocationConfig {
     /// Returns a new configuration builder with all default settings
     /// configured.
     pub fn new() -> PoolingAllocationConfig {
-        return PoolingAllocationConfig::default();
+        PoolingAllocationConfig::default()
     }
 
     /// Configures the maximum number of "unused warm slots" to retain in the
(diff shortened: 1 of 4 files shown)


crates/wasmtime/src/config.rs
```diff
@@ -2613,6 +2624,12 @@ pub struct PoolingAllocationConfig {
 
 #[cfg(feature = "pooling-allocator")]
 impl PoolingAllocationConfig {
+    /// Returns a new configuration builder with all default settings
+    /// configured.
+    pub fn new() -> PoolingAllocationConfig {
+        return PoolingAllocationConfig::default();
```
abrown, line 2630:
```suggestion
        PoolingAllocationConfig::default()
```

crates/wasmtime/src/config.rs
```diff
@@ -36,6 +36,7 @@ pub use wasmtime_environ::CacheStore;
 
 /// Represents the module instance allocation strategy to use.
 #[derive(Clone)]
+#[non_exhaustive]
```
abrown, line 39: Not sure I get why this must now be non-exhaustive?

abrown (approved):

crates/wasmtime/src/config.rs
alexcrichton: Ah this should have been non_exhaustive before because one variant is conditionally defined, so I figured I'd go ahead and fix it while I was here

PR #9554, opened 2024-11-05
winch: Sync registers and locals before fuel check

saulecabrera:
This commit fixes a fuzz bug in which the stack was misaligned when calling the out-of-fuel builtin function.

The misalignment was introduced by a erroneous handling of the the control flow merge introduced by the fuel check conditional. In general, prior to every branch emission, a spill to memory is needed to avoid issues at the control flow merge.

Note that we don't have many cases like this one in Winch's codebase (3 in total), however as a follow-up, it's probably worth considering introducing a stronger abstraction around branching to ensure that this case is handled whenever an arbitrary branch needs to be introduced. This change solely focuses on the fix and does not introduce any refactoring. I plan to follow-up with investigating a better branching strategy, since we would need to introduce a similar pattern for epoch handling.

I used `wasm-tools shink` to shrink the original program, which I decided to add as part of an integration test.


tests/all/fuel_stack_alignment.wat
saulecabrera, line 1: @alexcrichton I'm not sure if you have thoughts on introducing test fixtures like this as I didn't see this pattern used for the rest of the integration tests. I tried to shrink the original program using `wasm-tools shrink` and it was able to shrink it by 30%, if you think the test programs should be as small as possible I can definitely spend a bit more time trying to reduce it further.

alexcrichton:
@saulecabrera would you be ok backporting this to the release-27 branch? If you're busy I'm happy to take that on myself

alexcrichton:
Indeed, and thanks!

PR #9565, opened 2024-11-05
Refactor `vm::Instance` for better `Store` safety

fitzgen:
At various times, for example in libcalls, we need to go from a raw `vmctx` pointer that Wasm gave us to both an `&mut Instance` and an `&mut Store{Opaque,Context}`. This is a pretty fundamental unsafe aspect of implementing a language runtimne in Rust, but we can tighten things up a bit and make them a little safer than they currently are. This commit is such an attempt and is notably tackling the issue of creating multiple store borrows after we have an instance borrow.

This commit makes the following changes:

* `Instance` doesn't expose a method to get the raw `*mut dyn VMStore` pointer or otherwise create a store borrow.

* You cannot construct an `Instance` directly from a `vmctx` pointer.

* There is now an `InstanceAndStore` type that represents unique access to both an `Instance` and a `Store`.

* You can (with `unsafe`) create an `InstanceAndStore` from a raw `vmctx` pointer. This generally only happens inside a couple "bottlenecks", not all throughout the codebase, like the shared plumbing code for libcalls.

* The `InstanceAndStore` can be unpacked into a mutable borrow of an `Instance` and a mutable borrow of a `Store`. This unpacking takes holds a mutable borrow of the original `InstanceAndStore`, so double borrows of the store are no longer a concern, so long as you don't use `unsafe` to create new `InstanceAndStore`s for the same `vmctx`.

* All `Instance` methods and functions that previously would unsafely turn the `Instance`'s internal store pointer into a store borrow to perform some action now take a store argument, and this store is threaded around as necessary.

Altogether, I feel that we've ended up with an architecture that is a safety improvement over where we were previously, and will help us properly avoid Rust UB.

For what its worth, I do not think this is the end state. I foresee at least two additional follow ups that I unfortunately do not currently have time for:

1. Create `ComponentInstanceAndStore`, which is the same thing that this commit did but for components. I started on this but ran out of fuel while trying to update macros for all the component shims and transcoders.

2. Stop dealing with `&mut vm::Table`s and `&mut vm::Global`s and all that in the runtime. Instead just deal with indices, similar to how things are structured in the host API level. This refactoring was not previously possible due to (1) the `wasmtime` versus `wasmtime-runtime` crate split and (2) the lack of `StoreOpaque`s threaded through the VM internals. The first blocker was addressed a few months ago, this commit removes the second blocker. This will still be a pretty large refactoring though, but I think ultimately will be worth it.

### DIFF
--- a/crates/wasmtime/src/runtime/vm/instance.rs
+++ b/crates/wasmtime/src/runtime/vm/instance.rs
@@ -125,17 +125,10 @@ impl InstanceAndStore {
         f(&mut *ptr)
     }
 
-    /// Get the underlying `dyn VMStore` trait object for this
-    /// `InstanceAndStore`.
-    #[inline]
-    pub(crate) fn store_mut(&mut self) -> &mut (dyn VMStore + 'static) {
-        unsafe { &mut *self.store_ptr() }
-    }
-
     /// Unpacks this `InstanceAndStore` into its underlying `Instance` and `dyn
     /// VMStore`.
     #[inline]
-    pub(crate) fn unpack_mut(&mut self) -> (&mut Instance, &mut (dyn VMStore + 'static)) {
+    pub(crate) fn unpack_mut(&mut self) -> (&mut Instance, &mut dyn VMStore) {
         unsafe {
             let store = &mut *self.store_ptr();
             (&mut self.instance, store)
@@ -168,7 +161,7 @@ impl InstanceAndStore {
     /// functions are shared amongst threads and don't all share the same
     /// store).
     #[inline]
-    fn store_ptr(&self) -> *mut (dyn VMStore + 'static) {
+    fn store_ptr(&self) -> *mut dyn VMStore {
         let ptr = unsafe {
             *self
                 .instance
--- a/crates/wasmtime/src/runtime/vm/instance.rs
+++ b/crates/wasmtime/src/runtime/vm/instance.rs
@@ -2,7 +2,6 @@
 //! wasm module (except its callstack and register state). An
 //! `InstanceHandle` is a reference-counting handle for an `Instance`.
 
-use crate::prelude::*;
 use crate::runtime::vm::const_expr::{ConstEvalContext, ConstExprEvaluator};
 use crate::runtime::vm::export::Export;
 use crate::runtime::vm::memory::{Memory, RuntimeMemoryCreator};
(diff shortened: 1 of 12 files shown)


crates/wasmtime/src/runtime/vm/instance.rs
```diff
@@ -35,9 +36,149 @@ use wasmtime_environ::{
...
+
+    /// Unpacks this `InstanceAndStore` into its underlying `Instance` and `dyn
+    /// VMStore`.
+    #[inline]
+    pub(crate) fn unpack_mut(&mut self) -> (&mut Instance, &mut (dyn VMStore + 'static)) {
+        unsafe {
+            let store = &mut *self.store_ptr();
+            (&mut self.instance, store)
+        }
+    }
```
alexcrichton, line 143: Can the `'static` here be removed? In theory that's a "lie" because we don't constrain `T: 'static` in the `Store<T>`.

alexcrichton (commented):
Looks great 👍 

One comment about the `'static` though on the `dyn VMStore` type though.

crates/wasmtime/src/runtime/vm/instance.rs
fitzgen: Unfortunately, I don't think it can be or else the borrow is invariant and the `InstanceAndStore` is unusable after calling unpack once, which isn't acceptable for `Caller::with`, which needs to access the pair multiple times. It miiight have been an issue in certain libcalls too, I can't quite remember.

Morally, I see this as a "smaller lie" than our current status quo where we just whip up a borrow from a raw pointer whenever we feel like it...

I can try poking at it again, but if I can't figure out a way to remove the `'static`, do you think this is a blocker to landing this?

alexcrichton: Hm this is surprising. I'm worried about labeling this `'static` when it wasn't since that to me isn't a failure condition we worried about before (I think at least?). Mind trying to remove it and see what happens?

I'm also surprised if `InstanceAndStore` is unusable after one call, as that probably means we have this whole tying typed incorrectly anyway...

fitzgen: I guess we could maybe get rid of `InstanceAndStore` and just have

```rust
impl Instance {
    pub(crate) unsafe fn from_vmctx<R>(vmctx: *mut VMContext, f: impl FnOnce(&mut StoreOpaque, &mut Instance) -> R) -> R {
        // ...
    }
}
```

(And another version for `StoreInner<T>`)

fitzgen: This would require moving all the `VMStore` trait methods to `StoreOpaque`

alexcrichton (approved):

Issue #9568, opened 2024-11-05
Refactor the `MemoryCreator` trait to support immovable memories


alexcrichton:
In #9569 the `memory_may_move` configuration knob was plumbed into memories themselves to limit the growth of the memory. This knob is not surfaced in the custom `MemoryCreator` trait, however, and memories created from that trait are pessimistically assumed to always possibly move in memory. This issue is intended to track refactoring this trait to improve this situation and enable plumbing the information a bit further.

Ideally the `MemoryCreator` and `RuntimeMemoryCreator` traits are merged. After looking at this for a bit this'll require some deeper refactoring to keep things making sense.

PR #9569, opened 2024-11-05
Don't use `MemoryStyle` for heap base pointer relocations

### LINKED ISSUE #4267: shared memory: allow threads and memory64 proposals to coexist
Currently, threads and memory64 don't work together: not all memory64 memories will fit into 4GB so Wasmtime will classify them as dynamic to allow the OS to move them. Shared memories cannot be moved; their base pointer must remain the same throughout their lifetime, so Wasmtime only generates shared memories that are static. When memory64 is enabled, however, a memory could be created with a large enough size that its memory plan would be dynamic. If this happens to be a shared memory, Wasmtime returns an error.

@alexcrichton's proposed fix is to create static memory plans for all shared memories and limit the static memory size to the engine's configuration. At runtime, a threads + memory64 program attempting to use above this limit will fail to `memory.grow` once the engine-enforced static memory limit is reached, regardless of the memory size specified by the Wasm module.

### LINKED ISSUE #9523: Cannot instantiate a shared memory with a custom page size
Given this module:

```wasm
(module
  (memory 1 2 shared (pagesize 1))
)
```

this cannot be instantiated:

```
$ wasmtime run -Wcustom-page-sizes foo.wat

Caused by:
    0: failed to instantiate "foo.wat"
    1: shared memory can only be built from a static memory allocation
```

This is due to the fact that custom page sizes currently force a "dynamic" memory where the base pointer can change while shared memories are required to be "static" where the base pointer cannot change. Fixing this will require refactoring exactly how memories are represented internally within Wasmtime, but in the meantime I wanted to file an issue about this to have on-hand.

alexcrichton:
Instead add a helper method to `Memory` which indicates whether the base pointer of memory can move. Use this and plumb the result around to the various locations that require it. This improves the `readonly` application of the base pointer in Cranelift by having the optimization kick in in more scenarios. It additionally fixes combining shared linear memories with 64-bit addressing or a page size of 1 by adding a runtime check before relocation of a linear memory that it's allowed to do so.

A few codegen tests are added to ensure that `readonly` is applied where it wasn't previously and additionally a new `*.wast` test was added with the cross product of various features of linear memories and some basic tests to ensure that the memories all work as expected.

This refactoring fixes two preexisting issues about `shared` memories requiring a "static" memory style. The restriction is now based on whether the pointer can relocate or not and that's upheld via the new trait method added here.

To account for these bug fixes the fuzzers have been updated to allow mixing the `threads` proposal with `memory64` or `custom-page-sizes`.

Closes #4267
Closes #9523


alexcrichton:
I'll also notably point out that this is incompatible with `MemoryCreator` so I've filed #9568 to track that. I think there needs to be a good bit of refactoring of the internals of how we represent memory so I hope to tackle that next-ish.

fitzgen (approved):
Really like this clean up, everything is much easier to follow now. Thanks!

PR #9571, opened 2024-11-06
Add `Module::deserialize_open_file`

adambratschikaye:
Add an API to deserialize a `Module` from an already opened file. This can be useful in situations where `wasmtime` is running in a sandboxed environment with limited access to the file system. Then another process can handle opening the files and passing them to `wasmtime`.

### DIFF
--- a/crates/wasmtime/src/runtime/module.rs
+++ b/crates/wasmtime/src/runtime/module.rs
@@ -444,6 +444,13 @@ impl Module {
     ///
     /// [`deserialize_file`]: Module::deserialize_file
     ///
+    /// Note that the corresponding will be mapped as private writeable
+    /// (copy-on-write) and executable. For `windows` this means the file needs
+    /// to be opened with at least `FILE_GENERIC_READ | FILE_GENERIC_EXECUTE`
+    /// [`access_mode`].
+    ///
+    /// [`access_mode`]: https://doc.rust-lang.org/std/os/windows/fs/trait.OpenOptionsExt.html#tymethod.access_mode
+    ///
     /// # Unsafety
     ///
     /// All of the reasons that [`deserialize_file`] is `unsafe` applies to this
--- a/crates/wasmtime/src/runtime/module.rs
+++ b/crates/wasmtime/src/runtime/module.rs
@@ -1,4 +1,6 @@
 use crate::prelude::*;
+#[cfg(feature = "std")]
+use crate::runtime::vm::open_file_for_mmap;
 use crate::runtime::vm::{CompiledModuleId, MmapVec, ModuleMemoryImages, VMWasmCallFunction};
 use crate::sync::OnceLock;
 use crate::{
@@ -15,7 +17,7 @@ use core::fmt;
 use core::ops::Range;
 use core::ptr::NonNull;
 #[cfg(feature = "std")]
-use std::path::Path;
+use std::{fs::File, path::Path};
 use wasmparser::{Parser, ValidPayload, Validator};
 use wasmtime_environ::{
     CompiledModuleInfo, EntityIndex, HostPtr, ModuleTypes, ObjectKind, TypeTrace, VMOffsets,
@@ -343,7 +345,8 @@ impl Module {
     /// state of the file.
     #[cfg(all(feature = "std", any(feature = "cranelift", feature = "winch")))]
     pub unsafe fn from_trusted_file(engine: &Engine, file: impl AsRef<Path>) -> Result<Module> {
-        let mmap = MmapVec::from_file(file.as_ref())?;
+        let open_file = open_file_for_mmap(file.as_ref())?;
+        let mmap = MmapVec::from_file(open_file)?;
         if &mmap[0..4] == b"\x7fELF" {
             let code = engine.load_code(mmap, ObjectKind::Module)?;
             return Module::from_parts(engine, code, None);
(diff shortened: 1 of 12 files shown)


crates/wasmtime/src/runtime/module.rs
```diff
@@ -343,14 +345,35 @@ impl Module {
...
+    /// All the of the reasons [`from_trusted_file`] is `unsafe` apply here as
+    /// well.
+    ///
+    /// [`from_trusted_file`]: Module::from_trusted_file
+    ///
+    #[cfg(all(feature = "std", any(feature = "cranelift", feature = "winch")))]
+    pub unsafe fn from_trusted_open_file(
+        engine: &Engine,
+        file: File,
+        wasm_path: Option<&Path>,
```
bjorn3, line 367: It is not clear to me if this argument needs to match the actual location of the file you pass in or not. Or what happens when it doesn't match.

adambratschikaye: Yeah I'm still cleaning things up a bit. It's there so that we can still use it when calling from `from_trusted_file`, but I think we can remove it from the public version of this one.

crates/wasmtime/src/runtime/vm/sys/windows/mmap.rs
```diff
@@ -10,6 +10,22 @@ use windows_sys::Win32::Foundation::*;
 use windows_sys::Win32::Storage::FileSystem::*;
 use windows_sys::Win32::System::Memory::*;
 
+/// Open a file so that it can be mmap'd for executing.
+///
+/// Open the file with read/execute access and only share for
+/// read. This will enable us to perform the proper mmap below
+/// while also disallowing other processes modifying the file
+/// and having those modifications show up in our address space.
+pub fn open_file_for_mmap(path: &Path) -> Result<File> {
```
adambratschikaye, line 19: Creating this function seems a bit messy and I'm wondering if a cross-platform `open` is fine for Windows anyway. We do say that `deserialize_file` is unsafe if the file is modified while in use.

fitzgen (approved):
LGTM, but I am not sure I understand what you are getting at with regards to windows, so we should resolve that before merging. Thanks!

crates/wasmtime/src/runtime/vm/sys/windows/mmap.rs
adambratschikaye: Yeah I was suggesting we remove this additional code since we say it's `unsafe` to modify the file. But I'm good leaving it in.

alexcrichton: IIRC this is required on Windows not necessarily for safety but just to get anything working. I have a vague recollection that "just" doing `File::open` causes some mmap operation down the line to return an error or similar. I could be misremembering though (and CI should turn up any issues if there are some)

fitzgen: Gotcha. In general, when we can easily and without imposing overhead protect against bad things happening when an `unsafe` contract is violated, we should do it. So let's leave these bits in, even if it turns out they aren't strictly necessary for correctness.

alexcrichton:
Not necessarily obvious to find but the failure is [a Windows-specific one](https://github.com/bytecodealliance/wasmtime/actions/runs/11729597501/job/32675639869#step:5:297)

tests/all/module_serialize.rs
```diff
@@ -105,7 +105,15 @@ fn test_deserialize_from_file() -> Result<()> {
         assert_eq!(func.call(&mut store, ())?, 42);
 
         // Try an alreeady opened file as well.
```
bjorn3, line 107: *already

adambratschikaye: fixed

adambratschikaye:
Yeah @alexcrichton is right - it looks like a simple `File::open` doesn't work for windows because the open options need to be compatible with the options used when creating the mapping. I also had to fix up the new test for this and I added a comment on `deserialize_open_file` on how the file should be opened for windows.

Also ran the full CI on the branch now, so the previous windows failure should be good.

crates/wasmtime/src/runtime/module.rs
```diff
@@ -426,7 +429,35 @@ impl Module {
...
+    /// [`deserialize_file`] in situations where `wasmtime` is running with
+    /// limited file system permissions. In that case a process
+    /// with file system access can pass already opened files to `wasmtime`.
+    ///
+    /// [`deserialize_file`]: Module::deserialize_file
+    ///
+    /// Note that the corresponding will be mapped as private writeable
+    /// (copy-on-write) and executable. For `windows` this means the file needs
+    /// to be opened with at least `FILE_GENERIC_READ | FILE_GENERIC_EXECUTE`
+    /// [`access_mode`].
```
fitzgen, line 450: Does it need to be *writable*? Seems like it should be (and maybe actually is?) *readable* and executable?

fitzgen (commented):
I think there is a writable vs readable swaparoo in the docs. AFAICT, it is readable and executable, not writable and executable, and if that is true then this is good to merge with that doc comment fixed.

If my understanding is off, and we are actually mapping as writable and executable, then I think we want to revisit what we're doing here, since that is a pretty scary widget to expose!

alexcrichton:
Ah I think Nick is on vacation right now so I'll take this over and flag for merge. I think the last doc bit is fine for now and we can always update it as necessary later on.

PR #9574, opened 2024-11-06
Don't use `MemoryStyle` in `MmapMemory`

alexcrichton:
This commit removes the use fo `MemoryStyle` for determining the size of the allocation of a `MmapMemory`. This should semanatically be the same as before but it's hoped that the new version is clearer to understand as it doesn't require going through the `MemoryStyle` abstraction to understand how tunables interact with linear memory. The semantics for the allocation of linear memory are:

    reservation = tunables.memory_reservation
    growth = tunables.memory_reservation_for_growth
    if maximum <= reservation {
        growth = 0
    } else if minimum <= reservation {
        // use settings above
    } else {
        reservation = minimum + growth
    }

This means that the initial memory allocation is always `tunables.memory_reservation` unless that's not large enough to fit the linear memory. In such a situation the other settings, such as `memory_reservation_for_growth`, kick in.

The logic of clamping the maximum and/or dealing with static/dynamic are now all unnecessary and/or tracked elsewhere. For example the clamping nature of `maximum` now happens implicitly by rejecting growth when `memory_may_move` is false which should help keep that complexity localized where necessary rather than conflating multiple settings together.

### DIFF
--- a/crates/wasmtime/src/runtime/vm/memory.rs
+++ b/crates/wasmtime/src/runtime/vm/memory.rs
@@ -13,7 +13,7 @@ use alloc::sync::Arc;
 use core::ops::Range;
 use core::ptr::NonNull;
 use core::time::Duration;
-use wasmtime_environ::{MemoryStyle, Trap, Tunables};
+use wasmtime_environ::{Trap, Tunables};
 
 /// A memory allocator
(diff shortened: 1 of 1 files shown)


fitzgen (approved):
:+1:

PR #9577, opened 2024-11-06
Refactor the internals of `vm::Memory` to simplify the `RuntimeLinearMemory` trait

### LINKED ISSUE #9568: Refactor the `MemoryCreator` trait to support immovable memories

alexcrichton:
This is a series of commits which have the end-goal of removing the `RuntimeLinearMemory` trait and replacing it with `wasmtime::LinearMemory`. In doing this though I wanted to simplify the trait to move more responsibilities up a layer so it's clearer what the safety contracts are and what exactly is required on behalf of embedders. This moves many of the responsibilities to a new `LocalMemory` type rather than having most of the functionality implemented through trait methods. This provides, for example, a vector of having CoW support in Wasmtime without supporting it through `LinearMemory` (or not officially at least).

Each commit here is probably best viewed one-at-a-time since this adds up to a number of changes but I've tried to keep things separated and have tests passing for each individual commit.

Fixes #9568


crates/wasmtime/src/runtime/vm/memory.rs
```diff
@@ -1,6 +1,78 @@
...
+//! │                     │     │              │
+//! └──────────┬──────────┘     └──────────────┘
+//!            │
+//!            │
+//!            ▼
+//! ┌─────────────────────┐
+//! │                     │
+//! │ RuntimeLinearMemory ├─────┬───────┐
+//! │                     │     │       │
+//! └──────────┬──────────┘     └───────┼───────────────┐
```
fitzgen, line 29:
```suggestion
//! │ RuntimeLinearMemory ├─────────────┬───────────────┐
//! │                     │             │               │
//! └──────────┬──────────┘             │               │
```

fitzgen (approved):
Nice! One suggestion on the ascii diagram below, that's it.

PR #9578, opened 2024-11-06
Do not GC `DW_AT_declaration` `DW_TAG_subprogram`s

### LINKED ISSUE #9512: Class member function declarations lost in translation to native DWARF (leading to debugger failures)
Reproduction:
```cpp
; main.cpp
#include "someclass.h"

int main()
{
    return SomeClass::Main();
}

; someclass.h
class SomeClass
{
public:
    int SomeMemberVariable;

    static int Main()
    {
        int x = 1;
        int y = SomeFunc(x);
        return y;
    }
    
    static int SomeFunc(int y)
    {
        int s = SeparateFunction();
        int z = y + 1;
        return z + s;
    }

    static int SeparateFunction();
};

; satellite.cpp
#include "someclass.h"

int SomeClass::SeparateFunction()
{
    SomeClass x = SomeClass();
    return x.SomeMemberVariable;
}
```
```
> & $env:WASI_SDK_PATH/bin/clang main.cpp satellite.cpp -o main.wasm  -g3 -target wasm32-unknown-wasip2
> lldb wasmtime -Ccache=n -Ddebug-info -Oopt-level=0 main.wasm
> (lldb) b Main
> (lldb) b SeparateFunction
> (lldb) r
> (lldb) p x + x ; While broken in Main
> (lldb) c
> (lldb) p x.SomeMemberVariable
```
Expected result: evaluation succeeds.

Actual result:
```
(lldb) p x.SomeMemberVariable
warning: (x86_64) JIT(0x1550de50e60)(0x000001550de50e60) 0x000002a4: DW_AT_specification(0x0000000000000291) has no decl

(lldb) error: <lldb wrapper prefix>:45:11: use of undeclared identifier '$__lldb_local_vars'
   45 |     using $__lldb_local_vars::x;
      |           ^
error: <user expression 1>:1:1: use of undeclared identifier 'x'
    1 | x.SomeMemberVariable
      | ^
```
The tentative root case of this is that debug info translation discards `DW_AT_declaration` DIEs for the member functions, leading to the two CUs (that both have `SomeClass` type DIEs) diverging (and LLDB ultimately caching information from the one first accessed, here `main.cpp`).

SingleAccretion:
Declarations must be consistent across compilation units for any given type. This is the same situation as with, e. g., class member variables (fields).

Furthermore, even for ordinary C declarations, deleting them can theoretically mess up the expression evaluator (since a function that was supposed to be in scope no longer is). This is mostly a theoretical concern, however, as calling functions in the context of translated DWARF is currently deeply broken anyway.

Fixes #9512.

I've also tweaked the dumping of DIE offsets for easier cross-referencing with llvm-dwarfump.


alexcrichton (approved):
Thanks!

PR #9580, opened 2024-11-07
Make source links relative to OUT_DIR

### LINKED ISSUE #9553: Avoid printing full file paths in cranelift generated code
#### Feature

It would be great to avoid printing full paths of source files in the generated code.

#### Benefit

This would allow for better or more correct caching in third party build systems like Bazel or Nix.

#### Implementation

1. The naive solution is to remove references to the source files:

```diff
diff --git a/cranelift/codegen/meta/src/srcgen.rs b/cranelift/codegen/meta/src/srcgen.rs
index d3c321e5b..5b94ddd19 100644
--- a/cranelift/codegen/meta/src/srcgen.rs
+++ b/cranelift/codegen/meta/src/srcgen.rs
@@ -94,7 +94,6 @@ impl Formatter {
         directory: &std::path::Path,
     ) -> Result<(), error::Error> {
         let path = directory.join(&filename);
-        eprintln!("Writing generated file: {}", path.display());
         let mut f = fs::File::create(path)?;
 
         for l in self.lines.iter().map(|l| l.as_bytes()) {
diff --git a/cranelift/isle/isle/src/codegen.rs b/cranelift/isle/isle/src/codegen.rs
index 158285832..d292e43c0 100644
--- a/cranelift/isle/isle/src/codegen.rs
+++ b/cranelift/isle/isle/src/codegen.rs
@@ -127,9 +127,6 @@ impl<'a> Codegen<'a> {
             "// Generated automatically from the instruction-selection DSL code in:",
         )
         .unwrap();
-        for file in &self.files.file_names {
-            writeln!(code, "// - {file}").unwrap();
-        }
 
         if !options.exclude_global_allow_pragmas {
             writeln!(
@@ -641,12 +638,11 @@ impl<L: Length, C> Length for ContextIterWrapper<L, C> {{
                             stack.push((Self::validate_block(ret_kind, body), "", scope));
                         }
 
-                        &ControlFlow::Return { pos, result } => {
+                        &ControlFlow::Return { pos: _pos, result } => {
                             writeln!(
                                 ctx.out,
-                                "{}// Rule at {}.",
+                                "{}",
                                 &ctx.indent,
-                                pos.pretty_print_line(&self.files)
                             )?;
                             write!(ctx.out, "{}", &ctx.indent)?;
                             match ret_kind {
```

2. The behavior could be enabled/disabled based on a crate feature or environment variable.
3. The path could be made relative to `CARGO_MANIFEST_DIR`; alternatively only the `basename` of the source file could be retained and written to the generated files.

Not very familiar with the crate so can't make an educated decision! But this would be a nice improvement in the way of system-independent reproducible builds.

nmattia:
Fixes #9553

When generating code, `cranelift-isle` records the path to the source files. These paths were sometimes absolute, meaning it kept a trace of the system it was run on.

These changes here add helper functions in `Files` to try and strip the value of `$OUT_DIR` (if set) from the file names.

This also removes a log line (which included the full path of files) from `cranelift-codegen-meta`.

### DIFF
--- a/cranelift/isle/isle/src/files.rs
+++ b/cranelift/isle/isle/src/files.rs
@@ -100,12 +100,33 @@ impl Files {
         self.file_names.get(file).map(|x| x.as_str())
     }
 
+    /// Same as `file_name` but try to make the file relative to the project root. Otherwise,
+    /// return the original file name (if found).
+    pub fn file_name_relative(&self, file: usize) -> Option<&str> {
+        self.file_name(file).map(|f| relative(f))
+    }
+
+    /// Try to make file names relative to the project root. Otherwise, return the original file
+    /// names.
+    pub fn file_names_relative(&self) -> Vec<&str> {
+        self.file_names.iter().map(|f| relative(f)).collect()
+    }
+
     pub fn file_text(&self, file: usize) -> Option<&str> {
         self.file_texts.get(file).map(|x| x.as_str())
     }
 
     pub fn file_line_map(&self, file: usize) -> Option<&LineMap> {
         self.file_line_maps.get(file)
+    }
+}
+
+/// If OUT_DIR is set, strips it from the file name.
+/// Otherwise returns the original file name.
+pub fn relative(file_name: &str) -> &str {
+    match std::env::var("OUT_DIR") {
+        Err(_) => file_name,
+        Ok(root) => file_name.strip_prefix(&root).unwrap_or(file_name).into(),
     }
 }
 
--- a/cranelift/codegen/meta/src/srcgen.rs
+++ b/cranelift/codegen/meta/src/srcgen.rs
@@ -94,7 +94,6 @@ impl Formatter {
         directory: &std::path::Path,
     ) -> Result<(), error::Error> {
         let path = directory.join(&filename);
-        eprintln!("Writing generated file: {}", path.display());
         let mut f = fs::File::create(path)?;
 
         for l in self.lines.iter().map(|l| l.as_bytes()) {
--- a/cranelift/isle/isle/src/codegen.rs
+++ b/cranelift/isle/isle/src/codegen.rs
@@ -127,7 +127,7 @@ impl<'a> Codegen<'a> {
             "// Generated automatically from the instruction-selection DSL code in:",
         )
         .unwrap();
-        for file in &self.files.file_names {
+        for file in &self.files.file_names_relative() {
             writeln!(code, "// - {file}").unwrap();
         }
 
--- a/cranelift/isle/isle/src/lexer.rs
+++ b/cranelift/isle/isle/src/lexer.rs
@@ -39,7 +39,7 @@ impl Pos {
     pub fn pretty_print_line(&self, files: &Files) -> String {
         format!(
             "{} line {}",
-            files.file_name(self.file).unwrap(),
+            files.file_name_relative(self.file).unwrap(),
             files.file_line_map(self.file).unwrap().line(self.offset)
         )
     }


nmattia:
@cfallin here are some changes related to our discussion [here](https://github.com/bytecodealliance/wasmtime/issues/9553).

> This also removes a log line (which included the full path of files) from cranelift-codegen-meta.

About this ☝️ I'm happy to remove it/submit it separately/make it relative.

@fitzgen you [mentioned](https://github.com/bytecodealliance/wasmtime/issues/9553#issuecomment-2460565124) that, in the past, files were not printed with absolute paths. Is it worth investing some time to ensure this doesn't happen again in the future?

cranelift/isle/isle/src/files.rs
```diff
@@ -109,6 +121,15 @@ impl Files {
...
 }
 
+/// If OUT_DIR is set, strips it from the file name.
+/// Otherwise returns the original file name.
+pub fn relative(file_name: &str) -> &str {
+    match std::env::var("OUT_DIR") {
+        Err(_) => file_name,
+        Ok(root) => file_name.strip_prefix(&root).unwrap_or(file_name).into(),
+    }
+}
```
bjorn3, line 131: I wonder if the build script that invokes cranelift-isle can read this instead. Cranelift-isle can technically be used to build arbitrary code. `Files::from_paths` could get a set of paths to make make input paths relative to, or alternatively a set of paths against which to resolve relative paths and then have the build script only pass the relative paths for individual isle files instead.

nmattia: Sorry, not too familiar with the code/cranelift. Are you suggesting that we should make the paths relative [here](https://github.com/bytecodealliance/wasmtime/blob/06377eb08a649619cc8ac9a934cb3f119017f3ef/cranelift/codegen/build.rs#L225) instead?

bjorn3: Yes. There is already a `make_isle_source_path_relative` function which gets used, but that only affects isle files that are part of the source rather than those that get generated by the build script itself. You did still have to separately pass a (set of) prefix(s) to join with the relative path to get a path that can be passed to `fs::read`, but this prefix can be ignored when printing the source locations in the generated file.

nmattia: >  only affects isle files that are part of the source rather than those that get generated by the build script itself. 

Unrelated to this thread, but that explains why I had to use `OUT_DIR` instead of `CARGO_MANIFEST_DIR` !

> Yes. 

That makes sense. I'll try to find some time later to update the PR.

nmattia: sorry, didn't have much time to look into this!

After playing with this for a bit, I think it might make sense to update the existing function (`make_isle_source_path_relative`) to work with paths that are not subpaths of the current directory. Borrowing some code from [pathdiff](https://github.com/Manishearth/pathdiff/blob/695a66be96e679616c2982a971ae59905739811b/src/lib.rs), here's what this would look like:

```diff
diff --git a/cranelift/codegen/build.rs b/cranelift/codegen/build.rs
index 2cbf2c9e5..cfe6d37d7 100644
--- a/cranelift/codegen/build.rs
+++ b/cranelift/codegen/build.rs
@@ -140,10 +140,11 @@ fn make_isle_source_path_relative(
     cur_dir: &std::path::Path,
     filename: &std::path::Path,
 ) -> std::path::PathBuf {
-    if let Ok(suffix) = filename.strip_prefix(&cur_dir) {
-        suffix.to_path_buf()
-    } else {
-        filename.to_path_buf()
+
+    match diff_paths(filename, cur_dir) {
+        Some(relative) => relative,
+        None => filename.to_path_buf(),
+
     }
 }
 
@@ -253,3 +254,48 @@ fn rustfmt(code: &str) -> std::io::Result<String> {
 
     Ok(String::from_utf8(data).expect("rustfmt always writes utf-8 to stdout"))
 }
+
+pub fn diff_paths<P, B>(path: P, base: B) -> Option<std::path::PathBuf>
+where
+    P: AsRef<std::path::Path>,
+    B: AsRef<std::path::Path>,
+{
+    let path = path.as_ref();
+    let base = base.as_ref();
+
+    if path.is_absolute() != base.is_absolute() {
+        if path.is_absolute() {
+            Some(std::path::PathBuf::from(path))
+        } else {
+            None
+        }
+    } else {
+        let mut ita = path.components();
+        let mut itb = base.components();
+        let mut comps: Vec<std::path::Component> = vec![];
+        loop {
+            match (ita.next(), itb.next()) {
+                (None, None) => break,
+                (Some(a), None) => {
+                    comps.push(a);
+                    comps.extend(ita.by_ref());
+                    break;
+                }
+                (None, _) => comps.push(std::path::Component::ParentDir),
+                (Some(a), Some(b)) if comps.is_empty() && a == b => (),
+                (Some(a), Some(b)) if b == std::path::Component::CurDir => comps.push(a),
+                (Some(_), Some(b)) if b == std::path::Component::ParentDir => return None,
+                (Some(a), Some(_)) => {
+                    comps.push(std::path::Component::ParentDir);
+                    for _ in itb {
+                        comps.push(std::path::Component::ParentDir);
+                    }
+                    comps.push(a);
+                    comps.extend(ita.by_ref());
+                    break;
+                }
+            }
+        }
+        Some(comps.iter().map(|c| c.as_os_str()).collect())
+    }
+}
```

Can confirm that this also fixes the issue (no more non-determinism), plus the fix is a bit more self contained. Thoughts?

nmattia: @bjorn3 what do you think?

nmattia: @cfallin or @bjorn3 do you think it makes sense to extend `make_isle_source_path_relative` as suggested above?

cfallin: Hmm -- I'll admit I'm having a somewhat visceral reaction to this much complexity in a `build.rs` script. Yes, it works, and it is conceptually well-defined, but it's just... a lot of logic that seems like it shouldn't be necessary.

Popping up a few levels in the "why are we doing this" hierarchy, another thought occurs to me: we would also have deterministic output from the ISLE compiler if we omitted the comments with filenames in them; those comments are mainly meant for times when humans are trying to debug the generated code; we already have an environment variable to place the ISLE generated code in a non-build-outputs location for a human to be able to browse it (`ISLE_SOURCE_DIR`); perhaps we could add another environment variable `ISLE_COMMENTS` such that if set, it adds the comments, and we omit the comments otherwise?

We would need to add an option to [CodegenOptions](https://github.com/bytecodealliance/wasmtime/blob/1e58f1b8e24dc2fc013c4d657202af4b4ba51215/cranelift/isle/isle/src/codegen.rs#L16) in the ISLE compiler (call it, say, `include_comments`), then set that option appropriately in this `build.rs`, and follow it in `codegen.rs` -- and let's omit all comments except the "this is machine generated" one at the top if unset. How does that sound?

alexcrichton:
I believe this ended up getting folded into https://github.com/bytecodealliance/wasmtime/pull/10650 (morally, not literally), so closing.

PR #9581, opened 2024-11-07
Remove the use of `MemoryStyle` in Winch

alexcrichton:
This applies a similar change as #9576 but to Winch. Similar to the previous PR one case needed some reshuffling, but otherwise all current behavior is maintained and this should otherwise be a refactoring.


saulecabrera (approved):
Thanks!

PR #9585, opened 2024-11-07
Use blocks instead of instructions in DominatorTree

amartosch:
Since all branching instructions are terminators now, it makes sense to use blocks in `DominatorTree` as a unit of granularity. This makes interfaces a bit cleaner and allows to avoid `idom(..).inst_block(..).expect(..)` chain in a couple of places.

This is an independent part of work on adding new dominator tree computation algorithm.
Discussed in [zulip](https://bytecodealliance.zulipchat.com/#narrow/channel/217117-cranelift/topic/Dominator.20trees).


fitzgen (approved):
Nice clean up! Thanks!

PR #9592, opened 2024-11-11
Don't allow jump instructions after a terminator in is_block_basic

bjorn3:
We merged brz/brnz+jump into br_if a while ago.


cfallin (approved):
Good catch, thanks!

Issue #3573, opened 2021-11-30
ISLE: support bool types and constants properly


cfallin:
ISLE's prototype compiler was built with the simplifying assumption that all constants are integers. "Symbolic constants" (imported opaque values) like `$MYVALUE` were added later. It's possible to define `bool` as a primitive type, but the syntax for bool constants, `#t` and `#f`, results in generated code that uses integer values `1` and `0` because... everything is (was) an integer. This is clearly suboptimal! In #3572 @alexcrichton used the clever workaround of `$false`, leveraging the support for arbitrary passthrough of constant names to actually get a `false` in the code.

We should (i) add a `ConstBool` alongside `ConstInt` in the IR (`PatternInst` and `ExprInst`), and (ii) codegen these properly as Rust bool types. We might also consider at some point making our "primitive" type hierarchy a bit richer and distinguishing ints and bools, but that's lower-priority.

fitzgen:
FWIW, we use `$true` and `$false` elsewhere, eg:

https://github.com/bytecodealliance/wasmtime/blob/0580c84405cb5808518b187d1851dc17e7538108/cranelift/codegen/src/isa/x64/lower.isle#L24-L30

Not convinced that `#t` and `#f` is really that much of an improvement.

cfallin:
Yep, we'd want to clean up said uses all at the same time.

I'm not as concerned about syntax; we could even just accept tokens `true` and `false`; my concern is more that `$true` and `$false` are abusing a feature meant for other things (externally-defined constants) and if we have bool values, we should represent them as such in the IR. Getting the type right there would also allow us to use them in the future in other DSL-compiler optimizations.

PR #9593, opened 2024-11-11
ISLE: Add proper bool type

Kmeakin:
Partially solves https://github.com/bytecodealliance/wasmtime/issues/3573

### DIFF
--- a/cranelift/isle/isle/src/sema.rs
+++ b/cranelift/isle/isle/src/sema.rs
@@ -246,7 +246,7 @@ pub struct Term {
 }
 
 /// Flags from a term's declaration with `(decl ...)`.
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub struct TermFlags {
     /// Whether the term is marked as `pure`.
     pub pure: bool,
@@ -254,6 +254,15 @@ pub struct TermFlags {
     pub multi: bool,
     /// Whether the term is marked as `partial`.
     pub partial: bool,
+}
+
+impl TermFlags {
+    /// Return a new `TermFlags` suitable for a term on the LHS of a rule.
+    pub fn on_lhs(mut self) -> Self {
+        self.pure = true;
+        self.partial = true;
+        self
+    }
 }
 
 /// The kind of a term.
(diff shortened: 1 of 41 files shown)


cranelift/isle/isle/src/sema.rs
```diff
@@ -80,9 +80,36 @@ pub struct TypeEnv {
     pub errors: Vec<Error>,
 }
 
+/// A built-in type.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum BuiltinType {
```
cfallin, line 85: I'm not sure I understand the reason for having a notion of "built-in types" for bool but not for other types that we have constants for (namely integers) -- while on the other hand it adds some complication to the IR. Maybe we could separate out this part into another PR and discuss separately?

cfallin: (To make it more explicit: I think we probably *should* have this eventually; but for integers too, and then we should check that integer literals are used only in contexts that expect them.)

cfallin (commented):
Thanks for working on this -- it's nice to see motion on the "nice-to-have" upgrades we've had in our backlog for a while!

Mostly looks fine; my main thought is about the `BuiltinType` abstraction and whether we can separate the change out.

cranelift/isle/veri/README.md
```diff
@@ -1,15 +1,15 @@
 # Crocus: An SMT-based ISLE verification tool
 
-This directory contains Crocus, a tool for verifying instruction lowering and transformation rules written in ISLE. Crocus uses an underlying SMT solver to model values in ISLE rules in as logical bitvectors, searching over all possible inputs to find potential soundness counterexamples. The motivation and context project are described in detail in our ASPLOS 2024 paper: [Lightweight, Modular Verification for WebAssembly-to-Native Instruction Selection](https://dl.acm.org/doi/10.1145/3617232.3624862).
```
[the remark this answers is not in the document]
Kmeakin: I'm not sure how this slipped in, I don't recall editing it. Anyway, I'll revert it

cranelift/isle/isle/src/sema.rs
Kmeakin: The intention is to extend `BuiltinType` to include other types (like `u{8,16,32,64,size}, `i{8,16,32,64,size}` or `str`) in future PRs

cfallin: OK, yeah, on further thought I suppose it's totally fine to migrate integers over as a second step -- as long as the intention is there to finish out the refactor. Thanks!

cfallin (approved):

PR #9598, opened 2024-11-12
Use enum for directory/file permissions to make generated bindings easier to use

jder:
As part of https://github.com/bytecodealliance/wasmtime-py/issues/251, I want to use the File/Directory permission constants from the C API in Python. As defines, this is a bit tricky. Rather than hard-code them on the Python side, this PR makes them into an enum, and the existing C-to-Python logic is able to see & emit them. I have some work in progress which does this, see the [resulting generated code](https://github.com/bytecodealliance/wasmtime-py/compare/main...jder:wasmtime-py:update-dev?expand=1#diff-1ed8724e6e665b1b666545ea15177bcff0c3dbf8e2d7159ac115babd8c29e32bR1975-R1985) and the [nicer wrappers around them](https://github.com/bytecodealliance/wasmtime-py/compare/main...jder:wasmtime-py:update-dev?expand=1#diff-c4520ab85798819dc0b98db7de4f96f8e8b6ce0f2e7ab9c330a949f69836920c).

### DIFF
--- a/crates/c-api/include/wasi.h
+++ b/crates/c-api/include/wasi.h
@@ -166,7 +166,7 @@ WASI_API_EXTERN void wasi_config_inherit_stderr(wasi_config_t *config);
 /**
  * \brief The permissions granted for a directory when preopening it.
  */
-enum wasi_dir_perms_enum {
+enum wasi_dir_perms_flags {
   /**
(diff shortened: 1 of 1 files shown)


alexcrichton (approved):
Thanks for this!

Issue #9481, opened 2024-10-17
Use a linear time dominators algorithm


fitzgen:
Right now we use the classic simple-fast algorithm (*A Simple, Fast Dominance Algorithm* by Cooper et al) and while this works well most of the time in practice, it has quadratic worst case time.

Linear-time alternatives exist: https://dl.acm.org/doi/10.5555/1123869

We should consider using them instead.

SpiderMonkey found the Semi-NCA algorithm to be a big speed up: https://spidermonkey.dev/blog/2024/10/16/75x-faster-optimizing-the-ion-compiler-backend.html

amartosch:
I'd like to have a try at it. The plan would be to implement the Semi-NCA algorithm, because LLVM also seems to use it.

PR #9603, opened 2024-11-13
Compute dominator tree using semi-NCA algorithm

amartosch:
Compute dominator tree using semi-NCA algorithm, described in `Linear-Time Algorithms for Dominators and Related Problems. Loukas Georgiadis, Princeton University, November 2005.` 
The algorithm has quadratic worst-case complexity, but usually is much faster. The speedup is minor, but can be noticed on large functions, i.e. in [this file](https://bugzilla.mozilla.org/attachment.cgi?id=9426114).
Also add a differential fuzzing job for the new algorithm, with the old one as baseline. This addition is dubious, because it couldn't find anything overnight.
Described in #9481.


cranelift/codegen/src/dominator_tree.rs
```diff
@@ -32,7 +423,7 @@ struct DomNode {
 }
 
 /// The dominator tree for a single function.
-pub struct DominatorTree {
+pub struct SimpleDominatorTree {
```
Amanieu, line 426: This should be moved to a separate file to make the code cleaner. At the moment it looks messy with 2 separate algorithms in the same file.

amartosch: Moved to a separate file

cranelift/codegen/src/dominator_tree.rs
```diff
@@ -242,88 +290,118 @@ impl DominatorTree {
...
+                    if node.pre_number != 0 {
+                        continue;
+                    }
+
+                    self.dfs_worklist.push(TraversalEvent::Exit(block));
+
+                    let pre_number = self.stree.push(parent, block);
+                    node.pre_number = pre_number;
+
+                    // Use the same traversal heuristics as in traversals.rs
```
cfallin, line 317: Rather than refer to another source file here (which may change or be removed in the future), can we say what the heuristics are, briefly?

cranelift/codegen/src/dominator_tree.rs
```diff
@@ -242,88 +290,118 @@ impl DominatorTree {
...
+                    self.dfs_worklist.extend(
+                        func.block_successors(block)
+                            .rev()
+                            .filter(|successor| self.nodes[*successor].pre_number == 0)
```
cfallin, line 321: The `0` sentinel keeps occurring and I think it'd be clearer if we had a name for it -- could you define `const NOT_VISITED: u32 = 0;` or similar in this module?

cranelift/codegen/src/dominator_tree.rs
```diff
@@ -242,88 +290,118 @@ impl DominatorTree {
...
-            //
-            // Since `compute_idom` will only look at nodes with an assigned RPO number, the
-            // function will never see an uninitialized predecessor.
-            //
-            // Due to the nature of the post-order traversal, every node we visit will have at
-            // least one predecessor that has previously been visited during this RPO.
-            self.nodes[block] = DomNode {
-                idom: self.compute_idom(block, cfg).into(),
-                rpo_number: (rpo_idx as u32 + 3) * STRIDE,
+    /// Eval-compress
```
cfallin, line 331: This name ("eval-compress") doesn't mean anything to the reader -- could you give a more descriptive doc-comment?

cranelift/codegen/src/verifier/mod.rs
```diff
@@ -300,14 +300,14 @@ pub fn verify_context<'a, FOI: Into<FlagsOrIsa<'a>>>(
 struct Verifier<'a> {
     func: &'a Function,
     expected_cfg: ControlFlowGraph,
-    expected_domtree: DominatorTree,
+    expected_domtree: SimpleDominatorTree,
```
cfallin, line 303: Is there any reason we have to use the old ("simple") implementation here? Or is the idea that when doing verification, we want to fall back to the more trusted algorithm? (Could we document this reasoning in a comment if so?)

Actually, thinking aloud a bit: I think I'm OK with moving everything over to the new algorithm, including the verifier; it should be enough to have the fuzz-target that checks the algorithm itself against the old algorithm. (If we don't trust that, then we shouldn't use the new algorithm anywhere!)

cfallin (approved):
Thanks very much for this implementation! Some thoughts mostly on comments below but nothing major; given that you've done a fuzzing-based verification of the new algorithm against the old, and seen some compile-time speedups, I'm otherwise confident in it.

cranelift/codegen/src/dominator_tree.rs
```diff
@@ -5,50 +5,130 @@ use crate::flowgraph::{BlockPredecessor, ControlFlowGraph};
...
+
+/// Spanning tree node, used during domtree computation.
+#[derive(Clone, Default)]
+struct SpanningTreeNode {
+    block: PackedOption<Block>,
+    ancestor: u32,
+    label: u32,
+    semi: u32,
+    idom: u32,
+}
```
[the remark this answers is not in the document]
amartosch: Added.

cranelift/codegen/src/dominator_tree.rs
```diff
@@ -5,50 +5,130 @@ use crate::flowgraph::{BlockPredecessor, ControlFlowGraph};
...
+
+impl std::ops::IndexMut<u32> for SpanningTree {
+    fn index_mut(&mut self, idx: u32) -> &mut Self::Output {
+        &mut self.nodes[idx as usize]
+    }
+}
+
+/// Traversal event to compute both preorder spanning tree
+/// and postorder block list. Don't use `Dfs` from traversals.rs
```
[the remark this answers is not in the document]
amartosch: Clarified.

cranelift/codegen/src/dominator_tree.rs
```diff
@@ -5,50 +5,130 @@ use crate::flowgraph::{BlockPredecessor, ControlFlowGraph};
...
-    /// Unreachable nodes get number 0, all others are positive.
-    rpo_number: u32,
-
-    /// The immediate dominator of this block.
-    ///
-    /// This is `None` for unreachable blocks and the entry block which doesn't have an immediate
-    /// dominator.
+struct DominatorTreeNode {
     idom: PackedOption<Block>,
+    /// 0 for unreachable blocks
```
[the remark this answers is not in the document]
amartosch: Fixed.

cranelift/codegen/src/dominator_tree.rs
```diff
@@ -5,50 +5,130 @@ use crate::flowgraph::{BlockPredecessor, ControlFlowGraph};
...
-    /// Scratch traversal state used by `compute_postorder()`.
-    dfs: Dfs,
+    dfs_worklist: Vec<TraversalEvent>,
+    eval_worklist: Vec<u32>,
 
     valid: bool,
 }
 
-/// Methods for querying the dominator tree.
+/// Query methods
```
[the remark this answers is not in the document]
amartosch: Yes, this is a mistake. Fixed.

cranelift/codegen/src/dominator_tree.rs
amartosch: Added.

cranelift/codegen/src/verifier/mod.rs
amartosch: Yes, the idea was to compare this tree to the more trusted algorithm. Changed to the new.

amartosch:
Gentle ping. @cfallin,  @Amanieu , is there anything else to be done here?

cfallin:
> Gentle ping. @cfallin, @Amanieu , is there anything else to be done here?

Sorry about that, I didn't notice updates had been made (I watch @-pings and review requests but not always every commit) -- will take a look shortly!

cfallin (approved):
Thanks for this and sorry again about the delay!

PR #9604, opened 2024-11-14
Store one fiber stack in a `Store<T>`

alexcrichton:
This commit stores a single fiber stack in `Store<T>` as a cache to be used throughout the lifetime of the `Store`. This should help amortize the cost of allocating a stack for use in a store because the same stack can be used continuously throughout the lifetime of the `Store<T>`. This notably reduces contention on the lock used to manage the pooling allocator when possible.


pchickey (approved):
Love it!

PR #9606, opened 2024-11-14
Decrease default guard size from 2G to 32M

alexcrichton:
This commit follows in the footsteps of SpiderMonkey to reduce the size of the default guard region from 2GiB to 32MiB. SpiderMonkey performance an analysis of some wasm modules and found the largest static offset was 20MiB so 32 is the rounded up version of that.

This will reduce the size of the virtual memory reservation per linear-memory by default. Previously it was 8G due to guards being both before and after linear memory being 2G in size. Now it'll be 4G+64M with before/after guards taken into account. This should in theory make it easier to pack more instances in the pooling allocator for example and overall reduce the virtual memory footprint.

This is not expected to have any major impact on the performance of wasm modules as all bounds checks should still practically be elided. We've been fuzzing differently sized guard regions for quite a long time as well so there should be a low risk of this having any issues specifically connected to a smaller guard region.

### DIFF
--- a/crates/environ/src/tunables.rs
+++ b/crates/environ/src/tunables.rs
@@ -199,10 +199,13 @@ impl Tunables {
             // address space reservations liberally by default, allowing us to
             // help eliminate bounds checks.
             //
-            // Coupled with a 2 GiB address space guard it lets us translate
-            // wasm offsets into x86 offsets as aggressively as we can.
+            // A 32MiB default guard size is then allocated so we can remove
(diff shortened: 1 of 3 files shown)


alexcrichton:
For reference this is from discussions with Ryan this week at WasmCon and the SpiderMonkey reference is [here](https://searchfox.org/mozilla-central/rev/30afd6631333bf83d5c62ca48c313d96a21b4bbd/js/src/wasm/WasmMemory.h#189-193)

Issue #5431, opened 2022-12-13
cranelift-isle: richer integer types


jameysharp:
Lift a bit more of the integer type system into ISLE. Namely

```lisp
(type u64 (primitive u64 integer 64 unsigned))
(type i64 (primitive i64 integer 64 signed))
```

such that in a `u64` context we accept literals in 0..2^64 and in an `i64` context we accept -2^63..2^63. Then we also add some syntax

```lisp
(imm $I64 (u64 -1:i64))  ;; sketching some syntax, please bikeshed something better
```

where `u64` used in term position is a cast (compiles to `as u64`, but also works with typechecking at ISLE level) and `:i64` on a constant sets the type of the constant.

This is all a bit uncertain but I feel there's a better system for constants, and for integer types, buried in here somewhere...

_Originally posted by @cfallin in https://github.com/bytecodealliance/wasmtime/issues/5423#issuecomment-1347701158_

PR #9610, opened 2024-11-15
ISLE: built-in integer types

Kmeakin:
Follow up to https://github.com/bytecodealliance/wasmtime/pull/9593, adding all the built-in integer types (`u8` .. `u128`, `i8` .. `i128` plus `usize` and `isize`).

Integer literal expressions and patterns are still allowed to type-check against primitive types, because there are a lot of places in the lowering/optimization code that uses `(type Foo (primitive Foo))` where `Foo` is declared as a type alias of an integer type in Rust code. Not sure what to do about this. Perhaps it would be clearer to rename "primitive" types to "opaque" or "extern" types (and also have a mechanism for declaring transparent type aliases in ISLE)?

Closes https://github.com/bytecodealliance/wasmtime/issues/5431 and https://github.com/bytecodealliance/wasmtime/issues/3573

### DIFF
--- a/cranelift/isle/isle/src/sema.rs
+++ b/cranelift/isle/isle/src/sema.rs
@@ -22,6 +22,7 @@ use std::collections::hash_map::Entry;
 use std::collections::BTreeMap;
 use std::collections::BTreeSet;
 use std::collections::HashMap;
+use std::fmt;
 
 declare_id!(
     /// The id of an interned symbol.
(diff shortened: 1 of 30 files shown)


cranelift/isle/isle/src/sema.rs
```diff
@@ -82,26 +83,144 @@ pub struct TypeEnv {
...
+            IntType::I32 => "i32",
+            IntType::I64 => "i64",
+            IntType::I128 => "i128",
+            IntType::ISize => "isize",
+        }
+    }
+
+    /// Is this integer type signed?
+    pub fn is_signed(&self) -> bool {
+        matches!(
```
cfallin, line 144: Can we write a `match` with explicit arms for all types here? I'd prefer that so we are forced to think about any new types we add explicitly. The pattern as written makes that easy to miss...

cranelift/isle/isle/src/sema.rs
```diff
@@ -1166,7 +1290,7 @@ impl TypeEnv {
         self.errors.push(err);
     }
 
-    fn intern_mut(&mut self, ident: &ast::Ident) -> Sym {
+    fn get_or_intern(&mut self, ident: &ast::Ident) -> Sym {
```
cfallin, line 1293: I'm not sure I like this rename -- semantically I think of interning as the action of getting the existing ID, or creating a new one -- that is, "to intern" covers both cases; and `intern_mut` indicates we're permitting new cases to update the table, vs. `intern` that only looks up existing cases.

(To see this another way, consider `intern` below -- it should be renamed to something else if we take this rename, because this rename implies that adding to the table is the "intern" bit, which is exactly the bit that the non-mut version below does not do.)

cfallin (commented):
Thanks! Some comments below but nothing major. This is a nice refactor and makes the language semantically a little cleaner; I like it.

cfallin (approved):

cfallin:
Something is broken in the build in CI; once it's green I'm happy to merge.

PR #9611, opened 2024-11-15
Cranelift: add option to use new single-pass register allocator.

### LINKED ISSUE #9596: Add flag to select the fast regalloc impl
#### Feature

Add a flag to enable the regalloc impl introduced by https://github.com/bytecodealliance/regalloc2/pull/181.

#### Benefit

Improves compilation time.

#### Implementation

Trivial

cfallin:
In bytecodealliance/regalloc2#181, @d-sonuga added a fast single-pass algorithm option to regalloc2, in addition to its existing backtracking allocator. This produces code much more quickly, at the expense of code quality. Sometimes this tradeoff is desirable (e.g. when performing a debug build in a fast-iteration development situation, or in an initial JIT tier).

This PR adds a Cranelift option to select the RA2 algorithm, plumbs it through to a Wasmtime option, and adds the option to Wasmtime fuzzing as well.

An initial compile-time measurement in Wasmtime: `spidermonkey.wasm` builds in 1.383s with backtracking (existing algorithm), and 1.065s with single-pass. The resulting binary runs a simple Fibonacci benchmark in 2.060s with backtracking vs. 3.455s with single-pass.

Hence, the single-pass algorithm yields a 23% compile-time reduction, at the cost of a 67% runtime increase.

Fixes #9596.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -56,6 +56,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "allocator-api2"
+version = "0.2.20"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "45862d1c77f2228b9e10bc609d5bc203d86ebc9b87ad8d5d5167a6c9abf739d9"
+
+[[package]]
 name = "ambient-authority"
 version = "0.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


alexcrichton (approved):
🎉 nice!

I'm ambivalent myself on the defaults for O0 and could go either way.

cfallin:
For now at least, I think I'd prefer to keep it an opt-in default -- let's let it bake in wasmtime's continuous fuzzing for a little longer. We can always switch the default later.

@alexcrichton updated to add cargo-vet, could you rubber-stamp the new commit? Also fixed silly issues in fuzz build (which I never test beforehand because Ocaml; I should fix my setup!).

alexcrichton (approved):
Looks good!

For fuzzing you can also build the fuzzers with `--no-default-features` to turn off the ocaml integration.

PR #9613, opened 2024-11-15
loop analysis simplification

KGrewal1:
Some small simplifications I saw when looking through the loop analysis code, mostly regarding using inbuilt iterator method for greater simplicity

### DIFF
--- a/cranelift/codegen/src/loop_analysis.rs
+++ b/cranelift/codegen/src/loop_analysis.rs
@@ -10,7 +10,7 @@ use crate::ir::{Block, Function, Layout};
 use crate::packed_option::PackedOption;
 use crate::timing;
 use alloc::vec::Vec;
-use smallvec::{smallvec, SmallVec};
+use smallvec::SmallVec;
 
 /// A opaque reference to a code loop.
 #[derive(Copy, Clone, PartialEq, Eq, Hash)]
@@ -294,7 +294,7 @@ impl LoopAnalysis {
     }
 
     fn assign_loop_levels(&mut self) {
-        let mut stack: SmallVec<[Loop; 8]> = smallvec![];
+        let mut stack: SmallVec<[Loop; 8]> = SmallVec::new();
         for lp in self.loops.keys() {
             if self.loops[lp].level == LoopLevel::invalid() {
                 stack.push(lp);
--- a/cranelift/codegen/src/loop_analysis.rs
+++ b/cranelift/codegen/src/loop_analysis.rs
@@ -191,14 +191,14 @@ impl LoopAnalysis {
     }
 
     // Determines if a block dominates any predecessor
-    // and thus is a loop header
-    fn check_loop_header(
+    // and thus is a loop header.
+    fn is_block_loop_header(
         block: Block,
         cfg: &ControlFlowGraph,
         domtree: &DominatorTree,
         layout: &Layout,
     ) -> bool {
-        // loop header if it dominates any of its predecessors
+        // A block is a loop header if it dominates any of its predecessors.
         cfg.pred_iter(block)
             .any(|pred| domtree.dominates(block, pred.inst, layout))
     }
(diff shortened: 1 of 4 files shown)


cranelift/codegen/src/loop_analysis.rs
```diff
@@ -198,21 +198,22 @@ impl LoopAnalysis {
...
+            .into_iter()
+            .rev() // We traverse the CFG in reverse postorder
+            .filter(|&&block| {
+                // If the block dominates any one of its predecessors it is a back edge
+                cfg.pred_iter(block).any(
+                    |BlockPredecessor {
+                         inst: pred_inst, ..
+                     }| domtree.dominates(block, pred_inst, layout),
+                )
+            })
```
cfallin, line 212: This iterator chain is a little too convoluted to be easily readable, IMHO. One thing that might help is to pull out an `is_loop_header` helper?

It might also be nice to have a method on the domtree that returns an RPO iterator directly (`impl Iterator...` in return is fine). Then we could do something like:

`for &block in domtree.cfg_rpo().filter(|&block| is_loop_header(cfg, domtree, block)) { ... }`

cranelift/codegen/src/loop_analysis.rs
```diff
@@ -229,16 +230,18 @@ impl LoopAnalysis {
...
+                    .filter(
+                        |BlockPredecessor {
+                             inst: pred_inst, ..
+                         }| {
+                            // We follow the back edges
+                            domtree.dominates(self.loops[lp].header, *pred_inst, layout)
+                        },
+                    )
+                    .map(|BlockPredecessor { block, .. }| block),
+            );
```
cfallin, line 244: I'm not sure how I feel about this one either -- is this really more readable than the for-loop with an if-statement?

cfallin (commented):
Thanks for the PR. A few thoughts below. I think there may be some aspect of subjective taste here; but I'm not a huge fan of very large iterator chain expressions, because they take a lot of "mental type-inference" to understand intermediate steps and how the chain works, and they're harder to modify and maintain over time. They're not a huge amount *worse* than explicit for-loops, but here they don't really seem better either; so I'm not sure I see the "simplification" aspect. Happy to hear your thoughts though.

cranelift/codegen/src/loop_analysis.rs
KGrewal1: I think some of the complexity is in how I wrote the de structuring - changing it to `pred.inst`, as opposed to doing it between the pipes. Also would a collect be most appropriate here - used extend to ensure exact same semantics as before with push, but the `while let should` drain  `stack` and so it is being extended from empty each time (in which case collect may make this more clear)

cranelift/codegen/src/loop_analysis.rs
```diff
@@ -190,6 +190,19 @@ impl LoopAnalysis {
         self.valid = false;
     }
 
+    // Determines if a block dominates any predecessor
+    // and thus is a loop header
+    fn check_loop_header(
```
cfallin, line 195: Let's call this `is_loop_header` (`check` doesn't tell me anything about what it returns; usually implies an invariant check or something like that instead). And the doc-comment should end with a period.

cranelift/codegen/src/loop_analysis.rs
```diff
@@ -190,6 +190,19 @@ impl LoopAnalysis {
...
+        block: Block,
+        cfg: &ControlFlowGraph,
+        domtree: &DominatorTree,
+        layout: &Layout,
+    ) -> bool {
+        // loop header if it dominates any of its predecessors
```
cfallin, line 201: Full sentence is preferable: "A block is a loop header if it dominates any of its predecessors."

cranelift/codegen/src/loop_analysis.rs
```diff
@@ -232,15 +236,11 @@ impl LoopAnalysis {
...
-                            // We follow the back edges
-                            domtree.dominates(self.loops[lp].header, *pred_inst, layout)
-                        },
-                    )
-                    .map(|BlockPredecessor { block, .. }| block),
+                    .filter(|pred| {
+                        // We follow the back edges
+                        domtree.dominates(self.loops[lp].header, pred.inst, layout)
+                    })
+                    .map(|pred| pred.block),
```
cfallin, line 243: I think this would be a little clearer if we had a comment above the `stack.extend`, something like "Push all predecessors of this header that this header dominates onto the stack."

cranelift/codegen/src/loop_analysis.rs
KGrewal1: Was going to use `is_loop_header` but already in use by another function - check was best synonym but not ideal

cfallin (approved):

PR #9618, opened 2024-11-18
Expose component parameter names in types

### LINKED ISSUE #9595: Expose Parameter Names method for ComponentFuncs
#### Feature

Provide an interface for deriving a ComponentFunc params original names

#### Benefit

Presently, `wasmtime` provides a number tools that can be leveraged for something that resembles Reflection, but the details for Function Parameters are lossy. Especially for more complex runtime parameter mapping, it would be great if the names of function parameters were preserved, and an interface provided for retrieving them.

#### Implementation

In order to maintain backwards compatibility, it could likely be something like a `param_names()` method.

```rust
impl ComponentFunc

  pub fn param_names(&self) -> impl ExactSizeIterator<Item = String> + '_`

      Iterates over names of function parameters
```

#### Alternatives

There really doesn't seem to be one that's built-in to `wasmtime`. In order to grab the parameter names, I'd likely need to do something far more complex.

Based on `wasm-tools`, though, it seems like the name of the params are preserved in the `.wasm`, which is intriguing.

To my knowledge, ComponentFunc params are the entities that do not have some way of deriving their original names.

alexcrichton:
This commit plumbs through the parameter names since they're available in the component for consumers of the type information exposed by Wasmtime.

Closes #9595

### DIFF
--- a/crates/fuzzing/src/oracles.rs
+++ b/crates/fuzzing/src/oracles.rs
@@ -988,7 +988,7 @@ pub fn dynamic_component_api_target(input: &mut arbitrary::Unstructured) -> arbi
     while input.arbitrary()? {
         let params = param_tys
             .iter()
-            .map(|ty| component_types::arbitrary_val(ty, input))
+            .map(|(_, ty)| component_types::arbitrary_val(ty, input))
             .collect::<arbitrary::Result<Vec<_>>>()?;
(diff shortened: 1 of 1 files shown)


pchickey (approved):
Looks great, thanks!

Issue #9544, opened 2024-11-02
Improving wasmtime performance on illumos via mmapping /dev/null


sunshowers:
(I'm still a novice to the wasmtime code so please let me know if I've made any mistakes. Thanks!)

## Motivation

In #9535 we added initial support for illumos. (Thanks!) As noted there, we did notice some performance issues, particularly around freeing memory at shutdown.

Some of my colleagues and I investigated the situation today, and we have a pretty good sense of what's going on (some of my colleagues will be writing up the related illumos issues at some point, but it's essentially that `MAP_NORESERVE` allocations of size N take up O(N) CPU, rather than apparently O(1) as on some other platforms. The constant factor is very small, but wasmtime allocates enough memory that it is noticeable.)

However, we've found that there's an alternative mmapping strategy that works well, bringing illumos perf to within the same ballpark as Linux (at least as far as the wast test suite goes). The alternative strategy is:

* For `PROT_NONE` mappings, rather than creating a large amount of _anonymous_ memory, create a map to `/dev/null`.
* When a particular part of memory needs to be made accessible in any way, we need to create an anonymous mmap. Accessible memory is a very small portion of the overall address space map, so this is fine.
* The regions that have "real" anonymous mmaps associated with them need to be tracked via a data structure, ideally an interval tree. Essentially, some of the bookkeeping the kernel currently does makes its way into userspace.

I have a quick and dirty prototype that shows how `/dev/null` mapping might work, though it needs a lot of work to be made shippable. The prototype is at https://github.com/sunshowers/wasmtime/pull/1 -- `mmap.rs` has the meaty interval tree logic if you're interested.

With this PR,

```
% cargo +beta test --test wast -- --test-threads 1 Cranelift/pooling/tests/spec_testsuite/load.wast
<snip>
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 2491 filtered out; finished in 0.08s
```

(Note that this strategy doesn't work on Linux -- `mmap`ing `/dev/null` produces `ENODEV`. Naturally, if we go this route, this will be a platform-specific impl of mmapping on illumos, and possibly other systems where `/dev/null`-based mapping might be faster.)

## Challenges

The most important implication of this approach is that all memory management *must* go through the mmap. This means that nothing outside the mmap code should call `mmap`, `mmap_anonymous` or `mprotect` directly.

[...]

PR #9620, opened 2024-11-19
Add HostAlignedByteCount to enforce alignment at compile time

sunshowers:
As part of the work to allow mmaps to be backed by other implementations (#9544), I realized that we didn't have any way to track whether a particular usize is host-page-aligned at compile time.

Add a `HostAlignedByteCount` which tracks that a particular usize is aligned to the host page size. This also does not expose safe unchecked arithmetic operations, to ensure that overflows are always handled.

With `HostAlignedByteCount`, a lot of runtime checks can go away thanks to the type-level assertion.

In the interest of keeping the diff relatively small, I haven't converted everything over yet. More code can be converted over as we go along.

### DIFF
--- a/crates/wasmtime/src/runtime/vm.rs
+++ b/crates/wasmtime/src/runtime/vm.rs
@@ -86,6 +86,8 @@ mod module_id;
 pub use module_id::CompiledModuleId;
 
 #[cfg(feature = "signals-based-traps")]
+mod byte_count;
+#[cfg(feature = "signals-based-traps")]
 mod cow;
 #[cfg(not(feature = "signals-based-traps"))]
 mod cow_disabled;
@@ -94,6 +96,7 @@ mod mmap;
 
 cfg_if::cfg_if! {
     if #[cfg(feature = "signals-based-traps")] {
+        pub use crate::runtime::vm::byte_count::*;
         pub use crate::runtime::vm::mmap::Mmap;
         pub use self::cow::{MemoryImage, MemoryImageSlot, ModuleMemoryImages};
     } else {
@@ -365,6 +368,8 @@ pub fn host_page_size() -> usize {
 }
 
 /// Is `bytes` a multiple of the host page size?
+///
+/// (Deprecated: consider switching to `HostAlignedByteCount`.)
 #[cfg(feature = "signals-based-traps")]
 pub fn usize_is_multiple_of_host_page_size(bytes: usize) -> bool {
     bytes % host_page_size() == 0
@@ -373,6 +378,8 @@ pub fn usize_is_multiple_of_host_page_size(bytes: usize) -> bool {
 /// Round the given byte size up to a multiple of the host OS page size.
 ///
 /// Returns an error if rounding up overflows.
+///
+/// (Deprecated: consider switching to `HostAlignedByteCount`.)
 #[cfg(feature = "signals-based-traps")]
 pub fn round_u64_up_to_host_pages(bytes: u64) -> Result<u64> {
     let page_size = u64::try_from(crate::runtime::vm::host_page_size()).err2anyhow()?;
@@ -386,6 +393,8 @@ pub fn round_u64_up_to_host_pages(bytes: u64) -> Result<u64> {
 }
 
 /// Same as `round_u64_up_to_host_pages` but for `usize`s.
+///
+/// (Deprecated: consider switching to `HostAlignedByteCount`.)
 #[cfg(feature = "signals-based-traps")]
 pub fn round_usize_up_to_host_pages(bytes: usize) -> Result<usize> {
     let bytes = u64::try_from(bytes).err2anyhow()?;
--- a/crates/wasmtime/src/runtime/vm/instance/allocator/pooling/unix_stack_pool.rs
+++ b/crates/wasmtime/src/runtime/vm/instance/allocator/pooling/unix_stack_pool.rs
@@ -1,13 +1,11 @@
 #![cfg_attr(asan, allow(dead_code))]
 
-use super::{
-    index_allocator::{SimpleIndexAllocator, SlotId},
-    round_up_to_pow2,
-};
+use super::index_allocator::{SimpleIndexAllocator, SlotId};
 use crate::prelude::*;
 use crate::runtime::vm::sys::vm::commit_pages;
 use crate::runtime::vm::{
-    mmap::AlignedLength, round_usize_up_to_host_pages, Mmap, PoolingInstanceAllocatorConfig,
+    mmap::AlignedLength, round_usize_up_to_host_pages, HostAlignedByteCount, Mmap,
+    PoolingInstanceAllocatorConfig,
 };
 
 /// Represents a pool of execution stacks (used for the async fiber implementation).
@@ -23,9 +21,9 @@ use crate::runtime::vm::{
 #[derive(Debug)]
 pub struct StackPool {
     mapping: Mmap<AlignedLength>,
-    stack_size: usize,
+    stack_size: HostAlignedByteCount,
     max_stacks: usize,
-    page_size: usize,
+    page_size: HostAlignedByteCount,
     index_allocator: SimpleIndexAllocator,
     async_stack_zeroing: bool,
     async_stack_keep_resident: usize,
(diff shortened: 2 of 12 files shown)


crates/wasmtime/src/runtime/vm/instance/allocator/pooling/unix_stack_pool.rs
```diff
@@ -152,11 +158,11 @@ impl StackPool {
...
 
         // Remove the guard page from the size
-        let stack_size = self.stack_size - self.page_size;
+        let stack_size = self.stack_size.byte_count() - self.page_size.byte_count();
         let bottom_of_stack = top - stack_size;
-        let start_of_stack = bottom_of_stack - self.page_size;
+        let start_of_stack = bottom_of_stack - self.page_size.byte_count();
         assert!(start_of_stack >= base && start_of_stack < (base + len));
-        assert!((start_of_stack - base) % self.stack_size == 0);
+        assert!((start_of_stack - base) % self.stack_size.byte_count() == 0);
```
sunshowers, line 165: This code and the similar section below would be good to convert over, but I decided to do this in a followup.

crates/wasmtime/src/runtime/vm.rs
```diff
@@ -336,6 +336,216 @@ impl ModuleRuntimeInfo {
     }
 }
 
+/// A number of bytes that's guaranteed to be aligned to the host page size.
+///
+/// This is used to manage page-aligned memory allocations.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct HostAlignedByteCount(
```
sunshowers, line 343: I first named this `AlignedByteCount`, but then realized that the VM page size might not be the same as the host page size, so named it `HostAlignedByteCount` instead. Happy to rename this to something else if it makes sense.

crates/wasmtime/src/runtime/vm.rs
```diff
@@ -336,6 +336,216 @@ impl ModuleRuntimeInfo {
...
+impl PartialEq<HostAlignedByteCount> for usize {
+    #[inline]
+    fn eq(&self, other: &HostAlignedByteCount) -> bool {
+        *self == other.0
+    }
+}
+
+impl fmt::Display for HostAlignedByteCount {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
```
sunshowers, line 483: Should this print the size in hex or as a decimal?

sunshowers:
I started working on a followup as well: https://github.com/bytecodealliance/wasmtime/commit/03efe42fb4d8fdd760d945ce6ceb95b8ca338974

Unfortunately GitHub's review flow does not let you manage stacked PRs unless you have push access to the destination repo, so I'll have to put these up one at a time. Sigh :(

crates/wasmtime/src/runtime/vm.rs
alexcrichton: Same seems reasonable to me 👍

crates/wasmtime/src/runtime/vm.rs
```diff
@@ -336,6 +336,216 @@ impl ModuleRuntimeInfo {
...
+/// A number of bytes that's guaranteed to be aligned to the host page size.
+///
+/// This is used to manage page-aligned memory allocations.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct HostAlignedByteCount(
+    // Invariant: this is always a multiple of the host page size.
+    usize,
+);
+
+impl HostAlignedByteCount {
```
alexcrichton, line 348: Mind putting this type/impls/errors/etc in a dedicated submodule of the `vm` module?

crates/wasmtime/src/runtime/vm.rs
```diff
@@ -336,6 +336,216 @@ impl ModuleRuntimeInfo {
...
+#[cfg(feature = "std")]
+impl std::error::Error for ByteCountOutOfBounds {}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum ByteCountOutOfBoundsKind {
+    RoundUp(usize),
+    U64(u64),
+    Add(usize, usize),
+    Sub(usize, usize),
+    Mul(usize, usize),
```
alexcrichton, line 524: For here I think it's probably ok to discard the precise values that were being operated on as otherwise that makes the errors here pretty large for a low-level type such as this.

crates/wasmtime/src/runtime/vm.rs
alexcrichton: Let's go with hex perhaps to clearly show that the low bits are all zero

alexcrichton (approved):
Thanks for this!

crates/wasmtime/src/runtime/vm/sys/unix/mmap.rs
```diff
@@ -134,8 +138,10 @@ impl Mmap {
     }
 
     #[inline]
-    pub fn len(&self) -> usize {
-        self.memory.as_ptr().len()
+    pub fn len(&self) -> HostAlignedByteCount {
+        // SAFETY: When self.memory was created it was always aligned to a page
+        // boundary.
+        unsafe { HostAlignedByteCount::new_unchecked(self.memory.as_ptr().len()) }
```
alexcrichton, line 144: Reviewing some other code elsewhere, I think that the `from_file` constructor may cause problems here because file lengths aren't always page-aligned. 

To help catch this mind adding a `debug_assert!` to `new_unchecked`? That I think should make the test suite panic

alexcrichton:
As a heads up I'm adding https://github.com/bytecodealliance/wasmtime/pull/9614 to the merge queue which is going to have a number of conflicts with this. It should (hopefully) be just a few `#[cfg(feature = "signals-based-traps")]` in a few places but if anything is overly difficult or onerous lemme know and I can help out

crates/wasmtime/src/runtime/vm.rs
sunshowers: Done.

crates/wasmtime/src/runtime/vm/sys/unix/mmap.rs
sunshowers: This ended up becoming #9639.

crates/wasmtime/src/runtime/vm.rs
sunshowers: Hmm good point, done.

sunshowers:
(Apologies for the force push -- had to do it as part of the rebase. GH makes me cry)

alexcrichton (approved):
Thanks again for this!

sunshowers:
All right, think I fixed the Miri issue.

crates/wasmtime/src/runtime/vm/memory/mmap.rs
```diff
@@ -178,23 +190,20 @@ impl RuntimeLinearMemory for MmapMemory {
...
-            assert!(
-                new_accessible <= self.mmap.len() - self.offset_guard_size - self.pre_guard_size,
-            );
 
             // If the Wasm memory's page size is smaller than the host's page
             // size, then we might not need to actually change permissions,
             // since we are forced to round our accessible range up to the
             // host's page size.
-            if new_accessible > self.accessible() {
+            if let Ok(difference) = new_accessible.checked_sub(self.accessible()) {
```
alexcrichton, line 200: I think this may be the source of the bug, albeit it's a bit indirect. Previously the `>` condition is now morally becoming `>=` since `new_accessible == self.accessible()` didn't run this block before but it's now running this block with `difference == 0`. I think then it may be the case that mprotect for 0 bytes is succeeding on unix but failing on Windows.

On Linux I see `mprotect(0x740d36001000, 0, PROT_READ|PROT_WRITE) = 0` in strace with this patch when running the failing spec test and while I haven't executed this on Windows this is what I would suspect is going on.

Another option is to update the `Mmap` type to skip `make_accessible` when the byte size is zero which I think would also be reasonable.

sunshowers: Oh that's terrifying -- I'm going to fix this (probably by making `make_accessible` ignore zero-sized regions) and go through this code again to make sure I didn't miss any other sign changes. Thanks for diagnosing this!

sunshowers: Made `make_accessible` a no-op on zero-length allocations, as well as the other mprotect calls (`make_executable` and `make_readonly`). They also had this as a lurking bug because they had assertions that `range.start <= range.end`, not `range.start < range.end`.

I've also added tests for zero-length calls, and re-reviewed everything -- didn't see any other changes.

PR #9626, opened 2024-11-20
Skip the rlimited-memory test under QEMU

iii-i:
After the rework of the QEMU memory management, which landed in v8.0.0 and introduced interval trees instead of individual page descriptors - similar to how the Linux kernel does it - it became possible to run the WASMTIME_TEST_NO_HOG_MEMORY tests under it.

The only exception is the rlimited-memory test, which relies on the underlying OS enforcing the RLIMIT_DATA limit. QEMU chose to ignore this limit, because it interferes with its JIT.

Skip this test when using emulation.


alexcrichton (approved):
Thanks!

PR #9627, opened 2024-11-20
Refactor the `#[wasmtime_test]` macro

alexcrichton:
* Start tests with a blank slate of features instead of with the default set of features enables (ensures each test explicitly specifies required features)

* Reuse test features from `wasmtime_wast_util` to avoid duplicating listings of features. Also shares logic for "should this compiler fail this test because of unsupported features".

* Move logic in `tests/wast.rs` to apply test configuration to a `Config` to a new location that can be shared across suites.

* Add a new feature for `simd` and flag tests that need it with the feature.

This is done in preparation for adding a new compiler strategy of Pulley to be able to flag tests as passing for pulley or not.


crates/test-macros/src/lib.rs
```diff
@@ -259,21 +231,37 @@ fn expand(test_config: &TestConfig, func: Fn) -> Result<TokenStream> {
...
         );
 
-        let config_setup = test_config.wasm_features.iter().map(|f| {
-            let method_name = Ident::new(&format!("wasm_{f}"), f.span());
-            quote! {
-                config.#method_name(true);
-            }
-        });
+        let ignore = if strategy.should_fail(&test_config.flags) {
+            quote!(#[ignore])
```
abrown, line 235: Maybe this should be `#[should_panic]`? I get that this keeps the original behavior but more just wondering: if we check for panics and then it stops panicking then we know more quickly which Winch tests can be enabled.

abrown (approved):
This improvement looks good!

crates/test-macros/src/lib.rs
alexcrichton: Good point! Works well 👍

Issue #9588, opened 2024-11-09
Give a bit more informative error message related to deleting `isle_generated_code/`


larry0x:
PR #4143 added [this error](https://github.com/bytecodealliance/wasmtime/blob/main/cranelift/codegen/build.rs#L76-L88) if a `isle_generated_code/` directory exists and an "isle-in-source-tree" feature isn't enabled:

```rust
#[cfg(not(feature = "isle-in-source-tree"))]
{
    if explicit_isle_dir.is_dir() {
        eprintln!(concat!(
            "Error: directory isle_generated_code/ exists but is only used when\n",
            "`--feature isle-in-source-tree` is specified. To prevent confusion,\n",
            "this build script requires the directory to be removed when reverting\n",
            "to the usual generated code in target/. Please delete the directory and\n",
            "re-run this build.\n",
        ));
        std::process::exit(1);
    }
}
```

It says to delete the directory. However, for users who are not familiar with the inner workings of cranelift, it's not clear where this directory is. It took me a long while to find it:

```
~/.cargo/registry/src/index.crates.io-{hash}/cranelift-codegen-{version}/isle_generated_code
```

It may be helpful to add the `explicit_isle_dir` value to the error message so that users are informed exactly which directory they are supposed to delete.

cfallin:
Hmm, something is going wrong here -- you shouldn't ever need to delete a directory in `~/.cargo`. The intent of this feature was to allow Cranelift developers to see a version of the code in-tree when developing Cranelift, and should not be used when Cranelift is pulled in as a dependency from crates.io. Could you say more about your environment and how you're using Cranelift?

larry0x:
@cfallin I don't use cranelift directly so not sure what's supposed to happen... I use wasmer. At wasmer 5.0.0 my program compiles with no problem. Once bumped to wasmer 5.0.1, compiling my program results in this error:

```
error: failed to run custom build command for `cranelift-codegen v0.110.2`

Caused by:
  process didn't exit successfully: `/Users/[...]/target/release/build/cranelift-codegen-725296997bf7694e/build-script-build` (exit status: 1)
  --- stdout
  cargo:rerun-if-changed=build.rs

  --- stderr
  Error: directory isle_generated_code/ exists but is only used when
  `--feature isle-in-source-tree` is specified. To prevent confusion,
  this build script requires the directory to be removed when reverting
  to the usual generated code in target/. Please delete the directory and
  re-run this build.
```

[...]

PR #9633, opened 2024-11-20
Cranelift: rework isle-in-source-tree functionality.

### LINKED ISSUE #9625: More context on a cranelift-codegen build script error
Currently if an explicit isle directory is found when the "isle-in-source-tree" feature is not enabled, the following error message is shown by the build script for cranelift-codegen:

```
 --- stderr
  Error: directory isle_generated_code/ exists but is only used when
  `--feature isle-in-source-tree` is specified. To prevent confusion,
  this build script requires the directory to be removed when reverting
  to the usual generated code in target/. Please delete the directory and
  re-run this build.
```
as emitted by [https://github.com/bytecodealliance/wasmtime/blob/main/cranelift/codegen/build.rs#L78C9-L87C10](https://github.com/bytecodealliance/wasmtime/blob/main/cranelift/codegen/build.rs#L78C9-L87C10).

However this is misleading as deleting the `target/` directory (atleast on windows) does not fix the issue. The directory, for me, was located in `%HOME%\\.cargo\\registry\\src\\index.crates.io-6f17d22bba15001f\\cranelift-codegen-0.113.1\\isle_generated_code`. Could there possibly be a better check/error message and/or the actual directory located printed out to stderr?

cfallin:
Per #9625 and #9588, a downstream user of Cranelift was misusing the `isle-in-source-tree` feature, enabling it indiscriminately even in published crates (wasmerio/wasmer#5202). This went against the intent of the feature, to be a way for local developers to observe the generated source. As such, it ran afoul of some safety checks for that purpose, and also polluted users' Cargo caches in a way that we cannot now fix except in future releases.

The best we can do is probably to take away features that are liable to be misused. Following discussion in today's Cranelift weekly meeting, we decided to provide this functionality under an environment variable instead. This allows folks who know what they're doing to get the same behavior, but does not expose a feature to crate users.

Fixes #9625.

### DIFF
--- a/cranelift/codegen/build.rs
+++ b/cranelift/codegen/build.rs
@@ -63,37 +63,24 @@ fn main() {
 
     println!("cargo:rerun-if-changed=build.rs");
 
-    let explicit_isle_dir = &crate_dir.join("isle_generated_code");
-    #[cfg(feature = "isle-in-source-tree")]
-    let isle_dir = explicit_isle_dir;
-    #[cfg(not(feature = "isle-in-source-tree"))]
-    let isle_dir = &out_dir;
+    let isle_dir = if let Ok(path) = std::env::var("ISLE_SOURCE_DIR") {
+        // This will canonicalize any relative path in terms of the
+        // crate root, and will take any absolute path as overriding the
+        // `crate_dir`.
+        crate_dir.join(&path)
+    } else {
+        out_dir.into()
+    };
 
-    #[cfg(feature = "isle-in-source-tree")]
-    {
-        std::fs::create_dir_all(isle_dir).expect("Could not create ISLE source directory");
-    }
-    #[cfg(not(feature = "isle-in-source-tree"))]
-    {
-        if explicit_isle_dir.is_dir() {
-            eprintln!(concat!(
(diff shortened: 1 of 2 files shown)


alexcrichton (approved):

cfallin:
I think the merge queue run raced with the release; cargo-vet failed because it found a version on crates.io it didn't expect. Re-running, but noting in case we see it again later -- interesting corner case!

cranelift/codegen/build.rs
```diff
@@ -63,37 +63,24 @@ fn main() {
...
-                "Error: directory isle_generated_code/ exists but is only used when\n",
-                "`--feature isle-in-source-tree` is specified. To prevent confusion,\n",
-                "this build script requires the directory to be removed when reverting\n",
-                "to the usual generated code in target/. Please delete the directory and\n",
-                "re-run this build.\n",
-            ));
-            std::process::exit(1);
-        }
-    }
+    let isle_dir = if let Ok(path) = std::env::var("ISLE_SOURCE_DIR") {
```
bjorn3, line 66: Shouldn't this use `cargo::rerun-if-env-changed`?

PR #9636, opened 2024-11-20
 Speed up a "big stack" test, and test Winch 

alexcrichton:
This test currently takes 50s locally in debug mode locally so optimize
it by disabling optimizations in Cranelift and additionally using the
single-pass register allocator. This drops the test time to ~3s locally.

### DIFF
--- a/tests/all/stack_overflow.rs
+++ b/tests/all/stack_overflow.rs
@@ -2,6 +2,7 @@
 
 use std::sync::atomic::{AtomicUsize, Ordering::SeqCst};
 use wasmtime::*;
+use wasmtime_test_macros::wasmtime_test;
 
 #[test]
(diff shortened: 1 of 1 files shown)


pchickey (approved):
Just to help me understand: does this mean we lose test coverage for cranelift? if we're not losing coverage, why does this test exist?

alexcrichton:
A good question! I did a bit more digging and I've now changed this to:

* This is tested on both Winch and Cranelift now.
* Cranelift disables egraph optimizations
* Cranelift uses the single-pass register allocation pass

That makes the test much speedier now so it will continue to test Cranelift.

pchickey:
Thanks. Comment needs a fix, then its good to go

alexcrichton:
oops, thanks!

PR #9639, opened 2024-11-20
Add a type-level distinction between aligned and possibly-unaligned Mmaps

sunshowers:
In the wasmtime codebase there are two kinds of mmaps: those that are always backed by anonymous memory and are always aligned, and those that are possibly file-backed and so the length might be unaligned. In https://github.com/bytecodealliance/wasmtime/pull/9620 I accidentally mixed the two up in a way that was a ticking time bomb.

To resolve this, add a type parameter to `Mmap` to distinguish between the cases. All of wasmtime's users are clearly either one or the other, so it's quite simple in the end.

### DIFF
--- a/crates/wasmtime/src/runtime/vm/sys/unix/mmap.rs
+++ b/crates/wasmtime/src/runtime/vm/sys/unix/mmap.rs
@@ -20,7 +20,7 @@ pub struct Mmap {
 cfg_if::cfg_if! {
     if #[cfg(any(target_os = "illumos", target_os = "linux"))] {
         // On illumos, by default, mmap reserves what it calls "swap space" ahead of time, so that
-        // memory accesses are guaranteed not to fail once mmap succeeds. NORESERVE is for cases
+        // memory accesses a`re guaranteed not to fail once mmap succeeds. NORESERVE is for cases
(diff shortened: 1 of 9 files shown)


alexcrichton (approved):

crates/wasmtime/src/runtime/vm/sys/unix/mmap.rs
```diff
@@ -20,7 +20,7 @@ pub struct Mmap {
 cfg_if::cfg_if! {
     if #[cfg(any(target_os = "illumos", target_os = "linux"))] {
         // On illumos, by default, mmap reserves what it calls "swap space" ahead of time, so that
-        // memory accesses are guaranteed not to fail once mmap succeeds. NORESERVE is for cases
+        // memory accesses a`re guaranteed not to fail once mmap succeeds. NORESERVE is for cases
```
sunshowers, line 23: whoops, will fix in the followup!

sunshowers: Included a fix in https://github.com/bytecodealliance/wasmtime/pull/9620.

Issue #7037, opened 2023-09-14
Memory checker (wmemcheck) completeness


cfallin:
This issue is meant to track what remains to make `wmemcheck` feature-complete and robust enough for practical use:

- [ ] Handle SIMD loads and stores (#6829)
- [ ] Hook `calloc` as well as existing `malloc` and `free`
- [ ] Provide hooks for custom allocators to use
- [ ] Allow for more flexibility in heap and stack layout (don't assume stack is above `.data`/`.bss` and everything below initial TOS is accessible)
- [ ] Provide an alternate way to hook `malloc`/`free`, rather than requiring a name section in the `.wasm` module
- [ ] Provide an option to print warnings as violations occur, rather than halting execution (like Valgrind)
- [ ] Provide a mechanism to define "suppressions" for unwanted warnings (like Valgrind)
- [ ] Track definedness at byte or bit level?

cc @iximeow for more thoughts, and others from today's Wasmtime meeting?

iximeow:
@alexcrichton noted earlier that it would be good to have:
- [ ] an end-to-end test starting with some clearly-UB .c and test expectations for a stack trace pointing to wasmtime trapping a bad access

and since wmemcheck doesn't support multiple memories (though we ought to make it so!),
- [ ] warn if a user tries to wmemcheck a multi-memory program
or add that support multiple memories. but i think that's maybe a step after "robust enough for practical use" in terms of effort, warning seems more expedient for now :)

PR #9641, opened 2024-11-20
wmemcheck support for additional allocation functions and granular memory

cobbal:
I was recently tracking down a memory corruption bug in swift and found the wmemcheck tool to be very helpful. Since the runtime of swift is fairly large, it required some extra features to remove false positives. I was able to modify wmemcheck to be useful enough for the job.

I'd like to contribute the changes, many of which overlap with parts of issue https://github.com/bytecodealliance/wasmtime/issues/7037 . In particular, I needed to add the following:

- support for the other memory functions exported by wasi-libc (calloc, realloc, posix_memalign, aligned_alloc, malloc_usable_size) from https://github.com/WebAssembly/wasi-libc/blob/c47daaf6c69f82835fc92d9e14361803dfa794fa/dlmalloc/src/dlmalloc.c#L67
- tracking memory in 4-byte chunks instead of 1-byte. e.g. wasi-libc's strlen loads strings 4 bytes at a time, and possibly beyond the end of the string for efficiency.
- disabling the check for reading uninitialized memory.

The last 2 possibly should be hidden behind options, but I'm not familiar enough with the code base or how others use this tool to know what the right approach would be.

### DIFF
--- a/crates/cranelift/src/func_environ.rs
+++ b/crates/cranelift/src/func_environ.rs
@@ -909,10 +909,10 @@ impl<'module_environment> FuncEnvironment<'module_environment> {
     }
 
     #[cfg(feature = "wmemcheck")]
-    fn hook_before_allocator(&mut self, builder: &mut FunctionBuilder) {
-        let memcheck_off = self.builtin_functions.before_allocator(builder.func);
+    fn check_allocator_start(&mut self, builder: &mut FunctionBuilder) {
+        let allocator_start = self.builtin_functions.allocator_start(builder.func);
         let vmctx = self.vmctx_val(&mut builder.cursor());
-        builder.ins().call(memcheck_off, &[vmctx]);
+        builder.ins().call(allocator_start, &[vmctx]);
     }
 
     #[cfg(feature = "wmemcheck")]
@@ -3071,7 +3071,7 @@ impl<'module_environment> crate::translate::FuncEnvironment
                 Some("posix_memalign") |
                 Some("aligned_alloc") |
                 Some("malloc_usable_size") =>
-                    self.hook_before_allocator(builder),
+                    self.check_allocator_start(builder),
                 _ => ()
             }
         }
--- a/crates/cranelift/src/func_environ.rs
+++ b/crates/cranelift/src/func_environ.rs
@@ -586,6 +586,112 @@ impl<'module_environment> FuncEnvironment<'module_environment> {
     }
 
     #[cfg(feature = "wmemcheck")]
+    fn hook_calloc_exit(&mut self, builder: &mut FunctionBuilder, retvals: &[ir::Value]) {
+        let check_calloc = self.builtin_functions.check_calloc(builder.func);
+        let vmctx = self.vmctx_val(&mut builder.cursor());
+        let func_args = builder
+            .func
+            .dfg
+            .block_params(builder.func.layout.entry_block().unwrap());
+        let (count, size) = if func_args.len() < 4 {
+            return;
+        } else {
+            (func_args[2], func_args[3])
+        };
+        let retval = if retvals.len() < 1 {
+            return;
+        } else {
+            retvals[0]
+        };
+        builder.ins().call(check_calloc, &[vmctx, retval, count, size]);
+    }
+
+    #[cfg(feature = "wmemcheck")]
+    fn hook_realloc_exit(&mut self, builder: &mut FunctionBuilder, retvals: &[ir::Value]) {
+        let check_realloc = self.builtin_functions.check_realloc(builder.func);
+        let vmctx = self.vmctx_val(&mut builder.cursor());
+        let func_args = builder
+            .func
+            .dfg
+            .block_params(builder.func.layout.entry_block().unwrap());
+        let (ptr, len) = if func_args.len() < 4 {
+            return;
+        } else {
+            // If a function named `realloc` has at least two arguments, we assume the
+            // first arguments are the pointer and requested allocation size.
+            (func_args[2], func_args[3])
+        };
+        let retval = if retvals.len() < 1 {
+            return;
+        } else {
+            retvals[0]
+        };
+        builder.ins().call(check_realloc, &[vmctx, retval, ptr, len]);
+    }
+
+    #[cfg(feature = "wmemcheck")]
+    fn hook_malloc_usable_size_exit(&mut self, builder: &mut FunctionBuilder, retvals: &[ir::Value]) {
+        let check_malloc_usable_size = self.builtin_functions.check_malloc_usable_size(builder.func);
+        let vmctx = self.vmctx_val(&mut builder.cursor());
+        let func_args = builder
+            .func
+            .dfg
+            .block_params(builder.func.layout.entry_block().unwrap());
+        let ptr = if func_args.len() < 3 {
+            return;
+        } else {
+            func_args[2]
+        };
+        let retval = if retvals.len() < 1 {
+            return;
+        } else {
+            retvals[0]
+        };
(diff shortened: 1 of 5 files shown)


crates/cranelift/src/func_environ.rs
```diff
@@ -3088,11 +3187,22 @@ impl<'module_environment> crate::translate::FuncEnvironment
 
         #[cfg(feature = "wmemcheck")]
         if self.wmemcheck {
-            let func_name = self.current_func_name(builder);
-            if func_name == Some("malloc") {
-                self.check_malloc_start(builder);
-            } else if func_name == Some("free") {
-                self.check_free_start(builder);
+            match self.current_func_name(builder) {
+                Some("__wrap_malloc") | Some("malloc") =>
```
cfallin: Can we add a comment here noting where the `__wrap_*` variants come from? That's useful both for its own sake (readers' understanding) but also so that if things change in the future with the external thing causing this need, we can re-evaluate.

crates/environ/src/builtin.rs
```diff
@@ -53,19 +65,17 @@ macro_rules! foreach_builtin_function {
...
-            #[cfg(feature = "wmemcheck")]
-            free_start(vmctx: vmctx);
             // Invoked when wasm stack pointer is updated.
             #[cfg(feature = "wmemcheck")]
             update_stack_pointer(vmctx: vmctx, value: i32);
             // Invoked before memory.grow is called.
             #[cfg(feature = "wmemcheck")]
             update_mem_size(vmctx: vmctx, num_bytes: i32);
 
+            // Invoked before stuff is called.
```
cfallin: can we be more specific than "stuff" here?

crates/wasmtime/src/runtime/vm/libcalls.rs
```diff
@@ -1137,6 +1137,40 @@ unsafe fn check_malloc(
     Ok(0)
 }
 
+// Hook for validating calloc using wmemcheck_state.
+#[cfg(feature = "wmemcheck")]
+unsafe fn check_calloc(store: &mut dyn VMStore, instance: &mut Instance, addr: u32, count: u32, size: u32) -> Result<u32> {
+    check_malloc(store, instance, addr, count * size)
```
cfallin: Can we add a note here that since we aren't tracking undefined-ness, we don't have to care about the other semantic difference between `calloc` and `malloc`, namely that memory is zeroed?

crates/wasmtime/src/runtime/vm/libcalls.rs
```diff
@@ -1137,6 +1137,40 @@ unsafe fn check_malloc(
...
+unsafe fn check_realloc(store: &mut dyn VMStore, instance: &mut Instance, end_addr: u32, start_addr: u32, len: u32) -> Result<u32> {
+    check_free(store, instance, start_addr)?;
+    check_malloc(store, instance, end_addr, len)
+}
+
+// Hook for validating malloc_usable_size using wmemcheck_state.
+#[cfg(feature = "wmemcheck")]
+unsafe fn check_malloc_usable_size(store: &mut dyn VMStore, instance: &mut Instance, len: u32, addr: u32) -> Result<u32> {
+    check_free(store, instance, addr)?;
+    check_malloc(store, instance, addr, len)
```
cfallin: Is this a correct implementation of `malloc_usable_size`'s semantics? The man page on my system says that it returns the size of a memory block; unclear to me whether we need a hook for that, even, to track alloc'd/free state?

crates/wasmtime/src/runtime/vm/libcalls.rs
```diff
@@ -1137,6 +1137,40 @@ unsafe fn check_malloc(
...
+}
+
+
+// Hook for validating posix_memalign using wmemcheck_state.
+#[cfg(feature = "wmemcheck")]
+unsafe fn check_posix_memalign(store: &mut dyn VMStore, instance: &mut Instance, outptr: u32, size: u32) -> Result<u32> {
+    for (_, entry) in instance.exports() {
+        if let wasmtime_environ::EntityIndex::Memory(mem_index) = entry {
+            let mem = instance.get_memory(*mem_index);
+            let out = *(mem.base.offset(outptr as isize) as *mut u32);
```
cfallin: We definitely need to do a bounds-check here -- could you use the safe APIs for the memory to access the out-pointer?

This is also little-endian-specific code; we'll want to get a `[u8; 4]` and do an explicit conversion with `u32::from_le_bytes`. (Wasmtime supports one big-endian architecture, s390x!)

crates/wasmtime/src/runtime/vm/libcalls.rs
```diff
@@ -1158,6 +1192,30 @@ unsafe fn check_free(_store: &mut dyn VMStore, instance: &mut Instance, addr: u3
...
+            for (_, entry) in instance.exports() {
+                if let wasmtime_environ::EntityIndex::Memory(mem_index) = entry {
+                    let mem = instance.get_memory(*mem_index);
+                    for i in 0..prev_len {
+                        if i > 0 && i % 40 == 0 {
+                            println!();
+                        }
+                        unsafe {
+                            print!("{:02x} ", *mem.base.offset((prev_malloc + i) as isize));
+                        }
```
cfallin: This looks like code that could be useful for debugging, but as above with the `posix_memalign` out-pointer fetch, the memory accesses are unsafe, and also dumping the contents of previously allocated blocks might be too verbose. I think I'd prefer to omit this code for now; ultimately the right answer for any problem that needs examination of memory state like this is probably using a debugger (we have better support planned!) possibly with breakpoints on allocation events or trap-to-debugger on wmemcheck errors.

crates/wmemcheck/src/lib.rs
```diff
@@ -140,6 +158,9 @@ impl Wmemcheck {
 
     /// Updates memory checker memory state metadata when free is called.
     pub fn free(&mut self, addr: usize) -> Result<(), AccessError> {
+        if addr == 0 {
```
cfallin, line 192: Good find! Can we add a comment here noting the implemented semantics (`free` of a null pointer is a no-op)?

crates/wmemcheck/src/lib.rs
```diff
@@ -101,12 +119,12 @@ impl Wmemcheck {
                         len: len,
                     });
                 }
-                MemState::ValidToWrite => {
+                /* MemState::ValidToWrite => {
```
cfallin: Why is this commented out? (If the checks were causing issues, could we put the exclusion under an appropriately named and documented option?)

crates/wmemcheck/src/lib.rs
```diff
@@ -51,7 +66,10 @@ impl Wmemcheck {
     }
 
     /// Updates memory checker memory state metadata when malloc is called.
-    pub fn malloc(&mut self, addr: usize, len: usize) -> Result<(), AccessError> {
+    pub fn malloc(&mut self, addr: usize, start_len: usize) -> Result<(), AccessError> {
+        // round up to multiple of 4
+        let len = (start_len + 3) & !3;
```
cfallin: Can we put this under an option? Perhaps "wmemcheck granularity" or similar?

cfallin (commented):
Thanks very much for this! Some initial thoughts below.

I'm really happy `wmemcheck` was actually useful for you, and I think most of this can be upstreamed with no problem. Just a few thoughts below.

efJerryYang:
Hi @cobbal , I came across this PR while exploring the related tracking issue #7037 . Could you please share the current status of this PR? Would you mind if I took over the remaining steps from cfallin's review to help get it merged?

PR #9652, opened 2024-11-22
move most of runtime/vm/cow.rs over to aligned byte counts

sunshowers:
As part of attempting to move some of these operations over to `Mmap` instances, it is nice to have type-level checking for aligned sizes. In upcoming PRs, APIs like `map_at` will be switched to using `Mmap` instances with aligned counts.

There are a couple of spots where I have questions -- will flag them in review comments.

### DIFF
--- a/crates/wasmtime/src/runtime/vm/cow.rs
+++ b/crates/wasmtime/src/runtime/vm/cow.rs
@@ -545,28 +545,7 @@ impl MemoryImageSlot {
     ) {
         match &self.image {
             Some(image) => {
-                let image_end = image
-                    .linear_memory_offset
-                    .checked_add(image.len)
-                    .expect("image is in bounds");
-                let mem_after_image = self
-                    .accessible
-                    .checked_sub(image_end)
-                    .expect("image_end falls before self.accessible");
-
-                let keep_resident_minus_image_start =
-                    if let Ok(excess) = keep_resident.checked_sub(image.linear_memory_offset) {
-                        // The .is_zero() check is carried over from a previous
-                        // iteration of the logic, which had > 0 and <= 0 cases.
-                        //
-                        // XXX: are the two paths identical? If so, the is_zero
-                        // check can go away.
-                        (!excess.is_zero()).then_some(excess)
-                    } else {
-                        None
-                    };
-
-                if let Some(excess) = keep_resident_minus_image_start {
+                if image.linear_memory_offset < keep_resident {
                     // If the image starts below the `keep_resident` then
                     // memory looks something like this:
                     //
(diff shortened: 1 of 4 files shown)


crates/wasmtime/src/runtime/vm/cow.rs
```diff
@@ -528,13 +540,33 @@ impl MemoryImageSlot {
...
+                    .checked_sub(image_end)
+                    .expect("image_end falls before self.accessible");
+
+                let keep_resident_minus_image_start =
+                    if let Ok(excess) = keep_resident.checked_sub(image.linear_memory_offset) {
+                        // The .is_zero() check is carried over from a previous
+                        // iteration of the logic, which had > 0 and <= 0 cases.
+                        //
+                        // XXX: are the two paths identical? If so, the is_zero
+                        // check can go away.
```
sunshowers, line 563: Thoughts about this? I tried to be super cautious about signs this time.

crates/wasmtime/src/runtime/vm/cow.rs
```diff
@@ -610,26 +651,30 @@ impl MemoryImageSlot {
...
-                    ptr::write_bytes(self.base.as_ptr(), 0u8, keep_resident);
+                    ptr::write_bytes(self.base.as_ptr(), 0u8, keep_resident.byte_count());
 
                     // This is madvise (2)
                     self.restore_original_mapping(
                         keep_resident,
-                        self.accessible - keep_resident,
+                        self.accessible
+                            .checked_sub(keep_resident)
+                            .expect("XXX what should happen if this underflows?"),
```
sunshowers, line 661: Wasn't clear to me whether this was possible -- there are several offsets here that all interact with each other.

crates/wasmtime/src/runtime/vm/cow.rs
alexcrichton: Would it be possible to keep the same `<` condition as before? I find that one a bit easier to understand in terms of the diagrams below personally.

alexcrichton (commented):
Thanks for this!

crates/wasmtime/src/runtime/vm/cow.rs
sunshowers: Okay -- to be honest I'm not entirely sure that this is the case in reality, it seems like in configuration you could define keep_resident to be more then the byte size. But in any case this should panic (and if there is such a condition it should be handled at a higher level)

crates/wasmtime/src/runtime/vm/cow.rs
```diff
@@ -528,13 +540,33 @@ impl MemoryImageSlot {
...
-                assert!(self.accessible >= image.linear_memory_offset + image.len);
-                if image.linear_memory_offset < keep_resident {
+                let image_end = image
+                    .linear_memory_offset
+                    .checked_add(image.len)
+                    .expect("image is in bounds");
+                let mem_after_image = self
+                    .accessible
+                    .checked_sub(image_end)
+                    .expect("image_end falls before self.accessible");
```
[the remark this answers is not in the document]
sunshowers: Yeah good idea, much more understandable with this and the below change.

sunshowers:
ugh, attempting to fix the no-std build

crates/wasmtime/src/runtime/vm/cow.rs
sunshowers: Done.

alexcrichton (approved):

PR #9658, opened 2024-11-22
pulley: Implement `get_frame_pointer`

alexcrichton:
This commit implements the CLIF `get_frame_pointer` instruction which will be needed by trampolines in Wasmtime. This is implemented by generalizing the preexisting `GetSp` `MInst` into a "get special". This the additionally removes the extended `get_sp` opcode in Pulley as it's not necessary as `xmov` can operate on this register.


alexcrichton:
ping @cfallin, mind taking a look at this?

cranelift/codegen/src/isa/pulley_shared/inst.isle
```diff
@@ -39,8 +39,11 @@
     ;; Nothing.
     (Nop)
 
-    ;; Get the stack pointer.
-    (GetSp (dst WritableXReg))
+    ;; Move a special register (e.g. sp, fp, lr, etc) in to a general-purpose
+    ;; register
```
cfallin, line 43: nit but end comment with `.`?

cranelift/codegen/src/isa/pulley_shared/inst/mod.rs
```diff
@@ -79,8 +79,12 @@ fn pulley_get_operands(inst: &mut Inst, collector: &mut impl OperandVisitor) {
             collector.reg_use(src2);
         }
 
-        Inst::GetSp { dst } => {
+        Inst::GetSpecial { dst, reg } => {
             collector.reg_def(dst);
+            // Note that this is explicitly ignored as this is only used for
+            // special register that don't participate in register allocation
```
cfallin, line 85: s/register/registers/

cranelift/codegen/src/isa/pulley_shared/inst/mod.rs
```diff
@@ -79,8 +79,12 @@ fn pulley_get_operands(inst: &mut Inst, collector: &mut impl OperandVisitor) {
...
+            // such as the stack pointer, frame pointer, etc.
+            let _ = reg;
```
cfallin, line 87: Can we add an assert here that `reg` is one of the special registers?

cfallin (approved):
Sorry, missed this review request! LGTM with a few minor bits below.

fitzgen (commented):
FWIW, I kind of wanted to have the special registers not accessible by regular `xmov`/`xadd`/etc... so we could have 32 regular x registers (and the compact 5 bit representaton for all regular x registers) and then for our handful of special registers we would just have special instructions for manipulating.

This would trade using additional (presumably extended) opcodes for gaining more GPRs, which should help us avoid that many more regalloc spills. Spills result in more instructions, which means more turns of the interpreter loop, which is (expected to be) the most expensive part of wasm execution under pulley.

Not something we need to do now but it's something to think about as we design the instruction set, and maybe we should avoid similar changes as those included in this PR which move us away from that vision in the future where we can.

alexcrichton:
Yeah I've been thinking similarly that I'd prefer to take these registers out of the gpr set, I've opened https://github.com/bytecodealliance/wasmtime/issues/9747 and added this there.

PR #9659, opened 2024-11-22
pulley: Implement a new `br_table` instruction

alexcrichton:
This is intended to match WebAssembly's `br_table` and Cranelift's version as well. This is implemented as a new `br_table32` opcode where a 32-bit number of branch targets are encoded after `br_table32` all as a `PcRelOffset`, a 32-bit offset. This helps bake in a more "macro opcode" into the interpreter rather than a handful of more primitive opcodes that would achieve the same result with loads/indirect jumps/comparisons/etc.

### DIFF
--- a/cranelift/codegen/src/isa/pulley_shared/inst.isle
+++ b/cranelift/codegen/src/isa/pulley_shared/inst.isle
@@ -22,7 +22,7 @@
     (Unwind (inst UnwindInst))
 
     ;; Implementation of `br_table`, uses `idx` to jump to one of `targets` or
-    ;; jumps to `default` is it's out-of-bounds.
+    ;; jumps to `default` if it's out-of-bounds.
     (BrTable
       (idx XReg)
       (default MachLabel)
--- a/cranelift/codegen/src/isa/pulley_shared/inst.isle
+++ b/cranelift/codegen/src/isa/pulley_shared/inst.isle
@@ -21,6 +21,13 @@
     ;; A pseudo-instruction to update unwind info.
     (Unwind (inst UnwindInst))
 
+    ;; Implementation of `br_table`, uses `idx` to jump to one of `targets` or
+    ;; jumps to `default` is it's out-of-bounds.
+    (BrTable
+      (idx XReg)
+      (default MachLabel)
+      (targets BoxVecMachLabel))
+
     ;;;; Actual Instructions ;;;;
 
     ;; Raise a trap.
@@ -546,6 +553,10 @@
       (let ((dst WritableXReg (temp_writable_xreg))
             (_ Unit (emit (MInst.BitcastIntFromFloat64 dst src))))
         dst))
+
+(decl gen_br_table (XReg MachLabel BoxVecMachLabel) Unit)
+(rule (gen_br_table idx default labels)
+      (emit (MInst.BrTable idx default labels)))
 
 ;;;; Helpers for Emitting Calls ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
--- a/cranelift/codegen/src/isa/pulley_shared/inst/emit.rs
+++ b/cranelift/codegen/src/isa/pulley_shared/inst/emit.rs
@@ -104,8 +104,8 @@ where
         // (with an `EmitIsland`). We check this in debug builds. This is `mut`
         // to allow disabling the check for `JTSequence`, which is always
         // emitted following an `EmitIsland`.
-        let start = sink.cur_offset();
-        pulley_emit(self, sink, emit_info, state, start);
+        let mut start = sink.cur_offset();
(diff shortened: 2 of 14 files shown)


cranelift/codegen/src/isa/pulley_shared/inst.isle
```diff
@@ -21,6 +21,13 @@
     ;; A pseudo-instruction to update unwind info.
     (Unwind (inst UnwindInst))
 
+    ;; Implementation of `br_table`, uses `idx` to jump to one of `targets` or
+    ;; jumps to `default` is it's out-of-bounds.
```
abrown, line 25:
```suggestion
    ;; jumps to `default` if it's out-of-bounds.
```

cranelift/codegen/src/isa/pulley_shared/inst/emit.rs
```diff
@@ -484,6 +484,47 @@ fn pulley_emit<P>(
...
+            // jump around the island before the `br_table32` itself gets
+            // emitted.
+            let amt = u32::try_from(targets.len() + 1).expect("too many branch targets");
+            let br_table_size = amt * 4 + 6;
+            if sink.island_needed(br_table_size) {
+                let label = sink.get_label();
+                <InstAndKind<P>>::from(Inst::Jump { label }).emit(sink, emit_info, state);
+                sink.emit_island(br_table_size, &mut state.ctrl_plane);
+                sink.bind_label(label, &mut state.ctrl_plane);
+            }
```
abrown, line 511: This is confusing: if Pulley is an interpreter, we're not going to need an island for the same reasons we did during JIT compilation. I haven't looked below, but typically an interpreter sets the `pc`, the dispatch loop loads the bytecode at `pc`, etc. We don't have to jump an arbitrary `u32` amount. Or is there something that I don't understand here about the tail-calling version of Pulley?

tests/disas.rs
```diff
@@ -459,14 +454,15 @@ where
             disassembly: disas,
         } in disas(bytes, sym.address())?.into_iter()
         {
-            if write_offsets || (prev_jump && !is_jump) {
-                write!(result, "{address:>4x}: ")?;
-            } else {
-                write!(result, "      ")?;
+            for (i, line) in disas.lines().enumerate() {
+                if i == 0 && (write_offsets || (prev_jump && !is_jump)) {
```
abrown, line 458: It seems like we're trying to only have the targets of jumps be prefixed by address labels but then the test output ends up printing labels for everything after `0x33`?

cranelift/codegen/src/isa/pulley_shared/inst/emit.rs
```diff
@@ -104,8 +104,8 @@ where
         // (with an `EmitIsland`). We check this in debug builds. This is `mut`
         // to allow disabling the check for `JTSequence`, which is always
         // emitted following an `EmitIsland`.
-        let start = sink.cur_offset();
-        pulley_emit(self, sink, emit_info, state, start);
+        let mut start = sink.cur_offset();
```
abrown, line 107: Not having worked on Pulley from the beginning, I'm stepping out on a limb here: it's worrying that once we introduce mutable state during codegen (and later during disassembly) we will have difficulty going back from this. This feels like a better discussion to have with @fitzgen since he may have stronger opinions about this (or not).

cranelift/codegen/src/isa/pulley_shared/inst/emit.rs
alexcrichton: Heh your confusion is not unwarranted! Pulley is roughly modeled after our existing backends which is why this is still a problem. For example a jump will be "jump this distance forward or backward" in the bytecode stream (e.g. relative offset to pc). The jumps may have limited distance like other backends (right now they're all 32-bit jumps though) which is why islands might be necessary. Basically Pulley still has the same issues as other backends in terms of handling jumps b/c it's all relative to the current `pc` like native code and we'll probably add smaller-than-32-bit jumps at some point.

cranelift/codegen/src/isa/pulley_shared/inst/emit.rs
alexcrichton: FWIW this is modeled after the [aarch64 backend](https://github.com/bytecodealliance/wasmtime/blob/2af831a6defed4378a0c6a324d9ecd5fbab58eff/cranelift/codegen/src/isa/aarch64/inst/emit.rs#L720)

abrown (approved):
I don't want to stand in the way of progress: Pulley is getting better!

cranelift/codegen/src/isa/pulley_shared/inst/emit.rs
alexcrichton: For now yeah that's definitely the constraint, in the future though I think we might add 8-bit or smaller jump targets  in which case we'd definitely want the island bits. So in some sense this is future-proofing if not maintaining consistency across the backends.

PR #9663, opened 2024-11-22
pulley: Initial support for indirect calls

alexcrichton:
Not thoroughly tested at runtime yet but this should be enough to get the Cranelift backend to not panic and start testing more and more wasm instructions and modules.


abrown (approved):
Makes sense to me.

PR #9665, opened 2024-11-22
pulley: Implement interpreter-to-host calls

alexcrichton:
This commit is an initial stab at implementing interpreter-to-host communication in Pulley. The basic problem is that Pulley needs the ability to call back into Wasmtime to implement tasks such as `memory.grow`, imported functions, etc. For native platforms this is a simple `call_indirect` operation in Cranelift but the story for Pulley must be different because it's effectively switching from interpreted code to native code.

The initial idea for this in #9651 is replaced here and looks mostly similar but with a few changes. The overall structure of how this works is:

* A new `call_indirect_host` opcode is added to Pulley.
  * Function signatures that can be called from Pulley bytecode are statically enumerated at build-time.
  * This enables the implementation of `call_indirect_host` to take an immediate of which signature is being used and cast the function pointer to the right type.
* A new pulley-specific relocation is added to Cranelift for this opcode.
  * `RelocDistance::Far` calls to a name trigger the use of `call_indirect_host`.
  * The relocation is filled in by Wasmtime after compilation where the signature number is inserted.
  * A new `NS_*` value for user-function namespaces is reserved in `wasmtime-cranelift` for this new namespace of functions.
* Code generation for Pulley in `wasmtime-cranelift` now has Pulley-specific handling of the wasm-to-host transition where all previous `call_indirect` instructions are replaced with a call to a "backend intrinsic" which gets lowered to a `call_indirect_host`.

Note that most of this still isn't hooked up everywhere in Wasmtime. That means that the testing here is pretty light at this time. It'll require a fair bit more work to get everything fully integrated from Wasmtime in Pulley. This is expected to be one of the significant remaining chunks of work and should help unblock future testing (or make those diffs smaller ideally).

### DIFF
--- a/crates/cranelift/src/compiler.rs
+++ b/crates/cranelift/src/compiler.rs
@@ -146,7 +146,7 @@ impl Compiler {
         {
             use cranelift_codegen::ir::types::{I32, I64, I8};
 
-            // If pulley is enabled, even if we're not targetting it, determine
+            // If pulley is enabled, even if we're not targeting it, determine
             // what pulley signature that the input `signature` maps to. This is
             // done to ensure that even on native platforms we've always got a
             // signature listed in pulley for all platform intrinsics. In theory
--- a/crates/cranelift/src/lib.rs
+++ b/crates/cranelift/src/lib.rs
@@ -212,7 +212,11 @@ pub const NS_WASM_FUNC: u32 = 0;
 /// function through an indirect function call loaded by the `VMContext`.
 pub const NS_WASMTIME_BUILTIN: u32 = 1;
 
-/// TODO
+/// Namespace used to when a call from Pulley to the host is being made. This is
+/// used with a `colocated: false` name to trigger codegen for a special opcode
+/// for pulley-to-host communication. The index of the functions used in this
+/// namespace correspond to the function signature of `for_each_host_signature!`
+/// in the pulley_interpreter crate.
 pub const NS_PULLEY_HOSTCALL: u32 = 2;
 
 /// A record of a relocation to perform.
--- a/crates/cranelift/src/obj.rs
+++ b/crates/cranelift/src/obj.rs
@@ -212,10 +212,16 @@ impl<'a> ModuleTextBuilder<'a> {
                 // and the goal is to insert the static signature number, `n`,
                 // into the instruction.
                 //
-                // At this time the instructions is 4-bytes large (3 bytes for
-                // the opcode, one for the one-byte payload), so we target the
-                // third byte. The value `n` here should always fit within a
-                // byte.
+                // At this time the instruction looks like:
+                //
+                //      +------+------+------+------+
+                //      | OP   | OP_EXTENDED |  N   |
+                //      +------+------+------+------+
+                //
+                // This 4-byte encoding has `OP` indicating this is an "extended
+                // opcode" where `OP_EXTENDED` is a 16-bit extended opcode.
+                // The `N` byte is the index of the signature being called and
+                // is what's b eing filled in.
                 //
                 // See the `test_call_indirect_host_width` in
                 // `pulley/tests/all.rs` for this guarantee as well.


crates/cranelift/src/compiler.rs
```diff
@@ -119,6 +119,105 @@ impl Compiler {
...
+        // When calling the host we should always be using the platform's
+        // default calling convention since it'll be calling Rust code in
+        // Wasmtime itself.
+        assert_eq!(signature.call_conv, self.isa.default_call_conv());
+
+        #[cfg(feature = "pulley")]
+        {
+            use cranelift_codegen::ir::types::{I32, I64, I8};
+
+            // If pulley is enabled, even if we're not targetting it, determine
```
abrown, line 149:
```suggestion
            // If pulley is enabled, even if we're not targeting it, determine
```

crates/cranelift/src/lib.rs
```diff
@@ -212,6 +212,9 @@ pub const NS_WASM_FUNC: u32 = 0;
 /// function through an indirect function call loaded by the `VMContext`.
 pub const NS_WASMTIME_BUILTIN: u32 = 1;
 
+/// TODO
```
abrown, line 215: Missing docs...

crates/cranelift/src/obj.rs
```diff
@@ -198,12 +199,30 @@ impl<'a> ModuleTextBuilder<'a> {
...
+
+                // This relocation is used to fill in which hostcall signature
+                // is desired within the `call_indirect_host` opcode of Pulley
+                // itself. The relocation target is the start of the instruction
+                // and the goal is to insert the static signature number, `n`,
+                // into the instruction.
+                //
+                // At this time the instructions is 4-bytes large (3 bytes for
+                // the opcode, one for the one-byte payload), so we target the
+                // third byte. The value `n` here should always fit within a
```
abrown, line 217:
```suggestion
                // fourth byte. The value `n` here should always fit within a
```

Or describe which 0-based index.

abrown (approved):
Makes sense to get something working.

cranelift/codegen/src/binemit/mod.rs
```diff
@@ -119,6 +119,10 @@ pub enum Reloc {
     S390xTlsGd64,
     /// s390x TLS GDCall - marker to enable optimization of TLS calls
     S390xTlsGdCall,
+
+    /// Pulley - call a host function indirectly where the embedder resolving
+    /// this relocation needs to fill in the expected signature.
+    PulleyCallIndirectHost,
```
fitzgen, line 125: Does the signature actually need to be resolved at reloc time? It can't be done at compile time and embedded in the instruction itself?

The *address* of any host function obviously needs to be reloc time (this is a bit of an aside because my understanding is that we aren't actually embedding any host function addresses in the pulley bytecode) however the *signature* doesn't seem like it should need to be resolved at reloc time.

cranelift/codegen/src/isa/pulley_shared/abi.rs
```diff
@@ -541,36 +541,29 @@ where
...
-                &CallDest::Reg(reg) => smallvec![Inst::IndirectCall {
-                    info: Box::new(info.map(|()| XReg::new(*reg).unwrap()))
-                }
-                .into()],
+    fn gen_call(
+        dest: &CallDest,
+        _tmp: Writable<Reg>,
+        info: CallInfo<()>,
+    ) -> SmallVec<[Self::I; 2]> {
+        match dest {
```
fitzgen, line 549: Should we also check the calling conventions at all here? That was what I was (hackily) using to distinguish between pulley-to-pulley and pulley-to-host before. I like reloc-distance better but maybe we should be asserting that pulley-to-pulley always uses `tail` and pulley-to-host always uses `systemv` (which is a bit of a lie) or something like that?

cranelift/codegen/src/binemit/mod.rs
alexcrichton: I think what you're thinking is already done actually, but the phrasing here is ambiguous. The "reloc time" technically happens twice -- once when linking things into artifacts and again when loading the artifacts. Putting the signature into the instruction happens in the first of these, during linking time. The relocation here is needed because the `UserExternalName` isn't available during compilation, only after the compile has finished, so that level of relocation processing is required to stuff it in.

Otherwise though there's no runtime relocation when we load the bytecode itself, it's all frozen and loaded as-is from disk or the compile artifact.

alexcrichton: Good point yeah, this is also something that changed halfway through the design and I didn't get around to updating all the docs

PR #9666, opened 2024-11-22
pulley: Remove special registers as candidates from regalloc

alexcrichton:
This commit fixes a minor mistake where regalloc was allowed to allocate "special registers" such as the stack pointer, the frame pointer, etc. These register shouldn't participate in general-purpose register allocation, so they're removed from the list of non-preferred registers when metadata is provided to regalloc2.

### DIFF
--- a/cranelift/codegen/src/isa/pulley_shared/inst/args.rs
+++ b/cranelift/codegen/src/isa/pulley_shared/inst/args.rs
@@ -122,6 +122,12 @@ newtype_of_reg!(XReg, WritableXReg, RegClass::Int);
 newtype_of_reg!(FReg, WritableFReg, RegClass::Float);
 newtype_of_reg!(VReg, WritableVReg, RegClass::Vector);
 
+impl XReg {
+    /// Index of the first "special" register, or the end of which registers
+    /// regalloc is allowed to use.
+    pub const SPECIAL_START: u8 = pulley_interpreter::regs::XReg::SPECIAL_START;
+}
+
 pub use super::super::lower::isle::generated_code::ExtKind;
 
 pub use super::super::lower::isle::generated_code::Amode;
--- a/cranelift/codegen/src/isa/pulley_shared/abi.rs
+++ b/cranelift/codegen/src/isa/pulley_shared/abi.rs
@@ -807,7 +807,9 @@ fn create_reg_enviroment() -> MachineEnv {
     };
 
     let non_preferred_regs_by_class: [Vec<PReg>; 3] = {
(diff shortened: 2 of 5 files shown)


cranelift/codegen/src/isa/pulley_shared/inst/args.rs
```diff
@@ -122,6 +122,12 @@ newtype_of_reg!(XReg, WritableXReg, RegClass::Int);
 newtype_of_reg!(FReg, WritableFReg, RegClass::Float);
 newtype_of_reg!(VReg, WritableVReg, RegClass::Vector);
 
+impl XReg {
+    /// Index of the first "special" register, or the end of which registers
+    /// regalloc is allowed to use.
+    pub const SPECIAL_START: u8 = pulley_interpreter::regs::XReg::SPECIAL_START;
```
cfallin, line 128: Do we have constants for sp and fp? Can we define them like `SPECIAL_START + 0` and `SPECIAL_START + 1` to ensure this stays in sync?

cfallin (approved):
Good catch!

cranelift/codegen/src/isa/pulley_shared/inst/args.rs
alexcrichton: Does [this test](https://github.com/bytecodealliance/wasmtime/pull/9666/files#diff-577507af7de53d99c50170886b2c15e77f63375efaeb450058275f4eae3fc1d9R103) do what you're thinking?

cfallin: That tests the other direction: it asserts that every register from `SPECIAL_START` onward is special, but it doesn't test that every special register is beyond `SPECIAL_START`. I guess I'm suggesting a different way to define fp and sp, where we use the constant (plus a nonnegative offset) directly to have this by construction.

alexcrichton: Ah good point! The way these are defined in pulley is a bit special right now where each register is an `enum XReg { x0, ... sp, ... }` for example. That's uniform across x/f/v registers so I'm wary to change that much. That being said I'll enhance the test to assert that everything before `SPECIAL_START` is non-special.

cfallin (approved):

cranelift/codegen/src/isa/pulley_shared/inst/args.rs
cfallin: Looks good!

PR #9670, opened 2024-11-25
wasi-http: make the buffer and budget capacity of the OutgoingBody writer configurable

### LINKED ISSUE #9653: wasi-http: can we call the `blocking_write_and_flush` method of the `OutgoingRequest` body multiple times?
My scenario is simple: I want to send a request with a large body, but the `blocking_write_and_flush` method can only write 4096 bytes at a time. Naturally, I decided to call this method multiple times to write the entire body:

```rust
let req = OutgoingRequest::new(...);

let outgoing_body = req.body().unwarp();
let request_body = outgoing_body.write().unwarp();

let chunks = buf.chunks(4096);
for chunk in chunks {
    request_body.blocking_write_and_flush(chunk).expect("writing response");
}

OutgoingBody::finish(outgoing_body, None).unwarp();
```

However, I found that the program gets stuck forever on the second call to `blocking_write_and_flush`. After debugging, I found that it actually gets stuck during the `ready` check in the second call:

https://github.com/bytecodealliance/wasmtime/blob/5af89308dc0229ca404cd7000eec694201022e2d/crates/wasi-http/src/body.rs#L662-L669

https://github.com/bytecodealliance/wasmtime/blob/5af89308dc0229ca404cd7000eec694201022e2d/crates/wasi-http/src/body.rs#L472

At this point, the writer has no capacity left, but my writing process hasn't finished, so the reader hasn't started consuming the data (?).

For my issue, I could stop using this method and instead combine `check-write`, `subscribe`, `write`, and `flush` manually to solve it. However, I'm curious whether we are inclined to allow or not allow this behavior. Because the current behavior is strange—the program doesn't report an error, it just hangs indefinitely.

iawia002:
Please refer to issue #9653 for the background and reasons for the changes.

fix #9653

cc @alexcrichton @pchickey

### DIFF
--- a/crates/test-programs/src/bin/http_outbound_request_large_post.rs
+++ b/crates/test-programs/src/bin/http_outbound_request_large_post.rs
@@ -3,8 +3,11 @@ use std::io::{self, Read};
 use test_programs::wasi::http::types::{Method, Scheme};
 
 fn main() {
-    // TODO: ensure more than 700 bytes is allowed without error
-    const LEN: usize = 700;
+    // Make sure the final body is larger than 1024*1024, but we cannot allocate
+    // so much memory directly in the wasm program, so we use the `repeat`
+    // method to increase the body size.
+    const LEN: usize = 1024;
+    const REPEAT: usize = 1025;
     let mut buffer = [0; LEN];
     let addr = std::env::var("HTTP_SERVER").unwrap();
     io::repeat(0b001).read_exact(&mut buffer).unwrap();
@@ -13,7 +16,7 @@ fn main() {
         Scheme::Http,
         &addr,
         "/post",
-        Some(&buffer),
+        Some(&buffer.repeat(REPEAT)),
         None,
         None,
         None,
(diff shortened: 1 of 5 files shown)


pchickey (changes requested):
Thanks, we never got back to making these buffer depths configurable in WasiHttpView but this is the right path.

Some style notes:

Ideally we'd rewrite the internals to just give the user a single bound on the amount buffered, rather than a combination of number of chunks and chunk size. If you are up for doing that, I believe its the best path forward, but it is trickier than what we have. (Wanting that ideal implementation is mostly why we didn't make this configurable yet.)

Instead of calling it `writer_buffer` call it `outgoing_body_buffer_chunks`, described as the number of distinct write calls to the outgoing body's output-stream that the implementation will buffer. `write_budget` should be `outgoing_body_chunk_size`, the maximum size allowed in a write call to the outgoing body's output-stream.

The minimum of 2 chunks is an implementation detail related to the way channel reservations work, lets make the default configuration `1`,  assert! that any value given is greater than 1, and then always add 1 when creating the channel because one empty slot is required by the way this implementation uses the channel.

Also, I don't think we should change the configuration in `wasi-http/tests/all/main.rs`, but instead fix the implementation of `test_programs::http::request` which has the exact same bug I pointed out in your guest in https://github.com/bytecodealliance/wasmtime/issues/9653#issuecomment-2504699703.

pchickey:
Thanks! Its a holiday weekend here and I’ll review this next week.

alexcrichton (approved):
Looks reasonable to me, thanks! (Pat's taking a break for a bit)

I think this is reasonable enough that we can follow-up with further iterations if necessary.

PR #9673, opened 2024-11-25
Simplify some trap handling in `wasmtime` runtime

alexcrichton:
The `needs_backtrace` field on `TrapReason` is an old artifact of having the `wasmtime` and `wasmtime-runtime` crates split and there's no longer any need for that. This commit removes the field and directly queries it from the error as necessary when capturing backtraces.


pchickey (approved):
🚫 🍝

PR #9675, opened 2024-11-25
Update the host ABI of Wasmtime to return failure

alexcrichton:
This commit updates the "array call" ABI of Wasmtime used to transition from wasm to the host to explicitly return a `bool` indicating whether the call succeeded or not. Previously functions would implicitly unwind via `longjmp` and thus no explicit checks were necessary. The burden of unwinding is now placed on Cranelift-compiled code itself instead of the caller.

There are a few pieces of rationale for this change:

* Primarily I was implementing initial integration of Pulley where the host `setjmp` and `longjmp` cannot be used to maintain the Pulley interpreter state. My initial plan for handling this was to handle traps a bit differently in Pulley where having direct access to whether a function trapped or not in the interpreter bytecode is easier to deal with.

* Additionally it's generally not safe to call `longjmp` from Rust as it will not run on-stack destructors. This is ok today in the sense that we shouldn't have these in Wasmtime, but directly returning to compiled code improves our safety story here a bit where we just won't have to deal with the problem of skipping destructors.

* In the future I'd like to move away from `setjmp` and `longjmp` entirely in the host to a Cranelift-native solution. This change would be required for such a migration anyway so it's my hope to make such a Cranelift-based implementation easier in the future. This might be necessary, for example, when implementing the `exception-handling` proposal for Wasmtime.

Functionally this commit does not remove all usages of call-`longjmp`-from-Rust. Notably all libcalls and builtins still use this helper in the trampolines generated in Rust. I plan on going through the libcalls and updating their ABIs and signatures to reflect this in follow-up commits. As a result a few preexisting functions that should go away are marked `#[deprecated]` for internal use in this commit. I'll be cleaning that up as follow-ups. For now this commit handles the "hard part" of host functions by ensuring that the new `bool` return value is plumbed in all the locations correctly.


alexcrichton:
I'll note that this has a bit of https://github.com/bytecodealliance/wasmtime/pull/9673 mixed in but doesn't depend on it.

Also to expand on:

> I plan on going through the libcalls and updating their ABIs and signatures to reflect this in follow-up commits

The categories of libcalls are:

* Some libcalls, but not all, need to be able to raise traps. For example we support `memory.grow` raising a trap because of the embedder API but we don't support table-lazy-init raising a trap.
* All libcalls, however, need to catch Rust panics and do something with them.

My plan is to update the ABI of all libcalls that return traps to return some encoding of the current return value plus whether it trapped. That'll suffice for catching panics and handling traps. For all other libcalls which can't actually trap my plan is to abort the process. I plan on doing this by removing the `catch_unwind` entirely. That is no longer necessary for safety reasons in recent versions of Rust. That means that only libcalls which can raise a trap will have panics shepherded across them, and panics in other libcalls will abort the process.

crates/wasmtime/src/runtime/vm/libcalls.rs
```diff
@@ -109,6 +109,7 @@ pub mod raw {
                 ) $( -> libcall!(@ty $result))? {
                     $(#[cfg($attr)])?
                     {
+                        #[allow(deprecated)] // FIXME: need to update this
```
pchickey, line 112: Still a FIXME here and below in same file

crates/wasmtime/src/runtime/vm/traphandlers.rs
```diff
@@ -72,11 +83,16 @@ pub unsafe fn raise_user_trap(error: Error, needs_backtrace: bool) -> ! {
 /// the nearest `setjmp`. The panic will then be resumed from where it is
 /// caught.
 ///
+/// FIXME: this function should get removed in favor of
+/// `catch_unwind_and_record_trap` or a variant thereof.
+///
```
pchickey, line 88: FIXME

crates/wasmtime/src/runtime/vm/libcalls.rs
alexcrichton: Ah yeah this is intentional where it's deferred to a future PR to handle the usages for libcalls

cfallin (approved):
Cranelift compilation bits look good to me.

PR #9678, opened 2024-11-25
Move unix_stack_pool over to aligned byte counts

sunshowers:
This removes the last remaining use of `round_usize_up_to_host_pages`. I've added a couple of extra asserts that I believe are justified.


alexcrichton (approved):

sunshowers:
Looks like a flake.

PR #9681, opened 2024-11-26
[RFC] Move MemoryImageSource::map_at to mmap module

sunshowers:
This is part of the work to centralize memory management into the `mmap`
module. This commit introduces a few structures which aid in that process, and
starts converting one of the functions (`MemoryImageSource::map_at`) into this
module.

The structures introduced are:

* `MemoryBase`: `RuntimeLinearMemory::base_ptr` is now
  `RuntimeLinearMemory::base`, which returns a `MemoryBase`. This is either a
  raw pointer or an mmap + an offset into it.

* `MmapOffset`: A combination of a reference to an mmap and an offset into it.
  Logically represents a pointer into a mapped section of memory.

* `MmapOffsetRaw`: Some components like `MemoryImageSlot` logically work on
  borrowed memory, but adding lifetime parameters to them would introduce
  self-reference issues. Instead, store a raw form of the `MmapOffset` such
  that it can be reconstructed at runtime. This should work for most future
  work here, but not all of it -- I've written out some comments along with
  ideas.

On Zulip there was a suggestion to use `Arc<Mmap>` rather than a lifetime
parameter. To be honest it's quite appealing! One of the challenges though
is that `Mmap` has several `&mut` methods. The methods fall into two categories:

* Methods like [`make_accessible`](https://github.com/bytecodealliance/wasmtime/blob/0e3e863a57cd28e1ce9be628fd0d8df192077305/crates/wasmtime/src/runtime/vm/mmap.rs#L139), which can be changed to be `&self` since the OS performs synchronization of mapped memory.
* Methods like `slice_mut`, which are quite difficult to turn into `&self` methods. It would be too easy to call `self.slice_mut(0..host_page_size()); self.slice_mut(0..host_page_size());` and cause an insta-UB. I spent some time looking at how to do this but found it too difficult :(

### DIFF
--- a/crates/wasmtime/src/runtime/trampoline/memory.rs
+++ b/crates/wasmtime/src/runtime/trampoline/memory.rs
@@ -3,7 +3,7 @@ use crate::prelude::*;
 use crate::runtime::vm::mpk::ProtectionKey;
 use crate::runtime::vm::{
     CompiledModuleId, GcHeapAllocationIndex, Imports, InstanceAllocationRequest, InstanceAllocator,
-    InstanceAllocatorImpl, Memory, MemoryAllocationIndex, ModuleRuntimeInfo,
+    InstanceAllocatorImpl, Memory, MemoryAllocationIndex, MemoryBase, ModuleRuntimeInfo,
     OnDemandInstanceAllocator, RuntimeLinearMemory, RuntimeMemoryCreator, SharedMemory, StorePtr,
     Table, TableAllocationIndex,
 };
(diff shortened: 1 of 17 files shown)


sunshowers:
> Methods like slice_mut, which are quite difficult to turn into &self methods. It would be too easy to call self.slice_mut(0..host_page_size()); self.slice_mut(0..host_page_size()); and cause an insta-UB. I spent some time looking at how to do this but found it too difficult :(

Ah interesting, it looks like the only seriously affected use is in some tests.

sunshowers:
Moving to draft for now -- I think `Arc` is possible to make work.

sunshowers:
Yeah `Arc` is definitely possible -- going to re-do this after moving over to `Arc`.

PR #9682, opened 2024-11-26
Put MemoryPool and MmapMemory mmaps behind Arc

sunshowers:
In upcoming work we're going to centralize memory management inside `Mmap`.  In order to do that, we need memory that logically borrows from the `Mmap` to have access to it. That turns out to be possible to do by passing in `&Mmap` in some situations but very difficult in others (see #9681 for an attempt that was a real hassle to get working and still had a bunch of incomplete questions within it).

To prepare for this work, start moving these uses to be behind `Arc`. These aren't cloned at the moment, but they will be in the future.

### DIFF
--- a/crates/wasmtime/src/runtime/vm/cow.rs
+++ b/crates/wasmtime/src/runtime/vm/cow.rs
@@ -817,9 +817,33 @@ mod test {
         }
     }
 
-    fn mmap_4mib_inaccessible() -> Mmap<AlignedLength> {
+    fn mmap_4mib_inaccessible() -> Arc<Mmap<AlignedLength>> {
         let four_mib = HostAlignedByteCount::new(4 << 20).expect("4 MiB is page aligned");
-        Mmap::accessible_reserved(HostAlignedByteCount::ZERO, four_mib).unwrap()
+        Arc::new(Mmap::accessible_reserved(HostAlignedByteCount::ZERO, four_mib).unwrap())
+    }
+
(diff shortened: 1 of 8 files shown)


crates/wasmtime/src/runtime/vm/cow.rs
```diff
@@ -817,9 +817,33 @@ mod test {
...
+    /// function is done.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that during this function call, the only way this
+    /// region of memory is not accessed by (read from or written to) is via the
+    /// reference. Making the callback `'static` goes some way towards ensuring
+    /// that, but it's still possible to squirrel away a reference into global
+    /// state. So don't do that.
+    unsafe fn with_slice_mut(
```
sunshowers, line 837: This was somewhat annoying but I think this approach makes sense -- basically ensure that any `&mut` references are very short-lived as required by Rust's [aliasing rules](https://doc.rust-lang.org/std/ptr/index.html#pointer-to-reference-conversion).

alexcrichton:
I think this can make sense but I'm having a hard time understanding the end goal without the second piece here of using the `Arc` itself. Would you be ok delaying merging this until the second half is in place? It'd be easier for me to review once that's there (I'm ok reviewing the two together as two separate commits myself)

alexcrichton (approved):
Ok thanks for #9687! That looks reasonable to me so let's go ahead and land this 👍

PR #9687, opened 2024-11-26
Move MemoryImageSource::map_at to mmap module

sunshowers:
This is part of the work to centralize memory management into the `mmap`
module. This commit introduces a few structures which aid in that process, and
starts converting one of the functions (`MemoryImageSource::map_at`) into this
module.

The structures introduced are:

* `MemoryBase`: `RuntimeLinearMemory::base_ptr` is now
  `RuntimeLinearMemory::base`, which returns a `MemoryBase`. This is either a
  raw pointer or an `MmapOffset` as described below.

* `MmapOffset`: A combination of a reference to an mmap and an offset into it.
  Logically represents a pointer into a mapped section of memory.

In future work, we'll move more image-mapping code over to `Mmap` instances.

### DIFF
--- a/crates/wasmtime/src/runtime/vm/memory.rs
+++ b/crates/wasmtime/src/runtime/vm/memory.rs
@@ -77,11 +77,11 @@
 use crate::prelude::*;
 use crate::runtime::vm::vmcontext::VMMemoryDefinition;
 #[cfg(feature = "signals-based-traps")]
-use crate::runtime::vm::HostAlignedByteCount;
-use crate::runtime::vm::{MemoryImage, MemoryImageSlot, VMStore, WaitResult};
+use crate::runtime::vm::{HostAlignedByteCount, MmapOffset};
+use crate::runtime::vm::{MemoryImage, MemoryImageSlot, SendSyncPtr, VMStore, WaitResult};
 use alloc::sync::Arc;
-use core::ops::Range;
 use core::time::Duration;
+use core::{ops::Range, ptr::NonNull};
 use wasmtime_environ::{Trap, Tunables};
 
 #[cfg(feature = "signals-based-traps")]
@@ -160,7 +160,10 @@ pub trait RuntimeLinearMemory: Send + Sync {
     fn grow_to(&mut self, size: usize) -> Result<()>;
 
     /// Returns a pointer to the base of this linear memory allocation.
-    fn base_ptr(&self) -> *mut u8;
+    ///
+    /// This is either a raw pointer, or a reference to an mmap along with an
+    /// offset within it.
+    fn base(&self) -> MemoryBase;
 
     /// Internal method for Wasmtime when used in conjunction with CoW images.
     /// This is used to inform the underlying memory that the size of memory has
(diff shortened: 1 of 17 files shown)


crates/wasmtime/src/runtime/vm/memory.rs
```diff
@@ -488,14 +522,28 @@ impl LocalMemory {
...
+                    // was mapped on top of, so there is no need for the slot to
+                    // wipe it with an anonymous mapping first.
+                    //
+                    // Note that this code would be incorrect if clear-on-drop
+                    // were enabled. That's because in the struct definition,
+                    // `memory_image` above is listed after `alloc`. Rust drops
+                    // fields in the order they're defined, so `memory_image`
+                    // would be dropped after `alloc`. If clear-on-drop were
+                    // enabled, `memory_image` would end up remapping a bunch of
+                    // memory after it was unmapped by the `alloc` drop.
```
sunshowers, line 546: This was a subtle dependency that took me a bit to understand.

sunshowers: (And this is only dangerous when `alloc` represents an owned memory allocation. It's fine if `alloc` is borrowed. This is a subtle point.)

alexcrichton: Do you think it'd perhaps be better to remove this `no_clear_on_drop` and instead perform it in `Drop for LocalMemory`? (where that could `.take()` and explicitly do the image stuff before dropping the mmap)

alexcrichton (approved):
👍

crates/wasmtime/src/runtime/vm/memory.rs
sunshowers: Oh that's a good idea, similar to `MemoryPool`. I think it's correct.

sunshowers: Oh actually there's another subtle issue here, I think if `slot.instantiate` below fails, we should clear on drop (that's part of the point of clear-on-drop, right?) So I think it would be best to make this change in a followup since it's a behavior change, not a refactor.

sunshowers: I've expanded this comment -- it would definitely be good to move this to the destructor, but it would be a behavior change so it should be a followup. This PR itself should not contain any behavior changes.

sunshowers: Oh, and adding a `Drop` impl identified the problem with `unwrap_static_image`, haha, since you can't move out of a type when there's a `Drop` impl.

alexcrichton: Ah ok that all makes sense to me yeah, thanks for checking!

sunshowers:
Fixing the issues

sunshowers:
Hmm, actually I want to rename the os-specific Mmap `as_ptr` instances.

crates/wasmtime/src/runtime/vm/sys/windows/mmap.rs
```diff
@@ -133,7 +134,13 @@ impl Mmap {
...
             // remove the execute bit)
             let mut old = 0;
-            if VirtualProtect(ret.as_mut_ptr().cast(), ret.len(), PAGE_WRITECOPY, &mut old) == 0 {
+            if VirtualProtect(
+                ret.as_send_sync_ptr().as_ptr().cast(),
+                ret.len(),
+                PAGE_WRITECOPY,
+                &mut old,
+            ) == 0
+            {
```
sunshowers, line 143: Changed some of this code's pointer typology (though not the arithmetic) -- would be worth a look.

PR #9689, opened 2024-11-27
Port wasmtime-fiber to `no_std` and allow `async` feature in `no_std` Wasmtime.

cfallin:
This PR allows a `no_std` Wasmtime build to be configured with the `async` feature. (Previously, a minimal `no_std` configuration could only run with sync entry points, without suspending of stacks.)

The main hurdle to this support was the `wasmtime-fiber` crate. Fortunately, the "unix" variant of fibers was almost entirely portable to a `no_std` environment, owing to the fact that it implements stack-switching manually in assembly itself. I moved the per-ISA implementations to a shared submodule and built the nostd platform backend for `wasmtime-fiber` with a stripped-down version of the unix backend.

The nostd backend does not support mmap'd stacks, does not support custom stack allocators, and does not propagate panics.

I've updated the `min-platform` example to ensure this builds but have not yet actually tested it (I am in the middle of a larger porting effort); hence, a draft PR for initial feedback.


cfallin:
cc @alexcrichton -- I think you're on vacation now but I'd be very curious what you think of this in general!

crates/fiber/src/lib.rs
```diff
@@ -220,7 +239,17 @@ impl<Resume, Yield, Return> Suspend<Resume, Yield, Return> {
...
-        let result = panic::catch_unwind(AssertUnwindSafe(|| (func)(initial, &mut suspend)));
+
+        let result;
+        #[cfg(feature = "std")]
+        {
+            result = panic::catch_unwind(AssertUnwindSafe(|| (func)(initial, &mut suspend)));
+        }
+        #[cfg(not(feature = "std"))]
+        {
+            result = Ok((func)(initial, &mut suspend));
```
bjorn3, line 250: If you do unwind when the std feature is disabled, are you guaranteed to hit an `extern "C"` boundary to abort unwinding? Or will you get arbitrary corruption?

cfallin: Excellent question! (@alexcrichton when you're around it'd be great to get your canonical answer on this)

Currently in my porting context I have a `#[panic_handler]` that will abort, so I haven't thought through this in much detail. One option is certainly to say that panic unwinding is not supported in `no_std` builds with `async`, and the build must provide a custom aborting panic handler or specify `panic=abort`. However if we wanted to support propagation of unwinds I think we might be able to do something custom here (the `catch_unwind`/`resume_unwind` functions in `std::panic` don't exist in `core::panic` but one could roll one's own unwinding). I wonder if there's a way to make this safe (build only when `panic=abort` for example) in the meantime?

cfallin: @sunfishcode helpfully pointed out that `#[cfg(panic = "unwind")]` and `#[cfg(panic = "abort")]` work, so I've added a `compile_error!()` to the fiber library when built with `panic=unwind` without `std`. So this is at least now safe!

.github/workflows/main.yml
```diff
@@ -355,12 +355,12 @@ jobs:
               -p wasmtime --no-default-features --features wat
               -p wasmtime --no-default-features --features profiling
               -p wasmtime --no-default-features --features cache
-              -p wasmtime --no-default-features --features async
+              -p wasmtime --no-default-features --features async,std
```
alexcrichton, line 358: Was this necessary because it was otherwise broken? Or just keeping the same behavior tested?

If broken, could `wasmtime` enable the `std` feature as part of the async feature for now? (until that's fixed)

Otherwise if just keeping the same behavior mind also adding `-p wasmtime-fiber --features std` and `-p wasmtime-fiber` to this matrix?

crates/fiber/Cargo.toml
```diff
@@ -33,3 +33,10 @@ wasmtime-versioned-export-macros = { workspace = true }
 
 [dev-dependencies]
 backtrace = "0.3.68"
+
+[features]
+default = []
```
alexcrichton, line 38: It's ok to drop this if it's empty, and due to this being an internal crate I think it's reasonable to not have default features here anyway.

crates/wasmtime/Cargo.toml
```diff
@@ -302,6 +301,7 @@ std = [
   'wasmtime-environ/std',
   'object/std',
   'once_cell',
+  'wasmtime-fiber/std',
```
alexcrichton, line 304: I might recommend `wasmtime-fiber?/std` (note the `?`) to only enable the `wasmtime-fiber` dependency if it's already activated via some other means.

examples/min-platform/embedding/Cargo.toml
```diff
@@ -14,7 +14,7 @@ anyhow = { workspace = true }
 # Note that default-features of wasmtime are disabled and only those required
 # are enabled, in this case compilation is done in the guest from the wasm text
 # format so `cranelift` and `wat` are enabled.
-wasmtime = { workspace = true, features = ['runtime'] }
+wasmtime = { workspace = true, features = ['runtime', 'async'] }
```
alexcrichton, line 17: I think it's ok to leave this out for now and mostly just ensure this is `cargo check`'d somewhere in CI (to avoid someone copy/pasting this `Cargo.toml` example and forgetting to turn this back off)

crates/fiber/src/lib.rs
```diff
@@ -96,7 +118,7 @@ pub unsafe trait RuntimeFiberStackCreator: Send + Sync {
     ///
     /// This is useful to plugin previously allocated memory instead of mmap'ing a new stack for
     /// every instance.
-    fn new_stack(&self, size: usize) -> Result<Box<dyn RuntimeFiberStack>, Error>;
+    fn new_stack(&self, size: usize) -> core::result::Result<Box<dyn RuntimeFiberStack>, Error>;
```
alexcrichton, line 121: A nice trick to avoid `core::result::Result` here is to, in each module, change:

```rust
type Result<T> = some::Result<T>;
```

to 

```rust
type Result<T, E = some::Error> = core::result::Result<T, E>;
```

That way you can use any `Result` type-alias as the actual real `Result` without worry

crates/fiber/src/lib.rs
```diff
@@ -1,13 +1,32 @@
+#![no_std]
+
+// no_std support: we can only support `panic=abort` builds, because the
+// `core::panic` module otherwise does not provide us the catch/resume
+// unwinding primitives we need to propagate unwinding across stacks.
+#[cfg(all(panic = "unwind", not(feature = "std")))]
+compile_error!(
+    "no_std builds with wasmtime-fiber (e.g., Wasmtime async) only support `panic=abort`."
+);
```
alexcrichton, line 9: Personally I think it's reasonable to skip the `compile_error!` here and just let this happen as normal. With `panic = "abort"` nothing should be unsafe, it just means the extra unwind bits here aren't too useful (and they'll get optimized out)

crates/fiber/src/lib.rs
```diff
@@ -1,13 +1,32 @@
...
-use std::cell::Cell;
-use std::io;
-use std::marker::PhantomData;
-use std::ops::Range;
+use core::any::Any;
+use core::cell::Cell;
+use core::marker::PhantomData;
+use core::ops::Range;
+#[cfg(feature = "std")]
 use std::panic::{self, AssertUnwindSafe};
```
alexcrichton, line 23: Mind scoping this import to just where it's needed below? (avoid the duplicate `#[cfg]`)

crates/fiber/src/lib.rs
```diff
@@ -175,7 +197,12 @@ impl<'a, Resume, Yield, Return> Fiber<'a, Resume, Yield, Return> {
                 Err(y)
             }
             RunResult::Returned(r) => Ok(r),
-            RunResult::Panicked(payload) => std::panic::resume_unwind(payload),
+            RunResult::Panicked(_payload) => {
+                #[cfg(feature = "std")]
+                panic::resume_unwind(_payload);
+                #[cfg(not(feature = "std"))]
+                unreachable!()
+            }
```
alexcrichton, line 205: Could this be avoided by having a `#[cfg]` for std/unwind on the `Panicked` variant? That way we can statically provide that it's never constructed

crates/fiber/src/lib.rs
alexcrichton: I commented below as well, but I think it's fine to rely on unconditional aborts via `extern "C"` boundary in the panic=unwind std=false mode and skip the `compile_error!`

crates/fiber/src/lib.rs
bjorn3: This error only happens when using `panic = "unwind"`, which is unsound if not using `catch_unwind`.

alexcrichton: Ah yeah sorry that's just "faking" a non-inlined call to something that can theoretically unwind. You don't actually need `C-unwind` but with `panic=unwind` all normal rust functions can possibly unwind so a catch-and-panic should be inserted for all calls. Basically I'm still not sure how `#![no_std]` + `-Cpanic=unwind` is unsound myself

cfallin: I see: my confusion is the dual to your confusion I think, that is, I'm still not sure how `#![no_std]` + `-Cpanic=unwind` is sound :-)

Restating mostly for my own benefit, but to verify we're on the same page: so with the current state of `main`, we wrap a call to the entrypoint on the new stack in a call to `catch_unwind` (e.g. [here](https://github.com/bytecodealliance/wasmtime/blob/db956669a5bc7de1ac247c63b4ad895965469144/crates/fiber/src/lib.rs#L225)). If the fiber's body panics while on the fiber stack, system `libunwind` reaches this point, the panic is reified as a `RunResult`, and that `RunResult` is passed back to the caller on the original stack and rethrown with `resume_unwind`.

With this PR, in `no_std`, we don't have `catch_unwind`/`resume_unwind`. So a panic happens in a fiber's body; say that there is a custom unwinder (there must be I think, for one to be doing unwinding at all in `no_std`?); this unwinder iterates up the stack frames on the fiber stack. It reaches the top, there's no callsite in a `catch_unwind`; this is an unwind error / UB / something bad. Or is this prevented some other way?

alexcrichton: Oh sorry! Good point getting on the same page (can also talk more during the Cranelift meeting tomorrow if helpful)

I think the main thing to add is that `extern "C" fn foo() { panic!() }` unconditionally, and soundly, aborts the process. That happens irrespective of `-Cunwind=...` settings. Starting in Rust 1.81 all `extern "C"` functions have an implicit catch-and-abort. That means that even if we were to remove the `catch_unwind` happening in wasmtime-fiber the crate would still be sound, just not too helpful on panics. The `catch_unwind` is mostly still there to assist with debugging panics by shepherding the panic across the boundary. That's always optional though since it's sound to abort the process at any time (just better to not do so if possible).

That to me is why it's currently sound if we skip the `catch_unwind`. A bit of a historical note though is that Rust 1.81 is very recent and prior versions of rustc didn't catch-and-abort at `extern "C"` boundaries. That means that historically `catch_unwind` was indeed required for soundness but that's no longer the case.

cfallin: Ah! OK, that's the detail I had missed; thanks for your patience here. I guess since our MSRV is now 1.81, we can rely on this behavior unconditionally. I'll remove the compile-error and add a comment detailing this.

crates/fiber/src/lib.rs
cfallin: Done!

examples/min-platform/embedding/Cargo.toml
cfallin: Done.

crates/wasmtime/Cargo.toml
cfallin: Done.

crates/fiber/Cargo.toml
cfallin: Done!

.github/workflows/main.yml
cfallin: I think I had added it at some point while debugging failures but indeed it should work with just `async`, no `std` -- so, restored that, added `std` separately, and added the `wasmtime-fiber` variants.

crates/fiber/src/lib.rs
cfallin: Done.

crates/fiber/src/lib.rs
cfallin: Done -- refactored a bit in `Suspend::execute()` to statically avoid constructing the panic case as well.

cfallin:
Updated and out of draft mode now; thanks!

alexcrichton (approved):
Is it possible to add a `cargo test -p wasmtime-fiber --no-default-features to CI as well?

cfallin:
Sure thing, done!

cfallin:
The CI failures are fairly perplexing -- the only failing job I see is "report failure on cancelation", and a bunch of other jobs are canceled midway through their runs. @alexcrichton any ideas?

alexcrichton:
This happened on another PR the other night and I couldn't figure out what was happening. I then re-queued in the morning and it merged so I assumed it was a transient failure in CI infrastructure... Let's see if that still holds true?

.github/workflows/main.yml
```diff
@@ -432,6 +439,18 @@ jobs:
...
+    name: wasmtime-fiber tests
+    runs-on: ubuntu-latest
+    env:
+      CARGO_NDK_VERSION: 2.12.2
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        submodules: true
+    - uses: ./.github/actions/install-rust
+    - run: cargo test -p wasmtime-fiber --no-default-features
```
alexcrichton, line 452: Oh when adding a new job like this you'll need to also add a "join" to the `ci-status` below so it's handled for failure-vs-not.

alexcrichton:
I've pushed a commit to run full CI on this PR where the PR run can't cancel itself, so let's see what happens...

.github/workflows/main.yml
cfallin: Done!

PR #9697, opened 2024-11-30
cranelift-entity: Make PackedOption and EntityList have guaranteed layout

caelunshun:
Adds optional implementations of `Pod` and `Zeroable` for `EntityList` and `PackedOption`.

This helps avoid some unsafe code in my downstream crates.

### DIFF
--- a/cranelift/entity/src/list.rs
+++ b/cranelift/entity/src/list.rs
@@ -45,7 +45,7 @@ use serde_derive::{Deserialize, Serialize};
 /// The `EntityList` itself is designed to have the smallest possible footprint. This is important
 /// because it is used inside very compact data structures like `InstructionData`. The list
 /// contains only a 32-bit index into the pool's memory vector, pointing to the first element of
-/// the list.
+/// the list. A zero value represents the empty list, which is returned by `EntityList::default`.
 ///
 /// The pool is just a single `Vec<T>` containing all of the allocated lists. Each list is
 /// represented as three contiguous parts:
--- a/cranelift/entity/src/list.rs
+++ b/cranelift/entity/src/list.rs
@@ -64,6 +64,7 @@ use serde_derive::{Deserialize, Serialize};
 /// reserved for the empty list which isn't allocated in the vector.
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
+#[repr(C)]
 pub struct EntityList<T: EntityRef + ReservedValue> {
     index: u32,
     unused: PhantomData<T>,
--- a/cranelift/entity/src/packed_option.rs
+++ b/cranelift/entity/src/packed_option.rs
@@ -22,6 +22,9 @@ pub trait ReservedValue {
 }
 
 /// Packed representation of `Option<T>`.
+///
+/// This is a wrapper around a `T`, using `T::reserved_value` to represent
+/// `None`.
 #[derive(Clone, Copy, PartialEq, PartialOrd, Eq, Ord, Hash)]
 #[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
 #[repr(transparent)]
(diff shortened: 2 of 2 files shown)


abrown:
@cfallin, you want to take this review?

cfallin:
@abrown Will do!

cranelift/entity/src/list.rs
```diff
@@ -639,6 +640,19 @@ impl<T: EntityRef + ReservedValue> EntityList<T> {
...
 
+#[cfg(feature = "bytemuck")]
+mod bytemuck_impl {
+    use crate::packed_option::ReservedValue;
+    use crate::{EntityList, EntityRef};
+    use bytemuck::{Pod, Zeroable};
+
+    // Soundness: `EntityList` is a `repr(C)` struct
+    // containing only a `u32` and a zero-sized `PhantomData`,
+    // so it can always implement `Pod` regardless of `T`.
```
cfallin, line 651: This is a valid argument for the soundness of `Pod`, but not of `Zeroable`, as far as I can tell, right? In fact an all-zero-bits `EntityList` is not guaranteed to be valid at all, and we don't otherwise allow the existence of `EntityList` values that are "dangling"...

cfallin (commented):
Thanks for the PR, @caelunshun!

I have a specific comment below, but also a higher-level thought: in general we try to avoid taking on dependencies, even optional ones, because it adds to our vetting overhead (you can see the `cargo vet` failure in CI here; one of us would have to go and read the `bytemuck` sources to vet it) and just in general adds complexity. In particular here I'm somewhat concerned that this crate permits unsafe low-level casting and is claiming additional properties about our data structures; they may be true today, or may not (see comment), but we're also taking on additional burden reasoning about and upholding these properties going forward.

For your particular use-case, would it be a possible alternative for you to wrap whatever types need these trait impls in your own types, and impl them locally? That way you could carry the maintenance task of ensuring the properties still hold, and of vetting dependencies, without adding those overheads here. Curious what you think.

caelunshun (commented):
Thanks for the review. Yes, not wanting to take on extra dependencies makes sense; I wasn't aware of the vetting policy.

From my impression `bytemuck` is a fairly central ecosystem crate, seeing use in security-conscious projects like Servo and wgpu, and is also [vetted for use in Google's projects](https://github.com/google/rust-crate-audits/blob/93e07327e17465f183d18a825639880f876227fc/manual-sources/google3-audits.toml#L270). And given `cranelift-entity`'s applications outside of Wasmtime, integrating with the broader ecosystem would make sense to me. But it's not a big deal to me one way or the other.

As for your comment about guaranteeing a particular representation, the docs already do so implicitly, with `EntityList` being described as a single `u32` and `PackedOption<T>` being a trivial wrapper around `T`. So implementing these traits in `cranelift-entity` just allows downstream crates to take advantage of this layout without unsafe code that could silently break with a version bump.

The reason I couldn't write a newtype around `cranelift-entity` types and implement `Pod` is because their layout is not currently guaranteed; the upstream types would need to be `repr(C)` or `repr(transparent)` (as changed in this PR) for that to be sound. I could change the PR to simply add these `repr` annotations and not the `bytemuck` dependency if you like.

cranelift/entity/src/list.rs
cfallin: Ah, yes, you're right; I had forgotten that we use zero as an empty-list sentinel here. Thanks!

cfallin:
Thanks for the additional detail! Given all that, I think the deciding factor for me is the vetting overhead: we apparently don't pull in that Google vet (or it's not sufficient for our required level?) and I don't have cycles to go read all of `bytemuck` and convince myself of its correctness right now, so I think I'd prefer to see this be a minimal PR that adds the necessary `repr` directives to structs instead and documents invariants (zero-valued sentinels, etc) appropriately. I'd be happy to approve and merge such a PR.

caelunshun:
@cfallin Great, done.

cfallin (approved):

cfallin:
Thanks!

PR #9700, opened 2024-12-01
[DWARF] Synthetic type centralization

SingleAccretion:
The DWARF transform needs to add some "synthetic" types to the output, mainly to represent the VMContext structure.

So far, these types were being added to each CU, which is unnecessary overhead since they're exactly the same in each CU in a given module.

This change moves them to a single per-module CU that is references via DWARF's `DW_FORM_ref_addr` mechanism.

This is an optimization in both time (the debugger needs fewer types to keep track of) and space (the `.debug_info` section is ~5% smaller on some DWARF input I have).

### DIFF
--- a/crates/cranelift/src/debug/transform/simulate.rs
+++ b/crates/cranelift/src/debug/transform/simulate.rs
@@ -1,7 +1,7 @@
 use super::expression::{CompiledExpression, FunctionFrameInfo};
-use super::utils::{add_internal_types, append_vmctx_info};
+use super::utils::append_vmctx_info;
 use super::AddressTransform;
-use crate::debug::{Compilation, ModuleMemoryOffset};
+use crate::debug::Compilation;
 use crate::translate::get_vmctx_value_label;
 use anyhow::{Context, Error};
 use cranelift_codegen::isa::TargetIsa;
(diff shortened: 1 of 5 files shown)


crates/cranelift/src/debug/transform/synthetic.rs
```diff
@@ -0,0 +1,179 @@
...
+        let unit = out_units.get_mut(unit_id);
+        let unit_die = unit.get_mut(unit.root());
+        unit_die.set(
+            gimli::DW_AT_name,
+            AttributeValue::StringRef(out_strings.add("WasmtimeModuleSyntheticUnit")),
+        );
+        unit_id
+    }
+
+    fn create_vmctx_ptr_die(
```
SingleAccretion, line 75: Below code was simply moved from utils.rs without functional changes.

crates/cranelift/src/debug/transform/simulate.rs
```diff
@@ -345,13 +343,7 @@ pub fn generate_simulated_dwarf(
...
     };
 
-    let mut module_wasm_types = PrimaryMap::new();
-    for (module, memory_offset) in compilation.module_memory_offsets.iter() {
-        let wasm_types = add_wasm_types(unit, root_id, out_strings, memory_offset);
-        let i = module_wasm_types.push(wasm_types);
-        assert_eq!(i, module);
-    }
-
+    let wasm_types = add_wasm_types(unit, root_id, out_strings);
```
SingleAccretion, line 346: Since all of the remaining `wasm_types` are module-invariant, there is no longer a need for them to be per-module.

PR #9702, opened 2024-12-02
Rely on `core::error::Error`

alexcrichton:
With Wasmtime's new MSRV at 1.81 this means that `core::error::Error` is available which means that in `no_std` mode the `Error` trait can be used. This has been integrated into `anyhow::Error` already upstream and means that we can remove our own local hacks such as the `Err2Anyhow` trait.

This commit removes the `Err2Anyhow` trait and all usage, going back to idiomatic Rust error propagation and conversion even in the `no_std` world. This should make code more portable by default and remove some weird idioms we had for supporting this.


alexcrichton:
I'll note that this is a draft as this depends on a few upstream PRs:

* https://github.com/gimli-rs/object/pull/747
* https://github.com/bytecodealliance/wasm-tools/pull/1928
* An update in this repo of wasm-tools (blocked on https://github.com/bytecodealliance/wasm-tools/pull/1927)

pchickey (approved):

alexcrichton:
Ok everything's good to go except `object`. I've also added an update of `anyhow` and `serde` to pull in their latest versions which have support for `core::error::Error`. A small workaround for `object::Error` has been added temporarily which can be removed once the upstream PR is merged and/or the feature is settled one way or another.

alexcrichton:
Ok through much trial and error I've discovered that the s390x-specific failure on CI with this PR is due to the `anyhow` update. That's specifically the 1.0.76 to 1.0.77 update (which we're jumping from 75 to 93 here). This update uses `std::backtrace` by default and respects `RUST_BACKTRACE=1`. For the longest time I couldn't reproduce this locally, and turns out I copied over all env vars in CI *except* `RUST_BACKTRACE=1`.

I've disabled backtraces in CI for now. If that works and this lands I'll file an issue and cc Ulrich.

PR #9703, opened 2024-12-02
Update wasm-tools crates

alexcrichton:
Mostly comes with a split in `wasmparser` of simd/other ops since it's now possible to disable the `simd` feature at compile time (which we don't use in Wasmtime right now).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1364,6 +1364,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 
 [[package]]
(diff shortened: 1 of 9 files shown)


pchickey (approved):

alexcrichton:
Upstream in wasm-tools the test suite for spec tests [has stopped asserting](https://github.com/bytecodealliance/wasm-tools/pull/1909) that the spec-provided error message matches the wasmparser-provided error message. That means that the error-message-matching here in Wasmtime also needs to go since the same "failures" were cropping up there. I've adjusted the function to match errors in a similar way now which caused some tests to start passing. I think the new tests "passing" are probably where the error before was "wasmtime doesn't support this wasm instruction" and the expected error is something like "type mismatch" which would explain that.

Issue #8091, opened 2024-03-12
winch: Epoch based interruption


saulecabrera:
Winch is missing support for some of Wasmtime's features. The objective of this issue is to track progress of adding epoch based interruption to Winch.

PR #9708, opened 2024-12-02
winch: Improve frame handling

saulecabrera:
This commit addresses issues identified while working on issue #8091. It improves the frame handling in Winch to prevent subtle bugs and enhance the robustness of the code generation process.

Previously, there was no clear mechanism to verify when the frame was fully set up and safe to access the local slots allocated for register arguments, including the special slots used for the `VMContext`. As a result, it was possible to inadvertently read from uninitialized memory if calls were made before the frame was properly set up and sealed.

This commit introduces two main changes with the objective to help reduce the risk of introducing bugs related to the above:

* A `CodeGenPhase` trait, used via the type state pattern to clearly gate the operations allowed during each phase of the code generation process.

* Improve the semantics of locals, by clearly separating the notion of Wasm locals and special locals used by the compiler. This specialization allows a more accurate representation of the semantics of Wasm locals and their index space.

### DIFF
--- a/winch/codegen/src/abi/mod.rs
+++ b/winch/codegen/src/abi/mod.rs
@@ -212,21 +212,6 @@ impl ABIOperand {
             _ => unreachable!(),
         }
     }
-
-    /// Get the register associated to this [`ABIOperand`].
-    pub fn get_reg(&self) -> Option<Reg> {
-        match *self {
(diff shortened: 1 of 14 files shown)


abrown (approved):
Makes sense; sorry for missing this review earlier.

saulecabrera:
Thanks for the review @abrown -- and no worries!

PR #9709, opened 2024-12-02
winch: Prefer using `Context::without` for fuel checks

saulecabrera:
This commit is a follow-up to
https://github.com/bytecodealliance/wasmtime/pull/9472. It ensures that the fuel checking code makes use of `Context::without` to retrieve the current fuel register.

Even though no issues have been found, it is preferable to use `Context::without` in scenarios in which we have to introduce special out-of-band control flow and function calls to ensure that all the register allocation invariants are respected i.e., avoid holding onto to registers that: are not  in the stack and that might be needed for function calls.


cfallin (approved):
LGTM!

PR #9710, opened 2024-12-02
 Migrate libcalls to checking for traps 

alexcrichton:
This PR is the equivalent of https://github.com/bytecodealliance/wasmtime/pull/9675 for libcalls used in core wasm and components.
All libcalls now communicate whether or not they trapped through their
return value instead of implicitly calling `longjmp` to exit from the
libcall. This is to make integration with Pulley easier to avoid the
need to `longjmp` over Pulley execution.

Libcall definitions have changed where appropriate and the
`catch_unwind_and_record_trap` function introduced in https://github.com/bytecodealliance/wasmtime/pull/9675 was
refactored to better support multiple types of values being returned
from libcalls (instead of just `Result<()>`).

Note that changes have been made to both the Cranelift translation layer
and the Winch translation layer for this as the ABI of various libcalls
are all changing.

### DIFF
--- a/crates/cranelift/src/gc/enabled/drc.rs
+++ b/crates/cranelift/src/gc/enabled/drc.rs
@@ -283,8 +283,8 @@ fn emit_gc_raw_alloc(
         .call(gc_alloc_raw_builtin, &[vmctx, kind, ty, size, align]);
 
     let gc_ref = builder.func.dfg.first_result(call_inst);
+    let gc_ref = builder.ins().ireduce(ir::types::I32, gc_ref);
     builder.declare_value_needs_stack_map(gc_ref);
-
     gc_ref
 }
 
--- a/winch/codegen/src/codegen/mod.rs
+++ b/winch/codegen/src/codegen/mod.rs
@@ -992,6 +992,7 @@ where
             continuation,
             OperandSize::S64,
         );
+        self.context.free_reg(fuel_var);
         // Out-of-fuel branch.
         FnCall::emit::<M>(
             &mut self.env,
@@ -999,9 +1000,10 @@ where
             &mut self.context,
             Callee::Builtin(out_of_fuel.clone()),
         );
+        self.context.pop_and_free(self.masm);
+
         // Under fuel limits branch.
         self.masm.bind(continuation);
-        self.context.free_reg(fuel_var);
     }
 
     /// Emits a series of instructions that load the `fuel_consumed` field from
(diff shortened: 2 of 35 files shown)


winch/codegen/src/codegen/mod.rs
```diff
@@ -948,9 +948,9 @@ where
             &mut self.context,
             Callee::Builtin(out_of_fuel.clone()),
         );
+        self.raise_if_host_trapped();
         // Under fuel limits branch.
         self.masm.bind(continuation);
-        self.context.free_reg(fuel_var);
```
alexcrichton, line 953: Similarly @saulecabrera I had to move this up since `fuel_var` was assigned rax but rax was needed as the return value of the hostcall, so this was needed to fix some panics during compilation.

sunfishcode:
Does this significantly change overall code size of Cranelift-generated code, due to needing inline trap check code at each libcall callsite?

alexcrichton:
Yes as-is all libcalls that can trap have a check after them, so it's inflating code by at least that much.

In thinking though this can be pretty easily optimized. We already generate a trampoline-per-libcall so I'll move the check-for-trap logic to the trampoline instead of the callsite of the trampoline.

alexcrichton:
Ok I've updated this where the trampoline itself now checks for a trap, so the code size impact on wasm itself should be minimal.

crates/cranelift/src/gc/enabled/drc.rs
```diff
@@ -283,8 +283,8 @@ fn emit_gc_raw_alloc(
         .call(gc_alloc_raw_builtin, &[vmctx, kind, ty, size, align]);
 
     let gc_ref = builder.func.dfg.first_result(call_inst);
+    let gc_ref = builder.ins().ireduce(ir::types::I32, gc_ref);
     builder.declare_value_needs_stack_map(gc_ref);
-
```
fitzgen, line 287: Wait, we should only ever have `i32` values in stack maps, but this will make it include an `i64` that represents a `Result<u32>` IIUC, so this change looks pretty suspect to me. Is this intentional?

fitzgen (approved):
LGTM modulo one comment below

crates/cranelift/src/gc/enabled/drc.rs
alexcrichton: Oh I thought that the `ireduce` preceding this would take care of that, but is that not the case?

fitzgen: Ah whoops I misread this as moving the `ireduce` *after* the needs-stack-map declaration. Carry on!

Issue #9698, opened 2024-12-01
Consider integrating build provenance attestation into our CD pipeline


tschneidereit:
#### Feature

GitHub provides an [action](https://github.com/actions/attest-build-provenance) for creation build provenance attestation for Actions-built artifacts, which we could integrate into our pipeline with, I think, low effort.

#### Benefit

Build provenance attestation would provide consumers with verifiable guarantees of the artifacts' provenance.

#### Implementation

Based on the Action's description, implementation seems fairly straight-forward and should involve not much besides following the steps documented there.

#### Alternatives

There might be alternative forms of attestation. This one has the advantage of being officially suppported by GitHub, and showing up in the project's [attestations tab](https://github.com/bytecodealliance/wasmtime/attestations).

alexcrichton:
Well we have [attestations now](https://github.com/bytecodealliance/wasmtime/attestations) as they're rolling in for the artifacts produced for the `dev` tag. I downloaded a random one and ran:

```
$ ./gh_2.63.1_linux_amd64/bin/gh attestation verify --owner bytecodealliance ./wasmtime-dev-riscv64gc-linux.tar.xz
Loaded digest sha256:525d1ac2051ce2c903b6697c7875f85f27f1696e24993494b6d6506e8777d7be for file://wasmtime-dev-riscv64gc-linux.tar.xz
Loaded 1 attestation from GitHub API
✓ Verification succeeded!

sha256:525d1ac2051ce2c903b6697c7875f85f27f1696e24993494b6d6506e8777d7be was attested by:
REPO                       PREDICATE_TYPE                  WORKFLOW
bytecodealliance/wasmtime  https://slsa.dev/provenance/v1  .github/workflows/publish-artifacts.yml@refs/heads/main
```

so it looks like... at least something is working!

PR #9711, opened 2024-12-02
Test out GitHub build attestations

alexcrichton:
Unsure what will happen, but it probably can't break things too much... right?

cc #9698


tschneidereit (approved):
This looks right to me per the documentation for the action. Beyond that, I guess all we can do is land it and see what the result looks like?

PR #9713, opened 2024-12-03
Update QEMU used in CI

alexcrichton:
I'm getting segfaults on s390x in #9702 and I can't reproduce them locally in QEMU. Try updating QEMU to see if it's a transient bug. Historical updates haven't always gone well, so let's see what CI says.


alexcrichton:
Oh hey all green! Should be good to go. Doesn't fix https://github.com/bytecodealliance/wasmtime/pull/9702 though, so must be something else...

cfallin (approved):

PR #9715, opened 2024-12-03
Test out ubuntu-24.04 in CI

alexcrichton:
According to https://github.com/actions/runner-images/issues/10636 this'll be rolling out soon, gonna try to weed out problems in this PR ahead of time if there are any.


alexcrichton:
answer: much breakage!

PR #9717, opened 2024-12-03
Pin some runner images

alexcrichton:
For the larger test/build jobs use some variables in the `*.js` files calculating the matrix to pin the images to fixed versions of GitHub actions images. For the `main.yml` file instead of changing everything over to a pin only do the ones that are currently failing on the `ubuntu-24.04` update. The hope is that there's only a few locations to update pinned versions in the future, and we'll need to keep an eye on CI warnings and such to know when to update these pins in the future.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -80,7 +80,7 @@ jobs:
   # Note that `wasmtime-platform.h` is excluded here as it's auto-generated.
   clangformat:
     name: Clang format
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04 # FIXME: fails on `ubuntu-24.04` right now
     steps:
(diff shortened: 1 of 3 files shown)


alexcrichton:
(result of learnings from https://github.com/bytecodealliance/wasmtime/pull/9715)

alexcrichton:
For some longer-form rationale, my thinking here is it's not an unreasonable default to stick to `*-latest` by default, but for the "big" jobs it's probably more reasonable to use versioned images by default and explicitly update them to catch issues. That's where the `*.js` files all use versioned images and unfortunately we need two versioned images in `main.yml`, but I'm thinking we shouldn't go "version all the things" in the `main.yml` file just yet.

cfallin (approved):

PR #9718, opened 2024-12-03
pulley: Fill out more bits and pieces of stack allocation

alexcrichton:
* Use `push_frame` and `pop_frame` in the Cranelift backend.
* Use dedicated instructions for modifying the stack pointer
  * decrementing the stack pointer is always checked
  * incrementing the stack pointer is checked in debug mode
* Disregard stack probes and stack_limit checks as they're both unnecessary for Pulley (all stack allocations are checked).


abrown (approved):
LGTM!

PR #9721, opened 2024-12-03
Add basic pulley disassembly support to `clif-util`

alexcrichton:
Helpful when running `clif-util compile -D --target pulley64` over the input.

### DIFF
--- a/pulley/src/disas.rs
+++ b/pulley/src/disas.rs
@@ -274,7 +274,7 @@ impl<'a> OpVisitor for Disassembler<'a> {
                 write!(&mut self.disas, "{space}{byte:02x}").unwrap();
                 need_space = true;
             }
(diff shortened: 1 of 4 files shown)


pulley/src/disas.rs
```diff
@@ -274,7 +274,7 @@ impl<'a> OpVisitor for Disassembler<'a> {
                 write!(&mut self.disas, "{space}{byte:02x}").unwrap();
                 need_space = true;
             }
-            for _ in 0..11_usize.saturating_sub(size) {
+            for _ in 0..12_usize.saturating_sub(size) {
```
alexcrichton, line 277: This was done to make the disassembly a bit prettier since otherwise the hexdump and instructions might be side-by-side (no whitespace between)

abrown (approved):

PR #9722, opened 2024-12-03
pulley: Remove virtual sp offset

alexcrichton:
This fixes the output of the pulley cranelift backend with respect to virtual stack offsets and adjustments for calls with stack parameters and results. Previously the virtual stack pointer was updated but the rest of the backend assumed the real stack pointer was updated. This updates the pulley backend to match the other backends in this respects and fixes a test on my "more complete pulley integration" branch.

### DIFF
--- a/cranelift/codegen/src/isa/pulley_shared/abi.rs
+++ b/cranelift/codegen/src/isa/pulley_shared/abi.rs
@@ -254,6 +254,10 @@ where
     }
 
     fn gen_sp_reg_adjust(amount: i32) -> SmallInstVec<Self::I> {
+        if amount == 0 {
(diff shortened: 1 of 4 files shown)


alexcrichton:
I'll fully admit I don't know what a virtual sp offset is so I could be completely off-base here. The main data I'm going on is "fixes a bug in another branch".

cfallin (approved):
LGTM! Virtual-SP was a mechanism we used to handle frame layout relative to sp while sp moved during a callsite sequence; it went away in #8631, so probably it's a vestigial artifact here from Pulley's early development history (?). In any case, LGTM.

PR #9724, opened 2024-12-03
Fix `wasmtime::FieldType::matches` for subtyping and mutability

fitzgen:
Similar to https://github.com/bytecodealliance/wasm-tools/pull/1934 but for `wasmtime::FieldType`.

### DIFF
--- a/crates/wasmtime/src/runtime/types.rs
+++ b/crates/wasmtime/src/runtime/types.rs
@@ -1362,7 +1362,14 @@ impl StorageType {
     /// Panics if either type is associated with a different engine from the
     /// other.
     pub fn eq(a: &Self, b: &Self) -> bool {
-        a.matches(b) && b.matches(a)
+        match (a, b) {
+            (StorageType::I8, StorageType::I8) => true,
+            (StorageType::I8, _) => false,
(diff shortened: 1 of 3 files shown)


crates/wasmtime/src/runtime/types.rs
```diff
@@ -1676,13 +1696,9 @@ impl StructType {
     pub fn matches(&self, other: &StructType) -> bool {
         assert!(self.comes_from_same_engine(other.engine()));
 
-        // Avoid matching on structure for subtyping checks when we have
-        // precisely the same type.
-        if self.type_index() == other.type_index() {
-            return true;
-        }
```
alexcrichton, line 1683: This is still a valid optimization to do though, right? IIRC `signatures()` below grabs a lock so ideally we'd keep this.

alexcrichton (approved):

crates/wasmtime/src/runtime/types.rs
fitzgen: `signatures()` doesn't grab a lock, the methods on it do, and the `is_subtype` method already has this optimization as a fast path, so no need to duplicate it all over the place.

PR #9735, opened 2024-12-04
pulley: Track faulting opcode in stack overflow better

alexcrichton:
This commit updates the pulley `Encode` trait to have a `WIDTH` associated with it to be able to calculate the size of an instruction by name rather than hard-coding instruction details in multiple locations. This constant is procedurally generated per-instruction given the definition of the instruction. This for now assumes there are no variable-width instructions.

### DIFF
--- a/pulley/src/op.rs
+++ b/pulley/src/op.rs
@@ -1,5 +1,6 @@
 //! Pulley bytecode operations with their operands.
 
+use crate::encode::Encode;
 use crate::imms::*;
 use crate::regs::*;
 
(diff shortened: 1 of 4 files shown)


pulley/src/op.rs
```diff
@@ -128,6 +130,17 @@ macro_rules! define_op_encode {
...
             }
+
+            /// Returns the encoded size of this op.
+            #[cfg(feature = "encode")]
+            pub fn width(&self) -> u8 {
+                match self {
+                    $(
+                        Self::$name(_) => <$name as Encode>::WIDTH,
+                    )*
+                    Self::ExtendedOp(op) => op.width(),
```
fitzgen, line 141: Does this need to add one for the extended op prefix byte?

fitzgen (approved):

PR #9736, opened 2024-12-04
Remove the `FuncEnvironment` trait from `wasmtime-cranelift`

alexcrichton:
This is a leftover from the `cranelift-wasm` days and it's not necessary any more. Remove it and have an inherent impl for the trait methods instead.


fitzgen (approved):
Very nice! I wanted to get to this eventually, happy to have been beaten in the race

PR #9737, opened 2024-12-04
winch: Epoch-based interruption

saulecabrera:
Closes https://github.com/bytecodealliance/wasmtime/issues/8091

This commit introduces support for epoch interruption to Winch. The heuristics around epoch check emission are identical to Cranelift's except for the fact that the current implementation doesn't introduce a local-based cache for the current epoch deadline. This is an intentional decision given Winch's focus on compilation performance. However, if needed in the future, knobs could be introduced to optionally introduce a local cache at the cost of reduced compilation performance.

### DIFF
--- a/winch/codegen/src/codegen/mod.rs
+++ b/winch/codegen/src/codegen/mod.rs
@@ -276,9 +276,7 @@ where
             self.emit_fuel_check();
         }
 
-        if self.tunables.epoch_interruption {
-            self.emit_epoch_check();
-        }
+        self.maybe_emit_epoch_check();
 
         // Once we have emitted the epilogue and reserved stack space for the locals, we push the
         // base control flow block.
@@ -1028,9 +1026,15 @@ where
         );
     }
 
-    /// Emit series of instructions that check the current epoch against its
-    /// deadline.
-    pub fn emit_epoch_check(&mut self) {
+    /// Checks if epoch interruption is configured and emits a series of
+    /// instructions that check the current epoch against its deadline.
+    pub fn maybe_emit_epoch_check(&mut self) {
+        if !self.tunables.epoch_interruption {
+            return;
+        }
+
+        // The continuation branch if the current epoch hasn't reached the
+        // configured deadline.
         let cont = self.masm.get_label();
         let new_epoch = self.env.builtins.new_epoch::<M::ABI, M::Ptr>();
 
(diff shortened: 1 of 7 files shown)


winch/codegen/src/visitor.rs
```diff
@@ -1717,6 +1717,11 @@ where
         if self.tunables.consume_fuel {
             self.emit_fuel_check();
         }
+
+        // Emit epoch check right after binding the loop header.
+        if self.tunables.epoch_interruption {
+            self.emit_epoch_check();
+        }
```
fitzgen, line 1724: Maybe move this `if self.tunables.epoch_interruption` into the `emit_epoch_check` helper? (And rename it to `maybe_emit_epoch_check` or something like that? Feel free to bikeshed.)

This code motion would simplify callers, which are all doing the exact same conditional.

winch/codegen/src/codegen/mod.rs
```diff
@@ -1022,6 +1026,93 @@ where
...
+        let new_epoch = self.env.builtins.new_epoch::<M::ABI, M::Ptr>();
+
+        // Checks for runtime limits (e.g., fuel, epoch) are special since they
+        // require inserting abritrary function calls and control flow.
+        // Special care must be taken to ensure that all invariants are met. In
+        // this case, since `new_epoch` takes an argument and returns a value,
+        // we must ensure that any registers used to hold the current epoch
+        // value and deadline are not going to be needed later on by the
+        // function call.
+        let (epoch_deadline_var, epoch_counter_var) = self.context.without::<(Reg, Reg), M, _>(
```
fitzgen, line 1042: Is the `_var` suffix fairly established in Winch? If so, ignore the following.

If not, then I find it a little surprising. I thought maybe there was a `Var` type that abstracts over registers and stack slots or something. Or introduces a virtual register concept or something like that. I think a `_reg` suffix would have avoided me going down that line of thought and inquiry, which would make it slightly easier to follow what's going on, as a code reader.

winch/codegen/src/codegen/mod.rs
```diff
@@ -1022,6 +1026,93 @@ where
...
+        self.masm.bind(cont);
+
+        self.context.free_reg(epoch_deadline_var);
+        self.context.free_reg(epoch_counter_var);
+    }
+
+    fn emit_load_epoch_deadline_and_counter(
+        &mut self,
+        epoch_deadline_var: Reg,
+        epoch_counter_var: Reg,
```
fitzgen, line 1081: And then I'd expect these to also be `_reg` if the last suggestion is taken.

winch/codegen/src/codegen/mod.rs
```diff
@@ -1022,6 +1026,93 @@ where
         );
     }
 
+    /// Emit series of instructions that check the current epoch against its
+    /// deadline.
+    pub fn emit_epoch_check(&mut self) {
+        let cont = self.masm.get_label();
```
fitzgen, line 1032: I think a one-line comment describing what this label is could also be helpful to code readers.

fitzgen (approved):
Looks great! Just a couple nitpicks inline.

Also I know that most of the testing will be covered by fuzzing and the existing epoch tests, but it may make sense to throw in a single disas test with epochs enabled as well that contains a some `if`s, `block`s, and a `loop` just to show that we insert epoch checks only on the loop headers and body prologue. Could also be used to see that deadlines are actually cached, if we ever get around to doing that here.

saulecabrera:
@fitzgen agreed with all the suggestions,thanks! I've also opened https://github.com/bytecodealliance/wasmtime/pull/9737, which applies the suggestions in this PR to fuel checks.

PR #9739, opened 2024-12-05
 Shuffle fields in `VMRuntimeLimits`

alexcrichton:
Right now this structure has a pointer-sized field, two 64-bit integers,
and then three pointer-sized fields. This structure's layout is
nontrivial to calculate on 32-bit platforms as it needs to take the
alignment of 64-bit integers into account which differs between ARM and
x86 for example. To make this easier shuffle all the 64-bit integers are now
first, which means we don't have to worry about alignment.

I'll note that this particular ordering is still a bit brittle because
we might need to shuffle things again if more fields are added. That
being said any misalignment is caught during testing of the `wasmtime`
crate so there's not much danger in adding more things, it'll just
require updating a few more locations.


alexcrichton:
I'll note that the diff here looks scary but it's practically pretty small. The first commit is the actual change and the second commit is purely test expectation updates. This is because the offset of the stack-limit field has changed, and functions all have an explicit check of the stack limit.

fitzgen (approved):

PR #9740, opened 2024-12-05
 Tweak rules for traps in CLIF translation 

alexcrichton:
This commit updates how the `signals_based_traps` setting is read for
various instruction generation during the wasm-to-CLIF translation
phase. Specifically this splits the `signals_based_traps` configuration
into one for "general CLIF instructions can trap" and another for
"memory traps are allowed". These settings are based on
`signals_based_traps` but then differ for Pulley. For example Pulley
allows general instructions to trap since that just halts the
interpreter, but memory-related instructions can't trap since segfaults
aren't caught during the execution of Pulley. These settings then take
Pulley into account to assist in translating Pulley compatibly by
default.


alexcrichton:
I'll note that this is temporarily based on https://github.com/bytecodealliance/wasmtime/pull/9736, so only the second commit is part of this PR

tests/disas/pulley/loads.wat
fitzgen, line 1: Was this test file intentionally deleted?

fitzgen (approved):

tests/disas/pulley/loads.wat
alexcrichton: Ah yes sorry forgot to call this out, the changes in this PR change codegen for pulley where it uses `uadd_overflow_trap` now instead of add/compare which isn't supported on Pulley yet, so this test no longer passes.

PR #9743, opened 2024-12-05
Enumerate all host calls in `wasmtime_environ::HostCall`

alexcrichton:
This commit is a continuation of the plan of implementing host calls in Pulley through #9665, #9675, and #9693. Here the `Compiler::call_indirect_host` method is updated to take a new type, `HostCall`, which indicates what type of host call is being performed. This is then serialized to a 32-bit integer which will be present in the pulley instruction being generated. This 32-bit integer will then be used to perform a dispatch (the dispatch is left for a future PR with more Pulley integration).

This new `HostCall` structure is defined with `BuiltinFunctionIndex` internally. Additionally a new `ComponentBuiltinFunctionIndex` is added to enumerate the same set of indexes for components as well. Along the way the split between component transcoders/builtins were removed and they're now all lumped together in one macro for builtins. (no need to have two separate macros).

This new `HostCall` is used to implement the `call_indirect_host` instruction for Pulley to fill out an unimplemented piece of code.


fitzgen:
> Along the way the split between component transcoders/builtins were removed and they're now all lumped together in one macro for builtins. (no need to have two separate macros).

Yessssss

crates/cranelift/src/func_environ.rs
```diff
@@ -33,15 +33,14 @@ use wasmtime_environ::{FUNCREF_INIT_BIT, FUNCREF_MASK};
...
-        [Option<ir::FuncRef>; BuiltinFunctionIndex::builtin_functions_total_number() as usize],
+    builtins: [Option<ir::FuncRef>; BuiltinFunctionIndex::max() as usize],
 }
 
 impl BuiltinFunctions {
     fn new(compiler: &Compiler) -> Self {
         Self {
             types: BuiltinFunctionSignatures::new(compiler),
-            builtins: [None; BuiltinFunctionIndex::builtin_functions_total_number() as usize],
+            builtins: [None; BuiltinFunctionIndex::max() as usize],
```
fitzgen, line 43: Nitpick: If `max` returns the maximum index then this should be `max + 1`. If `max` returns the number of builtins then we should rename it to `len` or `count` or something.

Ditto just above.

crates/environ/src/builtin.rs
```diff
@@ -259,7 +268,7 @@ macro_rules! declare_indexes {
         $len:expr;
     ) => {
         /// Returns the total number of builtin functions.
-        pub const fn builtin_functions_total_number() -> u32 {
+        pub const fn max() -> u32 {
```
fitzgen, line 271: Yeah let's call this `count` or `len` or something instead of `max`.

fitzgen (commented):
LGTM modulo a nitpick around naming

fitzgen (approved):

PR #9744, opened 2024-12-05
 Support executing Pulley in Wasmtime 

alexcrichton:
This commit is the initial implementation of executing the Pulley
interpreter from the `wasmtime` crate. This gives access to all of the
`wasmtime` crate's runtime APIs backed by execution of bytecode in
Pulley. This builds on the previous PRs I've been making for support in
Pulley to culminate in testing on CI in this PR. This PR handles some
final tidbits related to producing a runnable image that can be
interpreted by the `wasmtime` crate such as:

* Pulley compilation artifacts are no longer marked as natively
  executable, just read-only.
* Pulley compilation artifacts include wasm-to-array trampolines like
  normal platforms (removes a pulley special-case).
* Dispatch of host calls from Pulley to the Wasmtime runtime are
  implemented.
* Pulley's list of panicking wasm features is slimmed down as most are
  covered by "this lowering isn't supported" errors.
* Execution of wasm code now has an `if` to see whether Pulley is
  enabled within a `Store` or not.
* Traps and basic "unwinding" of the pulley stack are implemented (e.g.
  a "pulley version" of `setjmp` and `longjmp`, sort of)
* Halting the interpreter has been refactored to help shrink the size of
  `ControlFlow<Done>` and handle metadata with each done state.

Some minor refactorings are also included here and there along with a
few fixes here and there necessary to get tests passing.

The next major part of this commit is updates to our `wast` test suite
and executing all `*.wast` files. Pulley is now executed by default for
all files as a new execution engine. This means that all platforms in CI
will start executing Pulley tests. At this time almost all tests are
flagged as "expected to fail" but there are a small handful of
allow-listed tests which are expected to pass. This exact list will
change over time as CLIF lowerings are implemented and the interpreter
is extended.

Follow-up PRs will extend the testing strategy further such as:

* Extending `#[wasmtime_test]` to include Pulley in addition to Winch.
* Getting testing set up on CI for 32-bit platforms.


alexcrichton:
I'll note that this is in draft form as it currently is based on https://github.com/bytecodealliance/wasmtime/pull/9743. I'm going ahead and opening this up to see what CI has to say about this.

alexcrichton:
@fitzgen ok should be rebased and updated, I believe s390x CI is failing but I think most of the rest should be passing. Once results come in I'll figure out how to best ignore s390x for now.

crates/wasmtime/src/runtime/vm/interpreter.rs
```diff
@@ -0,0 +1,324 @@
...
+            TrapTest::HandledByEmbedder => unreachable!(),
+
+            // Trap was handled, yay! We don't use `jmp_buf`.
+            TrapTest::Trap { jmp_buf: _ } => {}
+        }
+
+        // Perform a "longjmp" by restoring the "setjmp" context saved when this
+        // started.
+        //
+        // FIXME: this is not restoring callee-save state. For example if
```
alexcrichton, line 148: I've added this to https://github.com/bytecodealliance/wasmtime/issues/9747

crates/wasmtime/src/runtime/vm/vmcontext.rs
```diff
@@ -720,8 +720,35 @@ impl VMFuncRef {
     /// # Unsafety
     ///
     /// This method is unsafe because it can be called with any pointers. They
-    /// must all be valid for this wasm function call to proceed.
+    /// must all be valid for this wasm function call to proceed. Fore xample
```
fitzgen, line 723:
```suggestion
    /// must all be valid for this wasm function call to proceed. For example
```

crates/wast-util/src/lib.rs
```diff
@@ -229,18 +229,52 @@ pub struct WastConfig {
...
+    /// primarily supports x86_64.
     Winch,
+
+    /// Pulley interpreter.
+    ///
+    /// This tests the Cranelift pulley backend plus the pulley execution
+    /// environment of the output bytecode. Note that this is separate from
+    /// `Cranelift` above to be able to test both on platforms where Cranelift
+    /// has native codegen support.
+    Pulley,
```
fitzgen, line 255: Possible bikeshed: Maybe rename `Cranelift` to `CraneliftNative` and `Pulley` to `CraneliftPulley`? In addition to clarifying the native code execution bits, this would future proof for a possible `WinchPulley` in the future.

crates/wast-util/src/lib.rs
```diff
@@ -252,10 +286,72 @@ impl Compiler {
...
+                    || threads.unwrap_or(false)
+                    || gc.unwrap_or(false)
+                    || function_references.unwrap_or(false)
+                    || relaxed_simd.unwrap_or(false)
+                    || reference_types.unwrap_or(false)
+                    || tail_call.unwrap_or(false)
+                    || extended_const.unwrap_or(false)
+                    || wide_arithmetic.unwrap_or(false)
+                    || simd.unwrap_or(false)
+                    || gc_types.unwrap_or(false)
```
fitzgen, line 325: Maybe we should

```rust
impl TestConfit {
    fn custom_page_sizes(&self) -> bool {
        self.custom_page_sizes.unwrap_or(false)
    }

    // etc...
}
```

if we are doing this dance every time we access these fields

pulley/src/interp.rs
```diff
@@ -77,7 +76,21 @@ impl Vm {
...
+            DoneReason::Trap(pc) => DoneReason::Trap(pc),
+            DoneReason::CallIndirectHost { sig, resume } => {
+                DoneReason::CallIndirectHost { sig, resume }
+            }
+        }
+    }
+
+    /// Peforms the initial part of [`Vm::call`] in setting up the `args`
+    /// provided in registers according to Pulley's ABI.
+    pub unsafe fn call_start<'a>(&'a mut self, args: &[Val]) {
```
fitzgen, line 93:
```suggestion
    /// provided in registers according to Pulley's ABI.
    ///
    /// # Unsafety
    ///
    /// All the same unsafety as `call` and additiionally, you must
    /// invoke `call_run` and then `call_end` after calling `call_start`.
    /// If you don't want to wrangle these invocations, use `call` instead
    /// of `call_{start,run,end}`.
    pub unsafe fn call_start<'a>(&'a mut self, args: &[Val]) {
```

pulley/src/interp.rs
```diff
@@ -101,14 +114,34 @@ impl Vm {
                 },
             }
         }
+    }
 
-        self.run(func)?;
+    /// Peforms the internal part of [`Vm::call`] where bytecode is actually
+    /// executed.
+    pub unsafe fn call_run(&mut self, pc: NonNull<u8>) -> DoneReason<()> {
```
fitzgen, line 121:
```suggestion
    /// executed.
    ///
    /// # Unsafety
    ///
    /// In addition to all the invariants documented for `call`, you
    /// may only invoke `call_run` after invoking `call_start` to
    /// initialize this call's arguments.
    pub unsafe fn call_run(&mut self, pc: NonNull<u8>) -> DoneReason<()> {
```

pulley/src/interp.rs
```diff
@@ -101,14 +114,34 @@ impl Vm {
...
+            state: &mut self.state,
+            pc: UnsafeBytecodeStream::new(pc),
+        };
+        let done = interpreter.run();
+        self.state.done_decode(done)
+    }
+
+    /// Peforms the tail end of [`Vm::call`] by returning the values as
+    /// determined by `rets` according to Pulley's ABI.
+    pub unsafe fn call_end<'a>(
```
fitzgen, line 133:
```suggestion
    /// Peforms the tail end of [`Vm::call`] by returning the values as
    /// determined by `rets` according to Pulley's ABI.
    ///
    /// # Unsafety
    ///
    /// In addition to the invariants documented for `call`, this may
    /// only be called after `call_run`.
    pub unsafe fn call_end<'a>(
```

pulley/src/interp.rs
```diff
@@ -574,20 +588,72 @@ impl MachineState {
...
+    }
+
+    /// Reason that the pulley interpreter has ceased execution.
+    pub enum DoneReason<T> {
+        /// A trap happened at this bytecode instruction.
+        Trap(NonNull<u8>),
+        /// The `call_indirect_host` instruction was executed.
+        CallIndirectHost {
+            /// The payload of `call_indirect_host`.
+            sig: u8,
```
fitzgen, line 612:
```suggestion
            id: u8,
```

fitzgen (approved):
Very nice! Super excited to see this progress!

A few nitpicks and suggestions below, but nothing major

cranelift/codegen/src/isa/pulley_shared/lower.isle
```diff
@@ -77,14 +77,14 @@
 
 ;; Fold `(trap[n]z (icmp ...))` together.
 
-(rule 1 (lower (trapz (icmp cc a b @ (value_type (fits_in_64 ty))) code))
+(rule 1 (lower (trapz (icmp cc a b @ (value_type (ty_32_or_64 ty))) code))
```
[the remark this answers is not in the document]
alexcrichton: Ah this is due to some other rule only accepting 32 or 64-bit inputs which causes ISLE to panic as something not a partial constructor. Basically tests panic unless this is here, so a future refactoring will need to flesh out more lowering rules for 8/16-bit types.

crates/wasmtime/src/runtime/vm/interpreter_disabled.rs
```diff
@@ -0,0 +1,50 @@
...
+use core::ptr::NonNull;
+
+pub struct Interpreter {
+    empty: Void,
+}
+
+const _: () = assert!(mem::size_of::<Interpreter>() == 0);
+const _: () = assert!(mem::size_of::<Option<Interpreter>>() == 0);
+
+enum Void {}
```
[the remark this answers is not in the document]
alexcrichton: Thanks I always forget what we have, where it's defined, and what it's called heh.

PR #9745, opened 2024-12-05
Test Pulley on 32-bit platforms in CI

alexcrichton:
This PR is based on https://github.com/bytecodealliance/wasmtime/pull/9743 and https://github.com/bytecodealliance/wasmtime/pull/9744 and those should be reviewed first.

Opening this to get a full run of CI to see what happens on these platforms...

### DIFF
--- a/ci/build-test-matrix.js
+++ b/ci/build-test-matrix.js
@@ -36,7 +36,7 @@ function supports32Bit(pkg) {
   if (pkg.indexOf("pulley") !== -1)
     return true;
 
-  return pkg == 'wasmtime-fiber';
+  return pkg == 'wasmtime-fiber' || pkg == 'wasmtime';
 }
 
 // This is the full, unsharded, and unfiltered matrix of what we test on
@@ -231,7 +231,9 @@ async function shard(configs) {
   for (const config of configs) {
     // Special case 32-bit configs. Only some crates, according to
     // `supports32Bit`, run on this target. At this time the set of supported
-    // crates is small enough that they're not sharded.
+    // crates is small enough that they're not sharded. A second shard, however,
+    // is included which runs `--test wast` to run the full `*.wast` test suite
+    // in CI on 32-bit platforms, at this time effectively testing Pulley.
     if (config["32-bit"] === true) {
       sharded.push(Object.assign(
         {},
@@ -241,6 +243,11 @@ async function shard(configs) {
             .map(c => supports32Bit(c) ? `--package ${c}` : `--exclude ${c}`)
             .join(" "),
         }
+      ));
+      sharded.push(Object.assign(
+        {},
+        config,
+        { bucket: '--test wast' },
       ));
       continue;
     }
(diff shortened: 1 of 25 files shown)


ci/build-test-matrix.js
```diff
@@ -242,6 +244,11 @@ async function shard(configs) {
             .join(" "),
         }
       ));
+      sharded.push(Object.assign(
+        {},
+        config,
+        { bucket: '--test wast' },
+      ));
```
fitzgen, line 251: Is this not going to break for the crates which support 32-bit but don't have a `tests/wast.rs`? I.e. all of them other than `wasmtime`? But even the `wasmtime` crate doesn't have that test, only `wasmtime-cli` does... I guess I am pretty confused around the logic here...

crates/wasmtime/src/runtime/values.rs
```diff
@@ -1021,26 +1021,14 @@ mod tests {
...
-        );
+        assert!(std::mem::size_of::<Val>() <= 24);
     }
 
     #[test]
     fn size_of_ref() {
         // Try to keep tabs on the size of `Ref` and make sure we don't grow its
         // size.
-        assert_eq!(std::mem::size_of::<Ref>(), 24);
+        assert!(std::mem::size_of::<Ref>() <= 24);
```
fitzgen, line 1031: Can we make these precise for 64- and 32-bit archs? Because otherwise we might shrink it, forget to update these tests, and then regress the size again back to its current size.

crates/wasmtime/src/runtime/vm/sys/unix/unwind.rs
```diff
@@ -9,12 +9,29 @@ pub struct UnwindRegistration {
...
+        }
+    } else {
+        extern "C" {
+            // libunwind import
+            fn __register_frame(fde: *const u8);
+            fn __deregister_frame(fde: *const u8);
+            #[wasmtime_versioned_export_macros::versioned_link]
+            fn wasmtime_using_libunwind() -> bool;
+        }
+    }
```
fitzgen, line 34: @Kmeakin, happen to know how we can properly get access to these JIT code registration functions so that profiling and debugging and all that with the system tools is a better experience?

fitzgen (approved):

ci/build-test-matrix.js
alexcrichton: This is abusing the fact that `cargo test --workspace --test wast` only actually executes tests named `wast` and everything else is ignored. So this effectively runs `cargo test --test wast` and nothing else is named `wast`. 

In the long-run I want to get more 32-bit tests running, just needs more support in Pulley for that.

crates/wasmtime/src/runtime/values.rs
alexcrichton: Good point yeah, I'll try. It's not quite as easy as 32-or-64 since the size differs between x86 and arm due to the alignment of u64 being different, but I'll probably just make these tests 64-bit only.

PR #9751, opened 2024-12-06
winch: Implement aarch64 call, trapz, address_at_vmctx

MarinPostma:
Implement the following MASM instruction for the aarch64 platform:
- call
- address_at_vmctx
- trapz

I initially planned to only implement call, but implementing the tests uncovered the necessity to implement the other two as prerequisite for indirect calls.

Right now the calling convention is hardcoded to SystemV, but maybe we also want to support AppleAarch64?

#8321

### DIFF
--- a/tests/disas/winch/aarch64/call/simple.wat
+++ b/tests/disas/winch/aarch64/call/simple.wat
@@ -19,7 +19,7 @@
 ;;       stp     x29, x30, [sp, #-0x10]!
 ;;       mov     x29, sp
 ;;       ldur    x16, [x2, #8]
-;;       ldur    x16, [x16]
+;;       ldur    x16, [x16, #0x10]
 ;;       add     x16, x16, #0x10
 ;;       cmp     sp, x16
 ;;       b.lo    #0x38
--- a/tests/disas/winch/aarch64/call/simple.wat
+++ b/tests/disas/winch/aarch64/call/simple.wat
@@ -1,5 +1,5 @@
 ;;! target = "aarch64"
-;;! test = "compile"
+;;! test = "winch"
 
 (module
   (func $main (result i32)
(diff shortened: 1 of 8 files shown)


saulecabrera:
I can help taking a look at this one.

winch/codegen/src/isa/aarch64/asm.rs
```diff
@@ -901,4 +908,32 @@ impl Assembler {
...
+                CallConv::SystemV,
+            )),
+        })
+    }
+
+    pub fn call_with_reg(&mut self, callee: Reg) {
+        self.emit(Inst::CallInd {
+            info: Box::new(cranelift_codegen::CallInfo::empty(
+                callee.into(),
+                CallConv::SystemV,
```
saulecabrera, line 925: > Right now the calling convention is hardcoded to SystemV, but maybe we also want to support AppleAarch64?

I see why this might be confusing. Even though the `CallConv` is hard coded here, there are two things to note:

* All the infrastructure to handle the calling convention differences, is handled very early on in the compilation process in Winch (and in Cranelift as well) e.g., when constructing the signature for the callee, at that point,[ you'll note that Winch correctly handles the default and apple-aarch64 calling conventions.](https://github.com/bytecodealliance/wasmtime/blob/main/winch/codegen/src/isa/aarch64/abi.rs#L33)
* Winch relies on Cranelift's binary emission infrastructure, and in this concrete case, as you can note, you're constructing a `CallInfo::empty()`, which signals that none of the other Cranelift-specific invariants must be taken into account when emitting calls (some of them depend on the calling convention that is passed here). 

I've opened https://github.com/bytecodealliance/wasmtime/pull/9757 to refactor call emission a bit with the hope of making this part more consistent.

MarinPostma: That refactor makes a lot of sense to me 👍

winch/codegen/src/isa/aarch64/asm.rs
```diff
@@ -724,6 +723,14 @@ impl Assembler {
         });
     }
 
+    /// Trap if rn is 0
```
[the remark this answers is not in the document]
MarinPostma: fixed in [f1a56f1](https://github.com/bytecodealliance/wasmtime/pull/9751/commits/f1a56f1685decf816cfd001c655c3b1ca4dd44f4)

winch/codegen/src/isa/aarch64/asm.rs
```diff
@@ -901,4 +908,32 @@ impl Assembler {
...
+
+    pub fn call_with_lib(&mut self, lib: LibCall, dst: Reg) {
+        let name = ExternalName::LibCall(lib);
+        self.emit(Inst::LoadExtName {
+            rd: writable!(dst.into()),
+            name: name.into(),
+            offset: 0,
+        });
+        self.call_with_reg(dst)
+    }
```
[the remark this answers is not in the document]
MarinPostma: fixed in [f1a56f1](https://github.com/bytecodealliance/wasmtime/pull/9751/commits/f1a56f1685decf816cfd001c655c3b1ca4dd44f4)

winch/codegen/src/isa/aarch64/asm.rs
MarinPostma: updated in f4a6cf7

saulecabrera:
There are some failures in CI -- I'm trying to figure out if they are related to your changes or not.

tests/disas/winch/aarch64/call/simple.wat
```diff
@@ -0,0 +1,40 @@
+;;! target = "aarch64"
+;;! test = "compile"
```
saulecabrera, line 2: I'm not entirely sure if this is the source of the failure, but here you need to specify 

```suggestion
;;! test = "winch"
```

Else this will use Cranelift

MarinPostma: Some submodule update sneaked into one of my commits. That's what's messing with the CI, I'll fix that tommorow

MarinPostma:
Alright, I updated the submodule, ci looks good now

saulecabrera (approved):
Thanks!

PR #9755, opened 2024-12-06
Use `runner.os` in CI configuration more

alexcrichton:
Instead of using `matrix.os` which can change over time as images change instead use `runner.os` to both name jobs and have conditions on jobs. This couples it more closely to what's intended, the choice of OS, rather than image version.


alexcrichton:
Ok so surprise to me but apparently you can't use `runner.os` in the name of a job

fitzgen (approved):

PR #9756, opened 2024-12-06
cpp demangling check for demangling error

### LINKED ISSUE #9753: use of cpp_demangle misses check if demangling will actually succeed
Probably related: #4162

I came across this issue while trying to use _perf_ with wasmtime [following the documentation](https://docs.wasmtime.dev/examples-profiling-perf.html).
When I used the param `--profile=perfmap` or `--profile=jitdump`, I got the following error message:
```
thread 'main' panicked at crates/wasmtime/src/runtime/instantiate.rs:73:76:
called `Result::unwrap()` on an `Err` value: Error
```

I tried using a simple hello-world example wasm file and found that the issue was not present there.
After some trying around, I think, I might have found the cause.

The wasm file I tried to use with perf initially has a symbol named "RC4". When the _demangle_function_name_ function is called in [instantiate.rs:73](https://github.com/bytecodealliance/wasmtime/blob/66989d9d6cfabb4c0781134b8d78bcbd81c6e511/crates/wasmtime/src/runtime/instantiate.rs#L73) unwrapping fails, because the [if-let statement](https://github.com/bytecodealliance/wasmtime/blob/66989d9d6cfabb4c0781134b8d78bcbd81c6e511/crates/environ/src/demangling.rs#L10) that passes the function name to cpp_demangle crate only creates the [Symbol](https://docs.rs/cpp_demangle/latest/cpp_demangle/struct.Symbol.html), which apparently is not safe to demangle actually. Even if its Result is Ok.
I'm not quite sure if this is their intended behavior. What do you think?

This gist shows the isolation of the problem: https://gist.github.com/chobermaier/3adf8f2a004ccbe17865bfc20df5c6d3

I would propose to change _demangle_function_name_ this way ([commit with diff](
https://github.com/chobermaier/wasmtime/commit/2852fc633a65f97661fbab2033fc7cedcb010cf2)):

```rs
pub fn demangle_function_name(writer: &mut impl core::fmt::Write, name: &str) -> core::fmt::Result {
    #[cfg(feature = "demangle")]
    if let Ok(demangled) = rustc_demangle::try_demangle(name) {
        return write!(writer, "{demangled}");
    } else if let Ok(symbol) = cpp_demangle::Symbol::new(name) {
        let options = cpp_demangle::DemangleOptions::default();
        if let Ok(demangled) = symbol.demangle(&options) {
            return write!(writer, "{demangled}");
        }
    }

    write!(writer, "{name}")
}
```
This would call [demangle](https://docs.rs/cpp_demangle/latest/cpp_demangle/struct.Symbol.html#method.demangle) explicitly and provide the information if it's possible through the Result returned. Instead of the implicit call through the write macro. So this should not introduce any performance overhead.

### Versions and Environment

Wasmtime version or commit: `wasmtime 27.0.0 (8eefa236f 2024-11-20)`

Operating system: Linux 6.12.1-arch1-1

Architecture: amd64

chobermaier:
The use of cpp_demangle in `crates/environ/src/demangling.rs` missed a check if demangling would actually succeed.

This PR's wraps the use of cpp_demangle with an explicitly call of [demangle](https://docs.rs/cpp_demangle/latest/cpp_demangle/struct.Symbol.html#method.demangle) and can therefore check its Result.  
The current implicit demangle call did not check this result. In some cases this lead to failed unwrapping in [instantiate.rs](https://github.com/bytecodealliance/wasmtime/blob/66989d9d6cfabb4c0781134b8d78bcbd81c6e511/crates/wasmtime/src/runtime/instantiate.rs#L73).

Closes #9753

### DIFF
--- a/crates/environ/src/demangling.rs
+++ b/crates/environ/src/demangling.rs
@@ -7,8 +7,11 @@ pub fn demangle_function_name(writer: &mut impl core::fmt::Write, name: &str) ->
     #[cfg(feature = "demangle")]
     if let Ok(demangled) = rustc_demangle::try_demangle(name) {
         return write!(writer, "{demangled}");
-    } else if let Ok(demangled) = cpp_demangle::Symbol::new(name) {
-        return write!(writer, "{demangled}");
+    } else if let Ok(symbol) = cpp_demangle::Symbol::new(name) {
+        let options = cpp_demangle::DemangleOptions::default();
+        if let Ok(demangled) = symbol.demangle(&options) {
+            return write!(writer, "{demangled}");
+        }
     }
 
     write!(writer, "{name}")


alexcrichton (approved):
Thanks for this!

PR #9759, opened 2024-12-06
pulley: Support big-endian targets

alexcrichton:
This commit fixes the Pulley interpreter and Cranelift backend to properly support big-endian targets. The problem beforehand was that loads/stores in Pulley were defined as the native endianness which means that big and little-endian platforms differed. Additionally the previously set of understood pulley targets were implicitly always little-endian which was not appropriate for a big-endian host where JIT data structures are stored in big-endian format.

This commit fixes all of these issues with a few changes:

* Pulley loads/stores are always little-endian now.
* Pulley now has bswap32/bswap64 instructions.
* Wasmtime/Cranelift now understand big-endian pulley targets (e.g. `pulley32be`).
* CLIF translation of loads/stores now properly handles the endianness flags on `MemFlags`.

In the future if necessary we could natively add a macro-op for big-endian loads/stores to Pulley but for now it's more minimal to just have `bswap32` and `bswap64`.

The result of this commit is that Pulley tests are now run and executed on s390x like all other platforms.

### DIFF
--- a/cranelift/codegen/src/isa/pulley_shared/lower.isle
+++ b/cranelift/codegen/src/isa/pulley_shared/lower.isle
@@ -235,11 +235,9 @@
 
 ;;;; Rules for `store` ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
-(decl littleendian () MemFlags)
-(extern extractor littleendian littleendian)
-
-(decl bigendian () MemFlags)
-(extern extractor bigendian bigendian)
+(type Endianness extern (enum Big Little))
+(decl pure endianness (MemFlags) Endianness)
+(extern constructor endianness endianness)
 
 (rule (lower (store flags src @ (value_type ty) addr (offset32 offset)))
       (side_effect (pulley_store (Amode.RegOffset addr (i32_as_i64 offset))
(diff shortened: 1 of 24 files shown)


alexcrichton:
While this doesn't code-wise depend on https://github.com/bytecodealliance/wasmtime/pull/9758 it test-wise depends on that PR as otherwise unwinding doesn't actually work on s390x. With both PRs, however, tests pass for Pulley.

cranelift/codegen/src/isa/pulley_shared/lower.isle
```diff
@@ -226,19 +226,32 @@
...
+      (xswap_if_be le ty flags)))
+
 
 ;;;; Rules for `store` ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
+(decl littleendian () MemFlags)
+(extern extractor littleendian littleendian)
+
+(decl bigendian () MemFlags)
+(extern extractor bigendian bigendian)
```
fitzgen, line 242: Can we instead have a pure constructor like

```lisp
(type Endian (enum Little Big))

(decl endian (MemFlags) Endian)
(extern pure constructor endian endian)
```

cranelift/codegen/src/isa/pulley_shared/lower.isle
```diff
@@ -226,19 +226,32 @@
...
       (side_effect (pulley_store (Amode.RegOffset addr (i32_as_i64 offset))
-                                 src
+                                 (xswap_if_be src ty flags)
                                  ty
                                  flags)))
 
+(decl xswap_if_be (XReg Type MemFlags) XReg)
+(rule 0 (xswap_if_be val _ty (littleendian)) val)
+(rule 1 (xswap_if_be val $I32 (bigendian)) (pulley_bswap32 val))
+(rule 1 (xswap_if_be val $I64 (bigendian)) (pulley_bswap64 val))
```
fitzgen, line 253: ... which we can then use like

```suggestion
(decl xswap_if_be (XReg Type MemFlags) XReg)
(rule 0 (xswap_if_be val _ty flags)
      (if-let (endian flags) (Endian.Little))       
      val)
(rule 1 (xswap_if_be val $I32 flags)
      (if-let (endian flags) (Endian.Big))
      (pulley_bswap32 val))
(rule 1 (xswap_if_be val $I64 flags)
      (if-let (endian flags) (Endian.Big))
      (pulley_bswap64 val))
```

and I think we could remove the explicit priorities at that point because isle's overlap checking would be able to see that the rules do not overlap, which would also result in better isle-generated rust code.

pulley/src/lib.rs
```diff
@@ -208,6 +208,11 @@ macro_rules! for_each_op {
             sext16 = Sext16 { dst: XReg, src: XReg };
             /// `dst = sext(low32(src))`
             sext32 = Sext32 { dst: XReg, src: XReg };
+
+            /// `dst = byteswap(low32(src))`
+            bswap32 = Bswap32 { dst: XReg, src: XReg };
+            /// `dst = byteswap(src)`
+            bswap64 = Bswap64 { dst: XReg, src: XReg };
```
fitzgen, line 215: Want to move these to extended ops? I doubt they will be common.

fitzgen (approved):
r=me with the two below suggestions

PR #9760, opened 2024-12-06
Procedurally generate Pulley Cranelift boilerplate

alexcrichton:
This commit is an integration of the `for_each_op!` macro (and extended ops) for Cranelift. This procedurally generates a few new items for Cranelift's Pulley backend to use:

* `RawInst` - a raw enumeration of instructions as-is.
* ISLE constructors (e.g. `pulley_*` ctors) - generated for all of the `RawInst` variants.
* Register allocation methods for `RawInst`
* Pretty printing methods for `RawInst`
* Emission methods for `RawInst`

The `Inst` enum now has a `Raw` variant which contains a `RawInst`. In this manner the main `Inst` enum can still have pseudo-insts like `Call`, polymorphic instructions like loads/stores (probably gonna get refactored in the future though), and slightly different representations such as `Inst::Trap` having a `TrapCode` and `RawInst::Trap` wouldn't.

The goal of this commit is to lower the amount of effort to quickly add and experiment with new instructions in Pulley. Ideally it's now just (a) define them in the pulley macro, (b) implement a direct lowering rule, and (c) implement it in the interpreter. Ideally no need to implement anything else inside of Cranelift as everything should be auto-generated.

Many existing `Inst` variants have been deleted in favor of their equivalents in `RawInst` now. This undeniably increases the complexity of the Pulley backend but at least for me I find it well worth it to have all this boilerplate generated automatically.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -760,6 +760,7 @@ name = "cranelift-codegen-meta"
 version = "0.116.0"
 dependencies = [
  "cranelift-codegen-shared",
+ "pulley-interpreter",
 ]
 
 [[package]]
--- a/cranelift/codegen/Cargo.toml
+++ b/cranelift/codegen/Cargo.toml
@@ -86,7 +86,12 @@ x86 = []
 arm64 = []
 s390x = []
 riscv64 = []
-pulley = ["dep:pulley-interpreter", "pulley-interpreter/encode", "pulley-interpreter/disas"]
+pulley = [
+  "dep:pulley-interpreter",
+  "pulley-interpreter/encode",
+  "pulley-interpreter/disas",
+  "cranelift-codegen-meta/pulley",
+]
 # Enable the ISA target for the host machine
 host-arch = []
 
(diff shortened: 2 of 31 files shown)


alexcrichton:
Heh I was going to cc you @abrown on this as it's definitely related to your assembler work for x64 which I haven't reviewed myself yet. You're tagged as a reviewer here though so works well :)

If you've tried out various patterns for x64 that work well I'd be happy to update Pulley to match said patterns as well.

alexcrichton:
The final commit of https://github.com/bytecodealliance/wasmtime/pull/9765 is an example of how adding a new instruction after this PR requires relatively minimal boilerplate (yay!)

fitzgen (approved):
This looks great -- I'm going to go ahead and merge and if @abrown has feedback on the meta generation approach, we can address it in follow ups. This unblocks more pulley work, and the guide, and all that though, so I think it is important to move forward here now.

abrown:
Yeah, I recognize a lot of the same needs as in what I'm doing ([`assembler-collaborate`](https://github.com/bytecodealliance/wasmtime/compare/main...abrown:wasmtime:assembler-collaborate)). The assembler might get more involved but essentially I'll be doing similar things; I need to remember that [this](https://github.com/bytecodealliance/wasmtime/pull/9760/files#diff-b90f210df8431df9e158dba4609b6e273186773171c6d1d57c7b7bc28117aec6R68) is where I put my newly-generated ISLE!

PR #9761, opened 2024-12-06
Add required `shell: bash` to new action

alexcrichton:
Fixing a mistake in #9754 not caught til it was actually executed during a failed run on #9758.


saulecabrera:
This seems to also fix part of what’s going on here https://github.com/bytecodealliance/wasmtime/pull/9751

saulecabrera (approved):

PR #9763, opened 2024-12-07
Fix some off-by-one comparisons in assertions

alexcrichton:
Fuzzing found a small issue with #9687 and this commit relaxes a few off-by-one checks to allow addressing one-byte-beyond-the-end of a linear memory.

### DIFF
--- a/crates/wasmtime/src/runtime/vm/mmap.rs
+++ b/crates/wasmtime/src/runtime/vm/mmap.rs
@@ -135,7 +135,7 @@ impl Mmap<AlignedLength> {
 
     /// Return a struct representing a page-aligned offset into the mmap.
     ///
-    /// Returns an error if `offset >= self.len_aligned()`.
+    /// Returns an error if `offset > self.len_aligned()`.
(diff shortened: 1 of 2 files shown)


alexcrichton:
cc @sunshowers (you're likely interested, but this of course wasn't your fault)

crates/wasmtime/src/runtime/vm/mmap.rs
```diff
@@ -363,7 +363,7 @@ impl MmapOffset {
         // end of the mmap. We may need to change this if that becomes
```
sunshowers, line 363: Ahh thank you! Haha this comment anticipated it — needs to be changed.

crates/wasmtime/src/runtime/vm/mmap.rs
```diff
@@ -137,7 +137,7 @@ impl Mmap<AlignedLength> {
     ///
     /// Returns an error if `offset >= self.len_aligned()`.
```
sunshowers, line 138: This as well.

I really wasn't sure which one was correct—glad it got caught

fitzgen (approved):

PR #9765, opened 2024-12-08
 Add a guide for adding instructions to Pulley 

alexcrichton:
In helping others contribute to Pulley I've written up the steps I took
to make this commit itself. I've picked a random test in `misc_testsuite`
that previously didn't work on Pulley and was relatively small. I
implemented a few lowerings and some new instructions and now the test
passes. I've added `pulley/CONTRIBUTING.md` to document the experience
along the way in case anyone else is interested to help out in the
future.

### DIFF
--- a/pulley/CONTRIBUTING.md
+++ b/pulley/CONTRIBUTING.md
@@ -1,0 +1,233 @@
+# Contributing
+
+For general contribution to Wasmtime, see Wasmtime's [contributing docs][docs].
+
+[docs]: https://docs.wasmtime.dev/contributing.html
+
+## Adding an instruciton to Pulley
+
+So you want to add an instruction to Pulley. If you're reading this in the
+not-so-distant future Pulley probably doesn't support all of WebAssembly yet and
+you're interesting in helping to improve the situation. This is intended to be a
+small guide about how to add an instruction to Pulley through an early example
+of doing so.
+
+#### Choose a test to get passing
+
+First off find a test in this repository, probably a `*.wast` test, which isn't
+currently passing. At the time of this writing almost no tests are passing, but
+for an up-to-date list check out `crates/wast-util/src/lib.rs`. Here we're going
+to select `./tests/misc_testsuite/control-flow.wast` as it's a reasonably small
+test.
+
+#### See the test failure
+
+Run this command:
+
+```
+$ cargo run --features pulley wast --target pulley64 ./tests/misc_testsuite/control-flow.wast
+```
+
+This builds the `wasmtime` CLI with Pulley support enabled (`--features
+pulley`), runs the `wast` subcommand, executes with the pulley target
+(`--target pulley64`), and then runs our test. As of now this shows:
+
+```
+$ cargo run --features pulley wast --target pulley64 ./tests/misc_testsuite/control-flow.wast
+    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.08s
+     Running `target/debug/wasmtime wast --target pulley64 ./tests/misc_testsuite/control-flow.wast`
+Error: failed to run script file './tests/misc_testsuite/control-flow.wast'
+
+Caused by:
+    0: failed directive on ./tests/misc_testsuite/control-flow.wast:77:1
+    1: Compilation error: Unsupported feature: should be implemented in ISLE: inst = `v5 = sdiv.i32 v2, v3`, type = `Some(types::I32)`
+
+```
+
+Note that if you run `cargo test --test wast control-flow.wast` it'll also run
+the same test, but the test is expected to fail. You can update
+`crates/wast-util/src/lib.rs` to say the test is expected to pass, and then you
+can see a similar failure.
+
+#### Adding an instruction: Pulley
+
+Here the failure is what's the most common failure in Pulley right now -- the
+Pulley Cranelift backend is not yet complete and is missing a lowering. This
+means that there is CLIF that cannot be lowered to Pulley bytecode just yet.
+
+The first thing to do is to probably add a new opcode to Pulley itself as
+Pulley probably can't execute this operation just yet. Here we're interested in
+signed 32-bit division.
+
+Pull up `pulley/src/lib.rs` and you'll be editing the `for_each_op!` macro
+definition. If this is a "rare" opcode you can edit the `for_each_extended_op!`
+macro instead. The syntax is the same between the two macros.
+
+Here this is a simple instruction, so let's add it directly:
+
+```rust
+/// `dst = src1 / src2` (signed)
+xdiv32_s = XDiv32S { operands: BinaryOperands<XReg> };
+```
+
+This defines the disassembled name of the instruction (`i32_div_s`), the Rust
+structure for the instruction (`I32DivS`), and immediates in the instruction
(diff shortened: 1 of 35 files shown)


pulley/CONTRIBUTING.md
```diff
@@ -0,0 +1,233 @@
...
+next part...
+
+#### Adding a Cranelift Lowering
+
+Next up we need to actually fix the error at hand, a new lowering rule needs to
+be added to Cranelift. Here we'll be working in
+`cranelift/codegen/src/isa/pulley_shared/lower.isle`. Cranelift instructions and
+ISLE rules for our new `xdiv32_s` instruction are automatically generated from
+the `for_each_op!` macro, so all we need to do is to add a new lowering rule.
+That'll look like so:
```
fitzgen, line 134:
```suggestion
`cranelift/codegen/src/isa/pulley_shared/lower.isle`.

Our ISLE lowering rules are generally written like this:

\```lisp
(rule (lower <CLIF>)
      <Pulley>)
\```

This means "when lowering Cranelift's mid-level IR down to Pulley bytecode, and we match the snippet `<CLIF>`, replace it with `<Pulley>`". (For more details, see the [ISLE Language Reference](https://github.com/bytecodealliance/wasmtime/blob/main/cranelift/isle/docs/language-reference.md) and [How ISLE is Integrated with Cranelift](https://github.com/bytecodealliance/wasmtime/blob/main/cranelift/docs/isle-integration.md).)

In our case, we need to match an `sdiv` CLIF instruction that is operating on 32-bit values (this is the `(has_type $I32 ...)` bit) and then replace it with our new Pulley `xdiv32_s` instruction:
```

pulley/CONTRIBUTING.md
```diff
@@ -0,0 +1,233 @@
...
+```
+;;;; Rules for `idiv` ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(rule (lower (has_type $I32 (sdiv a b)))
+      (pulley_xdiv32_s a b))
+```
+
+Here the `lower` "rule" is what we're adding and is the main function used to
+lower from CLIF to Pulley bytecode. THe `pulley_xdiv32_s` constructor was
+auto-generated for us as part of `cranelift/codegen/meta`'s build.
```
fitzgen, line 145:
```suggestion
Note that ISLE constructors for Pulley instructions, including the `pulley_xdivs32_s`
constructor for our new `xdiv32_s` Pulley instruction, are automatically
generated from the `for_each_op!` macro.
```

pulley/CONTRIBUTING.md
```diff
@@ -0,0 +1,233 @@
...
+    CraneliftPulley/pooling/./tests/misc_testsuite/control-flow.wast
+
+test result: FAILED. 4 passed; 2 failed; 0 ignored; 0 measured; 4086 filtered out; finished in 0.05s
+
+error: test failed, to rerun pass `--test wast`
+```
+
+This indicates that the test was previously flagged as "should fail", but that
+assertion is no longer true! Update `crates/wast-util/src/lib.rs` about the new
+test passing and we'll see:
```
fitzgen, line 201:
```suggestion
assertion is no longer true! Update the `WastTest::should_fail` method in `crates/wast-util/src/lib.rs` so that it expects the test to pass and we'll see:
```

pulley/CONTRIBUTING.md
```diff
@@ -0,0 +1,233 @@
...
+$ cargo test --test wast control-flow.wast
+    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.74s
+     Running tests/wast.rs (target/debug/deps/wast-f83a3ee5e5dbacde)
+
+running 6 tests
+......
+test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 4086 filtered out; finished in 0.05s
+```
+
+Success! Let's see if we caused any other tests to start passing...
```
fitzgen, line 213:
```suggestion
Success!

But we aren't quite done yet: the new lowerings we added might have made additional tests that previously failed start passing as well. If so, then we also want to mark those tests as expected to pass now. We can double check by running the full `wast` test suite:
```

pulley/src/interp.rs
```diff
@@ -1439,6 +1439,26 @@ impl OpVisitor for Interpreter<'_> {
...
+    fn xdiv32_s(&mut self, operands: BinaryOperands<XReg>) -> ControlFlow<Done> {
+        let me = self.current_pc::<crate::XDiv32S>();
+        let a = self.state[operands.src1].get_i32();
+        let b = self.state[operands.src2].get_i32();
+        match a.checked_div(b) {
+            Some(result) => {
+                self.state[operands.dst].set_i32(result);
+                ControlFlow::Continue(())
+            }
+            None => ControlFlow::Break(self.done_trap(me)),
```
fitzgen, line 1452: Could we make `done_trap` generic over an instruction and then avoid the need to do `let me = self.current_pc::<...>();` and instead do

```rust
ControlFlow::Break(self.done_trap::<...>())
```

or even fold the `ControlFlow::Break` into the helper?

This both delays the computation until its demanded (which LLVM can probably do, but might be foiled if there are zero/overflow whatever checks hidden in there) and is a little more concise at usage/call sites.

fitzgen (approved):
Looks great!

I have a bunch of nitpicks and suggestions to make things flow a little smoother IMO or to link to other references and resources, but those are tiny nitpicks and this is fantastic.

Excited to start onboarding new contributors with this!

pulley/CONTRIBUTING.md
```diff
@@ -0,0 +1,233 @@
...
+Rerun our test command and we see:
+
+```
+$ cargo run --features pulley wast --target pulley64 ./tests/misc_testsuite/control-flow.wast
+   Compiling pulley-interpreter v29.0.0 (/home/alex/code/wasmtime/pulley)
+error[E0046]: not all trait items implemented, missing: `i32_div_s`
+   --> pulley/src/interp.rs:807:1
+    |
+807 | impl OpVisitor for Interpreter<'_> {
+    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `i32_div_s` in implementation
```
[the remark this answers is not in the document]
alexcrichton: Heh you can see I switched halfway through :)

pulley/src/interp.rs
alexcrichton: Excellent idea 👍

pulley/CONTRIBUTING.md
```diff
@@ -0,0 +1,273 @@
...
+```
+
+Alas, maybe next time!
+
+#### Clean up and make a PR
+
+All that's left now is to clean things up, document anything necessary, and make
+a pull request.
+
+Thanks for helping out!
```
fitzgen, line 273: Maybe also link to this PR at the end?

```suggestion
To view the complete pull request that implemented `xdiv32_s` and `xand32` Pulley instructions and got `./tests/misc_testsuite/control-flow.wast` passing (and also introduced this documentation) check out https://github.com/bytecodealliance/wasmtime/pull/9765

Thanks for helping out!
```

fitzgen (approved):

PR #9770, opened 2024-12-09
wasmtime_instance_pre_instantiate should take a wasmtime_context_t

simonvik:
The current definition of wasmtime_instance_pre_instantiate takes a wasmtime_store_t but the rust-code seem to want a wasmtime_context_t.


alexcrichton (approved):
Thanks!

Issue #9747, opened 2024-12-05
Pulley TODO List


alexcrichton:
I'd like a place to throw small items as I see them and to collect various TODO items during the development of Pulley. For now this will serve that location. My hope is that this list is TODO items which pertain to code in-tree and are less-so about wishlist items of what's yet to fill in. For example this won't track missing pulley work (e.g. all wasm proposals at this time). Unsure if this style of issue will work out.

- [ ] Evaluate 32-bit platforms and the decision to "only write low 32-bits". If we write the full 64-bit width of registers does it actually have a performance loss? Is it worth burning opcode space for opcodes that only write the low-32 instead of high 64? Should we split opcodes like `xconst8` into one that writes 64-bits and one that writes 32-bits? Should we remove `xload8_s32_offset32` instead?
- [x] Add support for Pulley to `#[wasmtime_test]` https://github.com/bytecodealliance/wasmtime/pull/10057
- [x] Move sp/fp/lr/special regs out of the general purpose register set; see https://github.com/bytecodealliance/wasmtime/pull/9658#pullrequestreview-2482824188 for details (also cc https://github.com/bytecodealliance/wasmtime/pull/9806)
- [x] Explore adding new addressing modes (e.g. register + register). Should be based on an evaluation of what wasm loads/stores do probably. Perhaps even fold a trapping arithmetic into one macro-op for "do the wasm load on 32-bit" and "do the wasm load on 64-bit" https://github.com/bytecodealliance/wasmtime/pull/9994
- [x] Assert in call lowering that pulley<->pulley is always tail-to-tail or similar; see https://github.com/bytecodealliance/wasmtime/pull/9665#discussion_r1871969050 for details -- https://github.com/bytecodealliance/wasmtime/pull/9836
- [x] Better debugging support, e.g. print the register state between instructions and print each instruction. https://github.com/bytecodealliance/wasmtime/pull/9796
- [x] Support big-endian - will need to add `pulley{64,32}be` targets to target-lexicon, then add big-endian loads/stores, then update the CLIF backend to delegate to the right endianness based on the target
- [x] Need to restore callee-save state when a trap happens somehow (either via explicit saves or by removing callee-save state in that situation) https://github.com/bytecodealliance/wasmtime/pull/9837
- [x] Enable Cranelift `runtest` support https://github.com/bytecodealliance/wasmtime/pull/9795
[...]

PR #9775, opened 2024-12-10
pulley: Implement more of loads/stores

alexcrichton:
This commit gets the `address.wast` spec test working by filling out more load/store infrastructure in Pulley. In doing so I've done a refactoring of the existing load/store methods. Changes here are:

* All load/stores are `*_offset32` now instead of optionally having no offset, a 8-bit offset, or a 64-bit offset.
* All x-register loads/stores are prefixed with `x` now.
* All loads/stores have "le" for little-endian in their name.
* Loads/stores are refactored to have 8 and 16-bit variants.
* Sign-extending loads now either extend to 32 or 64 depending on the opcode.
* Float loads/stores are added.
* Big-endian is handled with explicit big-endian loads/stores instead `bswap` to handle this transparently in the backend (e.g. for stores not ISLE-generated) and to handle floats.


pulley/src/lib.rs
```diff
@@ -15,6 +15,68 @@ extern crate std;
...
+///
+/// * Most instructions are suffixed or otherwise contain the bit width they're
+///   operating on. For example `xadd32` is a 32-bit addition.
+///
+/// * Instructions operate on either 32 or 64-bit parts of a register.
+///   Instructions modifying only 32-bits of a register always modify the "low"
+///   part of a register and leave the upper part unmodified. This is intended
+///   to help 32-bit platforms where if most operations are 32-bit there's no
+///   need for extra instructions to sign or zero extend and modify the upper
+///   half of the register.
```
alexcrichton, line 55: Curious on your thoughts on this in particular @fitzgen.

I'll note that we don't currently do this well, e.g. `xconst8` always writes the full 64-bits and `br_if` has no variant that only reads 32-bits. I think we'll want to change that over time though.

alexcrichton:
At a high level the `spec_testsuite/*.wast` still can't be executed, but that's the next PR after this to get that executing on CI (and a large number of tests already pass)

pulley/fuzz/src/interp.rs
alexcrichton, line 1: Also curious on your thoughts here @fitzgen of the deletion of this fuzzer. Personally I feel like it's a bit of a pain to maintain while not adding much more over the wasm-smith fuzzing we have already.

cranelift/codegen/src/isa/pulley_shared/lower.isle
```diff
@@ -167,6 +167,20 @@
 (rule (lower (has_type $I64 (iadd a b)))
       (pulley_xadd64 a b))
 
+;;;; Rules for `isub` ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(rule (lower (has_type $I8 (isub a b)))
+      (pulley_xsub32 a b))
+
+(rule (lower (has_type $I16 (isub a b)))
+      (pulley_xsub32 a b))
```
alexcrichton, line 176: Also double-check me on this, is this right?

fitzgen: I don't think so, since (unlike addition) subtraction lets high bits (which should be invisible) move down to the low bits (and become visible), right?

pulley/src/lib.rs
```diff
@@ -15,6 +15,68 @@ extern crate std;
...
+/// of design questions. Note that this is explicitly intended to be only ever
+/// used for Pulley where there are a different set of design constraints than
+/// other instruction sets (e.g. general-purpose CPU ISAs). Some examples of
+/// constraints for Pulley are:
+///
+/// * Instructions must be portable to many architectures.
+/// * The Pulley ISA is mostly target-independent as the compilation target is
+///   currently only parameterized on pointer width and endianness.
+/// * Pulley instructions must be fast to decode at runtime (e.g. it's theorized
+///   we can't get super fancy with bit-packing)
```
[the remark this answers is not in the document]
alexcrichton: Good point, I'll add notes about the balance. I was mostly thinking here of things like RISC-V instructions where the registers are spread throughout the instruction in various locations to fit around encodings of other instructions. That's the "super fancy" part but I also don't think we're at risk of getting there any time soon.

cranelift/codegen/src/isa/pulley_shared/lower.isle
fitzgen: let's merge this and get further in spec tests, and eventually differential fuzzing, which will definitely catch any issues here

pulley/src/lib.rs
alexcrichton: This I feel is a bit entangled with the question below of "do we want `xload8_s32_offset32`". I'm a bit worried about `xconst8` always filling the entire register mostly from the perspective of 32-bit platforms. If most constants in a program are actually 32-bits then we're in theory spending a lot of instructions to fill the upper half of the register that's never read. Cache-wise that's probably not too problematic but it's possibly problematic from an icache perspective of the implementation of `xconst8` itself.

I definitely agree though that for 64-bit platforms it doesn't matter at all.

For now I'm going to go ahead with this PR as-is (with `xload8_s32_offset32` for example) and these docs since we don't have a great answer one way or another and I think we're both on the fence. I'm going to add some notes to https://github.com/bytecodealliance/wasmtime/issues/9747 to come back to reevaluate though.

PR #9778, opened 2024-12-10
Wasmtime: support a notion of "custom code publisher".

cfallin:
In some `no_std` environments, virtual memory usage is *generally* prohibited for performance-predictability reasons, but the MMU hardware is still in use for permissions (e.g., `W^X` write-xor-execute). Occasional changes to page mapping permissions are thus necessary when new modules are loaded dynamically, and are acceptable in that context. Wasmtime needs a way to support "publishing" code (making it executable) in such environments.

Rather than try to segment the `signals-based-traps` divide further, and piece out the code-publishing parts from the heap parts, and backdoor a path to `mprotect` in an otherwise `no_std` build, in this PR I have opted to add a trait an impl of which the embedder can provide to the `Config` to implement custom actions for "code publish". This otherwise operates properly in a
no-`signals-based-traps` environment, e.g., the module backing memory itself is regularly allocated rather than mmap'd (but is now aligned to the degree requested by the trait impl).


crates/wasmtime/src/runtime/vm/mmap_vec.rs
```diff
@@ -68,13 +94,49 @@ impl MmapVec {
...
+            return Ok(MmapVec::new_mmap(Mmap::with_at_least(size)?, size));
+        }
+        #[cfg(not(feature = "signals-based-traps"))]
+        {
+            return Ok(MmapVec::new_vec_aligned(
+                vec![0; size + alignment],
+                size,
+                alignment,
+            ));
+        }
```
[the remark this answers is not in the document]
cfallin: Unfortunately that's not an option in my embedding: memory must come from the `alloc` allocator.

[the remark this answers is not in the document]
cfallin: Ah, actually, do you mean the lower-level `alloc` APIs? I'll look into that.

crates/wasmtime/Cargo.toml
```diff
@@ -364,3 +364,14 @@ wave = ["dep:wasm-wave"]
...
+# Adds support for a custom mechanism to make memory executable.
+#
+# When running in a `no_std` environment that nominally does not
+# permit modification of virtual memory mappings, but that does have
+# page permissions and a mechanism of updating them that can be used
+# off the critical path, the embedder will want to provide that
+# mechanism to allow loading of modules and making their code
+# executable. This feature defines a trait that can be provided to the
+# Engine to permit this.
+custom-code-memory = []
```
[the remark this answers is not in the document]
cfallin: Sure, happy to fold it in; was worried the extra checks might cause some consternation but you're right, it's probably not too bad (and loading modules is relatively not a hot path).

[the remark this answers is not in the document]
cfallin: Done.

crates/wasmtime/src/runtime/vm/mmap_vec.rs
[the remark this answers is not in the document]
cfallin: Done!

crates/wasmtime/src/runtime/vm/mmap_vec.rs
```diff
@@ -68,13 +94,49 @@ impl MmapVec {
...
+    /// `slice`, with a minimum alignment.
+    ///
+    /// `align` must be a power of two. This is useful when page
+    /// alignment is required when the system otherwise does not use
+    /// virtual memory but has a custom code publish handler.
+    ///
+    /// A new `MmapVec` is allocated to hold the contents of `slice` and then
+    /// `slice` is copied into the new mmap. It's recommended to avoid this
+    /// method if possible to avoid the need to copy data around.  pub
+    pub fn from_slice_with_alignment(slice: &[u8], align: usize) -> Result<MmapVec> {
```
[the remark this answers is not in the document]
cfallin: Done.

crates/wasmtime/src/engine.rs
```diff
@@ -61,6 +63,8 @@ struct EngineInner {
     signatures: TypeRegistry,
     #[cfg(feature = "runtime")]
     epoch: AtomicU64,
+    #[cfg(feature = "runtime")]
+    custom_code_memory: Option<Arc<dyn CustomCodeMemory>>,
```
alexcrichton, line 67: This is already stored as-is in `Config` so might be able to access from there (stored above) instead of also storing here too?

crates/wasmtime/src/engine.rs
```diff
@@ -655,6 +661,11 @@ impl Engine {
         &self.inner.signatures
     }
 
+    #[cfg(feature = "runtime")]
+    pub(crate) fn custom_code_memory(&self) -> Option<Arc<dyn CustomCodeMemory>> {
```
alexcrichton, line 665: Would it be possible to return `Option<&dyn ...>` here to avoid forcing the clone on callers? Or perhaps `Option<&Arc<dyn ...>>` if callers should optionally have the ability to create an owned copy

crates/wasmtime/src/runtime/vm/mmap_vec.rs
```diff
@@ -35,6 +44,38 @@ pub enum MmapVec {
...
 }
 
+#[cfg(not(feature = "signals-based-traps"))]
+mod base_ptr {
+    use core::ptr::NonNull;
+
+    /// A base pointer to an allocation, guaranteed not to be null,
+    /// and `Send`/`Sync`.
+    #[derive(Clone, Copy)]
+    pub struct BasePtr(NonNull<u8>);
```
alexcrichton, line 54: Would `SendSyncPtr` suffice for this?

crates/wasmtime/src/runtime/vm/mmap_vec.rs
```diff
@@ -20,13 +22,20 @@ use std::fs::File;
...
+/// and the methods are not available. However, the `CustomCodeMemory`
+/// mechanism may be used by the embedder to set up and tear down
+/// executable permissions on parts of this storage.
 pub enum MmapVec {
     #[doc(hidden)]
     #[cfg(not(feature = "signals-based-traps"))]
-    Vec(Vec<u8>),
+    Alloc {
+        base: base_ptr::BasePtr,
+        len: usize,
```
alexcrichton, line 36: I think `len` can be dropped here in favor of `layout.size()`?

crates/wasmtime/src/compile/runtime.rs
```diff
@@ -137,7 +137,7 @@ impl FinishedObject for MmapVecWrapper {
 
             fn reserve(&mut self, additional: usize) -> Result<(), ()> {
                 assert!(self.mmap.is_none(), "cannot reserve twice");
-                self.mmap = match MmapVec::with_capacity(additional) {
+                self.mmap = match MmapVec::with_capacity_and_alignment(additional, 1) {
```
alexcrichton, line 140: I think this may not work for the use case of combining `CustomCodeMemory` and compilation at runtime, so the `Engine` and its custom alignment might need to be plumbed down here?

alexcrichton (approved):
Would it be possible to add a test exercising this trait as well? Perhaps x86_64-linux-specific to avoid rewriting a bunch of logic to make things executable or something like that.

crates/wasmtime/src/runtime/vm/mmap_vec.rs
cfallin: Done!

crates/wasmtime/src/engine.rs
cfallin: Ah, right, good point -- updated.

crates/wasmtime/src/compile/runtime.rs
cfallin: Done -- this was slightly gnarly to plumb through (required an extra `State` arg on the `FinishedObject` trait in `environ` and propagated through all the delegated builder stuff and cache handling and ...) but seems to work.

cfallin:
Updated, thanks for feedback! Added a test that uses `rustix` on non-Windows targets to delegate to `mprotect` ("custom but actually the same as always" implementation).

PR #9779, opened 2024-12-10
 pulley: Ungate SIMD proposal to enable more tests 

alexcrichton:
This commit removes the simd proposal from "known panicking compiler
features" for Pulley by filling out some minor ABI bits around
loads/stores of v128 in the Cranelift ABI code. This unlocks running
`spec_testsuite/*.wast` in CI and expecting it to pass as otherwise
everything was rejected since that test suite requires SIMD and it never
even got to the Cranelift backend.

This then fixes a few minor issues here and there to get existing tests
passing now that they're actually being run. Some examples are:

* Fix some copy/paste typos around callee-save clobbers and float
  registers.
* Add vector load/store support to avoid panicking in ABI code (not well
  tested yet though).
* Zero-extend the condition in `brif`'s lowering to ensure that the
  entire register is defined. In the future more variants of `br_if` in
  Pulley are probably desired.
* Limit `store` lowerings to just integer types, not vector types.
* Fix `Func::call` with Pulley when the callee is a host function.


alexcrichton:
Note that this is currently based on https://github.com/bytecodealliance/wasmtime/pull/9775

fitzgen (approved):
Nice

PR #9780, opened 2024-12-10
 pulley: Ungate memory64 feature

alexcrichton:
This commit is similar to https://github.com/bytecodealliance/wasmtime/pull/9779 in that it's removing a proposal from
the "known list of panicking features" for Pulley to allow more tests to
run on Pulley. This then fills out a few miscellaneous instructions to
get a full suite of tests passing in Pulley related to memory64 and
other instructions.

cc #9783

### DIFF
--- a/crates/cranelift/src/translate/code_translator/bounds_checks.rs
+++ b/crates/cranelift/src/translate/code_translator/bounds_checks.rs
@@ -499,6 +499,10 @@ fn cast_index_to_pointer_ty(
     //
     // Also note that at this time this branch doesn't support pcc nor the
     // value-label-ranges of the below path.
+    //
+    // Finally, note that the returned `low_bits` here are still subject to an
+    // explicit bounds check in wasm so in terms of Spectre speculation on
+    // either side of the `trapnz` should be ok.
     if index_ty.bits() > pointer_ty.bits() {
         assert_eq!(index_ty, ir::types::I64);
         assert_eq!(pointer_ty, ir::types::I32);
--- a/crates/cranelift/src/translate/code_translator/bounds_checks.rs
+++ b/crates/cranelift/src/translate/code_translator/bounds_checks.rs
@@ -491,11 +491,25 @@ fn cast_index_to_pointer_ty(
     if index_ty == pointer_ty {
         return index;
     }
-    // Note that using 64-bit heaps on a 32-bit host is not currently supported,
-    // would require at least a bounds check here to ensure that the truncation
-    // from 64-to-32 bits doesn't lose any upper bits. For now though we're
-    // mostly interested in the 32-bit-heaps-on-64-bit-hosts cast.
-    assert!(index_ty.bits() < pointer_ty.bits());
+
+    // If the index size is larger than the pointer, that means that this is a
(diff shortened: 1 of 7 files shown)


alexcrichton:
> **Note**: currently stacked on https://github.com/bytecodealliance/wasmtime/pull/9779

crates/cranelift/src/translate/code_translator/bounds_checks.rs
```diff
@@ -491,11 +491,25 @@ fn cast_index_to_pointer_ty(
...
+    // value-label-ranges of the below path.
+    if index_ty.bits() > pointer_ty.bits() {
+        assert_eq!(index_ty, ir::types::I64);
+        assert_eq!(pointer_ty, ir::types::I32);
+        let low_bits = pos.ins().ireduce(pointer_ty, index);
+        let c32 = pos.ins().iconst(pointer_ty, 32);
+        let high_bits = pos.ins().ushr(index, c32);
+        let high_bits = pos.ins().ireduce(pointer_ty, high_bits);
+        pos.ins()
+            .trapnz(high_bits, ir::TrapCode::HEAP_OUT_OF_BOUNDS);
```
cfallin, line 510: I think we should make a note re: Spectre here -- `trapnz` internally uses a branch and so one might expect there to be some exposure here. I think we're actually safe because the worst that happens is that the guest accesses an OOB address like `0x1_0000_1000` where `0x1000` is in-bounds, and gets its own in-bounds data; in other words, this check is layered on top of the actual bounds-check on the lower 32 bits, so there is still not any visibility outside the sandbox in the misspeculated path. But it's... worth noting, if only because the guest might in turn rely on OOBs not to speculatively read valid data (niche but possible).

cfallin (approved):
Last commit LGTM, with note about Spectre below (happy to discuss further re: how much of an issue and/or how or if to document this).

crates/wasmtime/src/runtime/vm/memory.rs
```diff
@@ -290,22 +290,6 @@ impl Memory {
...
-            1 << 32
-        } else {
-            u64::try_from(absolute_max).unwrap()
-        };
-        if let Ok(size) = ty.minimum_byte_size() {
-            assert!(size <= absolute_max64);
-        }
-        if let Ok(max) = ty.maximum_byte_size() {
-            assert!(max <= absolute_max64);
-        }
```
fitzgen, line 307: Why were these asserts removed?

fitzgen (approved):
LGTM modulo @cfallin's comment

crates/wasmtime/src/runtime/vm/memory.rs
alexcrichton: They were basically a pain to maintain on 32-bit because I already had to modify the minimum check to use `1<<32` rather than `absolute_max` (due to `absolute_max` being a `usize` and not being able to represent `1<<32`) and then this PR uncovered an assertion in the next one where a 64-bit linear memory's maximum size far exceeds `1<<32` as well. Given that these are just debug asserts that are already basically checked in many other places it seemed best to just remove them instead of trying to contort ourselves to keep them in all portable conditions.

PR #9782, opened 2024-12-10
winch: relax some FnMut constraints

MarinPostma:
while working with the `emit_binop` operator, I found myself needed to derive copy and clone for some types, because the callback was `FnMut`. After inspection, the closure doesn't need to be `FnMut`, and `FnOnce` suffice. This is not much, but I think it improves the ergonomics slightly.


saulecabrera (approved):
Thanks!

PR #9784, opened 2024-12-11
winch(aarch64): Handle division signedness

saulecabrera:
This commit fixes how signedness division is hanlded in aarch64. Prior to this commit, sign-extension was emitted unconditionally.

This commit ensures that the correct extension is emitted depending on the division kind.

Additionally this commit prefers making use of existing assembler helpers.


fitzgen (approved):
:+1:

Issue #9783, opened 2024-12-11
Get all `.wast` tests passing under Pulley


fitzgen:
Let's get all `.wast` tests passing under Pulley!

[Pulley](https://github.com/bytecodealliance/wasmtime/blob/main/pulley/README.md) is Wasmtime's portable, optimizing interpreter. When using Pulley, Wasmtime still uses Cranelift (our optimizing compiler) for optimizations, but Cranelift ultimately emits Pulley bytecode instead of machine code. Then, Wasmtime uses Pulley's interpreter to interpret the bytecode, instead of executing native machine code.

<details> <summary>Diagram of the compiler pipeline with Pulley</summary>

```
         +------+
         | Wasm |
         +------+
            |
            |
  Wasm-to-CLIF translation (Wasmtime)
            |
            V
         +------+
         | CLIF |
         +------+
            |
            |
   mid-end optimizations (Cranelift)
            |
            V
         +------+
         | CLIF |
         +------+
            |
            |
    ISLE lowering rules (Cranelift)
            |
            V
   +-----------------+
   | Pulley bytecode |
   +-----------------+
```

</details>

This is a good first issue for new contributors that are already somewhat familiar with Rust development and are excited to dive into Wasmtime/Cranelift/compilers/interpreters/etc. To help out with this effort, follow this guide:

### [Guide: How to Add Missing Instructions to Pulley and Get `.wast` Tests Passing](https://github.com/bytecodealliance/wasmtime/blob/main/pulley/CONTRIBUTING.md)

Note that we have [docs for contributing to Wasmtime and Cranelift in general](https://docs.wasmtime.dev/contributing.html) as well.

For the most up-to-date list of which `.wast` tests are passing and which are failing under Pulley, see [the `WastTest::should_fail` method in `crates/wast-util/src/lib.rs`](https://github.com/bytecodealliance/wasmtime/blob/main/crates/wast-util/src/lib.rs#L405).

Xuanwo:
For newcomers to this issue:

- [ ] "misc_testsuite/simd/canonicalize-nan.wast"
- [ ] "misc_testsuite/simd/issue_3327_bnot_lowering.wast"
- [ ] "misc_testsuite/simd/v128-select.wast"
- [ ] "spec_testsuite/proposals/annotations/simd_lane.wast"
- [ ] "spec_testsuite/proposals/relaxed-simd/i16x8_relaxed_q15mulr_s.wast"
[...]

PR #9785, opened 2024-12-11
 pulley: Ungate the GC_TYPES feature from Pulley 

alexcrichton:
Similar to https://github.com/bytecodealliance/wasmtime/pull/9779 this remove the `GC_TYPES` feature from the list of
panicking features in Pulley. In doing so this then additionally fixes a
number of panics and then adds more tests that are working. Some other
minor instructions are filled out to ensure that tests are working on
both 32 and 64-bit platforms.

cc #9783

PR #9786, opened 2024-12-11
 pulley: Update how tests are expected to fail or pass 

alexcrichton:
With enough basic support here and there switch the listing of tests for
Pulley from known-passing to known-failing instead. This trims the list
of unsupported features to just those that are truly unsupported and
then additionally rewrites the list of tests to be an exhasutive list of
unsupported tests instead of a list of supported tests.

This should make it a bit easier to track the progress as now pulley is
"complete" once this list is empty.

cc #9783


alexcrichton:
Note that this is currently based on https://github.com/bytecodealliance/wasmtime/pull/9785 which is itself based on https://github.com/bytecodealliance/wasmtime/pull/9780

fitzgen (approved):
Nice

PR #9788, opened 2024-12-11
Revamp wasi example and related docs

### LINKED ISSUE #9777: Renovate the WASI example
#### Feature

Please renovate the WASI example in the [documentation](https://docs.wasmtime.dev/examples-rust-wasi.html) and the [code](https://github.com/bytecodealliance/wasmtime/blob/main/examples/wasi-async/main.rs), as they are using old `wasi_common`  and super outdated `wasmtime_wasi `.

#### Benefit

Updating the doc and example should allow newcomers find latest stable way to run WASI components, instead of prehistoric workarounds.

#### Implementation

I can make a PR.

ifsheldon:
Closes #9777

~~For now, this is a draft PR because there's a runtime issue in the revamp example. Please see the FIXME in `examples/wasi/main.rs`. There I want to invoke the exported `wasi:cli/run@0.2.0` function, which is the main function in `examples/wasi/wasm/wasi.rs`. Probably @alexcrichton or @fitzgen can help a bit? Thanks!~~

~~When the runtime issue is fixed, I will try to revamp more sections in the related doc in following commits~~

Updated:

This PR revamps the examples `wasi` and `wasi-async` with latest updates from `wasmtime` and `wasmtime-wasi` and uses WASIp2 APIs by default. Related documentation is also updated.


docs/examples-rust-wasi.md
```diff
@@ -5,33 +5,41 @@ repository to run the example locally.
...
 
 `wasi.rs`
 ```rust
 {{#include ../examples/wasi/wasm/wasi.rs}}
 ```
 
-Building this program generates `target/wasm32-wasip1/debug/wasi.wasm`, used below.
+> Building instructions:
+> 1. Update your Rust toolchain if it is not >= 1.83.0, as Rust 1.83.0 brings Tier 2 support for WASI Preview 2.
+>   `rustup update stable`
```
alexcrichton, line 26: I might recommend not using an exact rustc version here or just omitting this entirely. All relatively recent rust toolchains will give an error "this is too old" already so it's probably sufficient to just say "have rust installed" here

examples/wasi/main.rs
```diff
@@ -6,31 +6,52 @@ You can execute this example with:
...
+        wasi_ctx: wasi,
+        resource_table: ResourceTable::new(),
+    };
+    let mut store = Store::new(&engine, state);
 
+    // Instantiate our component with the imports we've created, and run it.
+    let component = Component::from_file(&engine, "target/wasm32-wasip2/debug/wasi.wasm")?;
+    let instance = linker.instantiate(&mut store, &component)?;
+    let exp_idx= instance.get_export(&mut store, None, "wasi:cli/run@0.2.0").unwrap();
+    // FIXME: Why can't I get an exported function here? The exp_idx is valid
```
alexcrichton, line 52: For this, would [this example](https://docs.rs/wasmtime-wasi/latest/wasmtime_wasi/bindings/sync/struct.Command.html) help?

docs/examples-rust-wasi.md
ifsheldon: updated

examples/wasi/main.rs
ifsheldon: Yes, thanks!

ifsheldon:
But now there comes another issue: this hello-world example is just too specific to wasi:cli, since we used a specific `wasmtime_wasi::bindings::sync::Command` to run a `wasi:cli/command`. For a more general component, we need to add another example or at least give a pointer.

BTW, I still don't know why I could not get a `Func` to execute the `run` function in `wasi:cli/run`.

alexcrichton:
Personally I think it's reasonable to start with hello-world like this and then graduate up to other examples. For example this page could link to the `bindgen_examples` page and that could get expanded with a custom world for example (IIRC there may already be one there?)

> BTW, I still don't know why I could not get a Func to execute the run function in wasi:cli/run.

You needed another invocation of `get_export`. You loaded the exported instance but you'll need to load the exported function of that instance as well.

ifsheldon:
OK, I think this draft PR seems good to me now, so I turned it into a formal PR.

ifsheldon:
> You needed another invocation of get_export. You loaded the exported instance but you'll need to load the exported function of that instance as well.

OK, got that. I also added related code in the example to demonstrate how to do that. 

BTW, do you know why [`ComponentNamedList`](https://docs.rs/wasmtime/latest/wasmtime/component/trait.ComponentNamedList.html) is implemented for tuple `(A1,)` but not `A1`? This is bit counter-intuitive when I want to write something like `func.typed::<(), Result<(), ()>>(&store)`, which `rustc` complains about not implementing `ComponentNamedList`. I really didn't know how to fix this error until I lookup `Implementations on Foreign Types` section of [`ComponentNamedList`](https://docs.rs/wasmtime/latest/wasmtime/component/trait.ComponentNamedList.html).

docs/examples-rust-wasi.md
```diff
@@ -1,101 +1,61 @@
-# WASI
+# WASI (Preview 2)
```
alexcrichton, line 1: Nowadays we try to call this "WASIp2", so mind renaming that here?

docs/examples-rust-wasi.md
```diff
@@ -1,101 +1,61 @@
...
 
-[`wasi-common`]: https://crates.io/crates/wasi-common
+[`wasmtime-wasi`]: https://crates.io/crates/wasmtime-wasi
 [`Linker`]: https://docs.rs/wasmtime/*/wasmtime/struct.Linker.html
 
-### WebAssembly module source code
+## WebAssembly Component Source Code
 
-For this WASI example, this Hello World program is compiled to a WebAssembly module using the WASI Preview 1 API.
+For this WASI example, this Hello World program is compiled to a WebAssembly module using the WASI Preview 2 (WASIp2) API.
```
alexcrichton, line 17: Same here as above, it's ok to replace "WASI Preview 1" with "WASIp2"

docs/examples-rust-wasi.md
```diff
@@ -1,101 +1,61 @@
...
-## WASI Preview 2
-
-An experimental implementation of the WASI Preview 2 API is also available, along with an adapter layer for  WASI Preview 1 WebAssembly modules. In future this `preview2` API will become the default. There are some features which are currently only accessible through the `preview2` API such as async support and overriding the clock and random implementations.
-
 ### Async example
 
-This [async example code][code2] shows how to use the [wasmtime-wasi::preview2][`preview2`] module to
-execute the same WASI Preview 1 WebAssembly module from the example above. This example requires the `wasmtime` crate `async` feature to be enabled.
+This [async example code][code2] shows how to use the [wasmtime-wasi][`wasmtime-wasi`] module to
+execute the same WASI Preview 2 WebAssembly component from the example above. This example requires the `wasmtime` crate `async` feature to be enabled.
```
alexcrichton, line 43: Another `s/WASI Preview 2/WASIp1/`

examples/wasi/main.rs
```diff
@@ -6,31 +6,79 @@ You can execute this example with:
...
+    // Get the index for the exported interface
+    let interface_idx = instance
+        .get_export(&mut store, None, "wasi:cli/run@0.2.0")
+        .unwrap();
+    // Get the index for the exported function in the exported interface
+    let parent_export_idx = Some(&interface_idx);
+    let func_idx = instance
+        .get_export(&mut store, parent_export_idx, "run")
+        .unwrap();
+    let func = instance.get_func(&mut store, func_idx).unwrap();
```
alexcrichton, line 74: Could the errors be handled here instead of using `.unwrap()`?

alexcrichton (commented):
Thanks!

> This is bit counter-intuitive when I want to write something like func.typed::<(), Result<(), ()>>(&store),

This is currently intentional in the sense that it's in theory preferred to use the output of `bindgen!` where you don't have to deal with this. In that sense the raw component APIs weren't designed to be the most ergonomic and easy-to-use since that's where `bindgen!` comes in to close the gap.

IIRC there were coherence with that trait impl, but I could be misremembering as well.

docs/examples-rust-wasi.md
ifsheldon: done

alexcrichton (approved):

ifsheldon:
Oops. I digged into `examples/wasi/main.c` and related `c-api` crate. It seems `c-api` crate also needs some renovation.........

For example, `c-api` exposes this function, which links wasip1 interfaces, but no other functions link wasip2 interfaces.
https://github.com/bytecodealliance/wasmtime/blob/db4bd219fae182cdebd8ea77b1afbd1bef6ea5c0/crates/c-api/src/linker.rs#L111

I think removing `preview1` feature from this line and fixing whatever breaks should be enough. This introduces API breaking changes, though.

https://github.com/bytecodealliance/wasmtime/blob/db4bd219fae182cdebd8ea77b1afbd1bef6ea5c0/crates/c-api/Cargo.toml#L36 

I'd rather like to skip this error first and open another tracking issue.

ifsheldon:
Can you take a look at the CI error [Test C-API macos-latest](https://github.com/bytecodealliance/wasmtime/pull/9788#logs)? I think I made a right correction on CMakeList.txt but I don't know what else needs to be fixed.

alexcrichton:
Have you tried running the example locally?

docs/examples-rust-wasi.md
```diff
@@ -1,101 +1,61 @@
...
-
-## WASI Preview 2
-
-An experimental implementation of the WASI Preview 2 API is also available, along with an adapter layer for  WASI Preview 1 WebAssembly modules. In future this `preview2` API will become the default. There are some features which are currently only accessible through the `preview2` API such as async support and overriding the clock and random implementations.
-
 ### Async example
 
-This [async example code][code2] shows how to use the [wasmtime-wasi::preview2][`preview2`] module to
-execute the same WASI Preview 1 WebAssembly module from the example above. This example requires the `wasmtime` crate `async` feature to be enabled.
+This [async example code][code2] shows how to use the [wasmtime-wasi][`wasmtime-wasi`] component to
```
dicej, line 42:
```suggestion
This [async example code][code2] shows how to use the [wasmtime-wasi][`wasmtime-wasi`] crate to
```

Here we're talking about a Rust crate rather than a Wasm component.

dicej (approved):

alexcrichton:
Ah I better understand what was going on now. Instead of replacing the current example which I think is still useful for wasm32-wasip1 targets, could this perhaps rename the prior wasi example to wasip1 and add this new example under wasip2? There then wouldn't be a C example of wasip2 because that's not supported yet.

docs/examples-rust-wasi.md
ifsheldon: yeah, that's a typo due to replacing "module" to "component"

ifsheldon:
OK, I've brought back wasip1 examples.

alexcrichton (approved):

ifsheldon:
This [test failure](https://github.com/bytecodealliance/wasmtime/actions/runs/12434560618/job/34718541405) is weird to me. I have cleaned all caches and run the commands in this test manually but I saw no failures.

ifsheldon:
Yes. Here is my complete log of my terminal outputs. Search for "$" to see my commands:
1. `cargo clean`
2. `cmake -Sexamples -Bexamples/build -DBUILD_SHARED_LIBS=OFF` copied from CI
3. `cmake --build examples/build --config Debug` copied from CI
4. `cmake -E env CTEST_OUTPUT_ON_FAILURE=1 cmake --build examples/build --config Debug --target test` copied from CI

[log.txt](https://github.com/user-attachments/files/18322068/log.txt)

ifsheldon:
I think CI silently swallow the errors caused by a target that was not added. 

https://github.com/bytecodealliance/wasmtime/blob/fc7ef8d62e196df90522b26ed055bc21667cd19a/.github/actions/install-rust/action.yml#L67

This is used by CI, but it does not add `wasm32-wasip2` target. CI should have given better errors stating that a target is not found.

alexcrichton:
Could the wasip2 target be installed for just this one CI job perhaps instead of all of them?

PR #9794, opened 2024-12-11
pulley: Refactor conditional branches and table codegen 

alexcrichton:
This commit first fixed an issue with table access codegen to disable
spectre mitigations on Pulley targets like how spectre is disabled for
memory accesses as well. This unblocked many tests related to tables
which then led to a different error about a `trapnz` with an 8-bit value
not being supported.

In fixing `trapnz` with 8-bit values this PR went ahead and did a
general-purpose refactoring for how conditional branches are managed.
Previously conditional traps and conditional branches had some
duplicated logic and the goal was to unify everything. There is now a
single `Cond` which represents the condition of a conditional jump which
is used uniformly for all locations such as `select`, `brif`, and
`trap[n]z`. This new type represents all the sorts of conditional
branches that can be done in Pulley, for example integer comparisons and
whether or not a register is zero. This `Cond` type has various helpers
for printing it, inverting it, collecting operands, emission, etc.

The end result is that it's a bit wordy to work with `Cond` right now
due to the size of the variants but all locations working with
conditional traps are deduplicated and now it's just repetitive logic
rather than duplicated logic.

Putting all of this together gets a large batch of spec tests working.

I'll note that this does remove a feature where `trapnz` was turned into
nothing or an unconditional trap if the argument was a constant, but
that feels like an optimization perhaps best left for the middle-end
rather than doing it in the backend.


alexcrichton:
Note this is currently based on https://github.com/bytecodealliance/wasmtime/pull/9793

fitzgen:
> I'll note that this does remove a feature where `trapnz` was turned into
> nothing or an unconditional trap if the argument was a constant, but
> that feels like an optimization perhaps best left for the middle-end
> rather than doing it in the backend.

Agreed that ideally that would happen in the mid-end, but FWIW the mid-end doesn't have the power to do that currently. I think it is ultimately fine to remove this now tho, and if it is ever an issue in the future we can address it again at that point.

fitzgen (approved):

Issue #9766, opened 2024-12-09
winch(aarch64): Improve 32-bit {s,u}div


saulecabrera:
While reviewing https://github.com/bytecodealliance/wasmtime/pull/9762, I realized that in the case of 32-bit operands a sign-extension is required given that `cranelift-codegen`, only accepts 64-bit operands. 

The ARM64 documentation states [1], [2] that 32-bit operands for division are allowed.

[1]: https://developer.arm.com/documentation/ddi0596/2020-12/Base-Instructions/SDIV--Signed-Divide-
[2]: https://developer.arm.com/documentation/ddi0602/2024-09/Base-Instructions/UDIV--Unsigned-divide-

Introducing support for 32-bit operands means that in Winch we can skip the extension sequence when dealing with 32-bit operands.

As a side note, it seems that when migrating to ISLE [the intention was to add support for 32-bit operands](https://github.com/bytecodealliance/wasmtime/pull/3572/files#diff-bd652d5af8795961ba7456cf427fdae37bb82500e9f18fa6fc1621281bd1ab58L103). 

cc @alexcrichton / @cfallin I'm not entirely familiar with the development history of the Aarch64 backend, is there any other reason to be aware of when considering support for 32-bit operands?

MarinPostma:
I'd like to take a stab at that after I'm done with winch. Any guidance on how I should tackle that?

saulecabrera:
One approach could be to take a look at how similar instructions have been added in the past e.g., https://github.com/bytecodealliance/wasmtime/pull/8453

PR #9798, opened 2024-12-12
aarch64: support udiv for 32bit integers

MarinPostma:
This PR takes first steps towards completing #9766, by implementing support for 32bit unsigned division in cranelift and winch.
we had to take the following steps to enable that:

-  extend the ISLE lowering rules in cranelift to properly handle 32bits udiv, by have a rule that matches 64bit division, and another for anything else that fits 32bit. The 32 bits basically is the old 64bits rule but for 32bits.
- stop extending registers to 64bits in the ISLE rules.
- patch the emit code to emit the correct instruction for 32bits division.
- in winch: stop emitting extend operations for urem and udiv.
- Make `trap_if` size aware: this avoids having to perform extensions before `cbz`, shaving more instructions. This is reflected in the lowering rules for checking division by zero for udiv.

### DIFF
--- a/cranelift/codegen/src/isa/aarch64/inst.isle
+++ b/cranelift/codegen/src/isa/aarch64/inst.isle
@@ -870,7 +870,6 @@
        ;; *execute* the embedded `Inst`. (In the emitted code, we use the inverse
        ;; of this condition in a branch that skips the trap instruction.)
        (TrapIf
-        (size OperandSize)
         (kind CondBrKind)
         (trap_code TrapCode))
 
@@ -1993,10 +1992,10 @@
 (decl nzcv (bool bool bool bool) NZCV)
 (extern constructor nzcv nzcv)
 
-(decl cond_br_zero (Reg) CondBrKind)
+(decl cond_br_zero (Reg OperandSize) CondBrKind)
 (extern constructor cond_br_zero cond_br_zero)
 
-(decl cond_br_not_zero (Reg) CondBrKind)
+(decl cond_br_not_zero (Reg OperandSize) CondBrKind)
 (extern constructor cond_br_not_zero cond_br_not_zero)
 
 (decl cond_br_cond (Cond) CondBrKind)
@@ -3507,7 +3506,7 @@
       (side_effect
        (with_flags_side_effect flags
         (ConsumesFlags.ConsumesFlagsSideEffect
-         (MInst.TrapIf (operand_size $I64) (cond_br_cond cond) trap_code)))))
+         (MInst.TrapIf (cond_br_cond cond) trap_code)))))
 
 ;; Helpers for lowering `trapz` and `trapnz`.
 (type ZeroCond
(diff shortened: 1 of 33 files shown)


cranelift/codegen/src/isa/aarch64/inst.isle
```diff
@@ -870,6 +870,7 @@
        ;; *execute* the embedded `Inst`. (In the emitted code, we use the inverse
        ;; of this condition in a branch that skips the trap instruction.)
        (TrapIf
+        (size OperandSize)
```
MarinPostma, line 873: I think we mostly had two choices here:
- either make the size part of the `CondBrKind`, which is tighter, but yields more changes in the code
- or setting the size in the op itself (current approach), which required less changes

cfallin: I think I might prefer the first option actually -- I was a bit confused at first how the `OperandSize` would affect the `TrapIf` if we're just talking about conditional branch kinds like `b.lo`, `b.hi`, etc; until I realized that `CondBrKind::Zero(reg)` and `NonZero(reg)` exist and encode `cbz`/`cbnz`. If you don't mind, do you think you could try the refactor out? On the upside, it seems like it should get rid of a bunch of extraneous `OperandSize`s on branches emitted in various places that are condcode-based (e.g. in `abi.rs` above and in many of the emit-tests below).

cranelift/codegen/src/isa/aarch64/inst/regs.rs
```diff
@@ -179,6 +179,18 @@ pub fn pretty_print_reg(reg: Reg) -> String {
     show_reg(reg)
 }
 
+fn show_reg_sized(reg: Reg, size: OperandSize) -> String {
+    match reg.class() {
+        RegClass::Int => show_ireg_sized(reg, size),
+        RegClass::Float => todo!(),
```
cfallin, line 185: `todo!()` here -- can we just show the reg (and below for vector too) with `show_reg` for now?

cranelift/codegen/src/isa/aarch64/inst.isle
cfallin: Also, separately, if we modify `CondBrKind`, this is toward the direction that we want for supporting 32-bit zero/nonzero tests in regular conditional branches too. We don't have to implement that now (you can `unimplemented!()` it in the `CondBr` case; but if you wanted to, it would certainly be fine with me) but it's nice to lay the groundwork.

cfallin (commented):
Thanks! One thought below on the way that `TrapIf` is extended, but generally I like how this is turning out.

MarinPostma:
Thanks @cfallin, will do the refactor :+1:

cranelift/codegen/src/isa/aarch64/inst.isle
MarinPostma: This make sense to me, I'll refactor

cfallin (approved):
Looks great; thanks a bunch!

PR #9797, opened 2024-12-12
pulley: Get fused.wast passing 

alexcrichton:
Implement float-to-int bitcasts as well as the xor operation for
integers.

cc https://github.com/bytecodealliance/wasmtime/issues/9783

PR #9801, opened 2024-12-12
pulley: Get strings.wast test passing 

alexcrichton:
Needed the `imul` CLIF instruction to get implemented so 32/64-bit
multiplication have now been added.


alexcrichton:
Note that this is currently based on https://github.com/bytecodealliance/wasmtime/pull/9797

pulley/src/opcode.rs
```diff
@@ -26,7 +26,7 @@ macro_rules! define_opcode {
 
         impl Opcode {
             /// The value of the maximum defined opcode.
-            pub const MAX: u8 = define_opcode!( @max $( $name )* ) + 1;
+            pub const MAX: u8 = Opcode::ExtendedOp as u8;
```
alexcrichton, line 29: This change was needed because otherwise we overran the recursion limit and rustc asked for a `#![recursion_limit = "bigger"]` value which I wanted to avoid so I switch this to using the final enum here instead.

fitzgen (approved):

pulley/src/opcode.rs
fitzgen: Might be an off by one here? Does this need a `+ 1`?

alexcrichton: Double-checked, but this doesn't have the `+1` for now. (it's at least the same as before)

PR #9803, opened 2024-12-12
pulley: Fill out remaining 32/64-bit integer operations 

alexcrichton:
This required some extra plumbing to shepherd the precise reason why
signed division trapped to Wasmtime which is done through an extra
`TrapKind` side channel now added.

This then additionally fixes the signed remainder interpreter function
to return 0 on `MIN % -1` which is different from what Rust specifies
(which is to return `None` or panic).


alexcrichton:
Note this is currently built on https://github.com/bytecodealliance/wasmtime/pull/9801

pulley/src/interp.rs
```diff
@@ -1619,31 +1650,41 @@ impl OpVisitor for Interpreter<'_> {
...
-            None => self.done_trap::<crate::XDiv64U>(),
+            None => self.done_trap_kind::<crate::XDiv64U>(Some(TrapKind::DivideByZero)),
         }
     }
 
     fn xrem32_s(&mut self, operands: BinaryOperands<XReg>) -> ControlFlow<Done> {
         let a = self.state[operands.src1].get_i32();
         let b = self.state[operands.src2].get_i32();
-        match a.checked_rem(b) {
+        let result = if a == i32::MIN && b == -1 {
```
alexcrichton, line 1660: I'll note that this is an interesting case of a divergence in Rust semantics and WebAssembly semantics.

fitzgen (approved):
:+1:

PR #9806, opened 2024-12-12
pulley: Move `fp`/`lr` out of `XReg` set

alexcrichton:
This commit moves the `fp` and `lr` registers out of the `XReg` register set and into the `MachineState` per-VM. These are automatically modified and read with `push_frame` and `pop_frame`. Dedicated `xmov_{fp,lr}` instructions were added for use in Wasmtime's trampolines which directly read these registers.

cc #9747

### DIFF
--- a/cranelift/codegen/src/isa/pulley_shared/inst/regs.rs
+++ b/cranelift/codegen/src/isa/pulley_shared/inst/regs.rs
@@ -89,10 +89,10 @@ define_registers! {
     x_reg(24) => x24, writable_x24;
     x_reg(25) => x25, writable_x25;
     x_reg(26) => x26, writable_x26;
+    x_reg(27) => x27, writable_x27;
+    x_reg(28) => x28, writable_x28;
 
-    x_reg(27) => stack_reg, writable_stack_reg;
-    x_reg(28) => lr_reg, writable_lr_reg;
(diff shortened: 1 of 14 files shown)


alexcrichton:
I'd like to get there with `sp` as well too but that'll require more special care since it's used in memory-related instructions as a base for stack loads/stores.

cranelift/codegen/src/isa/pulley_shared/inst/regs.rs
```diff
@@ -89,10 +89,10 @@ define_registers! {
...
     x_reg(26) => x26, writable_x26;
+    x_reg(27) => x27, writable_x27;
+    x_reg(28) => x28, writable_x28;
 
-    x_reg(27) => stack_reg, writable_stack_reg;
-    x_reg(28) => lr_reg, writable_lr_reg;
-    x_reg(29) => fp_reg, writable_fp_reg;
+    x_reg(29) => stack_reg, writable_stack_reg;
     x_reg(30) => spilltmp_reg, writable_spilltmp_reg;
     x_reg(31) => spilltmp2_reg, writable_spilltmp2_reg;
```
fitzgen, line 97: Do we actually use spilltmp2 in pulley? AFAICT, we don't. Could free up one more GPR I think.

fitzgen (approved):
Nice!

PR #9808, opened 2024-12-12
pulley: Implement float arithmetic operations

alexcrichton:
Fill out enough to get `f32.wast` and `f64.wast` spec tests working. A minor ABI issue was discovered along the way which is also required to get a new test working on both 32 and 64-bit platforms.

cc #9783


fitzgen (approved):
Nice

alexcrichton:
@fitzgen mind reviewing the last commit? Turns out `a.trunc()` differs on MSVC and Linux (e.g. glibc) so I've extracted all the logic to a new crate to share between the runtime libcalls and Pulley. In theory should have been done already but it was getting complicated enough now that it was definitely worth it I think.

fitzgen (approved):
New commit looks good to me :+1:

PR #9811, opened 2024-12-13
Create wasmtime::Config from toml

ludfjig:
This PR makes it possible to create a `wasmtime::Config` from a toml file and is a first step toward resolving #8784. Overall, I think the path forward is as follows:

1. This PR: Enables loading `wasmtime::Config` from a toml file, by making `CommonOptions` derive `Deserialize`. For example, `opt_level` is deserialized from values 0, 1, 2, "s" (same as cli flags) instead of "None", "Speed", "SpeedAndSize". Same applies to `regalloc_algorithm`, `compiler` and `collector`. Feedback appreciated on this.
2. Combine cache-config toml with this new toml.
3. Have one source of truth for cli flags and config options.

Feedback greatly appreciated, especially regarding how to deal with `nn_graph`, as well `config_var` and `keyvalue_in_memory_data`, which are all marked `serde(skip)` for now

### DIFF
--- a/crates/cli-flags/src/lib.rs
+++ b/crates/cli-flags/src/lib.rs
@@ -1,9 +1,13 @@
 //! Contains the common Wasmtime command line interface (CLI) flags.
 
-use anyhow::Result;
+use anyhow::{Context, Result};
 use clap::Parser;
-use std::fmt;
-use std::time::Duration;
+use serde::Deserialize;
+use std::{
+    fmt, fs,
+    path::{Path, PathBuf},
+    time::Duration,
+};
 use wasmtime::Config;
 
 pub mod opt;
@@ -38,12 +42,17 @@ fn init_file_per_thread_logger(prefix: &'static str) {
 }
 
 wasmtime_option_group! {
-    #[derive(PartialEq, Clone)]
+    #[derive(PartialEq, Clone, Deserialize)]
+    #[serde(deny_unknown_fields)]
     pub struct OptimizeOptions {
         /// Optimization level of generated code (0-2, s; default: 2)
+        #[serde(default)]
+        #[serde(deserialize_with = "crate::opt::cli_parse_wrapper")]
         pub opt_level: Option<wasmtime::OptLevel>,
 
         /// Register allocator algorithm choice.
+        #[serde(default)]
+        #[serde(deserialize_with = "crate::opt::cli_parse_wrapper")]
         pub regalloc_algorithm: Option<wasmtime::RegallocAlgorithm>,
 
         /// Do not allow Wasm linear memories to move in the host process's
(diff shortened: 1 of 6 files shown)


crates/cli-flags/src/lib.rs
```diff
@@ -924,3 +961,144 @@ impl CommonOptions {
...
+        Self: Sized;
+}
+
+impl FromFile for Config {
+    fn from_file<P: AsRef<Path>>(path: P) -> Result<Self> {
+        let file_contents = fs::read_to_string(path.as_ref())?;
+        let mut common_options: CommonOptions = toml::from_str(&file_contents)?;
+        common_options.config(None)
+    }
+}
```
alexcrichton, line 977: This might perhaps be best as an inherent method on `CommonOptions` as an alternative constructor? That would still require the caller call `.config(...)` but that seems like a reasonable tradeoff to me.

crates/cli-flags/src/lib.rs
```diff
@@ -924,3 +961,144 @@ impl CommonOptions {
...
+
+pub trait FromFile {
+    fn from_file<P: AsRef<Path>>(path: P) -> Result<Self>
+    where
+        Self: Sized;
+}
+
+impl FromFile for Config {
+    fn from_file<P: AsRef<Path>>(path: P) -> Result<Self> {
+        let file_contents = fs::read_to_string(path.as_ref())?;
```
alexcrichton, line 973: Mind updating the error using `.with_context(...)` to attach the name of the file here and below?

alexcrichton (commented):
Thanks for this! I'm liking how this is shaping up pretty cleanly so far. Would you be up for integrating this into a CLI flag itself as well? For example `wasmtime --config cfg.toml` or something like that

crates/cli-flags/src/lib.rs
ludfjig: I agree

crates/cli-flags/src/lib.rs
ludfjig: addressed.

ludfjig:
> Thanks for this! I'm liking how this is shaping up pretty cleanly so far. Would you be up for integrating this into a CLI flag itself as well? For example `wasmtime --config cfg.toml` or something like that

Thanks for looking at this. I added a --config flag. Any suggestions on how to best test this?

alexcrichton:
This is looking good to me, was there something else you wanted to address before landing though?

ludfjig:
Sorry for the delay @alexcrichton. I was wondering if I need to update any documentation, and if the current tests are sufficient. If not, I am ready for merge/reviews.

ludfjig:
@alexcrichton I had some problems getting the test suite with the i686 target to pass, but it should hopefully be fixed now by using `wasm-max-stack=0` which is explicitly disallowed and will cause an error.

alexcrichton:
Let's see if that's spurious...

ludfjig:
@alexcrichton no worries. The only question I have is about the fmt::Display impl for CommonOptions. If configured using a `--config <path>`, then the output will potentially be incorrect, since it will look like all options were provided on cli instead of from a toml. Is this acceptable?

alexcrichton:
I think that's reasonable yeah, the Display implementation here is sort of best-effort so printing out `--config $path` for when the field is `Some` I think is good enough

PR #9818, opened 2024-12-13
pulley: Disable the WebAssembly `threads` proposal

alexcrichton:
Forcibly disable the `threads` proposal for Pulley. This is done because I don't believe there's any way that we can, in Rust, implement non-atomic loads and stores in a manner that isn't undefined behavior. Until this is available leave this proposal as flagged as unsupported.


alexcrichton:
I've also [opened a topic on the rust-lang/rust zulip for this](https://rust-lang.zulipchat.com/#narrow/channel/213817-t-lang/topic/Writing.20a.20WebAssembly.20interpreter.20in.20Rust.20for.20wasm.20threads/near/488915886)

fitzgen (approved):

Issue #9799, opened 2024-12-12
Allow `wasmtime serve` command's LogStream prefix to be customizable


jzhn:
#### Feature

Currently, the `wasmtime serve` always prefixes the stdout and stderr output of http handlers. This prevents users from implementing structured logs like json.

##### Expected behaviour

logging from http handler in json will be logged by `wasmtime serve` as-is, without any prefixes.

```
{"timestamp":"2024-12-12T02:01:04.794485Z","level":"INFO","fields":{"message":"hello world","log.target":"webservice","log.module_path":"webservice","log.file":"webservice/src/lib.rs","log.line":237},"target":"webservice"}
```

##### Actual behaviour

handler logs are prefixed with `stdout [{req_id}] :: `.

```
stdout [176] :: {"timestamp":"2024-12-12T02:01:04.794485Z","level":"INFO","fields":{"message":"hello world","log.target":"webservice","log.module_path":"webservice","log.file":"webservice/src/lib.rs","log.line":237},"target":"webservice"}
```

#### Benefit

If `wasmtime serve` allows customizing log prefixes, it would allow users to implement structured logging at their http handlers, like json, so the stdout and stderr logs from wasmtime serve can be easily parsed by popular log management tools like [Grafana Loki](https://grafana.com/docs/loki/latest/send-data/promtail/stages/json/) and [AWS CloudWatch](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/CWL_AnalyzeLogData-discoverable-fields.html#CWL_AnalyzeLogData-discoverable-JSON-logs)

#### Implementation

Hopefully it's as simple as adding a coupld CLI flags, to make https://github.com/bytecodealliance/wasmtime/blob/9aa048b0eb0caf2bbd85a24bdada33928e42ad3b/src/commands/serve.rs#L141-L149 configurable.

#### Alternatives

No workarounds/alternatives I could think of.

alexcrichton:
Sounds reasonable to me to have some CLI flags for this yeah, would you be interested in sending a PR? Perhaps we could start out with a flag to just turn off the prefixes?

jzhn:
@alexcrichton : I'm fairly new to Rust and WASM, but with your encouragement I've created a PR for this. Please take a look and let me know your feedbacks. Thank you!

alexcrichton:
Resolved by https://github.com/bytecodealliance/wasmtime/pull/9821

PR #9821, opened 2024-12-14
allow customizing log prefixes for wasmtime serve command

jzhn:
### What

As discussed in feature request: https://github.com/bytecodealliance/wasmtime/issues/9799, `wasmtime serve` command always prefixes stdout and stderr logs of wasi-http handler with hardcoded format. This prevents user to implement consistent logging formats like json logging.

This PR adds an optional flags, `--no-logging-prefix`, to stop adding log prefixes. When the new flags are not provided, existing hehaviour is preserved.

### Test

- Added test case to cover new flags
- [Existing test case](https://github.com/bytecodealliance/wasmtime/blob/da93f647974518e111561b3b451ef8c4a576bf20/tests/all/cli_tests.rs#L1868) covers the existing behaviour when flags are absent.

### Alternatives

I know adding flags to CLI creates long term promise, so I'm trying to be careful here and have considered a few alternatives.

#### boolean flags vs string

(original implementation in this PR) To allow better flexibility and future proof, we could add two string flags, `--logging-prefix-stdout` and `--logging-prefix-stderr`, which allows users to customize the prefix, and possibly removing the prefix by providing empty strings. This enables future possibilities like allow users to specify log prefixes with templates, as the `{ req_id }` could be useful. __this is replaced with simple boolean flag of `--no-logging-prefix` to avoid over engineering.__

#### Debug options 

The existing `DebugOptions` already contains some customization about logging
https://github.com/bytecodealliance/wasmtime/blob/da93f647974518e111561b3b451ef8c4a576bf20/crates/cli-flags/src/lib.rs#L241-L244

While we could add the flags there, I feel the new flags in this PR are specific to `wasmtime serve` command. So adding them to the common flags might confuse people for other commands.

### DIFF
--- a/src/commands/serve.rs
+++ b/src/commands/serve.rs
@@ -88,8 +88,17 @@ pub struct ServeCommand {
     run: RunCommon,
 
     /// Socket address for the web server to bind to.
-    #[arg(long = "addr", value_name = "SOCKADDR", default_value_t = DEFAULT_ADDR )]
+    #[arg(long = "addr", value_name = "SOCKADDR", default_value_t = DEFAULT_ADDR)]
     addr: SocketAddr,
+
+    /// The custom prefix of stdout logs from wasi-http handlers.
+    /// if unspecified, default log of 'stdout [{req_id}] :: ' will be used.
+    #[arg(long = "logging-prefix-stdout", value_name = "PREFIX")]
+    logging_prefix_stdout: Option<String>,
+    /// The custom prefix of stderr logs from wasi-http handlers.
+    /// if unspecified, default log of 'stderr [{req_id}] :: ' will be used.
+    #[arg(long = "logging-prefix-stderr", value_name = "PREFIX")]
+    logging_prefix_stderr: Option<String>,
 
     /// The WebAssembly component to run.
     #[arg(value_name = "WASM", required = true)]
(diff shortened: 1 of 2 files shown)


alexcrichton (commented):
Thanks! I think the documentation may want to be updated though because using `{req_id}` could be confusing as the prefix here doesn't actually support interpolation of `req_id` or other variables. If there's no need to have acustom prefixes it might also be reasonable to have a single flag to disable prefixes perhaps?

jzhn:
> Thanks! I think the documentation may want to be updated though because using `{req_id}` could be confusing as the prefix here doesn't actually support interpolation of `req_id` or other variables. If there's no need to have acustom prefixes it might also be reasonable to have a single flag to disable prefixes perhaps?

Thanks for the review. I see that you are suggesting two directions, one to keep the custom prefix flags but update the documentation to remove mention of `{req_id}` to avoid confusion, the other to replace those custom prefix flags and use a simple boolean flag like `--no-logging-prefix` to disable the prefix.

I'm fine with either approach, so please just let me know which is your preference and I'll update accordingly. thanks :)

alexcrichton:
Ah yes sorry, I also don't have much of a preference. Since your goal is to just disable the prefix why don't we add that for now and consider adding customizable ones later.

alexcrichton (approved):

Issue #9828, opened 2024-12-15
[DWARF] Inhertiance `DW_TAG`s dropped by GC


SingleAccretion:
Reproduction:
1\) Compile the following C++ to WASM with `-g`:
```cpp
struct Base { int BaseValue; };
struct Derived : public Base { int DerivedValue; };

int main()
{
    Derived x;
    x.BaseValue = 2;
    x.DerivedValue = 1;
    return 0;
}
```
2\) Run under wasmtime+LLDB:
```
> lldb wasmtime -Ddebug-info -Oopt-level=0 main.wasm
> (lldb) b main
> (lldb) ...
> (lldb) v x.BaseValue
```
Expected result:
```
(int) x.BaseValue = 1
```
Actual result:
```
error: "BaseValue" is not a member of "(Derived) x"
```
Cause: DI GC drops correctness-bearing `DW_TAG_inheritance` DIEs.

Working on a fix...

PR #9829, opened 2024-12-15
[DWARF] Make the GC of child tags conservative by default

SingleAccretion:
Only allow for GCing of a known set of 'standalone' or 'structural' DIEs.

This is still likely not _fully_ correct, but at least it is _more_ correct than what was there before.

Fixes https://github.com/bytecodealliance/wasmtime/issues/9828.


fitzgen (approved):
Makes sense, thanks!

PR #9830, opened 2024-12-16
pulley: Lower umulhi and smulhi in CLIF 

alexcrichton:
This is not directly reachable from wasm but can be created through
optimizations.


alexcrichton:
Note this is built on https://github.com/bytecodealliance/wasmtime/pull/9825

fitzgen (approved):
👍

PR #9831, opened 2024-12-16
Implement initial-cwd for wasi:cli/environment

### LINKED ISSUE #9695: implement initial-cwd for wasi:cli/environment
For https://github.com/WebAssembly/wasi-cli/blob/main/wit/environment.wit, the `initial-cwd` implementation currently just returns `None`. Adding this issue highlighted by the `FIXME`:

https://github.com/bytecodealliance/wasmtime/blob/66910067642ce2ddf5509845306508f89a24fc9e/crates/wasi/src/host/env.rs#L14-L17

stephenmuss:
This implements initial-cwd for wasi:cli/environment.

My biggest struggle was finding the appropriate places to add tests. Any guidance would be appreciated. Thank you.

Closes #9695

### DIFF
--- a/crates/cli-flags/src/lib.rs
+++ b/crates/cli-flags/src/lib.rs
@@ -424,6 +424,8 @@ wasmtime_option_group! {
         pub config_var: Vec<KeyValuePair>,
         /// Preset data for the In-Memory provider of WASI key-value API.
         pub keyvalue_in_memory_data: Vec<KeyValuePair>,
+        /// Set the working directory to the path provided.
+        pub cwd: Option<String>,
     }
 
     enum Wasi {
--- a/crates/wasi/src/ctx.rs
+++ b/crates/wasi/src/ctx.rs
@@ -33,6 +33,7 @@ use wasmtime::component::ResourceTable;
 /// wasi.arg("./foo.wasm");
 /// wasi.arg("--help");
 /// wasi.env("FOO", "bar");
+/// wasi.cwd("/home/steve/foo");
 ///
 /// let wasi: WasiCtx = wasi.build();
 /// ```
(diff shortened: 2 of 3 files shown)


stephenmuss:
I assume I should be able to use something like this in `crates/test-programs/bin/*.rs` to test

```rs
fn main() -> std::io::Result<()> {
    let cwd = std::env::current_dir()?;
    assert_eq!(cwd.to_str(), Some("/tmp"));
    Ok(())
}
```

and run with `-S cwd=/tmp`.

Is that understanding correct? 

However, I always get `Some("/")` indicating that I still need to set the path somewhere I've missed, presumably in `src/commands/run.rs` and `src/commands/serve.rs` etc.

Timmmm:
Does the WASI API actually exist yet? [Your issue about it is still open](https://github.com/WebAssembly/wasi-filesystem/issues/24). Also as far as I can tell `initial_cwd()` isn't actually called from anywhere (though I may have missed it if there's some proc macro magic or something).

alexcrichton:
It exists insofar as the `*.wit` function [exists and is stable](https://github.com/WebAssembly/wasi-cli/blob/e4c71d8cbc6c1cefc5d43f25f83348f90c9d2346/wit/environment.wit#L18-L21). The issue you linked is from a long time ago and was intended for `*.witx` or effectively WASIp1. WASIp1 isn't going to change any more but this function was added to enhance the functionality and largely solve that issue in WASIp2.

To the extent the function exists, however, I'm also not familar with any integration with it. For example `wasi-libc` does not call the function and try to use the return value (nor does Rust's libstd). Hooking it up to Wasmtime, however, is step 1 in filling out ecosystem support for it though as otherwise it can't even be tested.

alexcrichton:
This ended up being done in https://github.com/bytecodealliance/wasmtime/pull/13468

PR #9832, opened 2024-12-16
pulley: Implement SIMD `splat` instruction

alexcrichton:
Gets a few spec tests and CLIF tests passing

cc #9783

### DIFF
--- a/cranelift/codegen/src/isa/pulley_shared/lower.isle
+++ b/cranelift/codegen/src/isa/pulley_shared/lower.isle
@@ -771,3 +771,12 @@
 
 (rule 0 (lower (has_type (fits_in_32 _) (iabs a))) (pulley_xabs32 (sext32 a)))
 (rule 1 (lower (has_type $I64 (iabs a))) (pulley_xabs64 a))
+
+;;;; Rules for `split` ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(rule (lower (has_type $I8X16 (splat a))) (pulley_vsplatx8 a))
+(rule (lower (has_type $I16X8 (splat a))) (pulley_vsplatx16 a))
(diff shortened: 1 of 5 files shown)


cranelift/codegen/src/isa/pulley_shared/lower.isle
```diff
@@ -771,3 +771,12 @@
 
 (rule 0 (lower (has_type (fits_in_32 _) (iabs a))) (pulley_xabs32 (sext32 a)))
 (rule 1 (lower (has_type $I64 (iabs a))) (pulley_xabs64 a))
+
+;;;; Rules for `split` ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
```
abrown, line 775:
```suggestion
;;;; Rules for `splat` ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
```

pulley/src/interp.rs
```diff
@@ -2695,6 +2695,42 @@ impl OpVisitor for Interpreter<'_> {
         self.state[dst].set_u128(val);
         ControlFlow::Continue(())
     }
+
+    fn vsplatx8(&mut self, dst: VReg, src: XReg) -> ControlFlow<Done> {
+        let val = self.state[src].get_u32() as u8;
```
abrown, line 2700: We may need `get_u8` and `get_u16` eventually?

abrown: ...just mulling over whether discarding the upper bits here is going to be problematic &mdash; probably not?

abrown (approved):
LGTM!

pulley/src/interp.rs
alexcrichton: I've been wondering the same yeah, I'll probably add these in a follow-up

PR #9834, opened 2024-12-16
pulley: Implement `return_call` instructions

alexcrichton:
This commit fleshes out the Cranelift lowerings of tail calls which gets the wasm tail call proposal itself working on Pulley. Most of the bits and pieces here were copied over from the riscv64 backend and then edited to suit Pulley.

### DIFF
--- a/tests/misc_testsuite/tail-call/loop-across-modules.wast
+++ b/tests/misc_testsuite/tail-call/loop-across-modules.wast
@@ -42,5 +42,5 @@
   (start $start)
 )
 
-(assert_return (invoke $B "g" (i32.const 100000000))
+(assert_return (invoke $B "g" (i32.const 100000))
                (i32.const 42))
--- a/cranelift/codegen/src/isa/pulley_shared/abi.rs
+++ b/cranelift/codegen/src/isa/pulley_shared/abi.rs
@@ -611,12 +611,45 @@ where
     P: PulleyTargetKind,
 {
     pub fn emit_return_call(
-        self,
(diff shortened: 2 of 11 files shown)


tests/misc_testsuite/tail-call/loop-across-modules.wast
```diff
@@ -42,5 +42,5 @@
   (start $start)
 )
 
-(assert_return (invoke $B "g" (i32.const 100000000))
+(assert_return (invoke $B "g" (i32.const 100000))
```
alexcrichton, line 45: I wasn't sure how best to handle this, basically the test takes forever on an unoptimized interpreter (aka `cargo test`) so I turned this down to be more reasonable, but that means this probably isn't testing what it originally wanted to test since the number is significantly smaller.

fitzgen: I think this is probably fine as-is, since 100,000 is still pretty big, but we could add an import like `(import "env" "tail-recursion-count" (func (result i32))` and configure it appropriately in the wast utils crate.

fitzgen (approved):
:+1:

tests/misc_testsuite/tail-call/loop-across-modules.wast
alexcrichton: I'm a bit wary to extend the namespace of builtins just yet so I'm follow your gut and leave this at 100k for now. We might want a little more "official-ness" around extending the set of names `wast` can import by default than a one-off thing such as this.

PR #9836, opened 2024-12-16
pulley: Assert calling-conventions match instructions

alexcrichton:
Use the "tail" calling convention for pulley<->pulley calls and "systemv" is used for calling the host.


alexcrichton:
Hm actually no this doesn't work because of `*.clif` filetests which aren't all using `tail` and `systemv`, so I'm going to close this.

PR #9837, opened 2024-12-16
Run the full test suite on 32-bit platforms

alexcrichton:
This commit switches to running the full test suite in its entirety (`./ci/run-tests.sh`) on 32-bit platforms in CI in addition to 64-bit platforms. This notably adds i686 and armv7 as architectures that are tested in CI.

Lots of little fixes here and there were applied to a number of tests. Many tests just don't run on 32-bit platforms or a platform without Cranelift support, and they've been annotated as such where necessary. Other tests were adjusted to run on all platforms a few minor bug fixes are here as well.

### DIFF
--- a/cranelift/filetests/src/function_runner.rs
+++ b/cranelift/filetests/src/function_runner.rs
@@ -611,6 +611,7 @@ extern "C" fn __cranelift_x86_pshufb(a: __m128i, b: __m128i) -> __m128i {
 }
 
 #[cfg(test)]
+#[cfg(target_pointer_width = "64")] // 32-bit platforms not supported at this time
 mod test {
     use super::*;
     use cranelift_reader::{parse_functions, parse_test, ParseOptions};
--- a/cranelift/filetests/src/function_runner.rs
+++ b/cranelift/filetests/src/function_runner.rs
@@ -611,7 +611,6 @@ extern "C" fn __cranelift_x86_pshufb(a: __m128i, b: __m128i) -> __m128i {
 }
 
 #[cfg(test)]
-#[cfg(target_pointer_width = "64")] // 32-bit platforms not supported at this time
 mod test {
     use super::*;
     use cranelift_reader::{parse_functions, parse_test, ParseOptions};
@@ -622,6 +621,10 @@ mod test {
 
     #[test]
     fn nop() {
+        // Skip this test when cranelift doesn't support the native platform.
+        if cranelift_native::builder().is_err() {
+            return;
+        }
         let code = String::from(
             "
             test run
@@ -656,6 +659,10 @@ mod test {
 
     #[test]
     fn trampolines() {
+        // Skip this test when cranelift doesn't support the native platform.
+        if cranelift_native::builder().is_err() {
+            return;
+        }
         let function = parse(
             "
             function %test(f32, i8, i64x2, i8) -> f32x4, i64 {
--- a/cranelift/filetests/src/test_run.rs
+++ b/cranelift/filetests/src/test_run.rs
@@ -50,12 +50,12 @@ fn build_host_isa(
 /// Checks if the host's ISA is compatible with the one requested by the test.
 fn is_isa_compatible(
     file_path: &str,
-    host: &anyhow::Result<OwnedTargetIsa>,
+    host: Option<&dyn TargetIsa>,
     requested: &dyn TargetIsa,
 ) -> Result<(), String> {
     let host_triple = match host {
-        Ok(host) => host.triple().clone(),
-        Err(_) => target_lexicon::Triple::host(),
+        Some(host) => host.triple().clone(),
+        None => target_lexicon::Triple::host(),
     };
     // If this test requests to run on a completely different
     // architecture than the host platform then we skip it entirely,
@@ -101,8 +101,12 @@ fn is_isa_compatible(
             None => unimplemented!("ISA flag {} of kind {:?}", req_value.name, req_value.kind()),
         };
         let host_isa_flags = match host {
-            Ok(host) => host.isa_flags(),
-            Err(e) => return Err(format!("{e:?}")),
+            Some(host) => host.isa_flags(),
+            None => {
+                return Err(format!(
+                    "host not available on this platform for isa-specific flag"
+                ))
+            }
         };
         let available_in_host = host_isa_flags
             .iter()
@@ -229,8 +233,8 @@ impl SubTest for TestRun {
         }
 
         // Check that the host machine can run this test case (i.e. has all extensions)
-        let host_isa = build_host_isa(true, flags.clone(), vec![]);
-        if let Err(e) = is_isa_compatible(file_path, &host_isa, isa.unwrap()) {
+        let host_isa = build_host_isa(true, flags.clone(), vec![]).ok();
+        if let Err(e) = is_isa_compatible(file_path, host_isa.as_deref(), isa.unwrap()) {
             log::info!("{}", e);
             return Ok(());
         }
--- a/cranelift/filetests/src/test_run.rs
+++ b/cranelift/filetests/src/test_run.rs
@@ -34,28 +34,33 @@ fn build_host_isa(
     infer_native_flags: bool,
     flags: settings::Flags,
     isa_flags: Vec<settings::Value>,
-) -> OwnedTargetIsa {
+) -> anyhow::Result<OwnedTargetIsa> {
     let mut builder = cranelift_native::builder_with_options(infer_native_flags)
-        .expect("Unable to build a TargetIsa for the current host");
+        .map_err(|e| anyhow::Error::msg(e))?;
 
     // Copy ISA Flags
     for value in isa_flags {
-        builder.set(value.name, &value.value_string()).unwrap();
+        builder.set(value.name, &value.value_string())?;
     }
 
-    builder.finish(flags).unwrap()
+    let isa = builder.finish(flags)?;
+    Ok(isa)
 }
 
 /// Checks if the host's ISA is compatible with the one requested by the test.
 fn is_isa_compatible(
     file_path: &str,
-    host: &dyn TargetIsa,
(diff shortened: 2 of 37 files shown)


cranelift/filetests/src/function_runner.rs
```diff
@@ -611,6 +611,7 @@ extern "C" fn __cranelift_x86_pshufb(a: __m128i, b: __m128i) -> __m128i {
 }
 
 #[cfg(test)]
+#[cfg(target_pointer_width = "64")] // 32-bit platforms not supported at this time
```
cfallin, line 614: The non-support here has less to do with "32-bit" than "no Cranelift backend for the host architecture", right? I wonder if we could add some build-script logic to add a `cranelift-has-host-backend` feature and then `cfg` on that here? My concern is otherwise that a bunch of "exclude 32-bit" conditions scattered through tests get missed if/when we do add a native backend for say arm32 or riscv32.

cranelift/filetests/src/test_run.rs
```diff
@@ -34,28 +34,33 @@ fn build_host_isa(
...
-    builder.finish(flags).unwrap()
+    let isa = builder.finish(flags)?;
+    Ok(isa)
 }
 
 /// Checks if the host's ISA is compatible with the one requested by the test.
 fn is_isa_compatible(
     file_path: &str,
-    host: &dyn TargetIsa,
+    host: &anyhow::Result<OwnedTargetIsa>,
```
cfallin, line 53: This is fairly counterintuitive -- could we take an `Option` or a custom enum here? `Err` turning into some other meaning below seems to go against what I'd expect `Err` to mean.

cranelift/src/run.rs
```diff
@@ -126,6 +126,7 @@ fn create_target_isa(isa_spec: &IsaSpec) -> Result<OwnedTargetIsa> {
 }
 
 #[cfg(test)]
+#[cfg(target_pointer_width = "64")] // 32-bit platforms not supported yet
```
cfallin, line 129: likewise here as above re: "has native backend" feature. Or alternatively, allowlist the natively supported ISAs; then at least when we add a new ISA we can grep for all the places ISAs are mentioned...

cfallin (commented):
I'll defer to @fitzgen on the Pulley interpreter changes but a few thoughts on the conditional logic -- my main thought is that I'd prefer we didn't conflate "32-bit host" with "has no backend" and instead of a specific feature (or other method) meaning "we have a native codegen backend for the host platform"; seems less likely to cause unintentional issues later.

cranelift/filetests/src/function_runner.rs
alexcrichton: I'll try to think of something better yeah, but I also don't want to go too too overboard here because the tests handled in this PR are about 1% of our tests and are probably all largely redundant with all other tests we do from a "run this on all platforms" perspective. I definitely agree with what you're thinking though, I'm also not a fan of writing these #[cfg] and I like the idea of a `build.rs`-based thing (although it may be difficult to propagate that out beyond cranelift-codegen...)

cfallin: Alternate idea: could we add a method to `Engine` (or `Config` or ...) that lets the user query whether Wasmtime has a native compilation backend? Then the affected tests could early-out succeed. Such a query is probably useful for other cases too...

cfallin: (Likewise at the `cranelift-native` level, or whatever other level, for Cranelift.)

alexcrichton: It might help yeah, but the problem is that we have tests all over the place in all sorts of crates that need this logic. That leads to either duplication of the logic, moving tests, or something like that. Basically it's just really difficult wrangling a ubiquitous dependency on platform-specific logic when testing across platforms.

crates/wasmtime/build.rs
```diff
@@ -3,6 +3,16 @@ fn main() {
...
+
+    // Flag pulley as enabled unconditionally on 32-bit targets to ensure that
+    // wasm is runnable by default like it is on other 64-bit native platforms.
+    // Note that this doesn't actually enable the Cargo feature, it just changes
+    // the cfg's passed to the crate, so for example care is still taken in
+    // `Cargo.toml` to handle pulley-specific dependencies on 32-bit platforms.
+    let target_pointer_width = std::env::var("CARGO_CFG_TARGET_POINTER_WIDTH").unwrap();
+    if target_pointer_width == "32" {
+        println!("cargo:rustc-cfg=feature=\"pulley\"");
+    }
```
fitzgen, line 15: Here I think it is worth explicitly checking against our ISAs supported by Cranelift.

crates/wasmtime/Cargo.toml
```diff
@@ -97,6 +97,14 @@ rustix = { workspace = true, optional = true }
 [target.'cfg(target_arch = "s390x")'.dependencies]
 psm = { workspace = true, optional = true }
 
+# On 32-bit platforms where Cranelift currently doesn't have any supported host
+# unconditionally pull in the `pulley-interpreter` crate. Runtime support is
+# enabled via `build.rs` which prints `feature="pulley"` by default here and
+# compilation support, if enabled via the `cranelift` feature, is handled in the
+# `wasmtime-cranelift` crate enabling Pulley on 32-bit platforms.
+[target.'cfg(target_pointer_width = "32")'.dependencies]
+pulley-interpreter = { workspace = true }
```
fitzgen, line 106: Shouldn't this happen automatically when the build script enables the feature?

fitzgen (approved):
Just looked at the runtime and pulley changes themselves since Chris looked at the other parts.

cranelift/filetests/src/function_runner.rs
fitzgen: I guess we could push and dedupe the check into some common dependency, which is a bit annoying, but would otherwise work. `wasmtime-environ`?

There is someone actively working on a riscv32 backend for cranelift, they've been commenting on zulip, so it would be a shame to build in new cranelift-doesn't-support-32-bit assumptions at this point...

alexcrichton:
Ok I've been thinking about how to clean this up. For the tests that @cfallin flagged I added logic to them all to check `cranelift_native::builder()` and see whether that's an error. That indicates whether the native platform is supported by Cranelift and tests will early-return if that's an error.

@fitzgen for your points though I'm not sure how to make it better. I'll note that the `cargo:rustc-cfg` in the build script is not enabling the `pulley` feature, it's just affecting the `--cfg` flags to the compiler. Accurately reflecting Cranelift's supported architecture would mean having:

```toml
[target.'cfg(not(any(target_arch = "x86_64", target_arch = "aarch64", target_arch = "s390x", target_arch = "riscv64")))'.dependencies]
pulley-interpreter = { workspace = true }
```

And then reflecting that same condition into `build.rs`

This feels "zoomed out" enough to me that leaving `target_pointer_width = "32"` as a proxy there for now is probably sufficient, but I want to confirm with you and/or see if you know of a way to simplify this.

fitzgen:
That cfg is certainly annoying but seems like it is probably better than baking in new assumptions that we know are unlikely to be true for too long?

alexcrichton:
To clarify though the precise assumption here is that 32-bit platforms don't have a Cranelift backend so pulley is pulled in as a dependency. This doesn't actually change any logic internally within Cranelift or Wasmtime, it just compiles in Pulley by default without a way to turn that off. That's only a problem with respect to binary size.

If all of Cranelift's currently supported architectures are listed out here then that's duplicated in two places in Wasmtime (not counting the ones in Cranelift) and they're just going to get blindly added to in the future. Alternatively if Cranelift gets a 32-bit backend then that would be added as an exception to the cfg as-is when binary size becomes a problem.

To me this is six-to-one-half-dozen-or-the-other where I have a preference for what's as-is as it's a managable cfg that's by-default workable for most platforms (it doesn't work for 64-bit platforms without a Cranelift backend). Listing out architectures individually feels "just as bad" to me basically.

The main alternative to me is that we leave pulley disabled by default on all platforms, regardless of Cranelift support. That feels like a worse experience to me though because it means that Wasmtime only works by default on Cranelift-supported platforms, which if someone's new to Wasmtime they might not understand to enable the Pulley backend.

alexcrichton:
ping @fitzgen on ^

fitzgen (approved):
I don't feel strongly enough on the issue of `cfg`s that I want to hold this up from landing. I do think writing out all the targets is slightly better than using pointer-width=32, but even better is running the full test suite on 32-bit platforms.

alexcrichton:
Doing a bit of thinking on this, @fitzgen mind looking at the last commit? (https://github.com/bytecodealliance/wasmtime/pull/9837/commits/b476a2e7e414413a8c6a72de9c85db77477f93d6) It's a different strategy for enabling Pulley but it pulls the on-vs-off out of `Cargo.toml` which makes it manageable to list out targets IMO (and it's also going to be hard to forget to update because if it's not updated then the native Cranelift backend isn't used by default which would almost surely get noticed)

alexcrichton:
I'll also call out the now-currently-last commit in this PR which is new and (somewhat) substantial. It resolves a segfault on CI which I could reproduce locally and was one of the items listed in https://github.com/bytecodealliance/wasmtime/issues/9747 namely that the setjmp/longjmp implementation for Pulley needed to restore callee-save state and this was the first time that turned up.

fitzgen (approved):
Latest commit seems like a good balance.

Did you happen to verify that code size didn't increase under this set up when not using Pulley?

alexcrichton:
Code size is expected to increase due to the Pulley backend being unconditionally enabled in Cranelift now, but otherwise for Wasmtime itself I've confirmed that Pulley support in Wasmtime is not compiled in on x86_64 (I introduced a compile failure in the pulley-specific integration and it didn't get compiled with default features)

Scutua (approved):

PR #9842, opened 2024-12-17
pulley: Get `simd_conversions.wast` test working

alexcrichton:
Lots of narrowing/extending/conversion-related opcodes implemented. Note that these opcodes are all in the "extended" namespace as the 1-byte namespace has started to overflow.


fitzgen (approved):
Looks great!

PR #9845, opened 2024-12-17
Optimize constant offset loads/stores in translation

alexcrichton:
This commit implements an optimization when lowering Wasm bytecode to CLIF to skip a bounds check when the offset in memory is statically known. This comes up in C/C++/Rust when `static` memory is addressed for example where LLVM emits an `i32.const 0` as the base address and puts the address of the variable in the `offset` instruction field. This isn't necessary for 64-bit platforms but when explicit bounds-checks are required this can help to eliminate a constant-true bounds-check.

### DIFF
--- a/cranelift/codegen/src/ir/immediates.rs
+++ b/cranelift/codegen/src/ir/immediates.rs
@@ -10,6 +10,7 @@ use core::fmt::{self, Display, Formatter};
 use core::ops::{Add, BitAnd, BitOr, BitXor, Div, Mul, Neg, Not, Sub};
 use core::str::FromStr;
 use core::{i32, u32};
+use cranelift_entity::{Signed, Unsigned};
 #[cfg(feature = "enable-serde")]
 use serde_derive::{Deserialize, Serialize};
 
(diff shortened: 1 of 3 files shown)


tests/disas/pulley/memory-inbounds.wat
```diff
@@ -0,0 +1,31 @@
...
+  (memory 1)
+
+  (func (result i32)
+    (i32.const 0)
+    i32.load
+  )
+
+  (func (result i32)
+    (i32.const 100)
+    i32.load
```
sunfishcode, line 14: Could you add a few tests with interesting non-zero offsets as well?

tests/disas/pulley/memory-inbounds.wat
```diff
@@ -0,0 +1,31 @@
+;;! target = "pulley64"
+;;! test = "compile"
+
+(module
+  (memory 1)
```
sunfishcode, line 5: Would it make sense to specify a max, like`(memory 1 2)`, so that this can test for loads that are in-bounds on the minimum but not the maximum?

fitzgen: > so that this can test for loads that are in-bounds on the minimum but not the maximum?

I think this should be "in bounds of the maximum but not the minimum", but yeah +1

fitzgen (approved):
LGTM modulo a nitpick below and Dan's comments as well.

cranelift/codegen/src/ir/immediates.rs
```diff
@@ -107,6 +108,25 @@ impl Imm64 {
         let sign_extended = (self.0 << delta) >> delta;
         Imm64(sign_extended)
     }
+
+    /// Zero extend this immediate as if it were a signed integer of the given
+    /// power-of-two width.
```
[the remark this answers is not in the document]
alexcrichton: Oops this is a copy/paste typo

PR #9847, opened 2024-12-17
Move memories first in `VMContext`

alexcrichton:
This commit shuffles the fields of `VMContext` to move memories first above all other items. This is done to ensure that the offset to memory-related fields is one of the smaller offsets and ideally fits within an 8-byte offset which helps fit into some special cases on various ISAs. For example x64 has a smaller instruction encoding for 8-bit offsets than 32-bit offsets, and I'm planning on doing a similar optimization for Pulley.

### DIFF
--- a/crates/environ/src/vmoffsets.rs
+++ b/crates/environ/src/vmoffsets.rs
@@ -15,13 +15,13 @@
 //      gc_heap_data: *mut T, // Collector-specific pointer
 //      store: *mut dyn Store,
 //      type_ids: *const VMSharedTypeIndex,
-//      imported_functions: [VMFunctionImport; module.num_imported_functions],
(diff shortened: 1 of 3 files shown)


crates/environ/src/vmoffsets.rs
```diff
@@ -15,13 +15,13 @@
 //      gc_heap_data: *mut T, // Collector-specific pointer
 //      store: *mut dyn Store,
 //      type_ids: *const VMSharedTypeIndex,
+//      imported_memories: [VMMemoryImport; module.num_imported_memories],
```
sunfishcode, line 18: Could you add a comment here saying somthing like "Put memories close to the front, so instructions referencing them can use smaller immediate fields, as load/store instructions tend to be frequent"?

sunfishcode (approved):

PR #9850, opened 2024-12-18
cranelift: 32bit div_s, rem_u, rem_s for aarch64

### LINKED ISSUE #9766: winch(aarch64): Improve 32-bit {s,u}div

MarinPostma:
followup to #9798 where I did the ground work for 32bit division, this PR extends 32bits optimizations to rem_u, rem_s and div_s.

This should close #9766.


cranelift/codegen/src/isa/aarch64/lower.isle
```diff
@@ -1028,43 +1028,48 @@
...
+        (Signed)
+        (Unsigned)))
 
-(rule udiv (lower (has_type (fits_in_32 ty) (udiv x y)))
-      (a64_udiv $I32 (put_in_reg_zext32 x) (put_nonzero_in_reg y)))
-
-;; helpers for udiv:
 ;; Helper for placing a `Value` into a `Reg` and validating that it's nonzero.
-(decl put_nonzero_in_reg (Value) Reg)
+;; It takes a value and entension type, and perform emits the appropriate checks.
```
cfallin, line 1038: s/entension/extension/
s/perform/performs/

cranelift/codegen/src/isa/aarch64/lower.isle
```diff
@@ -1028,43 +1028,48 @@
...
+;; TODO: restore spec
```
cfallin, line 1039: cc @avanhatt @mmcloughlin -- maybe the first instance of active work on the aarch64 backend that needs to update a spec. I definitely don't think we should block this PR on it (so don't worry about this, @MarinPostma!) but it's worth thinking what our short and medium term approaches will be to this since we're upstreamed but don't have a nice CI-integrated workflow yet -- should we keep a queue of such TODOs somewhere?

cranelift/codegen/src/isa/aarch64/lower/isle.rs
```diff
@@ -206,24 +207,44 @@ impl Context for IsleContext<'_, '_, MInst, AArch64Backend> {
...
-            } else {
-                value & !(u64::MAX << bits)
-            }
-        } else {
-            value
+
+        let value = match extend_to {
+            OperandSize::Size32 => {
+                if bits < 32 {
+                    if *extend == generated_code::ImmExtend::Sign {
```
cfallin, line 222: can we make this a `match` (with `Sign` and `Zero` cases)?

cranelift/codegen/src/isa/aarch64/lower/isle.rs
```diff
@@ -206,24 +207,44 @@ impl Context for IsleContext<'_, '_, MInst, AArch64Backend> {
...
+                    } else {
+                        value & !((u32::MAX as u64) << bits)
+                    }
+                } else {
+                    value
+                }
+            },
+            OperandSize::Size64 => {
+                if bits < 64 {
+                    if *extend == generated_code::ImmExtend::Sign {
```
cfallin, line 236: Likewise here, and we can pull the `if bits < 64` in as a guard on one of the match arms as well.

cranelift/codegen/src/isa/aarch64/pcc.rs
```diff
@@ -438,6 +438,7 @@ fn check_addr<'a>(
                 if ctx.subsumes_fact_optionals(loaded_fact.as_ref(), result_fact) {
                     Ok(())
                 } else {
+                    dbg!();
```
cfallin, line 441: Debugging code left in?

cfallin (commented):
Looks good, thanks a bunch! Very nice to see the test expectations get shorter. A few thoughts below but nothing major.

MarinPostma:
hey @cfallin, fixign a bunch of stuff, that's why I put it in draft, but I'll include your review, as soon I manage to fix the tests :)

cranelift/codegen/src/isa/aarch64/lower.isle
MarinPostma: about that: I meant to restore them at some point, but I don't know how to run the verification

cfallin: Yep, the integration with the normal dev workflow is still very much an open question; we could have you ramp up on that but I don't think it's at the point that we want to require that of everyone yet.

cfallin (approved):
Thanks!

PR #9851, opened 2024-12-18
winch: Gracefully handle compilation errors

saulecabrera:
Closes: https://github.com/bytecodealliance/wasmtime/issues/8096

This commit threads `anyhow::Result`  through most of Winch's compilation process in order to gracefully handle compilation errors instead of panicking.

The error classification is intentionally very granular, to avoid string allocation which could impact compilation performance.

The errors largely fit in two categories:

* Unimplemented/Unsupported
* Internal

The firs category signals partial or no support for Wasmtime features and or Wasm proposals. These errors are meant to be temporary while such features or proposals are in development.

The second category signals that a compilation invariant was not met. These errors are considered internal and their presence usually means a bug in the compiler.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4741,6 +4741,7 @@ dependencies = [
  "regalloc2",
  "smallvec",
  "target-lexicon",
+ "thiserror",
  "wasmparser 0.221.2",
  "wasmtime-cranelift",
  "wasmtime-environ",
--- a/winch/codegen/Cargo.toml
+++ b/winch/codegen/Cargo.toml
@@ -25,6 +25,7 @@ regalloc2 = { workspace = true }
 gimli = { workspace = true }
 wasmtime-environ = { workspace = true }
 wasmtime-cranelift = { workspace = true }
+thiserror = { workspace = true }
 
 [features]
 x64 = ["cranelift-codegen/x86"]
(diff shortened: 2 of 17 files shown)


fitzgen:
> The second category signals that a compilation invariant was not met. These errors are considered internal and their presence usually means a bug in the compiler.

This category seems like it should probably result in panics? At least, in the outer winch compilation driver code?

saulecabrera:
Ultimately it should I think: thinking through a bit more I don't think there's a case in which we want to bubble these errors up as recoverable. Given the classification done in this PR I think it'll easy to check if the contained error is internal and panic at that point. I'll add that in.

fitzgen (approved):
LGTM. 

Don't really need to address the comment about internal errors before landing this (or even at all, if there is good motivation I am missing).

FWIW, if you are only using custom errors, it may make sense to avoid using `anyhow::Result` and `anyhow::Error` all through winch and just define a custom `type Result<T, E = WinchError> = core::result::Result<T, E>;`.

saulecabrera:
> FWIW, if you are only using custom errors, it may make sense to avoid using anyhow::Result and anyhow::Error all through winch and just define a custom type Result<T, E = WinchError> = core::result::Result<T, E>;

Good point, I should've mentioned the rationale for going with `anyhow`  rather than a custom `Result<T, WinchError>` in my commit message. That was my initial intention, however, since the code generation pass is intertwined with validation of other crates (e.g., `wasmparser`), I found it easier to handle cross-crate error definitions through anyhow, at the cost of a bit more verbosity in Winch's code base.

PR #9854, opened 2024-12-18
pulley: Shuffle opcodes to free some 1-byte opcodes

alexcrichton:
In the near future the set of opcodes here are going to be expanded along a number of axes such as:

* More modes of addressing loads/stores other than just `reg + offset32`.
* Opcodes with immediate operands rather than unconditional register operands.

The 1-byte opcode namespace is already filling up and there's a bit of a mishmash of what's 1-byte and what's "extended" for now. To help bridge this gap in the interim shuffle all float/vector-related opcodes into the "extended" opcode namespace. This frees up a large chunk of the 1-byte opcode namespace for future expansion with extensions like above.

I'll note that I'm not 100% sure that the opcodes will all stay here after this reshuffling. I haven't done any profiling/performance analysis to gauge the impact of this change. The immediate goal is to start experimenting with non-float/vector programs and get them profiling well. This will require more "macro opcodes" such as new addressing modes and opcodes-with-immediates. These are expected to be relatively hot opcodes and thus probably want to be in the "fast" 1-byte namespace, hence the shuffling here.

My plan is to in the future do a bit of an evaluation with a float/vector program and see whether it make sense to shuffle some of them into this 1-bytecode space as well. More radically it might make sense to remove the split between ops/extended ops and instead just have a 2-byte opcode space entirely. That's all left for future evaluations though.

### DIFF
--- a/pulley/src/opcode.rs
+++ b/pulley/src/opcode.rs
@@ -28,10 +28,7 @@ macro_rules! define_opcode {
             /// The value of the maximum defined opcode.
             pub const MAX: u8 = Opcode::ExtendedOp as u8;
         }
-    };
-
-    ( @max $x:ident ) => { 0 };
-    ( @max $x:ident $( $xs:ident )* ) => { 1 + define_opcode!(@max $( $xs )* ) };
+    }
 }
 for_each_op!(define_opcode);
 
@@ -77,7 +74,9 @@ macro_rules! define_extended_opcode {
 
         impl ExtendedOpcode {
             /// The value of the maximum defined extended opcode.
-            pub const MAX: u16 = define_opcode!( @max $( $name )* ) + 1;
+            pub const MAX: u16 = $(
+                if true { 1 } else { ExtendedOpcode::$name as u16 } +
+            )* 0;
         }
     };
 }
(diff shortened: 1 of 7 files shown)


pulley/src/opcode.rs
```diff
@@ -77,7 +74,9 @@ macro_rules! define_extended_opcode {
 
         impl ExtendedOpcode {
             /// The value of the maximum defined extended opcode.
-            pub const MAX: u16 = define_opcode!( @max $( $name )* ) + 1;
+            pub const MAX: u16 = $(
+                if true { 1 } else { ExtendedOpcode::$name as u16 } +
+            )* 0;
```
fitzgen, line 79: Leftover from debugging or something?

fitzgen (approved):
:+1:

pulley/src/opcode.rs
alexcrichton: Ah no this was intentional. The recursive definition required upping `#[recursion_limit]` which I'd prefer to not do. In lieu of fancier `macro_rules!` features which will make this nicer in the future I opted to have a chain of `1 + 1 + ...` for each opcode name. That requires using `$name` in the repetition somewhere though so I came up with this funky way of always evaluating to `1` but still using `$name` somehow.

alexcrichton:
I'm waffling on whether we want the 1-byte opcode namespace myself. I'm starting to lean towards "yes" so in the end I think we will want benchmarking one way or another (or size analysis) to show this.

PR #9858, opened 2024-12-18
pulley: Add offset8 integer loads/stores

alexcrichton:
This commit extends the set of opcodes to load/stores from memory with integer registers. Previously the only addressing mode supported was a base register plus a 32-bit signed immediate. This immediate frequently doesn't need 32-bits though and can often fit in a much smaller range. Looking at `spidermonkey.cwasm` a large number of loads/stores can fit within an unsigned 8-bit integer instead so this commit adds an `offset8` mode in addition to the preexisting `offset32` mode. Empirically this commit shrinks `spidermonkey.cwasm` for pulley64 from 33M to 31M.

This notably, at this time, does not extend general addressing modes in Pulley nor does it extend all loads/stores. For example float/vector/big-endian loads and stores all continue to only support a 32-bit signed offset from the base pointer. This is done under the assumption that integer loads/stores dominate both performance/code-size, but this is not empirically proven just yet.

Additionally at this time the choice is being made to add an opcode-per-addressing-mode rather than having a single load opcode take a general addressing mode. The assumption here is that decoding a fully general addressing mode and processing it is probably slower at runtime than specializing opcodes per addressing mode. This is currently an unproven assumption however and the cost of this is increased complexity in the Cranelift backend as it has to have many branches for all loads/stores supported.

### DIFF
--- a/cranelift/codegen/src/isa/pulley_shared/inst/emit.rs
+++ b/cranelift/codegen/src/isa/pulley_shared/inst/emit.rs
@@ -119,6 +119,26 @@ where
     }
 }
 
+/// Representation of a static offset from a pointer.
+///
+/// In VCode this is always represented as an `i32` and then just before
+/// lowering this is used to determine which instruction to emit.
+enum Offset {
+    /// An unsigned 8-bit offset.
+    U8(u8),
+    /// A signed 32-bit offset.
+    I32(i32),
+}
+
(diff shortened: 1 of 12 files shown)


fitzgen (approved):
LGTM. Agree with both intuitions, looking forward to seeing empirical evidence.

PR #9860, opened 2024-12-18
Add Default and Debug impls to SparseMap

### LINKED ISSUE #9856: impl Debug and/or Default for SparseMap
Hello, is there a reason that [SparseMap](https://github.com/bytecodealliance/wasmtime/blob/main/cranelift/entity/src/sparse.rs) doesn't impl `std::fmt::Debug` and/or `Default`? I would like to use it inside my own data structures that derive those impls.

(I read [these contribution guidelines](https://docs.wasmtime.dev/contributing.html) which encouraged me to open an issue, please let me know if there's a more appropriate place to ask, thanks!)

wtachau:
https://github.com/bytecodealliance/wasmtime/issues/9856

Add `Default` and `Debug` trait impls to `SparseMap` (for the latter, only impl `Debug` if `K` and `V` impl `Debug`).

closes #9856

### DIFF
--- a/cranelift/entity/src/sparse.rs
+++ b/cranelift/entity/src/sparse.rs
@@ -220,9 +220,8 @@ where
     V: SparseMapValue<K> + fmt::Debug,
 {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_struct("SparseMap")
-            .field("sparse", &self.sparse)
-            .field("dense", &self.dense)
+        f.debug_map()
+            .entries(self.values().map(|v| (v.key(), v)))
             .finish()
     }
 }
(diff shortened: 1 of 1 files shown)


cranelift/entity/src/sparse.rs
```diff
@@ -203,6 +204,29 @@ where
...
+impl<K, V> fmt::Debug for SparseMap<K, V>
+where
+    K: EntityRef + fmt::Debug,
+    V: SparseMapValue<K> + fmt::Debug,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("SparseMap")
+            .field("sparse", &self.sparse)
+            .field("dense", &self.dense)
+            .finish()
```
bjorn3, line 226: The sparse and dense fields are not exposed to the user and rather useless unless you know how the implementation lf SparseMap works. I would much rather prefer the Debug impl being like that of HashMap showing the logical mapping from keys to values. So for example `{ v0: "a", v2: "b" }` for a `SparseMap<Value, String>`. This can be done using something like `f.debug_map().entries(entries).finish()` where `entries` is an iterator over the map entries.

wtachau: makes sense! updated.

cfallin (approved):
LGTM, thanks!

PR #9862, opened 2024-12-18
pulley: Remove unwind metadata from Cranelift backend

alexcrichton:
This is a copy/paste from the riscv64 backend originally but there's no need to integrate with native unwinders on Pulley so there's no need to track this information. This removes the `Unwind` pseudo-inst entirely.

### DIFF
--- a/cranelift/codegen/src/isa/pulley_shared/abi.rs
+++ b/cranelift/codegen/src/isa/pulley_shared/abi.rs
@@ -4,7 +4,7 @@ use super::{inst::*, PulleyFlags, PulleyTargetKind};
 use crate::isa::pulley_shared::{PointerWidth, PulleyBackend};
 use crate::{
     ir::{self, types::*, MemFlags, Signature},
-    isa::{self, unwind::UnwindInst},
+    isa::{self},
     machinst::*,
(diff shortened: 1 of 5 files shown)


cranelift/codegen/src/isa/pulley_shared/abi.rs
```diff
@@ -4,7 +4,7 @@ use super::{inst::*, PulleyFlags, PulleyTargetKind};
 use crate::isa::pulley_shared::{PointerWidth, PulleyBackend};
 use crate::{
     ir::{self, types::*, MemFlags, Signature},
-    isa::{self, unwind::UnwindInst},
+    isa::{self},
```
cfallin, line 7: strange artifact not caught by `rustfmt` maybe? would prefer `isa::self` to `isa::{self}`

cfallin (commented):
Nice!

cfallin (approved):

cranelift/codegen/src/isa/pulley_shared/abi.rs
alexcrichton: oh oops I totally didn't notice this at all!

PR #9864, opened 2024-12-19
pulley: Add macro instructions for function prologue/epilogue

alexcrichton:
This commit adds two new instructions to Pulley to combine the operations of setting up a frame, allocating stack, and saving clobbered registers. This is all combined into a single instruction which is relatively large but is much smaller than each of these individual operations exploded out.

This is a size win on `spidermonkey.cwasm` by about 1M and locally in a small `fib.wat` test this is also a good speedup by reducing the number of instructions executed.

### DIFF
--- a/pulley/src/interp.rs
+++ b/pulley/src/interp.rs
@@ -1798,6 +1798,50 @@ impl OpVisitor for Interpreter<'_> {
         ControlFlow::Continue(())
     }
 
+    #[inline]
+    fn push_frame_save(&mut self, amt: u32, regs: RegSet<XReg>) -> ControlFlow<Done> {
+        // Decrement the stack pointer `amt` bytes plus 2 pointers more for
+        // fp/lr.
+        let ptr_size = size_of::<usize>();
+        let full_amt = usize::try_from(amt).unwrap() + 2 * ptr_size;
+        let new_sp = self.state[XReg::sp].get_ptr::<u8>().wrapping_sub(full_amt);
+        self.set_sp::<crate::PushFrameSave>(new_sp)?;
+
+        unsafe {
+            // Emulate `push_frame` by placing `lr` and `fp` onto the stack, in
+            // that order, at the top of the allocated area.
+            self.store(XReg::sp, (full_amt - 1 * ptr_size) as i32, self.state.lr);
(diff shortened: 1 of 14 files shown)


pulley/src/interp.rs
```diff
@@ -1694,6 +1694,50 @@ impl OpVisitor for Interpreter<'_> {
         ControlFlow::Continue(())
     }
 
+    #[inline]
+    fn push_frame_save(&mut self, amt: u32, regs: RegSet<XReg>) -> ControlFlow<Done> {
+        // Decrement the stack pointer `amt` bytes plus 2 pointers more for
+        // fp/lr.
+        let ptr_size = size_of::<usize>();
+        let full_amt = usize::try_from(amt).unwrap() + 2 * ptr_size;
```
fitzgen, line 1702: Shouldn’t we be saving the full 64 bits of each register, not the usize, for when we are on 32-bit platforms? do/lr seem fine to be usize but not GPRs.

pulley/src/lib.rs
```diff
@@ -360,6 +360,16 @@ macro_rules! for_each_op {
...
 
+            /// Macro-instruction to enter a function, allocate some stack, and
+            /// then save some registers.
+            ///
+            /// This is equivalent to `push_frame`, `stack_alloc32 amt`, then
+            /// saving all of `regs` to the top of the stack just allocated.
+            push_frame_save = PushFrameSave { amt: u32, regs: RegSet<XReg> };
+            /// Inverse of `push_frame_save`. Restores `regs` from the top of
+            /// the stack, then runs `stack_free32 amt`, then runs `pop_frame`.
+            pop_frame_save = PopFrameSave { amt: u32, regs: RegSet<XReg> };
```
fitzgen, line 371: Maybe this should be called pop_frame_restore?

fitzgen (approved):
Nice! Wanted to put the clobbers and the frame setup all in the same instruction but the Cranelift abi traits weren’t quite making it natural. I like the way you handled that here.

pulley/src/interp.rs
alexcrichton: Oh the `ptr_size` bits are just for fp/lr which are pointer-sized, the `amt` which Cranelift decides handles the 8-bytes-per-integer-register

PR #9866, opened 2024-12-19
[DWARF] Fix debug intrinsics on Linux

SingleAccretion:
Work in progress.


crates/wasmtime/src/runtime/vm/helpers.c
```diff
@@ -163,6 +163,15 @@ __attribute__((weak, noinline))
     void __jit_debug_register_code() {
 #ifndef CFG_TARGET_OS_windows
   __asm__("");
+#ifdef FEATURE_DEBUG_BUILTINS
+  // Make sure these symbols do not get stripped by the compiler or linker.
+  // Once/if the Rust toolchain used gets new enough, this can be replaced by
+  // __attribute__((used, retain)).
```
alexcrichton, line 169: Question on this, what's the toolchain needed for this? I'm assuming the comment here about the Rust toolchain might be mistaken because Rust isn't involved in compiling this file, but do you know the minimum gcc/clang/etc version needed to process this?

We could look into CI and see which toolchain is lagging behind perhaps (I think x64-linux is one of the oldest due to glibc compat so probably nothing to do about that in the end alas)

.github/workflows/main.yml
```diff
@@ -860,7 +860,7 @@ jobs:
         # workaround for https://bugs.launchpad.net/ubuntu/+source/llvm-defaults/+bug/1972855
         sudo mkdir -p /usr/lib/local/lib/python3.10/dist-packages/lldb
         sudo ln -s /usr/lib/llvm-15/lib/python3.10/dist-packages/lldb/* /usr/lib/python3/dist-packages/lldb/
-        cargo test --test all -- --ignored --test-threads 1 debug::
+        cargo test --release --test all -- --ignored --test-threads 1 debug::
```
alexcrichton, line 863: If these tests are fast enough mind adding both `--release`-and-not?

crates/wasmtime/src/runtime/vm/helpers.c
SingleAccretion: > I'm assuming the comment here about the Rust toolchain might be mistaken because Rust isn't involved in compiling this file

(I was secretly hoping this uses some bundled clang, but if it's using the host/found in PATH toolchain...)

> Question on this, what's the toolchain needed for this? , but do you know the minimum gcc/clang/etc version needed to process this?

It is a relatively recent feature.

> __attribute__((retain)) (GCC 11 with modern binutils, Clang 13).

https://github.com/llvm/llvm-project/blob/28865769440756138a88a9c8e8b72b1f5d8db715/lld/docs/ELF/start-stop-gc.rst#annotate-c-identifier-name-sections

I think the best resolution here is to actually simply delete the comment with `retain`. `volatile` is a more robust than trying to ensure everything has the right toolchain.

.github/workflows/main.yml
SingleAccretion: Added. Numbers follow.

The tests themselves are fairly quick (because there are so few of them :)), what takes its time is building everything. 

This `--release` version now takes ~8min, and the previous version took ~5min. Combining these, we conservatively get ~13min, which is on the order of the longest job in this PR (~12min).

alexcrichton: Ah ok good point, in that case let's go ahead and drop the debug build then and only test `--release`. Mind adding a comment as to why we're only testing `--release` as well?

alexcrichton (approved):

PR #9867, opened 2024-12-19
Remove async-trait in bindgen

### LINKED ISSUE #9823: Bindgen improvement: Remove the use of `async_trait`
#### Feature

Remove the use of `async_trait` in `bindgen!` as per https://github.com/bytecodealliance/wasmtime/issues/9776#issuecomment-2539971806. Use async fn in traits, RPITIT and `trait_variant` instead.

#### Benefit

Removing `async_trait` simplifies the generated code by `bindgen!` a lot. This can also drastically improve IDE experience as IDEs have better support for async fn in traits and RPITIT. Coding experience should be much better as well, as programmers can just use `async fn` when implementing autogenerated traits by `bindgen` in async mode.

#### Implementation

Note: I don't have time to dig into details now, so this issue is primary for tracking. Probably someone can tag this as "good first issue" :)

Just search for `::async_trait` in the source code of `wit-bindgen` crate:

https://github.com/search?q=repo%3Abytecodealliance%2Fwasmtime+path%3A%2F%5Ecrates%5C%2Fwit-bindgen%5C%2F%2F+%3A%3Aasync_trait&type=code

There are not many occurrences to my surprise.

image

In theory, we need to generate something like:
```rust
#[trait_variant::make(HostX: Send)]
pub trait LocalHostX {
//...
}
```
instead of 
```
#[async_trait]
pub trait HostX {
//...
}
```

cc @alexcrichton

ifsheldon:
Closes #9823 

I ran `cargo test -p wasmtime-wasi` and all tests passed, so now I'd like to go through the whole CI.

There are only 2 cases remaining unresolved due to the use of dyn objects, i.e., `wasmtime::runtime::store::CallHookHandler` and `wasmtime::runtime::limits::ResourceLimiterAsync`, but they are related to the runtime, so I don't think they are very relevant to the issue and I left them unchanged.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -73,6 +73,7 @@ humantime = { workspace = true }
 tempfile = { workspace = true, optional = true }
 
 async-trait = { workspace = true }
+trait-variant = { workspace = true }
 bytes = { workspace = true }
 cfg-if = { workspace = true }
 tokio = { workspace = true, optional = true, features = [ "signal", "macros" ] }
@@ -96,6 +97,7 @@ criterion = { workspace = true }
 num_cpus = "1.13.0"
 memchr = "2.4"
 async-trait = { workspace = true }
+trait-variant = { workspace = true }
 wat = { workspace = true }
 rayon = "1.5.0"
 wasmtime-wast = { workspace = true, features = ['component-model'] }
@@ -313,6 +315,7 @@ tracing = "0.1.26"
 bitflags = "2.0"
 thiserror = "1.0.43"
 async-trait = "0.1.71"
+trait-variant = "0.1.2"
 heck = "0.5"
 similar = "2.1.0"
 toml = "0.8.10"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4435,6 +4435,7 @@ dependencies = [
  "tokio",
  "tracing",
  "tracing-subscriber",
+ "trait-variant",
  "url",
  "wasmtime",
  "wiggle",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3456,6 +3456,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "trait-variant"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "70977707304198400eb4835a78f6a9f928bf41bba420deb8fdb175cd965d77a7"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.90",
+]
+
+[[package]]
 name = "try-lock"
 version = "0.2.4"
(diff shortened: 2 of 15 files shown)


Cargo.toml
```diff
@@ -313,6 +315,7 @@ tracing = "0.1.26"
 bitflags = "2.0"
 thiserror = "1.0.43"
 async-trait = "0.1.71"
+trait-variant = "0.1.2"
```
dicej, line 318: Per [this section](https://docs.wasmtime.dev/contributing-coding-guidelines.html?highlight=vet#cargo-vet-for-contributors) of the contributing docs, the bar for adding new dependencies to Wasmtime is a bit higher than for other projects (hence the `cargo vet` CI failure), so we need to make sure this one is worth including.

From what I can tell, it's only being used to annotate `wasmtime-wit-bindgen`-generated traits.  Given that we're generating those traits anyway, I wonder if we should modify the generator to emit a `-> impl Future<Output = T> + Send` return type directly rather than rely on `trait-variant` to transform the code it just generated.  That would avoid the extra dependency and remove a layer of abstraction.  What do you think?

dicej (changes requested):
Thanks for this!

LGTM overall, but see my inline comment about avoiding the `trait-variant` dependency, which would address the `cargo vet` CI failure.

The other CI failure should be easy to fix: just run `BINDGEN_TEST_BLESS=1 cargo test -p wasmtime-component-macro --test expanded`.

Cargo.toml
ifsheldon: I would strongly recommend adding `trait-variant` because:
1. it is maintained officially by the Rust project, so the code quality is reassured.
2. this is the recommended way to add `Send` bound by the [Rust announcement about async fn in traits](https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html)
3. in the announcement, it's said future version of `trait-variant` will support dyn trait object. Using `trait-variant` will enable dyn object support in the future without the need to modify `bindgen`
4. compared to writing `-> impl Future<Output = T> + Send` ourselves, it's better to leave `async fn` signatures for users to debug when users have IDEs like RustRover that can expand proc_macro layer by layer. So when they want to debug, they actually see `#[trait_variant::make]` and `async fn` instead of  `fn ... -> impl Future<Output = T> + Send`
5. from my perspective, the use of `#[trait_variant::make]` is sort of a marker indicating better support of async fn in traits is needed, which may benefit contributions.

In general, I think `trait-variant` is a better solution than `async-trait` when we take the future evolution of Rust into account.

PS: replacing `async-trait` with `trait-variant` is a breaking change because:
* traits by `async-trait` support dyn trait object while traits with `trait-variant` for now do not
* when implementing async traits, users need to remove `#[async_trait::async_trait]` on the impl blocks

dicej (approved):

PR #9872, opened 2024-12-19
pulley: Refactor and reuse `imm` for floats

alexcrichton:
This was originally added long ago and this brings the method up-to-date with the current pulley opcode set.


fitzgen (approved):
Good catch

PR #9874, opened 2024-12-19
pulley: Add macro `CallN` instructions

alexcrichton:
This commit adds new macro instructions to assist with speeding up calls between functions. Pulley's previous `Call` instruction was similar to native call instructions where arguments/results are implicitly in the right location according to the ABI, but movement between registers is more expensive with Pulley than with native architectures. The `CallN` instructions here enable listing a few arguments (only integer registers) in the opcode itself. This removes the need for individual `xmov` instructions into individual registers and instead it can all be done within the opcode handlers.

This additionally enables passing the same argument twice to a function to reside only in one register. Finally parallel-copies between these registers are supported as the interpreter loads all registers and then stores all registers.

These new instructions participate in register allocation differently from before where the first few arguments are allowed to be in any register and no longer use `reg_fixed_use`. All other arguments (and all float arguments for example) continue to use `reg_fixed_use`.

Locally sightglass reports this change speeding up `pulldown-cmark` by 2-10%. On a `fib(N)` micro-benchmark it didn't help as much as I was hoping that it was going to.

### DIFF
--- a/cranelift/codegen/src/isa/pulley_shared/inst/emit.rs
+++ b/cranelift/codegen/src/isa/pulley_shared/inst/emit.rs
@@ -172,16 +172,36 @@ fn pulley_emit<P>(
         Inst::LoadExtName { .. } => todo!(),
 
         Inst::Call { info } => {
-            sink.put1(pulley_interpreter::Opcode::Call as u8);
-            sink.add_reloc(
+            let offset = sink.cur_offset();
+
+            // If arguments happen to already be in the right register for the
+            // ABI then remove them from this list. Otherwise emit the
+            // appropriate `Call` instruction depending on how many arguments we
+            // have that aren't already in their correct register according to
+            // ABI conventions.
+            let mut args = &info.dest.args[..];
+            while !args.is_empty() && args.last().copied() == XReg::new(x_reg(args.len() - 1)) {
+                args = &args[..args.len() - 1];
+            }
+            match args {
+                [] => enc::call(sink, 0),
+                [x0] => enc::call1(sink, x0, 0),
+                [x0, x1] => enc::call2(sink, x0, x1, 0),
+                [x0, x1, x2] => enc::call3(sink, x0, x1, x2, 0),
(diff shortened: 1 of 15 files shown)


cranelift/codegen/src/isa/pulley_shared/inst/emit.rs
```diff
@@ -172,16 +172,36 @@ fn pulley_emit<P>(
...
+            while !args.is_empty() && args.last().copied() == XReg::new(x_reg(args.len() - 1)) {
+                args = &args[..args.len() - 1];
+            }
+            match args {
+                [] => enc::call(sink, 0),
+                [x0] => enc::call1(sink, x0, 0),
+                [x0, x1] => enc::call2(sink, x0, x1, 0),
+                [x0, x1, x2] => enc::call3(sink, x0, x1, x2, 0),
+                [x0, x1, x2, x3] => enc::call4(sink, x0, x1, x2, x3, 0),
+                _ => unreachable!(),
```
fitzgen, line 192: Do we not have a fallback instruction for the `n >= 5` case?

cranelift/codegen/src/isa/pulley_shared/inst/mod.rs
```diff
@@ -151,7 +151,29 @@ fn pulley_get_operands(inst: &mut Inst, collector: &mut impl OperandVisitor) {
...
+
+            // Remaining arguments (and return values) are all in fixed
+            // registers according to Pulley's ABI, however.
+            for CallArgPair { vreg, preg } in uses {
+                collector.reg_fixed_use(vreg, *preg);
+            }
+            for CallRetPair { vreg, preg } in defs {
+                collector.reg_fixed_def(vreg, *preg);
+            }
+            collector.reg_clobbers(info.clobbers);
```
fitzgen, line 174: Ahhhh, I see what's going on. Argument `i` for `i >= 5` are in fixed registers and arguments `i <= 4` go in any register and that is encoded in the call instructions. Clever.

fitzgen (approved):

PR #9878, opened 2024-12-20
make canonicalize-nan.wast test work

Lordworms:
part of https://github.com/bytecodealliance/wasmtime/issues/9783


pulley/src/interp.rs
```diff
@@ -2612,25 +2612,47 @@ impl ExtendedOpVisitor for Interpreter<'_> {
 
     fn bitcast_int_from_float_32(&mut self, dst: XReg, src: FReg) -> ControlFlow<Done> {
         let val = self.state[src].get_f32();
-        self.state[dst].set_u32(val.to_bits());
+        let bits = if val.is_nan() {
+            0x7fc00000
+        } else {
+            val.to_bits()
+        };
```
cfallin, line 2619: I'm surprised this is necessary: the intent of `canonicalize-nan.wast` is to test Cranelift's NaN-canonicalization pass, and this pass operates on the machine-independent IR (CLIF) so it should work for compilation to Pulley as well as any other target. The Pulley virtual machine shouldn't have to unconditionally canonicalize NaNs.

Could you check whether the pass is producing NaN-canonicalization sequences (which look like compares to test for NaN, then a select operator)?

I'm wondering whether this might have something to do with the `has_vector_support` boolean in `Context::canonicalize_nans()`; the comment there (that riscv64 is the only architecture without vector support) is no longer accurate now that we have Pulley in-tree; I'm not sure but perhaps that's related?

alexcrichton: Oh I think this might be my fault, in the [contribution docs for pulley](https://github.com/bytecodealliance/wasmtime/blob/main/pulley/CONTRIBUTING.md) it recommends running the test suite directly via `wasmtime wast .../*.wast`. When `tests/misc_testsuite/simd/canonicalize-nan.wast` is executed that's actually expected to fail (it even fails on native) and it requires an extra `-W nan-canonicalization` flag. 

Chris is right though in that the manual canonicalization here shouldn't be necessary. @Lordworms you'll find though that `-W nan-canonicalization` isn't fully supported just yet:

```
Error: failed to run script file './tests/misc_testsuite/simd/canonicalize-nan.wast'

Caused by:
    0: failed directive on ./tests/misc_testsuite/simd/canonicalize-nan.wast:7:1
    1: Compilation error: Unsupported feature: should be implemented in ISLE: inst = `v10 = fcmp.f32x4 uno v7, v7`, type = `Some(types::I32X4)`
```

That being said this PR still gets the `simd_f32x4_rounding.wast` test passing, so with the canonicalization removed it should still be good to flag that test as passing

pulley/src/lib.rs
```diff
@@ -805,14 +805,30 @@ macro_rules! for_each_extended_op {
             fminimum32 = Fminimum32 { operands: BinaryOperands<FReg> };
             /// `low32(dst) = ieee_trunc(low32(src))`
             ftrunc32 = Ftrunc32 { dst: FReg, src: FReg };
+            /// `low128(dst) = ieee_trunc(low128(src))`
+            ftrunc32x4 = Ftrunc32x4 { dst: VReg, src: VReg };
```
alexcrichton, line 809: As a minor stylistic note, mind prefixing these instructions with `v` (e.g. `vtrunc32x4`) instead? I've been trying to use the `v` prefix for all simd/vector-related instructions so far

Lordworms: yep

pulley/src/interp.rs
Lordworms: I am new to this project and didn't find it. Sorry for that.

crates/wast-util/src/lib.rs
```diff
@@ -509,12 +507,10 @@ impl WastTest {
                 "spec_testsuite/simd_f32x4_arith.wast",
                 "spec_testsuite/simd_f32x4_cmp.wast",
                 "spec_testsuite/simd_f32x4_pmin_pmax.wast",
-                "spec_testsuite/simd_f32x4_rounding.wast",
                 "spec_testsuite/simd_f64x2.wast",
                 "spec_testsuite/simd_f64x2_arith.wast",
                 "spec_testsuite/simd_f64x2_cmp.wast",
                 "spec_testsuite/simd_f64x2_pmin_pmax.wast",
-                "spec_testsuite/simd_f64x2_rounding.wast",
```
alexcrichton, line 517: Oh this'll want to stay as these are test exemptions for a different compiler, Winch, unrelated to Pulley. Only the ones above are the ones affecting Pulley.

Also it looks like a merge conflict has creeped in due to this file, so if you want to rebase that should help resolve the conflict.

Lordworms: resolved it.

crates/wast-util/src/lib.rs
```diff
@@ -419,11 +419,10 @@ impl WastTest {
                 "spec_testsuite/simd_f32x4_arith.wast",
                 "spec_testsuite/simd_f32x4_cmp.wast",
                 "spec_testsuite/simd_f32x4_pmin_pmax.wast",
-                "spec_testsuite/simd_f32x4_rounding.wast",
+                "spec_testsuite/simd_f64x2.wast",
```
alexcrichton, line 422: Oh I think this test popped back in by accident during the rebase

alexcrichton (approved):

PR #9884, opened 2024-12-20
pulley: Fill out most remaining simd float ops

alexcrichton:
Get most simd/float-related tests passing. Mostly reusing preexisting scalar ops for the simd implementation.


fitzgen (approved):
:+1:

alexcrichton:
A bit of churn here at the end realizing that MinGW's `fma` routine is different and so we unconditionally use the `libm` crate on MinGW now.

PR #9885, opened 2024-12-20
Refactor release builds, add aarch64-musl

### LINKED ISSUE #9875: Publishing musl aarch64 build products
#### Feature

I would like wasmtime releases to be built for musllinux on aarch64 platform.

#### Benefit

Improved portability of wasmtime-py when installed from PyPI (https://github.com/bytecodealliance/wasmtime-py/issues/263).

#### Implementation

As far as I know this target should build fine, so only CI scripts need to be updated.

alexcrichton:
This commit refactors the way release builds are done in CI in terms of configuration and then additionally adds aarch64-musl release artifacts as requested in #9875. The refactoring here is done to reduce the number of locations to understand release builds. Notably the `binary-compatible-builds` action was removed in favor of direct environment configuration in conjunction with docker images used to build. The `aarch64-musl` build itself happens in a container provided by the `cross` project to ensure that the right toolchain is configured.

Closes #9875


alexcrichton:
@whitequark from [this CI page](https://github.com/bytecodealliance/wasmtime/actions/runs/12437964798?pr=9885) if you scroll to the bottom and find `bins-aarch64-musl` (here's a [direct link](https://github.com/bytecodealliance/wasmtime/actions/runs/12437964798/artifacts/2350223056), not sure if that'll work) can you test that out? Mostly just make sure `./wasmtime --help` does something and maybe poke at the commands a bit. If that works I think this is probably good to go to eventually get integrated into wasmtime-py

whitequark:
OK so here is what I get if I run wasmtime in alpine edge on aarch64:
Screenshot_20241220_210142

alexcrichton:
Ok perfect, thanks! I still don't know how to best manage libgcc_s on musl but it's at least no worse-off than the x64 target which I believe has the same behavior.

dicej (approved):

PR #9892, opened 2024-12-23
Override component model `Lower::store_list` and `Lift::load_list` for f32/f64

kaivol:
Added overrides for `Lower::store_list` and `Lift::load_list` for floating point numbers to reduce overhead when passing or returning `list<f32>` and `list<f64>` to/from components. 

I based this on the respective implementation for integers.

### DIFF
--- a/crates/wasmtime/src/runtime/component/func/typed.rs
+++ b/crates/wasmtime/src/runtime/component/func/typed.rs
@@ -12,6 +12,7 @@ use core::marker;
 use core::mem::{self, MaybeUninit};
 use core::ptr::NonNull;
 use core::str;
+use std::iter;
 use wasmtime_environ::component::{
     CanonicalAbiInfo, ComponentTypes, InterfaceType, StringEncoding, VariantInfo, MAX_FLAT_PARAMS,
     MAX_FLAT_RESULTS,
@@ -902,7 +903,7 @@ integers! {
 }
 
 macro_rules! floats {
-    ($($float:ident/$get_float:ident = $ty:ident with abi:$abi:ident)*) => ($(const _: () = {
+    ($($float:ident/$get_float:ident = $ty:ident with abi:$abi:ident $integer:ident)*) => ($(const _: () = {
         unsafe impl ComponentType for $float {
             type Lower = ValRaw;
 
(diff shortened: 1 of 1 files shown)


crates/wasmtime/src/runtime/component/func/typed.rs
```diff
@@ -914,6 +915,44 @@ macro_rules! floats {
...
+                // alignment for this type which should be true for all
+                // platforms.
+                assert!((Self::ALIGN32 as usize) >= mem::align_of::<Self>());
+
+                // Slice `cx`'s memory to the window that we'll be modifying.
+                // This should all have already been verified in terms of
+                // alignment and sizing meaning that these assertions here are
+                // not truly necessary but are instead double-checks.
+                //
+                // Note that we're casting a `[u8]` slice to `[$integer]` (with
```
dicej, line 937: This paragraph (and the code that follows) refers to `$integer`, but we really care about `$float`, right?  We should rewrite this to talk about floating point numbers instead of integers, e.g. "all `u8` patterns are valid `$float` patterns since `$float` is a IEEE 754 floating point type."

crates/wasmtime/src/runtime/component/func/typed.rs
```diff
@@ -929,13 +968,46 @@ macro_rules! floats {
                 debug_assert!((bytes.as_ptr() as usize) % Self::SIZE32 == 0);
                 Ok($float::from_le_bytes(bytes.try_into().unwrap()))
             }
+
+            fn load_list(cx: &mut LiftContext<'_>, list: &WasmList<Self>) -> Result<Vec<Self>> where Self: Sized {
+                // See comments in `WasmList::get` for the panicking indexing
+                let byte_size = list.len * mem::size_of::<$integer>();
+                let bytes = &cx.memory()[list.ptr..][..byte_size];
+
+                // The canonical ABI requires that everything is aligned to its
```
dicej, line 977: Again, these comments should be updated to discuss floating point types rather than integer types.

crates/wasmtime/src/runtime/component/func/typed.rs
```diff
@@ -914,6 +915,44 @@ macro_rules! floats {
...
+                // alignment and sizing meaning that these assertions here are
+                // not truly necessary but are instead double-checks.
+                //
+                // Note that we're casting a `[u8]` slice to `[$integer]` (with
+                // $integer having the same size as Self) with `align_to_mut`
+                // which is not safe in general but is safe in our specific
+                // case as all `u8` patterns are valid `$integer` patterns
+                // since `$integer` is an integral type.
+                let dst = &mut cx.as_slice_mut()[offset..][..items.len() * Self::SIZE32];
+                let (before, middle, end) = unsafe { dst.align_to_mut::<$integer>() };
```
dicej, line 943:
```suggestion
                let (before, middle, end) = unsafe { dst.align_to_mut::<$float>() };
```

Presumably the alignment of e.g. `f32` and `u32` should be the same, but we might as well be precise here (and in similar code below).

dicej (changes requested):
Thanks for this!  Looks reasonable to me overall; see my comments inline regarding the details.

kaivol:
Thanks for the review!
I'm a bit busy at the moment but I'll look into it when i find some time.

crates/wasmtime/src/runtime/component/func/typed.rs
kaivol: The reason I used integer types and `align_to[_mut]::<$integer>` was to handle endianness correctly.
I felt like using `align_to[_mut]::<$float>` and thereby transmuting the component's memory to `$float` is incorrect, because (in my understanding) `$float` represents floating point numbers in native endianness, while `dst` is a slice of floating point numbers with little-endian byte-order. 

Therefore, I thought using `$float` for floating point numbers with different endianess would also cause confusion, in particular when converting endianess. 

I also considered using only accordingly aligned bytes (e.g. `[u8; 4]`) for the memory of the component, but I was unsure if that would enable the same optimizations. 
Is there any easy (built-in) way to benchmark Wasmtime in specific scenarios?

Let me know what you think.

kaivol:
I modified the code to work with byte slices (`[u8]`) when referring to the component's memory. I think this is the best way to show which variables are host and which are component memory. 

I also got rid of the `unsafe` blocks, without compromising on performance (at least in my experiments).

Let me know if the code looks good, and if further comments would help clarify it.

dicej (approved):
Thanks!

crates/wasmtime/src/runtime/component/func/typed.rs
```diff
@@ -12,6 +12,7 @@ use core::marker;
 use core::mem::{self, MaybeUninit};
 use core::ptr::NonNull;
 use core::str;
+use std::iter;
```
dicej, line 15:
```suggestion
use core::iter;
```

That should fix https://github.com/bytecodealliance/wasmtime/actions/runs/14112166505/job/39533533593

dicej:
One last thing to make CI happy (I hope): please run `cargo fmt` and push the result.

PR #9896, opened 2024-12-23
pulley: Make canonicalize-nan.wast pass

eagr:
Helps #9783


eagr:
```
cargo run --features pulley -- wast --target pulley64 ./tests/misc_testsuite/simd/canonicalize-nan.wast -W nan-canonicalization
```
Seems passing. Is it how the test supposed to run?

About the fn naming, I'm just kinda going along with `veq32x4` and friends. If you'd prefer, say, `vlef32x4` over `vlteqf32x4`, or `vueqf64x2` over `vunoeqf64x2`, I'm happy to change them.

And I haven't implemented all the float condcodes, just enough to get the tests pass. I did skim through the docs, but I don't believe I understand half of the implications. It'd great to hear your inputs before I proceed implementing others.

alexcrichton:
Thanks for this! I think it might be best to cut down on some of these opcodes in Pulley though since wasm doesn't actually use many of the `FloatCC::Unordered*` orderings for example. I'd recommend having a pulley opcode for wasm-based orderings but those needed only for nan canonicalization for example I think should be implemented like other lowering rules on other platforms (e.g. using combinations of opcodes instead of one). If this ever becomes performance sensitive we can of course revisit.

I'd recommend taking a look at how [the riscv64 backend does this](https://github.com/bytecodealliance/wasmtime/blob/6e955180deddccfc252abab7a8ad5ce7e891b23e/cranelift/codegen/src/isa/riscv64/inst_vector.isle#L1803-L1834) for example

alexcrichton:
Oh I'll also note that another example is in https://github.com/bytecodealliance/wasmtime/pull/9884 (which I'm happy to merge your PR before mind and rebase around yours), which has another possible strategy for implementing some comparison-related opcodes here.

eagr:
I didn't notice your PR. I couldn't see this one is contributing anything on top of yours, so I guess it's faster to just close this one to get out of the way. :)

PR #9898, opened 2024-12-24
[DWARF] Transform instance methods into static methods with a `__this` parameter

### LINKED ISSUE #9855: [DWARF] Translate `this` in a more accessible way
Currently, for instance methods, `this` pointers are translated like all other pointers - wrapped in the synthetic types. However, this breaks LLDB's ability to do anything with them as it expects `this` to be a proper pointer.

(There are still ways to inspect the underlying data, of course, but they're more clunky than just `p Member`.)

The proposed fix, consistent with the existing system, would be to rename `this` and effectively transform all instance methods into static ones.

The user experience would then be `p __this->Member` instead of `p Member`.

A more ambitious alternative is to translate it into a proper pointer. Unlike for a number of other cases, we _can_ do that (since we have the frame context available), but it's a bit more complicated (need to intelligently intersect ranges of `__vmctx` with those of `this`) and more importantly, it wouldn't be consistent with how the rest of the system operates.

SingleAccretion:
This lets one use `this` (as `__this`) in the debugger's expression evaluator.

Closes #9855.


alexcrichton (approved):
Thanks!

PR #9902, opened 2024-12-25
Add `WASMTIME_LOG_NO_CONTEXT`

SingleAccretion:
When using trace logging for compiler debugging, the usual 'context': time, level, target, are just visual noise.

Add an environment variable to disable them. There is no change to default behavior.

Sample output diff:
```diff
- 2024-12-25T20:07:32.459831Z TRACE debug-info-transform: === Begin DIE at 0x0000229a (depth = 2):
+ === Begin DIE at 0x0000229a (depth = 2):
  DW_TAG_GNU_call_site
    DW_AT_abstract_origin (0x000022a4)
    DW_AT_low_pc (ffffffff)
```


crates/cranelift/src/debug/transform/debug_transform_logging.rs
```diff
@@ -123,7 +123,10 @@ impl<'a> fmt::Debug for DieDetailedSummary<'a> {
                 AttributeValue::CallingConvention(value) => write!(f, "{value}"),
                 AttributeValue::Inline(value) => write!(f, "{value}"),
                 AttributeValue::Ordering(value) => write!(f, "{value}"),
-                AttributeValue::UnitRef(offset) => write!(f, "0x{:08x}", offset.0),
+                AttributeValue::UnitRef(offset) => {
+                    let section_offset = offset.to_unit_section_offset(unit);
+                    write!(f, "0x{:08x}", get_offset_value(section_offset))
+                }
```
pchickey, line 129: Just calling out this change as unrelated - we can merge it in this PR. I don’t understand cranelift well enough to review it properly but since you seem to know this stuff inside out I’m fine with it.

pchickey (approved):

crates/cranelift/src/debug/transform/debug_transform_logging.rs
SingleAccretion: Thanks!

This is a bugfix for the debug logging you can turn on to see how DIEs get processed - it now prints the section (global) offsets for all referenced DIEs, allowing one to easily search for them in the editor.

PR #9903, opened 2024-12-26
ISLE: improve syntax BNF

eagr:
* correct `<pattern-arg>` syntax(?)
* re-org items in order to improve readability

### DIFF
--- a/cranelift/isle/docs/language-reference.md
+++ b/cranelift/isle/docs/language-reference.md
@@ -1442,12 +1442,14 @@ The grammar accepted by the parser is as follows:
                | "\t"
                | "\n"
                | "\r"
+               | "\b"
+               | "\f"
 
 <comment> ::= <line-comment> | <block-comment>
 
 <line-comment> ::= ";" <line-char>* (<newline> | eof)
-<newline> ::= "\n" | "\r"
 <line-char> ::= <any character other than "\n" or "\r">
+<newline> ::= "\n" | "\r"
 
 <block-comment> ::= "(;" <block-char>* ";)"
 <block-char> ::= <any character other than ";" or "(">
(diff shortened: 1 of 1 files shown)


cranelift/isle/docs/language-reference.md
```diff
@@ -1442,12 +1442,14 @@ The grammar accepted by the parser is as follows:
                | "\t"
                | "\n"
                | "\r"
+               | "\b"
+               | "\f"
```
eagr, line 1446: for sake of completeness. I suppose they're also ignored by the parser?

cranelift/isle/docs/language-reference.md
```diff
@@ -1458,45 +1460,43 @@ The grammar accepted by the parser is as follows:
 <ISLE> ::= <def>*
 
 <def> ::= "(" "pragma" <pragma> ")"
-        | "(" "type" <typedecl> ")"
+        | "(" "type" <type> ")"
```
eagr, line 1463: let's just use the dumbest mapping :)

cranelift/isle/docs/language-reference.md
```diff
@@ -1458,45 +1460,43 @@ The grammar accepted by the parser is as follows:
...
-        | "(" "extractor" <etor> ")"
+        | "(" "extractor" <extractor> ")"
         | "(" "extern" <extern> ")"
         | "(" "convert" <converter> ")"
 
-// No pragmas are defined yet
+;; No pragmas are defined yet
 <pragma> ::= <ident>
 
-<typedecl> ::= <ident> [ "extern" | "nodebug" ] <typevalue>
```
eagr, line 1471: technically not a *value*, could be confusing

cranelift/isle/docs/language-reference.md
```diff
@@ -1458,45 +1460,43 @@ The grammar accepted by the parser is as follows:
...
-        | [ "-" ] "0o" ( "0".."7" | "_" )+
-        | [ "-" ] "0b" ( "0".."1" | "_" )+
-
-<typevalue> ::= "(" "primitive" <ident> ")"
-              | "(" "enum" <enumvariant>* ")"
+<typedef> ::= "(" "primitive" <ident> ")"
+            | "(" "enum" <enum-variant>* ")"
 
-<enumvariant> ::= <ident>
-                | "(" <ident> <enumfield>* ")"
```
eagr, line 1488: technically it's a field of the variant

cranelift/isle/docs/language-reference.md
```diff
@@ -1508,7 +1508,7 @@ The grammar accepted by the parser is as follows:
             | "(" <ident> <pattern-arg>* ")"
 
 <pattern-arg> ::= <pattern>
-                | "<" <expr>  ;; in-argument to an extractor
```
eagr, line 1511: I could be very wrong about this. But having this kind of infix operator defined in S-expression feels weird and I couldn't find it in the extractor parser, so I think this is a typo?

cranelift/isle/docs/language-reference.md
cfallin: Actually no -- the full list of whitespace characters is [here](https://github.com/bytecodealliance/wasmtime/blob/41ff8e7b6b47ffca748db6fc5ec433d6c69fe486/cranelift/isle/isle/src/lexer.rs#L114). The original list was correct.

cranelift/isle/docs/language-reference.md
cfallin: Eh, there is a reason I had chosen `typedecl` -- `type` is too close to `ty` (the latter reads as an abbreviation of the former with no semantic difference), the existing nonterminal for type expressions, and there is a semantic difference between uses of a type and the form that declares them. Let's stick with `typedecl`.

cranelift/isle/docs/language-reference.md
cfallin: Well, here I would disagree again. You can call it a "type expression" if you want, but it's a value in the universe of types. It is definitely not a `typedef`: `typedef` is, in my mind, another name for the whole form that defines a new type (like a typedef in C/C++ for example). Perhaps "type body" if you want to distinguish the semantic place this occupies -- an expanded form that can be named in a `type` declaration form's right-hand side but not in-line wherever types are given.

cranelift/isle/docs/language-reference.md
cfallin: This is outdated syntax, sorry -- we removed argument polarity a while ago. The whole `<expr>` case needs to go away and you can inline `<pattern>` for `<pattern-arg>` since we don't have the distinction anymore.

cfallin (changes requested):
Thanks for the PR! Unfortunately this introduces some inaccuracies and I have a lot of comments -- see below.

cranelift/isle/docs/language-reference.md
eagr: Should have double checked first, sorry about that.

cranelift/isle/docs/language-reference.md
eagr: Yeah, typedef is worse than typevalue.. I did mean something like "type body".

eagr:
Out of curiosity, is there a reason for `convert` instead of `convertor`, given the `extractor`?

cfallin (approved):

cfallin:
> Out of curiosity, is there a reason for `convert` instead of `convertor`, given the `extractor`?

Locally it made sense as "convert T to U using foo"; one could debate the many options further but we already had that discussion three(?) years ago and it's unlikely to change now absent good reason.

PR #9904, opened 2024-12-27
pulley: Implement 16x8 arithmetics

eagr:
Helps #9783


pulley/src/interp.rs
```diff
@@ -4282,4 +4328,15 @@ impl ExtendedOpVisitor for Interpreter<'_> {
         self.state[operands.dst].set_f64x2(a);
         ControlFlow::Continue(())
     }
+
+    fn vavground16x8(&mut self, operands: BinaryOperands<VReg>) -> ControlFlow<Done> {
+        let mut a = self.state[operands.src1].get_u16x8();
+        let b = self.state[operands.src2].get_u16x8();
+        for (a, b) in a.iter_mut().zip(&b) {
+            // rounding average
+            *a = (*a & *b) + ((*a ^ *b) >> 1) + ((*a ^ *b) & 1);
```
alexcrichton, line 4337: Could this perhaps be written as `*a = (u32::from(a) + u32::from(b) + 1) / 2`? Basically the more classical version of an average which explicitly takes advantage of wider-precision arithmetic to avoid overflow issues.

alexcrichton (approved):
Thanks for this!

pulley/src/interp.rs
eagr: instantly convinced :))

PR #9905, opened 2024-12-27
ISLE: Polish reference

eagr:
Basically just typo fixing and more succinct wording, without any intent to change the original meaning.

### DIFF
--- a/cranelift/isle/docs/language-reference.md
+++ b/cranelift/isle/docs/language-reference.md
@@ -2,7 +2,7 @@
 
 This document will describe ISLE (Instruction Selection Lowering
 Expressions), a DSL (domain-specific language) that we have developed
-in order to help us express certain parts of the Cranelift compiler
+to help us express certain parts of the Cranelift compiler
 backend more naturally. ISLE was first [described in RFC
 #15](https://github.com/bytecodealliance/rfcs/pull/15) and now is used
 by and lives in the Cranelift tree in
@@ -14,9 +14,9 @@ Documentation on how ISLE is used in Cranelift can be found
 ## Intro and Whirlwind Tour: DSL for Instruction Lowering
 
 The goal of ISLE is to represent *instruction lowering patterns*. An
-instruction lowering pattern is a specification that a certain
-combination of operators in the IR (CLIF), when combined under certain
-conditions, can be compiled down into a certain sequence of machine
+instruction lowering pattern is a specification describing certain
+operators in the IR (CLIF) when combined under certain
+conditions, can be compiled down to a certain sequence of machine
 instructions. For example:
 
 - An `iadd` (integer add) operator can always be lowered to an x86
@@ -26,7 +26,7 @@ instructions. For example:
   be lowered to an x86 `ADD` instruction with a register and an
   immediate.
 
-One could write something like the following in ISLE (simplified from
+One can write something like the following in ISLE (simplified from
 the real code [here](https://github.com/bytecodealliance/wasmtime/blob/main/cranelift/codegen/src/isa/x64/lower.isle)):
 
 ```lisp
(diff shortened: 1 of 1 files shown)


eagr:
Couldn't seem to find the spec for `decl multi` in the reference. Is it forgotten or like TBD?

cranelift/isle/docs/language-reference.md
```diff
@@ -14,9 +14,9 @@ Documentation on how ISLE is used in Cranelift can be found
 ## Intro and Whirlwind Tour: DSL for Instruction Lowering
 
 The goal of ISLE is to represent *instruction lowering patterns*. An
-instruction lowering pattern is a specification that a certain
-combination of operators in the IR (CLIF), when combined under certain
-conditions, can be compiled down into a certain sequence of machine
+instruction lowering pattern is a specification describing certain
+operators in the IR (CLIF) when combined under certain
+conditions, can be compiled down to a certain sequence of machine
```
cfallin, line 19: This rewording is incorrect. "A ... pattern is a specification that a combination of operators ... can be compiled" can't be rewritten into "A ... pattern is a specification describing certain operators ... can be compiled". The first (current) form has a subordinate clause describing what the specification does, and the verb indicates that this specification can be compiled. Your rewrite somehow attaches two actions onto "specification" -- it describes, and it can be compiled. These aren't semantically equivalent.

cranelift/isle/docs/language-reference.md
```diff
@@ -26,7 +26,7 @@ instructions. For example:
   be lowered to an x86 `ADD` instruction with a register and an
   immediate.
 
-One could write something like the following in ISLE (simplified from
+One can write something like the following in ISLE (simplified from
```
cfallin, line 29: This rewrite loses the meaning: the whole stanza is of the form "one could ... [but] ISLE lets you ..." and changing "could" to "can" completely changes the meaning.

cranelift/isle/docs/language-reference.md
```diff
@@ -262,19 +262,19 @@ subterms:
 * `_` is a wildcard and matches anything, without capturing it.
 
 * Primitive constant values, such as `42` or `$Symbol`, match only if
-  the term is exactly equal to this constant.
+  the term is an exact equivalent.
```
cfallin, line 265: "an exact equivalent" does not read the same as "is exactly equal" to me: the former implies something about an equivalence class, different forms with the same underlying denotation, etc., whereas for primitive values (e.g. integers) we really do want to say "equal". I'm not sure what "equivalent" does better here.

cranelift/isle/docs/language-reference.md
```diff
@@ -262,19 +262,19 @@ subterms:
...
 
-These pattern-matching operators can be combined, so we could write,
+These pattern-matching operators can be combined, so we can write,
 for example, `(A (B x _) z)`. This pattern would match the term `(A (B
 1 2) 3)` but not `(A (C 4 5) 6)`.
 
-A pattern can properly be seen as a partial function from input term
-to captured (bound) variables: it either matches or it doesn't, and if
+A pattern can probably be seen as a partial function from the input term
```
cfallin, line 271: One cannot replace "properly" with "probably" -- that completely changes the meaning.

cfallin (changes requested):
Hi @eagr -- some comments below; I stopped reading after the last one, about 15% of the way through. Most of these edits are changing meaning in fairly important ways, and the ones that don't seem to be minor style things / subjective preferences, so I think I'd prefer not to accept an edit pass on this document -- given the magnitude of your changes I don't want to spend the time it would take to iterate back to fully correct wording. Thanks anyway, I'm sure you intended well. I'll go ahead and close this PR.

PR #9906, opened 2024-12-27
fix typo

JackDrogon:
fix typo

### DIFF
--- a/cranelift/codegen/src/isa/pulley_shared/abi.rs
+++ b/cranelift/codegen/src/isa/pulley_shared/abi.rs
@@ -297,7 +297,7 @@ where
     /// common cases and we don't want to spread the logic over multiple
     /// functions.
     ///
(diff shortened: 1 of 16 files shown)


pchickey (approved):
Thanks. Do you have automation that finds these, and if so, how many false positives did you filter out? We could stand to add something like that to CI if it’s not going to flag too many false positives.

JackDrogon:
> Thanks. Do you have automation that finds these, and if so, how many false positives did you filter out? We could stand to add something like that to CI if it’s not going to flag too many false positives.谢谢。您是否有自动功能来查找这些内容，如果有，您过滤掉了多少误报？如果 CI 不会标记太多误报，我们可以向 CI 添加类似的东西。

I use misspell and spellchecker to do this， only filter `strate` keyword of proptest and some typo in `crates/wasi-nn/examples/`.

PR #9910, opened 2024-12-28
pulley: implement vector shuffle

eagr:
part of #9783


cranelift/codegen/src/isa/pulley_shared/lower.isle
```diff
@@ -1369,7 +1369,12 @@
 (rule (lower (scalar_to_vector a @ (value_type $F64)))
   (pulley_vinsertf64 (pulley_vconst128 0) a 0))
 
+;;;; Rules for `shuffle` ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(rule (lower (has_type $I8X16 (shuffle a b (u128_from_immediate mask))))
+  (let ((vmask VReg (pulley_vconst128 mask)))
+    (pulley_vshuffle a b vmask)))
```
alexcrichton, line 1376: Instead of putting the `mask` into a `VReg` could it be a `u128` immediate to the `vshuffle` instruction itself? That should cut down on opcodes slightly and more closely matches the wasm semantics of the instruction.

eagr: > could it be a `u128` immediate to the `vshuffle` instruction itself?

I remember not, but let me try again

eagr: `vshuffle(VReg, VReg, VReg, u128)` would raise this error

```
thread '<unnamed>' panicked at cranelift/codegen/src/isa/pulley_shared/inst/emit.rs:109:9:
encoded inst InstAndKind { inst: Raw { raw: VShuffle { dst: Writable { reg: VReg(p0v) }, src1: VReg(p0v), src2: VReg(p1v), mask: 41362427191743139026751447860679676176 } }, kind: PhantomData<cranelift_codegen::isa::pulley64::Pulley64> } longer than worst-case size: length: 22, Inst::worst_case_size() = 20
```

eagr: If it's ok, I guess I could raise the inst. size limit.

alexcrichton (approved):

eagr:
Do you happen to have something along the Pully pipeline that could use an extra pair of hands? I should probably leave these exercises for others. Btw, this is a wonderful setup for helping people learn about how Pulley works :)

alexcrichton:
Sure yeah, whatever suits you best! Perhaps the next-lowest-hanging-fruit is performance optimizations which will involve running benchmarks and understanding the performance of Pulley and where hot-spots are. The main goal would be identifying macro-opcodes that can be added for lowering. The first one off the top of my head would be a macro opcode related to bounds checks to make memory accesses faster, but I haven't done performance profiling one way or another on that yet. (we also don't have many docs yet on how to do this so it's a choose-your-own-adventure at the moment, so not necessarily great as a first project)

PR #9911, opened 2024-12-29
pulley: Implement vector sqmul_round_sat

eagr:
part of #9783


pulley/src/interp.rs
```diff
@@ -3835,6 +3835,19 @@ impl ExtendedOpVisitor for Interpreter<'_> {
         ControlFlow::Continue(())
     }
 
+    fn vqmulrsi16x8(&mut self, operands: BinaryOperands<VReg>) -> ControlFlow<Done> {
+        let mut a = self.state[operands.src1].get_i16x8();
+        let b = self.state[operands.src2].get_i16x8();
+        const MIN: i32 = i16::MIN as i32;
+        const MAX: i32 = i16::MAX as i32;
+        for (a, b) in a.iter_mut().zip(b) {
+            let r = i32::from(*a) * i32::from(b) + (1 << 14) >> 15;
```
alexcrichton, line 3844: I'm a bit surprised by the lack of parentheses around this, I would have expected:

```rust
let r = (i32::from(*a) * i32::from(b) + (1 << 14)) >> 15;
```

as otherwise the final term `(1 << 14) >> 15` looks like it currently evaluates to 0.

Tests, however, are passing with this PR which means that either the precedence rules of parsing in Rust are different from what I expect or that the tests don't actually cover the case which needs the parentheses to be right here. Would you be up for debugging this a bit to figure out what's going on test-wise?

eagr: You must have always been in the good habit of applying parentheses when in doubt so you don't have to remember the precedence rules :)

I think this is equivalent to what you expect. Let me just add those parentheses and you'd see the same result. The code reads better that way anyways. And btw the tests are looking good to me.

alexcrichton: Oh nice, thanks for checking!

alexcrichton (approved):

PR #9912, opened 2024-12-29
pulley: Implement iadd_pairwise

eagr:
part of #9783


pulley/src/interp.rs
```diff
@@ -3299,6 +3299,40 @@ impl ExtendedOpVisitor for Interpreter<'_> {
...
+        let a = self.state[operands.src1].get_i16x8();
+        let b = self.state[operands.src2].get_i16x8();
+        let result = a
+            .chunks(2)
+            .chain(b.chunks(2))
+            .map(|pair| {
+                let [h, t]: [_; 2] = pair.try_into().unwrap();
+                h.wrapping_add(t)
+            })
+            .collect::<Vec<_>>()
```
alexcrichton, line 3312: For these instructions we'll definitely want to avoid an intermediate allocation of a `Vec`. It might be reasonable to just use manual indices here and/or "macro expanding" some code. I don't think the standard library helpers for all this are necessarily the best way to go.

Xuanwo: We only have limited numbers of items to operate, maybe we can just write them out like:

```rust
fn vaddpairwisei16x8_s(&mut self, operands: BinaryOperands<VReg>) -> ControlFlow<Done> {
    let a = self.state[operands.src1].get_i16x8();
    let b = self.state[operands.src2].get_i16x8();
    
    let mut result = [0i16; 8];
    
    // Process first 4 elements from a
    result[0] = a[0].wrapping_add(a[1]);
    result[1] = a[2].wrapping_add(a[3]);
    result[2] = a[4].wrapping_add(a[5]);
    result[3] = a[6].wrapping_add(a[7]);
    
    // Process first 4 elements from b 
    result[4] = b[0].wrapping_add(b[1]);
    result[5] = b[2].wrapping_add(b[3]);
    result[6] = b[4].wrapping_add(b[5]);
    result[7] = b[6].wrapping_add(b[7]);

    self.state[operands.dst].set_i16x8(result);
    ControlFlow::Continue(())
}
```

eagr: Guess I need a mind shift for this :)

alexcrichton (approved):

PR #9913, opened 2024-12-29
Fix incorrect frame code offset in coredump

xtuc:
Currently, coredump include incorrect code offsets, the generation relies on `func_offset()` which returns the instruction's offset relative to the function.

This change uses `module_offset()` to get the absolute code offset and, using a new method `get_relative_codeoffset`, converts the offset to be relative to the code section.

Also, the `name` property has been removed because it's misleading when displayed and not actually useful.


xtuc:
Sorry seems like I misremembered https://github.com/WebAssembly/tool-conventions/blob/main/Coredump.md#threads-and-stack-frames... Code offsets are relative to the function start. Closing this since it's already correct.

PR #9919, opened 2025-01-04
jit module return code slice

MarinPostma:
It can be usefull to get the compiled function as a slice from a `JITModule`, for example, if one wants to disassemble it, to inspect the assembled code. It is currently possible to get pseudo assembly from the module context, by explicitely requesting `ctx.want_disasm`, but it is not very practical and it is not the _actual_ machine code.

### DIFF
--- a/cranelift/jit/src/backend.rs
+++ b/cranelift/jit/src/backend.rs
@@ -408,11 +408,7 @@ impl JITModule {
         }
     }
 
-    /// Returns the address of a finalized function.
-    ///
-    /// The pointer remains valid until either [`JITModule::free_memory`] is called or in the future
-    /// some way of deallocating this individual function is used.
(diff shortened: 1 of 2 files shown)


bjorn3:
If you want the actual machine code you can already find it in `context.compiled_code().unwrap().buffer.data()` after you use `define_function`. You can also find all relocations in `context.compiled_code().unwrap().buffer.relocs()` to know which functions you are actually calling. cranelift-jit throws all relocations away after applying them as it doesn't need them anymore after applying them.

MarinPostma:
@bjorn3 but that's only for the current function no? Is there a way to get the buffer for any function in a module?

bjorn3:
Yeah, you have to do that right after calling define_function for the function which you want machine code for.

MarinPostma:
I'll close then 👍

Issue #9932, opened 2025-01-06
CI silently swallows errors and tests forget to check return values 


ifsheldon:
Here is a simple example that has blocked my PR #9788 for weeks. 

In this PR, I need to use `wasm32-wasip2` target, which was not added here
https://github.com/bytecodealliance/wasmtime/blob/fc7ef8d62e196df90522b26ed055bc21667cd19a/.github/actions/install-rust/action.yml#L67
which is used by CI.

My PR added L69 to compile an example with `wasm32-wasip2` target like below:

<img width="1199" alt="image" src="https://github.com/user-attachments/assets/f5780e6a-0e2e-4b67-a85c-b576ec431b38" />

This `create_rust_wasm(wasi wasm32-wasip2)` should have failed CI but not, because it didn't checked the return code of the command that was run.

The command should have returned something like 
```shell
cargo build -p example-wasi-wasm --target wasm32-wasip2
   Compiling example-wasi-wasm v0.0.0 (/Users/zhiqiu/offline_code/opensource/wasmtime/examples/wasm)
error[E0463]: can't find crate for `std`
  |
  = note: the `wasm32-wasip2` target may not be installed
  = help: consider downloading the target with `rustup target add wasm32-wasip2`

For more information about this error, try `rustc --explain E0463`.
error: could not compile `example-wasi-wasm` (bin "wasi") due to 1 previous error
```

This error was not checked, so a test later failed with misleading error messages, because this step didn't compiled a required manifest file (`wasi.wasm`).

I don't know if this case is a special case or there are loopholes in CI in which CMake is used but it does not checked a return code of a command that is run.

PR #9933, opened 2025-01-06
Add return code checking in CMakeList.txt in examples

ifsheldon:
This attempts to partially solve #9932, adding checking return code of the executed command `cargo build`, which can fail, for example, if a target has not been added.

I don't know whether there are similar cases somewhere else, in which the return code of an executed command is simply ignored.

In this file, however, when I asked Claude "are there anything else that will cause silent failures in this file", it says the following. I think what it says makes sense, but I don't know much about cmake, so I don't know if it's correct.
<details>
Yes, there are a few potential sources of silent failures in this CMake file:

1. The `create_rust_test` function also uses `cargo run` without checking return codes:
```cmake
function(CREATE_RUST_TEST EXAMPLE)
    # These tests might fail silently during test execution
    if(ARGC GREATER 1)
        add_test(NAME ${EXAMPLE}-rust COMMAND cargo run --example ${EXAMPLE} --features ${ARGV1} WORKING_DIRECTORY ../..)
    else()
        add_test(NAME ${EXAMPLE}-rust COMMAND cargo run --example ${EXAMPLE} WORKING_DIRECTORY ../..)
    endif()
endfunction()
```

2. The `add_subdirectory` command at the beginning might fail silently if the directory doesn't exist:
```cmake
add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/../crates/c-api ${CMAKE_CURRENT_BINARY_DIR}/wasmtime)
```

3. In the `CREATE_TARGET` function, the linking against CoreFoundation on Apple platforms doesn't verify if the framework exists:
```cmake
if(APPLE)
    target_link_libraries(wasmtime-${TARGET} PRIVATE "-framework CoreFoundation")
endif()
```

To make the file more robust, you could:

1. Add existence checks for required directories and files
2. Add `REQUIRED` to framework findings
3. Set test properties to fail on non-zero return codes:

```cmake
function(CREATE_RUST_TEST EXAMPLE)
    if(ARGC GREATER 1)
        add_test(NAME ${EXAMPLE}-rust 
                COMMAND cargo run --example ${EXAMPLE} --features ${ARGV1}
                WORKING_DIRECTORY ../..)
    else()
        add_test(NAME ${EXAMPLE}-rust 
                COMMAND cargo run --example ${EXAMPLE}
                WORKING_DIRECTORY ../..)
    endif()
    # Make the test fail if cargo returns non-zero
    set_tests_properties(${EXAMPLE}-rust PROPERTIES WILL_FAIL FALSE)
endfunction()
```

Also, it's worth noting that the relative paths (`../..`) in `WORKING_DIRECTORY` could be fragile if the directory structure changes. It would be more robust to use absolute paths or CMake variables to reference directories.
</details>

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -38,7 +38,13 @@ function(CREATE_RUST_TEST EXAMPLE)
 	endif()
 endfunction()
 function(CREATE_RUST_WASM EXAMPLE TARGET)
-	execute_process(COMMAND cargo build -p example-${EXAMPLE}-wasm --target ${TARGET})
+	execute_process(
+        COMMAND cargo build -p example-${EXAMPLE}-wasm --target ${TARGET}
+        RESULT_VARIABLE CARGO_RESULT
+    )
+    if(NOT CARGO_RESULT EQUAL 0)
+        message(FATAL_ERROR "Failed to build Rust WASM target: ${EXAMPLE} for ${TARGET}")
+    endif()
 endfunction()
 
 # Enable testing


alexcrichton:
Thanks! I think that `execute_process` probably isn't the right thing here but rather it's best to set up a "target" (I think that's the right CMake terminology) which is based on executing the process instead. That would also defer execution to the test step rather than running it during the configure step as well.

If you're not comfortable with CMake though I'm happy to make this change too. (I don't know what/who Claude is myself)

ifsheldon:
Yeah, please help, as this change might affect many test cases checking examples working.

PR #9937, opened 2025-01-06
Enable the memory64 proposal by default

alexcrichton:
The memory64 proposal for WebAssembly moved to phase 4 early last November which was the final remaining blocker for enabling it in Wasmtime. I've gone ahead and enabled it here with all the other checkboxes being ticked such as:

* Tests - all spec tests are enabled and we have a few tests here and there for memory64 behavior throughout the unit test suite.

* Finished - this proposal's memory-facing bits have been done for quite some time and the final 64-bit table bits have been done for a bit now as well.

* Fuzzed - this is enabled in `wasm-smith` and additionally has custom fuzzing via the `memory_accesses` fuzzer.

* API - all APIs related to memory work with `u64` values to accommodate 64-bit memories and the 64-bit-ness is reflected in the type of memories.

* C API - the C API's functions for working with memories all reflect 64-bit indices like the Rust API.

### DIFF
--- a/crates/wasmtime/src/config.rs
+++ b/crates/wasmtime/src/config.rs
@@ -2023,6 +2023,7 @@ impl Config {
         features |= WasmFeatures::RELAXED_SIMD;
         features |= WasmFeatures::TAIL_CALL;
         features |= WasmFeatures::EXTENDED_CONST;
+        features |= WasmFeatures::MEMORY64;
         // NB: if you add a feature above this line please double-check
         // https://docs.wasmtime.dev/stability-wasm-proposals.html
         // to ensure all requirements are met and/or update the documentation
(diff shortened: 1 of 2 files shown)


dicej (approved):
LGTM.  Do we need to do anything special (that hasn't already been done) to indicate that memory64 is not yet compatible with components?  Or will that be caught by `wasmparser` at validation time?

alexcrichton:
Oh you know that's a good point. I haven't checked up on the state of memory64 and components in awhile. I know some parts support some things but not all, and I suspect that there are a number of memory64-related panics in Wasmtime which I should convert to errors first before landing this.

alexcrichton:
On review looks like less has support for this than I thought. I've opted to go with a [wasmparser validation change](https://github.com/bytecodealliance/wasm-tools/pull/1970) for now since components [don't currently specify](https://github.com/WebAssembly/component-model/issues/22) how everything works anyway.

alexcrichton:
I'm going to flag this for merging now that wasmparser has been updated and components reject memory64-based memories explicitly.

PR #9941, opened 2025-01-07
Replace `signals-based-traps` with auto-detection

alexcrichton:
This commit refactors the platform support of the `wasmtime` crate itself to remove the previously added `signals-based-traps` feature in favor of auto-detecting whether it's there or not. The `build.rs` script for the `wasmtime` crate will now detect the target platform and auto-enable this feature as necessary.

The `signals-based-traps` cargo feature is removed and split into two custom `#[cfg]` directives that the build script sets:

* `has_virtual_memory` - this is used to gate mmap implementations for example. This is enabled on `unix || windows` and will be off for `no_std` targets for example. This is split out of "signals-based-traps" to better handle platforms like iOS which have virtual memory but don't execute native code (removing the need for native signals).

* `has_native_signals` - gates signal handlers on Unix for example. This is disabled on MIRI but otherwise enabled for `unix || windows`. This is intended to in the future get disabled for iOS by default for example since it's not necessary when using Pulley. This is additionally off-by-default for `no_std` platforms.

Two new crate features were added for `no_std` or "custom" platforms to opt-in to the `wasmtime-platform.h` C APIs for implementing virtual memory and signals. These are used in the `min-platform` embedding example.

This commit additionally updates some various documentation here and there to be more up-to-date.


alexcrichton:
One of the rough high-level goals here is that in conjunction with https://github.com/bytecodealliance/wasmtime/pull/9837 I hope to disable `has_native_signals` on Pulley-only platforms in the near future to statically remove all that code. That should make Wasmtime more portable by default since when there's no Cranelift backend there's no need to install signal handlers which often has lots of platform/architecture-specific implications.

crates/wasmtime/build.rs
```diff
@@ -1,27 +1,52 @@
...
+    }
+
+    let unix = cfg("unix");
+    let windows = cfg("windows");
+    let miri = cfg("miri");
+    let supported_platform = unix || windows;
+
+    let has_native_signals =
+        !miri && (supported_platform || cfg!(feature = "custom-native-signals"));
+    let has_virtual_memory = supported_platform || cfg!(feature = "custom-virtual-memory");
```
bjorn3, line 17: wasm32-unknown-emscripten is marked as unix, but has neither signals, nor virtual memory. I guess it is not all too important that Wasmtime works on Emscripten though.

alexcrichton: That's a good point yeah, although this `unix || windows` logic matches [what's currently in the `sys` module](https://github.com/bytecodealliance/wasmtime/blob/51e28c757e7dbeb434afddcf3961c35959408ba9/crates/wasmtime/src/runtime/vm/sys/mod.rs#L49-L63) so we technically already didn't support Emscripten. It should be reasonable though to add support here though in the future.

fitzgen (approved):
LGTM! I'm glad that this leaves the ability to turn off signals usage at engine config-time, even if the target supports it, as I think that is a use case we need to support.

PR #9943, opened 2025-01-07
pulley: Add some macro-instructions related to bounds-checks

alexcrichton:
This commit starts down the path of optimizing wasm loads/stores in Pulley with macro-instructions. It's expected that these instructions are so common that it's worth putting them in the 1-byte namespace of opcodes.

Locally this gets a 10% speedup on the sightglass bz2 benchmark.

### DIFF
--- a/tests/disas/pulley/pulley32_memory32.wat
+++ b/tests/disas/pulley/pulley32_memory32.wat
@@ -1,0 +1,109 @@
+;;! target = "pulley32"
+;;! test = "compile"
+
+(module
+  (memory 1)
+
+  (func $load8 (param i32) (result i32)
+    (i32.load8_u (local.get 0)))
+
+  (func $load16 (param i32) (result i32)
+    (i32.load16_u (local.get 0)))
+
+  (func $load32 (param i32) (result i32)
+    (i32.load (local.get 0)))
+
+  (func $load64 (param i32) (result i64)
+    (i64.load (local.get 0)))
+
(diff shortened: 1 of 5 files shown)


alexcrichton:
This additionally shrinks spidermonkey.cwasm from 28M to 21M

tests/disas/pulley/pulley32_memory32.wat
```diff
@@ -0,0 +1,110 @@
...
+  (func $load32_offset (param i32) (result i32)
+    (i32.load offset=32 (local.get 0)))
+
+  (func $load64_offset (param i32) (result i64)
+    (i64.load offset=32 (local.get 0)))
+)
+;; wasm[0]::function[0]::load8:
+;;       push_frame
+;;       xload32le_offset8 x6, x0, 52
+;;       br_if_xulteq32 x6, x2, 0x14    // target = 0x19
```
abrown, line 34: We don't want this one to use the new `xbc` macro-op?

alexcrichton: No we probably do but I wanted to add this here for complete-ness to show that it's not currently using the new op. One-byte loads/stores are generated a bit differently with bounds checks and use a `>=` condition instead of `>` + arithmetic to test whether something is out of bounds. That made it so it doesn't fall into the patterns I added and would need a specific opcode just for the one-byte access. I figured it's probably fine to leave that off for now and add in the future if necessary

PR #9944, opened 2025-01-07
pulley: Implement the wide-arithmetic proposal

alexcrichton:
Add a few minor instructions/lowerings for the new operations added as part of the wide-arithmetic proposal. These are all part of the "extended" opcode set since they shouldn't be common and if they're performance critical you probably want a native backend instead.

### DIFF
--- a/pulley/src/lib.rs
+++ b/pulley/src/lib.rs
@@ -1279,7 +1279,7 @@ macro_rules! for_each_extended_op {
             /// `dst = ieee_fma(a, b, c)`
             vfma64x2 = Vfma64x2 { dst: VReg, a: VReg, b: VReg, c: VReg };
 
-            /// `dst_lo:dst_hi = lhs_lo:lhs_hi + rhs_lo:rhs_hi`
+            /// `dst_hi:dst_lo = lhs_hi:lhs_lo + rhs_hi:rhs_lo`
             xadd128 = Xadd128 {
                 dst_lo: XReg,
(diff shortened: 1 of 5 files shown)


dicej (commented):
I'm new to Pulley and don't have much context for this, but LGTM.  I'll leave it for someone with more experience to to approve.

pulley/src/lib.rs
```diff
@@ -1278,6 +1278,39 @@ macro_rules! for_each_extended_op {
             vfma32x4 = Vfma32x4 { dst: VReg, a: VReg, b: VReg, c: VReg };
             /// `dst = ieee_fma(a, b, c)`
             vfma64x2 = Vfma64x2 { dst: VReg, a: VReg, b: VReg, c: VReg };
+
+            /// `dst_lo:dst_hi = lhs_lo:lhs_hi + rhs_lo:rhs_hi`
```
cfallin, line 1282: maybe a style thing, or Verilog influence, or something, but usually I write `a:b` concat syntax MSB-first (as the concrete values would be written) -- so `dst_hi:dst_lo = ...` maybe?

cfallin (approved):
LGTM!

pulley/src/lib.rs
alexcrichton: It's ok I have no influence so I think I swap lo:hi every time I write it, happy to canonicalize to hi:lo!

PR #9947, opened 2025-01-08
Make module validation deterministic in case of multiple errors

michael-weigelt:
If a Wasm binary is invalid in several ways, parallel compilation will return a non-deterministic result. We would like to get a deterministic result even in the presence of multiple errors. 

This PR addresses that by first materializing all validation results and then returning the first error, if any ('first' with regard to the order of the input vector).

The downside is that we cannot return early when an error is encountered. This slows down the validation for invalid binaries. Since that is not the happy path anyway, I propose to accept the slowdown.  

The provided test does not prove determinism, which is hard to do. But it gives _some_ confidence and ensures the behaviours of parallel and sequential compilation do not deviate.

### DIFF
--- a/crates/wasmtime/src/engine.rs
+++ b/crates/wasmtime/src/engine.rs
@@ -154,9 +154,14 @@ impl Engine {
             #[cfg(feature = "parallel-compilation")]
             {
                 use rayon::prelude::*;
+                // If we collect into Result<Vec<B>, E> directly, the returned error is not
+                // deterministic, because any error could be returned early. So we first materialize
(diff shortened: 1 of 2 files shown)


alexcrichton (approved):
Thanks for this!

alexcrichton:
Ah the timeout there looks like the new `validate_deterministic` test is taking awhile in miri, so mind putting `#[cfg(not(miri))]` on the test?

alexcrichton:
er, actually, `#[cfg_attr(miri, ignore)]` would probably be better

michael-weigelt:
@alexcrichton done, and I made the test smaller too.

Issue #8093, opened 2024-03-12
winch(x64): SIMD


saulecabrera:
Winch is missing support for most of finished WebAssembly proposals. The goal of this issue is to track support for the [SIMD proposal](https://github.com/webassembly/simd)

- [x] `v128.load8x8_s` @jeffcharles 
- [x] `v128.load8x8_u` @jeffcharles 
- [x] `v128.load16x4_s` @jeffcharles 
- [x] `v128.load16x4_u` @jeffcharles 
- [x] `v128.load32x2_s` @jeffcharles 
- [x] `v128.load32x2_u` @jeffcharles 
- [x] `v128.load8_splat` @jeffcharles 
- [x] `v128.load16_splat` @jeffcharles 
- [x] `v128.load32_splat` @jeffcharles 
- [x] `v128.load64_splat` @jeffcharles 
- [x] `i8x16.shuffle` @jeffcharles 
- [x] `i8x16.swizzle` @jeffcharles 
- [x] `i8x16.splat` @jeffcharles 
- [x] `i16x8.splat` @jeffcharles 
- [x] `i32x4.splat` @jeffcharles 
- [x] `i64x2.splat` @jeffcharles 
- [x] `f32x4.splat` @jeffcharles 
- [x] `f64x2.splat` @jeffcharles 
- [x] `i8x16.extract_lane_s` @jeffcharles 
- [x] `i8x16.extract_lane_u`  @jeffcharles 
- [x] `i8x16.replace_lane`  @jeffcharles 
- [x] `i16x8.extract_lane_s`  @jeffcharles
- [x] `i16x8.extract_lane_u`  @jeffcharles 
- [x] `i16x8.replace_lane`  @jeffcharles 
- [x] `i32x4.extract_lane`  @jeffcharles 
- [x] `i32x4.replace_lane`  @jeffcharles 
- [x] `i64x2.extract_lane`  @jeffcharles 
- [x] `i64x2.replace_lane`  @jeffcharles 
- [x] `f32x4.extract_lane`  @jeffcharles 
- [x] `f32x4.replace_lane`  @jeffcharles 
- [x] `f64x2.extract_lane`  @jeffcharles 
- [x] `f64x2.replace_lane`  @jeffcharles 
- [x] `i8x16.eq`@jeffcharles 
- [x] `i8x16.ne`@jeffcharles 
- [x] `i8x16.lt_s`@jeffcharles 
- [x] `i8x16.lt_u` @jeffcharles 
[...]

PR #9950, opened 2025-01-08
Winch: Add SIMD load and extend and load and splat instructions for x64

jeffcharles:
Part of #8093. Adds support on x64 with AVX2 for:
- `v128.load8x8_s`
- `v128.load8x8_u`
- `v128.load16x4_s`
- `v128.load16x4_u`
- `v128.load32x2_s`
- `v128.load32x2_u`
- `v128.load8_splat`
- `v128.load16_splat`
- `v128.load32_splat`
- `v128.load64_splat`

I've changed `wasm_load` on the macroassemblers to take a `LoadKind` instead of `Option<ExtendKind>` to model the additional vector operations (vector extends and splats), but I'm open to changing that to a different abstraction.


winch/codegen/src/isa/x64/asm.rs
```diff
@@ -471,6 +472,90 @@ impl Assembler {
...
+        assert!(dst.to_reg().is_float());
+
+        let op = match ext {
+            VectorExtendKind::V128Extend8x8S => SseOpcode::Pmovsxbw,
+            VectorExtendKind::V128Extend8x8U => SseOpcode::Pmovzxbw,
+            VectorExtendKind::V128Extend16x4S => SseOpcode::Pmovsxwd,
+            VectorExtendKind::V128Extend16x4U => SseOpcode::Pmovzxwd,
+            VectorExtendKind::V128Extend32x2S => SseOpcode::Pmovsxdq,
+            VectorExtendKind::V128Extend32x2U => SseOpcode::Pmovzxdq,
+        };
```
saulecabrera, line 492: Given that we're starting with avx2 as our baseline, I wonder if we should emit the `vpmov*` variants here.

winch/codegen/src/masm.rs
```diff
@@ -226,6 +226,36 @@ impl ExtendKind {
...
+    /// Sign extends two 32 bit integers to two 64 bit lanes.
+    V128Extend32x2S,
+    /// Zero extends two 32 bit integers to two 64 bit lanes.
+    V128Extend32x2U,
+}
+
+/// Kinds of behavior supported by Wasm loads.
+pub(crate) enum LoadKind {
+    /// Do not extend or splat.
+    None,
```
saulecabrera, line 250: Nit: can we remove this enum entry and keep `Option<LoadKind>` instead? Or where you envisioning this approach to be easier to consume in the callee as opposed to using an `Option`?

saulecabrera (commented):
Looking good! Left some comments inline.

winch/codegen/src/isa/x64/asm.rs
jeffcharles: We could. I figured if they are equivalent instructions that are supported on older chipsets, we could emit those so there would be less work to support non-AVX CPUs in the future given we only need to support 128-bit vectors. One thing I'm not clear on is if `Vpmov*` instructions are faster or not. If they are faster, then it would probably make sense to emit those instead.

saulecabrera: My expectation is that they are probably more performant, but I haven't performed any meaningful benchmarks. Aside from the potential performance benefit, my comment here is for consistency mostly: given that we aren't planning on doing any checks/emission for anything other than avx2 for SIMD at least initially, I'd prefer if we could default to emitting avx2 only to start. My expectation is that when thinking about fallbacks for anything < avx2, it will require substantial work regardless given that we'd probably want to find a way to share that implementation with what already exists in Cranelift.

alexcrichton: It's been a bit since I last checked, and I also haven't benchmarked, but IIRC the Intel optimization guide recommended either using exclusively VEX instructions (ones with the `v*` prefix) or exclusively non-VEX instructions. It's got something subtle to do with CPU state and such. It generally has to do with the fact that VEX instructions mutate the entire register contents as opposed to SSE which only mutate the lower 128-bits. This means that SSE instructions can accidentally create false dependencies between instructions.

The rule of thumb I used in Cranelift at least was to try to unconditionally prefer `v*` VEX instructions when AVX is enabled as I think that's what Intel assumes compilers are doing to make things faster

saulecabrera (approved):
Thanks!

crates/wast-util/src/lib.rs
```diff
@@ -526,6 +525,16 @@ impl WastTest {
             if unsupported.iter().any(|part| self.path.ends_with(part)) {
                 return true;
             }
+
+            // SIMD on Winch requires AVX instructions.
+            #[cfg(target_arch = "x86_64")]
+            if !(std::is_x86_feature_detected!("avx") && std::is_x86_feature_detected!("avx2")) {
+                let unsupported = ["spec_testsuite/simd_align.wasm"];
```
saulecabrera, line 532:
```suggestion
                let unsupported = ["spec_testsuite/simd_align.wast"];
```

This is why the PR is still failing in CI. I tested the change in my arm64 machine FWIW.

PR #9958, opened 2025-01-09
Remove usage of `gen` as an identifier

alexcrichton:
This commit is a start to some preparation for the Rust 2024 edition for Wasmtime and this workspace. The `rust-2024-compatibility` lint group in rustc, currently off-by-default, is intended to assist with migrating code to prepare for the 2024 edition. Some of those lints though are, in my opinion, far too noisy to be turned on so this PR doesn't turn on the whole group. Instead though I plan on enabling individual lints over time in our `Cargo.toml` before the 2024 edition is enabled. This should hopefully provide a relatively smooth and less churn-y path to enabling the 2024 edition in the future.

The first lint enabled here in this commit is the `keyword_idents_2024` lint which warns against usage of identifiers that will become keywords in the 2024 edition. The only one affecting Wasmtime is the `gen` identifier (soon to be keyword) and we had quite a few instances of it. Where possible I've renamed to `generator` or `generate` or `generated` but when used as methods from upstream crate traits (e.g. `RngCore::gen`) we're forced to use `r#gen`. The `rand` crate will be updated in 0.9.0 to avoid this keyword so this shouldn't be permanent.

### DIFF
--- a/crates/wasi-common/src/sync/mod.rs
+++ b/crates/wasi-common/src/sync/mod.rs
@@ -131,7 +131,7 @@ impl WasiCtxBuilder {
 
 pub fn random_ctx() -> Box<dyn RngCore + Send + Sync> {
     let mut rng = cap_rand::thread_rng(cap_rand::ambient_authority());
-    Box::new(cap_rand::rngs::StdRng::from_seed(rng.gen()))
+    Box::new(cap_rand::rngs::StdRng::from_seed(rng.r#gen()))
 }
 
 #[cfg(feature = "wasmtime")]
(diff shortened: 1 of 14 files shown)


crates/wasi-common/src/sync/mod.rs
```diff
@@ -131,7 +131,7 @@ impl WasiCtxBuilder {
 
 pub fn random_ctx() -> Box<dyn RngCore + Send + Sync> {
     let mut rng = cap_rand::thread_rng(cap_rand::ambient_authority());
-    Box::new(cap_rand::rngs::StdRng::from_seed(rng.gen()))
+    Box::new(cap_rand::rngs::StdRng::from_seed(rng.r#gen()))
```
alexcrichton, line 134: cc @sunfishcode on this in case you weren't already aware, but eventually `cap-rand` will probably want to rename (or at least provide an alternative to) its `gen` method to be easier to use in the 2024 edition.

pchickey (approved):

PR #9959, opened 2025-01-09
Winch: i8x16.shuffle for x64 with AVX

jeffcharles:
Part of #8093. Implements `i8x16.shuffle` on x64 with AVX extensions.


saulecabrera:
I can take this review as well.

winch/codegen/src/isa/x64/masm.rs
```diff
@@ -1258,6 +1258,36 @@ impl Masm for MacroAssembler {
...
+
+            self.asm.xmm_vpshufb_rrm(dst, lhs, &mask_lhs);
+            let scratch = writable!(regs::scratch_xmm());
+            self.asm.xmm_vpshufb_rrm(scratch, rhs, &mask_rhs);
+            self.asm.vpor(dst, dst.to_reg(), scratch.to_reg());
+        } else {
+            bail!(CodeGenError::UnimplementedForNoAvx)
+        }
+        Ok(())
+    }
```
saulecabrera, line 1290: One small suggestion perhaps to improve readability, given that the then branch is somewhat lengthy, could we invert the check so that we return early in case there's no avx support?

```rust
if !self.flags().has_avx() {
  bail!(...);
}

// ....
Ok(())
```

saulecabrera (approved):

PR #9968, opened 2025-01-09
Update `Config::wasm_simd` docs

saulecabrera:
To reflect that the base CPU feature requirement on x86_64 is SSE2


alexcrichton (approved):
Thanks!

Issue #9734, opened 2024-12-04
winch(x64): Threads


saulecabrera:
The Wasm Threads proposal is considered Tier 1 according to [Wasmtime's Tiers of support](https://docs.wasmtime.dev/stability-tiers.html#tier-1).

Winch currently doesn't support this proposal.

- [x] `i32.atomic.load8_u` @MarinPostma 
- [x] `i32.atomic.load16_u` @MarinPostma 
- [x] `i32.atomic.load` @MarinPostma 
- [x] `i64.atomic.load8_u` @MarinPostma 
- [x] `i64.atomic.load16_u` @MarinPostma 
- [x] `i64.atomic.load32_u` @MarinPostma 
- [x] `i64.atomic.load` @MarinPostma 
- [x] `i32.atomic.store8` @MarinPostma 
- [x] `i32.atomic.store16` @MarinPostma 
- [x] `i32.atomic.store` @MarinPostma 
- [x] `i64.atomic.store8` @MarinPostma 
- [x] `i64.atomic.store16` @MarinPostma 
- [x] `i64.atomic.store32` @MarinPostma 
- [x] `i64.atomic.store` @MarinPostma 
- [x] `i32.atomic.rmw8.add_u` @MarinPostma 
- [x] `i32.atomic.rmw16.add_u` @MarinPostma 
- [x] `i32.atomic.rmw.add` @MarinPostma 
- [x] `i64.atomic.rmw8.add_u` @MarinPostma 
- [x] `i64.atomic.rmw16.add_u` @MarinPostma 
- [x] `i64.atomic.rmw32.add_u` @MarinPostma 
- [x] `i64.atomic.rmw.add` @MarinPostma 
- [x] `i32.atomic.rmw8.sub_u` @MarinPostma 
- [x] `i32.atomic.rmw16.sub_u` @MarinPostma 
- [x] `i32.atomic.rmw.sub` @MarinPostma 
- [x] `i64.atomic.rmw8.sub_u` @MarinPostma 
- [x] `i64.atomic.rmw16.sub_u` @MarinPostma 
- [x] `i64.atomic.rmw32.sub_u` @MarinPostma 
- [x] `i64.atomic.rmw.sub` @MarinPostma 
- [x] `i32.atomic.rmw8.and_u` @MarinPostma 
- [x] `i32.atomic.rmw16.and_u` @MarinPostma 
- [x] `i32.atomic.rmw.and` @MarinPostma 
[...]

PR #9970, opened 2025-01-09
Winch atomic loads x64

MarinPostma:
this PR enable the thread feature for x64 in Winch, and implements atomic loads. This includes:
- `i32.atomic.load8_u`
- `i32.atomic.load16_u`
- `i32.atomic.load`
- `i64.atomic.load8_u`
- `i64.atomic.load16_u`
- `i64.atomic.load32_u`
- `i64.atomic.load`

It also enabled shared memory for winch.

reopened from https://github.com/bytecodealliance/wasmtime/pull/9954 because the branch was wrong

### DIFF
--- a/winch/codegen/src/codegen/mod.rs
+++ b/winch/codegen/src/codegen/mod.rs
@@ -847,8 +847,29 @@ where
         size: OperandSize,
         sextend: Option<ExtendKind>,
     ) -> Result<()> {
-        let addr = self.emit_compute_heap_address(&arg, size)?;
-        if let Some(addr) = addr {
+        self.emit_wasm_load_full(arg, ty, size, sextend, false)
+    }
+
+    /// Emit a WebAssembly atomic load.
+    pub fn emit_wasm_load_atomic(
+        &mut self,
+        arg: &MemArg,
+        ty: WasmValType,
+        size: OperandSize,
+        sextend: Option<ExtendKind>,
+    ) -> Result<()> {
+        self.emit_wasm_load_full(arg, ty, size, sextend, true)
+    }
+
+    pub fn emit_wasm_load_full(
+        &mut self,
+        arg: &MemArg,
+        ty: WasmValType,
+        size: OperandSize,
+        sextend: Option<ExtendKind>,
+        atomic: bool,
+    ) -> Result<()> {
+        if let Some(addr) = self.emit_compute_heap_address(&arg, size)? {
             let dst = match ty {
                 WasmValType::I32 | WasmValType::I64 => self.context.any_gpr(self.masm)?,
                 WasmValType::F32 | WasmValType::F64 => self.context.any_fpr(self.masm)?,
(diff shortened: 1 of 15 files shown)


winch/codegen/src/codegen/mod.rs
```diff
@@ -847,8 +847,29 @@ where
...
+        self.emit_wasm_load_full(arg, ty, size, sextend, true)
+    }
+
+    pub fn emit_wasm_load_full(
+        &mut self,
+        arg: &MemArg,
+        ty: WasmValType,
+        size: OperandSize,
+        sextend: Option<ExtendKind>,
+        atomic: bool,
```
cfallin, line 870: Rather than a `bool`, which can be error-prone, could you define a small enum `WasmLoadKind { Regular, Atomic }` or similar and match on it in this function?

cfallin (approved):

winch/codegen/src/isa/aarch64/masm.rs
```diff
@@ -870,6 +870,16 @@ impl Masm for MacroAssembler {
...
     }
+
+    fn wasm_load_atomic(
+        &mut self,
+        _src: Self::Address,
+        _dst: WritableReg,
+        _size: OperandSize,
+        _sextend: Option<ExtendKind>,
+    ) -> Result<()> {
+        todo!()
```
saulecabrera, line 881: Could you return a proper `CodeGenError` instead of panicking? I'm in the process of enabling spec tests for aarch64, in general recoverable errors are preferred as it makes  it easier to define when a particular proposal is not fully supported in a particular backend (like threads in `aarch64`)

I believe we have `CodeGenError::unimplemented_masm_instruction()`

winch/codegen/src/isa/x64/masm.rs
```diff
@@ -1211,6 +1211,18 @@ impl Masm for MacroAssembler {
...
+
+    fn wasm_load_atomic(
+        &mut self,
+        src: Self::Address,
+        dst: WritableReg,
+        size: OperandSize,
+        kind: Option<ExtendKind>,
+    ) {
+        // The guarantees of the x86-64 memory model ensure that `SeqCst`
+        // loads are equivalent to normal loads.
```
saulecabrera, line 1223: Given this comment and for the purposes of this pull request, I wonder if it's necessary to introduce a new method in the MacroAssembler. I'd prefer if possible, if we could instead extend the current load definition to take in a `load_kind` argument, which categorizes if the load is atomic or not, similar to Chris' comment in https://github.com/bytecodealliance/wasmtime/pull/9970/files#r1909577300, that would also simplify the indirection happening in the `CodeGen` module:

We could simplify to:

* Extending `emit_wasm_load` to take in a load kind
* Performing the right dispatch directly from that method to the MacroAssembler, by passing the in-scope load kind.

MarinPostma: sure i'll do that

MarinPostma:
comments addressed

MarinPostma:
I'll fix the test tommorow.

MarinPostma:
I have enabled shared memory in winch, and updated the spec tests

saulecabrera (approved):
Thanks!

PR #9971, opened 2025-01-10
Update to the Rust 2024 edition

alexcrichton:
This is a test to see what breaks in CI when updating to the 2024 edition. The edition won't actually be released until February of this year and we won't be able to use it until two releases after that. This is just a test in the meantime to smooth the future transition.


alexcrichton:
Learnings here:

* A few miscellaneous things were split out here and there (e.g. failing doctests)
* Looks like this is going to be a "reformat the world" transition, presumably due to newer defaults in rustfmt
* The windows-specific failures are due to https://github.com/rust-lang/rust/pull/131072, not the edition itself, and that's coming in 1.85. We'll just need to update some tests to expect success instead of expecting failure.

Otherwise I think we're bsaically good to go 👍

PR #9973, opened 2025-01-10
ISLE: fix more glitches in BNF

eagr:
Address some issues in BNF
* \<ident\> overlapped with \<const-ident\>
* `(spec (signature ...))` should be `(spec (signature) ...)`
* spec expression def. missing `result`

### DIFF
--- a/cranelift/isle/docs/language-reference.md
+++ b/cranelift/isle/docs/language-reference.md
@@ -1533,7 +1533,7 @@ The grammar accepted by the parser is as follows:
        | "(" "form" <form> ")"
        | "(" "instantiate" <instantiation> ")"
 
-<spec> ::= "(" <ident> <ident>* <provide> [ <require> ] ")"
+<spec> ::= "(" <ident> <ident>* ")" <provide> [ <require> ]
 <provide> ::= "(" "provide" <spec-expr>* ")"
 <require> ::= "(" "require" <spec-expr>* ")"
 
--- a/cranelift/isle/docs/language-reference.md
+++ b/cranelift/isle/docs/language-reference.md
@@ -1554,6 +1554,7 @@ The grammar accepted by the parser is as follows:
 
 <spec-expr> ::= <int>
               | <spec-bv>
+              | "result"
               | "true" | "false"
               | <ident>
               | "(" "switch" <spec-expr> <spec-pair>* ")"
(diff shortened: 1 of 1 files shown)


cranelift/isle/docs/language-reference.md
```diff
@@ -1554,6 +1554,7 @@ The grammar accepted by the parser is as follows:
 
 <spec-expr> ::= <int>
               | <spec-bv>
+              | "result"
```
cfallin, line 1557: This is considered an identifier and is not a separate keyword.

cfallin (commented):
Thanks -- two of the three grammar fixes here look good. Last one is a bit of a convention question (are identifiers with special meanings distinguished at the syntax level?) but I think I'd prefer to go the other way to minimize the grammar.

cranelift/isle/docs/language-reference.md
eagr: If it is an integral part of the verification and if produced value only bound to ident with the exact spelling "result", I think it deserves a place in the spec. Your call :))

cfallin: It's currently special, but the intent has never really been to build up a set of keywords -- in the future we may find a better way. I don't see it quite the same as a privileged name like `this` in C++ or `self` in Rust, for example.

I think you may be placing too much weight on the BNF here, too: the purpose of BNF is usually not to give a parse tree with the full semantics of a language embedded at the parse level, but rather to describe the accepted language syntax accurately and give names to concepts that can then be given meaning by higher levels of the spec. That's more or less what is going on here.

eagr: I see. I certainly did and shouldn't. Thanks for pointing that out!

cfallin (approved):
Thanks!

cfallin:
@eagr I see this is still marked as a draft; if you're ready for me to merge it, I'm happy to do so, you'll just need to mark it as ready.

eagr:
Sure! I'm writing a tree-sitter parser for ISLE and picking up inconsistencies from the BNF. Thought it might be nice to correct those along the way. I'll flag it as ready when I'm done.

eagr:
Nitpick

* \<int\> isn't accurate enough, e.g., "0x_" is invalid
* "switch" probably shouldn't be \<spec_op\> if \<spec_pair\> isn't a \<spec_expr\>

Do you want to address these too?

cfallin (approved):
Let's get this merged as-is and then any spec-related grammar fixes I think I might delegate to @avanhatt or @mmcloughlin to review -- their project is really authoritative on how that sublanguage is designed. Thanks for these fixes!

PR #9974, opened 2025-01-10
chore: update `hashbrown` to 0.15

JonasKruckenberg:
This change updates `hashbrown` from 0.14.x to 0.15.x replacing the usage of the removed  unstable `hashbrown::raw` module in `cranelift-codegen/src/ctxhash.rs` with `hashbrown::hashtable` which should not make a difference (afaik).


alexcrichton (approved):
Thanks for this!

PR #9975, opened 2025-01-10
refactor: `no_std` `rustc_hash::{FxHashMap, FxHashSet}` aliases

JonasKruckenberg:
As part of `no_std` support in cranelift this PR fixes the usage of `rustc_hash::{FxHashMap, FxHashSet}` which are not exported on `no_std` targets. It adds type aliases to `lib.rs` similar to the already existing `std::collections::HashMap` alias.


cranelift/codegen/src/scoped_hash_map.rs
```diff
@@ -4,12 +4,12 @@
...
 //! values inserted while inside a scope aren't visible outside the scope.
 
+use crate::FxHashMap;
 use core::hash::Hash;
-use rustc_hash::FxHashMap;
 use smallvec::{smallvec, SmallVec};
 
 #[cfg(not(feature = "std"))]
-use crate::fx::FxHasher;
+use rustc_hash::FxHasher;
```
cfallin, line 12: It doesn't look like `FxHasher` or `Hasher` are used anywhere in this module, and they aren't exported -- can we remove them instead?

cfallin (approved):
Thanks! LGTM modulo some potential dead code below.

alexcrichton:
I believe cranelift-codegen nowadays compiles for no_std, so I'm going to close this.

PR #9976, opened 2025-01-10
fix: `no_std` timing module

JonasKruckenberg:
This change moves code around in the `cranelift` `timing` module so that it compiles on `no_std` targets where `thread_local` isn't available.

### DIFF
--- a/cranelift/codegen/src/timing.rs
+++ b/cranelift/codegen/src/timing.rs
@@ -2,12 +2,9 @@
 //!
 //! This modules provides facilities for timing the execution of individual compilation passes.
 
+use alloc::boxed::Box;
+use core::any::Any;
 use core::fmt;
-use std::any::Any;
-use std::boxed::Box;
-use std::cell::RefCell;
-use std::mem;
-use std::time::Duration;
 
 // Each pass that can be timed is predefined with the `define_passes!` macro. Each pass has a
 // snake_case name and a plain text description used when printing out the timing report.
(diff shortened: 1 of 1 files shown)


cranelift/codegen/src/timing.rs
```diff
@@ -273,14 +278,18 @@ mod enabled {
...
 mod disabled {
     use super::{DefaultProfiler, Pass, Profiler};
-    use std::any::Any;
-    use std::boxed::Box;
+    use alloc::boxed::Box;
+    use core::any::Any;
 
     impl Profiler for DefaultProfiler {
         fn start_pass(&self, _pass: Pass) -> Box<dyn Any> {
             Box::new(())
```
bjorn3, line 286: This should probably return a private type to ensure it can't be downcasted.

cranelift/codegen/src/timing.rs
```diff
@@ -273,14 +278,18 @@ mod enabled {
...
         }
     }
+
+    pub fn start_pass(_pass: Pass) -> Box<dyn Any> {
+        Box::new(())
+    }
```
bjorn3, line 292: 1. where is this function used?
2. It should probably be `pub(crate)` instead.

cranelift/codegen/src/timing.rs
JonasKruckenberg: okay maybe this is stupid of me, but why would that be an issue?

cranelift/codegen/src/timing.rs
JonasKruckenberg: yup, good catch 👍🏻

cranelift/codegen/src/timing.rs
bjorn3: It is not that much of an issue, but if in the future we were to add an actually working profiler (eg by using atomics and aggregating statistics for all threads), it would technically be a breaking change to change the type stored in the `Box<dyn Any>` as now `.downcast::<()>()` for consumers would fail. I don't think anyone will write code like that though.

fitzgen (approved):

PR #9977, opened 2025-01-10
fix: `no_std` fix use without `unwind`

JonasKruckenberg:
As part of no_std support in cranelift this PR fixes uses of the `isa/unwind` module through `cranelift` so that `cranelift` can be compiled with the `unwind` feature (which depends on `stdlib` through `gimli`) disabled.

### DIFF
--- a/cranelift/codegen/src/machinst/vcode.rs
+++ b/cranelift/codegen/src/machinst/vcode.rs
@@ -1064,8 +1064,14 @@ impl<I: VCodeInst> VCode<I> {
         }
 
         self.monotonize_inst_offsets(&mut inst_offsets[..], func_body_len);
+
+        #[cfg(feature = "unwind")]
         let value_labels_ranges =
             self.compute_value_labels_ranges(regalloc, &inst_offsets[..], func_body_len);
+        // when unwind information is disabled, ValueLabelsRanges are meaningless anyway
+        #[cfg(not(feature = "unwind"))]
+        let value_labels_ranges = ValueLabelsRanges::default();
+
         let frame_size = self.abi.frame_size();
 
         EmitResult {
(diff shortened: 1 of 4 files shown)


cranelift/codegen/src/machinst/vcode.rs
```diff
@@ -1158,8 +1158,13 @@ impl<I: VCodeInst> VCode<I> {
                 let slot = alloc.as_stack().unwrap();
                 let slot_offset = self.abi.get_spillslot_offset(slot);
                 let slot_base_to_caller_sp_offset = self.abi.slot_base_to_caller_sp_offset();
+
+                #[cfg(feature = "unwind")]
                 let caller_sp_to_cfa_offset =
                     crate::isa::unwind::systemv::caller_sp_to_cfa_offset();
+                #[cfg(not(feature = "unwind"))]
+                let caller_sp_to_cfa_offset = 0; // TODO is this right?
```
bjorn3, line 1166: Maybe `compute_value_labels_ranges` should be made to `return ValueLabelsRanges::default();` when the unwind feature is disabled? Except for values which just so happen to be in registers, I don't think it is possible to generate valid debuginfo for locals when the unwind table is not also emitted.

SingleAccretion: Arguably, the register values are just as 'meaningless' in the no-unwind case as you need the unwind info to recover them (unless you are willing to disassemble code to guide your unwinder). Ultimately, the debug info production 'feature' is tightly coupled to the unwinding feature.

JonasKruckenberg: I marked `compute_value_labels_ranges` as available with the `unwind` feature too and made the `Default` changes at its only callsite, I think that's clearer than having the function exist but return a meaningless default value.

alexcrichton:
I believe cranelift-codegen now compiles for no_std, so I'm going to close this.

PR #9983, opened 2025-01-10
Enable `pulley` by default for the `wasmtime` CLI

alexcrichton:
Now that pulley is nearing completion this makes it more useful to test out the interpreter on the CLI by having it available by default on all platforms.


dicej (approved):

pchickey (approved):
🎉

PR #9984, opened 2025-01-10
WasiCtx paramaterized by WasiExecutor

pchickey:


pchickey:
I decided against this design direction

PR #9986, opened 2025-01-11
pulley: Add special instructions for `dst = 0` and 1

alexcrichton:
This commit adds a special instruction for setting a register to the value 0 or the value 1. This extends to the full width of the register and accounts for the majority of all `xconst` instructions found in `spidermonkey.cwasm`. It's not a major size decrease, but helps a bit.

### DIFF
--- a/cranelift/codegen/src/isa/pulley_shared/inst.isle
+++ b/cranelift/codegen/src/isa/pulley_shared/inst.isle
@@ -368,6 +368,9 @@
 ;; Lower a constant into a register.
 (decl imm (Type u64) Reg)
 
+(rule 4 (imm (ty_int _) 0) (pulley_xzero))
+(rule 5 (imm (ty_int _) 1) (pulley_xone))
+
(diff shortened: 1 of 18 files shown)


cfallin (approved):
Seems reasonable. Does adding `-1` too help any more with code size? I imagine it could appear with and-masks and such relatively often...

alexcrichton:
Looking at `spidermonkey.cwasm` before and after this PR:

* Before: 119494 instances of `xconst*` instructions
* After: 58502 instances of `xconst*` instructions

The most common instances of constants were:

* Before: 41014 `0`, 19979 `1`
* After: 3091 `-4294967296`, 3041 `32`

My guess is that `-1` is folded into instructions commonly and doesn't show up as a dedicated `xconst`. The main use for `xconst` right now I think is to store a constant to memory or pass it as an ABI argument. I should probably look into adding macro-instructions for the former case...

In any case though at least for spidermonkey adding `-1` I don't think will help too too much.

PR #9987, opened 2025-01-12
winch: x64 atomic stores

MarinPostma:
This PR implements x64 store operations:
- `i32.atomic.store8`
- `i32.atomic.store16`
- `i32.atomic.store`
- `i64.atomic.store8`
- `i64.atomic.store16`
- `i64.atomic.store32`
- `i64.atomic.store`

### DIFF
--- a/winch/codegen/src/codegen/mod.rs
+++ b/winch/codegen/src/codegen/mod.rs
@@ -648,12 +648,7 @@ where
         &mut self,
         memarg: &MemArg,
         access_size: OperandSize,
-        check_align: bool,
     ) -> Result<Option<Reg>> {
-        if check_align {
-            self.check_align(memarg, access_size)?;
-        }
-
         let ptr_size: OperandSize = self.env.ptr_type().try_into()?;
         let enable_spectre_mitigation = self.env.heap_access_spectre_mitigation();
         let add_offset_and_access_size = |offset: ImmOffset, access_size: OperandSize| {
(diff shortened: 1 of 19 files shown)


winch/codegen/src/isa/x64/masm.rs
```diff
@@ -215,8 +217,27 @@ impl Masm for MacroAssembler {
...
+        size: OperandSize,
+        op_kind: MemOpKind,
+    ) -> Result<()> {
+        match op_kind {
+            MemOpKind::Atomic => {
+                if size == OperandSize::S128 {
+                    // TODO: we don't support 128-bit atomic store yet.
+                    bail!(CodeGenError::unexpected_operand_size());
+                }
+                // To stay consistent with cranelift, we emit a normal load followed by a mfence,
```
saulecabrera, line 233:
```suggestion
                // To stay consistent with cranelift, we emit a normal store followed by a mfence,
```

saulecabrera (commented):
One thing that we might want to do as part of this change before proceeding further with the development of the rest of the instructions in the proposal,  is to ensure that addresses are aligned for atomic loads/stores. See the [alignment](https://github.com/WebAssembly/threads/blob/main/proposals/threads/Overview.md#alignment) of the proposal. 

To check for alignment, I'd recommend creating a new method in the `CodeGen` module (e.g., `emit_check_alignment`) and either calling it before calling `emit_compute_heap_addr` or enhancing `emit_compute_heap_addr` to check alignment internally if the load/store is atomic. 

For reference, [this is how alignment is handled in Cranelift.](https://github.com/bytecodealliance/wasmtime/blob/main/crates/cranelift/src/translate/code_translator.rs#L3306)

MarinPostma:
@saulecabrera I have added the align check

winch/codegen/src/codegen/mod.rs
```diff
@@ -840,6 +840,45 @@ where
         Ok(addr)
     }
 
+    /// Emit checks to ensure that the address at `memarg` is correctly aligned for `size`.
+    fn emit_check_align(&mut self, memarg: &MemArg, size: OperandSize) -> Result<()> {
+        if size.bytes() > 1 {
+            let addr = self.context.pop_to_reg(self.masm, None)?;
+            let tmp = scratch!(M);
+            // self.context.move_val_to_reg(&addr, tmp, self.masm)?;
```
saulecabrera, line 848:
```suggestion

```

MarinPostma:
@saulecabrera it's not working yet I had to move to a x86 machine to debug, I'll ping you when i fixed it

MarinPostma:
@saulecabrera, I have made the changes you requested, I also responded to your comments, let me know if you want me to make the changes I suggested in the comments instead.

winch/codegen/src/codegen/mod.rs
```diff
@@ -840,6 +845,36 @@ where
         Ok(addr)
     }
 
+    /// Emit checks to ensure that the address at `memarg` is correctly aligned for `size`.
+    fn check_align(&mut self, memarg: &MemArg, size: OperandSize) -> Result<()> {
+        if size.bytes() > 1 {
+            let addr = *self.context.stack.peek().unwrap();
+            let tmp = self.context.any_gpr(self.masm)?;
+            self.context.move_val_to_reg(&addr, tmp, self.masm)?;
```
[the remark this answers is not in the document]
saulecabrera: Oh that's a good point yeah. My main concern with `peek` is that nothing is preventing any other method to accidentally pop the wrong value from the stack. However, since this pattern is local to this method, I don't think there's a huge risk, so yeah, let's try the peek approach. One comment to your original implementation, could you return an error instead of doing `peek().unwrap()`? We have `CodeGenError::missing_values_in_stack` for this kind of situation.

MarinPostma:
@saulecabrera I made the final changes

PR #9989, opened 2025-01-12
feat: Add hexadecimal notation support for integer arguments

VolodymyrBg:
Add support for parsing hexadecimal integers (prefixed with 0x or 0X) when passing arguments to WebAssembly functions. This enhancement allows users to specify integer arguments in both decimal and hexadecimal formats.

For example:
- Decimal: 42
- Hexadecimal: 0x2A or 0x2a

The change affects both i32 and i64 argument types while maintaining backward compatibility with decimal notation.


alexcrichton:
Thanks! Mind adding a test for this as well? (probably somewhere in `tests/all/cli_tests.rs`)

VolodymyrBg:
> Thanks! Mind adding a test for this as well? (probably somewhere in `tests/all/cli_tests.rs`)

Thank you for the answer. I did what you asked

alexcrichton:
(I think there's a test failure though?)

fitzgen (approved):
Thanks!

fitzgen:
Looks like there are still some failing CI jobs that need to be fixed before this can merge.

VolodymyrBg:
> Looks like there are still some failing CI jobs that need to be fixed before this can merge.

Working on it))

PR #9992, opened 2025-01-13
refactor: remove `OnceLock`

JonasKruckenberg:
This PR is another attempt at #8489 (superseding it) this time by making the necessary changes in `regalloc2` (see [here](https://github.com/bytecodealliance/regalloc2/pull/208)) to allow `MachineEnv` to be created in const expressions and therefore placed into statics without the need for lazy initialization (as mentioned in [this comment](https://bytecodealliance.zulipchat.com/#narrow/channel/217117-cranelift/topic/Use.20in.20no_std/near/493078060))
This PR also makes the necessary changes in register creation functions to make them `const`.

While I agree with @cfallin that this is the "correct" move, it is slightly annoying that Rusts const limitation makes the static initializers so verbose where a `1..=16` range iterator was used previously, any ideas here are welcome :/

Edit: marked the PR as draft until the regalloc2 upstream PR is merged

### DIFF
--- a/cranelift/codegen/src/isa/aarch64/abi.rs
+++ b/cranelift/codegen/src/isa/aarch64/abi.rs
@@ -16,8 +16,6 @@ use alloc::boxed::Box;
 use alloc::vec::Vec;
 use regalloc2::{MachineEnv, PReg, PRegSet};
 use smallvec::{smallvec, SmallVec};
-use std::borrow::ToOwned;
-use std::sync::OnceLock;
 
 // We use a generic implementation that factors out AArch64 and x64 ABI commonalities, because
 // these ABIs are very similar.
@@ -1115,11 +1113,9 @@ impl ABIMachineSpec for AArch64MachineDeps {
 
     fn get_machine_env(flags: &settings::Flags, _call_conv: isa::CallConv) -> &MachineEnv {
         if flags.enable_pinned_reg() {
-            static MACHINE_ENV: OnceLock<MachineEnv> = OnceLock::new();
-            MACHINE_ENV.get_or_init(|| create_reg_env(true))
+            &PINNED_MACHINE_ENV
         } else {
-            static MACHINE_ENV: OnceLock<MachineEnv> = OnceLock::new();
-            MACHINE_ENV.get_or_init(|| create_reg_env(false))
+            &DEFAULT_MACHINE_ENV
         }
     }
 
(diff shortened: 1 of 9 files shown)


cranelift/codegen/src/isa/aarch64/abi.rs
```diff
@@ -1525,16 +1520,99 @@ fn create_reg_env(enable_pinned_reg: bool) -> MachineEnv {
...
-    if !enable_pinned_reg {
-        debug_assert_eq!(PINNED_REG, 21); // We assumed this above in hardcoded reg list.
-        env.non_preferred_regs_by_class[0].push(preg(xreg(PINNED_REG)));
     }
+};
 
-    env
-}
+static DEFAULT_MACHINE_ENV: MachineEnv = {
+    debug_assert!(PINNED_REG == 21); // We assumed this below in hardcoded reg list.
```
cfallin, line 1531: This probably goes in the `PINNED_MACHINE_ENV` above, no? (It's load-bearing there that we get it right, because we *excluded* the pinned register from allocation; the default environment is used when pinning is disabled so doesn't care what the pinned reg is)

cranelift/codegen/src/machinst/reg.rs
```diff
@@ -48,10 +48,26 @@ pub fn first_user_vreg_index() -> usize {
 pub struct Reg(VReg);
 
 impl Reg {
+    /// Create a register from a physical register
+    // FIXME(const-hack) remove in favor of `From` impl
```
cfallin, line 52: Can you say more here -- I'm not that familiar with the current state of constification of everything; is there some feature this is waiting on to stabilize or...?

cfallin (commented):
Thanks a bunch for working on this! This all looks like the shape I expected; I'm glad there were no other hidden snags. A few minor comments below but otherwise happy to see this merged once we get the RA2 PR merged and released.

cranelift/codegen/src/machinst/reg.rs
JonasKruckenberg: well, the feature that would be waiting for is const associated functions in traits (therefore allowing `From::from` to be const) but that afaik is in "never type land". I can remove the note as it doesn't make much sense to hold our breath on it

cfallin: Yeah, I think that's best -- usually when we have a FIXME we want it to be actionable, or at least give sufficient context so that we know what we're waiting on (i.e.: if someone else walked up to this code tomorrow, would they know what they need in order to fix it?). It's fine to evaluate questions like this once new Rust features land in the future.

JonasKruckenberg: the issue with constness is that theres no good way to detect constness workarounds after the fact (the whole reason for the const-hack comments) but usually you'd tie them to specific RFCs yeah

hanna-kruppe: Relevant RFC fresh off the press: https://github.com/rust-lang/rfcs/pull/3762

alexcrichton:
I believe this has since been resolved, so closing.

PR #9993, opened 2025-01-13
winch(x64): Fix `MemOpKind` for `i64_atomic_load8`

saulecabrera:
Change from `MemOpKind::Normal` to `MemOpKind::Atomic`. In the case of x64 this change doesn't meaningfully change the output of the instruction. However, for correctness and for the purposes of other backends, we need to pass the right `MemOpKind`.


alexcrichton (approved):

saulecabrera:
The failure was due to a missing rebase.

PR #9994, opened 2025-01-13
pulley: Add more addressing modes for loads/stores

alexcrichton:
This commit adds a new "g32" addressing mode to Pulley that matches the pattern emitted by Cranelift for 32-bit wasm guests running on hosts. The general idea here is that this addressing mode encompasses an addition of a host-width value to a zero-extended (optionally) 32-bit value. On 32-bit hosts there's no zero-extension but on 64-bit hosts there's a zero-extension. The wasm address is always 32-bits though which enables using a single instruction for both 32 and 64-bit hosts.

New "g32" loads and stores are added to Pulley with varying sizes and options according to what seems to be common in wasm. The `disas` test suite was updated to showcase using these instructions for wasm loads/stores on 32 and 64-bit hosts.

An additional change in this commit is to deduplicate the 32/64-bit bounds-check macro-ops. The trick in this commit works for those as well meaning that only a single instruction is needed instead of one-per-host-pointer-width. Additionally the load of the bound from the `VMContext` is folded into the bounds check itself as it was found that this was always present anyway before the bounds check.

Overall this shrinks the size of `spidermonkey.cwasm` from 21M to 20M and the runtime of `pulldown-cmark`, `bz2`, and `spidermonkey` on Sightglass have all been reduced by 10%. Not as big wins as I was hoping for but alas.

### DIFF
--- a/cranelift/codegen/src/isa/pulley_shared/inst.isle
+++ b/cranelift/codegen/src/isa/pulley_shared/inst.isle
@@ -172,6 +172,42 @@
 
 (type VExtKind (enum None S8x8 U8x8 S16x4 U16x4 S32x2 U32x2))
 
+;; Helper to determine the endianness of `MemFlags` taking the current target
+;; into account.
+(decl pure endianness (MemFlags) Endianness)
+(extern constructor endianness endianness)
+(type Endianness extern (enum Little Big))
+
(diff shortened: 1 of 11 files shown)


fitzgen (approved):
Nice!

PR #9999, opened 2025-01-13
pulley: Shrink frame save/restore instructions

alexcrichton:
This commit shrinks the size of the `PushFrameSave` and `PopFrameRestore` functions which are used in almost all wasm functions. Previously these instructions allowed for 32-bits of stack space in addition to saving/restoring all 32 X-registers. In reality though it's quite uncommon to need more than 16-bits of stack space and ABI-wise the most commonly saved registers are the upper 16 registers of the X register set.

This commit therefore shrinks the frame size to 16 bits and only has the ability to save/restore the upper 16 X-registers. Note that any clobbered registers and frame sizes are still supported, they'll just use more pessimal encodings which aren't a single opcode. If a function uses >64KiB of stack space though it's probably not too important what the dispatch cost is at the beginning.

The overall result of this change is that each instruction shaves of 4 bytes (2 from the frame size and 2 from the registers being saved/restored). This results in a 4% faster execution time on the bz2 Sightglass benchmark, ~1% on pulldown-cmark, and while it shrinks `spidermonkey.cwasm` slightly it's not significant.

While here this commit also removes the Cranelift-unused instructions of `x{push,pop}{,_many}` since the `RegSet` type was repurposed as an `UpperRegSet` instead. It's possible to bring back these instructions but they're otherwise not used right now.

### DIFF
--- a/cranelift/codegen/meta/src/pulley.rs
+++ b/cranelift/codegen/meta/src/pulley.rs
@@ -69,9 +69,9 @@ impl Inst<'_> {
                     Operand::Binop { dst, src1, src2 }
                 }
                 (name, ty) if name.starts_with("dst") => Operand::Writable { name, ty },
-                (name, "RegSet < XReg >") => Operand::Normal {
+                (name, "UpperRegSet < XReg >") => Operand::Normal {
                     name,
-                    ty: "XRegSet",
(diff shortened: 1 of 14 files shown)


fitzgen (approved):
Nice!

PR #10004, opened 2025-01-14
Cranelift: dedupe `trap[n]z` instructions

fitzgen:
This commit extends our existing support for merging idempotently side-effectful instructions that produce exactly one value to those that produce zero or one value, and marks the `trap[n]z` instructions as having idempotent side effects. This cleans up a lot test cases in our `disas` test suite, particularly those related to explicit bounds checks and GC.

As an aside, it seems like it should be easy to extend this to idempotently side-effectful instructions that produce multiple values as well, but I don't believe we have any such instructions, so I didn't bother.

### DIFF
--- a/cranelift/codegen/meta/src/shared/instructions.rs
+++ b/cranelift/codegen/meta/src/shared/instructions.rs
@@ -142,6 +142,11 @@ fn define_control_flow(
             Operand::new("code", &imm.trapcode),
         ])
         .can_trap()
+        // When one `trapz` dominates another `trapz` and they have identical
+        // conditions and trap codes, it is safe to deduplicate them (like GVN,
+        // although there is not actually any value being numbered). Either the
+        // first `trapz` raised a trap and execution halted, or it didn't and
+        // therefore the dominated `trapz` will not raise a trap either.
         .side_effects_idempotent(),
     );
 
@@ -160,6 +165,7 @@ fn define_control_flow(
             Operand::new("code", &imm.trapcode),
         ])
         .can_trap()
+        // See the above comment for `trapz` and idempotent side effects.
         .side_effects_idempotent(),
     );
 
--- a/cranelift/codegen/meta/src/shared/instructions.rs
+++ b/cranelift/codegen/meta/src/shared/instructions.rs
@@ -141,7 +141,8 @@ fn define_control_flow(
             Operand::new("c", ScalarTruthy).with_doc("Controlling value to test"),
             Operand::new("code", &imm.trapcode),
         ])
-        .can_trap(),
+        .can_trap()
+        .side_effects_idempotent(),
     );
 
     ig.push(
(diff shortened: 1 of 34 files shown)


cranelift/codegen/meta/src/shared/instructions.rs
```diff
@@ -141,7 +141,8 @@ fn define_control_flow(
             Operand::new("c", ScalarTruthy).with_doc("Controlling value to test"),
             Operand::new("code", &imm.trapcode),
         ])
-        .can_trap(),
+        .can_trap()
+        .side_effects_idempotent(),
```
cfallin, line 145: Can we add a comment here describing the rationale a bit more (for our future selves as much as others)?

Something like: "`side_effects_idempotent` indicates that when we have more than one copy of this instruction, it is safe to merge (GVN) them. Traps have this characteristic: as long as we don't remove them, if we have already passed through (computed) one trap instruction on a given condition, it is safe to consider other traps on the same condition as redundant. In other words, a trap on value X is idempotent: once we observe its side effect (which will either terminate execution or do nothing), we don't need to observe it again."

cranelift/codegen/src/egraph.rs
```diff
@@ -436,14 +436,25 @@ where
...
-                    trace!(" -> merges result {} to {}", result, orig_result);
+                    match (result, orig_result) {
+                        (Some(result), Some(orig_result)) => {
+                            // Hit in GVN map -- reuse value.
+                            self.value_to_opt_value[result] = orig_result;
+                            trace!(" -> merges result {} to {}", result, orig_result);
+                        }
+                        (None, None) => {
+                            trace!(" -> merges with dominating instruction");
+                            // Nothing else to do here.
```
cfallin, line 447: "Hit in GVN map but for instruction with zero results: may be a side-effect-only idempotent instruction like `trapnz`/`trapz`. Nothing else to do here." perhaps?

cranelift/filetests/filetests/egraph/idempotent-traps.clif
```diff
@@ -0,0 +1,92 @@
...
+; nextln:     br_table v1, block1, [block2, block3]
+; check:  block1:
+; nextln:     return v1
+; check:  block2:
+; nextln:     trapz.i32 v0, user1
+; nextln:     return v2
+; check:  block3:
+; nextln:     trapz.i32 v0, user1
+; nextln:     return v3
+; nextln: }
```
cfallin, line 92: Can we also add a test with a `trapz` and `trapnz` on the same condition, to verify that (if we ever change the instruction encoding) we're disambiguating on the condition as well and not merging them? (i.e., arrange such that they would merge if they were both the same opcode)

cfallin (approved):
This is great -- thanks a bunch for working this out!

My only thoughts are for a few more comments and one more test, otherwise LGTM.

PR #10011, opened 2025-01-14
Emit post-optimization CLIF with `--emit-clif`

alexcrichton:
This commit updates the implementation of the `wasmtime` CLI flag `--emit-clif` to emit the IR post-optimization rather than pre-optimization. This is now emitting the IR that's directly fed to the backend for lowering which reflects any mid-level transformation that's performed. It should still be possible to recover pre-optimized CLIF with `-O opt-level=0`.

This then additionally refactors functions to emit CLIF in a more "core" location so all trampolines get their CLIF emitted as well in case those are needed for debugging.

### DIFF
--- a/crates/cranelift/src/compiler.rs
+++ b/crates/cranelift/src/compiler.rs
@@ -273,18 +273,10 @@ impl wasmtime_environ::Compiler for Compiler {
             &mut func_env,
         )?;
 
-        if let Some(path) = &self.clif_dir {
-            use std::io::Write;
-
-            let mut path = path.to_path_buf();
-            path.push(format!("wasm_func_{}", func_index.as_u32()));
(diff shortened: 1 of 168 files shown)


dicej (approved):

alexcrichton:
Looks like all `disas` tests are changing notably for a few minor differences:

* Running through the compilation pipeline removes all `global_value` instructions (replaces them with their equivalents)
* Unnecessary block params are removed

bjorn3:
> It should still be possible to recover pre-optimized CLIF with -O opt-level=0.

Doesn't that still perform legalizations?

alexcrichton:
That's true yeah, `--emit-clif` won't work for that any more

PR #9990, opened 2025-01-13
Winch: implement rmw add ops

MarinPostma:
impl read-modify-write add for winch:
- `i32.atomic.rmw8.add_u`
- `i32.atomic.rmw16.add_u`
- `i32.atomic.rmw.add`
- `i64.atomic.rmw8.add_u`
- `i64.atomic.rmw16.add_u`
- `i64.atomic.rmw32.add_u`
- `i64.atomic.rmw.add`

https://github.com/bytecodealliance/wasmtime/issues/9734

PR #10012, opened 2025-01-14
winch: Simplify the handling of `wasm_load`s

saulecabrera:
This commit aims to simplify how `wasm_load`s are currently handled in the compiler by reducing the number of invariants that need to be encoded at callsites which make their emission error prone.

Concretely this change:

* Augments `ExtendKind`, to account for unsigned extends; prior to this commit it wasn't explicit which loads required unsigned extends.
* Derive `OperandSize` from `ExtendKind`, removing from the caller the responsibility of encoding the operand size to use.

### DIFF
--- a/winch/codegen/src/codegen/mod.rs
+++ b/winch/codegen/src/codegen/mod.rs
@@ -844,23 +844,23 @@ where
     pub fn emit_wasm_load(
         &mut self,
         arg: &MemArg,
-        ty: WasmValType,
-        size: OperandSize,
+        target_type: WasmValType,
         kind: LoadKind,
         op_kind: MemOpKind,
     ) -> Result<()> {
-        if let Some(addr) = self.emit_compute_heap_address(&arg, size)? {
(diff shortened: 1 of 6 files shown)


alexcrichton (approved):

MarinPostma:
In the context of atomic operations, which only accept zero-extends, it seems to me that encoding signed and zero extends in a single enum is maybe a bit too loose. Maybe we could split it in two enums instead:
```rust
enum ExtendKind {
    Signed(ExtendSize),
    Zero(ExtendSize),
}

enum ExtendSize {
   I32extends8,
   //...
}
```

WIth that, we could avoid this check, for example: https://github.com/bytecodealliance/wasmtime/pull/9990/commits/53d2d25a91926751d650467700a3b536d296c26e#diff-a9f4061f11240b9ea041bd84fbaf01290ff679313f87e5085eeebdecf300f5bdR1374-R1377

PR #10017, opened 2025-01-14
pulley: Support smaller int-to-float conversions

alexcrichton:
This adds support for converting 8/16-bit integers to floats to Pulley. This is not directly accessible from wasm instructions but is possible through various optimizations to create. A new `*.clif` runtest was added exercising many combinations of scalars-to-scalars for int-to-floats of varying widths.


cfallin (approved):
LGTM, thanks!

PR #10022, opened 2025-01-15
pulley: Refactor register restores on tail-calls

alexcrichton:
Use helpers from `abi.rs` where possible to avoid duplicating logic.


cfallin (approved):
LGTM, thanks!

PR #10023, opened 2025-01-15
Winch: implement rmw `and`, `xor` and `or` for x64

MarinPostma:
Implement related atomic rmw for winch x64 backend:
- `and`:
    - `i32.atomic.rmw8.and_u`
    - `i32.atomic.rmw16.and_u`
    - `i32.atomic.rmw.and`
    - `i64.atomic.rmw8.and_u`
    - `i64.atomic.rmw16.and_u`
    - `i64.atomic.rmw32.and_u`
    - `i64.atomic.rmw.and`
- `or`:
    - `i32.atomic.rmw8.or_u`
    - `i32.atomic.rmw16.or_u`
    - `i32.atomic.rmw.or`
    - `i64.atomic.rmw8.or_u`
    - `i64.atomic.rmw16.or_u`
    - `i64.atomic.rmw32.or_u`
    - `i64.atomic.rmw.or`
- `xor`:
    - `i32.atomic.rmw8.xor_u`
    - `i32.atomic.rmw16.xor_u`
    - `i32.atomic.rmw.xor`
    - `i64.atomic.rmw8.xor_u`
    - `i64.atomic.rmw16.xor_u`
    - `i64.atomic.rmw32.xor_u`
    - `i64.atomic.rmw.xor`


winch/codegen/src/isa/x64/masm.rs
```diff
@@ -1335,11 +1335,25 @@ impl Masm for MacroAssembler {
...
                     .lock_xadd(addr, operand.to_reg(), operand, size, flags);
             }
+            RmwOp::And | RmwOp::Or | RmwOp::Xor => {
+                let op = match op {
+                    RmwOp::And => AtomicRmwSeqOp::And,
+                    RmwOp::Or => AtomicRmwSeqOp::Or,
+                    RmwOp::Xor => AtomicRmwSeqOp::Xor,
+                    _ => unreachable!(
+                        "invalid op for atomic_rmw_seq, should be one of `or`, `and` or `xor`"
+                    ),
```
saulecabrera, line 1345: Could you return an error here instead of panicking?

cranelift/codegen/src/isa/x64/inst/mod.rs
```diff
@@ -30,6 +30,7 @@ use args::*;
 // Instructions (top level): definition
 
 // `Inst` is defined inside ISLE as `MInst`. We publicly re-export it here.
+pub use super::lower::isle::generated_code::AtomicRmwSeqOp;
```
saulecabrera, line 33: I don't see an issue with exposing this. However, given that I'm not deeply familiar with the history of x64 backend in Cranelift, I'll kindly ask @cfallin for confirmation.

saulecabrera (commented):
Winch pieces look good to me, modulo the panic comment below.

winch/codegen/src/isa/x64/masm.rs
MarinPostma: does it make sense to return an error here? we explicitely check that we only match the right variants above, this branch should be legitmately unreachable. No `CodeGenError` seem to correspond, so I was reluctant to add an error variant for a trivially unreachable statement.

saulecabrera: Yeah you're right, it's probably fine as is. I'm trying to error on the side of caution when it comes to panics, because we've put a lot of effort to make sure that they are minimal, particularly if they affect partial development of features at the visitor/masm layer. Here I got confused by the nested match which led me to think that we were missing an implementation.

saulecabrera (approved):

saulecabrera:
@MarinPostma just a heads up that there are a couple of conflicts that need to be resolved before merging.

MarinPostma:
@saulecabrera I squashed to make conflict resolution more manageable

saulecabrera (approved):

PR #10024, opened 2025-01-15
Disable sigaltstack overriding in asan builds

alexcrichton:
This commit is an attempt to fix a number of flaky crashes that we've been seeing on OSS-Fuzz for some time now. These crashes only reproduce under ASAN and even then have been spotty to reproduce. The current thinking is that a test with threads (e.g. only `wast_tests` using some of the threads spec tests) is required to run some wasm which will register a `sigaltstack`. Destruction of this `sigaltstack` happens with TLS destructors which seems to have a bad interaction with ASAN state additionally being destroyed around that time.

This whole interaction means that no one test case is enough to reproduce the corruption. Many crashes on OSS-Fuzz are likely due to "some historical test case spawned a thread" which corrupted something to crash later. The test case that I can reproduce with locally requires rerunning it in the same process a few thousand times to get a reproduction.

The purpose of the `sigaltstack` is to ensure that we have a big enough stack, primarily in debug mode, for testing if a trap is wasm. The hope is that this extra size of the Rust-standard-library-default's stack size is not necessary in release mode with ASAN. In the end time will tell with OSS-Fuzz to see if we can keep this or if we need to both install a bigger sigaltstack in addition to managing them differently in ASAN builds.

### DIFF
--- a/crates/wasmtime/src/runtime/vm/sys/unix/signals.rs
+++ b/crates/wasmtime/src/runtime/vm/sys/unix/signals.rs
@@ -393,8 +393,61 @@ unsafe fn set_pc(cx: *mut libc::c_void, pc: usize, arg1: usize) {
 /// always large enough for our signal handling code. Override it by creating
 /// and registering our own alternate stack that is large enough and has a guard
 /// page.
+///
+/// Note that one might reasonably ask why do this at all? Why not remove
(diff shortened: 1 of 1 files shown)


cfallin (approved):
LGTM -- Alex, Nick and I talked about this offline for a bit today. Thanks a bunch for root-causing this -- was very mysterious!

PR #10028, opened 2025-01-15
Winch: Add splat instructions to x64 using AVX2

jeffcharles:
Part of https://github.com/bytecodealliance/wasmtime/issues/8093. Implements the following instructions on x64 with AVX2 support:
- `i8x16.splat`
- `i16x8.splat`
- `i32x4.splat`
- `i64x2.splat`
- `f32x4.splat`
- `f64x2.splat`

### DIFF
--- a/winch/codegen/src/isa/x64/asm.rs
+++ b/winch/codegen/src/isa/x64/asm.rs
@@ -4,7 +4,7 @@ use crate::{
     isa::{reg::Reg, CallingConvention},
     masm::{
         DivKind, ExtendKind, IntCmpKind, MulWideKind, OperandSize, RemKind, RoundingMode,
-        ShiftKind, SplatKind, VectorExtendKind,
+        ShiftKind, VectorExtendKind,
     },
 };
 use cranelift_codegen::{
@@ -164,18 +164,6 @@ impl From<OperandSize> for Option<ExtMode> {
             S8 => Some(ExtMode::BQ),
             S16 => Some(ExtMode::WQ),
             S32 => Some(ExtMode::LQ),
-        }
-    }
-}
-
-impl SplatKind {
-    /// Get the operand size for `vpbroadcast`.
-    pub(super) fn vpbroadcast_operand_size(&self) -> OperandSize {
-        match self {
-            SplatKind::S8 => OperandSize::S8,
-            SplatKind::S16 => OperandSize::S16,
-            SplatKind::S32 => OperandSize::S32,
-            SplatKind::S64 => OperandSize::S64,
         }
     }
 }
--- a/winch/codegen/src/isa/x64/asm.rs
+++ b/winch/codegen/src/isa/x64/asm.rs
@@ -4,7 +4,7 @@ use crate::{
     isa::{reg::Reg, CallingConvention},
     masm::{
         DivKind, ExtendKind, IntCmpKind, MulWideKind, OperandSize, RemKind, RoundingMode,
-        ShiftKind, VectorExtendKind,
+        ShiftKind, SplatKind, VectorExtendKind,
     },
 };
 use cranelift_codegen::{
@@ -164,6 +164,18 @@ impl From<OperandSize> for Option<ExtMode> {
             S8 => Some(ExtMode::BQ),
             S16 => Some(ExtMode::WQ),
             S32 => Some(ExtMode::LQ),
+        }
+    }
+}
+
+impl SplatKind {
+    /// Get the operand size for `vpbroadcast`.
+    pub(super) fn vpbroadcast_operand_size(&self) -> OperandSize {
+        match self {
+            SplatKind::S8 => OperandSize::S8,
+            SplatKind::S16 => OperandSize::S16,
+            SplatKind::S32 => OperandSize::S32,
+            SplatKind::S64 => OperandSize::S64,
         }
     }
 }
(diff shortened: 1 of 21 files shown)


winch/codegen/src/isa/x64/asm.rs
```diff
@@ -168,6 +168,18 @@ impl From<OperandSize> for Option<ExtMode> {
     }
 }
 
+impl SplatKind {
+    /// Get the operand size for `vpbroadcast`.
+    pub(super) fn vpbroadcast_operand_size(&self) -> OperandSize {
```
saulecabrera, line 173: Instead of making this specific to x64, can we instead call this `lane_size` or something along those lines and move the implementation to `masm.rs`? I don't think we'd only need this for x64.

winch/codegen/src/isa/x64/masm.rs
```diff
@@ -1286,6 +1286,81 @@ impl Masm for MacroAssembler {
...
+    fn splat(&mut self, dst: WritableReg, src: RegImm, size: SplatKind) -> Result<()> {
+        if size == SplatKind::S64 {
+            if !self.flags.has_avx() {
+                bail!(CodeGenError::UnimplementedForNoAvx);
+            }
+            // Results in the first 4 bytes and second 4 bytes being
+            // swapped and then the swapped bytes being copied.
+            // [d0, d1, d2, d3, d4, d5, d6, d7, ...] yields
+            // [d4, d5, d6, d7, d0, d1, d2, d3, d4, d5, d6, d7, d0, d1, d2, d3].
+            let mask = 0b0100_0100;
```
saulecabrera, line 1331: If I'm not wrong, this is the exact same comment/mask value used in `load_splat`. Could we extract this as a helper?

winch/codegen/src/masm.rs
```diff
@@ -1257,6 +1269,14 @@ pub(crate) trait MacroAssembler {
...
         -> Result<()>;
 
+    /// Takes the int in a `src` operand and replicates it across lanes of
+    /// `size` in a destination result.
+    fn splat_int(&mut self, context: &mut CodeGenContext<Emission>, size: SplatKind) -> Result<()>;
+
+    /// Takes the `src` operand and replicates it across lanes of `size` in
+    /// `dst`.
+    fn splat(&mut self, dst: WritableReg, src: RegImm, size: SplatKind) -> Result<()>;
+
```
saulecabrera, line 1279: Taking a look at the callsites, and trying to think to future backends, I'm tempted to suggest adding a single entry here, `splat`, defined as 

```rust
fn splat(&mut self, context: &mut CodeGenContext, kind: SplatKind) {
  ...
}
```

That's potentially the most flexible approach, and that would potentially make the definition more uniform across backends: as far as I can tell, the only reason why we have a special `splat_int` is because of the requirements in x64. 

For that suggestion to work though, we'd probably need to extend `SplatKind` to encode the source type, (e.g., `SplatKind::I8x16`) -- I don't think this changes anything fundamental, since we'd still be able to derive the `lane_size`.

winch/codegen/src/masm.rs
```diff
@@ -478,6 +479,17 @@ impl Imm {
...
     }
+
+    /// Get a little endian representation of the immediate.
+    pub fn to_bytes(&self) -> Vec<u8> {
+        match self {
+            Imm::I32(n) => n.to_le_bytes().to_vec(),
+            Imm::I64(n) => n.to_le_bytes().to_vec(),
+            Imm::F32(n) => n.to_le_bytes().to_vec(),
+            Imm::F64(n) => n.to_le_bytes().to_vec(),
+            Imm::V128(n) => n.to_le_bytes().to_vec(),
```
saulecabrera, line 490: Could you return a `slice` here instead? The constant pool will copy the bytes into a `Vec<u8>`, but returning a slice will at least prevent heap allocations in case this gets used for anything else that's not constant pool related.

saulecabrera: To be more specific, using `as_slice()` instead of `to_vec()`, similar to the other uses of `add_constant`.

jeffcharles: If I use `as_slice()` and return a `&[u8]`, I get this message from the compiler: `cannot return value referencing temporary value`. Likely because the array that's created by `to_le_bytes()` gets dropped when the method returns. The arrays are also of different lengths between the variants. We could maybe use some unsafe code to create a reference to the underlying bytes. Otherwise I'm not sure how to avoid the heap allocation.

winch/codegen/src/masm.rs
jeffcharles: The one thing with changing `SplatKind` to encode an `I` or `F` is the `v128.load*_splat` instructions aren't necessarily operating on integers or floats. It's operating on whatever state was stored in the memory. That said, it should still be fine to have the visitor specify a `SplatKind::I32x4` for `v128.load32_splat` even if it's operating on 4 floats, it'll still have the same result.

saulecabrera: To be explicit, we could rename the current `SplatKind` to `SplatLoadKind` and introduce the types in `SplatKind`. That would tighten the implementation and keep the semantics as close as possible to the Wasm semantics.

winch/codegen/src/masm.rs
saulecabrera: Given that this is not used for anything that the constant pool, I don't think we need to introduce any unsafe code. Perhaps adding a comment stating that this method heap allocates and it's intended to be used mostly with the constant pool?

saulecabrera (approved):

PR #10034, opened 2025-01-16
Add basic support for profiling Pulley

alexcrichton:
This commit adds basic support for profiling the Pulley interpreter. This is partially achievable previously through the use of native profilers, but the downside of that approach is that you can find hot instructions but it's not clear in what context the hot instructions are being executed nor what functions are hot. The goal of this profiler is to show pulley bytecode and time spent in bytecode itself to better understand the shape of code around a hot instruction to identify new macro opcodes for example.

The general structure of this new profiler is:

* There is a compile-time feature for Pulley which is off-by-default where, when enabled, Pulley will record its current program counter into an `AtomicUsize` before each instruction.

* When the CLI has `--profile pulley` Wasmtime will spawn a sampling thread in the same process which will periodically read from this `AtomicUsize` to record where the program is currently executing.

* The Pulley profiler additionally records all bytecode through the use of the `ProfilingAgent` trait to ensure that the recording has access to all bytecode as well.

* Samples are taken throughout the process and emitted to a `pulley-$pid.data` file. This file is then interpreted and printed by an "example" program `profiler-html.rs` in the `pulley/examples` directory.

The end result is that hot functions of Pulley bytecode can be seen and instructions are annotated with how frequently they were executed. This enables finding hot loops and understanding more about the whole loop, bytecodes that were selected, and such.


pulley/src/interp/tail_loop.rs
```diff
@@ -29,9 +29,14 @@
...
 use crate::opcode::Opcode;
+use crate::profile::ExecutingPcRef;
 use crate::ExtendedOpcode;
 
-type Handler = fn(Interpreter<'_>) -> Done;
+/// ABI signature of each opcode handler.
+///
+/// Note that this "explodes" the internals of `Interpreter` to individual
+/// arguments to help get them all into registers.
+type Handler = fn(&mut MachineState, UnsafeBytecodeStream, ExecutingPcRef<'_>) -> Done;
```
alexcrichton, line 39: I'll note that this change was done to ensure/guarantee that these three components of `Interpreter` are passed in registers. I was worried about crossing a threshold where if `Interpeter` got too big it would be passed by-ref instead of "exploded" into components like we want.

pulley/src/interp/tail_loop.rs
```diff
@@ -59,20 +64,47 @@ macro_rules! tail_call {
...
 
 impl Interpreter<'_> {
-    pub fn run(mut self) -> Done {
-        // Perform a dynamic dispatch through a function pointer indexed by
-        // opcode.
-        let opcode = unwrap_uninhabited(Opcode::decode(self.bytecode()));
-        let handler = OPCODE_HANDLER_TABLE[opcode as usize];
-        tail_call!(handler(self));
+    pub fn run(self) -> Done {
+        dispatch(self.state, self.pc, self.executing_pc)
```
alexcrichton, line 68: I'll note that this change is required to appease `become` because `run`'s signature differed from the handler's signature (`Interpreter`-by-value vs exploded-to-components)

pulley/src/interp/tail_loop.rs
```diff
@@ -59,20 +64,47 @@ macro_rules! tail_call {
...
+
+fn dispatch(
+    state: &mut MachineState,
+    pc: UnsafeBytecodeStream,
+    executing_pc: ExecutingPcRef<'_>,
+) -> Done {
+    // Perform a dynamic dispatch through a function pointer indexed by
+    // opcode.
+    let mut debug = debug(state, pc, executing_pc);
+    debug.before_visit();
```
alexcrichton, line 92: I'll note here that we were forgetting this before, which caused the profiler to not work with the tail loop initially.

pulley/src/interp/tail_loop.rs
alexcrichton: I'll also note that `ExecutingPcRef` is a zero-sized-type when `profile` is disabled, otherwise it's a pointer-large.

fitzgen: Makes sense, thanks for the explanation.

pulley/src/profile.rs
```diff
@@ -0,0 +1,248 @@
...
+        }
+    }
+
+    /// Returns whether the interpreter has been destroyed and will no longer
+    /// execute any code.
+    pub fn is_done(&self) -> bool {
+        // TODO: this doesn't work with `Clone` above in general and only works
+        // given the strict pattern of how everything is integrated today.
+        // Should probably integrate an `AtomicBool` or something like that and
+        // hook into `Drop for Vm` or something like that.
```
fitzgen, line 62: This seems like a fairly trivial change in terms of effort, but good payoff in terms of longer-term maintainability. Mind doing it now?

fitzgen (approved):
Super excited for this!

PR #10037, opened 2025-01-16
Winch: avoid emitting 32->64 extend on rmw ops on x64

MarinPostma:
I got confused by [this comment](https://github.com/bytecodealliance/wasmtime/pull/9990#discussion_r1916804597), and read the code too fast, but we actually emit an unnecessary `mov` when extending from 32 to 64bits on x64 in winch.

### DIFF
--- a/winch/codegen/src/masm.rs
+++ b/winch/codegen/src/masm.rs
@@ -252,35 +252,6 @@ impl ExtendKind {
             Self::I32Extend8S | Self::I32Extend8U | Self::I32Extend16U | Self::I32Extend16S => 32,
         }
     }
-
-    pub(crate) fn src_operand_size(&self) -> OperandSize {
-        match self {
-            ExtendKind::I32Extend8S
-            | ExtendKind::I32Extend8U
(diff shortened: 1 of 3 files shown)


winch/codegen/src/masm.rs
```diff
@@ -252,6 +252,35 @@ impl ExtendKind {
...
+            | ExtendKind::I32Extend16S
+            | ExtendKind::I32Extend16U
+            | ExtendKind::I64Extend8S => OperandSize::S32,
+            ExtendKind::I64Extend8U
+            | ExtendKind::I64Extend16S
+            | ExtendKind::I64Extend16U
+            | ExtendKind::I64Extend32S
+            | ExtendKind::I64Extend32U => OperandSize::S64,
+        }
+    }
```
saulecabrera, line 283: `from_bits` and `to_bits` already exist in `ExtendKind`, I believe that should do it here instead of introducing new methods?

MarinPostma: sure

saulecabrera (approved):

PR #10040, opened 2025-01-17
Move `*mut dyn VMStore` out of `VMContext`

alexcrichton:
There's no need for this type to be accessible from compiled code, so store and manage it entirely from Rust instead.


fitzgen (approved):
Nice

alexcrichton:
I believe the crashes in CI are due to https://github.com/bytecodealliance/wasmtime/pull/10069, so I'm going to re-queue this once that merges (didn't want to mix that up in over here)

Issue #9060, opened 2024-08-01
Idea: put VM data structures in arena with relative pointers for security and strict provenance


cfallin:
As part of the discussion on #9015 / #9026, we discussed handling of VM data structures -- the vmctx struct, tables, function references, and the like -- that are touched both by runtime code (in Rust) and by generated code compiled from the Wasm. There are issues related to strict pointer provenance because pointers to these data structures are exposed to the generated code, and/or the Pulley interpreter, without strict provenance (either through the Pulley bytecode, or through the machine code we invoke that is entirely outside of the domain of Rust's semantics).

It occurs to me that one way to solve this would be to make all VM data structures use *relative* pointers -- e.g., `u32` offsets -- in an arena (per store? per engine?) whose base pointer is a parameter both to the generated code and to the Pulley interpreter. We then trivially have strict provenance because there is *only one pointer* -- and whatever we need to do to preserve provenance (keep it as a pointer in the Pulley interpreter loop; and "expose" it as we pass it to generated code) is localized and manageable.

If provenance were the only benefit, that may not be so interesting; but there are a few others as well:

- Security: maintaining more discipline around raw pointers, and carefully dereferencing offsets into an arena instead (for which we can reserve guard regions just as we do heaps today), is a layer of mitigation/defense-in-depth against engine bugs. For example, if a miscompile or a bug in the generated CLIF caused a pointer-type confusion in VM data structures today, one could plausibly find a memory-read gadget or control-flow escape more easily. We will still have some raw pointers -- the actual code pointer that we invoke in funcrefs, or the memory base address in an imported or owned memory descriptor -- but fewer of them is less exposure.

  This angle is not new: WebKit has the [Gigacage](https://phakeobj.netlify.app/posts/gigacage/), I believe V8 has something similar, and I had suggested we build (and now we are building) our Wasm GC implementation in the same way with relative-offset-pointers; so it's a proven mitigation and the overhead seems to be minimal.

  Note also that relative-pointer loads/stores can be implemented with fully safe code in the Pulley interpreter. We almost certainly still need unsafe code still for the Wasm heap dereferences and such (though, then again, maybe there's a way around that by either externalizing a table of alternative raw pointer bases, or putting memories in a large gigacage -- teracage? exacage? -- as well).

[...]

PR #10043, opened 2025-01-17
 Provenance preparation for Pulley 

alexcrichton:
This commit is an internal refactoring of Wasmtime's runtime to prepare
to execute Pulley in MIRI. Currently today this is not possible because
Pulley does not properly respect either strict or permissive provenance
models. The goal of this refactoring is to enable fixing this in a
future commit that doesn't touch everything in the codebase. Instead
everything is touched here in this commit.

The basic problem with Pulley is that it is incompatible with the strict
provenance model of Rust which means that we'll be using "exposed
provenance" APIs to satisfy Rust's soundness requirements. In this model
we must explicitly call `ptr.expose_provenance()` on any pointers
which are exposed to compiled code. Arguably we should also be already
doing this for natively-compiled code but I am not certain about how
strictly this is required.

Currently in Wasmtime today we call `ptr.expose_provenance()` nowhere.
It also turns out, though, that we share quite a few pointers in quite a
few places with compiled code. This creates a bit of a problem! The
solution settled on in this commit looks like:

* A new marker trait, `VmSafe`, is introduced. This trait is used to
  represent "safe to share with compiled code" types and enumerates some
  properties such as defined ABIs, primitives wrappers match primitive
  ABIs, and notably "does not contain a pointer".

* A new type, `VmPtr<T>`, is added to represent pointers shared with
  compiled code. Internally for now this is just `SendSyncPtr<T>` but in
  the future it will be `usize`. By using `SendSyncPtr<T>` it shouldn't
  actually really change anything today other than requiring a lot of
  refactoring to get the types to line up.

* The core `vmctx_plus_offset*` methods are updated to require
  `T: VmSafe`. Previously they allowed any `T` which is relatively
  dangerous to store any possible Rust type in Cranelift-accessible
  areas.

These three fundamental changes were introduced in this commit. All
further changes were refactoring necessary to get everything working
after these changes. For example many types in `vmcontext.rs`, such as
`VMFuncRef`, have changed to using `VmPtr<T>` instead of `NonNull<T>` or
`*mut T`. This is a pretty expansive change which resulted in touching a
lot of places.

One premise of `VmPtr<T>` is that it's non-null. This was an
additional refactoring that updated a lot of places where previously
`*mut T` was used and now either `VmPtr<T>` or `NonNull<T>` is used.

In the end the intention is that `VmPtr<T>` is used whenever pointers
are store in memory that can be accessed from Cranelift. When operating
inside of the runtime `NonNull<T>` or `SendSyncPtr<T>` is preferred
instead.

As a final note, no provenance changes have actually happened yet. For
example this doesn't fix Pulley in MIRI. What it does enable, though, is
that the future commit to fix Pulley in MIRI will be much smaller with
this having already landed.

### DIFF
--- a/crates/wasmtime/src/runtime/func.rs
+++ b/crates/wasmtime/src/runtime/func.rs
@@ -1,7 +1,7 @@
 use crate::prelude::*;
 use crate::runtime::vm::{
     ExportFunction, InterpreterRef, SendSyncPtr, StoreBox, VMArrayCallHostFuncContext, VMContext,
-    VMFuncRef, VMFunctionImport, VMOpaqueContext,
+    VMFuncRef, VMFunctionImport, VMOpaqueContext, VmPtr,
 };
 use crate::runtime::Uninhabited;
 use crate::store::{AutoAssertNoGc, StoreData, StoreOpaque, Stored};
@@ -1060,13 +1060,14 @@ impl Func {
         let mut store = store.as_context_mut();
         let data = &store.0.store_data()[self.0];
         let func_ref = data.export().func_ref;
+        let params_and_returns = NonNull::new(params_and_returns).unwrap_or(NonNull::from(&mut []));
         Self::call_unchecked_raw(&mut store, func_ref, params_and_returns)
     }
 
     pub(crate) unsafe fn call_unchecked_raw<T>(
         store: &mut StoreContextMut<'_, T>,
         func_ref: NonNull<VMFuncRef>,
-        params_and_returns: *mut [ValRaw],
+        params_and_returns: NonNull<[ValRaw]>,
     ) -> Result<()> {
         invoke_wasm_and_catch_traps(store, |caller, vm| {
             func_ref.as_ref().array_call(
@@ -1320,7 +1321,7 @@ impl Func {
             };
             VMFunctionImport {
                 wasm_call: if let Some(wasm_call) = f.as_ref().wasm_call {
-                    wasm_call
+                    wasm_call.into()
                 } else {
                     // Assert that this is a array-call function, since those
                     // are the only ones that could be missing a `wasm_call`
@@ -1331,7 +1332,7 @@ impl Func {
                     module.wasm_to_array_trampoline(sig).expect(
                         "if the wasm is importing a function of a given type, it must have the \
                          type's trampoline",
-                    )
+                    ).into()
                 },
                 array_call: f.as_ref().array_call,
                 vmctx: f.as_ref().vmctx,
@@ -1594,7 +1595,7 @@ impl Func {
 /// can pass to the called wasm function, if desired.
 pub(crate) fn invoke_wasm_and_catch_traps<T>(
     store: &mut StoreContextMut<'_, T>,
-    closure: impl FnMut(*mut VMContext, Option<InterpreterRef<'_>>) -> bool,
+    closure: impl FnMut(NonNull<VMContext>, Option<InterpreterRef<'_>>) -> bool,
 ) -> Result<()> {
     unsafe {
         let exit = enter_wasm(store);
@@ -2026,7 +2027,7 @@ pub struct Caller<'a, T> {
 }
 
 impl<T> Caller<'_, T> {
-    unsafe fn with<F, R>(caller: *mut VMContext, f: F) -> R
+    unsafe fn with<F, R>(caller: NonNull<VMContext>, f: F) -> R
     where
         // The closure must be valid for any `Caller` it is given; it doesn't
         // get to choose the `Caller`'s lifetime.
@@ -2034,7 +2035,6 @@ impl<T> Caller<'_, T> {
         // And the return value must not borrow from the caller/store.
         R: 'static,
     {
-        debug_assert!(!caller.is_null());
         crate::runtime::vm::InstanceAndStore::from_vmctx(caller, |pair| {
             let (instance, mut store) = pair.unpack_context_mut::<T>();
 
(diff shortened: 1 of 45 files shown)


alexcrichton:
Note that this is temporarily based on https://github.com/bytecodealliance/wasmtime/pull/10040 which accounts for the first few commits.

cfallin:
> The basic problem with Pulley is that it is incompatible with the strict
provenance model of Rust 

I'll link here that we have at least one idea (#9060) that would make Pulley compatible with strict provenance, albeit as a major refactor to the way that Wasmtime's data structures work. No argument that with the design today it is basically incompatible with strict provenance.

alexcrichton:
Oh true good point! I'm a bit skeptical of the feasibility of such an implementation route though...

One thing I've also been wondering about though, which is a bit related, is let's ignore Pulley and what's our provenance story for Cranelift-compiled x64 code? For example is it possible in Rust's model to still have strict provenance when JIT code is in the mix? Are we required regardless to use "expose provenance" as a primitive to say "modifications to this memory are happening behind you're back". I'm not sure myself.

cfallin:
It seems to me at least (with my compiler optimizer hat on) that separate compilation is the ultimate "provenance black hole" -- calling JIT code is no different from, say, calling a C library, which could legally return any pointer derived from anything reachable that we pass into it. As long as Rust supports linking to code that is not compiled by the same rustc invocation (i.e., does not adopt a fully closed-world assumption), its provenance model has to support the "outside world" escape hatch that calling JIT code would also fall into. And unless strict provenance is going to say "you are no longer allowed to link C libraries, and all indirect calls have to be statically devirtualizable" that's as far as an analysis is going to get.

bjorn3:
> For example is it possible in Rust's model to still have strict provenance when JIT code is in the mix? Are we required regardless to use "expose provenance" as a primitive to say "modifications to this memory are happening behind you're back". I'm not sure myself.

Calling jitted code is no different from FFI. The foreign code may only access pointers for which it could have received provenance (either through exposed provenance or through a pointer that was passed in) Unluke with Pulley the compiler doesn't have visibility into what the foreign code does, so it for exanple doesn't know if you are loading an integer (erasing provenance) or loading a pointer (preserving provenance if it exists) or which of the available provenances was used for the memory operation. As such it has to assume that the foreign code picked the right provenance to avoid UB if the right provenance could have been picked in the first place due to being exposed to the foreign code.

alexcrichton:
Ok that makes sense to me yeah. If this becomes a problem in the future we can perhaps have a different implementation of `VmPtr` depending on if pulley is turned on or not. Without pulley we'd require strict provenance but with pulley we'd require only permissive provenance.

alexcrichton:
Ok this is now rebased and I believe should be ready for review @dicej

alexcrichton:
(I can also tag someone else in if you'd prefer too)

dicej:
> (I can also tag someone else in if you'd prefer too)

I think that would be a good idea.  I don't have any experience with the Pulley code so far and don't feel qualified to review something like this -- at least not yet.

dicej:
...although browsing the code, maybe there's not much here that's Pulley-specific?  In which case I could give it a shot.

alexcrichton:
Happy to go either way yeah, it's true yeah that while the main motivation here is Pulley that's mostly for a subsequent PR running miri and pulley stuff. This PR itself is mostly just refactoring vm internals towards a `VmPtr<T>` type and a `VmSafe` trait. I've tagged @fitzgen here but if you'd like to review I'm sure Nick won't mind heh

crates/wasmtime/src/runtime/vm/provenance.rs
```diff
@@ -0,0 +1,215 @@
...
+use core::fmt;
+use core::ptr::NonNull;
+use core::sync::atomic::{AtomicU64, AtomicUsize};
+use wasmtime_environ::VMSharedTypeIndex;
+
+/// A pointer that is used by compiled code, or in other words is accessed
+/// outside of Rust.
+///
+/// This is intended to be the fundamental data type used to share
+/// pointers-to-things with Cranelift-compiled code for example. An example of
```
fitzgen, line 54: In these comments there are a lot of references to Cranelift, but it could also be Winch. Might be worth rewording "Cranelift-compiled code" to "compiled Wasm code" or something.

crates/wasmtime/src/runtime/vm/provenance.rs
```diff
@@ -0,0 +1,215 @@
...
+/// pointers stored within `VMContext` itself).
+///
+/// In general usage of this type should be minimized to only where absolutely
+/// necessary when sharing data structures with compiled code. Prefer to use
+/// `NonNull` or `SendSyncPtr` where possible.
+///
+/// > **Note**: at this time this type was just introduced to Wasmtime. This
+/// > doesn't actually do anything with provenance just yet as the original
+/// > pointer used to create a `VmPtr<T>` is preserved even at-rest. That will
+/// > change in the near future as more refactorings are completed.
```
fitzgen, line 90: This comment is going to be out of date as soon as your next PR lands. Is it even worth including here? Seems like it will be easy to forget to remove down the line, might as well not even add it in the first place...

fitzgen (approved):
LGTM modulo a couple nitpicks

crates/wasmtime/src/runtime/func.rs
```diff
@@ -1060,13 +1060,14 @@ impl Func {
         let mut store = store.as_context_mut();
         let data = &store.0.store_data()[self.0];
         let func_ref = data.export().func_ref;
+        let params_and_returns = NonNull::new(params_and_returns).unwrap_or(NonNull::from(&mut []));
```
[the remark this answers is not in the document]
alexcrichton: To the best of my knowledge, which is limited, this is sound even if the compiler doesn't use a `'static` lifetime here. My understanding is that there's no soundness issue having a pointer to something that's deallocated, for example malloc'ing some data, saving the pointer, and then freeing the data. So even if the array here isn't `'static` it's safe to have a pointer just pointing to some random address on the stack.

The other reason I think this is safe is because this is creating `NonNull<[T]>` which not only has a pointer but also a length. For `&mut []` the length is always zero, and that means that the pointer, invalid though it may be, is never accessed.

So all in all I think it's safe because the pointer is never accessed due to the zero length on the pointer, regardless of whether rustc extends this to `'static` or not

PR #10044, opened 2025-01-17
async/stream/future support for wasmtime-wit-bindgen

dicej:
I've split this out of #9582 to make review easier.

This patch adds async/stream/future/error-context support to the host binding generator, along with placeholder type and function definitions in the `wasmtime` crate which the generated bindings can refer to.  See https://github.com/dicej/rfcs/blob/component-async/accepted/component-model-async.md#componentbindgen-updates for the design and rationale.

Note that I've added temporary `[patch.crates-io]` overrides in Cargo.toml until https://github.com/bytecodealliance/wit-bindgen/pull/1130 and https://github.com/bytecodealliance/wasm-tools/pull/1978 have been released.

Also note that we emit a `T: 'static` bound for `AsContextMut<Data = T>` when generating bindings with `concurrent_imports: true`.  This is only because `rustc` insists that the closure we're passing to
`LinkerInstance::func_wrap_concurrent` captures the lifetime of `T` despite my best efforts to convince it otherwise.  Alex and I suspect this is a limitation in the compiler, and I asked about it on the rust-lang Zulip, but we haven't been able to determine a workaround so far.

### DIFF
--- a/crates/component-macro/src/bindgen.rs
+++ b/crates/component-macro/src/bindgen.rs
@@ -1,14 +1,15 @@
 use proc_macro2::{Span, TokenStream};
 use quote::ToTokens;
-use std::collections::HashMap;
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 use std::env;
 use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicUsize, Ordering::Relaxed};
 use syn::parse::{Error, Parse, ParseStream, Result};
 use syn::punctuated::Punctuated;
 use syn::{braced, token, Token};
-use wasmtime_wit_bindgen::{AsyncConfig, Opts, Ownership, TrappableError, TrappableImports};
+use wasmtime_wit_bindgen::{
+    AsyncConfig, CallStyle, Opts, Ownership, TrappableError, TrappableImports,
+};
 use wit_parser::{PackageId, Resolve, UnresolvedPackageGroup, WorldId};
 
 pub struct Config {
@@ -20,10 +21,19 @@ pub struct Config {
 }
 
 pub fn expand(input: &Config) -> Result<TokenStream> {
-    if !cfg!(feature = "async") && input.opts.async_.maybe_async() {
+    if let (CallStyle::Async | CallStyle::Concurrent, false) =
+        (input.opts.call_style(), cfg!(feature = "async"))
+    {
         return Err(Error::new(
             Span::call_site(),
             "cannot enable async bindings unless `async` crate feature is active",
+        ));
+    }
+
+    if input.opts.concurrent_imports && !cfg!(feature = "component-model-async") {
+        return Err(Error::new(
+            Span::call_site(),
+            "cannot enable `concurrent_imports` option unless `component-model-async` crate feature is active",
         ));
     }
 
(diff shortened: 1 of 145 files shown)


rvolosatovs:
> Also note that we emit a `T: 'static` bound for `AsContextMut<Data = T>` when generating bindings with `concurrent_imports: true`. This is only because `rustc` insists that the closure we're passing to
> `LinkerInstance::func_wrap_concurrent` captures the lifetime of `T` despite my best efforts to convince it otherwise. Alex and I suspect this is a limitation in the compiler, and I asked about it on the rust-lang Zulip, but we haven't been able to determine a workaround so far.

In case it helps:

Since `async` is involved, any chance the error looks anything like https://github.com/rust-lang/rust/issues/63033 ?
If so, a pretty simple fix I have found looks as follows:
- https://github.com/bytecodealliance/wrpc/blob/200518efc47aaa13f8397fc80bdd891d7382f143/crates/transport/src/lib.rs#L44-L51
- https://github.com/bytecodealliance/wrpc/blob/200518efc47aaa13f8397fc80bdd891d7382f143/crates/transport/src/serve.rs#L190-L205

That's the same workaround used in `rustc` internally: https://github.com/rust-lang/rust/blob/ebbe63891f1fae21734cb97f2f863b08b1d44bf8/compiler/rustc_data_structures/src/captures.rs

Another `async` bug I've encountered, which appears when `Send` bound is involved is https://github.com/rust-lang/rust/issues/96865, it sometimes manifests as lifetime issues, but looks like the following when minimized: https://docs.rs/send-future/latest/send_future/trait.SendFuture.html
The fix is straightforward: https://github.com/rvolosatovs/send-future/blob/5637a5982b16b1938680ae9ecac659bdad28b5a7/src/lib.rs#L53-L60

dicej:
Thanks for the links, @rvolosatovs.  The compiler error I'm seeing doesn't match the ones in those issues, so I suspect it's something different.  Here's a (somewhat) minimal example of what I'm seeing:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=cd04abfec0133f3a4c0ed7ab53b04d4b

crates/component-macro/src/bindgen.rs
```diff
@@ -20,13 +21,22 @@ pub struct Config {
 }
 
 pub fn expand(input: &Config) -> Result<TokenStream> {
-    if !cfg!(feature = "async") && input.opts.async_.maybe_async() {
+    if let (CallStyle::Async | CallStyle::Concurrent, false) =
+        (input.opts.call_style(), cfg!(feature = "async"))
```
alexcrichton, line 25: TIL this is a neat idiom for lack of `if let foo = bar && some_other_cond { ... }`

crates/wasmtime/Cargo.toml
```diff
@@ -62,6 +62,7 @@ semver = { workspace = true, optional = true }
 smallvec = { workspace = true, optional = true }
 hashbrown = { workspace = true, features = ["default-hasher"] }
 bitflags = { workspace = true }
+futures = { workspace = true, features = ["alloc"] }
```
alexcrichton, line 65: Could this be made an optional dependency activated by `component-model-async`?

alexcrichton (approved):

PR #10046, opened 2025-01-17
Try using ARM64 runners from GitHub

alexcrichton:
Let's see if we can run our aarch64 tests natively instead of through QEMU emulation...

prtest:linux-arm64


alexcrichton:
Well that was ... exciting!

* For the runners that passed, looks like this is about 2x faster (probably in large part due to not building an optimized test suite on top of removing emulation costs)
* One runner died, with not really much of an indication as to why
* From what I can tell I can't see in the metadata of the runners whether it's arm64 or x64. That being said it downloaded/uses arm64 rustc's, so seems like it's arm64

Seems like it's good to have in our pocket and when it's more generally available we'll use it, but otherwise for the time being perhaps a bit rocky to use.

PR #10052, opened 2025-01-20
Fuzzing: Keep AVX flags enabled for Winch

jeffcharles:
Winch relies on AVX and AVX2 CPU features for implementing SIMD instructions on x64. The fuzzer sometimes disables these flags in Cranelift but the SIMD Wast tests are marked as supported if the Rust standard library says the current CPU has AVX and AVX2 support. So when the fuzzer disables these flags, the test is still marked as supported and then fails with an error saying AVX or AVX2 is required when the test is executed.

### DIFF
--- a/crates/fuzzing/src/generators/codegen_settings.rs
+++ b/crates/fuzzing/src/generators/codegen_settings.rs
@@ -31,6 +31,15 @@ impl CodegenSettings {
             }
         }
     }
+
+    /// Returns the flags used for codegen.
+    pub(crate) fn flags(&self) -> &[(String, String)] {
+        if let Self::Target { flags, .. } = self {
+            flags
+        } else {
+            &[]
+        }
+    }
 }
 
 impl<'a> Arbitrary<'a> for CodegenSettings {
--- a/crates/fuzzing/src/generators/config.rs
+++ b/crates/fuzzing/src/generators/config.rs
@@ -386,21 +386,6 @@ impl Config {
             }
         }
 
-        if self.wasmtime.compiler_strategy == CompilerStrategy::Winch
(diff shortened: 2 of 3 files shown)


alexcrichton:
Thanks for this! Figuring out how to navigate the fuzzers around this I always find a bit of a challenge. For example I thought we [already disabled the simd proposal for winch](https://github.com/bytecodealliance/wasmtime/blob/cd67a6090d515189a9c286bec2d993a591da3739/crates/fuzzing/src/generators/config.rs#L582), so how come Winch got some simd instructions?

Once we know simd is in the picture it also might make more sense to either accept this as a "this is allowed error" or fail earlier during the test case itself. We try to keep the fuzz test cases meaning the same thing across machines so instead of force-enabling AVX it might make more sense to see if AVX is either explicitly disabled or unavailable on hardware and in that case bail out and discard the fuzz test case basically

jeffcharles:
I had to add at least one method to the config to get access to the codegen flags in the oracle. Or at least I couldn't figure out another way to get access to the flags. The checks are also a little ugly because the value for the flag is typed as a `String` as opposed to a boolean.

alexcrichton (approved):
Looks good to me, and thanks! I was just looking at a similar-ish thing over at https://github.com/bytecodealliance/wasmtime/pull/10068 which jogged my memory and leads me to ask, mind adding a `log::warn!` call that the test case is being skipped like [this log here](https://github.com/bytecodealliance/wasmtime/blob/a018a5a9addb77d5998021a0150192aa955c71bf/crates/fuzzing/src/generators/codegen_settings.rs#L76-L77)?

PR #10053, opened 2025-01-20
winch: Tighten the definition of `ExtendKind`

saulecabrera:
Some instructions, like the atomic ones, part of the threads proposal, we want to ensure through the type system that only supported extend kinds are passed, that way the emission layer can ignore the extend kind validation at runtime.

This commit divides the existing `ExtendKind` enum into the signed and unsigned variants, making the definition more amenable to atomic operations.


saulecabrera:
Thanks for the suggestion @abrown, when with implementing conversions from `{Signed|Unsigned}Extend` -> `ExtendKind`

saulecabrera:
Hmm the failure is related  to `cargo vet`.

PR #10057, opened 2025-01-21
Add Pulley support to wasmtime_test macro

irrio:
This PR updates the `#[wasmtime_test]` macro to add support for Pulley.

This was one of the last remaining items in the Pulley TODO List #9747, I figured I'd try to knock it out.

I've renamed the `Cranelift` compilation strategy to `CraneliftNative` in order to disambiguate it from `CraneliftPulley` and to directly match the symbol in the `Compiler` enum.

`CraneliftPulley` is now included by default in the compilation strategies used by `#[wasmtime_test]`.

In order to support the tests that are targeting Winch specifically, I've added an `only(...)` specifier. So instead of `#[wasmtime_test(strategies(not(Cranelift)))]` it would be `#[wasmtime_test(strategies(only(Winch)))]`.

Now that the `wrap_and_typed_i31ref` test was running multiple times with different strategies, the `static HITS` counter was leaking state into the other test and causing things to fail. I moved the counter into the Store to get rid of the static variable.

### DIFF
--- a/crates/test-macros/src/lib.rs
+++ b/crates/test-macros/src/lib.rs
@@ -15,6 +15,15 @@
 //! }
 //! ```
 //!
+//! To use just one specific compiler strategy:
+//!
+//! ```rust
+//! #[wasmtime_test(strategies(only(Winch)))]
+//! fn my_test(config: &mut Config) -> Result<()> {
+//!     Ok(())
+//! }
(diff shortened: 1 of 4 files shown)


alexcrichton (approved):
Thanks for this!

PR #10059, opened 2025-01-21
Add audit for `wasmtime-math`

saulecabrera:
I noticed that CI is failing given that an audit and policy for `wasmtime-math` is missing.

`wasmtime-math` was introduced in
https://github.com/bytecodealliance/wasmtime/pull/9808/files.

I followed a similar approach to what it's used for all the other `wasmtime-*` crates.


abrown (approved):

alexcrichton:
Thanks! I still don't know how to prevent this from happening when we add a new crate alas...

PR #10060, opened 2025-01-21
Winch: fix atomic and/or/xor bug

MarinPostma:
This PR fixes atomic `and`/`or`/`xor`. The issue was that I was using the operand as a destination for the `AtomicRmwSeq`, but it is a macro-instruction, and the operand ended up being overwritten. This was caught when I enabled the `atomic.wast` spec tests in https://github.com/bytecodealliance/wasmtime/pull/10039#issuecomment-2603318507.

Investigating into the issue I realized that `AtomicRmwSeq` also had requirement for one of it's registers, but that didn't show up initially, because I got lucky, and the value was popped in the right register every time I tested.

The fix is to pop-push the operand to compute the address, and then pass the codegen context to masm to allocate the registers.

### DIFF
--- a/crates/environ/src/types.rs
+++ b/crates/environ/src/types.rs
@@ -221,6 +221,17 @@ impl WasmValType {
             | WasmValType::V128 => *self,
         }
     }
+
+    /// Attempt to build a `WasmValType` with the passed number of bits.
+    ///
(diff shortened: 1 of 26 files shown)


fitzgen (approved):
LGTM, thanks!

PR #10062, opened 2025-01-21
pulley: Implement full 128-bit multiplication

alexcrichton:
While Pulley has lowering rules for widening multiplication it didn't have a rule for a full 128-bit multiplication which is possible to generate through CLIF optimizations given wasm input. This commit adds such a lowering to the Cranelift backend but doesn't add any new instructions yet under the assumption this probably isn't perf-critical at this time.

### DIFF
--- a/cranelift/codegen/src/isa/pulley_shared/inst.isle
+++ b/cranelift/codegen/src/isa/pulley_shared/inst.isle
@@ -168,14 +168,6 @@
   )
 )
 
-(decl pure partial amode_add (Amode i32) Amode)
-(rule (amode_add (Amode.RegOffset base offset) amt)
-  (if-let new_offset (s32_add_fallible offset amt))
-  (Amode.RegOffset base amt))
-(rule (amode_add (Amode.SpOffset offset) amt)
-  (if-let new_offset (s32_add_fallible offset amt))
-  (Amode.SpOffset amt))
-
 (type ExtKind (enum None Sign32 Sign64 Zero32 Zero64))
 
 (type VExtKind (enum None S8x8 U8x8 S16x4 U16x4 S32x2 U32x2))
--- a/cranelift/codegen/src/isa/pulley_shared/inst.isle
+++ b/cranelift/codegen/src/isa/pulley_shared/inst.isle
@@ -168,6 +168,14 @@
   )
 )
 
+(decl pure partial amode_add (Amode i32) Amode)
+(rule (amode_add (Amode.RegOffset base offset) amt)
+  (if-let new_offset (s32_add_fallible offset amt))
+  (Amode.RegOffset base amt))
+(rule (amode_add (Amode.SpOffset offset) amt)
+  (if-let new_offset (s32_add_fallible offset amt))
+  (Amode.SpOffset amt))
+
 (type ExtKind (enum None Sign32 Sign64 Zero32 Zero64))
 
 (type VExtKind (enum None S8x8 U8x8 S16x4 U16x4 S32x2 U32x2))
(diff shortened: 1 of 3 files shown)


cranelift/codegen/src/isa/pulley_shared/inst.isle
```diff
@@ -168,6 +168,14 @@
   )
 )
 
+(decl pure partial amode_add (Amode i32) Amode)
+(rule (amode_add (Amode.RegOffset base offset) amt)
```
cfallin, line 172: Perhaps a comment here noting that we control `offset` (constants in our lowering rules) so we don't have to worry about the case where the `if-let` fails?

cranelift/codegen/src/isa/pulley_shared/lower.isle
```diff
@@ -331,11 +331,36 @@
...
   (pulley_xwidemul64_u (zext64 a) (zext64 b)))
-(rule (lower (has_type $I128 (imul (sextend a) (sextend b))))
+(rule 4 (lower (has_type $I128 (imul (sextend a) (sextend b))))
   (pulley_xwidemul64_s (sext64 a) (sext64 b)))
 
+;; for I128
+(rule (lower (has_type $I128 (imul x y)))
+  (let
+    ((x_regs ValueRegs x)
+      (x_lo XReg (value_regs_get x_regs 0))
```
cfallin, line 343: indented one space too far here? (sorry, minor issue, my Lisp sensibilities are just causing an eyeball twitch)

cranelift/codegen/src/isa/pulley_shared/lower.isle
```diff
@@ -1054,6 +1079,30 @@
 (rule 2 (lower (store flags src @ (value_type (ty_vec128 ty)) addr offset))
   (side_effect (pulley_vstore (amode addr offset) src ty flags)))
 
+;; i128 stores
+
+(rule 3 (lower (store flags src @ (value_type $I128) addr offset))
+  (let
+    ((src_regs ValueRegs src)
+      (src_lo XReg (value_regs_get src_regs 0))
```
cfallin, line 1087: likewise here

cranelift/codegen/src/isa/pulley_shared/inst.isle
cfallin: Actually -- we're adding to a user-controlled initial offset -- should we have a generic fallback here?

cfallin (approved):

cranelift/codegen/src/isa/pulley_shared/inst.isle
alexcrichton: This is only used for i128 stores which is only used by Cranelift trampolines in tests, and technically it's universally correct just perhaps not as useful as it could be in the sense that it'll fail if the offset addition overflows in the future. Basically I wasn't sure if it was worth trying to be 100% correct here when nothing known will hit this, so I opted for a "will loudly fail" in the future rather than filling out all the particulars to get this working optimally. I'll look at it again though and see if I can't think of an always-works solution.

cranelift/codegen/src/isa/pulley_shared/lower.isle
alexcrichton: Currently ISLE backends differ a bit on style where the x64 backend likes to line things up by the previous line in ISLE (I think the style you're expecting here) but the riscv64 backend takes a different approach where the next line is just indented by one more than the previous line at all times (which is the style I was writing here). Personally I prefer the riscv64 style because I find it easier to work with editors and I don't have to think too hard about lining things up all the time. 

That being said I don't really have a strong preference either way, but in the absence of a style guide/formatter I'll probably leave this as-is since it doesn't deviate too far from the current style (AFAIK)

cfallin: If you prefer a different bikeshed color then don't take my comment too seriously -- in the fullness of time maybe we'll have an autoformatter (and then everything will be in the One True S-expr Indent Style) but no worries otherwise :-)

PR #10066, opened 2025-01-21
Winch: Even tighter Extends

MarinPostma:
Followup to https://github.com/bytecodealliance/wasmtime/pull/10053

This PR has two commit:

1) The first commit makes `mov*` instruct receive the adequate `Extend` operation. And it makes extend conversion  operation harder to misuse.
2) The second commit is **optional**. This is an alternative to `Signed/UnsignedExtend`, that merges the two into a `Extend<T>`, and factorizes common behaviour, avoiding repetition. The `T` is either `Signed` or `Zero`, with precautions in the design so that it's hard to misuse.

### DIFF
--- a/winch/codegen/src/masm.rs
+++ b/winch/codegen/src/masm.rs
@@ -199,81 +199,76 @@ pub(crate) enum ShiftKind {
 /// lowering to machine code.
 #[derive(Copy, Clone)]
 pub(crate) enum ExtendKind {
-    Signed(SignedExtend),
-    Unsigned(UnsignedExtend),
+    Signed(Extend<Signed>),
+    Unsigned(Extend<Zero>),
 }
 
 #[derive(Copy, Clone)]
-pub(crate) enum UnsignedExtend {
-    /// 8 to 32 bit unsigned extend.
-    I32Extend8U,
-    /// 16 to 32 bit unsigned extend.
-    I32Extend16U,
-    /// 8 to 64 bit unsigned extend.
-    I64Extend8U,
-    /// 16 to 64 bit unsigned extend.
-    I64Extend16U,
-    /// 32 to 64 bit unsigned extend.
-    I64Extend32U,
+pub(crate) enum Signed {}
+#[derive(Copy, Clone)]
+pub(crate) enum Zero {}
+
+pub(crate) trait ExtendType {}
+
+impl ExtendType for Signed {}
+impl ExtendType for Zero {}
+
+#[derive(Copy, Clone)]
+pub(crate) enum Extend<T: ExtendType> {
+    /// 8 to 32 bit extend.
+    I32Extend8,
+    /// 16 to 32 bit extend.
+    I32Extend16,
+    /// 8 to 64 bit extend.
+    I64Extend8,
+    /// 16 to 64 bit extend.
+    I64Extend16,
+    /// 32 to 64 bit extend.
+    I64Extend32,
+
+    /// Variant to hold the kind of extend marker.
+    ///
+    /// This is `Signed` or `Zero`, that are empty enums, which means that this variant cannot be
+    /// constructed.
+    __Kind(T),
 }
 
-impl From<UnsignedExtend> for ExtendKind {
-    fn from(value: UnsignedExtend) -> Self {
+impl From<Extend<Zero>> for ExtendKind {
+    fn from(value: Extend<Zero>) -> Self {
         ExtendKind::Unsigned(value)
     }
 }
 
-impl UnsignedExtend {
-    pub fn from_bits(&self) -> u8 {
+impl<T: ExtendType> Extend<T> {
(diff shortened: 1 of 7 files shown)


winch/codegen/src/masm.rs
```diff
@@ -199,81 +199,76 @@ pub(crate) enum ShiftKind {
...
+    /// 16 to 64 bit extend.
+    I64Extend16,
+    /// 32 to 64 bit extend.
+    I64Extend32,
+
+    /// Variant to hold the kind of extend marker.
+    ///
+    /// This is `Signed` or `Zero`, that are empty enums, which means that this variant cannot be
+    /// constructed.
+    __Kind(T),
```
saulecabrera, line 233: To avoid an unused member, could you use `PhantomData` to hold the type instead? https://doc.rust-lang.org/std/marker/struct.PhantomData.html

winch/codegen/src/masm.rs
```diff
@@ -199,81 +199,76 @@ pub(crate) enum ShiftKind {
...
+impl From<Extend<Zero>> for ExtendKind {
+    fn from(value: Extend<Zero>) -> Self {
         ExtendKind::Unsigned(value)
     }
 }
 
-impl UnsignedExtend {
-    pub fn from_bits(&self) -> u8 {
+impl<T: ExtendType> Extend<T> {
+    pub fn src_size(&self) -> OperandSize {
```
saulecabrera, line 243: Small naming convention/suggestion: we usually use `src`, `dst` for explicit moves, could we keep `from_bits` and `to_bits`? Although given that this now returns `OperandSize` it could maybe be `from_size` and `to_size`?

saulecabrera (commented):
Thanks for this. Left two minor comments inline.

winch/codegen/src/masm.rs
MarinPostma: do you mean for every variant? like `I64ExtendI8(PhantomData<T>)`? I think that would make construction pretty cumbersome, no?
If you mean `__Kind(PhamtomData<T>)`, then that defeats the purpose of this variant being unconstructible, since you can't construct `Signed` and `Zero`

winch/codegen/src/masm.rs
MarinPostma: No problem. It's just that it kinda overlaps with usual rust naming convention for `from_` and `to_`, which I found a bit confusing. I'll make the change now.

winch/codegen/src/masm.rs
saulecabrera: I meant more around the lines of:

```rust
pub enum ExtendBits {
    /// 8 to 32 bit extend.
    I32Extend8,
    /// 16 to 32 bit extend.
    I32Extend16,
    /// 8 to 64 bit extend.
    I64Extend8,
    /// 16 to 64 bit extend.
    I64Extend16,
    /// 32 to 64 bit extend.
    I64Extend32,
}

pub struct Extend<T> {
    bits: ExtendBits,
    marker: PhantomData<T>,
}
```

Which is the kind of job that `PhantomData` is for. 

That way you can avoid the unused enum member. In your approach `Signed` and `Zero` are mostly used as type flags to ensure that the extension is associated correctly, but even though your comment states that the `__Kind(T)` variant cannot be constructed, from a practical perspective, you could do `Extend::__Kind(Signed)` I believe.

saulecabrera: The general disadvantage that I see is that this approach adds a bit more complexity to the implementation.

saulecabrera: But we can always iterate on this after landing this change, I think regardless of the unused member, extends are in a better place now.

saulecabrera (approved):

winch/codegen/src/masm.rs
MarinPostma: | from a practical perspective, you could do Extend::__Kind(Signed) I believe.

that's in fact not possible :) You would need to exhibit a variant, but there are none, c.f: https://doc.rust-lang.org/reference/items/enumerations.html#zero-variant-enums

MarinPostma: I think your proposal could work nicely too, if we implemented `Deref<Target = ExtendBits>` for `Extend<T: ExtendType>`

MarinPostma: we'd need to add contructors to make instanciation simpler, though, which may hurt readability

saulecabrera: Oh I see, I overlooked, I read this as `struct` not `enum` (for `Signed` and `Zero`), but this makes sense. Thanks for clarifying. 

> we'd need to add contructors to make instanciation simpler, though, which may hurt readability


Yeah, I think the current approach is simple enough, if there's a strong reason to change it in the future then we can. For now I'd suggest we keep it as is.

PR #10068, opened 2025-01-21
Disable Winch fuzzing on non-x86_64 architectures

fitzgen:


alexcrichton (approved):
Thinking again about this, maybe add a `log::warn!` as well here? (I think we do that for CPU features where if one is requested and the local CPU doesn't have it we log to that effect)

alexcrichton:
Ah [this is the warning](https://github.com/bytecodealliance/wasmtime/blob/a018a5a9addb77d5998021a0150192aa955c71bf/crates/fuzzing/src/generators/codegen_settings.rs#L76-L77) I'm thinking of

PR #10069, opened 2025-01-21
Fix argument extension on riscv64 for Wasmtime builtins

alexcrichton:
This commit fixes a crash found in the CI of #10040. That PR itself isn't the fault per-se but rather uncovered a preexisting issue on riscv64. According to riscv64's ABI docs it looks like arguments are all expected to be sign-extended, whereas currently in Wasmtime all host signatures are zero-extended on all platforms. This commit applies two changes to fix this:

* A new `TargetIsa::default_argument_extension` method was added which is now used instead of unconditionally using `uext` to apply to all arguments/results.

* While I was here I went ahead and split apart things where the wasm signature of a builtin doesn't use `uext` or `sext`. The host signature still does, however, which means that any extension necessary happens in the trampoline we generate per-libcall as opposed to at all callsites.

I'm not certain that all platforms require `uext` but I've left the `TargetIsa` implementation as `uext` for now with a comment explaining why. Currently the only non-`uext` platforms are riscv64, which is `sext` to fix the issue from #10040, and Pulley which is "none" as things work differently there.

### DIFF
--- a/cranelift/codegen/src/isa/aarch64/mod.rs
+++ b/cranelift/codegen/src/isa/aarch64/mod.rs
@@ -1,7 +1,7 @@
 //! ARM 64-bit Instruction Set Architecture.
 
 use crate::dominator_tree::DominatorTree;
-use crate::ir::{Function, Type};
+use crate::ir::{self, Function, Type};
 use crate::isa::aarch64::settings as aarch64_settings;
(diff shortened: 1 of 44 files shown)


abrown (approved):
Thanks for the comments!

PR #10079, opened 2025-01-22
Attempt to speed up windows runners

alexcrichton:
I read somewhere on the internet that `C:\` may be slow for I/O. Let's try moving the Rust target directory to somewhere else and see if it has any affect on performance

prtest:full


alexcrichton:
I'm not seeing a significant change, so closing.

PR #10081, opened 2025-01-22
pulley: Add a multiply-and-add macro instruction

alexcrichton:
This is present in riscv64 and aarch64 native ISAs and was found in a benchmark I was looking at so let's add a macro-op as well to help cases where this crops up in the wild.


alexcrichton:
For runtests for this new instruction looks like we [already have a number of runtests](https://github.com/bytecodealliance/wasmtime/blob/ca9557646aac2d625a6d1d040e6caf065d0340f6/cranelift/filetests/filetests/runtests/arithmetic.clif#L200-L267) for multiply-and-add.

fitzgen (approved):

PR #10082, opened 2025-01-22
Winch: Add replace_lane instructions for x64 with AVX

jeffcharles:
Part of #8093. Implements the `*.replace_lane` set of instructions.


saulecabrera:
I can take this review.

winch/codegen/src/isa/x64/asm.rs
```diff
@@ -166,6 +166,19 @@ impl From<ExtendKind> for ExtMode {
...
+        match self {
+            OperandSize::S8 => AvxOpcode::Vpinsrb,
+            OperandSize::S16 => AvxOpcode::Vpinsrw,
+            OperandSize::S32 => AvxOpcode::Vpinsrd,
+            OperandSize::S64 => AvxOpcode::Vpinsrq,
+            _ => unimplemented!(),
+        }
+    }
+}
+
```
saulecabrera, line 181: I'd like to keep `OperandSize` as machine independent as possible. Would it be possible to move this to a helper in the x64 masm/asm?

saulecabrera (commented):
LGTM, one minor remark.

winch/codegen/src/isa/x64/asm.rs
jeffcharles: The `vpinsr_opcode` method isn't accessible outside of the x86 assembler module and shouldn't show up in autocomplete when using the Rust Analyzer. I can put on a static function instead in any case.

If someone tried to call it outside of that module, say in the macroassembler or code context, they would get the following compiler error:

```
error[E0624]: method `vpinsr_opcode` is private
    --> winch/codegen/src/isa/x64/masm.rs:1554:36
     |
1554 |         let foo = kind.lane_size().vpinsr_opcode();
     |                                    ^^^^^^^^^^^^^ private method
     |
    ::: winch/codegen/src/isa/x64/asm.rs:171:5
     |
171  |     pub(self) fn vpinsr_opcode(&self) -> AvxOpcode {
     |     ---------------------------------------------- private method defined here

For more information about this error, try `rustc --explain E0624`
```

saulecabrera (approved):

PR #10083, opened 2025-01-22
add component-model-async/lift.wast test

dicej:
This is another piece of #9582 which I'm splitting out to make review easier.

This test includes two components: one which exports a function using the async-with-callback ABI, and another which uses the async-without-callback ABI. It doesn't actually instantiate or run either component yet.

The rest of the changes fill in some TODOs to make the test pass.

### DIFF
--- a/crates/wasmtime/src/runtime/vm/component.rs
+++ b/crates/wasmtime/src/runtime/vm/component.rs
@@ -375,6 +375,16 @@ impl ComponentInstance {
         }
     }
 
+    /// Same as `set_runtime_memory` but for async callback function pointers.
+    pub fn set_runtime_callback(&mut self, idx: RuntimeCallbackIndex, ptr: NonNull<VMFuncRef>) {
+        unsafe {
+            let storage =
+                self.vmctx_plus_offset_mut::<VmPtr<VMFuncRef>>(self.offsets.runtime_callback(idx));
+            debug_assert!((*storage).as_ptr() as usize == INVALID_PTR);
+            *storage = ptr.into();
+        }
+    }
+
     /// Same as `set_runtime_memory` but for post-return function pointers.
     pub fn set_runtime_post_return(
         &mut self,
@@ -491,6 +501,11 @@ impl ComponentInstance {
             for i in 0..self.offsets.num_runtime_reallocs {
                 let i = RuntimeReallocIndex::from_u32(i);
                 let offset = self.offsets.runtime_realloc(i);
+                *self.vmctx_plus_offset_mut(offset) = INVALID_PTR;
+            }
+            for i in 0..self.offsets.num_runtime_callbacks {
+                let i = RuntimeCallbackIndex::from_u32(i);
+                let offset = self.offsets.runtime_callback(i);
                 *self.vmctx_plus_offset_mut(offset) = INVALID_PTR;
             }
             for i in 0..self.offsets.num_runtime_post_returns {
@@ -732,6 +747,11 @@ impl OwnedComponentInstance {
     /// See `ComponentInstance::set_runtime_realloc`
     pub fn set_runtime_realloc(&mut self, idx: RuntimeReallocIndex, ptr: NonNull<VMFuncRef>) {
         unsafe { self.instance_mut().set_runtime_realloc(idx, ptr) }
+    }
+
+    /// See `ComponentInstance::set_runtime_callback`
+    pub fn set_runtime_callback(&mut self, idx: RuntimeCallbackIndex, ptr: NonNull<VMFuncRef>) {
+        unsafe { self.instance_mut().set_runtime_callback(idx, ptr) }
     }
 
     /// See `ComponentInstance::set_runtime_post_return`
--- a/crates/fuzzing/src/generators/config.rs
+++ b/crates/fuzzing/src/generators/config.rs
@@ -139,6 +139,7 @@ impl Config {
             extended_const,
             wide_arithmetic,
             component_model_more_flags,
+            component_model_async,
             simd,
 
             hogs_memory: _,
@@ -151,6 +152,7 @@ impl Config {
         self.module_config.function_references_enabled =
             function_references.or(gc).unwrap_or(false);
         self.module_config.component_model_more_flags = component_model_more_flags.unwrap_or(false);
+        self.module_config.component_model_async = component_model_async.unwrap_or(false);
 
         // Enable/disable proposals that wasm-smith has knobs for which will be
         // read when creating `wasmtime::Config`.
@@ -266,6 +268,7 @@ impl Config {
             .wasm_wide_arithmetic(self.module_config.config.wide_arithmetic_enabled)
             .wasm_extended_const(self.module_config.config.extended_const_enabled)
             .wasm_component_model_more_flags(self.module_config.component_model_more_flags)
+            .wasm_component_model_async(self.module_config.component_model_async)
             .native_unwind_info(cfg!(target_os = "windows") || self.wasmtime.native_unwind_info)
             .cranelift_nan_canonicalization(self.wasmtime.canonicalize_nans)
             .cranelift_opt_level(self.wasmtime.opt_level.to_wasmtime())
--- a/crates/fuzzing/src/generators/module.rs
+++ b/crates/fuzzing/src/generators/module.rs
@@ -16,6 +16,7 @@ pub struct ModuleConfig {
     // config-to-`wasmtime::Config` translation.
     pub function_references_enabled: bool,
     pub component_model_more_flags: bool,
+    pub component_model_async: bool,
 }
 
 impl<'a> Arbitrary<'a> for ModuleConfig {
@@ -62,6 +63,7 @@ impl<'a> Arbitrary<'a> for ModuleConfig {
 
         Ok(ModuleConfig {
             component_model_more_flags: false,
+            component_model_async: false,
             function_references_enabled: config.gc_enabled,
             config,
         })
--- a/crates/misc/component-test-util/Cargo.toml
+++ b/crates/misc/component-test-util/Cargo.toml
@@ -11,7 +11,7 @@ publish = false
 env_logger = { workspace = true }
 anyhow = { workspace = true }
 arbitrary = { workspace = true, features = ["derive"] }
-wasmtime = { workspace = true, features = ["component-model", "async"] }
+wasmtime = { workspace = true, features = ["component-model", "async", "component-model-async"] }
 wasmtime-environ = { workspace = true }
 wasmtime-wast-util = { path = '../../wast-util' }
 target-lexicon = { workspace = true }
--- a/crates/misc/component-test-util/src/lib.rs
+++ b/crates/misc/component-test-util/src/lib.rs
@@ -166,6 +166,7 @@ pub fn apply_test_config(config: &mut Config, test_config: &wasmtime_wast_util::
         extended_const,
         wide_arithmetic,
         component_model_more_flags,
+        component_model_async,
         nan_canonicalization,
         simd,
 
@@ -184,6 +185,7 @@ pub fn apply_test_config(config: &mut Config, test_config: &wasmtime_wast_util::
     let extended_const = extended_const.unwrap_or(false);
     let wide_arithmetic = wide_arithmetic.unwrap_or(false);
     let component_model_more_flags = component_model_more_flags.unwrap_or(false);
+    let component_model_async = component_model_async.unwrap_or(false);
     let nan_canonicalization = nan_canonicalization.unwrap_or(false);
     let relaxed_simd = relaxed_simd.unwrap_or(false);
 
@@ -210,5 +212,6 @@ pub fn apply_test_config(config: &mut Config, test_config: &wasmtime_wast_util::
         .wasm_extended_const(extended_const)
         .wasm_wide_arithmetic(wide_arithmetic)
         .wasm_component_model_more_flags(component_model_more_flags)
+        .wasm_component_model_async(component_model_async)
         .cranelift_nan_canonicalization(nan_canonicalization);
 }
--- a/crates/wasmtime/src/config.rs
+++ b/crates/wasmtime/src/config.rs
@@ -1111,6 +1111,19 @@ impl Config {
         self
     }
 
+    /// Configures whether components support the async ABI [proposal] for
+    /// lifting and lowering functions, as well as `stream`, `future`, and
+    /// `error-context` types.
+    ///
+    /// Please note that Wasmtime's support for this feature is _very_ incomplete.
+    ///
+    /// [proposal]: https://github.com/WebAssembly/component-model/blob/main/design/mvp/Async.md
+    #[cfg(feature = "component-model-async")]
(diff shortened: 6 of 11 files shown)


crates/wasmtime/src/runtime/vm/component/states.rs
```diff
@@ -0,0 +1,134 @@
+use {
```
dicej, line 1: This file was shamelessly copied from resources.rs and modified for use with stream/future/error-context state.  Those handles are different enough from resource handles that I didn't want to try to address everything with a single data structure.  I'm open to other approaches, though.

crates/wasmtime/src/runtime/component/mod.rs
```diff
@@ -672,3 +672,158 @@ pub mod bindgen_examples;
 #[cfg(not(any(docsrs, test, doctest)))]
 #[doc(hidden)]
 pub mod bindgen_examples {}
+
+#[cfg(not(feature = "component-model-async"))]
+pub(crate) mod concurrent {
```
alexcrichton, line 677: To confirm, this'll eventually be a `concurrent.rs` in a future PR? So just a placeholder for now?

crates/wasmtime/src/runtime/component/values.rs
```diff
@@ -1000,3 +1031,12 @@ fn unexpected<T>(ty: InterfaceType, val: &Val) -> Result<T> {
...
 }
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct FutureAny(pub(crate) u32);
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct StreamAny(pub(crate) u32);
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ErrorContextAny(pub(crate) u32);
```
alexcrichton, line 1042: How temporary is the representation of these structures? E.g. is the `u32` what it'll be in the long run?

crates/wasmtime/src/runtime/component/values.rs
```diff
@@ -198,9 +202,9 @@ impl Val {
 
                 Val::Flags(flags.into())
             }
-            InterfaceType::Future(_)
-            | InterfaceType::Stream(_)
-            | InterfaceType::ErrorContext(_) => todo!(),
+            InterfaceType::Future(_) => FutureReader::<()>::lift(cx, ty, next(src))?.into_val(),
```
alexcrichton, line 205: The `::<()>` bit here and below gives me a bit of pause. Is that temporary or final-into-the-future?

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -1,14 +1,404 @@
...
+        Self {
+            rep,
+            _phantom: PhantomData,
+        }
+    }
+
+    /// Convert this `FutureReader` into a [`Val`].
+    pub fn into_val(self) -> Val {
+        Val::Future(FutureAny(self.rep))
+    }
```
alexcrichton, line 89: Could this perhaps be `impl<T> From<FutureReader<T>> for Val`?

crates/wasmtime/src/runtime/component/func/options.rs
```diff
@@ -196,7 +196,7 @@ pub struct LowerContext<'a, T> {
     /// into.
     ///
     /// This pointer is required to be owned by the `store` provided.
-    instance: *mut ComponentInstance,
+    pub(crate) instance: *mut ComponentInstance,
```
alexcrichton, line 199: For this and the below change I had to reread and re-figure-out what I was originally intending with this. I think it would be preferable to not add `pub(crate)` here as that expands the number of locations that need to be audited when working with this field.

I think it should be ok though to add accessors like `fn instance(&self) -> &ComponentInstance` and `fn instance_mut(&mut self) -> &mut ComponentInstance` to these types (lift/lower contexts). That would remove the need for `unsafe` at access sites and might clean up some preexisting access sites as well

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -1,14 +1,404 @@
...
+
+unsafe impl func::ComponentType for ErrorContext {
+    const ABI: CanonicalAbiInfo = CanonicalAbiInfo::SCALAR4;
+
+    type Lower = <u32 as func::ComponentType>::Lower;
+
+    fn typecheck(ty: &InterfaceType, _types: &InstanceType<'_>) -> Result<()> {
+        match ty {
+            InterfaceType::ErrorContext(_) => Ok(()),
+            other => bail!("expected `error`, found `{}`", func::desc(other)),
```
alexcrichton, line 367: s/error/error-context/

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -1,14 +1,404 @@
...
+                match state {
+                    StreamFutureState::Local => {
+                        *state = StreamFutureState::Write;
+                    }
+                    StreamFutureState::Read => {
+                        state_table.remove_by_index(index)?;
+                    }
+                    StreamFutureState::Write => bail!("cannot transfer write end of future"),
+                    StreamFutureState::Busy => bail!("cannot transfer busy future"),
+                }
```
alexcrichton, line 120: Personally I know I'm in a state of "I'm not a component-model async expert" so coming across bits of nontrivial code like this I'm left to go find where in the spec this behavior is implementing and trying to cross-reference that for example. Basically I find myself unable to reason about this locally and understand, just from what's here, what's going on.

Could you add some comments for any sort of nontrivial transformations like this? For example with resources I [tried to leave comments](https://github.com/bytecodealliance/wasmtime/blob/8a969897f76e9725778f713a31024651f7e3e72b/crates/wasmtime/src/runtime/component/resources.rs#L646-L669) about all the various state transitions in lifts/lowers.

I suspect the majority of future readers of this code are also unlikely to be component-model async experts and comments can help quite a bit in understanding what's going on. The more I read in various places in this PR this is also applicable in general to any nontrivial bits here as opposed to just this one function. For example `get_mut_by_index_from` above is easy enough to understand as it's pretty mechanical internally but in isolation I'm not really sure what it's doing in a broader sense just from reading it. It may not be too helpful to understand at-a-glance what it's doing but at the same time having a comment I think would still be helpful.

(aka this is sort of a general request for more comments on anything nontrivial)

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -1,14 +1,404 @@
...
+            };
+            if *actual_ty != ty {
+                bail!("invalid future handle");
+            }
+            (rep, state)
+        }
+    })
+}
+
+fn host_close_reader<U, S: AsContextMut<Data = U>>(store: S, rep: u32) -> Result<()> {
```
alexcrichton, line 67: I'm not sure how much this will persist int he future but you can drop the `<Data = U>` here in the sense that it's not necessary as part of this definition. In the future if you need bounds on `U` it should be possible to do `S::Data: Bound` I think too

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -1,14 +1,404 @@
...
+    _ = (store, rep);
+    todo!()
+}
 
 /// Represents the readable end of a Component Model `future`.
 pub struct FutureReader<T> {
+    rep: u32,
```
alexcrichton, line 74: Mind adding some comments as to what this `rep` is? My naive understanding of component model async is probably showing here but I'm familiar with "rep" of a resource where it's user-supplied but I didn't think that there was anything user-supplied here but rather this is an index to something (but I'm not sure)

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -1,14 +1,404 @@
...
+                            "error context table index present in (sub)component table during lift",
+                        )
+                        .get_mut_by_index(index)?
+                };
+
+                Ok(Self { rep })
+            }
+            _ => func::bad_type_info(),
+        }
+    }
```
alexcrichton, line 356: I'll definitely admit that I'm very lost when it comes to error-context here. I see a local refcount table and a global refcount table but I don't actually see the global refcount table being used. I'm also not really sure what `rep` is in the context of an error-context so I'm pretty lost here.

While comments would help here is this something that may be best deferred to a future PR? For example ideally there'd be a test exercising this lifting and lowering and testing various errors here and things like that, but I suspect such a test isn't quite ready yet?

crates/wasmtime/src/runtime/vm/component.rs
```diff
@@ -148,6 +184,37 @@ pub struct VMComponentContext {
...
+    /// Represents a future handle.
+    Future(TypeFutureTableIndex, StreamFutureState),
+}
+
+/// Represents the state associated with an error context
+#[derive(Debug, PartialEq, Eq, PartialOrd)]
+pub struct ErrorContextState {
+    /// Debug message associated with the error context
+    pub(crate) debug_msg: String,
+}
```
alexcrichton, line 216: This might be one part about error-context I don't understand b/c I think I'm missing where this is used...

crates/wasmtime/src/runtime/vm/component/states.rs
alexcrichton: Could you expand the documentation for this struct as to what problem it's solving? For example what extra functionality it has over `ResourceTable`. I'm a bit surprised for example of the `reps_to_indexes` reverse mapping where most of the component model stuff doesn't require hash maps today and that seems like it's bordering on needing something like that.

Also this is something I probably should have done for `ResourceTable` but if it's possible this'd be a good place to have some `#[test]` unit tests for this data structure doing some simple pushes/pops/etc.

alexcrichton (commented):
I'm definitely pretty far out of my comfort zone reviewing this in the sense that there's enough pieces that I don't see how they're connected just in this PR that I'm not confident I have the ability to review without going over to the spec. At a high level the test added here seems like it probably doesn't require the lift/lower impls, but then again I suspect that removing those impls would remove all the meat of the PR. Would it be possible to add some tests exercising the impls added here in this PR?

crates/wasmtime/src/runtime/component/mod.rs
dicej: `concurrent.rs` is [already in main](https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasmtime/src/runtime/component/concurrent.rs) and was added in https://github.com/bytecodealliance/wasmtime/pull/10044 (and will be expanded in subsequent PRs).  This is the stub module to be used when the `component-model-async` feature is disabled, so it will only go away when that feature is removed.

Alternatively, I could remove this stub module and make `concurrent.rs` the only implementation, but that would mean adding more `#[cfg(...)]` annotations elsewhere, so it's kind of a tradeoff.  Happy to do whichever you think is best, though.

crates/wasmtime/src/runtime/component/values.rs
dicej: They're like resource reps in that they'll be `u32`s for wasm32 but become `u64`s when we add wasm64 support for the component model.

crates/wasmtime/src/runtime/component/values.rs
dicej: It's "final" in that I wasn't planning to change it, but I agree it looks weird.  I had already written the static API `Lift`/`Lower` impls for these types, so I figured I'd reuse those implementations here since they're payload-type-agnostic.  I'll move the lifting and lowering code into separate, top-level functions which both the static API and dynamic API can use, which will avoid this.

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
dicej: Agreed and apologies for forgetting to add comments; I have "Add doc and code comments to new, non-trivial code (especially concurrent.rs and futures_and_streams.rs)" in the task list on #9582, and that was my next priority before I switched my focus to breaking the PR up into smaller pieces.  I'll start doing both now.

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
dicej: Good catch.  This is just a stub, with the signature copied from https://github.com/dicej/wasmtime/blob/d1596df0361a4a8c956ba7ef34ac8d3c5e0140e1/crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs#L578, but it looks like the real version doesn't use `U` either, so I'm guessing I had originally been using `U`, then stopped using it, but forgot to update the signature.

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
dicej: Yeah, I can replace this stuff with stubs.  My approach with this PR was to run the lift.wast test, hit a `todo!()`, replace the `todo!()` with code copied from #9582, let the compiler tell me what _other_ code _that_ code needed, then either copying more code or adding stubs to make the compiler happy.  Looks like I copied over too much, and that's making the review more difficult; sorry about that.  I'll trim it down.

crates/wasmtime/src/runtime/vm/component.rs
dicej: I don't think it _is_ being used yet; again, I may have copied over more than necessary from the Big PR.

dicej:
@alexcrichton I've "rebooted" this PR from scratch, this time only pulling in the minimum code needed for the `lift.wast` test.  It's much smaller now.

As a consequence, most of your review comments no longer apply, but I'll address them in future PRs.  Sorry for the churn; I'll try to keep future PRs more focused.

crates/wasmtime/src/runtime/component/mod.rs
alexcrichton: Ah I see, sorry I missed the `not(...)`!

I'll take a closer look in a future PR to see how it fits together 👍

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
alexcrichton: No worries of course! I myself basically never write comments as I write code itself and then I have to force myself to go back afterwards and comment things, which I often forget to do. I just wanted to highlight this as something I think will be important to do as pieces incrementally land as opposed to only once at the end

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
alexcrichton: How about something like `rep: WaitableIndex`? Basically using a type here to indicate more clearly what it's pointing to. My history with resources makes me think that "rep" comes from guests themselves and needs to be preserved and handed back to the guest but I don't think that's what's going on here, so this isn't so much a "representation" internally but moreso an index somewhere else, which is where I think a custom newtype might help to distinguish

alexcrichton (approved):
Looks good to me, thanks!

Issue #9980, opened 2025-01-10
Divergent behavior between regalloc_algorithm={backtracking,single_pass}


alexcrichton:
Found via oss-fuzz in https://issues.oss-fuzz.com/issues/387110342 I've minimized this to:

```wasm
(module
  (func (export "")
    call 1
    f64.const 0
    f64.const 0
    f64.ne
    if
    end
    call 0
  )

  (func
    f64.const nan
    f64.const 0
    f64.eq
    if
      loop
      end
    end
  )
)
```

which I can show different behavior with:

```
$ wasmtime run --invoke '' -W fuel=$((1<<62)) -C cranelift-regalloc-algorithm=backtracking foo.wat
Error: failed to run main module `foo.wat`

Caused by:
    0: failed to invoke ``
    1: error while executing at wasm backtrace:
           0: <unknown>!<wasm function 1>
           1:   0x1e - <unknown>!<wasm function 0>
           2:   0x36 - <unknown>!<wasm function 0>
...
         16365:   0x36 - <unknown>!<wasm function 0>
         16366:   0x36 - <unknown>!<wasm function 0>
    2: wasm trap: call stack exhausted
```

That's expected, this infinitely recurses. With `single_pass` though:

```
$ wasmtime run --invoke '' -W fuel=$((1<<62)) -C cranelift-regalloc-algorithm=single_pass foo.wat
Error: failed to run main module `foo.wat`

Caused by:
    0: failed to invoke ``
    1: error while executing at wasm backtrace:
           0:   0x1d - <unknown>!<wasm function 0>
           1:   0x36 - <unknown>!<wasm function 0>
    2: wasm trap: all fuel consumed by WebAssembly
```

cfallin:
For context: Alex and I briefly discussed this offline and concluded this was not a security issue as `backtracking` (the production algorithm) is performing the correct behavior, and single-pass is off by default and not part of any support tier (or perhaps tier 3 by default).

cfallin:
Additional context: it seems this only reproduces with x86-64; cannot reproduce with aarch64 (also infinitely recurses correctly). Successfully reproduces on macOS/x86-64 (via Rosetta 2), though, in addition to Linux.

alexcrichton:
Another test that came up in @Robbepop's differential fuzzing of wasmi vs wasmtime:

```wasm
(module
  (type (;0;) (func (result i32)))
  (global (;0;) (mut i32) i32.const 0)
[...]

PR #10086, opened 2025-01-23
Cranelift/x64 backend: do not use one-way branches.

cfallin:
In #9980, we saw that code compiled with the single-pass register allocator has incorrect behavior. We eventually narrowed this down to the fact that the single-pass allocator is inserting code meant to be at the end of a block, just before its terminator, *between* two branches that form the terminator sequence. The allocator is correct; the bug is with Cranelift's x64 backend.

When we produce instructions into a VCode container, we maintain basic blocks, and we have the invariant (usual for basic block-based IR) that only the last -- terminator -- instruction is a branch that can leave the block. Even the conditional branches maintain this invariant: though VCode is meant to be "almost machine code", we emit *two-target conditionals* that are semantically like "jcond; jmp". We then are able to optimize this inline during binary emission in the `MachBuffer`: the buffer knows about unconditional and conditional branches and will "chomp" branches off the tail of the buffer whenever they target the fallthrough block. (We designed the system this way because it is simpler to think about BBs that are order-invariant, i.e., not bake the "fallthrough" concept into the IR.) Thus we have a simpler abstraction but produce optimal terminator sequences.

Unfortunately, when adding a branch-on-floating-point-compare lowering, we had the need to branch to a target if either of *two* conditions were true, and rather than add a new kind of terminator instruction, we added a "one-armed branch": conditionally branch to label or fall through. We emitted this in sequence right before the actual terminator, so semantically it was almost equivalent.

I write "almost" because the register allocator *is* allowed to insert spills/reloads/moves between any two instructions. Here the distinct pieces of the terminator sequence matter: the allocator might insert something just before the last instruction, assuming the basic-block "single in, single out" invariant means this will always run with the block. With one-armed branches this is no longer true.

The backtracking allocator (our original RA2 algorithm, and still the default today) will never insert code at the end of a block when it has multiple terminators, because it associates such block-start/end insertions with *edges*; so in such conditions it inserts instructions into the tops of successor blocks instead. But the single-pass allocator needs to perform work at the end of every block, so it will trigger this bug.

This PR removes `JmpIf` and converts the br-of-fcmp lowering to use `JmpCondOr` instead, which is a pseudoinstruction that does `jcc1; jcc2; jmp`. This maintains the BB invariant and fixes the bug.

Note that Winch still uses `JmpIf`, so we cannot remove it entirely: this PR renames it to `WinchJmpIf` instead, and adds a mechanism to assert failure if it is ever added to `VCode` (rather than emitted directly, as Winch's macro-assembler does). We could instead write Winch's `jmp_if` assembler function in terms of `JmpCond` with a fallthrough label that is immediately bound, and let the MachBuffer always chomp the jmp; I opted not to regress Winch compiler performance by doing this. If one day we abstract out the assembler further, we can remove `WinchJmpIf`.

This is one of two instances of a "one-armed branch"; the other is s390x's `OneWayCondBr`, used in `br_table` lowerings, which we will address separately. Once we do, that will address #9980 entirely.

### DIFF
--- a/cranelift/codegen/src/isa/x64/inst/emit.rs
+++ b/cranelift/codegen/src/isa/x64/inst/emit.rs
@@ -1622,7 +1622,8 @@ pub(crate) fn emit(
             inst.emit(sink, info, state);
 
             // jne  .loop_start
-            // TODO: Encoding the JmpIf as a short jump saves us 4 bytes here.
+            // TODO: Encoding the conditional jump as a short jump
+            // could save us us 4 bytes here.
             one_way_jmp(sink, CC::NZ, loop_start);
 
             // The regular prologue code is going to emit a `sub` after this, so we need to
@@ -1891,7 +1892,7 @@ pub(crate) fn emit(
             sink.put4(0x0);
         }
 
-        Inst::JmpIf { cc, taken } => {
+        Inst::WinchJmpIf { cc, taken } => {
             let cond_start = sink.cur_offset();
             let cond_disp_off = cond_start + 2;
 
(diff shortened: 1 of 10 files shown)


cfallin:
(Thanks to @alexcrichton and others for bottoming this out into the root cause; discussed earlier today in the Cranelift meeting.)

cranelift/codegen/src/isa/x64/inst/emit.rs
```diff
@@ -1936,6 +1937,76 @@ pub(crate) fn emit(
...
+            //   jcc1 taken
+            //   jcc2 taken
+            //   jmp not_taken
+            //
+            // Note that we enroll both conditionals in the
+            // branch-chomping mechanism because MachBuffer
+            // simplification can continue upward as long as it keeps
+            // chomping branches. In the best case, if taken ==
+            // not_taken and that one block is the fallthrough block,
+            // all three branches can disappear.
```
fitzgen, line 1956: Mind writing a test for this case where all three branches can get chomped and replaced with a fallthrough?

fitzgen (approved):
Very nice! Thanks for trying to be defensive with the low-level branching instruction bits as well. Kind of annoying that we are dealing with that. Hopefully a separate assembler library will help here...

PR #10087, opened 2025-01-23
Cranelift/s390x: do not use one-way conditional branches.

### LINKED ISSUE #9980: Divergent behavior between regalloc_algorithm={backtracking,single_pass}

cfallin:
This is a followup to #10086, this time removing the one-armed branch variant for s390x. This branch was only used as the default-target branch in the `br_table` lowering. This PR incorporates the branch into the `JTSequence` pseudo-instruction. Some care is needed to keep the `ProducesBool` abstraction; it is unwrapped into its `ProducesFlags` and the `JTSequence` becomes a `ConsumesFlags`, so the compare for the jump-table bound (for default target) is not part of the pseudoinst. (This is OK because regalloc-inserted moves never alter flags, by explicit contract; the same reason allows cmp/branch terminators.)

Along the way I noticed that the comments on `JTSequence` claimed that `targets` included the default, but this is (no longer?) the case, as the targets are unwrapped by `jump_table_targets` which peels off the first (default) separately. Aside from comments, this only affected pretty-printing; codegen was correct.

With this, we have no more one-armed branches; hence, this fixes #9980.


cfallin:
cc @uweigand if you are available to review? Otherwise we're probably OK with the usual reviewer rotation...

cranelift/codegen/src/isa/s390x/inst.isle
```diff
@@ -3358,11 +3352,6 @@
 (rule (cond_br_bool (ProducesBool.ProducesBool producer cond) taken not_taken)
       (with_flags_side_effect producer (cond_br taken not_taken cond)))
 
-;; Emit a one-way conditional branch based on a boolean condition.
-(decl oneway_cond_br_bool (ProducesBool MachLabel) SideEffectNoResult)
-(rule (oneway_cond_br_bool (ProducesBool.ProducesBool producer cond) dest)
-      (with_flags_side_effect producer (oneway_cond_br dest cond)))
```
uweigand, line 3364: Couldn't we have a `jt_sequence_bool` that implicitly decomposes the `ProducesBool` like the above?  Then we wouldn't have to introduce the `bool_producer` / `bool_cond` accessors just for this case ...

cranelift/filetests/filetests/isa/s390x/jumptable.clif
```diff
@@ -29,34 +29,33 @@ block5(v5: i32):
 ; VCode:
 ; block0:
 ;   llgfr %r3, %r2
+;   sllg %r4, %r3, 2
 ;   clgfi %r3, 3
-;   jghe label4
-;   sllg %r3, %r3, 2
```
uweigand, line 34: The generated code is slightly worse now as the shifting is done before the branch.  This probably doesn't have a very significant impact on the pipeline, but it does increase register pressure.  One alternative would be to compare the shifted value instead (assuming there are no overflow considerations).

But I guess this isn't serious enough that it should prevent merging this.

uweigand (approved):
Looks generally good to me, but see inline for a couple of comments.

As a longer-term improvement, I'm wondering if the whole default case handling couldn't just be done by common code using regular basic blocks and branch instructions, enabling all the normal optimization machinery?  The back-end would then only have to do the actual tablejump (simply assuming that no overflow can happen).

cranelift/codegen/src/isa/s390x/inst.isle
cfallin: Great idea, done!

cranelift/filetests/filetests/isa/s390x/jumptable.clif
cfallin: Hmm, that's true -- I did the code motion without looking at the instruction semantics in detail to optimize. We can optimize in a followup if needed I suppose.

cfallin:
Could I bother someone for a rubber-stamp here (@uweigand's not an official approver) -- @alexcrichton or @fitzgen maybe? Thanks!

alexcrichton (approved):

PR #10092, opened 2025-01-23
Winch: x64 `wait`, `notify` and `fence`

MarinPostma:
This PR implements the following instruction in winch for the x64 backend:
- `atomic.fence`
- `memory.atomic.wait32`
- `memory.atomic.wait64`
- `memory.atomic.notify`


saulecabrera:
I can help with this one.

winch/codegen/src/codegen/mod.rs
```diff
@@ -1417,6 +1424,146 @@ where
...
+        self.context
+            .stack
+            .push(TypedReg::new(WasmValType::I32, mem).into());
+
+        // compute the offset if necessary.
+        self.masm.extend(
+            writable!(addr.reg),
+            addr.reg,
+            ExtendKind::Unsigned(Extend::I64Extend32),
+        )?;
```
saulecabrera, line 1470: I think this extend shouldn't be emitted unconditionally: in the case of `memory64`, the address is expected to be `64` bits, so there shouldn't be a need to emit this. 

Also, 32-bit loads are zero extend by default, so I think we can omit this extend entirely?

winch/codegen/src/codegen/mod.rs
```diff
@@ -1417,6 +1424,146 @@ where
...
+        self.context
+            .stack
+            .push(TypedReg::new(WasmValType::I32, mem).into());
+
+        // compute offset if necessary.
+        self.masm.extend(
+            writable!(addr.reg),
+            addr.reg,
+            ExtendKind::Unsigned(Extend::I64Extend32),
+        )?;
```
saulecabrera, line 1540: Similar comments here regarding the immediate and the stack and also the extend.

winch/codegen/src/codegen/mod.rs
MarinPostma: Let me check that again, if I recall correctly I added it because it didn't work like I expected with negative offsets

MarinPostma: nevermind, I must have imagined that...

MarinPostma:
I made the suggested changes. I think we could avoid popping the stack in the case where we don't have to compute an offset, and just insert the target memory instead, but it complicates the implementation slightly. Happy to do if you think it's better.

saulecabrera (approved):
LGTM, thanks.

PR #10096, opened 2025-01-23
pulley: Execute a wasm module under miri

alexcrichton:
This commit adds a test to CI and a script locally to execute which will run an entire wasm module under Pulley. The goal of this commit is to add The Test for miri execution of wasm. In general miri is too slow to run for the full test suite and even for this single test it takes a very long time to compile the one small module here. To help with this the module is precompiled on native for Pulley and then deserialized in miri itself, meaning that we skip miri execution of Cranelift entirely.

The goal of this commit is to eventually expand this test to cover lots of little and basic operations of wasm which touch VM state. For now it's just a simple smoke test that doesn't run much but it will be expanded over time. Making it much larger than now already turns up miri violations so I wanted to land an initial scaffold first before expanding later.

Getting this test to pass requires changing the `VmPtr<T>` introduced in #10043 to use a `NonZeroUsize` internally rather than `NonNull<T>`. This is because Pulley is only compatible with exposed provenance which means we need to actually expose the provenance of pointers.

Both Pulley and Wasmtime need to deal with exposed provenance APIs, but such APIs are not available in Wasmtime's current MSRV of 1.82. These APIs were instead introduced as stable in Rust 1.84. In lieu of waiting a few months because I'm impatient I've added a small build script to both crates to detect the rustc version and see whether provenance APIs are available. These build script modifications will no longer be necessary once our MSRV is 1.84+.

prtest:miri

### DIFF
--- a/crates/wasmtime/build.rs
+++ b/crates/wasmtime/build.rs
@@ -1,5 +1,10 @@
+use std::process::Command;
+use std::str;
+
 fn main() {
     println!("cargo:rerun-if-changed=build.rs");
+
+    enable_features_based_on_rustc_version();
 
     // NB: duplicating a workaround in the wasmtime-fiber build script.
     println!("cargo:rustc-check-cfg=cfg(asan)");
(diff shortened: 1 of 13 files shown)


crates/wasmtime/build.rs
```diff
@@ -92,3 +97,26 @@ fn build_c_helpers() {
...
+        println!("cargo:rustc-cfg=has_provenance_apis");
+    }
+    println!("cargo:rustc-check-cfg=cfg(has_provenance_apis)");
+}
+
+fn rustc_minor_version() -> Option<u32> {
+    let rustc = std::env::var("RUSTC").unwrap();
+    let output = Command::new(rustc).arg("--version").output().ok()?;
+    let version = str::from_utf8(&output.stdout).ok()?;
+    let mut pieces = version.split('.');
```
dicej, line 117: Nit: Could use `version.split_once('.')?` here (and in pulley/build.rs) to simplify the code a bit.

dicej (approved):

crates/wasmtime/build.rs
alexcrichton: Sort of but it'd end up requiring two invocations: one for looking at `rustc 1` and the next for looking at the N in `rustc 1.N.0`. Given that I think I'll leave it as `.split('.')`, but I'm also just [copy/pasting this detection code around](https://github.com/bytecodealliance/wasm-tools/blob/7530629a3a127a1429b4837a1b1632f11d3dfeea/crates/wasmparser/build.rs#L18-L27) so I could also be missing something

PR #10098, opened 2025-01-24
pulley: Run more functions in the miri provenance test

alexcrichton:
This commit executes a few more functions in Pulley which uncovered a bug where the pulley stack was not properly aligned on the host. This commit refactors things to ensure that the stack is 16-byte aligned on the host. While here the stack is additionally updated to be allocated as uninitialized to avoid paying the initialization cost on each Pulley interpreter being created.


fitzgen (approved):
:+1:

PR #10106, opened 2025-01-24
add component-model-async/{fused|futures|streams}.wast tests

dicej:
This is another piece of #9582 which I'm splitting out to make review easier.

The fused.wast test exercises fused adapter generation for various flavors of
intercomponent async->async, async->sync, and sync->async calls.

The futures.wast and streams.wast tests exercise the various intrinsics
(e.g. `stream.read`, `future.close_writable`, etc.) involving `future`s and
`stream`s.

The remaining changes fill in some TODOs to make the tests pass, plus plumbing
for a few intrinsics which aren't needed for these tests but which set the
foundation for future tests.


crates/cranelift/src/compiler/component.rs
```diff
@@ -262,7 +258,163 @@ impl<'a> TrampolineCompiler<'a> {
...
+            ),
+        );
+
+        let call = self.compiler.call_indirect_host(
+            &mut self.builder,
+            index,
+            host_sig,
+            host_fn,
+            &[vmctx, ty, values_vec_ptr, values_vec_len],
+        );
```
alexcrichton, line 318: You might be able to replace `host::task_return` + `load_libcall` + this with the `call_libcall` helper perhaps?

crates/cranelift/src/compiler/component.rs
```diff
@@ -262,7 +258,163 @@ impl<'a> TrampolineCompiler<'a> {
...
+            i64::from(
+                params
+                    .and_then(|params| {
+                        self.types
+                            .get_task_return_type(&TypeTaskReturn { params })
+                            .map(|v| v.as_u32())
+                    })
+                    .unwrap_or(u32::MAX),
+            ),
+        );
```
alexcrichton, line 310: Two things on this:

* For `let params = ...` above, it's unclear to me what this is doing. For example I'm not sure why this would switch to `None` for unknown types as opposed to return an error or panic or something. Additionally this looks like something that's better done in perhaps a `translate.rs` or `inline.rs` step? 
* For `let ty = ...` this looks like it's doing a hash map lookup of a calculated type. Would it be possible to calculate that when the type is originally created early on in `translate.rs` and/or `inline.rs` and then thread that through to here? For example I'd expect that this function here would have `ty: Option<SomeIndex>` as an argument and then that's what's encoded into a `u32` here.

crates/cranelift/src/compiler/component.rs
```diff
@@ -262,7 +258,163 @@ impl<'a> TrampolineCompiler<'a> {
...
+        // remaining parameters
+        callee_args.extend(args[2..].iter().copied());
+
+        let call = self.compiler.call_indirect_host(
+            &mut self.builder,
+            index,
+            host_sig,
+            host_fn,
+            &callee_args,
+        );
```
alexcrichton, line 392: Like above, the `call_libcall` helper might be able to help clean this up a little further

crates/cranelift/src/compiler/component.rs
```diff
@@ -799,7 +951,7 @@ impl<'a> TrampolineCompiler<'a> {
         self.raise_if_host_trapped(succeeded);
     }
 
-    fn raise_if_resource_trapped(&mut self, ret: ir::Value) -> ir::Value {
+    fn raise_if_i32_trapped(&mut self, ret: ir::Value) -> ir::Value {
```
alexcrichton, line 954: Perhaps something like `raise_if_negative_one`? It looks a bit odd above that an i64 return value is passed into a method that says `if_i32`.

crates/environ/src/fact/trampoline.rs
```diff
@@ -456,12 +881,21 @@ impl Compiler<'_, '_> {
...
-        let dst_flat =
+        let src_flat = self.types.flatten_types(
+            lift_opts,
+            if lift_opts.async_ {
+                MAX_FLAT_PARAMS
+            } else {
+                MAX_FLAT_RESULTS
+            },
+            src_tys.iter().copied(),
+        );
```
alexcrichton, line 892: This looks pretty similar to some code in `signature.rs` (and I'm similarly a bit confused about params/results a little). Is this similar enough that it might warrant a helper of some kind to do this internally?

(similar for `dst_flat` below, some sort of helper that takes `lower_opts` or `async_` which handles async-vs-not)

crates/environ/src/fact/trampoline.rs
```diff
@@ -404,16 +817,28 @@ impl Compiler<'_, '_> {
...
-            // destination module for the parameters via its `realloc` function.
-            let abi = CanonicalAbiInfo::record(dst_tys.iter().map(|t| self.types.canonical_abi(t)));
-            let (size, align) = if lift_opts.memory64 {
-                (abi.size64, abi.align64)
+            if lift_opts.async_ {
+                let align = dst_tys
+                    .iter()
+                    .map(|t| self.types.align(lift_opts, t))
+                    .max()
+                    .unwrap_or(1);
```
alexcrichton, line 825: How come the `::record` constructor couldn't be used like before?

alexcrichton (commented):
I think this is as far as I'm gonna get today. I haven't gotten to most of `trampoline.rs` yet which I realize is most of the guts of this PR, but I'll do that on Monday

crates/environ/src/fact/signature.rs
```diff
@@ -59,7 +93,70 @@ impl ComponentTypesBuilder {
...
+        let ty = &self[options.ty];
+        let ptr_ty = options.options.ptr();
+
+        let mut params = vec![ptr_ty];
+
+        let mut results_indirect = false;
+        let results = match self.flatten_types(
+            &options.options,
+            MAX_FLAT_PARAMS,
+            self[ty.params].types.iter().copied(),
```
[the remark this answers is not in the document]
dicej: Yeah, it's a little confusing since `task.return` allows you to return up to `MAX_FLAT_PARAMS` on the stack.  I'll leave a comment to explain.

crates/wasmtime/src/runtime/vm/component/libcalls.rs
```diff
@@ -9,6 +9,12 @@ use core::ptr::NonNull;
 use core::slice;
 use wasmtime_environ::component::TypeResourceTableIndex;
 
+#[cfg(feature = "component-model-async")]
+use {
+    crate::{vm::VMFuncRef, ValRaw},
+    wasmtime_environ::component::{RuntimeComponentInstanceIndex, TypeTaskReturnIndex},
+};
+
```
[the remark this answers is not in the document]
dicej: Sure, no problem.  I originally _did_ match the existing style, but then found it convenient to use this style so I could put all the imports under a single `#[cfg(...)]` annotation.  I'll switch it back.

crates/wasmtime/src/runtime/vm/component/libcalls.rs
```diff
@@ -572,6 +578,109 @@ unsafe fn trap(_vmctx: NonNull<VMComponentContext>, code: u8) -> Result<Infallib
     Err(wasmtime_environ::Trap::from_u8(code).unwrap().into())
 }
 
+unsafe fn task_return(
+    vmctx: NonNull<VMComponentContext>,
+    ty: u32,
+    storage: *mut u8,
+    storage_len: usize,
+) -> Result<()> {
+    #[cfg(feature = "component-model-async")]
```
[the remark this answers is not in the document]
dicej: Yeah, I was wondering if there was a cleaner way to do this.  I'm open to suggestions.

crates/environ/src/fact.rs
```diff
@@ -73,6 +78,17 @@ pub struct Module<'a> {
...
+    /// Track allocated global variables by type.
+    ///
+    /// globals_by_type[0] => offsets into `globals` representing I32 globals
+    /// globals_by_type[1] => offsets into `globals` representing I64 globals
+    /// globals_by_type[2] => offsets into `globals` representing F32 globals
+    /// globals_by_type[3] => offsets into `globals` representing F64 globals
+    globals_by_type: [Vec<u32>; 4],
+    globals: Vec<ValType>,
+
+    exports: Vec<(u32, String)>,
```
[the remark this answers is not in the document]
dicej: Correct, it's all about `ref.func` validation.  If there's a better way to address that, I'm definitely open to it.

crates/cranelift/src/compiler/component.rs
```diff
@@ -262,7 +258,163 @@ impl<'a> TrampolineCompiler<'a> {
...
+    fn translate_task_return_call(&mut self) {
+        match self.abi {
+            Abi::Wasm => {}
+
+            // These trampolines can only actually be called by Wasm, so
+            // let's assert that here.
+            Abi::Array => {
+                self.builder.ins().trap(TRAP_INTERNAL_ASSERT);
+                return;
+            }
```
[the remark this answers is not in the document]
dicej: I believe we discussed this at some point when pairing in a Slack huddle and decided then to use the trap approach.  So far it seems to be working for every test I throw at it (including cases that don't involve fused adapters at all), so if there's a scenario where this is the wrong behavior, I'd like to understand what it is and add a test case for it.

crates/cranelift/src/compiler/component.rs
dicej: Yeah, I've already done that (and removed other duplication) in a follow-up PR I'm currently finishing up.  Can it wait for that PR?

crates/cranelift/src/compiler/component.rs
dicej: Agreed about the `None` default; I don't remember why I did that.  I'll switch it to a panic.

And yes, I can move this code to an earlier stage, e.g. `translate.rs` or `inline.rs`.

crates/cranelift/src/compiler/component.rs
dicej: Per my response above: I've already done this in a follow-up PR.

crates/environ/src/fact/trampoline.rs
dicej: Ah, I think this is an artifact of when Luke changed the async lift ABI; previously parameter passing worked differently for async vs. sync lifts, and now they're the same.  In the churn I neglected to unify the code.  I'll do that.

crates/cranelift/src/compiler/component.rs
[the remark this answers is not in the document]
dicej: OK, now I remember the conversation.  I guess the scenario where this would be a problem is if a guest exported its `task.return` intrinsic to the host and the host called it.  I'll see if I can create a test case for that.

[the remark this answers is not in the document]
dicej: I won't be able to actually test this until I've filled more of the runtime code (coming in future PRs).  Would it be okay if I left a TODO comment here for the purposes of this PR?

I'd also like to propose changing the component model spec to prohibit re-exporting imported intrinsics.  I'm guessing we can't catch that reliably during validation due to e.g. `call_indirect` shenanigans, but we'd at least be justified in trapping at runtime.  @lukewagner do you have an opinion?

crates/wasmtime/src/runtime/vm.rs
```diff
@@ -189,6 +221,9 @@ pub unsafe trait VMStore {
     /// Metadata required for resources for the component model.
     #[cfg(feature = "component-model")]
     fn component_calls(&mut self) -> &mut component::CallContexts;
```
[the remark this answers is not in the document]
dicej: Sounds like we're agreed having a separate trait makes sense, so I'm leaving this as-is.

crates/cranelift/src/compiler/component.rs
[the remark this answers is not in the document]
dicej: Alternatively, if it's straightforward to support `Abi::Array` properly here we can just do that; I'll probably need some guidance on how to do that, though.

crates/environ/src/fact.rs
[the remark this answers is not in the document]
alexcrichton: Ah ok, if it's only about `ref.func`, then what you'll want to use is a "declare" element segment. You'll call [this function](https://docs.rs/wasm-encoder/latest/wasm_encoder/struct.ElementSection.html#method.declared) in wasm-encoder basically. That way you won't have to export the functions, but it will still be valid to `ref.func` the functions.

crates/cranelift/src/compiler/component.rs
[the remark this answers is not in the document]
alexcrichton: Ah yeah I think you booted up on the context here. It's the reexport case that I'm worried about (and you're right that I don't think we can reasonably forbid it at the embedder level). I think it's probably pretty easy to add `Abi::Array` support, it's mostly just around refactoring to use helpers to access the arguments/store returns, otherwise it's all the same internals. 

I think it'd be reasonable to leave this as a TODO, though, in the interim. Mind filing an issue about that? (can probably start a `component-model-async` label too perhaps)

crates/cranelift/src/compiler/component.rs
alexcrichton: Sounds good 👍

crates/environ/src/fact/trampoline.rs
```diff
@@ -80,50 +85,168 @@ struct Compiler<'a, 'b> {
...
+            types: module.types,
+            module,
+            code: Vec::new(),
+            nlocals: sig.params.len() as u32,
+            free_locals: HashMap::new(),
+            traps: Vec::new(),
+            result,
+            fuel: INITIAL_FUEL,
+            emit_resource_call: false,
+        }
```
alexcrichton, line 170: Mind adding a private `Compiler::new` constructor? That should help deduplicate this and the `Compiler` constructions above too

crates/environ/src/fact/trampoline.rs
```diff
@@ -80,50 +85,168 @@ struct Compiler<'a, 'b> {
...
+                result,
+                fuel: INITIAL_FUEL,
+                emit_resource_call,
+            },
+            lower_sig,
+            lift_sig,
+        )
+    }
+
+    let start_adapter = |module: &mut Module, param_globals| {
```
alexcrichton, line 128: Mind throwing "async" in the name of these local variables (and `return_adapter` below) to clarify they're async-related things?

crates/environ/src/fact/trampoline.rs
```diff
@@ -80,50 +85,168 @@ struct Compiler<'a, 'b> {
...
+        result
+    };
+
     match (adapter.lower.options.async_, adapter.lift.options.async_) {
-        (false, false) => {}
+        (false, false) => {
+            let (compiler, lower_sig, lift_sig) = compiler(module, adapter);
+            compiler.compile_sync_to_sync_adapter(adapter, &lower_sig, &lift_sig)
+        }
         (true, true) => {
```
alexcrichton, line 181: Mind leaving some high-level comments on each of these cases about what sort of adapters are being created? Mostly to help guide what adapters are calling what and what the purpose of the various adapter functions are.

alexcrichton: (I'm also not sure at-a-glance, for example, what the `param_globals`/`result_globals` variables are, so either docs here or on the methods below I think would be useful)

crates/environ/src/fact/trampoline.rs
```diff
@@ -257,7 +380,294 @@ struct Memory<'a> {
...
+        let param_locals = sig
+            .params
+            .iter()
+            .enumerate()
+            .map(|(i, ty)| (i as u32, *ty))
+            .collect::<Vec<_>>();
+
+        self.set_flag(adapter.lower.flags, FLAG_MAY_LEAVE, false);
+        self.translate_results(adapter, &param_locals, &param_locals);
+        self.set_flag(adapter.lower.flags, FLAG_MAY_LEAVE, true);
```
alexcrichton, line 656: I'm a bit perplexed by this and I've been trying to think about this but to no avail.

High-level request: mind writing up some longer-form documentation in this module or on a function here about adapters in the async world? For example I see the use of globals here and it makes me wonder:

* Why are globals needed here? For example what function will pick up these globals later?
* What prevents these globals from being set again before the "other side" picks up the results?

I suspect there's answers to these and it's where I think some docs could help. I realize as well that I didn't do a great job of documenting most of this file the first time I wrote it, so I'm also sort of using this opportunity to request some docs while it's still fresh in your head as well.

At a lower-level: why is `param_locals` passed twice here? It feels surprising and I'm not sure the code below was meant to handle the same values in/out, but I suspect you've been testing with this as well. Mind leaving some comments as to why it's ok to do that here? I'm lacking the high-level picture of what these adapters are doing so that might also help piece together why it's ok to pass the two sides in here too.

(also if you're up for it adding docs to preexisting methods like `translate_results` would be most welcome, but I understand if you'd prefer to not or defer that to later)

crates/environ/src/fact/trampoline.rs
```diff
@@ -257,7 +380,294 @@ struct Memory<'a> {
...
+
+        if let Some(globals) = result_globals {
+            self.flush_code();
+            for global in globals {
+                self.module.funcs[self.result]
+                    .body
+                    .push(Body::GlobalSet(*global));
+            }
```
alexcrichton, line 664: Mind adding a helper method like `self.global_set(u32)` which automatically does the `flush_code` plus `Body::GlobalSet`?

alexcrichton (commented):
Ok I'm trying to decipher `trampoline.rs` but if it's ok with you I'd prefer to ask for more documentation first before diving further into these methods. Currently there's very little documentation on how things are set up and understanding enough to be able to review this PR I think would require cross-referencing both all the details in the spec along with Wasmtime internal implementation details. A lot of this is also pretty specific to Wasmtime itself in the sense that it's all internal adapter details and we're implementing halves of the spec in some places.

Would you be ok writing up some docs for this? Ideally I'd find it most helpful to have a high-level description of how the adapters piece together and what the expected flow between them is. My hope is that such documentation would also be pretty valuable for future readers to understand this as well.

crates/cranelift/src/compiler/component.rs
dicej: I just pushed an update which uses https://github.com/bytecodealliance/wasm-tools/pull/1989 and makes this code obsolete.  Now we compare `TypeTupleIndex`es, so no need to intern core function types anymore.

crates/cranelift/src/compiler/component.rs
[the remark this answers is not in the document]
dicej: https://github.com/bytecodealliance/wasmtime/issues/10143

crates/environ/src/fact/trampoline.rs
dicej: I just pushed an update that adds some comments.  Let me know if anything needs to be clarified or added.

crates/environ/src/fact/trampoline.rs
dicej: Yeah, it's confusing and needs docs; I'll add them.

> Why are globals needed here? For example what function will pick up these globals later?

Quoting the comment I just pushed:

```
            // Like the async->async case above, for the sync->async case we
            // also need `async-start` and `async-return` helper functions to
            // allow the callee to asynchronously "pull" the parameters and
            // "push" the results when it is ready.
            //
            // However, since the caller is using the synchronous ABI, the
            // parameters may have been passed via the stack rather than linear
            // memory.  In that case, we use global variables to store them such
            // that they can be retrieved by the `async-start` function.
            // Similarly, the `async-return` function may write its result to
            // global variables from which the adapter function can read and
            // return them via the stack to the caller.
```

> What prevents these globals from being set again before the "other side" picks up the results?

Since these globals are only used for sync-lowered imports, and the caller instance can only call one of those at a time, we know nobody will touch them.

> At a lower-level: why is param_locals passed twice here?

Hmm, I think you found a bug (and a hole it the test coverage).  I think my _intention_ was that `translate_results` would use the last element of its `param_locals` parameter as the destination pointer in the case where the result does not fit on the stack.  But that doesn't make sense because it's a pointer to the callee's memory -- it only makes sense as a source pointer.  I think I was just confused, and the code is wrong.

I added tests to tests/all/component_model/import.rs to cover the various combinations of passing parameters and returning results via the stack and the heap, but am now realizing I'm missing component composition tests for those scenarios.

Also now realizing that testing async-lifted exports that return more than MAX_FLAT_RESULTS is not sufficient.  Since `task.return` accepts up to `MAX_FLAT_PARAMS`, I need to exceed that also.

I'll add those tests and fix any issues I find.

dicej: > > What prevents these globals from being set again before the "other side" picks up the results?
> 
> Since these globals are only used for sync-lowered imports, and the caller instance can only call one of those at a time, we know nobody will touch them.

Actually, I'm going to take this back.  For sync-lowered imports called from an async-without-callback-lifted export, we _could_ have more than one running concurrently, in which case we'd have a problem.  The good news is that we've decided not to include the async-without-callback ABI in WASIp3, so I don't need to solve that right away, but we do need to make sure we reject such exports until we've solved it.

This shouldn't be a problem for sync-lowered imports called from sync-lifted exports or async-with-callback-lifted exports since neither can be reentered until the import returns.

crates/environ/src/fact/trampoline.rs
dicej: Are you concerned at all about the cost of calling `flush_code` for each `GlobalSet` rather than once for all of them?

dicej: Oh, I see -- it doesn't matter because pushing the `GlobalSet`s won't add anything to `self.code`, so `flush_code` will be a no-op anyway.

crates/environ/src/fact/trampoline.rs
dicej: FYI, I'm currently tackling the arduous but inevitable task of rebasing my `async` branch onto this one.  Tons of conflicts, as expected.  Once I've finished that, I'll make sure all the existing tests pass, then add new tests and docs per the above discussion, and finally copy any relevant changes/fixes back into this PR.

crates/environ/src/fact/trampoline.rs
```diff
@@ -80,50 +84,192 @@ struct Compiler<'a, 'b> {
...
+            // e.g. resource enter and exit calls as needed.
+            let (compiler, lower_sig, lift_sig) = compiler(module, adapter);
+            compiler.compile_sync_to_sync_adapter(adapter, &lower_sig, &lift_sig)
+        }
         (true, true) => {
-            todo!()
+            // In the async->async case, we must compile a couple of helper functions:
+            //
+            // - `async-start`: copies the parameters from the caller to the callee
+            // - `async-return`: copies the result from the callee to the caller
```
alexcrichton, line 162: One question I'm left with after reading these docs is what's done with async-{start,return}, e.g. who's calling them. Definitely feel free to say "go read the CM docs Alex" for this, but it's at least to me still a big missing piece of the puzzle in understanding these trampolines.

crates/environ/src/fact/trampoline.rs
```diff
@@ -39,6 +40,9 @@ use wasmtime_component_util::{DiscriminantSize, FlagsSize};
 const MAX_STRING_BYTE_LENGTH: u32 = 1 << 31;
 const UTF16_TAG: u32 = 1 << 31;
 
+const EXIT_FLAG_ASYNC_CALLER: i32 = 1 << 0;
+const EXIT_FLAG_ASYNC_CALLEE: i32 = 1 << 1;
```
alexcrichton, line 44: These constants seem like they might be used by the host runtime as well, is that right? If so should they be `pub` and somewhere accessible from the outside of this crate?

crates/environ/src/fact/trampoline.rs
```diff
@@ -256,8 +402,282 @@ struct Memory<'a> {
...
+        self.instruction(Call(enter.as_u32()));
+
+        // TODO: As an optimization, consider checking the backpressure flag on the callee instance and, if it's
+        // unset _and_ the callee uses a callback, translate the params and call the callee function directly here
+        // (and make sure `exit` knows _not_ to call it in that case).
+
+        self.module.exports.push((
+            adapter.callee.as_u32(),
+            format!("[adapter-callee]{}", adapter.name),
+        ));
```
alexcrichton, line 456: This in theory can move to a "declare" `elem` segment, right? 

If no, mind commenting why it's exported?

If so, feel free to leave this here but can you add a comment/TODO saying it should become an element segment?

crates/environ/src/fact/trampoline.rs
```diff
@@ -256,8 +402,282 @@ struct Memory<'a> {
...
+        adapter: &AdapterData,
+        start: FunctionId,
+        return_: FunctionId,
+        param_count: i32,
+    ) {
+        let enter = self.module.import_async_enter_call();
+        let exit = self
+            .module
+            .import_async_exit_call(adapter.lift.options.callback, None);
+
```
alexcrichton, line 431: Mind adding a comment here as a small recap of what these intrinsics are doing? For example it looks a little odd to immediately call one then the other but I presume the host is doing something as part of enter/exit which may continue to execute wasm code?

crates/environ/src/fact/trampoline.rs
```diff
@@ -256,8 +402,282 @@ struct Memory<'a> {
...
+            .body
+            .push(Body::RefFunc(return_));
+        self.instruction(I32Const(
+            i32::try_from(adapter.lower.instance.as_u32()).unwrap(),
+        ));
+        self.instruction(I32Const(
+            i32::try_from(self.types[adapter.lift.ty].params.as_u32()).unwrap(),
+        ));
+        self.instruction(LocalGet(0));
+        self.instruction(LocalGet(1));
```
alexcrichton, line 446: Mind leaving comments as to what these two locals are?

crates/environ/src/fact/trampoline.rs
```diff
@@ -256,8 +402,282 @@ struct Memory<'a> {
...
+
+        self.instruction(I32Const(
+            i32::try_from(adapter.lower.instance.as_u32()).unwrap(),
+        ));
+        self.instruction(RefFunc(adapter.callee.as_u32()));
+        self.instruction(I32Const(
+            i32::try_from(adapter.lift.instance.as_u32()).unwrap(),
+        ));
+        self.instruction(I32Const(param_count));
+        self.instruction(I32Const(1)); // leave room for the guest context result
```
alexcrichton, line 466: Mind leaving a comment as to what these are doing? For example why nparams is being passed and expanding a bit on what it means to leave room for the guest context?

crates/environ/src/fact/trampoline.rs
```diff
@@ -256,8 +402,282 @@ struct Memory<'a> {
...
+        self.instruction(I32Const(EXIT_FLAG_ASYNC_CALLEE));
+        self.instruction(Call(exit.as_u32()));
+        self.instruction(Drop);
+
+        if let Some(globals) = result_globals {
+            for global in globals {
+                self.global_set(*global);
+            }
+        }
```
alexcrichton, line 549: I fear I'm personally still pretty lost here. In a sync->async call what I was expecting is that the results would be stored in globals and async-start would read from those globals when called by somone (the host? unsure). The results would then be stored-to in async-return. Here I was expecting something to block (as the caller is sync) until the nested async call was finished and then the globals would be read.

That doesn't appear to be what's happening though? I commented above that's it's not clear what async-start/async-return are doing but the setting of globals here rather than reading is what really threw me astray... Mind leaving more comments as to what the host intrinsics are doing and the various protocol that this is adhering to?

crates/environ/src/fact/trampoline.rs
```diff
@@ -256,8 +402,282 @@ struct Memory<'a> {
...
+        ));
+        self.instruction(I32Const(param_count));
+        self.instruction(I32Const(result_count));
+        self.instruction(I32Const(EXIT_FLAG_ASYNC_CALLER));
+        self.instruction(Call(exit.as_u32()));
+
+        self.finish()
+    }
+
+    fn compile_async_start_adapter(
```
alexcrichton, line 602: I'm personally pretty nervous about the use of `global`s here because it seems like it'll be easy to get wires crossed by accident and overwrite globals or use uninitialized globals. To help alleviate this concern though, what do you think about some debug-only code?

In wasmtime-cranelift we perform some extra debug checks in `cfg!(debug_assertions)` in the generated code itself. I'm wondering if we could do something similar here where in `debug_assertions` mode there's an "in_use" global which is asserted to be zero when we store to it and then it's asserted to be 1 when we read from it. (and stores switch from 0->1 where reads switch from 1->0). That'd help make me more confident that although I don't fully understand everything here we have at least a layer of defense against bugs in debug mode for testing/fuzzing.

crates/environ/src/fact/trampoline.rs
```diff
@@ -256,8 +402,282 @@ struct Memory<'a> {
...
+                } else {
+                    None
+                })
+                .collect::<Vec<_>>()
+        } else {
+            sig.params
+                .iter()
+                .enumerate()
+                .map(|(i, ty)| (i as u32, *ty))
+                .collect::<Vec<_>>()
```
alexcrichton, line 637: Ok I'll admit I'm still very lost as to what's going on here. My current understanding of async-start is that an async task calls it to get the parameters of the original async call. Here it looks like the adapter is performing the "translate from one component to another call", but I'm pretty confused why this signature's locals would ever be used. Why would a parameter be passed in if you're trying to learn the original parameters of the initial caller?

The globals bits make sense to me where the initial caller put some stuff in globals and that's read here, but I'm not sure how this works without the globals?

alexcrichton (commented):
I'm sorry I know I sound like a broken record but I'm personally still very lost trying to understand this. On one hand one way I can fix this is to go study the component-model specification, try to piece together what an expected implementation would be, and then try to connect those dots back to this implementation. On the other hand though I also think it'd be valuable to have enough local documentation to not require that because although I can do that it would also be required for any future readers as well.

If you feel like I'm requesting too much documentation though or there's something that is well outside the purview of this module's documentation and it's more basic fundamental understanding please let me know though. I don't think we should just mirror the entire specification in comments into this repository, but at the same time I still think there are critical implementation pieces lacking comments such as the protocol between the async-start/return adapters as well as the host-provided async-enter/exit functions.

dicej:
To be clear: I absolutely agree that more docs and comments are needed -- I've only added a bit of that so far but plan to add more.  Feel free to ignore this PR until that's done.

crates/environ/src/fact/signature.rs
```diff
@@ -59,7 +93,102 @@ impl ComponentTypesBuilder {
...
+    pub(super) fn async_start_signature(&self, options: &AdapterOptions) -> Signature {
+        let ty = &self[options.ty];
+        let ptr_ty = options.options.ptr();
+
+        let mut params = vec![ptr_ty];
+
+        let mut results_indirect = false;
+        let results = match self.flatten_types(
+            &options.options,
+            // Async functions return results by calling `task.return`, which
```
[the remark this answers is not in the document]
dicej: I've reworked this code and added clarifying comments.

crates/environ/src/fact/trampoline.rs
```diff
@@ -80,50 +84,192 @@ struct Compiler<'a, 'b> {
...
+                let mut counts = [0; 4];
+                Some(
+                    lower_sig
+                        .params
+                        .iter()
+                        .take(if lower_sig.results_indirect {
+                            lower_sig.params.len() - 1
+                        } else {
+                            lower_sig.params.len()
+                        })
```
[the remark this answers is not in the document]
dicej: I've refactored the code to no longer use global variables to stash params and results, so this code no longer exists.

crates/environ/src/fact/trampoline.rs
```diff
@@ -80,50 +84,192 @@ struct Compiler<'a, 'b> {
...
+                        .map(|ty| module.allocate_global(&mut counts, *ty))
+                        .collect::<Vec<_>>(),
+                )
+            };
+            let result_globals = if lower_sig.results_indirect {
+                None
+            } else {
+                let mut counts = [0; 4];
```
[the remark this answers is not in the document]
dicej: This code no longer exists, so marking this resolved.

crates/environ/src/fact/trampoline.rs
dicej: TIL about declarative element segments.  I've added a TODO comment.

crates/environ/src/fact/trampoline.rs
dicej: I've refactored this to avoid using globals, and have added explanatory comments to the new code.

crates/environ/src/fact/trampoline.rs
dicej: I've added doc comments to `fact.rs` to describe the purpose of these intrinsics.  Let me know if they need more work.

crates/environ/src/fact/trampoline.rs
dicej: The globals are gone 🎉

crates/environ/src/fact/trampoline.rs
dicej: The globals are gone, and I've added a bunch of comments; let me know if further clarification is needed.

dicej:
@alexcrichton I've rebooted this PR based on the latest code in the `wasip3-prototyping` repo.  There are lot more comments in the fused adapter code now; let me know if more are needed or the existing ones are unclear.

This PR now includes two more tests (`streams.wast` and `futures.wast`) and the `wasmtime-cranelift` and `wasmtime-environ` intrinsic plumbing needed to make them pass.  All the intrinsics follow the same pattern, and the Cranelift codegen for them share a lot of code, so I figure putting them all in a single PR might be easiest.  Let me know if it's too much.

crates/cranelift/src/compiler/component.rs
```diff
@@ -262,7 +328,337 @@ impl<'a> TrampolineCompiler<'a> {
...
+
+    fn translate_intrinsic_libcall(
+        &mut self,
+        vmctx: ir::Value,
+        get_libcall: fn(
+            &dyn TargetIsa,
+            &mut ir::Function,
+        ) -> (ir::SigRef, ComponentBuiltinFunctionIndex),
+        args: &[ir::Value],
+        result: ir::types::Type,
```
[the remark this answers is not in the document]
alexcrichton: Or, somewhat alternatively, I'm a bit uneasy having the return type imply the meaning of the value of the return in terms of whether it traps or not. Another option would be to thread around `TrapSentinel` perhaps? (or something akin to that)

crates/cranelift/src/compiler/component.rs
```diff
@@ -94,166 +96,230 @@ impl<'a> TrampolineCompiler<'a> {
...
         }
     }
 
-    fn translate_lower_import(
-        &mut self,
-        index: LoweredIndex,
-        options: &CanonicalOptions,
-        lower_ty: TypeFuncIndex,
-    ) {
+    fn flat_stream_element_info(&self, ty: TypeStreamTableIndex) -> Option<CanonicalAbiInfo> {
```
alexcrichton, line 281: Mind adding a comment here clarifying that this is just an optimization and it's always valid to return `None` here?

crates/cranelift/src/compiler/component.rs
```diff
@@ -262,7 +328,337 @@ impl<'a> TrampolineCompiler<'a> {
...
+        }
+
+        let args = self.builder.func.dfg.block_params(self.block0).to_vec();
+        let vmctx = args[0];
+
+        let pointer_type = self.isa.pointer_type();
+        let wasm_func_ty = &self.types[self.signature].unwrap_func();
+
+        let param_offset = 5;
+        let spill_offset = param_offset + 2;
```
alexcrichton, line 412: If you can comment pointing to a location in the fact trampoline bits to where these constants come from (and vice versa ideally) that'd be good to keep the two in sync. (and/or extracting this to constants or similar)

crates/cranelift/src/compiler/component.rs
```diff
@@ -262,7 +328,337 @@ impl<'a> TrampolineCompiler<'a> {
...
+            if let Some(post_return) = post_return {
+                callee_args.push(self.builder.ins().load(
+                    pointer_type,
+                    MemFlags::trusted(),
+                    vmctx,
+                    i32::try_from(self.offsets.runtime_post_return(post_return)).unwrap(),
+                ));
+            } else {
+                callee_args.push(self.builder.ins().iconst(pointer_type, 0));
+            }
```
alexcrichton, line 555: Similar to down below handling these canonical ABI arguments I think would be good to have as helper functions

alexcrichton (approved):
Thanks for adding these comments!

dicej:
Thanks for the review, @alexcrichton!  I'll address your feedback first thing tomorrow.

crates/cranelift/src/compiler/component.rs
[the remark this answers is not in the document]
alexcrichton: You'll want `value_type` the method, like this -- https://github.com/bytecodealliance/wasmtime/blob/132a490d72e6381c593a409aeca50a765b54ca28/crates/cranelift/src/translate/table.rs#L66 (there's a number of hits you can explore throughout `crates/cranelift` too, this is just a random one)

crates/cranelift/src/compiler/component.rs
```diff
@@ -94,166 +96,230 @@ impl<'a> TrampolineCompiler<'a> {
...
+                | InterfaceType::S16
+                | InterfaceType::U16
+                | InterfaceType::S32
+                | InterfaceType::U32
+                | InterfaceType::S64
+                | InterfaceType::U64
+                | InterfaceType::Float32
+                | InterfaceType::Float64
+                | InterfaceType::Char),
+            ) => Some(self.types.canonical_abi(&payload).clone()),
```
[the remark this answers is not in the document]
dicej: Yes, I was able to do this, but I ended up needing to clone it at the call site anyway to avoid conflicting borrows of `self` (which perhaps could be resolved via some kind of split borrow approach with heavy refactoring, but doesn't seem worth it in this case).

crates/fuzzing/src/generators/component_types.rs
```diff
@@ -138,7 +157,7 @@ where
         .root()
         .func_wrap(
             IMPORT_FUNCTION,
-            |cx: StoreContextMut<'_, Box<dyn Any>>, params: P| {
+            |cx: StoreContextMut<'_, Box<dyn Any + Send>>, params: P| {
```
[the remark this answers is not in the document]
dicej: I'll revert the changes to this file; they're not needed (yet).

crates/environ/src/fact.rs
```diff
@@ -407,20 +456,198 @@ impl<'a> Module<'a> {
...
+                    .insert(suffix.to_owned(), v)
+                    .is_none())
+            },
+        )
+    }
+
+    /// Import a host built-in function to start a subtask for a sync-lowered
+    /// import call to an async-lifted export.
+    ///
+    /// This call with block until the subtask has produced result(s) via the
```
FrankReh, line 525: "will block"

crates/environ/src/fact/trampoline.rs
```diff
@@ -80,50 +81,176 @@ struct Compiler<'a, 'b> {
...
+                lower_sig.params.len() as u32,
+                emit_resource_call,
+            ),
+            lower_sig,
+            lift_sig,
+        )
+    }
+
+    // This closure compiles a function to be exported to the host which host to
+    // lift the parameters from the caller and lower them to the callee.
```
FrankReh, line 120: to be exported to the host which host to lift the parameters ... maybe correct but I didn't know how to parse that part of the sentence.

crates/environ/src/fact/trampoline.rs
```diff
@@ -256,8 +383,309 @@ struct Memory<'a> {
...
+            .body
+            .push(Body::RefFunc(return_));
+        self.instruction(I32Const(
+            i32::try_from(adapter.lower.instance.as_u32()).unwrap(),
+        ));
+        self.instruction(I32Const(
+            i32::try_from(self.types[adapter.lift.ty].results.as_u32()).unwrap(),
+        ));
+        // Async-lowered imports pass params and receive results via linear
+        // memory, and those pointers are in the the first and second params to
```
FrankReh, line 441: the the

crates/environ/src/fact/trampoline.rs
```diff
@@ -256,8 +383,309 @@ struct Memory<'a> {
...
+        ));
+        self.instruction(I32Const(param_count));
+        self.instruction(I32Const(result_count));
+        self.instruction(I32Const(0));
+        self.instruction(Call(exit.as_u32()));
+
+        self.finish()
+    }
+
+    /// Compiles a function to be exported to the host which host to lift the
```
FrankReh, line 629: exported to the host which host to lift ... apologies if this makes sense to folks more in the know, maybe just hard for me to decipher

PR #10107, opened 2025-01-24
Improve support for completely unknown architectures

alexcrichton:
This commit is a step in the direction of trying to make Wasmtime more portable by default. The goal here is to enable Wasmtime to compile for architectures that it has no prior knowledge of. There's a few miscellaneous locations through Wasmtime where we need architecture-specific intrinsics and such but that's all in service of Cranelift itself. Without Cranelift support none of them are necessary.

This commit plumbs a custom `#[cfg]` from Wasmtime's `build.rs` script into the crate about whether there's a supported Cranelift backend. If this isn't available some architecture-specific intrinsics are turned off and not included. An example is that `vm::arch` entirely disappears which is only in service of `UnwindHost`, which also disappears. Furthermore the `helpers.c` file also entirely disappears as it's not necessary on unknown architectures.

To help keep this working I've added CI to build Wasmtime for `powerpc64le-unknown-linux-gnu`. Wasmtime currently has no support for this architecture, although if it grows such support in the future this'll need to be changed to some other unsupported architecture.


crates/wasmtime/src/runtime/func.rs
```diff
@@ -1646,7 +1646,19 @@ fn enter_wasm<T>(store: &mut StoreContextMut<'_, T>) -> Option<usize> {
         return None;
     }
 
+    // When Cranelift has support for the host then we might be running native
+    // compiled code meaning we need to read the actual stack pointer. If
+    // Cranelift can't be used though then we're guaranteed to be running pulley
+    // in which case this stack poitner isn't actually used as Pulley has custom
+    // mechanisms for stack overflow.
+    #[cfg(has_cranelift_host_backend)]
```
fitzgen, line 1654: Can we rename this `has_host_compiler_backend`? While we don't have any target supported on Winch but not Cranelift now, it doesn't seem like it would be impossible in the future, so we should just try to be a little more precise about what we are saying.

fitzgen (approved):

alexcrichton:
I probably should have done `prtest:full` sooner but as can probably be seen there's a lot of fixup commits here to get tests passing on all platforms in CI and such. Nontrivial changes include:

* Validation of `Tunables` for execution at runtime is deferred until a module is loaded. This enables 32-bit platforms, which disable `signals_based_traps`, to compile for 64-bit platforms (e.g. the `disas` test suite)
* Auto-configuration of `Tunables` based on crate features is skipped if `--target` is passed. This fixes `disas` cross-compilation tests from 32-to-64-bit where 64-bit hosts expect `signals-based-traps`.
* Some standalone `tests/*.rs` files have moved from `/tests/*.rs` to `/crates/wasmtime/tests/*.rs` to be able to use the `#[cfg]` directives printed by `crates/wasmtime/build.rs`
* Default Tunables for Pulley targets have been auto-adjusted to assume no signals-based-traps and no memory reservation. This fixes `--target pulley32` on 32-bit platforms where the second bullet above wasn't otherwise applying to auto-disable from the defaults.
* Stack overflow messages for overflowing the async stack no longer happen with Pulley. That's because we no longer (intentionally) install signal handlers.

Juggling compile options I feel still isn't great. Ideally everything "just works" but it's a narrow needle to thread trying to juggle different crate features and targetting different environments. Ideally we'd be able to query `has_native_signals` and `has_virtual_memory` based on `target_lexicon::Triple`, but that involves inferring `cfg(unix)` or `cfg(windows)` which is a bit of a stretch at this time. I've attempted to have at least a somewhat reasonable middle-ground in the meantime.

PR #10108, opened 2025-01-24
Disallow dead code/unused imports by default in `wasmtime` crate

alexcrichton:
This PR is a change to the crate attributes of the `wasmtime` crate. Quite some time ago when Wasmtime had far fewer crate features I added a directive to allow dead code and unused imports when the `default` feature was disabled. The original intention was to not worry too much about unused imports and dead code in every possible configuration of Wasmtime to ease the burden of dealing with unused imports or compile-time conditionally used code.

The primary consequence of this change though is that `cargo build` doesn't actually warn about dead code or unused imports in this repository during development. The reason for this is that the `default` feature is disabled by default and never actually enabled by any dependency inside the repository. Instead each crate, like `wasmtime-cli`, has its own `default` feature.

If I were to start from the problem of "I want dead code warnings by default in this repository" there's two possible fixes:

1. Change `wasmtime-cli`'s `default` feature to include `wasmtime/default`.
2. Remove this `#[allow]` directive.

While (1) would indeed work this is an attempt to get (2) working to see how bad the changes are necessary within the crate.

prtest:full

### DIFF
--- a/crates/wasmtime/src/lib.rs
+++ b/crates/wasmtime/src/lib.rs
@@ -281,7 +281,6 @@
 #![doc(test(attr(deny(warnings))))]
 #![doc(test(attr(allow(dead_code, unused_variables, unused_mut))))]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
-#![cfg_attr(not(feature = "default"), allow(dead_code, unused_imports))]
 // Allow broken links when the default features is disabled because most of our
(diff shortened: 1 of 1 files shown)


alexcrichton:
Oof that's a lot of red

alexcrichton:
Moving to https://github.com/bytecodealliance/wasmtime/pull/10131

PR #10109, opened 2025-01-24
Winch: `v128` logical ops for x64

MarinPostma:
implement `v128` simd logical operations:

- `v128.not`
- `v128.and`
- `v128.andnot`
- `v128.or`
- `v128.xor`
- `v128.bitselect`
- `v128.any_true`
- `v128.load8_lane`
- `v128.load16_lane`
- `v128.load32_lane`
- `v128.load64_lane`
- `v128.store8_lane`
- `v128.store16_lane`
- `v128.store32_lane`
- `v128.store64_lane`

https://github.com/bytecodealliance/wasmtime/issues/8093


alexcrichton:
Moving my review over to @saulecabrera who knows this better than I

winch/codegen/src/codegen/mod.rs
```diff
@@ -1417,6 +1417,30 @@ where
...
+        let src = self.context.pop_to_reg(self.masm, None)?;
+        if let Some(addr) = self.emit_compute_heap_address(&arg, size)? {
+            let dst = self.masm.address_at_reg(addr, 0)?;
+            self.masm.store_lane(src.reg, dst, lane, size)?;
+            self.context.free_reg(addr);
+            self.context.free_reg(src);
+        }
+
+        Ok(())
+    }
```
saulecabrera, line 1443: Instead of introducing new load/stored methods, I'd suggest calling `wasm_load` and `wasm_store`. This approach would probably require modifying the existing `LoadKind` definition in case the current definition can't represent the semantics of lane loading/storing. That's the approach we've followed for other vector loads (e.g., `load_splat`)

1 main reason:
- Wasm loads/stores are critical, from the sandboxing perspective, and we're trying to keep the implementation as tight as possible, reusing the existing methods makes auditing them and maintaining them much easier.

winch/codegen/src/masm.rs
```diff
@@ -1482,4 +1482,52 @@ pub(crate) trait MacroAssembler {
...
+    fn xor128v(&mut self, src1: Reg, src2: Reg, dst: WritableReg) -> Result<()>;
+
+    /// Given two 128bits vectors `src1` and `src2`, and a 128bits bitmask `mask`, selects bits
+    /// from `src1` when mask is 1, and from `src2` when mask is 0.
+    ///
+    /// This is equivalent to: `v128.or(v128.and(src1, mask), v128.and(src2, v128.not(mask)))`.
+    fn bitselect128v(&mut self, src1: Reg, src2: Reg, mask: Reg, dst: WritableReg) -> Result<()>;
+
+    /// If any bit in `src` is 1, set `dst` to 1, or 0 otherwise.
+    fn any_true128v(&mut self, src: Reg, dst: WritableReg) -> Result<()>;
```
saulecabrera, line 1514: A small remark on naming: to be consistent with the rest of the naming in the MacroAssembler for type-specific instructions, perhaps we could consider prefixing all these methods with `v128_`? e.g., `v128_or`, similar to float operations, e.g., `float_abs`

winch/codegen/src/isa/x64/masm.rs
```diff
@@ -1640,6 +1638,78 @@ impl Masm for MacroAssembler {
...
+        self.or128v(dst.to_reg(), tmp, dst)?;
+
+        Ok(())
+    }
+
+    fn any_true128v(&mut self, src: Reg, dst: WritableReg) -> Result<()> {
+        self.asm.xmm_vptest(src, src);
+        self.asm.setcc(IntCmpKind::Ne, dst);
+        Ok(())
+    }
```
saulecabrera, line 1686: We need to check that the host/target supports AVX for all of these methods: 

```rust
if !self.flags.has_avx() {
  bail!(CodeGenError::UnimplementedForNoAvx);
}
```

MarinPostma:
@saulecabrera it should be good now. I have take the opportunity to tighten store/loads even more regarding atomics operation. Having the `MemOpKind` as a separate arg meant that we could declare many opeartion as atomics, but that would have no effect.

saulecabrera:
There's a conflict, once resolved, we can land this one.

MarinPostma:
done

alexcrichton:
it's a bit hidden (alas) but I think that failure was non-spurious: https://github.com/bytecodealliance/wasmtime/actions/runs/13016537423/job/36307106911

saulecabrera:
Yeah, doesn't seem to be spurious. You need to update the spec tests here https://github.com/bytecodealliance/wasmtime/blob/main/crates/wast-util/src/lib.rs#L492 to include the ones that you enabled, so that it's marked as expected that these tests should fail on architectures that don't support AVX+. I believe this is mostly for for darwin x86_64, which runs via Rosetta and therefore there's no AVX support.

MarinPostma:
I think I have fixed it, but I don't have a mac to try it right now. Thanks!

winch/codegen/src/isa/x64/masm.rs
```diff
@@ -368,7 +342,14 @@ impl Masm for MacroAssembler {
                         .xmm_vpbroadcast_mr(&src, dst, size, UNTRUSTED_FLAGS);
                 }
             }
+            LoadKind::VectorLane(LaneSelector { lane, size }) => {
+                let byte_tmp = regs::scratch();
```
saulecabrera, line 346: The failures that CI are related to the `vpinsr*`instruction, which are emitted here. We need to add `self.ensure_has_avx()?` to get CI green.

MarinPostma: I setup the action on my fork, all tests pass now. worry for the back and forth :sweat:

Issue #41, opened 2019-01-12
Missing stdbool.h on `cargo build`


buckle2000:
```
$ RUST_BACKTRACE=1 cargo build
```

Error message: 

```
   Compiling wasmtime-runtime v0.1.0 (/tmp/wasmtime/lib/runtime)
error: failed to run custom build command for `wasmtime-runtime v0.1.0 (/tmp/wasmtime/lib/runtime)`
process didn't exit successfully: `/tmp/wasmtime/target/debug/build/wasmtime-runtime-d60ae619fe4012af/build-script-build` (exit code: 101)
--- stdout
running: "cmake" "/tmp/wasmtime/lib/runtime/signalhandlers" "-DCMAKE_INSTALL_PREFIX=/tmp/wasmtime/target/debug/build/wasmtime-runtime-e2adaa410bffafe9/out" "-DCMAKE_C_FLAGS= -ffunction-sections -fdata-sections -fPIC -m64" "-DCMAKE_C_COMPILER=/usr/bin/cc" "-DCMAKE_CXX_FLAGS= -ffunction-sections -fdata-sections -fPIC -m64" "-DCMAKE_CXX_COMPILER=/usr/bin/c++" "-DCMAKE_BUILD_TYPE=Debug"
-- Configuring done
-- Generating done
-- Build files have been written to: /tmp/wasmtime/target/debug/build/wasmtime-runtime-e2adaa410bffafe9/out/build
running: "cmake" "--build" "." "--target" "install" "--config" "Debug" "--"
[100%] Built target SignalHandlers
Install the project...
-- Install configuration: "Debug"
-- Up-to-date: /tmp/wasmtime/target/debug/build/wasmtime-runtime-e2adaa410bffafe9/out/./libSignalHandlers.a
cargo:root=/tmp/wasmtime/target/debug/build/wasmtime-runtime-e2adaa410bffafe9/out
cargo:rustc-link-search=native=/tmp/wasmtime/target/debug/build/wasmtime-runtime-e2adaa410bffafe9/out
cargo:rustc-link-lib=static=SignalHandlers

--- stderr
signalhandlers/SignalHandlers.h:7:10: fatal error: 'stdbool.h' file not found
signalhandlers/SignalHandlers.h:7:10: fatal error: 'stdbool.h' file not found, err: true
thread 'main' panicked at 'Unable to generate bindings: ()', src/libcore/result.rs:1009:5
[...]

PR #10110, opened 2025-01-24
asm: introduce a new x64 assembler

abrown:
This is a first step to providing an external assembler for `cranelift-codegen` as described in https://github.com/bytecodealliance/wasmtime/issues/41. Each commit has further details, but the summary is that this direction would eventually move all assembler logic out into crates designed for easier checking (e.g., fuzzing).


cranelift/codegen/src/isa/x64/inst/emit.rs
```diff
@@ -4693,6 +4692,19 @@ pub(crate) fn emit(
...
+        Inst::External { inst } => {
+            let mut known_offsets = [0, 0];
+            // These values are transcribed from is happening in
+            // `SyntheticAmode::finalize`. This, plus the `Into` logic converting a
+            // `SyntheticAmode` to its external counterpart, are
+            let frame = state.frame_layout();
+            known_offsets[external::offsets::KEY_INCOMING_ARG] =
+                i32::try_from(frame.tail_args_size + frame.setup_area_size).unwrap();
+            known_offsets[external::offsets::KEY_SLOT_OFFSET] =
+                i32::try_from(frame.outgoing_args_size).unwrap();
```
abrown, line 4705: @cfallin, I'm not a big fan of this `KnownOffsetTable` approach: it seems like we may want to just add appropriate `CodeSink`/`MachBuffer` methods to propagate this kind of thing.

cranelift/filetests/filetests/isa/x64/simd-arith-avx.clif
```diff
@@ -1648,8 +1646,6 @@ block0(v0: i8x16, v1: i32):
 ;   addb %al, (%rax)
 ;   addb %al, (%rax)
 ;   addb %al, (%rax)
-;   addb %al, (%rax)
-;   addb %bh, %bh
```
abrown, line 1652: Looking at this CLIF diffs, I see some potential problems: removed instructions, different immediates, potential `and` -> `or` miscompilations... This bears another look.

abrown: (Pretty sure those `add` -> `or` "miscompilations" were just bits changing in the constant pool... not sure why they're changing, though).

cranelift/assembler-x64/meta/src/generate/features.rs
```diff
@@ -0,0 +1,29 @@
...
+        generate_derive(f);
+        fmtln!(f, "pub enum Feature {{");
+        f.indent(|f| {
+            // N.B.: it is critical that this list contains _all_ variants of
+            // the `Flag` enumeration here at the `meta` level so that we can
+            // accurately transcribe them to a structure available in the
+            // generated layer above. If this list is incomplete, we will
+            // (fortunately) see compile errors for generated functions that use
+            // the missing variants.
+            const ALL: &[dsl::Feature] = &[_64b, compat];
```
cfallin, line 22: Could we put this const slice alongside the `Feature` enum definition so it's easier to update/keep in sync?

cranelift/assembler-x64/meta/src/generate/format.rs
```diff
@@ -0,0 +1,193 @@
...
+//! Rust code.
+
+use super::{fmtln, Formatter};
+use crate::dsl;
+
+impl dsl::Format {
+    /// Re-order the Intel-style operand order to accommodate ATT-style
+    /// printing.
+    #[must_use]
+    pub fn generate_att_style_operands(&self) -> String {
```
cfallin, line 11: Oh no! I think we all prefer Intel syntax and are kind of stuck with AT&T only because of legacy/inertia. It might be worth listing here or somewhere else why we're doing this for now and what the plan is: e.g. will we remove this swap and update all the compile test expectations once we've migrated all instructions to the new DSL?

cranelift/assembler-x64/meta/src/generate/format.rs
```diff
@@ -0,0 +1,193 @@
...
+    fn generate_legacy_prefix(&self, f: &mut Formatter, rex: &dsl::Rex) {
+        use dsl::LegacyPrefixes::*;
+        if rex.prefixes != NoPrefix {
+            f.empty_line();
+            f.comment("Emit legacy prefixes.");
+            match rex.prefixes {
+                NoPrefix => unreachable!(),
+                _66 => fmtln!(f, "buf.put1(0x66);"),
+                _F0 => fmtln!(f, "buf.put1(0xf0);"),
+                _66F0 => fmtln!(f, "buf.put1(0x66); buf.put1(0xf0);"),
```
cfallin, line 43: minor thing but could we put these on separate lines?

cranelift/assembler-x64/meta/src/generate/formatter.rs
```diff
@@ -0,0 +1,137 @@
+//! Logic for writing generated code to a file.
+//!
+//! Use [`fmtln`] by default since it will append Rust comments with the original source location.
+//! For more control, use the [`Formatter`] directly. The [`Formatter`] logic has been borrowed
+//! extensively from `cranelift/codegen/meta/src/srcgen.rs`.
```
cfallin, line 5: You've probably thought about this already but just to check: is there any way we could share this code directly? E.g. depend on `cranelift-codegen-meta`, or factor out another crate (`cranelift/meta-utils`), or something like that?

cranelift/assembler-x64/meta/src/instructions/and.rs
```diff
@@ -0,0 +1,29 @@
...
+use crate::dsl::{Feature::*, Inst, LegacyPrefixes::*, Location::*};
+
+pub fn list() -> Vec<Inst> {
+    vec![
+        inst("andb", fmt("I", [rw(al), r(imm8)]), rex(0x24).ib(), _64b | compat),
+        inst("andw", fmt("I", [rw(ax), r(imm16)]), rex(0x25).prefix(_66).iw(), _64b | compat),
+        inst("andl", fmt("I", [rw(eax), r(imm32)]), rex(0x25).id(), _64b | compat),
+        inst("andq", fmt("I_SXLQ", [rw(rax), sxq(imm32)]), rex(0x25).w().id(), _64b),
+        inst("andb", fmt("MI", [rw(rm8), r(imm8)]), rex(0x80).digit(4).ib(), _64b | compat),
+        //inst("andb", fmt("MI_W", [rw(rm8), r(imm8)]), rex(0x80).w().digit(4).ib(), _64b),
```
cfallin, line 11: What's the reason for the commented-out variant here -- could we add a comment describing why in the code? Or does it still need to be disabled?

cranelift/assembler-x64/src/api.rs
```diff
@@ -0,0 +1,155 @@
...
+/// When encoding instructions, Cranelift may not know all of the information
+/// needed to construct an immediate. Specifically, addressing modes that
+/// require knowing the size of the tail arguments or outgoing arguments (see
+/// `SyntheticAmode::finalize`) will not know these sizes until emission.
+///
+/// This table allows up to do a "late" look up of these values by their
+/// `KnownOffset`.
+pub trait KnownOffsetTable: Index<KnownOffset, Output = i32> {}
+impl KnownOffsetTable for Vec<i32> {}
+impl KnownOffsetTable for [i32; 2] {}
```
cfallin, line 92: Is this for testing? If so can we add a comment saying so? (The specificity of only 2-elem arrays otherwise is somewhat surprising)

cranelift/assembler-x64/src/api.rs
```diff
@@ -0,0 +1,155 @@
...
+
+/// A `KnownOffset` is a unique identifier for a specific offset known only at
+/// emission time.
+pub type KnownOffset = usize;
```
cfallin, line 96: Can we make this a newtype wrapper (`pub struct KnownOffset(usize)`) instead?

cranelift/assembler-x64/src/imm.rs
```diff
@@ -0,0 +1,209 @@
...
+#![allow(clippy::module_name_repetitions)]
+#![allow(unused_comparisons)] // Necessary to use maybe_print_hex! with `u*` values.
+#![allow(clippy::cast_possible_wrap)] // Necessary to cast to `i*` for sign extension.
+
+use crate::api::{KnownOffset, KnownOffsetTable};
+use arbitrary::Arbitrary;
+
+/// This helper function prints the hexadecimal representation of the immediate
+/// value, but only if the value is greater than or equal to 10. This is
+/// necessary to match how Capstone pretty-prints immediate values.
```
cfallin, line 12: Egads -- is there a way to configure Capstone to always print in decimal or hex? (Necessary evil otherwise, not your fault, just ugly)

cranelift/assembler-x64/src/reg.rs
```diff
@@ -0,0 +1,247 @@
...
+    /// A 64-bit access.
+    Quadword,
+}
+
+/// Like [`Gpr`], but with `%rsp` disallowed.
+///
+/// This is due to avoid special cases of REX encodings, see Intel SDM Vol. 2A,
+/// table 2-5.
+#[derive(Clone, Copy, Debug)]
+pub struct MinusRsp<R: AsReg>(R);
```
cfallin, line 82: `MinusRsp` evokes some sort of subtraction operation in my mind; maybe `NonRspGpr` or something like that?

cranelift/codegen/src/isa/x64/inst/emit.rs
```diff
@@ -4693,6 +4692,19 @@ pub(crate) fn emit(
         Inst::DummyUse { .. } => {
             // Nothing.
         }
+
+        Inst::External { inst } => {
+            let mut known_offsets = [0, 0];
+            // These values are transcribed from is happening in
+            // `SyntheticAmode::finalize`. This, plus the `Into` logic converting a
+            // `SyntheticAmode` to its external counterpart, are
```
cfallin, line 4700: "from is" and ending with "are..." -- incomplete sentence?

cranelift/codegen/src/isa/x64/inst/emit.rs
cfallin: Perhaps! Were you thinking we'd have specific trait methods for e.g. `incoming_arg` and `slot_offset`? Or that we'd still be generic over the set of known offsets but have a query method?

I'm totally fine with the latter, in fact it's probably even cleaner; for the former I'd be concerned about baking details of Cranelift's ABI impl into the lower-level library and would still want to follow the analogy of a real assembler letting one define symbolic constants and use them with specific values plugged in. Happy to look at proposed diff if you have one.

cranelift/codegen/src/isa/x64/pcc.rs
```diff
@@ -980,6 +980,8 @@ pub(crate) fn check(
         Inst::Unwind { .. } | Inst::DummyUse { .. } => Ok(()),
 
         Inst::StackSwitchBasic { .. } => Err(PccError::UnimplementedInst),
+
+        Inst::External { .. } => Ok(()), // TODO: unsure what to do about this!
```
cfallin, line 984: Totally fine to do `Err(PccError::UnimplementedInst)` for now -- I'll need to go and update this once we've moved instructions over that PCC cares about, but PCC isn't fuzzed at the moment anyway until I can make an update/fix pass...

cfallin (approved):
This looks really really good -- thanks for spawning this idea and working to build it out!

Most of my feedback you already got offline as this was being built and the general shape is therefore what I'd expect and am overall quite happy with. A bunch of nits below but nothing major -- let's get this merged and we can iterate and start the migration process.

Thanks again!

cranelift/assembler-x64/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "cranelift-assembler-x64"
+description = "A Cranelift-specific x64 assembler"
+version = "0.117.0"
+license = "Apache-2.0 WITH LLVM-exception"
+edition.workspace = true
```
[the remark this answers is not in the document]
abrown: I don't think any of the Cranelift crates use `version.workspace`; I assume that's because they exist within the overall Wasmtime workspace?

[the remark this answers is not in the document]
cfallin: Ah right, I forgot about the Wasmtime-vs-Cranelift-version distinction; you're right, thanks!

cranelift/assembler-x64/meta/src/dsl/encoding.rs
```diff
@@ -0,0 +1,276 @@
...
+        Self { imm: Imm::id, ..self }
+    }
+
+    #[must_use]
+    pub fn io(self) -> Self {
+        Self { imm: Imm::io, ..self }
+    }
+
+    fn validate(&self, operands: &[Operand]) {
+        assert!(self.digit < 8);
```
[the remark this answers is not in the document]
abrown: This condition is in section 2.1.2 of the "Instruction Format" chapter of volume 2 of, etc., etc.: "an additional 3-bit opcode field is sometimes encoded..." That chapter may be a good start and I'll link it here but, honestly, there's a lot there and it isn't always easy to pull out all the various conditions. I just started with some pretty standard ones here to catch trivial mistakes early; the real safety is in fuzzing against a known-good disassembler.

cranelift/assembler-x64/meta/src/dsl/encoding.rs
```diff
@@ -0,0 +1,276 @@
...
+    }
+}
+
+impl fmt::Display for Rex {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self.prefixes {
+            LegacyPrefixes::NoPrefix => {}
+            LegacyPrefixes::_66 => write!(f, "0x66 + ")?,
+            LegacyPrefixes::_F0 => write!(f, "0xF0 + ")?,
+            LegacyPrefixes::_66F0 => write!(f, "0x66F0 + ")?,
```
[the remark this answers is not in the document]
abrown: There are no hex hints: e.g., `66 0F 54 /r ANDPD xmm1, xmm2/m128`. I'll switch this, retaining the `0x` hints.

cranelift/assembler-x64/meta/src/dsl/encoding.rs
```diff
@@ -0,0 +1,276 @@
...
+        }
+        if self.imm != Imm::None {
+            write!(f, " {}", self.imm)?;
+        }
+        Ok(())
+    }
+}
+
+#[derive(PartialEq)]
+pub enum LegacyPrefixes {
```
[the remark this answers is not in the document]
abrown: Yeah, I just didn't need it at the moment. I would say that it could be completely valid to also expand the opcode width, but let's see what we think when we get to those instructions.

cranelift/assembler-x64/meta/src/dsl/features.rs
```diff
@@ -0,0 +1,83 @@
...
+///
+/// IA-32e mode is the default mode for 64-bit x86 processors. It consists of
+/// two sub-modes:
+/// - __64-bit mode__: uses the full 64-bit address space
+/// - __compatibility mode__: allows use of legacy 32-bit code
+#[derive(Clone, Copy, PartialEq)]
+#[allow(non_camel_case_types)]
+pub enum Feature {
+    _64b,
+    compat,
```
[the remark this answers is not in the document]
abrown: Yeah, true 32-bit support requires a bit more work. Since I had received the request to at least consider it, though, I thought that the first and quite crucial step is to set up instructions to be gated on 32-bit support.

cranelift/codegen/src/isa/x64/inst/emit.rs
abrown: I agree, we should avoid Cranelift-izing this assembler as much as possible. I guess things I don't like about the known offset table approach is that (a) we rebuild this table for every emissions and (b) building it the way I'm doing it here feels quite fragile. How long before things get out of sync with the keys used during amode conversion?!

cranelift/assembler-x64/src/api.rs
abrown: Leaving it as a `type` enables us to just tag on the `Index<KnownOffset, ...>` to the trait and no implementation is needed for vectors and slices. Since I wasn't really in love with this part (see my initial comments above) I took a pretty minimal approach. Let's figure out the big picture for this and then we can decide whether to flesh this out more.

cranelift/assembler-x64/src/imm.rs
abrown: I don't think so... it all seems to be hidden away behind a variant in [`ArchSyntax`](https://docs.rs/capstone/latest/capstone/arch/x86/enum.ArchSyntax.html). I suspect there's more configurability possible with a more x86-specific tool like XED but we're just not ready for that yet. The upside to the fuzzing approach here, though, is that the minute we switch to another disassembler that doesn't have this representation we will see fuzz bugs and be able to remove this code.

cranelift/codegen/src/isa/x64/pcc.rs
abrown: I tried that initially but I see test failures?

```
FAIL filetests/filetests/pcc/succeed/imul-fuzzbug.clif: compile

Caused by:
    Proof-carrying-code validation error: UnimplementedInst
```

To reproduce with the `Err(...)` version: `cd cranelift && cargo test --test filetests`.

abrown: No `and` in that test but it's probably used in lowering some other instruction...

cfallin: Oh, I had forgotten we also had that test-suite still active -- OK, no worries, this is fine (and leave the TODO) and I'll get to it in my update pass!

cranelift/assembler-x64/src/imm.rs
cfallin: Yep, sounds good, it's at least well-tested!

cranelift/assembler-x64/meta/src/generate/formatter.rs
abrown: Well, due to the ISLE bits we need `cranelift-codegen-meta` to depend on `cranelift-assembler-x64-meta`, but I guess we could pull the formatting bits into a third crate they both depend on. I think at this point the `Formatter` and `fmtln!` logic has diverged a bit since on the assembler side I both simplified and then added interesting bits like the "append the location as a comment" logic.

I'm fine with trying to re-synchronize this formatter code into a third crate though I think I should commit to doing that in a future PR to avoid re-working everything here. But another option is to use a pre-existing crate: a casual glance online suggests `quote` + `prettyplease` or even `genco` could be an alternative. I feel like we're in an in-between place where we don't need the full power of something like `quote` but now we would be re-inventing parts of it into a new crate...

cranelift/codegen/src/isa/x64/lower/isle.rs
```diff
@@ -940,6 +954,79 @@ impl Context for IsleContext<'_, '_, MInst, X64Backend> {
...
+    ////////////////////////////////////////////////////////////////////////////
+
+    fn is_imm8(&mut self, src: &GprMemImm) -> Option<AssemblerImm8> {
+        match src.clone().to_reg_mem_imm() {
+            RegMemImm::Imm { simm32 } if simm32 <= u8::MAX as u32 => {
+                Some(AssemblerImm8::new(simm32 as u8))
+            }
+            _ => None,
+        }
+    }
```
alexcrichton, line 969: Question on this: how does this know that an unsigned value is desired? For example if you have `(a: u8) & -2` that means it would fail `is_imm8`, right? But the x64 instruction should be able to encode that?

With the assembler immediates all annotated with if they're sign/zero-extended should there be `Simm8` and `Uimm8` and the extractors here would be `is_simm8` and `is_uimm8`?

(also as a side note I find `u8::try_from(...)` handy in cases of conversion like this so you don't have to do any `as` casts yourself)

cranelift/filetests/filetests/isa/x64/i128.clif
```diff
@@ -440,15 +440,15 @@ block0(v0: i128, v1: i128):
...
-;   andl %r15d, %eax
+;   andb %r9b, %al
+;   andb %r11b, %r8b
+;   andb %r14b, %r10b
+;   andb %bl, %r13b
+;   andb %dil, %r15b
+;   andb %r8b, %al
+;   andb %r13b, %r10b
+;   andb %r10b, %al
+;   andb %r15b, %al
```
alexcrichton, line 451: (sorry if this was already covered elsewhere) but how come these are changing from a 32-bit and to an 8-bit and?

cranelift/filetests/filetests/isa/x64/simd-arith-avx.clif
alexcrichton: This one looks like 3 bytes were shaved off the encoding (the 0x19 => 0x16 in the diff), are the new assembler methods perhaps slightly more optimal in encoding size?

cranelift/assembler-x64/src/api.rs
```diff
@@ -0,0 +1,156 @@
...
+///
+/// This assembler is parameterizable over register types; this allows the
+/// assembler users (e.g., Cranelift) to define their own register types
+/// independent of this crate.
+pub trait Registers {
+    /// An x64 general purpose register that may be read.
+    type ReadGpr: AsReg + for<'a> Arbitrary<'a>;
+
+    /// An x64 general purpose register that may be read and written.
+    type ReadWriteGpr: AsReg + for<'a> Arbitrary<'a>;
```
alexcrichton, line 109: If possible could the `Arbitrary` methods be left off? And the `Arbitrary` impl for each instruction have `where R::ReadGpr: Arbitrary`?

alexcrichton:
(feel free to ignore my comments and respond after landing)

alexcrichton:
Oh one blocking thing if that's ok: this adds a hard dependency of cranelift on capstone which I believe wasn't previously there. Would it be possible to move that to a dev-dependency or a feature'd dependency?

cranelift/assembler-x64/meta/src/instructions/and.rs
abrown: There's a couple of issues that caused us to leave this disabled. There are instructions under the `and` family that, when a `REX` prefix is prepended, end up sign-extending their operands. When this encoding is used, they have limitations on their registers: "In 64-bit mode, r/m8 can not be encoded to access the following byte registers if a REX prefix is used: AH, BH, CH, DH." Since we won't be using the high 8-bit slots here, we can probably disregard that, but that was one of the initial concerns. The second concern is that, in documenting these instructions, it is not clear if there are mistakes: some of these `REX + ...` indicate sign extension for certain instructions but not others (e.g., compare `and`'s `RM` format and `or`'s `RM` format. The third is that, with both Capstone and XED, we don't really see a difference in the disassembled string, so our fuzzing cannot check that we correctly emit one instruction and not the other. Currently thinking about what to do...

alexcrichton:
(I sent a PR for that change as https://github.com/abrown/wasmtime/pull/10)

cranelift/filetests/filetests/isa/x64/simd-arith-avx.clif
abrown: Yeah, that's what I concluded.

cranelift/assembler-x64/meta/src/instructions/and.rs
abrown: I think I need to ask some questions internally about this part of the reference manual. I'll add a TODO here with a comment.

cranelift/filetests/filetests/isa/x64/i128.clif
abrown: I'm a bit mystified: CLIF's `icmp` should return a truthy `Int` which reduces down to an 8-bit value here:

https://github.com/bytecodealliance/wasmtime/blob/a0338af84f66cb452fdf4b692d4facb5d052c12d/cranelift/codegen/meta/src/cdsl/typevar.rs#L433-L435

So those `band` instructions should always have lowered to an `andb`? Perhaps the original encoding just modified a larger width...

cranelift/codegen/src/isa/x64/lower/isle.rs
abrown: Ok, I see what you mean. The assembler doesn't care too much about the sign, it just wants 8 bits, so I chose to use `u8` to pass those along. What I was trying to do here (incorrectly) is check that we can fit whatever was in the `simm32: i32` into those 8 bits.

cranelift/assembler-x64/src/api.rs
alexcrichton: (I pushed up a commit for this over at https://github.com/abrown/wasmtime/pull/10)

cranelift/filetests/filetests/isa/x64/i128.clif
alexcrichton: I talked a bit more about this with @abrown and the reason for this is that Cranelift today [clamps operand sizes to 32-bits](https://github.com/bytecodealliance/wasmtime/blob/a0338af84f66cb452fdf4b692d4facb5d052c12d/cranelift/codegen/src/isa/x64/inst.isle#L2061) in the `alu_rmi_r` constructor. That means that all 8-bit `and`s are actually 32-bit `and`s when emitted to x64. Andrew's PR is (IMO correctly) changing the 8-bit and to use `andb` instead. 

@cfallin do you remember what `operand_size_of_type_32_64` was doing there? For example why clamp it to 32-bit? (and if so, are said reasons still applicable today?)

cranelift/assembler-x64/src/api.rs
abrown: Thanks!

cranelift/filetests/filetests/isa/x64/i128.clif
abrown: Ok, see 2c8a276.

abrown:
This latest [CI failure](https://github.com/bytecodealliance/wasmtime/actions/runs/13124822832/job/36618948362#step:18:987) is under QEMU-emulated s390x. Have we ever observed a segfault here before? I can't imagine this has much to do with the new `cranelift-assembler-x64` crate since that is excluded here...

PR #10112, opened 2025-01-25
pulley: Expand provenance test some more

alexcrichton:
Cover more styles of function calls and a trap or two. This required fixing some issues with libcalls, notably ensuring that `expose_provenance()` is called at some point on the function pointers and additionally ensuring that the signedness of integer arguments were the same. The internal macro previously used `i32` and `i64` but that was actually used everywhere on the host as `u32` and `u64`. There was some confusion in some signatures so this commit fixes things to force everyone to use unsigned integers everywhere to match type we ascribe, the type it's defined with, and how it's called with Pulley.


fitzgen (approved):
Nice

Issue #9667, opened 2024-11-23
wasi-io: Missing `would-block` in `stream-error` (Returning 0 from TcpReader::read)


Heap-Hop:
https://github.com/bytecodealliance/wasmtime/blob/67674881db5fbdbba8594feb52655aaa351a5f77/crates/wasi/src/tcp.rs#L719-L723

Returning `0` for `std::io::ErrorKind::WouldBlock` causes downstream to interpret it as a closed stream, see: 
https://github.com/yoshuawuyts/wstd/issues/25#issuecomment-2493294087.
https://github.com/yoshuawuyts/wstd/blob/5ce367add5e7bcb569b6487453cb9ba94468dc99/src/io/copy.rs#L12

This is also found in:
https://github.com/bytecodealliance/wasmtime/blob/67674881db5fbdbba8594feb52655aaa351a5f77/crates/test-programs/src/bin/preview2_tcp_streams.rs#L17-L22

Given that the `wit` files already include many `would-block` errors, would it make sense to extend `stream-error` to include a `WouldBlock`?

pchickey:
Wasi streams operations are nonblocking, unless they have `blocking_` in their name, so read should always return an empty list instead of blocking. The `blocking_` functions are equivalent to waiting for readiness on the stream's pollable, and then calling the nonblocking variant.

bjorn3:
How can you distinguish in WASI between an EOF stream (no data will be returned in the future because the connection was closed or the end of the file is reached) and a stream where you have to wait until the stream is ready before you can read more data? The distinction is very important for `std::io::copy`/`tokio::io::copy` which needs to continue reading until EOF and only returns when EOF is reached.

pchickey:
When the stream is EOF, it retusn Err(StreamError::Closed) as seen in the above snippet line 21.

Heap-Hop:
Thanks for your kind explanation!

https://github.com/bytecodealliance/wasmtime/blob/67674881db5fbdbba8594feb52655aaa351a5f77/crates/wasi/wit/deps/io/streams.wit#L47-L54

>The pollable given by `subscribe` will be ready when `more bytes` are available. 

I think it would be helpful to update the doc to explicitly note this situation for users.

Heap-Hop:
Additionally, there's a potential **bug** in `HostInputStream::blocking_read`(also affects `blocking_skip`).
The doc states: `it blocks until at least one byte can be read.` 

[...]

PR #10113, opened 2025-01-25
wasmtime-wasi: Fix spurious wake-ups in `blocking_` of `InputStream` and `OutputStream`

Heap-Hop:
issue #9667 

The `InputStream::ready` can be prematurely triggered by `io::ErrorKind::WouldBlock`. This PR ensure that `blocking_` functions return a valid non-empty result.

### DIFF
--- a/crates/wasi-io/src/streams.rs
+++ b/crates/wasi-io/src/streams.rs
@@ -3,6 +3,15 @@ use alloc::boxed::Box;
 use anyhow::Result;
 use bytes::Bytes;
 
+/// `Pollable::ready()` for `InputStream` and `OutputStream` may return
+/// prematurely due to `io::ErrorKind::WouldBlock`.
+///
+/// To ensure that `blocking_` functions return a valid non-empty result,
+/// we use a loop with a maximum iteration limit.
+///
+/// This constant defines the maximum number of loop attempts allowed.
+const MAX_BLOCKING_ATTEMPTS: u8 = 10;
+
 /// Host trait for implementing the `wasi:io/streams.input-stream` resource: A
 /// bytestream which can be read from.
 #[async_trait::async_trait]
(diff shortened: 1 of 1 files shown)


pchickey (approved):

pchickey:
Thanks. As you said in the other thread, there's no real way to test this operation from inside because it blocks. I am fine landing this without a test.

badeend:
The change seems fine indeed.

Aside from `blocking_read`, its counterpart of the writable side (`write_ready`) probably needs similar treatment as well. I think it should loop while `check_write` returns `Ok(0)`.

Heap-Hop:
```rust
        loop {
            // This `ready` call may return early due to `io::ErrorKind::WouldBlock`.
            self.ready().await;
            let data = self.read(size)?;
            if data.is_empty() {
                continue;
            }
            return Ok(data);
        }
```
Can we ensure that there won't be infinite or excessive invalid loops here?

badeend:
As a fail-safe, you could keep track of the number of iterations and return a trap if it exceeds _{some number}_. The reason I suggest to trap as opposed to return an empty buffer, is because after a certain threshold it's more likely to be a bug in the `InputStream` implementation.

Heap-Hop:
https://github.com/bytecodealliance/wasmtime/blob/5dfccc078bb3dcad152d37013dde8067f197e6e7/crates/wasi/src/tcp.rs#L844-L850

I noticed that `io::ErrorKind::WouldBlock` is already handled here. 
However, considering that this trait might still be implemented elsewhere, it indeed makes sense to perform a loop in `check_write`.

Heap-Hop:
> you could keep track of the number of iterations and return a trap if it exceeds {some number}. 

I believe this requires further discussion before implementation.
How to determine the maximum value and how to define the error.

alexcrichton:
I think it'd be reasonable to pick a small arbitrary number, for example 10, as the limit of turns of the loop. If something is spuriously readable/writable 10 times in a row that seems indicative of a bug

alexcrichton:
I believe the test failure can be fixed by checking if the blocking_read size is zero, and in such a case avoiding the loop entirely or basically doing the read once and returning the result instead of turning the loop again.

Issue #10116, opened 2025-01-26
Wrong link to the Standards Compliant on the main page


AdamStrojek:
On the main page we can find list of features provided by wasmtime. One of them is:

> - [Standards Compliant](https://docs.wasmtime.dev/stability-wasm-proposals-support.html). Wasmtime passes the [official WebAssembly test suite](https://github.com/WebAssembly/testsuite), implements the [official C API of wasm](https://github.com/WebAssembly/wasm-c-api), and implements [future proposals to WebAssembly](https://github.com/WebAssembly/proposals) as well. Wasmtime developers are intimately engaged with the WebAssembly standards process all along the way too.

First link targets `https://docs.wasmtime.dev/stability-wasm-proposals-support.html` which page currently doesn't exist

I think correct address should be: `https://docs.wasmtime.dev/stability-wasm-proposals.html`

Here you can find page that I mentioned:
https://wasmtime.dev

Additionally, footnote for [On-by-default proposals](https://docs.wasmtime.dev/stability-wasm-proposals.html#on-by-default-proposals) is unordered

alexcrichton:
Fixed in https://github.com/bytecodealliance/wasmtime/pull/10117

PR #10117, opened 2025-01-26
docs: Fix Standards Compliant link

saulecabrera:
See https://github.com/bytecodealliance/wasmtime/issues/10116


saulecabrera:
One thing that I noticed is that the top level README, *Standards Compliant* links to https://docs.wasmtime.dev/stability-tiers.html, however, in the `wasmtime` crate it links to https://docs.wasmtime.dev/stability-wasm-proposals.html and I'm not sure if this is intended. It seems that the main README should also point to https://docs.wasmtime.dev/stability-wasm-proposals.html?

abrown (approved):

abrown:
> It seems that the main README should also point to https://docs.wasmtime.dev/stability-wasm-proposals.html?

Yeah, that would make more sense to me as well.

PR #10123, opened 2025-01-27
 Make it easier to reuse fuzzing configuration on the CLI

alexcrichton:
During fuzzing we emit a debug log of configured options to assist with
reproducing fuzz test cases outside of the fuzzing harness. Mapping
options the fuzzer is using though to CLI flags, for example, is a bit
of an art though and not obvious. Just today we've got a fuzz bug and I
couldn't figure out how to reproduce on the CLI and it turns out the
issue was that I was forgetting a flag that was being configured in
response to another flag. I got a bit fed up with constantly trying to
map one to the other, so I've decided to fix things.

This commit adds a `Display for CommonOptions` implementation to the
`wasmtime-cli-flags` crate. This is built on the same macro-construction
infrastructure of all our flags making it a relatively low one-time-cost
to implement this. Each option value now implements not only parsing but
printing as well.

Next the `wasmtime-fuzzing` crate was updated to create a
`CommonOptions` first which is then in turn used to create a
`wasmtime::Config`. This provides a layer to insert a log statement with
to emit all configuration options in a form that can be easily
copy/pasted to the CLI to reproduce.

Overall after doing this I was able to quickly reproduce the bug in
question (yay!). The CLI flag logging is pretty verbose right now since
the fuzzing infrastructure sets many settings redundantly to their
defaults, but reducing flags to a minimum is expected to be relatively
easy compared to otherwise trying to extract the options.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4507,6 +4507,7 @@ dependencies = [
  "wasmparser",
  "wasmprinter",
  "wasmtime",
+ "wasmtime-cli-flags",
  "wasmtime-wast",
  "wasmtime-wast-util",
  "wat",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -445,6 +445,7 @@ wmemcheck = ["wasmtime/wmemcheck"]
 trace-log = ["wasmtime/trace-log"]
 memory-protection-keys = ["wasmtime-cli-flags/memory-protection-keys"]
 profile-pulley = ["wasmtime/profile-pulley"]
+component-model-async = ["wasmtime-cli-flags/component-model-async", "component-model"]
 
 # This feature, when enabled, will statically compile out all logging statements
 # throughout Wasmtime and its dependencies.
--- a/crates/cli-flags/Cargo.toml
+++ b/crates/cli-flags/Cargo.toml
@@ -25,6 +25,7 @@ humantime = { workspace = true }
 async = ["wasmtime/async"]
 pooling-allocator = ["wasmtime/pooling-allocator"]
 component-model = ["wasmtime/component-model"]
+component-model-async = ["wasmtime/component-model-async"]
 cache = ["wasmtime/cache"]
 parallel-compilation = ["wasmtime/parallel-compilation", "dep:rayon"]
 logging = ["dep:file-per-thread-logger", "dep:tracing-subscriber"]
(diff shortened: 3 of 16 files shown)


alexcrichton:
(sorry missed I had two commits here so I've updated the PR title/description with what I intended)

alexcrichton:
An example of this:

**Before**

```
[2025-01-27T17:16:09Z DEBUG wasmtime_fuzzing::oracles] running "/home/alex/code/wasmtime/tests/spec_testsuite/simd_load_extend.wast"
[2025-01-27T17:16:09Z DEBUG wasmtime_fuzzing::generators::config] creating wasmtime config with WasmtimeConfig {
        opt_level: SpeedAndSize,
        regalloc_algorithm: Backtracking,
        debug_info: false,
        canonicalize_nans: false,
        interruptable: false,
        consume_fuel: true,
        epoch_interruption: false,
        memory_config: CustomUnaligned,
        force_jump_veneers: false,
        memory_init_cow: false,
        memory_guaranteed_dense_image_size: 2314885530818453536,
        use_precompiled_cwasm: false,
        async_stack_zeroing: false,
        strategy: OnDemand,
        codegen: Native,
        padding_between_functions: None,
        generate_address_map: false,
        native_unwind_info: false,
        compiler_strategy: Winch,
        collector: Null,
        table_lazy_init: true,
        pcc: true,
        async_config: Disabled,
        signals_based_traps: true,
    }
```

**After**

```
[2025-01-27T17:15:44Z DEBUG wasmtime_fuzzing::oracles] running "/home/alex/code/wasmtime/tests/spec_testsuite/simd_load_extend.wast"
[2025-01-27T17:15:44Z DEBUG wasmtime_fuzzing::generators::config] a custom unaligned host memory will be in use
[2025-01-27T17:15:44Z DEBUG wasmtime_fuzzing::generators::config] creating wasmtime config with CLI options:
    -Ccompiler=winch -Ccollector=null -Cparallel-compilation=n -Cnative-unwind-info=n -Oopt-level=s -Oregalloc-algorithm=backtracking -Omemory-reservation=0 -Omemory-reservation-for-growth=0 -Omemory-guard-size=0 -Oguard-before-linear-memory=n -Omemory-init-cow=n -Omemory-guaranteed-dense-image-size=16777216 -Osignals-based-traps=y -Wnan-canonicalization=n -Wfuel=18446744073709551615 -Wepoch-interruption=n -Wasync-stack-zeroing=n -Wbulk-memory=y -Wmulti-memory=n -Wmulti-value=y -Wreference-types=y -Wsimd=y -Wtail-call=n -Wthreads=n -Wmemory64=n -Wcomponent-model-more-flags=n -Wcomponent-model-async=n -Wfunction-references=n -Wgc=n -Wcustom-page-sizes=n -Wwide-arithmetic=n -Wextended-const=n -Daddress-map=n
```

abrown (approved):
Seems pretty useful!

PR #10127, opened 2025-01-27
Winch: Fix Wasm vector loads to operate on 64-bits

### LINKED ISSUE #10115: Corruption (?) with simd + winch + fuel on spec tests
Found via OSS-Fuzz this fails locally:

```
$ cargo run -q wast -C compiler=winch ./tests/spec_testsuite/simd_load_extend.wast -W fuel=$((1<<62))
Error: failed to run script file './tests/spec_testsuite/simd_load_extend.wast'

Caused by:
    0: failed directive on ./tests/spec_testsuite/simd_load_extend.wast:163:1
           0: <unknown>!<wasm function 0>
    2: wasm trap: all fuel consumed by WebAssembly
```

while this passes:

```
$ cargo run -q wast -C compiler=winch ./tests/spec_testsuite/simd_load_extend.wast
```

Bisection points at https://github.com/bytecodealliance/wasmtime/pull/9950 so I think this isn't a regression and it was just discovered via fuzzing recently. (cc @jeffcharles)

jeffcharles:
Fixes #10115. Wasm vector loads operate on 64 bits of memory so update `LoadKind::derive_operand_size` to return that result for vector loads. This will result in [the value containing the offset with access size](https://github.com/bytecodealliance/wasmtime/blob/1bd66bf65ba96a397d75e51f44c9bc79a1615ff8/winch/codegen/src/codegen/mod.rs#L673) to be set so that bounds are enforced correctly.

### DIFF
--- a/tests/disas/winch/x64/load/v128_load32x2_s_oob_avx.wat
+++ b/tests/disas/winch/x64/load/v128_load32x2_s_oob_avx.wat
@@ -1,0 +1,40 @@
+;;! target = "x86_64"
+;;! test = "winch"
+;;! flags = [ "-Ccranelift-has-avx=true", "-Omemory-reservation=0" ]
+
+(module
+  (memory (data "\00\00\00\00\00\00\00\00\00\00\00\00\00\00\a0\7f"))
+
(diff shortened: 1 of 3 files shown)


saulecabrera (approved):
Thanks!

PR #10128, opened 2025-01-27
add wasmtime-wasi-io and custom async executor to min-platform example

pchickey:
This example demonstrates how to use wasmtime-wasi-io and a custom async executor as the basis for a custom WASI implementation in a restricted environment. 

It is built on the existing min-platform example. Its enabled by default, but can be disabled to demonstrate that, without component and WASI support, the min-platform example can run on a much smaller memory footprint (see examples/min-platform/embedding/src/allocator.rs `INITIAL_HEAP_SIZE`) and without any of the wasmtime crate features required for component-model and async.

This provides some test coverage that wasmtime-wasi-io executes in a no_std target (previously CI only built it), and that wasmtime-wit-bindgen can generate exports for a no_std target (see https://github.com/bytecodealliance/wasmtime/pull/10105)

### DIFF
--- a/examples/min-platform/embedding/src/wasi.rs
+++ b/examples/min-platform/embedding/src/wasi.rs
@@ -1,0 +1,859 @@
+//! This example demonstrates how wasmtime-wasi-io can be used in a #![no_std]
+//! target as the basis for a WASI implementation.
+//!
+//! This example can execute a wasi:cli/command component on a custom async
+//! executor with no dependencies on the environment: execution is
+//! deterministic, and no sources of input are provided to the component. The
+//! WASI implementation is deliberately limited and incomplete, and many WASI
+//! components will not even instantiate, or execute correctly, because this
+//! is not a fully fleshed-out example.
+//!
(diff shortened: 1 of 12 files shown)


examples/min-platform/embedding/src/wasi.rs
```diff
@@ -0,0 +1,859 @@
...
+
+    // No special waker needed for this executor.
+    let waker = noop_waker();
+    let mut cx = Context::from_waker(&waker);
+    let mut f = core::pin::pin!(f);
+
+    // Drive the Future to completion in the following loop
+    let r = 'outer: loop {
+        const POLLS_PER_CLOCK: usize = 200; // Arbitrary. Could be as little
+                                            // as 1. Doesn't hurt to kee[
```
dicej, line 449: Incomplete/truncated sentence in comment?

dicej (approved):

examples/min-platform/embedding/src/wasi.rs
pchickey: Thanks

Issue #10118, opened 2025-01-27
Cranelift: IR Reference example triggers debug assert on x86_64


iwanders:
### `.clif` Test Case

The Cranelift IR `average` example code from the [docs/ir.md](https://github.com/bytecodealliance/wasmtime/blob/5dfccc078bb3dcad152d37013dde8067f197e6e7/cranelift/docs/ir.md#overall-structure). I've already added that to the filetests test directory in caeafe27c69804e14c10d28d25ce511488afd212. 

Using `clif-util`;
```
../target/debug/clif-util  bugpoint  filetests/filetests/isa/x64/average.clif x86_64
After pass 0, remaining insts/blocks: 22/3 (will keep reducing)
After pass 1, remaining insts/blocks: 13/3 (will keep reducing)
After pass 2, remaining insts/blocks: 13/3 (stop reducing)
████████████████████████████████████████████████████████████ pass 2 phase merge blocks                   2/   2 Remove unused global values
Crash message: assertion `left == right` failed
  left: types::I32
 right: types::I64

function %average() -> f32 system_v {
    ss0 = explicit_slot 8

block1:
    v0 = iconst.i32 0
    v1 = iconst.i32 0
    v5 = iconst.i32 0
    v6 = iadd v0, v5  ; v0 = 0, v5 = 0
    v9 = f64const 0.0
    v100 = f32const +NaN
    brif v1, block2, block5  ; v1 = 0

block2:
    v7 = load.f32 v6
    v8 = fpromote.f64 v7
    v10 = fadd v8, v9  ; v9 = 0.0
    stack_store v10, ss0
    trap user1

block5:
    return v100  ; v100 = +NaN
}

5 blocks 22 insts -> 3 blocks 13 insts
```

Manually, I reduced it to the following:
```
function %average(i32, i32) -> f32 system_v {
    ss0 = explicit_slot 8        ; Stack slot for `sum`

block1(v0: i32, v1: i32):
    v20 = f64const 0x0.0         ; Create a f64 as 0.0, just to initialise ss0.
    stack_store v20, ss0         ; Store f64 into the ss0 stack slot, just to initialise ss0.
    v6 = iadd v0, v1             ; Adds the input together as integers, output is i32?
    v7 = load.f32 v6             ; Converts v7 to f32 from i32.
    v8 = fpromote.f64 v7         ; converts v7 to f64 into v8.
[...]

PR #10130, opened 2025-01-28
IR documentation improvements and improve debug assert

iwanders:
Improvements discussed in #10118, fyi @cfallin.

Changes are two-fold, commits are split out.

In 681c8c8fb5e286afb58b97f84cd23b0b0af70d54 adds a message to the debug assert stating the reason why the assert failed. Other asserts with string were capitalised, I tried to match it, reads like:
```
assertion `left == right` failed: Address width of 64 expected, got i32
  left: types::I32
 right: types::I64
```
Technically the type assignment and print isn't necessary, but I found it reads more clear than left and right only, though the fmt made it a multi-line :/ 

In c5f31de428e5f7bb0ea84a64a7e228c9ea9e7574 the size of the first and second argument is changed to 64 bits, as well as the `v3` and `v4`, which hold the offset into the array, I also added one sentence above it pointing at the pointer size is assumed to be 64 bits for the conversion of the function to IR. I did the changes manually, confirmed the changes do not trigger the debug assert and verified the created symbol does work to calculate an average.

Meta suggestion: The PR template contains:
> Our development process is documented in the Wasmtime book:
> https://docs.wasmtime.dev/contributing-development-process.html

That should really be linked in [CONTRIBUTING.md](https://github.com/bytecodealliance/wasmtime/blob/main/CONTRIBUTING.md) as well, because I ended up looking at the commit history to figure out the commit message preference, only to discover that it'll be squashed anyway :D

### DIFF
--- a/cranelift/codegen/src/isa/x64/lower.rs
+++ b/cranelift/codegen/src/isa/x64/lower.rs
@@ -230,7 +230,12 @@ fn lower_to_amode(ctx: &mut Lower<Inst>, spec: InsnInput, offset: i32) -> Amode 
     // We now either have an add that we must materialize, or some other input; as well as the
     // final offset.
     if let Some(add) = matches_input(ctx, spec, Opcode::Iadd) {
-        debug_assert_eq!(ctx.output_ty(add, 0), types::I64);
+        let output_ty = ctx.output_ty(add, 0);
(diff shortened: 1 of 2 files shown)


cfallin (approved):
Thanks!

PR #10132, opened 2025-01-28
Make TOML config keys kebab-case, and add docs

ludfjig:
Updates the doc for flag introduced in #9811.

Feedback appreciated, happy to change things around. The reason I added some stuff on the --optimize, --debug things were to be able to reference those options in the TOML section.

### DIFF
--- a/docs/cli-options.md
+++ b/docs/cli-options.md
@@ -126,3 +126,58 @@ display what Cranelift settings are inferred for the host:
 ```sh
 $ wasmtime settings
 ```
+
+# Additional options
+Many of the above subcommands also take additional options. For example,
+- run 
+- serve
+- compile
+- explore
+- wast
+
+are all subcommands which can take additional CLI options of the format 
+
+```sh
+Options:
+  -O, --optimize <KEY[=VAL[,..]]>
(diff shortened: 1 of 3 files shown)


docs/cli-options.md
```diff
@@ -126,3 +126,58 @@ display what Cranelift settings are inferred for the host:
...
+$ wasmtime compile --config config.toml
+``` 
+would be the same as 
+```sh
+$ wasmtime compile --optimize opt-level=0
+```
+assuming the TOML file is called `config.toml`. Of course you can put as many
+key-value pairs as you want in the TOML file.
+
+> Note: When specifying names in TOML, you must replaces dashes with underscores. In the example above, `opt-level` becomes `opt_level`.
```
alexcrichton, line 183: For this, could we actually use `#[serde(rename_all = "...")]` to blanket rename everything to `kebab-case`? That way it'd be the same on the CLI and in TOML?

alexcrichton (commented):
Thanks for this!

docs/cli-options.md
ludfjig: This is a good idea

alexcrichton (approved):

PR #10144, opened 2025-01-28
Add a style guideline for conditional compilation

alexcrichton:
I've been doing a fair amount of work recently that's touched on `#[cfg]` in many ways throughout Wasmtime. I've personally got opinions about how to structure everything as well, and these opinions are not always obvious to others or discerned from just reading snippets. To assist with this I figured it would be nice to have a style guideline for Wasmtime's conditionally compiled code explaining at least at a high level what's going on and some rough basic principles.

I've attempted to give this a stab and have added a page to the contributing documentation about the style guidelines for conditional compilation. I'm sure I've forgotten something here but my hope is that we can evolve this over time.

### DIFF
--- a/docs/contributing-conditional-compilation.md
+++ b/docs/contributing-conditional-compilation.md
@@ -1,0 +1,277 @@
+# Conditional Compilation in Wasmtime
+
+The `wasmtime` crate and workspace is both quite configurable in terms of
+runtime configuration (e.g. `Config::*`) and compile-time configuration (Cargo
+features). Wasmtime also wants to take advantage of native hardware features on
+specific CPUs and operating systems to implement optimizations for executing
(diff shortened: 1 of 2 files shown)


docs/contributing-conditional-compilation.md
```diff
@@ -0,0 +1,277 @@
...
+```rust,ignore
+use std::ptr::NonNull; //~ WARNING: unused import when `foo` is turned off
+
+#[cfg(foo)]
+fn my_function() -> NonNull<u8> {
+    // ...
+}
+```
+
+**Repetetive Attributes**
```
cfallin, line 104: s/repetetive/repetitive/

cfallin (approved):
Thanks for writing this up!

PR #10147, opened 2025-01-29
Winch: packed integer basic arithmetic for x64

MarinPostma:
implements the following instructions for winch x64:
- `i8x16.add`
- `i8x16.add_sat_u`
- `i8x16.add_sat_s`
- `i16x8.add`
- `i16x8.add_sat_u`
- `i16x8.add_sat_s`
- `i32x4.add`
- `i64x2.add`
- `i8x16.sub`
- `i8x16.sub_sat_u`
- `i8x16.sub_sat_s`
- `i16x8.sub`
- `i16x8.sub_sat_u`
- `i16x8.sub_sat_s`
- `i32x4.sub`
- `i64x2.sub`
- `i16x8.mul`
- `i32x4.mul`
- `i64x2.mul`

#8093


winch/codegen/src/masm.rs
```diff
@@ -1520,4 +1530,30 @@ pub(crate) trait MacroAssembler {
 
     /// If any bit in `src` is 1, set `dst` to 1, or 0 otherwise.
     fn v128_any_true(&mut self, src: Reg, dst: WritableReg) -> Result<()>;
+
+    /// Perform a vector add between `lsh` and `rhs`, placing the result in `dst`, where each lane
+    /// is interpreted to be `size` long.
+    fn v128_add(
```
MarinPostma, line 1536: naming-wise, I'm completely sure what to call those. Maybe `vector_add` is more appropriate?

saulecabrera:
I can take this review.

winch/codegen/src/isa/x64/masm.rs
```diff
@@ -1670,6 +1670,98 @@ impl Masm for MacroAssembler {
...
+            this.ensure_has_avx()?;
+            this.asm.xmm_rmi_rvex(op, lhs, rhs, dst);
+            Ok(())
+        };
+
+        let mul_avx512 = |this: &mut Self, op| {
+            this.ensure_has_avx512vl()?;
+            this.ensure_has_avx512dq()?;
+            this.asm.xmm_rm_rvex3(op, lhs, rhs, dst);
+            Ok(())
```
saulecabrera, line 1755: I don't think we should make this a hard requirement, given that our baseline is AVX.

Given that Intel [suggests that there's no penalty on mixing AVX with AVX512 instructions ](https://www.intel.com/content/www/us/en/developer/articles/technical/intel-avx-512-instructions.html), we could emit AVX512 if they are available, however in case they aren't we still need to emit a fallback to avoid bumping our baseline for this operation. For reference: https://github.com/bytecodealliance/wasmtime/blob/main/cranelift/codegen/src/isa/x64/lower.isle#L1121

winch/codegen/src/masm.rs
saulecabrera: I think the naming is fine here, as far as I can tell, we don't have any other `vector` naming convention.

winch/codegen/src/isa/x64/masm.rs
MarinPostma: my bad, I misunderstood your DM about what we expected to support. I have ported cranelift fallback implementation.

MarinPostma: hold on I just foudn a bug

MarinPostma: Fixed

saulecabrera (approved):
LGTM, thanks!

PR #10149, opened 2025-01-29
Enable warnings if `gc` is disabled

alexcrichton:
Continuation of work in #10131. This additionally handles turning off `gc-null` and `gc-drc` and the various combinations within.

### DIFF
--- a/crates/wasmtime/src/runtime/store.rs
+++ b/crates/wasmtime/src/runtime/store.rs
@@ -76,22 +76,22 @@
 //! contents of `StoreOpaque`. This is an invariant that we, as the authors of
 //! `wasmtime`, must uphold for the public interface to be safe.
 
-use crate::hash_set::HashSet;
 use crate::instance::InstanceData;
 use crate::linker::Definition;
 use crate::module::RegisteredModuleId;
 use crate::prelude::*;
 use crate::runtime::vm::mpk::ProtectionKey;
+#[cfg(feature = "gc")]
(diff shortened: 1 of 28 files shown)


crates/wasmtime/src/runtime/store.rs
```diff
@@ -76,22 +76,22 @@
...
 //! `wasmtime`, must uphold for the public interface to be safe.
 
-use crate::hash_set::HashSet;
 use crate::instance::InstanceData;
 use crate::linker::Definition;
 use crate::module::RegisteredModuleId;
 use crate::prelude::*;
 use crate::runtime::vm::mpk::ProtectionKey;
+#[cfg(feature = "gc")]
+use crate::runtime::vm::GcRootsList;
```
abrown, line 85: I thought you wanted to fully qualify these at the use site?

abrown: I guess there are multiple uses in this module or something like that?

abrown (approved):

crates/wasmtime/src/runtime/store.rs
alexcrichton: Ideally yeah I generally prefer avoiding `#[cfg]` on imports since it feels pretty noisy, but this one was used so much it was getting even noisier to have fully qualified use-sites.

Probably an indicator to create `crates/wasmtime/src/runtime/store/gc.rs` perhaps...

Issue #10102, opened 2025-01-24
Pulley Performance Tracking


alexcrichton:
I wasnted to open a meta-issue about tracking the performance of Pulley. There's a few items I've identified about improving Pulley's performance which I'm unable to tackle today myself so I'm hoping to track both the meta-topic here of Pulley's performance.

Overall Pulley is in a relatively good spot with respect to performance right now, but Pulley is by no means outstripping other wasm interpreters. Pulley's chief downside from what I can tell is that it's starting from a much lower level of abstraction than other interpreters, CLIF, instead of wasm. This particularly hurts memory accesses where Pulley fundamentally uses two opcodes per memory access: one for the bounds check and one for the actual load/store. In terms of interpreter performance is this pretty costly to turn the interpreter loop twice per load/store where other interpreters are likely only turning the loop once.

That's not to say that Pulley is fundamentally less performant than other interpreters, however. Pulley also has the strengths of an optimizing compiler such as Cranelift to perform relatively advanced optimizations such as hoisting loads/stores out of loops. Pulley also can relatively easily add macro-ops as necessary to improve performance as well.

Locally though what I've seen are performance issues with Pulley are:

* [ ] As mentioned above, loads/stores are two opcodes. This is because one opcode is "rooted" in the CLIF `trapnz` node for the bounds check. Another node is rooted in the load/store itself. Lowering in ISLE has no way currently to merge these two instructions together into a single trapping instruction. Improving this would need the ability to lower two terms at once (sort of?) or something like a peephole optimization pass after VCode is created.
* [ ] Pulley register moves are not cheap, unlike native platforms, and regalloc2's allocation decisions don't always minimize register moves. One example is https://github.com/bytecodealliance/wasmtime/issues/9942. Other decisions I've found hard to isolate into reproducible examples, but I've seen when benchmarking [this binary](https://github.com/wasmi-labs/wasmi-benchmarks/blob/d045a88246d3ac9b0b80b188feda54b89ca126b5/benches/res/wasm/coremark-minimal.wasm) that a hot loop both starts and ends with the same `xmov`, so I'm not sure why it's there. I've tried to isolate some small examples in `tests/disas/pulley/coremark-1.wat` and `tests/disas/pulley/fib.wat` in-tree.
[...]

PR #10154, opened 2025-01-29
pulley: Reimplement wasm loads/stores & memory opcodes

alexcrichton:
This commit is a large refactoring to reimplement how WebAssembly loads/stores are translated to Pulley opcodes when using the interpreter. Additionally the functionality related to memory support has changed quite a bit with the interpreter as well. This is all based off comments on #10102 with the end goal of folding the two Pulley opcodes today of "do the bounds check" and "do the load" into one opcode. This is intended to reduce the number of opcodes and overall improve interpreter throughput by minimizing turns of the interpreter loop.

The basic idea behind this PR is that a new basic suite of loads/stores are added to Pulley which trap if the address is zero. This provides a route to translate trapping loads/stores in CLIF to Pulley bytecode without actually causing segfaults at runtime. WebAssembly translation to CLIF is then updated to use the `select` trick for wasm loads/stores where either 0 is loaded from or the actual address is loaded from. Basic support for translation and such is added for this everywhere, and this ensures that all loads/stores for wasm will be translated successfully with Pulley.

The next step was to extend the "g32" addressing mode preexisting in Pulley to support a bounds check as well. New pattern-matches were added to ISLE to search for a bounds check in the address of a trapping load/store. If found then the entire chain of operations necessary to compute the address are folded into a single "g32" opcode which ends up being a fallible load/store at runtime.

To fit all this into Pulley this commit contains a number of refactorings to shuffle around existing opcodes related to memory and extend various pieces of functionality here and there:

* Pulley now uses a `AddrFoo` types to represent addressing modes as a single immediate rather than splitting it up into pieces for each method. For example `AddrO32` represents "base + offset32". `AddrZ` represents the same thing but traps if the address is zero. The `AddrG32` mode represents a bounds-checked 32-bit linear memory access on behalf of wasm.

* Pulley loads/stores were reduced to always using an `AddrFoo` immediate. This means that the old `offset8` addressing mode was removed without replacement here (to be added in the future if necessary). Additionally the suite of sign-extension modes supported were trimmed down to remove 8-to-64, 16-to-64, and 32-to-64 extensions folded as part of the opcode. These can of course always be re-added later but probably want to be added just for the `G32` addressing mode as opposed to all addressing modes.

* The interpreter itself was refactored to have an `AddressingMode` trait to ensure that all memory accesses, regardless of addressing modes, are largely just copy/pastes of each other. In the future it might make sense to implement these methods with a macro, but for now it's copy/paste.

* In ISLE the `XLoad` generic instruction removed its `ext` field to have extensions handled exclusively in ISLE instead of partly in `emit.rs`.

* Float/vector loads/stores now have "g32" addressing (in addition to the "z" that's required for wasm) since it was easy to add them.

* Translation of 1-byte accesses on Pulley from WebAssembly to CLIF no longer has a special case for using `a >= b` instead of `a > b - 1` to ensure that the same bounds-check instruction can be used for all sizes of loads/stores.

* The bounds-check which folded a load-of-the-bound into the opcode is now present as a "g32bne" addressing mode. with its of suite of instructions to boo.

Overall this PR is not a 1:1 replacement of all previous opcodes with exactly one opcode. For example loading 8 bits sign-extended to 64-bits is now two opcodes instead of one. Additionally some previous opcodes have expanded in size where for example the 8-bit offset mode was remove in favor of only having 32-bit offsets. The goal of this PR is to reboot how memory is handled in Pulley. All loads/stores now use a specific addressing mode and currently all operations supported across addressing modes are consistently supported. In the future it's expected that some features will be added to some addressing modes and not others as necessary, for example extending the "g32" addressing mode only instead of all addressing modes.

For an evaluation of this PR:

* Code size: `spidermonkey.cwasm` file is reduced from 19M to 16M.
* Sightglass: `pulldown-cmark` is improved by 15%
* Sightglass: `bz2` is improved by 20%
* Sightglass: `spidermonkey` is improved by 22%
* Coremark: score improved by 40%

Overall this PR and new design looks to be a large win. This is all driven by the reduction in opcodes both for compiled code size and execution speed by minimizing turns of the interpreter loop. In the end I'm also pretty happy with how this turned out and I think the refactorings are well worth it.


alexcrichton:
I'll note that one thing I'm having difficulty adding a `disas` test for is something which generates a "g32bne" lowering. I can see the instructions getting emitted in `spidermonkey.cwasm` but I am having a difficult time reducing things down to something that can be added as an example. Most loads aren't "sinkable" which is why it's not used super often at least.

cranelift/codegen/src/prelude_opt.isle
```diff
@@ -120,3 +120,35 @@
...
+(extractor (slt ty x y) (icmp ty (IntCC.SignedLessThan) x y))
+
+(decl sle (Type Value Value) Value)
+(extractor (sle ty x y) (icmp ty (IntCC.SignedLessThanOrEqual) x y))
+
+(decl sgt (Type Value Value) Value)
+(extractor (sgt ty x y) (icmp ty (IntCC.SignedGreaterThan) x y))
+
+(decl sge (Type Value Value) Value)
+(extractor (sge ty x y) (icmp ty (IntCC.SignedGreaterThanOrEqual) x y))
```
fitzgen, line 154: Can we put this in the common prelude, rather than duplicating it in both the opt and lower preludes?

crates/cranelift/src/func_environ.rs
```diff
@@ -3421,6 +3421,20 @@ impl FuncEnvironment<'_> {
     pub fn clif_instruction_traps_enabled(&self) -> bool {
         self.tunables.signals_based_traps || self.isa.triple().is_pulley()
     }
+
+    /// Returns whether loads from the null address are allowed as signals of
+    /// whether to trap or not.
+    pub fn load_from_zero_allowed(&self) -> bool {
+        // Pulley allows loads-from-zero and otherwise this is only allowed with
+        // traps + spectre mitigations.
+        self.isa.triple().is_pulley()
```
fitzgen, line 3430: Might as well use the helper if you're gonna define it:

```suggestion
        self.is_pulley()
```

tests/disas/pulley/memory-inbounds.wat
fitzgen, line 1: All the instruction deltas in this file are equal or worse compared to before -- is that expected?

fitzgen (approved):
LGTM but some questions that should be resolved below.

cranelift/codegen/src/prelude_opt.isle
alexcrichton: It was originally, but the problem is that they're slightly different where the helpers for egraphs have a `ty` variable and the helpers for backends don't. I'm not sure how to otherwise unify those two myself

tests/disas/pulley/memory-inbounds.wat
alexcrichton: That's a good point, and one where I mostly shrugged this off but shouldn't have! I focused a bit more on this and improved a few cases in https://github.com/bytecodealliance/wasmtime/pull/10154/commits/63e2c5a92a3a209b466a8e3a7627bcb9cd3ef159

The remaining ones that got worse are:

* `offset_just_bad_v2` - looking into this it looks like optimizations shuffle constants around enough that the output CLIF is quite different from normal loads/stores, so I think this is something we'll basically just take a hit on for now and improve later if necessary
* `maybe_inbounds_v2` - this doesn't codegen well because egraphs can't optimize away the `uadd_overflow_trap`-with-`iconst`-inputs that the frontend generates. Another possible fix is to optimize the frontend to skip `uadd_overflow_trap` in this case (constant address).

cranelift/codegen/src/prelude_opt.isle
cfallin: FWIW, we want to add the type parameters in the lowering prelude's versions of the bindings eventually, for exactly this reason -- the only reason they aren't there is that lowering rules were written first and we didn't realize we would need them on the constructor side. @mmcloughlin apparently has a private branch where this change has been made -- Michael, would you be willing to think about upstreaming this?

mmcloughlin: Yes, I think I could upstream it. Our fork has diverged a bit so it's not a clean patch, but thankfully I did keep the Python script I used to do most of the fixups and hopefully that procedure would work again.

A couple of things to note about the change in our fork:

* The `ty` parameter is not always present on instruction extractors. In the CDSL code generator, a `ty` parameter added only when the instruction is statically known to always have at least one result. In that case, `ty` is the type of the first result value.
* The existing code uses `has_type` extractor in cases that would now be more naturally be done with the provided `ty` argument. Replacing `has_type` uses with the `ty` argument is hard/impossible to do automatically. This is fine as long as you're okay with migrating away from `has_type` (where possible) more slowly over time.

PR #10158, opened 2025-01-30
Always allow dead_code/unused_import warnings in `wasmtime`

alexcrichton:
Finishes the work started in #10131 and this means that the crate is no longer special and has the same default settings of all other crates in all situations.


alexcrichton:
This is going to be particularly gnarly, so I'm deferring this to a later date.

PR #10177, opened 2025-02-03
Support for stack switching proposal

frank-emrich:
This PR adds support for the [Wasm stack switching proposal](https://github.com/WebAssembly/stack-switching). The explainer document for the proposal is [here](https://github.com/WebAssembly/stack-switching/blob/main/proposals/stack-switching/Explainer.md).

This means that the following instructions are added: `cont.new`, `resume`, `suspend`, `switch`, and `cont.bind`.
The instruction `resume.throw`, which is part of the proposal but relies on exception handling, is currently unsupported.

This was developed together with @dhil.

### General implementation notes

In Wasm, continuations are represented by values of type `(ref $ct)`, where `$ct` is a new composite type/heap type for continuations.
In the implementation, these are represented by values of type `VMContObj`. These are fat pointers, consisting of a sequence counter, and a pointer to a `VMContRef`. The latter type is used for the actual representation of continuations.
The sequence counter part of `VMContObj`s is used to check that every continuation value can only be used once.

The `StoreOpaque` is extended to contain a "stack chain": It indicates what stack we are currently executing on. Logically, this is a linked list of stacks, since each continuation has a parent field. The chain stored in the `StoreOpaque` always ends with a value representing the initial stack. This is the stack we were on when entering Wasm, which will usually be the main stack. 

### Memory Management

Currently, memory management is very basic: The `StoreOpaque` provides a new method for allocation a new `VMContRef`, and keeps track of all continuations created this way. Continuations are never deallocated at the moment, meaning that they live until the store itself is deallocated. 

The stack memory used by each allocation (represented by type `ContinuationStack`) is always `mmap`-ed, with a guard page at the end. There is currently no support for growing stacks, or segmented stacks.

### Backtrace generation

The existing backtrace generation mechanism is extended to be fully aware of continuations: When generating backtraces, the entire chain of continuation is traversed, not just the stack frames of the currently active stack/continuation.

### Integration with GC

Integration with the GC proposal is limited, but should at least be safe:
- Programs where continuation objects would be stored inside GC objects (structs, arrays, ...) are rejected
- During GC, all stack frames of all continuations are inspected when looking for GC roots.

### Entering/Exiting Wasm

Prior to this PR, there were two separate mechanisms that save and restore some state of the runtime when entering/exiting Wasm:
1. The functions `enter_wasm` and `exit_wasm` in `func.rs`
2. `CallThreadState` saves and restores (on `drop`) parts of the `VMRuntimeLimits`

This PR consolidates these two mechanism, because it requires some additional state to be updated and restored on enter/exit:
the type `wasmtime::runtime::func::RuntimeEntryState` now stores all of the required runtime state and ensures that it's restored when exiting Wasm.

### Tags
Since the stack switching proposal extends the notion of tags introduced by the exception handling proposal, it adds the necessary machinery for importing, exporting, and defining tags.

### Limitations

The limitations are as follows, some of which were mentioned above:
- `resume.throw` needs to be implemented once exception handling is there
- Full integration with GC and support for deallocating continuations
- The only supported platform at the moment are unix-like x64 systems. I'm planning to change that soon as follow-up PRs, this won't require big changes to what's in this PR at the moment.
- Neither Winch or Pulley are supported at the moment.


frank-emrich:
There are currently a few test failures that I 'm hoping to get some help with:
- I don't understand `cargo vet` well enough to know what's up, but the problem should be fairly benign: We are using `memoffset` in a different crate now, but from what I can tell, it has already been vetted for use in this project.
- There are some build failures due to the new instructions not being supported on all platforms. In general, we've added a Wasm feature and a cargo feature for stack switching. My plan was that all stack switching tests would be skipped if the `stack-switching` cargo feature is not enabled. That's easy to do for the `all` test suite, but I'm not sure what the best approach is for the `wast` tests to achieve this. Am I understanding correctly that the `wast` test suite always currently runs all tests, independently from what cargo features are enabled?

As a general, so far most of the stack switching code is compiled even if the cargo feature is disabled. It would be possible to get more aggressive with conditional compilation, but this requires some ugly special casing in `backtrace.rs`. Happy for suggestions if people would like to see more conditional compilation or are happy with keeping it minimal.

fitzgen:
> * Am I understanding correctly that the `wast` test suite always currently runs all tests, independently from what cargo features are enabled?

That is correct, and we assert that tests that "should" fail do in fact fail. This way, if we fix bugs or expand support, we won't accidentally and silently miss out on tests for those things.

This should be fine for your stack switching tests, as long as you

* mark the tests as "should fail" on supported platforms, and
* return an error or some sort (rather than unsafely segfault or whatever) when processing Wasm that uses this proposal on an unsupported platform/configuration.

See https://github.com/bytecodealliance/wasmtime/blob/d7d605c236a857a4019aa39850e9dd6a6597ece1/crates/wast-util/src/lib.rs#L280

alexcrichton:
Oh sorry I should have left a note here but I think I forgot. I was out last week for a company offsite and I'm out this week for the CG meeting, but I plan on getting to this next week. I want to make sure I've got plenty of time to sit down and go through this, but wanted to give a heads-up that I'll be looking into it next week at the earliest.

Also to reiterate: thank you so much for this! Thanks for wrestling with CI as well, much appreciated!

crates/wasmtime/Cargo.toml
```diff
@@ -309,6 +310,11 @@ threads = [
   "std",
 ]
 
+stack-switching = [
```
alexcrichton, line 313: Mind adding some docs above this explaining what this is gating?

crates/fuzzing/src/oracles/dummy.rs
```diff
@@ -23,6 +23,7 @@ pub fn dummy_extern<T>(store: &mut Store<T>, ty: ExternType) -> Result<Extern> {
         ExternType::Global(global_ty) => Extern::Global(dummy_global(store, global_ty)?),
         ExternType::Table(table_ty) => Extern::Table(dummy_table(store, table_ty)?),
         ExternType::Memory(mem_ty) => Extern::Memory(dummy_memory(store, mem_ty)?),
+        ExternType::Tag(_) => todo!(),
```
alexcrichton, line 26: Mind tagging this with a FIXME pointing to an issue?

crates/fuzzing/src/generators/config.rs
```diff
@@ -145,6 +145,8 @@ impl Config {
             hogs_memory: _,
             nan_canonicalization: _,
             gc_types: _,
+            exceptions: _,
+            stack_switching: _,
```
alexcrichton, line 149: Mind tagging this with a `// FIXME` and an issue to touch it up later?

crates/environ/src/vmoffsets.rs
```diff
@@ -150,6 +159,18 @@ pub trait PtrSize {
...
+    #[inline]
+    fn size_of_vmtag_definition(&self) -> u8 {
+        4
+    }
+
+    /// This is the size of the largest value type (i.e. a V128).
+    #[inline]
+    fn maximum_value_size(&self) -> u8 {
+        self.size_of_vmglobal_definition()
+    }
```
alexcrichton, line 172: This feels a bit suspicious to me so I'm leaving a note to myself here to revisit this once I've seen more of this PR

crates/wasmtime/src/config.rs
```diff
@@ -127,6 +129,14 @@ pub struct Config {
     profiling_strategy: ProfilingStrategy,
     tunables: ConfigTunables,
 
+    /// Runtime configuration for the stack switching feature.
+    /// The structure is defined in the
+    /// `wasmtime_environ::stack_switching` module, so that we can
+    /// hand out the configuration object in the interface of
+    /// `wasmtime_runtime::Store` trait, where the full `Config` type
+    /// is not in scope.
+    pub stack_switching_config: StackSwitchingConfig,
```
alexcrichton, line 138: This'll definitely want to avoid `pub`, but `pub(crate)` is fine to access it throughout this crate.

crates/wasmtime/src/config.rs
```diff
@@ -777,6 +788,12 @@ impl Config {
         self
     }
 
+    /// Configures the size of the stacks created with cont.new instructions.
+    pub fn stack_switching_stack_size(&mut self, size: usize) -> &mut Self {
+        self.stack_switching_config.stack_size = size;
+        self
+    }
```
alexcrichton, line 795: Would it make sense to tag this with `#[cfg(feature = "stack-switching")]`?

crates/wasmtime/src/config.rs
```diff
@@ -1087,6 +1104,32 @@ impl Config {
...
+    ///
+    /// Note that this feature is a work-in-progress and is incomplete.
+    ///
+    /// This is `false` by default.
+    ///
+    /// [proposal]: https://github.com/WebAssembly/stack-switching
+    pub fn wasm_stack_switching(&mut self, enable: bool) -> &mut Self {
+        self.wasm_feature(WasmFeatures::STACK_SWITCHING, enable);
+        self
+    }
```
alexcrichton, line 1131: Like above, mind tagging this as `#[cfg(feature = "stack-switching")]`?

crates/wasmtime/src/config.rs
```diff
@@ -2423,6 +2485,27 @@ impl Config {
             bail!("cannot disable the simd proposal but enable the relaxed simd proposal");
         }
 
+        if features.contains(WasmFeatures::STACK_SWITCHING) {
```
alexcrichton, line 2488: One other piece of validation (which might belong in `Engine::_check_compatible_with_native_host` perhaps) is to ensure that if stack-switching is enabled that `#[cfg(feature = "stack-switching")]` is also enabled. Otherwise that represents a host that has enabled the wasm feature but not compile-time support so it'd probably be good to return a first-class error in such a situation.

crates/wasmtime/src/runtime/externals/global.rs
```diff
@@ -130,6 +130,9 @@ impl Global {
                                 })
                                 .into(),
                         ),
+                        HeapType::NoCont | HeapType::ConcreteCont(_) | HeapType::Cont => todo!(
+                            "Embedder support for continuations has not yet been implemented!"
+                        ),
```
alexcrichton, line 135: This location unfortunately can't easily be switched to returning a `Result` so panicking makes sense to me, but mind tagging this with `// FIXME(...)` pointing to an issue?

crates/wasmtime/src/runtime/externals/table.rs
```diff
@@ -188,6 +188,8 @@ impl Table {
                         ty => unreachable!("not a top type: {ty:?}"),
                     }
                 }
+
+                runtime::TableElement::ContRef(_c) => todo!(), // TODO(dhil): Required for the embedder API.
```
alexcrichton, line 192: Like above for globals the panic here is fine for now but this'd be good to tag with a `// FIXME` pointing to an issue

crates/wasmtime/src/runtime/externals/tag.rs
```diff
@@ -0,0 +1,44 @@
...
+                    instance.engine_type_index(module_index)
+                })
+            });
+
+        Tag(store.store_data_mut().insert(wasmtime_export))
+    }
+
+    pub(crate) fn ty(&self, _store: impl AsContext) -> TagType {
+        todo!()
+    }
```
alexcrichton, line 31: Since this is just an internal method it might make sense to delete this and update callers to not use it?

crates/wasmtime/src/runtime/externals/tag.rs
```diff
@@ -0,0 +1,44 @@
...
+        mut wasmtime_export: crate::runtime::vm::ExportTag,
+        store: &mut StoreOpaque,
+    ) -> Self {
+        use wasmtime_environ::TypeTrace;
+        wasmtime_export
+            .tag
+            .canonicalize_for_runtime_usage(&mut |module_index| {
+                crate::runtime::vm::Instance::from_vmctx(wasmtime_export.vmctx, |instance| {
+                    instance.engine_type_index(module_index)
+                })
```
alexcrichton, line 23: As a heads-up I think this may no longer be necessary after https://github.com/bytecodealliance/wasmtime/pull/10223 landed.

crates/wasmtime/src/runtime/values.rs
```diff
@@ -280,6 +280,10 @@ impl Val {
 
                     HeapType::NoFunc => Ref::Func(None),
 
+                    HeapType::NoCont | HeapType::ConcreteCont(_) | HeapType::Cont => {
+                        unimplemented!()
+                    } // TODO(dhil): Need to do this for the embedder API.
```
alexcrichton, line 285: Mind tagging this with a `// FIXME` pointing to an issue?

alexcrichton (commented):
Ok I think I've seen about half of the files of this PR (excluding auto-updates in tests) so wanted to leave an initial round of comments/thoughts.

At a high-level everything here looks reasonable to me so far. I haven't dug into the actual implementation details of stack switching though and have for this been focusing on the integration points with the rest of Wasmtime. I left a lot of comments along the lines of "file an issue to point a comment to this" and I think it might make the most sense to have a single tracking issue for follow-up work on stack-switching. That can have a lot of various checkboxes for the logical work items remaining (e.g. contref in the gc heap, in the embedding API, windows support, etc).

Additionally at a high-level I want to also ask you what your own time allowance is for landing this? This is, as expected, a large PR which is going to take some time to land. I suspect it'll take me on the order of ~days of review to get through the the real "meat" of stack switching and I'll likely have comments from all that as well. If, however, you're running short on time for this that's ok too!

Basically what I want to ask is: do you have time allocated for the back-and-forth of review here? It's hard to predict exactly how long it will take but my gut is that it'll be a nontrivial amount of time. If you don't have time for this I think that's also totally ok. I'll probably still continue to review all this and I'll try to find time in the coming future to work on addressing many of these items myself. 

In an ideal world this PR would be split up into chunks and landed piecemeal, for example parsing support, then compiling support, then maybe an instruction-at-a-time for each piece, etc. Splitting this up is no easy task though so I don't want to ask this lightly of you, and you also know much better than I about if it even could be split up and where it could be split.  Part of this is that GitHub does not make reviewing large PRs easy (e.g. every time I add a comment it takes ~5 seconds for the comment box to appear), and overall it's easier to focus on smaller changes to ensure that as little as possible slips through the cracks.

Basically tl;dr; I've left a lot of comments below but it doesn't have to be you alone addressing them. If you're running low on time to allocate to this work I think we can try to be accomodating in assisting to landing this.

frank-emrich:
@alexcrichton

Thanks so much for having a look at this already!

Regarding my own time commitment for this PR: I'm starting a new job in mid March, so things are looking as follows:
- Until end of February: Focusing exclusively on getting this PR in shape
- First two weeks of March: Working on this PR, but some distractions due to moving, etc.
- From mid March: There will be an initial period of me settling in at my new job during which I can't do much on this. After that the Wasmtime/stack switching work will not part of my day job anymore, but I'll still be around to help out, address bugs, etc. I also have a few things on top of the current PR already done or planned that I want to add at that stage (aarch64 support, Windows support, better DWARF support, ....).

@dhil should also be able to help out here and there.

> In an ideal world this PR would be split up into chunks and landed piecemeal, for example parsing support, then compiling support, then maybe an instruction-at-a-time for each piece, etc.

I think there was just a bit of a miscommunication between me and @dhil. I thought you wanted to see this as one big chunk. There's a simple way in which this PR could be split up, leading to a sequence of PRs as follows:
1. Integrate tag support, with nothing stack switching related, yet.
2. Add dummy field to VMContext with the size and position that the stack switching field will later occupy. This way, all the changes to the disas tests will happen in this PR. I'm not sure if the disas tests actually contribute to Github choking on this PR, but at least getting them out of the way will reduce the amount of conflicts coming in.
3. Everything except the actual Wasm -> CLIF translation. This will be a big chunk where most of the integration of stack switching with the rest of Wasmtime goes.
4. The Wasm -> CLIF translation. This will mostly be what's currently in `crates/cranelift/src/stack_switching/`, in particular `instructions.rs`. The latter is 2500 LOC, so reviewing that separately seems worthwhile. I may even be able to split this up further.

This is far from perfect because PR 3 and 4 will still be large and not really split along separate features of the stack switching implementation, but at least that split up would be easy for me to do. I may discover some more chunks that can be factored out into separate PRs on the way.

My only concerns about this is the following: Given that Github doesn't really support stacked PRs, we would work through these 4+ PRs linearly, making it difficult to point at why we need something in the next PR.  
On the other hand, I could just put a branch somewhere that shows everything that's still missing at any point so you can have a look at how something will be used in the remaining PRs.

Please let me know what you think. I will also make sure to work through the comments you already added to this PR so they don't get lost if we do decide to split this PR.

crates/cranelift/src/func_environ.rs
```diff
@@ -91,7 +91,8 @@ wasmtime_environ::foreach_builtin_function!(declare_function_signatures);
 /// The `FuncEnvironment` implementation for use by the `ModuleEnvironment`.
 pub struct FuncEnvironment<'module_environment> {
     compiler: &'module_environment Compiler,
-    isa: &'module_environment (dyn TargetIsa + 'module_environment),
+    /// NOTE(frank-emrich) pub for use in crate::wasmfx::* modules
```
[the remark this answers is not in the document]
frank-emrich: This was just a note to ourselves that we made this field `pub(crate)` because we access it from our code. For example, we call `crate::value_type` a few times, which requires a `&dyn TargetIsa`. But I'm happy to leave a mental note here that we should double-check if that's really necessary

crates/environ/src/builtin.rs
```diff
@@ -206,6 +206,32 @@ macro_rules! foreach_builtin_function {
...
+            // FIXME(frank-emrich) The next three builtins are used by the debug printing mechanism.
+            // They are not supposed to be part of the final upstreamed code.
+            //
+            // Prints a 'static str, represented as a
+            // pointer and a length.
+            delete_me_print_str(vmctx: vmctx, s: pointer, len : u64);
+            // Prints integer
+            delete_me_print_int(vmctx: vmctx, arg : u64);
+            // Prints pointer, formatted as hex.
+            delete_me_print_pointer(vmctx: vmctx, arg : pointer);
```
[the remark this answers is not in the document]
frank-emrich: Sorry, there's one thing I forgot to mention in the overall PR description: There is currently some debugging logic left in this PR, you will run into it when looking at `instructions.rs`: We have added `emit_debug_println!` and `emit_debug_assert!`. Their implementation is too hacky to keep in the final version that we merge, but I'm hoping to keep them until the very end and only rip them out when the rest of the code is ready. They are incredibly helpful while I may need to make any kind of changes to the code.

Anything that has something like "delete me" in the name is related to this debugging code.

crates/environ/src/stack_switching.rs
```diff
@@ -0,0 +1,345 @@
...
+#[macro_export]
+macro_rules! debug_println {
+    ($( $args:expr ),+ ) => {
+        #[cfg(debug_assertions)]
+        if ENABLE_DEBUG_PRINTING {
+            #[cfg(feature = "std")]
+            println!($($args),*);
+        }
+    }
+}
```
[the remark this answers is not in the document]
frank-emrich: This is also an instance of "debugging code that I would like to keep around until the rest of the PR is ready to merge"

crates/environ/src/stack_switching.rs
```diff
@@ -0,0 +1,345 @@
...
+        use crate::stack_switching::*;
+        use core::mem::offset_of;
+
+        /// Offset of `capacity` field
+        pub const CAPACITY: usize = offset_of!(Array<()>, capacity);
+        /// Offset of `data` field
+        pub const DATA: usize = offset_of!(Array<()>, data);
+        /// Offset of `length` field
+        pub const LENGTH: usize = offset_of!(Array<()>, length);
+    }
```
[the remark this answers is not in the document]
frank-emrich: Yes, that's absolutely the reason why some of the types that should morally live in the runtime/wasmtime crate are defined here, just to make these offset definitions easier.

I actually suspected that the module here would have to be merge into (something like) `VMOffsets`, there's a comment a few lines above saying that.

I'm happy to do that refactoring, but just a few questions:
Should I actually integrate all the offsets for these types here into `VMOffsets` (or more accurately, the `PtrSize` trait)? Or was your suggestion to re-create similar infrastructure elsewhere? I'm mostly asking because AFAIK all the types that `vmoffsets.rs` talks about have the `VM*` prefix. There's nothing really stopping us from adding that prefix to our own types, but I'm unsure if that prefix carries more meaning and may be inappropriate for our types.

crates/environ/src/trap_encoding.rs
```diff
@@ -92,6 +92,15 @@ pub enum Trap {
...
     NoAsyncResult,
+
+    /// We are suspending to a tag for which there is no active handler.
+    UnhandledTag,
+
+    /// Attempt to resume a continuation twice.
+    ContinuationAlreadyConsumed,
+
+    /// For debug assertions in generated code.
+    DebugAssertion,
```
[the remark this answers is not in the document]
frank-emrich: Yes, definitely! See my comment on the `delete_me_*` libcalls

crates/environ/src/vmoffsets.rs
frank-emrich: Yes, this is ugly. This is basically just a way of saying "give me the size of a `ValRaw`", without having access to that type here.
The reason we need that size is that all continuation payloads (arguments to continuations, return values from continuations, values provided by `suspend`, etc) are currently written to and then read from buffers containing `ValRaw`. Let's keep this comment open, we will certainly be re-visiting that discussion when you have a look at `instructions.rs`.

alexcrichton:
> Regarding my own time commitment for this PR

Ok that all sounds good, thanks! (very helpful to set expectations!)

> I think there was just a bit of a miscommunication between me and @dhil.

Oh no that's on me. I remember telling @dhil that and I think I was just basically wrong, so that's on me for leading y'all astray. Given the size of this (not just the auto-updated tests but also the scale of the implementation) along with the subtelty of the implementation in retrospect I just didn't account for that and misled y'all. If possible I think the best way to handle this would be to peel off PRs from this one and submit them independently.

Overall my goal is to ensure that we get a good chance to review everything coming in and reducing the chance of something being lost by accident due to being overlooked. For coordination what I would imagine is that it's completely reasonable to have lightly, if at all, tested intermediate states. Additionally it's totally reasonable to leave a "hole" in the code with something like `// more coming soon ...` (e.g. making a hole in the `VMContext` for the stack-switching state necessary for a future PR). Let's keep this PR open so it can be cross-referenced if needed, and this can be rebased occasionally as smaller pieces are landed.

Does that seem ok? It's a similar strategy we're taking for landing [async component model support in Wasmtime](https://github.com/bytecodealliance/wasmtime/pull/9582) where it's landing piece-by-piece. Each piece is not 100% thoroughly tested but the  "final PR" has lots of tests. We try to keep track of outstanding TODO items in issues occasionally and otherwise are ok deferring work to a future PR which has more tests and more fully justifies a prior design.

Your sequence of PRs makes sense to me as well. I'd mostly defer to y'all for pieces to split up as you're the most familiar with the implementation, and if something seems too big I can bring it up during the review but by naturally splitting this up I think most slicing/dicing will be quite reasonable.

frank-emrich:
> Overall my goal is to ensure that we get a good chance to review everything coming in and reducing the chance of something being lost by accident due to being overlooked.

Yes, absolutely. I'm happy to do the necessary chopping. I'll create a tracking issue describing the bigger picture and where we currently are regarding the PRs that have and haven't landed at any time.

The outline of PRs I mentioned in my previous comment will still lead to fairly large ones, but I will see if there's potential to split this up further and wait for your feedback once I've split off a candidate for a PR.

Also, sorry for making you wade through lots of `todo!()`,  `unimplemented!()` and other little things that we hadn't resolved before creating this PR.

alexcrichton:
Sounds good! It's ok to skip the step of writing out a plan in an issue and then executing on the plan, we can work on making sure things are documented before you wind down but up until then it's ok if it's just in a few heads. In that sense no need to exhaustively document "here's the round peg that's gonna fit in that round hole" and we can just have one issue for tracking known items that are already in-tree.

And please no apologies necessary! Landing big changes I've found is always a bit of an art and if anything I'm the one who led you astray by being overconfident!

crates/wasmtime/src/runtime/vm/vmcontext.rs
```diff
@@ -263,6 +263,41 @@ mod test_vmglobal_import {
...
+/// tag imported from another instance.
+#[derive(Debug, Copy, Clone)]
+#[repr(C)]
+pub struct VMTagImport {
+    /// A pointer to the imported tag description.
+    pub from: VmPtr<VMTagDefinition>,
+    /// A pointer to the owning instance.
+    pub vmctx: VmPtr<VMContext>,
+}
+
```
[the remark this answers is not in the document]
dhil: A tag is more than a wasm type. It is a nominal entity and its unique identity is important. Basically we get a unique identity by picking backing on pointer equality.

I think we can get rid of the vmctx, potentially. Though we can discuss the best course of action here. I have pulled the tag stuff into its own patch which I will open tomorrow when I have had a chance to look at the tests.

frank-emrich:
Converting this back to a draft now because it will be split up into multiple PRs.

Over the next few weeks, I expect to rebase this PR whenever parts of it have landed. This way, this PR will always contain whatever is left.

I've also created issue #10248 to track the overall status. I will also use that issue in various FIXMEs for missing features.

crates/environ/src/stack_switching.rs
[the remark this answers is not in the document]
alexcrichton: > Should I actually integrate all the offsets for these types here into VMOffsets (or more accurately, the PtrSize trait)?

Yeah let's do that. That's the conventional location we have for dealing with offsets-of-things in compiled code, so I think keeping it all in one place is best.

> There's nothing really stopping us from adding that prefix to our own types, but I'm unsure if that prefix carries more meaning and may be inappropriate for our types.

Oh no special meaning yeah, just our convention for "this datatype is read by both compiled code and the host" and then we try to group things together where it makes sense

crates/environ/src/stack_switching.rs
```diff
@@ -0,0 +1,345 @@
...
+
+/// Runtime configuration options for stack switching that can be set
+/// via the command line.
+///
+/// Part of wasmtime::config::Config type (which is not in scope in this crate).
+#[derive(Debug, Clone)]
+pub struct StackSwitchingConfig {
+    /// The (fixed) size of a continuation stack.
+    pub stack_size: usize,
+}
```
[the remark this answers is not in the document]
dhil: I think we need some knobs for continuation stacks, e.g. limit the number of allocated stacks. I think the stack size can be useful, but perhaps its use is rather limited until we have implemented growable stacks. I think we also want to land a pooling allocator for continuation stacks, as it is necessary to obtain decent performance on the workloads we have been benchmarking. 

Though, I'd be quite happy to simply use `wasm_stack_size` and keep the specific stack switching config downstream for now.

alexcrichton:
I believe this PR was split up and has now since landed, so I'm going to close this.

PR #10178, opened 2025-02-04
asm: prepare docs for publishing

abrown:
This change is motivated by staring at the output of `cargo doc` for the `cranelift-assembler-x64` crate. In order to write a sensible top-level example, I felt it was best to refactor how we construct the instructions: this removes the `build` module and adds conventional `<inst_name>::new` functions to each instruction. Also, a generated `From` implementation makes it easier to convert to an `Inst`.

This change has other doc-related refactorings and tweaks, but should not change any functionality.


fitzgen (approved):
nice, thanks!

PR #10180, opened 2025-02-04
Winch: Add SIMD conversion operators for x64 with AVX

jeffcharles:
Part of #8093. Adds the following instructions:
- `*x*.convert_*`
- `*x*.demote_*`
- `*x*.extend_*`
- `*x*.narrow_*`


saulecabrera:
I can help with this review.

winch/codegen/src/isa/x64/asm.rs
```diff
@@ -2082,6 +2226,155 @@ impl Assembler {
...
+            OperandSize::S64 => AvxOpcode::Vsubpd,
+            _ => unimplemented!(),
+        };
+
+        let address = Self::to_synthetic_amode(
+            src2,
+            &mut self.pool,
+            &mut self.constants,
+            &mut self.buffer,
+            MemFlags::trusted(),
```
MarinPostma, line 2272: Is it generally fine to consider those memory access trusted? It seems to me that, in this specific case, this is fine, but the callee cannot assume that address will indeed be aligned and non-trapping, no?

saulecabrera: > Is it generally fine to consider those memory access trusted?

It is -- as a rule of thumb, any accesses emitted by the compiler as part of loading and storing/loading spilled values are considered trusted.

On the other hand, any accesses emitted by the compiler by virtue of translating any Wasm instruction which explicitly interacts with linear memory is considered untrusted and _must_ comply with the semantics dictated by such instruction, like alignment and trapping.


Practically speaking what that means is that if the operand used in this instruction is an incorrect address, this memory access will likely result in a segmentation fault rather than in a Wasm trap.

saulecabrera (approved):

PR #6135, opened 2023-04-03
simple_gvn: recognize commutative operators

Kmeakin:
Normalize instructions with commutative opcodes by sorting the arguments. This means instructions like `iadd v0, v1` and `iadd v1, v0` will be considered identical by GVN and de-duplicated.

~~This is not quite ready to merge: I need someone with more knowledge of the IR to confirm that `fcmp` and `fmin`/`fmax` are commutative in all cases~~

PR #10183, opened 2025-02-04
Standardize the arg order of commutative instructions in pursuit of a higher GVN hit ratio.

erikrose:
This stands on the shoulders of #6135 but reorders only while computing GVN map lookup keys. It doesn't affect the codegenned instructions themselves, which will hopefully dodge the performance regression that led to that PR's reversion.

### DIFF
--- a/cranelift/codegen/src/egraph.rs
+++ b/cranelift/codegen/src/egraph.rs
@@ -116,10 +116,10 @@ pub(crate) enum NewOrExistingInst {
 impl NewOrExistingInst {
     fn get_inst_key<'a>(&'a self, dfg: &'a DataFlowGraph) -> (Type, InstructionData) {
         match self {
-            NewOrExistingInst::New(data, ty) => (*ty, *data),
+            NewOrExistingInst::New(data, ty) => (*ty, data.with_sorted_args()), // sort here too?
             NewOrExistingInst::Existing(inst) => {
                 let ty = dfg.ctrl_typevar(*inst);
-                (ty, dfg.insts[*inst])
+                (ty, dfg.insts[*inst].with_sorted_args())
             }
         }
     }
@@ -222,8 +222,11 @@ where
                 union_find: self.eclasses,
                 value_lists: &self.func.dfg.value_lists,
             };
-            self.gvn_map
-                .insert((ty, self.func.dfg.insts[inst]), opt_value, &gvn_context);
+            self.gvn_map.insert(
+                (ty, self.func.dfg.insts[inst].with_sorted_args()),
+                opt_value,
+                &gvn_context,
+            );
             self.value_to_opt_value[result] = opt_value;
             opt_value
         }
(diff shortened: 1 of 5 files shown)


cfallin:
@erikrose interesting, thanks for this PR. I'm curious if you've observed a benefit (e.g. on our Sightglass suite) in runtime, and also what the impact on compile time is?

My general thought is I'm a little leery of adding extra magic to the egraph optimizer -- it's already a carefully-tuned combo stage of all of the core optimizations, and I'd rather make it simpler than more complex, all other things being equal -- but we can certainly think about ways to clean this up if it shows a benefit.

erikrose:
I benched this patch last night (and it took all night, because I ran the whole "all" suite). It didn't have a huge impact on execution speed either way: 7 benchmarks were faster, 9 slower, and 94 no-difference. Caveats: I ran this on macOS, so it's ARM-only and cycle-count-only—no perf counters. Here are the trimmed-down results, to get a sense of impact:

Ones where the PR was faster than `main`:
```
target/release/libwasmtime_bench_api.dylib is 1.02x to 1.03x faster than tmp/wasmtime_main.dylib!
target/release/libwasmtime_bench_api.dylib is 1.02x to 1.02x faster than tmp/wasmtime_main.dylib!
target/release/libwasmtime_bench_api.dylib is 1.00x to 1.02x faster than tmp/wasmtime_main.dylib!
target/release/libwasmtime_bench_api.dylib is 1.01x to 1.02x faster than tmp/wasmtime_main.dylib!
target/release/libwasmtime_bench_api.dylib is 1.00x to 1.02x faster than tmp/wasmtime_main.dylib!
target/release/libwasmtime_bench_api.dylib is 1.00x to 1.01x faster than tmp/wasmtime_main.dylib!
target/release/libwasmtime_bench_api.dylib is 1.00x to 1.01x faster than tmp/wasmtime_main.dylib!
```

Ones where the PR was slower:
```
tmp/wasmtime_main.dylib is 1.00x to 1.05x faster than target/release/libwasmtime_bench_api.dylib!
tmp/wasmtime_main.dylib is 1.00x to 1.04x faster than target/release/libwasmtime_bench_api.dylib!
tmp/wasmtime_main.dylib is 1.02x to 1.02x faster than target/release/libwasmtime_bench_api.dylib!
tmp/wasmtime_main.dylib is 1.00x to 1.04x faster than target/release/libwasmtime_bench_api.dylib!
tmp/wasmtime_main.dylib is 1.01x to 1.02x faster than target/release/libwasmtime_bench_api.dylib!
tmp/wasmtime_main.dylib is 1.00x to 1.01x faster than target/release/libwasmtime_bench_api.dylib!
tmp/wasmtime_main.dylib is 1.00x to 1.01x faster than target/release/libwasmtime_bench_api.dylib!
tmp/wasmtime_main.dylib is 1.00x to 1.01x faster than target/release/libwasmtime_bench_api.dylib!
tmp/wasmtime_main.dylib is 1.00x to 1.00x faster than target/release/libwasmtime_bench_api.dylib!
```

I expect most of the suite, embodied in .wasm files, already made a trip through LLVM's optimization pipeline and thus had all the GVN opportunities used up already. It may be worth waking this up again if ever we encounter a use case where that's not true. In the meantime, I'll close this. Here are the full benchmark results for posterity: [Full Sightglass Results.txt](https://github.com/user-attachments/files/18698493/Full.Sightglass.Results.txt).

If we ever pick it up again, I would…
- [ ] Push GVN key generation to someplace where we don't have to look at it in the main egraph loop: perhaps in the GVN hash itself.
- [ ] Remove the `IntCompare` and `FloatCompare` `==` arms of the `match`, since they'd be more at home as ISLE rules (and are probably already implemented as such).

Issue #10171, opened 2025-02-03
Incorrect `ref.test` result


apoorvaanand1998:
```WebAssembly
(module
    (type $appNode 
        (struct (field $left anyref)
                (field $right anyref)))
    
    (type $comb (struct (field $asciiTag i32)))

    (func $prob1 (export "prob1") (param $p (ref null $appNode)) (result i32)
        (local.get $p)
        (struct.get $appNode $left)
        (ref.cast (ref null $appNode))
        (struct.get $appNode $right)
        (ref.test (ref null $comb))
    )

    (func $prob2 (export "prob2") (param $p (ref null $appNode)) (result i32)
        (local.get $p)
        (struct.get $appNode $left)
        (ref.cast (ref null $appNode))
        (struct.get $appNode $left)
        (ref.test (ref null $comb))
    )

    (func $main (export "main") (result i32)
        (struct.new $appNode (struct.new $appNode (struct.new $comb (i32.const 67))
                                                  (struct.new $comb (i32.const 73)))
                             (ref.i31 (i32.const 2)))
        (call $prob2)                             
    )
)
```
Hey Wasmtime team! `call`ing `$prob1` returns 1 but `$prob2` returns 0. Why is that? Shouldn't it be reading the `(struct.new $comb (i32.const 67))` which indeed is a `comb`? I've tried reading the GC spec but haven't been able to figure out why the runtime behaves this way.

fitzgen:
This looks like a bug in Wasmtime, I think `prob2` should indeed return `1`.

FWIW, it also returns `1` in Firefox.

fitzgen:
Interestingly enough, removing `$right` makes the test pass.

fitzgen:
Haven't managed to meaningfully reduce it past this:

```wat
(module
  (type $tree (struct (field $left anyref)
                      (field $right anyref)))
  (type $s (struct))
  (func (export "main") (result i32)
    (struct.new $tree
                (struct.new $tree
                            (struct.new $s)
                            (ref.i31 (i32.const 1)))
                (ref.i31 (i32.const 2)))
    struct.get $tree $left
    ref.cast (ref null $tree)
    struct.get $tree $left
    ref.test (ref null $s)
  )
)
```

Interestingly, changing `$right` to be an `i31ref` also makes the test pass.

fitzgen:
And removing one layer of tree nesting also makes the test pass.

fitzgen:
[...]

PR #10191, opened 2025-02-05
Add a bunch of trace logging for CLIF and Wasm-to-CLIF translation

fitzgen:
This is some general trace logging to help with debugging issues like https://github.com/bytecodealliance/wasmtime/issues/10171.

This adds trace logging for:

* `InstructionBuilder` methods
* Switching `FunctionBuilder`s between blocks
* A ton of GC-related Wasm-to-CLIF translation bits

The result is that it is wayyyyyyyy easier to tell what CLIF is generated for what purpose when staring at trace logs, particularly for Wasm GC things where a single Wasm instruction might become many blocks of CLIF.

At the same time, this consolidates some `translate_{array,struct}_get{,_s,_u}` helpers so that there is less code duplication (purely mechanical; should not change any actual translations or instructions we emit) just so that there were fewer places to add trace logs to.


abrown (approved):
Makes sense to me!

PR #10193, opened 2025-02-05
[DWARF] Basic infrastructure for compiling test assets from source

SingleAccretion:
Work in progress.


alexcrichton:
This looks awesome, thank you for working on this!

What do you think about folding this directly into the `test-programs-artifacts` crate? That build script could build both Rust and C/C++ artifacts perhaps? 

> Of course, as with any from-source scheme, determinism is an issue

One mitigating factor here is that we're going to pin  to a particular wasi-sdk in CI. Updating that will be a discrete step and while others may have different versions locally at least for our own CI it should remain deterministic until we update wasi-sdk in which case we'll at least know what was the cause.

SingleAccretion:
> What do you think about folding this directly into the test-programs-artifacts crate?

I considered this, but decided against for the following two reasons:
1) `test-programs-artifacts` can be built with just the standard Rust tools. It is very easy to install the additional WASM pieces if they're missing. In comparison, this C/C++ setup is much more complex - you need to procure the (correct) WASI SDK and need to understand what `WASI_SDK_PATH` is. So it's very "different" from that perspective.
2) It has potentially better build time characteristics. While currently the debug tests are under the all-encompassing `all` target, I want to move them to a separate test target, so that they're quicker to build. In such a scheme, the `--test=all` workflow won't need to build the C/C++ assets at all.

And I couldn't find clear downsides, beside the inherent cost of adding one more crate. But it seems in Rust it is not unusual to have a large number of crates.

SingleAccretion:
Sounds reasonable. I'll work on folding it.

alexcrichton (approved):
👍

PR #10197, opened 2025-02-05
support pulley even if host triplet also works

cpetig:
Previously executing pulley required winch or cranelift even if pulley would be available - because the host triplet was used unconditionally if possible.

### DIFF
--- a/crates/wasmtime/src/config.rs
+++ b/crates/wasmtime/src/config.rs
@@ -2124,6 +2124,11 @@ impl Config {
             return target_lexicon::Triple::pulley_host();
         }
 
+        #[cfg(feature = "pulley")]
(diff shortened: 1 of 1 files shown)


alexcrichton (commented):
Thanks for this! I think though it might be best to ungate the condition above this for `if let Some(target)` perhaps? Previously that was only available with cranelift/winch because it was part of `self.compiler_config` but refactoring lately has made `self.target` always available regardless of crate features and the cfg's haven't really caught up yet. I think that would remove the need for the addition here too?

fitzgen:
I think this was subsumed by #10618 so I'm going to close it. If this was mistaken, please feel free to reopen.

PR #10201, opened 2025-02-06
winch(aarch64): Correct treatment for stores and other trapping ops

saulecabrera:
This commit is one more in the series of executing spec tests for aarch64.

It's mostly a small follow-up to https://github.com/bytecodealliance/wasmtime/pull/10146, in which I omitted contextualizing the memory flags for stores as well as ensuring that the SP is aligned when emitting other trapping instructions like `checked_uadd`.

### DIFF
--- a/winch/codegen/src/isa/aarch64/masm.rs
+++ b/winch/codegen/src/isa/aarch64/masm.rs
@@ -225,14 +225,14 @@ impl Masm for MacroAssembler {
             RegImm::Reg(reg) => reg,
         };
 
-        self.asm.str(src, dst, size);
+        self.asm.str(src, dst, size, TRUSTED_FLAGS);
         Ok(())
     }
 
     fn wasm_store(&mut self, src: Reg, dst: Self::Address, op_kind: StoreKind) -> Result<()> {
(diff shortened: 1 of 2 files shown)


winch/codegen/src/isa/aarch64/masm.rs
```diff
@@ -385,9 +385,11 @@ impl Masm for MacroAssembler {
         size: OperandSize,
         trap: TrapCode,
     ) -> Result<()> {
-        self.add(dst, lhs, rhs, size)?;
-        self.asm.trapif(Cond::Hs, trap);
-        Ok(())
+        self.with_aligned_sp(|masm| {
```
cfallin, line 388: Can we add a comment here noting why the SP alignment is needed? (Is it required to be aligned at all potentially trapping ops?)

cfallin (approved):
Looks good, sorry for the delay!

winch/codegen/src/isa/aarch64/masm.rs
saulecabrera: Done, thanks!

PR #10203, opened 2025-02-06
Winch: Clean up Wast SIMD tests

jeffcharles:
Clean up some of the Winch SIMD Wast tests. I had initially set these up as copies of the spec testsuite with some instructions commented out so there would be some spec test coverage of what I was implementing. But now it looks like the actual spec testsuite can be used for most of them so I'm deleting them. The exception is `simd_multivalue` which doesn't have a corresponding spec test so I've renamed it to remove the leading underscore. For the remaining files, I've uncommented tests that should now pass.


alexcrichton (approved):

alexcrichton:
It's unfortunately buried, but the failure is [here](https://github.com/bytecodealliance/wasmtime/actions/runs/13192553693/job/36828100570)

alexcrichton:
notably:

```
---- Winch/pooling/./tests/misc_testsuite/winch/_simd_load.wast ----
failed to run spec test with default engine

Caused by:
    0: failed directive on ./tests/misc_testsuite/winch/_simd_load.wast:20:1
    1: Compilation error: Instruction not implemented for CPUs without AVX support

---- Winch/./tests/misc_testsuite/winch/_simd_load.wast ----
failed to run spec test with default engine

Caused by:
    0: failed directive on ./tests/misc_testsuite/winch/_simd_load.wast:20:1
    1: Compilation error: Instruction not implemented for CPUs without AVX support


failures:
    Winch/pooling/./tests/misc_testsuite/winch/_simd_load.wast
    Winch/./tests/misc_testsuite/winch/_simd_load.wast
```

jeffcharles:
That makes sense. The tests that were already uncommented probably didn't rely on AVX support, but the tests I uncommented in this PR do rely on it, so the test needs to be added to the unsupported with no AVX list.

Issue #10153, opened 2025-01-29
Wasi_nn PyTorch GPU support


nerdola-de-cartola:
#### Feature

Add GPU support for wasi_nn with PyTorch backend

#### Benefit

Improve performance over nn operations

#### Implementation

The PyTorch backend uses tch, which is a PyTorch wrapper for rust, so the basic implementation would be to add cada as a compilation target when compiling the tch crate.

nerdola-de-cartola:
If you are a contributor to wasmtime/wasi_nn please leave your comments and sugestions. This is my first contribution in this project, so and help would be nice.

nerdola-de-cartola:
I created a fork that solves the issue, at least for me. Maybe is interested for some one else. [My fork](https://github.com/nerdola-de-cartola/wasmtime/tree/wasi-nn-pytorch-gpu)

abrown:
@nerdola-de-cartola, thanks for looking into this. Looking at the top commit from your fork, I think the change makes sense to me (at a glance!). Are you willing to create a PR here that I can review?

nerdola-de-cartola:
Yes, for sure, just tell me what I need to do

abrown:
Can you just create a PR from that fork you reference above? (If you're coming from a version control environment other than GitHub, [these instructions](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request) may help?).

nerdola-de-cartola:
PR [10204](https://github.com/bytecodealliance/wasmtime/pull/10204) crated! Thanks for the guidance.

PR #10204, opened 2025-02-07
Add GPU support for wasi-nn/pytorch

nerdola-de-cartola:
# Solves #10153 

Add CUDA support for feature wasi-nn/pytorch


abrown (approved):
Makes sense to me. Of course, we can't readily test this in CI so if you're a future observer looking at this understand that this added support is experimental. @nerdola-de-cartola, thanks for the contribution!

nerdola-de-cartola:
@abrown thanks for your work! I will be also closing the corresponding issue #10153

PR #10206, opened 2025-02-07
threads: add feature flags

abrown:
This change adds the minimal, internal flags necessary for testing the shared-everything-threads proposal. It chooses to skip the user-facing CLI flags for now and, as suggested, keeps the Cargo `threads` feature as the sole feature flag gating compilation of threads-related code, which will eventually include `shared-everything-threads`.


alexcrichton:
What do you think about using the preexisting `threads` Cargo feature for this? That's mostly avoiding runtime deps of threads/libstd/etc and having that gate both `threads` and `shared-everything-threads` seems reasonable to me

alexcrichton:
Oh and do directly answer this:

> The feedback I'm look for is: what we should be gating with this new Cargo feature?

I wrote up some basic [thoughts here](https://docs.wasmtime.dev/contributing-conditional-compilation.html) but the tl;dr; IMO is "anything that requires a runtime dependency" aka something outside of libcore. In that sense the current `threads` proposal is gated because it requires mutexes/etc to implement atomic notify/wait.  I don't think that `shared-everything-threads` will require any more than `threads`, though, but any nontrivial dependency above the `core` crate should be gated.

(or do you have specific things you're curious about too?)

abrown:
The latest commits remove the CLI flags since we probably don't want to expose that yet to users _and_ adopt @alexcrichton's suggestion to use the `threads` Cargo feature instead of creating a new, specialized `shared-everything-threads` feature. What is now added is all the internal flag bits necessary for running WAST tests, the absence of which is one blocker to #10518. I'm going to alter the PR message to reflect this so we get a more accurate commit message in the Git history.

alexcrichton (approved):
Technically this won't actually work if the test actually used shared-everything-threads in the sense that [this block](https://github.com/bytecodealliance/wasmtime/blob/f1383ec6a09f84e214d78d9ce76fe351d48fac9f/crates/fuzzing/src/generators/config.rs#L279-L302) needs to be updated for fuzzing to propagate the need of shared-everything-threads to `wasmtime::Config`, and that's going through CLI flags which means this'll also have to add `-Wshared-everything-threads`.

That being said such a change is a superset of what you've got here, and this seems fine otherwise.

PR #10221, opened 2025-02-12
Wasm GC: Fix is-null-or-i31ref checks

### LINKED ISSUE #10171: Incorrect `ref.test` result

fitzgen:
In the case where we need to check for either null or i31refs, we were attempting to fold the two checks together with a bitwise-and and a bitmask, but that was incorrect and buggy. We need to actually do the two separate checks and bitwise-or them together.

Fixes #10171

### DIFF
--- a/crates/environ/src/gc.rs
+++ b/crates/environ/src/gc.rs
@@ -25,10 +25,6 @@ use core::alloc::Layout;
 /// Discriminant to check whether GC reference is an `i31ref` or not.
 pub const I31_DISCRIMINANT: u64 = 1;
 
-/// A mask that can be used to check for non-null and non-i31ref GC references
-/// with a single bitwise-and operation.
-pub const NON_NULL_NON_I31_MASK: u64 = !I31_DISCRIMINANT;
-
 /// The size of the `VMGcHeader` in bytes.
 pub const VM_GC_HEADER_SIZE: u32 = 8;
 
--- a/crates/cranelift/src/gc/enabled.rs
+++ b/crates/cranelift/src/gc/enabled.rs
@@ -8,12 +8,13 @@ use cranelift_codegen::{
     ir::{self, condcodes::IntCC, InstBuilder},
 };
 use cranelift_entity::packed_option::ReservedValue;
+use cranelift_entity::Signed;
 use cranelift_frontend::FunctionBuilder;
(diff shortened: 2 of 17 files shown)


crates/environ/src/gc.rs
```diff
@@ -25,10 +25,6 @@ use core::alloc::Layout;
 /// Discriminant to check whether GC reference is an `i31ref` or not.
 pub const I31_DISCRIMINANT: u64 = 1;
```
alexcrichton, line 26: Curiosity: why is this typed `u64`? This discriminant is technically only for 32-bit values right?

alexcrichton (approved):

crates/environ/src/gc.rs
fitzgen: I think it was useful at some point but might not be useful anymore, I’d have to look at uses again

fitzgen-f5: Looks like there's no issue changing it to a u32, so I'll do that in a follow up

PR #10223, opened 2025-02-12
Wasm GC: Fix an incorrect assertion and canonicalize types for runtime usage in ExternType::from_wasmtime

fitzgen:
see commit messages for details


alexcrichton:
Nick and I talked a bit about this in person and the conclusion was that we'll instead try canonicalizing `ModuleTypes` when the signature collection is created instead of later when type-checking happens to avoid threading through the new state

alexcrichton (approved):

PR #10228, opened 2025-02-13
Winch: Implement avgr instructions for x64 with AVX

jeffcharles:
Part of #8093. Implements the following instructions:
- `i8x16.avgr_u`
- `i16x8.avgr_u`


winch/codegen/src/isa/x64/masm.rs
```diff
@@ -2714,6 +2714,17 @@ impl Masm for MacroAssembler {
...
     }
+
+    fn v128_avgr(&mut self, lhs: Reg, rhs: Reg, dst: WritableReg, size: OperandSize) -> Result<()> {
+        self.ensure_has_avx()?;
+
+        match size {
+            OperandSize::S8 | OperandSize::S16 => self.asm.xmm_vpavg_rrr(lhs, rhs, dst, size),
+            _ => unimplemented!(),
+        }
+
```
saulecabrera, line 2725: Could we avoid this match here entirely and instead delegate the operand size invariant checks to the assembler?

saulecabrera: It's currently duplicated in both places, and I'm not sure if such duplication is worth it.

saulecabrera (approved):

PR #10229, opened 2025-02-13
aarch64: Add support for the `extr` instruction

alexcrichton:
This is pattern-matched from `bor` patterns of a specific shape. I found this when doing some benchmarking of Wasmtime on aarch64 and I saw LLVM generating this pattern but Wasmtime didn't. I didn't perform any benchmarking between wasmtime/native though, so I'm just relying on this reducing the number of instructions to probably be a wee bit faster.

### DIFF
--- a/cranelift/codegen/src/isa/aarch64/inst/emit.rs
+++ b/cranelift/codegen/src/isa/aarch64/inst/emit.rs
@@ -759,7 +759,7 @@ impl MachInstEmit for Inst {
                     ALUOp::AddS => 0b00101011_000,
                     ALUOp::SubS => 0b01101011_000,
                     ALUOp::SDiv | ALUOp::UDiv => 0b00011010_110,
-                    ALUOp::RotR | ALUOp::Lsr | ALUOp::Asr | ALUOp::Lsl => 0b00011010_110,
+                    ALUOp::Extr | ALUOp::Lsr | ALUOp::Asr | ALUOp::Lsl => 0b00011010_110,
                     ALUOp::SMulH => 0b10011011_010,
                     ALUOp::UMulH => 0b10011011_110,
                 };
@@ -768,7 +768,7 @@ impl MachInstEmit for Inst {
                 let bit15_10 = match alu_op {
                     ALUOp::SDiv => 0b000011,
                     ALUOp::UDiv => 0b000010,
-                    ALUOp::RotR => 0b001011,
+                    ALUOp::Extr => 0b001011,
                     ALUOp::Lsr => 0b001001,
                     ALUOp::Asr => 0b001010,
                     ALUOp::Lsl => 0b001000,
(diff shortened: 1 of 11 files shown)


cranelift/codegen/src/isa/aarch64/inst/emit.rs
```diff
@@ -906,6 +906,7 @@ impl MachInstEmit for Inst {
                     ALUOp::OrrNot => 0b001_01010001,
                     ALUOp::EorNot => 0b010_01010001,
                     ALUOp::AndNot => 0b000_01010001,
+                    ALUOp::RotR => 0b000_10011100,
```
cfallin, line 909: It looks like (at least by my reading) `rotr` is an alias for / special case of `extr`, not the other way around, so perhaps we could rename the `ALUOp`?

cranelift/codegen/src/isa/aarch64/lower.isle
```diff
@@ -1426,6 +1426,29 @@
...
+                                 (ushr y (u8_from_iconst ys)))))
+  (if-let shift (a64_extr_shift ty xs ys))
+  (a64_extr ty x y shift))
+(rule 5 (lower (has_type ty (bor (ushr y (u8_from_iconst ys))
+                                 (ishl x (u8_from_iconst xs)))))
+  (if-let shift (a64_extr_shift ty xs ys))
+  (a64_extr ty x y shift))
+
+;; Helper in Rust to test whether the pair of shifts of `u8` add up to the type
+;; width specified. Optionally returns the second value as `ImmShift`.
```
cfallin, line 1448: Is it possible to build this out of the add/eq/... primitives we have? Nothing against custom ctors when needed, just curious if we can reduce our FFI surface somewhat though.

cranelift/codegen/src/isa/aarch64/lower.isle
```diff
@@ -1426,6 +1426,29 @@
...
+;; starting from an immediate index. This is pattern-matched here as a `bor` of
+;; the high/low halves of two values shifted around.
+;;
+;; The immediate used for the `extr` instruction itself is the N for the
+;; shift-right. Two patterns are used here to detect either ordering of the
+;; `bor`.
+(rule 5 (lower (has_type ty (bor (ishl x (u8_from_iconst xs))
+                                 (ushr y (u8_from_iconst ys)))))
+  (if-let shift (a64_extr_shift ty xs ys))
+  (a64_extr ty x y shift))
```
cfallin, line 1441: Can we write the formula in the comment too?

```
;; (x << xs) | (y >> ys) if (xs + ys == widthof(ty)) => extr(x, y, ys)
```

cfallin (approved):
Good find, thanks!

cranelift/codegen/src/isa/aarch64/lower.isle
alexcrichton: Good point, I also prefer fewer ISLE FFI functions!

PR #10233, opened 2025-02-14
x64: Add support for the `shld` instruction

alexcrichton:
This commit is similar to #10229 except that it uses the x64 `shld` instruction. This instruction was not previously supported by Cranelift so I opted to add support via the new `cranelift-assembler-x64` crate instead of trying to fit it into the existing `emit.rs`/`MInst` infrastructure. This was a good learning experience for myself and was also fun to do!

### DIFF
--- a/cranelift/assembler-x64/meta/src/dsl/encoding.rs
+++ b/cranelift/assembler-x64/meta/src/dsl/encoding.rs
@@ -195,7 +195,10 @@ impl Rex {
 
         if self.opcodes.prefix.contains_66() {
             assert!(
-                operands.iter().all(|&op| op.location.bits() == 16),
+                operands
+                    .iter()
+                    .all(|&op| matches!(op.location.kind(), OperandKind::Imm(_) | OperandKind::FixedReg(_))
+                        || op.location.bits() == 16),
                 "when we encode the 66 prefix, we expect all operands to be 16-bit wide"
             );
         }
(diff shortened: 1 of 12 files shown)


alexcrichton:
I'm opening this as a draft PR because https://github.com/bytecodealliance/wasmtime/pull/10232 should land first and this should rebase on top of that. In talking with Andrew I temporarily came up with the "solution" in this PR but our conclusion was that #10232 was the better way to go in terms of robustly supporting this encoding pattern in x64 instructions.

alexcrichton:
(also cc @abrown)

cranelift/assembler-x64/meta/src/generate/format.rs
```diff
@@ -15,15 +15,12 @@ impl dsl::Format {
     /// (TODO).
     #[must_use]
     pub fn generate_att_style_operands(&self) -> String {
-        let mut ordered_ops: Vec<_> = self
+        let ordered_ops: Vec<_> = self
             .operands
             .iter()
+            .rev()
```
abrown, line 21: Huh... let's try this but I'm not sure.

cranelift/assembler-x64/meta/src/dsl/encoding.rs
```diff
@@ -332,7 +335,7 @@ impl From<[u8; 3]> for Opcodes {
     fn from(bytes: [u8; 3]) -> Opcodes {
         let [a, b, c] = bytes;
         match (LegacyPrefix::try_from(a), b, c) {
-            (Ok(prefix), 0x0f, primary) => Opcodes { prefix, escape: false, primary, secondary: None },
+            (Ok(prefix), 0x0f, primary) => Opcodes { prefix, escape: true, primary, secondary: None },
```
abrown, line 338: Good catch!

abrown (approved):
Makes sense to me!

cranelift/assembler-x64/meta/src/generate/format.rs
alexcrichton: Heh agreed! I figure we'll need to add more heuristics here eventually to figure out how to print and still match att/capstone ordering.

PR #10235, opened 2025-02-14
Update wasm-tools crates to 226

alexcrichton:
This commit updates the wasm-tools crates to the 226 track of versions. The most notable change here is the removal and refactoring within wasm-tools of support for multiple and named returns in the component model along with the support for 33+ flags in the `flags` type. This has been gated for some time now and it's just now being removed now that it's been awhile an no issues have been raised.


alexcrichton:
I'll note that this is a draft because I  haven't actually published anything yet on the wasm-tools side of things.

alexcrichton:
Ok should be ready to go now.

pchickey (approved):

PR #10237, opened 2025-02-14
asm: add implementations for more ALU operations

abrown:
This adds DSL definitions in the assembler _and_ uses the new instructions for ISLE lowering for the following instructions:
- `add`
- `adc`
- `or`
- `sub`
- `sbb`
- `xor`

The original `AluRmiROpcode` variants are not yet gone; perhaps in a future pass.

### DIFF
--- a/cranelift/assembler-x64/meta/src/generate/format.rs
+++ b/cranelift/assembler-x64/meta/src/generate/format.rs
@@ -88,7 +88,8 @@ impl dsl::Format {
             [FixedReg(dst), Imm(_)] => {
                 // TODO: don't emit REX byte here.
                 fmtln!(f, "let {dst} = {};", dst.generate_fixed_reg().unwrap());
-                fmtln!(f, "let digit = 0;"); // No digit for this pattern.
+                assert_eq!(rex.digit, None, "we expect no digit for operands: [FixedReg, Imm]");
+                fmtln!(f, "let digit = 0;");
                 fmtln!(f, "rex.emit_two_op(buf, digit, {dst}.enc());");
             }
             [RegMem(dst), Imm(_)] => {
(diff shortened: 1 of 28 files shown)


cranelift/assembler-x64/meta/src/generate/format.rs
```diff
@@ -88,21 +88,20 @@ impl dsl::Format {
             [FixedReg(dst), Imm(_)] => {
                 // TODO: don't emit REX byte here.
                 fmtln!(f, "let {dst} = {};", dst.generate_fixed_reg().unwrap());
-                fmtln!(f, "let digit = 0x{:x};", rex.digit);
+                fmtln!(f, "let digit = 0;"); // No digit for this pattern.
```
cfallin, line 91: Should we `assert_eq!(rex.digit, None)` here?

cfallin (approved):
Looks good, thanks!

cranelift/assembler-x64/meta/src/generate/format.rs
abrown: Yeah, good point.

abrown:
@cfallin, you might want to take another glance at 9b8aecb; it looks like we're emitting a slightly longer encoding somewhere? Not sure if that's ok...

abrown:
Ok, looking into this more, we had some extra 8-bit lowering rules up in `cranelift-codegen` that I hadn't ported to the new assembler--no new instructions needed, just some extra ISLE rules. This undoes any extra bytes added in prior disassembly tests and should match what we had in `cranelift-codegen` previously.

cfallin:
Updates look good -- thanks for tracking down that divergence!

Issue #10089, opened 2025-01-23
Implement wasi-tls


badeend:
[wasi-tls](https://github.com/WebAssembly/wasi-tls) has [recently been accepted](https://github.com/WebAssembly/WASI/pull/639) as a phase 1 proposal.

We'd like to start implementing this in wasmtime. There already exists some prior efforts:
- @dicej created a standalone WASI host with TLS support at: https://github.com/dicej/dotnet-wasi-tls
    - This implements a minimal subset of the [draft spec](https://github.com/WebAssembly/wasi-sockets/pull/104).
    - Uses the `native-tls` crate.
- @jsturtevant modified the .NET runtime to use Joel's implementation at: https://github.com/dotnet/runtime/compare/main...jsturtevant:runtime:wasi-tls-2
    - This successfully runs their SqlClient.
- Based on Joel's work, I (@badeend) started integrating it into wasmtime itself: https://github.com/badeend/wasmtime/tree/wasi-tls
    - This too is just a minimal subset of the [draft spec](https://github.com/WebAssembly/wasi-sockets/pull/104). Although the WIT in this branch has evolved slightly.
    - Uses the `rustls` crate.
    - The changes are part of the wasi-sockets interface, which is not the right place.
    - And some boilerplate:
        - threading through a new `tls` CLI flag
        - Change `rustls` from a `wasi-http`-only dependency to a workspace dependency.

---

My suggestion is to add a new standalone `wasi-tls` crate:
- under the `/crates` folder, similar to the other proposals.
- that uses the WIT file from [this PR](https://github.com/WebAssembly/wasi-tls/pull/1)
- that targets WASI v0.2 for the time being. Work on a v0.3 interface can happen in parallel in the future.
- that is completely experimental and therefore behind a `tls` flag.
- uses `rustls` for its implementation? (up for discussion, see next comment)

Thoughts?

---

CC @dicej @jsturtevant

badeend:
I did an inventory of the two most popular rust TLS crates to see how suitable they are to implement the [draft spec](https://github.com/WebAssembly/wasi-sockets/pull/104):

**`rustls`:**
- doesn't support post-handshake authentication.
- supports programmatic client certificate selection, but the callback isn't async.
- supports programmatic peer certificate verification, but the callback isn't async.

**`native-tls`:**
- does not support TLS1.3 (yet)
- uses MacOS Secure Transport, which is deprecated and will likely never receive TLS1.3 support.
[...]

PR #10249, opened 2025-02-19
Initial implementation of Wasi-tls (Transport Layer Security) 

jsturtevant:
fixes: https://github.com/bytecodealliance/wasmtime/issues/10089

This adds a crate that provides the Wasmtime host implementation for the [wasi-tls](https://github.com/WebAssembly/wasi-tls) API.
The [wasi-tls](https://github.com/WebAssembly/wasi-tls) world allows WebAssembly modules to perform SSL/TLS operations,
such as establishing secure connections to servers. 

This initially implements it with rustls which was already included in the repository.

There are a few items that still need to be resolved but are not specially related to the implementation but wanted to start the review to make sure things moving in the correct direction.

- [x] add a `close-notify` function to the wasi-tls wit in the proposal to properly close down the https://github.com/WebAssembly/wasi-tls/pull/5 
- [x] resolve importing stable types into unstable features (https://github.com/bytecodealliance/wasm-tools/issues/1995). Write now I have a workaround on a branch that is being used in rust cargo patch.

Thanks goes out to @badeend  and @dicej as much of this was based on their prior work.


badeend:
Looking great!

CI fails because Clippy being Clippy, and I suspect the build fails because there's some missing conditional compilation.

Aside from that, I;ve left behind some remarks:

crates/test-programs/src/lib.rs
```diff
@@ -12,15 +13,17 @@ wit_bindgen::generate!({
...
             include wasi:config/imports@0.2.0-draft;
             include wasi:keyvalue/imports@0.2.0-draft;
+            include wasi:tls/imports@0.2.3;
         }
     ",
     path: [
         "../wasi-http/wit",
         "../wasi-config/wit",
         "../wasi-keyvalue/wit",
+        "../wasi-tls/wit/world.wit",
```
jsturtevant, line 23: The reason for this being the wit file directly is https://github.com/bytecodealliance/wasm-tools/issues/1996#issuecomment-2640748051

Cargo.toml
```diff
@@ -566,3 +573,18 @@ opt-level = 's'
...
+wat =  { git = "https://github.com/jsturtevant/wasm-tools.git", branch="unstable-refrence-stable" }
+wast =  { git = "https://github.com/jsturtevant/wasm-tools.git", branch="unstable-refrence-stable" }
+wasmprinter =  { git = "https://github.com/jsturtevant/wasm-tools.git", branch="unstable-refrence-stable" }
+wasm-encoder =  { git = "https://github.com/jsturtevant/wasm-tools.git", branch="unstable-refrence-stable" }
+wasm-smith =  { git = "https://github.com/jsturtevant/wasm-tools.git", branch="unstable-refrence-stable" }
+wasm-mutate =  { git = "https://github.com/jsturtevant/wasm-tools.git", branch="unstable-refrence-stable" }
+wit-parser =  { git = "https://github.com/jsturtevant/wasm-tools.git", branch="unstable-refrence-stable" }
+wit-component =  { git = "https://github.com/jsturtevant/wasm-tools.git", branch="unstable-refrence-stable" }
+wasm-wave =  { git = "https://github.com/jsturtevant/wasm-tools.git", branch="unstable-refrence-stable" }
+wasm-metadata =  { git = "https://github.com/jsturtevant/wasm-tools.git", branch="unstable-refrence-stable" }
```
jsturtevant, line 590: this branch contains the temp fix for https://github.com/bytecodealliance/wasm-tools/issues/1995

crates/wasi-tls/src/lib.rs
```diff
@@ -0,0 +1,637 @@
...
+    Pending(Pin<Box<dyn Future<Output = T> + Send>>),
+    Closed,
+}
+
+struct TcpStreams {
+    input: StreamState<BoxInputStream>,
+    output: StreamState<BoxOutputStream>,
+}
+
+impl AsyncWrite for TcpStreams {
```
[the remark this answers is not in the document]
jsturtevant: correct, not specific to TLS it just enables passing wasi-io streams to libraries that require those traits (i.e. in this case the `connect()` method for rustls)

crates/wasi-tls/src/lib.rs
```diff
@@ -0,0 +1,637 @@
...
+    async fn ready(&mut self) {
+        match &self.0 {
+            StreamState::Pending(_) => (),
+            StreamState::Ready(_) | StreamState::Closed => return,
+        }
+
+        let StreamState::Pending(future) = mem::replace(&mut self.0, StreamState::Closed) else {
+            unreachable!()
+        };
+        self.0 = StreamState::Ready(future.await);
```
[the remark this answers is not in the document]
jsturtevant: Will take a look, is there an example of how to test that?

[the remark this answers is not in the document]
badeend: To reproduce, you could try adjusting the timeout of `blocking_finish` in `crates/test-programs/src/tls.rs` to something non-0, but very small, so that both pollables are polled at least once, but the monotonic_clock pollable will finish first. After that, I suspect that the ClientHandshake::finish pollable will never resolve.

crates/wasi-tls/src/lib.rs
```diff
@@ -0,0 +1,637 @@
...
+    opts: &mut LinkOptions,
+    f: impl Fn(&mut T) -> WasiTlsCtx + Send + Sync + Copy + 'static,
+) -> Result<()> {
+    generated::types::add_to_linker_get_host(l, &opts, f)?;
+    Ok(())
+}
+///  Represents the ClientHandshake which will be used to configure the handshake
+pub struct ClientHandShake {
+    server_name: String,
+    streams: TcpStreams,
```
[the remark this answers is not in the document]
jsturtevant: updated

crates/wasi-tls/src/lib.rs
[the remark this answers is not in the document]
jsturtevant: you are indeed correct.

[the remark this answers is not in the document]
jsturtevant: I've pushed an update for this, I tested it locally but working getting a test written for this scenario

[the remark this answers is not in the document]
jsturtevant: added test that covers this in https://github.com/bytecodealliance/wasmtime/pull/10249/commits/f13938ee922ef11f1893957d77465d46ae442c14

crates/wasi-http/Cargo.toml
```diff
@@ -31,9 +31,9 @@ wasmtime = { workspace = true, features = ['component-model'] }
 
 # The `ring` crate, used to implement TLS, does not build on riscv64 or s390x
 [target.'cfg(not(any(target_arch = "riscv64", target_arch = "s390x")))'.dependencies]
```
alexcrichton, line 33: For this I might recommend adding a commit with `prtest:full` somewhere in the commit message to ensure that this can get past CI. I vaguely recall that rustls and/or ring and/or something doesn't compile on riscv64/s390x, hence the conditional here. That'll make the integration of `wasmtime-wasi-tls` into the CLI a bit hairier if so, but probably still workable.

alexcrichton (commented):
This looks great to me 👍 

In terms of testing, would it be possible to also add a test (or multiple) that deal with the server-side as well. Ideally there'd either be a pre-generated cert check-in to the repo that they use or something generated at test-time perhaps per-test. That might help enable removing the dependency on `example.com` as well which has been flaky in the past, and while it's not the end of the world to have some flakiness it'd be best to insulate ourselves and have all the networking be purely local. Although I realize it's also good to test "is the system cert store actually used?" which necessarily requires external communication basially.

It might also be interesting, if local certs work, to test a variety of configurations related to invalid certificates to ensure that they do indeed all return an error instead of succeeding the authentication and such.

jsturtevant:
> In terms of testing, would it be possible to also add a test (or multiple) that deal with the server-side as well.

Just to clarify, you mean to implement a server similar to what is done on [wasi-http](https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasi-http/tests/all/http_server.rs) but with certs?  We are going to need this for more advance scenarios later so works for me.  I'll get started on it here

alexcrichton:
That's one route yeah, another being something similar to [what wasi-sockets does](https://github.com/bytecodealliance/wasmtime/blob/2ad3bea6ae1762eb5491051a425fcb8ff881811e/crates/test-programs/src/bin/preview2_tcp_streams.rs#L134-L157) where the client/server are both purely in-wasm and doens't require any external integration. The latter is a bit nicer from a "share this test suite in more places" perspective as well, but regardless I think either way is fine in the long run

jsturtevant:
sounds good. We didn't add the server side parts to the wasi-tls spec yet but it shouldn't be to hard to do a minimal version as well now that the client side is done.  @badeend wdyt?

alexcrichton:
Oh I didn't look too closely at the tls bits here but if the server parts aren't implemented yet I don't mean to add more work here, so I think it's totally reasonable to land without that and plan to add more tests later too.

jsturtevant:
> (aka landing the wasm-tools fixes and integrating that here).

I tried pulling the fix  (https://github.com/bytecodealliance/wasm-tools/pull/2076) in from the main branch but looks like due to https://github.com/bytecodealliance/wasm-tools/pull/2073, there were a few missing cases that need to be added to wit-bindgen and here.  Would it be helpful to open PR's to stub those out or is there work being done on them?  No rush/pressure, just want to be helpful if I can.

alexcrichton:
Good point! I started the update a few days ago but didn't publish it -- https://github.com/bytecodealliance/wasmtime/pull/10314 -- I'll work on landing that once all the wasm-tools changes are in, but if you want to rebase on that it can give a preview of what things are likely to look like

crates/wasi-http/Cargo.toml
jsturtevant: made some updates to address this.  but now getting `register mapping is currently only implemented for x86_64` on some of the other platforms which I were working previously.  I am not sure what this means yet, as I wouldn't have expected the changes to make a difference.

jsturtevant: Looked a bit more at this. It wasn't passing prior to my changes.  I am not sure what causes this, if it is expected, or how to debug it.

@alexcrichton could you give me a pointer to what might be going wrong here?

alexcrichton: Ah that's a bad error, but it bottoms out in `config.debug_info(true);` which isn't supported on Pulley. I'll go try and improve that error.

jsturtevant: Thanks! Removing that line fixed the issue.

Cargo.toml
jsturtevant: fixed in https://github.com/bytecodealliance/wasm-tools/releases/tag/v1.227.0

jsturtevant:
@badeend @alexcrichton We've merged all the outstand issues and I've rebased to bring in all the changes.  I think this is ready for a final review. Thanks!

alexcrichton (approved):
I pushed a fix for the CI failure, and I double-checked and it looks like everything might actually work on s390x/riscv64 nowadays so I've updated to drop all the #[cfg] related to that. Otherwise looks good 👍

Issue #10248, opened 2025-02-19
Tracking issue: Stack switching proposal implementation status


frank-emrich:
This is a tracking issue for the in-progress implementation of the [Wasm stack switching proposal](https://github.com/WebAssembly/stack-switching). The explainer document for the proposal is [here](https://github.com/WebAssembly/stack-switching/blob/main/proposals/stack-switching/Explainer.md).

### Progress of initial implementation

This just tracks what's happening in the sequence of PRs that's providing an initial implementation of the proposal. To get an idea of the extent of the initial implementation, see draft PR #10177.
- [ ] Support for tags
- [ ] Type definitions for stack switching support
- [ ] Runtime support for stack switching
- [ ] Compiling stack switching instructions to CLIF

### Limitations of initial implementation
These are things that won't be part of the initial implementation:

- Support for ISAs other than x64
- Support for `resume.throw` instructions (requires exception handling)
- Support for deallocating continuations
- Integration with GC (i.e., allowing continuation references to appear in GC objects, such as arrays an structs)
- Windows support
- Pulley support
- Winch support

dhil:
Codegen was added by #11003.

smarcd:
Two embedder-API requirements, in case they're useful input while the shape of that API is still open (I saw @fitzgen's "add an item to the meta task list for the embedder API" note on #10388 — doesn't look like that's been fleshed out yet). Both come up for any embedder targeting real POSIX/C compatibility on top of Wasm:

**1. `setjmp`/`longjmp`**

Wasm has no native non-local control transfer, so this currently has to be faked — either via Asyncify's whole-program stack-copying transform, or an equivalent host-driven suspend/resume shim. That's a real, generalizable cost (code size, compile time, and runtime overhead on every function that *might* be unwound through, not just the ones that are). WASIX is a concrete existing example of this: it implements POSIX `longjmp` (and `fork`) on Wasm this way, via Asyncify in Wasmer, as far as I'm aware the only runtime to have actually shipped that spec so far — which suggests the cost is being paid for real, not just hypothetically. It'd be good to confirm, as the embedder API takes shape, that its suspend/resume primitives are actually sufficient to implement `setjmp`/`longjmp`-style non-local jumps directly on top of continuations — i.e. that this is a design goal, not just an accident of what the API happens to support.

**2. Duplicating a live continuation (fork-style semantics)**

[...]

PR #10251, opened 2025-02-20
Exception and control tags

dhil:
This patch is a subset of PR #10177. It adds support for tags in Wasmtime. Tags are used by the exception handling and stack switching proposals for typing and matching transfers of control.

This patch does not implement component model or C API support for tags. Issue #10252 tracks these two things.

CC @frank-emrich

### DIFF
--- a/crates/environ/src/vmoffsets.rs
+++ b/crates/environ/src/vmoffsets.rs
@@ -676,16 +676,10 @@ impl<P: PtrSize> VMOffsets<P> {
         0 * self.pointer_size()
     }
 
-    /// The offset of the `vmctx` field.
-    #[inline]
-    pub fn vmtag_import_vmctx(&self) -> u8 {
-        1 * self.pointer_size()
-    }
-
     /// Return the size of `VMTagImport`.
     #[inline]
     pub fn size_of_vmtag_import(&self) -> u8 {
-        2 * self.pointer_size()
+        1 * self.pointer_size()
     }
 }
 
--- a/crates/environ/src/vmoffsets.rs
+++ b/crates/environ/src/vmoffsets.rs
@@ -35,8 +35,8 @@
 // }
 
 use crate::{
-    DefinedGlobalIndex, DefinedMemoryIndex, DefinedTableIndex, FuncIndex, FuncRefIndex,
-    GlobalIndex, MemoryIndex, Module, OwnedMemoryIndex, TableIndex,
+    DefinedGlobalIndex, DefinedMemoryIndex, DefinedTableIndex, DefinedTagIndex, FuncIndex,
+    FuncRefIndex, GlobalIndex, MemoryIndex, Module, OwnedMemoryIndex, TableIndex, TagIndex,
 };
 use cranelift_entity::packed_option::ReservedValue;
 
@@ -69,6 +69,8 @@ pub struct VMOffsets<P> {
     pub num_imported_memories: u32,
     /// The number of imported globals in the module.
     pub num_imported_globals: u32,
+    /// The number of imported tags in the module.
+    pub num_imported_tags: u32,
     /// The number of defined tables in the module.
     pub num_defined_tables: u32,
     /// The number of defined memories in the module.
@@ -77,6 +79,8 @@ pub struct VMOffsets<P> {
     pub num_owned_memories: u32,
     /// The number of defined globals in the module.
     pub num_defined_globals: u32,
+    /// The number of defined tags in the module.
+    pub num_defined_tags: u32,
     /// The number of escaped functions in the module, the size of the func_refs
     /// array.
     pub num_escaped_funcs: u32,
@@ -86,10 +90,12 @@ pub struct VMOffsets<P> {
     imported_tables: u32,
     imported_memories: u32,
     imported_globals: u32,
+    imported_tags: u32,
     defined_tables: u32,
     defined_memories: u32,
     owned_memories: u32,
     defined_globals: u32,
+    defined_tags: u32,
     defined_func_refs: u32,
     size: u32,
 }
@@ -148,6 +154,12 @@ pub trait PtrSize {
     #[inline]
     fn size_of_vmglobal_definition(&self) -> u8 {
         16
+    }
+
+    /// Return the size of `VMTagDefinition`.
+    #[inline]
+    fn size_of_vmtag_definition(&self) -> u8 {
+        4
     }
 
     // Offsets within `VMRuntimeLimits`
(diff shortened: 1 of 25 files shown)


crates/wasmtime/src/config.rs
```diff
@@ -1068,6 +1068,30 @@ impl Config {
...
+    /// This feature is `false` by default.
+    ///
+    /// # Errors
+    ///
+    /// [proposal]: https://github.com/webassembly/stack-switching
+    pub fn wasm_stack_switching(&mut self, enable: bool) -> &mut Self {
+        // FIXME(dhil): Once the config provides a handle
+        // for turning on/off exception handling proposal support,
+        // this ought to only enable stack switching.
+        self.wasm_feature(
```
alexcrichton, line 1088: I added a "miscellaneous TODO items" section on https://github.com/bytecodealliance/wasmtime/issues/10248 and added this there too (trying to ensure FIXME's like this are written down in issues)

alexcrichton (approved):
Thank you again for being willing to split up the PR, it's  very helpful!

crates/environ/src/vmoffsets.rs
```diff
@@ -35,8 +35,8 @@
 // }
```
fitzgen, line 35: When you have a moment, can you update the above struct pseudo-definition to reflect the changes from this PR? Thanks!

crates/wasmtime/src/runtime/types.rs
```diff
@@ -2457,6 +2466,34 @@ impl GlobalType {
...
+#[derive(Debug, Clone, Hash)]
+pub struct TagType {
+    ty: FuncType,
+}
+
+impl TagType {
+    /// Creates a new global descriptor of the specified type.
+    pub fn new(ty: FuncType) -> TagType {
+        TagType { ty }
+    }
```
fitzgen, line 2484: This should validate that the function type only has parameters and not results, right? I'll add an item to https://github.com/bytecodealliance/wasmtime/pull/10177 for this, but feel free to remove it if my understanding is incorrect.

crates/wasmtime/src/runtime/vm/vmcontext.rs
```diff
@@ -263,6 +263,39 @@ mod test_vmglobal_import {
...
 }
 
+/// The fields compiled code needs to access to utilize a WebAssembly
+/// tag imported from another instance.
+#[derive(Debug, Copy, Clone)]
+#[repr(C)]
+pub struct VMTagImport {
+    /// A pointer to the imported tag description.
+    pub from: VmPtr<VMTagDefinition>,
+}
```
fitzgen, line 273: Do we actually need a separate `VMTagImport` from `VMTagDefinition`?

Because each tag is immutable (right?) we can inline the original `VMTagDefinition` for imported tags, removing the distinction between imported vs defined tags at the vmctx level. And `VMTagDefinition` happens to be smaller than a pointer on 64-bit architectures, so between that and removing an indirection, it is a double win.

crates/wasmtime/src/runtime/types.rs
dhil: Only for exception tags. But currently we don't have an exceptions feature. When stack switching is enabled the results can be nonempty.

crates/wasmtime/src/runtime/vm/export.rs
```diff
@@ -106,3 +110,22 @@ impl From<ExportGlobal> for Export {
...
 }
+
+/// A tag export value.
+#[derive(Debug, Clone)]
+pub struct ExportTag {
+    /// The address of the global storage.
+    pub definition: NonNull<VMTagDefinition>,
+    /// The global declaration, used for compatibility checking.
+    pub tag: Tag,
+}
```
fitzgen, line 121: Similarly it seems like we could just inline the tag's `VMSharedTypeIndex` here and avoid an indirection.

crates/wasmtime/src/runtime/vm/vmcontext.rs
dhil: The key thing as far as I am aware is that tags are nominal. So we need to equip them with an unique identity, one per module instance.

crates/environ/src/vmoffsets.rs
frank-emrich: I actually added those in a follow up PR [here](https://github.com/bytecodealliance/wasmtime/pull/10265/files#diff-6419ff9508ee5b42d1de13f9509eed5a52ae455d696dedc2b74e090e279757dfR32-R36) when I was touching the VMContext again.

crates/wasmtime/src/runtime/types.rs
fitzgen: Gotcha, thanks!

crates/wasmtime/src/runtime/vm/vmcontext.rs
fitzgen: Ah, okay, that is good to know, I didn't realize they were nominal.

crates/wasmtime/src/runtime/vm/export.rs
dhil: I think that'd be unsound. Consider this example:
```wast
(module $m1
  (type $ft (func))
  (tag $foo (export "foo"))
  (func (export "catch-foo") (param $f (ref $ft))
    (block $on_foo
      (try_table (catch $foo $on_foo) (call_ref $ft (local.get $f)))
      (unreachable)
    ))
)
(register "m1")

(module $m2
  (type $ft (func))

  (func $catch-foo (import "m1" "catch-foo") (param (ref $ft)))

  (tag $foo (import "m1" "foo"))
  (tag $bar)

  (func $throw_foo
    (throw $foo))

  ;; Handle the imported foo.
  (func $f (export "main-1")
    (block $on_foo
      (try_table (catch $foo $on_foo) (call $g))
      (unreachable)
    ))

  ;; Handle this module's bar.
  (func $g
    (block $on_bar
      (try_table (catch $bar $on_bar) (call $throw_foo))
      (unreachable)
    )
    (unreachable))
   (elem declare func $g)

  ;; Handle the imported foo with the imported function
  (func $h (export "main-2")
    (call $catch-foo (ref.func $g)))
)
(register "m2")
(assert_return (invoke "main-1"))
(assert_return (invoke "main-2"))
```
It is important here that `foo` and `bar` are different. And it is important that the identity of `foo` is preserved after the import.

fitzgen: Thanks for the example, I made this comment before I realized that tags were nominal.

PR #10256, opened 2025-02-20
Don’t panic when http `handle` function returns before invoking `set`

primoly:
`wasmtime serve` currently presumes that a failure to call `response-outparam::set` inside `handle` means the component trapped. But it can also happen by returning before invoking `set`. This PR accommodates for that.


alexcrichton:
Thanks! Could you add a test for this as well?

alexcrichton:
I think the difference between binaries is `panic=abort` as the `wasmtime` executable has that set in CI which is probably what causes an abort.

For the test here could the output of `finish()` be inspected to ensure that "panicked" doesn't show up? Or basically assert that it has the expected output? (ideally empty or just a warning or similar)

alexcrichton:
@primoly would you like some help in debugging the CI failure?

PR #10263, opened 2025-02-20
winch(aarch64): Sync SP with SSP when claiming stack

saulecabrera:
This commit is a follow-up to
https://github.com/bytecodealliance/wasmtime/pull/10146  and represents another step toward fixing the remaining issues discovered through spec tests in the same vein as https://github.com/bytecodealliance/wasmtime/pull/10201

Specifically, this commit ensures that the stack pointer is always in sync with the shadow stack pointer. The previous approach was lossy because it only performed the sync when reserving stack space. While this approach worked in some cases, it failed to account for situations where the shadow stack pointer might be adjusted and aligned for calls. As a result, the stack pointer could become unaligned when claiming stack space, leading to issues at call sites.

It is possible to avoid the unconditional move and perform it only when alignment is needed, i.e., at call sites and when the real stack pointer is unaligned. However, as of now, the simplest solution is to always perform the sync, which integrates best with the current infrastructure.

### DIFF
--- a/tests/disas/winch/aarch64/br/as_br_if_cond.wat
+++ b/tests/disas/winch/aarch64/br/as_br_if_cond.wat
@@ -16,5 +16,6 @@
 ;;       stur    x1, [x28]
 ;;       add     x28, x28, #0x10
 ;;       mov     sp, x28
+;;       mov     sp, x28
 ;;       ldp     x29, x30, [sp], #0x10
 ;;       ret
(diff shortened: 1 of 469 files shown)


saulecabrera:
The diff is considerably large, since this change touches all of the disassembly tests for aarch64. The relevant changes are in the first commit https://github.com/bytecodealliance/wasmtime/pull/10263/commits/97d7f740b8180358fd1aef63a5f12cc35d59374c

alexcrichton (approved):

PR #10266, opened 2025-02-21
ci: use semicolons for path separation on Windows

abrown:
The new assembler uses `:` to separate a list of paths containing generated code. We received a report that on Windows this could lead to build issues; this change uses `;` for Windows instead.

prtest:full

### DIFF
--- a/cranelift/assembler-x64/build.rs
+++ b/cranelift/assembler-x64/build.rs
@@ -13,12 +13,16 @@ fn main() {
         meta::generate_isle_definitions(out_dir.join("assembler-definitions.isle")),
     ];
 
+    #[cfg(not(target_os = "windows"))]
+    const SEPARATOR: &str = ":";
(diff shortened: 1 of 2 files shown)


abrown:
It's still pretty unclear to me how this could have got past our Windows CI...

abrown:
Perhaps something like the following is "fixing up" the paths?

https://github.com/bytecodealliance/wasmtime/blob/692490b81176010a084a01797a2e2bece912c0a1/.github/workflows/main.yml#L29-L32

cfallin (approved):

alexcrichton:
This may technically still be broken in cross compilation scenarios due to the os of the build script differing from the os of the target. To sidestep all encoding issues could the array be generated in a rust file in the build script which is included with a macro?

alexcrichton:
Superseded by https://github.com/bytecodealliance/wasmtime/pull/10267

PR #10271, opened 2025-02-21
Winch: Add min and max for SIMD floats for x64 with AVX

jeffcharles:
Part of #8093. Adds support for the following instructions:
- `f32x4.min`
- `f64x2.min`
- `f32x4.max`
- `f64x2.max`

I explored putting the float implementations for min and max in the same helper method given how similar they are but the flow is different enough between them that I found it hard to write accurate comments and track what values were in which registers when debugging spec test failures so I've opted to just factor out the NaN canonicalization logic which is identical.


saulecabrera (approved):
LGTM, thanks.

PR #10273, opened 2025-02-21
asm: sse orpd implementation

rahulchaphalkar:
This PR adds `xmm` registers for SSE instructions, and 1 SSE instruction `orpd`. Includes 2 new locations `xmm` and `rm128` and the relevant logic.


cranelift/assembler-x64/meta/src/generate/operand.rs
```diff
@@ -84,6 +90,8 @@ impl dsl::Location {
                 Some(size) => format!("self.{self}.to_string({size})"),
                 None => unreachable!(),
             },
+            xmm => format!("self.{self}.to_string()"),
+            rm128 => format!("self.{self}.to_string()"),
```
abrown, line 94:
```suggestion
            xmm | m128 => format!("self.{self}.to_string()"),
```

cranelift/assembler-x64/src/reg.rs
```diff
@@ -64,6 +104,8 @@ pub enum Size {
     Doubleword,
     /// A 64-bit access.
     Quadword,
+    /// A 128-bit access.
+    DoubleQuadword,
```
abrown, line 108: A thought: if we move all of the XMM bits to their own module, `xmm.rs`, then we don't need this variant here and we can separate the GPR and XMM logic from the pretty-printing below. It's a bit confusing that we ask `reg::enc::to_string` to print register 0 sometimes as `%rax` and other times as `%xmm0`. If we separate this out, then we could refactor to `gpr::enc::to_string(enc: u8, size: OperandSize)` and `xmm::enc::to_string(enc: u8)`.

abrown (commented):
This all makes sense to me. I think the main thing here is to fix CI; the minor refactoring I suggest below shouldn't significantly change this, but #10276 may.

cranelift/assembler-x64/meta/src/dsl/encoding.rs
```diff
@@ -197,7 +197,10 @@ impl Rex {
                 operands
                     .iter()
                     .all(|&op| matches!(op.location.kind(), OperandKind::Imm(_) | OperandKind::FixedReg(_))
```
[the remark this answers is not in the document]
rahulchaphalkar: Changed to check if 128 bits

cranelift/assembler-x64/meta/src/generate/format.rs
```diff
@@ -124,13 +124,31 @@ impl dsl::Format {
...
+                fmtln!(f, "}}");
+            }
+            [XmmReg(dst), XmmRegMem(src)] => {
+                fmtln!(f, "let {dst} = self.{dst}.enc();");
+                fmtln!(f, "match &self.{src} {{");
+                f.indent(|f| {
+                    fmtln!(f, "XmmMem::Xmm({src}) => rex.emit_two_op(buf, {dst}, {src}.enc()),");
+                    fmtln!(f, "XmmMem::Mem({src}) => {src}.emit_rex_prefix(rex, {dst}, buf),");
+                });
+                fmtln!(f, "}}");
```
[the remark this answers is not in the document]
rahulchaphalkar: Folded the xmm bits under reg

cranelift/assembler-x64/src/reg.rs
rahulchaphalkar: Folded xmm bits under reg. It does add several match statements in different places to check 128 bits.

cranelift/assembler-x64/meta/src/dsl/format.rs
```diff
@@ -311,6 +320,8 @@ pub enum OperandKind {
     Imm(Location),
     Reg(Location),
     RegMem(Location),
+    XmmReg(Location),
+    XmmRegMem(Location),
```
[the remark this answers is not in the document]
rahulchaphalkar: Folded xmm bits under reg.

rahulchaphalkar:
Thanks for the review, @abrown. I've addressed all comments. I'll now look at resolving conflicts.

rahulchaphalkar:
@abrown I've rebased latest main, please take a look.

cranelift/assembler-x64/meta/src/generate/operand.rs
```diff
@@ -167,6 +208,7 @@ impl dsl::Location {
             r16 | rm16 => Some("Size::Word"),
             r32 | rm32 => Some("Size::Doubleword"),
             r64 | rm64 => Some("Size::Quadword"),
+            xmm | rm128 => Some("Size::DoubleQuadword"),
```
abrown, line 211: We should be able to get rid of this now, right? Perhaps this becomes a `panic!("no need to generate a size for XMM-sized access")`.

cranelift/assembler-x64/src/api.rs
```diff
@@ -123,8 +130,11 @@ pub trait AsReg: Clone + std::fmt::Debug {
     fn enc(&self) -> u8;
 
     /// Return the register name.
-    fn to_string(&self, size: reg::Size) -> &str {
-        reg::enc::to_string(self.enc(), size)
+    fn to_string(&self, size: Option<reg::Size>) -> &str {
+        match size {
+            Some(_s) => reg::enc::to_string(self.enc(), size.unwrap()),
```
abrown, line 135:
```suggestion
            Some(size) => reg::enc::to_string(self.enc(), size),
```

cranelift/assembler-x64/src/mem.rs
```diff
@@ -272,6 +273,28 @@ impl<R: AsReg, M: AsReg> GprMem<R, M> {
...
+pub enum XmmMem<R: AsReg, M: AsReg> {
+    Xmm(R),
+    Mem(Amode<M>),
+}
+
+impl<R: AsReg, M: AsReg> XmmMem<R, M> {
+    /// Pretty-print the operand.
+    pub fn to_string(&self) -> String {
+        match self {
+            XmmMem::Xmm(xmm) => xmm::enc::to_string(xmm.enc()).to_owned(),
```
abrown, line 292:
```suggestion
            XmmMem::Xmm(xmm) => enc::to_string(xmm.enc()).to_owned(),
```

cranelift/assembler-x64/src/mem.rs
```diff
@@ -3,6 +3,7 @@
 use crate::api::{AsReg, CodeSink, Constant, KnownOffset, KnownOffsetTable, Label, TrapCode};
 use crate::reg::{self, NonRspGpr, Size};
 use crate::rex::{encode_modrm, encode_sib, Imm, RexFlags};
+use crate::xmm::{self};
```
abrown, line 6:
```suggestion
```

abrown (approved):
Makes sense to me with a couple small refactorings.

cranelift/assembler-x64/meta/src/generate/operand.rs
rahulchaphalkar: Done

cranelift/assembler-x64/src/api.rs
rahulchaphalkar: Done

cranelift/assembler-x64/src/mem.rs
rahulchaphalkar: I need `enc` from xmm module, similar to `let base = reg::enc::to_string(base.enc(), Size::Quadword);` needs enc from reg. I could just do `use crate::xmm::enc` but its easier to distinguish between `reg::enc` and `xmm::enc` this way I think.

cranelift/assembler-x64/src/mem.rs
rahulchaphalkar: From above comment, I need to have `use crate::xmm`

cranelift/assembler-x64/src/mem.rs
abrown: Oh, ok, I thought we were already in `xmm.rs` here. In that case, why not just use the existing `to_string` function?

```suggestion
            XmmMem::Xmm(xmm) => xmm.to_string(),
```

rahulchaphalkar: Andrew and I had a chat, and we're going to leave this as is for now. There is some work to be done on `to_string` front in `api.rs` in `AsReg`, where `size` is an optional argument.

Issue #10244, opened 2025-02-19
Conditionally remove page-alignment padding in `*.cwasm`


alexcrichton:
Currently Wasmtime will allocate space in `*.cwasm` files to ensure that sections are page-aligned on disk. This is done for two reasons:

*  When mmap-ing a `*.cwasm` into the address space the `.text` section needs to be made executable and we don't want to make non-code executable, meaning that the start/end of the `.text` section are both page aligned. The base alignment is set [here](https://github.com/bytecodealliance/wasmtime/blob/362b0dbf7ef5f1e4944266b85ff4fdf54dcc4dd0/crates/cranelift/src/obj.rs#L283-L285) and the alignment for the end is [set here](https://github.com/bytecodealliance/wasmtime/blob/362b0dbf7ef5f1e4944266b85ff4fdf54dcc4dd0/crates/cranelift/src/obj.rs#L440-L444).
* When `memory_init_cow` is enabled then it's necessary to mmap the `.rodata` section from the `*.cwasm` into the address space, meaning the start of the data section is also aligned. The data section alignment is set [here](https://github.com/bytecodealliance/wasmtime/blob/362b0dbf7ef5f1e4944266b85ff4fdf54dcc4dd0/crates/environ/src/compile/module_artifacts.rs#L138). The `data_align` value is set [here](https://github.com/bytecodealliance/wasmtime/blob/362b0dbf7ef5f1e4944266b85ff4fdf54dcc4dd0/crates/environ/src/compile/module_environ.rs#L1121)

When writing up this issue I thought we'd need to fix both these locations but it turns out we [already conditionally set `data_align`](https://github.com/bytecodealliance/wasmtime/blob/362b0dbf7ef5f1e4944266b85ff4fdf54dcc4dd0/crates/wasmtime/src/compile.rs#L779) by skipping that method if `memory_init_cow` isn't called. That means that I think the only thing we need to do here is to update compilation to Pulley to use a different alignment for the text section. We already have [some logic](https://github.com/bytecodealliance/wasmtime/blob/362b0dbf7ef5f1e4944266b85ff4fdf54dcc4dd0/crates/wasmtime/src/runtime/code_memory.rs#L330) for conditionally making the `.text` section executable, so we mostly just need to propagate similar logic back to the actual creation of the `*.cwasm` file.

alexcrichton:
Done in https://github.com/bytecodealliance/wasmtime/pull/10285

PR #10285, opened 2025-02-24
Don't pad .text to page size for pulley

posborne:
When targeting pulley we aren't directly emitting executable code in the .text section and we don't have a good idea of the true target page size so we end up with ELF files that can have a significant amount of extra padding around the .text section with no benefit to the consumer.

The change here aligns with the already present section flag SH_WASMTIME_NOT_EXECUTED.  Padding elimination for the .rodata.wasm section is already handled by the presence/absence of the memory-init-on-cow configuration.

For a basic wasip1 hello-world rust program with the combination of this patch and `-O memory-init-cow=n` I saw a reduction in the compiled wasm ELF from 421KB to 189KB with the sections no longer showing as being padded out significantly:

```
$ objdump --section-headers target/wasm32-wasip1/release/hello-wasm-world-0x00.cwasm

target/wasm32-wasip1/release/hello-wasm-world-0x00.cwasm:       file format elf64-littleriscv

Sections:
Idx Name              Size     VMA              Type
  0                   00000000 0000000000000000
  1 .wasmtime.engine  00000353 0000000000000000 DATA
  2 .wasmtime.bti     00000001 0000000000000000 DATA
  3 .text             00011bdc 0000000000000000
  4 .wasmtime.addrmap 00011c5c 0000000000000000 DATA
  5 .wasmtime.traps   00000ae0 0000000000000000 DATA
  6 .rodata.wasm      000019e8 0000000000000000 DATA
  7 .name.wasm        000027a6 0000000000000000 DATA
  8 .wasmtime.info    000010f9 0000000000000000 DATA
  9 .symtab           00001788 0000000000000000
 10 .strtab           000040f0 0000000000000000
 11 .shstrtab         00000089 0000000000000000
```

Addresses #10244, CC @ia0 @tschneidereit 

---

Feedback very welcome as this is a bit of a starter issue for me within wasmtime.  Things I considered but didn't end up going with with this patch:

* Directly inlining the conditional logic for the text align in the two callsites.
* Using the SH_WASMTIME_NOT_EXECUTED instead of looking at the triple again for the alignment decision.
* Possibly looking to see if it made sense to reuse memory-init-on-cow or some other flagging for this decision rather than using the triple.

I'm also open to any feedback regarding if there is any additional testing that should be added in tree for this patch

### DIFF
--- a/crates/cranelift/src/obj.rs
+++ b/crates/cranelift/src/obj.rs
@@ -30,6 +30,16 @@ use wasmtime_environ::{Compiler, TripleExt, Unsigned};
 
 const TEXT_SECTION_NAME: &[u8] = b".text";
 
+fn text_align(compiler: &dyn Compiler) -> u64 {
+    // text pages will not be made executable with pulley, so the section
+    // doesn't need to be padded out to page alignment boundaries.
+    if compiler.triple().is_pulley() {
+        0x1
+    } else {
+        compiler.page_size_align()
+    }
+}
+
 /// A helper structure used to assemble the final text section of an executable,
 /// plus unwinding information and other related details.
 ///
@@ -282,7 +292,7 @@ impl<'a> ModuleTextBuilder<'a> {
         let text = self.text.finish(&mut self.ctrl_plane);
         self.obj
             .section_mut(self.text_section)
-            .set_data(text, self.compiler.page_size_align());
+            .set_data(text, text_align(self.compiler));
 
         // Append the unwind information for all our functions, if necessary.
         self.unwind_info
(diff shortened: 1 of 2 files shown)


alexcrichton (approved):
This all looks great to me, thanks for working on this!

Using `objdump --section-headers` also makes me think that we should probably also provide a `Config::*` option for dropping the `.symtab`, `.strtab`, and `.shstrtab` sections. I don't think we have anything relying on that other than various debugging/perf utilities meaning it should be possible to strip them and still have a functioning `*.cwasm` binary. 

I'll file an issue about this...

posborne:
> @posborne you were probably busy with https://github.com/bytecodealliance/wasmtime/pull/10321 and other work, but in the case that you're stuck debugging the CI failure here I can try to help out if you'd like.

Ah, I missed that the checks failed due to a CI failure on this one, so I haven't looked yet.  I'll take a look in depth tomorrow, hopefully won't be too hard to resolve but I'll reach out if it's not making sense.

PR #10296, opened 2025-02-25
Winch: Turn on SIMD for Winch fuzzing

jeffcharles:
Part of #8093. Enables emitting modules with SIMD instructions for Winch when performing differential fuzzing if AVX and AVX2 extensions are available.


alexcrichton (approved):
🎉

PR #10297, opened 2025-02-26
winch(aarch64): Fix effective address calculation

saulecabrera:
~~This patch builds on top of https://github.com/bytecodealliance/wasmtime/pull/10263, only the last commit is relevant.~~

`load_addr` should load the effective address, not the contents of the
address.

This was causing issues with indirect function calls.

### DIFF
--- a/winch/codegen/src/isa/aarch64/masm.rs
+++ b/winch/codegen/src/isa/aarch64/masm.rs
@@ -315,7 +315,12 @@ impl Masm for MacroAssembler {
         })
     }
 
-    fn load_addr(&mut self, src: Self::Address, dst: WritableReg, size: OperandSize) -> Result<()> {
+    fn compute_addr(
+        &mut self,
+        src: Self::Address,
+        dst: WritableReg,
+        size: OperandSize,
+    ) -> Result<()> {
         let (base, offset) = src.unwrap_offset();
         self.asm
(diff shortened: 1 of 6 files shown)


winch/codegen/src/isa/aarch64/masm.rs
```diff
@@ -316,7 +316,9 @@ impl Masm for MacroAssembler {
     }
 
     fn load_addr(&mut self, src: Self::Address, dst: WritableReg, size: OperandSize) -> Result<()> {
-        self.asm.uload(src, dst, size, TRUSTED_FLAGS);
+        let (base, offset) = src.unwrap_offset();
+        self.asm
+            .add_ir(u64::try_from(offset).unwrap(), base, dst, size);
```
cfallin, line 321: Perhaps we could call this masm method `compute_addr` to make it clearer that it's returning the address itself, not loading anything? (i.e., avoid the ambiguity where "load" could mean "load from memory" or "load a value into a register")

cfallin (approved):
Good catch!

winch/codegen/src/isa/aarch64/masm.rs
saulecabrera: Good point. Renamed, thanks!

PR #10303, opened 2025-02-27
Improve rebuild detection of test-programs

alexcrichton:
This commit improves the logic of detecting when to rebuild the `test-programs` artifacts used during test by parsing the `*.d` files that Cargo emits as part of its compilation and using that as the `cargo:rerun-if-changed` directive. This not only includes what was previously depended on but additionally includes features such as `path` dependencies which might temporarily be used during development.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,4 +27,4 @@ examples/.cache
 cranelift/isle/veri/veri_engine/test_output
 crates/explorer/node_modules
 tests/all/pulley_provenance_test.cwasm
-artifacts
+/artifacts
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -27,4 +27,4 @@ examples/.cache
 cranelift/isle/veri/veri_engine/test_output
 crates/explorer/node_modules
 tests/all/pulley_provenance_test.cwasm
-artifacts
+/artifacts
```
pchickey, line 30: This doesn't look right

pchickey (approved):
Thanks, all good except for the gitignore

.gitignore
alexcrichton: Oh this is actually intentional, although I forgot to mention it. We `./crates/test-programs/artifacts` in the repo which was otherwise being ignored when new changes were being added to it and I was editing things around there so figured I'd fix it up here as well. This was added originally because our c-api documentation generated this folder in the root of the repo

PR #10306, opened 2025-02-27
Test natively on AArch64, not emulated

alexcrichton:
This commit updates the aarch64 tests to use the new `*-arm` images from GitHub instead of running through QEMU emulation. This should be a bit faster but primarily helps build confidence that it runs on real hardware.

I'll note that release builds aren't updated at this time to run on native hardware since that's moreso about glibc compatibility and it's a bit easier to keep the setup we currently have for that. If `*-arm` machines are noticably faster than the default x64 machines then we could in theory move everything over to an aarch64-based machine as opposed to just the aarch64 build.

### DIFF
--- a/ci/build-test-matrix.js
+++ b/ci/build-test-matrix.js
@@ -95,12 +95,8 @@ const FULL_MATRIX = [
     "filter": "mingw-x64"
   },
   {
-    "os": ubuntu,
+    "os": ubuntu + '-arm',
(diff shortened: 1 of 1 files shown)


fitzgen (approved):
Does this also run the NO_HOG_MEMORY tests that we were skipping under qemu?

alexcrichton:
It does indeed!

PR #6648, opened 2023-06-27
[wit-bindgen] provide more control over type ownership

dicej:
This replaces the `duplicate_if_necessary` parameter to `wasmtime::component::bindgen` with a new `ownership` parameter which provides finer-grained control over whether and how generated types own their fields.

The default is `Ownership::Owning`, which means types own their fields regardless of how they are used in functions.  These types are passed by reference when used as parameters to guest-exported functions.  Note that this also affects how unnamed types (e.g. `list<list<string>>`) are passed: using a reference only at the top level (e.g. `&[Vec<String>]` instead of `&[&[&str]]`, which is more difficult to construct when using non-`'static` data).

The other option is `Ownership::Borrowing`, which includes a `duplicate_if_necessary` field, providing the same code generation strategy as was used prior to this change.

If we're happy with this approach, I'll open another PR in the `wit-bindgen` repo to match.

This also fixes a bug that caused named types to be considered owned and/or borrowed when they shouldn't have been due to having fields with unnamed types which were owned and/or borrowed in unrelated interfaces.

PR #10311, opened 2025-02-28
wasmtime-wit-bindgen: emit a definition for all types in a wit interface

### LINKED ISSUE #10090: Option to Generate Unused Types with `bindgen!()` Macro
Hi, first of all, I want to thank you all for the great work and effort you’ve put into this project.

#### Feature Request

It would be incredibly helpful to have the option to generate unused types when using the `wasmtime::component::bindgen!()` macro, similar to the functionality provided by the `wit_bindgen::generate!()` macro.  
I’ve included a code example below to illustrate the desired functionality:

**WIT File:**
```wit
package example:example;

interface types {
  record referenced-type {
    field: string
  }

  record no-ref-type {
    filed: string
  }

  fn: func(t: referenced-type);
}

world example {
  import types;
}
```
**Code in Rust:**
```rust
mod wsm {
    wasmtime::component::bindgen!({
        path: "wit",
        // NOTE: no equivalent for this option provided
        // generate_unused_types: true,
    });

    use self::example::example::types::{
        ReferencedType,
        // NoRefType: This type won't get generated
    };
}

mod wit {
    wit_bindgen::generate!({
        path: "wit",
        generate_unused_types: true,
    });

    use self::example::example::types::{NoRefType, ReferencedType};
}
```
#### Benefit

Generating unused types can be really useful when we aim to group shared types into a single package and generate them only once for use across multiple other packages.

#### Alternatives

The current workaround involves adding unused worlds, interfaces, or functions that reference the desired types, ensuring they are generated.

pchickey:
Rust type definitions for wit types are created for the set returned by `modes_of`, when an empty list is returned the type never gets a definition in Rust.

The calculation of TypeInfo only reaches types which are passed to or from a function. For types which are not reachable, default to the defining them according to the ownership setting given to bindgen.

I have my doubts that `with`-reuse of bindgen types actually works properly when bindgen is set to Ownership::Borrowing but thats out of scope for this PR, which is to fix #10090

### DIFF
--- a/crates/component-macro/tests/expanded/dead-code.rs
+++ b/crates/component-macro/tests/expanded/dead-code.rs
@@ -248,6 +248,56 @@ pub mod a {
         pub mod interface_with_dead_type {
             #[allow(unused_imports)]
             use wasmtime::component::__internal::{anyhow, Box};
+            pub type LiveType = super::super::super::a::b::interface_with_live_type::LiveType;
+            const _: () = {
+                assert!(4 == < LiveType as wasmtime::component::ComponentType >::SIZE32);
+                assert!(
+                    4 == < LiveType as wasmtime::component::ComponentType >::ALIGN32
+                );
+            };
+            #[derive(wasmtime::component::ComponentType)]
+            #[derive(wasmtime::component::Lift)]
+            #[derive(wasmtime::component::Lower)]
+            #[component(record)]
+            #[derive(Clone, Copy)]
+            pub struct DeadType {
+                #[component(name = "a")]
+                pub a: u32,
+            }
+            impl core::fmt::Debug for DeadType {
+                fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+                    f.debug_struct("DeadType").field("a", &self.a).finish()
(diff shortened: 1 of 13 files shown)


dicej:
> ping @dicej, who last touched this logic in #6648 . Does this still work for your needs?

#6648 Was motivated by fuzz testing the host bindings; the nested borrows we had before made generating arbitrary test cases really difficult, so I wanted to avoid that.  As long as the fuzz tests in this repo still work, then we're good 👍

alexcrichton (approved):
This logic is originally derivative of [this logic in wit-bindgen for guests](https://github.com/bytecodealliance/wit-bindgen/blob/e0d2b889411acf60738dcd976ed5373f7a81e651/crates/rust/src/interface.rs#L1927), and while they were originally the same I think the guest has since moved on. IIRC that part of the guest bindings was extremely tricky and probably still isn't fully correct even today. I also think that there's various different concerns on the guest than there are for the host, which means we probably don't want them to be the exact same.

Regardless though this seems reasonable to me. I'm not actually sure if it's correct so I'm mostly with Joel where if it compiles seems reasonable to me

PR #10314, opened 2025-02-28
Update wasm-tools to the latest revision

alexcrichton:

### DIFF
--- a/crates/wasmtime/src/compile.rs
+++ b/crates/wasmtime/src/compile.rs
@@ -373,7 +373,8 @@ impl<'a> CompileInputs<'a> {
                     symbol: trampoline.symbol_name(),
                     function: compiler
                         .component_compiler()
-                        .compile_trampoline(component, types, idx, tunables)?
+                        .compile_trampoline(component, types, idx, tunables)
+                        .with_context(|| format!("failed to compile {}", trampoline.symbol_name()))?
                         .into(),
                     info: None,
                 })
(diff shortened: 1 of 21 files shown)


alexcrichton:
Draft for now as they're git dependencies

alexcrichton:
Ok this is now ready for review. It's going to have a lot of conflicts with https://github.com/bytecodealliance/wasmtime/pull/10106, however, so the plan is to land that first currently.

pchickey (approved):

alexcrichton:
@dicej mind doing a once-over on this with an eye towards the component-model-async bits?

crates/wasmtime/src/compile.rs
```diff
@@ -482,18 +486,17 @@ impl<'a> CompileInputs<'a> {
                 if translation.module.functions[func_index].is_escaping() {
                     self.push_input(move |compiler| {
                         let func_index = translation.module.func_index(def_func_index);
-                        let trampoline = compiler.compile_array_to_wasm_trampoline(
-                            translation,
-                            types,
-                            def_func_index,
-                        )?;
+                        let symbol = format!(
```
dicej, line 489: Any concerns about the cost of doing these `format!`s unconditionally despite only using the result if there's an error?  If so, perhaps we could make this a lambda that we call from within the `with_context` closure.

dicej (approved):
Thanks for doing this!

crates/wasmtime/src/compile.rs
alexcrichton: Nah they were gonna all happen anyway, I just moved them a bit earlier so cranelift failures now have the function name so I don't have to go hunting down which intrinsic is the one that needs updating

dicej: Oops, nevermind, I didn't notice we use it in the non-error case.  Disregard!

PR #10316, opened 2025-03-01
asm: SSE ALU Instructions

rahulchaphalkar:
Implement the following SSE instructions:

- `xorpd`
- `xorps`
- `andpd`
- `andps`
- `addpd`
- `addps`
- `subpd`
- `subps`
- `orps`

The opcode for `*ps` variants of these instructions is `NP 0F xx`. NP has the following explanation in manual:
```
NP — Indicates the use of 66/F2/F3 prefixes (beyond those already part of the instructions opcode) are not
allowed with the instruction. Such use will either cause an invalid-opcode exception (#UD) or result in the
encoding for a different instruction.
```
There is an existing `NoPrefix` prefix, but I have not reused that to serve above `NP` as there might be some difference in the intended meaning. Also, `NP` doesn't classify as a prefix technically. I am thinking to address this in a later patch which would deal with prefix parsing. Any thoughts or feedback on this is welcome. 
My thinking is `NP` doesn't functionally do anything, so it is fine to ignore it. We just need a way for `validate` function to check this.

### DIFF
--- a/cranelift/assembler-x64/meta/src/instructions/add.rs
+++ b/cranelift/assembler-x64/meta/src/instructions/add.rs
@@ -21,6 +21,8 @@ pub fn list() -> Vec<Inst> {
         inst("addw", fmt("RM", [rw(r16), r(rm16)]), rex([0x66, 0x3]).r(), _64b | compat),
         inst("addl", fmt("RM", [rw(r32), r(rm32)]), rex(0x3).r(), _64b | compat),
         inst("addq", fmt("RM", [rw(r64), r(rm64)]), rex(0x3).w().r(), _64b),
+        inst("addpd", fmt("A", [rw(xmm), r(rm128)]), rex([0x66, 0x0F, 0x58]).r(), _64b),
+        inst("addps", fmt("A", [rw(xmm), r(rm128)]), rex([0x0F, 0x58]).r(), _64b),
         // Add with carry.
         inst("adcb", fmt("I", [rw(al), r(imm8)]), rex(0x14).ib(), _64b | compat),
         inst("adcw", fmt("I", [rw(ax), r(imm16)]), rex([0x66, 0x15]).iw(), _64b | compat),
--- a/cranelift/assembler-x64/meta/src/instructions/add.rs
+++ b/cranelift/assembler-x64/meta/src/instructions/add.rs
@@ -21,8 +21,9 @@ pub fn list() -> Vec<Inst> {
         inst("addw", fmt("RM", [rw(r16), r(rm16)]), rex([0x66, 0x3]).r(), _64b | compat),
         inst("addl", fmt("RM", [rw(r32), r(rm32)]), rex(0x3).r(), _64b | compat),
         inst("addq", fmt("RM", [rw(r64), r(rm64)]), rex(0x3).w().r(), _64b),
-        inst("addpd", fmt("A", [rw(xmm), r(rm128)]), rex([0x66, 0x0F, 0x58]).r(), _64b),
+        // SSE vector instructions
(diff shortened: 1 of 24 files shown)


cranelift/codegen/src/isa/x64/inst.isle
```diff
@@ -3833,19 +3840,21 @@
...
 (decl x64_subpd (Xmm XmmMem) Xmm)
-(rule 0 (x64_subpd src1 src2)
-      (xmm_rm_r (SseOpcode.Subpd) src1 src2))
-(rule 1 (x64_subpd src1 src2)
+(rule 2 (x64_subpd src1 src2)
       (if-let true (use_avx))
       (xmm_rmir_vex (AvxOpcode.Vsubpd) src1 src2))
+(rule 1 (x64_subpd src1 (is_xmm_mem src2)) (x64_subpd_a src1 src2))
+(rule 0 (x64_subpd src1 src2)
+      (xmm_rm_r (SseOpcode.Subpd) src1 src2))
```
abrown, line 3857: Let's remove the old emission. In a recent PR I began removing these for scalar instructions. It ensures that our `is_xmm_mem` is operating as expected.

cranelift/codegen/src/isa/x64/inst.isle
```diff
@@ -3833,19 +3840,21 @@
...
 
 ;; Helper for creating `subpd` instructions.
 (decl x64_subpd (Xmm XmmMem) Xmm)
-(rule 0 (x64_subpd src1 src2)
-      (xmm_rm_r (SseOpcode.Subpd) src1 src2))
-(rule 1 (x64_subpd src1 src2)
+(rule 2 (x64_subpd src1 src2)
       (if-let true (use_avx))
       (xmm_rmir_vex (AvxOpcode.Vsubpd) src1 src2))
+(rule 1 (x64_subpd src1 (is_xmm_mem src2)) (x64_subpd_a src1 src2))
```
abrown, line 3855: And on that note, I believe that the ISLE we generate already has the signature `(... (Xmm XmmMem) Xmm)` which is what we need here. We can drop the `is_xmm_mem` extractor for these instructions.

cranelift/assembler-x64/meta/src/instructions/add.rs
```diff
@@ -21,6 +21,8 @@ pub fn list() -> Vec<Inst> {
         inst("addw", fmt("RM", [rw(r16), r(rm16)]), rex([0x66, 0x3]).r(), _64b | compat),
         inst("addl", fmt("RM", [rw(r32), r(rm32)]), rex(0x3).r(), _64b | compat),
         inst("addq", fmt("RM", [rw(r64), r(rm64)]), rex(0x3).w().r(), _64b),
+        inst("addpd", fmt("A", [rw(xmm), r(rm128)]), rex([0x66, 0x0F, 0x58]).r(), _64b),
+        inst("addps", fmt("A", [rw(xmm), r(rm128)]), rex([0x0F, 0x58]).r(), _64b),
```
abrown, line 25: Some minor nits about organizing these new instructions:
- let's put them at the bottom of the file with their own comment header, e.g., `// SSE vector addition.` or the like
- can we order them from smallest to largest? I.e., packed single-precision (`ps = f32x4`) first and then packed double-precision (`pd = f64x2`).

rahulchaphalkar: Done

cranelift/codegen/src/isa/x64/inst.isle
rahulchaphalkar: Done

abrown (approved):

Issue #10245, opened 2025-02-19
Enable deserializing code owned by some external source


alexcrichton:
Currently in Wasmtime we have `Module::deserialize` and `Module::deserialize_file`. Given these APIs though it has the fundamental requirements that `deserialize` will copy the bytes into Wasmtime (e.g. some new map) and `deserialize_file` requires `std` support. That means that these APIs are not compatible with a `#![no_std]` situation which already has bytes loaded elsewhere in memory. There's no need to mutate the bytes so we just need a read-only view of the bytes.

One possibility would be to add a `deserialize_raw` API which takes `*const [u8]` or similar. That would mean that we're not responsible for owning the memory at all and the embedder is responsible for ensuring it stays alive long enough as well as is able to update page permissions (which isn't applicable for Pulley but may be applicable for other platforms). This might be done by copying [this API](https://github.com/bytecodealliance/wasmtime/blob/362b0dbf7ef5f1e4944266b85ff4fdf54dcc4dd0/crates/wasmtime/src/engine.rs#L767-L784) to have raw pointers and then adding a [new variant to `MmapVec`](https://github.com/bytecodealliance/wasmtime/blob/362b0dbf7ef5f1e4944266b85ff4fdf54dcc4dd0/crates/wasmtime/src/runtime/vm/mmap_vec.rs#L36-L41) which represents a non-owned pointer.

The end goal is that Wasmtime should be able to load bytes from an external source without needing to copy out of that source in a `no_std` context.

alexcrichton:
Addressed in https://github.com/bytecodealliance/wasmtime/pull/10321

PR #10321, opened 2025-03-03
Add Engine/Component::deserialize_raw

posborne:
For targets without virtual memory support, the only mechanism for
load code previously was to copy bytes from a provided slice into
a separate, owned, allocated buffer containing a copy of the serialized
module contents.  This is expensive for constrained targets and
prohibits embedded runtimes like doing things like running code
directly out of memory-mapped devices such as NOR flash.

The tradeoff for directly using the externally owned memory is that
the caller must ensure that the code memory will not be written
to for the lifetime of the CodeMemory.

Loading code from externally owned memory is supported for all
configurations but is expected to fail to deserialize on platforms
that support virtual memory and are attempting to load modules
or components that require that the memory be made executable
(native code rather than pulley).

https://github.com/bytecodealliance/wasmtime/issues/10245


crates/wasmtime/src/runtime/vm/mmap_vec.rs
```diff
@@ -74,6 +80,20 @@ impl MmapVec {
...
+    fn new_externally_owned(ptr: *const u8, len: usize, alignment: usize) -> MmapVec {
+        let layout = Layout::from_size_align(len, alignment)
+            .expect("Invalid size or alignment for MMapVec allocation");
+
+        // NonNull doesn't allow descrimination between mut and non-mut pointers.
+        // `as_mut` on SendSyncPtr is unsafe and to ensure soundness we avoid
+        // its use here.
+        let base = SendSyncPtr::new(
+            NonNull::new(ptr as *mut u8).expect("Attempted to create MmapVec from null pointer"),
+        );
```
posborne, line 93: Reusing `NonNull`/`SendSyncPtr` here and casting the const ptr to mut definitely feels hacky so feedback very welcome on a better approach if one would make sense for this.

crates/wasmtime/src/engine.rs
```diff
@@ -783,6 +783,29 @@ impl Engine {
...
+    ///
+    /// The `expected` marker here is whether the bytes are expected to be
+    /// a precompiled module or a component
+    ///
+    /// # Safety
+    ///
+    /// As there is no copy here, the runtime will be making direct use
+    /// of the provided memory and may read or write to this memory.  As such,
+    /// any outside writes to this memory region will result in undefined and
+    /// likely very undesirable behavior.
```
fitzgen, line 796: Maybe also mention that the pointer must be valid for `size` bytes and must point to the contents of a `.cwasm` file or result of `{Module,Component}::serialize`-style APIs.

crates/wasmtime/src/runtime/vm/mmap_vec.rs
fitzgen: There is no difference between `*mut T` and `*const T` in Rust's semantics, so it isn't a big deal.

Soft preference for using `.cast::<T>()` and `.cast_{const,mut}()` methods for conversions, since they are a little more explicit and future proof, since they cast just one thing at a time, where `as` can change multiple things.

fitzgen (approved):
Looks great! Just a couple nitpicks below.

crates/wasmtime/src/engine.rs
```diff
@@ -783,6 +783,34 @@ impl Engine {
...
+    ///
+    /// The memory provided must be valid for `size` bytes following `ptr`.
+    /// As there is no copy here, the runtime will be making direct use
+    /// of the provided memory and may read or write to this memory.  As such,
+    /// any outside writes to this memory region will result in undefined and
+    /// likely very undesirable behavior.
+    pub(crate) unsafe fn load_code_raw(
+        &self,
+        ptr: *const u8,
+        size: usize,
```
alexcrichton, line 805: A small bikeshed on this, one thing we could do here is take `NonNull<[u8]>` which (a) bundles ptr/len into one, and (b) helps clarify the contract that it must be non-null

posborne: I'm definitely not against this; I think the counterpoint is that, at least with the type signature, you lose the ability to document that the usage of the memory is going to be immutable (which I do believe would be an important invariant for the intended use case) and we end up pushing the `*mut` transmute to callers who have a reference to an immutable chunk of memory.

So, we gain the ability to model with types that we require a non-null pointer but we lose the ability to inform the caller that we're using the memory immutably. 🤷

fitzgen: That's already true as soon as you're using raw pointers instead of safe references though, so we are already in that world here.

posborne: Pushed a new commit making the change to `NonNull<[u8]>`.

alexcrichton (approved):
Oh one last comment, for testing could something be added to `tests/all/module.rs` perhaps? It doesn't need to be super comprehensive but I think it'd be good to have at least a small test (maybe Pulley-only?) which loads a module, asserts the image is pointing within the original pointer, and maybe does some basic arithmetic in-wasm to ensure it works.

crates/wasmtime/src/runtime/vm/mmap_vec.rs
```diff
@@ -101,6 +110,35 @@ impl MmapVec {
...
+    /// will result in undefined and likely very undesirable behavior.
+    pub unsafe fn from_raw(memory: NonNull<[u8]>, alignment: usize) -> Result<MmapVec> {
+        #[cfg(not(has_virtual_memory))]
+        {
+            return Ok(MmapVec::new_externally_owned(memory, alignment));
+        }
+        #[cfg(has_virtual_memory)]
+        {
+            return MmapVec::from_slice_with_alignment(memory.as_ref(), alignment);
+        }
```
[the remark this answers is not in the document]
posborne: > Could this unconditionally use the new_externally_owned variant here?

Yeah, should be doable.  I'll do some testing to ensure we don't simply panic or crash but are able to nicely reject modules that will not be able to be execute without mmap but agree that it's probably best to provide this on all platforms in a consistent fashion.

posborne:
@alexcrichton Hopefully the updates address your latest comments, I did a little bit of extra in order to ensure that we bailed out with hopefully reasonable errors rather than panics for cases where virtual memory is required.

I wasn't able to find a clear way to assert in the test context that the memory offsets were identical as a lot of the APIs that actually make it through to the CodeMemory and mmap are either not public or they work on copy types.  It's likely that I missed something that I can assert on but I did at least test that we can do some basic math.

alexcrichton:
👍 tests and changes look great to me, no need to test what I was thinking if it's difficult

posborne:
Fixed the compile error, missed doing a final compile check on a target with virtual memory.

posborne:
Ok, I missed that for platforms like i686 that the default engine config falls back on pulley so those wouldn't expect to fail -- I should be able to handle that just fine.  I'll try to do some testing for those cases and fix that up tomorrow morning.  Will probably just do a check around the assertion on the engine target triple containing "pulley".

With that small change I'll probably amend the original commit and squash down as the initial title about only adding this support for platforms without virtual memory support is no longer true as well.

PR #10332, opened 2025-03-05
Enable the GC proposal during general fuzzing

### LINKED ISSUE #10328: Enable general/differential fuzzing with the Wasm GC proposal

fitzgen:
This allows us to fuzz Wasm GC in our fuzz targets that use the common config-generation infrastructure, such as the differential fuzz target.

Fixes #10328

---------------

I've been running this locally for about 20 minutes since I last hit a bug. Let's let it run longer before we merge it, just in case. Figured that there was no reason I couldn't make the PR now and optimistically and speculatively pipeline with review and CI potentially and all that.

### DIFF
--- a/crates/wasmtime/src/runtime/vm/traphandlers.rs
+++ b/crates/wasmtime/src/runtime/vm/traphandlers.rs
@@ -33,6 +33,7 @@ pub use self::tls::{AsyncWasmCallState, PreviousAsyncWasmCallState};
 
 pub use traphandlers::SignalHandler;
 
+#[derive(Debug)]
 pub(crate) struct TrapRegisters {
     pub pc: usize,
     pub fp: usize,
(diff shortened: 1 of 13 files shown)


alexcrichton (approved):

fitzgen:
Found a couple issues that need to be fixed before this lands.

fitzgen:
Fuzzing has successfully run overnight on my machine locally, so I'll clean up what I have and update this PR so it can merge. There are still outstanding GC bugs, but I think we're at the point where the fuzzers run long enough that it makes sense to move forward with this PR and start exercising GC programs in OSS-Fuzz.

crates/wasmtime/src/runtime/vm/traphandlers.rs
```diff
@@ -668,8 +669,11 @@ impl CallThreadState {
         faulting_addr: Option<usize>,
         call_handler: impl Fn(&SignalHandler) -> bool,
     ) -> TrapTest {
+        log::trace!("testing if trap: {faulting_addr:?}, {regs:?}");
```
alexcrichton, line 672: I think we'll want to remove all the logging here because this is called from a signal handler. Or maybe a per-module debugging macro which can be toggle locally without committing it or something like that?

alexcrichton (approved):

PR #10340, opened 2025-03-05
Add "pure" flag to `ir::MemFlags`

fitzgen:
This flag represents whether the memory operation's safety (e.g. the validity of its `notrap` and `readonly` claims) is purely a function of its data dependencies.

If this flag is `true`, then it is okay to code motion this instruction to arbitrary locations, in the function, including across blocks and conditional branches, so long as data dependencies (and trap ordering, if relevant) are upheld.

If this flag is `false`, then the memory operation's safety potentially relies upon invariants that are not reflected in its data dependencies, and therefore it is not safe to code motion this operation. For example, this operation could be in a block that is dominated by a control-flow bounds check that makes this operation safe, and that invariant is not reflected in its operands. It would be unsafe to code motion such an instruction above its associated bounds check, even if its data dependencies would still be satisfied.

I've added this flag because we were doing exactly that kind of code motion where we moved a `readonly` and `notrap` memory operation past its associated null-check and therefore it was no longer safe to perform and we would get a segfault. This could only be triggered when the Wasm typed-function-references proposal was enabled, which is not a tier-1 proposal, so it is not considered a vulnerability. Nonetheless, it is a pretty scary kind of bug, and other code paths weren't affected due to pretty subtle interactions. And this is the motivation for the new "pure" flag: without needing to explicitly opt into data-dependency-based code motion (i.e. set the "pure" flag), it is too easy to accidentally move loads past their control-flow-based safety guards.

### DIFF
--- a/cranelift/codegen/src/ir/memflags.rs
+++ b/cranelift/codegen/src/ir/memflags.rs
@@ -73,7 +73,7 @@ pub struct MemFlags {
     // * 4 - checked flag
     // * 5/6 - alias region
     // * 7/8/9/10/11/12/13/14 - trap code
-    // * 15 - pure flag
+    // * 15 - can_move flag
     //
     // Current properties upheld are:
     //
@@ -112,11 +112,12 @@ const ALIAS_REGION_OFFSET: u16 = 5;
 const MASK_TRAP_CODE: u16 = 0b1111_1111 << TRAP_CODE_OFFSET;
 const TRAP_CODE_OFFSET: u16 = 7;
 
-/// Whether the load's/store's safety is purely a function of its data
-/// dependencies (i.e. operands) and is not guarded by
-/// outside-the-data-flow-graph properties, like implicit bounds-checking
+/// Whether this memory operation may be freely moved by the optimizer so long
+/// as its data dependencies are satisfied. That is, by setting this flag, the
+/// producer is guaranteeing that this memory operation's safety is not guarded
+/// by outside-the-data-flow-graph properties, like implicit bounds-checking
 /// control dependencies.
-const BIT_PURE: u16 = 1 << 15;
+const BIT_CAN_MOVE: u16 = 1 << 15;
 
 impl MemFlags {
     /// Create a new empty set of flags.
@@ -204,7 +205,7 @@ impl MemFlags {
                 self.with_alias_region(Some(AliasRegion::Vmctx))
             }
             "checked" => self.with_checked(),
-            "pure" => self.with_pure(),
+            "can_move" => self.with_can_move(),
 
             other => match TrapCode::from_str(other) {
                 Ok(code) => self.with_trap_code(Some(code)),
@@ -272,7 +273,7 @@ impl MemFlags {
     /// This flag does *not* mean that the associated instruction can be
     /// code-motioned to arbitrary places in the function so long as its data
     /// dependencies are met. This only means that, given its current location
-    /// in the function, it will never trap. See the `pure` method for more
+    /// in the function, it will never trap. See the `can_move` method for more
     /// details.
     pub const fn notrap(self) -> bool {
         self.trap_code().is_none()
@@ -289,8 +290,8 @@ impl MemFlags {
         self.with_trap_code(None)
     }
 
-    /// Is this memory operation's safety (e.g. the validity of its `notrap` and
-    /// `readonly` claims) purely a function of its data dependencies?
+    /// Is this memory operation safe to move so long as its data dependencies
+    /// remain satisfied?
     ///
     /// If this is `true`, then it is okay to code motion this instruction to
     /// arbitrary locations, in the function, including across blocks and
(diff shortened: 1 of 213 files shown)


cfallin (approved):
Thanks for finding this and for the excellent documentation of semantics of the new flag. I think this is a reasonable solution.

Would you mind adding two filetests (e.g. in `cranelift/filetests/filetests/egraph/`) testing that a non-pure load won't LICM, and that a pure/readonly/notrap one will? (This will also force the parser bit below)

crates/cranelift/src/func_environ.rs
```diff
@@ -2474,7 +2480,7 @@ impl FuncEnvironment<'_> {
             }
         };
 
-        let mut flags = MemFlags::trusted().with_checked();
+        let mut flags = MemFlags::trusted().with_checked().with_pure();
         if !memory.memory_may_move(self.tunables) {
             flags.set_readonly();
         }
```
alexcrichton, line 2486: This raises an eyebrow from me. Functionally I don't think this is an issue, but it seems kind of weird to have a pure, but not readonly, `MemFlags`. I could sort of extend that to "what is a `pure` but trapping load?". 

To be conservative though, do you think it would be reasonable to validate that pure memflags are always notrap+readonly? Or is there a use case for a trapping pure operation or non-readonly pure operation?

crates/cranelift/src/gc/enabled.rs
```diff
@@ -680,9 +680,12 @@ pub fn translate_array_len(
         // don't know its length yet. Chicken and egg problem.
         BoundsCheck::Access(ir::types::I32.bytes()),
     );
-    let result = builder
-        .ins()
-        .load(ir::types::I32, ir::MemFlags::trusted(), len_field, 0);
+    let result = builder.ins().load(
+        ir::types::I32,
+        ir::MemFlags::trusted().with_readonly(),
```
alexcrichton, line 685: How come this is readonly? Can't the length change over time?

cranelift/codegen/src/ir/memflags.rs
```diff
@@ -382,6 +424,9 @@ impl fmt::Display for MemFlags {
         if self.readonly() {
             write!(f, " readonly")?;
         }
+        if self.pure() {
+            write!(f, " pure")?;
```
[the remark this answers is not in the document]
fitzgen: That all happens via `ir::MemFlags::set_by_name`, so the parser itself doesn't actually need any updating.

fitzgen:
> Would you mind adding two filetests (e.g. in `cranelift/filetests/filetests/egraph/`) testing that a non-pure load won't LICM, and that a pure/readonly/notrap one will? (This will also force the parser bit below)

Extended `load-hoist.clif` to test both hoisting and non-hoisting. Good idea, thanks!

alexcrichton (approved):

crates/cranelift/src/gc/enabled.rs
fitzgen: Nope, GC arrays have fixed length.

crates/cranelift/src/func_environ.rs
fitzgen: The semantics of `pure` are "this memory operation's safety, including the validity of its other flags, depend only on its data dependencies". There is nothing about that statement that requires the memory operation also be `readonly` or also be `notrap`. Therefore, I'd rather not add additional validation code that ultimately isn't necessary.

fitzgen: We could potentially rename it to `data-pure` or something if you think that would be better?

alexcrichton: What I'm worried about is that it's a pretty subtle definition and you basically can't do anything if you only check `pure`, it's only if pure + readonly + notrap are all set that something can actually be done about moving the instruction around. I do realize it's more validation though... Do you feel that there's not that much risk of someone in the future using `pure` by accident?

In terms of bikeshedding I suppose my concern does indeed stem from the term "pure" being pretty overloaded and my gut feeling about "pure" things is "oh hey you can put that anywhere and it's fine". Basically I do think my concern might be assuaged with a different name. That being said "data-pure" doesn't feel all that different from "pure", but perhaps different enough?

I'm trying to think of other ideas... 

* `notrap_always`
* `notrap_everywhere`
* `semantically_ok_to_put_anywhere` (ok definitely not)
* `cfg_independent` 

I'm reading up on `readonly` and that already says that nothing about the memory can change from when the function starts to when it ends,  so maybe `notrap_foo` as a name? Maybe `notrap_pure`? Something like that maybe...

cfallin: I actually agree that it probably doesn't make sense to have a load that is `pure` but not `readonly`: `pure` means that we can sink or hoist it anywhere (constrained by its address input data dependency); that should only be possible if the memory is unchanging. (Right? Or are there cases where we're ok with hoisting but the value is changing, and we might load a different value if hoisted?)

`pure` is a bit different than (and orthogonal to) `notrap` though: it's ok to have an address that maybe traps, but traps everywhere the same. In other words, `pure` says "the trappiness of this address is unchanging" while `notrap` says "this address is not trappy at this location".

cfallin: Maybe we should name flags based on actions that we expect or permit the compiler to do instead:

- `can_trap`: add trap metadata. This is the inverse of `notrap` (why invert it? because we're describing actions, not lack of actions)
- `can_skip`: allow dead-code elimination if unused
- `can_move`: can hoist or sink as data dependencies allow
- `can_merge`: can reuse values from equivalent loads, even without alias analysis proving this (i.e., value is readonly)

Is that sufficient to express what we need? `is_pure_for_egraph` is then `can_skip` + `can_move` + `can_merge` (since the egraph can do all three); DCE uses `can_skip`; if we do other things in the future we have the specific action-permissions defined for those.

Alternately-alternately, we could look at the predicates we define today and give specific names to those: e.g., redefine `pure` to be what `is_pure_for_egraph` tests; and then `notrap` (or `can_trap`) for DCE; and I think that's it?

fitzgen: I like `can_{trap,skip,move,merge}` and think that is a good place to aim towards ending up at. This PR's `pure` is equivalent to `can_move`, but the rest of those flags don't align exactly with what we have today. For example, `can_skip` would be `is-not-a-store && notrap` today -- would it ever make sense to have `can_skip` on a store?

Anyways, what do we think of, in order to land this PR, renaming this PR's `pure` to `move`, `can_move`, or `movable` and then kicking the can down the road for the rest of the potential flags overhaul?

cfallin: SGTM!

And yes, I think at least `can_skip` and `can_merge` apply only to loads. Stores can definitely vary on trappiness, and I think `can_move` makes sense too -- some store that we want to happen at some point but don't care where (analogous to eventual consistency or relaxed-consistency atomics).

PR #10345, opened 2025-03-06
Remove hotswapping support from cranelift-jit

bjorn3:
It was originally introduced for cg_clif. cg_clif recently removed it's use of hotswapping as the way it is implemented in cranelift-jit has various issues like leaking memory, panicking when the memory allocator decided to put two functions more than 2GB away from each other and only supporting x86_64. Better hotswapping support will likely require a fundamentally different implementation.

Also fix a bug that prevents cg_clif from switching to is_pic=false for the jit mode.

cc https://github.com/bytecodealliance/wasmtime/issues/2735

### DIFF
--- a/cranelift/jit/src/backend.rs
+++ b/cranelift/jit/src/backend.rs
@@ -767,15 +767,16 @@ impl Module for JITModule {
             align,
         } = data;
 
-        let size = init.size();
-        let ptr = if size == 0 {
-            // Return a correctly aligned non-null pointer to avoid UB in write_bytes and
-            // copy_nonoverlapping.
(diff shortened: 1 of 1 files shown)


bjorn3:
If the second commit could be backported to the release-31.0.0 branch that would be appreciated.

alexcrichton:
For the backport I think that's reasonable, would you be ok sending a PR yourself? (no need for a subset of folks to send backports to older branches, it's ok for anyone to open a PR to the branch)

abrown (approved):
This change seems sensible to me but I'll let @alexcrichton comment as well since he's here.

PR #10348, opened 2025-03-06
meta: deduplicate source generation infrastructure

abrown:
When building the new `cranelift-assembler-x64` crate, I duplicated some `Formatter` functionality that we were using in `cranelift-codegen-meta` to generate both Rust and ISLE code. This change splits out that functionality into a separate crate that both `cranelift-assembler-x64` and `cranelift-codegen-meta` can depend on without depending on each other.

While doing this, I rationalized the divergence that had happened between both copies of `Formatter`; more details are noted in the commit messages. As a part of all of this, I will highlight that the `Formatter` now knows which language it is emitting (Rust or ISLE) and how to emit Rust brace-delimited blocks, a wide-ranging change.

All of this is to support the same goal as https://github.com/bytecodealliance/wasmtime/pull/10334. To avoid `OUT_DIR` issues, I believe we should move the ISLE generation from `cranelift-assembler-x64` to `cranelift-codegen-meta` (with the rest of the ISLE wrangling). This PR is a first step in that direction by sorting out the crate dependencies.


Cargo.toml
```diff
@@ -274,6 +274,7 @@ cranelift-fuzzgen = { path = "cranelift/fuzzgen" }
 cranelift-bforest = { path = "cranelift/bforest", version = "0.119.0" }
 cranelift-bitset = { path = "cranelift/bitset", version = "0.119.0" }
 cranelift-control = { path = "cranelift/control", version = "0.119.0" }
+cranelift-srcgen = { path = "cranelift/srcgen" }
```
abrown, line 277: It felt to me that this kind of crate isn't the same as the other Cranelift crates in terms of versioning, but I suspect we will have to give it a `0.119.0` version just to get it to pass the `publish.rs` script? I noticed that some fuzzing infrastructure crate had a `0.0.0` version but perhaps that isn't published?

cfallin (approved):
This is really nice; thanks for the cleanups/refactors!

PR #9901, opened 2024-12-24
Stop formatting ISLE generated code

SingleAccretion:
The tool stack overflows in default configurations (Windows on my machine (tm) / CI).

Example from a recent CI run: https://github.com/bytecodealliance/wasmtime/actions/runs/12472519540/job/34811457587?pr=9897:
```
warning: cranelift-codegen@0.116.0: Failed to run `rustfmt` on ISLE-generated code: Os { code: 32, kind: BrokenPipe, message: "Broken pipe" }
```

Closes #9899.

PR #10352, opened 2025-03-07
meta: move all ISLE-related generated code to `cranelift-codegen-meta`

abrown:
Previously, the new `cranelift-assembler-x64` crate would generate the assembler Rust code, the ISLE integration code, and the ISLE's accompanying Rust macro all in the same crate. Because these last two files were generated in the assembler's `OUT_DIR`, they caused issues for deterministic builds that expect those files in `cranelift-codegen`'s `OUT_DIR` (see #10348). This change fixes that by moving all of the assembler's generated integration code, the ISLE instruction definitions and their accompanying Rust macro, to the `cranelift-codegen-meta` crate.

This has the added advantage that `cranelift-assembler-x64` now can focus exclusively on being an assembler and not being concerned with the ISLE integration, which is substantial.


cranelift/assembler-x64/meta/src/lib.rs
```diff
@@ -11,42 +11,17 @@ use std::process::Command;
...
     eprintln!("Generating {}", out.display());
-    let mut fmt = Formatter::new(lang);
-    generator(&mut fmt, &instructions::list());
+    let mut fmt = Formatter::new(Language::Rust);
+    generate::rust_assembler(&mut fmt, &instructions::list());
     fmt.write(file, dir.as_ref()).unwrap();
-    if matches!(lang, Language::Rust) {
-        rustfmt(&out);
-    }
+    rustfmt(&out);
```
cfallin, line 24: FWIW, we removed the invocation of `rustfmt` in #9901 for the main meta crate; it's an artifact of an earlier time when we had the code checked into git, but we don't now, and it can cause issues in certain build environments -- probably ok to remove it here too?

cfallin (approved):
Nice!

PR #10354, opened 2025-03-07
chore: remove redundant words in comment

NinaLua:
remove redundant words in comment


winch/codegen/src/masm.rs
```diff
@@ -2110,7 +2110,7 @@ pub(crate) trait MacroAssembler {
     fn v128_neg(&mut self, op: WritableReg, kind: V128NegKind) -> Result<()>;
 
     /// Perform the shift operation specified by `kind`, by the shift amount specified by the 32-bit
-    /// integer at the top the the stack, on the 128-bit vector specified by the second value
+    /// integer at the top the stack, on the 128-bit vector specified by the second value
```
saulecabrera, line 2113:
```suggestion
    /// integer at the top of the stack, on the 128-bit vector specified by the second value
```

saulecabrera (commented):
LGTM, one minor thing that needs fixing. Not introduced by your changes, but might as well fix it while we are at it.

winch/codegen/src/masm.rs
NinaLua: Thanks! Modified.

saulecabrera (approved):
Thanks!

PR #10355, opened 2025-03-07
Stack switching: Remove stack_chain field from VMContext

frank-emrich:
This PR is part of a series that adds support for the [Wasm stack switching proposal](https://github.com/WebAssembly/stack-switching). The explainer document for the proposal is [here](https://github.com/WebAssembly/stack-switching/blob/main/proposals/stack-switching/Explainer.md). There's a tracking issue for the overall progress: #10248

This PR is effectively a revert of #10265: As @fitzgen suggested [there](https://github.com/bytecodealliance/wasmtime/pull/10265#issuecomment-2688899292), it is actually more natural to store the actual stack chain in the `VMStoreContext` (née `VMRuntimeLimits`) instead of having it be a separate field of the `StoreOpaque`.
Therefore, we don't actually need a pointer to it in the `VMContext`. Instead, can now just rely on the fact that the `VMContext` already has a pointer to the `VMStoreContext`, where the stack chain will now we stored in subsequent PRs.


alexcrichton (approved):

fitzgen:
Thanks!

PR #10360, opened 2025-03-08
feat: Add hexadecimal integer parsing in CLI arguments

VolodymyrBg:
Previously, the CLI interface would only accept base-10 (decimal) integers when 
passing arguments to WebAssembly functions. This change adds support for 
hexadecimal notation with the '0x' prefix.

Now users can pass hexadecimal values directly to functions that expect integer
arguments, making it more convenient to work with bit patterns, addresses, and
other values commonly represented in hexadecimal format.

Fixes the TODO comment in the run.rs file that requested this feature.

### DIFF
--- a/src/commands/run.rs
+++ b/src/commands/run.rs
@@ -526,11 +526,9 @@ impl RunCommand {
                 .to_str()
                 .ok_or_else(|| anyhow!("argument is not valid utf-8: {val:?}"))?;
             values.push(match ty {
-                // TODO: integer parsing here should handle hexadecimal notation
-                // like `0x0...`, but the Rust standard library currently only
-                // parses base-10 representations.
-                ValType::I32 => Val::I32(val.parse()?),
-                ValType::I64 => Val::I64(val.parse()?),
+                // Supports both decimal and hexadecimal notation (with 0x prefix)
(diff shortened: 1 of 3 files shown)


alexcrichton:
I think this may have lost the tests from https://github.com/bytecodealliance/wasmtime/pull/9989?

VolodymyrBg:
@alexcrichton corrected

src/commands/run.rs
```diff
@@ -526,11 +526,15 @@ impl RunCommand {
...
             values.push(match ty {
-                // TODO: integer parsing here should handle hexadecimal notation
-                // like `0x0...`, but the Rust standard library currently only
-                // parses base-10 representations.
-                ValType::I32 => Val::I32(val.parse()?),
-                ValType::I64 => Val::I64(val.parse()?),
+                // Supports both decimal and hexadecimal notation (with 0x prefix)
+                ValType::I32 => Val::I32(
+                    val.parse::<i32>()
+                        .or_else(|_| i32::from_str_radix(&val.trim_start_matches("0x"), 16))?,
```
alexcrichton, line 532: This unfortunately has the behavior that `ff` is interpreted as 255 since the `trim_start_matches` is an optional trim rather than a required trim. Could this be refactored to require the `0x` prefix for hexadecimal arguments?

alexcrichton (approved):
Thanks!

PR #10370, opened 2025-03-11
Winch: Fix issue with multivalue returns

### LINKED ISSUE #10357: winch simd: Diverging behavior from Cranelift with params/results
This input:

```wasm
(module
  (func (export "") (result i64 v128 v128)
    i32.const 0
    v128.const i32x4 0x00000000 0x00000000 0x68732efe 0x74727473
    i64.const 1
    call 1
  )

  (func (param i32 v128 i64) (result i64 v128 v128)
    i64.const 0
    local.get 1
    v128.const i32x4 0x00000000 0x00000000 0x00000000 0x00000000
  )
)
```

shows:

```
$ cargo run -q run -C compiler=winch testcase0.shrunken.wat | shasum
warning: using `--invoke` with a function that returns values is experimental and may break in the future
bc51dd722ba91bc34e6d9650b9682c8c11e4acec  -
$ cargo run -q run testcase0.shrunken.wat | shasum
warning: using `--invoke` with a function that returns values is experimental and may break in the future
a9612e752ea5fb1d25ae07d25dcbd6f5084334ab  -
```

cc @jeffcharles

jeffcharles:
Fixes #10357. There's a subtle bug with the existing `memmove` implementation where if a `HighToLow` `memmove` is being performed and the source and destination offsets overlap and more than 8 bytes are being moved, then part of the data is copied multiple times into different destination offsets. Performing the move in the opposite direction fixes the problem.

### DIFF
--- a/tests/disas/winch/x64/local/v128_multivalue.wat
+++ b/tests/disas/winch/x64/local/v128_multivalue.wat
@@ -1,0 +1,44 @@
+;;! target = "x86_64"
+;;! test = "winch"
+
+(module
+  (func (export "") (param v128) (result i64 v128 i64)
+    i64.const 0
+    local.get 0
+    i64.const 0
+  )
+)
+;; wasm[0]::function[0]:
+;;       pushq   %rbp
+;;       movq    %rsp, %rbp
+;;       movq    8(%rsi), %r11
+;;       movq    0x10(%r11), %r11
+;;       addq    $0x48, %r11
+;;       cmpq    %rsp, %r11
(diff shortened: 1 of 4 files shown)


saulecabrera:
I'll take a look at this one.

saulecabrera (approved):
LGTM,thanks. I'll run the fuzzer locally for a bit before landing.

saulecabrera:
Ran the fuzzer over night and nothing was reported.

Issue #10181, opened 2025-02-04
Wasm GC: internal assertion failure in JIT code, leading to illegal instruction fault


vouillon:
### Test Case

[constprop.zip](https://github.com/user-attachments/files/18657925/constprop.zip)

### Steps to Reproduce

Run the following command:
```
./target/debug/wasmtime  -W=all-proposals=y constprop.wasm
```

### Expected Results

This works fine if one add the `-C collector=null` option:
```
$ ./target/debug/wasmtime -C collector=null  -W=all-proposals=y constprop.wasm
booleans: passed
integers: passed
floats: passed
32-bit integers: passed
native integers: passed
64-bit integers: passed
integer conversions: passed
32-bit integer conversions: passed
native integer conversions: passed
64-bit integer conversions: passed
```

### Actual Results

This crashes with an `Illegal instruction` error.

### Versions and Environment

Wasmtime commit: 0e0560087beff704bc111a1abb1d6a80c1b5da70

Operating system: Linux

Architecture: x64

fitzgen:
Thanks for the bug report!

If you have time, minimizing the test case `wat` with `creduce` would be very appreciated and make diagnosing and fixing this easier. If not, I'll get around to that eventually.

just-an-engineer:
I'll take this.
After some initial looking, it has to do with the async stuff, I'm guessing it pops the wrong address somehow.

just-an-engineer:
Where is this wasm file from, and how did you compile it?
I suspect it may be from an invalid wasm file, which the code _should_ check for currently, but doesn't seem to. Running `wasm-validate` on it gives me an error of invalid type a few bytes in, but I imagine that tools stops after the first error, so no idea how many more.
Anyways, it could be that there's some invalid stuff here that operates outside of the specification, that leds to an incorrect address being jumped to during the async execution stuffs.

vouillon:
This code relies on the Wasm GC proposal. This proposal is not supported by `wasm-validate`. That's why you get errors.
The code is generated using [wasm_of_ocaml](https://github.com/ocsigen/js_of_ocaml/blob/7c6c20949ebff14ba0d069ba1b5ff99a0f3be434/README_wasm_of_ocaml.md), which compiles OCaml code to Wam.
[...]

PR #10371, opened 2025-03-11
Fix an issue with DRC collector's barriers and `(extern.convert_any (ref.i31 ...))`

fitzgen:
Found during local fuzzing. This might be a fix for issue https://github.com/bytecodealliance/wasmtime/issues/10181 as well, but I want to get confirmation before marking it fixed, and IIUC there isn't a minimal test case in there at the moment.

### DIFF
--- a/crates/cranelift/src/gc/enabled.rs
+++ b/crates/cranelift/src/gc/enabled.rs
@@ -1388,8 +1388,14 @@ impl FuncEnvironment<'_> {
             | WasmHeapType::ConcreteStruct(_)
             | WasmHeapType::None => false,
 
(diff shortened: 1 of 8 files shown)


alexcrichton (approved):

vouillon:
There was a reduced test case (https://github.com/bytecodealliance/wasmtime/issues/10181#issuecomment-2642608130), but it seems it is no longer crashing. The initial large test case is still crashing.

fitzgen:
@vouillon Thanks, looks like there are still some outstanding internal assertions in JIT code failing then. As always, a reduced test case would be very appreciated and would make debugging and fixing much easier!

Issue #10369, opened 2025-03-11
Tracking issues for CI spurious network failures with example.com


alexcrichton:
CI just [spuriously failed](https://github.com/bytecodealliance/wasmtime/actions/runs/13790687014/job/38569638778) where a test hits `example.com` as a live testing service. This was known that it was probably going to be flaky so I wanted to have an issue here tracking this so we know what's going on. I've also opened an issue at https://github.com/WebAssembly/wasi-tls/issues/9 for more error information

cc @jsturtevant and @badeend 

Currently the failure looks like this:

```
running 1 test

thread 'main' panicked at crates/test-programs/src/bin/tls_sample_application.rs:33:14:
test tls_sample_application ... FAILED
called `Result::unwrap()` on an `Err` value: ()

note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
failures:
error: test failed, to rerun pass `-p wasmtime-wasi-tls --test main`

---- tls_sample_application stdout ----
Error: error while executing at wasm backtrace:
    0: 0x1cd96 - tls_sample_application-3d73890cb2b55afd.wasm!__rust_start_panic
    1: 0x1cce8 - tls_sample_application-3d73890cb2b55afd.wasm!rust_panic
    2: 0x1ccbb - tls_sample_application-3d73890cb2b55afd.wasm!std::panicking::rust_panic_with_hook::he1beb51ce54153b5
    3: 0x1bf2e - tls_sample_application-3d73890cb2b55afd.wasm!std::panicking::begin_panic_handler::{{closure}}::hdbff82bd9d56f6f9
    4: 0x1be9a - tls_sample_application-3d73890cb2b55afd.wasm!std::sys::backtrace::__rust_end_short_backtrace::h5bb26fcae04a79f5
    5: 0x1c64f - tls_sample_application-3d73890cb2b55afd.wasm!rust_begin_unwind
    6: 0x21b2b - tls_sample_application-3d73890cb2b55afd.wasm!core::panicking::panic_fmt::h2781e55e5f70e742
    7: 0x2339b - tls_sample_application-3d73890cb2b55afd.wasm!core::result::unwrap_failed::he956e17bd675892a
    8: 0x5fea - tls_sample_application-3d73890cb2b55afd.wasm!tls_sample_application::test_tls_sample_application::h85c84c9b597e7424
[...]

PR #10377, opened 2025-03-12
wasi-tls: Add a second endpoint to make the test less flakey

jsturtevant:
Helps address https://github.com/bytecodealliance/wasmtime/issues/10369 by adding a second endpoint.  One reason for testing the external endpoint is to make sure the set up of the system cert store works when calling external sites, so this preserves the external call but should also help the tests be less flakey.

### DIFF
--- a/crates/test-programs/src/bin/tls_sample_application.rs
+++ b/crates/test-programs/src/bin/tls_sample_application.rs
@@ -1,25 +1,24 @@
+use anyhow::{Context, Result};
 use core::str;
-
 use test_programs::wasi::sockets::network::{IpSocketAddress, Network};
 use test_programs::wasi::sockets::tcp::{ShutdownType, TcpSocket};
 use test_programs::wasi::tls::types::ClientHandshake;
 
-fn test_tls_sample_application() {
+fn make_tls_request(domain: &str) -> Result<String> {
     const PORT: u16 = 443;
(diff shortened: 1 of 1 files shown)


crates/test-programs/src/bin/tls_sample_application.rs
```diff
@@ -28,19 +27,32 @@ fn test_tls_sample_application() {
...
-    assert!(response.contains("HTTP/1.1 200 OK"));
+fn test_tls_sample_application() {
+    // since this is testing remote endpoint to ensure system cert store works
+    // the test uses a couple different endpoints to reduce the number of flakes
+    const DOMAIN1: &'static str = "example.com";
+    const DOMAIN2: &'static str = "api.github.com";
+
+    let response1 = make_tls_request(DOMAIN1).unwrap();
+    let response2 = make_tls_request(DOMAIN2).unwrap();
+
```
alexcrichton, line 51: Should this avoid a `.unwrap()` here since I think that would translate to both domains being required to succeed instead of just one?

jsturtevant: yes, I had that at initially and was trying something else and forgot to go back🤦

jsturtevant: updated

alexcrichton (approved):

PR #10384, opened 2025-03-12
Change `allow(missing_docs)` to `expect(..)`

alexcrichton:
This wasn't possible when `expect` was first introduced due to a change being required in upstream rust-lang/rust. That change rust-lang/rust#130025) has now rode enough trains to be in our MSRV, so we can expect missing docs now instead of just allowing it.

### DIFF
--- a/cranelift/assembler-x64/src/rex.rs
+++ b/cranelift/assembler-x64/src/rex.rs
@@ -121,7 +121,6 @@ impl RexFlags {
 }
 
 #[derive(Copy, Clone)]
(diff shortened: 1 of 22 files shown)


cranelift/foo.clif
```diff
@@ -0,0 +1,7 @@
+function u0:0(i64, i64) -> i64 {
```
dicej, line 1: How is this file related to the rest of the PR?

alexcrichton: Oops it is not, just me being bad at git

PR #10386, opened 2025-03-12
Update Wasmi fuzzing oracle to v0.42

Robbepop:
Wasmi v0.42 now supports the following Wasm proposals:

- `custom-page-sizes`
- `memory64`
- `wide-arithmetic`

I think they are a welcome addition for Wasmtime's differential fuzzing. :)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -35,18 +35,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ahash"
-version = "0.8.11"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


fitzgen (approved):
Nice! Thanks!

fitzgen:
This will need a `cargo vet` before landing. Don't have time at this exact moment but can maybe get to it later today if no one else beats me to it.

Robbepop:
If it helps we could even disable some crate features that are not strictly needed for differential fuzzing, such as `wat` (Wat files support) and we could enable `prefer-btree-collections` to avoid the dependency on `string-interner` and transitively also `hashbrown`.

Issue #9350, opened 2024-10-01
Reuse linear memories, and our code to emit their bound checks, for Wasm GC heaps


fitzgen:
We have a bunch of knobs and associated optimizations and special cases for linear memory bounds checks. We should reuse all that for Wasm GC heaps, letting them be more configurable and emitting better code for their bounds checks.

PR #10388, opened 2025-03-12
Stack switching: Infrastructure and runtime support

frank-emrich:
This PR is part of a series that adds support for the [Wasm stack switching proposal](https://github.com/WebAssembly/stack-switching). The explainer document for the proposal is [here](https://github.com/WebAssembly/stack-switching/blob/main/proposals/stack-switching/Explainer.md). There's a tracking issue describing the overall progress and limitations here: #10248
The draft PR #10177 contains the whole (initial) implementation of the proposal. 

This PR contains all of the necessary infrastructure and runtime support. In other words, this contains the entire implementation, except for codegen. Tests are added at the end, after codegen has also been added in a follow-up PR.

This was developed together with @dhil.

### General implementation notes

In Wasm, continuations are represented by values of type `(ref $ct)`, where `$ct` is a new composite type/heap type for continuations.
In the implementation, these are represented by values of type `VMContObj`. These are fat pointers, consisting of a sequence counter, and a pointer to a `VMContRef`. The latter type is used for the actual representation of continuations.
The sequence counter part of `VMContObj`s is used to check that every continuation value can only be used once.

The `VMStoreContext` is extended to contain a "stack chain": It indicates what stack we are currently executing on. Logically, this is a linked list of stacks, since each continuation has a parent field. The chain stored in the `VMStoreContext` always ends with a value representing the initial stack. This is the stack we were on when entering Wasm, which will usually be the main stack. 

### Memory Management

Currently, memory management is very basic: The `StoreOpaque` provides a new method for allocation a new `VMContRef`, and keeps track of all continuations created this way. Continuations are never deallocated at the moment, meaning that they live until the store itself is deallocated. 

The stack memory used by each allocation (represented by type `ContinuationStack`) is always `mmap`-ed, with a guard page at the end. There is currently no support for growing stacks, or segmented stacks.

### Backtrace generation

The existing backtrace generation mechanism is extended to be fully aware of continuations: When generating backtraces, the entire chain of continuation is traversed, not just the stack frames of the currently active stack/continuation.

### Entering/Exiting Wasm

Prior to this PR, there were two separate mechanisms that save and restore some state of the runtime when entering/exiting Wasm:
1. The functions `enter_wasm` and `exit_wasm` in `func.rs`
2. `CallThreadState` saves and restores (on `drop`) parts of the `VMStoreContext`

This PR consolidates these two mechanism, because it requires some additional state to be updated and restored on enter/exit:
the type `wasmtime::runtime::func::EntryStoreContext` now stores all of the required runtime state and ensures that it's restored when exiting Wasm.

### Tables

- The runtime part for resizing and filling tables containing continuations is added.
- Note that there is a potentially controversial change to how the pooling allocator works for tables: Since a `VMContObj` is 2 pointers wide, the maximum size of an entry in a table doubles. As a result, this PR doubles the amount of virtual memory space occupied by the table pool (but the amount of pages actually used stays the same).


fitzgen:
Haven't dug into the details of the PR yet, but I have some initial questions after reading the OP.

> The stack memory used by each allocation (represented by type `ContinuationStack`) is always `mmap`-ed, with a guard page at the end. There is currently no support for growing stacks, or segmented stacks.

Is there a particular reason this does not use the [`InstanceAllocatorImpl::[de]allocate_fiber_stack`](https://github.com/bytecodealliance/wasmtime/blob/9da52ede33a9f8996832b0210579ef15296addbd/crates/wasmtime/src/runtime/vm/instance/allocator.rs#L297-L309) methods via `[store.engine().allocator().[de]allocate_fiber_stack()`](https://github.com/bytecodealliance/wasmtime/blob/9da52ede33a9f8996832b0210579ef15296addbd/crates/wasmtime/src/engine.rs#L659)? That would make it so that this feature automatically integrates with the `wasmtime::Config`, its various knobs for stack sizes and such, and whether the pooling or on-demand allocator is in use.

> Note that there is a potentially controversial change to how the pooling allocator works for tables: Since a `VMContObj` is 2 pointers wide, the maximum size of an entry in a table doubles. As a result, this PR doubles the amount of virtual memory space occupied by the table pool (but the amount of pages actually used stays the same).

Is this because `(ref cont)`/`VMContObj` tables are being shoe-horned into the same underlying representation as `anyref`/`externref`/`VMGcRef` tables? Because `VMContObj` has a different representation from `VMGcRef`, I think it would be best to add a new `wasmtime::runtime::vm::Table` variant for `VMContObj`, the same way that we have [different variants for `VMGcRef` tables versus `*const VMFuncRef` tables](https://github.com/bytecodealliance/wasmtime/blob/9da52ede33a9f8996832b0210579ef15296addbd/crates/wasmtime/src/runtime/vm/table.rs#L144-L147).

crates/c-api/include/wasmtime/linker.h
```diff
@@ -66,6 +66,16 @@ WASM_API_EXTERN void wasmtime_linker_delete(wasmtime_linker_t *linker);
...
 
+/**
+ * \brief Configures whether the given Linker will allow unknown exports from
+ * command modules.
+ *
+ * By default this setting is `false`.
+ */
+WASM_API_EXTERN void
+wasmtime_linker_allow_unknown_exports(wasmtime_linker_t *linker,
+                                      bool allow_unknown_exports);
```
fitzgen, line 77: I don't see a corresponding function definition in `crates/c-api`, I think it is missing?

crates/c-api/include/wasmtime/config.h
```diff
@@ -250,6 +250,22 @@ WASMTIME_CONFIG_PROP(void, wasm_wide_arithmetic, bool)
 
 #ifdef WASMTIME_FEATURE_COMPILER
 
+/**
+ * \brief Configures whether the WebAssembly function references
+ * proposal is enabled.
+ *
+ * This setting is `false` by default.
+ */
+WASMTIME_CONFIG_PROP(void, wasm_function_references, bool)
```
fitzgen, line 259: Similarly, I don't see a function definition for this one either.

crates/cli-flags/src/lib.rs
```diff
@@ -331,6 +331,8 @@ wasmtime_option_group! {
         pub trap_on_grow_failure: Option<bool>,
         /// Maximum execution time of wasm code before timing out (1, 2s, 100ms, etc)
         pub timeout: Option<Duration>,
+        /// Size of stacks created with cont.new instructions
+        pub stack_switching_stack_size: Option<usize>,
```
fitzgen, line 335: At least for now, I'm not sure it is worth having separate config knobs for stacks created via `cont.new` versus API-level fibers. This effectively creates two different kinds of stacks in the system, and I'd prefer to try and keep things simple for as long as we can by having a single kind of stack that we config and allocate and handle in the system.

crates/cranelift/src/gc/enabled.rs
```diff
@@ -148,7 +148,12 @@ fn read_field_at_addr(
                         .call(get_interned_func_ref, &[vmctx, func_ref_id, expected_ty]);
                     builder.func.dfg.first_result(call_inst)
                 }
-                WasmHeapTopType::Cont => todo!(), // FIXME: #10248 stack switching support.
+                WasmHeapTopType::Cont => {
+                    // TODO(#10248) GC integration for stack switching
+                    return Err(wasmtime_environ::WasmError::Unsupported(
+                        "Stack switching feature not compatbile with GC, yet".to_string(),
+                    ));
```
fitzgen, line 155: Not something that needs to be handled in this PR, but there is no reason we can't enable GC and also maintain the existing memory management for continuations, where they have the same lifetime as the `Store`. We don't have to implement full GC of continuations before we allow enabling both GC and stack switching at the same time.

crates/cranelift/src/lib.rs
```diff
@@ -61,6 +61,12 @@ pub const TRAP_HEAP_MISALIGNED: TrapCode =
     TrapCode::unwrap_user(Trap::HeapMisaligned as u8 + TRAP_OFFSET);
 pub const TRAP_TABLE_OUT_OF_BOUNDS: TrapCode =
     TrapCode::unwrap_user(Trap::TableOutOfBounds as u8 + TRAP_OFFSET);
+pub const TRAP_UNHANDLED_TAG: TrapCode =
+    TrapCode::unwrap_user(Trap::UnhandledTag as u8 + TRAP_OFFSET);
+pub const TRAP_CONTINUATION_ALREADY_CONSUMED: TrapCode =
+    TrapCode::unwrap_user(Trap::ContinuationAlreadyConsumed as u8 + TRAP_OFFSET);
+pub const TRAP_DELETE_ME_DEBUG_ASSERTION: TrapCode =
+    TrapCode::unwrap_user(Trap::DeleteMeDebugAssertion as u8 + TRAP_OFFSET);
```
fitzgen, line 69: FWIW, we already have `TRAP_INTERNAL_ASSERT` that we use for debug-only assertions inside JIT code.

crates/environ/src/builtin.rs
```diff
@@ -206,6 +206,32 @@ macro_rules! foreach_builtin_function {
...
+            // FIXME(frank-emrich) The next three builtins are used by the debug printing mechanism.
+            // They are not supposed to be part of the final upstreamed code.
+            //
+            // Prints a 'static str, represented as a
+            // pointer and a length.
+            delete_me_print_str(vmctx: vmctx, s: pointer, len : u64);
+            // Prints integer
+            delete_me_print_int(vmctx: vmctx, arg : u64);
+            // Prints pointer, formatted as hex.
+            delete_me_print_pointer(vmctx: vmctx, arg : pointer);
```
fitzgen, line 222: Were these going to be removed in the next PR, or should they be removed now? Maybe throw an item into the meta task list if not now.

crates/environ/src/stack_switching.rs
```diff
@@ -0,0 +1,78 @@
+//! This module contains basic type definitions used by the implementation of
+//! the stack switching proposal.
+
+/// FIXME(frank-emrich) Will remove in the final upstreamed version
+#[allow(dead_code, reason = "Only accessed in debug builds")]
+pub const ENABLE_DEBUG_PRINTING: bool = false;
```
fitzgen, line 6: Ditto to the debug printing stuff in this file.

crates/environ/src/stack_switching.rs
```diff
@@ -0,0 +1,78 @@
...
+#[macro_export]
+macro_rules! debug_println {
+    ($( $args:expr ),+ ) => {
+        #[cfg(debug_assertions)]
+        if ENABLE_DEBUG_PRINTING {
+            #[cfg(feature = "std")]
+            println!($($args),*);
+        }
+    }
+}
```
fitzgen, line 18: fyi, I tend to do `log::trace!(...)` for debug prints (both temporary ones that I will remove, and for ones that are useful enough to keep in tree).

when running the CLI, you can see them by setting the `WASMTIME_LOG=trace` env var, or only a specific crate/module via `WASMTIME_LOG=wasmtime::runtime::vm=trace`, or multiple crates/modules at different levels via `WASMTIME_LOG=debug,regalloc2=info,cranelift_codegen::machinst=trace`.

When running tests instead of the CLI, `WASMTIME_LOG` becomes `RUST_LOG`. Most tests initialize the env-logger, but if you aren't seeing the logs, then you can put `let _ = env_logger::try_init();` at the start of the test (or ideally in a shared test helper function that a bunch of related tests already call at the beginning of their execution).

fitzgen (commented):
(review is still in progress, but I need to context switch for a little bit and might as well share what I have so far)

crates/environ/src/stack_switching.rs
fitzgen: * https://docs.wasmtime.dev/cli-logging.html?highlight=wasmtime_log#logging-in-the-wasmtime-cli
* https://docs.rs/log/latest/log/
* https://docs.rs/env_logger/latest/env_logger/

crates/wasmtime/Cargo.toml
```diff
@@ -308,6 +309,11 @@ threads = [
   "std",
 ]
 
+stack-switching = [
+  "std",
+  "runtime"
+]
```
fitzgen, line 315: Two things:

1. Why does stack-switching need `std`?

2. We should be able to enable stack-switching for a compiler-without-runtime build to be able to compile stack-switching Wasm to be run elsewhere, and similarly should be able to enable stack-switching in a runtime-without-compiler build to run elsewhere-compiled stack-switching Wasm modules. As written, enabling the stack-switching feature will always enable the runtime, making the compiler-without-runtime version impossible. That is, unless compilation of stack-switching isn't gated on any cargo feature, but that is kind of funky given that this cargo feature exists and I expect users would think "I want to compile Wasm programs that use stack-switching, therefore I will enable this feature". This is a long way of basically saying, I think this should not unconditionally enable the runtime and, once compilation is added in the next PR, we ultimately end up with something like this:

   ```rust
   stack-switching = [
     "wasmtime-cranelift?/stack-switching",
   ]
   ```

Neither of these things necessarily needs to be addressed now, before this PR merges, but it may end up being easier doing it now before more stuff gets entangled, making it harder to pull the `cfg`s apart.

fitzgen: You'll also want to add some CI checks for these features being enabled/disabled in various combinations like this:

https://github.com/bytecodealliance/wasmtime/blob/main/.github/workflows/main.yml#L363-L365

crates/wasmtime/src/runtime/externals/table.rs
```diff
@@ -188,6 +188,11 @@ impl Table {
                         ty => unreachable!("not a top type: {ty:?}"),
                     }
                 }
+
+                runtime::TableElement::ContRef(_c) => {
+                    // TODO(#10248) Required to support stack switching in the embedder API.
+                    unimplemented!()
```
fitzgen, line 194: Can you add an item to the meta task list for the embedder API, if you haven't already?

crates/wasmtime/src/runtime/func.rs
```diff
@@ -1597,103 +1600,173 @@ pub(crate) fn invoke_wasm_and_catch_traps<T>(
     closure: impl FnMut(NonNull<VMContext>, Option<InterpreterRef<'_>>) -> bool,
 ) -> Result<()> {
     unsafe {
-        let exit = enter_wasm(store);
+        // The `enter_wasm` call below will reset the store's `stack_chain` to
+        // a new `InitialStack`, pointing to the stack-allocated
+        // `initial_stack_csi`.
+        let mut initial_stack_csi = VMCommonStackInformation::running_default();
```
fitzgen, line 1606: Can you run the function call microbenchmarks on `main` and on this branch? I want to make sure that we aren't regressing performance here, as this path is very hot, and if we are we may need to figure out how to `cfg(...)` some of this stuff so that it only happens when the `stack-switching` feature is enabled.

This will run the microbenchmarks and will check the results for statistical significance:

```shell-session
$ git checkout main
$ cargo bench --bench call -- --save-baseline main
$ git checkout stack-switching-infra
$ cargo bench --bench call -- --baseline main
```

I can help figure things out if we need to do something here.

crates/wasmtime/src/runtime/func.rs
```diff
@@ -1597,103 +1600,173 @@ pub(crate) fn invoke_wasm_and_catch_traps<T>(
...
+        let previous_runtime_state = EntryStoreContext::enter_wasm(store, &mut initial_stack_csi);
 
         if let Err(trap) = store.0.call_hook(CallHook::CallingWasm) {
-            exit_wasm(store, exit);
             return Err(trap);
         }
-        let result = crate::runtime::vm::catch_traps(store, closure);
-        exit_wasm(store, exit);
+        let result = crate::runtime::vm::catch_traps(store, &previous_runtime_state, closure);
+        core::mem::drop(previous_runtime_state);
```
fitzgen, line 1617: This `mem::drop` dance is kind of funky -- can we instead pass ownership into `catch_traps` and let it handle the dropping, if necessary? Or are there callers that need to do different things?

crates/wasmtime/src/runtime/func.rs
```diff
@@ -1597,103 +1600,173 @@ pub(crate) fn invoke_wasm_and_catch_traps<T>(
...
-        debug_assert!(store.engine().target().is_pulley());
-        usize::MAX
-    };
+/// This type helps managing the state of the runtime when entering and exiting
+/// Wasm. To this end, it contains a subset of the data in `VMStoreContext`..
+/// Upon entering Wasm, it updates various runtime fields and their
+/// original values saved in this struct. Upon exiting Wasm, the previous values
+/// are restored.
+// FIXME(frank-emrich) Do the fields in here need to be (Unsafe)Cells?
+pub struct EntryStoreContext {
```
fitzgen, line 1629: If they are written to directly from Wasm JIT code, then probably yes. This struct should also be `#[repr(C)]` in that case.

If it is not written to directly from Wasm, then it would only need to be unsafe cells if you are ever violating Rust's mutable/exclusive xor immutable/shared semantics.

I haven't read enough of these changes to say any more than that yet.

fitzgen (commented):
(again, review still in progress, leaving intermediate notes)

crates/c-api/include/wasmtime/linker.h
dhil: I think this is a stray declaration from some of our downstream experiments. It has been needed to capture the C shadow stack pointer, that clang generates. It must be kept in sync with the stack switching done by the engine. To be clear, this is just due to the fact that clang doesn't know about stack switching.

crates/c-api/include/wasmtime/config.h
dhil: Same reason as above.

crates/c-api/include/wasmtime/linker.h
fitzgen: FWIW, exposing these things the C API would be good in general, so if you want to do it here or in a new PR, you're more than welcome to!

dhil: Yes sure! I am leaning towards adding this in a follow-up PR focused on the C API. What do you think @frank-emrich ?

crates/environ/src/stack_switching.rs
```diff
@@ -0,0 +1,78 @@
...
+
+/// Runtime configuration options for stack switching that can be set
+/// via the command line.
+///
+/// Part of wasmtime::config::Config type (which is not in scope in this crate).
+#[derive(Debug, Clone)]
+pub struct StackSwitchingConfig {
+    /// The (fixed) size of a continuation stack.
+    pub stack_size: usize,
+}
```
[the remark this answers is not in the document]
dhil: As I discussed with @alexcrichton (https://github.com/bytecodealliance/wasmtime/pull/10177/files#r1965733715), I'd be OK with no bespoke stack switching configuration for now.

crates/cli-flags/src/lib.rs
dhil: This is an interesting question, which Frank and I have discussed in the past. If my memory serves me right, then our layout is slightly different. However perhaps we can unify the layouts (or we can perhaps simply adopt the fibers' layout). I am not too sure about the implications, I think @frank-emrich has the key knowledge to best answer this question.

fitzgen: The automatic pooling allocator integration would also get you fast allocation of new stacks from a pool, which would look nice on your benchmarks and what have you ;)

crates/environ/src/builtin.rs
dhil: My take is that they shouldn't make it into the `main` branch -- probably a stray include?

frank-emrich:
(Sorry for the long delay, I've moved to London and started a new job, where I needed to settle in first)

@fitzgen 
> Is there a particular reason this does not use the [InstanceAllocatorImpl::[de]allocate_fiber_stack](https://github.com/bytecodealliance/wasmtime/blob/9da52ede33a9f8996832b0210579ef15296addbd/crates/wasmtime/src/runtime/vm/instance/allocator.rs#L297-L309) methods via [store.engine().allocator().[de]allocate_fiber_stack()](https://github.com/bytecodealliance/wasmtime/blob/9da52ede33a9f8996832b0210579ef15296addbd/crates/wasmtime/src/engine.rs#L659)? That would make it so that this feature automatically integrates with the wasmtime::Config, its various knobs for stack sizes and such, and whether the pooling or on-demand allocator is in use.

The reason we do not use these in verbatim is that our stacks are not the same as `wasmtime_fiber::FiberStack`. Trying to make our stacks compatible with that type would be quite difficult and IMO not worth the effort.
However, our medium-term plan was to add similar allocation and deallocation functions to `InstanceAllocatorImpl`. This shouldn't be too much work, but something we didn't see through for now. In particular, @dhil is very interested in restoring support for pooled allocation of stack. We had implemented our own interface for that (outside of `InstanceAllocatorImpl`) and it greatly improves performance in those use cases where you frequently (de-) allocate continuations.

> Is this because (ref cont)/VMContObj tables are being shoe-horned into the same underlying representation as anyref/externref/VMGcRef tables? Because VMContObj has a different representation from VMGcRef, I think it would be best to add a new wasmtime::runtime::vm::Table variant for VMContObj, the same way that we have [different variants for VMGcRef tables versus *const VMFuncRef tables](https://github.com/bytecodealliance/wasmtime/blob/9da52ede33a9f8996832b0210579ef15296addbd/crates/wasmtime/src/runtime/vm/table.rs#L144-L147).

We actually extend `StaticTable` and `TableElement` in that file with appropriate variants for our types. The problem is that in [TablePool::new](https://github.com/bytecodealliance/wasmtime/blob/9da52ede33a9f8996832b0210579ef15296addbd/crates/wasmtime/src/runtime/vm/instance/allocator/pooling/table_pool.rs#L32) we need to `mmap` enough memory to hold `limits.total_tables` * `limits.table_elements` entries, independent from what data is stored in those tables. In order to support tables storing continuations pointers, this effectively doubles the required size of the mmapped area.

crates/cli-flags/src/lib.rs
frank-emrich: From a technical perspective, there is no need for us to have our own config setting for the stack size. 
I guess the reason we didn't just re-use the existing `stack_size` option is that the latter always felt like a soft limit to me: AFAIK you can still occupy stack space beyond that limit, for example with host functions.
But I have no strong feelings about re-using the existing `stack_size` config option, or merging our `stack_switching_stack_size` with `async_stack_size`.

crates/environ/src/builtin.rs
frank-emrich: Good question! These only exist for some hacky debug printing code that should not end up in `main`. Initially, when we thought that there would only be a single, big PR, the plan was to keep them during reviewing, and then remove them at the last moment before merging. Now that we will land this in smaller pieces, we could either remove them now, or once all PRs from the series have landed.

The reason why we don't want to land this is explained [here](https://github.com/bytecodealliance/wasmtime/pull/10177/files#diff-0c7bbb4f13d0c3d3403d05e6d5f32beee56eb14e2f64f37f5b2a8b2ee6e78fc2R61): On the codegen side, we just emit the addresses of Rust string literals as literals into generated code, with no relocation information. So this crashes and burns if you ever try to re-use the same generated code from a different invocation of `wamtime`.

crates/wasmtime/src/runtime/vm/stack_switching.rs
```diff
@@ -0,0 +1,705 @@
...
+    Resume = wasmtime_environ::stack_switching::CONTROL_EFFECT_RESUME_DISCRIMINANT,
+    /// Used to signal that a continuation has invoked a `suspend` instruction.
+    Suspend {
+        /// The index of the handler to be used in the parent continuation to
+        /// switch back to.
+        handler_index: u32,
+    } = wasmtime_environ::stack_switching::CONTROL_EFFECT_SUSPEND_DISCRIMINANT,
+    /// Used to signal that a continuation has invoked a `suspend` instruction.
+    Switch = wasmtime_environ::stack_switching::CONTROL_EFFECT_SWITCH_DISCRIMINANT,
+}
```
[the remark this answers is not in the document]
frank-emrich: Haha, these were in `wasmtime_environ` originally, but then @alexcrichton suggested moving them into the runtime. I don't mind it either way.

crates/wasmtime/src/runtime/func.rs
frank-emrich: Yeah, that `drop` call is definitely not pretty. We could get away with not explicitly calling `drop`, but managing the scopes here slightly differently. But that seems equally funky.

The main reason for all this new `EntryStoreContext` business is the following:  Previously, there were two places that do some updates to the runtime on entering and exiting Wasm: In `catch_traps` (together with `CallThreadState`'s `drop`) and here in `func.rs` (using `enter_wasm`/`exit_wasm`). I believe @alexcrichton mentioned that this split is mostly for legacy reasons due to how to things were separated into different crates until recently.

For stack switching, we now require an additional update to the runtime on entering and exiting. Intuitively, it felt to me like these changes should be made in `enter_wasm`/`exit_wasm`, but that doesn't quite work: We need to undo these changes if you `panic` out of a Wasm execution, so I would indeed need to do all of this in `catch_traps` instead, so that `CallThreadState`'s `drop` can do the steps needed on exiting the runtime.

But updating the stack chain in `catch_traps` to signal that we are entering Wasm seemed awkward to me, so I decided to just consolidate all the logic for changes that need to happen on entering and exiting Wasm into one place, which resulted in the birth of `EntryStoreContext`.

TL;DR Technically, all that's needed for stack switching could all be done in `catch_traps`, but that felt like it's morally the wrong spot to me.

Any suggestions? Should I just move it to `catch_traps`?

fitzgen:
> (Sorry for the long delay, I've moved to London and started a new job, where I needed to settle in first)

No worries. Congrats on the move and the new gig!

> The reason we do not use these in verbatim is that our stacks are not the same as `wasmtime_fiber::FiberStack`. Trying to make our stacks compatible with that type would be quite difficult and IMO not worth the effort.

I'd like to dive more into this because I would (perhaps naively?) assume that there is no fundamental reason they couldn't be the same. This is also one of the biggest concerns I have about the the stack-switching implementation: the complexity of having multiple kinds of stacks that the runtime must manage; when there are different kinds of resources, we will eventually need different config knobs for tuning them; and etc. Basically, this feels like a bunch of cognitive overhead for both maintainers and users. Keeping things uniform keeps things simpler. For example, https://github.com/bytecodealliance/wasmtime/issues/9350 (which is admittedly a beast of a PR; I have been splitting it into smaller pieces and landing those incrementally) will ultimately reduce complexity in Wasmtime by unifying our handling of Wasm linear memories and GC heaps. I just point this out so that you know I'm not trying to pick on you, and that I am attempting to hold myself to these standards as well.

But perhaps I am missing some key point that makes it so that these stack-switching stacks and fiber stacks really do have different requirements that do not overlap enough that sharing an implementation makes sense?

You say it would be quite difficult to unify them, can you expand on what difficulties your foresee?

> In particular, @dhil is very interested in restoring support for pooled allocation of stack. We had implemented our own interface for that (outside of `InstanceAllocatorImpl`) and it greatly improves performance in those use cases where you frequently (de-) allocate continuations.

Totally, that's exactly what I'd expect: a free list pop should be much faster than `mmap`, especially under load in a concurrent, multi-threaded embedding of Wasmtime.

> We actually extend `StaticTable` and `TableElement` in that file with appropriate variants for our types. The problem is that in [TablePool::new](https://github.com/bytecodealliance/wasmtime/blob/9da52ede33a9f8996832b0210579ef15296addbd/crates/wasmtime/src/runtime/vm/instance/allocator/pooling/table_pool.rs#L32) we need to `mmap` enough memory to hold `limits.total_tables` * `limits.table_elements` entries, independent from what data is stored in those tables. In order to support tables storing continuations pointers, this effectively doubles the required size of the mmapped area.

Ah okay, I get it now. You're right that we definitely don't want to cut the capacity of all existing tables in half in order to support a proposal that 99% of users won't be leveraging yet.

Two things:

1. And this is a pre-existing issue, that calculation of table size should be a little more robust, and use something like `mem::size_of::<TableElementUnion>()` so that it is more future-proof for changes like we have here where we are adding new kinds of table elements.
2. We could plumb through a boolean for whether the stack-switching proposal is enabled to this method, or the whole `WasmFeatures` or something, and then adjust the size of a single element in the table size calculation based on that value.

In the meantime, halving the capacity just for continuation tables (rather than all other existing tables) is fine. And we can do the above suggestions in follow up PRs.

crates/environ/src/builtin.rs
fitzgen: > The reason why we don't want to land this is explained [here](https://github.com/bytecodealliance/wasmtime/pull/10177/files#diff-0c7bbb4f13d0c3d3403d05e6d5f32beee56eb14e2f64f37f5b2a8b2ee6e78fc2R61): On the codegen side, we just emit the addresses of Rust string literals as literals into generated code, with no relocation information. So this crashes and burns if you ever try to re-use the same generated code from a different invocation of `wamtime`.

Ah okay! Yes, definitely should not land these bits in that case.

crates/wasmtime/src/runtime/vm/stack_switching.rs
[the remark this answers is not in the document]
fitzgen: Ha, lets keep it as-is for now then. I don't want to make you bounce back and forth.

frank-emrich:
@fitzgen 

> I'd like to dive more into this because I would (perhaps naively?) assume that there is no fundamental reason they couldn't be the same. This is also one of the biggest concerns I have about the the stack-switching implementation: the complexity of having multiple kinds of stacks that the runtime must manage; when there are different kinds of resources, we will eventually need different config knobs for tuning them; and etc. Basically, this feels like a bunch of cognitive overhead for both maintainers and users. Keeping things uniform keeps things simpler. For example, #9350 (which is admittedly a beast of a PR; I have been splitting it into smaller pieces and landing those incrementally) will ultimately reduce complexity in Wasmtime by unifying our handling of Wasm linear memories and GC heaps. I just point this out so that you know I'm not trying to pick on you, and that I am attempting to hold myself to these standards as well.
> 
> But perhaps I am missing some key point that makes it so that these stack-switching stacks and fiber stacks really do have different requirements that do not overlap enough that sharing an implementation makes sense?
> 
> You say it would be quite difficult to unify them, can you expand on what difficulties your foresee?

I totally get the wish to make things more maintainable by avoiding duplicating as much stack management logic as possible.

**The short answer is:** If we had a `FiberStack` type that means "just a big chunk of memory used for an independent stack, allocated with the virtual memory mechanism of the current platform" we could share them between our stack switching implementation and `wasmtime_fiber`. But in its current form, the `wasmtime_fiber::FiberStack` is just naturally tailored to be used together with `wasmtime::Fiber`, and bakes in details about how stack switching is done there.

**Longer answer:** Currently, `wasmtime_fiber::FiberStack`  is just a wrapper around a platform specific implementation. If we look at `unix::FiberStack`, the documentation there shows how a `unix::FiberStack` is used by a `unix::Fiber`. This influences the layout of the data at the top of the stack (see the diagram in crates/fiber/src/unix.rs). In contrast, our layout of the header near the top of a stack is different (see crates/wasmtime/src/runtime/vm/stack_switching/stack/unix.rs).

For example, `wasmtime_fiber::FiberStack` expects that the `wasmtime_fiber_switch` function stores the IP and RP on the actual stack where we left off on the stack we depart. In contrast, when you, me and @alexcrichton discussed the  `stack_switch` CLIF instruction we decided that it would be nicer to store only the data GPRs on the stack itself, and write the remaining data (like IP, RBP) elsewhere. We now write this to the header of the stack near its top.

It may be possible to move both implementations (i.e., `wasmtime_fiber` and our stack switching) closer to each other, so that they used the exact same layout of `FiberStack`, but that may require some compromising, ultimately making things more brittle. For example, both implementations would need to agree on a type for passing status information and payloads (see `wasmtime_fiber::RunResult`). Do let me know if you would like me to elaborate on the details of what would be required further. 

If we leave the Unix world, things become even more complicated: The Windows implementation of `wasmtime_fiber::FiberStack` doesn't manage the stack memory directly (making `wasmtime_fiber::FiberStack` mostly just a dummy), and `wasmtime_fiber::Fiber` uses the Windows Fiber API to create and run stacks. My plan for the Windows implementation of the stack switching proposal was not to use that API, but do the stack switching in generated code. Thus, there is quite a difference between what `wasmtime_fiber::FiberStack` currently does on Windows, and what we'd need.

frank-emrich:
@fitzgen Regarding tables with continuations

> Ah okay, I get it now. You're right that we definitely don't want to cut the capacity of all existing tables in half in order to support a proposal that 99% of users won't be leveraging yet.
>
> [...] We could plumb through a boolean for whether the stack-switching proposal is enabled to this method, or the whole WasmFeatures or something, and then adjust the size of a single element in the table size calculation based on that value.
> 
> In the meantime, halving the capacity just for continuation tables (rather than all other existing tables) is fine. And we can do the above suggestions in follow up PRs.

Just halving the max allowed capacity for the time being and then plumbing through whether the stack switching feature is enabled in the longer terms sounds good to me.

crates/wasmtime/src/runtime/func.rs
frank-emrich: I've run these commands, the results are here:
https://gist.github.com/frank-emrich/f7f80adc8005900a3e245d2400149b38

(But no comparing the  `stack-switching-infra` branch against actual `main`, but the merge base with `main`, since `main` has moved forward quite a bit since the last merge)

I'm not quite sure how to interpret the output, would you mind having a look?

I will re-run this once I follow through with your [suggestion](https://github.com/bytecodealliance/wasmtime/pull/10388#discussion_r2031832400) of splitting out most of the `EntryStoreContext` into a separate PR.

crates/wasmtime/src/runtime/func.rs
frank-emrich: No problem, making the changes to this critical bit of code more reviewable is perfectly sensible. I've created a PR with the `EntryStoreContext` stuff minus stack-switching fields here: #10626
Once that is landed it will be easy to see what additional stuff this PR needs to add to `EntryStoreContext` (not much, really)

crates/wasmtime/src/runtime/func.rs
fitzgen: Are these benchmark results for this whole branch, or for just the `EntryStoreContext` changes in https://github.com/bytecodealliance/wasmtime/pull/10626?

Unfortunately, it looks like there are a bunch of regressions, and enough of them that it doesn't seem likely to be a statistical fluke. But that doesn't mean that this is all a non-starter. What this likely means is that there are some functions that need to be marked `#[inline]` and/or split into fast and slow paths, roughly like

```rust
#[inline]
fn foo(&self, ...) -> Option<Bar> {
    if !self.is_foo_enabled() {
        return None;
    }

    self.foo_slow_path()
}

#[inline(never)]
fn foo_slow_path(&self) -> Option<Bar> {
    // ...
}
```

A good place to start addressing this would be in https://github.com/bytecodealliance/wasmtime/pull/10626, since the set of changes is smaller and easier to get a handle on the total effect of changes. I can help out and do a little profiling later today.

The workflow is basically run a regressing benchmark [in profiling mode](https://bheisler.github.io/criterion.rs/book/user_guide/profiling.html) and attach `perf`/`samply`/instruments/vtune/whatever your profiler of choice is, look at the top self-time functions, and then check whether it is necessary to be called (i.e. can't be conditional on some feature being used), whether it is being inlined, whether it can be split into fast and slow paths, etc...

Again, I can do a bit of this later today, so don't stress out too much. I don't want to hold these PRs up.

fitzgen:
> Just halving the max allowed capacity for the time being and then plumbing through whether the stack switching feature is enabled in the longer terms sounds good to me.

Just making sure we are on the same page: specifically only continuation-ref tables should have their capacity halved. `funcref` and `VMGcRef` tables should be unaffected.

fitzgen:
> It may be possible to move both implementations (i.e., `wasmtime_fiber` and our stack switching) closer to each other, so that they used the exact same layout of `FiberStack`, but that may require some compromising, ultimately making things more brittle. For example, both implementations would need to agree on a type for passing status information and payloads (see `wasmtime_fiber::RunResult`). Do let me know if you would like me to elaborate on the details of what would be required further.

I guess that what I am imagining is that, despite whatever headers or whatnot that they layer onto the stack, they share the same underying mechanism (and `Config` knobs and all that) to allocate/deallocate them and manage them in the runtime (i.e. they come from the same pool when using the pooling allocator). This would *not* require that we unify their stack switching code paths (since one is in a mix of Rust code and inline assembly and the other is implemented in JIT code) but it would mean that we would have some kind of `VMRawStack` type or some such that manages the underlying memory allocation/`mmap`/fiber, is allocated/deallocated via the `InstanceAllocator` trait, the pooling allocator (for example) would pool these objects, and then both `wasmtime_fiber::FiberStack` and `ContinuationStack` would be thin-ish wrappers around that `VMRawStack` allocation.

That said, this refactoring doesn't need to happen now, before this PR merges. I think it can be left for follow ups. However, we wouldn't do things like enable the stack-switching feature by default in Wasmtime, for example, until after this was resolved.

> If we leave the Unix world, things become even more complicated: The Windows implementation of `wasmtime_fiber::FiberStack` doesn't manage the stack memory directly (making `wasmtime_fiber::FiberStack` mostly just a dummy), and `wasmtime_fiber::Fiber` uses the Windows Fiber API to create and run stacks. My plan for the Windows implementation of the stack switching proposal was not to use that API, but do the stack switching in generated code. Thus, there is quite a difference between what `wasmtime_fiber::FiberStack` currently does on Windows, and what we'd need.

My understanding of the constraints for Windows is that, even if we emit inline code for bookkeeping the metadata required in Windows, the allocation and deallocation of fiber stacks must happen through `CreateFiber[Ex]` (or `ConvertThreadToFiber`) and `DeleteFiber`. In this case, having a shared `VMRawStack` that understands and abstracts this stuff away should hopefully actually make things easier for us here.

-------------------------

Anyways, let me know when all the `TODO`s and previous comments have been addressed, and this is ready for another round of review. Thanks!

crates/wasmtime/src/runtime/vm/traphandlers/backtrace.rs
```diff
@@ -176,6 +236,100 @@ impl Backtrace {
...
+                // continuation. Otherwise, if `continuation` is the last
+                // continuation (i.e., its parent is the initial stack), this is
+                // None.
+                let parent_continuation = if i + 1 < continuations_vec.len() {
+                    Some(&*continuations_vec[i + 1])
+                } else {
+                    None
+                };
+                (continuation, parent_continuation, parent_limits)
+            };
```
fitzgen, line 307: Can this be rewritten to use [slice::chunks](https://doc.rust-lang.org/nightly/std/primitive.slice.html#method.chunks) and `zip`? And also narrow the `unsafe` to just the relevant blocks (because I had a hard time figuring out why it was needed at a glance). I think this would make the code a bit safer and easier to read. Something like this (although preserving comments where appropriate):

```suggestion
        for (conts, parent_limits) in continuations_vec.chunks(2).zip(stack_limits_vec.iter().skip(1)) {
            let continuation = conts[0];
            let continuation = unsafe { &*continuation };
            
            let parent_limits = unsafe { &*parent_limits };
            
            let parent_continuation = conts.get(1).map(|p| unsafe { &*p });
            
            // ...
```

fitzgen (commented):
Looking good! I have a ton of comments below, they are all nitpicks really, so this should be good to land as soon as they are fixed.

Thanks everyone!

crates/wasmtime/src/runtime/vm/stack_switching/stack/unix.rs
```diff
@@ -0,0 +1,356 @@
...
+}
+
+unsafe extern "C" {
+    #[allow(dead_code)] // only used in inline assembly for some platforms
+    fn wasmtime_continuation_start();
+}
+
+/// This function is responsible for actually running a wasm function inside a
+/// continuation. It is only ever called from `wasmtime_continuation_start`. Hence, it
+/// must never return.
```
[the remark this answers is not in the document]
frank-emrich: Ha, well spotted! I must have forgotten updating the comment here, `fiber_start` is allowed to return now because `wasmtime_continuation_start` contains the logic for going back to the parent these days.

crates/wasmtime/src/runtime/vm/stack_switching.rs
```diff
@@ -0,0 +1,700 @@
...
+/// - A continuation (i.e., created with cont.new).
+/// - The initial stack. This is the stack that we were on when entering
+///   Wasm (i.e., when executing
+///   `crate::runtime::func::invoke_wasm_and_catch_traps`).
+///   This stack never has a parent.
+///   In terms of the memory allocation that this stack resides on, it will
+///   usually be the main stack, but doesn't have to: If we are running
+///   inside a continuation while executing a host call, which in turn
+///   re-renters Wasm, the initial stack is actually the stack of that
+///   continuation.
```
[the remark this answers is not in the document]
frank-emrich: Ah, sorry, I must have overlooked your last sentence in the linked thread:

> [...] and I would also expect that we would call libcalls, but never host functions, on stack-switching stacks, same way we do things on our async fibers 

We do indeed allow calling host functions from our continuation stacks at the moment, and even re-entering Wasm from such a host function leading to calls chains chains of the following shape:

```
   Wasm function  (on main stack) 
-> Wasm function  (on continuation stack 1) 
-> host function  (on continuation stack 1) 
-> Wasm function  (on continuation stack 1)
-> Wasm function  (on continuation stack 2)
-> host function  (on continuation stack 2)
... 
```

I think @dhil had a use-case where allowing such re-entrance is useful, but I see how running host functions on continuation stacks has its problems.

It's easy to check if we are currently running on a continuation stack, but that would have to happen at every call to a host function, right? But I guess there must already be such a check for whether we are on a Fiber stack (for the async feature) that we can adapt?

crates/wasmtime/src/runtime/vm/libcalls.rs
```diff
@@ -261,6 +264,41 @@ unsafe fn table_grow_gc_ref(
...
+    // A None value is indicated by the pointer being null.
+    init_value_contref: *mut u8,
+    init_value_revision: u64,
+) -> Result<Option<AllocationSize>> {
+    use core::ptr::NonNull;
+    let init_value = if init_value_contref.is_null() {
+        None
+    } else {
+        // SAFETY: We just checked that the pointer is non-null
+        let contref = NonNull::new_unchecked(init_value_contref as *mut VMContRef);
```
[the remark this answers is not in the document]
posborne: Patch will land in a new `VMContObj::from_raw_parts` impl.

crates/wasmtime/src/runtime/vm/stack_switching.rs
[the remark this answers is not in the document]
fitzgen: Those sequence of calls should definitely be supported, but ideally we would automatically switch to the native host stack and then back again while executing the actual host functions. That said, I think this PR is fine for now, and this is something we can address in the future.

> It's easy to check if we are currently running on a continuation stack, but that would have to happen at every call to a host function, right? But I guess there must already be such a check for whether we are on a Fiber stack (for the async feature) that we can adapt?

We don't dynamically check, we have different paths for constructing async and sync `wasmtime::Func`s (and we have various assertions that the constructor you use matches the engine's config):

* The sync version invokes the host function directly after marshalling arguments
* The async version [uses `block_on`](https://github.com/bytecodealliance/wasmtime/blob/90a2351b4f9aa70bfdda81ad4f4713e1b97f6936/crates/wasmtime/src/runtime/func.rs#L566) to [automatically suspend](https://github.com/bytecodealliance/wasmtime/blob/90a2351b4f9aa70bfdda81ad4f4713e1b97f6936/crates/wasmtime/src/runtime/store/async_.rs#L798) the current fiber and switch back to the native host stack (via coordination with the original [`on_fiber`](https://github.com/bytecodealliance/wasmtime/blob/90a2351b4f9aa70bfdda81ad4f4713e1b97f6936/crates/wasmtime/src/runtime/store/async_.rs#L221) caller; we have similar sync vs async paths for calling into Wasm, each of which also assert that the version used matches the engine's config, and the async path takes care to [alway use `on_fiber`](https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasmtime/src/runtime/func.rs#L1170))

crates/cli-flags/src/lib.rs
```diff
@@ -923,6 +942,8 @@ impl CommonOptions {
             );
         }
 
+        #[cfg(any(feature = "async", feature = "stack-switching"))]
+
```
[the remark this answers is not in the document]
posborne: Looking at this again, this whole `match_feature!` should be dropped at is now covered by an earlier chunk of code (basically doing the macro by hand to support both features).  Good catch.

crates/wasmtime/src/runtime/vm/stack_switching/stack/unix.rs
```diff
@@ -0,0 +1,356 @@
...
+            };
+
+            let args = &mut *args;
+            let args_capacity = std::cmp::max(parameter_count, return_value_count);
+            // The args object must currently be empty.
+            debug_assert_eq!(args.capacity, 0);
+            debug_assert_eq!(args.length, 0);
+
+            // The size of the args buffer
+            let s = args_capacity as usize * std::mem::size_of::<ValRaw>();
```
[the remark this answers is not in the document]
posborne: @fitzgen  In addressing the latest review items, I'm trying to work out what I should target in terms of conversions of types into usize.  I see a few cases with what I believe is the requested conversion.

- `T -> usize` where `T: TryInto<usize>`: For any of these types (e.g. u32, u8, ...) we want to do a `usize::try_from(value).unwrap()` rather than a cast.  I'm assuming that the these will compile down to a cast when optimized for types (like u32 on any architecture with a 32-bit+ word size) but we prefer a panic if somehow the code were compiled for a target with a 16-bit pointer width (and a value that doesn't fit).
- Pointers: A cast using `as usize` is OK for these.

I'll admit the `TryFrom`/`unwrap` for types like u32 feels gross to me with it being preferred to fail compilation for types that can't be converted always be cast correctly.

As an aside, I did to a quick check in godbolt to confirm that the `TryFrom` for `u32 -> usize` does optimize out, so shouldn't be a perf impact, so I suppose this is mostly a readabilty/cosmetic concern.

---

In this particular case `tos` is a `*mut u8` so the `TryFrom` approach isn't available and I'm assuming it should remain an `as usize` cast.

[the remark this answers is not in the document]
fitzgen: Here are the general rules to follow:

First, if there are infallible `From`/`Into` impls available, use those.

For conversions with `usize`, Rust doesn't add/remove infallible conversions to `u32`/`u64` depending on the target, so that source code is more portable by default. In these cases, we should use `TryFrom`/`TryInto` and `unwrap()`, and the compiler will boil it all away, as you mentioned.

For other technically-fallible conversions that shouldn't actually fail due to program invariants, like `u64` to `u32` where we know we never exceed `u32::MAX`, we should also use `TryFrom`/`TryInto` and `unwrap()` so that things fail loudly if our invariants aren't upheld.

For conversions to `usize` from pointers, we use `as` today, but eventually should be using `ptr.addr()` and strict provenance APIs once we can rely on Rust 1.84 and greater.

-----------------------

If I messed up any specific review comment, feel free to just leave a comment on that explaining my whoopsie, as long as we are abiding by the above general rules.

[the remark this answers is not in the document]
posborne: Thanks for the guidance, makes sense to me; I talked a bit with @sunfishcode in our 1:1 about strict provenance this afternoon as well, so thanks for covering that.  I'll update based on those recommendations.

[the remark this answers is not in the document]
posborne: It does appear that the wasmtime MSRV on main is now 1.85.0 which should mean that we could rely on the stabilized strict provenance APIs (or use `sptr::Strict`).

crates/wasmtime/src/runtime/vm/traphandlers/backtrace.rs
posborne: Adopted this general approach, worked out well.

crates/wasmtime/src/runtime/vm/vmcontext.rs
```diff
@@ -1078,6 +1079,10 @@ pub struct VMStoreContext {
     /// Used to find the end of a contiguous sequence of Wasm frames when
     /// walking the stack.
     pub last_wasm_entry_fp: UnsafeCell<usize>,
+
+    /// Stack information used by stack switching instructions. See documentation
+    /// on `VMStackChain` for details.
+    pub stack_chain: UnsafeCell<VMStackChain>,
```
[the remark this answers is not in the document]
posborne: The tests in the environ crate for stack switching seemed to cover the other cases already; I did land a commit reorganizing the location of the tests there to be in their own module to aid discovery and be a bit more in line with the prevailing pattern for test inclusion.

Let me know if those don't cover what you had in mind for assertions for the environ bits.

posborne:
@fitzgen I think there's now changes to address the latest round of feedback; with how gh permissions work I can't directly modify the description for the tracking issue but I have notes and will add a comment with the new items to add that a maintainer or @frank-emrich  should be able to pull into the description body.

crates/wasmtime/src/runtime/store.rs
```diff
@@ -1623,67 +1633,115 @@ impl StoreOpaque {
...
+            //   chain starting at either the running one, or the suspended
+            //   continuations below.
+            // - For `Fresh` continuations, we know that there are no GC values
+            //   on their stack, yet.
+            if state == VMStackState::Suspended {
+                Backtrace::trace_suspended_continuation(self, continuation.deref(), |frame| {
+                    self.trace_wasm_stack_frame(gc_roots_list, frame);
+                    core::ops::ControlFlow::Continue(())
+                });
+            }
```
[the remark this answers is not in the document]
posborne: Missed this in my first pass; should land a patch shortly with this change.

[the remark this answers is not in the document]
posborne: Updated now; there wasn't a previous document comment on `VMStackState::Returned` which I grouped with Fresh but didn't dig into that case in detail not being particularly familiar with the gc impl at this point.

crates/environ/src/trap_encoding.rs
```diff
@@ -93,6 +93,16 @@ pub enum Trap {
...
 
+    /// We are suspending to a tag for which there is no active handler.
+    UnhandledTag,
+
+    /// Attempt to resume a continuation twice.
+    ContinuationAlreadyConsumed,
+
+    /// FIXME(frank-emrich) Only used for stack switching debugging code, to be
+    /// removed from final upstreamed code.
+    DeleteMeDebugAssertion,
```
alexcrichton, line 104: Should this be left out?

alexcrichton (commented):
This all looks good to me as well, thanks again for pushing on this to everyone involved!

One minor thing I noticed reading over this and I also filed a few points in https://github.com/bytecodealliance/wasmtime/issues/10248 under miscellaneous TODOs at the end

alexcrichton:
(I'm going to go ahead and flag this for merge though to carry through Nick's approval, my small thing can be fixed after)

crates/environ/src/trap_encoding.rs
posborne: Yeah, missed these when I removed most of the other bespoke debug assertion bits; will add to #10248 TODOs.

Issue #10361, opened 2025-03-10
Cranelift: new assembler panics in unreachable code when dumping trace-log of compilation


cfallin:
With the following test-case

```
test compile
target x86_64

function %f0(i32, i32) -> i32 {
block0(v0: i32, v1: i32):
  v2 = band.i32 v0, v1
  return v2
}
```

when running the following command

```
$ RUST_BACKTRACE=1 RUST_LOG=trace cargo run -p cranelift-tools -- test ./test.clif
```

I get a panic like so:

```
 TRACE cranelift_codegen::machinst::lower      > lowering: inst inst0: v2 = band.i32 v0, v1
[ ... ]

thread 'worker #2' panicked at cranelift/codegen/src/isa/x64/inst/external.rs:95:9:
internal error: entered unreachable code
stack backtrace:
[ ... ]
   3: cranelift_codegen::isa::x64::inst::external::enc_gpr
             at ./cranelift/codegen/src/isa/x64/inst/external.rs:95:9
   4: <cranelift_codegen::isa::x64::inst::external::PairedGpr as cranelift_assembler_x64::api::AsReg>::enc
             at ./cranelift/codegen/src/isa/x64/inst/external.rs:35:20
   5: cranelift_assembler_x64::reg::Gpr<R>::enc
             at ./cranelift/assembler-x64/src/reg.rs:27:19
   6: cranelift_assembler_x64::reg::Gpr<R>::to_string
             at ./cranelift/assembler-x64/src/reg.rs:34:24
   7: <cranelift_assembler_x64::inst::andl_rm<R> as core::fmt::Display>::fmt
             at ./target/debug/build/cranelift-assembler-x64-0a4a19fac84309b6/out/assembler.rs:4336:19
[ ... ]
  22: cranelift_codegen::isa::x64::inst::<impl cranelift_codegen::machinst::reg::PrettyPrint for cranelift_codegen::isa::x64::lower::isle::generated_code::MInst>::pretty_print
             at ./cranelift/codegen/src/isa/x64/inst/mod.rs:1983:17
  23: cranelift_codegen::isa::x64::inst::<impl cranelift_codegen::machinst::MachInstEmit for cranelift_codegen::isa::x64::lower::isle::generated_code::MInst>::pretty_print_inst
             at ./cranelift/codegen/src/isa/x64/inst/mod.rs:2967:9
[...]

PR #10389, opened 2025-03-13
asm: allow printing of virtual registers

abrown:
#10361 identifies a situation in which we want to print assembly instructions that have not yet been register allocated. Because of this, the register slots hold virtual registers, which the assembler does not know how to pretty-print. This change overrides `AsReg::to_string` in a few places so that `cranelift-codegen`'s view of registers can be pretty-printed using the existing code for doing so.


abrown:
@cfallin, I was going to add some tests to keep this working correctly: can you remind me how to create a virtual `Reg` out of thin air?

cfallin:
Thanks for addressing this!

> @cfallin, I was going to add some tests to keep this working correctly: can you remind me how to create a virtual Reg out of thin air?

Sounds good; something like `VReg::new(200, RegClass::Int)` should work I think?

cfallin (approved):
This approach is definitely workable -- I'm fine with landing this. However I note that `AsReg` in the assembler crate carries a `Debug` constraint already, meaning that we're requiring the user of the assembler library to provide a `Debug`-printable type -- is there any way we could make `to_string` handle this case at the assembler crate level by delegating to that (e.g. `format!("{:?}", self)`) and differeniate between the real and virtual cases by making `enc()` return an `Option<u8>`? I also feel this is a little more "honest" in the sense that not every `AsReg` value will be a physical register at all times. `.enc().expect("need physreg for encoding")` in actual emission code seems fine in that case.

If that turns out to be too complex, no worries, this existing solution is fine too.

cfallin:
I poked at this a little myself and I think I'm actually fine with the way this works currently. Modifying `AsReg::enc()` to return `Option<u8>` is in some sense closer to reality -- the register may or may not have a physical encoding -- but almost all users use it only when it does, and only pretty-printing differs, so it's probably just as good to simply document that it can panic when the reg is virtual, and make pretty-printing work explicitly instead, as you've done here. I also played a bit with the plumbing on pretty-printing -- one could invert so that `to_string()` in the assembler crate calls back to the embedder (Cranelift) for vregs but again it's about the same amount of code, and this works fine. So I'll say LGTM, and we can always refactor later if needed.

Thanks for adding the test as well!

PR #10393, opened 2025-03-13
Winch: Fix issue with floats and replace_lane

jeffcharles:
Fixes an issue with `f64x2.replace_lane` where it always zeroes out the second lane if the operand containing the replacement value is a `f64.const`.


saulecabrera (approved):
Good catch, thanks!

jeffcharles:
Right, I forgot to add the new tests to the exclusion list. I'll fix that

PR #10394, opened 2025-03-13
Add graceful shutdown to `wasmtime serve`, fix flaky CI

alexcrichton:
This commit fixes the spurious CI failure found [here][fail]. The problem here is that the infrastructure for testing `wasmtime serve` was not properly waiting for all output in `finish`. There's some infrastructure in the tests for spawning a subprocess and managing it, but prior to this commit there was no way to shut down the server gracefully and thus read all pending output from the child tasks.

The specific problem is that a specific error message was expected in the logs after a request had been processed, but the `finish` method wasn't reading the message. The reason for this is that `finish` had to resort to `kill -9` on the child process as there was no other means of shutting it down. This meant that the print, which happened after request completion, might be killed and never happen.

The solution ended up in this commit is to (a) add a `--shutdown-addr` CLI flag to `wasmtime serve` and (b) beef up handling of graceful shutdown. Previously ctrl-c was used to exit the server but it didn't do anything but drop all in-progress work. Instead graceful shutdown is now handled by breaking out of the `accept` loop and then waiting for all child tasks to complete, meaning that no http requests once received are cancelled. In addition to ctrl-c the `--shutdown-addr` is used to listen for a TCP connection which is a second means of cancellation. The reason for this is that sending ctrl-c to a process is not nearly as trivial on Windows as it is on Unix, so I didn't want to deal with all the console bits necessary to get that aligned.

[fail]: https://github.com/bytecodealliance/wasmtime/actions/runs/13833291117/job/38702374924?pr=10390

### DIFF
--- a/src/commands/serve.rs
+++ b/src/commands/serve.rs
@@ -6,9 +6,10 @@ use std::{
     path::PathBuf,
     sync::{
         atomic::{AtomicBool, AtomicU64, Ordering},
-        Arc,
+        Arc, Mutex,
     },
 };
+use tokio::sync::Notify;
 use wasmtime::component::Linker;
 use wasmtime::{Engine, Store, StoreLimits};
 use wasmtime_wasi::{IoView, StreamError, StreamResult, WasiCtx, WasiCtxBuilder, WasiView};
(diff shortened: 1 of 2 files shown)


src/commands/serve.rs
```diff
@@ -533,7 +629,7 @@ async fn handle_request(
             // which should more clearly tell the user what went wrong. Note
             // that we assume the task has already exited at this point so the
             // `await` should resolve immediately.
-            let e = match task.await {
+            let e = match dbg!(task.await) {
```
dicej, line 632: Looks like this was included by mistake?

tests/all/cli_tests.rs
```diff
@@ -1548,61 +1549,73 @@ mod test_programs {
...
-            // and then reap it. Assume that if it's still running then the test
-            // has otherwise passed so no need to print the output.
-            let known_failure = if child.try_wait()?.is_some() {
-                false
-            } else {
-                child.kill()?;
-                true
-            };
+            // If the child process has already exited, then great! Otherwise
+            // the server is still running so send it a graceful ctrl-c shutdown
```
dicej, line 1610: It doesn't look like we send a ctrl-c here, regardless of platform, so maybe change this comment to read "...so connect to the shutdown port..." or something similar.

dicej (approved):

tests/all/cli_tests.rs
alexcrichton: Oops sorry I was halfway through updating this comment when I context switched at some point and forgot to update the other half of the comment...

PR #10400, opened 2025-03-14
[DWARF] Fix the loclist to exprloc optimization

### LINKED ISSUE #9900: [DWARF] Incorrect location descriptor
Reproduction steps:
1) Check out #9898.
2) Modify the added test to stop passing `-Oopt-level=0` to wasmtime.
3) Run the test.

Expected result: the test passes or fails with 'expression unavailable'.

Actual result:
```
error: Execution was interrupted, reason: Exception 0xc0000005 encountered at address 0x7ff6fcf15f90: Access violation reading location 0x1ae198763ec.
The process has been returned to the state before expression evaluation.

* thread #1, name = 'main', stop reason = breakpoint 1.1
    frame #0: 0x0000028d720c15b8 JIT(0x28d72210080)`DerivedType::InstanceMethod(__this=(__ptr = <read memory from 0x28de65f5f0c failed (0 of 4 bytes read)>)) at generic.cpp:26:5
   23     long long DerivedValue = 2;
   24
   25     int InstanceMethod() {
-> 26       debug_break();
   27       return BaseValue + DerivedValue;
   28     }
   29
(lldb) v
(WasmtimeVMContext *) __vmctx = 0x0000028d72165f00
(WebAssemblyPtrWrapper<DerivedType>) __this = (__ptr = <read memory from 0x28de65f5f0c failed (0 of 4 bytes read)>)
```
It means the location descriptor for `__this` is incorrect.

SingleAccretion:
When using an exprloc to describe a location, we must ensure the underlying expression computes the correct value over the whole parent scope.

```
; DWARF5

Single location descriptions

They are sufficient for describing the location of any object
as long as its lifetime is either static or the same as the lexical
block that owns it, and it does not move during its lifetime.
```
Fixes #9900.

### DIFF
--- a/crates/cranelift/src/debug/transform/attr.rs
+++ b/crates/cranelift/src/debug/transform/attr.rs
@@ -241,6 +241,7 @@ pub(crate) fn clone_die_attributes<'a>(
                                 frame_info,
                                 isa,
                             )
+                            .expressions
                             .filter(|i| {
                                 // Ignore empty range
                                 if let Ok((_, 0, _)) = i {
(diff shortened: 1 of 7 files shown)


fitzgen (approved):
Looks great, thanks!

PR #10404, opened 2025-03-15
Store stack maps in an ELF section

alexcrichton:
This commit moves the storage of stack maps from being embedded within serde-encoded information to instead being stored in a separate ELF section in the final executable. The motivation for this is to make this more easily debuggable with a `wasmtime objdump` command in the future but this additionally should have the nice side effect of making non-stack-maps modules have smaller encoded information (no need to encode an empty list) and additionally make stack-maps-using-modules faster to decode (no serde decoding, it's already "decoded").

This implements a scheme similar to the address map section where there's a "builder" for the section and then a separate half to decode the section. The same basic encoding, a bit map, is used. This is likely going to make accessing stack maps slightly slower, but if that's an issue we can tweak the representation and align things and/or use `usize` or such.

### DIFF
--- a/cranelift/bitset/src/compound.rs
+++ b/cranelift/bitset/src/compound.rs
@@ -76,7 +76,7 @@ impl CompoundBitSet {
 }
 
 impl<T: ScalarBitSetStorage> CompoundBitSet<T> {
-    const BITS_PER_WORD: usize = mem::size_of::<T>() * 8;
+    const BITS_PER_SCALAR: usize = mem::size_of::<T>() * 8;
 
     /// Construct a new, empty bit set with space reserved to store any element
     /// `x` such that `x < capacity`.
@@ -146,7 +146,7 @@ impl<T: ScalarBitSetStorage> CompoundBitSet<T> {
     /// assert!(bitset.capacity() >= 999);
     ///```
     pub fn capacity(&self) -> usize {
-        self.elems.len() * Self::BITS_PER_WORD
+        self.elems.len() * Self::BITS_PER_SCALAR
     }
 
     /// Is this bitset empty?
@@ -174,8 +174,8 @@ impl<T: ScalarBitSetStorage> CompoundBitSet<T> {
     /// `ScalarBitSet<usize>` at `self.elems[word]`.
     #[inline]
     fn word_and_bit(i: usize) -> (usize, u8) {
-        let word = i / Self::BITS_PER_WORD;
-        let bit = i % Self::BITS_PER_WORD;
+        let word = i / Self::BITS_PER_SCALAR;
+        let bit = i % Self::BITS_PER_SCALAR;
         let bit = u8::try_from(bit).unwrap();
         (word, bit)
     }
@@ -185,8 +185,8 @@ impl<T: ScalarBitSetStorage> CompoundBitSet<T> {
     #[inline]
     fn elem(word: usize, bit: u8) -> usize {
         let bit = usize::from(bit);
-        debug_assert!(bit < Self::BITS_PER_WORD);
-        word * Self::BITS_PER_WORD + bit
+        debug_assert!(bit < Self::BITS_PER_SCALAR);
+        word * Self::BITS_PER_SCALAR + bit
     }
 
     /// Is `i` contained in this bitset?
@@ -509,7 +509,7 @@ impl<T: ScalarBitSetStorage> CompoundBitSet<T> {
     /// ```
     pub fn iter_words(&self) -> impl Iterator<Item = T> + '_ {
         let nwords = match self.max {
-            Some(n) => 1 + (n as usize / Self::BITS_PER_WORD),
+            Some(n) => 1 + (n as usize / Self::BITS_PER_SCALAR),
             None => 0,
         };
         self.elems.iter().map(|b| b.0).take(nwords)
--- a/cranelift/bitset/src/compound.rs
+++ b/cranelift/bitset/src/compound.rs
@@ -1,6 +1,6 @@
 //! Compound bit sets.
 
-use crate::scalar::{self, ScalarBitSet};
+use crate::scalar::{self, ScalarBitSet, ScalarBitSetStorage};
 use alloc::boxed::Box;
 use core::{cmp, iter, mem};
 
@@ -45,8 +45,8 @@ use core::{cmp, iter, mem};
     feature = "enable-serde",
     derive(serde_derive::Serialize, serde_derive::Deserialize)
 )]
-pub struct CompoundBitSet {
-    elems: Box<[ScalarBitSet<usize>]>,
+pub struct CompoundBitSet<T = usize> {
+    elems: Box<[ScalarBitSet<T>]>,
     max: Option<u32>,
 }
 
@@ -56,8 +56,6 @@ impl core::fmt::Debug for CompoundBitSet {
         f.debug_set().entries(self.iter()).finish()
     }
 }
-
-const BITS_PER_WORD: usize = mem::size_of::<usize>() * 8;
 
 impl CompoundBitSet {
     /// Construct a new, empty bit set.
@@ -75,6 +73,10 @@ impl CompoundBitSet {
     pub fn new() -> Self {
         CompoundBitSet::default()
     }
+}
+
+impl<T: ScalarBitSetStorage> CompoundBitSet<T> {
+    const BITS_PER_WORD: usize = mem::size_of::<T>() * 8;
 
     /// Construct a new, empty bit set with space reserved to store any element
     /// `x` such that `x < capacity`.
@@ -86,14 +88,14 @@ impl CompoundBitSet {
     /// ```
     /// use cranelift_bitset::CompoundBitSet;
     ///
-    /// let bitset = CompoundBitSet::with_capacity(4096);
+    /// let bitset = CompoundBitSet::<u32>::with_capacity(4096);
     ///
     /// assert!(bitset.is_empty());
     /// assert!(bitset.capacity() >= 4096);
     /// ```
     #[inline]
     pub fn with_capacity(capacity: usize) -> Self {
-        let mut bitset = Self::new();
+        let mut bitset = Self::default();
         bitset.ensure_capacity(capacity);
         bitset
     }
@@ -144,7 +146,7 @@ impl CompoundBitSet {
     /// assert!(bitset.capacity() >= 999);
     ///```
     pub fn capacity(&self) -> usize {
-        self.elems.len() * BITS_PER_WORD
+        self.elems.len() * Self::BITS_PER_WORD
     }
 
     /// Is this bitset empty?
@@ -172,8 +174,8 @@ impl CompoundBitSet {
     /// `ScalarBitSet<usize>` at `self.elems[word]`.
     #[inline]
     fn word_and_bit(i: usize) -> (usize, u8) {
-        let word = i / BITS_PER_WORD;
-        let bit = i % BITS_PER_WORD;
+        let word = i / Self::BITS_PER_WORD;
+        let bit = i % Self::BITS_PER_WORD;
         let bit = u8::try_from(bit).unwrap();
         (word, bit)
     }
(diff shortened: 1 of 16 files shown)


crates/environ/src/compile/stack_maps.rs
```diff
@@ -0,0 +1,175 @@
...
+/// Builder for the `ELF_WASMTIME_STACK_MAP` section in compiled executables.
+///
+/// This format is parsed by `crate::stack_map`.
+///
+/// The current layout of the format is:
+///
+/// * A 4-byte little-endian count of how many stack maps there are: `N`.
+/// * `N` 4-byte little endian program counters, in ascending order.
+/// * `N` 4-byte little endian offsets.
+/// * Stack map data as 4-bit little endian integers.
```
abrown, line 16:
```suggestion
/// * Stack map data as 4-byte little endian integers.
```
Without looking below to see what this should be... is this correct?

crates/environ/src/stack_map.rs
```diff
@@ -1,22 +1,92 @@
...
 ///
 /// Note that this is currently primarily documented as cranelift's
 /// `binemit::StackMap`, so for detailed documentation about this please read
 /// the docs over there.
-#[derive(Debug, Serialize, Deserialize)]
-pub struct StackMap {
-    bits: CompoundBitSet,
+pub struct StackMap<'a> {
     frame_size: u32,
+    data: &'a [U32Bytes<LittleEndian>],
```
abrown, line 67: Are we sure we want to re-export `U32Bytes` in the public API? I suspect the answer is "yes" but just wanted to note this.

abrown (approved):
Makes sense to me! I just noted some doubts I had while reading this for those who are more familiar with stack maps and their usage.

crates/environ/src/compile/stack_maps.rs
alexcrichton: Whoops, yes!

crates/environ/src/stack_map.rs
alexcrichton: Yeah it's fine here in that `wasmtime-environ` is a "private crate" where we don't consider it part of Wasmtime's public API, just an internal implementation detail. We can probably do more to document that at the crate itself though.

crates/environ/src/compile/stack_maps.rs
```diff
@@ -0,0 +1,175 @@
...
+    pub fn push(&mut self, pc: u64, frame_size: u32, offsets: impl ExactSizeIterator<Item = u32>) {
+        // NB: for now this only supports <=4GB text sections in object files.
+        // Alternative schemes will need to be created for >32-bit offsets to
+        // avoid making this section overly large.
+        let pc = u32::try_from(pc).unwrap();
+
+        // Sanity-check to ensure that functions are pushed in-order, otherwise
+        // the `pcs` array won't be sorted which is our goal.
+        assert!(pc >= self.last_offset);
+        self.last_offset = pc;
```
fitzgen, line 47: Something to consider for the future: if we frequently have multiple sequential entries for different PCs but which have the same stack slots, eg

```
...
0x1dc: offset of [8, 12]
0x124: offset of [8, 12, 24] copy 1
0x142: offset of [8, 12, 24] copy 2
0x15a: offset of [8, 12, 24] copy 3
0x166: offset of [8]
...
```

then it may make sense for each entry in the index to store non-overlapping PC ranges, rather than exact PCs, and we could effectively dedupe the index entries and the stack map data. That is, the previous example would become

```
...
0x1dc..0x1dd: offset of [8, 12]
0x124..0x15b: offset of [8, 12, 24] (only copy)
0x166..0x167: offset of [8]
...
```

The downsides are that

1. We would need to change Cranelift to actually emit empty stack maps for safepoints without any live GC refs, otherwise if we have `(pc=0x1234, [8]); (pc=0x1238, []); (pc=0x123b, [8])` and we don't see that middle entry in this builder, then we risk using `[8]` as our stack map at pc `0x1238`, which is extending a dead gc ref's lifetime at best and is giving the collector uninitialized data at worst.
2. Relatedly, we lose our ability to catch bugs where the return address PC we are tracing isn't an exact match for a stack map entry.

These are actually pretty scary, so maybe we don't want to do this, even if it would let us make these binary search indices much smaller.

------------------------

All that said, we can actually already dedupe the stack map _data_ if we want to, and have multiple index entries point to the same stack map data (even if they aren't contiguous!) with the encoding scheme already in use in this PR. We just need to hash cons and cache stack-map-data to encoded offset in this builder. This doesn't have any of the downsides from above. Seems like it would be a pure win.

crates/environ/src/compile/stack_maps.rs
```diff
@@ -0,0 +1,175 @@
...
+        for offset in offsets {
+            assert!(offset % 4 == 0);
+            bits.insert((offset / 4) as usize);
+        }
+        let count = bits.iter_words().count();
+        self.stack_map_data
+            .push(U32Bytes::new(LittleEndian, count as u32));
+        for word in bits.iter_words() {
+            self.stack_map_data.push(U32Bytes::new(LittleEndian, word));
+        }
```
fitzgen, line 75: So back to `iter_words`: the usage here makes me think we should instead just expose a method like

```rust
impl CompoundBitSet {
    pub fn as_raw(&self) -> &[u32] { ... }
}
```

method or something, have it always operate on `u32`, and maybe (I haven't looked at the updated usage yet) add something like the following to allow zero-copy usage: 

```rust
impl CompoundBitSet {
    pub fn ref_from_raw<F, T>(raw: &[u32], f: F) -> T
    where
        F: for<'a> FnOnce(&'a CompoundBitSet) -> T
    {
        // unsafely create a CompoundBitSet wrapping the raw data,
        // pass a ref to the bitset into the closure,
        // mem::forget the bitset to avoid Vec dtors
    }
}
```

rather than making it generic over the internal storage and doing a whole `iter_words` thing.

crates/environ/src/stack_map.rs
```diff
@@ -57,4 +128,26 @@ impl StackMap {
...
+                } else {
+                    let pos = remaining.trailing_zeros();
+                    let bit = 1 << pos;
+                    remaining ^= bit;
+                    Some(i * 32 + pos)
+                }
+            })
+        });
+
+        bit_positions.map(|pos| pos * 4)
```
fitzgen, line 151: The `CompoundBitSet::ref_from_raw` thing would help tidy up some of this stuff too, so that it doesn't have to have a copy of the bitset logic inline here.

fitzgen (changes requested):
Overall, I really like this approach. Have opinions on some of the details tho.

crates/environ/src/compile/stack_maps.rs
fitzgen: Could avoid the branded lifetime (so that the thing could be stored in the usage site's iterator) by doing a `std::cell::Ref`-esque thing instead:

```
pub struct BorrowedCompoundBitSet<'a> {
    bitset: core::mem::ManuallyDrop<CompoundBitSet>,
    _borrow: core::marker::PhantomData<&'a [u32]>,
}

impl core::ops::Deref for BorrowedCompoundBitSet<'_> {
    type Target = CompoundBitSet;
    // ...
}
```

cranelift/bitset/src/compound.rs
```diff
@@ -45,8 +45,8 @@ use core::{cmp, iter, mem};
     feature = "enable-serde",
     derive(serde_derive::Serialize, serde_derive::Deserialize)
 )]
-pub struct CompoundBitSet {
-    elems: Box<[ScalarBitSet<usize>]>,
+pub struct CompoundBitSet<T = usize> {
+    elems: Box<[ScalarBitSet<T>]>,
     max: Option<u32>,
 }
```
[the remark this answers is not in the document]
alexcrichton: This is an attempt to multiplex the (IMO correct) default behavior of pointer-sized-by-default with the cross-compilation behavior of "always use fixed-size things as it's easier". I don't want to use `usize` in the compiled ELF as that makes cross-compilation and debugging cross-compiled artifacts a bit harder.

cranelift/bitset/src/compound.rs
```diff
@@ -461,19 +463,63 @@ impl CompoundBitSet {
...
+    ///     [0x3, 0x1],
+    /// );
+    /// ```
+    pub fn iter_words(&self) -> impl Iterator<Item = T> + '_ {
+        let nwords = match self.max {
+            Some(n) => 1 + (n as usize / Self::BITS_PER_WORD),
+            None => 0,
+        };
+        self.elems.iter().map(|b| b.0).take(nwords)
+    }
```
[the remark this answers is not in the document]
alexcrichton: I'm not sure I understand your confusion, but I'm happy to rename this and/or document it better. IMO `CompoundBitSet` is by definition a `Vec<ScalarBitSet<T>>` and the point of it is to do all the vector indexing for you so you don't have to think about it. In that sense I wouldn't want to work with `Vec<ScalarBitSet<T>>` raw and it also makes sense to me to expose an iterator or slice accessor as that's the definition of the representation.

crates/cranelift/src/compiler.rs
```diff
@@ -1054,45 +1059,37 @@ impl FunctionCompiler<'_> {
...
-        let stack_map = wasmtime_environ::StackMap::new(mapped_bytes, bitset);
-        stack_maps.push(StackMapInformation {
-            code_offset,
-            stack_map,
-        });
+        section.push(
+            range.start + u64::from(*offset),
+            *frame_size,
+            offsets.into_iter(),
+        );
```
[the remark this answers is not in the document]
alexcrichton: For (1) agreed and that happens within `StackMapSection` building. For (2) I'll add some assertions, good point!

crates/environ/src/compile/stack_maps.rs
alexcrichton: I like your latter idea, having a "pool of sets" and when we insert a stack map we shove it in a hash map and dedupe based on that. Agreed it's a pure win with no real degree of complexity overhead. Are you thinking that should be done here? (I'd sort of prefer to defer that to analyze a "real world module" with a lot of stack maps)

crates/environ/src/compile/stack_maps.rs
alexcrichton: I realize I can probably use `ScalarBitSet::iter` to simplify a lot of this. I'd prefer to avoid too wonky of an abstraction to use `CompoundBitSet` literally somehow, but do you think that's an acceptable complexity tradeoff?

alexcrichton: Er, sorry.

To answer your original thoughts on this comment, I'd sort of defer to my response [above](https://github.com/bytecodealliance/wasmtime/pull/10404#discussion_r1999727160). My literal response just now above this one is actually a response to your thought [below](https://github.com/bytecodealliance/wasmtime/pull/10404#discussion_r1999477527)

crates/environ/src/compile/stack_maps.rs
fitzgen: Deduping the stack map data can definitely be a separate PR

crates/environ/src/compile/stack_maps.rs
```diff
@@ -0,0 +1,175 @@
...
+/// * `N` 4-byte little endian program counters, in ascending order.
+/// * `N` 4-byte little endian offsets.
+/// * Stack map data as 4-bit little endian integers.
+///
+/// The "offset" is an offset into the "stack map data" field which are encoded
+/// as:
+///
+/// * A 4-byte little-endian frame size
+/// * A 4-byte little-endian count of remaining bits: `M`
+/// * `M` 4-byte little-endian integers as a bit map.
```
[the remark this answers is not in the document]
alexcrichton: I've tried to document this better with some more art/boxes, but let me know if I can improve anything

cranelift/bitset/src/compound.rs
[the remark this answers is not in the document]
fitzgen: How about this:

* we call this method `iter_scalars`
* the iterator yields `ScalarBitSet<T>` items instead of `T` directly

alexcrichton:
@fitzgen did you want to review this over again after the recent changes? Or have more thoughts?

PR #10436, opened 2025-03-20
Remove `wasmtime_fuzzing::oracles::dummy::dummy_linker`

fitzgen:
Since that helper was written, we now have this same functionality built into `wasmtime::Linker`.

Unfortunately, we can't remove most of the `dummy` module, because it is used by the `api_calls` fuzzer to create dummy `wasmtime::Extern`s and we don't have off-the-shelf helpers for doing that in the core `wasmtime` API for that yet.

### DIFF
--- a/crates/fuzzing/src/oracles.rs
+++ b/crates/fuzzing/src/oracles.rs
@@ -366,14 +366,18 @@ fn compile_module(
 /// also yield the same values for each configuration, and we should assert
 /// that.
 pub fn instantiate_with_dummy(store: &mut Store<StoreLimits>, module: &Module) -> Option<Instance> {
+    assert!(Engine::same(store.engine(), module.engine()));
     // Creation of imports can fail due to resource limit constraints, and then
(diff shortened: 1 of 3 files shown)


alexcrichton:
I think this may not reduce the efficacy of fuzzing because the `Linker`-based version only handles functions while `dummy` handles everything?

fitzgen:
> I think this may not reduce the efficacy of fuzzing because the `Linker`-based version only handles functions while `dummy` handles everything?

Ah, I didn't realize that the linker method only handled functions. Is there a particular reason we didn't do all extern types?

alexcrichton:
IIRC it's because it was added from a CLI-centric perspective at the time where most modules don't import globals/memories/etc. I think it'd be reasonable to add support though.

fitzgen:
Ah it is because we don't have a `Store` in the `Linker` but we need one to create e.g. `Global`s.

fitzgen:
Probably not worth pursuing this PR any further then.

PR #10438, opened 2025-03-20
Add some infrastructure for keeping old CI working

alexcrichton:
This commit is preparation for the infrastructure to be used when supporting [Wasmtime LTS releases][rfc]. The goal here is to add some automation and infrastructure to perform a weekly build of all active release branches which will file an issue on failure. This should ideally keep branches up-to-date and ensure that we don't forget to backport any fixes to older branches. Or rather when we do forget to backport fixes this'll be a reminder to go do that anyway.

The general architecture here is:

* A new `ci-cron-trigger.yml` workflow is added.
* This new workflow runs once-a-week and runs a small script that triggers CI for all active release branches.
* The main CI, `main.yml`, is updated to file an issue on failure when triggered in this fashion.

While I was here I additionally removed the `schedule:` from the `main.yml` to instead fold the daily scheduling of CI runs into this new script as well. That way all our cron CI jobs are gated in workflows that require this exact repository meaning that forks won't be running cron jobs.

[rfc]: https://github.com/bytecodealliance/rfcs/pull/42

### DIFF
--- a/ci/trigger-release-branch-ci.rs
+++ b/ci/trigger-release-branch-ci.rs
@@ -1,0 +1,67 @@
+//! Helper script used by `.github/workflows/ci-cron-trigger.yml`
+
+use std::process::Command;
+
+fn main() {
+    let output = Command::new("git")
+        .arg("for-each-ref")
+        .arg("refs/remotes/origin")
+        .arg("--format")
+        .arg("%(refname)")
+        .output()
+        .unwrap();
+    assert!(output.status.success());
+    let mut releases = std::str::from_utf8(&output.stdout)
+        .unwrap()
+        .lines()
+        .filter_map(|l| l.strip_prefix("refs/remotes/origin/release-"))
+        .filter_map(|l| {
+            let mut parts = l.split('.');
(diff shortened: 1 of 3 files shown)


ci/trigger-release-branch-ci.rs
```diff
@@ -0,0 +1,67 @@
...
+    // keeping some older branch's CI working, which shouldn't be too hard.
+    to_trigger.extend(iter.by_ref().take(3));
+
+    // We support two LTS channels 12 versions apart. If one is already included
+    // in the above set of 3 latest releases, however, then we're only picking
+    // one historical LTS release.
+    let mut lts_channels = 2;
+    if to_trigger.iter().any(|(major, _, _)| *major % 12 == 0) {
+        lts_channels -= 1;
+    }
```
alexcrichton, line 45: This is technically somewhat inaccurate. If an LTS branch has been created, but not released, then we need to still look back 2 releases. The risk of that causing issues is pretty small though as that's a 2-week window and I otherwise don't know how to easily test for whether the latest branch has been released or not as part of this script.

fitzgen (approved):

PR #10439, opened 2025-03-20
Add `--stack-maps` to `wasmtime objdump`

alexcrichton:
Follow-up to #10404 and #10405

### DIFF
--- a/src/commands/objdump.rs
+++ b/src/commands/objdump.rs
@@ -36,11 +36,11 @@ pub struct ObjdumpCommand {
     address_jumps: bool,
 
     /// What functions should be printed (all|wasm|trampoline|builtin|libcall, default: wasm)
-    #[arg(long, value_parser = Func::from_str)]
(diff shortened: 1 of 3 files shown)


src/commands/objdump.rs
```diff
@@ -62,6 +62,10 @@ pub struct ObjdumpCommand {
     /// Whether or not to show information about what instructions can trap.
     #[arg(long)]
     traps: bool,
+
+    /// Whether or not to show information about stack maps.
+    #[arg(long)]
+    stack_maps: bool,
```
fitzgen, line 68: This one also seems like it should probably be on by default and won't be too noisy or anything like that.

fitzgen (approved):

src/commands/objdump.rs
alexcrichton: Ok I've done some rejiggering of the types and such for clap and I think I got something working, namely `--traps` turns things on, `--traps=false` turns things off, and it's on-by-default.

PR #10450, opened 2025-03-21
[24.0.0]  Pin CI runner images for some builds

alexcrichton:
This is an attempt to get CI working to address https://github.com/bytecodealliance/wasmtime/issues/10447

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -78,7 +78,7 @@ jobs:
   # Note that `wasmtime-platform.h` is excluded here as it's auto-generated.
   clangformat:
     name: Clang format
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
     - uses: actions/checkout@v4
       with:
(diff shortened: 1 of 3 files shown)


alexcrichton:
I temporarily tried to update the openvino installation action but abandoned that when it caused other failures. The original failure was resolved by pinning to a specific CI images, so I've stuck with that. All green now!

abrown:
I'm pretty sure that OpenVINO failure is related to an upstream breaking change in the OpenVINO C API, fixed in https://github.com/intel/openvino-rs/pull/149. Not sure the best way to resolve that in this branch, though: should we be backporting dependency updates? Seems not in the spirit of our LTS concept.

abrown (approved):
Makes sense to pin all this down for stability.

alexcrichton:
Yeah in the spirit of older branches I think it's best to leave dependency updates as a last resort. Still possible to do but only done when other avenues have been exhausted.

PR #10455, opened 2025-03-21
Keep Config options for disabling compiled-out features

### LINKED ISSUE #10454: Provide a mechanism for disabling on-by-default config options regardless of the enabled cargo feature flags
#### Feature

Provide some mechanism for disabling on-by-default config options without having to enable their associated cargo feature flags.

#### Benefit

The wasmtime [Config](https://docs.rs/wasmtime/latest/wasmtime/struct.Config.html) has a bunch of options like [Config::wasm_reference_types](https://docs.rs/wasmtime/latest/wasmtime/struct.Config.html#method.wasm_reference_types) that are enabled by-default if the corresponding cargo feature is enabled. Unfortunately, these config options are gated by the same feature flags making it impossible to turn them off without enabling the underlying cargo feature.

This is fine when building an application, but is a problem for library authors. If a library author wants to configure wasmtime in a specific way, they need to enable features (in cargo) they don't actually want so they can disable them in the config. Unfortunately, there's no way to actually _detect_ if the feature has been enabled.

#### Implementation

The simplest solution is to provide some form of `Config::minimal()` constructor (or something like that) that turns off all optional features. New versions of wasmtime may make some of these features mandatory, but, IMO, that's fine. I'm mostly looking for predictability.

#### Alternatives

Another option is to let users call all config methods even if the associated capabilities are disabled via cargo features:

1. Remove the `#[cfg(feature = ...)]` gates on all `Config` methods and their associated fields in the config.
2. Return an error from `Engine::new` if unsupported features are requested.

The main drawback is that we'll lose the compile-time check on whether or not the feature is enabled.

Stebalien:
This patch keeps config options for disabling default-enabled wasmtime features, even when those features are disabled via crate feature flags. That way, these wasm features can be disabled by libraries even if the crate-level features are later enabled by some other crate in the build-tree.

Specifically, the following methods are available on the `wasmtime::Config` type, regardless of the enabled crate features:

1. `Config::wasm_threads`.
2. `Config::wasm_reference_types`.
3. `Config::wasm_component_model`.
4. `Config::parallel_compilation`.

Of these, enabling any of the first three while the corresponding crate feature is disabled will lead to a runtime error when constructing the `wasmtime::Engine`. Enabling the last one, parallel compilation, will simply do nothing.

fixes #10454


Stebalien:
I've tried to keep this change as minimal as possible and stuck to default-enabled features. I'm also happy to extend it to (any of the) features that are currently default-disabled:

- `Config::async_support`
- `Config::wasm_function_references`
- `Config::wasm_gc`
- `Config::component_model_async`
- `Config::coredump_on_trap`
- `Config::wmemcheck`
- `Config::debug_adapter_modules`

Also note: I didn't bother with cranelift/winch options. As I understand it, those will only matter if you actually use cranelift/winch so they shouldn't affect any creates that aren't using the cranelift/winch features.

crates/wasmtime/src/config.rs
```diff
@@ -857,15 +857,15 @@ impl Config {
...
+    /// This feature is enabled by default as long as the `gc` crate feature (enabled
+    /// by default) is also enabled.
     ///
     /// # Errors
     ///
     /// The validation of this feature are deferred until the engine is being built,
     /// and thus may cause `Engine::new` fail if the `bulk_memory` feature is disabled.
     ///
     /// [proposal]: https://github.com/webassembly/reference-types
-    #[cfg(feature = "gc")]
```
Stebalien, line 868: Judging by the test failures, this feature flag shouldn't be here in the first place and this feature is enabled by-default regardless, right?

alexcrichton: Oh this is a bit subtle, and this also changed historically with Wasmtime. In essence though `REFERENCE_TYPES` should always be enabled regardless of what the crate features are, it's the `GC_TYPES` feature which is actually gated based on `feature = "gc"`. There's no `wasm_gc_types(...)` configuration method, though. 

I think this can be fixed by updating the validator to allow reference-types being enabled with `feature = "gc"` being turned off. It's `feature = "gc"` being turned off plus `GC_TYPES` being turned on that should be an error.

alexcrichton:
I think it might make sense to un-gate `wasm_*` configuration while you're here, but for the others I think it's ok to leave them as-is and deal with them as they come up later. Although if you'd like to handle them here that's also ok.

crates/wasmtime/src/config.rs
Stebalien: - What happens if I turn off `wasm_reference_types` but enable the GC feature?
- Should `wasm_gc` toggle `GC_TYPES`?

Stebalien: Also, does this apply to function references? That is, can I enable function references _without_ the GC feature?

Stebalien:
I've updated the PR per the new description to:

1. Always define the `Config::wasm_*` functions.
2. Allow setting `Config::wasm_reference_types` and `Config::wasm_function_references` (I'm not sure if this second one is correct, pending discussion in https://github.com/bytecodealliance/wasmtime/pull/10455#discussion_r2008436743).

crates/wasmtime/src/config.rs
alexcrichton: If reference-types is turned off but GC is enabled, then that means GC-using modules will fail to load/compile because the feature is turned off, and the active feature at compile time is basically bloat/inert code and doesn't get used.

For now the hope is we can internalize GC_TYPES to exclusively the `feature = "gc"` toggle, so `wasm_gc` should not turn on `GC_TYPES`. If `wasm_gc` is turned on though and `GC_TYPES` is off then you can still load modules, but only those that don't use types that require a GC.

For function-references we don't require a GC for typed function references so you should be able to use such a module without the GC feature. 

The end goal is to have `wasm_gc` on-by-default in the future, and in such a world we still want GC-disabled builds of Wasmtime to still load modules, so that's what `GC_TYPES` is doing, basically disallowing loading modules at runtime which use features disabled at compile time.

crates/wasmtime/src/config.rs
```diff
@@ -2165,6 +2159,26 @@ impl Config {
...
+
+        #[cfg(not(feature = "component-model-async"))]
+        if features.component_model_async() {
+            bail!("component-model-async support was requested but is not enabled in this build (requires the `component-model-async` crate feature)");
+        }
+
+        #[cfg(not(feature = "gc"))]
+        if features.gc_types() || features.gc() {
+            bail!("gc support was requested but is not enabled in this build (requires the `gc` crate feature)");
+        }
```
alexcrichton, line 2175: For this we should be able to remove this check entirely. I commented a bit above but on this specifically it's where `gc` is  safe to enable and it's just GC-using-things that aren't safe to enable. This is mostly because the GC proposal was larger than *just* GC references, it also introduced many other highly-related but still somewhat orthogonal things. The goal is to use `GC_TYPES` to separate out the runtime requirement but otherwise all the other stuff in the GC proposal is still valid.

crates/wasmtime/src/config.rs
```diff
@@ -1837,8 +1831,8 @@ impl Config {
     /// Disabling this will result in a single thread being used to compile
     /// the wasm bytecode.
     ///
-    /// By default parallel compilation is enabled.
-    #[cfg(feature = "parallel-compilation")]
+    /// This feature is enabled by default as long as the `parallel-compilation`
+    /// crate feature (enabled by default) is also enabled.
     pub fn parallel_compilation(&mut self, parallel: bool) -> &mut Self {
         self.parallel_compilation = parallel;
```
alexcrichton, line 1837: Mind adding a check below to return an error if this is explicitly enabled but disabled at compile time?

Stebalien:
Status update: I'm no longer working on the project that needed this PR so I'm going to close it for now. @alexcrichton thanks for taking the time to review this.

In case anyone else is interested in implementing this, the main remaining part is to resolve https://github.com/bytecodealliance/wasmtime/pull/10455#discussion_r2008491047.

PR #10456, opened 2025-03-21
Fix needs-stack-map propagation for variables' SSA block params

### LINKED ISSUE #10397: [GC] id from different slab
### Test Case

[slab-mismatch.zip](https://github.com/user-attachments/files/19246260/slab-mismatch.zip)

### Steps to Reproduce

Run the following command:
```
./target/debug/wasmtime  -W=all-proposals=y slab-mismatch.wasm
```

### Expected Results

It should not crash.

### Actual Results

We get a panic:
```
thread 'main' panicked at /home/jerome/sources/wasmtime/crates/slab/src/lib.rs:387:14:
id from different slab
```

### Versions and Environment

Wasmtime commit: a3381e48775912a3e1a68c05180932f3ce74c5b4

Operating system: Linux

Architecture: x64

### LINKED ISSUE #10398: [GC] assertion failed: `types.is_subtype(actual_ty, expected_ty)`
### Test Case

[subtype.zip](https://github.com/user-attachments/files/19246790/subtype.zip)

### Steps to Reproduce

Run the following command:
```
./target/debug/wasmtime  -W=all-proposals=y subtype.wasm
```

### Expected Results

It should not crash.

### Actual Results

We get a panic:
```
thread 'main' panicked at crates/wasmtime/src/runtime/vm/gc/func_ref.rs:81:13:
assertion failed: types.is_subtype(actual_ty, expected_ty)
```

### Versions and Environment

Wasmtime commit: a3381e48775912a3e1a68c05180932f3ce74c5b4

Operating system: Linux

Architecture: x64

### LINKED ISSUE #10399: [gc] Panic: every on-stack gc_ref inside a Wasm frame should have an entry in the VMGcRefActivationsTable
### Test Case

[not-in-table.zip](https://github.com/user-attachments/files/19248004/not-in-table.zip)

### Steps to Reproduce

Run the following command:
```
./target/debug/wasmtime  -W=all-proposals=y not-in-table.wasm
```

### Expected Results

It should not crash.

### Actual Results

We get a panic:
```
thread 'main' panicked at crates/wasmtime/src/runtime/vm/gc/enabled/drc.rs:239:13:
every on-stack gc_ref inside a Wasm frame should have an entry in the VMGcRefActivationsTable; 0x110 is not in the table
```

### Versions and Environment

Wasmtime commit: a3381e48775912a3e1a68c05180932f3ce74c5b4

Operating system: Linux

Architecture: x64

fitzgen:
We were previously propagating the needs-inclusion-in-stack-maps bit from a variable to the values resulting from `builder.use_var(my_var)` and passed to `builder.def_var(my_var, my_val)`. However, that does not cover every `ir::Value` generated for a given `ir::Variable`: it is missing block parameters that the SSA builder inserted and which are *not* then used directly as the result of `use_var`, but instead passed as arguments another block, and that block's parameter is then directly or indirectly used as the result of `builder.use_var(...)`. This chaining can go arbitrarily deep, based on the program's control flow. By not marking these block parameters as needing inclusion in stack maps, our stack maps would be incomplete, which meant that the GC could too-eagerly collect objects, which meant that we would then get use-after-free GC heap corruption bugs (still limited to within the GC heap because of our sandboxing approach) that would ultimately become (safe) panics when runtime code operated on corrupted objects.

The solution is for the `SSABuilder` to report all the block parameters that it inserted, and for which `ir::Variable` it inserted them. We already have a summary of the mutations that an `SSABuilder` performed, `SideEffects`, and which must be handled by the `FunctionBuilder` after, e.g., sealing a block, so we add this block-param-and-var information to it. Now, after the `FunctionBuilder` has the `SSABuilder` do its thing, it handles fixing up not only the modified blocks' status, which it was doing before, but also propagates the needs-inclusion-in-stack-maps metadata from variables to all those freshly-inserted block parameter values.

After this change, we successfully include *all* of a needs-inclusion-in-stack-maps variable's values in stack maps, which avoids the over-eager object reclamation and associated GC heap corruption.

Fixes #10397
Fixes #10398
Fixes #10399

### DIFF
--- a/cranelift/frontend/src/frontend.rs
+++ b/cranelift/frontend/src/frontend.rs
@@ -473,10 +473,10 @@ impl<'a> FunctionBuilder<'a> {
         };
         self.handle_ssa_side_effects(side_effects);
 
-        // If the variable was declared as needing stack maps, then propagate
-        // that requirement to all values derived from using the variable.
-        if self.func_ctx.stack_map_vars.contains(var) {
-            self.declare_value_needs_stack_map(val);
+        // If the variable was declared as needing stack maps, then we should
+        // have propagated that to the value as well.
+        if cfg!(debug_assertions) && self.func_ctx.stack_map_vars.contains(var) {
+            assert!(self.func_ctx.stack_map_values.contains(val));
         }
 
         Ok(val)
@@ -1178,9 +1178,24 @@ impl<'a> FunctionBuilder<'a> {
     }
 
     fn handle_ssa_side_effects(&mut self, side_effects: SideEffects) {
-        for modified_block in side_effects.instructions_added_to_blocks {
+        let SideEffects {
+            instructions_added_to_blocks,
+            params_added_to_blocks,
+        } = side_effects;
+
+        for modified_block in instructions_added_to_blocks {
             if self.is_pristine(modified_block) {
                 self.func_ctx.status[modified_block] = BlockStatus::Partial;
+            }
+        }
+
+        // Propagate needs-inclusion-in-stack-maps metadata from variables to
+        // their block parameter SSA values.
+        if !self.func_ctx.stack_map_vars.is_empty() {
+            for (val, var) in params_added_to_blocks {
+                if self.func_ctx.stack_map_vars.contains(var) {
+                    self.func_ctx.stack_map_values.insert(val);
+                }
             }
         }
     }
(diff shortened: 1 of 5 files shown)


cfallin (approved):
LGTM -- thanks for the thorough unit test!

PR #10461, opened 2025-03-24
Remove some unnecessary `UnsafeCell`s in GC code

fitzgen:
None of these are actually mutated while behind shared references. Everything, including interactions with JIT code, obeys Rust's borrowing rules. Running tests under MIRI still passes, verifying these claims. Therefore, we can remove these `UnsafeCell`s, which makes the code much easier to read and much less footgun-y since there is less `unsafe` code and less working with `UnsafeCell`s' raw pointers.

### DIFF
--- a/crates/wasmtime/src/runtime/vm/gc/enabled/drc.rs
+++ b/crates/wasmtime/src/runtime/vm/gc/enabled/drc.rs
@@ -58,7 +58,6 @@ use core::{
     num::NonZeroUsize,
     ops::{Deref, DerefMut, Range},
     ptr::NonNull,
-    slice,
 };
 use wasmtime_environ::drc::DrcTypeLayouts;
 use wasmtime_environ::{GcArrayLayout, GcStructLayout, GcTypeLayouts, VMGcKind, VMSharedTypeIndex};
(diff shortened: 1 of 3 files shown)


crates/wasmtime/src/runtime/vm/gc/enabled/drc.rs
```diff
@@ -934,20 +911,20 @@ impl VMGcRefTableAlloc {
...
     }
 }
 
-// This gets around the usage of `UnsafeCell` throughout the internals of this
-// allocator, but the storage should all be Send/Sync and synchronization isn't
+// This gets around the usage of `NonNull` throughout the internals of this
+// type, but the storage should all be Send/Sync and synchronization isn't
 // necessary since operations require `&mut self`.
 unsafe impl Send for VMGcRefTableAlloc {}
 unsafe impl Sync for VMGcRefTableAlloc {}
```
alexcrichton, line 930: Drive-by review for unrelated code, `SendSyncPtr` can remove the need for these

crates/wasmtime/src/runtime/vm/gc/enabled/null.rs
```diff
@@ -202,8 +201,8 @@ unsafe impl GcHeap for NullHeap {
         self.no_gc_count -= 1;
     }
 
-    fn heap_slice(&self) -> &[UnsafeCell<u8>] {
-        let ptr = self.heap.as_ptr().cast();
+    fn heap_slice(&self) -> &[u8] {
+        let ptr = self.heap.as_ptr();
         let len = self.heap.len();
         unsafe { core::slice::from_raw_parts(ptr, len) }
```
alexcrichton, line 207: More drive-by review: perhaps adding an `unsafe fn as_slice(&self)` to `Mmap`? (same for above for the drc heap)

tests/all/gc.rs
```diff
@@ -1089,7 +1089,7 @@ fn drc_transitive_drop_cons_list() -> Result<()> {
     let pre = StructRefPre::new(&mut store, struct_ty.clone());
     let num_refs_dropped = Arc::new(AtomicUsize::new(0));
 
-    let len = if cfg!(miri) { 2 } else { 100 };
+    let len = if true || cfg!(miri) { 2 } else { 100 };
```
alexcrichton, line 1092: debugging code I think?

alexcrichton (approved):

PR #10468, opened 2025-03-25
`cranelift-frontend`: Propagate needs-stack-map from variables to values during finalization

fitzgen:
Rather than trying to propagate the needs-stack-map bit from variables to values online, while we are building the IR and defining and using variables, wait until the function is being finalized, and then propagate everything all at once. This avoids potential repeated work and is easier to ensure that it is complete and covers all values associated with a variable, since by the time we are finalizing the function, we won't add any new values for a variable that we will need to keep track of and propagate this information to.

This also means that we can remove the `params_added_to_blocks` vector from the SSA side effects structure, since it was only used to online-update the `stack_map_values` set.

### DIFF
--- a/cranelift/frontend/src/ssa.rs
+++ b/cranelift/frontend/src/ssa.rs
@@ -67,21 +67,14 @@ pub struct SideEffects {
     /// This field signals if it is the case and return the `Block` to which the initialization has
     /// been added.
     pub instructions_added_to_blocks: Vec<Block>,
-
-    /// When a variable does not have a single, dominating definition of a use,
-    /// we must insert block parameters to control-flow join points. This field
-    /// contains all of the block parameters we inserted, along with the
-    /// variable that they are associated with.
-    pub params_added_to_blocks: Vec<(Value, Variable)>,
 }
 
 impl SideEffects {
     fn is_empty(&self) -> bool {
         let Self {
             instructions_added_to_blocks,
-            params_added_to_blocks,
         } = self;
-        instructions_added_to_blocks.is_empty() && params_added_to_blocks.is_empty()
+        instructions_added_to_blocks.is_empty()
     }
 }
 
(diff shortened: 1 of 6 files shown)


cranelift/frontend/src/ssa.rs
```diff
@@ -195,6 +188,12 @@ fn emit_zero(ty: Type, mut cur: FuncCursor) -> Value {
 /// Phi functions.
 ///
 impl SSABuilder {
+    /// Get all of the values associated with the given variable that we have
+    /// inserted in the function thus far.
+    pub fn values_for_var(&self, var: Variable) -> impl Iterator<Item = Value> + '_ {
+        self.variables[var].values().filter_map(|v| v.expand())
```
bjorn3, line 194: This is quadratic, right? `O(#vars * #vals)`

fitzgen: The `values_for_var` function is not `O(vars * vals)` but the loop calling `values_for_vars` in finalization is. I'm unsure which part of the code you are referring to with this comment.

And while that finalization loop is `O(vars * vals)` in the worse case where every variable is marked as needing a stack map and every value is associated with not just one but every single variable, I wouldn't really call that quadratic because `vars` and `vals` are two independent big-O variables. Similarly, I wouldn't really call an `O(edges * vertices)` graph algorithm quadratic.

Either way, I don't think it actually matters in this case. It shouldn't be possible to write a Wasm program that constructs that worst case input or anything like it. And it would be pretty difficult to accidentally generate this code with any typical `cranelift_frontend` usage (even in the face of translating untrusted code to CLIF). It would require using the `cranelift_frontend` API directly in a fairly malicious way to create that worst case.

cfallin: Maybe one way of seeing this is that values are subordinate to variables: so we aren't taking a cross-product of all vars with all vals, but performing some work for each value, and we get to all the values by traversing the per-var bins of values.

Said more simply: this is linear in the size of the IR; but all of compilation is already at least linear in the size of the IR, so we aren't introducing any additional blowup here, I think.

cfallin (approved):
LGTM; this is simpler as well as fixing the issue, so thanks for revisiting!

PR #10473, opened 2025-03-26
Fix some `unreachable!()` with wasm components

alexcrichton:
This commit resolves a few panics in Wasmtime's compilation of components with respect to tags and the exception-handling wasm proposal. Despite exception-handling not being enabled in Wasmtime at this time these panics were still reachable due to an issue in wasm-tools's validation not being exhaustive enough for this proposal feature combination: bytecodealliance/wasm-tools#2114.

This is not full support for exceptions since it can't be tested yet, but this should be enough to ensure that the validator will get far enough to flag components as otherwise invalid due to using tags that can't be present (as core wasm tags can't be present in core modules right now, the `EXCEPTIONS` feature is always disabled).


alexcrichton:
Thanks to @0xalpharush for discovering this issue and letting us know!

dicej (approved):

PR #10475, opened 2025-03-26
asm: improve DSL's understanding of prefixes

abrown:
Previously, Cranelift's x64 backend only understood a subset of the available prefixes that can be encoded and the documentation for these was quite sparse. This change upgrades the assembler to understand all kinds of prefixes, even those as yet unused, and drastically increases the amount of documentation around these prefixes. This, along with some new `Group`-based typing, should prevent accidental misuse but also increase confidence that this implementation is correctly covering the possible cases.

### DIFF
--- a/cranelift/assembler-x64/meta/src/dsl/encoding.rs
+++ b/cranelift/assembler-x64/meta/src/dsl/encoding.rs
@@ -98,12 +98,6 @@ pub struct Rex {
     /// the instruction's opcode."
     pub digit: Option<u8>,
     /// The number of bits used as an immediate operand to the instruction.
-    ///
-    /// From the reference manual: "a 1-byte (ib), 2-byte (iw), 4-byte (id) or
-    /// 8-byte (io) immediate operand to the instruction that follows the
-    /// opcode, ModR/M bytes or scale-indexing bytes. The opcode determines if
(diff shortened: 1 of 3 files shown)


cranelift/assembler-x64/meta/src/dsl/encoding.rs
```diff
@@ -301,117 +298,256 @@ pub struct Opcodes {
...
+    pub group3: Option<Group3Prefix>,
+    pub group4: Option<Group4Prefix>,
+}
+
+impl Prefixes {
+    /// Parse a slice of `bytes` into a set of prefixes, returning both the
+    /// configured [`Prefixes`] as well as any remaining bytes.
+    fn parse(mut bytes: &[u8]) -> (Self, &[u8]) {
+        let mut prefixes = Self::default();
+        while prefixes.try_assign(bytes[0]).is_ok() {
```
cfallin, line 342: `while bytes.len() > 0 && prefixes.try_assign(bytes[0]).is_ok()` ?

(I think this won't matter as long as you only pass in slices that end in true opcodes, and no opcode has the same value as a prefix, but it caught my eye as that's quite subtle)

cfallin (approved):
Looks good!

PR #10478, opened 2025-03-27
asm: fix trace logging again

### LINKED ISSUE #10476: Panic in trace logging output for tests on x64 again
For example,

```
$ RUST_LOG=trace cargo test --test wast issue-10353.wast
...
thread '<unnamed>' panicked at cranelift/codegen/src/isa/x64/inst/external.rs:134:9:
internal error: entered unreachable code
stack backtrace:
   0: rust_begin_unwind
             at /rustc/21fe748be15271ea5804e0507cd699b675efe038/library/std/src/panicking.rs:681:5
   1: core::panicking::panic_fmt
             at /rustc/21fe748be15271ea5804e0507cd699b675efe038/library/core/src/panicking.rs:75:14
   2: core::panicking::panic
             at /rustc/21fe748be15271ea5804e0507cd699b675efe038/library/core/src/panicking.rs:145:5
   3: cranelift_codegen::isa::x64::inst::external::enc_gpr
             at ./cranelift/codegen/src/isa/x64/inst/external.rs:134:9
   4: cranelift_codegen::isa::x64::inst::external::<impl cranelift_assembler_x64::api::AsReg for cranelift_codegen::isa::x64::inst::args::Gpr>::enc
             at ./cranelift/codegen/src/isa/x64/inst/external.rs:92:9
   5: <cranelift_assembler_x64::mem::Amode<R> as core::fmt::Display>::fmt
             at ./cranelift/assembler-x64/src/mem.rs:172:48
   6: <T as alloc::string::ToString>::to_string
             at /home/nick/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/string.rs:2688:9
   7: cranelift_assembler_x64::mem::GprMem<R,M>::to_string
             at ./cranelift/assembler-x64/src/mem.rs:260:35
   8: <cranelift_assembler_x64::inst::addq_rm<R> as core::fmt::Display>::fmt
...
```

cc @abrown

abrown:
As pointed out in #10476, there are a few more locations in which the raw encoding of a register is used when pretty-printing memory operands. Using the raw encoding before register allocation results in a panic. This change ensures that the additional locations delegate to the `to_string()` method which is overriden in `cranelift-codegen` to know how to pretty-print virtual registers. Closes #10476.

### DIFF
--- a/cranelift/assembler-x64/src/gpr.rs
+++ b/cranelift/assembler-x64/src/gpr.rs
@@ -79,6 +79,11 @@ impl<R: AsReg> NonRspGpr<R> {
         Self(reg)
     }
 
+    /// See [`Gpr::to_string`].
+    pub fn to_string(&self, size: Size) -> String {
+        self.0.to_string(Some(size))
+    }
+
     /// See [`Gpr::enc`].
     ///
     /// # Panics
--- a/cranelift/assembler-x64/src/mem.rs
+++ b/cranelift/assembler-x64/src/mem.rs
@@ -3,7 +3,6 @@
 use crate::api::{AsReg, CodeSink, Constant, KnownOffset, KnownOffsetTable, Label, TrapCode};
 use crate::gpr::{self, NonRspGpr, Size};
 use crate::rex::{encode_modrm, encode_sib, Imm, RexFlags};
-use crate::xmm;
 
 /// x64 memory addressing modes.
 #[derive(Clone, Debug)]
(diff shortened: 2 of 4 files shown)


cfallin (approved):
Thanks!

Would it be possible to add some sort of test to make sure that trace-logging works for a whole compilation? Perhaps programmatically setting the log-level to trace with the `log` crate or related and then compiling some simple function body in a unit test, or similar.

abrown:
> Would it be possible to add some sort of test to make sure that trace-logging works for a whole compilation? Perhaps programmatically setting the log-level to trace with the `log` crate or related and then compiling some simple function body in a unit test, or similar.

Should it just be a CLI test?

alexcrichton:
One thing I'll note is that this'd be easiest during `tests/all/cli_tests.rs` as that's where we can scope setting an env var to just one process, but that's disconnected from wast testing. What you can do is pick any of the test-programs modules/components and compile that with the env var and ensure that it doesn't panic.

alexcrichton (approved):

PR #10481, opened 2025-03-27
Add documentation for Wasmtime's LTS releases

alexcrichton:
With Wasmtime's [LTS
releases](https://github.com/bytecodealliance/rfcs/pull/42) this commit documents the various process changes and updates to our release process. Additionally some improvements are made to the release documentation with respect to showing current versions.


docs/stability-release.md
```diff
@@ -1,18 +1,46 @@
...
+along with the dates around their release process. Rows in **bold** are
+actively supported at this time.
+
+| Version    | LTS | Branch Date | Release Date | EOL Date |
+|------------|-----|-------------|--------------|----------|
+<!-- JS edits in the table -->
+
+In more visual form this is a gantt chart of the current release trains:
+
+<!-- JS edits in the chart -->
```
alexcrichton, line 32: An example of what this generates:

Screenshot 2025-03-27 at 3 08 49 PM

alexcrichton:
I originally wanted to document the property of "when backporting must update all branches" and got a bit carried away and went ahead and documented the LTS updates.

docs/stability-release.md
cfallin: That is some impressive DOM-tree-editing hackery! The end result looks great.

At the risk of complexifying a little bit: could we add a placeholder ("view this document as HTML in a browser with JavaScript enabled to see contents" or somesuch) so that if the Markdown is processed in some other way -- for example, converted to print pdf by pandoc -- or viewed by a user with JS disabled, they aren't confused?

fitzgen (approved):
Can you put version numbers in the JS's file paths so that if we ever want/need to upgrade, things are a little easier?

docs/stability-release.md
alexcrichton: Sounds good yeah, added some `<noscript>` tags to indicate that it's empty without JS

PR #10483, opened 2025-03-28
Improve errors with pooling-by-default `serve` command

### LINKED ISSUE #10482: Wasmtime -W cli options do not seem to have any effect? max-table-elements, max-memory-size, Error - table index exceeds the limit of 20000
I'm having trouble with increasing the allowed table elements per table in the pooling allocator - i'm hitting the hard limit of 20k

Image

### Test Case

Too big of a wasm file to upload here.... (32Mb), it is accessible here: https://github.com/MA3CIN/wasmtime-error/blob/main/wasm-server.wasm

### Steps to Reproduce

Run the wasm file with wastime server: "wasmtime serve wasm-server.wasm"

### Expected Results

I expected the allowed table elements per table in the pooling allocator limit to be increased when using the -W max-table-elements cli arguments

### Actual Results

I'm still hitting the limit

Image

### Versions and Environment

Wasmtime version or commit: wasmtime 31.0.0 (7a9be587f 2025-03-20)

Operating system: Ubuntu 22.04.3 LTS

Architecture: N/A, single wasm instance

### Extra Info

Works fine when i run it with wasmtime run, i only hit the limit with wasmtime serve..

Maybe a similar problem to https://github.com/bytecodealliance/wasmtime/issues/8504

alexcrichton:
This commit is intended to address #10482 where the defaults of the `wasmtime serve` subcommand produced a confusing and surprising error. Specifically the `-Wmax-table-elements` option, prior to this change, only affected the store limiter used and didn't actually affect the pooling allocator settings. That meant that if a module exceeded the limits of the pooling allocator it would produce an error message that seemed like `-Wmax-table-elements` would fix but it wouldn't actually.

Two changes in this commit are meant to address this:

* Errors from the pooling allocator have been update to mention "pooling allocator" within them somewhere to surface where the error is coming from.
* The `-Wmax-memory-size` and `-Wmax-table-elements` configuration are now applied to the pooling allocator automatically if the corresponding `-Opooling-*` option isn't passed.

That should mean that the original error should be a bit easier to debug while the attempted solution will also work.

Closes #10482

### DIFF
--- a/crates/cli-flags/src/lib.rs
+++ b/crates/cli-flags/src/lib.rs
@@ -826,7 +826,9 @@ impl CommonOptions {
                     if let Some(limit) = self.opts.pooling_total_tables {
                         cfg.total_tables(limit);
                     }
-                    if let Some(limit) = self.opts.pooling_table_elements {
+                    if let Some(limit) = self.opts.pooling_table_elements
+                        .or(self.wasm.max_table_elements)
+                    {
                         cfg.table_elements(limit);
                     }
                     if let Some(limit) = self.opts.pooling_max_core_instance_size {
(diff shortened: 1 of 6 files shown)


abrown (approved):

abrown:
...but it looks like there are a few more tests to fix up.

alexcrichton:
Turns out the test I added is the first test of the CLI using the pooling allocator. Also turns out that with the default configuration the pooling allocator couldn't work on 32-bit platforms. I've adjusted the pooling allocator defaults on 32-bit platforms to work-by-default (but it's still not that useful)

Issue #10479, opened 2025-03-27
`cranelift_codegen`: panic with `shift left with overflow`


Robbepop:
- Version: Wasmtime v31.0.0
- Architecture: `aarch64`
- OS: MacOS

While differentially fuzzing Wasmi and Wasmtime I encountered this Cranelift panic:

```
NOTE: libFuzzer has rudimentary signal handlers.
      Combine libFuzzer with AddressSanitizer or similar for better crash reports.
SUMMARY: libFuzzer: deadly signal
MS: 2 EraseBytes-EraseBytes- DE: "asmtime.info"-; base unit: 0000000000000000000000000000000000000000
artifact_prefix='/Users/me/Documents/GitHub/wasmi/fuzz/artifacts/differential/'; Test unit written to /Users/me/Documents/GitHub/wasmi/fuzz/artifacts/differential/minimized-from-de00244b92a84f05f080a7fdff69dee621206cd4
*********************************
CRASH_MIN: minimizing crash input: '/Users/me/Documents/GitHub/wasmi/fuzz/artifacts/differential/minimized-from-de00244b92a84f05f080a7fdff69dee621206cd4' (466 bytes)
CRASH_MIN: executing: target/aarch64-apple-darwin/release/differential -artifact_prefix=/Users/me/Documents/GitHub/wasmi/fuzz/artifacts/differential/ -runs=255 /Users/me/Documents/GitHub/wasmi/fuzz/artifacts/differential/minimized-from-de00244b92a84f05f080a7fdff69dee621206cd4 2>&1
CRASH_MIN: '/Users/me/Documents/GitHub/wasmi/fuzz/artifacts/differential/minimized-from-de00244b92a84f05f080a7fdff69dee621206cd4' (466 bytes) caused a crash. Will try to minimize it further
CRASH_MIN: executing: target/aarch64-apple-darwin/release/differential -artifact_prefix=/Users/me/Documents/GitHub/wasmi/fuzz/artifacts/differential/ -runs=255 /Users/me/Documents/GitHub/wasmi/fuzz/artifacts/differential/minimized-from-de00244b92a84f05f080a7fdff69dee621206cd4 -minimize_crash_internal_step=1 -exact_artifact_path=/Users/me/Documents/GitHub/wasmi/fuzz/artifacts/differential/minimized-from-6bc04e97d6ade6982223726c65803bbebce42325 2>&1
INFO: Running with entropic power schedule (0xFF, 100).
[...]

PR #10487, opened 2025-03-29
fuzz: enable `relxed-simd` for the Wasmi oracle

Robbepop:
This enables `relaxed-simd` for the Wasmi differential fuzzing oracle.

Note that the `relaxed-simd` proposal support is very new for Wasmi. I have only ran Wasmi's differential fuzzing for ~45 minutes before it crashed with [the already known Wasmtime bug](https://github.com/bytecodealliance/wasmtime/issues/10479).

Feel free to accept or close this PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4062,9 +4062,9 @@ dependencies = [
 
 [[package]]
 name = "wasmi"
-version = "0.43.0"
+version = "0.44.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "60160ffa66a3f95ae969aaf9cac28591b919a677d68faf29810c4989d5b0cad8"
+checksum = "7bdd371943f36290dbb9abe6c66d6f134d0d04a04e841db1350323f9ce0bf853"
 dependencies = [
  "arrayvec",
  "multi-stash",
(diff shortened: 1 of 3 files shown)


alexcrichton:
Unfortunately I don't think this will do anything at this time due to [this line](https://github.com/bytecodealliance/wasmtime/blob/ae69c92e756890ef7d15826b471ee083ff51ed5b/crates/fuzzing/src/generators/config.rs#L103-L105). I'm not sure how we can change that because while wasmi/wasmtime may be able to be differentially fuzzed against each other it assumes certain wasmtime configuration which isn't plumbed yet and differential execution against v8 or even the spec interpreter in the future may not be possible

Robbepop:
@alexcrichton thanks for the clarifications.

I agree, this won't be very useful for Wasmtime at the moment. I currently do perform `relaxed-simd` differential fuzzing in Wasmi's fuzzer and it seems to match Wasmtime's semantics. However, I just got the weird feeling that even with `config.relaxed_simd_deterministic()` enbabled this might change since `relaxed-simd` instructions can remain deterministic while changing semantics as long as the new semantics are deterministic across platforms.

Robbepop:
> I think the intention is that the deterministic profile of relaxed-simd is spec'd as one thing and that "thing" won't change. Officially though there aren't profiles at the moment, so it's more of a handshake between folks at this point that the deterministic profile won't be changing AFAIK

Ah I didn't know about the deterministic profile but it seems I got it right by accident.

Robbepop:
I think this can be closed for now.

PR #10497, opened 2025-03-31
wasmtime-environ: add initial table support

abrown:
In order to lower the [`thread.spawn_indirect`] canonical builtin from the component model, we need to be able to refer to the tables exported from instance. This change adds the initial plumbing to access such tables.

[`thread.spawn_indirect`]: https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md#-canon-threadspawn_indirect

### DIFF
--- a/crates/environ/src/component/vmcomponent_offsets.rs
+++ b/crates/environ/src/component/vmcomponent_offsets.rs
@@ -44,6 +44,8 @@ pub struct VMComponentOffsets<P> {
     pub num_lowerings: u32,
     /// The number of memories which are recorded in this component for options.
     pub num_runtime_memories: u32,
+    /// The number of tables which are recorded in this component for options.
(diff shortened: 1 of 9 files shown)


crates/environ/src/component/vmcomponent_offsets.rs
```diff
@@ -262,6 +267,20 @@ impl<P: PtrSize> VMComponentOffsets<P> {
         self.runtime_memories() + index.as_u32() * u32::from(self.ptr.size())
     }
 
+    /// The offset of the base of the `runtime_tables` field
+    #[inline]
+    pub fn runtime_tables(&self) -> u32 {
+        self.tables
+    }
```
alexcrichton, line 274: Mind updating the comment at the top of this file with the pseudo-struct definition to include the table section?

PR #10499, opened 2025-04-01
Improve table support for components

abrown:
See commit messages for details.

### DIFF
--- a/crates/wasmtime/src/runtime/vm/vmcontext.rs
+++ b/crates/wasmtime/src/runtime/vm/vmcontext.rs
@@ -1486,3 +1486,47 @@ impl VMOpaqueContext {
         ptr.cast()
     }
 }
+
+/// The fields a component needs to access to use a WebAssembly table imported
+/// from another instance; see the identically-defined [`VMTableImport`].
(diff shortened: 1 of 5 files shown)


crates/wasmtime/src/runtime/vm/vmcontext.rs
```diff
@@ -1486,3 +1486,47 @@ impl VMOpaqueContext {
         ptr.cast()
     }
 }
+
+/// The fields a component needs to access to use a WebAssembly table imported
+/// from another instance; see the identically-defined [`VMTableImport`].
+#[derive(Debug, Copy, Clone)]
+#[repr(C)]
+pub struct VMTableUse {
```
abrown, line 1494: I'm not really sure what to call this: it is structurally equivalent to a `VMTableImport` but I didn't want to reuse that `struct` since it could confuse things. Suggestions?

abrown: See a2332f7. What do you think about a `VMTableImport` to `VMTable` renaming?

alexcrichton (approved):

Issue #10437, opened 2025-03-20
Define default-value constructors for various Wasm types in the host API


fitzgen:
We already have code to create default values in order to implement https://docs.rs/wasmtime/latest/wasmtime/struct.Linker.html#method.define_unknown_imports_as_default_values

We have duplicates of that code in the fuzzing utility crate: https://github.com/bytecodealliance/wasmtime/blob/main/crates/fuzzing/src/oracles/dummy.rs

We should just expose those constructors in the public API of the `wasmtime` crate, move the various fuzzing oracles over to these methods, and then delete 95% of that `wasmtime_fuzzing::oracles::dummy` module.

I am thinking that we will want to expose these as `pub fn default_value(&self, store: impl AsContextMut) -> Option<T>` methods on `wasmtime::{ExternType,GlobalType,MemoryType,ValType,...}` where `T` is `wasmtime::Memory` for the `wasmtime::MemoryType` method, etc...

shashforge:
Hi all,

I’ve traced through `crates/wasmtime/src/runtime/types.rs` and can confirm that all five `default_value` constructors are already exposed in the public API:

- `ExternType::default_value(&self, store) -> Result<Extern>`
- `GlobalType::default_value(&self, store) -> Result<RuntimeGlobal>`
- `MemoryType::default_value(&self, store) -> Result<RuntimeMemory>`
- `TableType::default_value(&self, store) -> Result<RuntimeTable>`
- `FuncType::default_value(&self, store) -> Result<Func>`

This is my first issue in Wasmtime, and I’d love your suggestions and input on what to do next in this PR:

- Are there any missing corner-case tests you’d like me to add?  
- Should I draft an example or update docs to showcase these APIs?  
- Is there anything else I can improve or clean up before we close this issue?

Looking forward to your guidance—thanks for your help!

fitzgen:
Hi @shashi1687,

It looks like `{Global,Table,Memory}Type::default_value` do not have any any doc comments. Adding a couple sentences to those docs would be fantastic. Once that is done, I think we can close this issue.

shashi1687:
Hi @fitzgen , thanks for the feedback! I’ve added doc comments to:

- `GlobalType::default_value`
- `TableType::default_value`
- `MemoryType::default_value`

Could you please take another look and let me know if there’s anything else to improve? Thanks! 😊

shashforge:
[...]

Issue #10252, opened 2025-02-20
Implement WebAssembly Tags (exceptions + stack-switching)


dhil:
PR #10251 adds support for exception and control tags in Wasmtime. At the time of writing it lacks:

- [ ] C API support
- [ ] Component model support
- [ ] Fuzzing support

PR #10500, opened 2025-04-01
Expose default value constructors for Wasm types in host API

Gonzalosilvalde:
## Description:

This pull request addresses issue #10437 by adding default_value(&self, store: impl AsContextMut) -> Option<T> methods to various Wasm types (e.g., ExternType, GlobalType, MemoryType, ValType, etc.), enabling default value construction for the host API.

## Changes:

- Implemented new default_value in different types in types.rs.
- Refactored duplicate code from the fuzzing utility by moving default value construction to these new methods.
- Removed the majority of the code from the wasmtime_fuzzing::oracles::dummy module.
- Modified code that used dummy functions

## Additional Notes:

- No documentation comments were added.
- The error in oracles.rs was commented out because the variable e no longer exists.
- For dummy.rs/dummy_linker, a placeholder error “ERROR” has been added, because I didn't know what error to put on it.
- The comment // FIXME #10252 was relocated to the corresponding default_value method implementation.

### DIFF
--- a/crates/wasmtime/src/runtime/types.rs
+++ b/crates/wasmtime/src/runtime/types.rs
@@ -345,7 +345,7 @@ impl ValType {
             WasmValType::Ref(r) => Self::Ref(RefType::from_wasm_type(engine, r)),
         }
     }
-    ///
+    /// Construct a default value. Returns None for non-nullable Ref types, which have no default.
     pub fn default_value(&self) -> Option<Val> {
         match self {
             ValType::I32 => Some(Val::I32(0)),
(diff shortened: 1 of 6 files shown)


crates/wasmtime/src/runtime/types.rs
```diff
@@ -341,6 +345,23 @@ impl ValType {
             WasmValType::Ref(r) => Self::Ref(RefType::from_wasm_type(engine, r)),
         }
     }
+    ///
```
pchickey, line 348:
```suggestion
    /// Construct a default value. Returns None for non-nullable Ref types, which have no default.
```

pchickey (approved):
Thanks. Can you fill in the remaining empty doc comments along the lines of my suggestions?

PR #10501, opened 2025-04-01
Update wasm-tools crates to 1.228.0

abrown:
See details in comments.

### DIFF
--- a/crates/fuzzing/src/generators/config.rs
+++ b/crates/fuzzing/src/generators/config.rs
@@ -139,13 +139,13 @@ impl Config {
             extended_const,
             wide_arithmetic,
             component_model_async,
+            component_model_async_builtins,
+            component_model_async_stackful,
             simd,
 
             hogs_memory: _,
             nan_canonicalization: _,
             gc_types: _,
-            component_model_async_builtins: _,
-            component_model_async_stackful: _,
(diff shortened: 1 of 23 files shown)


abrown:
I'm opening this as a draft so we can figure out what to do about the failing `component-model-async` WAST tests. Some failures I've noticed:
- some canonical builtins have changed their return type from `i32` to `i64`... this might mean changes to the CLIF lowering
- we need to enable the right `CM_ASYNC_*` flags for some of these tests to continue to run (or disable them temporarily?)

crates/fuzzing/src/generators/config.rs
```diff
@@ -144,6 +144,8 @@ impl Config {
             hogs_memory: _,
             nan_canonicalization: _,
             gc_types: _,
+            component_model_async_builtins: _,
+            component_model_async_stackful: _,
```
abrown, line 148: To respond to the helpful CI reminder bot, I just want to point out that this change does not in fact exercise these new CM-level features.

alexcrichton (approved):

crates/fuzzing/src/generators/config.rs
abrown: ...but no longer: 326ed15.

Issue #10488, opened 2025-03-29
Cranelift: remove multiple return value support in order to allow feasible implementation of exception handling?


cfallin:
I am currently implementing exception handling for Cranelift ([WIP branch](https://github.com/cfallin/wasmtime/tree/lets-be-truly-exceptional)) and I believe I have hit a complexity wall that merits further discussion about Cranelift's IR design.

Currently, we permit signatures to have arbitrary numbers of return values, and we lower these to use stack slots. This adds a nontrivial bit of complexity to ABI handling, but we manage. It means that after a true `call` instruction, we may have loads that are nominally part of the callsite but are separate VCode instructions.

However, adding `try_call` instructions, which define return values as block-call args for their successors, has led me into a difficult spot when these extra "return-value instructions" exist -- I have tried all of the below approaches:

- We cannot emit these as instructions after the` try_call` in the same block, because the `try_call` is a terminator; it acts as a branch to either the normal-return target or exceptional return targets; this is a hard constraint and unlikely to change (e.g. via EBBs).
- Perhaps we could emit the loads as part of the one call VCode instruction. This is the path that was taken for a post-call SP adjustment for tail-call support. However:
  - We can't hold a `Vec<Inst>` inside the `CallInfo` or whatever: Pulley's instruction type is parameterized on a generic parameter that selects between 32- and 64-bit targets, we need to define the inst enum in ISLE, and ISLE doesn't have generics.
  - We can't add a notion of "PReg or stack offset" to the `CallRetPair`s, and emit loads directly: we need to support an arbitrarily large number of vreg definitions, and if they're all constrained as registers, this breaks down after we fill the machine registers (regalloc panics because the program is un-allocatable per its constraints).
    - We can't add these loads with "any" constraints because they we need to codegen loads from stack offsets to spillslots and at this point we are re-inventing regalloc's handling of the various cases of memory-to-memory moves with or without temporaries available.
- We can't split control-flow edges with critical-edge blocks and emit the retval instructions into the edge block when processing the main branch because VCode emission has very tight constraints about instruction lowering order, contiguous instruction IDs, and instruction ID ordering being the same as block ordering.
[...]

PR #10502, opened 2025-04-01
Cranelift: remove return-value instructions after calls at callsites.

cfallin:
This PR addresses the issues described in #10488 in a more head-on way: it removes the use of separate "return-value instructions" that load return values from the stack, instead folding these loads into the semantics of the call VCode instruction.

This is a prerequisite for exception-handling: we need calls to be workable as terminators, meaning that we cannot require any other (VCode) instructions after the call to define the return values.

In principle, this PR starts simply enough: the return-locations list on the `CallInfo` that each backend uses to provide regalloc metadata is updated to support a notion of "register or stack address" as the source of each return value, and this list is now used for both kinds of returns, not just returns in registers. Shared code is defined in `machinst::abi` used by all backends to perform the requisite loads.

In order to make this work with more defined values than fit in registers, however, this PR also had to add support for "any"-constrained registers to Cranelift, and handling allocations that may be spillslots. This has always been supported by RA2, but this is the first time that Cranelift uses them directly (previously they were used only internally in RA2 as lowerings from other kinds of constraints like safepoints). This requires encoding a spillslot index in our `Reg` type.

There is a little bit of complexity around handling the loads/defs as well: if we have a return value on-stack, and we need to put it in a spillslot, we cannot do a memory-to-memory move directly, so we need a temporary register. Earlier versions of this PR allocated another temp as a vreg on the call, but this doesn't work with all calling conventions (too many clobbers). For simplicity I picked a particular register that is (i) clobbered by calls and (ii) not used for return values for each architecture (x86-64's tailcall needed to lose one return-in-register slot to make this work).

This removes retval insts from the shared ABI infra completely. s390x is different, still, because it handles callsite lowering from ISLE; we will need to address that separately for exception support there.


cfallin:
I'll note that this appears to pessimize codegen somewhat for many-return-values (more than fit in registers, so more than 8? on aarch64/x86-64), since this does a manual load-from-stack/store-to-spillslot rather than letting regalloc handle it; that seems acceptable given the complexity reduction and unblocking exception support, IMHO.

cranelift/codegen/src/machinst/abi.rs
```diff
@@ -2438,6 +2464,75 @@ impl<M: ABIMachineSpec> CallSite<M> {
...
+                                i64::from(stackslots_size)
+                                    + i64::from(M::word_bytes())
+                                        * ((spillslot.index() as i64) + (part as i64)),
+                            ),
+                            temp.to_reg(),
+                            M::word_type(),
+                        ));
+                    }
+                } else {
+                    emit(M::gen_load_stack(*amode, *vreg, *ty));
```
alexcrichton, line 2529: Could this case assert that `vreg` was not allocated to `temp`?

alexcrichton: Also, is it possible to somewhat easily assert that `temp` is callee-saved? (or caller-save, I forget which would be appropriate here)

alexcrichton (commented):
Overall this looks reasonable to me, and I have a passing thought, but I'm going to defer to @fitzgen for review rather than myself as I think he's been chatting with you more about this

cranelift/codegen/src/machinst/abi.rs
cfallin: Done (to assert), good idea. It's a little trickier to assert anything about whether this is a caller-saved (volatile) register we can use: this helper is called after we've thrown away all the ABI info except for what's in the `CallInfo`, so we'd have to somehow preserve that.

cfallin:
Just pushed a fix for aarch64 -- it turns out our optimization to avoid unnecessary clobber-saves for the lower half of float registers (`is_included_in_clobbers`) can no longer apply if we have other defs (the stack-carried values) that allocate into other registers; those other registers are indeed clobbered by our function body! The pessimization there is, again, only on very-many-returned-values cases; the happy path of returns-that-fit-in-regs is unaffected.

alexcrichton:
> this helper is called after we've thrown away all the ABI info

Could the `temp` parameter  to `CallInfo::emit_retval_loads` turn into an associated const and/or function on `ABIMachineSpec`? That way it wouldn't have to be passed as a parameter and it'd also be available for assertions if necessary elsewhere? No need really to bend over backwards for this, but I do think it'd be good to have a debug assertion that it's in the right set of caller or callee

cfallin:
Ah, yes, that's a much better idea, thanks! Done.

uweigand:
Hi @cfallin I've had a quick look at this, and it should be relatively straightforward to implement this for s390x as well; I'd be happy to do that (either after this has landed, or else I could provide a patch ahead of time?).

However, I do have one comment about the new `has_non_abi_defs` mechanism.  Not only is this further complicating the `is_included_in_clobbers` mechanism, it also introduces performance regressions.

The original reason for `is_included_in_clobbers` was that for partially callee-saved registers (e.g. on s390x, only the top half of some vector registers, which overlap the old floating-point registers, is callee-saved), in a sequence of A calling B calling C, we have to conservatively assume that the call B->C clobbers them, while the call A->B must preserve them, and therefore B will have to save/restore them all.  Of course, as long as the A->B calling convention is the same as the B->C calling convention, we can avoid this, which is implemented via `is_included_in_clobbers`.

Now, assuming the call B->C needs another def as it has to load a return value into a callee-saved register.  We have to consider this, which is what the new `has_non_abi_defs` mechanism achieves.  However, that will then also have the side effect of once again forcing B to save/restore all floating-point registers - there's really no good reason for this.

Quite a while ago, I made the suggestion that `is_included_in_clobbers` should really only apply to the *clobber* list, not the *def* list.  If we were to make that change, then the extra defs for this case would just happen to work transparently without any new `has_non_abi_defs` mechanism, and it would not introduce the above regression.

cfallin:
@uweigand thanks, that is a great suggestion -- much better than what I did, and I will update as suggested!

fitzgen (approved):
LGTM!

alexcrichton:
A number of fuzz bugs have this in the regression range and I suspect might be related, would you be up for taking a look at these when you're back on Monday?

* https://oss-fuzz.com/testcase-detail/5433312476987392
* https://oss-fuzz.com/testcase-detail/4819793142415360
* https://oss-fuzz.com/testcase-detail/4967284097744896

cfallin:
Yes, just bisected and commented on one of them -- will take a look!

cfallin:
Fuzzbug fix in RA2 (https://github.com/bytecodealliance/regalloc2/pull/214) -- this PR creates more challenging constraints and the bundle-splitting logic had to get a little bit more refined.

PR #10503, opened 2025-04-01
Reuse Wasm linear memories code for GC heaps

### LINKED ISSUE #9350: Reuse linear memories, and our code to emit their bound checks, for Wasm GC heaps

fitzgen:
Instead of bespoke code paths and structures for Wasm GC, this commit makes it so that we now reuse VM structures like `VMMemoryDefinition` and bounds-checking logic. Notably, we also reuse all the associated bounds-checking optimizations and, when possible, virtual-memory techniques to completely elide them.

Furthermore, this commit adds support for growing GC heaps, reusing the machinery for growing memories, and makes it so that GC heaps always start out empty. This allows us to properly delay allocating the GC heap's storage until a GC object is actually allocated.

Fixes #9350


tests/all/pooling_allocator.rs
```diff
@@ -44,10 +44,18 @@ fn memory_limit() -> Result<()> {
...
+            eprintln!("Got error: {msg}");
+            assert!(msg.contains(
+                "memory index 0 is unsupported in this pooling allocator \
+                 configuration"
+            ));
+            assert!(msg.contains(
+                "memory has a minimum byte size of 262144 which exceeds \
+                 the limit of 0x30000 bytes"
+            ));
+        }
```
alexcrichton, line 58: The `ErrorExt` trait I added in a PR yesterday means I think we can avoid this refactoring and/or you can use `assert_contains` instead of manually doing `e:?`

tests/all/main.rs
```diff
@@ -119,6 +123,99 @@ pub(crate) fn gc_store() -> wasmtime::Result<wasmtime::Store<()>> {
     Ok(wasmtime::Store::new(&engine, ()))
 }
 
+pub(crate) fn retry_after_gc<S, T, U>(
```
alexcrichton, line 126: I see these both. here and in `wasmtime-wast`, could they perhaps move to helper methods directly on `StoreContextMut` or similar?

tests/all/limits.rs
```diff
@@ -752,6 +752,7 @@ fn custom_limiter_detect_grow_failure() -> Result<()> {
         return Ok(());
     }
     let mut pool = crate::small_pool_config();
+    pool.total_memories(2); // 1 for the test, 1 for the GC heap.
```
alexcrichton, line 755: Note to myself: somehow I think we need to address this before merging. At minimum this should probably have a FIXME (or somewhere else), but changing all embeddings to reserve two slots for memory when basically no embedding using GC right now feels pretty bad as well.

crates/wasmtime/src/runtime/vm/libcalls.rs
```diff
@@ -829,13 +848,16 @@ unsafe fn array_new_elem(
 
         let array = match ArrayRef::_new_fixed(store, &pre, &vals) {
             Ok(a) => a,
-            Err(e) if e.is::<GcHeapOutOfMemory<()>>() => {
-                // Collect garbage to hopefully free up space, then try the
-                // allocation again.
-                store.maybe_async_gc(None)?;
-                ArrayRef::_new_fixed(store, &pre, &vals)?
+            Err(e) => {
+                if let Some(oom) = e.downcast_ref::<crate::GcHeapOutOfMemory<()>>() {
```
alexcrichton, line 852: Similar to tests, this pattern is 3 times in this file, would it be possible to refactor that to share code amongst the users?

crates/wasmtime/src/runtime/vm/instance/allocator.rs
```diff
@@ -92,7 +92,14 @@ pub struct InstanceAllocationRequest<'a> {
 /// InstanceAllocationRequest, rather than on a &mut InstanceAllocationRequest
 /// itself, because several use-sites require a split mut borrow on the
 /// InstanceAllocationRequest.
-pub struct StorePtr(Option<NonNull<dyn VMStore>>);
+pub struct StorePtr(Option<crate::vm::SendSyncPtr<dyn VMStore>>);
```
alexcrichton, line 95: Mind leaving this as `NonNull` to avoid the cognitive overhead of using `SendSyncPtr`?

crates/wasmtime/src/runtime/vm/gc/gc_runtime.rs
```diff
@@ -356,45 +382,39 @@ pub unsafe trait GcHeap: 'static + Send + Sync {
...
-    /// nothing were allocated in this heap (because nothing is allocated in
-    /// this heap anymore).
-    ///
-    /// This should retain any allocated memory from the global allocator and
-    /// any virtual memory mappings.
-    ///
-    /// This method is only used with the pooling allocator.
-    #[cfg(feature = "pooling-allocator")]
-    fn reset(&mut self);
+    /// TODO FITZGEN
```
alexcrichton, line 387: TODO (and just below)

crates/wasmtime/src/runtime/vm/gc/enabled/free_list.rs
```diff
@@ -28,17 +27,91 @@ impl FreeList {
...
+        } else {
+            None
+        };
+
+        free_list.free_block_index_to_len.extend(entire_range);
+
         free_list
     }
 
+    /// TODO FITZGEN
```
alexcrichton, line 60: TODO

crates/wasmtime/src/runtime/vm/cow.rs
```diff
@@ -406,7 +406,12 @@ impl MemoryImageSlot {
         tunables: &Tunables,
     ) -> Result<()> {
         assert!(!self.dirty);
-        assert!(initial_size_bytes <= self.static_size);
+        assert!(
+            initial_size_bytes <= self.static_size,
+            "initial_size_bytes <= self.static_size failed: \
```
alexcrichton, line 411: I think this line of the error message can be omitted as `assert!` should print out the expression that failed

crates/wasmtime/src/runtime/vm/const_expr.rs
```diff
@@ -269,7 +283,19 @@ impl ConstExprEvaluator {
                     let elem = Val::_from_raw(&mut store, self.pop()?, ty.element_type().unpack());
 
                     let pre = ArrayRefPre::_new(&mut store, ty);
-                    let array = ArrayRef::_new(&mut store, &pre, &elem, len)?;
+                    let array = match ArrayRef::_new(&mut store, &pre, &elem, len) {
```
alexcrichton, line 286: Ditto from libcalls, this is being duplicated enough that it should be centralized somehow to a smaller handful of functions than are in use today

crates/wasmtime/src/runtime/vm.rs
```diff
@@ -200,6 +200,21 @@ pub unsafe trait VMStore {
     /// as a trap to clean up Wasm execution.
     fn maybe_async_gc(&mut self, root: Option<VMGcRef>) -> Result<Option<VMGcRef>>;
 
+    /// TODO FITZGEN
+    fn maybe_async_grow_gc_heap(&mut self, bytes_needed: u64) -> Result<()>;
+
+    /// TODO FITZGEN
+    ///
+    /// TODO FITZGEN: not guaranteed to have space after this
```
alexcrichton, line 208: TODO

alexcrichton: And these should return an error if `bytes_needed` space wasn't actually free'd?

crates/wasmtime/src/runtime/store/async_.rs
```diff
@@ -506,6 +511,22 @@ impl StoreOpaque {
...
+            if self
+                .on_fiber(|store| store.maybe_async_grow_gc_heap(bytes_needed).is_ok())
+                .await
+                .unwrap_or(false)
+            {
+                return;
+            }
+        }
+
+        self.gc_async().await;
```
alexcrichton, line 527: Should this be inverted? First GC then maybe grow if necessary?

crates/wasmtime/src/runtime/store.rs
```diff
@@ -1492,8 +1567,98 @@ impl StoreOpaque {
         self.gc_roots.exit_lifo_scope(self.gc_store.as_mut(), scope);
     }
 
+    /// TODO FITZGEN
+    ///
+    /// FIBER NONSENSE
```
alexcrichton, line 1572: TODO

alexcrichton (commented):
I've gotten through a chunk of this, but if possible I do really think it would be nice to split this apart. I realize it's extra work over what's already here to split this apart but it's pretty hard to review this all-at-once even though I was helping out for at least a chunk of it. If it's too hard it doesn't have to be split but I tend to get nervous when big changes to core runtime bits happen all in one large PR

fitzgen:
Okay, after splitting out all of those PRs, I have this down to

```
51 files changed, 1650 insertions(+), 698 deletions(-)
```

when ignoring the `disas` test expectation updates, which isn't too bad.

Still need to fix the C API and address some of the review comments. Will leave a comment once I've done that and this is ready for re-review.

tests/all/main.rs
fitzgen: This is a little bit of an annoying combinatorial problem:

* we want `async` and regular versions
* we want an `externref` version that plumbs through the external value to the retry, and a non-`externref` version that doesn't have that value
* we want this method on all of `Store`, `StoreContextMut`, and `Caller`

And even after all that, the API is still not very good, because as a caller you don't actually want to think about this stuff, you just want to do `ExternRef::new(...)` and have everything else just work.

We originally didn't automatically trigger GC because that means we would need both `new` and `new_async` methods on `{Any,Extern,...}Ref`, as well as `new_manually_rooted` and `new_manually_rooted_async`. But I'm starting to think that is actually better than trying to expose generic `retry_after_gc` methods. And I also don't think we actually need `new_manually_rooted` variants at all: users can easily do `ExternRef::new(..).to_manually_rooted(..)` if they want to. Then we only need `FooRef::new` and `FooRef::new_async`, which really isn't too bad from a maintenance perspective, and is way better from a user API perspective.

So: what do you think of doing that larger clean up after this lands?

tests/all/limits.rs
fitzgen: Ah, this isn't necessary anymore. Removed.

crates/wasmtime/src/runtime/vm/cow.rs
fitzgen: Unfortunately, if you specify any custom message, then it doesn't show the original expression anymore: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=fb30232a93261b5bfa0bd219eac3ea7c

crates/wasmtime/src/runtime/store/async_.rs
fitzgen: We want to grow first because

1. it is usually cheaper (just updating the accessible region of a mapping; even copying bytes to a new region is also cheaper) and
2. if we swapped the order, then we could hit the following pathological case:
   
   ```rust
   // Assume: the GC heap has capacity for N objects, and we have N-1
   // objects allocated and held alive.

   // Allocate an object, now the GC heap is at capacity. We do not keep the
   // object live, however.
   drop(allocate());

   loop {
       // Upon entry of this loop, the GC heap is at capacity.

       // Allocate another object. This triggers a collect-or-else-grow, the
       // collection reclaims the previously allocated object, and then this
       // allocation succeeds without needing to grow the heap.
       drop(allocate());

       // At the end of this loop, the GC heap is still at capacity.
   }
   ```

   This scenario means we will trigger a GC *on every allocation inside the loop* which is super pathological. We should really only enter into this kind of behavior when there really is no more memory we can give the GC heap (and even then, the embedder probably wants to cancel the execution via async epochs/fuel at some point; may eventually make sense to add a GC limiter style API).

crates/wasmtime/src/runtime/vm/const_expr.rs
fitzgen: See my other comment about folding the retry-after-gc into the `FooRef` constructors.

crates/wasmtime/src/runtime/store/async_.rs
alexcrichton: That makes sense, but isn't this current behavior pathological too? GC heaps will, by default, balloon to 4G before a GC ever runs? 

This makes me think that there's not an ideal answer one way or the other, but picking either extreme seems bad?

cranelift/codegen/src/isa/x64/inst/external.rs
```diff
@@ -131,7 +131,7 @@ fn enc_gpr(gpr: &Gpr) -> u8 {
     if let Some(real) = gpr.to_reg().to_real_reg() {
         real.hw_enc()
     } else {
-        unreachable!()
+        0
```
alexcrichton, line 134: Outdated change now?  (although I understand why this was done I think https://github.com/bytecodealliance/wasmtime/pull/10538 is the "true" fix? or something like that?)

crates/wasmtime/src/runtime/vm/gc/gc_runtime.rs
```diff
@@ -356,45 +382,39 @@ pub unsafe trait GcHeap: 'static + Send + Sync {
...
-    /// This method is only used with the pooling allocator.
-    #[cfg(feature = "pooling-allocator")]
-    fn reset(&mut self);
+    /// Take the underlying memory storage out of this GC heap.
+    fn take_memory(&mut self) -> crate::vm::Memory;
 
-    ////////////////////////////////////////////////////////////////////////////
-    // Accessors for the raw bytes of the GC heap
+    /// Replace this GC heap's underlying memory storage.
+    fn replace_memory(&mut self, memory: crate::vm::Memory, delta_bytes_grown: u64);
```
alexcrichton, line 391: This carries an unsafe requirement that the memory being replaced has to be basically the same as `take_memory`, right? In that it must be at least the same size?

crates/wasmtime/src/runtime/vm.rs
```diff
@@ -202,6 +202,32 @@ pub unsafe trait VMStore {
     /// as a trap to clean up Wasm execution.
     fn maybe_async_gc(&mut self, root: Option<VMGcRef>) -> Result<Option<VMGcRef>>;
 
+    /// Callback invoked whenever an instance needs to grow the GC heap.
+    ///
+    /// Cooperative, async-yielding (if configured) is completely transparent.
+    ///
+    /// If the growth was canceled or fails, returns an error.
+    fn maybe_async_grow_gc_heap(&mut self, bytes_needed: u64) -> Result<()>;
```
alexcrichton, line 210: Would it be possible to avoid the split between either GC or grow? My hope would be that there's a single location which makes a determination of what to do. I might just not be familiar with the code and the method below is The Location, but I feel like the logic of exactly whether to grow or GC is spread out across multiple files at this time.

I understand we probably want a method-per-op at some point but  I would hope that would be buried in some guaranteed-module-private location rather than so prominently exposed to the entire codebase through the `VMStore` trait.

crates/wasmtime/src/runtime/store.rs
```diff
@@ -1518,8 +1584,102 @@ impl StoreOpaque {
...
+        // inside the GC heap.
+        self.unwrap_gc_store_mut()
+            .gc_heap
+            .replace_memory(memory, delta_bytes_grown);
+
+        grow_result
+    }
+
+    #[cfg(feature = "gc")]
+    pub fn grow_or_collect_gc_heap(&mut self, bytes_needed: Option<u64>) {
```
alexcrichton, line 1664: re my comment below: this was my worry, the method here duplicates `maybe_async_grow_or_collect_gc_heap` and I find it confusing to trace what method is used where. I'd prefer ideally if there were a single "do gc" method that everything funnels through rather than having a forest of methods with no narrow waist

crates/wasmtime/src/runtime/vm/const_expr.rs
```diff
@@ -73,7 +73,19 @@ impl<'a> ConstEvalContext<'a> {
...
 
         let allocator = StructRefPre::_new(store, struct_ty);
-        let struct_ref = StructRef::_new(store, &allocator, &fields)?;
+        let struct_ref = match StructRef::_new(store, &allocator, &fields) {
+            Ok(s) => s,
+            Err(e) => {
+                if let Some(oom) = e.downcast_ref::<crate::GcHeapOutOfMemory<()>>() {
+                    store
+                        .traitobj_mut()
+                        .maybe_async_grow_or_collect_gc_heap(oom.bytes_needed())?;
```
alexcrichton, line 82: To double-check, you've confirmed that all locations that evaluate const exprs are on fibers?

crates/wasmtime/src/runtime/gc.rs
```diff
@@ -76,20 +77,49 @@ impl<T> fmt::Debug for GcHeapOutOfMemory<T> {
...
+    /// calls.
+    pub fn take_inner(self) -> (T, GcHeapOutOfMemory<()>) {
+        #[cfg(feature = "gc")]
+        {
+            (self.inner, GcHeapOutOfMemory::new((), self.bytes_needed))
+        }
+        #[cfg(not(feature = "gc"))]
+        {
+            unreachable!()
+        }
```
alexcrichton, line 123: Can you add some comments here for why the `unreachable!()` isn't reachable? It wasn't immediately obvious to me until I saw that the `new` constructor is gated, so presumably it's not possible to call this method when `gc` is disabled, but I think a comment here might be good to have

crates/environ/src/gc.rs
```diff
@@ -367,6 +367,11 @@ pub enum VMGcKind {
     StructRef      = 0b10110 << 27,
 }
 
+/// The size of the `VMGcKind` in bytes.
+pub const VM_GC_KIND_SIZE: u8 = 4;
```
alexcrichton, line 371: Why not define this as `size_of::<VMGcKind>()` to avoid the const assertion?

crates/cranelift/src/bounds_checks.rs
```diff
@@ -31,26 +33,155 @@ use cranelift_frontend::FunctionBuilder;
...
+    /// Like `StaticWholeObject` but with dynamic offset and object size.
+    ///
+    /// It is *your* responsibility to ensure that the `offset + access_size <=
+    /// object_size` precondition holds.
+    #[cfg(feature = "gc")]
+    DynamicWholeObject {
+        offset: ir::Value,
+        object_size: ir::Value,
+    },
+}
```
alexcrichton, line 73: Bikeshedding a bit here, how about `StaticOffset`, `StaticObjectField`, and `DynamicObjectField`? I know you're in a GC-centric mindset right now but making normal linear memory accesses named `BoundsCheck::Field` feels a bit weird to me

crates/cranelift/src/bounds_checks.rs
```diff
@@ -31,26 +33,155 @@ use cranelift_frontend::FunctionBuilder;
...
+///
+/// Prefer `BoundsCheck::*WholeObject` over `BoundsCheck::Field` when possible,
+/// as that approach allows the mid-end to deduplicate bounds checks across
+/// multiple accesses to the same GC object.
+#[derive(Debug)]
+pub enum BoundsCheck {
+    /// Check that this one access in particular is in bounds:
+    ///
+    /// ```ignore
+    /// index + offset + access_size <= gc_heap_bound
```
alexcrichton, line 47: Mind renaming `s/gc_heap_bound/bound/`? (to emphasize this isn't just GC accesses)

crates/cranelift/src/bounds_checks.rs
```diff
@@ -31,26 +33,155 @@ use cranelift_frontend::FunctionBuilder;
...
+            assert_eq!(builder.func.dfg.value_type(object_size), ir::types::I32);
+
+            let index_and_object_size = builder.ins().uadd_overflow_trap(index, object_size, trap);
+            let ptr_just_after_obj = match bounds_check_field_access(
+                builder,
+                env,
+                heap,
+                index_and_object_size,
+                0,
+                0,
```
alexcrichton, line 150: AFAIK this is the first time this code has had to deal with `access_size == 0`, could you do an audit on the below function to ensure that everything works reasonably if the access size is zero?

crates/wasmtime/src/runtime/vm/instance/allocator.rs
fitzgen: Oh huh, I guess I lost this in a rebase or something, because I definitely undid this change at one point

fitzgen:
I'm going to implement the `new` and `new_async` stuff first, before this PR lands, because I am finding a long tail of ref-type allocations in things like doc tests, and fixing all of them is annoying. Would rather just fix the callee than every caller.

fitzgen:
> I'm going to implement the `new` and `new_async` stuff first, before this PR lands, because I am finding a long tail of ref-type allocations in things like doc tests, and fixing all of them is annoying. Would rather just fix the callee than every caller.

This is https://github.com/bytecodealliance/wasmtime/pull/10560

crates/wasmtime/src/runtime/vm/const_expr.rs
fitzgen: Yep :+1:

crates/environ/src/gc.rs
fitzgen: The idea is to do it this way to make sure that it doesn't actually vary depending on the target

crates/cranelift/src/bounds_checks.rs
fitzgen: looks good to me

crates/wasmtime/src/runtime/store/async_.rs
fitzgen: I think growing to the limit before doing GC is actually what you want: you don't *want* to do GC if you can avoid it. Growing the heap is faster than doing a GC. There are knobs for limiting the max size of memories, including the GC heap, so unbounded growth isn't really a concern (also there is a hard bound of 4 GiB since the GC heap is a 32-bit memory). And we are likely going to throw away the whole store, and the GC heap along with it, before we ever need to collect, and doing that is also much faster than doing a collection.

alexcrichton (approved):

PR #10504, opened 2025-04-01
Cranelift: update to regalloc2 v0.11.2.

cfallin:
This incorporates a fix for a panic caused by too many live registers that is not properly returned as a regalloc error. This condition should not be reachable from Cranelift (which should not generate constraints that trigger this condition), and was seen only during development where different constraints were added; but this PR is an update out of an abundance of caution.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2794,9 +2794,9 @@ dependencies = [
 
 [[package]]
 name = "regalloc2"
-version = "0.11.1"
+version = "0.11.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "145c1c267e14f20fb0f88aa76a1c5ffec42d592c1d28b3cd9148ae35916158d3"
+checksum = "dc06e6b318142614e4a48bc725abbf08ff166694835c43c9dae5a9009704639a"
 dependencies = [
  "allocator-api2",
  "bumpalo",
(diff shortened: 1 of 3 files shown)


alexcrichton (approved):
r=me, but this'll need a `cargo vet` too

cfallin:
Weirdly, the cargo vet job itself [succeeded](https://github.com/bytecodealliance/wasmtime/actions/runs/14203976402/job/39797257774?pr=10504); it's only the "Cargo vet failed on a Pull Request" job that failed?

Locally, `cargo vet` on this branch finds no issues either...

alexcrichton:
Oh the vet [actually failed](https://github.com/bytecodealliance/wasmtime/actions/runs/14203976402/job/39797257774?pr=10504) but it's [allowed to do so](https://github.com/bytecodealliance/wasmtime/blob/238ec908d485997a2b148fd4099971e62830c66b/.github/workflows/main.yml#L138) to avoid blocking PRs entering the merge queue on the CI check. That way for external contributors we can push vets separately and then merge a PR without requesting a rebase

alexcrichton:
Er, lemme try again with that [first link](https://github.com/bytecodealliance/wasmtime/actions/runs/14203976402/job/39797257774?pr=10504#step:5:1)

alexcrichton:
Ok thanks github actions, anyway if you expand the `cargo vet --locked` step you'll see the failure

cfallin:
Ah, that's a bit counterintuitive; thanks, updated.

PR #10506, opened 2025-04-01
Rename `VMTableImport` to `VMTable`

abrown:
In #10499, I used the `struct VMTableImport` to hold the necessary pointers for accessing a core WebAssembly table from a component. Since `VMTableImport` no longer was exclusively used for imported tables, it seems reasonable to rename it to something less specific: `VMTable`. This changes all the functions referencing `vmtable_import*` as well except those that are specifically dealing with imported tables.


fitzgen (approved):
SGTM

Issue #8773, opened 2024-06-11
Support guest profiling for components


yoshuawuyts:
#### Feature

Enable the following to work:

```sh
wasmtime serve ./my-component.wasm --profile=guest
```

#### Benefit

In general: profiling is useful and it's neat to be able to profile wasm components.

For me specifically though: I'm on MacOS and setting up native trace tools requires turning off security features in the kernel. I've not been able to do so successfully, so as a result I (somewhat embarrassingly) don't currently have a working `perf` command on my machine. Because wasmtime's guest profiler doesn't interact with the kernel at all, this seems like the ideal way to resolve it, basically by providing a portable, OS-agnostic profiling suite.

#### Implementation

There is a working guest profiler available already, but trying to run it on components gives the following error message:

```text
Error: Cannot use the guest profiler with components
```

I don't really have insight in what it would take to implement support for components. But I figured it would be desirable - and I didn't see it being tracked already, so I figured we should. Folks with more knowledge about the challenges here should probably chime in though.

posborne:
I'm starting to look at this now; in my brief test I'm not seeing the same error related to components but I'm also not seeing results that look correct which seems to be backed up by the code only registering directly present core modules rather than those used as part of components.

sunfishcode:
This is implemented in #10507.

Issue #7669, opened 2023-12-09
Annotate guest profiler stack frames with module info


jameysharp:
#### Feature

So far usually a guest has consisted of only one core WebAssembly module, but with the component model on its way, that's changing. For each stack frame in a profile, the guest profiler should indicate which module that frame's function came from. If the same module is instantiated multiple times, ideally the profile would also report which instance was running.

#### Benefit

In multi-module guests, seeing only a function name is confusing. Including the module identity will help the user understand which part of their code is running.

#### Implementation

I'm not sure how to identify a module in a human-meaningful way. I'm not sure if modules have useful names inside components, or how to find those names if they exist. Similarly I don't think instances have an identity that we can find while stack-walking or that means anything to the user; I believe this is an open issue for our core-dump support as well.

#### Alternatives

We could continue to make the user guess which module each frame is from. In many cases probably the important function names are unique across modules anyway. For the moment, while multi-module guests remain unusual, the _status quo_ is probably fine.

posborne:
As part of the work for #8773 I think we'll end up getting this as well; I'll need to go through some refining for how I'm going about this, but it generally seems to work.  This is with components and for this first pass I'm looking at each of the modules with text include pc and then finding the first one that has a func including pc which appears to work well but, I think, could be wrong in cases of code sharing.

![Image](https://github.com/user-attachments/assets/69cb7646-a66d-450f-afa1-84256394d4dd)

sunfishcode:
This is now implemented in #10507. @posborne mentioned to me that there are ways we could improve it further, but we can file separate issues as needed.

PR #10507, opened 2025-04-01
Guest Profiling suport for component model

posborne:
This change adds support for using guest profiling with the component model.  In addition to the core change to support attributing stack frames to constituent modules within a component, the cli `run` and `serve` commands are also updated to support using the `--profile=guest` cli option with components.

https://github.com/bytecodealliance/wasmtime/issues/8773
https://github.com/bytecodealliance/wasmtime/issues/7669

### DIFF
--- a/crates/wasmtime/src/runtime/profiling.rs
+++ b/crates/wasmtime/src/runtime/profiling.rs
@@ -1,7 +1,10 @@
+#[cfg(feature = "component-model")]
+use crate::component::Component;
 use crate::instantiate::CompiledModule;
 use crate::prelude::*;
 use crate::runtime::vm::Backtrace;
 use crate::{AsContext, CallHook, Module};
+use core::cmp::Ordering;
 use fxprof_processed_profile::debugid::DebugId;
 use fxprof_processed_profile::{
     CategoryHandle, Frame, FrameFlags, FrameInfo, LibraryInfo, MarkerLocation, MarkerSchema,
@@ -78,7 +81,14 @@ pub struct GuestProfiler {
     start: Instant,
 }
 
-type Modules = Vec<(Range<usize>, fxprof_processed_profile::LibraryHandle)>;
+#[derive(Debug)]
+struct ProfiledModule {
+    module: Module,
+    fxprof_libhandle: fxprof_processed_profile::LibraryHandle,
+    text_range: Range<usize>,
+}
+
+type Modules = Vec<ProfiledModule>;
 
 impl GuestProfiler {
     /// Begin profiling a new guest. When this function is called, the current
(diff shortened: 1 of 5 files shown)


posborne:
Here's a couple samples from some simple components for the CLI and wasi-http.  There's a couple frames missing from the CLI example so I think I am missing something from the component that isn't part of `static_modules()` individual functions which I'll try to find.

I'm also not at all confident on the API which would be best for all use cases for `GuestProfiler` in terms of how users are likely to want to register various components and modules.  Here we support 1 component + multiple additional modules which was biased by the potential registrations in the `run` and `serve` commands with prelinking but definitely welcome ideas.  I considered having a `ProfileTarget` enum or similar to abstract over things -- some approaches are a bit more challenging due to conditional compilation concerns.

[wasmtime-run-profile-data.json](https://github.com/user-attachments/files/19558979/wasmtime-run-profile-data.json)
[wasmtime-serve-profile-data.json](https://github.com/user-attachments/files/19558980/wasmtime-serve-profile-data.json)

image

crates/wasmtime/src/runtime/profiling.rs
```diff
@@ -96,21 +106,42 @@ impl GuestProfiler {
...
+    pub fn new(
+        module_name: &str,
+        interval: Duration,
+        modules: impl IntoIterator<Item = (String, Module)>,
+    ) -> Self {
         let zero = ReferenceTimestamp::from_millis_since_unix_epoch(0.0);
         let mut profile = Profile::new(module_name, zero, interval.into());
 
+        // past this point, we just need to think about modules as we pull out
+        // the disparate module information from components from this point forward.
```
fitzgen, line 118: Nitpick for here and elsewhere: we generally try to capitalize sentences and such in comments.

crates/wasmtime/src/runtime/profiling.rs
```diff
@@ -96,21 +106,42 @@ impl GuestProfiler {
...
         let mut modules: Vec<_> = modules
             .into_iter()
             .filter_map(|(name, module)| {
                 let compiled = module.compiled_module();
-                let text = compiled.text().as_ptr_range();
-                let address_range = text.start as usize..text.end as usize;
-                module_symbols(name, compiled).map(|lib| (address_range, profile.add_lib(lib)))
+                let text_range = {
+                    // assumption, within text, the code for a given module is packed linarly and
```
fitzgen, line 124: "linarly"

crates/wasmtime/src/runtime/profiling.rs
```diff
@@ -125,6 +156,26 @@ impl GuestProfiler {
...
+    /// See [`GuestProfiler::new`] for additional information; this function
+    /// works identically except that it takes a component and sets up
+    /// instrumentation to track calls in each of its constituent modules.
+    #[cfg(feature = "component-model")]
+    pub fn new_component(
+        component_name: &str,
+        interval: Duration,
+        component: Component,
+        extra_modules: impl IntoIterator<Item = (String, Module)>,
+    ) -> Self {
```
fitzgen, line 170: It probably makes sense to use a builder-style API in the limit here, to allow users to specify all the various components and modules they are profiling, but this is totally fine for the moment.

fitzgen (commented):
LGTM modulo a couple nitpicks :+1:

crates/wasmtime/src/runtime/profiling.rs
FrankReh: Super nits: Is "past this point, ... from this point forward" redundant? Maybe I should understand the context better. Also, while here, I'll point out the title has "suport" in it, probably a typo.

FrankReh:
Drive-by comment:

I cloned the branch and was successful in creating a wasmtime-guest-profile.json with it that profiler.firefox.com accepted. The flame graph aspect of that UI didn't seem to work but presumably that has nothing to do with what this PR is accomplishing.

Thanks for making this feature possible!

posborne:
@FrankReh thanks for doing a check.  In my testing, I was getting the flamegraph to work fine for both `run` and `serve`; what was your test setup?  Usually when I didn't get flame graphs in the UI it was due to some part of the sampling not working correctly (but I can't discount something else going sideways unrelated to the profile itself).

Does the call tree view work?

FrankReh:
My bad. Even the flamegraph aspect is working. Embarrassing but I didn't recognize the pane as a valid flamegraph simply because of the grey color and my test case having very little happening besides two function calls writing to stdout repeatedly. In my defense, first time running profiling for a wasm module and first time using profiler.firefox.com web app.

Thanks again for working to get this functionality added.

crates/wasmtime/src/runtime/profiling.rs
posborne: Yeah, I noticed that as well when I was making updates.  Should be addressed with the fixup.

posborne:
My fixups got autosquashed when I rebased to resolve a trivial conflict in `run.rs` but updates were all very minor:
- Couple small updates to hopefully make CI happy with dead_code conditional comp and type annotations from c-api code.
- Updates to comments as suggested (plus a few in other parts of the patch to match the convention.

For now, keeping the API as-is to keep the collateral for this change smaller and (mostly) not require updates for users of the API (may require very minor updates to type hint).

fitzgen (approved):
Thanks!

PR #10508, opened 2025-04-01
Analyze whether a module requires a GC heap during its compilation

fitzgen:
And avoid forcing the allocation of a GC heap for modules that will not use it.

Split out from https://github.com/bytecodealliance/wasmtime/issues/9350

### DIFF
--- a/crates/wasmtime/src/runtime/instance.rs
+++ b/crates/wasmtime/src/runtime/instance.rs
@@ -259,7 +259,7 @@ impl Instance {
         store.bump_resource_counts(module)?;
 
         // Allocate the GC heap, if necessary.
-        if store.engine().features().gc_types() {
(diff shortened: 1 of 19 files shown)


crates/wasmtime/src/runtime/instance.rs
```diff
@@ -259,7 +259,7 @@ impl Instance {
         store.bump_resource_counts(module)?;
 
         // Allocate the GC heap, if necessary.
-        if store.engine().features().gc_types() {
+        if store.engine().features().gc_types() && module.env_module().needs_gc_heap {
```
alexcrichton, line 262: I think here we might be able to entirely remove the first part of the condition?

alexcrichton (approved):
Thanks for splitting this out!

Issue #4986, opened 2022-09-30
Always use mmap instead of posix_memalign in cranelift-jit


bjorn3:
This currently behind the selinux-fix feature flag. SELinux doesn't allow making parts of the heap executable. In addition it seems that there is an issue in multithreaded environments causing permission issues leading to crashes when using posix_memalign instead of mmap. (see https://github.com/bytecodealliance/wasmtime/issues/4980) I don't know why using posix_memalign was chosen initially. Maybe there is some advantage to it that I don't know of. If not, I don't see any reason to stay with posix_memalign as default. @sunfishcode do you remember why you used posix_memalign?

sunfishcode:
It's `std::alloc::alloc` now, rather than `posix_memalign`, but the same question applies.

I think the reason was that `alloc`/`posix_memalign` can be faster, because it can reuse free memory that the allocator has, rather than always requiring an mmap system call to allocate and a munmap system call to free each time. However, I don't have any data on whether it's actually faster.

bjorn3:
Maybe we could make a simple memory allocator ourself? Unlike system allocators it doesn't have to deal with tiny allocations or allocations of fixed sizes, so complexity from that doesn't have to exist and it could be made aware of which page has which memory protection applied so that once a page is made executable it will only consider it allocatable once every function stored in it has been deallocated. This way it is also no longer necessary to allocate multiples of the page size.

bjorn3:
Another advantage of that would be that it allows keeping all code within 2GB even when allocating a lot of memory in between compiling functions. This is necessary to avoid the panics at https://github.com/bytecodealliance/wasmtime/blob/c1d6ca48a74ab6f356d08baa533b90b17ccb834a/cranelift/jit/src/compiled_blob.rs#L60 and elsewhere during applying of relocations.

Issue #4000, opened 2022-04-06
Cranelift: JIT relocations depend on system allocator behaviour


Mrmaxmeier:
Hey,

I'm seeing crashes during `finalize_definitions` calls related to x86\_64 `call` relocations:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: TryFromIntError(())', cranelift/jit/src/compiled_blob.rs:55:80
```

Cranelift [emits 32-bit relocations for calls on x86\_64](https://github.com/bytecodealliance/wasmtime/blob/d147802d513db045406c043740a77e0556d1dfdb/cranelift/codegen/src/isa/x64/inst/emit.rs#L1196), and thus can "only" address in the relative ±2GB range. Code memory is [allocated](https://github.com/bytecodealliance/wasmtime/blob/d147802d513db045406c043740a77e0556d1dfdb/cranelift/jit/src/memory.rs#L55) with the normal system allocator, which might place different allocations in distant parts of the address space.
I'm seeing irregular crashes in a heavily multithreaded program, but the problem can be reproduced with this abridged `jit-minimal.rs` example:

```rust
use cranelift::prelude::*;
use cranelift_codegen::settings;
use cranelift_jit::{JITBuilder, JITModule};
use cranelift_module::{default_libcall_names, Linkage, Module};

fn main() {
    let isa_builder = cranelift_native::builder().unwrap();
    let isa = isa_builder
        .finish(settings::Flags::new(settings::builder()))
        .unwrap();

    let mut m = JITModule::new(JITBuilder::with_isa(isa, default_libcall_names()));

    let mut ctx = m.make_context();
    let mut func_ctx = FunctionBuilderContext::new();

    let func_a = m
        .declare_function("a", Linkage::Local, &m.make_signature())
        .unwrap();
    let func_b = m
        .declare_function("b", Linkage::Local, &m.make_signature())
        .unwrap();

    // Define a dummy function `func_a`
    ctx.func.name = ExternalName::user(0, func_a.as_u32());
    {
        let mut bcx: FunctionBuilder = FunctionBuilder::new(&mut ctx.func, &mut func_ctx);
        let block = bcx.create_block();

        bcx.switch_to_block(block);
        bcx.ins().return_(&[]);
        bcx.seal_all_blocks();
        bcx.finalize();
    }
    m.define_function(func_a, &mut ctx).unwrap();
    m.clear_context(&mut ctx);

[...]

PR #10512, opened 2025-04-02
cranelift-jit: add bump memory allocator, make configurable

Mrmaxmeier:
Hey,
this is a stab at #4000 (and #4986) to resolve an issue with x86 relocation range that comes up in multi-threaded JITs.
I've added an option to swap out the memory allocator in `cranelift-jit` and have added an arena-based option that works with a pre-allocated `PROT_NONE` region.

I've marked this as draft because I could also see updating the PR to either
- just add the trait, have users provide their own option when in doubt?
- update the existing allocator to allow reserving an area up-front?
- split the changes into multiple commits if helpful

Let me know if this seems reasonable. Thanks! :slightly_smiling_face:

### DIFF
--- a/cranelift/jit/src/memory/arena.rs
+++ b/cranelift/jit/src/memory/arena.rs
@@ -178,6 +178,9 @@ impl ArenaMemoryProvider {
         for segment in &mut self.segments {
             segment.finalize(branch_protection);
         }
+
+        // Flush any in-flight instructions from the pipeline
+        wasmtime_jit_icache_coherence::pipeline_flush_mt().expect("Failed pipeline flush");
     }
 
     /// Frees the allocated memory region, which would be leaked otherwise.
--- a/cranelift/jit/src/memory/arena.rs
+++ b/cranelift/jit/src/memory/arena.rs
@@ -22,6 +22,9 @@ struct Segment {
 
 impl Segment {
     fn new(ptr: *mut u8, len: usize, target_prot: region::Protection) -> Self {
+        // Segments are created on page boundaries.
+        debug_assert_eq!(ptr as usize % region::page::size(), 0);
+        debug_assert_eq!(len % region::page::size(), 0);
         let mut segment = Segment {
             ptr,
             len,
@@ -117,14 +120,14 @@ impl ArenaMemoryProvider {
         protection: region::Protection,
     ) -> io::Result<*mut u8> {
         let align = usize::try_from(align).expect("alignment too big");
-        debug_assert!(
+        assert!(
             align <= region::page::size(),
             "alignment over page size is not supported"
         );
 
         // Note: Add a fast path without a linear scan over segments here?
 
-        // Can we fit this allocation into an existing segment.
+        // Can we fit this allocation into an existing segment?
         if let Some(segment) = self.segments.iter_mut().find(|seg| {
             seg.target_prot == protection && !seg.finalized && seg.has_space_for(size, align)
         }) {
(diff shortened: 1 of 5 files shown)


cranelift/jit/src/memory/arena.rs
```diff
@@ -0,0 +1,215 @@
...
+        if self.target_prot == region::Protection::READ_EXECUTE {
+            super::set_readable_and_executable(self.ptr, self.len, branch_protection)
+                .expect("unable to set memory protection for jit memory segment");
+        } else {
+            unsafe {
+                region::protect(self.ptr, self.len, self.target_prot)
+                    .expect("unable to change memory protection for jit memory segment");
+            }
+        }
+        self.finalized = true;
```
bjorn3, line 58: After finalization, `has_space_for` should unconditionally return `false` at least when this is an executable segment as any attempt to write to it will fail.

cranelift/jit/src/memory/arena.rs
```diff
@@ -0,0 +1,215 @@
...
+impl Segment {
+    fn new(ptr: *mut u8, len: usize, target_prot: region::Protection) -> Self {
+        let mut segment = Segment {
+            ptr,
+            len,
+            target_prot,
+            position: 0,
+            finalized: false,
+        };
+        // set setgment to read-write for initialization
```
bjorn3, line 32: *segment

cranelift/jit/src/memory/arena.rs
```diff
@@ -0,0 +1,218 @@
...
+                if self.position + additional_size <= self.size {
+                    segment.len += additional_size;
+                    segment.set_rw();
+                    self.position += additional_size;
+                    return Ok(segment.allocate(size, align));
+                }
+            }
+        }
+
+        // allocate new segment for given size and alignment
```
abrown, line 140: Nit: most comments throughout the project follow the convention of capitalizing sentences and ending them with a period.

abrown (approved):
I think this overall makes sense to me; I didn't look too, too closely at all the arithmetic so it would be useful to have some tests around this (if there aren't any already). Plus, @bjorn3 are you a +1 on this? I believe you may be the one most interested in this. (I checked and apparently there are some other [dependents](https://crates.io/crates/cranelift-jit/reverse_dependencies) of this so it isn't _just_ @bjorn3).

Mrmaxmeier:
> I think this overall makes sense to me; I didn't look too, too closely at all the arithmetic so it would be useful to have some tests around this (if there aren't any already). Plus, @bjorn3 are you a +1 on this? I believe you may be the one most interested in this. (I checked and apparently there are some other [dependents](https://crates.io/crates/cranelift-jit/reverse_dependencies) of this so it isn't _just_ @bjorn3).

I've added some simple tests and have cleaned up the PR a bit.
Bjorn3's sub-1hr review is just luck of the draw.  :slightly_smiling_face:

cranelift/jit/src/memory/system.rs
```diff
@@ -175,56 +165,16 @@ impl Memory {
...
-                        anyhow::Error::new(e).context("unable to make memory readable+executable"),
-                    )
-                })?;
-            }
-            Ok(())
-        };
-
         for &PtrLen { ptr, len, .. } in self.non_protected_allocations_iter() {
-            set_region_readable_and_executable(ptr, len)?;
+            super::set_readable_and_executable(ptr, len, branch_protection)?;
```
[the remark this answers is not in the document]
Mrmaxmeier: Yes, seems reasonable to avoid the extra calls.  I've moved the pipeline flush call back outside of this function :+1:

bjorn3 (approved):

Mrmaxmeier:
Whoops, sorry. Didn't realize that only a subset of the CI jobs runs on posted PRs ^^

Issue #10460, opened 2025-03-22
Winch x64 simd: disagreement with cranelift trampolines about ABI


alexcrichton:
This input:

```wasm
(module
  (func (export "x")
    (param i32 v128 i32)
    (param v128 i32 v128 i32 v128)
    (param i32 v128 i32 v128 i32)
    (param v128 i32 v128 i32)
    (param $last v128)
    (result v128)
    local.get $last
  )
)

(assert_return
  (invoke "x"
    (i32.const 0)
    (v128.const i64x2 0 0)
    (i32.const 0)
    (v128.const i64x2 0 0)
    (i32.const 0)
    (v128.const i64x2 0 0)
    (i32.const 0)
    (v128.const i64x2 0 0)
    (i32.const 0)
    (v128.const i64x2 0 0)
    (i32.const 0)
    (v128.const i64x2 0 0)
    (i32.const 0)
    (v128.const i64x2 0 0)
    (i32.const 0)
    (v128.const i64x2 0 0)
    (i32.const 0)
    (v128.const i64x2 0 0)
  )
  (v128.const i64x2 0 0))

```

yields:

```
$ cargo run -q wast -Ccompiler=winch repro.wat
Error: failed to run script file 'repro.wat'

Caused by:
    0: failed directive on repro.wat:13:1
    1: result 0 didn't match
    2: expected [                   0,                    0]
       actual   [      94750456038144,                    0]

       expected (hex) [0000000000000000, 0000000000000000]
       actual (hex)   [0000562ccf464300, 0000000000000000]
```

cc @jeffcharles @saulecabrera

saulecabrera:
Thanks for the report! We'll take a look into it.

jeffcharles:
I wasn't able to resolve this today but I narrowed down the repro case to:

```wat
(module
  (func (export "x")
    (param f32 f32 f32 f32 f32 f32 f32 f32 f32) (param $last v128)
    (result v128)
    local.get $last
  )
)

(assert_return
  (invoke "x"
    (f32.const 1)
    (f32.const 2)
    (f32.const 3)
    (f32.const 4)
    (f32.const 5)
    (f32.const 6)
    (f32.const 7)
    (f32.const 8)
    (f32.const 9)
    (v128.const i64x2 10 11)
  )
  (v128.const i64x2 10 11))
```

[...]

PR #10513, opened 2025-04-02
winch(x64): Fix stack parameter alignment

saulecabrera:
Closes https://github.com/bytecodealliance/wasmtime/issues/10460

This commit ensures stack arguments greater than 8 bytes are correctly aligned.

Prior to this commit this wasn't an issue since for the default calling convention, all arguments were aligned to a fixed slot size (e.g, 8 bytes). With the introduction of vector types, namely, v128, we must ensure that arguments greater than 8 bytes in size are correctly type-sized aligned.


fitzgen (approved):
Thanks!

PR #10515, opened 2025-04-02
cranelift: Restore 16-bit TrapCodes

Mrmaxmeier:
This allows users of Cranelift to have ~65k unique trap codes again.
The range of `TrapCode` was reduced in #9338 in order to be able to fit arbitrary trap codes inside of `MemFlags`.

I'm keeping MemFlags a u16 with this change and chose to only allow a _low subset_ of trap codes to be represented instead. Explicit traps can still use the full TrapCode range.

The implementation is a bit awkward but I've added exhaustive tests for the encoding part. Feel free to close this PR if the cons (panic in `MemFlags::with_trap_code`) outweigh the pros here (more flexibility with dynamically-assigned user traps). :slightly_smiling_face: 

Thanks!

### DIFF
--- a/cranelift/assembler-x64/src/api.rs
+++ b/cranelift/assembler-x64/src/api.rs
@@ -2,7 +2,7 @@
 
 use crate::gpr;
 use crate::xmm;
-use std::{num::NonZeroU8, ops::Index, vec::Vec};
+use std::{num::NonZeroU16, ops::Index, vec::Vec};
 
 /// Describe how an instruction is emitted into a code buffer.
 pub trait CodeSink {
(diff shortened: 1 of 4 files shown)


cfallin:
Thanks for the PR!

Looking it over, it indeed does what you say, but it's not clear to me that it is a clean, usable, improvement to the API to have only a subset of trap codes usable in some places. E.g., we may want the ability to turn a `trapif` into a virtual memory-based trap (which would use a load with `MemFlags`); that transform/optimization would no longer be possible with this distinction.

If the change were to make trap codes, say, 32 or 64 bits, I could understand the utility better, but u16 is still small enough that one has to worry about running out if one is allocating these codes dynamically. Could you clarify your use-case and requirements? Maybe there's another way to encode the information that you're encoding in the trap code?

Mrmaxmeier:
Ah yes, it doesn't seem worth it considering the possible optimization you mentioned.

In my specific use case I'm using Cranelift in a custom JIT that generates code with a fair amount of rare "error" cases. In my case the amount of unique errors roughly matches the amount of basic blocks in a source function.

I'm handling these through signals and Cranelift's ability to resolve the trapping instruction back to an IR-level trap code is quite useful here. Of course I could also call into a function that raises a passed code from the generated assembly, but I'm not sure if there's a way that is as _code_-space efficient.

PR #10516, opened 2025-04-02
Support adding additional capacity to `FreeList`s

fitzgen:
Split out from https://github.com/bytecodealliance/wasmtime/pull/10503

### DIFF
--- a/crates/wasmtime/src/runtime/vm/gc/enabled/free_list.rs
+++ b/crates/wasmtime/src/runtime/vm/gc/enabled/free_list.rs
@@ -5,6 +5,20 @@ use core::{alloc::Layout, num::NonZeroU32, ops::Bound};
 /// A very simple first-fit free list for use by our garbage collectors.
 pub(crate) struct FreeList {
     /// The total capacity of the contiguous range of memory we are managing.
+    ///
+    /// NB: we keep `self.capacity` unrounded because otherwise we would get
+    /// rounding errors where we lose track of the actual capacity we have when
+    /// repeatedly adding capacity `n` where `n < ALIGN`:
+    ///
+    ///     let mut free_list = FreeList::new(0);
+    ///     loop {
+    ///         free_list.add_capacity(1);
+    ///     }
+    ///
+    /// If we eagerly rounded capacity down to our alignment on every call to
+    /// `add_capacity`, the free list would always think it has zero capacity,
+    /// even though it would have enough capacity for many allocations after
+    /// enough iterations of the loop.
     capacity: usize,
     /// Our free blocks, as a map from index to length of the free block at that
     /// index.
@@ -46,10 +60,14 @@ impl FreeList {
             self.capacity
         );
 
+        // See the comment on `self.capacity` about why we need to do the
+        // alignment-rounding here, rather than keeping `self.capacity` aligned
+        // at rest.
+        let old_cap_rounded = round_usize_down_to_pow2(old_cap, ALIGN_USIZE);
+
         // If we are adding capacity beyond what a `u32` can address, then we
         // can't actually use that capacity, so don't bother adding a new block
         // to the free list.
-        let old_cap_rounded = round_usize_down_to_pow2(old_cap, ALIGN_USIZE);
         let Ok(old_cap_rounded) = u32::try_from(old_cap_rounded) else {
             return;
         };
--- a/crates/wasmtime/src/runtime/vm/gc/enabled/free_list.rs
+++ b/crates/wasmtime/src/runtime/vm/gc/enabled/free_list.rs
@@ -1,6 +1,5 @@
 use crate::prelude::*;
 use alloc::collections::BTreeMap;
-use core::cmp;
 use core::{alloc::Layout, num::NonZeroU32, ops::Bound};
 
 /// A very simple first-fit free list for use by our garbage collectors.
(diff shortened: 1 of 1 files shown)


crates/wasmtime/src/runtime/vm/gc/enabled/free_list.rs
```diff
@@ -28,17 +27,92 @@ impl FreeList {
...
             free_block_index_to_len: BTreeMap::new(),
         };
-        free_list.reset();
+
+        let end = u32::try_from(free_list.capacity).unwrap_or_else(|_| {
+            assert!(free_list.capacity > usize::try_from(u32::MAX).unwrap());
+            u32::MAX
+        });
+
+        // Don't start at `0`. Reserve that for "null pointers" and free_list way we
```
FrankReh, line 42: "and free_list way we" I was not able to parse.

crates/wasmtime/src/runtime/vm/gc/enabled/free_list.rs
```diff
@@ -28,17 +27,92 @@ impl FreeList {
...
+            None
+        };
+
+        free_list.free_block_index_to_len.extend(entire_range);
+
         free_list
     }
 
+    /// Add additional capacity to this free list.
+    #[allow(dead_code)] // TODO: becomes used in https://github.com/bytecodealliance/wasmtime/pull/10503
```
alexcrichton, line 40: Could the tests below exercise this new method?

crates/wasmtime/src/runtime/vm/gc/enabled/free_list.rs
```diff
@@ -47,21 +121,11 @@ impl FreeList {
...
-        if layout.size() > self.max_size() {
-            let trap = crate::Trap::AllocationTooLarge;
-            let err = anyhow::Error::from(trap);
-            let err = err.context(format!(
-                "requested allocation's size of {} is greater than the max supported size of {}",
-                layout.size(),
-                self.max_size(),
-            ));
-            return Err(err);
-        }
```
alexcrichton, line 63: To confirm, there's still a test below testing this behavior and that it fails?

crates/wasmtime/src/runtime/vm/gc/enabled/free_list.rs
fitzgen: I just reversed these changes, so that it still calls `reset`, since `reset` can't be removed until the rest of the changes from #10503 are present. Will make sure I don't accidentally remove words from the comment somehow when eventually inlining what used to be `reset` in that PR.

crates/wasmtime/src/runtime/vm/gc/enabled/free_list.rs
```diff
@@ -37,8 +36,62 @@ impl FreeList {
...
+        self.capacity = self.capacity.saturating_add(additional);
+        log::debug!(
+            "FreeList::add_capacity({additional:#x}): capacity growing from {old_cap:#x} to {:#x}",
+            self.capacity
+        );
+
+        // If we are adding capacity beyond what a `u32` can address, then we
+        // can't actually use that capacity, so don't bother adding a new block
+        // to the free list.
+        let old_cap_rounded = round_usize_down_to_pow2(old_cap, ALIGN_USIZE);
```
alexcrichton, line 52: If allocations could previously go up to `old_cap`, shouldn't this round up instead of round down?

crates/wasmtime/src/runtime/vm/gc/enabled/free_list.rs
fitzgen: Correct, we just return an `Ok(None)` (meaning that the allocation might succeed after growth/gc) instead of `Err(_)` (meaning that the allocation will never succeed no matter what)

crates/wasmtime/src/runtime/vm/gc/enabled/free_list.rs
fitzgen: No, because the old capacity was also rounded down the the alignment (all allocations are also a multiple of the alignment) so the first newly allocatable index for our additional capacity is that old capacity rounded down.

crates/wasmtime/src/runtime/vm/gc/enabled/free_list.rs
fitzgen: Done

crates/wasmtime/src/runtime/vm/gc/enabled/free_list.rs
alexcrichton: Should this assert that old_cap is already aligned in that case? It mostly just seems counter-inutitive to me that you'd round down the end to figure out the index of the next block, I'd expect that you'd either round up or assert rounding isn't necessary

fitzgen: If we keep `self.capacity` aligned down to our alignment, then a sequence of 100 `add_capacity(1)` calls will never actually result in the `FreeList` thinking it has capacity for an allocation, even though it actually should. That is, `self.capacity = round_down_to_align(self.capacity + additional)` will always result in `self.capacity ==0` when `self.capacity` is initially `0` and `additional` is `1`, even when it is called in a loop.

We have to keep track of the actual capacity, and just do the rounding when adding blocks into the free list.

alexcrichton: Ah ok yeah that makes sense, I was missing that subtelty. Mind adding some comments to that effect to this?

alexcrichton (approved):

crates/wasmtime/src/runtime/vm/gc/enabled/free_list.rs
fitzgen: Added

Issue #4311, opened 2022-06-23
Implement support in the component model for memory64


alexcrichton:
I'm going to split this work item from https://github.com/bytecodealliance/wasmtime/issues/4185 since I suspect it's going to be somewhat significant. The intention here is to track issues related to the memory64 proposal for WebAssembly and the component model. At the time of this writing neither of these proposals are "stable". Additionally the component model does not actually define what the canonical ABI is with respect to the memory64 proposal. Fully implementing this item will be a blend of "do the obvious thing" where possible such as making pointers 64-bits large, while at the same time also identifying question that don't have "obvious" answers such as "what's the maximum length of a string or list?"

Currently there are a lot of TODO items sprinkled throughout `typed.rs` which note various places that I suspect will need memory64 treatment. Fully implementing this issue, however, will likely require fuzzer integration because otherwise it'll be difficult to ensure that we've covered all cases.

Some examples of things that need updating are:

* Need to call `realloc` with a different signature
* All reads/writes of pointers to linear memory need to have a 32/64 bit switch
* Reading and writing pointers to `ValRaw` needs to be audited
* Size and alignment calculations all need to be duplicated or have a parameter for 64 bit 
* Somehow should probably confirm that this doesn't vastly affect performance for 32-bit

The last point here is an open question I don't know how best to answer.

PR #10518, opened 2025-04-02
Plumb initial translation of `thread.spawn_indirect`

abrown:
This change in no way implements actual spawning of threads; it simply translates the CM builtin `thread.spawn_indirect` to invoke a real Wasmtime libcall that does some preliminary checks and panics. It is useful, though, to (1) exercise all the new table-related functions and (2) show the limits of Wasmtime's current state: we will need a way of interacting with shared functions and possibly some form of shared store.

### DIFF
--- a/crates/cranelift/src/compiler/component.rs
+++ b/crates/cranelift/src/compiler/component.rs
@@ -279,6 +279,11 @@ impl<'a> TrampolineCompiler<'a> {
                     rets[0] = me.raise_if_negative_one_and_truncate(rets[0]);
                 })
             }
+            Trampoline::ThreadSpawnIndirect { ty: _, table } => {
+                // TODO: eventually pass through the `ty` argument to check the
+                // table's funcref signature.
+                self.translate_thread_spawn_indirect(*table)
+            }
         }
     }
 
@@ -1326,6 +1331,26 @@ impl<'a> TrampolineCompiler<'a> {
             vmctx,
             host::error_context_drop,
             &callee_args,
+            TrapSentinel::Falsy,
+        );
+    }
+
+    fn translate_thread_spawn_indirect(&mut self, table: RuntimeTableIndex) {
+        let args = self.builder.func.dfg.block_params(self.block0).to_vec();
+        let vmctx = args[0];
+        let element = args[1];
+        let context = args[2];
+
+        // table: u32
+        let table = self
+            .builder
+            .ins()
+            .iconst(ir::types::I32, i64::from(self.offsets.runtime_table(table)));
+
+        self.translate_intrinsic_libcall(
+            vmctx,
+            host::thread_spawn_indirect,
+            &[vmctx, table, element, context],
             TrapSentinel::Falsy,
         );
     }
--- a/crates/cranelift/src/compiler/component.rs
+++ b/crates/cranelift/src/compiler/component.rs
@@ -1345,7 +1345,7 @@ impl<'a> TrampolineCompiler<'a> {
         let table = self
             .builder
             .ins()
-            .iconst(ir::types::I32, i64::from(self.offsets.runtime_table(table)));
+            .iconst(ir::types::I32, i64::from(table.as_u32()));
 
         self.translate_intrinsic_libcall(
             vmctx,
--- a/crates/cranelift/src/compiler/component.rs
+++ b/crates/cranelift/src/compiler/component.rs
@@ -279,10 +279,8 @@ impl<'a> TrampolineCompiler<'a> {
                     rets[0] = me.raise_if_negative_one_and_truncate(rets[0]);
                 })
             }
-            Trampoline::ThreadSpawnIndirect { ty: _, table } => {
-                // TODO: eventually pass through the `ty` argument to check the
-                // table's funcref signature.
-                self.translate_thread_spawn_indirect(*table)
+            Trampoline::ThreadSpawnIndirect { ty, table } => {
+                self.translate_thread_spawn_indirect(*ty, *table)
             }
         }
     }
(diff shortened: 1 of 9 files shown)


crates/cranelift/src/compiler/component.rs
```diff
@@ -280,6 +280,11 @@ impl<'a> TrampolineCompiler<'a> {
                     rets[0] = me.raise_if_negative_one_and_truncate(rets[0]);
                 })
             }
+            Trampoline::ThreadSpawnIndirect { ty: _, table } => {
+                // TODO: eventually pass through the `ty` argument to check the
+                // table's funcref signature.
+                self.translate_thread_spawn_indirect(*table)
```
abrown, line 286: @alexcrichton, I feel like there is some way to actually materialize this `ty` index into a real type so we can check that the `funcref` has the expected type of the thread start function but I don't see it.

crates/environ/src/component.rs
```diff
@@ -154,6 +154,9 @@ macro_rules! foreach_builtin_component_function {
             #[cfg(feature = "component-model-async")]
             error_context_transfer(vmctx: vmctx, src_idx: u32, src_table: u32, dst_table: u32) -> u64;
 
+            #[cfg(feature = "threads")]
+            thread_spawn_indirect(vmctx: vmctx, table: u32, element: u32, context: u32 ) -> u64;
```
abrown, line 158: For 64-bit WebAssembly, we probably want `element` (and probably even `context`) to be a `u64`; how are we handling the 32-bit/64-bit distinction elsewhere?

crates/wasmtime/src/runtime/vm/component/libcalls.rs
```diff
@@ -1267,3 +1267,15 @@ unsafe fn error_context_drop(
...
+
+#[cfg(feature = "threads")]
+unsafe fn thread_spawn_indirect(
+    vmctx: NonNull<VMComponentContext>,
+    table: u32,
+    element: u32,
+    context: u32,
+) -> Result<u32> {
+    ComponentInstance::from_vmctx(vmctx, |instance| {
+        instance.thread_spawn_indirect(table, element, context)
```
abrown, line 1279: It's unclear to me which parts should live up at this level and which should live in `ComponentInstance::thread_spawn_indirect`. Any suggestions?

crates/cranelift/src/compiler/component.rs
alexcrichton: `TypeFuncIndex` is fine yeah, and the way to pass it to runtime would basically be to embed the constant in the generated code itself. Runtime would then be responsible for translating `TypeFuncIndex` into a `VMSharedTypeIndex` which can be done through various translation tables similar to how core wasm does it

crates/wasmtime/src/runtime/vm/component/libcalls.rs
alexcrichton: Stylistic an up-to-you,  but generally "simple things" and "ABI things" in this file and meat elsewhere. So e.g. converting `u32` to more typed things here, but leaving the meat of the implementation elsewhere is fine.

crates/cranelift/src/compiler/component.rs
```diff
@@ -1331,6 +1336,26 @@ impl<'a> TrampolineCompiler<'a> {
...
+        let args = self.builder.func.dfg.block_params(self.block0).to_vec();
+        let vmctx = args[0];
+        let element = args[1];
+        let context = args[2];
+
+        // table: u32
+        let table = self
+            .builder
+            .ins()
+            .iconst(ir::types::I32, i64::from(self.offsets.runtime_table(table)));
```
alexcrichton, line 1349: This is a disconnect between runtime and compile-time where here the u32 is the offset in the VMContext but at runtime it's interpreted as `RuntimeTableIndex`

crates/environ/src/component.rs
alexcrichton: Currently it's basically not-handled, Wasmtime [doesn't support 64-bit components at all](https://github.com/bytecodealliance/wasmtime/issues/4311)

crates/wasmtime/src/runtime/vm/component.rs
```diff
@@ -745,6 +745,75 @@ impl ComponentInstance {
...
+        let VMTable { vmctx, from } = self.runtime_table(RuntimeTableIndex::from_u32(table));
+        let element = u64::from(element);
+        let table = unsafe {
+            Instance::from_vmctx(vmctx.as_non_null(), |handle| {
+                let idx = handle.table_index(from.as_non_null().as_ref());
+                handle
+                    .get_defined_table_with_lazy_init(idx, iter::once(element))
+                    .as_ref()
+            })
+        };
```
alexcrichton, line 772: FWIW technically some and/or most of this is going to go away. Lazy initialization is probably not going to be possible with shared tables.

alexcrichton (commented):
This seems reasonable enough to me with some tweaks, but I do think we'll want a test for this. I don't think there's any viable path to testing the runtime bits at this time so I'd prefer to leave the implementation body of the intrinsic as `todo!()` for now, but testing at least the compile-time/Cranelift parts seems reasonable with a test that just compiles something but does nothing with the result.

crates/cranelift/src/compiler/component.rs
abrown: Ok, but actually the index being used is coming from a core function type. E.g., in the tests, for `(core type $start (shared (func (param $context i32))))`, we use `$start` as the index passed through to this canonical built-in. I see how to retrieve a `TypeFunc` using `ComponentTypes` but I don't think that's what we actually want here &mdash; that would be a component type. Is `TypeFuncIndex` the right index to be using here? How do we get to its core type (presumably backed by `VMSharedTypeIndex`)?

crates/wasmtime/src/runtime/vm/component.rs
abrown: Yup. What I'm trying to do is plumb through enough to (a) show how this is currently possible using unshared tables (while still immediately failing) and (b) create a use site here so that future refactoring considers this usage. I was thinking that tests could even check for the initial behavior that _is_ functional today: i.e., we can't run a shared function, but at least we can check that using the wrong index results in a failure.

crates/cranelift/src/compiler/component.rs
alexcrichton: Agreed yeah we don't want to deal with `TypeFunc` here as the type-check should be a single comparison ideally, via `VMShareTypeIndex`. Converting from `TypeFuncIndex` will require a table of some form stored in `VMComponentContext` which is filled in on initialization. That happens for core modules, for example, where basically `VMContext` has a table from `TypeFuncIndex` to `VMSharedTypeIndex`

crates/wasmtime/src/runtime/vm/component.rs
alexcrichton: I don't think it's possible to hit this at runtime though? All components would fail validation unless they use shared tables, and shared talbes aren't implemented in Wasmtime, so it's not possible to instantiate any component using this intrinsic?

PR #10524, opened 2025-04-04
Cranelift: simplify some side-effectful instructions in ISLE

fitzgen:
This commit adds a new top-level ISLE entrypoint specifically for instructions in the side-effectful skeleton: `simplify_skeleton`. While these rewrites are processed during the egraph pass, values from skeleton instructions still do not get inserted into the egraph. Indeed, `simplify_skeleton` operates on *instructions* rather than *values* because we do not represent side effects as values; values do not have side effects in CLIF, instructions do. Therefore, rather than doing a whole dynamic-programming style extraction of the best candidate simplification like we do with the egraph, we take an eager and greedy approach.

Furthermore, `simplify_skeleton` is limited only to skeleton instructions that do not involve control-flow or terminators right now. This is because changing the control-flow graph can change whether a use is dominated by a def or not, and we do not currently have the machinery to track and fix up invalidated uses. Addressing this is left for future commits.

### DIFF
--- a/cranelift/filetests/filetests/egraph/skeleton.clif
+++ b/cranelift/filetests/filetests/egraph/skeleton.clif
@@ -89,3 +89,62 @@ block0(v0: i32):
     return v2
 }
 ; check: return v0
+
+;;;;;;;;;; Tests for `simplify_skeleton` TODOs ;;;;;;;;;;;;
+;;
+;; What follows are tests for patterns that `simplify_skeleton` *should* clean
+;; up, but is unable to at this moment in time.
+
+function %int_min_sdiv_neg_one() -> i32 {
+block0():
+    v0 = iconst.i32 0x80000000
+    v1 = iconst.i32 -1
+    v2 = sdiv v0, v1
+    return v2
+}
+; check:  v2 = sdiv v0, v1
+; nextln: return v2
+
+function %sdiv_zero(i32) -> i32 {
+block0(v0: i32):
+    v1 = iconst.i32 0
+    v2 = sdiv v0, v1
+    return v2
+}
+; check:  v2 = sdiv v0, v1
+; nextln: return v2
+
+function %udiv_zero(i32) -> i32 {
+block0(v0: i32):
+    v1 = iconst.i32 0
+    v2 = udiv v0, v1
+    return v2
+}
+; check:  v2 = udiv v0, v1
+; nextln: return v2
+
+function %always_trapping_trapz() -> i32 {
+block0:
+    v0 = iconst.i32 0
+    trapz v0, user42
+    return v0
+}
+; check: trapz v0, user42
+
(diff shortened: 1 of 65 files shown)


fitzgen:
FWIW, I just did a sightglass run, and this had no statistically significant effect on any of our default benchmarks. If you look at the disas test expectation updates however, you’ll see that it is a big boon for gc codegen.

cranelift/filetests/filetests/egraph/skeleton.clif
```diff
@@ -0,0 +1,91 @@
...
+}
+; check: return v0
+
+function %sdiv_by_one(i32) -> i32 {
+block0(v0: i32):
+    v1 = iconst.i32 1
+    v2 = sdiv v0, v1
+    return v2
+}
+; check: return v0
```
cfallin, line 91: Can we add a test for `INT_MIN / -1` and show that it remains un-cprop'd?

(It would be even neater if it could change into an unconditional trap but I know that's a limitation at the moment)

cfallin (approved):
Looks great -- very cool to see the GC tests clean up with the new opts in particular!

I think I mentioned it last week but just to write it down: I suspect a good way to lift some of the CFG-mutation limitations around traps would be to make `trap` a non-terminator, and to add `unreachable` as a terminator. This would allow us to turn insts that we can prove always trap into `trap` (e.g. the `sdiv` case mentioned below, as well as trapnz-of-1 and trapz-of-0). The branch-folding cases are a bit harder; ideally we do actually skip traversing the newly-unreachable blocks, perhaps by tombstoning them with a flag set by the branch simplification. Would you mind creating a followup issue for this?

fitzgen:
> I think I mentioned it last week but just to write it down: I suspect a good way to lift some of the CFG-mutation limitations around traps would be to make `trap` a non-terminator, and to add `unreachable` as a terminator.

Yes, we could even avoid a new `unreachable` instruction if we wanted by adding a `terminator: bool` to `trap`, if we wanted, since we should have the bytes available in the `InstructionData`.

Or if we do want two different instructions, it is probably easier to make the new instruction the non-terminator, so that existing uses of `trap` don't need updating.

I had also floated the idea of `trap` returning a variable number of values, that we could use as dummy values in now-unreachable uses, just as a convenience and to make the branch-folding a little simpler to deal with (rather than requiring tracking reachability perfectly). This could be an intermediate step since it would allow us to avoid perfectly tracking and fixing up unreachable blocks, but keeping dominating defs for any now-unreachable blocks' uses.

> The branch-folding cases are a bit harder; ideally we do actually skip traversing the newly-unreachable blocks, perhaps by tombstoning them with a flag set by the branch simplification. Would you mind creating a followup issue for this?

We also would need to differentiate between blocks that do vs don't become unreachable, which is a further little wrinkle.

But yeah, happy to write up an issue.

Issue #6038, opened 2023-03-17
cranelift: ISLE wrapper for constructing constants


jameysharp:
#### Feature

In our mid-end optimization rules we sometimes need to construct new constants, where the constant's type comes from somewhere in the left-hand side pattern. We should add a helper term that constructs constants without having to know whether we need to use `iconst`, `vconst`, `f32const`, `f64const`, …

#### Benefit

We can write more general optimizations without needing a lot of special cases.

#### Implementation

I'm pretty sure this isn't quite right, but I'm thinking of an ISLE term something along these lines:

```lisp
(decl pure const (Type u64) Value)
(rule (const $I128 n) (uextend $I128 (iconst $I64 n))
(rule (const (ty_int ty) n) (iconst ty n))
(rule (const (ty_vec ty) n) (splat ty (const (lane_type ty) n)))
```

I don't know whether this should go in `prelude-opt.isle` or `prelude.isle`. It depends on whether it's useful in lowering as well as optimization. I think it probably isn't useful in lowering and should go in the optimization-specific prelude.

Thanks to @fitzgen for the general idea.

PR #10528, opened 2025-04-05
cranelift: ISLE wrapper for constructing constants

Gonzalosilvalde:
## Description:

This pull request addresses issue [#6038](https://github.com/bytecodealliance/wasmtime/issues/6038) by adding an ISLE wrapper for constructing constants. The goal is to allow mid-end optimization rules to generate constants generically without needing to know whether to use `iconst`, `vconst`, `f32const`, `f64const`, etc.

## Changes:

- Added a helper function `ty_vec` in `cranelift/codegen/src/isle_prelude.rs`
- Defined an extractor `ty_vec` in `cranelift/codegen/src/prelude.isle
- Introduced a new rule in `cranelift/codegen/src/prelude_opt.isle` for constructing constant nodes

## Additional Notes:

- It compiles without any problem and passes the `cranelift` test.

### DIFF
--- a/cranelift/codegen/src/isle_prelude.rs
+++ b/cranelift/codegen/src/isle_prelude.rs
@@ -529,6 +529,15 @@ macro_rules! isle_common_prelude_methods {
         }
 
         #[inline]
+        fn ty_vec(&mut self, ty: Type) -> Option<Type> {
+            if ty.is_vector() {
+                Some(ty)
+            } else {
+                None
+            }
+        }
+
+        #[inline]
         fn ty_vec64_int(&mut self, ty: Type) -> Option<Type> {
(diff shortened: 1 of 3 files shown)


cfallin:
@Gonzalosilvalde thanks for the PR!

In order to ensure this is tested and works for our lowerings, would you mind switching over at least a few existing uses of `iconst`/etc to use this instead? Looking at the issue(s) that spawned this might help to find cases to demonstrate it as well.

efJerryYang:
Hi @Gonzalosilvalde , I noticed this PR has been quiet for a while.

Would you mind if I pick this up and continue the remaining testing work in a separate PR to help move it forward?

Gonzalosilvalde:
Hi @efJerryYang ,
Yeah, continue what remains without any problem

PR #10537, opened 2025-04-07
Run CI tests through AddressSanitizer

alexcrichton:
This is similar to running tests in Valgrind (which we should perhaps also do...) but can be useful for catching use-after-free style bugs faster than when a process crashes. Given the unsafe nature of Wasmtime this is something we should have probably enabled awhile back but otherwise so long as it doesn't take too long to run on CI seems like an easy win of a boost-of-confidence.

prtest:asan

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -769,6 +769,14 @@ jobs:
         touch ${{ runner.tool_cache }}/qemu/built
       if: matrix.qemu != ''
 
+    - name: Configure ASAN
+      run: |
+        echo CARGO_PROFILE_DEV_OPT_LEVEL=2 >> $GITHUB_ENV
+        echo CARGO_PROFILE_TEST_OPT_LEVEL=2 >> $GITHUB_ENV
+        echo RUSTFLAGS=-Zsanitizer=address >> $GITHUB_ENV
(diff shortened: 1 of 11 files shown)


alexcrichton:
Looks like un-optimized this takes [over an hour in CI](https://github.com/bytecodealliance/wasmtime/actions/runs/14313411386/job/40113786700).

Trying optimized next.

alexcrichton:
Ok 25 minutes much more reasonable, I might look into sharding this test suite though like we do all the others

fitzgen:
> This is similar to running tests in Valgrind (which we should perhaps also do...) but can be useful for catching use-after-free style bugs faster than when a process crashes. Given the unsafe nature of Wasmtime this is something we should have probably enabled awhile back but otherwise so long as it doesn't take too long to run on CI seems like an easy win of a boost-of-confidence.

A note for anyone following along at home: this comment isn't *quite* as scary as it might otherwise seem, since we already run as many tests under [MIRI](https://github.com/rust-lang/miri) (an undefined behavior-detecting interpreter for Rust's MIR internal representation) as we can, which gives us many of the same (although not fully overlapping) benefits that asan and valgrind give. That said, more coverage with more tools is always welcome! See also https://bytecodealliance.org/articles/security-and-correctness-in-wasmtime for more details on our various approaches to security and correctness in Wasmtime.

dicej (approved):

PR #10540, opened 2025-04-07
Apply argument extensions to component intrinsics too

alexcrichton:
This commit fixes the definition of host signatures for component builtins to work in the same way that core builtins do which is to use platform-specific sign-extension-rules by default. This fixes a regression found in the wasip3-prototyping repository where new component intrinsics didn't have this set and subsequently were failing in riscv64 CI.


fitzgen (approved):
Is it possible to whip up a disas test for this?

PR #10542, opened 2025-04-07
Simplify handling of caller-vmctx and `dyn VMStore` pointers

fitzgen:
Just keep a copy of the trait object pointer in `StoreOpaque`, rather than trying to pluck it out of the default caller.

Clean up `CallThreadState` so that it doesn't need a caller-vmctx.

Split out from https://github.com/bytecodealliance/wasmtime/pull/10503

### DIFF
--- a/crates/wasmtime/src/runtime/store.rs
+++ b/crates/wasmtime/src/runtime/store.rs
@@ -584,6 +584,10 @@ impl<T> Store<T> {
             data: ManuallyDrop::new(data),
         });
 
+        // Note the erasure of the lifetime here into `'static`, so in general
+        // usage of this trait object must be strictly bounded to the `Store`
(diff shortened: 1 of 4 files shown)


crates/wasmtime/src/runtime/store.rs
```diff
@@ -605,19 +620,10 @@ impl<T> Store<T> {
                     })
                     .expect("failed to allocate default callee")
             };
-
-            // Note the erasure of the lifetime here into `'static`, so in
-            // general usage of this trait object must be strictly bounded to
-            // the `Store` itself, and this is an invariant that we have to
-            // maintain throughout Wasmtime.
```
alexcrichton, line 612: Can this comment be preserved?

alexcrichton (approved):

PR #10547, opened 2025-04-08
Beef up documentation about TLS management in Wasmtime

alexcrichton:
This comes out of discussions on the wasip3-prototyping repository where I was trying to re-learn what this all does. The goal here is to more exhaustively document how this module works.

### DIFF
--- a/crates/wasmtime/src/runtime/vm/traphandlers.rs
+++ b/crates/wasmtime/src/runtime/vm/traphandlers.rs
@@ -425,8 +425,27 @@ mod call_thread_state {
     use super::*;
     use crate::runtime::vm::Unwind;
 
-    /// Temporary state stored on the stack which is registered in the `tls` module
-    /// below for calls into wasm.
+    /// Temporary state stored on the stack which is registered in the `tls`
+    /// module below for calls into wasm.
+    ///
+    /// This structure is stored on the stack and allocated during the
+    /// `catch_traps` function above. The purpose of this structure is to track
+    /// the state of an "activation" or a sequence of 0-or-more contiguous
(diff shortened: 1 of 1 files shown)


crates/wasmtime/src/runtime/vm/traphandlers.rs
```diff
@@ -734,11 +772,84 @@ impl CallThreadState {
...
-// A private inner module for managing the TLS state that we require across
-// calls in wasm. The WebAssembly code is called from C++ and then a trap may
-// happen which requires us to read some contextual state to figure out what to
-// do with the trap. This `tls` module is used to persist that information from
-// the caller to the trap site.
+/// A private inner module managing the state of Wasmtime's thread-local storage
+/// (TLS) state.
+///
+/// Wasmtime at this time has a single pointer of TLS. This single pointer of
+/// TLS is the totality of all TLS required by Wasmtime. By keeping this as
```
dicej, line 779: FYI, this will no longer be true once CM async support is merged given that `concurrent.rs` has a couple of TLS variables (and `wasmtime-wit-bindgen` may also generate code which includes zero or more TLS variables).

dicej (approved):
Very helpful docs and diagram; thanks for doing this!

crates/wasmtime/src/runtime/vm/traphandlers.rs
alexcrichton: Hm ok yeah that's going to need to be carefully integrated, "just add a TLS var" breaks portability guarantees we have, so it's not as simple as just relaxing this documentation. We'll have to handle that at some point.

PR #10548, opened 2025-04-08
Define an RAII helper for generic take-and-replace borrow splitting

fitzgen:
Follow up to https://github.com/bytecodealliance/wasmtime/pull/10524#discussion_r2032164529

### DIFF
--- a/cranelift/codegen/src/take_and_replace.rs
+++ b/cranelift/codegen/src/take_and_replace.rs
@@ -1,0 +1,91 @@
+//! Helper for temporarily taking values out and then putting them back in.
+
+/// An RAII type to temporarily take a `U` out of a `T` and then put it back
+/// again on drop.
+///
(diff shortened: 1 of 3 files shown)


cranelift/codegen/src/take_and_replace.rs
```diff
@@ -0,0 +1,91 @@
...
+}
+
+impl<T, U, F> TakeAndReplace<T, U, F>
+where
+    F: Fn(&mut T) -> &mut U,
+    U: Default,
+{
+    /// Create a new `TakeAndReplace` that temporarily takes out
+    /// `proj(container)`.
+    pub fn new(mut container: T, proj: F) -> Self {
```
alexcrichton, line 78: Perhaps `container: &mut T` here to emphasize that this isn't intended to take `T` by ownership? (above it's only instantiated as `&mut Self` which means it should still compile I think)

PR #10553, opened 2025-04-08
Update wasm spec test suite, add exception feature flags

alexcrichton:
This commit performs an update of the spec test suite submodule to the next-to-latest commit. The latest commit will require updating the `wast` dependency which isn't published yet.

This update brings in the `wasm-3.0` folder of tests since it's been awhile since the last update. That update notably means that the exception-handling proposal is mixed in with all the others with various tests. Getting tests as flagged as passing or failing as a result was unexpectedly difficult.

The solution I ended up settling on was to preemptively implement some infrastructure for the exceptions proposal:

* `wasmtime::Config` methods
* `wasmtime` CLI flags
* integration with wast testing
* various updates to `should_fail`

It turns out we can run a few tests with the exception proposal, notably due to tags being implemented for stack switching. That meant that this couldn't blanket ignore the exceptions proposal and say it's expected to fail. Instead the proposal is said "this passes!" and tests are individually listed as "this is expected to fail".

This then required changing an `unsupported!` panic to plumbing errors around to avoid actually implementing the exceptions proposal here.

### DIFF
--- a/crates/cli-flags/src/lib.rs
+++ b/crates/cli-flags/src/lib.rs
@@ -382,7 +382,7 @@ wasmtime_option_group! {
         pub extended_const: Option<bool>,
         /// Configure support for the exceptions proposal.
         pub exceptions: Option<bool>,
-        /// Configure support for the legacy exceptions proposal.
+        /// DEPRECATED: Configure support for the legacy exceptions proposal.
         pub legacy_exceptions: Option<bool>,
     }
 
@@ -991,6 +991,7 @@ impl CommonOptions {
             config.wasm_exceptions(enable);
         }
         if let Some(enable) = self.wasm.legacy_exceptions.or(all) {
+            #[expect(deprecated, reason = "forwarding CLI flag")]
             config.wasm_legacy_exceptions(enable);
         }
 
--- a/crates/cli-flags/src/lib.rs
+++ b/crates/cli-flags/src/lib.rs
@@ -380,6 +380,10 @@ wasmtime_option_group! {
         pub wide_arithmetic: Option<bool>,
         /// Configure support for the extended-const proposal.
         pub extended_const: Option<bool>,
+        /// Configure support for the exceptions proposal.
+        pub exceptions: Option<bool>,
+        /// Configure support for the legacy exceptions proposal.
+        pub legacy_exceptions: Option<bool>,
     }
 
     enum Wasm {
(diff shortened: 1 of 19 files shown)


alexcrichton:
cc @cfallin here on how this touches on the exceptions proposal

crates/cli-flags/src/lib.rs
```diff
@@ -380,6 +380,10 @@ wasmtime_option_group! {
         pub wide_arithmetic: Option<bool>,
         /// Configure support for the extended-const proposal.
         pub extended_const: Option<bool>,
+        /// Configure support for the exceptions proposal.
+        pub exceptions: Option<bool>,
+        /// Configure support for the legacy exceptions proposal.
+        pub legacy_exceptions: Option<bool>,
```
fitzgen, line 386: This should be `doc(hidden)` and `clap(hidden)` or whatever as well, same as the `Config::method`.

crates/wasmtime/src/config.rs
```diff
@@ -1151,6 +1145,18 @@ impl Config {
...
+    pub fn wasm_exceptions(&mut self, enable: bool) -> &mut Self {
+        self.wasm_feature(WasmFeatures::EXCEPTIONS, enable);
+        self
+    }
+
+    #[doc(hidden)] // FIXME(#3427) - if/when implemented then un-hide this
+    pub fn wasm_legacy_exceptions(&mut self, enable: bool) -> &mut Self {
+        self.wasm_feature(WasmFeatures::LEGACY_EXCEPTIONS, enable);
+        self
+    }
```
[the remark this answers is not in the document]
fitzgen: Can we also mark it deprecated something like this:

```rust
#[deprecated = "This configuration option only exists for internal \
                usage with the spec testsuite. It may be removed at \
                any time and without warning. Do not rely on it!"]
```

and then in our couple of usages just do `#[allow(deprecated)]` on the use

fitzgen (approved):
Yeah we shouldn't ever implement the legacy exceptions on principle, but this knob makes sense for the `.wast` test suite.

crates/cli-flags/src/lib.rs
alexcrichton: We unfortunately don't have that here, but I can put "DEPRECATED" in the doc at least!

PR #10554, opened 2025-04-08
Cranelift/Wasmtime: disable fastalloc (single-pass) allocator for now.

cfallin:
Unfortunately, as discovered by a recent fuzzbug [1], the single-pass register allocator is not compatible with the approach to callsite defs that exception-handling support has forced us to take. In particular, we needed to move all call return-value defs onto the call instruction itself, so calls could be terminators; this unbounded number of defs is made to be a solvable allocation problem by using `any` constraints, which allow allocation directly into spillslots; but fastalloc appears to error out if it runs out of registers, regardless of this constraint.

Long-term, we should fix this, but unfortunately I don't have cycles to dive into fastalloc's internals at the moment, and it's (I think) a tier-3 feature. As such, this PR disables its use for now. I've filed a tracking issue in RA2 [2], and referenced this in the Cranelift configuration option docs at least.

To keep from shifting all fuzzbugs / fuzzing corpii by altering the `arbitrary` interpretation, I opted to keep the enum the same in the fuzzing crate, and remap `SinglePass` to `Backtracking` there. I'm happy to take the other approach and remove the option (thus invalidating all fuzzbugs) if we'd prefer that instead.

[1]: https://oss-fuzz.com/testcase-detail/5433312476987392
[2]: https://github.com/bytecodealliance/regalloc2/issues/217

### DIFF
--- a/cranelift/codegen/meta/src/shared/settings.rs
+++ b/cranelift/codegen/meta/src/shared/settings.rs
@@ -38,10 +38,12 @@ pub(crate) fn define() -> SettingGroup {
 
             - `backtracking`: A backtracking allocator with range splitting; more expensive
                               but generates better code.
-            - `single_pass`: A single-pass algorithm that yields quick compilation but
-                             results in code with more register spills and moves.
+
+            Note that the `single_pass` option is currently disabled because it does not
+            have adequate support for the kinds of allocations required by exception
+            handling (https://github.com/bytecodealliance/regalloc2/issues/217).
         "#,
-        vec!["backtracking", "single_pass"],
+        vec!["backtracking"],
     );
 
     settings.add_enum(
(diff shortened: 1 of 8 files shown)


fitzgen (approved):

Robbepop:
@cfallin are there plans to re-introduce `Singlepass` register allocator back into Wasmtime? I am using Wasmtime in differential fuzzing in Wasmi and was using this register allocator due to very annoying time-out issues during fuzzing.

cfallin:
@Robbepop the issue at bytecodealliance/regalloc2#217 describes the issue that is preventing us from re-enabling it. As of right now, I have no time to investigate or fix this; I am oversubscribed on too many other things that are higher-priority and that is unlikely to change for the rest of the year. If you or anyone else are willing to dig in and fix this, it would be very much appreciated!

PR #10557, opened 2025-04-09
s390x: Add try_call / try_call_indirect support

uweigand:
Following-up on https://github.com/bytecodealliance/wasmtime/pull/10510 this adds the corresponding support to the s390x target.

Due to the separate ABI implementation, this currently needs to duplicate some of the underlying logic.  I hope to be able to refactor ABI handling in the future to be able to share more of that code across all targets.

FYI @cfallin


cfallin (approved):
Looks great -- thanks for putting this together so quickly!

PR #10560, opened 2025-04-09
Automatically trigger GC in `{Array,Extern,Struct}Ref` allocation functions

fitzgen:
Rather than forcing all callers to check for `GcHeapOutOfMemory`, trigger a GC, and then try again. This does force us to define `*_async` variations for when async is enabled, however; it's ultimately worth it.

### DIFF
--- a/crates/wasmtime/src/runtime/gc/enabled/arrayref.rs
+++ b/crates/wasmtime/src/runtime/gc/enabled/arrayref.rs
@@ -307,11 +307,6 @@ impl ArrayRef {
         elem: &Val,
         len: u32,
     ) -> Result<Rooted<ArrayRef>> {
-        assert!(
-            !store.async_support(),
-            " use `ArrayRef::new_async` with asynchronous stores"
-        );
-
         store.retry_after_gc((), |store, ()| {
             Self::new_from_iter(store, allocator, RepeatN(elem, len))
         })
@@ -366,11 +361,6 @@ impl ArrayRef {
         elem: &Val,
         len: u32,
     ) -> Result<Rooted<ArrayRef>> {
-        assert!(
-            store.async_support(),
-            "use `ArrayRef::new` with synchronous stores"
-        );
-
         store
             .retry_after_gc_async((), |store, ()| {
                 Self::new_from_iter(store, allocator, RepeatN(elem, len))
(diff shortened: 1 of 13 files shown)


crates/wasmtime/src/runtime/gc/enabled/arrayref.rs
```diff
@@ -271,40 +311,88 @@ impl ArrayRef {
...
+        store: &mut StoreOpaque,
+        allocator: &ArrayRefPre,
+        elem: &Val,
+        len: u32,
+    ) -> Result<Rooted<ArrayRef>> {
+        assert_eq!(
+            store.id(),
+            allocator.store_id,
+            "attempted to use a `ArrayRefPre` with the wrong store"
+        );
```
alexcrichton, line 381: Could this assertion get folded into `new_unchecked`?

alexcrichton (approved):

crates/wasmtime/src/runtime/gc/enabled/arrayref.rs
```diff
@@ -385,14 +498,121 @@ impl ArrayRef {
             allocator.store_id,
             "attempted to use a `ArrayRefPre` with the wrong store"
         );
+        assert!(
+            !store.async_support(),
+            "use `ArrayRef::new_fixed_async` with asynchronous stores"
+        );
```
alexcrichton, line 504: Could this get sunk into `retry_after_gc`? (and the dual into `retry_after_gc_async`?)

crates/wasmtime/src/runtime/gc/enabled/arrayref.rs
```diff
@@ -385,14 +498,121 @@ impl ArrayRef {
...
+
+        // Type check the elements against the element type.
+        for elem in elems {
+            elem.ensure_matches_ty(store, allocator.ty.element_type().unpack())
+                .context("element type mismatch")?;
+        }
```
alexcrichton, line 510: Could this get sunk into `_new_unchecked`? (making it a bit less unchecked)

Basically trying to get these multiple entrypoints as a one-liner-ish around the "guts" and the same "guts" are used everywhere

alexcrichton (approved):

crates/wasmtime/src/runtime/gc/enabled/arrayref.rs
fitzgen: Not without either making the panic message worse for users or plumbing through extra arguments to the retry functions that are just for the panic message. Neither seems worth the trade off.

alexcrichton: Personally I think it's fine to make the panic messages worse given the predicted rarity they'll come up, but I'm ok either way.

PR #10561, opened 2025-04-09
Winch: Free dst register if computing heap address fails

jeffcharles:
Fixes a bug where a destination register is allocated and not freed when performing a vector load lane from a memory address that is invalid at compilation time. Not freeing the destination register causes an internal compiler error later when emitting code for any result value that uses that register.


saulecabrera (commented):
Looks good to me, thanks! Would you mind adding a wast test as well under `tests/misc_testsuite/winch`?

saulecabrera (approved):

PR #10566, opened 2025-04-11
c-api: Compile a component

MangoPeachGrape:
First little bit of #9812, ["compile a component"](https://github.com/bytecodealliance/wasmtime/pull/9812#pullrequestreview-2503060369), with some changes suggested in the original pr:
- [No longer depends on cranelift](https://github.com/bytecodealliance/wasmtime/pull/9812#discussion_r1884430669), feature gated compile function on `WASMTIME_FEATURE_COMPILER`.
- [Moved config switch to config.h](https://github.com/bytecodealliance/wasmtime/pull/9812#discussion_r1884432876).
- [Prepared the file structure for multiple header files](https://github.com/bytecodealliance/wasmtime/pull/9812#discussion_r1884460573), should the same be done for the rust side?

Based on the efforts of @lanfeust69 and @rockwotj.

### DIFF
--- a/crates/c-api/Cargo.toml
+++ b/crates/c-api/Cargo.toml
@@ -59,5 +59,6 @@ winch = ['wasmtime/winch']
 debug-builtins = ['wasmtime/debug-builtins']
 wat = ['dep:wat', 'wasmtime/wat']
 pooling-allocator = ["wasmtime/pooling-allocator"]
+component-model = ["wasmtime/component-model"]
 # ... if you add a line above this be sure to change the other locations
 # marked WASMTIME_FEATURE_LIST
--- a/crates/c-api/artifact/Cargo.toml
+++ b/crates/c-api/artifact/Cargo.toml
@@ -40,6 +40,7 @@ default = [
   'winch',
   'debug-builtins',
   'pooling-allocator',
+  'component-model',
   # ... if you add a line above this be sure to change the other locations
   # marked WASMTIME_FEATURE_LIST
 ]
@@ -62,5 +63,6 @@ cranelift = ["wasmtime-c-api/cranelift"]
 winch = ["wasmtime-c-api/winch"]
 debug-builtins = ["wasmtime-c-api/debug-builtins"]
 pooling-allocator = ["wasmtime-c-api/pooling-allocator"]
+component-model = ["wasmtime-c-api/component-model"]
 # ... if you add a line above this be sure to read the comment at the end of
 # `default`
--- a/crates/c-api/build.rs
+++ b/crates/c-api/build.rs
@@ -22,6 +22,7 @@ const FEATURES: &[&str] = &[
     "DEBUG_BUILTINS",
     "WAT",
     "POOLING_ALLOCATOR",
+    "COMPONENT_MODEL",
 ];
 // ... if you add a line above this be sure to change the other locations
 // marked WASMTIME_FEATURE_LIST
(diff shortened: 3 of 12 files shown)


alexcrichton (commented):
Thanks!

alexcrichton:
Also if you're up for it feel free to fill out this header to match the `module.h` header, e.g. with deserializing and such. Reflection on imports/exports though is going to be a "big deal" so deferring that is probably best.

MangoPeachGrape:
> Also if you're up for it feel free to fill out this header to match the `module.h` header, e.g. with deserializing and such. Reflection on imports/exports though is going to be a "big deal" so deferring that is probably best.

Added `wasmtime_component_serialize()` and `wasmtime_component_serialize()`. Should I do `wasmtime_module_deserialize_file()` as well, or any other functions?

For the documentation in the header file I did `s/module/component/`, so it might not be accurate. In general, do you want documentation on everything in the header files in these early commits, or is that something that can be added in PRs down the line?

alexcrichton:
`wasmtime_component_deserialize_file` sounds good to add yeah, and maybe `wasmtime_component_clone` as well, but otherwise it looks like other functions aren't as applicable yet so that's probably good for now.

For docs I find it best to add it when the functions come in, but I'll admit during review I generally gloss over docs. I can give them a once-over to confirm everything is right after deserialize_file is added

alexcrichton:
Ah yeah the const there is ok, and it's one where we aren't necessarily the best-principled, but adding `const` when it maps to `&T` in Rust doesn't hurt for sure.

For file organization, yeah if you're up for it I think it would be a good idea. We'll be filling this out more in the future so I think it makes sense to separate everything out by file and starting here feels a bit odd as it's so small but it's predicted to get larger and be more justified in the future.

MangoPeachGrape:
> For file organization, yeah if you're up for it I think it would be a good idea. We'll be filling this out more in the future so I think it makes sense to separate everything out by file and starting here feels a bit odd as it's so small but it's predicted to get larger and be more justified in the future.

Done!

What should the scope be for the next PR?

alexcrichton (approved):

MangoPeachGrape:
Yes, tests would be great!

All the tests would use GoogleTest, would you still seperate the tests for the C api and the C++ api, or would you have those combined?

Is this something you would want me to take a jab at? I'm not that familiar with the codebase, so I would need some guiding where to put everything.

alexcrichton:
GoogleTest would be fine yeah, and I think it'd be reasonable to test everything together. I'd expected we'd test C++ APIs by deafult (by moving `wasmtime.hh` to this repository) and then we could write one-off tests in C if necessary.

If you're willing to work on this that'd be great! Most of us in this repo aren't really all that well-versed in C/C++ repository management so if you know of practices/idioms around organization/testing/CI/etc that'd be very helpful. Most of the source would probably want to live in `crates/c-api` or some sub-folder, and I'm happy to help out with CI integration as well.

MangoPeachGrape:
@alexcrichton, Some questions that that came up while moving the C++ api into this repository:
- Should the C++ examples go to `examples/`, and the tests go to `crates/c-api/tests/`?
- Should we guard the C++ api to be only built if the consumer wants it (some consumers might not be able to build C++)?
- Should C/C++ tests be built by default, or only be enabled by a CMake option when desired?

alexcrichton:
Yeah let's put examples in `examples/{example}.{rs,c,cpp}` and in theory we can fill out that matrix over time. For tests I think `crates/c-api/tests` is a good spot too.

The C++ API build step from an end consumable artifact point of view should just be a header file so I think it's fine to always enable/copy that around. For the tests I think it's also reasonable to require C++ to build tests since that's just for developers which should be easier in general. Given that I'd hope we can skip CMake options to conditionalize the C++ usage and basically just always use it

PR #10568, opened 2025-04-11
wasmtime-slab: add PartialOrd and Ord to Ids.

cfallin:
This became necessary in a use-case where I am using slab IDs as keys in a BTreeMap (as a member of a tuple key type). The actual ordering doesn't carry a fundamental meaning, but overall I do need ordering as the first elements of the tuple have significant ordering and I need to process in order (namely: timer deadlines). The slab ID ordering is at least stable (since the Ids are integers under the hood), and this is useful for anyone who needs to build an ordered key type. The alternative is to use `.into_raw()` and manually write an `Ord` impl or use the raw `u32` in a key, which is much worse.

### DIFF
--- a/crates/slab/src/lib.rs
+++ b/crates/slab/src/lib.rs
@@ -131,7 +131,7 @@ use core::fmt;
 use core::num::NonZeroU32;
 
 /// An identifier for an allocated value inside a `slab`.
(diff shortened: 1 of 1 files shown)


alexcrichton (approved):

cfallin:
Merge queue job for Nightly tests failed due to `apt install` freezing for 25 minutes; retrying.

PR #10580, opened 2025-04-14
asm: replace flag-using ALU instructions

abrown:
This change replaces ISLE lowerings for the `ProducesFlags` and `ConsumesFlags` wrappers with instructions from the new assembler. This is a necessary step towards fully using the new assembler for ALU instructions (`AluRmiR` is only used for zeroing registers now).

A couple items of note:
- there are a few places where `lower.isle` wants to abstract over some subset of ALU operations to avoid repetition; this change retains that via some `x64_*_flags` helpers that take a new type (`ChainFlagsOp`, e.g.) that restricts which operations the helper can use. This is not a major refactor but it does show that we use these flag-producing instructions in quite limited ways.
- we previously decided to keep rules that used wider versions of registers for i8/i16 operations to avoid CPU dependencies; this is problematic for instructions that rely on overflow, though, so this change extracts those "wider register" rules into `x64_*_break_deps` helpers.

### DIFF
--- a/cranelift/codegen/src/isa/x64/inst.isle
+++ b/cranelift/codegen/src/isa/x64/inst.isle
@@ -2765,141 +2765,213 @@
 (rule (x64_xmm_load_const ty const)
       (x64_load ty (const_to_synthetic_amode const) (ExtKind.None)))
 
+
+;;;; Flag Helpers ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;
+;; These helpers are used to emit instructions that produce or consume flags.
+;; The operations used here are by no means the only ones possible; they are
+;; simply the ones currently used in Cranelift's lowerings.
+
+;; Some operations produce flags.
+(type ProduceFlagsOp (enum (Add) (Sub)))
+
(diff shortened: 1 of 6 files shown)


cranelift/codegen/src/isa/x64/inst.isle
```diff
@@ -2765,141 +2765,213 @@
...
+;; Some operations produce flags.
+(type ProduceFlagsOp (enum (Add) (Sub)))
+
+(decl x64_produce_flags (ProduceFlagsOp Type Gpr GprMemImm) ProducesFlags)
+(rule (x64_produce_flags (ProduceFlagsOp.Add) ty src1 src2)
+      (x64_add_with_flags_paired ty src1 src2))
+(rule (x64_produce_flags (ProduceFlagsOp.Sub) ty src1 src2)
+      (x64_sub_with_flags_paired ty src1 src2))
+
+(decl asm_produce_flags (AssemblerOutputs) ProducesFlags)
```
cfallin, line 2784: Can we add a comment here that this should only be provided an instruction that actually produces flags, and is meant to be used by helpers that bind it with explicit instructions? In other words, it's semantically "unsafe" in the Rust sense and has to be used in the right way to uphold invariants.

cfallin (approved):
Nice! I like the factoring here -- once the low-level helpers are out of the way, the application of this to each instruction is pretty straightforward.

PR #10582, opened 2025-04-15
c-api: Bring wasmtime-cpp into this repository

MangoPeachGrape:
See discussion in #10566 for more context.

Folloving files have been moved without edits:
- `wasmtime-cpp/include/wasmtime.hh` -> `/crates/c-api/include/`
- `wasmtime-cpp/tests/*` -> `/crates/c-api/tests/`
- `wasmtime-cpp/examples/*` -> `/examples/`

The CMakeLists.txt content is copied to `/crates/c-api/CMakeLists.txt` with slight adjustments. I haven't fully gone through and checked if everything from the old CMake file still makes sense.

The tests inside `/crates/c-api/tests/` seem to run when examples ("Test C-API" from CI) is executed.

Added a CMake option to disable tests, feel free to let me know if this is not desired.


crates/c-api/CMakeLists.txt
```diff
@@ -1,8 +1,9 @@
 cmake_minimum_required(VERSION 3.12)
-project(wasmtime C)
+project(wasmtime)
 
 set(WASMTIME_USER_CARGO_BUILD_OPTIONS "" CACHE STRING "Additional cargo flags (such as --features) to apply to the build command")
 option(BUILD_SHARED_LIBS "Build using shared libraries" OFF)
+option(BUILD_TESTS "Build tests" ON)
```
alexcrichton, line 6: Could this default to OFF but we be careful to pass it as ON in CI? Agreed though that this is good to have, thanks for adding this!

alexcrichton (commented):
Thanks for this! This looks pretty good to me, but I'd also want to test some more integration-y things before landing. To that end can you try a few things:

* Add a commit to this PR with "prtest:full" somewhere in the commit message so we can see the full CI run
* With a full run I want to double-check that the header file is present in the installation artifacts
* Mind updating `crates/c-api/README.md` with a mention that the C++ API is present as well? I can also do this as a follow-up if you'd prefer.
* With a full run I want to double-check that the docs here are present in the c-api docs.

alexcrichton:
Also, after merging, I'll take on the bits of archiving the old repo.

MangoPeachGrape:
CI doesn't seem to have C++ compiler? Should a C++ compiler be a requirement, even when we don't build any C++ code?

alexcrichton:
Could this be updated to only require a C++ compiler for tests? For a normal build it shouldn't be required I think?

alexcrichton:
Ok nice looks like everything's integrated well, I can see the C++ tests running and I see `wasmtime.hh` in the distribution artifacts as well. Can you double-check the docs though? I don't see the doxygen docs for the C++ API in the gh-pages artifact, just the C API. I suspect we need to tweak `doxygen.conf` a bit?

MangoPeachGrape:
Would it be preferred to only silence the undocumented warnings from `wasm.hh` instead of disabling it all together?

alexcrichton:
If possible, yeah, although I'm not sure if it's easy to configure doxygen as such.

MangoPeachGrape:
I was thinking of `EXCLUDE = wasm.hh`, but now I realize that it obviously removes it from the output as well. What's better, exclude `wasm.hh` and enable the warning, or keep it like it is now?

alexcrichton (approved):
Nice docs are looking great. Thank you again for working on this!

PR #10597, opened 2025-04-16
Component and Instance have corresponding get_export, get_export_index

pchickey:
Prior to this PR, there was `Component::export_index` which looked up `(ComponentItem, ComponentExportIndex)` and `Instance::get_export` which looked up `ComponentExportIndex`.

Now each of these structures has a `get_export` which looks up the `(ComponentItem, ComponentExportIndex)` and `get_export_index` which looks up just the `ComponentExportIndex`. The docs point to the sister methods, and mention that creating the ComponentItem is comparatively expensive so use the cheaper `get_export_index` if you don't need it.

This will be a breaking change to users, but the hope is that most users are only using these by way of wasmtime-wit-bindgen, so hopefully its not a big deal.

### DIFF
--- a/crates/wasmtime/src/runtime/component/instance.rs
+++ b/crates/wasmtime/src/runtime/component/instance.rs
@@ -1,7 +1,8 @@
 use crate::component::func::HostFunc;
 use crate::component::matching::InstanceType;
 use crate::component::{
-    Component, ComponentExportIndex, ComponentNamedList, Func, Lift, Lower, ResourceType, TypedFunc,
+    types::ComponentItem, Component, ComponentExportIndex, ComponentNamedList, Func, Lift, Lower,
+    ResourceType, TypedFunc,
 };
 use crate::instance::OwnedImports;
 use crate::linker::DefinitionType;
@@ -279,8 +280,9 @@ impl Instance {
     /// instance.
     ///
     /// This method will lookup the `name` provided within the `instance`
-    /// provided and return a [`ComponentExportIndex`] which can be used to
-    /// pass to other `get_*` functions like [`Instance::get_func`].
+    /// provided and return a [`ComponentItem`] describing the export,
+    /// and [`ComponentExportIndex`] which can be passed other `get_*`
+    /// functions like [`Instance::get_func`].
     ///
     /// # Panics
     ///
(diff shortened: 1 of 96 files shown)


crates/wasmtime/src/runtime/component/instance.rs
```diff
@@ -299,9 +305,48 @@ impl Instance {
...
+            Export::Type(ty) => ty,
+        };
+        let item = ComponentItem::from(&store.engine(), &ty, &data.ty());
+        Some((
+            item,
+            ComponentExportIndex {
+                id: data.component_id(),
+                index,
+            },
+        ))
```
alexcrichton, line 326: Since this is a copy of what's in `component.rs`, could there be a crate-private helper which takes the `InstanceType` as an argument and both this and the public component version call into that?

crates/wasmtime/src/runtime/component/instance.rs
```diff
@@ -299,9 +305,48 @@ impl Instance {
...
+    /// provided and return a [`ComponentExportIndex`] which can be passed
+    /// other `get_*` functions like [`Instance::get_func`].
+    ///
+    /// If you need the [`ComponentItem`] corresponding to this export, use
+    /// the [`Instance::get_export`] instead.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `store` does not own this instance.
+    pub fn get_export_index(
```
alexcrichton, line 342: Possible bikeshed: instead of `get_export{,_index}`, maybe `get_export{,_and_type}`? (e.g. favoring the just-the-index version)

pchickey: Eh, maybe, but get_export_and_type feels too smalltalky and 98% of users just use bindgen which hides all that so I'll respectfully decline

crates/wasmtime/src/runtime/component/instance.rs
pchickey: Good call, I'll make a `pub(crate) ComponentItem::from_export` that takes a wasmtime-environ::Export which should reduce the duplication to just a line or two.

src/commands/run.rs
```diff
@@ -567,7 +567,7 @@ impl RunCommand {
                 let mut export = None;
                 for name in names {
                     let (_item, ix) = component
-                        .export_index(export.as_ref(), name)
+                        .get_export(export.as_ref(), name)
```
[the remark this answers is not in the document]
pchickey: Fixed 947b90df

crates/wasmtime/src/runtime/component/instance.rs
pchickey: [2b13acd](https://github.com/bytecodealliance/wasmtime/pull/10597/commits/2b13acdf960f8106a20ca96f4928f7ffd9af9973)

alexcrichton (approved):

PR #10598, opened 2025-04-16
c-api: Create and instantiate a component model linker

MangoPeachGrape:
Next small chunk of #9812.

I'm not sure about the lifetime requirements of `wasmtime_component_linker_instance()`, is there any better way to express this?
See previous discussion: https://github.com/bytecodealliance/wasmtime/pull/9812#discussion_r1884466077


alexcrichton:
My thinking is that most of the tests could use the text format of WebAssembly, meaning the test wasm file would be in-line with the test itself (this is used a lot on the Rust side for testing too). Orchestrating native-language binaries compiled to wasm to be available in the tests is possible but will be a bit of a pain.

For location, yeah I'm thinking we'd drop things in `crates/c-api/tests/*.cc`. And yeah if you'd prefer to start by testing the C API I think that's reasonable.

crates/c-api/include/wasmtime/component/instance.h
```diff
@@ -0,0 +1,27 @@
+#ifndef WASMTIME_COMPONENT_INSTANCE_H
+#define WASMTIME_COMPONENT_INSTANCE_H
+
+#ifdef WASMTIME_FEATURE_COMPONENT_MODEL
```
alexcrichton, line 4: I may have missed this on other headers, but I think for this to work reliably you'll want to `#include <wasmtime/conf.h>` in these headers too? (basically fix the case where this header, and just this header, are included)

crates/c-api/include/wasmtime/component/linker.h
```diff
@@ -0,0 +1,95 @@
...
+#include <wasmtime/store.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct wasmtime_component_linker_t wasmtime_component_linker_t;
+
+typedef struct wasmtime_component_linker_instance_t
+    wasmtime_component_linker_instance_t;
```
alexcrichton, line 19: Mind deferring this type to later? I think there's going to be some tricky ownership questions we'll want to grapple with for this (e.g. handling the lifetime parameter in the Rust side of things)

crates/c-api/include/wasmtime/component/instance.h
```diff
@@ -0,0 +1,27 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct wasmtime_component_instance_t wasmtime_component_instance_t;
```
alexcrichton, line 10: I think we'll want this to be similar to `wasmtime_instance_t` where it's not an opaque struct but instead it's a known sturct of two integers (although embeddings won't use the integers much). That'll also obviate the need for deletion of this structure too

alexcrichton (approved):

MangoPeachGrape:
@alexcrichton What do you want in the next PR?

alexcrichton:
How about sketching out the `Linker` API? Basically being able to create recursive instances/etc. Perhaps hooking up the `wasmtime_module_t` type where you can supply a core module to a component? (in lieu of defining functions which I think is a bit harder to do as a single step)

My main concern there is that the `Linker` API in Rust has a lot of borrowed lifetimes which makes it nontrivial to map to other languages. Clear ownership semantics tend to work out best IMO, and it's also the case that performance is not of the utmost concern when creating a `Linker`. In that sense I think we may want to design this in such a way such that separate pointers in the C API aren't intrinsically linked but instead perhaps define an item-at-a-time where there's no intermediately-owned structures?

One example could be that the entire "path" to an item is passed to the C API, where each element in the path is an exported instance layer. That would then call `.instance(..)` as necessary in the Rust API and define the requested item at the leaf when the path is iterated over. Or... something like that, just some vague ideas.

It's not that we can't have inter-related objects in the C API, it's mostly just that I'd prefer to avoid it if possible.

PR #10600, opened 2025-04-16
Sketch out merging the C/C++ APIs

alexcrichton:
With the merging of the C++ API into this repository in #10582 it opens up some interesting questions about how to organize the C++ API. Externally it was all entirely a single file, but naturally this isn't great for evolution as it's just one giant tangled header. Instead this commit sketches out a possible different path forward which is to provide the C++ API as a sibling to the C API in preexisting header files. For example this moves the `Error` class to the `error.h` header file as an example.

My rough hope would be that in the long-term we could deprecate/remove the `wasmtime.hh` header file and instead "just" have all the C++ APIs in the normal header files (e.g. `wasmtime.h`). Additionally the split of the C API in separate header files would be amenable to a similar split of the C++ API too where the API you see is basically conditional on the language mode of whatever's including the headers.

I'll note though I've not seen prior art in doing this. I'm not aware of any other project which exports both a C and C++ API in its header files. That being said I'm not sure how many other projects would fall in such a bucket.


fitzgen (commented):
I'm a fan of splitting up the monolith C++ header and also the tests, but I am less sold on sharing headers for C and C++. What do you think of having

* `foo.h` -- the C API for `foo` things
* `foo.hh` -- includes the C API, adds the C++ layer on top

?

But that said, this is just a soft preference, and I also don't have a ton of experience designing C/C++ libraries, so if there is a better way I am also all ears.

ac000:
> I'm a fan of splitting up the monolith C++ header and also the tests, but I am less sold on sharing headers for C and C++. What do you think of having
> 
>     * `foo.h` -- the C API for `foo` things
> 
>     * `foo.hh` -- includes the C API, adds the C++ layer on top

FWIW that's probably the way I'd go...

alexcrichton:
Sounds reasonable to me, updated!

fitzgen (approved):
THanks!

PR #10603, opened 2025-04-16
Drop `shellexpand` from dependency tree

alexcrichton:
This is a historical dependency of `wiggle` which can be worked around in a different manner by making paths relative to `CARGO_MANIFEST_DIR` by default.


pchickey (approved):

pchickey:
Still one test to fix in crates/wiggle/tests/wasmtime_integration.rs

PR #10609, opened 2025-04-17
Cranelift: move exception-handler metadata into callsites.

cfallin:
This PR starts with a cherry-pick of @bjorn3's work at [1], which switches exception-handler metadata from a separate list of records (of callsite offset, tag, handler offset tuples) to per-callsite (tag, handler) tuples. This is cleaner for both the native-code use-case and for our planned Wasmtime use-case: in both cases one knows the specific callsite in each frame and wants to look at handlers only for that callsite. The old representation was designed for fast emission and flat storage, but we can get flat storage with this semantic structure as well...

... with the second commit, in which I've updated @bjorn3's implementation to keep a separate vector of handler tuples in the `MachBuffer`, and refer to ranges in this vector in the callsite metadata. It also provides an iterator that yields per-callsite information in a safe way.

No functional change in-tree (because Cranelift and Wasmtime both do not use this metadata yet) but should enable `cg_clif` to use our exception support for native unwind.

Co-authored-by: `bjorn3 <17426603+bjorn3@users.noreply.github.com>`

[1]: https://github.com/bjorn3/wasmtime/commit/72c68f2c68ee3fe3159e203c6277d5322844ad37#diff-f9524bac9108c13ddfe351160d287ab858d4e815181a867e56fa5382e5bd564e

### DIFF
--- a/cranelift/codegen/src/machinst/buffer.rs
+++ b/cranelift/codegen/src/machinst/buffer.rs
@@ -172,7 +172,7 @@
 
 use crate::binemit::{Addend, CodeOffset, Reloc};
 use crate::ir::function::FunctionParameters;
-use crate::ir::{ExternalName, RelSourceLoc, SourceLoc, TrapCode};
+use crate::ir::{ExceptionTag, ExternalName, RelSourceLoc, SourceLoc, TrapCode};
 use crate::isa::unwind::UnwindInst;
 use crate::machinst::{
     BlockIndex, MachInstLabelUse, TextSectionBuilder, VCodeConstant, VCodeConstants, VCodeInst,
@@ -259,8 +259,6 @@ pub struct MachBuffer<I: VCodeInst> {
     user_stack_maps: SmallVec<[(CodeOffset, u32, ir::UserStackMap); 8]>,
     /// Any unwind info at a given location.
     unwind_info: SmallVec<[(CodeOffset, UnwindInst); 8]>,
-    /// Any exception handler targets at a given location.
-    exception_handlers: SmallVec<[(CodeOffset, PackedOption<ir::ExceptionTag>, MachLabel); 8]>,
     /// The current source location in progress (after `start_srcloc()` and
     /// before `end_srcloc()`).  This is a (start_offset, src_loc) tuple.
     cur_srcloc: Option<(CodeOffset, RelSourceLoc)>,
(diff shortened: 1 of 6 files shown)


cranelift/codegen/src/machinst/buffer.rs
```diff
@@ -1937,6 +1967,25 @@ pub struct MachTrap {
...
+    exception_handler_range: Range<u32>,
+}
+
+/// A call site record resulting from a compilation.
+#[derive(Clone, Debug, PartialEq)]
+#[cfg_attr(
+    feature = "enable-serde",
+    derive(serde_derive::Serialize, serde_derive::Deserialize)
+)]
+pub struct FinalizedMachCallSite {
```
bjorn3, line 1982: This doesn't need to be a separate type anymore as `MachCallSite` no longer contains `MachLabel`s.

bjorn3: Or actually, maybe we should store `MachCallSite` and have `buffer.call_sites()` return owned `FInalizedMachCallSite<'a>` values with the `exception_handler_range` field replaced by `exception_handler: &'a [(PackedOption<ir::ExceptionTag>, CodeOffset)]`?

cranelift/codegen/src/machinst/buffer.rs
```diff
@@ -2494,10 +2546,14 @@ mod test {
             1,
         );
         buf.put1(4);
+        buf.bind_label(label(1), ctrl_plane);
+        buf.put1(0xff);
+        buf.bind_label(label(2), ctrl_plane);
```
bjorn3, line 2551: Looks like line 2525 needs to be changed to `buf.reserve_labels_for_blocks(3);`.

cfallin: Yep, that explains the failures in CI, thanks!

cranelift/codegen/src/machinst/buffer.rs
cfallin: Ah, yes, I like the latter option a lot -- updated, thanks!

cranelift/codegen/src/machinst/buffer.rs
```diff
@@ -1937,6 +1954,21 @@ pub struct MachTrap {
...
+}
+
+/// A call site record resulting from a compilation.
+#[derive(Clone, Debug, PartialEq)]
+pub struct FinalizedMachCallSite<'a> {
+    /// The offset of the call's return address, *relative to the containing section*.
+    pub ret_addr: CodeOffset,
+
+    /// Exception handlers at this callsite, with target offsets
+    /// *relative to the containing section*.
```
bjorn3, line 1970: Shouldn't these all say "relative to the containing function"? There can be multiple functions in a single section.

bjorn3 (approved):
Tested with cg_clif.

PR #10610, opened 2025-04-17
wasmtime-wit-bindgen: Typecheck exports at {Foo}Indices construction

### LINKED ISSUE #9155: wasmtime::component::bindgen: GuestPre should typecheck exports before instantiation
Context: `wasmtime::component::bindgen` creates a `GuestPre` struct (https://github.com/bytecodealliance/wasmtime/blob/main/crates/wit-bindgen/src/lib.rs#L578) for each exported interface.

`GuestPre::new` performs the `export_index` lookups to validate that each function in the interface exists. However, it does not typecheck the functions in `new` - instead, they are typechecked in `GuestPre::load`, which takes a store and therefore must occur at instantiation time.

The information to typecheck these exports exists in the `Component` (specifically in the `wasmtime_environ::component::Component` via `Component::env_component`) but right now there is no straightforward way to typecheck against that - `load` typechecks by way of `Func::typed` which requires a store argument, but ultimately ends up getting all the types out of that same env_component information.

We should fix this so that GuestPre can be used to validate that the component's exports are all typechecked. This is useful for two reasons:
1. it allows bindgen's `Pre` types to be used to fully validate a component is well typed without instantiating it - it already does this for imports, only exports are missing
2.  we could potentially eliminate type checking performed in the (`Instance::get_typed_func`)[https://github.com/bytecodealliance/wasmtime/blob/main/crates/wit-bindgen/src/lib.rs#L2577] performed in `GuestPre::load` which would make instantiation faster, and have one less path of possible failure.

pchickey:
Based on #10621

Currently, in bindgen-emitted component wrappers, the existence of export functions is checked at the construction of the {interface-name}Indices struct, but each export function is only finally typechecked on their invocation in {interface-name}::call_{func-name}.

The goal of this PR is to typecheck component exports at the construction of Indices. This is desirable because it gives a type error as early as possible - when using {interface-name}Indices by way of {interface-name}Pre, this can be performed with only an InstancePre, where the creation of an InstancePre requires only a Component and Linker, and typechecks all of the Component's imports. So, really, this is about getting exports on parity with imports.

Closes #9155 . I've worked around this with ugly hacks in embeddings for 2 different employers now and I'm not about to do it for a third.

## Addition to Wasmtime

`component::types::ComponentFunc` has a new method `typecheck<Params, Returns>(&self, instance_type: &InstanceType) -> Result<()>`. The underlying typecheck performed here is the exact same one performed in `Func::typed<Params, Returns>(&self, store: impl AsContextMut) -> Result<()>`, except rather than the type information coming out of the store, its provided by the InstanceType.

This addition is a pub interfaces but marked `#[doc(hidden)]` to make them "internal", because the prior art was that `InstanceType` and is internal, and there are no other public interfaces for retrieving `InstanceType` or typechecking, with the sole exception of `Func::typed`. I think that we should consider making these documented public interfaces, but I'll leave that to a future PR unless requested.

## Changes to wasmtime-wit-bindgen

The `resources-import` test case demonstates the extent of the changes: https://github.com/bytecodealliance/wasmtime/pull/10610/files#diff-0b0815115f3addd488fbf1f032466f210b9fcd29ec42f4b4a777e58f80abba16

* Since typechecking information is available from the InstancePre (thanks to #10621), we can unify the two variations on constructing an Indices struct: there is now a single constructor `fn new<T>(_instance_pre: InstancePre<T>) -> Result<Self>`. Deduplicating this eliminates both generated code and complexity in the generator.
* The new sigular constructor uses `_instance_pre.component()` to lookup exports (both their index and their ComponentItem), and `_instance_pre.instance_type()` is used to typecheck the export.
* Each place an export index is retrieved for a function, `.get_export_index` has been substituted out with `.get_export`, and the returned `ComponentItem` is checked to be a ComponentItem::ComponentFunc(func), and then `func.typecheck` is invoked with the type argument corresponding to the `TypedFunc` type parameters.

## Future work

Each export func is typechecked again in its bindgen `call_{funcname}` function, in which the `Func` member owned by the typed instance `{Foo}` struct is cast to a `TypedFunc` with the specific type parameters bindgen keeps track of for each func. Since the `{Foo}` struct is created from a `{Foo}Indices`, we can guarantee that the export funcs have been typechecked against a given `Linker` or `Store`, so we could potentially elide the repeated typecheck there, and make an unchecked cast to TypedFunc instead. (TypedFuncs are not stored in the `{Foo}` struct at least in part because their type parameters may use `&'a ...` which only apply to the site of the call, and not to the storage of that member in `{Foo}`. The `for<'a>` trick only works for trait objects, not concrete types like TypedFunc.)

### DIFF
--- a/crates/component-macro/tests/expanded/worlds-with-types_tracing_async.rs
+++ b/crates/component-macro/tests/expanded/worlds-with-types_tracing_async.rs
@@ -52,7 +52,7 @@ impl<_T> FooPre<_T> {
     pub fn new(
         instance_pre: wasmtime::component::InstancePre<_T>,
     ) -> wasmtime::Result<Self> {
-        let indices = FooIndices::new(instance_pre.component())?;
+        let indices = FooIndices::new(&instance_pre)?;
         Ok(Self { instance_pre, indices })
     }
     pub fn engine(&self) -> &wasmtime::Engine {
@@ -110,11 +110,6 @@ pub struct FooIndices {
 /// * If you've instantiated the instance yourself already
 ///   then you can use [`Foo::new`].
 ///
-/// * You can also access the guts of instantiation through
-///   [`FooIndices::new_instance`] followed
-///   by [`FooIndices::load`] to crate an instance of this
-///   type.
-///
 /// These methods are all equivalent to one another and move
 /// around the tradeoff of what work is performed when.
 ///
@@ -133,30 +128,26 @@ const _: () = {
         ///
         /// This method may fail if the component does not have the
         /// required exports.
-        pub fn new(
-            component: &wasmtime::component::Component,
+        pub fn new<_T>(
+            _instance_pre: &wasmtime::component::InstancePre<_T>,
         ) -> wasmtime::Result<Self> {
-            let _component = component;
-            let f = _component
-                .get_export_index(None, "f")
-                .ok_or_else(|| anyhow::anyhow!("no function export `f` found"))?;
-            Ok(FooIndices { f })
-        }
-        /// Creates a new instance of [`FooIndices`] from an
-        /// instantiated component.
-        ///
-        /// This method of creating a [`Foo`] will perform string
-        /// lookups for all exports when this method is called. This
-        /// will only succeed if the provided instance matches the
-        /// requirements of [`Foo`].
-        pub fn new_instance(
-            mut store: impl wasmtime::AsContextMut,
-            instance: &wasmtime::component::Instance,
-        ) -> wasmtime::Result<Self> {
-            let _instance = instance;
-            let f = _instance
-                .get_export_index(&mut store, None, "f")
-                .ok_or_else(|| anyhow::anyhow!("no function export `f` found"))?;
+            let _component = _instance_pre.component();
+            let _instance_type = _instance_pre.instance_type();
+            let f = {
+                let (item, index) = _component
+                    .get_export(None, "f")
+                    .ok_or_else(|| anyhow::anyhow!("no export `f` found"))?;
+                match item {
+                    wasmtime::component::types::ComponentItem::ComponentFunc(func) => {
+                        anyhow::Context::context(
+                            func.typecheck::<(), ((T, U, R),)>(&_instance_type),
+                            "type-checking export func `f`",
+                        )?;
+                        index
+                    }
+                    _ => Err(anyhow::anyhow!("export `f` is not a function"))?,
+                }
+            };
             Ok(FooIndices { f })
         }
         /// Uses the indices stored in `self` to load an instance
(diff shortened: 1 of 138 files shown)


crates/component-macro/tests/expanded/worlds-with-types_tracing_async.rs
```diff
@@ -137,9 +137,21 @@ const _: () = {
...
+                            func.typecheck::<(), ((T, U, R),)>(),
+                            "type-checking export func `f`",
+                        )?;
+                        index
+                    }
+                    _ => Err(anyhow::anyhow!("export `f` is not a function"))?,
+                }
+            };
             Ok(FooIndices { f })
         }
```
alexcrichton, line 156: Technically this isn't possible, type-checking at the component level. Components import abstract resources which don't match any types defined in Rust, which is what would be passed in as the type parameters here. The type-check can only be done once you have an `InstancePre` or `Linker` where the actual resource types are all assigned and we can match up the host-provided resources with our expectations.

That being said I do like the property that `FooIndices` is all you need to "quickly load the thing" and as such represents a type-checked version of the component. Given that one options is to switch this to using `InstancePre<T>` instead of `&Component` as a construction argument. In doing so though we're not really prepared to have that working. The problem is that we'd have to prevent crossing `FooIndices` values by accident.  Right now export indices are intrinsically tied to a `Component` which has a unique ID, but not a `Linker` which does not have a unique ID. That means you could technically create a `FooIndices` with one linker and then extract exports from a component created with another linker, and if we skip the type-checks with `unsafe` (which I assume this PR is building towards) that could be memory-unsafe.

That might be solvable though by putting a unique ID in linkers and plumbing that to `InstancePre` and `Instance` which is checked.

pchickey: > The type-check can only be done once you have an InstancePre or Linker where the actual resource types are all assigned and we can match up the host-provided resources with our expectations.
Agreed, thats exactly where I got to - something representing that resource assignment (__internals::InstanceType) needs to be passed into ComponentFunc::typecheck as an argument. I got pretty tangled up with InstanceType because it borrows arcs of Store's type tables instead of cloning the arcs, so I am going to either change the structure or make a variant of it that owns the arcs.

dicej (approved):

alexcrichton:
I'm a bit concerned about https://github.com/bytecodealliance/wasmtime/pull/10616 and the cost implications of having more `Arc`-clones on lifting/lowering paths, and after reading over this more I've got a possibilty which I forget if we already talked about and/or whether you tried, so let me know if I sound like a broken record. Could the various `*Indices` constructors take a `&InstancePre<T>` instead of a `&Component`? That way we could refactor the internals of `InstancePre<T>` to hold whatever is necessary to cheaply create the `InstanceType<T>` I think?

alexcrichton (approved):

pchickey:
I addressed the concerns above and rewrote the description accordingly.

PR #10612, opened 2025-04-18
Log module "DNA" during fuzzing

alexcrichton:
This commit updates the fuzzing infrastructure of the `wasmtime-fuzzing` crate to record the "DNA string" of a module used to generate a module in a `*.dna` file. This is accompanied with a `*.json` file to pass to `wasm-tools smith --config`. The end result is that it should be possible now to more easily reproduce a module generation outside of Wasmtime itself when reproducing bugs and such.

Creation of these files is gated on the debug log level in a similar manner to creation of normal wasm files is gated on the debug log level too.


fitzgen (approved):
Great idea

PR #10619, opened 2025-04-20
Minimal implementation of fixed-length lists to enable wit-bindgen runtime tests

cpetig:
Implement the necessary parts of fixed size lists (see https://github.com/WebAssembly/component-model/pull/384 ) to enable runtime testing in wit-bindgen.

### DIFF
--- a/crates/environ/src/component/types_builder.rs
+++ b/crates/environ/src/component/types_builder.rs
@@ -594,7 +594,13 @@ impl ComponentTypesBuilder {
         size: u32,
     ) -> TypeFixedSizeListIndex {
         let element_abi = self.component_types.canonical_abi(&element);
-        let abi = CanonicalAbiInfo::record((0..size).map(|_| element_abi));
+        let mut abi = element_abi.clone();
+        // this assumes that size32 is already rounded up to alignment
+        abi.size32 = element_abi.size32 * size;
+        abi.size64 = element_abi.size64 * size;
+        abi.flat_count = element_abi
+            .flat_count
+            .map(|c| c.saturating_mul(size.min(255) as u8));
         self.add_fixed_size_list_type(TypeFixedSizeList { element, size, abi })
     }
 
@@ -1115,7 +1121,21 @@ impl TypeInformation {
     }
 
     fn fixed_size_lists(&mut self, types: &ComponentTypesBuilder, ty: &TypeFixedSizeList) {
-        self.build_record((0..ty.size).map(|_| types.type_information(&ty.element)));
+        let element_info = types.type_information(&ty.element);
+        self.depth = 1 + element_info.depth;
+        self.has_borrow = element_info.has_borrow;
+        match element_info.flat.as_flat_types() {
+            Some(types) => {
+                'outer: for _ in 0..ty.size {
+                    for (t32, t64) in types.memory32.iter().zip(types.memory64) {
+                        if self.flat.push(*t32, *t64) {
+                            break 'outer;
+                        }
+                    }
+                }
+            }
+            None => self.flat.len = u8::try_from(MAX_FLAT_TYPES + 1).unwrap(),
+        }
     }
 
     fn enums(&mut self, _types: &ComponentTypesBuilder, _ty: &TypeEnum) {
--- a/crates/environ/src/component/types_builder.rs
+++ b/crates/environ/src/component/types_builder.rs
@@ -596,11 +596,12 @@ impl ComponentTypesBuilder {
         let element_abi = self.component_types.canonical_abi(&element);
         let mut abi = element_abi.clone();
         // this assumes that size32 is already rounded up to alignment
-        abi.size32 = element_abi.size32 * size;
-        abi.size64 = element_abi.size64 * size;
+        abi.size32 = element_abi.size32.saturating_mul(size);
+        abi.size64 = element_abi.size64.saturating_mul(size);
         abi.flat_count = element_abi
             .flat_count
-            .map(|c| c.saturating_mul(size.min(255) as u8));
+            .zip(u8::try_from(size).ok())
+            .and_then(|(flat_count, size)| flat_count.checked_mul(size));
         self.add_fixed_size_list_type(TypeFixedSizeList { element, size, abi })
     }
 
--- a/crates/environ/src/component/types_builder.rs
+++ b/crates/environ/src/component/types_builder.rs
@@ -459,7 +459,7 @@ impl ComponentTypesBuilder {
                 InterfaceType::Stream(self.stream_table_type(types, ty)?)
             }
             ComponentDefinedType::FixedSizeList(ty, size) => {
-                InterfaceType::FixedLengthList(self.fixed_size_list_type(types, ty, *size)?)
+                InterfaceType::FixedLengthList(self.fixed_length_list_type(types, ty, *size)?)
             }
             ComponentDefinedType::Map(..) => {
                 bail!("support not implemented for map type");
@@ -577,7 +577,7 @@ impl ComponentTypesBuilder {
         self.add_tuple_type(TypeTuple { types, abi })
     }
 
-    fn fixed_size_list_type(
+    fn fixed_length_list_type(
         &mut self,
         types: TypesRef<'_>,
         ty: &ComponentValType,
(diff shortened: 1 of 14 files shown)


crates/environ/src/fact/trampoline.rs
```diff
@@ -2628,6 +2628,31 @@ impl<'a, 'b> Compiler<'a, 'b> {
...
+        assert_eq!(src_ty.size, dst_ty.size);
+
+        let srcs = src
+            .record_field_srcs(self.types, (0..src_ty.size).into_iter().map(|_| src_ty.element));
+        let dsts = dst
+            .record_field_dsts(self.types, (0..dst_ty.size).into_iter().map(|_| dst_ty.element));
+        for (src, dst) in srcs.zip(dsts) {
+            self.translate(&src_ty.element, &src, &dst_ty.element, &dst);
+        }
+    }
```
alexcrichton, line 2654: This is a bit of an interesting case. Effectively what this is doing is unrolling `list<T, N>` unconditionally, but I don't think that's necessarily appropriate for large `N`. I think it's fine to unroll for small `N` (perhaps only when `T` is small?). Ideally this would have a budget of sorts where a fixed number `Cost` is divided by the cost of `T` , and that's the maximal unrolling. That way we don't generate exponential amounts of code here for `list<list<T, N>, N>`.

Basically what I'm saying is that I think we must implement the loop-based variant of translating lists, not just the fully-unrolled version of translating lists.

crates/wasmtime/src/runtime/component/types.rs
```diff
@@ -148,6 +148,7 @@ impl TypeChecker<'_> {
             (InterfaceType::Future(_), _)
             | (InterfaceType::Stream(_), _)
             | (InterfaceType::ErrorContext(_), _) => todo!(),
+            (InterfaceType::FixedSizeList(_), _) => todo!(),
```
alexcrichton, line 151: Since these are relatively targeted `todo!()` items mind filing an issue as a tracking issue and tagging these with `// FIXME(#NNNN)`?

cpetig: See https://github.com/bytecodealliance/wasmtime/issues/12279

crates/environ/src/fact/trampoline.rs
cpetig: This takes a bit more effort, I will complete this later.

cpetig:
The best test for this functionality are the tests added to wit-bindgen in https://github.com/bytecodealliance/wit-bindgen/pull/1277 . 

I can create a test for this one as well (I didn't find a good template in the existing tests), but perhaps it makes most sense to introduce tests together with the host support in #12315 .

crates/environ/src/component/types_builder.rs
```diff
@@ -575,6 +577,33 @@ impl ComponentTypesBuilder {
...
+    pub(crate) fn new_fixed_size_list_type(
+        &mut self,
+        element: InterfaceType,
+        size: u32,
+    ) -> TypeFixedSizeListIndex {
+        let element_abi = self.component_types.canonical_abi(&element);
+        let mut abi = element_abi.clone();
+        // this assumes that size32 is already rounded up to alignment
+        abi.size32 = element_abi.size32 * size;
+        abi.size64 = element_abi.size64 * size;
```
alexcrichton: Question for you (sorry it's been awhile since I looked at this) -- is there a validation predicate in wasm-tools to ensure that this multiplication doesn't overflow?

crates/environ/src/component/types_builder.rs
```diff
@@ -575,6 +577,33 @@ impl ComponentTypesBuilder {
...
+        abi.flat_count = element_abi
+            .flat_count
+            .map(|c| c.saturating_mul(size.min(255) as u8));
```
alexcrichton: This sould be an `.and_then` I think, right? Plus a `checked_mul` and/or `try_into`?

alexcrichton (approved):
Would it be possible to add a sample `*.wast` test which exercises this?

crates/environ/src/component/types_builder.rs
cpetig: I don't think so, so its probably better to use a saturating multiply here. I don't think that making the function fallible is worth the effort?

alexcrichton: Hm ok I agree that this doesn't need to be fallible, but we should probably add a reasonable size limits in wasm-tools as well, e.g. 1G or something like that, to ensure that calculations don't overflow

cpetig:
> Would it be possible to add a sample `*.wast` test which exercises this?

I guess it would be possible to make a two sub-component example which externally only offers a single boolean returning function, anything beyond this will depend on https://github.com/bytecodealliance/wasmtime/pull/12315 (e.g. passing an array to a function and checking the round-trip result).

Judging from the amount of errors uncovered by a single wit-bindgen test I think `list-minmax16: func(a: list<u16, 4>, b: list<s16, 4>) -> tuple<list<u16, 4>, list<s16, 4>>;` is the signature to test here :wink: (arguments are on stack, result is in memory and the tuple exercises offsets)

alexcrichton:
That sounds reasonable to me yeah, and I think it's fine to expand the test coverage over time but I also think it'd be good to test what we can in this PR itself (which is, as you say, the composed-in-the-middle use)

crates/environ/src/component/types_builder.rs
cpetig: https://github.com/bytecodealliance/wasm-tools/issues/2416

crates/environ/src/component/types_builder.rs
cpetig: `element_abi.flat_count .zip(u8::try_from(size).ok()) .and_then(|(flat_count, size)| flat_count.checked_mul(size))` doesn't feel ideal, but should be closer to what you proposed.

cpetig:
I added a moderately readable test (created in C++, then stripped down to 400 lines of wat).

Though, I was unable to get `;;! component_model_fixed_size_lists = true` recognized by the wast test runner. 

Could it be that it usually sets properties according to the path of the test? Or is it supposed to work and I just missed something?

cpetig:
Oh, that is a new to me error in CI: 
```
thread '<unnamed>' (24806) panicked at crates/environ/src/component/translate/adapt.rs:245:14:
invalid adapter module generated: multiple memories (at offset 0x5d)
``` 
I guess I need to take a deeper look tomorrow.

PS: The error on my computer is
```
Error: failed to run script file 'fixed_size_list.wast'

Caused by:
    0: failed directive on fixed_size_list.wast:17
    1: failed to parse WebAssembly module
    2: Fixed size lists require the component model fixed size list feature (at offset 0x16)
```

`wasmtime wast -W component-model-fixed-size-lists  fixed_size_list.wast` works for me...

alexcrichton:
Oh for that you'll want to add `multi_memory = true` at the top of the file. Not needed on the CLI because that inherits the default feature set, but during tests we turn off all features and require them to be explicitly turned on in files

cpetig:
> Oh for that you'll want to add `multi_memory = true` at the top of the file. Not needed on the CLI because that inherits the default feature set, but during tests we turn off all features and require them to be explicitly turned on in files

My assumption was that this is caused by the component to component copying code https://github.com/bytecodealliance/wasmtime/pull/10619/files#diff-32f96d31ce61f8177ede00faf17bc565a8d15a00d42eb20b28a948a7a257dd87R2885-R2953 - I would assume that all adapters need multi-memory to work.

cpetig:
> My assumption was that this is caused by the component to component copying code https://github.com/bytecodealliance/wasmtime/pull/10619/files#diff-32f96d31ce61f8177ede00faf17bc565a8d15a00d42eb20b28a948a7a257dd87R2885-R2953 - I would assume that all adapters need multi-memory to work.

I was looking for `wasmtime wast` to pick up the setting and didn't realize that the test runner takes care of this. Should work now. Thank you for the tip, Alex.

cpetig:
:blush: Oh, I took a look whether "fixed-size-list" or "fixed-size-lists" is the name in the component spec, it is "fixed-length lists". So, I made sure that wasmtime adheres to the standard name. 

Whether/how to change wasm-tools will be a separate discussion.

alexcrichton (approved):

PR #10628, opened 2025-04-21
add #10610 and #10597 to RELEASES.md

pchickey:
I know we usually do these all at the end, but I wanted to write these up myself while I still had them fresh.


alexcrichton (approved):
Very much appreciated, thanks!

PR #10629, opened 2025-04-21
asm: plumb fixed registers through

abrown:
This change introduces a new `Fixed` wrapper type that allows us to plumb fixed registers through the assembler. These instructions are not yet used by `cranelift-codegen`, but will be necessary at some point as described in [#10238]. The end result of this is that up at the `cranelift-codegen` level, fixed registers should look like run-of-the-mill virtual registers using the appropriate types, but down in `cranelift-assembler-x64` we panic if the register allocator does not give them the correct register.

[#10238]: https://github.com/bytecodealliance/wasmtime/issues/10238


cranelift/assembler-x64/src/fixed.rs
```diff
@@ -0,0 +1,44 @@
...
+/// # use cranelift_assembler_x64::{AsReg, Fixed, gpr};
+/// # let invalid_reg = 42;
+/// let fixed = Fixed::<u8, { gpr::enc::RAX }>(invalid_reg);
+/// fixed.enc(); // Will panic because `invalid_reg` does not match `RAX`.
+/// ```
+#[derive(Clone, Debug)]
+pub struct Fixed<R, const E: u8>(pub R);
+
+impl<R: AsReg, const E: u8> AsReg for Fixed<R, E> {
+    fn new(reg: u8) -> Self {
```
cfallin, line 29: Is there a reason (e.g. generic interface somewhere else) that we need to accept `reg` here as an argument, rather than just using the generic parameter `E` -- since there is only one valid value to pass in?

cfallin: Ah, I see below, this enables the idiom `Fixed(rax)` for example, where type inference carries `reg` through to the generic parameter.

Maybe there could be two entry points, `new()` and `from_const_reg(reg: u8)` or similar? Maybe not a huge deal; just trying to reduce confusing APIs.

cranelift/codegen/src/isa/x64/inst/mod.rs
```diff
@@ -57,7 +57,7 @@ pub struct ReturnCallInfo<T> {
 fn inst_size_test() {
     // This test will help with unintentionally growing the size
     // of the Inst enum.
-    assert_eq!(56, std::mem::size_of::<Inst>());
+    assert_eq!(48, std::mem::size_of::<Inst>());
```
cfallin, line 60: Nice! This is presumably because a reg that we carried as a (always the same value) `u32` previously now becomes a type parameter? Good to see reductions in any case.

cfallin (approved):
Looks good, thanks! Just some nits below.

cranelift/assembler-x64/src/fixed.rs
abrown: I don't think we need a different entry point and, in fact, I've been trying to remove this `AsReg::new` somehow. It's only during fuzzing that we need the ability to construct an `AsReg`. I just can't seem to pull the right Jenga blocks out to get rid of this without things falling down.

cfallin: Ah, I see -- a few thoughts then:

- At the very least, `pub(crate) fn new(...)` so it's only exposed to where fuzzing needs it (the `Arbitrary` impl);
- Perhaps removing the field (so it's a zero-sized struct, carrying `E` only at the type level); you still have `E` for `enc()` to work; the only part I'm not as clear on is where `AsRef` comes in and why you need a reference (vs. constructing an `R` on-the-fly as needed, as e.g. `fn to_reg(&self) -> R { R::new(E) }` or something like that)

abrown: > * At the very least, `pub(crate) fn new(...)` so it's only exposed to where fuzzing needs it (the `Arbitrary` impl);

What I meant is that I want to remove the trait function `AsReg::new`--we can't make trait functions `pub(crate)`, right?

> * Perhaps removing the field (so it's a zero-sized struct, carrying `E` only at the type level); you still have `E` for `enc()` to work; the only part I'm not as clear on is where `AsRef` comes in and why you need a reference (vs. constructing an `R` on-the-fly as needed, as e.g. `fn to_reg(&self) -> R { R::new(E) }` or something like that)

I believe we still need the field because we need a place for the virtual register to live until register allocation happens.

cfallin: Oh, d'oh, I had missed that `new` was part of the trait definition too, sorry! And you're of course right that we need an actual vreg too.

PR #10630, opened 2025-04-21
Add a bunch of cookbook-style "how do I do X?" examples to the book

fitzgen:
This tries to cover things that we repeatedly get questions about on Zulip.


pchickey (approved):

pchickey:
Thanks Nick! This is super helpful

PR #10632, opened 2025-04-22
asm: change how regalloc interacts with operands

abrown:
This PR changes the paradigm for how a user (like `regalloc2`) will interact with the operands of an assembler instruction. Instead of using a visitor pattern, calling functions for each of the operands, this approach builds up a list of operands that expose `&mut` references to the underlying operands of the instruction.

While this does have the advantage of being a bit more straightforward, the motivation for this is more subtle: some instructions have special register constraints that the visitor pattern does not handle. E.g., when switching `AluConstOp` to use the new assembler, I noticed that the calls it makes to the register allocator are quite different than the visitor expects: `AluConstOp` boils down to `xor %r*, %r*` and expects us a completely different call to `regalloc2`: a single `reg_def()`. We would need to build a completely different visitor implementation for this.

Also, this refactoring makes it possible for users of the assembler to examine and modify any operands to any instruction. This should be useful regardless of the `AluConstOp` motivation.

### DIFF
--- a/cranelift/assembler-x64/meta/src/generate.rs
+++ b/cranelift/assembler-x64/meta/src/generate.rs
@@ -14,7 +14,7 @@ pub fn rust_assembler(f: &mut Formatter, insts: &[dsl::Inst]) {
     generate_inst_enum(f, insts);
     generate_inst_display_impl(f, insts);
     generate_inst_encode_impl(f, insts);
-    generate_inst_visit_impl(f, insts);
+    generate_inst_operands_impl(f, insts);
     generate_inst_features_impl(f, insts);
 
     // Generate per-instruction structs.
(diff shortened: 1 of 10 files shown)


cranelift/assembler-x64/meta/src/generate.rs
```diff
@@ -102,24 +102,18 @@ fn generate_inst_encode_impl(f: &mut Formatter, insts: &[dsl::Inst]) {
...
-                fmtln!(f, "Self::{variant_name}(i) => i.visit(v),");
-            }
-            f.indent_pop();
-            fmtln!(f, "}}");
-        });
-        fmtln!(f, "}}");
+/// `impl Inst { fn operands... }`
+fn generate_inst_operands_impl(f: &mut Formatter, insts: &[dsl::Inst]) {
+    f.add_block("impl<R: Registers> Inst<R>", |f| {
+        f.add_block("pub fn operands<'a>(&'a mut self) -> Vec<Operand<'a, R>>", |f| {
```
abrown, line 108: We _could_ use a `smallvec` with a threshold of 4 and be a bit more efficient but I thought, for now, this was fine.

cfallin:
Hmm -- I have some serious concerns with this approach, and in particular, constructing a `Vec` of operands for each instruction during each visitor pass. We've worked hard to reduce allocations during compilation, and to operate on data in-place as much as possible; I would expect this to have a pretty nontrivial impact on all that, especially in the fullness of time when the migration of all instructions to the new assembler is complete. Basically: the "mutate in place" was hard-won, let's not give it up.

I suspect/hope there is an easier way to address the "differing constraints for some instructions" problem. In the xor case, we could adopt the approach we took with XMM zeroing, namely to have an "uninitialized value" virtual def instruction that assembles to nothing and then always emit that prior to (and only in the case of) an xor-zeroing idiom. What do you think?

abrown:
The now-merged #10648 takes a different approach to this problem.

PR #10633, opened 2025-04-22
asm: use `Formatter::add_block` for more generated code

abrown:
This propagates a helpful pattern from #10348 in more places; `Formatter::add_block` is a bit more clear and prevents some classes of generation mistakes.


abrown:
This should all be non-functional changes; I covered some of this in #10632 but this is the rest of the refactoring.

fitzgen (approved):

PR #10644, opened 2025-04-22
asm: plumb fixed registers through operand visitor

abrown:
This finishes plumbing out fixed registers using the visitor pattern in
`cranelift-codegen`. This was skipped initially in [#10629] because
[#10632] offered a different approach; while we discuss whether and how
that should proceed, this change at least fixes up the current state of
things.

[#10629]: https://github.com/bytecodealliance/wasmtime/pull/10629
[#10632]: https://github.com/bytecodealliance/wasmtime/pull/10632


cfallin (approved):
LGTM!

PR #10645, opened 2025-04-22
wasmtime serve: when handle_request fails, give a 500 response

pchickey:
Without this 500 response, hyper closes the connection without sending a response. Sending a 500 is better http hygiene.

### DIFF
--- a/src/commands/serve.rs
+++ b/src/commands/serve.rs
@@ -1,6 +1,8 @@
 use crate::common::{Profile, RunCommon, RunTarget};
 use anyhow::{anyhow, bail, Result};
 use clap::Parser;
+use http::{Response, StatusCode};
+use std::convert::Infallible;
 use std::net::SocketAddr;
 use std::time::Instant;
 use std::{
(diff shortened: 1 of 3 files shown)


alexcrichton (approved):
Mind adding a test [around here](https://github.com/bytecodealliance/wasmtime/blob/5f3aefa1570c55cf1a54425637be84b707b95f37/tests/all/cli_tests.rs#L1677) for this as well?

Also would it perhaps make sense to render the error in the response too?

pchickey:
CI revealed where existing tests covered the old behavior, so I updated that, rather than adding a test.

I don't think anyone should be using wasmtime-cli as a production server, but inevitably someone will, and in that case I don't want to put the crash information in the 500 response in case it gets exposed to untrusted users.

src/commands/serve.rs
```diff
@@ -432,7 +434,28 @@ impl ServeCommand {
...
+                                    unreachable!()
+                                }
+                                match handle_request(h, req, comp).await {
+                                    Ok(r) => Ok::<_, Infallible>(r),
+                                    Err(e) => {
+                                        eprintln!("error: {e:?}");
+                                        Ok(Response::builder()
+                                            .status(StatusCode::INTERNAL_SERVER_ERROR)
+                                            .body(
+                                                Full::new(bytes::Bytes::new())
```
bjorn3, line 451: Shouldn't this at show a basic html page with a user visible message that an internal server error happened like Apache and Nginx do? A completely empty page is even more confusing for end users than the browser telling the user that the connection got closed.

bjorn3: Opened https://github.com/bytecodealliance/wasmtime/pull/10851

PR #10648, opened 2025-04-22
asm: clear GPR with `xor <tmp>, <tmp>`

abrown:
This change uses an "uninitialized value" to resolve the register allocation issues that come up when clearing a GPR with `xor`; this tries out the approach suggested in #10632 and will replace that if merged. The problem is that, in `xor <tmp>, <tmp>`, we need some way to communicate to the register allocator that the previous values of `<tmp>` are completely cleared--i.e., not used. While #10632 started down the path of allowing the register allocator to make a different decision about `<tmp>`--the approach currently taken for `AluConstOp`--this change instead adopts the "uninitialized value" approach. As was done previously with `XmmUninitializedValue`, we create a new pseudo-instruction called `GprUninitializedValue` whose registers are defined out of thin air.

### DIFF
--- a/cranelift/codegen/src/isa/x64/inst/emit.rs
+++ b/cranelift/codegen/src/isa/x64/inst/emit.rs
@@ -3350,7 +3350,10 @@ pub(crate) fn emit(
 
         Inst::XmmUninitializedValue { .. } | Inst::GprUninitializedValue { .. } => {
             // These instruction formats only exist to declare a register as a
-            // `def`; no code is emitted.
+            // `def`; no code is emitted. This is always immediately followed by
+            // an instruction, such as `xor <tmp>, <tmp>`, that semantically
+            // reads this undefined value but arithmetically produces the same
+            // result regardless of its value.
         }
 
         Inst::XmmMovRM { op, src, dst } => {
--- a/cranelift/codegen/src/isa/x64/inst/emit.rs
+++ b/cranelift/codegen/src/isa/x64/inst/emit.rs
@@ -3348,9 +3348,9 @@ pub(crate) fn emit(
             sink.put1(*imm);
         }
 
-        Inst::XmmUninitializedValue { .. } => {
-            // This instruction format only exists to declare a register as a `def`; no code is
-            // emitted.
+        Inst::XmmUninitializedValue { .. } | Inst::GprUninitializedValue { .. } => {
+            // These instruction formats only exist to declare a register as a
+            // `def`; no code is emitted.
         }
 
         Inst::XmmMovRM { op, src, dst } => {
--- a/tests/disas/load-store/x64/load_store_static_kind_i32_index_0_guard_yes_spectre_i32_access_0_offset.wat
+++ b/tests/disas/load-store/x64/load_store_static_kind_i32_index_0_guard_yes_spectre_i32_access_0_offset.wat
@@ -21,33 +21,33 @@
 ;; wasm[0]::function[0]:
 ;;       pushq   %rbp
 ;;       movq    %rsp, %rbp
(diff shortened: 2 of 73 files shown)


cranelift/codegen/src/isa/x64/inst/emit.rs
```diff
@@ -3348,9 +3348,9 @@ pub(crate) fn emit(
             sink.put1(*imm);
         }
 
-        Inst::XmmUninitializedValue { .. } => {
-            // This instruction format only exists to declare a register as a `def`; no code is
-            // emitted.
+        Inst::XmmUninitializedValue { .. } | Inst::GprUninitializedValue { .. } => {
+            // These instruction formats only exist to declare a register as a
+            // `def`; no code is emitted.
```
cfallin, line 3353: I know pre-existing, but can we add a bit more here about how it's used (to ease concerns of the reader that this looks wildly unsafe)? Basically something like: "This is always immediately followed by an instruction, such as `xor reg, reg`, that *semantically* reads this undefined value but *arithmetically* produces the same result regardless of its value."

cfallin (approved):
LGTM, thanks!

tests/disas/load-store/x64/load_store_static_kind_i32_index_0_guard_yes_spectre_i32_access_0_offset.wat
```diff
@@ -21,33 +21,33 @@
...
+;;       movl    %ecx, (%r10)
 ;;       movq    %rbp, %rsp
 ;;       popq    %rbp
 ;;       retq
-;;   24: addb    %al, (%rax)
-;;   26: addb    %al, (%rax)
-;;   28: cld
+;;   23: addb    %al, (%rax)
+;;   25: addb    %al, (%rax)
+;;   27: addb    %bh, %ah
```
abrown, line 36: I do have some questions about the offset changes here (hence the draft status): here, for example, we've managed to shave a byte off our encodings somehow...

tests/disas/pcc-imported-memory.wat
```diff
@@ -37,22 +37,22 @@
...
-;;   3b: xorl    %r10d, %r10d
+;;       movq    0x38(%rdi), %rdx
+;;       movq    8(%rdx), %rcx
+;;       shrq    $0x10, %rcx
+;;       movq    %rcx, %rdx
+;;       shll    $0x10, %edx
+;;       leal    4(%rcx), %r8d
+;;       cmpl    %r8d, %edx
+;;       jbe     0x3d
+;;   23: testl   %ecx, %ecx
```
abrown, line 48: ...whereas here we've managed to gain two of them! It's not yet clear to me why this is.

tests/disas/winch/x64/epoch/func.wat
```diff
@@ -23,7 +23,7 @@
 ;;       cmpq    %rcx, %rdx
 ;;       jb      0x51
 ;;   44: movq    %r14, %rdi
-;;       callq   0x126
+;;       callq   0x127
```
abrown, line 26: And here, we've somehow lost a byte again, but without any other changes to the file?

abrown:
In the Cranelift meeting today we chalked these offset changes up to different register allocation (potentially adding or removing REX bytes in a few places) &mdash; no one sounded too concerned. We did discuss another approach: instead of continuing to rely on the `uninit` footgun, we could create a pseudo-instruction, much like the current `AluConstOp`, to differentiate this `xor <tmp>, <tmp>` pattern for register allocation. (I have built a branch that already walks out that approach: https://github.com/bytecodealliance/wasmtime/compare/main...abrown:wasmtime:asm-const).

PR #10652, opened 2025-04-23
Add inital support for `f16` without `Zfh` and `f128` to the riscv64 backend

beetrees:
On the riscv64 backend, this PR adds initial support for using `f16` without the `Zfh` extension being enabled, as well as initial support for `f128`. Support is added for the `load`, `store`, `bitcast`, `select`, `f16const`, `f128const`, `bnot`, `band`, `bor` and `bxor` CLIF instructions, as well as adding the `zfhmin` and `zvfh` target features (alongside the pre-existing `zfh` target feature).

cc @afonso360 who previously added initial `f16` support in #9135

`f16`/`f128` issue: #8312


cranelift/codegen/meta/src/isa/riscv64.rs
```diff
@@ -64,6 +64,13 @@ pub(crate) fn define() -> TargetIsa {
         false,
     );
 
+    let _has_zfhmin = setting.add_bool(
+        "has_zfhmin",
+        "has extension Zfhmin?",
+        "Zfmin: Minimal Half-Precision Floating-Point",
```
bjorn3, line 70:
```suggestion
        "Zfhmin: Minimal Half-Precision Floating-Point",
```

cranelift/codegen/meta/src/isa/riscv64.rs
```diff
@@ -78,6 +85,13 @@ pub(crate) fn define() -> TargetIsa {
         false,
     );
 
+    let _has_zvfh = setting.add_bool(
+        "has_zvfh",
+        "has extension Zvfh?",
+        "Zfmin: Vector Extension for Half-Precision Floating-Point",
```
bjorn3, line 91:
```suggestion
        "Zvfh: Vector Extension for Half-Precision Floating-Point",
```

cranelift/codegen/src/isa/riscv64/inst.isle
```diff
@@ -1856,15 +1881,14 @@
...
+  (if-let true (has_fli_for_type ty))
   (if-let const (fli_constant_from_negated_u64 ty imm))
   (rv_fneg ty (rv_fli ty const)))
 
 ;; Otherwise floats get loaded as integers and then moved into an F register.
-(rule 6 (imm (ty_supported_float $F16) c) (gen_bitcast (imm $I16 c) $I16 $F16))
-(rule 6 (imm (ty_supported_float $F32) c) (gen_bitcast (imm $I32 c) $I32 $F32))
-(rule 6 (imm (ty_supported_float $F64) c) (gen_bitcast (imm $I64 c) $I64 $F64))
+(rule 6 (imm (ty_supported_float_min ty) c) (gen_bitcast (imm (float_int_of_same_size ty) c) (float_int_of_same_size ty) ty))
+(rule 5 (imm (ty_supported_float_size $F16) c) (gen_bitcast (imm $I32 (u64_or c 0xffff0000)) $I32 $F32))
```
cfallin, line 1891: Could we add a comment here describing why we're OR'ing in all ones in the high bits? I'm not sure if I understand why myself -- does this allow more immediates to fit into better/smaller codegen patterns for `imm` perhaps?

cranelift/codegen/src/isa/riscv64/inst.isle
```diff
@@ -2928,30 +2958,33 @@
...
+(rule 6 (gen_bitcast r (ty_supported_vec _) (ty_supported_float_min dst_ty)) (rv_vfmv_fs r dst_ty))
+
+(rule 5 (gen_bitcast r (ty_int_ref_scalar_64 src_ty) (ty_supported_vec _)) (rv_vmv_sx r src_ty))
+(rule 4 (gen_bitcast r (ty_supported_vec _) (ty_int_ref_scalar_64 dst_ty)) (rv_vmv_xs r dst_ty))
+(rule 3 (gen_bitcast r (ty_supported_float_min $F16) $I16) (rv_fmvxh r))
+(rule 2 (gen_bitcast r (ty_supported_float_size $F16) $I16) (rv_fmvxw r))
+(rule 2 (gen_bitcast r (ty_supported_float_size $F32) $I32) (rv_fmvxw r))
+(rule 2 (gen_bitcast r (ty_supported_float_size $F64) $I64) (rv_fmvxd r))
+(rule 1 (gen_bitcast r $I16 (ty_supported_float_min $F16)) (rv_fmvhx r))
+(rule 0 (gen_bitcast r $I16 (ty_supported_float_size $F16)) (rv_fmvwx (rv_or r (imm $I32 0xffff0000))))
```
cfallin, line 2973: Likewise here as above -- where does the constant come from?

cranelift/filetests/filetests/runtests/f128-bitcast.clif
```diff
@@ -3,7 +3,7 @@ test run
 set enable_llvm_abi_extensions
 target x86_64
 target aarch64
-target aarch64 has_fp16
+target riscv64
```
cfallin, line 6: Is there any particular reason that we remove `aarch64 has_fp16` from the list of targets tested by this runtest and those below? Can we keep that while adding `riscv64` below it?

cfallin (commented):
A few nits but overall this looks plausible to me -- thanks for working through the tedious details!

cranelift/codegen/src/isa/riscv64/inst.isle
bjorn3: Does this rule only applies when 16bit float instructions are not supported? If so I think we could get away with only NaN-boxing on the ABI boundary and within the function leaving the high bits undefined. I don't know if that will be faster though.

cranelift/filetests/filetests/runtests/f128-bitcast.clif
cfallin: Ah, right, of course -- saw the seemingly unrelated change but that cleanup makes sense, thanks.

cfallin (approved):
Thanks, LGTM!

PR #10653, opened 2025-04-23
fix(docs): remove $ from console commands for easy copy/paste

vados-cosmonic:
This came up during the recent [`--invoke` blog post PR](https://github.com/bytecodealliance/bytecodealliance.org/pull/126), but IMO it would be better if the documentation made it easy to copy-paste commands rather than including `$` in command-only blocks.

This PR does a couple things to make docs a bit easier to use:

- Switch from `sh` to `console` for console commands
- Remove `$` from console commands that do not also have output

Generally, console commands that are standalone *should* be easy to copy and paste.

This PR doesn't make the changes everywhere in the docs, but that can be handled in follow-ups.


vados-cosmonic:
hey @pchickey I think it had a bad time with just a plain block (the `Options:` bit) -- changed it to `console`

PR #10654, opened 2025-04-23
Consolidate building guest programs for tests

alexcrichton:
This commit folds all C/C++ tests for debuginfo into the same location as Rust programs compiled to wasm. That means that all of `tests/all/debug/testsuite/*` is moved to
`crates/test-programs/src/bin/*`. Additionally the hardcoded per-test configuration in `crates/test-programs/artifacts/build.rs` was removed in favor of in-test configuration at the top of the file. This means that all precompiled binaries are removed from the tree for DWARF testing.

Various offsets were updated, a test was dropped as it wasn't used, and some commands/directives were updated to account for this adjustment.

prtest:debug


alexcrichton:
cc @SingleAccretion for the affect this has on debuginfo tests

crates/test-programs/artifacts/build.rs
```diff
@@ -1,293 +1,360 @@
...
-    let out_dir = PathBuf::from(env::var_os("OUT_DIR").unwrap());
+struct Artifacts {
+    out_dir: PathBuf,
+    deps: HashSet<String>,
+}
+
+struct Test {
+    /// Not all tests can be built at build-time, for example C/C++ tests require
+    /// the `WASI_SDK_PATH` environment variable which isn't available on all
+    /// machines. The `Option` here encapsulates tests that were unabled to be
```
pchickey, line 27:
```suggestion
    /// machines. The `Option` here encapsulates tests that were not able to be
```

crates/test-programs/artifacts/build.rs
```diff
@@ -1,293 +1,360 @@
...
+    /// built.
+    ///
+    /// For tests that wern't able to be built their error is deferred to
```
pchickey, line 30:
```suggestion
    /// For tests that were not able to be built their error is deferred to
```

pchickey (approved):
LGTM, but also a good idea to have someone with deeper understanding of the C/C++ tests spot check.

SingleAccretion (approved):
Changes LGTM. FWIW, I tend[ed] to view the hardcoded compile options as an explicit benefit due to fewer indirections required for understanding of where they come from. But this version is also completely fine.

tests/all/debug/translate.rs
```diff
@@ -4,9 +4,11 @@ use anyhow::{format_err, Result};
 use filecheck::{CheckerBuilder, NO_VARIABLES};
 use std::fs::read;
 use tempfile::NamedTempFile;
+use test_programs_artifacts::*;
 
 #[allow(dead_code)]
 fn check_wasm(wasm_path: &str, directives: &str) -> Result<()> {
+    println!("check {wasm_path}");
```
dicej, line 11: Was this meant to be included, or was it temporarily added for debugging?

dicej (approved):

PR #10657, opened 2025-04-23
Switch to using Wasmtime-style builtins for ceil, floor, etc.

sunfishcode:
With this patch, all emitted calls are Wasmtime-style builtins, rather than Cranelift-style libcalls. This ensures that all calls from Cranelift-generated code into Wasmtime host code use the same mechanism, and eliminates the relocation handling code for the libcall mechanism.

This is split out from #9093.

### DIFF
--- a/crates/wasmtime/src/runtime/vm/interpreter.rs
+++ b/crates/wasmtime/src/runtime/vm/interpreter.rs
@@ -342,8 +342,8 @@ impl InterpreterRef<'_> {
             (@set bool $reg:ident $val:ident) => (self.0[$reg].set_i32(i32::from($val)));
             (@set u32 $reg:ident $val:ident) => (self.0[$reg].set_u32($val));
             (@set u64 $reg:ident $val:ident) => (self.0[$reg].set_u64($val));
-            (@set f32 $reg:ident $val:ident) => (self.0[$reg].set_f32($val));
-            (@set f64 $reg:ident $val:ident) => (self.0[$reg].set_f64($val));
+            (@set f32 $reg:ident $val:ident) => (unreachable::<f32, _>(($reg, $val)));
+            (@set f64 $reg:ident $val:ident) => (unreachable::<f64, _>(($reg, $val)));
             (@set i8x16 $reg:ident $val:ident) => (unreachable::<i8x16, _>(($reg, $val)));
             (@set pointer $reg:ident $val:ident) => (self.0[$reg].set_ptr($val));
             (@set size $reg:ident $val:ident) => (self.0[$reg].set_ptr($val as *mut u8));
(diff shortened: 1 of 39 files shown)


pulley/src/interp.rs
```diff
@@ -472,6 +472,16 @@ impl XRegVal {
...
 
+    pub fn get_f32(&self) -> f32 {
+        let x = unsafe { self.0.u32 };
+        f32::from_bits(u32::from_le(x))
+    }
+
+    pub fn get_f64(&self) -> f64 {
+        let x = unsafe { self.0.u64 };
+        f64::from_bits(u64::from_le(x))
+    }
```
alexcrichton, line 483: This is technically incorrect because floats are stored in `FRegVal`, not `XRegVal`. I see this is added here for the various macro machinations for crossing between Pulley and the host. Could this be dropped coupled with the suggestion above as well?

crates/wasmtime/src/runtime/vm/interpreter.rs
```diff
@@ -319,6 +325,9 @@ impl InterpreterRef<'_> {
             (@get u8 $reg:ident) => (self.0[$reg].get_i32() as u8);
             (@get u32 $reg:ident) => (self.0[$reg].get_u32());
             (@get u64 $reg:ident) => (self.0[$reg].get_u64());
+            (@get f32 $reg:ident) => (self.0[$reg].get_f32());
+            (@get f64 $reg:ident) => (self.0[$reg].get_f64());
```
alexcrichton, line 329: In combination with dropping these methods below, could this be changed to `unreachable()` in the same manner as `i8x16`? Fixing this would otherwise require updates to the macro to have a separate "finger" for float registers as it has for integer registers right now, so this'll need more updates if floats are ever passed here.

alexcrichton (approved):

pulley/src/interp.rs
sunfishcode: Done.

crates/wasmtime/src/runtime/vm/interpreter.rs
sunfishcode: Done.

sunfishcode:
The failure was:

> 8.538 E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/liba/libarchive/libarchive13_3.6.0-1ubuntu1.4_amd64.deb  404  Not Found [IP: 91.189.91.83 80]
8.538 E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?

so, trying again.

pulley/src/interp.rs
```diff
@@ -493,6 +493,14 @@ impl XRegVal {
...
     }
 
+    pub fn set_f32(&mut self, x: f32) {
+        self.0.u32 = x.to_bits().to_le();
+    }
+
+    pub fn set_f64(&mut self, x: f64) {
+        self.0.u64 = x.to_bits().to_le();
+    }
+
```
alexcrichton, line 503: Mind also deleting these setters too? (fine to do in a follow-up too)

sunfishcode: Done.

PR #10659, opened 2025-04-23
x64: remove `AluConstOp`

abrown:
After #10648, this `Inst` variant is no longer used.

### DIFF
--- a/cranelift/codegen/src/isa/x64/pcc.rs
+++ b/cranelift/codegen/src/isa/x64/pcc.rs
@@ -173,16 +173,6 @@ pub(crate) fn check(
             RegMem::Reg { .. } => undefined_result(ctx, vcode, dst, 64, size.to_bits().into()),
         },
 
-        Inst::AluConstOp {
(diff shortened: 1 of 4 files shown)


cranelift/codegen/src/isa/x64/pcc.rs
```diff
@@ -173,16 +173,6 @@ pub(crate) fn check(
...
 
-        Inst::AluConstOp {
-            op: AluRmiROpcode::Xor,
-            dst,
-            ..
-        } => check_output(ctx, vcode, dst.to_writable_reg(), &[], |_vcode| {
-            Ok(Some(Fact::constant(64, 0)))
-        }),
-
-        Inst::AluConstOp { dst, .. } => undefined_result(ctx, vcode, dst, 64, 64),
```
abrown, line 184: @cfallin, something to think about: removing this here is unfortunate, but the more unfortunate thing for PCC will be the `AluRmiR` removal, coming soon.

Issue #10638, opened 2025-04-22
[wasmtime]: Construct CacheConfig programatically instead of from file


benbrandt:
#### Feature

`wasmtime::Config::cache_config` for loading a `CacheConfig` from a struct instead of a file.

#### Benefit

Right now, in order to use the cache, you have to provide a path to a file with the needed configuration.

However, there are cases where the wasmtime is embedded, rather than the CLI, where the cache is still helpful. It would be more convenient to create the configuration programmatically rather than loaded from a file, to allow the application to determine the cache setup based on its own configuration.

#### Implementation

1. Expose `wasmtime_cache::CacheConfig` from `wasmtime`

2. Add setters for `wasmtime_cache::CacheConfig` so individual fields can be configured.

```rust
let mut cache_config = wasmtime_cache::CacheConfig::default();
cache_config.with_directory("/path/to/cache");
```

3. Add `cache_config` method to set this config on the `wasmtime::Config`

```rust
let mut config = wasmtime::Config::default();
config.cache_config(cache_config);
```

#### Alternatives

Based on [the implementation for the method to load from a file](https://docs.rs/wasmtime/latest/src/wasmtime/config.rs.html#1369-1372), this seems like a straightforward way to allow for a wasmtime user to create their own cache.

If there are other ideas, we could explore those instead.

benbrandt:
I forgot to mention, but I am open to doing the implementation (with any necessary pointers), but wanted to get feedback on the idea before submitting the PR

alexcrichton:
I think this'd be a great idea to have, thanks for filing an issue and working on this!

The only thing I'd suggest is that we wouldn't want users to have to depend on the `wasmtime-cache` crate itself, so we'd reexport as `wasmtime::CacheConfig` to have accessible from the main `wasmtime` crate. Otherwise though  definitley agreed it'd be best to support progammatic configuration instead of exclusively file-based configuration.

PR #10665, opened 2025-04-24
Allow creation of a CacheConfig without loading from a file

benbrandt:
Addresses issue #10638

Since creating a `CacheConfig` also creates the worker thread, I opted for a Builder pattern, so that this worker thread, validation, and default values could be created in the `build` method, much like what happens when deserializing from a file.

If there is a preferred alternate approach though, let me know.

One thing that came to mind is that we may make it easier to create these worker threads, because it is easier to construct the `CacheConfig`, but since these threads get cleaned up once the `CacheConfig` is dropped, maybe this is not an issue?

I'm also not sure if there is any danger in having multiple `CacheConfig`s pointing to the same directory in terms of the worker threads. But this issue could have existed with multiple `Config` with the same default config, so I assume this is not a huge issue.

If there are any changes needed, let me know, especially around coding guidelines or best testing practices.

Thanks!

### DIFF
--- a/crates/cache/src/config.rs
+++ b/crates/cache/src/config.rs
@@ -37,6 +37,9 @@ macro_rules! generate_config_setting_getter {
 impl Cache {
     /// Builds a [`Cache`] from the configuration and spawns the cache worker.
     ///
+    /// If you want to load the cache configuration from a file, use [`CacheConfig::from_file`].
+    /// You can call [`CacheConfig::new`] for the default configuration.
+    ///
     /// # Errors
     /// Returns an error if the configuration is invalid.
     pub fn new(mut config: CacheConfig) -> Result<Self> {
@@ -46,6 +49,28 @@ impl Cache {
             config,
             state: Default::default(),
         })
+    }
+
+    /// Loads cache configuration specified at `path`.
+    ///
+    /// This method will read the file specified by `path` on the filesystem and
+    /// attempt to load cache configuration from it. This method can also fail
+    /// due to I/O errors, misconfiguration, syntax errors, etc. For expected
+    /// syntax in the configuration file see the [documentation online][docs].
+    ///
+    /// Passing in `None` loads cache configuration from the system default path.
+    /// This is located, for example, on Unix at `$HOME/.config/wasmtime/config.toml`
+    /// and is typically created with the `wasmtime config new` command.
+    ///
+    /// # Errors
+    ///
+    /// This method can fail due to any error that happens when loading the file
+    /// pointed to by `path` and attempting to load the cache configuration.
+    ///
+    /// [docs]: https://bytecodealliance.github.io/wasmtime/cli-cache.html
+    pub fn from_file(path: Option<&Path>) -> Result<Self> {
+        let config = CacheConfig::from_file(path)?;
+        Self::new(config)
     }
 
     generate_config_setting_getter!(worker_event_queue_size: u64);
--- a/crates/cache/src/config.rs
+++ b/crates/cache/src/config.rs
@@ -18,51 +18,49 @@ use std::time::Duration;
 /// Global configuration for how the cache is managed
 #[derive(Debug, Clone)]
 pub struct Cache {
-    directory: PathBuf,
-    worker_event_queue_size: u64,
-    baseline_compression_level: i32,
-    optimized_compression_level: i32,
-    optimized_compression_usage_counter_threshold: u64,
-    cleanup_interval: Duration,
-    optimizing_compression_task_timeout: Duration,
-    allowed_clock_drift_for_files_from_future: Duration,
-    file_count_soft_limit: u64,
-    files_total_size_soft_limit: u64,
-    file_count_limit_percent_if_deleting: u8,
-    files_total_size_limit_percent_if_deleting: u8,
+    config: CacheConfig,
     worker: Worker,
     state: Arc<CacheState>,
 }
 
-macro_rules! generate_setting_getter {
+macro_rules! generate_config_setting_getter {
     ($setting:ident: $setting_type:ty) => {
         /// Returns `$setting`.
         ///
         /// Panics if the cache is disabled.
         pub fn $setting(&self) -> $setting_type {
-            self.$setting
+            self.config.$setting()
         }
     };
 }
 
 impl Cache {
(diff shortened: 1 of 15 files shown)


crates/cache/src/config.rs
```diff
@@ -301,6 +301,10 @@ impl CacheConfig {
     generate_setting_getter!(file_count_limit_percent_if_deleting: u8);
     generate_setting_getter!(files_total_size_limit_percent_if_deleting: u8);
 
+    pub fn builder() -> CacheConfigBuilder {
```
alexcrichton, line 304: Mind adding some docs to this method since it'll be public-facing in the `wasmtime` crate API?

crates/cache/src/config.rs
```diff
@@ -579,6 +586,187 @@ impl CacheConfig {
...
+    allowed_clock_drift_for_files_from_future: Option<Duration>,
+    file_count_soft_limit: Option<u64>,
+    files_total_size_soft_limit: Option<u64>,
+    file_count_limit_percent_if_deleting: Option<u8>,
+    files_total_size_limit_percent_if_deleting: Option<u8>,
+}
+
+impl CacheConfigBuilder {
+    /// Specifies whether the cache system is used or not.
+    pub fn enabled(mut self, enabled: bool) -> Self {
```
alexcrichton, line 609: If you're ok I'd prefer to leave this not exposed, it's a bit weird to me to go through the rigamarole of creating a config just to say it's disabled. In that sense I think we should consider `CacheConfigBuilder` as "always create something with the `enabled` flag set to `true`".

crates/cache/src/config.rs
```diff
@@ -579,6 +586,187 @@ impl CacheConfig {
...
+        self.enabled = enabled;
+        self
+    }
+
+    /// Specifies where the cache directory is. Must be an absolute path.
+    pub fn directory(mut self, directory: impl Into<PathBuf>) -> Self {
```
alexcrichton, line 615: While we don't have hard-and-fast conventions in Wasmtime I at least personally prefer builders to take `&mut self` and return `&mut Self` (like `Config` methods). Would you be up to changing to that pattern?

crates/cache/src/config.rs
```diff
@@ -579,6 +586,187 @@ impl CacheConfig {
...
+    /// exceed files-total-size-soft-limit * files-total-size-limit-percent-if-deleting.
+    ///
+    /// This doesn't include files with metadata. To learn more, please refer to the cache system
+    /// section.
+    pub fn files_total_size_limit_percent_if_deleting(mut self, percent: u8) -> Self {
+        self.files_total_size_limit_percent_if_deleting = Some(percent);
+        self
+    }
+
+    pub fn build(self) -> Result<CacheConfig> {
```
alexcrichton, line 729: Mind adding some documentation here? I'm thinking specifically about some error cases for example.

Additionally if you're ok with the `&mut self` suggestion above, this would change to either `&self` (ideally) or `&mut self` if necessary.

alexcrichton (commented):
Thanks again for this! FWIW the caching system hasn't really been substantially touched since its inception in 2019 so while reading over this I've noticed some other things that I think might be worth implementing, but I also don't want to hitch you with unnecessary work you don't have time to do. In that sense I'd like to see if you're interested in some of these adjacent cleanups perhaps? If not no worries we can go ahead and land this and I can work on some cleanups later.

* All `Config::cache_*` methods I think can now be deleted in favor of your new `cache_config` method  with a minor adjustment to take `Option<CacheConfig>` where `None` means "disable". We could then delegate file parsing and "load default" to methods on `CacheConfig` which are already there as `CacheConfig::from_file`
* For the worker I think ideally creating a `Config` wouldn't actually spawn anything. Instead I think it would be best to defer that to `Engine`-creation time. For example we'd have some code in the engine creation where it'd call `config.cache_config.spawn()` or similar.
* To mirror configuration in Wasmtime we could perhaps rename `wasmtime_cache::CacheConfig` to `Cache` and then rename `wasmtime_cache::CacheConfigBuilder` to `CacheConfig`. Basically centralizing on "Thing" is the actual thing that was made where "ThingConfig" is configuration to produce the Thing. That means the new method could perhaps be `Config::cache(&mut self, cache: Option<Cache>)`.

Would you be up for those changes as well? I've got a few comments on this change in particular below too as well, and again I'm happy to defer the above improvements for later if you'd prefer to not take that on.

benbrandt:
@alexcrichton thanks for the thorough review! 
I think all of these changes definitely make sense, and also address some weirdness I felt in implementing this. I can give the changes a try and see how far I get!

crates/cache/src/config.rs
benbrandt: Agreed this indeed felt weird as well, I think we can go with an option on the main config and handle it that way

crates/cache/src/config.rs
benbrandt: I had a feeling this one might come up, and I figured it was a 50/50 chance I got it right :) 
Not a problem

benbrandt:
@alexcrichton Ok I think it is ready for review again. I also think this is all a lot, lot nicer than before, as well as being more flexible.

Notes of breaking changes:

- Previous cache methods on `Config` have been removed in favor of a single `cache` method that can be set to `None` to disable it, or passed a constructed `Cache`, either from a file or manually configured
- I removed `enabled` from `CacheConfig` (more controversial)

The latter felt kind of weird to leave as an option, since now there is either a `Cache` or there is not, and it simplified things to not have to worry if it was enabled or not.

However, because of `#[serde(deny_unknown_fields)]`, this will cause existing config files to not work.

If we want to support this old behavior, I am fine with doing so. Users who embed wasmtime will need to update their code regardless, so this might be acceptable. 
However, CLI users would have to fix this themselves.

alexcrichton (approved):
Thanks again for your work on this! All the changes here look reasonable to me, and thinking about it I think it's reasonable to leave `deny_unknown_fields` and drop `enabled = true` as well. That's a pretty old setting which predated the current CLI flags and configuration system, so I think it's reasonable to take this time to bring it forward to the current era.

benbrandt:
@dev-null-undefined good catch! Made a follow up PR here: https://github.com/bytecodealliance/wasmtime/pull/10859

PR #10670, opened 2025-04-24
Avoid using a `TruncF64` libcall in one more place.

sunfishcode:
When `!clif_instruction_traps_enabled()`, the float-to-int conversion would sometimes emit a Cranelift `trunc` instruction directly, without checking `has_round()`, which led to Cranelift emitting a libcall. Fix it by calling the `trunc_f32` and `trunc_f64` routines which do check `has_round()`.

This fixes a fuzzbug from #10657.

### DIFF
--- a/crates/cranelift/src/func_environ.rs
+++ b/crates/cranelift/src/func_environ.rs
@@ -1169,7 +1169,11 @@ impl<'module_environment> FuncEnvironment<'module_environment> {
         };
         let isnan = builder.ins().fcmp(FloatCC::NotEqual, val, val);
         self.trapnz(builder, isnan, ir::TrapCode::BAD_CONVERSION_TO_INTEGER);
-        let val = builder.ins().trunc(val);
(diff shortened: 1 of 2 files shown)


sunfishcode:
What's a good place to add a testcase for this? It needs a Wasm input, a custom target configuration, and it needs to check the generated Cranelift code or machine code.

cfallin:
> What's a good place to add a testcase for this?

We have the [disas tests](https://github.com/bytecodealliance/wasmtime/blob/main/tests/disas/) now; a good example with specific target settings might be [this one](https://github.com/bytecodealliance/wasmtime/blob/main/tests/disas/x64-bmi2.wat) (for the x86-64 BMI2 extensions). We have the env-var-based "blessing" infra here to autogenerate the expectation, too (`WASMTIME_TEST_BLESS=1`).

cfallin (approved):

Issue #10667, opened 2025-04-24
Async Yield may immediately reschedule when using Tokio


posborne:
Credit to @aturon for the discovery...

Today, the `async_yield_impl` behavior triggered an epoch deadline is hit or we run out of fuel does not match that of [`tokio::task::yield_now`](https://docs.rs/tokio/latest/tokio/task/fn.yield_now.html).  They are _mostly_ the same except that with the current implementation that wasm task effectively gets immediately rescheduled (it is not deferred and ends up at the *front* rather than the *back* of the pending queue).  This would not necessarily be the case with all task schedulers but is the case with tokio today in its default configuration with the lifo-slot enabled (see [`tokio::runtime::Builder::disable_lifo_slot`](https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.disable_lifo_slot)).

When targeting tokio, it would be preferred to have the async yield behavior map directly to whatever `tokio::task::yield_now` does.  There seems to be a couple ways to approach this:
1. Feature flagging to enable the direct call.
2. Allow embedder to configure the async yield behavior to allow for `yield_now` to be called.

Secondarily, even if the lifo-slot was disabled, the `yield_now` behavior in tokio has the added benefit of not allowing the yielding task to be immediately stolen as it is first moved to the deferred queue.  If the current worker does not end up picking up other blocking work then we'll end up resuming on the same thread and avoid potential thrashing.

fitzgen:
We jsut talked about this in the Wasmtime biweekly.

Gist was that we already support [calling a callback at epoch interruption points](https://docs.rs/wasmtime/latest/wasmtime/struct.Store.html#method.epoch_deadline_callback) and it makes sense to add an async variation of this callback API. This would allow embedders to call `tokio::task::yield_now()` from that async callback, while keeping the core of wasmtime free of any assumptions of which particular async executor is being used.

PR #10671, opened 2025-04-24
Support custom yield behavior for epoch interrupts

posborne:
As described in
https://github.com/bytecodealliance/wasmtime/issues/10667, the yield future provided in-tree when a epoch callback cannot always trigger the specific behavior desired in async schedulers like tokio.

This change introduces a new variant to the UpdateDeadline callback with the future implementing the desired task suspension behavior desired for yielding control within the async runtime.

Note that this change does not allow for controlling this behavior for fuel yielding at this time.

CC @aturon @fitzgen -- at first I looked at adding a separate async callback but this felt a little cleaner but happy to change things up.  There's definitely some overhead here in needing to frequently do the future boxing so would be happy to look at other paths to optimize if we have concerns (though my gut says it shouldn't be _too_ bad).

### DIFF
--- a/crates/wasmtime/src/runtime/store.rs
+++ b/crates/wasmtime/src/runtime/store.rs
@@ -252,6 +252,7 @@ enum CallHookInner<T> {
 
 /// What to do after returning from a callback when the engine epoch reaches
 /// the deadline for a Store during execution of a function using that store.
+#[non_exhaustive]
 pub enum UpdateDeadline {
     /// Extend the deadline by the specified number of ticks.
     Continue(u64),
@@ -268,7 +269,10 @@ pub enum UpdateDeadline {
     /// it is recommended to provide [`tokio::task::yield_now`](https://docs.rs/tokio/latest/tokio/task/fn.yield_now.html)
     /// here.
     #[cfg(feature = "async")]
-    YieldCustom(u64, Box<dyn ::core::future::Future<Output = ()> + Send>),
+    YieldCustom(
+        u64,
+        ::core::pin::Pin<Box<dyn ::core::future::Future<Output = ()> + Send>>,
+    ),
 }
 
 // Forward methods on `StoreOpaque` to also being on `StoreInner<T>`
@@ -2141,7 +2145,7 @@ unsafe impl<T> crate::runtime::vm::VMStore for StoreInner<T> {
                         unsafe {
                             self.async_cx()
                                 .expect("attempted to pull async context during shutdown")
-                                .block_on(::core::pin::Pin::new_unchecked(future))?
+                                .block_on(future)?
                         }
                         delta
                     }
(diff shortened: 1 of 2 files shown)


crates/wasmtime/src/runtime/store.rs
```diff
@@ -260,6 +260,15 @@ pub enum UpdateDeadline {
...
     Yield(u64),
+    /// Extend the deadline by the specified number of ticks after yielding to
+    /// the async executor loop. This can only be used with an async [`Store`]
+    /// configured via [`Config::async_support`](crate::Config::async_support).
+    ///
+    /// The yield will be performed by the future provided; when using `tokio`
+    /// it is recommended to provide [`tokio::task::yield_now`](https://docs.rs/tokio/latest/tokio/task/fn.yield_now.html)
+    /// here.
+    #[cfg(feature = "async")]
+    YieldCustom(u64, Box<dyn ::core::future::Future<Output = ()> + Send>),
```
alexcrichton, line 271: Drive-by-comment on this unrelated to this change, mind throwing `#[non_exhaustive]` on this enum definition? That'll help consumers work with this a bit more nicely with the `#[cfg]`'d variants

crates/wasmtime/src/runtime/store.rs
```diff
@@ -2116,6 +2124,27 @@ unsafe impl<T> crate::runtime::vm::VMStore for StoreInner<T> {
...
+                        // in from the host fiber. If this finished successfully then
+                        // we were resumed normally via a `poll`, so keep going.  If
+                        // the future was dropped while we were yielded, then we need
+                        // to clean up this fiber. Do so by raising a trap which will
+                        // abort all wasm and get caught on the other side to clean
+                        // things up.
+                        unsafe {
+                            self.async_cx()
+                                .expect("attempted to pull async context during shutdown")
+                                .block_on(::core::pin::Pin::new_unchecked(future))?
```
alexcrichton, line 2144: I realize this was probably copied from `async_yield_impl` but nowadays I think we can do `let mut future = core::pin::pin!(future);` to safely creat `Pin<Thing>` and then with some `future.as_mut()` or such invocations I think it can safely create `Pin<&mut T>` to pass here

alexcrichton (approved):
Seems like a reasonable API to me! I think it's fine to add similar configuration options for fuel in the future if necessary as a Fuel-specific configuration option

crates/wasmtime/src/runtime/store.rs
posborne: In this particular case, I think we can just change the type signature of the future to be a `Pin<Box<dyn Future ...>>>` to avoid having to do anything extra to mark the memory as pinned.

posborne:
I considered adding a new API such as `epoch_deadline_async_yield_and_update_custom` or modifying the existing API to take a duration or duration + future-yielding callback but that felt pretty heavy.  As-is, there's an option for embedders that have a need to customize the yield behavior but figured I would at least call things out.

The tokio behavior is subtle enough to probably not rightly be able to be considered a bug and most use cases will probably not run into issues with the stock yield behavior.

PR #10672, opened 2025-04-24
Simplify module features in serialization

alexcrichton:
This commit simplifies how wasm features are handled in module serialization. Instead of serializing a struct-of-bools this uses the bitflags representation that wasmparser has. There are prior checks to ensure that the wasmtime version for a serialized artifact is matching so we know that the bits mean the same things too.

This reduces the boilerplate in implementing new wasm proposals and should hopefully make it so that we get all the same benefit as before just with less maintenance over time.

### DIFF
--- a/crates/wasmtime/src/engine/serialization.rs
+++ b/crates/wasmtime/src/engine/serialization.rs
@@ -408,7 +408,12 @@ impl Metadata<'_> {
 
         impl fmt::Display for WasmFeature<'_> {
             fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-                write!(f, "support for WebAssembly feature {}", self.0)
+                write!(f, "support for WebAssembly feature `")?;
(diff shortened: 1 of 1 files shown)


crates/wasmtime/src/engine/serialization.rs
```diff
@@ -477,153 +381,33 @@ impl Metadata<'_> {
...
+            let found = module_features.contains(flag);
+            let expected = other.contains(flag);
+            if flag == wasmparser::WasmFeatures::GC_TYPES {
+                Self::check_cfg_bool(
+                    cfg!(feature = "gc"),
+                    "gc",
+                    found,
+                    expected,
+                    WasmFeature(name),
+                )?;
```
pchickey, line 396: Could you leave a comment here for why gc gets this special case?

pchickey (approved):
Thanks, appreciate the deduplication!

PR #10675, opened 2025-04-25
c-api: component-model: Lookup function from an instance

MangoPeachGrape:
Only `get_func()` as of now, should https://github.com/bytecodealliance/wasmtime/pull/9812#discussion_r1884473950 come in a later PR?

One thing of note, if name isn't UTF-8, it returns false instead of returning an error, is that fine?

### DIFF
--- a/crates/c-api/src/component/component.rs
+++ b/crates/c-api/src/component/component.rs
@@ -99,7 +99,7 @@ pub unsafe extern "C" fn wasmtime_component_get_export_index(
     name_len: usize,
 ) -> Option<Box<wasmtime_component_export_index_t>> {
     let name = unsafe { std::slice::from_raw_parts(name, name_len) };
-    let Ok(name) = str::from_utf8(name) else {
+    let Ok(name) = std::str::from_utf8(name) else {
         return None;
     };
 
--- a/crates/c-api/src/component/component.rs
+++ b/crates/c-api/src/component/component.rs
@@ -1,9 +1,9 @@
 use std::ffi::{c_char, CStr};
 
 use anyhow::Context;
-use wasmtime::component::Component;
+use wasmtime::component::{Component, ComponentExportIndex};
 
-use crate::{wasm_byte_vec_t, wasm_config_t, wasm_engine_t, wasmtime_error_t};
+use crate::{wasm_byte_vec_t, wasm_config_t, wasm_engine_t, wasm_name_t, wasmtime_error_t};
 
 #[unsafe(no_mangle)]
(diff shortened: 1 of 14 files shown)


alexcrichton:
Thanks! This has triggered some further thoughts which are both preexisting and related to this PR as well:

* For names I realize now we shouldn't use `const char*`, but instead we should be using `wasm_name_t*`. Unicode can't be represented as `const char*` due to the lack of ability to support interior nul characters, so we'll want to change this and other APIs to use `wasm_name_t` instead.
* For this I think we'll want to base lookup of an item on an `ComponentExportIndex` instead of a string, although a string can be used to acquire an `ComponentExportIndex`. Effectively I think we should bind `{Instance,Component}::get_export_index` and then an export index is used to lookup a function. That'll handle nested instances as well as top-level exports in the same manner.

crates/c-api/src/component/component.rs
```diff
@@ -85,3 +85,44 @@ pub unsafe extern "C" fn wasmtime_component_clone(
...
+pub unsafe extern "C" fn wasmtime_component_get_export_index(
+    component: &wasmtime_component_t,
+    instance_export_index: *const wasmtime_component_export_index_t,
+    name: &mut wasm_name_t,
+    out: &mut *mut wasmtime_component_export_index_t,
+) -> bool {
+    let name = name.take();
+    let Ok(name) = String::from_utf8(name) else {
+        return false;
+    };
```
alexcrichton, line 104: Oh for this I think we'll want to have this be `const wasm_name_t*` in C and not take ownership here.

Also I apologize again for yet more whiplash... Looking at `wasmtime_linker_t` it looks like the functions there idiomatically take a pointer/length instead of `&wasm_name_t` (e.g. [`wasmtime_linker_define`](https://github.com/bytecodealliance/wasmtime/blob/5614b80b58d946a9a5270b1c56934fec99f097d9/crates/c-api/src/linker.rs#L51-L66)). I had forgotten about that where `wasm_name_t` I think is mostly used for return values as opposed to arguments. In light of that could the idioms of the component linker be similar where a pointer/length is given? (sorry I should have checked this earlier)

alexcrichton (approved):

MangoPeachGrape:
Next up values, which are required for function invocation?

alexcrichton:
Yeah that seems reasonable. I'll be honest though in that I don't really know how to do values. Everything I've thought of historically is either extremely chatty over the C API boundary or extremely allocation-heavy, neither of which I feel is great... If you've got ideas though please dive in!

PR #10678, opened 2025-04-26
Update `F128` argument ABI for `windows_fastcall` to match LLVM 21

beetrees:
LLVM updated the `windows_fastcall` ABI for `f128`  in https://github.com/llvm/llvm-project/pull/128848 to be passed indirectly (like `__m128i`), as that is more consistent with the rest of the `windows_fastcall` ABI. This PR updates Cranelift to match (`f128` can only be used with `windows_fastcall` in Cranelift with `enable_llvm_abi_extensions`).


fitzgen (approved):
Thanks!

PR #10681, opened 2025-04-27
[DWARF] Fix "simulated" DWARF

### LINKED ISSUE #10676: [DWARF] Excluded prolog for "simulated" DWARF
We auto-generate some DWARF for functions missing WASM-level DWARF. The line program of this "simulated" DWARF lacks the prolog markers (`prolog_end`). This makes LLDB believe the function starts later than it actually does.

See also https://github.com/llvm/llvm-project/issues/137385 for some more details.

SingleAccretion:
We need to record the correct ranges for the CU and include an offset zero entry in the line table, otherwise LLDB won't class our functions as "having debug info".

Fixes #10676.

### DIFF
--- a/crates/cranelift/src/debug/transform/simulate.rs
+++ b/crates/cranelift/src/debug/transform/simulate.rs
@@ -69,24 +69,32 @@ fn generate_line_info(
 
     for (symbol, map) in maps {
         let base_addr = map.offset;
-        out_program.begin_sequence(Some(write::Address::Symbol { symbol, addend: 0 }));
+        out_program.begin_sequence(Some(write::Address::Symbol {
+            symbol,
+            addend: base_addr as i64,
+        }));
+
+        // Always emit a row for offset zero - debuggers expect this.
+        out_program.row().address_offset = 0;
+        out_program.row().file = file_index;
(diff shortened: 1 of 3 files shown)


crates/cranelift/src/debug/transform/simulate.rs
```diff
@@ -69,24 +69,32 @@ fn generate_line_info(
 
     for (symbol, map) in maps {
         let base_addr = map.offset;
-        out_program.begin_sequence(Some(write::Address::Symbol { symbol, addend: 0 }));
+        out_program.begin_sequence(Some(write::Address::Symbol {
+            symbol,
+            addend: base_addr as i64,
+        }));
+
+        // Always emit a row for offset zero - debuggers expect this.
```
SingleAccretion, line 77: > debuggers expect this

At least, LLDB expects this. When LLDB stops at a certain address (say due to a step-in), its "has debug info" check is implemented as:
```
[current address].debug_line_row.IsValid()
```
So even if we have the right DWARF DIE, etc, if we don't have a sequence in the line table that matches the current address, we get behaviors associated with functions that don't have DI.

alexcrichton (approved):

PR #10687, opened 2025-04-28
x64: remove `Inst::AluRmiR`

abrown:
This change completely removes the `Inst::AluRmiR` variant along with associated code. It updates ABI code to use new `Inst` helpers and switches Winch to use the new assembler directly. More details in each commit; note that this significantly changes the disassembly tests.


winch/codegen/src/isa/x64/asm.rs
```diff
@@ -31,6 +32,7 @@ use cranelift_codegen::{
 };
 
 use crate::reg::WritableReg;
+use cranelift_assembler_x64 as asm;
```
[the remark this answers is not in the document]
abrown: Ok, I edited the commits based on this plus some tweaks to make some immediates smaller.

cranelift/codegen/src/isa/x64/inst/mod.rs
```diff
@@ -209,27 +208,99 @@ impl Inst {
...
+                            let src = asm::Simm8::new(simm8);
+                            asm::inst::$q_mi_sxb::new(dst, src).into()
+                        } else {
+                            let src = asm::Simm32::new(simm32 as i32);
+                            asm::inst::$q_mi_sxl::new(dst, src).into()
+                        }
+                    }
+                }
+            }
+            RegMemImm::Mem { addr: _ } => unimplemented!(),
```
alexcrichton, line 265: Would it be possible to replace the `RegMemImm` argument with just a `Reg`? And if there's only a small handful of locations using an `Imm` switching them to using the external assembler directly?

cranelift/codegen/src/isa/x64/inst/emit.rs
```diff
@@ -3968,12 +3850,7 @@ pub(crate) fn emit(
             if *is_saturating {
                 // The input was "small" (< 2**(width -1)), so the only way to get an integer
                 // overflow is because the input was too small: saturate to the min value, i.e. 0.
-                let inst = Inst::alu_rmi_r(
-                    *dst_size,
-                    AluRmiROpcode::Xor,
-                    RegMemImm::reg(dst.to_reg()),
-                    dst,
-                );
+                let inst = Inst::xor(*dst_size, RegMemImm::reg(dst.to_reg()), dst); // TODO: uninit for regalloc?
```
alexcrichton, line 3853: This "TODO" should be ok because this is all post-regalloc so no need to worry about uninit/etc

cranelift/codegen/src/isa/x64/inst/mod.rs
alexcrichton: Or, alternatively, could these helpers be purged entirely in favor of using the external assembler directly?

winch/codegen/src/isa/x64/asm.rs
```diff
@@ -772,48 +787,55 @@ impl Assembler {
...
-            dst: dst.map(Into::into),
-        });
+        let dst = pair(dst).into();
+        let inst = match size {
+            OperandSize::S8 => asm::inst::subb_mi::new(dst, imm.try_into().unwrap()).into(),
+            OperandSize::S16 => asm::inst::subw_mi::new(dst, imm.try_into().unwrap()).into(),
+            OperandSize::S32 => asm::inst::subl_mi::new(dst, imm.into()).into(),
+            OperandSize::S64 => asm::inst::subq_mi_sxl::new(dst, imm.into()).into(),
+            OperandSize::S128 => unimplemented!(),
+        };
```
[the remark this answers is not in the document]
saulecabrera: Oh, taking a closer look those helpers seem to be the ones that are lowered through `lower_alu!`? If that's the case, I sense a bit of repetition, but I think it's fine to diverge and keep things the way you have them in this PR; that gives Winch the freedom to evolve the lowering rules more independently, plus I think there's probably going to be less indirection this way.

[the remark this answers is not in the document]
saulecabrera: > The disadvantage is that, for these immediate encodings, we could be using the "sign-extend when it fits in 8 bits" forms for smaller code size but I just chose not to

I think it's fine not to introduce this, if I remember correctly, we didn't have this optimization before anyway. Perhaps we could consider adding a `TODO` at the top of the module to track these potential optimizations?

cranelift/codegen/src/isa/x64/inst/mod.rs
abrown: Ok, b197c14 does just that, but I can't claim things are very pretty. There's an ergonomics issue here that can absolutely be solved later, but I would be interested in discussing this at some point. Here are some thoughts on making this _look_ nicer:
- to avoid repetitive `.into()` conversions for each operand, perhaps we could make the assembler constructor functions take parameters of `impl Into<...>`?
- to avoid repetitive `.into()` conversions to create `Inst::External`s, perhaps we could create an `Inst::external(...)` helper to do some of this more succinctly?
- to avoid repetitive checks for immediate sizes (i.e., "does this fit in an 8-bit immediate? no? ok, let's use a 32-bit...") we might be able to reuse the `ABIMachineSpec::gen_add_imm` helper? If that were to take an `i32` instead of a `u32`, we could probably replace all the ~15 `sub` and `add` instances everywhere with by flipping the sign for `sub`
- to avoid `asm::` everywhere... well, what is the conventional thing to do about that?
- to avoid wrapping immediates in `asm::Simm*` and `asm::Imm*`, we could refactor things in the assembler to just take the raw Rust types; right now the wrapper is used for pretty-printing and details like that but there's surely another way to do that.

Thoughts?

winch/codegen/src/isa/x64/asm.rs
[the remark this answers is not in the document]
abrown: > I'm still getting caught up with the structure of the new x64 assembler

Maybe I should improve the documentation! The `lower_alu!` macro is now gone and it was just a side helper anyways. These `new` functions are all being generated from instruction definitions like those in this file:

https://github.com/bytecodealliance/wasmtime/blob/28e26fdf7e389148c5725a58f0627693c525b170/cranelift/assembler-x64/meta/src/instructions/and.rs#L27-L30

The `cranelift-assembler-x64` exposes all that generated code. If you want to see it, run `cd cranelift/assembler-x64; cargo run` and open the file that gets spit out of that.

> I sense a bit of repetition, but I think it's fine to diverge and keep things the way you have them in this PR; that gives Winch the freedom to evolve the lowering rules more independently, plus I think there's probably going to be less indirection this way.

Yeah, that was what I was concerned about as well. But, yeah, "more independent" and "more direct" are the benefits here. I am still thinking of through the best way to do this so if you have any recommendations on making this more ergonomic and less repetitive, I'm interested!

winch/codegen/src/isa/x64/asm.rs
```diff
@@ -772,48 +787,55 @@ impl Assembler {
...
+            OperandSize::S16 => asm::inst::subw_rm::new(dst, src).into(),
+            OperandSize::S32 => asm::inst::subl_rm::new(dst, src).into(),
+            OperandSize::S64 => asm::inst::subq_rm::new(dst, src).into(),
+            OperandSize::S128 => unimplemented!(),
+        };
+        self.emit(Inst::External { inst });
     }
 
     /// Subtact immediate register.
     pub fn sub_ir(&mut self, imm: i32, dst: WritableReg, size: OperandSize) {
```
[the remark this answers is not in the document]
abrown: Yeah, good point; I almost switched everything around here but then thought, "let's keep this small if we can."

cranelift/codegen/src/isa/x64/inst/mod.rs
alexcrichton: I like the idea of using `impl Into<...>` in the ctor arguments to assist with host-level ergonomics, it's pretty similar to how things work with `(convert ...)` in ISLE. I could go either way on `Inst::external`, and another way to shorten things would be to use `inst::foo` instead of `asm::inst::foo`? Or perhaps `trait EmitExt { .. }` just to give an `emit` method?

For immediates I still think that one option we could do is, just before emission, to "compress" the immediate. For example [here](https://github.com/bytecodealliance/wasmtime/blob/28e26fdf7e389148c5725a58f0627693c525b170/cranelift/codegen/src/isa/x64/inst/emit.rs#L4681-L4695) we could match on a arithmetic instructions like add/sub which use the 32-bit immediate form and optionally compress them to the 8-bit form. I don't think that would have any meaningful impact on compile time, wouldn't affect any semantics, and would avoid the need to do this everywhere. Alternatively having `Inst::addq_mi` as a helper I also think would be reasonable where it's still a wrapper around the assembler but an even thinner wrapper around the assembler than the one that previously took `RegMemImm` and only does immediate-checking.

I also think it'd be reasonable to refactor the assembler to use raw Rust types instead of `asm::{Si,I}mm` types too to assist with verbosity.

Near-term what about the `impl Into<...>` plus op-with-immediate helpers on `Inst`? Longer-term could consider something like the mutate-the-instruction-just-before-emit or something like that.

abrown: Ok, latest two commits handle any near-term case.

alexcrichton (approved):

PR #10690, opened 2025-04-28
Test out Windows aarch64 runners in CI

alexcrichton:
Looks like they're new on GitHub, let's see if tests pass?

prtest:windows-aarch64


alexcrichton:
> let's see if tests pass?

nope: rustup is not installed

PR #10694, opened 2025-04-29
Test C/C++ headers can be included in isolation

alexcrichton:
I frequently make mistakes about this, so I figure it's good to at least try to test this. It's not possible to fully test AFAIK but this is hopefully at least something.


fitzgen (approved):
Good idea

PR #10695, opened 2025-04-29
Initial support for vex encoding for the new assembler.

jlb6740:
Patch is first, one or two more will be pushed for enabling.

### DIFF
--- a/cranelift/assembler-x64/meta/src/dsl/encoding.rs
+++ b/cranelift/assembler-x64/meta/src/dsl/encoding.rs
@@ -32,8 +32,24 @@ pub fn rex(opcode: impl Into<Opcodes>) -> Rex {
 
 /// An abbreviated constructor for VEX-encoded instructions.
 #[must_use]
-pub fn vex() -> Vex {
-    Vex {}
+pub fn vex(opcode: impl Into<Opcodes>) -> Vex {
+    Vex {
+        map: OpcodeMap::None,
+        opcodes: opcode.into(),
+        w: false,
+        r: false,
+        digit: 0,
+        is4: false,
+        size_e: 0,
+        wig: false,
+        rxb: 0,
+        length: VexLength::default(),
+        mmmmm: VexMMMMM::None,
+        pp: VexPP::None,
+        reg: 0x00,
+        vvvv: None,
+        imm: None,
+    }
 }
 
 /// Enumerate the ways x64 encodes instructions.
@@ -48,7 +64,7 @@ impl Encoding {
     pub fn validate(&self, operands: &[Operand]) {
         match self {
             Encoding::Rex(rex) => rex.validate(operands),
-            Encoding::Vex(vex) => vex.validate(),
+            Encoding::Vex(vex) => vex.validate(operands),
         }
     }
 }
@@ -57,7 +73,7 @@ impl fmt::Display for Encoding {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
             Encoding::Rex(rex) => write!(f, "{rex}"),
-            Encoding::Vex(_vex) => todo!(),
+            Encoding::Vex(vex) => write!(f, "{vex}"),
         }
     }
 }
@@ -383,6 +399,23 @@ impl Prefixes {
     pub fn is_empty(&self) -> bool {
         self.group1.is_none() && self.group2.is_none() && self.group3.is_none() && self.group4.is_none()
     }
+
+    pub fn bits(&self) -> u8 {
+        let mut bits = 0;
+        if self.group1.is_some() {
+            bits |= 0b0001;
+        }
+        if self.group2.is_some() {
+            bits |= 0b0010;
+        }
+        if self.group3.is_some() {
+            bits |= 0b0100;
+        }
+        if self.group4.is_some() {
+            bits |= 0b1000;
+        }
+        bits
+    }
 }
 
 pub enum Group1Prefix {
@@ -559,7 +592,7 @@ pub enum Imm {
 }
 
 impl Imm {
-    fn bits(&self) -> u8 {
+    fn bits(&self) -> u16 {
         match self {
             Imm::None => 0,
             Imm::ib => 8,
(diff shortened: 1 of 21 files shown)


cranelift/assembler-x64/src/vex.rs
```diff
@@ -0,0 +1,339 @@
+//! Encoding logic for VEX instructions.
+//use super::evex::{Register, RegisterOrAmode};
+
+use super::rex;
+use super::XmmMem;
+use crate::api::{AsReg, CodeSink, KnownOffsetTable, Registers};
+use crate::mem::emit_modrm_sib_disp;
+use crate::Amode;
+use cranelift_assembler_x64_meta::dsl::{OpcodeMap, Vex};
```
alexcrichton, line 9: Would it be possible to duplicate necessary types here? The intent is that the meta crate is only used at build-time and isn't a runtime dependency but this is otherwise adding the meta crate as a runtime dependency as well

jlb6740: Hi @alexcrichton, I think so. I did it this way as it seemed to be required for a previous iteration of this patch set, plus it assures consistency and avoids duplication, but I understand we don't want a dependency on the meta crate and that duplication is ok. If I do believe there is a dependency we want to keep, I'll follow @cfallin advise and see if a shared crate makes sense here. Also note, there are a few unnecessary elements in the vex struct in the meta crate that I am removing along with commented out code intended to help develop the follow-up patch 2 of 2 to be submitted later when we enable this.

jlb6740: @alexcrichton .. I believe this is addressed now.

cranelift/assembler-x64/meta/src/dsl/encoding.rs
```diff
@@ -582,10 +611,128 @@ impl fmt::Display for Imm {
...
+        Self { length, ..self }
+    }
+    pub fn pp(self, pp: VexPP) -> Self {
+        Self { pp, ..self }
+    }
+    pub fn mmmmm(self, mmmmm: VexMMMMM) -> Self {
+        Self { mmmmm, ..self }
+    }
+
+    fn validate(&self, _operands: &[Operand]) {}
```
abrown, line 719: It will be helpful to check here for invalid instruction definitions: that way we can find them early, at codegen time.

cranelift/assembler-x64/meta/src/dsl/format.rs
```diff
@@ -248,35 +256,51 @@ pub enum Location {
...
+            zmm1 | zmm2 | zmm3 | zmm_m512 => 512,
         }
     }
 
     /// Return the number of bytes accessed, for convenience.
     #[must_use]
-    pub fn bytes(&self) -> u8 {
-        self.bits() / 8
+    pub fn bytes(&self) -> u16 {
+        self.bits() / 16
```
abrown, line 293: This is not right:

```suggestion
        self.bits() / 8
```

cranelift/assembler-x64/meta/src/dsl/format.rs
```diff
@@ -248,35 +256,51 @@ pub enum Location {
...
+    xmm2,
+    xmm3,
+    ymm1,
+    ymm2,
+    ymm3,
+    zmm1,
+    zmm2,
+    zmm3,
+
+    xmm_m128,
```
abrown, line 269: This is the same as `rm128`... I'm not convinced adding these unused variants is necessary at this point, either.

cranelift/assembler-x64/meta/src/generate/format.rs
```diff
@@ -137,6 +140,22 @@ impl dsl::Format {
...
+    fn generate_vex(&self, f: &mut Formatter, vex: &dsl::Vex) {
+        f.empty_line();
+        f.comment("Emit VEX prefix.");
+        fmtln!(f, "let vex: VexInstruction<R> = VexInstruction::default();");
+        fmtln!(f, "let mut vex = vex.opcode(0x{:0x});", vex.opcodes.primary);
+        fmtln!(f, "vex.reg = self.xmm1.enc();");
+        fmtln!(f, "match &self.xmm_m128 {{");
+        fmtln!(f, "XmmMem::Xmm(r) => {{vex.rm =  XmmMem::Xmm(r.clone());}}");
+        fmtln!(f, "XmmMem::Mem(m) => {{vex.rm =  XmmMem::Mem(m.clone());}}");
+        fmtln!(f, "}}");
```
abrown, line 152: When you're adding brace-delimited blocks, you can use `Formatter::add_block` to get the indentation right, avoid forgetting to add a closing brace, and to match the rest of this crate:

```suggestion
        f.add_block("match &self.xmm_m128", |f| {
            fmtln!(f, "XmmMem::Xmm(r) => {{vex.rm =  XmmMem::Xmm(r.clone());}}");
            fmtln!(f, "XmmMem::Mem(m) => {{vex.rm =  XmmMem::Mem(m.clone());}}");
        });
```

abrown: This seems a bit redundant to match here in any case...

cranelift/assembler-x64/meta/src/instructions/addpd.rs
```diff
@@ -0,0 +1,15 @@
+use crate::dsl::{
+    align, fmt, inst, r, rex, rw, vex, w, Feature::*, Inst, Location::*, VexLength::*, VexMMMMM::*, VexPP::*,
+};
+
+pub fn list() -> Vec<Inst> {
+    vec![
+        inst("addpd", fmt("A", [rw(xmm1), r(align(rm128))]), rex([0x66, 0x0F, 0x58]).r(), _64b | compat | sse),
```
abrown, line 7: This duplicates one of the definitions in `add.rs`, right? Let's just keep all the addition in that file to avoid this kind of mistake.

cranelift/assembler-x64/src/fuzz.rs
```diff
@@ -274,6 +277,6 @@ mod test {
         .budget_ms(1_000);
 
         // This will run the `roundtrip` fuzzer for one second. To repeatably
-        // test a single input, append `.seed(0x<failing seed>)`.
+        // test a single0 input, append `.seed(0x<failing seed>)`.
```
abrown, line 280:
```suggestion
        // test a single input, append `.seed(0x<failing seed>)`.
```

cranelift/assembler-x64/src/vex.rs
```diff
@@ -0,0 +1,259 @@
...
+            map: OpcodeMap::None,
+            opcode: 0x00,
+            w: false,
+            reg: 0x00,
+            rm: XmmMem::default(),
+            vvvv: None,
+            imm: None,
+        }
+    }
+}
```
abrown, line 105: We shouldn't need to create a default (i.e., invalid) VEX instruction ever, right?

cranelift/assembler-x64/meta/src/generate/format.rs
```diff
@@ -137,6 +140,22 @@ impl dsl::Format {
         }
     }
 
+    fn generate_vex(&self, f: &mut Formatter, vex: &dsl::Vex) {
+        f.empty_line();
+        f.comment("Emit VEX prefix.");
+        fmtln!(f, "let vex: VexInstruction<R> = VexInstruction::default();");
```
abrown, line 146: Instead of creating a default (i.e., invalid) instruction here, why not use a `new` constructor or a builder pattern (e.g., `op(...).reg(...)...`) to create only valid instructions?

abrown (commented):
I got about halfway through reviewing this; here are some comments.

cranelift/assembler-x64/meta/src/dsl/encoding.rs
```diff
@@ -582,10 +611,128 @@ impl fmt::Display for Imm {
     }
 }
 
-pub struct Vex {}
+pub struct Vex {
+    pub opcodes: Opcodes,
+    pub w: bool,
+    pub r: bool,
+    pub wig: bool,
+    pub rxb: u8,
```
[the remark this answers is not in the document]
jlb6740: Yes, I am not positive, but I don't even think any of these bits will need to be set at the DSL level. For now I'll just remove the rxb and add the b and w bits as suggested here (but maybe we can remove later).

cranelift/assembler-x64/meta/src/instructions/addpd.rs
jlb6740: For the final vex support I removed the one in add.rs (but put it back for this pr). The reason is simple, I was organizing the instructions the same way they are organized in the manual by opcode (specifically Vol 2A 3-34). The ADDPD opcode instruction section covers rex, vex, and evex. What do you think about this organization of grouping instructions consistent with the manual?

cranelift/assembler-x64/src/vex.rs
jlb6740: I don't think so. Will remove this.

cranelift/assembler-x64/meta/src/dsl/encoding.rs
jlb6740: Yes agreed .. can we implement this validate in the next PR where we turn the plumbing on and we can see that the validation is actually doing the right thing against those first few instructions or would you like to see a little more of this filled out now?

cranelift/assembler-x64/meta/src/dsl/format.rs
jlb6740: I was wanting to be consistent with the manual here:

ADDPD xmm1, xmm2/m128
VADDPD xmm1, xmm2, xmm3/m128

and writing something like this:
```
inst(
            "vaddpd",
            fmt("B", [w(xmm1), r(xmm2), r(xmm_m128)]),
            vex(0x58).length(_128).pp(_66).mmmmm(_OF),
            _64b | compat | sse,
        ),

```
instead of like this:
```
inst(
            "vaddpd",
            fmt("B", [w(xmm1), r(xmm2), r(rm128)]),
            vex(0x58).length(_128).pp(_66).mmmmm(_OF),
            _64b | compat | sse,
        ),
```

I kept the rm128 because I didn't know if it would fly during review that I replaced in in all the rex encoded instructions but using the naming xmm_m128 is more consistent with the manual. What do you think?

jlb6740:
Hi @cfallin @alexcrichton @abrown I've attempted to address all comments or at least touch base with the reviewer offline. I've moved this PR to https://github.com/bytecodealliance/wasmtime/pull/10754 where instead of doing this in 2 parts, well just attempt to switch on the lowering of vaddpd and vaddps. I'll close this PR.

PR #10698, opened 2025-04-30
Make a non-deterministic log message deterministic

adambratschikaye:
This resolves the issue from the `eprintln!` line mentioned in https://github.com/bytecodealliance/wasmtime/issues/9553.

When generating a file we currently log the full path to that file which changes depending on where the build is run and can prevent cache hits in build systems like Bazel.

This change instead just prints the file name, along with a deterministic name for the parent directory (e.g. `OUT_DIR` or `ISLE_DIR`) which can be makes the log message deterministic and should still be useful for debugging.

### DIFF
--- a/cranelift/assembler-x64/build.rs
+++ b/cranelift/assembler-x64/build.rs
@@ -9,7 +9,11 @@ fn main() {
 
     let out_dir = env::var("OUT_DIR").expect("The OUT_DIR environment variable must be set");
     let out_dir = Path::new(&out_dir);
-    let built_files = [meta::generate_rust_assembler(out_dir, "assembler.rs")];
+    let built_files = [meta::generate_rust_assembler(
+        out_dir,
+        "OUT_DIR",
+        "assembler.rs",
+    )];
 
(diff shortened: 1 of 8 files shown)


cfallin (commented):
Thanks for the PR!

That said, I'm not sure how I feel about this: the original purpose of this output was to print the actual path so that someone debugging the build script steps could go read the generated file; so rewriting the printed output to use some artificial name defeats that purpose. A build tool that thinks a build-script run that prints different stderr output is different will also think that a build step that writes to a different file path is different, no? (In other words, are we really "deterministic" if Cargo is choosing a different path anyway for different runs?)

If for some reason only the diagnostic output is the issue, we could probably turn this into a `log::info!` or similar. Generally though I'd rather we don't remove useful functionality to satisfy (what appears to me at the moment to be) an odd/overly-strict build tool rule.

adambratschikaye:
Good points. It seems like the output was just being tracked by bazel's rules for rust crates as a means for debugging and we were able to have the rules modified so output isn't tracked by default. So this is no longer an issue.

PR #10699, opened 2025-04-30
Fix missing libcalls with simd float rounding

alexcrichton:
This commit fixes some more fallout found on oss-fuzz about the x64 generating rounding builtins when it shouldn't be. This situation is caused by simd float rounding instructions which the x64 backend lowers to libcall-per-element and now needs to move that logic to the frontend instead.


dicej (commented):
I don't have enough Cranelift context yet to evaluate this properly, so I'll let Dan review it before approving, but LGTM.

alexcrichton:
@fitzgen would you be able to take a look at this perhaps?

fitzgen (approved):

PR #10700, opened 2025-04-30
Temporarily disable simd differential fuzzing with wasmi

alexcrichton:
OSS-Fuzz has found a differential execution with wasmi/wasmtime which relates to simd which I've reported, so in the meantime to let the fuzzers keep making progress this disables simd when using wasmi as a differential fuzzer.


fitzgen (approved):

Robbepop:
Note for future: The reported Wasmi bug was this: https://github.com/wasmi-labs/wasmi/pull/1450
This has since been fixed and included in https://github.com/wasmi-labs/wasmi/releases/tag/v0.44.1.

PR #10701, opened 2025-04-30
Collapse resource tables to be per-instance, not per-resource

alexcrichton:
This commit is an update to how Wasmtime represents and processes resource handles for components. The upstream specification changed in WebAssembly/component-model#427 and while it's been awhile since that change this is Wasmtime catching up to that change. It's expected that these paths are going to get stressed more with component model async and more kinds of handles so this additionally refactors things to leave sort of "holes" where async resources are going to go.

This notably will result in different behavior for guest programs. Previously if there were two resources imported into a guest they could have overlapping indices and now they're going to all have disjoint indices since they're all in the same index space. It's expected that this is will have a minimal, if any, impact on component guests as in theory they're all mostly treating handles as opaque indexes today anyway.

### DIFF
--- a/crates/environ/src/component/dfg.rs
+++ b/crates/environ/src/component/dfg.rs
@@ -120,11 +120,6 @@ pub struct ComponentDfg {
     /// with what the corresponding runtime import is.
     pub imported_resources: PrimaryMap<ResourceIndex, RuntimeImportIndex>,
 
-    /// The total number of resource tables that will be used by this component,
-    /// currently the number of unique `TypeResourceTableIndex` allocations for
-    /// this component.
-    pub num_resource_tables: usize,
-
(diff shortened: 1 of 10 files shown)


alexcrichton:
cc @dicej

alexcrichton:
I'll also note that the original motivation for this spec change, https://github.com/WebAssembly/component-model/issues/395, also detailed improved error messages from runtime about mismatches in index types. For example the original issue was importing two versions of `wasi:io/poll` and accidentally crossing handle indexes between the two. This PR doesn't go so far to attach debugging names to resources so it's known how to inform the users/components what exactly a mismatch was, it only says "expected resource found different resource" effectively. I've left the improvement of error messages to a future PR.

PR #10709, opened 2025-05-02
Cranelift: fix invalid regalloc constraints on try-call with empty handler list.

cfallin:
(Reported by bjorn3 -- thanks!)

We handle edge splitting for regalloc moves in a way that is designed to split minimally (for compiler performance and codegen quality): we only split edges that are truly critical, i.e., come from a block with more than one successor and go to a block with more than one predecessor. In all other cases, there is always a place for these moves to go: if a block only has one successor, then edge-moves can go before its jump; and if a block only has one predecessor, then edge-moves can go at the beginning of that block (before the blockparams' parallel-move). The former case -- before the branch -- works because of a restriction that regalloc2 imposes: branches with only one target (that is, unconditional branches) cannot have any arguments. Otherwise, those uses would occur after the edge-moves have already shuffled the register state into a state suitable for the next block. Violating this constraint leads to a panic.

With ordinary unconditional branches, this is no problem. With `try_call`s with at least one handler listed, this is also no problem: such a terminator is seen as a branch with (at least) two targets by regalloc, so no edge-moves are placed before it, so it's allowed to have arguments, such as the arguments to the call itself. However, a `try_call` with *no* handler clauses, though pathological, appears to regalloc as an unconditional branch and so should not have any arguments. The included test-case triggers this issue with such a `try_call` together with a normal-return target branch that has more than one incoming edge, forcing the location for the moves into the `try_call`'s block. (The lack of actual edge-moves doesn't matter -- RA2 performs the check on the IR restriction first.) The result is a panic at compile time.

This PR fixes the issue by extending a similar fix for `br_table`s (which can trigger a very similar bug if they have only the default case, i.e., one target) to `try_call{,_indirect}` as well: the lowered-block order computation, where edge-splits are determined, pretends that they always have at least two successors. This ensures that edges will be split as necessary, satisfying the no-arguments-to-unconditional-terminators restriction.

### DIFF
--- a/cranelift/codegen/src/machinst/blockorder.rs
+++ b/cranelift/codegen/src/machinst/blockorder.rs
@@ -173,12 +173,19 @@ impl BlockLoweringOrder {
                 }
             });
 
-            // Ensure that blocks terminated by br_table instructions with an empty jump table are
-            // still treated like conditional blocks from the point of view of critical edge
-            // splitting.
+            // Ensure that blocks terminated by br_table instructions
+            // with an empty jump table are still treated like
+            // conditional blocks from the point of view of critical
+            // edge splitting. Also do the same for TryCall and
+            // TryCallIndirect: we cannot have edge moves before the
+            // branch, even if they have empty handler tables and thus
+            // would otherwise have only one successor.
             if let Some(inst) = f.layout.last_inst(block) {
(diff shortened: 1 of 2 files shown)


bjorn3:
Please remove the @ mention from the commit message and PR description to avoid pinging me every time someone pushes it to their repo.

PR #10714, opened 2025-05-02
Fix broken Component API URLs in documentation

GarmashAlex:
This commit fixes broken URLs in the pre-compiling-wasm documentation that 
were causing 404 errors. The URLs for Component::deserialize and 
Component::deserialize_file were incorrectly pointing to a non-existent 
path. The fix updates the paths to correctly reflect that the Component 
struct is in the component module, not in the root namespace.


fitzgen:
What is with

image

?

fitzgen:
Why the weird branch name?

pchickey:
I don't know enough to be confident the branch name warning from GitHub isn't something potentially harmful, so out of an abundance of caution I recreated this change in #10717

PR #10716, opened 2025-05-02
asm: refactor using `RegClass` and `xmm_m{32|64|128}`

abrown:
This change involves no change in functionality but does pave the way for defining more SSE instructions in two ways:
- it adds the `xmm_m32` and `xmm_m64` variants as locations and changes `rm128` to read `xmm_m128` to match; this is necessary for SSE `*ss` and `*sd` instructions that load 32- and 64-bit values into XMM registers
- by returning `RegClass` from `Location::reg_class` it is much easier to distinguish what kind of a register an operand is using: previously this matched on the number of bits (i.e., 128) which does no longer makes sense for `xmm_m32` and `xmm_m64`.


cfallin (approved):
Good generalizations; thanks!

PR #10718, opened 2025-05-02
x64: use new versions of SSE `add` and `sub`

abrown:
This set of changes builds on #10716 to clean up some of the old `SseOpcode` and related code. This adds some new instructions `{add|sub}{ss|sd}` and plumbs them throughout Cranelift and Winch.


cranelift/codegen/src/isa/x64/inst/external.rs
```diff
@@ -126,9 +128,59 @@ impl asm::AsReg for PairedGpr {
...
+}
+
+// For emission ergonomics.
+impl From<Reg> for asm::XmmMem<Xmm, Gpr> {
+    fn from(xmm: Reg) -> Self {
+        assert!(xmm.class() == RegClass::Float);
+        let xmm = Xmm::unwrap_new(xmm);
+        Self::Xmm(xmm)
+    }
+}
```
abrown, line 177: I'm not particularly happy with how much conversion code is needed. Having this here is certainly better than constructing all the necessary structs when trying to reason about some code sequence, but it still feels a bit much. This is not really anything that needs to be solved in this PR; in fact, it might be better solved once most instructions have been ported to the new assembler. But I bring it up to see if anyone has any good ideas for reducing this kind of thing.

cfallin: Yeah, I suspect we'll be able to think a little more clearly about this (or at least I know I will!) once we're able to DCE away all of the types that the old-style insts are keeping alive...

cfallin (approved):
This all looks pretty reasonable -- thanks!

PR #10720, opened 2025-05-02
x64: convert more logical vector instructions

abrown:
More details in each commit message.

### DIFF
--- a/cranelift/assembler-x64/meta/src/instructions/and.rs
+++ b/cranelift/assembler-x64/meta/src/instructions/and.rs
@@ -42,5 +42,7 @@ pub fn list() -> Vec<Inst> {
         // Vector instructions.
         inst("andps", fmt("A", [rw(xmm), r(align(xmm_m128))]), rex([0x0F, 0x54]).r(), _64b | compat | sse),
         inst("andpd", fmt("A", [rw(xmm), r(align(xmm_m128))]), rex([0x66, 0x0F, 0x54]).r(), _64b | compat | sse2),
+        inst("andnps", fmt("A", [rw(xmm), r(align(xmm_m128))]), rex([0x0F, 0x55]).r(), _64b | compat | sse),
+        inst("andnpd", fmt("A", [rw(xmm), r(align(xmm_m128))]), rex([0x66, 0x0F, 0x55]).r(), _64b | compat | sse2),
     ]
 }
--- a/cranelift/codegen/src/isa/x64/inst.isle
+++ b/cranelift/codegen/src/isa/x64/inst.isle
@@ -876,9 +876,7 @@
       (enum Rorx))
 
 (type SseOpcode extern
-      (enum Andps
-            Andpd
-            Andnps
+      (enum Andnps
             Andnpd
             Blendvpd
             Blendvps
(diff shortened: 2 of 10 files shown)


cranelift/assembler-x64/meta/src/instructions/and.rs
```diff
@@ -42,5 +42,7 @@ pub fn list() -> Vec<Inst> {
         // Vector instructions.
         inst("andps", fmt("A", [rw(xmm), r(align(xmm_m128))]), rex([0x0F, 0x54]).r(), _64b | compat | sse),
```
alexcrichton, line 43: Drive-by question: [here](https://www.felixcloutier.com/x86/andps) it says the encoding is: `NP 0F 54 /r ANDPS xmm1, xmm2/m128`, how does the "NP" map to this?

alexcrichton: Also drive-by-question, should `_64b | compat` be removed? I feel like that has been copy/pasted around but is mostly only relevant for gpr instructions, 64-bit ones?

alexcrichton (approved):

cranelift/assembler-x64/meta/src/instructions/and.rs
abrown: > Drive-by question: [here](https://www.felixcloutier.com/x86/andps) it says the encoding is: `NP 0F 54 /r ANDPS xmm1, xmm2/m128`, how does the "NP" map to this?

I think it boils down to "no prefix":

> NP — Indicates the use of 66/F2/F3 prefixes (beyond those already part of the instructions opcode) are not 
allowed with the instruction. Such use will either cause an invalid-opcode exception (#UD) or result in the 
encoding for a different instruction.

> should _64b | compat be removed?

I was just thinking about that today: for most instructions we want to have a shorter thing that means "both". Maybe `base`?

alexcrichton: Ah ok makes sense about NP.

For features [the normal `and` instruction](https://www.felixcloutier.com/x86/and) has columns for 64-bit and compat, so it makes sense to me to translate that to `_64b | compat`, I don't think it's weird to have to specify both vs having a combined option. For [`andnps` though](https://www.felixcloutier.com/x86/andps) it's easy to translate the "CPUID Feature Flag" column but how does the "64/32 bit Mode Support" column translate to this? Does "V/V" mean `_64b | compat`? Should it not require `_64b` or `compat` features?

abrown: > how does the "64/32 bit Mode Support" column translate to this? Does "V/V" mean _64b | compat?

For older instructions they split this out into two separate columns, "64-bit Mode" and "Compat/Leg Mode," but then they were coalesced into a single column with the space-efficient "V/V" instead of "Valid | Valid." So my reading is that `andps` is valid in both 64-bit and 32-bit (compatibility) modes, hence `_64b | compat`.

So think at some level we need to boil down to essentially `_64b | compat` but we could have an alias for that since it is quite, quite common. I'm going to merge this as-is and we can keep discussing elsewhere.

PR #10722, opened 2025-05-03
Refactor call ABI implementation

uweigand:
This refactors implementation of call ABI handling across architectures with the goal of bringing s390x in line with other platforms.

The main idea is to
- handle main call instruction selection and generation in ISLE (like s390x but unlike other platforms today)
- handle argument setup mostly outside of ISLE (like other platforms but unlike s390x today)
- handle return value processing as part of the call instructio (like all platforms today)

All platforms now emit the main call instruction directly from ISLE, which e.g. handles selection of the correct ISA instruction depending on the call destination.  This ISLE code calls out to helper routines to handle argument and return value processing.  These helpers are mostly common code and provided by the Callee and/or Lower layers, with some platform-specific additions via ISLE Context routines.

The old CallSite abstraction is no longer needed; most of the differences between call and return_call handling disappear. (There is still a common-code CallInfo vs. a platform-specifc ReturnCallInfo.  At this point, it should be relatively straight- forward to make CallInfo platform-specific as well if desired, but this is not done here.)

Some ISLE infrastructure for iterators / loops, which was only ever used by the s390x argument processing code, has been removed.

s390x now closely matches all other platforms, with only a few special cases (slightly different tail-call ABI requires some differences in stack offset computations; we still need to handle vector lane swaps for cross-ABI calls), which should simplify future maintenance.

FYI @cfallin

### DIFF
--- a/cranelift/codegen/src/isa/aarch64/abi.rs
+++ b/cranelift/codegen/src/isa/aarch64/abi.rs
@@ -6,7 +6,7 @@ use crate::ir::types::*;
 use crate::ir::MemFlags;
 use crate::ir::{dynamic_to_fixed, ExternalName, LibCall, Signature};
 use crate::isa;
-use crate::isa::aarch64::{inst::*, settings as aarch64_settings, AArch64Backend};
+use crate::isa::aarch64::{inst::*, settings as aarch64_settings};
 use crate::isa::unwind::UnwindInst;
 use crate::isa::winch;
 use crate::machinst::*;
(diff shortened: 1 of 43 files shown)


abrown (approved):
This makes sense to me; looking at the x64 backend especially, this refactoring seems fine. I'll defer to @cfallin, though, on any finer points since it may be that you all have discussed this change previously (?).

cfallin (approved):
This looks great -- thanks a bunch for doing this refactor and better unifying all the ABI implementations! I like the more functional approach (external helpers to build info, then emit the one call inst) a lot.

uweigand:
Looks like there's another internal error.

alexcrichton:
I'm not certain it's something wrong with the action, the failure is also for `actions/cache@v4`. I restarted again and it's making progress this time. I think github actions is just having a bad day so far and I don't think there's anything specific about abrown/install-vtune-action

PR #10724, opened 2025-05-03
docs: update link to target Markdown rather than HTML

GarmashAlex:
Fixes the link to the deterministic fuel section in 
   examples-deterministic-wasm-execution.md by changing the file extension 
   from .html to .md.

### DIFF
--- a/docs/examples-deterministic-wasm-execution.md
+++ b/docs/examples-deterministic-wasm-execution.md
@@ -76,5 +76,5 @@ You can make this deterministic in a variety of ways:
 ## Use Deterministic Interruption, If Any
 
 If you are making Wasm execution interruptible, use [deterministic fuel-based
-interruption](./examples-interrupting-wasm.html#deterministic-fuel) rather than
+interruption](./examples-interrupting-wasm.md#deterministic-fuel) rather than
 non-deterministic epoch-based interruption.


cfallin:
This link actually needs to be to the HTML -- the rendered content appears [here](https://docs.wasmtime.dev/examples-deterministic-wasm-execution.html), and changing the target of the link at the bottom from `examples-interrupting-wasm.html` to `examples-interrupting-wasm.md` creates a 404, unfortunately. I suspect this is a limitation in `mdbook`'s rendering: it can translate normal links to whole Markdown files to the corresponding HTML files but not links with anchor targets.

(This appears to be tracked in rust-lang/mdBook#167, which on cursory skim looks like one of those "will never be fixed" kinds of issues, unfortunately...)

cfallin:
Ah, actually, my experiment was incorrect -- I was playing with this in `SUMMARY.md` to ensure this kind of link worked, and apparently it doesn't there specifically (per the above issue) but it does in other files.

I wouldn't call this a "broken link" -- it does already work on the page I linked above -- but linking to the Markdown is more portable (will work with other rendering backends), so this PR seems reasonable -- thanks.

cfallin (approved):

PR #10727, opened 2025-05-05
pulley: Allow disabling SIMD in the interpreter at compile-time

alexcrichton:
This commit adds a new feature to Pulley which is used to reduce the compiled code size of the interpreter itself by disabling SIMD opcode interpretation at compile-time. The goal here is to be low-impact on Pulley itself to avoid needing `#[cfg]` all over the place and to additionally avoid the need to redesign Pulley's opcode macro for use in various parts of Wasmtime.

Methods are annotated with a custom macro in the interpreter which registers a `#[cfg]` that either does the listed implementation or switches to an implementation that emits a trap if executed. This means that it's safe to execute mismatched code where SIMD was enabled at compile time but disabled at runtime, it just means the semantics may be a bit surprising to debug.

Note that this SIMD is still enabled by default, and an explicit `--cfg` via `RUSTFLAGS` is required to compile-out the SIMD support. Cargo features aren't a great fit for this sort of feature so an explicit flag is used.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -624,6 +624,9 @@ jobs:
     - run: cargo check -p pulley-interpreter --all-features
       env:
         RUSTFLAGS: "--cfg pulley_tail_calls"
+    - run: cargo check -p pulley-interpreter --all-features
+      env:
+        RUSTFLAGS: "--cfg pulley_disable_interp_simd"
     - run: cargo test -p pulley-interpreter --all-features --release
       env:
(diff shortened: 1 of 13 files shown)


alexcrichton:
I gathered some numbers in https://github.com/google/wasefire/issues/458#issuecomment-2851490398 and for that particular embedding this is a 10% reduction in text size when compiled with `--release` and a 5% reduction in text size when compiled with `opt-level=z`

pchickey (approved):

PR #10729, opened 2025-05-05
x64: convert more `padd*` instructions

abrown:


abrown:
cc: @jlb6740, this is an example of how to plumb through new instructions: (1) define in the assembler, (2) connect up the right places in ISLE, (3) bless any file tests that may have any slight pretty-printing differences.

fitzgen (approved):

PR #10738, opened 2025-05-07
Run AArch64 Winch tests on CI

### LINKED ISSUE #9566: Winch support for AArch64 tough to integrate with `tests/wast.rs`
I've done some recent refactoring over the past month or so of the `tests/wast.rs` test suite. The general idea is that it should not only assert that passing tests pass but it should additionally assert that failing tests fail. This ensures that we keep an accurate picture of where tests are at and we know to enable spec tests once support is implemented.

This is problematic on Winch for AArch64, however. The problem is that when a test is supposed to fail it's run anyway and it's expected to return an error through `Result`. Failures such as panics and segfaults are not currently recoverable. Currently, though, tests on the AArch64 backend for Winch often panic (e.g. `todo!()` items) or segfault (likely because something isn't finished yet).

The way that `tests/wast.rs` works for now is that it pretends that all tests require the reference-types feature for wasm which is [hardcoded as "doesn't work on aarch64 winch"](https://github.com/bytecodealliance/wasmtime/blob/65181b360f0aae2d08961ac1f4ad6c8920f627c0/crates/wasmtime/src/config.rs#L1890-L1893). This causes all tests to fail early in `Config::validate` before the backend is used. This isn't great though because some tests do indeed pass on Winch right now and it might be reasonable to start ratcheting support to ensure it doesn't regress.

cc @saulecabrera and @vulc41n as y'all may be interested in this. I'll clarify that this is not an urgent issue by any means -- I don't intend to put pressure on either of y'all. Instead I wanted to make y'all aware of this if you're looking for future aarch64 work.

alexcrichton:
Most tests pass (yay!), some tests fail, and some tests crash and/or have nondeterministic results. A new `WastTest::should_skip_entirely` helper is added to avoid running crashing or nondeterministic tests in CI and that should enable all other tests to run in CI to ensure we have a ratchet to prevent future regressions. This should also in theory make testing easier locally with `cargo test --test wast`.

Closes #9566

### DIFF
--- a/crates/test-util/src/wast.rs
+++ b/crates/test-util/src/wast.rs
@@ -349,6 +349,8 @@ impl Compiler {
                     || config.gc_types()
                     || config.exceptions()
                     || config.legacy_exceptions()
+                    || (cfg!(target_arch = "aarch64")
+                        && (config.simd() || config.threads() || config.wide_arithmetic()))
             }
 
             Compiler::CraneliftPulley => config.threads() || config.legacy_exceptions(),
@@ -365,9 +367,7 @@ impl Compiler {
                     || cfg!(target_arch = "riscv64")
                     || cfg!(target_arch = "s390x")
             }
-            Compiler::Winch => {
-                cfg!(target_arch = "x86_64")
-            }
+            Compiler::Winch => cfg!(target_arch = "x86_64") || cfg!(target_arch = "aarch64"),
             Compiler::CraneliftPulley => true,
         }
     }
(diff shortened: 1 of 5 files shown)


alexcrichton:
After some follow-up discussion I think it's a bit premature for this so I'm going to close this.

saulecabrera:
I opened https://github.com/bytecodealliance/wasmtime/pull/10750, with some inspiration from this pull request and after our discussion in our meeting today. Let me know what you think!

PR #10739, opened 2025-05-07
Inline assembler-x64 `generated_files` in `main.rs`

adambratschikaye:
The public function `generated_files` in `cranelift-assembler-x64` makes the generated `rlib` non-deterministic because it contains the full paths of generated files. But this function is only used in `main.rs` of the same crate, so this change inlines it there to keep the library artifact deterministic while maintaining the same behavior.


adambratschikaye:
I think this should be the last PR needed to fix our build determinism/bazel caching issues.

abrown (approved):

PR #10742, opened 2025-05-07
[33.0.0] Add release notes

alexcrichton:

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -4,6 +4,40 @@ Unreleased.
 
 ### Added
 
+* Cranelift now has initial support for `try_call` and `try_call_indirect`
+  instructions, to be used in the future for the WebAssembly exception-handling
+  proposal. Wasmtime does not yet implement this proposal yet.
+  [#10510](https://github.com/bytecodealliance/wasmtime/pull/10510)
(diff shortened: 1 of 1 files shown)


RELEASES.md
```diff
@@ -20,6 +54,63 @@ Unreleased.
...
+  [#10503](https://github.com/bytecodealliance/wasmtime/pull/10503)
+
+* Wasmtime's implementation of WASIp2 has moved to `wasmtime_wasi::p2` from the
+  root of the crate.
+  [#10073](https://github.com/bytecodealliance/wasmtime/pull/10073)
+
+* Wasmtime will no longer emit calls to Cranelift-defined "libcalls" and instead
+  everything goes through Wasmtime's libcall mechanism instead, paving the way
+  for a future change for more efficient stack limit checking in wasm.
+  [#10657](https://github.com/bytecodealliance/wasmtime/pull/10657)
```
cfallin, line 79: should we note here that this also removes the need to perform relocations on the code, improving load times and allowing better use of cached pages when mapping from disk?

cfallin (approved):
Thanks as always for writing these up!

RELEASES.md
alexcrichton: Good idea!

PR #10746, opened 2025-05-07
Move the `GetHost` trait used in `bindgen!` into Wasmtime

alexcrichton:
Turns out we don't actually need to generate this `GetHost` trait, we can instead have it live in one location with extra documentation. There are already extra bounds on the `Host` associated type at all call-sites so there's no need to additionally have trait bounds in the trait definition, meaning the trait definition is always the same and it can move within Wasmtime.

This shouldn't have any impact on any embedders today, it's just moving things around.


crates/wasmtime/src/runtime/component/get_host.rs
```diff
@@ -0,0 +1,255 @@
...
+///
+///     fn type_annotate<F>(f: F) -> F
+///         where F: Fn(&mut MyStoreState) -> MyBorrowedState<'_>
+///     {
+///         f
+///     }
+/// }
+/// ```
+///
+/// # Too-much Detail
```
dicej, line 118:
```suggestion
/// # Too much Detail
```

Nit: remove unnecessary hyphen

crates/wasmtime/src/runtime/component/get_host.rs
```diff
@@ -0,0 +1,255 @@
...
+/// ```ignore
+/// # struct MyStoreType { the_impl: MyType, }
+/// # struct MyType { /* ... */ }
+/// let my_closure = for<'a> |x: &'a mut MyStoreType| -> &'a mut MyType {
+///     &mut x.the_impl
+/// };
+/// ```
+///
+/// This syntax technically compiles but [is
+/// unstable](https://github.com/rust-lang/rust/issues/97362). Alternatively on
```
dicej, line 215:
```suggestion
/// unstable](https://github.com/rust-lang/rust/issues/97362). Alternatively one
```

crates/wit-bindgen/src/lib.rs
```diff
@@ -1423,36 +1423,35 @@ impl Wasmtime {
...
-                impl<F, T, D, O> {world_camel}ImportsGetHost<T, D> for F
-                where
-                    F: Fn(T) -> O + Send + Sync + Copy + 'static,
-                    O: {get_host_bounds},
-                {{
-                    type Host = O;
-                }}
-            "
-        );
+        // let get_host_bounds = if let CallStyle::Concurrent = self.opts.call_style() {
```
dicej, line 1426: Can we just remove this code rather than comment it out?

dicej (approved):

crates/wit-bindgen/src/lib.rs
alexcrichton: Oops yes that's my mistake!

PR #10747, opened 2025-05-07
Cranelift: update to regalloc2 0.12.2; support many return values.

### LINKED ISSUE #10741: Codegen fails when exporting a function with long result type
### Test Case

```wat
;; sample.wat
(module
  (type $t
    (func
      (result
        i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 ;; 10
        i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 ;; 20
        i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 ;; 30
        i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 ;; 40
        i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 ;; 50
        i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 ;; 60
        i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 ;; 70
        i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 ;; 80
        i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 ;; 90
        i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 ;; 100
        i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 ;; 110
        i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 ;; 120
        i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 ;; 130
        i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 ;; 140
        i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 ;; 150
        i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 ;; 160
        i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 ;; 170
        i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 ;; 180
        i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 ;; 190
        i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 ;; 200
        i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 ;; 210
        i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 ;; 220
        i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 ;; 230
        i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 ;; 240
        i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 ;; 250
        i32 i32 i32 i32                         ;; 254
      )
    )
  )
  (export "f" (func $f))
  (func $f (type $t) (unreachable))
)
```

### Steps to Reproduce

Execute the `wat` file above using the following command:
```sh
target/debug/wasmtime sample.wat
```

### Expected Results

Terminates normally

### Actual Results

Thread panic:
```
thread '<unnamed>' panicked at cranelift/codegen/src/isa/aarch64/inst/emit.rs:107:31:
called `Option::unwrap()` on a `None` value
```

### Versions and Environment

Wasmtime version or commit: wasmtime 34.0.0 (303b836a4 2025-05-06)

Operating system: MacOS (Sonoma, 14.5)

Architecture: arm64 (Apple Silicon, M3)

(Also confirmed at Ubuntu 20.04.6 LTS, x86_64)

### Extra Info

Long result type with specific length (at least 254) and exporting the function with that type triggers this bug, even without calling the function.

cfallin:
Prior versions of regalloc2 could not support more than 255 operands on an instruction, and together with the integrated return-value loads on call instructions introduced in #10502, this caused issues with calls with many returns. This PR upgrades to a version of RA2 that supports up to `2^16 - 1` operands per instruction (well in excess of the maximum of 1000 return/result values per Wasm's implementation limits, for example).

Fixes #10741.


tests/disas/many-return-values.wat
```diff
@@ -0,0 +1,43 @@
+;;! target = "x86_64"
+;;! test = "compile"
+
+(module
```
alexcrichton, line 4: Could you add this as a `tests/misc_testsuite/*.wast` test as well so it gets tested on all architectures automatically? I figure it's ok to avoid running the test, just having it there compiled everywhere seems good

alexcrichton (approved):

cfallin:
In attempting to update `cargo vet` metadata, I'm seeing that perhaps we've failed to update for new crates:

```
% cargo vet
ERROR   × There are some issues with your policy.audit-as-crates-io entries

Error:   × Some non-crates.io-fetched packages match published crates.io versions
  │   cranelift-srcgen:0.120.0
  │   wasmtime-wasi-tls:33.0.0
```

will need to work out exactly how we handle this again...

alexcrichton:
Do you have the latest `cargo vet` version? (0.10.1 for me)

Otherwise, before this PR, if you run `cargo vet` does it change anything or have any errors?

cfallin:
Yes, I have v0.10.1; and I actually get the same errors on `main`. I need to context-switch and don't have time to dig into the innards of cargo-vet now so if someone else knows how to fix this it would be much appreciated...

alexcrichton:
Looks like it needed a merge/rebase with main first

cfallin:
Ah, awesome, thanks -- that's what I get for not updating my local `main` with regular development!

PR #10750, opened 2025-05-07
winch(aarch64): Run wast test in CI

saulecabrera:
Closes https://github.com/bytecodealliance/wasmtime/issues/9566

This commit takes inspiration from
https://github.com/bytecodealliance/wasmtime/pull/10738/files

During the 2025-05-07 Cranelift meeting we discussed potential avenues to start running wast tests for winch-aarch64. One potential idea was to declare an allowlist of tests that should be exectued and potentially ignore everyhing else. Although that idea works, I decided to diverge a bit from it, in favor of introducing a very strict classification of the state of tests for aarch64, namely, in this commit tests are classified as:

* Run and expected to pass
* Run and exepected to fail in a recoverable way
* Don't run since it's known that they produce an unrecoverable error e.g., segafault.

This approach is probably more verbose than the one discussed in the meeting, however, I think it's easier to have a global view of that status for aarch64 and potentially other backends in the future.

@alexcrichton curious to see what do you think.

### DIFF
--- a/crates/test-macros/src/wasmtime_test.rs
+++ b/crates/test-macros/src/wasmtime_test.rs
@@ -292,7 +292,7 @@ fn expand(test_config: &TestConfig, func: Fn) -> Result<TokenStream> {
         if wasmtime_test_util::wast::Compiler::#strategy_ident.should_fail(&#test_config) {
             assert!(result.is_err());
         } else {
-            assert!(result.is_ok());
+            result.unwrap();
         }
             }
         };
--- a/crates/test-macros/src/wasmtime_test.rs
+++ b/crates/test-macros/src/wasmtime_test.rs
@@ -253,20 +253,16 @@ fn expand(test_config: &TestConfig, func: Fn) -> Result<TokenStream> {
             (quote! {}, quote! {})
         };
         let func_name = &func.sig.ident;
-        let expect = match &func.sig.output {
-            ReturnType::Default => quote! {},
-            ReturnType::Type(..) => quote! { .expect("test is expected to pass") },
+        match &func.sig.output {
+            ReturnType::Default => {
+                return Err(syn::Error::new(func_name.span(), "Expected `Restult<()>`"))
+            }
+            ReturnType::Type(..) => {}
         };
         let test_name = Ident::new(
             &format!("{}_{}", strategy_name.to_lowercase(), func_name),
             func_name.span(),
         );
-
-        let should_panic = if strategy.should_fail(&test_config.flags) {
-            quote!(#[should_panic])
-        } else {
-            quote!()
-        };
 
         let test_config = format!("wasmtime_test_util::wast::{:?}", test_config.flags)
             .parse::<proc_macro2::TokenStream>()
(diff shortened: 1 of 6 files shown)


crates/test-util/src/wast.rs
```diff
@@ -430,125 +486,188 @@ impl WastTest {
...
+                    "spec_testsuite/local_tee.wast",
+                    "spec_testsuite/loop.wast",
+                    "spec_testsuite/table_copy.wast",
+                    "spec_testsuite/table_init.wast",
+                    "misc_testsuite/custom-page-sizes/custom-page-sizes.wast",
+                    "misc_testsuite/winch/table_grow.wast",
+                    "spec_testsuite/proposals/custom-page-sizes/custom-page-sizes.wast",
+                    "misc_testsuite/memory64/more-than-4gb.wast",
+                    "misc_testsuite/call_indirect.wast",
+                ];
```
alexcrichton, line 543: Could the platform-independent tests, the ones that fail on all platforms, be deduplicated between here and x64?

alexcrichton (approved):

crates/test-util/src/wast.rs
saulecabrera: Sure, yeah. Added that in https://github.com/bytecodealliance/wasmtime/pull/10750/commits/0d4784bc61f6b7a5350b2e91d0ec7df85880cae3

saulecabrera:
The failure in CI happens because `wasmtime_test_util::wast::Compiler::should_fail` gets invoked from `wasmtime_test` macro, and as far as I understand, `#[cfg(target_arch="...")]` in macro code will resolve to the host, not to the target, which ends up returning `true` for `should_fail`, but since we're cross compiling (aarch64 -> x64), in reality the test passes and the `#[should_panic]` predicate is not met. 

Prior to this change this was not an issue, since we were only taking x64 into account for Winch.

alexcrichton:
Bah I was afraid of this happening at some point...

Could we perhaps delay [this](https://github.com/bytecodealliance/wasmtime/blob/19000eea134cea26bcff281ef6c4f918503cfe9d/crates/test-macros/src/wasmtime_test.rs#L265-L269) to happening at runtime? We've got the two structures there at test-run-time so we should be able to run the same code. I'd prefer to not catch panics so it means though that all tests would have to return `Result<()>` and the runtime result of `should_fail` would result in expecting an error or not, however.

saulecabrera:
Yeah, I'll try that. I remember we had discussed this, I just wasn't sure if there was a strong preference regarding catching panics.

alexcrichton:
Personally I'd say I have a weak preference, if that's the best solution maintenance-wise I'd go with catching panics

saulecabrera:
Ok, ended up going with emitting the call to `Compiler::should_fail` at runtime. Would you mind taking another look?

crates/test-macros/src/wasmtime_test.rs
```diff
@@ -293,7 +288,12 @@ fn expand(test_config: &TestConfig, func: Fn) -> Result<TokenStream> {
                         collector: wasmtime_test_util::wast::Collector::Auto,
                     },
                 );
-                #func_name(&mut config) #await_ #expect
+                let result = #func_name(&mut config) #await_;
+        if wasmtime_test_util::wast::Compiler::#strategy_ident.should_fail(&#test_config) {
+            assert!(result.is_err());
+        } else {
+            assert!(result.is_ok());
```
alexcrichton, line 295: Could this use `result.unwrap()` to print the error on failure?

alexcrichton (approved):

PR #10752, opened 2025-05-08
x64: convert scalar negation instructions

abrown:
This replaces all uses of `neg` and `not` with instructions from the new assembler.


cfallin (approved):
LGTM, thanks!

PR #10755, opened 2025-05-09
Classify Winch with Tier 1 support

alexcrichton:
This commit classifies the Winch compiler in our documentation as a Tier 1 feature. This reflects the degree of work and investment put into it as well as the thorough quality vetting that has already happened in production. For Wasmtime this means that bugs in Winch are candidates for security releases as well.

This additionally refines our tiers-of-support documentation to reflect how not all compilers are required to support all features. Tables are added of per-architecture support for each compiler as well as refined wording of the various requirements of what is required for each tier as well.

### DIFF
--- a/docs/stability-tiers.md
+++ b/docs/stability-tiers.md
@@ -25,7 +25,8 @@ For explanations of what each tier means see below.
 | Target               | `x86_64-apple-darwin`                      |
 | Target               | `x86_64-pc-windows-msvc`                   |
 | Target               | `x86_64-unknown-linux-gnu`                 |
-| Compiler Backend     | Cranelift                                  |
+| Compiler             | Cranelift [^support]                       |
+| Compiler             | Winch [^support]                           |
 | WebAssembly Proposal | [`mutable-globals`]                        |
 | WebAssembly Proposal | [`sign-extension-ops`]                     |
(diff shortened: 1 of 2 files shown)


alexcrichton:
cc @saulecabrera

docs/stability-tiers.md
```diff
@@ -259,32 +410,43 @@ well-maintained, tested well, but don't necessarily meet the stringent criteria
...
+
+* **Target**: tests are run in CI for this platform with a compiler, either
+  directly or via emulation. This ensures that all changes to Wasmtime are
+  tested against this target, and this additionally means that all applicable
+  Tier 1/2 features are tested on this target in CI.
+
+* **Target**: Complete implementations for anything that's part of Tier 1. For
+  example all Tier 2 targets must implement all of the applicable Tier 1
+  WebAssembly proposals, and all Tier 2 features must be implemented on all Tier
+  1 targets.
```
saulecabrera, line 423: Small note on formatting: could these two be included under a single Target bullet point? e.g., 

* Target
  * Tests are run ... 
  * Complete implementation ...


In the rendered version, it took me a bit to understand the current formatting, but if I'm understanding correctly, they are meant to be two separate points under the same Target classification (similar above with Maintenance).

saulecabrera (approved):

PR #10756, opened 2025-05-09
[s390x] Use VCodeConstant literal pool

uweigand:
This switches s390x to use the common-code VCodeConstant literal pool instead of constants inlined into the instruction stream.


cfallin (approved):
LGTM -- thanks!

PR #10759, opened 2025-05-09
asm: implement write-only operands

abrown:
This change adds support to the new assembler for write-only operands. This implementation appeared first in [#10754] but is split out here to unblock implementation of instructions that require it: multiplication, conversions, moves, etc. This starts roughly the same as what was implemented for write-only XMMs in [#10754] but includes support for write-only GPRs as well and generates the temporary registers which are needed.

[#10754]: https://github.com/bytecodealliance/wasmtime/pull/10754

### DIFF
--- a/cranelift/codegen/meta/src/gen_asm.rs
+++ b/cranelift/codegen/meta/src/gen_asm.rs
@@ -132,7 +132,7 @@ pub fn generate_macro_inst_fn(f: &mut Formatter, inst: &Inst) {
                         OperandKind::Reg(r) => {
                             let ty = r.reg_class().unwrap().to_string();
                             let var = ty.to_lowercase();
-                            fmtln!(f, "let {var} = {r}.as_ref().clone();");
+                            fmtln!(f, "let {var} = {r}.as_ref().to_reg();");
                             fmtln!(f, "AssemblerOutputs::Ret{ty} {{ inst, {var} }}");
                         }
                         _ => unimplemented!(),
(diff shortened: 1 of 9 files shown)


cranelift/codegen/meta/src/gen_asm.rs
```diff
@@ -117,6 +126,17 @@ pub fn generate_macro_inst_fn(f: &mut Formatter, inst: &Inst) {
...
+                        // One write-only register output? Output the
+                        // instruction and that register.
+                        OperandKind::Reg(r) => {
+                            let ty = r.reg_class().unwrap().to_string();
+                            let var = ty.to_lowercase();
+                            fmtln!(f, "let {var} = {r}.as_ref().clone();");
+                            fmtln!(f, "AssemblerOutputs::Ret{ty} {{ inst, {var} }}");
+                        }
+                        _ => unimplemented!(),
+                    },
```
alexcrichton, line 139: Should this get collapsed into `ReadWrite` as well below? This has the same handling for `Reg` and all the other cases look applicable for one-operand-written instructions too (e.g. if something stores in memory or outputs to a fixed register)

alexcrichton (approved):

cranelift/codegen/meta/src/gen_asm.rs
abrown: Unfortunately, there's slightly different behavior for retrieving the underlying Cranelift type. The `ReadWrite` variant needs to access the `write` field; this one should actually be doing a `as_ref().to_reg()` (let me fix that). I think in the fullness of time we will want to merge some things like I did over in the assembler generator but let's leave this as-is until we have a few more write-only instructions implemented.

PR #10762, opened 2025-05-09
x64: convert some SSE shift instructions

abrown:


cranelift/codegen/src/isa/x64/lower/isle.rs
```diff
@@ -1035,6 +1035,13 @@ impl Context for IsleContext<'_, '_, MInst, X64Backend> {
         }
     }
 
+    fn is_imm8_xmm(&mut self, src: &XmmMemImm) -> Option<u8> {
+        match src.clone().to_reg_mem_imm() {
+            RegMemImm::Imm { simm32 } => Some(u8::try_from(simm32).ok()?),
+            _ => None,
+        }
+    }
```
abrown, line 1043: I don't really want to add this but, because of the type signatures of the `inst.isle` constructors I have to find a way to match on a `XmmMemImm` instead of `GprMemImm`. One alternative to adding this would be to have `is_imm8` examine the more general `RegMemImm` instead, but this would mean the constructor types would have to change. Thoughts?

alexcrichton: Poking around it looks like these shifting instructions are the only ones using `XmmMemImm`, so perhaps we could drop that type entirely? All the various helpers seem to be using `mask_xmm_shift` and the current lowerings could dispatch to a helper taking the `RegMemImm` result of `mask_xmm_shift` which handles the `*Imm` case to delegate to different constructors than when in the `Reg` or `Mem` case?

cfallin (approved):
LGTM with Alex's suggestion -- thanks!

alexcrichton:
I can also clarify I'm happy to see this landed as-is. @abrown mentioned he wanted to chat with me about my suggestion but I'm busy all this week and won't be back until next week so I think it's ok to land this as-is and follow-up as necessary if you'd like to @abrown

alexcrichton:
that was the wrong button...

PR #10763, opened 2025-05-10
winch: Implement check_stack for aarch64

graydon:
This is my first attempt at contributing to winch or cranelift at all -- wide open to any feedback/suggestions. I was just poking around looking for things I could do to help push winch over the finish line on aarch64. @saulecabrera suggested the missing `check_stack` implementation might be a welcome change.

(This is also my first time working with aarch64 instructions in any detail; hopefully I got the semantics not-too-wrong)

Before:

```
$ cargo run -- wast -Ccompiler=winch tests/spec_testsuite/call.wast
[...]
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.41s
     Running `target/debug/wasmtime wast -Ccompiler=winch tests/spec_testsuite/call.wast`
execution on async fiber has overflowed its stackzsh: abort      cargo run -- wast -Ccompiler=winch tests/spec_testsuite/call.wast
```

After:

```
$ cargo run -- wast -Ccompiler=winch tests/spec_testsuite/call.wast
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.41s
     Running `target/debug/wasmtime wast -Ccompiler=winch tests/spec_testsuite/call.wast`
Error: failed to run script file 'tests/spec_testsuite/call.wast'

Caused by:
    0: failed directive on tests/spec_testsuite/call.wast:285:1
           0: <unknown>!<wasm function 17>
    2: wasm trap: call stack exhausted
```


alexcrichton:
(moving review over to @saulecabrera)

Looks like you found `WASMTIME_TEST_BLESS=1` though which is indeed the recommend way to fix the original test failures 👍

winch/codegen/src/isa/aarch64/asm.rs
```diff
@@ -1284,3 +1288,81 @@ impl Assembler {
...
+    // The destination register for the add instruction.
+    reg: Reg,
+
+    // The temporary register used to hold the immediate value.
+    tmp: Reg,
+}
+
+impl PatchableAddToReg {
+    /// Create a new [`PatchableAddToReg`] by capturing a region in the output buffer where the
+    /// add-with-immediate sequence occurs. The [`MachBuffer`] will have and add-with-immediate instruction
```
saulecabrera, line 1308: Perhaps we should update the documentation here, since the implementation here differs a bit from x64, e.g., we're emitting `add rd, rm, rn`

winch/codegen/src/isa/aarch64/masm.rs
```diff
@@ -114,8 +141,43 @@ impl Masm for MacroAssembler {
...
+                ptr_size.vmstore_context_stack_limit().into(),
+            ),
+            writable!(scratch_stk_limit),
+        )?;
+
+        self.add_stack_max(scratch_stk_limit, scratch_tmp);
+
+        self.asm
+            .subs_rrr(scratch_stk_limit, regs::sp(), OperandSize::S64);
+        self.asm.trapif(Cond::Gt, TrapCode::STACK_OVERFLOW);
```
saulecabrera, line 179: Here we might want to use `self.masm.trapif` instead. Aarch64 requires the SP to be 16-byte aligned, and even though the SP is not used for memory addressing, if a trap is raised, signal handling code will segfault if SP is not correctly aligned. See https://github.com/bytecodealliance/wasmtime/pull/10146 for more details.

saulecabrera (commented):
This is looking really good, thanks! I left a couple of comments inline, which I think is probably worth addressing. 

In terms of tests, since our last discussion via Zulip, https://github.com/bytecodealliance/wasmtime/pull/10750 landed, which enables running some aarch64 tests in CI. It's probably a good idea to enable at least the `call.wast` tests here https://github.com/bytecodealliance/wasmtime/blob/main/crates/test-util/src/wast.rs#L422. Ideally, we should enable all of them, but it's possible that some of them will fail, due to some of the existing bugs that I'm currently in the process of fixing.

winch/codegen/src/isa/aarch64/masm.rs
graydon: Oh, sure (`self` already is a `MacroAssembler` so I assume you mean `self.trapif`).

winch/codegen/src/isa/aarch64/asm.rs
graydon: Oh, ok. I already did alter it (note it describes an instruction _sequence_) but I can alter it to be a little more explicit.

winch/codegen/src/isa/aarch64/asm.rs
```diff
@@ -1284,3 +1288,81 @@ impl Assembler {
...
+pub(crate) struct PatchableAddToReg {
+    /// The region to be patched in the [`MachBuffer`]. It contains
+    /// space for 3 32-bit instructions, i.e. it's 12 bytes long.
+    region: PatchRegion,
+
+    // The destination register for the add instruction.
+    reg: Reg,
+
+    // The temporary register used to hold the immediate value.
+    tmp: Reg,
```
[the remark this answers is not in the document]
graydon: Sure, done.

graydon:
@saulecabrera thanks for the review! one thing worth pointing out is that this change doesn't include the bit about emitting unwind instructions that's in the x86_64 masm. I got the impression that's a larger project though (there doesn't seem to be any unwind support in the aarch64 backend at all right now -- is that something else that needs doing?)

saulecabrera:
@graydon You're right that there is no unwind support right now, it's not expected to land that functionality as part of this PR. I think that once that (any passing) tests are updated here  https://github.com/bytecodealliance/wasmtime/blob/main/crates/test-util/src/wast.rs#L422, we should be able to land this one.

saulecabrera (approved):
LGTM, thanks!

PR #10765, opened 2025-05-11
c-api: new wasmtime_trap_new_code function

dotcarmen:
right now, returning wasmtime-recognized traps from functions is only available to the Rust API. this PR adds a new function `wasmtime_trap_new_code` to the c api that returns a new `wasm_trap_t` with the `Trap` value associated with the given code.

### DIFF
--- a/crates/c-api/include/wasmtime/trap.hh
+++ b/crates/c-api/include/wasmtime/trap.hh
@@ -132,9 +132,7 @@ public:
       : Trap(wasmtime_trap_new(msg.data(), msg.size())) {}
 
   /// Creates a new trap with the given wasmtime trap code.
-  Trap(wasmtime_trap_code_enum code) : Trap(wasmtime_trap_new_code(code)) {
-    assert(ptr != nullptr);
-  }
+  Trap(wasmtime_trap_code_enum code) : Trap(wasmtime_trap_new_code(code)) {}
 
   /// Returns the descriptive message associated with this trap.
   std::string message() const {
--- a/crates/c-api/include/wasmtime/trap.hh
+++ b/crates/c-api/include/wasmtime/trap.hh
@@ -5,8 +5,6 @@
 #ifndef WASMTIME_TRAP_HH
 #define WASMTIME_TRAP_HH
 
-#include "trap.h"
-#include <cstdlib>
 #include <wasmtime/error.hh>
 #include <wasmtime/trap.h>
 
--- a/crates/c-api/include/wasmtime/trap.hh
+++ b/crates/c-api/include/wasmtime/trap.hh
@@ -5,6 +5,8 @@
 #ifndef WASMTIME_TRAP_HH
 #define WASMTIME_TRAP_HH
 
+#include "trap.h"
+#include <cstdlib>
 #include <wasmtime/error.hh>
 #include <wasmtime/trap.h>
 
@@ -129,7 +131,12 @@ public:
   explicit Trap(std::string_view msg)
       : Trap(wasmtime_trap_new(msg.data(), msg.size())) {}
 
-  /// Returns the descriptive message associated with this trap
+  /// Creates a new trap with the given wasmtime trap code.
+  Trap(wasmtime_trap_code_enum code) : Trap(wasmtime_trap_new_code(code)) {
+    assert(ptr != nullptr);
+  }
+
+  /// Returns the descriptive message associated with this trap.
   std::string message() const {
     wasm_byte_vec_t msg;
     wasm_trap_message(ptr.get(), &msg);
--- a/crates/c-api/src/trap.rs
+++ b/crates/c-api/src/trap.rs
@@ -66,11 +66,11 @@ pub unsafe extern "C" fn wasmtime_trap_new(message: *const u8, len: usize) -> Bo
 }
 
 #[unsafe(no_mangle)]
-pub unsafe extern "C" fn wasmtime_trap_new_code(code: u8) -> Option<Box<wasm_trap_t>> {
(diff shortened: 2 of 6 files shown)


crates/c-api/include/wasmtime/trap.hh
```diff
@@ -129,7 +130,13 @@ public:
   explicit Trap(std::string_view msg)
       : Trap(wasmtime_trap_new(msg.data(), msg.size())) {}
 
-  /// Returns the descriptive message associated with this trap
+  /// Creates a new trap with the given wasmtime trap code.
+  Trap(wasmtime_trap_code_t code)
+      : Trap(wasmtime_trap_new_code(code)) {
+    assert(ptr != nullptr);
```
dotcarmen, line 136: i kinda figured this is the appropriate way to handle this, if a different way is preferred i can change it :)

crates/c-api/include/wasmtime/trap.hh
```diff
@@ -129,7 +131,13 @@ public:
   explicit Trap(std::string_view msg)
       : Trap(wasmtime_trap_new(msg.data(), msg.size())) {}
 
-  /// Returns the descriptive message associated with this trap
+  /// Creates a new trap with the given wasmtime trap code.
+  Trap(wasmtime_trap_code_enum code)
```
dotcarmen, line 135: i'm not sure if this is a good idea or if it's desired, i can change it if not :)

dotcarmen, line 135: i'm not sure if this is a good idea or if it's desired, i can change it if not :)

crates/c-api/include/wasmtime/trap.hh
```diff
@@ -129,7 +131,13 @@ public:
...
+      : Trap(wasmtime_trap_new_code(code)) {
+    assert(ptr != nullptr);
```
dotcarmen, line 137: same as above, i'm not sure if `assert` is the preferred way to handle this

crates/c-api/src/trap.rs
```diff
@@ -65,6 +65,14 @@ pub unsafe extern "C" fn wasmtime_trap_new(message: *const u8, len: usize) -> Bo
     })
 }
 
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn wasmtime_trap_new_code(code: u8) -> Option<Box<wasm_trap_t>> {
```
dotcarmen, line 69: returning a null pointer is kinda an anti-pattern in the wasmtime c-api afaict, but i also think it makes sense here. lmk if a different pattern is preferred

alexcrichton: Another alternative might be to `.unwrap()` here, effectively asserting that the code is valid?

alexcrichton (commented):
Thanks! Would you be up for sync-ing the implementation in C and Rust? The trap code listing [here](https://github.com/bytecodealliance/wasmtime/blob/388b3ff4264c5f9736ab2f5f2e879f8c38620b97/crates/c-api/src/trap.rs#L132) already doesn't match the [Rust definition](https://github.com/bytecodealliance/wasmtime/blob/9d40c6eb67d0d791c68fae46a2059647ca0e5dbc/crates/environ/src/trap_encoding.rs#L26), so using `from_u8` I don't think will work for the out-of-fuel trap for example. The convert-to-`u8` can probably use `trap as u8` nowadays, I'm not sure it was possible to do that when it was originally written. That should make the `from_u8` conversion work as well.

Additionally would you be up for adding a test for this? Somewhere within [this file](https://github.com/bytecodealliance/wasmtime/blob/388b3ff4264c5f9736ab2f5f2e879f8c38620b97/crates/c-api/tests/trap.cc) is probably a good place to go.

crates/c-api/tests/trap.cc
```diff
@@ -33,4 +33,10 @@ TEST(Trap, Smoke) {
               std::string::npos);
   EXPECT_EQ(func.call(store, {1}).err().message(),
             "expected 0 arguments, got 1");
+
+  Trap out_of_fuel(WASMTIME_TRAP_CODE_OUT_OF_FUEL);
+  EXPECT_EQ(out_of_fuel.code(),
+            WASMTIME_TRAP_CODE_OUT_OF_FUEL);
+  EXPECT_EQ(out_of_fuel.message(),
+            "so much fuel");
```
dotcarmen, line 41: hmmm apparently my test command doesn't run the tests in this dir, @alexcrichton what's the command to run these tests? i can document it in c-api/README as well :)

dotcarmen: nvm, found it from ci :)

crates/c-api/include/wasmtime/trap.hh
```diff
@@ -5,6 +5,8 @@
 #ifndef WASMTIME_TRAP_HH
 #define WASMTIME_TRAP_HH
 
+#include "trap.h"
```
alexcrichton, line 8: I'm not actually sure how this works, it probably means that our include paths are set up incorrectly? This should already be taken care of below by `<wasmtime/trap.h>` I think though?

alexcrichton (approved):

crates/c-api/include/wasmtime/trap.hh
dotcarmen: Ugh I think my LSP keeps sneaking that in. I removed this line before

dotcarmen: removed both accidental includes in 5e2dc3e

alexcrichton (approved):

PR #10768, opened 2025-05-12
Add support sqrtps, sqrtpd, sqrtss, sqrtsd

jlb6740:


fitzgen (approved):
LGTM, @abrown did you want to give a look at this, since you're much more familiar with the new x64 assembler?

jlb6740:
@abrown This is ready for review for sqrtps and sqrtpd. The sqrtss and sqrtsd will be included in a follow-up patch. Those isle instructions use two operands as a special case with the explanation that data dependencies needed to be captured for the non-packed case. It's not clear that those dependencies were being captured by my patch so instead of blessing the file-test update, I removed support for those instructions to get support for these instructions pushed now to avoid constant rebasing. Will follow-up quickly with another patch for sqrtss and sqrtsd once the changes needed there are understood.

abrown (approved):

PR #10774, opened 2025-05-14
[s390x] Add full support for f128

uweigand:
This adds all missing operations to fully support f128 at the same level as f64 and f32 on s390x.  The implementation is mostly a straightfoward extension of existing support, using the same vector instruction set.  The exception is conversion between integer and f128, where we have to use the older floating-point instructions that use FPR register pairs to hold a f128 value.  As regalloc does not support pairs, those registers are hard-coded, just as is done for the few instructions that require GPR register pairs.


cfallin (approved):
Thanks for this implementation -- looks good!

PR #10780, opened 2025-05-15
Add a `--generate-dwarf` flag to the `wast` subcommand

alexcrichton:
This commit adds support for the `wast` subcommand to generate DWARF debugging information for the text-to-binary translation that happens. This leans on the support within the `wast` crate already and enables more easily seeing where a trap is happening during development of tests due to the fact that line numbers are available as opposed to just having binary offsets with a stack trace.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1543,9 +1543,9 @@ dependencies = [
 
 [[package]]
 name = "gimli"
-version = "0.31.0"
+version = "0.31.1"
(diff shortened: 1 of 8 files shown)


alexcrichton:
On second thought this is now an on-by-default option, I'm not sure if there's a need to disable it...

alexcrichton:
Oops just realized I closed this by accident...

fitzgen (approved):
Nice

PR #10782, opened 2025-05-15
x64: convert all multiplication-related instructions

abrown:
Multiplication on x86 is a bit tricky: there are a few different instruction formats, some gaps, and some implicit registers holding different sets of bits (`rax`, `rdx`). This PR converts all x64 backend multiplications to use the new assembler and, in doing so, adds support for implicit registers: these are operands that need to be tracked during register allocation but may not show up in instruction disassembly. There is more that _could_ be done to refactor these multiplication rules but that could come later.


cranelift/codegen/src/isa/x64/inst.isle
```diff
@@ -2933,49 +2944,43 @@
...
-      (let ((dst WritableGpr (temp_writable_gpr))
-            (size OperandSize (raw_operand_size_of_type ty))
-            (_ Unit (emit (MInst.IMulImm size src1 src2 dst))))
-        dst))
+(rule 2 (x64_imul_imm $I16 src1 (is_rust_simm8 src2)) (x64_imulw_rmi_sxb src1 src2))
+(rule 2 (x64_imul_imm $I32 src1 (is_rust_simm8 src2)) (x64_imull_rmi_sxb src1 src2))
+(rule 2 (x64_imul_imm $I64 src1 (is_rust_simm8 src2)) (x64_imulq_rmi_sxb src1 src2))
+(rule 1 (x64_imul_imm $I16 src1 (is_rust_imm16 src2)) (x64_imulw_rmi src1 src2))
+(rule 1 (x64_imul_imm $I32 src1 (is_rust_imm32 src2)) (x64_imull_rmi src1 src2))
+(rule 1 (x64_imul_imm $I64 src1 src2)                 (x64_imulq_rmi_sxl src1 src2))
```
abrown, line 2969: Once again (as in #10762), we run into ISLE difficulties matching immediates. I did put up #10775, but I'm not really sure if it helps all that much. Another option to avoid these extra conversions would be to wrap up  the `i32` in a `RegMemImm.Imm` and then immediately unwrap it, but that confuses the type declaration. Not sure what to do, but it seems this is ripe for some refactoring...

cranelift/assembler-x64/meta/src/generate/inst.rs
```diff
@@ -194,7 +194,7 @@ impl dsl::Inst {
                 f.add_block(
                     "fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result",
                     |f| {
-                        for op in &self.format.operands {
+                        for op in self.format.operands.iter().filter(|o| !o.implicit) {
```
alexcrichton, line 197: One thing that can be a bit unfortunate here is that this implementation hides information of pre-register-allocation instructions that have implicit operands and the operands won't get printed. That seems like it could create a pretty surprising situation where an instruction uses an vreg operand but it doesn't show up in the disassembly.

Would it be possible to skip printing the operands post-regalloc but print them, maybe in a comment, pre-regalloc?

cranelift/codegen/src/isa/x64/inst.isle
```diff
@@ -1829,12 +1829,18 @@
 (extern extractor is_imm8_xmm is_imm8_xmm)
 (decl is_simm8 (i8) GprMemImm)
 (extern extractor is_simm8 is_simm8)
+(decl is_rust_simm8 (i8) i32)
+(extern extractor is_rust_simm8 is_rust_simm8)
```
alexcrichton, line 1833: We have a number of conversions along these lines in `prelude.isle` of varying names (e.g. `u8_try_from_u64`)

Could the conversions added here purely between integer types get moved over to `prelude.isle` and named similarly?

cranelift/codegen/src/isa/x64/inst.isle
alexcrichton: I'm a bit confused, how is this related to #10762? Here the argument is `i32`, not a `*MemImm`, so this to me at least looks like a normal what I'd expect "match the size/value of the immediate and emit an appropriate instruction"

abrown: What I mean is that it's not easy to line up all the types so that matching can be clear and simple; this is not a problem with ISLE per se but with the network of rules that currently exist in the `*.isle` files.

alexcrichton: Hm I'm not sure I understand still, but nevertheless this whole PR looks quite reasonable to me so I wouldn't have any qualms about landing it myself.

cranelift/assembler-x64/meta/src/generate/inst.rs
abrown: Yup, made up some new syntax: `... ;; implicit: %rax`.

cranelift/codegen/src/isa/x64/inst.isle
abrown: Moved.

alexcrichton (approved):

PR #10785, opened 2025-05-16
Update MSRV to 1.85.0

alexcrichton:
This makes us eligible to update to the 2024 Rust edition but I'm planning on deferring that to a separate follow-up change.

### DIFF
--- a/cranelift/filetests/src/function_runner.rs
+++ b/cranelift/filetests/src/function_runner.rs
@@ -114,7 +114,7 @@ impl TestFileCompiler {
             unsafe extern "C" {
                 safe fn cosf(f: f32) -> f32;
             }
-            let f = 1.2_f32;
+            let f = std::hint::black_box(1.2_f32);
             assert_eq!(f.cos(), cosf(f));
         }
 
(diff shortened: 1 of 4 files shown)


abrown (approved):
👍 but I guess there's a few extra warnings to handle now.

cranelift/filetests/src/function_runner.rs
```diff
@@ -114,7 +114,7 @@ impl TestFileCompiler {
             unsafe extern "C" {
                 safe fn cosf(f: f32) -> f32;
             }
-            let f = 1.2_f32;
+            let f = std::hint::black_box(1.2_f32);
```
alexcrichton, line 117: Digging into this, it looks like LLVM sees in the assertion below `assert_eq!(cosf(x), cosf(x))` and then LLVM has built-in knowledge that `cosf` is pure so it's optimized out entirely. That means libm doesn't get linked at all, meaning later `dlsym` calls fail.

Using `black_box` here is apparently enough, but unsure how long it'll last necessarily.

PR #10787, opened 2025-05-16
Update cc crate dependency

alexcrichton:
Looks to fix an issue coming up on CI and MinGW

prtest:full


pchickey (approved):
Doesn't look 100% resolved yet but once it goes green its good

alexcrichton:
Yeah this is not a valid fix, local testing fooled me into thinking it was but I've been duped

alexcrichton:
Ok the true fix is in https://github.com/bytecodealliance/wasmtime/pull/10790 but I figure may as well land this since I went ahead and did the audit. I'll rebase once that lands.

PR #10790, opened 2025-05-17
Migrate run-tests.sh to Python

alexcrichton:
Try to fix MinGW breakage in CI

prtest:full

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -812,9 +812,15 @@ jobs:
       if: matrix.filter == 'linux-x64' && contains(matrix.bucket, 'wasmtime-cli')
       uses: abrown/install-vtune-action@v1
 
-    # Build and test all features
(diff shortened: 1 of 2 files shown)


alexcrichton:
Another attempt after https://github.com/bytecodealliance/wasmtime/pull/10787

Some testing has shown that `gcc -E foo.c` fails when the `shell: bash` parameter is set in CI. All other shells work though. I don't know what's going on so I just want to get out of bash/MinGW and try to get to the point where we just execute the compiler in isolation.

alexcrichton:
Ok I'm clearly too tired for this.

alexcrichton:
Ok I wasn't able to use `shell: python` but I was able to use a Python script with the `pwsh` shell to avoid unduly modifying PATH on MinGW

cfallin (approved):
Wow -- everything about "we can't use `bash` because it breaks the environment on the pre-set-up runner" is cursed, but looks like this works, but this seems to work so ship it!

Issue #10792, opened 2025-05-18
Unclear documentation for `--invoke` flag syntax and return value display


koke1997:
## Description
While working with Wasmtime's CLI, I discovered undocumented behavior differences between the two syntaxes for invoking WebAssembly functions:

- `wasmtime add.wasm --invoke add 1 2` (space syntax) - executes the function but doesn't display its return value
- `wasmtime --invoke=add add.wasm 1 2` (equals sign syntax) - executes the function AND displays its return value

Both commands show warnings that this functionality is experimental, but the CLI documentation doesn't explain this syntax difference or mention that return value display is only enabled with the equals sign variant.

## Example
Using a simple WebAssembly module with an `add` function that returns the sum of two integers:

```wat
(module
  (func $add (param $lhs i32) (param $rhs i32) (result i32)
     (i32.add
       (local.get $lhs)
       (local.get $rhs)
     )
  )
  (export "add" (func $add))
)

```

Compiled with `wat2wasm add.wat`, then:

```bash
# This command shows no return value
$ wasmtime add.wasm --invoke add 1 2

# This command shows '3' as the return value
$ wasmtime --invoke=add add.wasm 1 2
warning: using `--invoke` with a function that takes arguments is experimental and may break in the future
warning: using `--invoke` with a function that returns values is experimental and may break in the future
3

```

## Suggested Documentation Improvement

I suggest updating the CLI documentation at https://docs.wasmtime.dev/cli-options.html to explain:

1.  The difference between space syntax (`--invoke function`) and equals sign syntax (`--invoke=function`)
2.  That function return values are only displayed with the equals sign syntax
3.  That this feature is experimental and may change in future versions

This would help new users understand the expected behavior and avoid confusion when trying to view function return values.

I noticed this while doing a course from Linux Foundation , since resource mentions the first way of executing command without equal sign - https://trainingportal.linuxfoundation.org/learn/course/webassembly-components-from-cloud-to-edge-lfd134/webassembly-from-scratch/exploring-webassembly?page=7

## Additional Context

-   Wasmtime version: 32.0.0 (the current version I'm using)
-   OS: Linux Ubuntu (but should be consistent across platforms)

alexcrichton:
Hello and thanks for the report! I think there's some confusion going on here, although I'm not sure how best to resolve it.

[...]

PR #10793, opened 2025-05-18
Update CLI `run` documentation

alexcrichton:
Try to take the spirit of #10792 to improve our documentation for the `run` command and how CLI arguments interact with the provided WebAssembly module.

### DIFF
--- a/docs/cli-options.md
+++ b/docs/cli-options.md
@@ -44,25 +44,92 @@ as well as the text format for WebAssembly (`*.wat`):
 wasmtime foo.wat
 ```
 
-**Wasm Modules**
+#### Running "Command" Wasms
 
-A Wasm **module** exports raw functions directly. The `run` command accepts an optional `--invoke` argument, which is the name of an exported raw function (of the module) to run:
+WebAssembly modules or components can behave like a "command" which means
+they're intended to look like a normal OS executable with a `main` function and
+run once to completion. This is the default mode of running a wasm provided to
+Wasmtime.
+
+For core WebAssembly modules this means that the function exported as an empty
+string, or the `_start` export, is invoked. For WebAssembly components this
+means that the `wasi:cli/run` interface is executed.
+
+For both core modules and components CLI arguments are passed via WASI. Core
+modules receive arguments via WASIp1 APIs and components receive arguments via
+WASIp2 or later APIs. Arguments, flags, etc, are passed to the WebAssembly file
+after the file itself. For example:
+
+```console
+wasmtime foo.wasm --bar baz
+```
+
+Will pass `["foo.wasm", "--bar", "baz"]` as the list of arguments to the module.
+Note that flags for Wasmtime must be passed before the WebAssembly file, not
+afterwards. For example
+
+```console
+wasmtime foo.wasm --dir .
+```
+
+will pass `--dir .` to the `foo.wasm` program, not Wasmtime. If you want to
+mount the current directory you instead need to invoke
+
+```console
+wasmtime --dir . foo.wasm
+```
+
+All Wasmtime options must come before the WebAssembly file provided. All
+arguments afterwards are passed to the WebAssembly file itself.
+
+#### Running Custom Module exports
+
+If you're not running a "command" but want to run a specific export of a
+WebAssembly core module you can use the `--invoke` argument:
 
 ```console
 wasmtime run --invoke initialize foo.wasm
 ```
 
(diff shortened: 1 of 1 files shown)


docs/cli-options.md
```diff
@@ -44,25 +44,92 @@ as well as the text format for WebAssembly (`*.wat`):
 wasmtime foo.wat
 ```
 
-**Wasm Modules**
+#### Running "Command" Wasms
 
-A Wasm **module** exports raw functions directly. The `run` command accepts an optional `--invoke` argument, which is the name of an exported raw function (of the module) to run:
+WebAssembly modules or components can behave like a "command" which means
```
abrown, line 49: Just wondering: isn't there some WASI repository page somewhere that we can link to here in case anyone wants to dig into the difference between command and reactor components?

docs/cli-options.md
```diff
@@ -44,25 +44,92 @@ as well as the text format for WebAssembly (`*.wat`):
...
+they're intended to look like a normal OS executable with a `main` function and
+run once to completion. This is the default mode of running a wasm provided to
+Wasmtime.
+
+For core WebAssembly modules this means that the function exported as an empty
+string, or the `_start` export, is invoked. For WebAssembly components this
+means that the `wasi:cli/run` interface is executed.
+
+For both core modules and components CLI arguments are passed via WASI. Core
```
abrown, line 58:
```suggestion
For both core modules and components, CLI arguments are passed via WASI. Core
```

docs/cli-options.md
```diff
@@ -44,25 +44,92 @@ as well as the text format for WebAssembly (`*.wat`):
...
+modules receive arguments via WASIp1 APIs and components receive arguments via
+WASIp2 or later APIs. Arguments, flags, etc, are passed to the WebAssembly file
```
abrown, line 60:
```suggestion
WASIp2 or later APIs. Arguments, flags, etc., are passed to the WebAssembly file
```

docs/cli-options.md
```diff
@@ -44,25 +44,92 @@ as well as the text format for WebAssembly (`*.wat`):
...
+after the file itself. For example:
+
+```console
+wasmtime foo.wasm --bar baz
+```
+
+Will pass `["foo.wasm", "--bar", "baz"]` as the list of arguments to the module.
```
abrown, line 67:
```suggestion
will pass `["foo.wasm", "--bar", "baz"]` as the list of arguments to the module.
```

docs/cli-options.md
```diff
@@ -44,25 +44,92 @@ as well as the text format for WebAssembly (`*.wat`):
...
+Wasmtime.
+
+For core WebAssembly modules this means that the function exported as an empty
+string, or the `_start` export, is invoked. For WebAssembly components this
+means that the `wasi:cli/run` interface is executed.
+
+For both core modules and components CLI arguments are passed via WASI. Core
+modules receive arguments via WASIp1 APIs and components receive arguments via
+WASIp2 or later APIs. Arguments, flags, etc, are passed to the WebAssembly file
+after the file itself. For example:
```
abrown, line 61:
```suggestion
after the file itself. For example,
```

docs/cli-options.md
```diff
@@ -44,25 +44,92 @@ as well as the text format for WebAssembly (`*.wat`):
...
+
+```console
+wasmtime foo.wasm --bar baz
+```
+
+Will pass `["foo.wasm", "--bar", "baz"]` as the list of arguments to the module.
+Note that flags for Wasmtime must be passed before the WebAssembly file, not
+afterwards. For example
```
abrown, line 69:
```suggestion
afterwards. For example,
```

abrown (approved):
Apologies for the nits!

docs/cli-options.md
koke1997: Hey @abrown ,

I think this was mentioned in this comment - https://github.com/WebAssembly/WASI/issues/13#issuecomment-1188574599

So i think this the right repository - https://github.com/WebAssembly/component-model

There is post on blog for this topic which helped me to understand this topic - https://wasmcloud.com/blog/webassembly-patterns-command-reactor-library/

alexcrichton: I'll update this to remove the "command" terminology since that's now-dated and instead just say "CLI program" in a few places here. I'd prefer to not have to bring up commands/reactors if possible since they're ideally only a historical feature.

Issue #10802, opened 2025-05-19
Panic during component instantiation


moldhouse:
### Test Case

Work in progress. Currently we only witness this in our production code base. We are currently working on a minimal example. Yet we found it may be valuable to share the back trace with you up front.

### Steps to Reproduce

See above.

### Expected Results

Instantiating the component without a panic.

### Actual Results

We encounter a panic.

```sh
thread 'tokio-runtime-worker' panicked at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rustix-1.0.5/src/backend/linux_raw/param/auxv.rs:302:68:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: core::option::unwrap_failed
   4: rustix::backend::param::auxv::init_auxv_impl
   5: rustix::backend::param::auxv::init_auxv
   6: <wasmtime::runtime::vm::instance::allocator::on_demand::OnDemandInstanceAllocator as wasmtime::runtime::vm::instance::allocator::InstanceAllocatorImpl>::allocate_fiber_stack
   7: wasmtime::runtime::component::instance::InstancePre<T>::instantiate_async::{{closure}}
   8: pharia_kernel::skills::v0_3::skill::SkillPre<_T>::instantiate_async::{{closure}}
   9: <pharia_kernel::skills::v0_3::skill::SkillPre<engine_room::LinkerImpl<alloc::boxed::Box<dyn pharia_kernel::csi::CsiForSkills+core::marker::Send>>> as pharia_kernel::skills::Skill>::run_as_function::{{closure}}
  10: pharia_kernel::skill_runtime::SkillRuntimeActor<C,S>::run::{{closure}}::{{closure}}
  11: <futures_util::stream::stream::select_next_some::SelectNextSome<St> as core::future::future::Future>::poll
  12: tokio::runtime::task::core::Core<T,S>::poll
  13: tokio::runtime::task::raw::poll
  14: tokio::runtime::scheduler::multi_thread::worker::Context::run_task
  15: tokio::runtime::task::raw::poll
```

### Versions and Environment

We see the panic in wasmtime 32, not in 31. We do not see the problem on all platforms.

We saw it on:

- MacOS (arm) running a Container with Ubuntu 24 (always)
- GitHub CI running the same Container (sometimes, flaky)

We did not see it on:

- MacOS (arm) without Container
[...]

PR #10803, opened 2025-05-19
Duplicate page size determination in `wasmtime-fiber`

alexcrichton:
Currently Wasmtime has a function `crate::runtime::vm::host_page_size`
but this isn't reachable from the `wasmtime-fiber` crate and instead tha
crate uses `rustix::param::page_size` to determine the host page size.
It looks like this usage of `rustix` is causing a panic in https://github.com/bytecodealliance/wasmtime/issues/10802.
Ideally `wasmtime-fiber` would be able to use the same function but the
crate separation does not currently make that feasible. For now
duplicate the logic of `wasmtime` into `wasmtime-fiber` as it's modest
enough to ensure that this does not panic.

Closes https://github.com/bytecodealliance/wasmtime/issues/10802

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4458,6 +4458,7 @@ dependencies = [
  "backtrace",
  "cc",
  "cfg-if",
+ "libc",
  "rustix 1.0.3",
  "wasmtime-asm-macros",
  "wasmtime-versioned-export-macros",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -88,7 +88,7 @@ http = { workspace = true, optional = true }
 http-body-util = { workspace = true, optional = true }
 
 [target.'cfg(unix)'.dependencies]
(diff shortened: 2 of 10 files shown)


pchickey (approved):

alexcrichton:
I don't believe the CI failure is easily addressable with the previous design so I opted to duplicate the definition of `host_page_size` from the `wasmtime` crate into the `wasmtime-fiber` crate instead of passing it in as a parameter. That I think is regrettable duplication but "at least it's now commented duplication" and also fixes the panic for now.

I've also re-rolled review since Pat is on vacation right now since this has now changed substantially enough that I don't want to carry over the previous approval.

alexcrichton:
I'm actually going to go ahead and flag this for merge. Apparently using this syscall prevents using `rr` with Wasmtime (rr crashes locally) so I'd like to get that fixed. If others have further comments on this I'm happy to handle them in follow-ups.

PR #10804, opened 2025-05-19
Use `Waker::noop` from Rust 1.85

alexcrichton:
Now possible after #10785


cfallin (approved):
Much cleaner without the manual implementations -- thanks!

PR #10806, opened 2025-05-19
Update Wasmtime to the 2024 Rust Edition

alexcrichton:
Now that our MSRV supports the 2024 edition it's possible to make this switch. This commit moves Wasmtime to the 2024 Edition to keep up-to-date with Rust idioms and access many of the edition features exclusive to the 2024 edition.

prtest:full


cfallin (approved):
From the commit messages it looks like this is just the metadata update then a `cargo fmt`? I didn't read through all the reformatting noise in detail but 👍 if so

alexcrichton:
Indeed!

PR #10808, opened 2025-05-19
Switch doxygen download locations again

alexcrichton:
In the never-ending quest to find the best way to download this switch to using a GitHub based URL to hopefully match GitHub's own uptime. This failed spuriously in the wasip3-prototyping repository so I figured we could try out updating here.


abrown (approved):
🤞

PR #10809, opened 2025-05-19
Add a `rustfmt.toml` to the repository

alexcrichton:
According to
https://doc.rust-lang.org/nightly/edition-guide/rust-2024/rustfmt-style-edition.html#migration and https://github.com/rust-lang/rustfmt?tab=readme-ov-file#tips it's now "strongly recommended" to have `rustfmt.toml` with and edition configuration. This looks to be because `cargo fmt` reads the `edition` field in `Cargo.toml`, but running `rustfmt` directly doesn't do that. Editors often run `rustfmt` directly meaning they format differently than `cargo fmt` without this configuration.


cfallin (approved):
👍

PR #10810, opened 2025-05-19
x64: Migrate "UnaryRmR" instructions to the new assembler

alexcrichton:
This includes `bsr`, `bsf`, `tzcnt`, `popcnt`, and `lzcnt`.

### DIFF
--- a/cranelift/codegen/src/isa/x64/inst/external.rs
+++ b/cranelift/codegen/src/isa/x64/inst/external.rs
@@ -59,13 +59,6 @@ impl From<Writable<Reg>> for asm::GprMem<PairedGpr, Gpr> {
 }
 
 // For ABI ergonomics.
-impl From<Gpr> for asm::GprMem<Gpr, Gpr> {
-    fn from(gpr: Gpr) -> Self {
-        Self::Gpr(gpr)
-    }
-}
-
-// For ABI ergonomics.
 impl From<Reg> for asm::GprMem<Gpr, Gpr> {
     fn from(gpr: Reg) -> Self {
         assert!(gpr.class() == RegClass::Int);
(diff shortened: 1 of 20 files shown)


cranelift/codegen/src/isa/x64/inst/external.rs
```diff
@@ -98,27 +91,6 @@ impl From<Writable<Reg>> for asm::Gpr<WritableGpr> {
...
-        Self::Gpr(pair)
-    }
-}
-
-// For Winch ergonomics.
-impl From<WritableGpr> for asm::Gpr<WritableGpr> {
-    fn from(gpr: WritableGpr) -> Self {
-        Self::new(gpr)
-    }
-}
```
abrown, line 120: Thanks for this. Putting the `From` over on the assembler side should be clearer!

winch/codegen/src/isa/x64/asm.rs
```diff
@@ -1455,35 +1457,41 @@ impl Assembler {
     /// Requires `has_lzcnt` flag.
     pub fn lzcnt(&mut self, src: Reg, dst: WritableReg, size: OperandSize) {
         assert!(self.isa_flags.has_lzcnt(), "Requires has_lzcnt flag");
-        self.emit(Inst::UnaryRmR {
-            size: size.into(),
-            op: args::UnaryRmROpcode::Lzcnt,
-            src: src.into(),
-            dst: dst.map(Into::into),
-        });
+        let dst = WritableGpr::from_reg(dst.to_reg().into());
```
abrown, line 1460: I was struggling with this today with multiplication: is there no better way?

abrown (approved):

winch/codegen/src/isa/x64/asm.rs
alexcrichton: Oh heh I was just copying from code you wrote, so I alas don't have great ideas here :(

"fixing" this would probably involve something more invasive in Winch to have typed registers throughout the entire implementation rather than just the low-level assembler would be my guess, but that's a pretty big change

PR #10811, opened 2025-05-19
Minor optimizations to assembler-generated code

alexcrichton:
This is a few minor changes in the hopes of optimizing the compile-time of the generated code itself for the x64 assembler crate, including:

* Add `derive(Copy)` to all instructions to benefit from a specialized implementation of `derive(Clone)` when a `Copy` implementation is also present (e.g. it's `*foo` instead of a structural `clone`-each-field).
* Don't use `write!` in `Display for Inst` and instead delegate directly with methods to avoid formatting machinery.
* Use helper methods in `RegisterVisitor` trait to have register allocation be a method-per-operand and shrink the methods a bit.


abrown (approved):
Makes sense to me. My sense is that, with all this generated code, any clarity improvements are worth their weight in gold due to easier debugging, regardless of the performance improvement.

Issue #9726, opened 2024-12-04
Use brk #0xf000 on arm64 for debug_trap


bjorn3:
#### Feature

See title

#### Benefit

This should make debuggers actually recognize it as breakpoint rather than as abort. 

cc https://github.com/rust-lang/rust/pull/133726#discussion_r1868295474

PR #10813, opened 2025-05-20
Make debugtrap generate `brk #0xf000` on aarch64 so debuggers recognize it

joshtriplett:
Debuggers treat `brk #0xf000` as a software breakpoint, so that they
know to skip over it when attempting to resume.

Fixes: https://github.com/bytecodealliance/wasmtime/issues/9726

### DIFF
--- a/cranelift/codegen/src/isa/aarch64/inst.isle
+++ b/cranelift/codegen/src/isa/aarch64/inst.isle
@@ -2966,7 +2966,7 @@
 (rule (csdb)
       (SideEffectNoResult.Inst (MInst.Csdb)))
 
-;; Helper for generating `brk` instructions.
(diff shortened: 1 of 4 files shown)


bjorn3:
https://github.com/bytecodealliance/wasmtime/blob/e82e178670945eb39ab337d2ec32acbdbb42eccf/cranelift/codegen/src/isa/aarch64/inst/emit.rs#L3122 needs to be updated too I think.

joshtriplett:
@bjorn3 Good catch, not sure how I missed that.

I just realized that I should have been running `cargo test --all` rather than `cargo test`.

Fixing.

alexcrichton (approved):
Thanks!

PR #10816, opened 2025-05-20
x64: Migrate `ShiftR` pseudo-instruction to the new assembler

alexcrichton:
Shifts and rotates were all represented with `ShiftR` and they're now all represented with individual instructions. Additionally the old `Imm8{Gpr,Reg}` abstractions were largely removed and/or replace to only be helpers in ISLE.

This does regress pcc a bit more than it already is at due to the shift instructions not being special-cased for pcc processing.


cfallin:
> This does regress pcc a bit more than it already is at due to the shift instructions not being special-cased for pcc processing.

No worries there -- it's already regressed since I last had time to try to push it forward so I'll need to make a comprehensive update in any case.

abrown (approved):
Nice; I'm loving all the removed code.

PR #10822, opened 2025-05-21
x64: convert all XMM minimum/maximum instructions

abrown:
This converts both the floating-point `min*` and `max*` instructions as well as all of the packed integer versions of those instructions.


cfallin (approved):
LGTM, thanks!

cfallin:
(It looks like there are some build failures due to some remaining uses of the removed functions?)

abrown:
> (It looks like there are some build failures due to some remaining uses of the removed functions?)

Yeah, apparently it's still used in tests so I moved it to `#[cfg(test)]`.

PR #10825, opened 2025-05-22
x64: Convert some lane-manipulation instructions to the new assembler

alexcrichton:
This was mostly straightforward but ended up requiring handling for a new style of instruction not previously before-seen: a write-only `GprMem` operand. This needed some extra treatment/helpers in the ISLE integration to ensure that this was all hooked up correctly from ISLE constructors down to the assembler instructions.


abrown (approved):
Let's do it! I struggled a bit over how this emits ISLE for write-only reg-mem operands; when and how we are emitting certain things can get a bit confusing (e.g., how we're intermixing this all with the trap codes). I concluded, though, that this is creating the constructors we would expect, even if it isn't particularly clear. And we can always refactor this later. Thanks for figuring this stuff out!

PR #10829, opened 2025-05-23
winch: Crank the ratchet on tests that can fail

alexcrichton:
This commit moves all "ignore these outright" tests for Winch and
AArch64 to "run the test and expect failure" after recent PRs which
removed most of the segfaults/etc which prevented running these tests.
One failure remained which was easy enough to fix here: the `cmov`
implementation only worked on integer registers, not floating-point
registers, leading to a panic. A helper was added for floating-point selects to
resolve this panic and get some tests passing.

After this fix the `ignore` method is no longer necessary as all tests
can be run and be flagged as either pass or fail.

### DIFF
--- a/crates/test-util/src/wast.rs
+++ b/crates/test-util/src/wast.rs
@@ -409,43 +409,6 @@ impl WastTest {
         spec_proposal_from_path(&self.path)
     }
 
-    /// Returns true for tests that should be ignored (not run) for
-    /// the given config.
-
-    /// The infrastructure for `.wast` tests is designed to enable the
-    /// execution of tests at all times, however, while compiler
-    /// backends are in partial development state, it becomes harder
-    /// to guarantee particular tests will run to completion and
-    /// return a recoverable error, in some cases tests might segfault
-    /// making it challenging to assert a failure status.
-    /// It's recommended to avoid ignoring tests as much as possible, instead
-    /// it's recommended to add tests to `WastTest::should_fail`.
-    pub fn ignore(&self, config: &WastConfig) -> bool {
-        if config.compiler == Compiler::Winch {
-            if cfg!(target_arch = "aarch64") {
-                let unsupported = [
-                    // Segfault
-                    "spec_testsuite/conversions.wast",
-                    "spec_testsuite/float_exprs.wast",
-                    "spec_testsuite/int_exprs.wast",
-                    "spec_testsuite/left-to-right.wast",
-                    "spec_testsuite/i32.wast",
-                    "spec_testsuite/traps.wast",
-                    "spec_testsuite/unreachable.wast",
-                    "spec_testsuite/unreached-valid.wast",
-                    "misc_testsuite/component-model/fused.wast",
-                    "misc_testsuite/component-model/strings.wast",
-                    "misc_testsuite/winch/select.wast",
-                    "misc_testsuite/sink-float-but-dont-trap.wast",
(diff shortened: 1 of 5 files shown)


winch/codegen/src/codegen/mod.rs
```diff
@@ -1005,9 +1005,9 @@ where
         base: Reg,
         table_data: &TableData,
     ) -> Result<M::Address> {
-        let scratch = scratch!(M);
         let bound = self.context.any_gpr(self.masm)?;
         let tmp = self.context.any_gpr(self.masm)?;
+        let tmp2 = self.context.any_gpr(self.masm)?;
```
saulecabrera, line 1010: One thing to note here is that this will potentially increase register pressure for x64, but in reality we're trying to fix an Aarch64 issue with immediate value handling. I don't think this is a blocker, but wanted to point this out since this will likely change in the near-future since I have a work-in-progress branch which fixes the scratch register handling more broadly and will hopefully allow us to keep the previous implementation for x64.

saulecabrera (approved):
Glad that we can avoid ignoring tests. Thanks for this! Left one small comment inline.

winch/codegen/src/codegen/mod.rs
alexcrichton: Oh nice! Would you prefer to hold off on landing this until said refactoring has landed?

saulecabrera: Sounds good to me, yeah.

bank8641 (approved):

winch/codegen/src/codegen/mod.rs
saulecabrera: I have part 2/N here https://github.com/bytecodealliance/wasmtime/pull/10989, once that one lands, I think the only remaining piece is to ensure that any scratch register usage in ISA-agnostic code goes through the proper path to allocate scratch registers and then we could land this one. Said differently, I think we're 1 change away from unblocking this.

alexcrichton: I've split out [a refactoring](https://github.com/bytecodealliance/wasmtime/pull/10997) from this PR but otherwise the final commit here updates the scratch usage in `emit_compute_table_elem_addr` which is then will unlock running more tests (yay!)

saulecabrera: I've opened the full migration here https://github.com/bytecodealliance/wasmtime/pull/10998, I opened it as draft for now since I still need to amend it to use the newer version of `with_scratch` and investigate why a disassembly test is failing to compile. I think it should be good after that! I had migrated `emit_compute_table_elem_addr` as well in that PR, before reading your message here, sorry about that! But the `csel` changes are still valuable though.

alexcrichton: Nice! And no please use your version, I was just curious to get a feel for things here, I'm not wed to these changes at all

alexcrichton:
Ok now a much smaller PR with all the previous fixes, thanks @saulecabrera! Now this just has a change for `fpu_csel` plus removing the `ignore` method, meaning all tests are run in aarch64 right now (although many are still expected to fail)

saulecabrera:
> all tests are run in aarch64 right now (although many are still expected to fail)

I'm currently looking into it.

alexcrichton:
Reproduced that failure locally and it required a new `self.with_aligned_sp` during codegen for the wasm `unreachable` instruction.

PR #10831, opened 2025-05-23
Remove mul support based on assembler

jlb6740:


jlb6740:
@abrown You implemented mul support. Not sure if there was a reason for not removing the older support so wanted to notify you of this pr.

cfallin:
(moving review over to @abrown as he has more context)

abrown (approved):
Thanks for catching this; LGTM!

PR #10832, opened 2025-05-23
asm: model a way to modify the lower three opcode bits

abrown:
The x86 ISA includes a special format that, for certain instructions, writes the low three bits of the destination register into the low three bits of the opcode. This is specified with `+rb`, `+rw`, `+rd`, and `+ro` in the reference manual. This change adds a way to model this when defining instructions, e.g., `rex(...).rb()`.

### DIFF
--- a/cranelift/assembler-x64/meta/src/dsl/encoding.rs
+++ b/cranelift/assembler-x64/meta/src/dsl/encoding.rs
@@ -27,6 +27,7 @@ pub fn rex(opcode: impl Into<Opcodes>) -> Rex {
         r: false,
         digit: None,
         imm: Imm::None,
+        opcode_mod: None,
     }
 }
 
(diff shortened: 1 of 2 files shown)


cranelift/assembler-x64/meta/src/dsl/encoding.rs
```diff
@@ -99,6 +100,19 @@ pub struct Rex {
     pub digit: Option<u8>,
     /// The number of bits used as an immediate operand to the instruction.
     pub imm: Imm,
+    /// Used for `+rb`, `+rw`, `+rd`, and `+ro` instructions, which encode `reg`
+    /// bits in the opcode byte; if `Some`, this contains the expected bit width
+    /// of `reg`.
```
cfallin, line 105: Would it make sense to make this field an enum with `None`, `Rb`, `Rw`, `Rd`, `Ro` as options and a `.bits()` method? Not too important either way but that might make this a little more self-documenting/nicer.

cfallin (approved):
LGTM, thanks!

cranelift/assembler-x64/meta/src/dsl/encoding.rs
abrown: 👍 I have another branch with that... abandoned because it was more lines. But, yeah, let me revive that since you mentioned it! I'll just add it to this PR and move the documentation around.

PR #10834, opened 2025-05-25
x64: Migrate xmm mov-family instructions to new assembler

alexcrichton:
This commit migrates xmm movement-style instructions, such as `mov{s{s,d},{a,u}p{s,d},dq{a,u}}` to the new assembler. This enables deleting the `XmmMovRM` class of instructions in ISLE. Along the way a number of notable changes were made:

* The emission of `XmmCmove` now properly uses `movss` for 32-bit floats instead of `movsd` (accident from #4317)
* ISLE constructors for assembler instructions now take `SyntheticAmode` instead of `Amode` since it's already supported anyway and it's a more flexible argument to take.
* Some `Into` impls were swapped to `From` impls since that's a more generic way of implementing the `Into` and `From` traits.
* The conversion from `SyntheticAmode` to the assembler `Amode<Gpr>` was fixed where one variant needed to use `rsp` instead of `rbp`.
* The `is_move` method was updated to ignore `movss` and `movsd` instructions since they're not true movement-related instructions in register-to-register situations.


alexcrichton:
@abrown IIRC you said you were working on mov instructions, which I'm realize now I assumed meant integer-related ones, but if this overlaps with what you've done I'm happy to close this and defer to your work

alexcrichton:
Ok just kidding, back to being ready. I ideally wanted to split out the `SyntheticAmode` and update from rbp to rsp, but there was nothing exercising it previously so I think it makes sense to include here instead.

abrown (approved):
Looks good to me! I'll add the scalar moves on top of this.

PR #10836, opened 2025-05-27
x64: Convert `compare` instructions to the new assembler

rahulchaphalkar:
Implement `compare` instructions.
Implement `Eflags` logic for DSL and wire up isle rules to set flags when appropriate.

### DIFF
--- a/cranelift/assembler-x64/meta/src/dsl/format.rs
+++ b/cranelift/assembler-x64/meta/src/dsl/format.rs
@@ -24,6 +24,7 @@ pub fn fmt(
     Format {
         name: name.into(),
         operands: operands.into_iter().map(Into::into).collect(),
+        eflags: EflagsMutability::default(),
     }
 }
 
@@ -134,6 +135,8 @@ pub struct Format {
     /// These operands should match the "Instruction" column in the reference
     /// manual.
     pub operands: Vec<Operand>,
+    /// This should match eflags description of an instruction.
+    pub eflags: EflagsMutability,
 }
 
 impl Format {
@@ -166,17 +169,38 @@ impl Format {
     pub fn operands_by_kind(&self) -> Vec<OperandKind> {
         self.locations().map(Location::kind).collect()
     }
+
+    /// Set the EFLAGS mutability for this instruction.
+    pub fn flags(mut self, eflags: EflagsMutability) -> Self {
+        self.eflags = eflags;
+        self
+    }
+
+    /// Return true if an instruction uses EFLAGS.
+    pub fn uses_eflags(&self) -> bool {
+        self.eflags != EflagsMutability::None
+    }
 }
 
 impl core::fmt::Display for Format {
     fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
-        let Format { name, operands } = self;
+        let Format {
+            name,
+            operands,
+            eflags,
+        } = self;
         let operands = operands
             .iter()
             .map(|operand| format!("{operand}"))
             .collect::<Vec<_>>()
             .join(", ");
-        write!(f, "{name}({operands})")
+        write!(f, "{name}({operands})")?;
+
+        if *eflags != EflagsMutability::None {
+            write!(f, "[flags:{eflags}]")?;
+        }
+
+        Ok(())
     }
 }
 
@@ -256,6 +280,7 @@ impl From<Location> for Operand {
 }
 
 /// The kind of register used in a [`Location`].
+#[derive(PartialEq)]
 pub enum RegClass {
     Gpr,
     Xmm,
@@ -534,6 +559,31 @@ impl core::fmt::Display for Extension {
             Extension::SignExtendQuad => write!(f, "sxq"),
             Extension::SignExtendLong => write!(f, "sxl"),
             Extension::SignExtendWord => write!(f, "sxw"),
+        }
+    }
+}
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum EflagsMutability {
+    None,
+    R,
+    W,
+    RW,
+}
+
+impl Default for EflagsMutability {
+    fn default() -> Self {
+        Self::None
+    }
+}
+
+impl core::fmt::Display for EflagsMutability {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::None => write!(f, ""),
+            Self::R => write!(f, "r"),
+            Self::W => write!(f, "w"),
+            Self::RW => write!(f, "rw"),
         }
     }
 }
--- a/cranelift/assembler-x64/meta/src/generate/inst.rs
+++ b/cranelift/assembler-x64/meta/src/generate/inst.rs
@@ -217,6 +217,54 @@ impl dsl::Inst {
                             let to_string = location.generate_to_string(op.extension);
                             fmtln!(f, "let {location} = {to_string};");
                         }
+                        // Fix the mnemonic for comparison instructions: we want to print pseudo-ops without
+                        // the immediate operand.
+                        if self.mnemonic.starts_with("cmp")
+                            && ["cmppd", "cmpps", "cmpsd", "cmpss"]
+                                .contains(&self.mnemonic.as_str())
+                            && self.format.operands.len() == 3
+                            && matches!(
+                                self.format.operands[2].location.kind(),
+                                dsl::OperandKind::Imm(_)
+                            )
+                        {
+                            let imm_operand = self.format.operands[2].location;
+                            f.add_block(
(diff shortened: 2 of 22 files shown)


cranelift/codegen/src/isa/x64/inst/emit.rs
```diff
@@ -2625,8 +2623,7 @@ pub(crate) fn emit(
                 _ => unreachable!(),
             };
 
-            let inst = Inst::xmm_cmp_rm_r(cmp_op, dst.to_reg(), RegMem::reg(lhs));
-            inst.emit(sink, info, state);
+            Inst::x64_ucomis(*size, dst.to_reg(), RegMem::reg(lhs)).emit(sink, info, state);
```
abrown, line 2626: Like we were talking about in the Cranelift meeting, I think we want to reduce these methods hanging off of `Inst`. We can probably just fit into the existing `match size` above, but if that doesn't work, I also started using this pattern:

https://github.com/bytecodealliance/wasmtime/blob/89419ec2671de863cc2a0743d29a094c91355a45/cranelift/codegen/src/isa/x64/inst/emit.rs#L3099-L3105

cranelift/assembler-x64/meta/src/instructions/cmp.rs
```diff
@@ -0,0 +1,34 @@
+use crate::dsl::{EflagsMutability::*, Feature::*, Inst, Location::*};
+use crate::dsl::{fmt, inst, r, rex, rw, sxl, sxq, sxw};
+
+#[rustfmt::skip] // Keeps instructions on a single line.
+pub fn list() -> Vec<Inst> {
+    vec![
+        inst("cmpb", fmt("I", [r(al), r(imm8)]), rex(0x3C).ib(), _64b | compat),
```
abrown, line 7: We don't end up lowering to these instructions yet, right? If so, a simplification here could be to keep these instructions out of this PR and add them in once we can do the ISLE conversions at the same time. (I mean, we could also leave them in this PR since we know we'll eventually use them, I'm just flagging that they're not yet used).

cranelift/codegen/src/isa/x64/inst/emit.rs
```diff
@@ -2599,9 +2600,13 @@ pub(crate) fn emit(
...
             let do_min_max = sink.get_label();
 
-            let (add_op, and_op, or_op, min_max_op) = match size {
+            let (add_op, cmp_op, and_op, or_op, min_max_op) = match size {
                 OperandSize::Size32 => (
                     asm::inst::addss_a::new(dst, lhs).into(),
+                    asm::Inst::<CraneliftRegisters>::from(asm::inst::ucomiss_a::new(
+                        dst.to_reg(),
+                        lhs,
+                    )),
```
abrown: How about this to match the others?

```suggestion
                    asm::inst::ucomiss_a::new(dst, lhs).into(),
```

cranelift/assembler-x64/meta/src/generate/inst.rs
```diff
@@ -217,6 +217,45 @@ impl dsl::Inst {
...
+                            fmtln!(f, ";");
+                            let op0 = self.format.operands[0].location.to_string();
+                            let op1 = self.format.operands[1].location.to_string();
+                            let imm = imm_operand.to_string();
+                            fmtln!(f, "let operands = if mnemonic != \"{}\" {{", self.mnemonic);
+                            fmtln!(f, "  format!(\"{{{op1}}}, {{{op0}}}\")",);
+                            fmtln!(f, "}} else {{");
+                            fmtln!(f, "  format!(\"{{{imm}}}, {{{op1}}}, {{{op0}}}\")");
+                            fmtln!(f, "}};");
+                            fmtln!(f, "write!(f, \"{{mnemonic}} {{operands}}\")");
```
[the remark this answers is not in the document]
abrown: Ok, let's leave this as-is for now but we'll need to refactor to something more like the `custom` logic introduced by @alexcrichton in https://github.com/bytecodealliance/wasmtime/pull/10887.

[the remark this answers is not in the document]
rahulchaphalkar: Hey, I agree with the idea that lets push the external printing patch first, and then rebase on that. So we can leave this PR open for a bit until I push the other PR, I'll take a look at at https://github.com/bytecodealliance/wasmtime/pull/10887

cranelift/assembler-x64/meta/src/dsl/format.rs
```diff
@@ -256,6 +280,7 @@ impl From<Location> for Operand {
 }
 
 /// The kind of register used in a [`Location`].
+#[derive(PartialEq)]
```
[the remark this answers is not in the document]
rahulchaphalkar: Removed

cranelift/assembler-x64/meta/src/instructions/cmp.rs
```diff
@@ -0,0 +1,34 @@
...
+        inst("cmpl", fmt("MI_SXB", [r(rm32), sxl(imm8)]), rex(0x83).digit(7).ib(), _64b | compat),
+        inst("cmpq", fmt("MI_SXB", [r(rm64), sxq(imm8)]), rex(0x83).w().digit(7).ib(), _64b),
+        inst("cmpb", fmt("MR", [r(rm8), r(r8)]), rex(0x38).r(), _64b | compat),
+        inst("cmpw", fmt("MR", [r(rm16), r(r16)]), rex([0x66, 0x39]).r(), _64b | compat),
+        inst("cmpl", fmt("MR", [r(rm32), r(r32)]), rex(0x39).r(), _64b | compat),
+        inst("cmpq", fmt("MR", [r(rm64), r(r64)]), rex(0x39).w().r(), _64b),
+        inst("cmpb", fmt("RM", [r(r8), r(rm8)]), rex(0x3A).r(), _64b | compat),
+        inst("cmpw", fmt("RM", [r(r16), r(rm16)]), rex([0x66, 0x3B]).r(), _64b | compat),
+        inst("cmpl", fmt("RM", [r(r32), r(rm32)]), rex(0x3B).r(), _64b | compat),
+        inst("cmpq", fmt("RM", [r(r64), r(rm64)]), rex(0x3B).w().r(), _64b),
```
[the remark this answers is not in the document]
rahulchaphalkar: Yes, removed these instructions for now since they weren't lowered

cranelift/codegen/src/isa/x64/inst/emit.rs
rahulchaphalkar: It needs type annotation, I don't know why only for these couple of instructions rust is unable to infer type, so was having issues with using Into even with types. I can retry with `into` with type annotations, but it still won't match exactly.

rahulchaphalkar:
@abrown 
I have rebased this on the `Custom` logic PR as well as latest main.
For compare instructions, we need to change operands as well (remove the immediate) as well as change name, so now we are passing `ordered_ops` string to `custom::Display` functions, which returns a concatenated string containing `name` and `ordered_ops`.

cranelift/assembler-x64/src/custom.rs
```diff
@@ -87,18 +87,97 @@ pub mod display {
...
+            _ => return format!("{} {ordered_ops}", inst.mnemonic().to_string()),
+        };
+        let ordered_ops = ordered_ops
+            .split_once(", ")
+            .map(|(_, rest)| rest.to_string())
+            .unwrap_or(ordered_ops);
+        format!("{name} {ordered_ops}")
+    }
+
+    pub fn cmppd_a<R: Registers>(inst: &inst::cmppd_a<R>, ordered_ops: String) -> String {
```
abrown: It seems like in the end we probably just want this to be:

```suggestion
    pub fn cmppd_a<R: Registers>(f: &mut std::fmt::Formatter, inst: &inst::cmppd_a<R>) -> std::fmt::Result {
```

The problem with that is that then we have to list out the operands of each instruction in order, so I guess we can go with this for now.

cranelift/codegen/src/isa/x64/inst/emit.rs
abrown: Do you need help with this part?

rahulchaphalkar: Done

cranelift/assembler-x64/src/custom.rs
rahulchaphalkar: Got it, will push those changes to a later PR.

abrown (approved):
LGTM! I think the one last thing we'll want to clean up the `ordered_ops` bits once https://github.com/bytecodealliance/wasmtime/pull/11017 is merged.

rahulchaphalkar:
@abrown ok i have rebased and pushed with changes, CI still running.

PR #10838, opened 2025-05-27
Enable the `useless_conversion` Clippy lint

alexcrichton:
We've got lots of types in Wasmtime and convert between them quite a lot, but often over time conversions become unnecessary through refactorings or similar. This will hopefully enable us to clean up some conversions as they come up to try to have as few as possible ideally.

### DIFF
--- a/crates/cache/src/worker.rs
+++ b/crates/cache/src/worker.rs
@@ -9,7 +9,10 @@
     not(test),
     expect(
         clippy::useless_conversion,
-        reason = "test/not have a different definition of `SystemTime`",
+        reason = "cfg(test) and cfg(not(test)) have a different definition \
+                  of `SystemTime`, so conversions below are needed in \
+                  one mode but not the other, just ignore the lint in this \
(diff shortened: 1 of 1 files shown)


crates/cache/src/worker.rs
```diff
@@ -5,6 +5,14 @@
 //! but we guarantee eventual consistency and fault tolerancy.
 //! Background tasks can be CPU intensive, but the worker thread has low priority.
 
+#![cfg_attr(
+    not(test),
+    expect(
+        clippy::useless_conversion,
+        reason = "test/not have a different definition of `SystemTime`",
```
abrown, line 12: ?

abrown (approved):
Makes sense!

crates/cache/src/worker.rs
abrown: What I meant (and should have written out) is that the "test/not" part of the `reason` message was hard to parse.

PR #10841, opened 2025-05-27
c-api: component-model: Function post return, more values

MangoPeachGrape:


alexcrichton (approved):
Thanks for this!

Random thought: do you know how we might run asan or something to check for memory leaks in CI? I'm not sure how to do that myself but it'd be nice to ensure that we're not accidentally leaking any memory with this representation anywhere

alexcrichton:
They should both bottom out in `malloc` and `free` by default, so I think ASAN might be enough (which IIRC also detects leaks in addition to other things).  If LSAN works though that also seems reasonable to me

pchickey:
I recently hooked up wasmtime and a C api project (though not wasmtime's C api) to asan to find my own bugs of that sort. I found that I needed to:

* compile the rust with nightly and `RUSTFLAGS=-Zsanitizer=address`
* install and build the C with an llvm/clang toolchain that is relatively recent so the sanitizer is compatible with the information emitted by rustc, since my CI package manager's was too old I used LLVM-20.1 which is the same version rust is using in nightly. I don't have a sense for when rust will drift far enough from that to break something, but it works right now anyway.
* build the C sources with `-fsanitize=address,undefined -fno-omit-frame-pointer` cflags
* set `ASAN_SYMBOLIZER_PATH=<install location of LLVM-20.1.0-Linux-X64>/bin/llvm-symbolizer` to get symbols in asan's traces

Also, beware that running debug builds of cranelift under asan will be egregiously, obscenely slow.

PR #10844, opened 2025-05-27
Merge wasip3-prototyping bindgen changes to main 

alexcrichton:
This commit wholesale copies the `wasmtime-wit-bindgen` changes from the
wasip3-prototyping repository to the Wasmtime repository here. This is
done to make the future merge with `wasip3-prototyping` smaller
complexity-wise.

This is intended to be a no-functional-change for non-"concurrent"
imports, or those for component model async. That means that it's
intended that no current embedding is affected by these changes.
Internally though there has been restructuring.

Namely internally the generation of an interface trait, a resource
trait, and a world trait are all unified in a single method now. In the
future with component model async these traits are being split into
async/sync versions and so it was much nicer to do this in one location
rather than across three different locations.

This required some minor renamings in the generated code and moving
around some impls, but otherwise the generated code should mostly be the
same as before (see golden output diff in the subsequent commit of this
PR)

### DIFF
--- a/crates/component-macro/tests/expanded/char.rs
+++ b/crates/component-macro/tests/expanded/char.rs
@@ -146,14 +146,14 @@ const _: () = {
         }
         pub fn add_to_linker<T, D>(
             linker: &mut wasmtime::component::Linker<T>,
-            get: fn(&mut T) -> D::Data<'_>,
+            host_getter: fn(&mut T) -> D::Data<'_>,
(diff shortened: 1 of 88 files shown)


alexcrichton:
Oops yeah that's an accident, and yeah I'll work on uncommenting this in wasip3-prototyping and then see what it would take to merge back those changes here

PR #10845, opened 2025-05-27
asm: refactor VEX implementation

abrown:
This change adapts the current VEX logic implemented in #10754 to look more like the current infrastructure. To do so:
- it factors out a `VexPrefix` from the current `VexInstruction`, allowing us to reuse existing logic for ModR/M and SIB bytes (these do are the same in both REX and VEX encodings)
- it uses compile-time constant bits like the `L` or `pp` fields directly, hopefully saving a bit on compile time
- all of this allows more consistent code generation

On the DSL side, this significantly changes how one defines VEX instructions. Many helper methods are added and now documented. It also makes use of a slightly unrelated change included here: I factored out how any encoding can affect the ModR/M byte with `ModRmKind`. This `enum` forces the DSL to pick only one of `/r` or `/digit` when defining instructions.


cranelift/assembler-x64/meta/src/dsl/encoding.rs
```diff
@@ -705,98 +751,309 @@ impl fmt::Display for OpcodeMod {
...
+    /// Map the `MMMMM` field encodings.
+    pub mmmmm: Option<VexEscape>,
+    /// The `W` bit.
+    ///
+    /// Since VEX-encoded instructions may ignore the `W` bit, we model this
+    /// with a three-state field:
+    /// - `None`: the `W` bit is ignored (i.e., `WIG` in the manual)
+    /// - `Some(false)`: the `W` bit is set to `0` (i.e., `W0` in the manual)
+    /// - `Some(true)`: the `W` bit is set to `1` (i.e., `W1` in the manual)
+    pub w: Option<bool>,
```
cfallin, line 881: Could we define a `WBit` (or pick your favorite name) enum here with `WIG`/`W0`/`W1`?

cfallin (approved):
Looks reasonable!

PR #10856, opened 2025-05-29
Add round variants to the new assembler

jlb6740:

### DIFF
--- a/cranelift/assembler-x64/meta/src/instructions/round.rs
+++ b/cranelift/assembler-x64/meta/src/instructions/round.rs
@@ -1,5 +1,5 @@
-use crate::dsl::{Feature::*, Inst, Location::*};
-use crate::dsl::{fmt, inst, r, rex, w};
+use crate::dsl::{Feature::*, Inst, Location::*, VexLength::*};
+use crate::dsl::{fmt, inst, r, rex, vex, w};
 
 #[rustfmt::skip] // Keeps instructions on a single line.
 pub fn list() -> Vec<Inst> {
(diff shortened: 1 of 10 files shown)


jlb6740:
Will add vround variants in a follow-up to this patch.

abrown (approved):

cranelift/assembler-x64/meta/src/instructions/round.rs
```diff
@@ -0,0 +1,19 @@
+use crate::dsl::{Feature::*, Inst, Location::*, VexLength::*};
+use crate::dsl::{fmt, inst, r, rex, vex, w};
+
+#[rustfmt::skip] // Keeps instructions on a single line.
+pub fn list() -> Vec<Inst> {
+    vec![
+        inst("roundpd", fmt("RMI", [w(xmm1), r(xmm_m128), r(imm8)]), rex([0x66, 0x0F, 0x3A, 0x09]).ib(), _64b | compat | sse41),
+        inst("roundps", fmt("RMI", [w(xmm1), r(xmm_m128), r(imm8)]), rex([0x66, 0x0F, 0x3A, 0x08]).ib(), _64b | compat | sse41),
```
alexcrichton, line 8: Should these be `align(xmm_m128)`?

(should we maybe have a check in the DSL that non-avx `xmm_m128` arguments are required to be `align`?)

jlb6740: I read this comment multiple times trying to interpret: 

"NB: the non-AVX variant of this instruction does not require that the operand
;; is aligned, but the instruction variant used here requires it to be aligned."

Perhaps I should have asked. So the non-AVX variant here does require alignment?

alexcrichton: Ah yeah I believe that comment is referring to `rounds{s,d}`, as opposed to `roundp{s,d}`. I'm under the impression that the f32/f64 versions do not require aligned memory but the 128-bit versions require aligned memory. The comment in ISLE I believe can be deleted with the new instructions since it's correctly modeled.

This reminds me as well, should `rounds{s,d}` below use `xmm_m{32,64}` instead of `xmm_m128`? We don't have anything actually checking that right now but with PCC in the future (or some other integration) it might matter to annotate that with the appropriate bit-width.

alexcrichton: Oh sorry fully reading your comment I can also clarify more:

The documentation there referred to the ISLE-specific abstraction of `XmmUnaryRmRImm` which internally stores `XmmMemAligned`. The `roundss` and `roundsd` instructions used `XmmUnaryRmRImm` but this was technically not a correct way to model these instructions. The `roundss` and `roundsd` instructions do not actually require alignment. The problem, and where the comment comes from, is that the way this happens in ISLE is that first a `Value` is converted to `XmmMem` as the load is "sinkable". Next it goes to the `x64_round` helper which takes `XmmMem`. Next, without AVX, it would go to `XmmUnaryRmRImm`, but this required converting the `XmmMem` to `XmmMemAligned`. That, for all WebAssembly loads/stores, would automatically do an unaligned load and then use a register operand. The problem with this is that the automatic load inserted would load 128-bits, not the appropriate number of bits for the f32/f64. 

So basically that long comment was an artifact of a precisely how ISLE is set up today. None of it is applicable under the new assembler, and I think that the `roundps` and `roundpd` instructions need `align(xmm_m128)` but `roundss` and `roundsd` can use `xmm_m32` and `xmm_m64` since they don't require alignment.

Also FWIW I see a new commit here but I don't believe it landed because it was pushed after this was already in the merge queue, so mind filing a follow-up commit with that?

PR #10863, opened 2025-05-29
Cranelift: properly reject unimplemented big-endian loads/stores.

### LINKED ISSUE #10861: Cranelift: Respect or error on big-endian flag for loads/stores on little-endian targets
#### Feature

See title

#### Benefit

Currently the endianness flags are only respected by the s390x backend. Ignoring it on other targets only causes confusion like in https://github.com/bytecodealliance/wasmtime/issues/10852 and is technically a miscompilation.

#### Implementation

Depends on which of the two options is picked.

#### Alternatives

Keep the status quo and keep confusing people/miscompiling code.

cfallin:
At some point during the development of the Cranelift backend infrastructure, to properly support big-endian architectures such as s390x, we added explicit endianness flags to `MemFlags`, which are given to all memory operations (e.g., loads, stores, and atomic ops). In s390x in particular, the backend carefully observes these flags, because a prominent use of Cranelift (as a Wasm backend) requires explicit little-endian behavior and the system is big-endian. However, all of our other supported ISAs are little-endian and so we did not implement explicit checks at the time, instead accepting all loads and stores as an artifact of our little-endian-only origins.

This PR adds explicit conditions to all ISLE rules that lower loads, stores, and atomic ops on x86-64, aarch64, and riscv64 to accept little or "native" (also little) endian operations only. Compilation of a big-endian operation will now result in a compilation error because no ISLE rule will match (no lowering exists). At some later point we could add these lowerings, but for now we at least do not miscompile.

Fixes #10861.

### DIFF
--- a/cranelift/codegen/src/isa/aarch64/lower.isle
+++ b/cranelift/codegen/src/isa/aarch64/lower.isle
@@ -1258,7 +1258,7 @@
 ;; Atomic loads will also automatically zero their upper bits so the `uextend`
 ;; instruction can effectively get skipped here.
 (rule 1 (lower (has_type (fits_in_64 out)
-                       (uextend x @ (and (value_type in) (atomic_load flags _)))))
+                       (uextend x @ (and (value_type in) (atomic_load (little_or_native_endian flags) _)))))
       (if-let mem_op (is_sinkable_inst x))
       (load_acquire in flags (sink_atomic_load mem_op)))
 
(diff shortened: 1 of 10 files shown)


cfallin:
(Note that the Pulley backend rules appear to have explicit consideration for endianness, so I skipped that backend here.)

cfallin:
Design note: I could have added checks to more central places in all three backends instead -- namely, `Amode` construction -- but opted instead to annotate every rule with this predicate, because the information logically belongs to the individual lowerings and if we want to add actual big-endian lowerings in the future, we'll want `Amode` to work for those too.

fitzgen (approved):

Issue #10740, opened 2025-05-07
Regression for custom LinearMemory implementations: entered unreachable code: memory_image is Some only for mmap-based memories


V0ldek:
Hello.

I've been using `wasmtime` with a custom memory implementation that is backed by my custom managed virtual mmap. I don't think the implementation details of my custom struct matter. I've been on wasmtime 25.0 and everything worked fine. After an update to wasmtime 32.0 I had to rewrite the `wasmtime::LinearMemory` impl, but it just got simplified. However, running my code now gets me this panic:

<details>
<summary>Stacktrace</summary>
```
thread 'main' panicked at /home/mat/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wasmtime-32.0.0/src/runtime/vm/memory.rs:536:29:
internal error: entered unreachable code: memory_image is Some only for mmap-based memories
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/a15cce2690e8fab72422515c9dc02c6fbc506733/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/a15cce2690e8fab72422515c9dc02c6fbc506733/library/core/src/panicking.rs:75:14
   2: wasmtime::runtime::vm::memory::LocalMemory::new
             at /home/mat/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wasmtime-32.0.0/src/runtime/vm/memory.rs:536:29
   3: wasmtime::runtime::vm::memory::Memory::new_dynamic
             at /home/mat/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wasmtime-32.0.0/src/runtime/vm/memory.rs:240:22
   4: <wasmtime::runtime::vm::instance::allocator::on_demand::OnDemandInstanceAllocator as wasmtime::runtime::vm::instance::allocator::InstanceAllocatorImpl>::allocate_memory
             at /home/mat/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wasmtime-32.0.0/src/runtime/vm/instance/allocator/on_demand.rs:119:22
   5: wasmtime::runtime::vm::instance::allocator::InstanceAllocator::allocate_memories
             at /home/mat/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wasmtime-32.0.0/src/runtime/vm/instance/allocator.rs:463:27
[...]

PR #10865, opened 2025-05-29
Extend CustomCodeMemory interface allowing for using previously-mapped images

Milek7:
`Module::deserialize_raw` and `CustomCodeMemory` were introduced with aim of supporting `no_std` platforms. With small tweaks they can also be used for using Wasmtime on platforms that have full-blown virtual memory capabilities, but doesn't allow for directly mapping executable pages from user code instead limiting that capability for system loader. Adding features necessary for such platforms was previously attempted by #8245.

There are currently two issues with using `Module::deserialize_raw` for using images either statically linked into embedder executable or dynamically loaded shared objects:

- ~~`CodeMemory::publish` will initially make entire image read-only, destroying executable permissions that cannot be restored by user code. This will happen even if `CustomCodeMemory` is provided.~~
- `CodeMemory::publish` will attempt to unconditionally register unwind information. As these are already properly registered by the system loader this is superfluous at best, or could fail module loading if system decides to return error on attempt for double-registration.

This commit solves these issues by:
- Moving responsibility for making image RO to `CustomCodeMemory` hook.
- Making `CustomCodeMemory` publishing hook return enum that tells what steps (only mapping, or mapping with registration) were performed, using default implementation only for actions not reported by the hook.
- Additionally `CustomCodeMemory` hooks also receive pointers to the entire image, not only executable section. This allows the embedder to keep track of the images and unload them in case they were loaded as dynamic shared object.

### DIFF
--- a/crates/wasmtime/src/engine.rs
+++ b/crates/wasmtime/src/engine.rs
@@ -3,6 +3,8 @@ use crate::prelude::*;
 #[cfg(feature = "runtime")]
 pub use crate::runtime::code_memory::CustomCodeMemory;
 #[cfg(feature = "runtime")]
+pub use crate::runtime::code_memory::HandledByCustomCodeMemory;
+#[cfg(feature = "runtime")]
 use crate::runtime::type_registry::TypeRegistry;
 #[cfg(feature = "runtime")]
 use crate::runtime::vm::GcRuntime;
--- a/crates/wasmtime/src/runtime/code_memory.rs
+++ b/crates/wasmtime/src/runtime/code_memory.rs
@@ -44,10 +44,17 @@ impl Drop for CodeMemory {
         // If there is a custom code memory handler, restore the
         // original (non-executable) state of the memory.
         if let Some(mem) = self.custom_code_memory.as_ref() {
-            if self.published && self.needs_executable {
-                let text = self.text();
-                mem.unpublish_executable(text.as_ptr(), text.len())
-                    .expect("Executable memory unpublish failed");
+            if self.published {
+                mem.unpublish_image(
+                    self.mmap.as_ptr(),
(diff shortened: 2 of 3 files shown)


Milek7:
Okay, I got a little ahead of myself.

`make_readonly` won't be actually get called for modules created by `Module::deserialize_raw`. So first issue is not true.
However double-registration is still a concern.
Providing address of entire image to hooks is still useful for keeping track for unloading purposes (and maybe for no virtual memory target if they actually want to make that RO mapping).

So I still think this make sense.

alexcrichton:
This seems reasonable to me to support, but what do you think of modeling this a bit differently? This is a pretty sensitive piece of the codebase which we want to ensure doesn't accidentally regress and having a lot of intertwined "if this else that" style logic can be relatively confusing to follow. Do you think it would be possible to internally model this as "assume there's always a custom memory publisher"? That way adding more hooks would mean adding more trait methods, and the "driver" would still be relatively straightforward of "do this, then that, then that".

A custom trait could then override any one particular step to a noop and that wouldn't involve a need for custom enums/etc which are pretty tightly bound to the exact structure of what we have today as well.

Milek7:
Do you think about trait methods that would return bool indicating whether default implementation should be called, or calling only trait methods with default implementation provided as default CustomCodeMemory? In second case should MmapVec become part of public API?

alexcrichton:
Ideally we'd model everything with the trait to allow embedders to customize all the hooks necessary, but that's also not easy as there's internal types, like `MmapVec`, which we don't want to expose. This came up in https://github.com/bytecodealliance/wasmtime/issues/10740 where we have an internal trait with private types and a public trait with less functionality and that became a problem as well. Unfortunately I'm not sure how best to thread this needle...

Issue #10869, opened 2025-05-30
Memory "leak" when running a GC


alexcrichton:
When running this program:

```rust
use wasmtime::*;

fn main() -> Result<()> {
    let mut config = Config::new();
    config.wasm_gc(true);
    let engine = Engine::new(&config)?;
    let mut store = Store::new(&engine, ());

    let module = Module::new(
        &engine,
        r#"
            (module
                (table 1 anyref)

                (func (result anyref)
                    i32.const 0
                    table.get)

            )
        "#,
    )?;

    Instance::new(&mut store, &module, &[]).unwrap();
    loop {
        store.gc(None);
    }
}
```

this will infinitely consume memory and eventually die. The problem here is that GC is written at the abstraction level of wasmtime-crate-exports but created those from internals requires pushing onto an internal vector. These vectors are never shrunk meaning that the memory is held until the store is dropped, meaning that a store's memory usage is O(gcs) which isn't great for long-running programs.

alexcrichton:
Fixed in https://github.com/bytecodealliance/wasmtime/pull/10903

Issue #10868, opened 2025-05-30
Memory "leak" when passing funcrefs to the host


alexcrichton:
This program:

```rust
use wasmtime::*;

fn main() -> Result<()> {
    let engine = Engine::default();
    let mut store = Store::new(&engine, ());

    let module = Module::new(
        &engine,
        r#"
            (module
                (import "" "" (func (param funcref)))

                (elem declare func $x)

                (func $x)
                (func (export "")
                    ref.func $x
                    call 0)
            )
        "#,
    )?;

    let func = Func::wrap(&mut store, |_: Option<Func>| {});
    let i = Instance::new(&mut store, &module, &[func.into()]).unwrap();
    let f = i.get_typed_func::<(), ()>(&mut store, "")?;
    loop {
        f.call(&mut store, ())?;
    }
}
```

will infinitely consume memory until killed. Basically what happens here is that each time a host function with a `funcref` parameter is invoked Wasmtime creates a new `Func` which pushes onto an internal store-local vector that is never shrunk. That means that the memory usage of a store is O(calls) which is generally not a desirable property we want to have.

This is not a regression but is something that's been around for quite a long time. I'm working on fixes in this area though and wanted to file an issue about this.

PR #10870, opened 2025-05-30
Start moving away from using `Stored<T>` for wasmtime exports

alexcrichton:
This PR is a series of commits which is intended to lead up to the conclusion of removing usage of `Stored<T>` from wasmtime exports. This is intended to be an internal refactoring of the implementation guts of Wasmtime that doesn't have an API-level impact on users of Wasmtime. The motivations for this change are:

* Fix issues such as https://github.com/bytecodealliance/wasmtime/issues/10869 and https://github.com/bytecodealliance/wasmtime/issues/10868. Effectively it's always been a cost that converting from an internal representation to the external representation pushes onto some internal vectors. Making this cost go away will fix the above issues.
* Enable using public exports in the internal implementation of wasmtime. Historically the internals of Wasmtime are implemented very differently than what the users of Wasmtime get to use. That creates this weird dichotomy where externally everyone gets to learn one API while internally it's completely different. Internally Wasmtime is already a bit of a mish mash of some bits using raw pointers and others using indices. The goal of this refactoring is to enable pushing more towards the indices style of internal implementation which is safer, easier to reason about, and matches the public API.

While the first bullet here is pretty concrete the second is pretty nebulous. The original motivation was myself working in the wasip3-prototyping branch and brainstorming ways to reduce the amount of unsafety. That led me to wanting to stop using `*mut ComponentInstance` for example and instead use `wasmtime::component::Instance` but it was pretty awkward to do so. While none of that has to do with the core wasm bits being changed here I wanted to game out what such changes might look like. Additionally I also wanted to ensure that the chasm between the style of implementation of core wasm and components does not diverge further than it already has.

This PR is a draft PR since it's pretty large and I'll be wanting to split it up over time. I'm going to peel off commits from this PR into separate PRs and land those separately, but I wanted to have this open as a sort of "north star" in case folks are interested to see where this is going.

Sequence-wise this is organized as going through each of the exported items from core wasm modules to remove their usage of `Stored<T>` one at a time. This will eventually culminate in removing `Stored<T>` for `wasmtime::Instance` which will fully migrate core wasm bits away from `Stored<T>`. Status-wise here's the progress

* [x] `wasmtime::Memory`
* [x] `wasmtime::Table`
* [x] `wasmtime::Tag`
* [x] `wasmtime::Global`
* [x] `wasmtime::Func`
* [x] `wasmtime::Instance`
* [ ] `wasmtime::component::Instance`
* [ ] `wasmtime::component::Func`

### DIFF
--- a/crates/wasmtime/src/runtime/externals/global.rs
+++ b/crates/wasmtime/src/runtime/externals/global.rs
@@ -348,7 +348,7 @@ impl Global {
     /// this hash key will be consistent across all of these globals.
     #[cfg(feature = "coredump")]
     pub(crate) fn hash_key(&self, store: &StoreOpaque) -> impl core::hash::Hash + Eq + use<> {
-        self.definition(store).as_ptr() as usize
+        self.definition(store).as_ptr().addr()
     }
 
     fn definition(&self, store: &StoreOpaque) -> NonNull<VMGlobalDefinition> {
--- a/crates/wasmtime/src/runtime/externals/table.rs
+++ b/crates/wasmtime/src/runtime/externals/table.rs
@@ -449,7 +449,7 @@ impl Table {
     /// this hash key will be consistent across all of these tables.
     #[allow(dead_code)] // Not used yet, but added for consistency.
     pub(crate) fn hash_key(&self, store: &StoreOpaque) -> impl core::hash::Hash + Eq + use<'_> {
-        store[self.instance].table_ptr(self.index).as_ptr() as usize
+        store[self.instance].table_ptr(self.index).as_ptr().addr()
     }
 }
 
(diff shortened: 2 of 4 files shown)


alexcrichton:
Sorry I hit submit too soon and I've now actually filled out the description/title for anyone interested.

alexcrichton:
I'll also clarify, I plan on using this to push lots of commit to, so if anyone would prefer please feel free to unsubscribe from this PR as it may be a bit noisy.

alexcrichton:
Final bits here split into https://github.com/bytecodealliance/wasmtime/pull/10908 and https://github.com/bytecodealliance/wasmtime/pull/10909, so closing now.

PR #10871, opened 2025-05-30
x64: convert `movq`

abrown:
This change represents a first part toward changing all scalar moves in x64 to use the new assembler. Recall that when moving sub-64-bit values, upper bits of a GPR remain unchanged, so Cranelift chooses to use zero-extending moves for those operations. This change does not touch that: for 64-bit moves, Cranelift _does_ use `movq` and this PR only touches that usage. Why? `movq` is used widely throughout Cranelift and many filetests will change, so it seems safer to change moves part by part.


cranelift/codegen/src/isa/x64/pcc.rs
```diff
@@ -165,11 +165,6 @@ pub(crate) fn check(
             }
         }
 
-        Inst::Mov64MR { ref src, dst } => {
-            check_load(ctx, Some(dst.to_writable_reg()), src, vcode, I64, 64)?;
-            Ok(())
-        }
-
```
abrown, line 172: @cfallin: I expect 3 PCC tests to fail due to this. I tried to look at how I might re-write this based on the assembler but at this point we only have assembler types to deal with. What to do?

cranelift/assembler-x64/src/fuzz.rs
```diff
@@ -195,12 +196,16 @@ fn replace_signed_immediates(dis: &str) -> std::borrow::Cow<str> {
...
             let (hex, rest) = rest.split_at(n); // Split at next non-hex character.
-            let simm = match hex.len() {
-                1 | 2 => hex_print_signed_imm!(hex, u8 => i8),
-                4 => hex_print_signed_imm!(hex, u16 => i16),
-                8 => hex_print_signed_imm!(hex, u32 => i32),
-                16 => hex_print_signed_imm!(hex, u64 => i64),
-                _ => panic!("unexpected length for hex: {hex}"),
+            let simm = if dis.starts_with("mov") {
+                u64::from_str_radix(hex, 16).unwrap().to_string()
+            } else {
```
abrown, line 201: Here's another Capstone-related weirdness: the assembler emits all immediates the same but this special logic "fixes" things up so we can match against Capstone. I will be very thankful when we can switch to a more regular disassembler.

cranelift/codegen/src/isa/x64/pcc.rs
cfallin: It's fine to remove the newly failing tests for now, I think -- PCC shouldn't get in the way of progress here (I'm happy to update it later when I next get a timeslice for it). Feel free to remove the test directives in the files and leave a note instead "disabled until PCC is migrated to understand new assembler instructions" or similar...

cranelift/filetests/filetests/isa/x64/call-conv.clif
```diff
@@ -193,8 +193,8 @@ block0(
 ;   movq    %rsp, %rbp
 ;   subq $0x90, %rsp
 ; block0:
-;   movq    rbp(stack args max - 32), %r10
-;   movq    rbp(stack args max - 24), %r11
+;   movq <offset:0>+-0x20(%rbp), %r10
+;   movq <offset:0>+-0x18(%rbp), %r11
```
cfallin, line 197: Pre-existing but I'm just noticing it now -- whether in this PR or separately, could we update the printing format for these synthetic amodes so that we don't have "x+-y" to mean "x-y"? (Slightly more complex in the formatting code but hopefully not too bad)

cfallin (approved):
This looks good overall -- thanks for the continued migration push!

cranelift/assembler-x64/meta/src/instructions/mov.rs
```diff
@@ -1,14 +1,45 @@
...
+        // overwrite all 64 bits are considered "write-only"; smaller-width
+        // moves indicate that upper bits are preserved by being "read-write."
+        inst("movb", fmt("MR", [rw(rm8), r(r8)]), rex(0x88).r(), _64b | compat),
+        inst("movw", fmt("MR", [rw(rm16), r(r16)]), rex([0x66, 0x89]).r(), _64b | compat),
+        inst("movl", fmt("MR", [rw(rm32), r(r32)]), rex(0x89).r(), _64b | compat),
+        inst("movq", fmt("MR", [w(rm64), r(r64)]), rex(0x89).w().r(), _64b),
+        inst("movb", fmt("RM", [rw(r8), r(rm8)]), rex(0x8A).r(), _64b | compat),
+        inst("movw", fmt("RM", [rw(r16), r(rm16)]), rex([0x66, 0x8B]).r(), _64b | compat),
+        inst("movl", fmt("RM", [rw(r32), r(rm32)]), rex(0x8B).r(), _64b | compat),
+        inst("movq", fmt("RM", [w(r64), r(rm64)]), rex(0x8B).w().r(), _64b),
```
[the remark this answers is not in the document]
cfallin: I know we do read-modify-write (load; add; store) in our lowering rules but I too am surprised we don't do store-of-immediate; seems worth adding?

cranelift/assembler-x64/meta/src/instructions/mov.rs
```diff
@@ -1,14 +1,45 @@
...
 
 #[rustfmt::skip] // Keeps instructions on a single line.
 pub fn list() -> Vec<Inst> {
     vec![
+        // Move integers to and from GPR and memory locations. Note that, in
+        // 64-bit mode, `r/m8` can not be encoded to access the following byte
+        // registers if a REX prefix is used: AH, BH, CH, DH. Only moves that
+        // overwrite all 64 bits are considered "write-only"; smaller-width
+        // moves indicate that upper bits are preserved by being "read-write."
+        inst("movb", fmt("MR", [rw(rm8), r(r8)]), rex(0x88).r(), _64b | compat),
```
cfallin, line 12: Actually, on second look -- does the `rw(rm8)` here mean that we have a read/write (modify) operand? This is somewhat surprising to me, per our Zulip thread -- to use this properly with narrow values we'd have to have an "undefined def" pseudoinst prior to it which would add a lot of bloat/compile-time slowdown otherwise.

cranelift/assembler-x64/meta/src/instructions/mov.rs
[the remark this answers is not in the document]
abrown: Yeah, exactly: my plan was was to keep an eye out for this as more rules get converted.

cranelift/assembler-x64/meta/src/instructions/mov.rs
abrown: Perhaps the rule is simple for GPR: every time we see `w(r*{8|16|32})` defined in the assembler, we actually emit code using the right `ReadWriteGpr` type.

For XMM it might be more complex: `w(xmm)` does not tell us that only some of the bits are modified so we may need to examine the other operands?

abrown: Another idea: we add a `partial` bit which gets set automatically for `w(r*{8|16|32})` but set explicitly for XMM, e.g., `w(partial(xmm))`. Then when we decide what type to use for an operand `op`, we would emit the `ReadWrite*` type when `op.is_read_write() || op.is_partial()`.

cfallin: It's possible I'm not fully grokking the proposed rules here, but the end state needs to be: writes to GPRs, even narrow writes, must be only `w`, not `rw`. This is a reflection of the semantic gap between CLIF / VCode data types and the machine: when we store an `i8` (say)-typed vreg in a particular register, it really is only 8 bits wide, and the rest of the machine register is unused. As long as uses and defs of vregs are properly typed, the undefined bits never affect anything.

(This is different than the Xmm case, as you note too -- a lane insertion is operating on a 128-bit input and producing a 128-bit output, so we don't have the same "narrow type in wide register" gap.)

We almost certainly don't want a full type system at the assembler layer, but in the absence of that, we do need to represent this gap somehow. There's a bit of tension with the "general assembler library" use-case, where the latter may actually want to represent that the writes are partial, etc. Perhaps we have a mode bit somewhere indicating that partial GPR writes are to be treated as full writes (a `high_gpr_bits_are_undefined()` method on the toplevel Regs trait so we can plug in a constant `true` impl from Cranelift, maybe?). Or there's always the option that we adopt the Cranelift convention and, w.r.t. regalloc metadata, assume that the embedder is only doing things in a properly typed way, so narrow writes leave undefined bits rather than previously-present bits...

alexcrichton: I'm a bit surprised because I thought `rw(rm8)` is basically almost exactly what we want here? It matches the raw instruction because it's a single operand, and it matches Cranelift because the actual operand is `api::Registers::ReadWriteGpr` which is modeled as two `VReg`s in Cranelift (which during emisson is required to be the same register). Additionally for generic assembler purposes there's nothing saying that `ReadWriteGpr` has to be different than `WriteGpr`, so I'd imagine a generic usage of this crate would have the same type plugged in there.

Otherwise I'm curious where your concern about compile-time comes from @cfallin? It seems like we're not actually using `movb` anywhere unless we're storing to memory, and register moves generated by regalloc always otherwise use `movq` to overwrite the entire destination.

---

Orthogonally though one possible change we may want to consider: should this instruction be split into two forms? If the destination is memory that's technically not a read/write operand, it's a pure write in this case. That doesn't actually get surfaced anywhere right now but with PCC integration in the future it might? Basically the read/write-ness is a property of the register-to-register encoding, not the register-to-memory encoding.

For the `movss` instruction for example I modeled this where memory-to-xmm is a write of the destination since it zeros upper bits, but xmm-to-xmm is a read/write of the destination as it only overwrites the lower bits. Should something similar be done here where the register-to-memory version of `movb` uses `w(m8)` while the register-to-register version uses `rw(r8)`?

cfallin: >  We should then define the width of the write as part of the metadata somehow

or rather, I suppose that's just the `8` in `rm8`. So `w(rm8)` is correct in this interpretation, and means "we're writing 8 bits" -- we don't read the 8 bits first, so it's not `rw(rm8)`.

alexcrichton: I like that, it makes a lot of sense to me 👍 

I particularly like the idea of "ok let's model how some bits are preserved" is a special-case variant of an instruction we may expose in some circumstances but aren't necessarily required to.

abrown: Yeah, after talking about this with @cfallin, I think the right thing to at the assembler level is to use `w`, i.e., " define the r/rw/w's "scope" to be only the width of the instruction." What I realized I was trying to do is avoid footguns up at the ISLE level, but we (1) are already relying on Cranelift to use the right widths and (2) we can be safer another way (e.g., we _know_ a `w(r8)` is actually a partial write so we could still generate a `ReadWriteGpr` parameter for that operand... or a `type PartialWriteGpr = ReadWriteGpr`?). I'll update this PR with all `w` for now and follow it up with a different PR that we can throw darts at.

PR #10875, opened 2025-05-30
Fix a MIRI stacked borrows violation with Pulley

alexcrichton:
This commit adds a new tests which MIRI flags a stacked borrows violation with and then fixes the test. This has to do with recursively calling WebAssembly and juggling the various bits of mutable state associated with the Pulley VM. I don't believe that this is a soundness issue in the sense of miscompiles, but it's nevertheless something good to fix.

### DIFF
--- a/crates/wasmtime/src/runtime/vm/interpreter.rs
+++ b/crates/wasmtime/src/runtime/vm/interpreter.rs
@@ -1,9 +1,9 @@
-use crate::prelude::*;
 use crate::runtime::vm::vmcontext::VMArrayCallNative;
 use crate::runtime::vm::{
-    TrapRegisters, TrapTest, VMContext, VMOpaqueContext, f32x4, f64x2, i8x16, tls,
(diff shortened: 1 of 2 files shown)


crates/wasmtime/src/runtime/vm/interpreter.rs
```diff
@@ -12,17 +12,45 @@ use wasmtime_environ::{BuiltinFunctionIndex, HostCall, Trap};
...
+    /// re-acquire from store state to re-assert that it has an exclusive
+    /// borrow.
+    ///
+    /// This in turn models how VM state could be modified as part of the
+    /// recursive function call, for example with another VM execution itself.
+    ///
+    /// Note that the safety of this all relies not only on correctly managing
+    /// this pointer but it also requires that this pointer is never
+    /// deallocated. This private field is never overwritten which ensures such
+    /// a guarantee.
```
fitzgen, line 45: Never deallocated *for the lifetime of the store*, right? We aren't leaking Pulley VMs indefinitely. Would be good to clarify that in this comment.

PR #10876, opened 2025-05-30
Update release builds from Ubuntu 16.04 to 18.04

alexcrichton:
We're getting quite a lot of spurious failures on CI due to a failure to install packages in Ubuntu 16.04 images. According to https://ubuntu.com/about/release-cycle Ubuntu 16.04 ended standard support 2 years ago and is nearing the end of "Pro Support". Overall it seems clear that this version is on the way out and mirrors are getting flakier, so this commit updates the images to 18.04 in an attempt to make CI more reliable under the assumption that mirrors for more recent versions of the distribution are more reliable.

This has the consequence though that release binaries are going to have a higher glibc requirement. That means if anyone is Downloading Wasmtime onto these older systems and running it then our precompiled binaries will no longer work.

### DIFF
--- a/ci/docker/aarch64-linux/Dockerfile
+++ b/ci/docker/aarch64-linux/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:16.04
+FROM ubuntu:18.04
 
 RUN apt-get update -y && apt-get install -y gcc gcc-aarch64-linux-gnu ca-certificates curl make git ninja-build
 RUN git config --global --add safe.directory '*'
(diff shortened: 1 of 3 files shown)


abrown (approved):

abrown:
This glibc change might be something to note in the Wasmtime meeting... or the release notes eventually?

alexcrichton:
I'm actually going to close this in favor of https://github.com/bytecodealliance/wasmtime/issues/10930 and https://github.com/bytecodealliance/wasmtime/issues/10931

PR #10877, opened 2025-05-30
Update the definition of `wasmtime::Memory`

alexcrichton:
Powered by the previous commit this change updates the internal implementation details of `wasmtime::Memory` to avoid the need for the use of `Stored`. This means that converting from a Wasmtime-internal representation of a memory to an external representation of a memory is largely a noop and no longer requires an allocation. This is intended to be more useful in the future to use the public types more pervasively inside of Wasmtime itself, but for now this is mostly the scaffolding necessary for such a change.


alexcrichton:
This is next up in the list from https://github.com/bytecodealliance/wasmtime/pull/10870

fitzgen (commented):
Very nice!

PR #10879, opened 2025-05-30
Add a `DefinedTableIndex` to table imports/exports

alexcrichton:
This commit extends `ExportTable` and `VMTableImport` (renamed from `VMTable`) to include a `DefinedTableIndex` in the same manner that memories carry their index they're defined at as well. The main goal of this change is to power the next change which will update how tables are stored in a store.


alexcrichton:
This is another step from https://github.com/bytecodealliance/wasmtime/pull/10870

PR #10880, opened 2025-05-30
Merge `vm::Instance::vmctx_plus_offset{,_mut}`

alexcrichton:
This commit merges two helper methods to calculate an internal pointer with a `VMContext`. Historically these two methods mapped different levels of mutability but nowadays there's no need for this. The provenance/mutability of `VMContext` memory is independent of `&self` meaning that it's safe to create/mutate through a `NonNull` in the vmctx area through a `&self`-derived pointer.

The purpose of this commit is to avoid needlessly requiring `&mut self` throughout methods on `Instance`. Currently many of them require mutability to only create a pointer, and this mutability is a burden that need not be required.


alexcrichton:
Hm ok thinking more about this is mainly means that `Sync` is no longer correctly inferred, as `Instance` should no longer be `Sync`. I feel like we probably do want to keep it `Sync`, though, so now I'm souring a bit on this approach.

I think I have an idea of how to tackle this, improve safety, and keep `Sync`, so converting this to a draft.

alexcrichton:
Ok I have now applied the refactorings intended and this I believe preserves the send/sync-ness by default a bit better. This is by no means all that's necessary to uphold that, but it's at least not regressing too much.

alexcrichton:
I should also note this is split from https://github.com/bytecodealliance/wasmtime/pull/10870

crates/wasmtime/src/runtime/vm/instance.rs
```diff
@@ -384,22 +384,72 @@ impl Instance {
...
+    /// memory must only happen where appropriate, for example a non-atomic
+    /// write (as most are) should never happen concurrently with another read
+    /// or write. It's generally on the burden of the caller to adhere to this.
+    ///
+    /// Also of note is that most of the time the usage of this method falls
+    /// into one of:
+    ///
+    /// * Something in the VMContext is being read or written. In that case use
+    ///   `vmctx_plus_offset` or `vmctx_plus_offset_raw` if possible due to
+    ///   that having a safer lifetime.
```
fitzgen:
```suggestion
    /// * Something in the VMContext is being read or written. In that case use
    ///   `vmctx_plus_offset` or `vmctx_plus_offset_mut` if possible due to
    ///   that having a safer lifetime.
```

Right?

alexcrichton: Oops, yes!

PR #10881, opened 2025-05-30
Add pavg vpavg variants to the new assembler

jlb6740:

### DIFF
--- a/cranelift/assembler-x64/meta/src/instructions/avg.rs
+++ b/cranelift/assembler-x64/meta/src/instructions/avg.rs
@@ -1,0 +1,16 @@
+use crate::dsl::{Feature::*, Inst, Location::*, VexLength::*};
+use crate::dsl::{align, fmt, inst, r, rex, vex, w};
+
+#[rustfmt::skip] // Keeps instructions on a single line.
+pub fn list() -> Vec<Inst> {
+    vec![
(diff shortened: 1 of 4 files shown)


cranelift/assembler-x64/meta/src/instructions/avg.rs
```diff
@@ -0,0 +1,16 @@
+use crate::dsl::{Feature::*, Inst, Location::*, VexLength::*};
+use crate::dsl::{align, fmt, inst, r, rex, vex, w};
+
+#[rustfmt::skip] // Keeps instructions on a single line.
+pub fn list() -> Vec<Inst> {
+    vec![
+        inst("pavgb", fmt("A", [w(xmm1), r(align(xmm_m128))]), rex([0x66, 0x0F, 0xE0]), _64b | compat | sse2),
+        inst("pavgw", fmt("A", [w(xmm1), r(align(xmm_m128))]), rex([0x66, 0x0F, 0xE3]), _64b | compat | sse2),
```
alexcrichton: I think these need to be `rw(xmm1)` along with updating the ISLE below to pass tests?

jlb6740: @Alex .. hadn't seen your comment until now, but I just discovered this while working on pcmp but using pavg as a template. Will try this out and see if it resolves.

PR #10886, opened 2025-05-31
x64: Flag unpack instructions as requiring aligned loads

alexcrichton:
This commit fixes a minor mistake from #10842 where the instructions weren't tagged as requiring aligned loads meaning that wasm loads could get folded into the instruction, producing unexpected alignment faults.

A test is added here which would trigger this a bit more quickly during fuzzing as fuzzing runs `*.wast` tests with a variety of codegen parameters, and would in theory eventually run this without AVX.


abrown (approved):
Good catch! We should probably add a `validate` rule for this at some point; I keep missing these in reviews.

PR #10887, opened 2025-05-31
x64: Migrate `mulx` to the new assembler

alexcrichton:
This is an interesting instruction as it has a relatively unique shape compared to many others. The VEX encoding is used to give it a 3-operand form, although it still has an implicit 4th operand as well. The other unique part about this instruction is that if the two write-only operands are the same then that has a different semantic meaning than if they are different.

Modeling the two-output form of the instruction was pretty easy, the only changes needed were to add the `r32a` and `r32b` locations as previously only `r32` was available. Modeling the one-output form of the instruction led to a "hook" where these instructions specify that they use a custom regalloc function. That skips the auto-generated regalloc entirely and defers to a new `custom` submodule in the assembler crate. These custom variants handle only having a single write operand on the instruction.


cranelift/assembler-x64/meta/src/instructions/mul.rs
```diff
@@ -23,6 +23,15 @@ pub fn list() -> Vec<Inst> {
...
         inst("imulq", fmt("RMI_SXL", [w(r64), r(rm64), sxq(imm32)]), rex(0x69).w().id(), _64b),
+        // MULX, a BMI2 extension
+        //
+        // Note that these have custom visitors for regalloc to handle the case
+        // where both output operands are the same. This is where the
+        // instruction only calculates the high half of the multiplication, as
+        // opposed to when the operands are different to calculate both the high
+        // and low halves.
+        inst("mulxl", fmt("RVM", [w(r32a), w(r32b), r(rm32), r(implicit(edx))]), vex(LZ)._f2()._0f38().w0().op(0xF6), _64b | compat | bmi2).custom_visit(),
+        inst("mulxq", fmt("RVM", [w(r64a), w(r64b), r(rm64), r(implicit(rdx))]), vex(LZ)._f2()._0f38().w1().op(0xF6), _64b | bmi2).custom_visit(),
```
abrown, line 34: I really like this `custom*` idea: I'd been mulling over how to do custom encodings for `nop` and custom printing for comparisons and felt we needed an escape hatch for very special instructions. And I like that we just expect `custom::visit_{inst}` to be present, which should be pretty clear in compile errors.

I don't know if you were thinking this way, but maybe in the future we can go to `.custom(Visit | Encode | Display)` which would result in an appropriate call to `custom::visit::{inst}`, `custom::encode::{inst}`, and `custom::display::{inst}`. Thoughts?

cranelift/codegen/src/isa/x64/inst.isle
```diff
@@ -2627,21 +2612,29 @@
...
+    (value_regs (value_regs_get regs 1) (value_regs_get regs 0))))
 
 (decl x64_mulx_hi (Type Gpr GprMem) Gpr)
-(rule (x64_mulx_hi (ty_32_or_64 ty) src1 src2)
-  (let ((dst WritableGpr (temp_writable_gpr))
-        (size OperandSize (raw_operand_size_of_type ty))
-        (_ Unit (emit (MInst.MulX size src1 src2 (writable_invalid_gpr) dst))))
-    dst))
+(rule (x64_mulx_hi $I32 src1 src2) (x64_mulxl_rvm_hi src2 src1))
+(rule (x64_mulx_hi $I64 src1 src2) (x64_mulxq_rvm_hi src2 src1))
```
abrown, line 2632: I guess there's no way to pass the same temporary register in... I wonder if eventually we need to allow even more fine-grained control from ISLE &mdash; a few new `gen_asm.rs` rules?

cranelift/assembler-x64/meta/src/instructions/mul.rs
alexcrichton: I think that's reasonable yeah!

cranelift/codegen/src/isa/x64/inst.isle
alexcrichton: I was wondering the same yeah, but I figured that if it's just mulx then there's no need to go to such lengths, so I figured I'd leave it as-is for now. If the number of instructions needing this grows though I can see how making new ISLE helpers would be reasonable

PR #10889, opened 2025-06-01
x64: Simplify ModR/M byte emission in new assembler 

alexcrichton:
Right now there are two intertwined `match` statements in the DSL
between `generate_{rex,vex}_prefix` and `generate_modrm_byte`. Whenever
one is updated it typically is accompanied with a modification of the
other. This commit is an attempt to simplify this and to also decouple
the two a bit more.

Specifically this commit adds a new `ModRmStyle` enum to the code
generator which is a structure returned from `generate_{r,v}ex_prefix`.
This is then passed to `generate_modrm_byte` and means that it doesn't
need to perform the same pattern-match on the operands. This simplifies
the `generate_modrm_byte` method and enables selection of the ModR/M
byte to be different in REX/VEX as will be required for the `blsi`
instruction. For `blsi` the same list of operands is encoded differently
with VEX than it is with REX, and this will enable that decoupling to be
easier to specify.

### DIFF
--- a/cranelift/assembler-x64/meta/src/generate/format.rs
+++ b/cranelift/assembler-x64/meta/src/generate/format.rs
@@ -148,42 +148,24 @@ impl dsl::Format {
             },
             [FixedReg(_), RegMem(mem)]
             | [FixedReg(_), FixedReg(_), RegMem(mem)]
-            | [RegMem(mem), FixedReg(_)] => {
+            | [RegMem(mem), FixedReg(_)]
+            | [Mem(mem), Imm(_)]
+            | [RegMem(mem), Imm(_)]
+            | [RegMem(mem)] => {
                 let digit = rex.unwrap_digit().unwrap();
                 fmtln!(f, "let digit = 0x{digit:x};");
                 fmtln!(f, "let rex = self.{mem}.as_rex_prefix(digit, {bits});");
                 ModRmStyle::RegMemWithDigit { mem: *mem, digit }
             }
-            [Mem(dst), Imm(_)] | [RegMem(dst), Imm(_)] | [RegMem(dst)] => {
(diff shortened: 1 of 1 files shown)


cranelift/assembler-x64/meta/src/generate/format.rs
```diff
@@ -3,6 +3,36 @@
...
+        reg: dsl::Location,
+        mem: dsl::Location,
+    },
+
+    /// The R/M bits are encoded with `reg`, and the Reg/Opcode bits are encoded
+    /// with `dst`.
+    RegReg {
+        reg: dsl::Location,
+        dst: dsl::Location,
+    },
```
abrown, line 33: I wonder if we should merge these two into `RegRm { reg: ..., rm: ... }` since that ends up being the way the manual describes this.

cranelift/assembler-x64/meta/src/generate/format.rs
```diff
@@ -159,97 +194,71 @@ impl dsl::Format {
...
-            | [RegMem(mem), Reg(reg)]
-            | [RegMem(mem), Reg(reg), Imm(_)]
-            | [RegMem(mem), Reg(reg), FixedReg(_)]
-            | [Reg(reg), Reg(_), RegMem(mem), Imm(_)]
-            | [Reg(reg), Mem(mem)] => {
+            ModRmStyle::RegMem { reg, mem } => {
                 fmtln!(f, "let reg = self.{reg}.enc();");
                 fmtln!(
                     f,
                     "self.{mem}.encode_rex_suffixes(buf, off, reg, {bytes_at_end});"
```
abrown, line 256: Ah, I see: I guess we do need to differentiate `RegMem` and `RegReg` for this kind of thing.

abrown (approved):
Good call!

PR #10894, opened 2025-06-02
x64: convert `movzx` and `movsx` to the new assembler

abrown:
This sequence of commits replaces all old uses of `movzx` and `movsx` to use instructions in the new assembler. The trickiest part of these lowerings is the use of `movl` for the hypothetical `movzxlq` instruction: I have documented why this is legal in the various places this is used.


fitzgen (approved):
Nice!

PR #10896, opened 2025-06-02
Add a new `ComponentInstanceId` type

alexcrichton:
This is intended to be similar to `InstanceId`, but used for components.

### DIFF
--- a/crates/wasmtime/src/runtime/component/store.rs
+++ b/crates/wasmtime/src/runtime/component/store.rs
@@ -1,5 +1,7 @@
+use crate::component::instance::InstanceData;
 use crate::prelude::*;
-use crate::store::{StoreData, StoredData};
+use crate::runtime::vm::component::ComponentInstance;
+use crate::store::{StoreData, StoreOpaque, Stored, StoredData};
 
 macro_rules! component_store_data {
     ($($field:ident => $t:ty,)*) => (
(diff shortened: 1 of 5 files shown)


alexcrichton:
This is split out of https://github.com/bytecodealliance/wasmtime/pull/10870 and is needed for a method of defining globals where globals can be flags for instances.

crates/wasmtime/src/runtime/component/store.rs
```diff
@@ -25,5 +27,43 @@ macro_rules! component_store_data {
...
+            data.instance().id(),
+            self.components.next_component_instance_id()
+        );
+        self.insert(Some(data))
+    }
+}
+
+impl StoreOpaque {
+    #[expect(dead_code, reason = "to be used soon")]
+    pub(crate) fn component_instance(&self, id: ComponentInstanceId) -> &ComponentInstance {
```
dicej, line 63: Will we also be adding a `component_instance_mut` method for getting a `&mut ComponentInstance` from a `&mut self`?

dicej (approved):

crates/wasmtime/src/runtime/component/store.rs
alexcrichton: Indeed! I'll go ahead and add that in to indicate that it's desired too.

alexcrichton: Hm ok no just kidding. There's not a safe `instance_mut` accessor due to my historical (and probably irrational paranoia). When we move more towards indices though I believe it should be safe to add that, but it'll need some other changes too.

PR #10897, opened 2025-06-02
Refactor the representation of `Func`

### LINKED ISSUE #10868: Memory "leak" when passing funcrefs to the host

alexcrichton:
This commit rewrites the internals of `wasmtime::Func`. The `Stored` type is no longer used meaning that it's now free to create a `wasmtime::Func` at any time. It is effectively a store-tagged `NonNull<VMFuncRef>`. This required a few internal changes to how functions are passed around:

* Previously the insertion of a `wasm_call`-less `VMFuncRef` was deferred until the raw pointer was loaded. Now the insertion happens immediately as soon as the function is placed within a store. This is done to ensure that a `Func` corresponds to one exact `VMFuncRef` and that's it, and lazily filled in versions within the store are still lazily filled in but they're more eagerly allocated. This isn't expected to have much of an impact perf-wise since all these lazily allocated functions were already almost guaranteed to get lazily allocated anyway.

* Filling in "holes" in `VMFuncRef`, notably the `wasm_call` field, no longer happens lazily when the `VMFuncRef` is demanded. Instead during instantiation a pass is made to fill in holes with the new module being instantiated (after registration). Additionally when a lazily-allocated `VMFuncRef` is created the module registry is checked immediately. This means that all store-local `VMFuncRef` values are either filled in immediately or filled in during instantiation. This notably means that the previous logic in `Func::vmimport` is now "just" an `.unwrap()` with a lot of comments saying why the unwrap shouldn't panic.

* To implement this commit a previous optimization for the `Func` API was removed as well, namely `Func::call` will become slower after this commit. The `Func::call` API is a dynamically-typed API which requires run-time type-checking of arguments. Previously a `FuncType` was loaded into a cache once-per-`Func` which helped amortize the cost of using `Func::call` repeatedly. Now, though, there's no natural place to put such a cache since `Func` no longer has dedicated storage within a `Store`. Historically this optimization was added for the C API before the `*_call_unchecked` APIs existed, but nowadays the `*_call_unchecked` APIs should suffice for performance-critical applications where needed. In the future it might also be possible to have a hash map in the `Store` of a `VMSharedTypeIndex` to `FuncType` which is lazily populated based on calls to `Func::call`, but that feels a bit overkill nowadays for a possibly rarely-used map.

* The no-longer-necessary `RootedHostFunc` type is now gone as its unsafety and various contracts are subsumed by other preexisting `unsafe` blocks.

* The specific drop order between `StoreOpaque::store_data` and `StoreOpaque::rooted_host_funcs` is removed. The `rooted_host_funcs` field now lives in the `func_refs` field and the `FuncKind` type, where the destructors came from before, is no more.

* The `func_refs` field has grown storage locations for a variety of "keep this thing alive as long as the store" related to functions and such.

Closes #10868

### DIFF
--- a/crates/wasmtime/src/runtime/func.rs
+++ b/crates/wasmtime/src/runtime/func.rs
@@ -1227,7 +1227,7 @@ impl Func {
         unsafe {
             let f = self.vm_func_ref(store);
             VMFunctionImport {
-                // Note that this is a load-bearing assertion here, but is
+                // Note that this is a load-bearing `unwrap` here, but is
                 // never expected to trip at runtime. The general problem is
                 // that host functions do not have a `wasm_call` function so
                 // the `VMFuncRef` type has an optional pointer there. This is
--- a/crates/wasmtime/src/runtime/func.rs
+++ b/crates/wasmtime/src/runtime/func.rs
@@ -4,11 +4,11 @@ use crate::runtime::vm::{
     ExportFunction, InterpreterRef, SendSyncPtr, StoreBox, VMArrayCallHostFuncContext, VMContext,
     VMFuncRef, VMFunctionImport, VMOpaqueContext, VMStoreContext,
 };
-use crate::store::{AutoAssertNoGc, StoreData, StoreOpaque, Stored};
+use crate::store::{AutoAssertNoGc, StoreId, StoreOpaque};
 use crate::type_registry::RegisteredType;
 use crate::{
-    AsContext, AsContextMut, CallHook, Engine, Extern, FuncType, Instance, Module, ModuleExport,
-    Ref, StoreContext, StoreContextMut, Val, ValRaw, ValType,
+    AsContext, AsContextMut, CallHook, Engine, Extern, FuncType, Instance, ModuleExport, Ref,
+    StoreContext, StoreContextMut, Val, ValRaw, ValType,
 };
 use alloc::sync::Arc;
 use core::ffi::c_void;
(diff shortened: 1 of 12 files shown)


alexcrichton:
I'll also note that this is split out of https://github.com/bytecodealliance/wasmtime/pull/10870

crates/wasmtime/src/runtime/store/func_refs.rs
```diff
@@ -53,40 +102,108 @@ impl FuncRefs {
...
+    ) -> NonNull<VMFuncRef> {
+        debug_assert!(func.func_ref().wasm_call.is_none());
+        let ret = self.push(func.func_ref().clone(), modules);
+        self.storage.push(Storage::BoxHost { func });
+        ret
+    }
+}
+
+/// Attempts to fill the `wasm_call` field of `func_ref` given `modules`
+/// registered.
```
fitzgen:
```suggestion
/// Attempts to fill the `wasm_call` field of `func_ref` given `modules`
/// registered and returns `true` if the field was filled, `false` otherwise.
```

crates/wasmtime/src/runtime/func.rs
```diff
@@ -261,77 +261,33 @@ impl NoFunc {
...
-}
+// Double-check that the C representation in `extern.h` matches our in-Rust
+// representation here in terms of size/alignment/etc.
+const _: () = {
+    #[repr(C)]
+    struct C(u64, *mut u8);
+    assert!(core::mem::size_of::<C>() == core::mem::size_of::<Func>());
+    assert!(core::mem::align_of::<C>() == core::mem::align_of::<Func>());
+    assert!(core::mem::offset_of!(Func, store) == 0);
+};
```
fitzgen, line 290: Kind of an aside but should all of these assertions actually live in the `wasmtime-c-api` crate so that we don't need to have a dummy struct which itself needs to be kept in sync with the struct that we actually care about? We've removed one syncing problem but added another, so we haven't really changed anything. I guess the only assertion that would remain in the `wasmtime` crate would be the one for the offset of the `store` field.

crates/wasmtime/src/runtime/func.rs
```diff
@@ -261,77 +261,33 @@ impl NoFunc {
...
-    /// host functions that are packaged up in an `Arc`, so the `Arc<[T]>` is
-    /// cloned once into the `Store` to avoid each individual function requiring
-    /// an `Arc::clone`.
-    ///
-    /// The lifetime management of this type is `unsafe` because
-    /// `RootedHostFunc` is a small wrapper around `NonNull<HostFunc>`. To be
-    /// safe this is required that the memory of the host function is pinned
-    /// elsewhere (e.g. the `Arc` in the `Store`).
-    RootedHost(RootedHostFunc),
-}
```
fitzgen, line 334: Sooooooo nice to get rid of `FuncKind`

crates/wasmtime/src/runtime/func.rs
```diff
@@ -1273,88 +1211,55 @@ impl Func {
...
-                    let sig = self.type_index(store.store_data());
-                    module
-                        .wasm_to_array_trampoline(sig)
-                        .expect(
-                            "if the wasm is importing a function of a given type, it must have the \
-                         type's trampoline",
-                        )
-                        .into()
-                },
+                // Note that this is a load-bearing assertion here, but is
```
fitzgen: Just to clarify, since there isn't any `assert!` here:

```suggestion
                // Note that this is a load-bearing `unwrap` here, but is
```

crates/wasmtime/src/runtime/func.rs
alexcrichton: Oh I think you might be assuming there's a `struct wasmtime_func_t { ... }` in Rust in `crates/c-api` but that doesn't exist. In that sense this `struct C` is the only C-version-written-in-Rust anywhere in the codebase, and I figured here was a good as place as any to test things out.

fitzgen: Ah, I was indeed

PR #10898, opened 2025-06-02
x64: convert `movlhps`

abrown:
This change continues the push to convert all `mov`-related instruction and converts `movlhps` to use the new assembler (at least the SSE variant). In doing so, it discovers that we were incorrectly allowing an `XmmMem` as one of the operands; this is fixed to be `Xmm` only.


tests/disas/pcc-insertlane-x64-avx.wat
```diff
@@ -138,9 +138,10 @@
 ;; wasm[0]::function[5]:
 ;;       pushq   %rbp
 ;;       movq    %rsp, %rbp
-;;       movl    %edx, %r9d
-;;       movq    0x40(%rdi), %r10
-;;       vmovhps (%r10, %r9), %xmm0, %xmm0
+;;       movl    %edx, %r10d
+;;       movq    0x40(%rdi), %r11
+;;       vmovsd  (%r11, %r10), %xmm7
+;;       vmovlhps %xmm7, %xmm0, %xmm0
```
abrown: Note how we're gaining an extra `vmovsd` here because we were probably doing some internal `emit`-time matching to emit `vmovhps`.

tests/disas/pcc-insertlane-x64.wat
```diff
@@ -143,8 +143,8 @@
 ;;       movq    %rsp, %rbp
 ;;       movl    %edx, %r10d
 ;;       movq    0x40(%rdi), %r11
-;;       movdqu  (%r11, %r10), %xmm6
```
abrown, line 146: The old version seems wrong... this _should_ be a `movsd` I believe.

tests/disas/pcc-insertlane-x64-avx.wat
alexcrichton: It looks like `movhps` has the exact same encoding as `movlhps`, which might be how this worked beforehand? Nevertheless from an assembler point of view I'd much rather match the Intel manual for sure.

Perhaps though `XmmMem` could stay as an argument on some handwritten ISLE-helper and that delegates to `movlhps` and `movhps` as appropriate?

tests/disas/pcc-insertlane-x64.wat
alexcrichton: Yeah definitely wrong, `XmmMem` auto-converted to `Xmm` always loads 128-bits, and this is just an incorrect translation and means that on-the-boundary-of-linear-memory loads of f32/f64 will be incorrectly identified as out-of-bounds.

alexcrichton: For this, mind adding [this test](https://github.com/alexcrichton/wasmtime/commit/3a34e2bac097ddce38da0fa0b3f3ab7886696c4a) to this PR?

tests/disas/pcc-insertlane-x64-avx.wat
abrown: Yeah, the more I stare at the manual, it could be that they are truly the same instruction and we _should_ make that second operand an `xmm_m64`. I suspect that the names were different for some historical reason and we'll just have to paper over the mnemonic difference with #10918 once that is done.

abrown: Ok, I ended up creating a separate instruction with the same encoding (avoids any customizations) and everything is fine. The only real user of `vmovhps` is Winch, though. I looked through the five uses of `x64_movlhps` in `lower.isle` and, unless I'm missing some auto-conversion somewhere, it appears to me that we are fine using the `movlhps` reg-to-reg form and don't really need `movhps` in Cranelift.

tests/disas/pcc-insertlane-x64.wat
abrown: Added. It doesn't fail, though?

abrown: (Maybe it would have failed with the old version?)

alexcrichton:
Failure is [here](https://github.com/bytecodealliance/wasmtime/actions/runs/15598622270/job/43934350268#step:19:1209) and will need a new exception [here](https://github.com/bytecodealliance/wasmtime/blob/fcf4e0e4f243485ed3ab133dc104c8c10cfd6ca9/crates/test-util/src/wast.rs#L484-L574)

PR #10899, opened 2025-06-02
x64: convert `pmov*x`

abrown:
This change converts all `pmovsx*` and `pmovzx*` instructions to use the new assembler.


abrown:
We can wait on merging this until #10881 is resolved to avoid merge conflicts for @jlb6740.

cranelift/assembler-x64/meta/src/dsl/format.rs
```diff
@@ -301,6 +301,10 @@ pub enum Location {
     rm64,
 
     // XMM registers, and their memory forms.
+    xmm1,
```
cfallin, line 308: I see these are pre-existing (this part of the diff is just code motion) but even so, I'm not sure I understand what these variants mean -- could we have doc-comments on `xmm1`, `xmm2`, `xmm3` describing what they're for and how they differ?

cfallin (approved):
Looks good overall, just a doc-comment request below; thanks!

cranelift/assembler-x64/meta/src/dsl/format.rs
jlb6740: @cfallin When describing an instruction a unique location is required for each reg/mem operand to avoid name clashes when generating lowering code in assembler.rs. Before vex, there was just a single xmm location which was adequate, but didn't breakdown xmm registers the way gpr registers were already broken down (eax, edx, etc). When describing vex instructions that use multiple xmm registers, just having "xmm" describing all xmm locations was no longer adequate due to name clashing. There was an attempt append subscripts to xmm locations while generating code in assembler.rs, but it became clear that it made the most sense to just enumerate the xmm registers the same way the gpr registers already were. We don't need to describe all xmm registers (nor do we for gpr), we only need about 3. When describing instructions, the naming reads the same as is used as the Intel manual.

cfallin: OK, I'm not sure I understand why we need to define new *types of operands* though -- can't we have DSL methods that somehow specify which slot the registers correspond to?

In essence this seems like a category error to me: the enum so far defines the kinds of things we can operate on; an XMM register, a reg-or-mem, that sort of thing. Whether we use that as input X or Y, it's still that kind of thing. The fact that we use the operand in a certain way should be a property of the *instruction*, not the operand kind. The concern I have with this kind of conflation is that it's not at all clear to me (or presumably others in the future) which of these variants to use or what they mean -- a recipe for subtle bugs and footguns in the future. So I'd rather we find a way to store the information elsewhere...

jlb6740:
> We can wait on merging this until #10881 is resolved to avoid merge conflicts for @jlb6740.

@abrown Thanks. #10881 should be resolved now.

cranelift/assembler-x64/meta/src/dsl/format.rs
abrown: Here's what is going on: this `enum Location` is actually the _concrete_ list of names accepted as operands, not a classification of operands. There are three reasons why this became concrete: (1) it makes it really simple for us to write `rax` or `m16` or `r32b` when defining instructions, (2) that name then matches how the manual defines the instruction, and (3) that name is then unique for the operand and we can use it as Rust field, e.g.

When it comes to _classifications_ of operands, we actually have two different ways of categorizing operands:
- `enum OperandKind`, which we use to distinguish between `Reg`, `Mem`, etc.
- `enum RegClass`, which we use to distinguish between `Xmm` and `Gpr`

The concrete `Location` names get categorized appropriately in `Location::kind()` and `Location::reg_class()`. Most of the assembler code generation uses their output. I think it would be helpful to summarize all of this this text above in the doc comments of `Location`, which is what @cfallin you were originally asking, right? I think that would be a good addition: do you have any specific wording choices that would be most helpful here?

abrown: For later reference: we discussed this in the Cranelift meeting today and @cfallin is planning to look into this a bit more. There are other ways to factor this that avoids having the concrete things in the same bucket with general ones while still "looking like the manual" and having unique Rust field names; @alexcrichton had some thoughts to write up in an issue. I'll go ahead and merge this for now since this discussion was unrelated to this PR.

abrown: See https://github.com/bytecodealliance/wasmtime/issues/10922 for future discussion.

PR #10902, opened 2025-06-02
Refactor globals to no longer use `Stored`

alexcrichton:
This commit refactors the `wasmtime::Global` to avoid the usage of `Stored<T>` internally. This makes conversion from internal global state to external global state a noop along the lines of previous commits. The end goal is to remove `Stored` entirely and enable more pervasively using external types internally within Wasmtime as well.

Globals were different than the prior iterations of memories, tags, and tables. Globals have three different ways of defining them: wasm instances, the host embedder, and component flags. Representing these in all the various locations required a bit of finesse in how everything is represented and stored at rest and such. In the end there's a small amount of "type punning" in a few instance/vmctx fields related to globals now since everything is squeezed into one slot. This is required because the `VMGlobalImport` structure must have a size known to wasm-compiled code and `wasmtime::Global` must have a known layout for C code.

In the end while this is more code to manage globals my hope is that the end result will be a net negative in terms of complexity by ensuring that the embedder API is additionally suitable for use internally within Wasmtime as well.

### DIFF
--- a/crates/wasmtime/src/runtime/externals/global.rs
+++ b/crates/wasmtime/src/runtime/externals/global.rs
@@ -1,12 +1,16 @@
 use crate::prelude::*;
+use crate::runtime::vm::{
+    self, ExportGlobalKind, VMGlobalDefinition, VMGlobalKind, VMOpaqueContext,
+};
 use crate::{
     AnyRef, AsContext, AsContextMut, ExternRef, Func, GlobalType, HeapType, Mutability, Ref,
     RootedGcRefImpl, Val, ValType,
-    store::{AutoAssertNoGc, StoreData, StoreOpaque, Stored},
+    store::{AutoAssertNoGc, InstanceId, StoreId, StoreOpaque},
     trampoline::generate_global_export,
 };
 use core::ptr;
-use wasmtime_environ::TypeTrace;
+use core::ptr::NonNull;
+use wasmtime_environ::{DefinedGlobalIndex, TypeTrace};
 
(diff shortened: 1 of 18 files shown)


alexcrichton:
Note that this is split out of https://github.com/bytecodealliance/wasmtime/pull/10870

dicej (approved):

crates/wasmtime/src/runtime/externals/global.rs
```diff
@@ -229,36 +266,111 @@ impl Global {
...
+
     /// Get a stable hash key for this global.
     ///
     /// Even if the same underlying global definition is added to the
     /// `StoreData` multiple times and becomes multiple `wasmtime::Global`s,
     /// this hash key will be consistent across all of these globals.
     #[cfg(feature = "coredump")]
     pub(crate) fn hash_key(&self, store: &StoreOpaque) -> impl core::hash::Hash + Eq + use<> {
-        store[self.0].definition.as_ptr() as usize
+        self.definition(store).as_ptr() as usize
```
fitzgen, line 352: We can avoid `as` for pointer-to-usize conversions now:

```suggestion
        self.definition(store).as_ptr().addr()
```

fitzgen (approved):

crates/wasmtime/src/runtime/externals/global.rs
alexcrichton: Oh nice, I'll do that in a follow-up

PR #10904, opened 2025-06-02
x64: Migrate `UnaryRmRVex*` instructions to new assembler

alexcrichton:
This migrates `blsi`, `blsmsk`, `blsr`, and `rorx` all to the new assembler. Some new formats but otherwise nothing major.


abrown (approved):
Nice!

PR #10905, opened 2025-06-02
Remove dead accessors for a field of `VM*Import`

alexcrichton:
These aren't actually read from compiled wasm code, only from the host.


alexcrichton:
Split out from https://github.com/bytecodealliance/wasmtime/pull/10870

pchickey (approved):

PR #10913, opened 2025-06-03
Remove `Stored` usage from `wasmtime::component::Instance`

alexcrichton:
This is a continuation of the work from https://github.com/bytecodealliance/wasmtime/pull/10870 to remove the usage of `Stored<T>` from Wasmtime entirely. Next on the chopping block is component instances as they feed into a number of other locations. This work is split across a number of commits to be more bite-sized as the integration of `InstanceData` was a bit sprawling. With this removal some other refactorings are now unblocked to use more first-class index types within a Store and eventual removal of `Stored` entirely with its final user, component functions.

### DIFF
--- a/crates/wasmtime/src/runtime/component/component.rs
+++ b/crates/wasmtime/src/runtime/component/component.rs
@@ -2,7 +2,6 @@ use crate::component::InstanceExportLookup;
 use crate::component::matching::InstanceType;
 use crate::component::types;
 use crate::prelude::*;
-use crate::runtime::vm::component::ComponentRuntimeInfo;
 #[cfg(feature = "std")]
 use crate::runtime::vm::open_file_for_mmap;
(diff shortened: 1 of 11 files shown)


crates/wasmtime/src/runtime/component/component.rs
```diff
@@ -93,7 +91,7 @@ struct ComponentInner {
     /// A cached handle to the `wasmtime::FuncType` for the canonical ABI's
     /// `realloc`, to avoid the need to look up types in the registry and take
     /// locks when calling `realloc` via `TypedFunc::call_raw`.
-    realloc_func_type: Arc<dyn Any + Send + Sync>,
+    realloc_func_type: Arc<FuncType>,
```
fitzgen, line 94: Nice

fitzgen (approved):
Nice

PR #10914, opened 2025-06-03
 Refactor the representation of component::Func 

alexcrichton:
This commit updates the implementation of `component::Func` to be
index-based like all other exported items are now in Wasmtime. This
necessitated a new `StoreComponentInstanceId` abstraction similar to
`StoreInstanceId`. Additionally the `component_instance_replace`
function was entirely removed as it's no longer necessary.

This is more continuation of work from https://github.com/bytecodealliance/wasmtime/pull/10870 and removing `Stored` after this should effectively be trivial.


pchickey (approved):

fitzgen (approved):
:tada:

PR #10917, opened 2025-06-03
Remove `Stored<T>`

alexcrichton:
Prior refactorings have removed the need for this abstraction, so delete it entirely (yay!).


fitzgen (approved):
:tada:

PR #10918, opened 2025-06-04
x64: Custom print logic

rahulchaphalkar:
This PR implements custom print logic for `lock` instruction, and reworks the existing logic for `custom_visit` to be more in the form of `.custom(Visit | Display)`. The `custom_visit` logic was originally implemented in https://github.com/bytecodealliance/wasmtime/pull/10887 and the changes are made based on some of these comments https://github.com/bytecodealliance/wasmtime/pull/10887#discussion_r2121785272

The `Display` logic will be used for compare instructions as well.
@abrown


cranelift/assembler-x64/meta/src/dsl/custom.rs
```diff
@@ -0,0 +1,85 @@
+use core::fmt;
+use std::ops::BitOr;
+
+#[derive(PartialEq, Debug)]
+pub enum CustomOperation {
+    Visit,
+    Display,
+    None,
```
abrown:
```suggestion
    Encode,
```

We don't need `None` here because `Custom` can be empty.

cranelift/assembler-x64/meta/src/dsl.rs
```diff
@@ -119,7 +120,7 @@ impl core::fmt::Display for Inst {
         if *has_trap {
             write!(f, " has_trap")?;
         }
-        if *custom_visit {
+        if custom.contains(CustomOperation::Visit) {
             write!(f, " custom_visit")?;
         }
```
abrown:
```suggestion
        if !custom.is_empty() {
            write!(f, " custom({custom})")?;
        }
```

Since we already have `Display` implemented for `Custom`.

cranelift/assembler-x64/meta/src/dsl/custom.rs
```diff
@@ -0,0 +1,85 @@
...
+            !self.0.contains(&rhs),
+            "duplicate custom operation: {rhs:?}"
+        );
+        self.0.push(rhs);
+        self
+    }
+}
+
+#[derive(PartialEq)]
+pub struct Custom(pub Vec<CustomOperation>);
```
abrown:
```suggestion
pub struct Custom(Vec<CustomOperation>);
```

No need for this to be `pub` if you have `iter` below.

cranelift/assembler-x64/meta/src/dsl.rs
```diff
@@ -93,11 +96,9 @@ impl Inst {
         self.has_trap = true;
         self
     }
-
-    /// Flags this instruction as having a custom visit function, skipping the
-    /// auto-generated one.
-    pub fn custom_visit(mut self) -> Self {
-        self.custom_visit = true;
+    /// Indicate this instruction as needing custom processing.
```
abrown, line 99:
```suggestion

    /// Indicate this instruction as needing custom processing.
```

cranelift/assembler-x64/meta/src/dsl.rs
```diff
@@ -32,7 +35,7 @@ pub fn inst(
         encoding,
         features: features.into(),
         has_trap: false,
-        custom_visit: false,
+        custom: CustomOperation::None.into(),
```
abrown:
```suggestion
        custom: Custom::default(),
```

We could `#[derive(Default)]` to create the empty set or we could create a `Custom::empty() -> Self` function if we want to be more clear.

cranelift/assembler-x64/meta/src/dsl.rs
```diff
@@ -4,10 +4,13 @@
 //! - compact--i.e., define an x64 instruction on a single line, and
 //! - a close-to-direct mapping of what we read in the x64 reference manual.
 
+pub(crate) mod custom;
```
abrown:
```suggestion
mod custom;
```

No need to publicize this; we re-publish the types we want to be public in the `pub use custom::...` below.

cranelift/assembler-x64/meta/src/dsl.rs
```diff
@@ -4,10 +4,13 @@
...
+pub(crate) mod display;
```
abrown:
```suggestion
```

We don't want any implementation of the actual custom functions to live in this crate; we want them to live in the parent crate, `cranelift-assembler-x64`, in the `custom` module. So that `custom` module, which you modified to contain a `visit` sub-module, should also contain a `display` sub-module and eventually an `encode` module, once we need it.

cranelift/assembler-x64/meta/src/dsl.rs
```diff
@@ -64,7 +67,7 @@ pub struct Inst {
     pub has_trap: bool,
     /// Whether or not this instruction has a custom visit function for register
     /// allocation.
```
abrown:
```suggestion
    /// Whether or not this instruction uses custom, external functions
    /// instead of Rust code generated by this crate.
```

cranelift/assembler-x64/meta/src/dsl/custom.rs
```diff
@@ -0,0 +1,85 @@
...
+    fn from(flag: Option<CustomOperation>) -> Self {
+        Custom(flag.into_iter().collect())
+    }
+}
+
+impl From<Vec<CustomOperation>> for Custom {
+    fn from(operations: Vec<CustomOperation>) -> Self {
+        Custom(operations)
+    }
+}
```
abrown: These might not be needed ever, but probably aren't used now (?). Can we remove them?

cranelift/assembler-x64/meta/src/dsl/display.rs
```diff
@@ -0,0 +1,6 @@
+// Fix up the mnemonic for locked instructions: we want to print
+// "lock <inst>", not "lock_<inst>".
+pub fn lock(mnemonic: &String) -> String {
+    let inst_name = format!("lock {}", &mnemonic[5..]);
+    inst_name
+}
```
abrown: As mentioned above, this should move to `cranelift_assembler_x64::custom::display`. And, just like `visit`, we will (1) want to have a function for each instruction that needs the custom display logic and (2) pass through the whole instruction, in case we need more information in the future:

```rust
pub mod display {
  pub fn lock_addb(inst: &lock_addb) -> String { ... }
  pub fn lock_addw(inst: &lock_addw) -> String { ... }
  ...
  pub fn lock_subb(inst: &lock_subb) -> String { ... }
  pub fn lock_subw(inst: &lock_subw) -> String { ... }
  ...
}
```

Two thoughts about this:
- this could be quite repetitive so we might want to use a macro
- eventually we probably want the signature `(&<inst>) -> String` to be a bit more expansive, i.e., to return the entire display string and not just the altered mnemonic, but this is probably OK for now.

cranelift/assembler-x64/meta/src/generate/inst.rs
```diff
@@ -221,13 +222,18 @@ impl dsl::Inst {
...
-                        // "lock <inst>", not "lock_<inst>".
-                        let inst_name = if self.mnemonic.starts_with("lock_") {
-                            &format!("lock {}", &self.mnemonic[5..])
+                        let inst_name = if self.custom.contains(Display) {
+                            if self.name().starts_with("lock") {
+                                display::lock(&self.mnemonic)
+                            } else {
+                                unimplemented!(
+                                    "Only lock instructions custom processing implemented for now"
+                                );
```
abrown: This should look more like the logic with `visit` above: if we have a custom `display` function, then we emit something like:

```rust
fmtln!(f, "let inst_name = crate::custom::display::{}(self)", self.name());
```

And then, in this case, we would use `inst_name` later in the generated code instead of the already-known `inst_name` we have here.

cranelift/assembler-x64/src/custom.rs
```diff
@@ -1,51 +1,46 @@
...
-        &mut mulx.r32a,
-        &mut mulx.r32b,
-        &mut mulx.rm32,
-        &mut mulx.edx,
-        visitor,
-    )
-}
+pub mod visit {
+    use crate::inst::{mulxl_rvm, mulxq_rvm};
+    use crate::{Fixed, Gpr, GprMem, RegisterVisitor, Registers, gpr};
```
abrown, line 90:
```suggestion
    use crate::{Fixed, Gpr, GprMem, RegisterVisitor, Registers, gpr};

```

cranelift/assembler-x64/src/lib.rs
```diff
@@ -47,7 +47,7 @@
 )]
 
 mod api;
-mod custom;
+pub mod custom;
```
abrown:
```suggestion
mod custom;
```

?

abrown (commented):
I think the important bits are mostly here, most of the comments below are smaller Rust issues. The one real issue here is _where_ we put the logic... see below.

rahulchaphalkar:
Ready for review, addressed feedback

abrown (approved):
Thanks! This looks good. I'll put this in the merge queue but if @alexcrichton has any further suggestions we could follow this up in the `compare` PR.

alexcrichton (approved):

PR #10919, opened 2025-06-04
Cranelift: implement an "unwinder" crate and exception throws in filetests.

cfallin:
This commit introduces the next major piece of machinery (after the previously-landed `try_call` support) that we will eventually use to implement Wasm exceptions in Wasmtime. In particular, it implements a generic unwinder as a new crate that supports (i) walking a stack produced by Cranelift code, (ii) serializing Cranelift exception metadata to compact tables (in a way very similar to address maps in Wasmtime, so they will be mappable directly from disk), (iii) using these serialized tables to find handlers during a stack-walk, and (iv) jumping to handlers (i.e., actually unwinding). This crate is currently used in the filetests runner, and will next be used in Wasmtime.

The commit first performs code-motion: it moves stack-walking code from Wasmtime to `cranelift-unwinder`. This itself has no functional effect, but isolates the code that understands contiguous sequences of Cranelift frames ("activations") from that which is specific to Wasmtime's activation delimiters and metadata.

It then implements a compact exception-table format. This format uses the `object` crate's mechanisms for directly referencing in-memory arrays of little-endian `u32`s in a way that will allow us to find handlers when mapping exception metadata directly from an ELF section in a `.cwasm` (for example). The format consists of four sorted `u32` arrays in a way that allows us to look up a callsite first, then search its sorted array of handler offsets by tags.

It next implements the actual unwind control flow: it contains an assembly stub for each supported architecture that transfers control to a PC, SP, and FP value "up the stack", with payload values placed in the payload registers we have defined per our exception ABI in Cranelift.

Finally, it puts these pieces together in the filetest runner. Note that the runtest does a lot "by hand": we don't have entry and exit trampolines as we do in Wasmtime, so the filetest contains three functions, with the middle one invoking the "throw hostcall" and entry and exit trampolines around it grabbing the appropriate entry/exit FPs and exit PC. The dance to call back to host code is also somewhat delicate, as we haven't done this before. The `JITModule`'s linking + relocation support does not seem sufficient to properly define a symbol, so instead we scan for `func_addr` instructions referencing a well-known name (`__cranelift_throw`) and replace them with `iconst`s with the function address at runtime, baking it in. This is somewhat ugly, but it works. All of these filetest-specific details will be handled much more nicely in the Wasmtime version of this functionality, as we have proper abstractions for entry/exit trampolines and hostcalls.


cranelift/filetests/filetests/runtests/throw.clif
```diff
@@ -0,0 +1,50 @@
...
+function %throw(i64, i32, i64, i64) tail {
+  sig0 = (i64, i64, i64, i32, i64, i64)
+  fn0 = %__cranelift_throw(i64, i64, i64, i32, i64, i64)
+
+block0(v0: i64, v1: i32, v2: i64, v3: i64):
+  v4 = get_frame_pointer.i64
+  v5 = get_return_address.i64
+  v6 = load.i64 v5 ; get caller's FP
+  v7 = func_addr.i64 fn0
+  call_indirect sig0, v7(v0, v4, v6, v1, v2, v3)
```
bjorn3, line 48: Making this a `return_call_indirect` would avoid the need to get any frame pointer, right? You only need to change the return address before returning from `__cranelift_throw`, not unwind the stack back to the caller of `%throw`. It would require a landingpad for every call even if it directly unwinds to the caller, but that is already necessary for the current implementation in this PR anyway.

cranelift/filetests/filetests/runtests/throw.clif
```diff
@@ -0,0 +1,50 @@
...
+  return
+}
```
bjorn3: nit: missing trailing newline

cranelift/filetests/src/function_runner.rs
```diff
@@ -563,6 +623,47 @@ fn make_trampoline(name: UserFuncName, signature: &ir::Signature, isa: &dyn Targ
     func
 }
 
+/// Hostcall invoked directly from a compiled function body to test
+/// exception throws.
+extern "C" fn __cranelift_throw(
```
bjorn3: `extern "C-unwind"` otherwise unwinding out of this is UB (you are skipping past the aborting landingpad due to using a different unwinding mechanism).

bjorn3:
By the way why is the unwind info format defined in the cranelift-unwinder crate? It is Wasmtime specific and I would assume that Winch and Pulley will reuse the same format.

cranelift/unwinder/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+authors = ["The Cranelift Project Developers"]
+name = "cranelift-unwinder"
+version = "0.121.0"
+description = "Unwinder that works with Cranelift's exception support"
```
fitzgen: I don't care too much what we name this crate, in some sense it could be either `wasmtime-unwinder` or `cranelift-unwinder` because the dependency graph is a little convoluted here, but I think this description is a bit misleading. People who are writing a random Cranelift project could see this and think it is a solution to their unwinding issues -- especially since using Cranelift means pulling in a bunch of `cranelift-foo` deps -- and *mayyyybe* it could be, but it is really tailored for Wasmtime's usage and isn't going to be very flexible to other use cases.

We should probably make it clear in this description that this is an internal implementation detail of Wasmtime and Cranelift's unit testing.

cranelift/unwinder/src/arch/mod.rs
```diff
@@ -34,11 +34,26 @@ pub fn get_stack_pointer() -> usize {
     imp::get_stack_pointer()
 }
 
+/// Resume execution at the given PC, SP, and FP, with the given
+/// payload values, according to the tail-call ABI's exception
+/// scheme. Note that this scheme does not restore any other
+/// registers, so the given state is all that we need.
+pub unsafe fn resume_to_exception_handler(
```
fitzgen: Maybe add an olive branch of documentation about the safety requirements here? We've been trying to improve this stuff in the core VM.

cranelift/unwinder/src/stackwalk.rs
```diff
@@ -0,0 +1,230 @@
...
+    }
+    fn assert_fp_is_aligned(&self, fp: usize) {
+        let expected = if cfg!(target_pointer_width = "32") {
+            8
+        } else {
+            16
+        };
+        assert_eq!(fp % expected, 0, "stack should always be aligned");
+    }
+}
```
alexcrichton: Not necessarily required for this PR, but could this be moved back to Wasmtime? It feels a bit odd to define it all the way over here when Pulley is defined somewhere else.

Placement-wise the only use of `UnwindPulley` is [here](https://github.com/bytecodealliance/wasmtime/blob/5603ee7b099c837aa7194f1052e7c73cf4c36750/crates/wasmtime/src/runtime/store.rs#L1854) so it might make sense to add an `unwinder` method on `struct Interpreter` which would enable moving all of this to [this file](https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasmtime/src/runtime/vm/interpreter.rs) (and a stub in [this file](https://github.com/bytecodealliance/wasmtime/blob/5603ee7b099c837aa7194f1052e7c73cf4c36750/crates/wasmtime/src/runtime/vm/interpreter_disabled.rs))

cranelift/unwinder/build.rs
```diff
@@ -0,0 +1,19 @@
+fn main() {
+    println!("cargo:rerun-if-changed=build.rs");
+
+    // Determine if the current host architecture is supported by Cranelift
+    // meaning that we might be executing native code.
+    let has_host_compiler_backend = match std::env::var("CARGO_CFG_TARGET_ARCH").unwrap().as_str() {
+        "x86_64" | "riscv64" | "s390x" | "aarch64" => true,
+        _ => false,
+    };
```
alexcrichton: Ideally I'd prefer to not duplicate this with the main Wasmtime crate as I suspect the two will diverge over time. For example on iOS we should probably disable cranelift by default despite having a backend for aarch64. Now that doesn't actually matter for this crate though since it'll still all compile just fine on iOS (in theory), it just wouldn't be useful to have at runtime.

Would it be possible to actually drop this build script entirely? For example the `arch` module would become naturally empty on unsupported architectures and structures like `UnwindHost` would only appear for supported architectures. That'd mean that this build script would basically fall out of the `cfg-if` if/else branches for the arch support

cranelift/filetests/src/function_runner.rs
```diff
@@ -563,6 +623,47 @@ fn make_trampoline(name: UserFuncName, signature: &ir::Signature, isa: &dyn Targ
...
+    entry_fp: usize,
+    exit_fp: usize,
+    exit_pc: usize,
+    tag: u32,
+    payload1: usize,
+    payload2: usize,
+) -> ! {
+    let compiled_test_file = unsafe {
+        std::mem::transmute::<*const CompiledTestFile, &CompiledTestFile>(COMPILED_TEST_FILE.get())
```
alexcrichton: FWIW you can use `&*` to translate `*const T` to `&T`, no `transmute` necessary

cranelift/filetests/src/function_runner.rs
alexcrichton: Could you elaborate on this? I was under the impression that `C-unwind` had to do with the system unwinder, or whatever the default is for the platform. For this unwinder landing pads aren't invoked no matter what, so by the same logic `C-unwind` is also UB because landing pads still aren't executed. 

To me it seems that the soundness of this function relies on there not actually being any landing pads, which is something we'd have to uphold?

cranelift/filetests/src/function_runner.rs
```diff
@@ -563,6 +623,47 @@ fn make_trampoline(name: UserFuncName, signature: &ir::Signature, isa: &dyn Targ
...
+            &mut unwind_host,
+            module_lookup,
+            exit_pc,
+            exit_fp,
+            entry_fp,
+            tag,
+        ) {
+            cranelift_unwinder::ThrowAction::Handler { pc, sp, fp } => {
+                cranelift_unwinder::resume_to_exception_handler(pc, sp, fp, payload1, payload2);
+                // Doesn't return.
```
alexcrichton: It should be ok to skip this comment since the type system basically all guarantees that for us (or perhaps move this beforehand as a lead-in explanation for what the function's doing?)

cranelift/unwinder/src/exception_table.rs
```diff
@@ -0,0 +1,279 @@
...
+///
+/// This corresponds to:
+/// - A callsite return address at `0x10` with:
+///  - A handler for exception tag `1` at offset `0x40`
+///  - A handler for exception tag `5` at offset `0x42`
+/// - A callsite return address at `0x50` with:
+///   - A handler for exception tag `1` at offset `0x6f`
+///   - A handler for any tag (sentinel `-1`) -- i.e., a catch-all -- at offset `0x71`
+/// - A callsite return address at `0xf0` with:
+///   - A handler for any tag at offset `0xf5`
```
fitzgen: I had to read this pretty closely to understand how the below "expanded" representation corresponds to the above encoding example. I think what would have allowed me to understand at a relatively light skim would have been inline comments in the `ranges` array expanding what actual `tags`/`handlers` ranges they represent, as well as having links the other way for the `tags` and `handlers`, something like this:

```
callsites: [0x10, 0x50, 0xf0],
ranges: [
  // tag/handlers range for callsite 0x10: 0..2
  2,
  // tag/handlers range for callsite 0x50: 2..4
  4,
  // tag/handlers range for callsite 0xf0: 4..5
  5,
]
tags: [
  // Tags for callsite 0x10
  1,
  5,
  // Tags for callsite 0x50
  1,
  -1,
  // Tags for callsite 0xf0
  -1
],
handlers: [
  // Handlers for callsite 0x10
  0x40,
  0x42,
  // Handlers for callsite 0x50
  0x6f,
  0x71,
  // Handlers for callsite 0xf0
  0xf5,
]
```

Basically, it wasn't immediately clear to me at a glance which arrays were part of the "same struct" (in a struct-of-arrays vs array-of-structs isomorphism) and for the ones which had separate indexing, how that indexing worked, and I ended up having to read this comment closely a couple times.

cranelift/unwinder/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+readme = "README.md"
+keywords = ["compile", "compiler", "jit"]
+edition.workspace = true
+rust-version.workspace = true
+
+[lints]
+workspace = true
+
+[dependencies]
+cranelift-codegen = { workspace = true, optional = true }
```
alexcrichton: I'm hesitating with this in that this feels like a new significant dependency edge that didn't exist prior, but I can't really articulate what exactly it is. Currently in all other locations though `cranelift-codegen` is isolated to `wasmtime-cranelift`/Winch and nowhere else. My assumption is that the `wasmtime` crate will depend on `cranelift-unwinder` (and maybe also the `wasmtime-cranelift` crate to pass in the cranelift data structures?). I realize that `wasmtime` itself won't enable this dependency, but WDYT about making this built-in to cranelift-codegen directly? Or maybe a separate crate?

I don't feel too too strong about this, but I think it sort of depends on how exactly this'll get integrated in to Wasmtime as well.

cranelift/unwinder/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+log = { workspace = true }
+cfg-if = { workspace = true }
+object = { workspace = true }
+anyhow = { workspace = true }
+
+[target.'cfg(target_arch = "s390x")'.dependencies]
+psm = { workspace = true }
```
alexcrichton: FWIW if you're feeling enterprising IIRC the only reason we have this is `get_stack_pointer` on s390x where inline assembly wasn't supported originally, but it is supported now, so a small inline assembly snippet would probably suffice for removing this crate dependency.

cranelift/filetests/src/function_runner.rs
alexcrichton: I've written up more thoughts on this in https://github.com/bytecodealliance/wasmtime/issues/10923

cranelift/unwinder/src/stackwalk.rs
```diff
@@ -0,0 +1,230 @@
...
+///
+/// We require that the initial PC, FP, and `trampoline_fp` values are
+/// non-null (non-zero).
+pub unsafe fn visit_frames<R>(
+    unwind: &dyn Unwind,
+    mut pc: usize,
+    mut fp: usize,
+    trampoline_fp: usize,
+    mut f: impl FnMut(Frame) -> ControlFlow<R>,
+) -> ControlFlow<R> {
```
fitzgen: If this function is already monomorphized over `R`, we might as well use an `&impl Unwind` instead of a `&dyn Unwind`

cranelift/filetests/filetests/runtests/throw.clif
cfallin: Fixed, thanks!

cranelift/filetests/src/function_runner.rs
cfallin: Updated to use `C-unwind` -- thanks for catching this.

cranelift/unwinder/build.rs
cfallin: Done as suggested in Cranelift meeting today -- there are now `pub use`s to re-export with doc-comments attached. `cfg`s minimized and put right next to each other -- only the one `cfg_if` chain for the implementations and one with `cfg(any(...))` to wrap all the `pub use`s and the `UnwindHost` impl.

I've verified that `wasmtime-unwinder` checks on `i686-unknown-linux-gnu`.

cranelift/unwinder/Cargo.toml
cfallin: Yep, renamed to `wasmtime-unwinder` -- thanks all for the feedback on this!

cranelift/unwinder/src/stackwalk.rs
cfallin: Interestingly, this ends up propagating all the way up to e.g. `store.unwinder()` -- the `dyn` is infectious and we'd have a monomorphize a bunch more to make this work (or build a proxy-to-dyn impl to monomorphize on). Since this seems like just an opportunistic thing, I'll leave it as `dyn Unwind` for now...

cranelift/unwinder/src/arch/mod.rs
cfallin: Done!

cranelift/jit/Cargo.toml
```diff
@@ -19,6 +19,7 @@ cranelift-native = { workspace = true }
 cranelift-codegen = { workspace = true, features = ["std"] }
 cranelift-entity = { workspace = true }
 cranelift-control = { workspace = true }
+wasmtime-unwinder = { workspace = true, features = ["cranelift"] }
```
bjorn3: Please make this an optional dependency. I don't want to pull this in for cg_clif.

cranelift/jit/src/backend.rs
```diff
@@ -358,10 +362,49 @@ impl JITModule {
...
+            code_ranges: Vec::new(),
             functions_to_finalize: Vec::new(),
             data_objects_to_finalize: Vec::new(),
         }
     }
+
+    /// Look up the ExceptionTable and corresponding base PC, if any,
+    /// for a given PC that may be within one of the CompiledBlobs in
+    /// this module.
+    pub fn lookup_exception_data<'a>(
```
bjorn3:
```suggestion
    pub fn lookup_wasmtime_exception_data<'a>(
```

cranelift/filetests/filetests/runtests/throw.clif
cfallin: Perhaps, but one of my goals here was to closely mirror how the trampolines work in Wasmtime, and also I don't know if we really need to optimize the testing infrastructure like this.

cranelift/unwinder/src/stackwalk.rs
cfallin: Done!

cranelift/unwinder/src/stackwalk.rs
alexcrichton: Yeah changing this away from `dyn Unwind`, while possible, I think is best to defer to later.

cranelift/jit/Cargo.toml
cfallin: Sorry about that, didn't think through all the use-cases -- done!

cranelift/jit/src/backend.rs
cfallin: Done!

cranelift/unwinder/Cargo.toml
cfallin: Done!

cranelift/filetests/src/function_runner.rs
cfallin: Done!

cranelift/filetests/src/function_runner.rs
cfallin: Ah, right, much simpler -- done!

cranelift/unwinder/Cargo.toml
cfallin: Now that the crate is called `wasmtime-unwinder` I hesitate to put it as a dependency (even optional) of `cranelift-codegen`; I suppose in the fullness of time we could have a `wasmtime-unwinder-cranelift` to split off the translation-from-Cranelift-metadata functionality but I also don't feel too strongly about it. For now at least it's under a feature and optional. Happy to see this refactored later if we want to go that way or have other ideas.

cranelift/unwinder/src/exception_table.rs
cfallin: Updated, thanks -- I like this way of writing it out a lot better.

cfallin:
This appears to have bounced off the merge queue because it raced with #10388 and lost; and that PR heavily modified backtracing as well...

alexcrichton:
github's not great about rendering all the time, but the failure was [here](https://github.com/bytecodealliance/wasmtime/actions/runs/15455099518/job/43505743615#step:19:313) (32-bit build)

cfallin:
Grr, now there's another merge conflict because the version bump went in ahead of this one.

cfallin:
Wow, CI *really* does not like this PR -- network error [here](https://github.com/bytecodealliance/wasmtime/actions/runs/15456051175/job/43508497144) downloading CMake.

PR #10924, opened 2025-06-04
Add intial porting of table_ops from arbitrary to mutatis

khagankhan:
This is the initial porting of `table_ops` from `arbitrary` to `mutatis`.

No new features are added. This is a direct port of the existing implementation using the `arbitrary` crate to the `mutatis` framework.

Open to review and feedback.

Initial work towards: bytecodealliance/wasmtime#10327  
cc @ospencer @fitzgen

### DIFF
--- a/crates/fuzzing/src/generators/table_ops.rs
+++ b/crates/fuzzing/src/generators/table_ops.rs
@@ -1,6 +1,7 @@
 //! Generating series of `table.get` and `table.set` operations.
 
-use arbitrary::{Arbitrary, Result, Unstructured};
+use mutatis::{Generate, DefaultMutate, Candidates, Mutate, Context, Result as MutResult};
+use mutatis::mutators as m;
 use std::ops::RangeInclusive;
 use wasm_encoder::{
     CodeSection, ConstExpr, EntityType, ExportKind, ExportSection, Function, FunctionSection,
@@ -137,32 +138,75 @@ impl TableOps {
     }
 }
 
-impl<'a> Arbitrary<'a> for TableOps {
-    fn arbitrary(u: &mut Unstructured<'a>) -> Result<Self> {
-        let mut result = TableOps {
-            num_params: u.int_in_range(NUM_PARAMS_RANGE)?,
-            num_globals: u.int_in_range(NUM_GLOBALS_RANGE)?,
-            table_size: u.int_in_range(TABLE_SIZE_RANGE)?,
-            ops: Vec::new(),
+#[derive(Debug)]
+pub struct TableOpsMutator;
+
+impl Mutate<TableOps> for TableOpsMutator {
+    fn mutate(&mut self, c: &mut Candidates<'_>, ops: &mut TableOps) -> mutatis::Result<()> {
+        c.mutation(|ctx| {
+            let mut stack = 0;
+            add_table_op_mutatis(ops, ctx, &mut stack)
+        })?;
+        Ok(())
+    }
+}
+
+impl DefaultMutate for TableOps {
+    type DefaultMutate = TableOpsMutator;
+}
+
+impl Default for TableOpsMutator {
+    fn default() -> Self {
+        TableOpsMutator
+    }
+}
+
+impl Generate<TableOps> for TableOpsMutator {
+    fn generate(&mut self, ctx: &mut Context) -> MutResult<TableOps> {
+        let num_params = m::range(NUM_PARAMS_RANGE).generate(ctx)?;
+        let num_globals = m::range(NUM_GLOBALS_RANGE).generate(ctx)?;
+        let table_size = m::range(TABLE_SIZE_RANGE).generate(ctx)?;
+        let mut ops = Vec::new();
+        let mut stack = 0u32;
+        let mut temp_ops = TableOps {
+            num_params,
+            num_globals,
+            table_size,
+            ops: vec![
+                TableOp::Null,
+                TableOp::Drop,
+                TableOp::Gc,
+                TableOp::LocalSet(0),
+                TableOp::LocalGet(0),
+                TableOp::GlobalSet(0),
+                TableOp::GlobalGet(0),
+            ],
         };
-
-        let mut stack = 0;
-        let mut choices = vec![];
-        while result.ops.len() < MAX_OPS && !u.is_empty() {
-            add_table_op(&mut result, u, &mut stack, &mut choices)?;
+        while ops.len() < MAX_OPS {
+            temp_ops.ops = ops.clone(); 
+            let add_result = add_table_op_mutatis(&mut temp_ops, ctx, &mut stack);
+            if let Ok(()) = add_result {
+                if let Some(last) = temp_ops.ops.last() {
+                    ops.push(*last);
+                }
+            } else {
+                break;
+            }
         }
-
-        // Drop any extant refs on the stack.
         for _ in 0..stack {
-            result.ops.push(TableOp::Drop);
+            ops.push(TableOp::Drop);
         }
-
-        Ok(result)
+        Ok(TableOps {
+            num_params,
+            num_globals,
+            table_size,
+            ops,
+        })
     }
 }
 
 macro_rules! define_table_ops {
-	(
+    (
         $(
             $op:ident $( ( $($limit:expr => $ty:ty),* ) )? : $params:expr => $results:expr ,
         )*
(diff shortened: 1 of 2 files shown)


crates/fuzzing/src/generators/table_ops.rs
```diff
@@ -135,70 +134,299 @@ impl TableOps {
...
+    pub fn abstract_stack_depth(&self) -> usize {
+        let mut stack: usize = 0;
+        for op in &self.ops {
+            let pop = op.operands_len();
+            let push = op.results_len();
+            stack = stack.saturating_sub(pop);
+            stack += push;
+        }
+        stack
+    }
```
fitzgen: I think we won't need this helper anymore, given the `fixup` helper function I suggested in private communication.

crates/fuzzing/src/generators/table_ops.rs
```diff
@@ -135,70 +134,299 @@ impl TableOps {
...
+#[derive(Debug)]
+pub struct TableOpsMutator;
+
+impl Mutate<TableOps> for TableOpsMutator {
+    fn mutate(&mut self, c: &mut Candidates<'_>, ops: &mut TableOps) -> mutatis::Result<()> {
+        //  1. Append
+        c.mutation(|ctx| {
+            let mut stack = ops.abstract_stack_depth();
+            add_table_op_mutatis(ops, ctx, &mut stack)
+        })?;
```
fitzgen: And this can become an arbitrary `insert` instead of `append` (which is a subset of an `insert` operation).

crates/fuzzing/src/generators/table_ops.rs
```diff
@@ -135,70 +134,299 @@ impl TableOps {
...
+                        idx,
+                        removed
+                    );
+                    stack += 1;
+                }
+            }
+            Ok(())
+        })?;
+        // 3. Replace
+        c.mutation(|ctx| {
```
fitzgen: Similarly, I think this function body will get cleaned up a ton with the `fixup` pass. Additionally, whenever possible we don't want to report mutations that we know will fail so if the ops are empty we shouldn't call `c.mutation(...)`. Putting these things together, I expect this will become something like:

```rust
if !ops.ops.is_empty() {
    c.mutation(|ctx| {
        let i = ctx.rng().gen_index(ops.ops.len()).unwrap();
        ops.ops[i] = choose_random_op();
        Ok(())
    })?;
}
```

crates/fuzzing/src/generators/table_ops.rs
```diff
@@ -135,70 +134,299 @@ impl TableOps {
...
+                    log::debug!(
+                        "Patched with ref.null at idx {} after removing {:?}",
+                        idx,
+                        removed
+                    );
+                    stack += 1;
+                }
+            }
+            Ok(())
+        })?;
```
fitzgen: And this can become something like

```rust
if !ops.ops.is_empty() {
    c.mutation(|ctx| {
        let i = ctx.rng().gen_index(ops.ops.len()).unwrap();
        ops.ops.remove(i);
        Ok(())
    })?;
}
```

crates/fuzzing/src/generators/table_ops.rs
fitzgen: And the `insert` mutation would be something like

```rust
c.mutation(|ctx| {
    let i = ctx.rng().gen_index(ops.ops.len() + 1).unwrap();
    ops.ops.insert(random_op());
    Ok(())
})?;
```

crates/fuzzing/src/generators/table_ops.rs
```diff
@@ -135,70 +134,299 @@ impl TableOps {
...
+                TableOp::Gc(),
+                TableOp::LocalSet(0),
+                TableOp::LocalGet(0),
+                TableOp::GlobalSet(0),
+                TableOp::GlobalGet(0),
+            ],
+        };
+        while ops.len() < MAX_OPS {
+            temp_ops.ops = ops.clone();
+            let mut stack = temp_ops.abstract_stack_depth();
```
fitzgen: This is accidentally `O(n*n)` because the helper walks every op in every iteration of this loop. You'll want to have a local `stack` variable that is adjusted as ops are generated here.

crates/fuzzing/src/generators/table_ops.rs
```diff
@@ -135,70 +134,299 @@ impl TableOps {
...
+                TableOp::Null(),
+                TableOp::Drop(),
+                TableOp::Gc(),
+                TableOp::LocalSet(0),
+                TableOp::LocalGet(0),
+                TableOp::GlobalSet(0),
+                TableOp::GlobalGet(0),
+            ],
+        };
+        while ops.len() < MAX_OPS {
```
fitzgen: And rather than generating the max every time, it probably makes sense to do something like

```rust
for _ in ctx.rng().gen_index(MAX_OPS).unwrap() {
   // ...
}
```

crates/fuzzing/src/generators/table_ops.rs
```diff
@@ -135,70 +134,299 @@ impl TableOps {
...
+            ops: vec![
+                TableOp::Null(),
+                TableOp::Drop(),
+                TableOp::Gc(),
+                TableOp::LocalSet(0),
+                TableOp::LocalGet(0),
+                TableOp::GlobalSet(0),
+                TableOp::GlobalGet(0),
+            ],
+        };
```
fitzgen: And this can be `TableOps::default()` where the default is an empty set of ops, since we will generate new ops just below. There is no need to start with this prologue every time.

crates/fuzzing/src/generators/table_ops.rs
```diff
@@ -135,70 +134,299 @@ impl TableOps {
...
 
-        #[expect(unused_comparisons, reason = "macro-generated code")]
-        fn add_table_op(
-            ops: &mut TableOps,
-            u: &mut Unstructured,
-            stack: &mut u32,
-            choices: &mut Vec<fn(&TableOps, &mut Unstructured, &mut u32) -> Result<TableOp>>,
-        ) -> Result<()> {
-            choices.clear();
+            pub fn operands_len(&self) -> usize {
```
fitzgen, line 309: Don't think this needs to be `pub` because I don't think it is used outside this module.

crates/fuzzing/src/generators/table_ops.rs
```diff
@@ -135,70 +134,299 @@ impl TableOps {
...
+            }
 
-            // Add all the choices of valid `TableOp`s we could generate.
-            $(
-                if $( $(($limit as fn(&TableOps) -> $ty)(&*ops) > 0 &&)* )? *stack >= $params {
-                    choices.push(|_ops, _u, stack| {
-                        *stack = *stack - $params + $results;
-                        Ok(TableOp::$op $( ( $(_u.int_in_range(0..=($limit as fn(&TableOps) -> $ty)(_ops) - 1)?),* ) )? )
-                    });
+            pub fn results_len(&self) -> usize {
```
fitzgen, line 317: Ditto.

crates/fuzzing/src/generators/table_ops.rs
```diff
@@ -273,100 +501,109 @@ impl TableOp {
...
+    macro_rules! empty_table_ops {
+        ($num_params:expr, $num_globals:expr, $table_size:expr) => {
+            TableOps {
+                num_params: $num_params,
+                num_globals: $num_globals,
+                table_size: $table_size,
+                ops: vec![],
             }
-        }
+        };
```
fitzgen: This macro should just be `TableOps::default()`, which can simply be derived on the type rather than written out.

crates/fuzzing/src/generators/table_ops.rs
```diff
@@ -1,13 +1,12 @@
...
 use std::ops::RangeInclusive;
 use wasm_encoder::{
     CodeSection, ConstExpr, EntityType, ExportKind, ExportSection, Function, FunctionSection,
     GlobalSection, ImportSection, Instruction, Module, RefType, TableSection, TableType,
     TypeSection, ValType,
 };
-
 /// A description of a Wasm module that makes a series of `externref` table
 /// operations.
 #[derive(Debug)]
```
fitzgen: Here we would add `Default` next to `Debug`.

crates/fuzzing/src/generators/table_ops.rs
```diff
@@ -273,100 +501,109 @@ impl TableOp {
...
+                    TableOp::Drop(),
+                    TableOp::Gc(),
+                    TableOp::LocalSet(0),
+                    TableOp::LocalGet(0),
+                    TableOp::GlobalSet(0),
+                    TableOp::GlobalGet(0),
+                ],
+            }
+        };
+    }
```
fitzgen: I don't think we should need this, and can simply use an empty `TableOps::default()` instead, but if we did want to create this particular sequence in a bunch of places we should just use a function rather than a macro.

crates/fuzzing/src/generators/table_ops.rs
```diff
@@ -273,100 +501,109 @@ impl TableOp {
...
-        eprintln!("actual WAT = {actual}");
+    #[test]
+    fn test_tableops_mutate_with() -> mutatis::Result<()> {
+        let _ = env_logger::try_init();
+        let mut res = default_table_ops![5, 5, 5];
+        let mut generator = TableOpsMutator;
+        let mut session = mutatis::Session::new();
+
+        for _ in 0..=1024 {
+            session.mutate_with(&mut generator, &mut res)?;
```
fitzgen: Because the `TableOpsMutator` is the default, we don't actually need this test; it is redundant with `mutate_table_ops_with_default_mutator`.

crates/fuzzing/src/generators/table_ops.rs
```diff
@@ -273,100 +501,109 @@ impl TableOp {
...
+                unseen_ops.remove(op.name());
+                if unseen_ops.is_empty() {
+                    break 'outer;
+                }
+            }
+        }
+        assert!(unseen_ops.is_empty(), "Failed to generate {unseen_ops:?}");
+        Ok(())
+    }
+    // TODO: Write a test for expected wat
```
fitzgen: Reminder for this TODO.

crates/fuzzing/src/generators/table_ops.rs
khagankhan: Yes! I am working on that and a `fixup` method results in much cleaner code.

crates/fuzzing/src/generators/table_ops.rs
khagankhan: Right!!

khagankhan:
Thanks, @fitzgen! currently working on it. Will integrate your suggestions in the next commit.

crates/fuzzing/src/generators/table_ops.rs
khagankhan: Removed.

crates/fuzzing/src/generators/table_ops.rs
khagankhan: Added

crates/fuzzing/src/generators/table_ops.rs
khagankhan: added new methods instead of macros

crates/fuzzing/src/generators/table_ops.rs
khagankhan: Done

crates/fuzzing/src/generators/table_ops.rs
khagankhan: Added new empty() method

PR #10928, opened 2025-06-05
Add pcmp variants to the new assembler

jlb6740:


cranelift/filetests/filetests/isa/x64/sqmul_round_sat.clif
jlb6740, line 1: This filetest update bothers me as the original result that includes "const(0)" is not valid syntax so I am not sure I can directly map this previous result to the current change and assume it is correct.

cranelift/filetests/filetests/isa/x64/sqmul_round_sat.clif
```diff
@@ -13,7 +13,7 @@ block0(v0: i16x8, v1: i16x8):
 ; block0:
 ;   pmulhrsw %xmm1, %xmm0
 ;   movdqa %xmm0, %xmm5
-;   pcmpeqw %xmm5, const(0), %xmm5
```
jlb6740, line 16: @abrown This filetest update bothers me since the original result that includes "const(0)" is not x86 syntax. Because it is not x86 syntax I am not sure how I can directly map this previous result to the current change to assume the blessed update is correct.

cranelift/filetests/filetests/isa/x64/sqmul_round_sat.clif
```diff
@@ -13,7 +13,7 @@ block0(v0: i16x8, v1: i16x8):
...
+;   pcmpeqw (%rip), %xmm5
```
jlb6740, line 16: @abrown This filetest update bothers me since the original result includes code (const(0)) that is not valid x86 syntax. Because of this, I am not sure how I can directly map this previous result to the current update to assume this blessed update is correct.

cfallin: The previous syntax was representing a relocation that we perform after emitting, to refer to the constant pool. I don't know that we have to use the same syntax, but perhaps we could come up with a way to print the same information around the RIP-relative amode rather than just `(%rip)`?

jlb6740: @cfallin, I see and thanks for the education. I looked into this and talked with @abrown and it looks like we don't have information at the time of display that would alert us to tweak the output here to read `const(0)` or anything else we choose. Perhaps we could figure out how to propagate that along but I am assuming the benefit would just be for highlighting a relocation reference during some stage of debugging. Do you think that is needed as a future todo? In any case for me, now that you've explained, it seems that task is unrelated to the correctness of pcmp* here, but are you also comfortable that I've bless the filetest solution update here in this pr?

cfallin: Sure, this should be well-covered by runtests so if it's green, it's good to ship. I do suspect one could do something with traits here -- we already need to carry the amode with relocations somewhere, so perhaps we could have a "pretty-print" hook that Cranelift plugs into the assembler. No need to do that in this PR though, I don't think.

abrown (approved):

Issue #10933, opened 2025-06-05
Technically `&mut Instance` is not safe


alexcrichton:
Currently in Wasmtime it's safe to acquire `&mut wasmtime::runtime::vm::Instance` in various locations and this is used pervasively. Technically though this alone is not sound as it allows `mem::swap`-ing two instances. This would mean that the offsets, used to calculate the size of the allocation used for deallocation, could get mixed up (among other issues). Ideally we would have an abstraction which allows mutable access but disallows usage of `mem::swap`. In a sense this is similar to `StoreContextMut` which basically entirely exists to prevent `mem::swap`-ing two stores. This approach has a lot of ergonomic downsides though and it's not fun to write against.

Another alternative is to assume that this basically doesn't happen (it's an internal type in Wasmtime, and it's pretty unlikely we reach for doing this naturally). That would mean we'd document on some `unsafe` block somewhere "this requires auditing the entire codebase" which practically wouldn't happen but would in-practice work alright.

Another possible alternative would be something like:

```rust
struct InstanceFields { /* ... */ } // what `struct Instance` is today

pub struct Instance {
    fields: InstanceFields,
    _something_forcing_mut_ref_to_be_dst: [u8],
}
```

We'd keep methods and such on `impl Instance { ... }` and `Instance` would itself be a dynamically sized type which would disallow overwriting the entire structure via the type system. Runtime-wise this would have a slight cost as passing it by reference passes two values and one value is always zero (the `_something_forcing_mut_ref_to_be_dst` would always be zero-length, we wouldn't want it to encompass the `VMContext` as that would inaccurately model how `&Instance` actually has mutable provenance to `VMContext`, but if `[u8]` covered the whole field then it would also seemingly be only shared provenance). LLVM might be able to figure that out though with enough inlining?

Perosnally I'm tempted to pursue this last solution of a dynamically-sized type. It would still require auditing the entire `vm/instance.rs` module (as you could still overwrite `InstanceFields` in there) but that's a much less daunting task than the entire codebase.

alexcrichton:
I'll also note that after https://github.com/bytecodealliance/wasmtime/pull/10934 this issue is equally applicable to components as well.

bjorn3:
So basically Instance needs to be pinned?

alexcrichton:
[...]

PR #10934, opened 2025-06-05
Start reducing unsafety of `ComponentInstance`

alexcrichton:
Work recently in the wasip3-prototyping repository has focused on reducing the amount of `unsafe` code and improving internal abstractions to facilitate this. One major thrust that's manifested in is the removal of the usage of `*mut ComponentInstance` where possible and instead using an index to safely borrow from the store to get the entire instance. This commit does not fully realize this vision just yet but lays some groundwork leading up to this.

This commit specifically removes the `*mut ComponentInstance` pointers in lift/lower contexts in favor of directly storing a `wasmtime::component::Instance`. When the raw `ComponentInstance` is needed it's acquired from the `StoreOpaque` in a safe fashion that borrows the entire store for the duration of the returned borrow. This in turn required pushing instances into the store earlier during instantiation because during instantiation an instance could call out to host APIs which do lifts/lowers.

There's still more work to be done to plumb this fully into libcalls and host function invocations but I wanted to upstream the wasip3 work piecemeal a chunk at a time.

### DIFF
--- a/crates/wasmtime/src/runtime/vm/component.rs
+++ b/crates/wasmtime/src/runtime/vm/component.rs
@@ -637,6 +637,11 @@ impl ComponentInstance {
         &self.resource_types
     }
 
+    /// Returns mutable a reference to the resource type information.
+    pub fn resource_types_mut(&mut self) -> &mut Arc<PrimaryMap<ResourceIndex, ResourceType>> {
+        &mut self.resource_types
+    }
+
     /// Returns whether the resource that `ty` points to is owned by the
     /// instance that `ty` correspond to.
     ///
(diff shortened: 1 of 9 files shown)


crates/wasmtime/src/runtime/vm/component.rs
```diff
@@ -976,98 +981,10 @@ impl OwnedComponentInstance {
         }
     }
 
-    // Note that this is technically unsafe due to the fact that it enables
-    // `mem::swap`-ing two component instances which would get all the offsets
-    // mixed up and cause issues. This is scoped to just this module though as a
-    // convenience to forward to `&mut` methods on `ComponentInstance`.
-    unsafe fn instance_mut(&mut self) -> &mut ComponentInstance {
-        &mut *self.ptr.as_ptr()
-    }
```
alexcrichton, line 985: I'll note that this comment is still equally applicable today, but I don't think it's worth it trying to keep this distinction. It's just too damn useful to be able to get a safe `&mut ComponentInstance` and too subtle to avoid trying to do so. I opened https://github.com/bytecodealliance/wasmtime/issues/10933 to track a better solution here.

fitzgen (approved):
Nice

PR #10935, opened 2025-06-05
Remove `unsafe` from `Instance::from_wasmtime`

alexcrichton:
I mistakenly thought that this would be an `unsafe` function due to reasons that aren't applicable, and I'm not aware of any reason why this function should be `unsafe`, so mark it safe.

### DIFF
--- a/crates/wasmtime/src/runtime/component/instance.rs
+++ b/crates/wasmtime/src/runtime/component/instance.rs
@@ -55,7 +55,7 @@ impl Instance {
     ///
     /// The safety of this function relies on `id` belonging to the instance
     /// within `store`.
-    pub(crate) unsafe fn from_wasmtime(store: &StoreOpaque, id: ComponentInstanceId) -> Instance {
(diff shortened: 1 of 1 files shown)


crates/wasmtime/src/runtime/component/instance.rs
```diff
@@ -55,7 +55,7 @@ impl Instance {
     ///
     /// The safety of this function relies on `id` belonging to the instance
     /// within `store`.
-    pub(crate) unsafe fn from_wasmtime(store: &StoreOpaque, id: ComponentInstanceId) -> Instance {
+    pub(crate) fn from_wasmtime(store: &StoreOpaque, id: ComponentInstanceId) -> Instance {
```
dicej, line 53: Is the sentence regarding safety in the doc comment above still relevant, or should it be removed?

dicej (approved):

crates/wasmtime/src/runtime/component/instance.rs
alexcrichton: Oops no it should be removed

PR #10938, opened 2025-06-05
[34.0.0] Revert "Stack switching: Infrastructure and runtime support"

alexcrichton:
This is a revert of #10388 for the 34.0.0 release branch. That commit was the last commit in on the release branch and there should be no harm in deferring it for one extra release (stack-switching is still a work-in-progress on `main`) and the hope is that this reduces the risk/need to backport any changes.


fitzgen (approved):

alexcrichton:
cc @posborne sorry I should have cc'd you on this to start. I'm not dead-set on doing this I just figured it'd be a reasonable idea. If you would prefer to get this infrastructure landed on the release branch though I think it'd be reasonable to re-land. (and to clarify I have no intention of reverting on `main`, this was just to de-risk the release branch a little)

posborne:
> cc @posborne sorry I should have cc'd you on this to start. I'm not dead-set on doing this I just figured it'd be a reasonable idea. If you would prefer to get this infrastructure landed on the release branch though I think it'd be reasonable to re-land. (and to clarify I have no intention of reverting on main, this was just to de-risk the release branch a little)

No, I'm happy with this not being included on the release branch for this cycle.

PR #10939, opened 2025-06-05
Simplify and optimize the way that `enum`s are passed in fused adapters

fitzgen:
instead of treating them like variants that don't have payloads, just check that the discriminant is in bounds.


alexcrichton (approved):
If you're feeling particularly enterprising, mind adding a test to `tests/misc_testsuite/component-model/*.wast`? I think it should be reasonably easy to hook this all up and we could assert that traps don't happen for in-bounds values but do happen for out-of-bounds values

fitzgen:
> If you're feeling particularly enterprising, mind adding a test to `tests/misc_testsuite/component-model/*.wast`?

I'll give this a shot in a follow up PR

PR #10942, opened 2025-06-05
x64: generate rules to pick between SSE or AVX instructions

abrown:
The new assembler gives us the opportunity to generate some extra ISLE rules to eliminate boilerplate like the following, hand-written around 180 times throughout `inst.isle`:

```
(decl x64_... (Xmm XmmMem) Xmm)
(rule 1 (x64_... src1 src2)
        (if-let true (use_avx))
        (<emit avx version>))
(rule 0 (x64_... src1 src2) (<emit sse version>))
```

This PR only attempts to solve that exact pattern (i.e., `Xmm XmmMem`) but the approach could scale to more patterns in the future.

### DIFF
--- a/cranelift/assembler-x64/meta/src/dsl/encoding.rs
+++ b/cranelift/assembler-x64/meta/src/dsl/encoding.rs
@@ -59,6 +59,21 @@ impl Encoding {
             Encoding::Vex(vex) => vex.validate(operands),
         }
     }
+
+    /// Return the primary opcode for this encoding.
+    ///
+    /// Note that [`Rex`]-encoded instructions have a more complex opcode scheme
+    /// (see [`Opcodes`]), and this returns the last opcode: the secondary
+    /// opcode if one is available and the primary otherwise.
+    pub fn opcode(&self) -> u8 {
+        match self {
+            Encoding::Rex(rex) => match rex.opcodes.secondary {
+                Some(secondary) => secondary,
+                None => rex.opcodes.primary,
+            },
+            Encoding::Vex(vex) => vex.opcode,
+        }
+    }
 }
 
 impl fmt::Display for Encoding {
--- a/cranelift/assembler-x64/meta/src/instructions.rs
+++ b/cranelift/assembler-x64/meta/src/instructions.rs
@@ -20,7 +20,8 @@ mod sub;
 mod unpack;
 mod xor;
 
-use crate::dsl::Inst;
+use crate::dsl::{Feature, Inst, Mutability, OperandKind};
+use std::collections::HashMap;
 
 #[must_use]
 pub fn list() -> Vec<Inst> {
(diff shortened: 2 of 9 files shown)


cranelift/assembler-x64/meta/src/instructions.rs
```diff
@@ -44,5 +45,104 @@ pub fn list() -> Vec<Inst> {
     all.extend(sub::list());
     all.extend(xor::list());
     all.extend(unpack::list());
+
+    // Automatically assign AVX alternates to SSE instructions (see
+    // `Inst::alternate`). This allows later code generation to
+    // instruction-select between the AVX and SSE versions of an instruction.
+    assign_avx_alternates(&mut all);
```
abrown: The implementation (below) _could_ all go live in a separate module (?).

cranelift/codegen/meta/src/gen_asm.rs
```diff
@@ -504,6 +504,38 @@ pub fn generate_isle_inst_decls(f: &mut Formatter, inst: &Inst) {
             f,
             "(rule ({rule_name} {param_names}) ({convert} ({raw_name} {implicit_params} {param_names})))"
         );
+
+        if let Some(avx_name) = &inst.alternate {
```
abrown: I'm not terribly happy about dropping this here; it will work fine but at some point we may need to refactor the function this is in. It is just quite handy that all the parameter types and names have been pre-calculated for us.

cranelift/assembler-x64/meta/src/dsl/encoding.rs
```diff
@@ -59,6 +59,21 @@ impl Encoding {
...
     }
+
+    /// Return the primary opcode for this encoding.
+    ///
+    /// Note that [`Rex`]-encoded instructions have a more complex opcode scheme
+    /// (see [`Opcodes`]), and this returns the last opcode: the secondary
+    /// opcode if one is available and the primary otherwise.
+    pub fn opcode(&self) -> u8 {
+        match self {
+            Encoding::Rex(rex) => match rex.opcodes.secondary {
```
cfallin: Can we put this nested `match` in an `opcode()` method on `rex.opcodes`?

Also, it's not totally clear to me what "primary opcode" means in the doc-comment above and why we return the *secondary* for the "primary" if it exists -- I read the doc-comment for `Opcodes` below too and I'm still a little fuzzy. Could we say what we mean here by "primary" and why the secondary is the primary?

cranelift/assembler-x64/meta/src/instructions.rs
```diff
@@ -44,5 +45,104 @@ pub fn list() -> Vec<Inst> {
...
+
     all
 }
+
+/// Assigns AVX alternates to SSE instructions.
```
cfallin: It's neat that this was made to work automagically -- all the same, I worry a tiny bit that it may be brittle (unexpectedly determine a match when some subtle condition means they're not quite equivalent; or on the other side, miss a match because of something). The note below about rounding and the fact this limits things to just one format for now as a result seems to bear this out a little.

In the spirit of the explicit DSL, I wonder if it would be better to have a clause on the instruction builder that sets the alternate? Something like

```
  inst("vpadd", ...).avx_for("padd"),
  inst("padd", ...),
  ...
```

(`avx_for`, `upgrades`, `equiv`, pick a verb phrase...)

Thoughts?

cfallin (commented):
Thanks -- neat that this was made to work! Some thoughts below. If this PR turns out to be the best way I'm fine with it going in but I'm curious what you think of the below...

cranelift/assembler-x64/meta/src/instructions.rs
abrown: Yeah, I did consider adding an explicit version: `inst(...).alt("vpadd")`. I can't remember what dissuaded me, but probably laziness (there will be a lot of places to update!). But it is more clear so I'm in favor of that &mdash; less magic. I'll rework this to use a DSL function and use the current "check the SSE-to-AVX link" logic to verify that the user has specified `.alt(...)` correctly.

cfallin: Sounds good, thanks!

cranelift/assembler-x64/meta/src/dsl/encoding.rs
abrown: The manual uses "primary" and "secondary" to explain all this but, yes, it kind of feels like the "primary" opcode should always be the last one. I'll add `Opcodes::opcode` and try to explain things there a bit better.

cranelift/assembler-x64/meta/src/instructions.rs
alexcrichton: Personally I'd agree with @cfallin as well in that I'm in favor of explicitly saying "this is the AVX thing for that other instruction" which then generates a panic/assertion/etc if something doesn't match. (I realize y'all've already reached this conclusion but wanted to echo my opinion here too to voice support)

alexcrichton (approved):
Nice 👍

PR #10943, opened 2025-06-05
Crack down on mutability and ownership of `vm::Instance`

alexcrichton:
This commit represents more effort to bring safety to `vm::Instance` and, eventually, `ComponentInstance`. This is specifically addressing two points of safety around `vm::Instance`:

* Previously ownership of this was murky where `InstanceHandle` sort of represented ownership but sort of didn't either through the `InstanceHandle::clone` method. Now `InstanceHandle` has a destructor for instances and no longer has `clone`, so there's one exclusive owner of an instance.

* Previously `&mut Instance` was liberally passed around, but this is not sound because certain fields cannot be mutated (e.g. runtime offset information). While not a perfect solution this PR switches to using `Pin<&mut Instance>` everywhere instead. This prevents safe access to `&mut Instance` and we hand-write accessors to individual fields. Notably we omit mutable access to the `runtime_info` field.

This naturally involved a lot of refactoring internally, but notably this started bringing up preexisting issues around how there are locations in the codebase that simultaneously have `&mut Instance` and `&mut StoreOpaque` which is technically not sound due to being able to get back to the instance from the store. Some issues here were address by passing around indices more often such as in instance initialization and const-expr evaluation.

Note that all proxy methods on `InstanceHandle` are also all removed now and there's now only two: `get` and `get_mut`. This reflects how `InstanceHandle` should in general no longer be used and instead `Instance` itself, and some pointer-to thereof, should be exclusively used.

cc #10933

### DIFF
--- a/crates/wasmtime/src/runtime/externals/global.rs
+++ b/crates/wasmtime/src/runtime/externals/global.rs
@@ -303,7 +303,7 @@ impl Global {
         match self.kind {
             VMGlobalKind::Instance(index) => {
                 let instance = InstanceId::from_u32(self.instance);
-                let module = store.instance(instance).module();
+                let module = store.instance(instance).env_module();
                 let index = module.global_index(index);
                 &module.globals[index]
             }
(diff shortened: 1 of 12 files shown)


fitzgen (approved):
nice

alexcrichton:
I'll also note that ownership-wise this brings core wasm in line with components, where components have `OwnedComponentInstance` which is now basically the exact same thing as `InstanceHandle` (and I'll try to eventually get around to aligning on naming conventions too)

alexcrichton:
While I have no reason to believe that's a flaky test failure I'm neverthless curious. If it's flaky that's quite worrisome. If it's not flaky that's also worrisome...

alexcrichton:
Ok spurious failures should be fixed by https://github.com/bytecodealliance/wasmtime/pull/10947

PR #10944, opened 2025-06-05
Cranelift: Remove unnecessary cursor position setting in legalization

fitzgen:


alexcrichton:
That flaky test failure should be fixed in https://github.com/bytecodealliance/wasmtime/pull/10947

PR #10855, opened 2025-05-28
x64: Convert Div instructions to new assembler

rahulchaphalkar:
Converts div instructions to new assembler. Remove old div impl from codegen for SSE variants, not Vex variants.

PR #10946, opened 2025-06-06
asm: follow up on DSL nits

abrown:
Some final review comments went unaddressed in #10855 that this change resolve, primarily using `LIG` for non-vector definitions. I was going to add `.wig()` everywhere but, after thinking about it, it _is_ the default value for `vex()` and leaving it out reduces the amount of information one has to parse. This also cleans up a few extra lines I saw lying around.


alexcrichton (approved):

alexcrichton:
That flaky test failure should be fixed in https://github.com/bytecodealliance/wasmtime/pull/10947

PR #10947, opened 2025-06-06
Align symbol names with `--emit-clif`

alexcrichton:
This commit fixes two issues: (1) is that `--emit-clif` generates file names that are different from symbol names and (2) is that `--emit-clif` for components today has collisions on filenames which means that output files can get corrupted. Currently (2) is causing spurious failures in CI due to the test added in #10939 being the first test with a component and functions in modules where they clash to emit the same file.


cfallin (approved):
going to go ahead and review to unblock CI -- thanks!

fitzgen:
Thanks!

PR #10953, opened 2025-06-06
Synchronize p3 intrinsics/translation

alexcrichton:
This commit synchronizes the list of intrinsics and notably libcalls with the wasip3-prototyping repository. This change alone does not really do all that much but the goal is to reduce the diff between this repository and the wasip3-prototyping repository. The main change included here is synchronizing the list of libcalls. That necessitates changing various concrete signatures of libcalls and all implementations are replaced with `todo!()` for now to get filled in in the future. Additionally this necessitates changes to the trampoline compilation as well which is brought wholesale from the wasip3-prototyping repository.

Note that this should all be well-tested and covered in the wasip3-prototyping repository so no extra tests are brought in at this time. The intention is that once the runtime bits start landing here upstream it'll be accompanied with appropriate tests.


fitzgen (approved):
lgtm

Issue #10000, opened 2025-01-13
Spurious (?) CI failures on s390x


alexcrichton:
I've seen two CI builds fail today on the s390x tests of the `wasmtime-cli` crate:

* https://github.com/bytecodealliance/wasmtime/actions/runs/12752790805/job/35543074991 (https://github.com/bytecodealliance/wasmtime/pull/9982#event-15909676078)
* https://github.com/bytecodealliance/wasmtime/actions/runs/12755184456/job/35550742014 (https://github.com/bytecodealliance/wasmtime/pull/9994#event-15912281849)

The first was re-queued and merged and the second is currently back in the queue. The first didn't touch anything in `wasmtime-cli` and the second does touch quite a bit that could affect it so it's currently an assumption of mine that the s390x-failure there was specific to s390x.

I tried running the test suite in a loop locally for a bit using s390x/qemu and so far it hasn't failed in the same way. I figured I'd open an issue to keep track of things if they happen in the future.

cc @uweigand

iii-i:
In my local runs I've seen segfaults due to an ABI problem in the capstone wrapper: https://github.com/capstone-rust/capstone-rs/pull/166. I wonder if this can be related?

abrown:
Not sure, but I believe I saw a similar failure over in #10110: https://github.com/bytecodealliance/wasmtime/actions/runs/13124822832/job/36618948362#step:18:987.

alexcrichton:
Another failure at https://github.com/bytecodealliance/wasmtime/actions/runs/13421426149/job/37494891337

alexcrichton:
[another](https://github.com/bytecodealliance/wasmtime/actions/runs/14223794316/attempts/1)

alexcrichton:
[another](https://github.com/bytecodealliance/wasmtime/actions/runs/15355939727/job/43215004725)

alexcrichton:
[another](https://github.com/bytecodealliance/wasmtime/actions/runs/15499429820/job/43643913057)

alexcrichton:
[another](https://github.com/bytecodealliance/wasmtime/actions/runs/23758391421/job/69219149673)

alexcrichton:
[another](https://github.com/bytecodealliance/wasmtime/actions/runs/24025790583/job/70063899590)

alexcrichton:
[another](https://github.com/bytecodealliance/wasmtime/actions/runs/24039708464/job/70108081361)

[...]

PR #10956, opened 2025-06-06
Cranelift: Refactor legalization such that traversal is separate from legalizing a particular instruction

fitzgen:
Notably, make sure that we aren't using the same `FuncCursor` for traversal and modification, as that makes it hard to follow when the traversal is being moved forward and backed up.

Additionally, this makes it so that when we legalize an instruction, we only back up the cursor to continue legalization until we reach a fixed-point when said legalization resulted in something that actually requires further legalization. This should be mildly more efficient, since we can avoid an unnecessary loop iteration and its `match` on the just-legalized instruction data and opcode.

### DIFF
--- a/cranelift/codegen/src/legalizer/mod.rs
+++ b/cranelift/codegen/src/legalizer/mod.rs
@@ -49,205 +49,259 @@ fn imm_const(pos: &mut FuncCursor, arg: Value, imm: Imm64, is_signed: bool) -> V
     }
 }
 
+/// A command describing how the walk over instructions should proceed.
+enum WalkCommand {
+    /// Continue to the next instruction, if any.
+    Continue,
+    /// Revisit the current instruction (presumably because it was legalized
+    /// into a new instruction that may also require further legalization).
+    Revisit,
+}
+
+/// A simple, naive forwards walk over every instruction in every block in the
+/// function's layout.
+///
+/// This does not guarantee any kind of post-order visitation or anything like
+/// that, it is just iterating over blocks in layout order, not any kind of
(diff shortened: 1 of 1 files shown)


cranelift/codegen/src/legalizer/mod.rs
```diff
@@ -49,205 +49,259 @@ fn imm_const(pos: &mut FuncCursor, arg: Value, imm: Imm64, is_signed: bool) -> V
...
+    func: &mut ir::Function,
+    mut f: impl FnMut(&mut ir::Function, ir::Inst) -> WalkCommand,
+) {
+    let mut pos = FuncCursor::new(func);
+    while let Some(_block) = pos.next_block() {
+        let prev_pos = pos.position();
+        while let Some(inst) = pos.next_inst() {
+            match f(pos.func, inst) {
+                WalkCommand::Continue => continue,
+                WalkCommand::Revisit => pos.set_position(prev_pos),
```
alexcrichton, line 83: I think this might be revisiting too much since `prev_pos` is never updated?

alexcrichton (approved):

cranelift/codegen/src/legalizer/mod.rs
fitzgen: Heh, yeah I actually fixed that in a later commit and then realized I never updated the PR and was in the middle of doing that rn

fitzgen: good eye!

alexcrichton:
I think that's https://github.com/bytecodealliance/wasmtime/issues/10000

PR #6025, opened 2023-03-15
x64: Improve codegen for splats

alexcrichton:
This commit goes through the lowerings for the CLIF `splat` instruction and improves the support for each operator. Many of these lowerings are mirrored from v8/SpiderMonkey and there are a number of improvements:

* AVX2 `v{p,}broadcast*` instructions are added and used when available.
* Float-based splats are much simpler and always a single-instruction
* Integer-based splats don't insert into an uninit xmm value and instead start out with a `movd` to move into an `xmm` register. This thoeretically breaks dependencies with prior instructions since `movd` creates a fresh new value in the destination register.
* Loads are now sunk into all of the instructions. A new extractor, `sinkable_load_exact`, was added to sink the i8/i16 loads.

PR #10958, opened 2025-06-06
x64: Migrate `movddup` to the new assembler

alexcrichton:
Enables removing the `XmmUnaryRmRUnaligned` instruction variant. Additionally this exposed a shortcoming where Cranelift didn't have access for testing for SSE3 in ISLE (not to be confused with SSSE3). Additionally `movddup` was incorrectly classified as SSSE3 so this commit fills out some infrastructure to use SSE3 instead of SSSE3. There are more instructions as well that are mis-classified under SSSE3 which I'll address in a follow-up.


alexcrichton:
On the topic of SSE3 testing lacking in Cranelift and instructions mis-classified as SSSE3 instead of SSE3 AFAIK it's benign and otherwise dates back to #6025 where I assume I misread something and got the classification wrong. I think that `movddup` is the only SSE3 instruction that Cranelift supports at this time too.

fitzgen (approved):

PR #10960, opened 2025-06-06
Minor updates to the `wasmtime-unwinder` crate

alexcrichton:
* Don't opt-out of the default set of lints.
* Minor changes to inline assembly which shouldn't affect functionality.


fitzgen (approved):
Nice

alexcrichton:
I'll file the s390x inline assembly changes as a follow-up since I'm either misunderstanding something or that's indicative of a bug somewhere else, but best to keep that a small isolated change.

alexcrichton:
I'm dropping inline assembly changes to figure out what's going on later.

PR #10962, opened 2025-06-06
cranelift: ISLE printer

mmcloughlin:
This PR adds a printer to the ISLE library. This is intended to support tooling which generates ISLE source code. In such cases, this printer library allows applications to work with ISLE's AST types rather than raw string manipulation, and produce more readable output due to the pretty-printing features.

The printer is tested with a new integration test that verifies the print-parse roundtrip property for provided ISLE files.


cranelift/isle/isle/src/parser.rs
```diff
@@ -12,12 +12,21 @@ pub fn parse(lexer: Lexer) -> Result<Vec<Def>> {
     parser.parse_defs()
 }
 
+/// Parse without positional information. Provided mainly to support testing, to
+/// enable equality testing on structure alone.
+pub fn parse_without_pos(lexer: Lexer) -> Result<Vec<Def>> {
+    let mut parser = Parser::new(lexer);
+    parser.disable_pos();
+    parser.parse_defs()
+}
```
mmcloughlin: Previous discussion still applies: https://github.com/bytecodealliance/wasmtime/pull/8263#discussion_r1543966705. I still tend to think this is the simplest of the options.

@cfallin regarding the suggestion "a helper that maps through all AST nodes and sets positions to e.g. zero?". I feel like this could be quite a bit of boilerplate to implement. Happy to do that if you prefer.

alexcrichton:
redirecting review to @cfallin who I believe has more context on this

cranelift/isle/isle/build.rs
```diff
@@ -19,7 +24,23 @@ fn main() {
     std::fs::write(output, out).unwrap();
 }
 
+fn isle_printer_tests(out_dir: &std::path::PathBuf) {
+    let mut out = String::new();
+
+    emit_tests(&mut out, "isle_examples/pass", "run_print");
+    emit_tests(&mut out, "../../codegen/src", "run_print");
```
cfallin: Unfortunately we can't refer to one crate from another with paths like this: the tests have to work even when the crates are vendored or in some other directory layout. Can you run the print-tests over the existing ISLE test files we have in this crate?

cranelift/isle/isle/src/parser.rs
```diff
@@ -12,12 +12,21 @@ pub fn parse(lexer: Lexer) -> Result<Vec<Def>> {
     parser.parse_defs()
 }
 
+/// Parse without positional information. Provided mainly to support testing, to
+/// enable equality testing on structure alone.
+pub fn parse_without_pos(lexer: Lexer) -> Result<Vec<Def>> {
+    let mut parser = Parser::new(lexer);
+    parser.disable_pos();
```
cfallin: Rather than a setter, can we have an alternative constructor, e.g. `Parser::new_without_pos_tracking()`?

cranelift/isle/isle/src/printer.rs
```diff
@@ -0,0 +1,560 @@
...
+            Def::Extern(ext) => ext.to_sexpr(),
+            Def::Converter(converter) => converter.to_sexpr(),
+        }
+    }
+}
+
+impl ToSExpr for Type {
+    fn to_sexpr(&self) -> SExpr {
+        let mut parts = vec![SExpr::atom("type"), self.name.to_sexpr()];
+        if self.is_extern {
```
cfallin: Here and elsewhere -- anywhere a node is a struct with fields -- a pattern I like to use is an infallible destructuring without an ellipsis, to ensure I've got all the fields and do something with them; e.g.:

```rust
let Type { is_extern, is_nodebug, ty } = self;
// ... print `is_extern`, `is_nodebug`, `ty`
```

This way, if/when we add new fields, we don't forget to add them to the printer, because we'll get a compile error otherwise.

cfallin (commented):
Looks generally good -- thanks for writing this out! A few things below but nothing too bad:

cranelift/isle/isle/src/parser.rs
mmcloughlin: Yes, done.

cranelift/isle/isle/src/printer.rs
mmcloughlin: This is a really nice technique, thanks for sharing. I've used it throughout the file.

cranelift/isle/isle/build.rs
mmcloughlin: Makes sense, removed. I also consolidated the print tests back into the existing `run_tests.rs`. The separate `printer_tests.rs` was a holdover from #8263, which had the printer guarded by a feature flag because of the pretty-printer dependency.

I did find the tests over the complete ISLE files to be useful/reassuring during development, and a useful backup to ensure the printer stays in sync with the ISLE syntax. So I wonder if there's a good way to do this that maintains the independence of the ISLE crate. Ideas:

* Dump a snapshot of ISLE files into `isle_examples/print`. For example, from the cranelift directory running: `find . -name '*.isle' | xargs cat >isle/isle/isle_examples/print/all.isle`.
* Some special environment variable signals that additional test cases should be generated (e.g. `WASMTIME_CRANELIFT_DIR`). This would be set in Wasmtime CI but not considered an error if it's missing.

cfallin: We could potentially add a test to the `cranelift-codegen` side to roundtrip-test the printer -- that's the natural point in the dep graph at which we have both those files and the ISLE compiler. I'm happy to review that as a followup if you'd like.

cfallin (approved):
Thanks!

PR #10963, opened 2025-06-06
More clearly flag internal crates as such

alexcrichton:
This commit is an attempt to more clearly flag internal crates in this project as internal and not intended for external use. Specifically:

* Many crates are renamed from `wasmtime-foo` to `wasmtime-internal-foo`.
* All of these crates now have `INTERNAL: ...` in their crates.io description.
* All of these crates now have a warning at the top of their documentation discouraging use.

This change is a result of rustsec/advisory-db#1999 where the goal is to be crystal clear from a project perspective that usage of these crates are highly discouraged and not supported. We'll still probably get such advisories but we won't be considering them CVEs from the project itself due to the internal nature of these crates and the discouraging warnings.

Some concrete changes used here are:

* Inter-crate dependencies still use `wasmtime_foo` for naming and do so with Cargo's package-renaming features.
* Crate renames are specified at the workspace level so the rename is only in one locations and all other inherit it.
* Contribution documentation now has some brief guidelines about crate organization.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -446,7 +446,7 @@ jobs:
       with:
         submodules: true
     - uses: ./.github/actions/install-rust
-    - run: cargo test -p wasmtime-fiber --no-default-features
+    - run: cargo test -p wasmtime-internal-fiber --no-default-features
     - run: cargo test -p cranelift-tools --test logged-filetests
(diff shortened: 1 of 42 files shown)


pchickey (approved):
Thanks, Alex

PR #10965, opened 2025-06-06
`cranelift-frontend`: Refactor how variables are declared

fitzgen:
Rather than requiring users to construct a `Variable` entity out of thin air and *then* declare it to the frontend, declaring a new variable now returns the new `Variable` entity. And instead of having a `SecondaryMap<Variable, Type>` inside of `FunctionBuilder` we now have a `PrimaryMap<Variable, Type>`. Using variables in `cranelift-frontend` should generally feel a little bit better and more idiomatic now.

### DIFF
--- a/cranelift/frontend/src/lib.rs
+++ b/cranelift/frontend/src/lib.rs
@@ -81,12 +81,9 @@
 //!     let block1 = builder.create_block();
 //!     let block2 = builder.create_block();
 //!     let block3 = builder.create_block();
-//!     let x = Variable::new(0);
-//!     let y = Variable::new(1);
-//!     let z = Variable::new(2);
-//!     builder.declare_var(x, I32);
-//!     builder.declare_var(y, I32);
-//!     builder.declare_var(z, I32);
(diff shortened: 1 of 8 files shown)


cranelift/frontend/src/lib.rs
```diff
@@ -81,12 +81,9 @@
...
 //!     let block3 = builder.create_block();
-//!     let x = Variable::new(0);
-//!     let y = Variable::new(1);
-//!     let z = Variable::new(2);
-//!     builder.declare_var(x, I32);
-//!     builder.declare_var(y, I32);
-//!     builder.declare_var(z, I32);
+//!     let x = builder.define_var(I32);
+//!     let y = builder.define_var(I32);
+//!     let z = builder.define_var(I32);
```
bjorn3: These should be declare_var, not define_var, right?

fitzgen: Whoops, yep, thanks

cfallin (approved):
Nice -- much cleaner!

PR #10968, opened 2025-06-06
Add InputFile for use with WasiCtxBuilder

jeffcharles:
This adds a `InputFile` for use with the `stdin` method available on `WasiCtxtBuilder`. `InputFile` corresponds with `OutputFile` available for the `stdout` and the `stderr` methods.


fitzgen:
Redirecting review to @pchickey since I'm not too familiar with this layer.

pchickey (approved):
Thank you!

PR #10972, opened 2025-06-06
 Re-enable concurrent bindings generation tests 

alexcrichton:
This commit re-enables tests for bindings generation for concurrent
calls in the main repo after all syncs have now finished with wasip3.
This additionally adds some skeleton APIs that the bindings generator
uses which are necessary to get tests passing.


dicej (approved):
Thanks!

PR #10973, opened 2025-06-06
Test out s390x inline assembly change in isolation

alexcrichton:
Extracted from #10960

prtest:linux-s390x

### DIFF
--- a/crates/unwinder/src/arch/s390x.rs
+++ b/crates/unwinder/src/arch/s390x.rs
@@ -30,20 +30,14 @@ pub unsafe fn resume_to_exception_handler(
 ) -> ! {
     unsafe {
         core::arch::asm!(
-            "lgr %r6, {}",
-            "lgr %r7, {}",
             "lgr %r15, {}",
             "br {}",
-            in(reg) payload1,
-            in(reg) payload2,
             in(reg) sp,
             in(reg) pc,
-            out("r6") _,
-            out("r7") _,
-            options(nostack, nomem),
+            in("r6") payload1,
+            in("r7") payload2,
+            options(nostack, nomem, noreturn),
         );
-
-        core::hint::unreachable_unchecked()
     }
 }
 


cfallin:
I'm pretty perplexed staring at this one too -- for reference here are Godbolt links for [before](https://godbolt.org/z/oYaYYWoM4) (working) and [after](https://godbolt.org/z/ccfjPxTrv) (apparently segfaulting). The clobber-saves come in different order wrt loading values into payload (r6/r7) but the dataflow from arguments to PC/SP (r15)/payload (r6, r7) looks completely equivalent...

alexcrichton:
cc @uweigand would you be able to help dig in to what's going on here? (we can provide more background context if necessary, but the general gist is that this PR in theory is a refactoring that shouldn't do anything but it causes tests to segfault)

uweigand:
I was finally able to reproduce this.  Turns out the crash only happens when building with `CARGO_INCREMENTAL=0`.

crates/unwinder/src/arch/s390x.rs
```diff
@@ -30,20 +30,14 @@ pub unsafe fn resume_to_exception_handler(
...
     unsafe {
         core::arch::asm!(
-            "lgr %r6, {}",
-            "lgr %r7, {}",
             "lgr %r15, {}",
             "br {}",
-            in(reg) payload1,
-            in(reg) payload2,
             in(reg) sp,
             in(reg) pc,
```
uweigand: Here's the problem.  The `br` instruction requires an address register operand (i.e. GPR 1-15, but not GPR 0).  In the failing case, the compiler chose to use GPR 0 to hold `pc`.  To fix this, you'll need to use the `reg_addr` register class instead of `reg` for `pc`.

alexcrichton: Nice thanks so much for helping to track this down! Changing to `reg_addr` looks to fix it indeed!

alexcrichton:
Ok this should be good to go now, thanks again @uweigand!

alexcrichton:
ping @pchickey, mind taking a look at this? (I can also reassign if you'd like too)

PR #10974, opened 2025-06-07
Fix ASAN in combination with throw intrinsic

alexcrichton:
Prior to this commit locally I would see a nondeterministic crash when executing:

    RUSTFLAGS=-Zsanitizer=address \
      cargo +nightly run --release test filetests

The crash has a scary-looking internal assertion within the ASAN runtime itself and local debugging seems to show that threads are required to reproduce this and it's nondeterministic. I don't know why CI is currently passing on all PRs when this is failing so frequently for myself locally.

In any case it seems like there's a magical ASAN intrinsic to call before `noreturn` things which I think is intended for thread-stopping events which isn't a great fit for exceptions here but it seems to resolve the crash.

Basically I have no idea why things crash locally, what this new function does, nor why it fixes things. I am but a slave to the CI gods in an attempt to make it so our CI is more reliable.

### DIFF
--- a/crates/unwinder/build.rs
+++ b/crates/unwinder/build.rs
@@ -1,0 +1,18 @@
+fn main() {
+    // NB: duplicating a workaround in the wasmtime-fiber build script.
+    custom_cfg("asan", cfg_is("sanitize", "address"));
+}
+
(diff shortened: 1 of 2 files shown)


cfallin (approved):
Thanks for catching this!

PR #10975, opened 2025-06-07
x64: Migrate some blend-related instructions

alexcrichton:
This commit migrates the `pblendvb`, `blendvps`, `blendvpd`, and AVX versions of each to the new assembler. This required a few minor features in the assembler:

* The fixed physical `xmm0` register is now a `Location`
* The "/is4" encoding format for VEX instructions is supported
* The `Fixed` type now supports both xmm and gpr registers (minor update to its `to_string` method).

In the end this enables deleting `XmmRmRBlend` and `XmmRmRBlendVex`. (yay!)


cranelift/filetests/filetests/isa/aarch64/nan-canonicalization.clif
```diff
@@ -21,7 +21,7 @@ block0(v0: f32x4, v1: f32x4):
 ;   movdqa %xmm0, %xmm2
 ;   movdqa %xmm1, %xmm0
 ;   movdqa %xmm2, %xmm1
-;   pblendvb %xmm1, %xmm7, %xmm1
+;   pblendvb %xmm0, %xmm7, %xmm1
```
alexcrichton, line 24: I'll note that there's no actual register allocation changes here, this is just an artifact of how the instruction used to be printed. The xmm0 argument was omitted by default and the src/destination, which are required to be the same register, were printed twice.

cranelift/assembler-x64/meta/src/instructions/lanes.rs
```diff
@@ -36,5 +36,12 @@ pub fn list() -> Vec<Inst> {
 
         inst("movddup", fmt("A", [w(xmm1), r(xmm_m64)]), rex([0xF2, 0x0F, 0x12]).r(), _64b | compat | sse3),
         inst("vmovddup", fmt("A", [w(xmm1), r(xmm_m64)]), vex(L128)._f2()._0f().op(0x12).r(), _64b | compat | avx),
+
+        inst("pblendvb", fmt("RM", [rw(xmm1), r(align(xmm_m128)), r(xmm0)]), rex([0x66, 0x0F, 0x38, 0x10]).r(), _64b | compat | sse41),
```
abrown, line 40:
```suggestion
        inst("pblendvb", fmt("RM", [rw(xmm1), r(align(xmm_m128)), r(implicit(xmm0))]), rex([0x66, 0x0F, 0x38, 0x10]).r(), _64b | compat | sse41),
```

Don't we want this to be implicit?

alexcrichton: I originally thought so yeah but apparently Capstone disassembles this with `%xmm0` show it shows up in the instruction listing. It looks like even Intel-style disassembly shows `xmm0` as well, so for whatever reason while it's implicit on instructions like `div` it's not implicit here...

abrown (approved):

PR #10977, opened 2025-06-07
 x64: Migrate push/pop to the new assembler 

alexcrichton:
Figured these would be some interesting instruction shapes. This adds a
helper to the ISLE generation to skip some instructions and the "push"
instructions are skipped as they're not needed in ISLE and would
otherwise require binding new instruction shapes which didn't seem worth
it. Additionally many items in `gen_asm.rs` were made private (removed
`pub`).


abrown (approved):
Makes sense!

PR #10983, opened 2025-06-09
[34.0.0] Update release notes

alexcrichton:
My take on 

```
git log `git merge-base HEAD origin/release-33.0.0`...HEAD
```

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -4,7 +4,82 @@ Unreleased.
 
 ### Added
 
+* Support for SIMD in the Pulley interpreter can now be disabled at compile-time
+  to shrink the size of the final binary.
+  [#10727](https://github.com/bytecodealliance/wasmtime/pull/10727)
+
+* The C API now has `wasmtime_trap_new_code` to create a `wasm_trap_t` from
+  its code.
+  [#10765](https://github.com/bytecodealliance/wasmtime/pull/10765)
+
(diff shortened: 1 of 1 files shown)


RELEASES.md
```diff
@@ -4,8 +4,83 @@ Unreleased.
...
+
+* Wasmtime and Cranelift's minimnum supported version of Rust (MSRV) is now
+  1.85.0.
+  [#10785](https://github.com/bytecodealliance/wasmtime/pull/10785)
+
+* Cranelift's `debugtrap` on aarch64 now generates `brk #0xf000` for debuggers
+  to recognize it.
+  [#10813](https://github.com/bytecodealliance/wasmtime/pull/10813)
+
+* The wasi-http implementation no longer generates a trap if the handle to
```
pchickey, line 60:
```suggestion
* The wasi-http implementation responds with a 500 server error if the handle to
```

pchickey (approved):

RELEASES.md
alexcrichton: Oh I believe that this is the other way around, where for the embedder API if the channel to receive the response is dropped then previously `response-outparam.set` would generate a trap but now it just ignores the error. So this is more of a guest-side change than a host-side change  (e.g. as it relates to `wasmtime serve`)

alexcrichton: I'm going to go ahead and land without this change, but happy to follow-up of course

Issue #10941, opened 2025-06-05
Optimize control-flow-y conditional traps into single CLIF `trapnz` instructions


fitzgen:
Wasm does not have conditional trap instructions, just control flow and unconditional traps. CLIF does have conditional trap instructions, but we do not translate the control-flow-y Wasm equivalent into these single conditionally-trapping instructions. This is unfortunate because Cranelift's mid-end cannot currently turn control-flow-y conditional traps into single `trap[n]z` instructions either, but the single instructions allow the optimizer to do more and also ultimately get better codegen in the backend.

I'm sure that conditional traps show up frequently in real Wasm code. They additionally show up frequently in the fused adapters we generate when linking components together.

Might be worth identifying blocks that contain a single unconditional trap instruction and rewrite conditional branches to these blocks during egraph construction or something.

https://github.com/bytecodealliance/wasmtime/pull/10940 introduced the following disas tests that highlight the problem:

```wat
(module
  ;; This function body should ideally get compiled down into a single `trapz`
  ;; CLIF instruction.
  (func (export "trapnz") (param i32)
    local.get 0
    if
      unreachable
    end
  )

  ;; And this one into a single `trapnz` instruction.
  (func (export "trapz") (param i32)
    local.get 0
    i32.eqz
    if
      unreachable
    end
  )
)

;; function u0:0(i64 vmctx, i64, i32) tail {
;;     gv0 = vmctx
;;     gv1 = load.i64 notrap aligned readonly gv0+8
;;     gv2 = load.i64 notrap aligned gv1+16
;;     stack_limit = gv2
;;
;;                                 block0(v0: i64, v1: i64, v2: i32):
;; @002f                               brif v2, block2, block3
;;
;;                                 block2:
;; @0031                               trap user11
;;
;;                                 block3:
;; @0033                               jump block1
;;
;;                                 block1:
;; @0033                               return
;; }
;;
;; function u0:1(i64 vmctx, i64, i32) tail {
;;     gv0 = vmctx
;;     gv1 = load.i64 notrap aligned readonly gv0+8
;;     gv2 = load.i64 notrap aligned gv1+16
;;     stack_limit = gv2
;;
;;                                 block0(v0: i64, v1: i64, v2: i32):
;;                                     v5 = iconst.i32 0
;; @0038                               v3 = icmp eq v2, v5  ; v5 = 0
;; @0038                               v4 = uextend.i32 v3
;; @0039                               brif v4, block2, block3
;;
[...]

PR #10988, opened 2025-06-09
Cranelift: Rewrite conditional branches to unconditional traps into conditional traps during legalization

fitzgen:
Given this instruction:

```clif
brif v0, block1, block2
```

If we know that `block1` does nothing but immediately trap then we can rewrite
that `brif` into the following:

```clif
trapz v0, <trapcode>
jump block2
```

(And we can do the equivalent with `trapz` if `block2` immediately traps).

This transformation allows for the conditional trap instructions to be GVN'd and
for our egraphs mid-end to generally better optimize the program. We
additionally have better codegen in our backends for `trapz` than branches to
unconditional traps.

Fixes https://github.com/bytecodealliance/wasmtime/issues/10941

### DIFF
--- a/cranelift/codegen/src/context.rs
+++ b/cranelift/codegen/src/context.rs
@@ -168,13 +168,13 @@ impl Context {
             self.func.display()
         );
 
-        self.compute_cfg();
         if isa.flags().enable_nan_canonicalization() {
             self.canonicalize_nans(isa)?;
         }
 
         self.legalize(isa)?;
 
+        self.compute_cfg();
         self.compute_domtree();
         self.eliminate_unreachable_code(isa)?;
         self.remove_constant_phis(isa)?;
@@ -291,6 +291,7 @@ impl Context {
         // TODO: Avoid doing this when legalization doesn't actually mutate the CFG.
         self.domtree.clear();
         self.loop_analysis.clear();
+        self.cfg.clear();
 
         // Run some specific legalizations only.
         simple_legalize(&mut self.func, isa);
--- a/cranelift/codegen/src/ir/layout.rs
+++ b/cranelift/codegen/src/ir/layout.rs
@@ -516,6 +516,12 @@ impl Layout {
         }
     }
 
+    /// Does the given block contain exactly one instruction?
+    pub fn block_contains_exactly_one_inst(&self, block: Block) -> bool {
+        let block = &self.blocks[block];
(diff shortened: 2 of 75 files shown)


alexcrichton:
Could you detail more why this is part of legalization and not a separate function/pass? It seems a bit odd to me to have this be intertwined with legalization as it's basically not realized to legalization at all and is more of an optimization pass. Was it a compile-speed-related concern?

Also, do you know why so many tests are changing? Is it also due to the nature of the first commit where it's switching to a backwards walk from a forwards walk which renumbers legalization-produced things which GVN might pick differently between now?

> We additionally have better codegen in our backends for trapz than branches to unconditional traps.

FWIW with `cold` blocks I don't think this is true, I believe if the trapping block is `cold` the codegen should basically be the same. That being said the GVN point carries this PR alone IMO in terms of motivation.

fitzgen:
> Could you detail more why this is part of legalization and not a separate function/pass? It seems a bit odd to me to have this be intertwined with legalization as it's basically not realized to legalization at all and is more of an optimization pass. Was it a compile-speed-related concern?

Yes, I did not want to introduce another IR traversal, and would consider that a non-starter for solving this issue. We have in general been trying to remove passes (eg we've talked a lot about how to fuse phi-removal into the egraph pass) rather than add them.

You can think of legalization as a type of canonicalization, and this PR is canonicalizing conditional-branch-to-unconditional-trap into conditional-trap, if that helps.

> Also, do you know why so many tests are changing? Is it also due to the nature of the first commit where it's switching to a backwards walk from a forwards walk which renumbers legalization-produced things which GVN might pick differently between now?

Yes, GVN is a forwards pass will choose the first value it sees as the canonical version. This used to be the legalized value with the lowest number because legalization was also a forwards pass so values at the start of the function would be legalized before those at the end. After the change to legalize in reverse, the legalized values at the end of the function are lower and at the start of the function are higher. So even if the IR is the "same" except for value numbers, the value numbers do change in the golden output.

> FWIW with `cold` blocks I don't think this is true, I believe if the trapping block is `cold` the codegen should basically be the same.

We don't mark any Wasm blocks `cold` during translation today, so we won't actually get to that scenario. But maybe we should start making translation smarter and marking blocks cold based on heuristics at some point.

alexcrichton (approved):
Sounds reasonable to me 👍

PR #10989, opened 2025-06-09
winch: Simplify constant handling, part 2/N

saulecabrera:
This patch is best reviewed commit by commit.

This is a follow-up to https://github.com/bytecodealliance/wasmtime/pull/10888

Scratch registers are used throughout Winch's codebase to handle constant values that cannot be directly encoded in specific instructions. Since the encoding rules are ISA-specific, selecting instructions for these operations often requires scratch registers. Ideally, these temporary registers should be acquired without adding any register pressure to prevent value spilling. 

The objective of this change is to make it generally safer to work with scratch registers and prevent accidental clobbering of such registers.

### DIFF
--- a/winch/codegen/src/isa/aarch64/masm.rs
+++ b/winch/codegen/src/isa/aarch64/masm.rs
@@ -448,18 +448,7 @@ impl Masm for MacroAssembler {
         size: OperandSize,
     ) -> Result<()> {
         let (base, offset) = src.unwrap_offset();
-        match Imm12::maybe_from_u64(offset as u64) {
-            Some(imm12) => self.asm.add_ir(imm12, base, dst, size),
-            None => {
-                self.with_scratch::<IntScratch, _>(|masm, scratch| {
-                    masm.asm
-                        .mov_ir(scratch.writable(), I::i64(offset), OperandSize::S64);
-                    masm.asm.add_rrr(scratch.inner(), base, dst, size);
-                    Ok(())
-                })?;
-            }
-        }
-        Ok(())
+        self.add_ir(dst, base, I::i64(offset), size)
     }
 
     fn pop(&mut self, dst: WritableReg, size: OperandSize) -> Result<()> {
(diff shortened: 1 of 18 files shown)


winch/codegen/src/isa/aarch64/masm.rs
```diff
@@ -402,8 +448,17 @@ impl Masm for MacroAssembler {
...
+            Some(imm12) => self.asm.add_ir(imm12, base, dst, size),
+            None => {
+                self.with_scratch::<IntScratch, _>(|masm, scratch| {
+                    masm.asm
+                        .mov_ir(scratch.writable(), I::i64(offset), OperandSize::S64);
+                    masm.asm.add_rrr(scratch.inner(), base, dst, size);
+                    Ok(())
+                })?;
+            }
+        }
```
alexcrichton: Would it make sense to have `add_ir` methods on `masm` which encapsulate this pattern? (it looks similar below in `add` for example)

winch/codegen/src/isa/aarch64/regs.rs
```diff
@@ -26,6 +27,7 @@ pub(crate) const fn ip0() -> Reg {
     xreg(16)
 }
 
+// TODO: delete this.
 /// Alias to the IP0 register.
 pub(crate) const fn scratch() -> Reg {
```
alexcrichton: Would it make sense to tag this as `#[deprecated]` and annotate the remaining uses with `#[expect(deprecated, reason = "...")]`?

alexcrichton (approved):
This all look sound and good to me, definitely makes handling scratch registers less brittle!

winch/codegen/src/isa/aarch64/masm.rs
saulecabrera: Done in https://github.com/bytecodealliance/wasmtime/pull/10989/commits/3508ec7bcaf7b3b4bbad8e3eb8fe33a984ec53c6

winch/codegen/src/isa/aarch64/regs.rs
saulecabrera: Yeah, great suggestion, added in https://github.com/bytecodealliance/wasmtime/pull/10989/commits/3508ec7bcaf7b3b4bbad8e3eb8fe33a984ec53c6

PR #10990, opened 2025-06-09
Allow embedders to override / define restricted headers

fibonacci1729:
This PR allows embeddings to wholesale override / define what headers for an outgoing request are acceptable.


crates/wasi-http/src/types.rs
```diff
@@ -121,8 +121,20 @@ pub trait WasiHttpView: IoView {
     }
 
     /// Whether a given header should be considered forbidden and not allowed.
-    fn is_forbidden_header(&mut self, _name: &HeaderName) -> bool {
-        false
+    fn is_forbidden_header(&mut self, name: &HeaderName) -> bool {
```
lann, line 124: While this behavior better matches the method doc comment it _is_ a breaking change, and in the unfortunate compiler-won't-help-you sense. :thinking:

crates/wasi-http/src/types.rs
```diff
@@ -121,8 +121,20 @@ pub trait WasiHttpView: IoView {
...
+        static FORBIDDEN_HEADERS: [HeaderName; 10] = [
+            hyper::header::CONNECTION,
+            HeaderName::from_static("keep-alive"),
+            hyper::header::PROXY_AUTHENTICATE,
+            hyper::header::PROXY_AUTHORIZATION,
+            HeaderName::from_static("proxy-connection"),
+            hyper::header::TE,
```
lann, line 131: Given that the immediate goal is just to allow `TE` maybe we should just propose removing this line?

tests/spec_testsuite
alexcrichton, line 1: I suspect this (and `wasi-common` below) were unintentional?

crates/wasi-http/src/types.rs
lann: OK not "just": it would probably make sense to also enforce `trailers` as the only acceptable value, which is actually pretty sensible anyway since it is the only value that is allowed across http versions today.

fibonacci1729:
Closing this to pursue another solution that isn't as surprising a breaking change.

PR #10991, opened 2025-06-09
x64: Migrate `XmmUnaryRmRImm{,Vex}` to the new assembler

alexcrichton:
This required definitions for `{v,}pshuf{lw,hw,d}` as well as reusing existing definitions for `vround*` instructions. While no other changes were needed in the assembler itself there was one minor ISLE change: the `vrounds{s,d}` instructions have a register operand that is used to copy to the upper bits of the result and this was not previously modeled in ISLE. Instead during emission the destination had its upper bits preserved, but this has proven to be problematic in the past where the upper bits being carried from previous instructions can cause unintended data dependencies and drastically slowing down tight loops. This commit fills in this operand with `xmm_zero` which means the upper bits will now be zeroed and any data dependencies from before will be severed.


cfallin:
(Handing off review to @abrown if you don't mind -- I'm traveling this week at a conference)

PR #10992, opened 2025-06-09
stack-switching: fix continuation chain walking for backtraces

posborne:
A refactor during the stack switching runtime changes introduced a defect not present in the base stack switching code.  The `chunks` iterator call ends up skipping elements and the `windows` call would require pulling in something like `itertools::zip_longest` to work, so I just restored an impl closer to the original (but with more fine-grained unsafe calls).

This code has coverage but only in the final round of stack switching changes.

---

Original Discussion: https://github.com/bytecodealliance/wasmtime/pull/10388#discussion_r2114747895.  I didn't look closely enough when I took on this approach and w/o testing this slipped through (though can't really be hit yet).

Failing Test with current main + cranelift changes + tests: https://github.com/posborne/wasmtime/actions/runs/15538863221/job/43744530801#step:19:1165

CC @frank-emrich, @dhil

### DIFF
--- a/crates/wasmtime/src/runtime/vm/traphandlers/backtrace.rs
+++ b/crates/wasmtime/src/runtime/vm/traphandlers/backtrace.rs
@@ -262,24 +262,17 @@ impl Backtrace {
         // are continuations, due to the initial stack having one, too.
         assert_eq!(stack_limits_vec.len(), continuations_vec.len() + 1);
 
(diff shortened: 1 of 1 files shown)


fitzgen (approved):

dhil (approved):
LGTM. Though, it may be good to have @frank-emrich give it a look too.

PR #10998, opened 2025-06-10
winch: Use `Masm::with_scratch` in ISA-agnostic code

saulecabrera:
This commit is a follow-up to
https://github.com/bytecodealliance/wasmtime/pull/10989; it migrates all the uses of the scratch register to use `Masm::with_scratch`. This commit also updates the tests that can fail / are ignored, since many of the bugs were introduced by accidental clobbers to scratch registers.

Even though this change doesn't introduce any functional changes, the disassembly changes in aarch64 are due to the usage of x17 as an extra scratch registers.

### DIFF
--- a/tests/disas/winch/x64/br_if/save_state_before_br_emission.wat
+++ b/tests/disas/winch/x64/br_if/save_state_before_br_emission.wat
@@ -44,7 +44,7 @@
 ;;       movq    0x10(%r11), %r11
 ;;       addq    $0x58, %r11
 ;;       cmpq    %rsp, %r11
-;;       ja      0x131
+;;       ja      0x12d
 ;;   1c: movq    %rsi, %r14
 ;;       subq    $0x40, %rsp
 ;;       movq    %rsi, 0x38(%rsp)
@@ -67,7 +67,7 @@
 ;;       movsd   (%rsp), %xmm0
 ;;       addq    $8, %rsp
 ;;       subq    $4, %rsp
-;;       movss   0x8e(%rip), %xmm15
+;;       movss   0x86(%rip), %xmm15
 ;;       movss   %xmm15, (%rsp)
 ;;       testl   %eax, %eax
 ;;       je      0xcd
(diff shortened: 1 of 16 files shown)


tests/disas/winch/x64/br_if/save_state_before_br_emission.wat
```diff
@@ -81,25 +81,23 @@
...
 ;;       addq    $8, %rsp
-;;       movsd   0x45(%rip), %xmm0
+;;       movsd   0x3d(%rip), %xmm0
 ;;       subq    $4, %rsp
-;;       movss   0x2d(%rip), %xmm15
+;;       movss   0x25(%rip), %xmm15
 ;;       movss   %xmm15, (%rsp)
 ;;       movq    0xc(%rsp), %rax
-;;       movss   (%rsp), %xmm15
+;;       movl    (%rsp), %r11d
```
saulecabrera: Oh this is wrong, it's using the wrong scratch type. I'll take a look.

saulecabrera: Fixed in https://github.com/bytecodealliance/wasmtime/pull/10998/commits/f6529682aae214d898e303ccbd46a4794ef30415

winch/codegen/src/masm.rs
```diff
@@ -15,6 +15,28 @@ use wasmtime_environ::PtrSize;
...
+            }) => $self.with_scratch::<$crate::masm::IntScratch, _>($callback),
+            wasmtime_environ::WasmValType::F32
+            | wasmtime_environ::WasmValType::F64
+            | wasmtime_environ::WasmValType::V128 => {
+                $self.with_scratch::<$crate::masm::FloatScratch, _>($callback)
+            }
+            _ => unimplemented!(),
+        }
+    };
+}
```
alexcrichton: WDYT about something like:

```rust
trait MacroAssembler {
    fn with_scratch_ty(&mut self, reg_class: RegClass);
    fn with_scratch<T: ScratchType>(...) { self.with_scratch_ty(T::reg_class(), ...) }
}
```

to avoid macros? That way assemblers would implement `with_scratch_ty` under the hood and `with_scratch` as-is today would become a convenience over that method

alexcrichton (approved):
Very nice 👍

winch/codegen/src/masm.rs
saulecabrera: Yeah, makes sense to me, I'll update.

saulecabrera: Got rid of the macro, but ended with a slightly different version than what you proposed, mainly because I think it make sense to have the wrapper derive the register class and delegate to `Masm::with_scratch<T,_>`, which is what I was trying to solve for.

alexcrichton: 👍

PR #11003, opened 2025-06-10
cranelift: stack-switching support

posborne:
These changes pull in the cranelift changes from #10177 with some additional stacked changes to resolve conflicts, align with previous changes in the stack-switching series, and address feedback items which were raised on previous iterations of the PR (but mostly not changing anything of significant substance).  Tracking Issue: #10248.

The `stack-switching` feature flag is retained and used minimally in this changeset in order to avoid compilation problems, but not really used beyond that.  There is at least one item in the tracking issue already related to likely finding a way to drop these compilation flags in most places but I think it is worth deferring that here as it will required touching code more broadly.

CC @frank-emrich @dhil


crates/cranelift/src/stack_switching/instructions.rs
```diff
@@ -0,0 +1,1910 @@
...
+            env: &mut crate::func_environ::FuncEnvironment<'a>,
+            builder: &mut FunctionBuilder,
+            values: &[ir::Value],
+            allow_smaller: bool,
+        ) {
+            let store_count = builder
+                .ins()
+                .iconst(I32, i64::try_from(values.len()).unwrap());
+
+            // TODO(posborne): allow_smaller only used in debug_assert!
```
posborne: @frank-emrich @dhil I put this in here as passing as the only use of `allow_smaller` here is in the debug assertions which seemed like a potential defect, though I didn't do a deep dive into the full usage.

cranelift/codegen/src/ir/function.rs
```diff
@@ -255,6 +255,13 @@ impl FunctionStencil {
         self.get_dyn_scale(dyn_ty)
     }
 
+    /// Find the data for the given stack slot
+    pub fn get_stack_slot_data(&self, stack_slot: StackSlot) -> &StackSlotData {
+        self.sized_stack_slots
+            .get(stack_slot)
+            .expect("undeclared stack slot: {stack_slot}")
+    }
```
fitzgen: This is fine to add, but fwiw the `sized_stack_slots` member is `pub` and implements `Index<StackSlot>` so you can always do

```rust
&func.sized_stack_slots[slot]
```

crates/cranelift/src/stack_switching/mod.rs
fitzgen, line 1: If we move the `stack_switching` module from the top-level to being a submodule of `func_environ` (similar to the `func_environ::gc` module) then we don't need to make all those members of `FuncEnvironment` be `pub(crate)`, which is a little thing but a nice thing (especially if we consider all the future evolutions of this code and any changes to `FuncEnvironment` members that will also need or not need to become `pub(crate)`).

crates/cranelift/src/func_environ.rs
```diff
@@ -1818,22 +1834,41 @@ impl FuncEnvironment<'_> {
         let mut pos = builder.cursor();
         let table = self.table(table_index);
         let ty = table.ref_type.heap_type;
+        let vmctx = self.vmctx_val(&mut pos);
+        let index_type = table.idx_type;
+        let delta = self.cast_index_to_i64(&mut builder.cursor(), delta, index_type);
+        let table_index_arg = builder.ins().iconst(I32, table_index.as_u32() as i64);
+        let mut args = vec![vmctx, table_index_arg, delta];
```
fitzgen: Nice to avoid the allocation when possible:

```suggestion
        let mut args: SmallVec<[_; 6]> = smallvec![vmctx, table_index_arg, delta];
```

(I think I counted the max args correctly but feel free to fix the inline capacity if I messed up)

crates/cranelift/src/func_environ.rs
```diff
@@ -1818,22 +1834,41 @@ impl FuncEnvironment<'_> {
...
+                        init_value,
+                    );
+                    args.extend_from_slice(&[contref, revision]);
+                    stack_switching::builtins::table_grow_cont_obj(self, &mut builder.func)?
+                }
 
-        let vmctx = self.vmctx_val(&mut pos);
+                _ => panic!("unsupported table type."),
+            }
+        };
```
fitzgen: Can we fold the `is_vmgcref_type()` branch into the `match` here? Something like this:

```rust
match ty.top() {
    WasmHeapTopType::Any | WasmHeapTopType::Extern => ...,
    WasmHeapTopType::Func => ...,
    ...
}
```

This should allow us to exhaustively match and remove the `_ => panic!(...)` arm.

crates/cranelift/src/func_environ.rs
```diff
@@ -1929,22 +1977,35 @@ impl FuncEnvironment<'_> {
...
+                        .table_fill_func_ref(&mut builder.func)
+                }
+                WasmHeapTopType::Cont => {
+                    let (revision, contref) =
+                        stack_switching::fatpointer::deconstruct(self, &mut builder.cursor(), val);
+                    args.extend_from_slice(&[contref, revision]);
+                    stack_switching::builtins::table_fill_cont_obj(self, &mut builder.func)?
+                }
+                _ => panic!("unsupported table type"),
+            }
```
fitzgen: Similar here regarding exhaustively matching on the top type.

crates/cranelift/src/func_environ.rs
```diff
@@ -1929,22 +1977,35 @@ impl FuncEnvironment<'_> {
...
-        let index_type = table.idx_type;
-        let dst = self.cast_index_to_i64(&mut pos, dst, index_type);
-        let len = self.cast_index_to_i64(&mut pos, len, index_type);
         let ty = table.ref_type.heap_type;
+        let vmctx = self.vmctx_val(&mut pos);
+        let index_type = table.idx_type;
+        let table_index_arg = builder.ins().iconst(I32, table_index.as_u32() as i64);
+        let dst = self.cast_index_to_i64(&mut builder.cursor(), dst, index_type);
+        let len = self.cast_index_to_i64(&mut builder.cursor(), len, index_type);
+        let mut args = vec![vmctx, table_index_arg, dst];
```
fitzgen: And this could be a `smallvec` too.

crates/cranelift/src/func_environ.rs
```diff
@@ -2274,9 +2338,19 @@ impl FuncEnvironment<'_> {
...
+            ty if ty == stack_switching::fatpointer::POINTER_TYPE => {
+                let (_revision, contref) =
+                    stack_switching::fatpointer::deconstruct(self, &mut pos, value);
+                pos.ins()
+                    .icmp_imm(cranelift_codegen::ir::condcodes::IntCC::Equal, contref, 0)
+            }
+            _ => pos
+                .ins()
+                .icmp_imm(cranelift_codegen::ir::condcodes::IntCC::Equal, value, 0),
+        };
```
fitzgen: Instead of looking at the value's CLIF type, can you modify `translate_ref_is_null` to pass in the actual wasm type as an argument? We already map multiple Wasm types onto the same CLIF type, it just so happens that is not currently the case for the particular type used for pointers to stacks, but this code as-written is pretty fragile to rely on that incidental property.

crates/cranelift/src/func_environ.rs
```diff
@@ -3209,6 +3283,133 @@ impl FuncEnvironment<'_> {
...
+            self,
+            builder,
+            type_index,
+            contobj,
+            resume_args,
+            resumetable,
+        )
+    }
+
+    #[allow(dead_code, reason = "TODO")]
```
fitzgen: TODO

crates/cranelift/src/stack_switching/fatpointer.rs
```diff
@@ -0,0 +1,71 @@
...
+
+/// The Cranelfift type used to represent all of the following:
+/// - wasm values of type `(ref null $ct)` and `(ref $ct)`
+/// - equivalently: runtime values of type `Option<VMContObj>` and `VMContObj`
+/// Note that a `VMContObj` is a fat pointer
+/// consisting of a pointer to `VMContRef` and a 64 bit sequence
+/// counter.
+/// We represent this here as a 128bit value, with the same representation as
+/// `core::mem::transmute::<i128, VMContObj>`.
+pub const POINTER_TYPE: ir::Type = ir::types::I128;
```
fitzgen: I'd prefer defining a type containing the fat pointer's parts, something like

```rust
pub struct ContObj {
    revision: ir::Value,
    contref: ir::Value,
}

impl ContObj {
    /// ...
    pub(crate) new(...) -> Self { /* ... */ }

    /// getters for the parts...
}
```

rather than trying to use `ir::types::I128` here.

Then we could also add `ContObj::load` and `ContObj::store` helper methods as needed.

This lets us avoid the sketchy `transmute` alluded to above (which I don't believe actually exists anyways, because we pass `VMContObj`s by parts to libcalls already) and we also don't need to do all the packing and unpacking and extending/reducing that packing into and unpacing from `ir::types::I128` requires either.

The only reason `ir::types::I128` exists is for ABI compatibility with LLVM code, which isn't a concern we have here.

crates/cranelift/src/stack_switching/instructions.rs
```diff
@@ -0,0 +1,1910 @@
...
+use cranelift_codegen::ir::condcodes::*;
+use cranelift_codegen::ir::types::*;
+use cranelift_codegen::ir::{self, MemFlags};
+use cranelift_codegen::ir::{Block, BlockCall, InstBuilder, JumpTableData};
+use cranelift_frontend::FunctionBuilder;
+use wasmtime_environ::{PtrSize, TagIndex, TypeIndex, WasmResult, WasmValType};
+
+// TODO(frank-emrich) This is the size for x64 Linux. Once we support different
+// platforms for stack switching, must select appropriate value for target.
+pub const CONTROL_CONTEXT_SIZE: usize = 24;
```
fitzgen: Can we put this into a helper function something like this:

```rust
fn control_context_size(triple: &target_lexicon::Triple) -> WasmResult<usize> {
    match (triple.architecture, triple.operating_system) {
        (target_lexicon::Architecture::X86_64, target_lexicon::OperatingSystem::Linux) => Ok(24),
        _ => Err(wasm_unsupported!("stack switching is not supported on {triple}")),
    }
}
```

This will future proof the code and give us a nice place to slot in updates when we expand platform support.

crates/cranelift/src/stack_switching/instructions.rs
```diff
@@ -0,0 +1,1910 @@
...
+
+        /// Adding this (statically) known offset gets us the overall address.
+        offset: i32,
+
+        /// The type parameter T is never used in the fields above. We still
+        /// want to have it for consistency with
+        /// `wasmtime_environ::Vector` and to use it in the associated
+        /// functions.
+        phantom: PhantomData<T>,
+    }
```
fitzgen: My suggestion for a `ContObj` type instead of `ir::types::I128` is basically to do the same as what is happening here.

fitzgen: This is not a `VMHostArray`, it is a pointer to a `VMHostArray`, right? Can we update the name to reflect that? `VMHostArrayPtr` or `VMHostArrayRef` perhaps?

Also, I don't think we need to separate the static offset and dynamic base address because nothing is (for example) determining whether we need to emit explicit bounds checks or can rely on guard pages for these accesses, so we can just have a single `address: ir::Value` field and trust Cranelift's instruction selection choose the best addressing mode it can given the input IR.

crates/cranelift/src/translate/code_translator.rs
```diff
@@ -2898,54 +2898,150 @@ pub fn translate_operator(
...
 
-        Operator::ContNew { cont_type_index: _ } => {
-            // TODO(10248) This is added in a follow-up PR
-            return Err(wasmtime_environ::WasmError::Unsupported(
-                "codegen for stack switching instructions not implemented, yet".to_string(),
-            ));
+        Operator::ContNew { cont_type_index } => {
+            let cont_type_index = TypeIndex::from_u32(*cont_type_index);
+            let arg_types = environ.continuation_arguments(cont_type_index).to_vec();
+            let result_types = environ.continuation_returns(cont_type_index).to_vec();
```
fitzgen: These should be able to be `SmallVec`s too.

crates/cranelift/src/translate/code_translator.rs
```diff
@@ -2898,54 +2898,150 @@ pub fn translate_operator(
...
+                // We can't bind GC objects using cont.bind at the moment: We
+                // don't have the necessary infrastructure to traverse the
+                // buffers used by cont.bind when looking for GC roots. Thus,
+                // this crude check ensures that these buffers can never contain
+                // GC roots to begin with.
+                if arg_type.is_vmgcref_type_and_not_i31() {
+                    return Err(wasmtime_environ::WasmError::Unsupported(
+                        "cont.bind does not support GC types at the moment".into(),
+                    ));
+                }
```
fitzgen, line 2977: Can you make sure there is an item for this in the tracking issue?

crates/cranelift/src/translate/code_translator.rs
```diff
@@ -2898,54 +2898,150 @@ pub fn translate_operator(
...
-                "codegen for stack switching instructions not implemented, yet".to_string(),
-            ));
+        Operator::Suspend { tag_index } => {
+            let tag_index = TagIndex::from_u32(*tag_index);
+            let param_types = environ.tag_params(tag_index).to_vec();
+            let return_types: Vec<_> = environ
+                .tag_returns(tag_index)
+                .iter()
+                .map(|ty| crate::value_type(environ.isa(), *ty))
+                .collect();
```
fitzgen: `SmallVec` here too.

crates/cranelift/src/translate/code_translator.rs
```diff
@@ -2898,54 +2898,150 @@ pub fn translate_operator(
...
-            resume_table: _,
+            cont_type_index,
+            resume_table,
         } => {
-            // TODO(10248) This is added in a follow-up PR
-            return Err(wasmtime_environ::WasmError::Unsupported(
-                "codegen for stack switching instructions not implemented, yet".to_string(),
-            ));
+            // We translate the block indices in the resumetable to actual Blocks.
+            let mut resumetable = vec![];
```
fitzgen: Can we name the local something else? Having both a `resume_table` and `resumetable` variable in scope at the same time is pretty confusing. Maybe we can have `wasm_resume_table` and `clif_resume_table`?

crates/cranelift/src/translate/code_translator.rs
```diff
@@ -2898,54 +2898,150 @@ pub fn translate_operator(
...
             cont_type_index: _,
             tag_index: _,
             resume_table: _,
-        } => {
-            // TODO(10248) This depends on exception handling
-            return Err(wasmtime_environ::WasmError::Unsupported(
-                "resume.throw instructions not supported, yet".to_string(),
-            ));
-        }
+        } => todo!("unimplemented stack switching instruction"),
```
fitzgen: Can we leave this as an error instead of a panic?

crates/cranelift/src/translate/code_translator.rs
```diff
@@ -2898,54 +2898,150 @@ pub fn translate_operator(
...
+            tag_index,
         } => {
-            // TODO(10248) This is added in a follow-up PR
-            return Err(wasmtime_environ::WasmError::Unsupported(
-                "codegen for stack switching instructions not implemented, yet".to_string(),
-            ));
+            // Arguments of the continuation we are going to switch to
+            let continuation_argument_types = environ
+                .continuation_arguments(TypeIndex::from_u32(*cont_type_index))
+                .to_vec();
```
fitzgen: `SmallVec`

crates/cranelift/src/translate/code_translator.rs
```diff
@@ -2898,54 +2898,150 @@ pub fn translate_operator(
...
+                        .expect("Only supporting module type indices on switch for now");
+
+                    environ
+                        .continuation_arguments(TypeIndex::from_u32(mti.as_u32()))
+                        .iter()
+                        .map(|ty| crate::value_type(environ.isa(), *ty))
+                        .collect()
+                }
+                _ => panic!("Invalid type on switch"),
+            };
```
fitzgen: `SmallVec`

fitzgen (commented):
Thanks! Lots of comments below but for the most part these should be pretty straightforward to fix.

crates/cranelift/src/stack_switching/fatpointer.rs
```diff
@@ -0,0 +1,71 @@
+use cranelift_codegen::ir;
+use cranelift_codegen::ir::InstBuilder;
+use cranelift_codegen::ir::types::I64;
+
+/// The Cranelfift type used to represent all of the following:
```
dhil:
```suggestion
/// The Cranelift type used to represent all of the following:
```

crates/cranelift/src/stack_switching/instructions.rs
dhil: I think this code is harmless. It seems to be doing some confidence testing that the byte-encoding of the values will fit into the allocated space. If I read the above comment correctly (and the code below) then each encoded element is stored as if it had type `T`, so for elements of type `S` such that `|S| < |T|`, then there will be some padding between each element. 

I am not sure `allow_smaller` is useful, perhaps we can just have `debug_assert!(size <= std::mem::size_of::<T>());`, or perhaps we can dispense with this particular confidence checking altogether.

crates/cranelift/src/stack_switching/mod.rs
dhil: Yes, good shout! I think I am to blame for the current state of affairs.

crates/cranelift/src/func_environ.rs
posborne: <strike>I think 7 here</strike> (6 is right with the branching), will do a commit with the smallvec changes.

crates/cranelift/src/translate/code_translator.rs
posborne: For args and returns, I guessed at a smallvec max of 8; I didn't see anywhere where we are bound to this, but smallvec will still fall back on heap allocations if we go beyond this.

crates/cranelift/src/translate/code_translator.rs
posborne: Added to tracking.

crates/cranelift/src/stack_switching/instructions.rs
posborne: For now, I'll keep a debug_assert! in place to check `size <= std::mem::size_of::<T>()`.

crates/cranelift/src/func_environ.rs
posborne: Tracing through this, just wanted to confirm that in order to do this we'll need to update wasm-tools/wasmparser to pass through the heap type information (similar to RefNull)?  Or am I off in the weeds in tracking through the changes required?

fitzgen: Argh. I forgot that `ref.is_null` doesn't have a type parameter. I'll see how hard it would be to add type information that can be used here (and elsewhere).

fitzgen: https://github.com/bytecodealliance/wasmtime/pull/11030 passes the wasm ref type through to `translate_ref_is_null`. A couple things need to happen before that can land tho, see the PR description.

crates/cranelift/src/func_environ.rs
posborne: Just removed this for now as there's already an error at where this callsite will eventually be and its addition should be obvious.  Probably to follow the work that @cfallin does to land other exception bits.

I confirmed the tracking issues already has an item to track `resume.throw` addition (outside of the initial impl).

posborne:
Pushed most of the updates but still need to make the suggested updates around the fat pointer stuff and resolve conflicts with upstream.

crates/cranelift/src/func_environ.rs
posborne: Updated to use the wasm type as part of merging upstream back into the branch.

crates/cranelift/src/stack_switching/mod.rs
posborne: Implemented this now.

crates/cranelift/src/stack_switching/fatpointer.rs
posborne: Discussed this change a bit more with @fitzgen and played around with an impl.  Currently, we are thinking that to facilitate having a single wasm value be modeled by multiple ir values (as we want here), we'll need to do some updates to the code translator to support having a single translated wasm value map to multiple ir values.

This will be a larger change, so I am proposing that we move forward with changing the stack-switching code to be updated to use that mechanism when introduced in a later change.

posborne:
@fitzgen I believe this is ready for review again, let me know if there's pieces I missed in my previous updates.  As noted, I chose to defer fatpointer changes for now to be potentially addressed as part of (or following) changes to the translation stack.

crates/cranelift/src/func_environ.rs
```diff
@@ -138,7 +139,7 @@ pub struct FuncEnvironment<'module_environment> {
     pcc_vmctx_memtype: Option<ir::MemoryType>,
 
     /// Caches of signatures for builtin functions.
-    builtin_functions: BuiltinFunctions,
+    pub(crate) builtin_functions: BuiltinFunctions,
```
fitzgen: This shouldn't need to be `pub(crate)` after the `stack_switching` module moved to `func_environ::stack_switching`, right?

crates/cranelift/src/func_environ.rs
```diff
@@ -174,6 +175,16 @@ pub struct FuncEnvironment<'module_environment> {
...
+
+    /// Used by the stack switching feature. If set, we have a allocated a
+    /// slot on this function's stack to be used for the
+    /// current stack's `handler_list` field.
+    pub(crate) stack_switching_handler_list_buffer: Option<ir::StackSlot>,
+
+    /// Used by the stack switching feature. If set, we have a allocated a
+    /// slot on this function's stack to be used for the
+    /// current continuation's `values` field.
+    pub(crate) stack_switching_values_buffer: Option<ir::StackSlot>,
```
fitzgen: Ditto

crates/cranelift/src/func_environ.rs
```diff
@@ -3219,6 +3285,119 @@ impl FuncEnvironment<'_> {
...
+    ) -> WasmResult<Vec<ir::Value>> {
+        stack_switching::instructions::translate_switch(
+            self,
+            builder,
+            tag_index,
+            contobj,
+            switch_args,
+            return_types,
+        )
+    }
```
fitzgen: I think it makes sense to [add `disas` tests](https://github.com/bytecodealliance/wasmtime/tree/main/tests/disas) for these operations in `tests/disas/stack-switching/*`. Should be as easy as writing a couple `.wat` files and then running `WASMTIME_TEST_BLESS=1 cargo test --test disas` to initialize the golden output.

Doesn't need to be in this PR, but if not in this PR, then should be added to the tracking issue.

crates/cranelift/src/func_environ/stack_switching/fatpointer.rs
```diff
@@ -0,0 +1,71 @@
...
+            let contref = pos.ins().ushr_imm(lsbs, pad_bits as i64);
+            (msbs, contref)
+        }
+    };
+    let contref = if env.pointer_type().bits() < I64.bits() {
+        pos.ins().ireduce(env.pointer_type(), contref)
+    } else {
+        contref
+    };
+    (revision_counter, contref)
```
fitzgen: Can this just be something like

```suggestion
    let ptr_ty = env.pointer_type();
    assert!(ptr_ty.bits() <= 64);
    
    let contref = pos.ins().ireduce(ptr_ty, contobj);
    
    let shifted = pos.ins().ushr_imm(contobj, 64);
    let revision_counter = pos.ins().ireduce(it::types::I64, shifted);
    
    (revision_counter, contref)
```

?

That is, we define the fat pointer as 128 bits where the upper 64 are the revision counter and the bottom `sizeof(pointer)` bits are the pointer to the continuation. This way I don't think we ever have to branch on endianness or pointer width.

crates/cranelift/src/func_environ/stack_switching/fatpointer.rs
```diff
@@ -0,0 +1,71 @@
...
+            let pad_bits = 64 - env.pointer_type().bits();
+            let lsbs = pos.ins().ishl_imm(contref_addr, pad_bits as i64);
+            (revision_counter, lsbs)
+        }
+    };
+
+    let lsbs = pos.ins().uextend(ir::types::I128, lsbs);
+    let msbs = pos.ins().uextend(ir::types::I128, msbs);
+    let msbs = pos.ins().ishl_imm(msbs, 64);
+    pos.ins().bor(lsbs, msbs)
```
fitzgen: And then construction would become something like this:

```suggestion
    assert!(env.pointer_type().bits() <= 64);
    let contref_addr = pos.ins().uextend(ir::types::I28, contref_addr);

    let revision_counter = pos.ins().uextend(ir::types::I128, revision_counter);
    let shifted_counter = pos.ins().ishl_imm(revision_counter, 64);
    
    pos.ins().bor(shifted_counter, contref_addr)
```

Note: if you want to switch the order of the contref pointer and the counter in the fat pointer, that's fine (I think I unwittingly switched it from what is in the code now), but the important point is that we should be able do all of this fat pointer construction and destruction without branching on endianness and pointer width.

crates/cranelift/src/func_environ/stack_switching/fatpointer.rs
posborne: Yeah, this makes sense to me, will adopt this approach (with a quick round of testing) until we can split into individual ir values.

posborne:
@fitzgen I am hunting down one regression with the latest, probably around the changes to table ops based on wasm types (though I haven't found a smoking gun yet).  These tests passed prior to some of the latest changes on this branch and the test is doing a table.grow of continuations followed by a resume of the first added to the table which traps on a null reference.

posborne:
Pulled in and resolved conflicts (some pretty mild ones from https://github.com/bytecodealliance/wasmtime/pull/11216.  @fitzgen Review would be appreciated; I think in our last convos I had played around with a different fatpointer impl but I'm fine with this as-is (possibly to revisit later on).

cranelift/codegen/src/ir/function.rs
```diff
@@ -255,6 +255,11 @@ impl FunctionStencil {
         self.get_dyn_scale(dyn_ty)
     }
 
+    /// Find the data for the given stack slot
+    pub fn get_stack_slot_data(&self, stack_slot: StackSlot) -> &StackSlotData {
+        &self.sized_stack_slots[stack_slot]
+    }
```
bjorn3: This field is already public.

cranelift/codegen/src/isa/x64/abi.rs
```diff
@@ -194,6 +200,7 @@ impl ABIMachineSpec for X64ABIMachineSpec {
             if param.value_type.bits() > 64
                 && !(param.value_type.is_vector() || param.value_type.is_float())
                 && !flags.enable_llvm_abi_extensions()
+                && !is_tail
```
bjorn3, line 203: Why is this necessary? Wasm doesn't have 128bit integers.

posborne: At least in this iteration of the changes, the stack switching code is using an i128 for it's vmcontobj fat pointer (consisting of vmcontref pointer and revision).  @fitzgen and I have discussed a bit about how we can get rid of this and just have two ir values through the transformation, but it will require some additional changes to how we model the relationship between wasm and clif values to support having one of the former map to two distinct ir values.

cranelift/codegen/src/ir/function.rs
posborne: I'll remove this in favor of using the pub field directly; probably just a miss on the visibility when the wasmfx contributors originally authored these changes (looks like field has been pub for a long time).

crates/cranelift/src/func_environ.rs
```diff
@@ -173,6 +174,16 @@ pub struct FuncEnvironment<'module_environment> {
     /// always present even if this is a "leaf" function, as we have to call
     /// into the host to trap when signal handlers are disabled.
     pub(crate) stack_limit_at_function_entry: Option<ir::GlobalValue>,
+
+    /// Used by the stack switching feature. If set, we have a allocated a
+    /// slot on this function's stack to be used for the
+    /// current stack's `handler_list` field.
+    pub(crate) stack_switching_handler_list_buffer: Option<ir::StackSlot>,
```
fitzgen: This field is only accessed in a submodule, so it does not need to be `pub(crate)`.

crates/cranelift/src/func_environ.rs
```diff
@@ -173,6 +174,16 @@ pub struct FuncEnvironment<'module_environment> {
...
+
+    /// Used by the stack switching feature. If set, we have a allocated a
+    /// slot on this function's stack to be used for the
+    /// current continuation's `values` field.
+    pub stack_switching_values_buffer: Option<ir::StackSlot>,
```
fitzgen: Similarly, this is only accessed in submodules as well, and not across crate boundaries at all, so it does not need to be `pub`.

crates/environ/src/vmoffsets.rs
```diff
@@ -332,6 +338,23 @@ pub trait PtrSize {
...
+    // Offsets within `VMContObj`
+
+    /// Return the offset of `VMContObj::contref`
+    fn vmcontobj_contref(&self) -> u8 {
+        0
+    }
+
+    /// Return the offset of `VMContObj::revision`
+    fn vmcontobj_revision(&self) -> u8 {
+        8
```
fitzgen: To work across non-64-bit ISAs, which we support today via, for example, the 32-bit Pulley interpreter, this needs to be

```suggestion
        self.size()
```

crates/environ/src/vmoffsets.rs
```diff
@@ -332,6 +338,23 @@ pub trait PtrSize {
...
+    }
+
+    /// Return the size of `VMHostArray`.
+    fn size_of_vmcontobj(&self) -> u8 {
+        16
```
fitzgen: And then this should be something like


```suggestion
        u8::try_from(align(
            u32::from(self.vmcontobj_revision())
                + u32::try_from(core::mem::size_of::<u64>()).unwrap(),
            u32::from(self.size()),
        ))
        .unwrap()
```

crates/wasmtime/src/runtime/vm/stack_switching.rs
```diff
@@ -39,15 +39,11 @@ pub use stack::*;
 /// For performance reasons, the VMContRef at the bottom of this chain
 /// (i.e., the one pointed to by the VMContObj) has a pointer to the
 /// other end of the chain (i.e., its last ancestor).
-// FIXME(frank-emrich) Does this actually need to be 16-byte aligned any
-// more? Now that we use I128 on the Cranelift side (see
-// [wasmtime_cranelift::stack_switching::fatpointer::pointer_type]), it
-// should be fine to use the natural alignment of the type.
 #[repr(C, align(16))]
```
fitzgen: I don't think this align(16) needs to be here anymore, as described in the deleted `FIXME` comment? But if we are leaving it, then we should have a follow up item to investigate in the meta issue and also should adjust the `size_of_vmcontobj` method above to align to 16 instead of the natural alignment that my suggestion comment sketched out.

fitzgen (approved):
LGTM! Couple final nitpicks before this lands, but once those are addressed, feel free to add it to the merge queue or ask me to do that (I forget if you have those permissions or not).

Thanks for sticking through with this one!

crates/wasmtime/src/runtime/vm/stack_switching.rs
posborne: I wasn't sure if we wanted to pad out this struct to always be packed as 128 bits or as two words.  I think two words makes more sense and aligns with what you are saying but means that the cranelift generated code will definitely need to be updated to be able to support 32-bit target architectures (probably along with changes discussed before to change how we do the fatpointer handling).

I'll land changes after thinking this a bit more and update the meta-issue with any notes on 32-bit targets.

fitzgen:
@posborne is this ready for review now?

posborne:
@fitzgen Yes, should be ready for review -- I'll just push a fixup for the CI issues.  I sent a side note in zulip but I'll repost here:

> Ok, I got around to updating the branch with (hopefully) a final set of changes. Since we had been going back-and-forth on VMContObj alignment a bit I did end up changing this to natural alignment with revision now being defined as a usize.
> 
> Along with that I updated the cranelift fatpointer code and libcalls to work with the usize and to use an appropriate fat pointer container ir type for two words -- for this I had to plumb through a size type for that layer matching what exists for components (https://github.com/bytecodealliance/wasmtime/pull/11003/commits/078002aa024fe825f148f9dbcf864049bcf708ba).
> 
> Since the changes were a bit more substantive I will wait for a fresh review of that -- there were some merge conflicts related to the other changes in func_environ but they were pretty straightforward to address.
> 
> In parallel, I'll work on a rebase of the test changes to get a bit of extra validation as there isn't a lot in tree to cover things end-to-end on its own.

posborne:
@fitzgen I did run the test changes now -- a subset of those tests are now failing as it looks like there are some codepaths hit due to upstream changes where we end up calling `Val::null_top` which currently panics for continuation types.

posborne:
> @fitzgen I did run the test changes now -- a subset of those tests are now failing as it looks like there are some codepaths hit due to upstream changes where we end up calling Val::null_top which currently panics for continuation types.

To get tests going with the upstream stuff, I did end up adding a minimal `Val::ContRef` member which doesn't seem avoidable.  I tried to keep this as small and I could bundle it with the test changes, but it does appear to be required in order to have ref.null globals for continuations and some other cases.

crates/wasmtime/src/runtime/externals/global.rs
```diff
@@ -274,6 +274,16 @@ impl Global {
                     let new = new.as_ref();
                     definition.write_gc_ref(&mut store, new);
                 }
+                Val::ContRef(None) => {
+                    // Allow null continuation references for globals - these are just placeholders
+                    // Non-null references are not supported yet
```
fitzgen: This should still zero the global's definition, no?

crates/wasmtime/src/runtime/trampoline/global.rs
```diff
@@ -70,6 +70,16 @@ pub fn generate_global_export(
                 let new = new.as_ref();
                 global.write_gc_ref(&mut store, new);
             }
+            Val::ContRef(None) => {
+                // Allow null continuation references for trampoline globals - these are just placeholders
+                // Non-null references are not supported yet
```
fitzgen: And similar here

fitzgen (approved):
LGTM with the below addressed, thanks!

crates/wasmtime/src/runtime/externals/global.rs
posborne: Yeah, definitely.

crates/cranelift/src/func_environ/stack_switching/instructions.rs
```diff
@@ -302,8 +304,16 @@ pub(crate) mod stack_switching_helpers {
...
+            // TODO: hack around host pointer size being mixed up with target
+            let (align, entry_size) =
+                if core::any::TypeId::of::<T>() == core::any::TypeId::of::<*mut u8>() {
+                    (env.pointer_type().bytes() as u8, env.pointer_type().bytes())
+                } else {
+                    (
+                        u8::try_from(std::mem::align_of::<T>()).unwrap(),
+                        u32::try_from(std::mem::size_of::<T>()).unwrap(),
+                    )
+                };
```
fitzgen: This stuff should be a method on the `PtrSize` trait that we define in `vmoffsets.rs` (maybe `fn vm_host_array_layout(&self) -> (u8, u32)`) and ultimately we should have access to that method here via `env.offsets.ptr`.

We should definitely avoid any kind of `std::mem::{size,align}_of` calls, as that ends up being pretty fragile (doesn't work for `f64` across 32- and 64-bit architectures, for example, where the alignment is different despite size being the same).

fitzgen (changes requested):

crates/cranelift/src/func_environ/stack_switching/instructions.rs
fitzgen: To clarify a little more:

To avoid the `std::mem::{size,align}_of` calls, you may have to have different `PtrSize` methods for each concrete `T` that is used, and then bound `T` by `T: HostArrayLayout` and define

```rust
trait HostArrayLayout {
    fn host_array_layout<P: PtrSize>(p: &P) -> (u8, u32);
}
```

and implement that for each concreate `T` such that it calls the appropriate `PtrSize` method.

posborne: Makes sense, I'll take a look at making changes to revise VMHostArray and any other problematic parts I can find (or see if @dhil can help out).  I wanted to get a basic version of the commit up to get this kind of feedback as doing operations on the generic T here (from the original impl) seemed problematic for the reasons you describe.

posborne: Addressed in 9343ab99f2c07fb3f2a4dd778142422da9202d3f if my interpretation was correct.

fitzgen: Looks good, thanks!

PR #11008, opened 2025-06-10
x64: Migrate `xchg` to the new assembler

alexcrichton:
This required a custom visit function because printing operands has to happen in one order to match Capstone but allocating operations has to happen in another order to satisfy regalloc, so so one of the two has to be custom.

### DIFF
--- a/cranelift/assembler-x64/meta/src/instructions.rs
+++ b/cranelift/assembler-x64/meta/src/instructions.rs
@@ -2,6 +2,7 @@
 
 mod add;
 mod and;
+mod atomic;
 mod avg;
 mod bitmanip;
 mod cmp;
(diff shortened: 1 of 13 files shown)


abrown:
Should we just bite the bullet and migrate the custom display functions to have a signature like `pub fn inst<R: Registers>(f: &mut std::fmt::Formatter, inst: &inst::inst<R>) -> std::fmt::Result`? It looks like the same thing @rahulchaphalkar is running into over in: https://github.com/bytecodealliance/wasmtime/pull/10836#discussion_r2138322890.

alexcrichton:
I figure this'll need a custom something, but I don't mind it being either a display or a visit. Seems reasonable to me to prefer a custom display so I'm happy to hold off until that merges

abrown (approved):

PR #11009, opened 2025-06-10
Add a dummy impl of fibers for Miri

alexcrichton:
This commit extends the `wasmtime-internal-fiber` crate with an implementation for Miri. Previously this was entirely unsupported because fibers use inline assembly. The implementation with Miri spawns a separate thread and keeps it in a suspended state with locks to model a suspended stack. This technically isn't correct because TLS variables will be wrong, but it's "correct enough" for our usage in Wasmtime. In the end this enables running more tests in Miri which is always a good thing, and a number of loose odds and ends were cleaned up relate to our unsafe management of async state.

### DIFF
--- a/crates/fiber/src/lib.rs
+++ b/crates/fiber/src/lib.rs
@@ -22,6 +22,9 @@ cfg_if::cfg_if! {
     if #[cfg(not(feature = "std"))] {
         mod nostd;
         use nostd as imp;
+    } else if #[cfg(miri)] {
+        mod miri;
+        use miri as imp;
     } else if #[cfg(windows)] {
         mod windows;
         use windows as imp;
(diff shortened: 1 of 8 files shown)


crates/fiber/src/lib.rs
```diff
@@ -139,11 +142,28 @@ pub struct Suspend<Resume, Yield, Return> {
...
+
+    /// Suspend: the fiber hasn't finished but has provided the following value
+    /// during its suspension.
     Yield(Yield),
+
+    /// Suspend: the fiber has completed with the provided value and can no
+    /// longer be resumed.
     Returned(Return),
+
+    /// Suspend: the fiber execution panicked.
```
pchickey:
```suggestion
    /// The fiber execution panicked.
```

crates/fiber/src/lib.rs
```diff
@@ -139,11 +142,28 @@ pub struct Suspend<Resume, Yield, Return> {
...
+
+    /// Resume: resume with this value. Called for each invocation of
+    /// `Fiber::resume`.
     Resuming(Resume),
+
+    /// Suspend: the fiber hasn't finished but has provided the following value
+    /// during its suspension.
     Yield(Yield),
+
+    /// Suspend: the fiber has completed with the provided value and can no
```
pchickey:
```suggestion
    /// The fiber has completed with the provided value and can no
```

crates/fiber/src/lib.rs
```diff
@@ -139,11 +142,28 @@ pub struct Suspend<Resume, Yield, Return> {
...
 enum RunResult<Resume, Yield, Return> {
+    /// The fiber is currently executing meaning it picked up whatever it was
+    /// resuming with and hasn't yet completed.
     Executing,
+
+    /// Resume: resume with this value. Called for each invocation of
+    /// `Fiber::resume`.
     Resuming(Resume),
+
+    /// Suspend: the fiber hasn't finished but has provided the following value
```
pchickey:
```suggestion
    /// The fiber hasn't finished but has provided the following value
```

crates/fiber/src/lib.rs
```diff
@@ -139,11 +142,28 @@ pub struct Suspend<Resume, Yield, Return> {
...
+/// A structure that is stored on a stack frame of a call to `Fiber::resume`.
+///
+/// This is used to both transmit data to a fiber (the resume step) as well as
+/// acquire data from a fiber (the suspension step).
 enum RunResult<Resume, Yield, Return> {
+    /// The fiber is currently executing meaning it picked up whatever it was
+    /// resuming with and hasn't yet completed.
     Executing,
+
+    /// Resume: resume with this value. Called for each invocation of
```
pchickey:
```suggestion
    /// Resume with this value. Called for each invocation of
```

pchickey (approved):

PR #11017, opened 2025-06-11
asm: customize `Mnemonic` and `Display` separately

abrown:
Sometimes we just want to fix up the printed name of the instruction in the disassembly; e.g., instructions with the `lock` prefix. But, as we're seeing in #10836 and elsewhere, other times we want to control the printing of the entire instruction--the full `Display` implementation. We'll need this kind of thing if we want to swap the order of operands or omit an immediate if it ends up being included in the printed instruction name.

This change splits this kind of customization in two: `Mnemonic` and `Display`. After renaming this enum to `Customization`, all existing uses of `Display` are migrated to `Mnemonic`.


abrown:
cc: @rahulchaphalkar

alexcrichton (approved):

PR #11020, opened 2025-06-11
Initial component model and GC support in fused adapters

fitzgen:
This lays down the initial infrastructure for support for GC in our fused adapters for the component model. We keep track of whether each lifted/lowered function wants args/results as GC values or in linear memory. We additionally plumb through the core function types of the functions being lifted and lowered for (eventual) use with GC adapters.

Ultimately, this commit is enough to fuse together lifted and lowered functions where one or both are using the GC variant of the canonical ABI. Attempting to actually pass arguments will hit `todo!()`s. The work of implementing those `todo!()`s is left to future commits.

### DIFF
--- a/crates/environ/src/component/translate.rs
+++ b/crates/environ/src/component/translate.rs
@@ -177,7 +177,6 @@ enum LocalInitializer<'data> {
     Lower {
         func: ComponentFuncIndex,
         lower_ty: ComponentFuncTypeId,
-        canonical_abi: ModuleInternedTypeIndex,
         options: LocalCanonicalOptions,
     },
     Lift(ComponentFuncTypeId, FuncIndex, LocalCanonicalOptions),
@@ -192,7 +191,6 @@ enum LocalInitializer<'data> {
         func: ModuleInternedTypeIndex,
     },
     TaskReturn {
-        func: ModuleInternedTypeIndex,
         result: Option<ComponentValType>,
         options: LocalCanonicalOptions,
     },
@@ -235,12 +233,10 @@ enum LocalInitializer<'data> {
     },
     StreamRead {
         ty: ComponentDefinedTypeId,
-        func: ModuleInternedTypeIndex,
         options: LocalCanonicalOptions,
     },
     StreamWrite {
         ty: ComponentDefinedTypeId,
-        func: ModuleInternedTypeIndex,
         options: LocalCanonicalOptions,
     },
     StreamCancelRead {
@@ -267,12 +263,10 @@ enum LocalInitializer<'data> {
     },
     FutureRead {
         ty: ComponentDefinedTypeId,
-        func: ModuleInternedTypeIndex,
         options: LocalCanonicalOptions,
     },
     FutureWrite {
         ty: ComponentDefinedTypeId,
-        func: ModuleInternedTypeIndex,
         options: LocalCanonicalOptions,
     },
     FutureCancelRead {
@@ -294,11 +288,9 @@ enum LocalInitializer<'data> {
         func: ModuleInternedTypeIndex,
     },
     ErrorContextNew {
-        func: ModuleInternedTypeIndex,
         options: LocalCanonicalOptions,
     },
     ErrorContextDebugMessage {
-        func: ModuleInternedTypeIndex,
         options: LocalCanonicalOptions,
     },
     ErrorContextDrop {
@@ -379,10 +371,7 @@ enum ClosedOverModule {
 #[derive(Debug, Clone, Hash, Eq, PartialEq)]
 pub enum LocalDataModel {
     /// Data is stored in GC objects.
-    Gc {
-        /// The type index of the core GC types signature.
-        core_type: Option<ModuleInternedTypeIndex>,
-    },
+    Gc {},
 
     /// Data is stored in a linear memory.
     LinearMemory {
@@ -399,6 +388,8 @@ struct LocalCanonicalOptions {
     post_return: Option<FuncIndex>,
     async_: bool,
     callback: Option<FuncIndex>,
+    /// The type index of the core GC types signature.
+    core_type: ModuleInternedTypeIndex,
     data_model: LocalDataModel,
 }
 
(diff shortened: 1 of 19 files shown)


crates/environ/src/component/translate/inline.rs
```diff
@@ -1314,6 +1325,28 @@ impl<'a> Inliner<'a> {
...
+                    }
+                }
+                ExportItem::Index(_) => None,
+                ExportItem::Name(_name) => todo!("FITZGEN"),
+            },
+            dfg::CoreDef::InstanceFlags(_) => None,
+            dfg::CoreDef::Trampoline(idx) => Some(self.result.trampolines[*idx].0),
+            dfg::CoreDef::Adapter(_) => None,
+        }
+    }
```
fitzgen: Marking this PR as a draft because I am pretty unsure about this method. It feels like there has to be a better way? Also I am not sure how to grab the type for imports or named exports.

cc @alexcrichton

crates/environ/src/fact.rs
```diff
@@ -200,6 +237,12 @@ enum HelperLocation {
     Stack,
     /// Located in linear memory as configured by `opts`.
     Memory,
+    /// Located in a GC struct field.
+    #[allow(dead_code, reason = "CM+GC is still WIP")]
```
alexcrichton: s/allow/expect/

(that way you're forced to delete this once it's used)

crates/environ/src/fact/signature.rs
```diff
@@ -25,8 +27,22 @@ impl ComponentTypesBuilder {
...
+                            results: f.returns().iter().map(|ty| self.val_type(ty)).collect(),
+                        };
+                    }
+                    _ => unreachable!(),
+                }
+            }
+        };
         let ty = &self[options.ty];
-        let ptr_ty = options.options.ptr();
+        let ptr_ty = mem_opts.ptr();
```
alexcrichton: Question on this, would it be possible to plumb the type information around to avoid ever needing to reimplement lowering here? In that if we always had a type available could we always run the case of the Gc case here?

crates/environ/src/component/translate.rs
```diff
@@ -375,14 +375,31 @@ enum ClosedOverModule {
     Upvar(ModuleUpvarIndex),
 }
 
+/// The data model for objects that are not unboxed in locals.
+#[derive(Debug, Clone, Hash, Eq, PartialEq)]
+pub enum LocalDataModel {
+    /// Data is stored in GC objects.
+    Gc {
+        /// The type index of the core GC types signature.
+        core_type: Option<ModuleInternedTypeIndex>,
```
alexcrichton: Possible alternative to this, and a number of other locations, could `Gc` be empty? The type of a function is something that's always known regardless of whether it's gc or linear memory, so could the type be adjacent to the "function item" and `Gc` here be empty? That way in the gc case there's still always a type available but it would also be available in the linear memory case too in case it's needed. (e.g. elsewhere when calculating the lowered signature we could avoid reimplementing algorithms in wasmtime I think?)

crates/environ/src/component/translate/inline.rs
alexcrichton: I'd recommend renaming this to `core_ty_of_func_def` and panicking internally if it's not a function (e.g. `InstanceFlags` would panic saying "not a function").

Either that or I might recommend replacing the one caller of this function (right?), `AliasExportFunc`, by inlining the definition of `core_def_of_module_instance_export` into that location. The `core_def_of_module_instance_export` has all the type information available to it and might be able to simplify this as you wouldn't need to match-then-match of sorts.

> Also I am not sure how to grab the type for imports or named exports.

Named exports aren't actually possible so you can ignore those. By inlining `core_def_of_module_instance_export` into `AliasExportFunc` that should become apparent as the only constructor of `ExportItem::Name` is with imports. For imports you'll also have the module type available and you can look up the named export of the module type and that'll be the type information you want.

crates/environ/src/fact/signature.rs
fitzgen: You mean plumb in the `wasmparser` types? I can't think of any reason that wouldn't be possible in theory, but I don't really feel comfortable enough with this corner of the code base to actually do that myself with confidence yet. FWIW, the GC signature is just the core signature that we *did* already plumb through, so this change isn't making the problem any worse, and is more similar to where we would like to eventually end up than the existing code is.

alexcrichton: Nah not `wasmparser::types::Types`, but `ModuleInternedTypeIndex`. Basically how the GC branch has the type available to it, no lowering necessary, so there's not really any reason we can't also carry along the type index in the linear memory case, removing the need for lowering.

fitzgen: I didn't remove this method yet, but it should be possible in follow ups.

fitzgen:
Making it so that the options always have a core type was actually really nice, and a bunch of initializers that previously had options and a core type no longer need a separate core type.

Anyways, I think this is ready for another round of review.

crates/wasmtime/src/runtime/component/component.rs
```diff
@@ -834,9 +834,19 @@ impl Component {
...
+        &CoreDef,
+        ModuleInternedTypeIndex,
+        &CanonicalOptions,
+    ) {
         match &self.env_component().export_items[export] {
-            Export::LiftedFunction { ty, func, options } => (*ty, func, options),
+            Export::LiftedFunction {
+                ty,
+                func,
+                func_ty,
```
alexcrichton: Given that this live in `CanonicalOptions` I don't think there's any need for it to live in `LiftedFunction` any more?

crates/environ/src/fact/signature.rs
```diff
@@ -25,8 +27,25 @@ impl ComponentTypesBuilder {
...
+                    .inner
+                {
+                    crate::WasmCompositeInnerType::Func(f) => {
+                        return Signature {
+                            params: f.params().iter().map(|ty| self.val_type(ty)).collect(),
+                            results: f.returns().iter().map(|ty| self.val_type(ty)).collect(),
+                        };
+                    }
+                    _ => unreachable!(),
+                }
```
alexcrichton: Would it be possible to use this branch independent of whatever data model is used now?

fitzgen: Ah I see what you were getting at this whole time now. Yes, done.

alexcrichton (approved):

PR #11025, opened 2025-06-12
 x64: Delete MovRR instruction variant 

alexcrichton:
This removes `MovRR` and no new instructions were needed in the
assembler as everything was already added. Lots of little updates here
and there but everything pretty straightforward. Much of the conversion
goop should get better over time as preexisting instructions should move
to `Gpr` instead of `Reg` which would avoid the need for conversions at
all.

### DIFF
--- a/cranelift/codegen/src/isa/x64/inst/emit.rs
+++ b/cranelift/codegen/src/isa/x64/inst/emit.rs
@@ -287,37 +287,17 @@ pub(crate) fn emit(
             }
         }
 
-        Inst::MovRR { size, src, dst } => {
-            let src = src.to_reg();
-            let dst = dst.to_reg().to_reg();
-            emit_std_reg_reg(
-                sink,
-                LegacyPrefixes::None,
-                0x89,
-                1,
-                src,
-                dst,
-                RexFlags::from(*size),
-            );
-        }
-
         Inst::MovFromPReg { src, dst } => {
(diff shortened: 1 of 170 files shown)


cranelift/codegen/src/isa/x64/inst/emit.rs
```diff
@@ -2697,8 +2676,12 @@ pub(crate) fn emit(
...
 
             // movq %rax, %r_temp
-            let i2 = Inst::mov_r_r(OperandSize::Size64, dst_old.to_reg(), temp);
-            i2.emit(sink, info, state);
+            let inst = asm::inst::movq_mr::new(
+                temp.map(Gpr::unwrap_new),
+                Gpr::unwrap_new(dst_old.to_reg()),
+            )
+            .into();
+            Inst::External { inst }.emit(sink, info, state);
```
fitzgen: Would be nice to have an `emit_asm` helper or something so that every call site doesn't need to do the whole `.into(); Inst::External { inst }.emit(...)` dance. Because the new version of everything here is a lot harder to read.

alexcrichton: I mentioned this in the OP but while some impedance mismatch is what you mention another big source is using `Reg` in pseudo-insts instead of `Gpr`. I'll work on a follow-up to try to fix that.

PR #11029, opened 2025-06-12
x64: Improve external assembler ergonomics

alexcrichton:
Especially in the case of pseudo-instructions in `x64/inst/emit.rs` the goal is to reduce the amount of boilerplate necessary to emit instructions. This is done via:

* A new helper trait was added that's implemented for all assembler instructions to avoid a `.into()` after constructors as well as to avoid `Inst::External { inst }` construction.

* Pseudo-instructions are moved to using `Gpr` and friends instead of `Reg` to more naturally convert to raw assembler types.

* Some usage of `Inst::*` helpers is removed when it's known exactly what assembler instruction is going to be used and it's easier to use the assembler instruction instead.


alexcrichton:
cc @fitzgen

abrown (approved):
I'm fine with this; let's see what @fitzgen thinks. I think there is more that can be done along this path in the future to make things more ergonomic.

fitzgen (approved):
Big improvement IMO :+1:

PR #11030, opened 2025-06-12
Pass operand types in to Wasm-to-CLIF translation

fitzgen:
This lets us generate better CLIF based on what we know of the Wasm types. There is a bunch we can do here, but this commit just lays down the initial infrastructure and plumbs the type info through to the `ref.is_null` implementation. We will now avoid actually performing null check when the input to a `ref.is_null` instruction is a non-nullable ref type.

Note that this depends on both an eventual `wasm-tools` version bump and that we fix a bug in the arity stuff in `wasmparser`, which I have not tracked down yet. (One of the `disas` tests currently panics because we fail to provide an arity for a valid, reachable `end` instruction.)


fitzgen:
With https://github.com/bytecodealliance/wasm-tools/pull/2237, this is now passing all tests locally, so I'll cut a wasm-tools release, update those crates in this repo, and then this PR can rebase and merge.

fitzgen:
wasm-tools dep update in https://github.com/bytecodealliance/wasmtime/pull/11034

crates/cranelift/src/translate/func_translator.rs
```diff
@@ -241,15 +241,34 @@ fn parse_function_body(
...
     let mut reader = OperatorsReader::new(reader);
+    let mut operand_types = vec![];
+
     while !reader.eof() {
         let pos = reader.original_position();
         builder.set_srcloc(cur_srcloc(&reader.get_binary_reader()));
+
         let op = reader.read()?;
+
+        // Get the operand types for this operator.
```
alexcrichton: Mind leaving a comment that the option-ality is due to the fact that we're not certain that `op` is valid yet? In that if it's invalid the arity may be `None`. Once we get past `validator.op`, though, then it's known that `operand_types` is valid. Although we also can't move this block because `validator.op` mutates the validator.

Also, can this `.unwrap()` after the validator to ensure that the signatures of the before/after/translate all take a slice, not an optional slice?

crates/cranelift/src/translate/code_translator.rs
```diff
@@ -2863,9 +2901,12 @@ pub fn translate_operator(
             from_ref_type: _,
```
alexcrichton, line 2897: There's a TODO just above where I'm commenting here which can be resolved now

alexcrichton (approved):

crates/cranelift/src/translate/func_translator.rs
fitzgen: > Also, can this `.unwrap()` after the validator to ensure that the signatures of the before/after/translate all take a slice, not an optional slice?

Unfortunately not because even if the op is valid, if it is in unreachable code, the op might want to pop more values from the stack than actually exist on the stack (which is allowed in unreachable code) so even if we can get arity, we are only guaranteed to have operand types for ops that are not only valid but also reachable.

We could instead pass `&[Option<WasmValType>]` through, which would only ever contain `None`s inside valid-but-unreachable code, but I think the ergonomics are even worse with this approach.

I've added comments explaining this however.

alexcrichton: Ah ok makes ense, and yeah I don't think `&[Option<_>]` is correct here, so threading around option but documenting the unreachable-ness sounds good to me.

PR #11031, opened 2025-06-12
winch(aarch64) Run integration tests in winch-aarch64

saulecabrera:
This commit closes https://github.com/bytecodealliance/wasmtime/issues/8321

With https://github.com/bytecodealliance/wasmtime/pull/11013, Winch passes all spec tests for Core Wasm on Aarch64.

### DIFF
--- a/crates/test-util/src/wast.rs
+++ b/crates/test-util/src/wast.rs
@@ -249,6 +249,7 @@ macro_rules! foreach_config_option {
             exceptions
             legacy_exceptions
             stack_switching
+        integration
         }
     };
 }
@@ -270,7 +271,6 @@ macro_rules! define_test_config {
                 }
             )*
         }
-
     }
 }
 
@@ -361,8 +361,8 @@ impl Compiler {
                 if cfg!(target_arch = "aarch64") {
                     return unsupported_base
                         || config.wide_arithmetic()
-                        || config.threads()
-                        || config.simd();
+                        || (config.simd() && config.integration())
+                        || config.threads();
                 }
 
                 false
--- a/crates/test-util/src/wast.rs
+++ b/crates/test-util/src/wast.rs
@@ -72,6 +72,7 @@ fn add_tests(tests: &mut Vec<WastTest>, path: &Path, has_config: bool) -> Result
 
 fn spec_test_config(test: &Path) -> TestConfig {
     let mut ret = TestConfig::default();
+    ret.spec_test = Some(true);
     match spec_proposal_from_path(test) {
         Some("multi-memory") => {
             ret.multi_memory = Some(true);
@@ -249,7 +250,7 @@ macro_rules! foreach_config_option {
             exceptions
             legacy_exceptions
             stack_switching
-        integration
+            spec_test
         }
     };
 }
@@ -361,7 +362,7 @@ impl Compiler {
                 if cfg!(target_arch = "aarch64") {
                     return unsupported_base
                         || config.wide_arithmetic()
-                        || (config.simd() && config.integration())
+                        || (config.simd() && !config.spec_test())
                         || config.threads();
                 }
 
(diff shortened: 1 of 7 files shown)


tests/all/memory.rs
```diff
@@ -114,8 +114,8 @@ fn offsets_static_dynamic_oh_my(config: &mut Config) -> Result<()> {
         }
     }
 
-    engines.par_iter().for_each(|engine| {
-        let module = module(&engine).unwrap();
+    engines.par_iter().try_for_each(|engine| {
+        let module = module(&engine)?;
```
saulecabrera, line 118: We need to gracefully handle compilation errors to correctly assert that a test should fail in case of missing features (e.g., SIMD on aarch64).

crates/test-util/src/wast.rs
```diff
@@ -359,7 +359,10 @@ impl Compiler {
                 }
 
                 if cfg!(target_arch = "aarch64") {
-                    return unsupported_base || config.wide_arithmetic() || config.threads();
+                    return unsupported_base
+                        || config.wide_arithmetic()
+                        || config.threads()
+                        || config.simd();
```
alexcrichton: I think that this actually causes lots of test to "fail" with:

```
...

---- Winch/./tests/spec_testsuite/float_exprs.wast ----
this test is flagged as should-fail but it succeeded

---- Winch/./tests/spec_testsuite/table_copy.wast ----
this test is flagged as should-fail but it succeeded

---- Winch/./tests/spec_testsuite/const.wast ----
this test is flagged as should-fail but it succeeded

...
```

the reason being that when we classify spec features required by spec tests we do that by directory as opposed to per-test so the entire spec test suite is considered as "requires simd". That means that all spec tests are expected to fail despite many of them passing.

One possible fix is we could split into `config.simd` into `config.really_needs_simd` and `config.maybe_needs_simd` so this could return "should fail" for `config.really_needs_simd` but for spec tests they'd just be flagged with `maybe_needs_simd` which means we'd still run them (and allow-list failures below)

saulecabrera: Ah right, I didn't test this change with ~~x64~~ wast and aarch64, I also had forgotten that the `should_fail` code is now shared. Let me explore your idea and see where I land.

saulecabrera: @alexcrichton I've pushed a variation of your suggestion in https://github.com/bytecodealliance/wasmtime/pull/11031/commits/99e648766a9ebb41b864b92f96fe2eb16669d397 -- let me know what you think. Given that this configuration struct is shared among two tests contexts, I thought it might be easier to reason through if we instead introduce a proper option to flag the test as integration. Your suggested approach works, but I found it a bit challenging to reason about since it modifies to a certain extent the semantic meaning of the Wasm config depending on the testing context. Happy to revisit that last commit if you think that dividing the `simd` option is better though.

alexcrichton (commented):
I like it 👍 

If you'll indulge a minor bikeshed though, what about inverting this from `integration: bool` to `spec_test: bool`? I'm a little worried that only the macro-defined tests are `integration: true` where all of `tests/misc_testsuite/*` should in theory be automatically `integration: true` as well. By inverting this to `spec_test: bool` it should be easy to audit the single location that classifies spec tests and ensure it sets that and then everything else can ignore it being set.

Also, as a minor nit, instead of adding `..` to patterns can you add `spec_test: _` to explicitly ignore the field? The exhaustive match helps ensure that we update the patterns/etc when a new field is added.

saulecabrera:
@alexcrichton all that make sense, I've pushed the fixes in https://github.com/bytecodealliance/wasmtime/pull/11031/commits/93bd9d4530203806e1ae1efe76359e879dda3733

saulecabrera:
Hmm there seems to be a legit segfault for `macOS-arm64`, taking a look.

saulecabrera:
It didn't seem like a segfault in the test itself, nor I was able to reproduce after running the job multiple times (locally and in CI).

PR #11035, opened 2025-06-13
x64: Migrate `cmpxchg{,16b}` to the new assembler

alexcrichton:
A few new operands were needed to define this as well as a number of new instruction "shapes" in a few locations, but nothing too too radical.


cranelift/codegen/src/isa/x64/inst/emit_tests.rs
```diff
@@ -827,7 +680,7 @@ fn test_x64_emit() {
             temp: w_r11.map(Gpr::unwrap_new),
             dst_old: w_rax.map(Gpr::unwrap_new)
         },
-        "490FB7014989C34D23DA66F0450FB1190F85EEFFFFFF",
+        "490FB7014989C34D23DAF066450FB1190F85EEFFFFFF",
```
alexcrichton, line 683: TIL there's multiple possible encodings for one instruction:

```
   0:   66 f0 45 0f b1 19       lock cmpxchg %r11w,(%r9)
   6:   f0 66 45 0f b1 19       lock cmpxchg %r11w,(%r9)
```

fitzgen (approved):

cranelift/codegen/src/isa/x64/inst/emit_tests.rs
abrown: I'm trying to document some of this stuff as I come across it:

https://github.com/bytecodealliance/wasmtime/blob/b4e86667187317d7cd62fe8cfaa2a7479babc3d9/cranelift/assembler-x64/meta/src/dsl/encoding.rs#L477-L485

PR #11037, opened 2025-06-13
x64: Migrate some no-operand instructions

alexcrichton:
Fences, hlt, and ud2 are now all in the new assembler.

### DIFF
--- a/cranelift/codegen/src/isa/x64/inst/emit.rs
+++ b/cranelift/codegen/src/isa/x64/inst/emit.rs
@@ -2369,8 +2369,7 @@ pub(crate) fn emit(
                 inst.emit(sink, info, state);
             } else {
                 // Trap.
-                let inst = Inst::trap(TrapCode::INTEGER_OVERFLOW);
-                inst.emit(sink, info, state);
+                asm::inst::ud2_zo::new(TrapCode::INTEGER_OVERFLOW).emit(sink, info, state);
             }
 
             // Now handle large inputs.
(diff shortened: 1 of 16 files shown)


cranelift/codegen/src/isa/x64/inst/emit.rs
```diff
@@ -2820,25 +2819,6 @@ pub(crate) fn emit(
...
-            sink.put1(0xAE);
-            match kind {
-                FenceKind::MFence => sink.put1(0xF0), // mfence = 0F AE F0
-                FenceKind::LFence => sink.put1(0xE8), // lfence = 0F AE E8
-                FenceKind::SFence => sink.put1(0xF8), // sfence = 0F AE F8
-            }
-        }
-
-        Inst::Hlt => {
-            sink.put1(0xcc);
```
abrown, line 2737: Just wanted to flag that we will now be using a new encoding here: `0xf4` versus `0xcc` originally. `0xCC` is the encoding for `INT3`.

abrown (commented):
Maybe we should look at this `INT3` versus `HLT` difference?

cranelift/codegen/src/isa/x64/inst/emit.rs
abrown: What I mean is: this `Inst` variant may be misnamed (?) and previously we were always using `INT3` when we emitted a `Inst::Hlt`. I think if we emit a `HLT` it's a bit more drastic...

alexcrichton: Oh good catch! I'll make sure the encoding doesn't change

Issue #10733, opened 2025-05-05
`--emit-clif` files do not include function names


MonoidMusician:
The IR included in `explore` includes function names with each section of IR, but this info doesn't end up in the similar files produced by `--emit-clif` which makes it hard to see what each function is (without cross-referencing with the explore output or something).

jameysharp:
As the person who asked you to open this issue, I want to add a couple details for whoever has time to poke at it:

`wasmtime explore` gets the function name here: https://github.com/bytecodealliance/wasmtime/blob/ede670dd937d73122ae36104fcea6a345c0890c1/crates/explorer/src/lib.rs#L222-L231

The implementation of `--emit-clif` is here: https://github.com/bytecodealliance/wasmtime/blob/a4700cb7eb86484ba4d5246cdb85bf07311a51a0/crates/cranelift/src/compiler.rs#L1000-L1004

In addition to `explore`, `tests/disas.rs` also uses emit-clif internally and IIRC also wants function names, so maybe this can all be unified a bit.

PhantomInTheWire:
hey @alexcrichton! This looks like a good first issue! I think I have a solution for this. Could you please assign it to me?

alexcrichton:
Sure thing, and one thing to be aware of perhaps is that https://github.com/bytecodealliance/wasmtime/pull/10947 just landed which changed things slightly in this area, but I believe this is still applicable with respect to `wasmtime explore`.

PhantomInTheWire:
Awesome, thanks for assigning it! I'll dig into this

PR #11040, opened 2025-06-13
Add function names to --emit-clif output

PhantomInTheWire:
#### Related Issues
Fixes: #10733

#### Explanation

This PR addresses issue #10733 by adding the function name as a comment to the beginning of each Cranelift IR file generated by the `--emit-clif` flag.

Previously, the files generated by `--emit-clif` contained only the raw Cranelift IR, making it difficult to identify which function the IR belonged to without cross-referencing with other tools like `wasmtime explore`. This change improves the usability of the `--emit-clif` output by clearly labeling each function's IR.

#### Input Needed
I'm unsure of where to place tests for this.

### DIFF
--- a/crates/wasmtime/src/runtime/component/func.rs
+++ b/crates/wasmtime/src/runtime/component/func.rs
@@ -167,7 +167,7 @@ impl Func {
         Params: ComponentNamedList + Lower,
         Return: ComponentNamedList + Lift,
     {
-        let cx = InstanceType::new(instance.unwrap_or_else(|| &store[self.instance.id()]));
+        let cx = InstanceType::new(instance.unwrap_or_else(|| self.instance.id().get(store)));
         let ty = &cx.types[self.ty(store)];
 
         Params::typecheck(&InterfaceType::Tuple(ty.params), &cx)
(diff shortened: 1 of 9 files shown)


crates/cranelift/src/compiler.rs
```diff
@@ -1002,6 +1002,12 @@ impl FunctionCompiler<'_> {
             path.set_extension("clif");
 
             let mut output = std::fs::File::create(path).unwrap();
+            write!(
+                output,
+                ";;Intermediate Representation of function <{}>:\n",
+                symbol.split("::").last().unwrap()
```
bjorn3: This would work too, right?

```suggestion
                ";; Intermediate Representation of function <{symbol}>:\n",
```

This avoids truncating a function name like `wasm[0]::function[1]::my_library::SomeType::get` (as would be used if the names section contains `my_library::SomeType::get` as name) to the pretty much useless `get`.

fitzgen (commented):
Nothing to add over what @bjorn3 pointed out.

crates/cranelift/src/compiler.rs
PhantomInTheWire: would using something like this function work better?

```rust
fn remove_first_two_segments(s: &str) -> Option<String> {
    let mut parts = s.split("::").skip(2);
    let rest: Vec<&str> = parts.collect();
    if rest.is_empty() {
        None
    } else {
        Some(format!("::{}", rest.join("::")))
    }
}
```

also wasmtime explore uses "Intermediate Representation of function <get>:" too should i figure out a way where it gives my_library::SomeType::get too?

PR #11042, opened 2025-06-13
Use `Pin<&mut ComponentInstance>`

### LINKED ISSUE #10933: Technically `&mut Instance` is not safe

alexcrichton:
This commit is the continuation of #10943 for component instances. The allocation/vmctx infrastructure was additionally refactored to be shared for both core and component instances since they behave the exact same way anyway. This further enables sharing various methods like `vmctx_plus_offset` which are pretty unsafe internally.

Like #10943 this necessitated removal of `Index` implementations because `IndexMut` is not compatible with the returned type being `Pin<&mut T>` so they were replaced by inherent `get` and `get_mut` methods on the component instance id type.

Closes #10933

### DIFF
--- a/crates/wasmtime/src/runtime/component/func.rs
+++ b/crates/wasmtime/src/runtime/component/func.rs
@@ -167,7 +167,7 @@ impl Func {
         Params: ComponentNamedList + Lower,
         Return: ComponentNamedList + Lift,
     {
-        let cx = InstanceType::new(instance.unwrap_or_else(|| &store[self.instance.id()]));
+        let cx = InstanceType::new(instance.unwrap_or_else(|| self.instance.id().get(store)));
         let ty = &cx.types[self.ty(store)];
 
         Params::typecheck(&InterfaceType::Tuple(ty.params), &cx)
@@ -181,7 +181,7 @@ impl Func {
     /// Get the parameter names and types for this function.
     pub fn params(&self, store: impl AsContext) -> Box<[(String, Type)]> {
         let store = store.as_context();
-        let instance = &store[self.instance.id()];
+        let instance = self.instance.id().get(store.0);
         let types = instance.component().types();
         let func_ty = &types[self.ty(store.0)];
         types[func_ty.params]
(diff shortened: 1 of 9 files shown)


alexcrichton:
cc @fitzgen you're probably also interested in these refactorings.

PR #11044, opened 2025-06-14
Handle `uextend` in conditional trap lowerings on x64

fitzgen:


fitzgen:
Looks like this needs some debug info tweaks, will update soon.

PR #11049, opened 2025-06-16
Update to WASI 0.2.6.

badeend:
Nothing much has changed since 0.2.3, except:
- https://github.com/WebAssembly/wasi-filesystem/pull/165: Documentation update
- https://github.com/WebAssembly/wasi-http/pull/139: Adds a new unstable `send-informational` function to the `response-outparam` resource.
  - I've left this `unimplemented!()`.
  - Since this is the first `@unstable` feature in the wasi-http interface, the bindgen has also generated a new LinkOptions type. Once somebody implements this method, they'll need to propagate a LinkOptions parameter up to the CLI options code.

### DIFF
--- a/crates/wasi-http/src/types_impl.rs
+++ b/crates/wasi-http/src/types_impl.rs
@@ -1,7 +1,7 @@
 //! Implementation for the `wasi:http/types` interface.
 
 use crate::{
-    HttpResult, WasiHttpImpl, WasiHttpView,
+    HttpError, HttpResult, WasiHttpImpl, WasiHttpView,
     bindings::http::types::{self, Headers, Method, Scheme, StatusCode, Trailers},
     body::{HostFutureTrailers, HostIncomingBody, HostOutgoingBody, StreamContext},
     types::{
@@ -10,7 +10,7 @@ use crate::{
         is_forbidden_header, remove_forbidden_headers,
     },
 };
-use anyhow::Context;
+use anyhow::{Context, anyhow};
 use std::any::Any;
 use std::str::FromStr;
 use wasmtime::component::{Resource, ResourceTable, ResourceTableError};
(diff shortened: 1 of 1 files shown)


crates/wasi-http/src/types_impl.rs
```diff
@@ -580,6 +580,15 @@ where
...
         Ok(())
     }
+
+    fn send_informational(
+        &mut self,
+        _id: Resource<HostResponseOutparam>,
+        _status: u16,
+        _headers: Resource<Headers>,
+    ) -> HttpResult<()> {
+        unimplemented!()
```
dicej: I think we should trap here rather than panic.

dicej (commented):
Thanks for doing this!  LGTM, but please see the note about panicking in `send_informational`.

crates/wasi-http/src/lib.rs
```diff
@@ -303,10 +303,13 @@ pub fn add_only_http_to_linker_async<T>(
 where
     T: WasiHttpView + 'static,
 {
+    let options = crate::bindings::LinkOptions::default(); // FIXME: Thread through to the CLI options.
```
rvolosatovs, line 306: Other modules typically define `with_options` counterparts, like so: https://github.com/bytecodealliance/wasmtime/blob/60269d642510263ccf5fe637c042d663fecfc43a/crates/wasi/src/p2/mod.rs#L469-L490

badeend: I'm aware.
I think it makes more sense to update the public API whenever this new method actually gets implemented. Because right now, there's no reason for the consumer to use this parameter, other than to enable a method that unconditionally ~panics~ traps.

WDYT?

dicej (approved):

crates/wasi-http/src/lib.rs
rvolosatovs: Absolutely!

rvolosatovs (approved):

PR #11055, opened 2025-06-16
c-api: component-model: Resource table, WASI

MangoPeachGrape:
I've gotten quite stuck on how to do resources, so I decided to try WASI first.

Some things I'm unsure about:
- Should the context be inside an `Option` and then `expect()` in the WASI traits, or is the default context fine?
- How to specify your own context, perhaps `wasmtime_context_set_wasi_p2()`?

---
On resources, I think I'm getting stuck on `ResourceTypeKind::Host` using `TypeId`.

Would all the resources use a `CApiResource` type on the rust side, and then use a string discriminant, or something like that?

Do you have any other ideas?

### DIFF
--- a/crates/c-api/include/wasmtime/component/linker.h
+++ b/crates/c-api/include/wasmtime/component/linker.h
@@ -142,7 +142,7 @@ WASM_API_EXTERN wasmtime_error_t *wasmtime_component_linker_instance_add_func(
  * \brief Add all WASI interfaces into the \p linker provided.
  */
 WASM_API_EXTERN wasmtime_error_t *
-wasmtime_component_linker_add_wasi(wasmtime_component_linker_t *linker);
+wasmtime_component_linker_add_wasip2(wasmtime_component_linker_t *linker);
 
 #endif // WASMTIME_FEATURE_WASI
 
--- a/crates/c-api/include/wasmtime/component/linker.h
+++ b/crates/c-api/include/wasmtime/component/linker.h
@@ -136,6 +136,16 @@ WASM_API_EXTERN wasmtime_error_t *wasmtime_component_linker_instance_add_func(
     size_t name_len, wasmtime_component_func_callback_t callback, void *data,
     void (*finalizer)());
 
+#ifdef WASMTIME_FEATURE_WASI
+
+/**
+ * \brief Add all WASI interfaces into the \p linker provided.
+ */
+WASM_API_EXTERN wasmtime_error_t *
+wasmtime_component_linker_add_wasi(wasmtime_component_linker_t *linker);
+
+#endif // WASMTIME_FEATURE_WASI
+
 /**
  * \brief Deletes a #wasmtime_component_linker_instance_t
  *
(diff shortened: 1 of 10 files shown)


alexcrichton:
> Should the context be inside an Option and then expect() in the WASI traits, or is the default context fine?

For `ResourceTable` I think it's ok to leave it there on-by-default, but for `wasmtime_wasi::p2::WasiCtx` yeah let's have that be an option which panics if it's not set.

> How to specify your own context, perhaps wasmtime_context_set_wasi_p2()?

Right now `wasi.h` is a binding to the wasip1 builder API, and I think we'll want a new API for the wasip2 builder API (e.g. bindings for `wasmtime_wasi::p2::WasiCtx::builder`). Once that structure is built yeah a new API of `wasmtime_context_set_wasi_p2` would take the builder and then insert it into the store. Then afterwards the Option in the store would be `Some` and everything should work.

> On resources, I think I'm getting stuck on ResourceTypeKind::Host using TypeId.

This has tripped me up in the past as well, I think it might be reasonable to add `ResourceTypeKind::HostId(u64)` where the host just provides a 64-bit integer. That could be added for the C API and that way the C API could take integers to differentiate resource types?

MangoPeachGrape:
> we'll want a new API for the wasip2 builder API (e.g. bindings for `wasmtime_wasi::p2::WasiCtx::builder`)

Where should I place this? Maybe `include/wasmtime/wasi/p2.h`? Or should I place it next to the `include/wasi.h`, i.e. `include/wasi_p2.h`?

And how should the naming convention be? Similar to how it is in `wasi.h`, e.g. `wasi_p2_config_inherit_stdout()`, or something more similar to the rust side, maybe `wasmtime_wasi_p2_builder_inherit_stdout()`?

> I think it might be reasonable to add ResourceTypeKind::HostId(u64) where the host just provides a 64-bit integer. That could be added for the C API and that way the C API could take integers to differentiate resource types?

Seems like a nice solution, I think I'll try that in a next PR?

alexcrichton:
I've long wanted to move `wasi.h` and scope it under the "wasmtime" namespace. Additionally we try to consider "wasipN" a unit as oppose to "wasi pN", so how about `include/wasmtime/wasip2.h` and `wasmtime_wasip2_config_*`?

MangoPeachGrape:
Does that look correct?

> I've long wanted to move wasi.h and scope it under the "wasmtime" namespace

Should I do that also here? Would that include adding  `wasmtime_wasip1_` prefix to everything?

crates/c-api/include/wasmtime/component/linker.h
```diff
@@ -136,6 +136,16 @@ WASM_API_EXTERN wasmtime_error_t *wasmtime_component_linker_instance_add_func(
     size_t name_len, wasmtime_component_func_callback_t callback, void *data,
     void (*finalizer)());
 
+#ifdef WASMTIME_FEATURE_WASI
+
+/**
+ * \brief Add all WASI interfaces into the \p linker provided.
+ */
+WASM_API_EXTERN wasmtime_error_t *
+wasmtime_component_linker_add_wasi(wasmtime_component_linker_t *linker);
```
alexcrichton: s/add_wasi/add_wasip2/

PR #11060, opened 2025-06-17
wasi-nn: add feature to use custom ONNX Runtime

yanwucai:
This change adds a new Cargo feature to allow the users to use a custom build of ONNX Runtime. 

This will also help the users who need to remove some crypto-related dependencies imported by the download-binaries feature of ort.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -778,7 +778,7 @@ jobs:
   test_wasi_nn:
     strategy:
       matrix:
-        feature: ["openvino", "onnx"]
+        feature: ["openvino", "onnx-download"]
         os: ["ubuntu-24.04", "windows-2025"]
         include:
           - os: windows-2025


crates/wasi-nn/Cargo.toml
```diff
@@ -69,7 +68,9 @@ default = ["openvino", "winml"]
 # OpenVINO is available on all platforms; it requires OpenVINO to be installed.
 openvino = ["dep:openvino"]
 # ONNX is available on all platforms.
-onnx = ["dep:ort"]
+onnx = ["ort/download-binaries"]
+# Use locally built ONNX Runtime.
+onnx-local-rt = ["dep:ort"]
```
abrown: What if we flipped this around?
- `onnx` would just have `dep:ort` and
- `onnx-download` would bring in the `ort/download-binaries` feature

(I'm no Cargo features expert but it does seem to conform better to the idea that they should be [additive](https://doc.rust-lang.org/cargo/reference/features.html)).

crates/wasi-nn/src/backend/mod.rs
```diff
@@ -2,7 +2,7 @@
 //! this crate. The `Box<dyn ...>` types returned by these interfaces allow
 //! implementations to maintain backend-specific state between calls.
 
-#[cfg(feature = "onnx")]
+#[cfg(any(feature = "onnx", feature = "onnx-local-rt"))]
```
abrown: If you flip the features around, then these `#cfg`s no longer need to check both cases, right?

crates/wasi-nn/Cargo.toml
yanwucai: I am open to both options. But this would be a breaking change. If we flipped this around, existing users who are using `onnx` feature would have to switch to `onnx-download` feature. Is this OK?

abrown: I think that's ok. This is all pretty experimental at this point so having to change features for a new release seems reasonable. We can document it in the release notes.

yanwucai: I flipped it around.

crates/wasi-nn/src/backend/mod.rs
yanwucai: Yes, no need to check both cases now.

abrown:
@yanwucai, it looks like this failure is due to some CI job that now needs the `onnx-download` feature enabled. Why didn't we see this earlier? Wasmtime only runs a subset of the CI jobs for a PR but runs all the CI jobs in the merge queue, including the wasi-nn jobs that failed here. If you want to run all the CI jobs in this PR to make sure things are working as intended before merge, add the text `prtest:full` to the body of one of the commit messages.

yanwucai:
Hi @abrown, the `onnx` feature is added back to CI in PR #11071 , I updated my branch and updated the feature. The CI jobs have passed now.

PR #11062, opened 2025-06-17
Fix Cranelift-native fuzzers

alexcrichton:
This commit fixes a typo mistake from #10588 where the option specified there was actually invalid. This error got covered up by accident when generating the fuzz test case, however, meaning that the error was not noticed. This changes builder configuration to using `.unwrap()` more liberally for options that are all known should be enabled.

### DIFF
--- a/cranelift/fuzzgen/src/lib.rs
+++ b/cranelift/fuzzgen/src/lib.rs
@@ -171,11 +171,11 @@ where
 
     /// Generate a random set of cranelift flags.
     /// Only semantics preserving flags are considered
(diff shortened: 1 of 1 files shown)


fitzgen (approved):
Maybe define a helper for set/enable that documents why we are unwrapping so that someone doesn't clean this up to use "proper" error propagation in a couple years?

alexcrichton:
That should already be done with the `arbitrary::Result` return type I think? That's what guided me to change some `?` to `.unwrap()` here

fitzgen:
Ah, I overlooked the function signature change. That makes perfect sense, nice little tick.

PR #11064, opened 2025-06-17
Add `native-tls` backend for wasi-tls

badeend:
During the [initial wasi-tls implementation discussion](https://github.com/bytecodealliance/wasmtime/issues/10089#issuecomment-2610420435), the suggestion was floated to support both `rustls` and `native-tls`. The [initial/current implementation](https://github.com/bytecodealliance/wasmtime/pull/10249) only supports `rustls`. This PR adds `native-tls` as an alternative TLS backend.

The default TLS provider remains `rustls`. Consumers can switch to the `native-tls` backend by enabling the corresponding feature flags and then use the newly created `WasiTlsCtxBuilder::provider` method. On the CLI, this is exposed as `-S tls-provider=nativetls`.

Adding this new backend immediately surfaced two bugs:

#### 1. Improved compatibility of `test_tls_sample_application`
The test used to perform a "half-close" after sending the HTTP request. This is a TLS1.3+ feature, though Rustls & OpenSSL already supported it for TLS1.2 and lower. Technically, that makes them non-spec compliant, but they chose to align with the semantics of the underlying TCP connection. I suspect the TLS1.3 spec was updated to match what was already happening in reality.

Anyhow, SChannel does not support half-closed connections and so the `read` call after the `close_output+shutdown` failed. I've reordered the test to first perform the HTTP conversation and _then_ do the TLS+TCP teardown.

#### 2. Implement flushing on the AsyncWriteStream.
The AsyncWriteStream implementation was copied from the TCP equivalent, which doesn't need flushing. The TLS implementations _do_ maintain an internal buffer, so the `flush` call needs to be hooked up. Without it, the test would hang after making the change mentioned above.

#### Misc: Reorganized wasi-tls folder
To make life easier for myself, I've shuffled some code around. I've isolated the rustls-specific bits from the rest of the implementation. The folder structure now looks like:
- `providers/*`: The rustls & native-tls specific parts.
- `bindings.rs`: The generated bindings.
- `host.rs`: The host types + impls.
- `io.rs`: WASI I/O conversion utility types.

CC: @jsturtevant

### DIFF
--- a/supply-chain/imports.lock
+++ b/supply-chain/imports.lock
@@ -622,6 +622,13 @@ user-id = 6743
 user-login = "epage"
 user-name = "Ed Page"
 
+[[publisher.core-foundation]]
+version = "0.9.3"
+when = "2022-02-07"
+user-id = 5946
+user-login = "jrmuizel"
+user-name = "Jeff Muizelaar"
+
 [[publisher.core-foundation-sys]]
 version = "0.8.4"
 when = "2023-04-03"
@@ -930,6 +937,13 @@ when = "2022-04-30"
 user-id = 189
 user-login = "BurntSushi"
 user-name = "Andrew Gallant"
+
+[[publisher.openssl-probe]]
+version = "0.1.6"
+when = "2025-01-23"
+user-id = 1
+user-login = "alexcrichton"
+user-name = "Alex Crichton"
 
 [[publisher.prettyplease]]
 version = "0.2.31"


alexcrichton:
This all looks great to me, thanks again for working on this!

My only comment would be could this update CI checks to perform a build of the `wasmtime-wasi-tls` crate with some feature combinations? I'd be interested to see `--no-default-features --features rustls` and one for native-tls to ensure the crate can built with only one of the two enabled.

I'll file a separate PR to perform the vets as well.

badeend:
Is [this](https://github.com/bytecodealliance/wasmtime/pull/11064/commits/4040567f21929a1d672ffdb1a587c74d5603a035) the right place to do that?

crates/wasi-tls/src/providers/mod.rs
```diff
@@ -0,0 +1,20 @@
...
+#[cfg(feature = "rustls")]
+pub use rustls::*;
+
+cfg_if::cfg_if! {
+    if #[cfg(feature = "rustls")] {
+        pub use RustlsProvider as DefaultProvider;
+    } else if #[cfg(feature = "nativetls")] {
+        pub use NativeTlsProvider as DefaultProvider;
+    } else {
+        compile_error!("At least one TLS provider must be enabled.");
```
jsturtevant: does it make sense to have to set two feature flags to enable the feature or should we provide a default to enable the feature or should we provide a default? 

Also a bit confused in the way to set this up. Do I need to configure both feature flags (`wasi-tls` and `nativetls` or `rusttls`) and then also set the provider with   `wasi_tls_ctx: WasiTlsCtxBuilder::new().provider(provider).build(),`?

badeend: `rustls` is part of the `default` feature. So by default, `DefaultProvider` will point to the rustls provider and there's no need to configure anything on the `WasiTlsCtxBuilder`.

Similarly, if you build with `--no-default-features --features nativetls`, then `DefaultProvider` will point to the native_tls provider and there's again no need to configure anything on the `WasiTlsCtxBuilder`.

The only reason to call `WasiTlsCtxBuilder::provider` is when the crate has been built with multiple providers and you want to switch between them at run time. This is what happens in e.g. the test suite and in the CLI code.

I've [added some docs](https://github.com/bytecodealliance/wasmtime/pull/11064/commits/240f5619541ae8ad0fa61cbe89831cd0d45c2995) to clarify this.

alexcrichton:
Hm ok so this is actually going to be kind of tricky. We test crates in CI with `--all-features` but we also test cross-compilations to various architectures, and native-tls on Linux wants OpenSSL which wants a system copy by default which isn't present for cross-compiled variants. 

One option might be to have a feature such as `openssl-vendored` on the `wasmtime-wasi-tls` crate which builds a vendored copy of `openssl-src`. That way `--all-features` of that would be activated an OpenSSL would be built from source?

That's not an ideal option though IMO as ideally we wouldn't test native-tls by default and we'd instead have a dedicated job for just testin native-tls, but with `--all-features` that's not easy :(

badeend:
How about moving `/crates/wasi-tls/src/providers/native_tls.rs` into a dedicated `wasi-tls-nativetls` crate and exclude that entire crate from the default test runs?

**Edit:** And set up a separate CI job to test `wasi-tls-nativetls` in isolation on a Linux, MacOS and Windows host.

**Edit2:** A downside of this approach is that the wasmtime CLI may not reference the `wasi-tls-nativetls` package anymore either.

alexcrichton:
I think what you're proposing is probably the best option for now. You're right that it means that the CLI won't be easily buildable with native-tls, but that seems like a reasonable state of affairs for the time being and we can address that later if need be

badeend:
Could you take a look at my latest changes? I've ran it with `prtest:full` and everything (except `cargo vet`) now succeeds.

PR #11065, opened 2025-06-17
Cranelift: Generate integer numeric ops and conversions for ISLE in the meta crate

fitzgen:
This automatically generates operations and conversions for integer types for use in ISLE.

Supported types are: `{i,u}{8,16,32,64,128}`

We generate

* Comparisons (eq, ne, lt, lt_eq, gt, gt_eq)
* Arithmetic operations (add, sub, mul, div, neg)
  * These each have checked, wrapping, and unwrapping variants
* Bitwise operations (and, or, xor, shifts, counting leading/trailing zeros/ones)
* A variety of predicates (is_zero, is_power_of_two, is_odd, etc...)
  * These generate both partial constructors and a handful of extractors
* Conversions
  * These come in a variety of flavors: fallible, infallible, truncating, unwrapping, sign-reinterpretation
  * Fallible conversions are also available as an extractor


abrown:
I like this. Side thought: should we add a `main.rs` to `cranelift-codegen-meta` that prints out all the paths of the generated files? Having all this generated code is great but sometimes it can be tricky to figure out which generated code in is being used. I've been thinking of doing some refactoring to see if we could get there...

fitzgen:
> sometimes it can be tricky to figure out which generated code in is being used

FWIW if you pass `-vv` to `cargo build` (or `cargo test` or whatever) it should print the outputs of any build scripts it runs.

abrown:
Yeah, for me wading through `cargo build -vv` is about as appetizing as wading through the `target` directory hierarchy to figure out which files are actually being used.

cranelift/codegen/meta/src/gen_isle.rs
```diff
@@ -495,13 +495,679 @@ fn gen_isle_enum(name: &str, mut variants: Vec<&str>, fmt: &mut Formatter) {
...
+            },
+            NumericOp {
+                name: "trailing_zeros",
+                body: "a.trailing_zeros()",
+                ret: "u32",
+                ..unop.clone()
+            },
+            NumericOp {
+                name: "trailing_ones",
+                body: "a.trailing_zeros()",
```
alexcrichton: s/zeros/ones/

cranelift/codegen/meta/src/gen_isle.rs
```diff
@@ -495,13 +495,679 @@ fn gen_isle_enum(name: &str, mut variants: Vec<&str>, fmt: &mut Formatter) {
...
+            },
+            NumericOp {
+                name: "leading_zeros",
+                body: "a.leading_zeros()",
+                ret: "u32",
+                ..unop.clone()
+            },
+            NumericOp {
+                name: "leading_ones",
+                body: "a.leading_zeros()",
```
alexcrichton: s/zeros/ones/

cranelift/codegen/meta/src/isle.rs
```diff
@@ -128,7 +129,11 @@ pub fn get_isle_compilations(
                     src_isa_x64.join("inst.isle"),
                     src_isa_x64.join("lower.isle"),
                 ],
-                untracked_inputs: vec![clif_lower_isle.clone(), gen_dir.join("assembler.isle")],
+                untracked_inputs: vec![
+                    numerics_isle.clone(),
```
alexcrichton, line 133: Is it concerning that we now have hundreds of helpers, probably generating ~thousands of lines of Rust, and we're compiling those N times, one for each backend?

cranelift/codegen/meta/src/gen_isle.rs
```diff
@@ -495,13 +495,679 @@ fn gen_isle_enum(name: &str, mut variants: Vec<&str>, fmt: &mut Formatter) {
...
+                body: "a.checked_add(b)",
+                ..partial_binop.clone()
+            },
+            NumericOp {
+                name: "wrapping_add",
+                body: "a.wrapping_add(b)",
+                ..binop.clone()
+            },
+            NumericOp {
+                name: "unwrapping_add",
```
alexcrichton: As a bikeshed what about renaming `unwrapping_add` to `add`? That matches Rust more closely and the only difference is "always debug_assertions semantics"

cranelift/codegen/src/isa/aarch64/inst.isle
```diff
@@ -3837,17 +3837,17 @@
...
-        (if-let new_offset (s32_add_fallible x offset))
+        (if-let new_offset (i32_checked_add x offset))
         (amode_no_more_iconst ty y new_offset))
 
 (decl amode_no_more_iconst (Type Value i32) AMode)
 ;; Base case: move the `offset` into a register and add it to `val` via the
 ;; amode
 (rule 0 (amode_no_more_iconst ty val offset)
-        (AMode.RegReg val (imm $I64 (ImmExtend.Zero) (i64_as_u64 offset))))
+        (AMode.RegReg val (imm $I64 (ImmExtend.Zero) (i64_reinterpret_as_u64 offset))))
```
alexcrichton: As a bikeshed I find "reinterpret" pretty wordy and overly verbose, and additionally it looks like `cast_{signed,unsigned}` are actually on stable Rust already, so could this perhaps switch to being `i64_cast_unsigned`? That makes it a little less wordy and additionally aligns with what one would expect in Rust too

cranelift/codegen/src/isa/pulley_shared/lower.isle
```diff
@@ -266,18 +266,18 @@
...
 ;; Helper extract a constant from a `Value`, negate it, and fit it within a
 ;; `u32`.
 (decl pure partial u32_from_negated_iconst (Value) u32)
 (rule (u32_from_negated_iconst (i32_from_iconst i))
-  (if-let neg_i64 (i64_neg (i32_as_i64 i)))
-  (if-let neg_u64 (u64_try_from_i64 neg_i64))
-  (if-let neg_u32 (u32_try_from_u64 neg_u64))
+  (if-let neg_i64 (i64_checked_neg i))
+  (if-let neg_u64 (i64_try_into_u64 neg_i64))
+  (if-let neg_u32 (u64_try_into_u32 neg_u64))
```
alexcrichton: It should be possible to collapse these sequential `*_try_into_*` calls now if you're up for it

cranelift/codegen/src/isa/pulley_shared/inst.isle
```diff
@@ -626,27 +626,27 @@
...
 (rule 2 (imm (ty_int _) x)
-      (if-let y (i16_try_from_u64 x))
+      (if-let y (u64_try_into_i16 x))
       (pulley_xconst16 y))
 
 ;; If a value can fit into 32 bits, then prioritize that.
 (rule 1 (imm (ty_int _) x)
-      (if-let y (i32_try_from_u64 x))
+      (if-let y (u64_try_into_i32 x))
       (pulley_xconst32 y))
```
alexcrichton: While you're here mind changing this by adding a `u64_reinterpret_as_i64` before the conversion to a smaller bit-width? I think that's a preexisting missing optimization we have.

cranelift/codegen/src/isa/riscv64/inst.isle
```diff
@@ -2352,18 +2346,18 @@
 (decl gen_bseti (Reg u64) Reg)
 (rule (gen_bseti val bit)
   (if-let false (has_zbs))
-  (if-let false (u64_le bit 12))
-  (let ((const XReg (imm $I64 (u64_shl 1 bit))))
+  (if-let false (u64_lt_eq bit 12))
```
alexcrichton, line 2349: Bikeshedding this a bit, `le` matches Rust as well, so maybe `le` instead of `lt_eq`?

cranelift/codegen/src/isa/riscv64/lower.isle
```diff
@@ -616,7 +616,8 @@
 (rule (safe_sdiv_divisor ty x y)
   (let (
       (y XReg (nonzero_divisor y))
-      (min XReg (imm $I64 (u64_shl 0xffffffff_ffffffff (u64_sub (ty_bits ty) 1))))
+      (min XReg (imm $I64 (u64_wrapping_shl 0xffffffff_ffffffff
+                                            (u64_truncate_into_u32 (u64_wrapping_sub (ty_bits ty) 1)))))
```
alexcrichton: I think you can probably tweak this a bit to avoid the truncate since the source, `ty_bits` I think returns `u32` instead of `u64`

cranelift/codegen/src/isa/x64/inst.isle
```diff
@@ -2422,25 +2422,25 @@
...
   (x64_movq_mi_sxl imm32))
 
 ;; Special case: a 64-bit immediate which zero extends from a 32-bit immediate.
 ;;
 ;; Note that `movl` here will zero-extend the destination register in 64-bit
 ;; mode which is the zero-extension we want.
 (rule 3 (imm $I64 imm)
-  (if-let imm32 (u32_try_from_u64 imm))
+  (if-let imm32 (u64_try_into_u32 imm))
   (x64_movl_oi imm32))
```
alexcrichton: This can be cleaned up to using an extractor instead of `if-let` if you're feeling up to it

cranelift/codegen/src/isa/x64/inst.isle
```diff
@@ -4244,12 +4244,12 @@
...
-      (if-let val2 (i32_try_from_i64 val))
+      (if-let val2 (i64_try_into_i32 val))
       (x64_addq_mi_sxl_mem addr val2))
-(rule 2 (x64_add_mem $I32 addr (i32_from_iconst (i8_try_from_i32 val)))
+(rule 2 (x64_add_mem $I32 addr (i32_from_iconst (i8_from_i32 val)))
       (x64_addl_mi_sxb_mem addr val))
 (rule 2 (x64_add_mem $I64 addr (i64_from_iconst val))
-      (if-let val2 (i8_try_from_i64 val))
+      (if-let val2 (i64_try_into_i8 val))
       (x64_addq_mi_sxb_mem addr val2))
```
alexcrichton: LIke above these can be cleaned up to using extractors as well

cranelift/codegen/src/prelude_opt.isle
```diff
@@ -134,9 +134,11 @@
 (decl iconst_s (Type i64) Value)
 (extractor (iconst_s ty c) (inst_data_value_tupled (iconst_sextend_etor ty c)))
 (rule 0 (iconst_s ty c)
-    (if-let c_masked (u64_and (i64_as_u64 c) (ty_umax ty)))
+    (if-let c_masked (u64_and (i64_reinterpret_as_u64 c)
+                              (ty_umax ty)))
     (if-let c_reextended (i64_sextend_u64 ty c_masked))
-    (if-let true (u64_eq (i64_as_u64 c) (i64_as_u64 c_reextended)))
+    (if-let true (u64_eq (i64_reinterpret_as_u64 c)
+                         (i64_reinterpret_as_u64 c_reextended)))
```
alexcrichton: Now that we have `i64_eq` these reinterprets I think can be removed?

alexcrichton (commented):
Some bikesheds here and there, along with some minor suggestions, but overall 👍

cranelift/codegen/meta/src/isle.rs
fitzgen: Ideally we would only generate what is actually used, but that seems difficult to track and also annoying to update when you are the first person who needed `u8_is_zero`, which is part of what this PR is trying to alleviate. Ultimately, I think we end up in a better situation than we were before. At least all the generated code is very simple, doesn't use generics, and hopefully shouldn't stress rustc.

cranelift/codegen/meta/src/gen_isle.rs
fitzgen: I don't feel super strongly, but lowering and optimization are places where we need to be really careful about preserving bits, so I wanted to push us towards consciously choosing the correct operation every time, rather than using the default `add` operation without thinking. But if you and others still prefer `add` over `unwrapping_add`, I can certainly make that change.

cranelift/codegen/src/isa/riscv64/inst.isle
fitzgen: I kind of like `lt_eq` because it isn't ambiguous with "little endian" but happy to change if everyone else prefers `le`

cranelift/codegen/src/isa/pulley_shared/inst.isle
```diff
@@ -621,28 +621,43 @@
...
+;; `xconst16`.
+(rule 4 (imm $I16 x) (pulley_xconst16 (u16_cast_signed (u64_truncate_into_u16 x))))
+
 ;; If a value can fit into 16 bits, then prioritize that.
-(rule 2 (imm (ty_int _) x)
-      (if-let y (u64_try_into_i16 x))
+(rule 3 (imm (ty_int ty) x)
+      (if-let y (i64_truncate_into_i16 (u64_cast_signed x)))
+      (if-let true (u64_eq (u64_and (i64_cast_unsigned (i16_into_i64 y)) (ty_mask ty))
+                           (u64_and x (ty_mask ty))))
```
alexcrichton, line 645: This seems perhaps a bit overly complicated, could this be `(if-let (i16_from_i64 imm) (u64_cast_signed x))` as the entirety of the conversion?

cranelift/codegen/meta/src/gen_isle.rs
alexcrichton: Personally I don't consider panics in the backend all that serious and relatively easily weeded out with fuzzing, so I'd prefer to keep aligned with Rust-native methods across these and drop the "unwrapping" prefix. I definitely agree it should have unwrapping semantics though, regardless.

cranelift/codegen/src/isa/riscv64/inst.isle
alexcrichton: Nah that makes sense to me

alexcrichton (approved):

cranelift/codegen/src/isa/pulley_shared/inst.isle
fitzgen: I originally had roughly this, but it doesn't work unfortunately, because the `imm` term is sometimes given `u64` values that are not sign extended to the type's width. Basically the upper bits are undefined and elsewhere in the system we do not canonicalize them to being sign extended in practice, even though that would simplify things for us here.

```
FAIL wasmtime/cranelift/filetests/filetests/isa/pulley64/iconst.clif: compile

Caused by:
    compilation of function on line 22 does not match
    the text expectation
    
    --- expected
    +++ actual
    @@ -1,10 +1,10 @@
     VCode:
     block0:
    -  xconst8 x0, -1
    +  xconst16 x0, -1
       xconst16 x1, 255
       ret
     
     Disassembled:
    -xconst8 x0, -1
    +xconst16 x0, -1
     xconst16 x1, 255
     ret
```

alexcrichton: Alas :(

I'll try to come through and clean this up after the PR lands

cranelift/codegen/meta/src/gen_isle.rs
fitzgen: Okay, I renamed `u64_unwrapping_add` to `u64_add`, etc...

PR #11068, opened 2025-06-18
fix: do not collect backtrace on each `gc_runtime` access

rvolosatovs:
This is something I've discovered profiling a Wasmtime embedding with `wasm32-unknown-unknown` modules. Currently, `gc_runtime` accessor returns an `anyhow::Result`, which triggers a backtrace collection, when `gc` is disabled (e.g. when `gc` feature is not set)

With `gc` feature disabled:
https://share.firefox.dev/3FTmlgN (note the `anyhow::error::msg` and related `_Unwind_IteratePhdrCallback`)

With `gc`, `gc-null`, `gc-drc` features enabled:
https://share.firefox.dev/3HGyEh5

To alleviate this, return an `Option` from the accessor, since in most cases, it's used as an `Option` anyway and there is only one case when error is returned when it is not set

### DIFF
--- a/crates/wasmtime/src/engine.rs
+++ b/crates/wasmtime/src/engine.rs
@@ -688,12 +688,8 @@ impl Engine {
         self.inner.allocator.as_ref()
     }
 
-    pub(crate) fn gc_runtime(&self) -> Result<&Arc<dyn GcRuntime>> {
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks!

PR #11070, opened 2025-06-18
Prefix `Lift` and `Lower` trait methods with `memory_`

fitzgen:
This is to distinguish them from the GC versions that will be added in follow up commits.

No functional changes here, just renaming.

### DIFF
--- a/crates/component-macro/src/component.rs
+++ b/crates/component-macro/src/component.rs
@@ -456,16 +456,20 @@ impl Expander for LiftExpander {
 
         for (i, syn::Field { ident, ty, .. }) in fields.iter().enumerate() {
             let field_ty = quote!(ty.fields[#i].ty);
-            lifts.extend(quote!(#ident: <#ty as #wt::component::Lift>::lift(
+            lifts.extend(
+                quote!(#ident: <#ty as #wt::component::Lift>::linear_lift_from_flat(
                 cx, #field_ty, &src.#ident
-            )?,));
+            )?,),
+            );
 
-            loads.extend(quote!(#ident: <#ty as #wt::component::Lift>::load(
+            loads.extend(
+                quote!(#ident: <#ty as #wt::component::Lift>::linear_lift_from_memory(
                 cx, #field_ty,
                 &bytes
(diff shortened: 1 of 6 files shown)


alexcrichton:
To motivate a bit of a bikeshed: these sorts of changes tend to have a very high sticking power where they stick around for a very long time so I'd ideally like to handle things in one fell swoop instead of going through multiple renames. To bikeshed: `memory_lift` and `memory_load` are IMO deceptively subtle in how they're different as "lift from memory" is something I would reasonably say should be synonymous with a load from memory. Given that I'd prefer to find different names that are more nuanced than just putting `memory_*` in front of what we currently have.

One observation as well is that I think there's probably not going to be a `gc_load` or a `gc_store`. Current design has lifting/lowering of GC values being mostly like the "flat" lifting/lowering of linear memory values. I realize the struct/array cases are a bit different but I think what we'll probably do is pass a parameter through which differentiates the local/struct/array source/destinations of values.

Naming-wise WDYT about `{lift,lower}_{linear,gc}` and otherwise leaving `load` and `store` as-is? (or maybe `{load,store}_linear`)

fitzgen:
For the GC methods I currently have:

* `Lift::gc_lift_val` for doing GC lifting from a value
* `Lift::gc_lift_storage` for doing GC lifting from a storage type (i.e. a struct field or array element)
* `Lower::gc_lower_val` for doing GC lowering into a value
* `Lower::gc_lower_storage` for doing GC lowering into a storage type

fitzgen:
How about

* `Lift::lift_from_flat_linear`
* `Lift::lift_from_memory_linear`
* `Lift::lift_from_val_gc`
* `Lift::lift_from_storage_gc`
* `Lower::lower_to_flat_linear`
* `Lower::lower_to_memory_linear`
* `Lower::lower_to_val_gc`
* `Lower::lower_to_storage_gc`

?

alexcrichton:
That sounds good to me 👍

alexcrichton (approved):

pchickey:
https://github.com/bytecodealliance/wasmtime/pull/11089 fixes CI

PR #10411, opened 2025-03-17
Remove testing ONNX from CI

alexcrichton:
This is blocking CI currently as a downloaded artifact is currently offline. This can be re-enabled once that's fixed.

PR #11071, opened 2025-06-19
Re-enable removed testing onnx from CI

TrumsZ:
This PR is raised to re-enable the testing of onnx in the CI according to the discussion in https://github.com/bytecodealliance/wasmtime/pull/10411.


abrown (approved):
Thanks for this PR; let's see if it works when I put it in the merge queue! The merge queue runs all of CI whereas by default a PR only runs a subset of the jobs (one can override this by adding `prtest:full` in the body of a commit message).

PR #11072, opened 2025-06-19
x64: Migrate cmp/test to the new assembler

alexcrichton:
This commit migrates the `cmp` and `test` family of instructions, the `CmpRmiR` variant in ISLE, to the new assembler. This required fiddling with various helpers for immediates and such throughout a few locations and refactoring various callsites of creating these instructions to fit nicely into these new idioms. EFLAGS-writing instructions with no other results are now also modeled in the assembler as `ProducesFlags` rather than `SideEffectNoResult` as their base case enabling using the generated constructors a bit more too.

No functional change is expected from this, it should just be internal refactoring.

### DIFF
--- a/cranelift/assembler-x64/meta/src/dsl/format.rs
+++ b/cranelift/assembler-x64/meta/src/dsl/format.rs
@@ -598,6 +598,17 @@ pub enum Eflags {
     RW,
 }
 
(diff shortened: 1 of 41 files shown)


abrown (approved):
Nice!

PR #11075, opened 2025-06-19
Avoid running wasm-opt when compiling test programs

philipc:
wasm-opt will strip DWARF debug info, which causes a test failure for `debug::lldb::dwarf_codegen_optimized_wasm_optimized`.

clang automatically runs wasm-opt if it is found in PATH and if optimization is enabled. clang-20 has a --no-wasm-opt option, but that doesn't work for wasi-sdk-25.

### DIFF
--- a/crates/test-programs/artifacts/build.rs
+++ b/crates/test-programs/artifacts/build.rs
@@ -290,6 +290,9 @@ impl Artifacts {
         cmd.args(&config.flags);
         cmd.arg("-o");
         cmd.arg(&wasm_path);
+        // If optimizations are enabled, clang will look for wasm-opt in PATH
+        // and run it. This will strip DWARF debug info, which we don't want.
+        cmd.env("PATH", "");
         println!("running: {cmd:?}");
         let result = cmd.status().expect("failed to spawn clang");
         assert!(result.success());


alexcrichton:
Is there a compiler flag to tell wasm-opt to preserve debuginfo we could pass? I'm a bit wary to null out PATH as I think that may cause issues on Windows and otherwise thwart detection of various tools clang might otherwise do?

philipc:
`wasm-opt` has a `-g` flag which may or may not be sufficient, but I can't see a way to pass that through clang. But even if that worked, I would still prefer to not run `wasm-opt`. I don't think tests should be changing behaviour based on whether `wasm-opt` happens to be in the path.

I could add something to wasmtime to build a binary called `wasm-opt` which does nothing, and place that first in the path when building the test program.

Or I could change my own system to avoid the problem, but that won't fix it for others.

philipc:
Maybe it's best to wait for wasi-sdk to use a newer clang version so we can use --no-wasm-opt.

alexcrichton:
Hey if it works it works, we can always clean up later if needed once wasi-sdk is upgraded 👍

PR #11080, opened 2025-06-19
Add pack convert (*packss*, *packus*) instructions to new assembler

jlb6740:

### DIFF
--- a/cranelift/codegen/src/isa/x64/inst.isle
+++ b/cranelift/codegen/src/isa/x64/inst.isle
@@ -493,10 +493,6 @@
 
 (type SseOpcode extern
       (enum Insertps
-            Packssdw
-            Packsswb
-            Packusdw
-            Packuswb
             Palignr
             Pmaddubsw
             Pmaddwd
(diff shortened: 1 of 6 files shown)


cranelift/codegen/src/isa/x64/inst.isle
```diff
@@ -3341,35 +3337,35 @@
 
 ;; Helper for creating `packsswb` instructions.
 (decl x64_packsswb (Xmm XmmMem) Xmm)
-(rule 0 (x64_packsswb src1 src2)
-      (xmm_rm_r (SseOpcode.Packsswb) src1 src2))
 (rule 1 (x64_packsswb src1 src2)
         (if-let true (use_avx))
         (xmm_rmir_vex (AvxOpcode.Vpacksswb) src1 src2))
+(rule 0 (x64_packsswb src1 src2)
+      (x64_packsswb_a src1 src2))
```
abrown: Now that the alternate stuff is built, we can just use one rule that uses the "or_avx" ISLE rule.

jlb6740: Something like this:

```
;; Helper for creating `packsswb` instructions.
(decl x64_packsswb (Xmm XmmMem) Xmm)
(rule (x64_packsswb src1 src2) (x64_packsswb_a_or_avx src1 src2))
```

And it's normal that the vex version for the isle does not include an explicit destination? Just the two source operands right?

abrown (approved):

PR #11084, opened 2025-06-19
 x64: Migrate `lea` to the new assembler 

alexcrichton:
This PR has two commits which culminate in migrating `lea` to the new assembler. The main trickiness about `lea` is that we lower all `iadd` nodes to `lea` but rely on an optimization to turn `lea` back into `add` where possible. The first commit here adds infrastructure to use the results of register allocation to refine instruction selection and then the `lea` commit builds on top of that and uses it. There are a few other examples of using the results of register allocation such as `AND RAX, imm32` which ISLE never otherwise generates but can be selected after register allocation for a slightly more compact encoding.


abrown (approved):
Nice! It looks like there are small code-size wins everywhere and there are probably opportunities for more of this kind of micro-optimization as well.

PR #11090, opened 2025-06-20
Make auto-publish script more robust

alexcrichton:
* Add a helper to run `curl` and get the output
* Print an error if `curl` fails to help diagnose what went wrong in the future
* Pass a custom `--user-agent` which is requested by crates.io's [data access policy](https://crates.io/data-access).

### DIFF
--- a/scripts/publish.rs
+++ b/scripts/publish.rs
@@ -444,19 +444,18 @@ fn publish(krate: &Crate) -> bool {
 
     // First make sure the crate isn't already published at this version. This
     // script may be re-run and there's no need to re-attempt previous work.
-    let output = cmd_output(Command::new("curl").arg(&format!(
+    let output = curl(&format!(
         "https://crates.io/api/v1/crates/{}/versions",
         krate.name
-    )));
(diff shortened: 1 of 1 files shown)


alexcrichton:
The auto-publish of the 34.0.0 release failed without these changes. I believe the failure is because currently crates.io is blocking requests from `curl` ([#t-crates-io > API requests bounced with 403 since ~3 hours ago @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/318791-t-crates-io/topic/API.20requests.20bounced.20with.20403.20since.20~3.20hours.20ago/near/524709398)). If a `curl` request fails the script currently prints nothing and then the continued logic does not handle its own failure, basically we expect `curl` to succeed. My hope is that by adding `--user-agent` it can indeed succeed, and then additionally extra info is printed on failure to help diagnose this in the future.

alexcrichton:
I'll also note that in the meantime I manually ran the publish locally and this release at least should be sorted out.

pchickey (approved):

PR #3608, opened 2021-12-16
aarch64: Migrate ishl/ushr/sshr to ISLE

alexcrichton:
This commit migrates the `ishl`, `ushr`, and `sshr` instructions to
ISLE. These involve special cases for almost all types of integers
(including vectors) and helper functions for the i128 lowerings since
the i128 lowerings look to be used for other instructions as well. This
doesn't delete the i128 lowerings in the Rust code just yet because
they're still used by Rust lowerings, but they should be deletable in
due time once those lowerings are translated to ISLE.

PR #11096, opened 2025-06-21
Add CI job to verify static library build configuration

thedevyashsaini:
This PR introduces a dedicated CI job (`build-static-lib`) that verifies the static library build of Wasmtime by passing `-DWASMEDGE_BUILD_STATIC_LIB=ON` to CMake.

### Changes:

* Added a standalone `build-static-lib` job in `.github/workflows/main.yml`, triggered alongside regular builds.
* This job uses a fixed Rust toolchain (`stable`) and target (`x86_64-unknown-linux-gnu`).
* It runs independently of the main matrix-based `build` job to avoid cache conflicts or artifact overrides.
* Static build logic is activated via `static=true` passed to `ci/build-release-artifacts.sh`.

### Motivation:

Static library builds can fail due to symbol collisions or source file name conflicts. CI coverage ensures we catch such regressions early.

### Related:

Fixes [#3608](https://github.com/WasmEdge/WasmEdge/issues/3608)
Refs [this comment](https://github.com/WasmEdge/WasmEdge/pull/3607#issuecomment-2254381977) suggesting the need for a `STATIC_LIB=ON` build in CI.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1079,6 +1079,41 @@ jobs:
         name: bins-${{ matrix.build }}
         path: dist
 
+  # Independent job to verify that static builds still work.
+  # This catches issues like duplicate symbols or bad linkage
+  # when building Wasmtime with -DWASMEDGE_BUILD_STATIC_LIB=ON.
+  #
+  # This job is intentionally NOT part of the main matrix build.
+  # It runs with hardcoded values and doesn't interfere with
+  # release artifacts or build cache.
+  build-static-lib:
+    name: Static Library Build
+    runs-on: ubuntu-latest
+    needs: determine
+    if: needs.determine.outputs.run-full
+    steps:
+      # Checkout source code with submodules
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+
+      # Install Ninja (required by some CMake builds)
+      - uses: ./.github/actions/install-ninja
+
+      # Install the Rust toolchain
+      - uses: ./.github/actions/install-rust
+        with:
+          toolchain: stable
+
+      # Add required Rust components and targets
+      - run: |
+          rustup component add rust-src
+          rustup target add x86_64-unknown-linux-gnu
+
+      # Build the project with the static lib flag enabled
+      - name: Build Wasmtime as a static library
+        run: static=true ./ci/build-release-artifacts.sh release x86_64-unknown-linux-gnu
+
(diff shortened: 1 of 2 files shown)


alexcrichton:
@thedevyashsaini did you mean to send this PR to the WasmEdge project instead of Wasmtime?

thedevyashsaini:
@alexcrichton yeaah, this started as a fix for WasmEdge and somewhere down the rabbit hole I ended up static linking Wasmtime by instinct 😅
That said - the CI test might still be useful here, but feel free to shut it down if it's out of scope...

alexcrichton:
Ok, in that case I'm going to close this beacuse I don't think this does what you think it does in terms of testing. AFAIK this doesn't actually do anything on CI as the cmake option you specify here isn't read anywhere. As a result I'm going to close this.

thedevyashsaini:
@alexcrichton 

> Totally understand closing this — thanks for reviewing!
>
> Just wanted to clarify one point you mentioned:
>
> > “the cmake option you specify here isn’t read anywhere…”
>
> I believe it *is* wired through the CI scripts. Here’s the flow I followed:
>
> * In `.github/workflows/main.yml` (line \~1115), the job runs:
>
>   ```bash
>   static=true ./ci/build-release-artifacts.sh release x86_64-unknown-linux-gnu
>   ```
>
> * Then inside `ci/build-release-artifacts.sh` the "static" option is checked, around line 69:
>
>   ```bash
>   if [[ "$static" == "true" ]]; then
>     cmake_flags="$cmake_flags -DWASMEDGE_BUILD_STATIC_LIB=ON"
>   fi
>   ```
>
> * These `$cmake_flags` are passed into the `cmake` command at line \~82:
>
>   ```bash
>   cmake \
>     -G Ninja \
>     ../../crates/c-api \
>     $cmake_flags \
>     ...
>   ```
>
> * So `-DWASMEDGE_BUILD_STATIC_LIB=ON` *should* be included in the build, unless I’ve misunderstood something else upstream (like the flag not being handled in the actual CMakeLists).
>
> In any case, totally fair to close this if it's not useful — just thought I'd clarify in case this ends up being worth revisiting later. Cheers!

thedevyashsaini:
Thanks for the review! Yep, just realized `-DWASMEDGE_BUILD_STATIC_LIB` isn’t defined in Wasmtime’s CMake. Appreciate the clarification!

PR #11097, opened 2025-06-21
x64: Refactor backend condition handling

alexcrichton:
This commit represents a refactor of the x64 backend's handling of condition codes for various instructions with a goal of simplifying conversion of a `Value` to a condition code used for `select` and `brif`. This additionally tries to deduplicate handling of condition codes across `icmp` and `fcmp` for example.

Previously the x64 backend had types such as `IcmpCondResult` and `FcmpCondResult` which respresented the condition codes for the `icmp` and `fcmp` values. When used in `select`, `brif`, or `trap{n,}z` each lowering had to special case `icmp` and `fcmp` to create these `*CondResult` values and call specialized helpers to lower the value back. Furthermore on the `select` instruction this created a matrix of ways-to-produce-the-condition-code along with
ways-to-handle-the-condition-code to conditionally move GPR, XMM, or 128-bit values. The end result was a fair bit of duplication across rules and optimizations on some rules but not others. For example `brif` on a `vall_true` value was optimized but `select` was not.

The design implemented in this commit is inspired/modeled after what the riscv64 backend is currently doing. There is a top-level helper `is_nonzero_cmp` which takes a `Value` and produces a `CondResult`. This `CondResult` is a merging of `IcmpCondResult` and `FcmpCondResult` into one. This centralizes the ability to take any value and produce condition codes, for example this handles `icmp`, `fcmp`, 128-bit integers, `v{all,any}_true`, etc. Once the condition is produced it's then handled separately at each location for jumps, selects, traps, etc. In effect `CondResult` serves as a "narrow waist" through which production of condition codes can all flow meaning that if an optimization is added for production of condition codes all instructions benefit instead of having to hand-update each one.

The goal of this refactoring was to not actually change any lowerings and only refactor internals, but this refactoring exposed a few non-optimal lowerings in the backend which were improved as a result of this change. In the future I plan on additionally adding more ways to pattern-match "produces a condition" which will now equally benefit all of these locations.

### DIFF
--- a/cranelift/codegen/src/isa/x64/inst.isle
+++ b/cranelift/codegen/src/isa/x64/inst.isle
@@ -2560,74 +2560,12 @@
 (rule 1 (cmove $I64 (CC.P) c a) (x64_cmovpq_rm a c))
 (rule 1 (cmove $I64 (CC.NP) c a) (x64_cmovnpq_rm a c))
 
-(decl cmove128 (CC ValueRegs ValueRegs) ConsumesFlags)
-(rule (cmove128 cc cons alt)
-  (consumes_flags_concat
-    (cmove $I64 cc (value_regs_get_gpr cons 0) (value_regs_get_gpr alt 0))
-    (cmove $I64 cc (value_regs_get_gpr cons 1) (value_regs_get_gpr alt 1))))
-
 (decl cmove_xmm (Type CC Xmm Xmm) ConsumesFlags)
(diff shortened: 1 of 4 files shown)


abrown (approved):
I think this makes sense to me but @fitzgen you may want to double-check the Spectre-related changes.

fitzgen (approved):
Nice!

PR #9213, opened 2024-09-07
#7954 alias analysis attempt

badumbatish:
Mentioned in #7954, attempted in #8535, reattempted in this pr.

Draft first for ci pipeline. Possible rebase for better commit msg

PR #11098, opened 2025-06-22
feat(cranelift): Use DominatorTreePreorder in more places

PhantomInTheWire:
**Motivation**
This PR is to audit and reduce usage of `DominatorTree` and its `dominates()` method, as tracked in https://github.com/bytecodealliance/wasmtime/issues/7954.The goal is to replace these with `DominatorTreePreorder` wherever applicable.

**Context**
Recent related work includes:

* [[#8534](https://github.com/bytecodealliance/wasmtime/pull/8534)](https://github.com/bytecodealliance/wasmtime/pull/8534)
* [[#8535](https://github.com/bytecodealliance/wasmtime/pull/8535)](https://github.com/bytecodealliance/wasmtime/pull/8535)
* [[#9213](https://github.com/bytecodealliance/wasmtime/pull/9213)](https://github.com/bytecodealliance/wasmtime/pull/9213)

### DIFF
--- a/cranelift/codegen/src/context.rs
+++ b/cranelift/codegen/src/context.rs
@@ -11,6 +11,7 @@
 
 use crate::alias_analysis::AliasAnalysis;
 use crate::dominator_tree::DominatorTree;
+use crate::dominator_tree::DominatorTreePreorder;
 use crate::egraph::EgraphPass;
 use crate::flowgraph::ControlFlowGraph;
 use crate::ir::Function;
@@ -46,6 +47,9 @@ pub struct Context {
     /// Dominator tree for `func`.
     pub domtree: DominatorTree,
 
+    /// Dominator tree with dominance stored implicitly via visit-order indices for `func`
+    domtree_preorder: DominatorTreePreorder,
+
     /// Loop analysis of `func`.
     pub loop_analysis: LoopAnalysis,
 
@@ -74,6 +78,7 @@ impl Context {
             func,
             cfg: ControlFlowGraph::new(),
             domtree: DominatorTree::new(),
+            domtree_preorder: DominatorTreePreorder::new(),
             loop_analysis: LoopAnalysis::new(),
             compiled_code: None,
             want_disasm: false,
@@ -305,7 +310,8 @@ impl Context {
 
     /// Compute dominator tree.
     pub fn compute_domtree(&mut self) {
-        self.domtree.compute(&self.func, &self.cfg)
+        self.domtree.compute(&self.func, &self.cfg);
+        self.domtree_preorder.compute(&self.domtree);
     }
 
     /// Compute the loop analysis.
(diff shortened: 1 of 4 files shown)


alexcrichton:
I'm going to reroll this on the compiler-reviewer rotation as I'm not familiar with this data structure, but I suspect others are!

cranelift/codegen/src/context.rs
```diff
@@ -45,6 +46,7 @@ pub struct Context {
 
     /// Dominator tree for `func`.
     pub domtree: DominatorTree,
+    domtree_preorder: DominatorTreePreorder,
```
cfallin, line 51: Could we have a doc-comment on this field, like the others? In particular let's note what's different ("dominator tree with dominance stored implicitly via visit-order indices").

cranelift/codegen/src/dominator_tree.rs
```diff
@@ -580,9 +580,37 @@ impl DominatorTreePreorder {
     pub fn dominates(&self, a: Block, b: Block) -> bool {
         let na = &self.nodes[a];
         let nb = &self.nodes[b];
+
+        // A block is considered to dominate itself
+        if a == b {
+            return true;
+        }
```
cfallin: I'm surprised you needed to add this case -- it should be implied by the below (`pre_number` and `post_number` will both be equal pairwise to the other nodes' values if `a == b`). Is this a performance optimization or...?

cranelift/codegen/src/dominator_tree.rs
```diff
@@ -580,9 +580,37 @@ impl DominatorTreePreorder {
...
+
+        // Dominance is ill defined for unreachable blocks
+        // Return false if either block is unreachable
+        if na.pre_number == 0 || nb.pre_number == 0 {
+            return false;
+        }
```
cfallin: Was this needed to address a failure otherwise? I'd prefer keeping the simple definition -- the original should also return false (I think?) if either is unreachable and the other is not, since the RHS of a `<=` will be zero.

cranelift/codegen/src/verifier/mod.rs
```diff
@@ -1036,12 +1042,9 @@ impl<'a> Verifier<'a> {
                         format!("{v} is defined by {block} which is not in the layout"),
                     ));
                 }
+                let user_block = self.func.layout.inst_block(loc_inst).unwrap();
```
cfallin: Let's use `expect()` here with a string indicating why this is infallible (we're traversing code that is in a block).

cfallin (commented):
Thanks for this! A number of thoughts below, but overall looks fine.

cranelift/codegen/src/context.rs
PhantomInTheWire: done.

cranelift/codegen/src/dominator_tree.rs
PhantomInTheWire: no it was actually to deal with some debug statements that the previous pr added, but since I removed them this is no longer necessary

cranelift/codegen/src/dominator_tree.rs
PhantomInTheWire: sorry i missed it, removed.

cranelift/codegen/src/verifier/mod.rs
PhantomInTheWire: my bad, i forgot to replace it. we should avoid unwrap() as much as we can.

cfallin (approved):
Looks great -- thanks!

PR #11113, opened 2025-06-23
Change `wasmtime wast` to be powered from JSON AST

alexcrichton:
This commit is a refactoring of the `wasmtime-wast` crate to use the `json-from-wast` crate added in bytecodealliance/wasm-tools#2247 instead of the `wast` crate directly. The primary motivation for this PR is to make spec tests more suitable for running inside of Miri. There are two primary factors in how Miri is slow with wast tests today:

* Compiling WebAssembly modules. These are all embedded throughout `*.wast` files and basically need to be precompiled to run in Miri.

* Parsing the `*.wast` script itself. The scripts are generally quite large in the upstream spec test suite and parsing the script requires parsing all modules as well, so this can take quite some time.

The goal of this commit was to leverage `json-from-wast` to perform much of the heavy lifting on the host and then have a "cheap" parse step in Miri which deserializes the JSON and runs precompiled `*.cwasm` files. The main change then required of `wasmtime-wast` was to use the JSON AST as its "IR" instead of `wast` directly. The actual transformation of the `wasmtime-wast` crate isn't too bad, mostly just some refactorings here and there.

A new `./ci/miri-wast.sh` script is added which first runs on native with `wasmtime wast --precompile-save` and then runs in Miri with `wasmtime wast --precompile-load`. In this manner I was able to get a number of spec test `*.wast` files running in Miri.

Unfortunately, though, Miri is still far too slow to run in CI. One major bottleneck is that the `*.json` files are pretty large and take a nontrivial amount of time to parse. Another issue is that these tests run a fair bit of code which with Miri's ~million-x slowdown. For the first problem I tried using other more-binary serialization formats but the JSON AST is unfortunately not compatible with many of them (e.g. `postcard`, which we use for Wasmtime, is not compatible with the JSON AST's structure in Rust types). For the latter I'm not sure what to do...

In the end I figured this might be a reasonable refactoring to go ahead and land anyway. It at least enables running `*.wast` tests in Miri while removing a large part of the runtime slowdown. We can in theory still allow-list some tests to run as they don't all take forever. Some future breakthrough is still going to be required though to run everything through Miri.

### DIFF
--- a/ci/miri-wast.sh
+++ b/ci/miri-wast.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+# Helper script to execute a `*.wast` test in Miri. This is only lightly used on
+# CI and is provided here to assist with development of anything that ends up
+# using unsafe for example.
+
+set -ex
+
+rm -rf ./miri-wast
+mkdir ./miri-wast
+cargo run -- wast --target pulley64 --precompile-save ./miri-wast "$@" \
+  -O memory-reservation=$((1 << 20)) \
+  -O memory-guard-size=0 \
+  -O signals-based-traps=n \
+  -O memory-init-cow=n
+
+MIRIFLAGS="$MIRIFLAGS -Zmiri-disable-isolation -Zmiri-permissive-provenance" \
+  cargo miri run -- wast -Ccache=n --target pulley64 --precompile-load ./miri-wast "$@" \
+  -O memory-init-cow=n
--- a/ci/miri-wast.sh
+++ b/ci/miri-wast.sh
@@ -3,6 +3,13 @@
 # Helper script to execute a `*.wast` test in Miri. This is only lightly used on
 # CI and is provided here to assist with development of anything that ends up
 # using unsafe for example.
+#
+# Example usage is:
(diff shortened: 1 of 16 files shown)


alexcrichton:
My rough hope here is that we can at least use this to run scripts during development, but I'm also curious what others' thoughts on this are.

ci/miri-wast.sh
fitzgen, line 1: Can you put a comment at the top with an example usage, showing the CLI args expected and all that?

fitzgen (commented):
In general, I like the approach.

We can potentially try `bincode` or `peekpoke` or something for the serialization.

alexcrichton:
Ok with the publication of wasm-tools this is now ready to go. I tried `bincode` and `postcard` but unfortunately neither works with the `#[serde]` attributes being used in `json-from-wast` which are done to match the `wast2json` output (mostly). This still ends up accelerating wast tests by a fair bit by skipping the whole compilation stage though, but there's still work to be done to improve the json parsing (or actually get it onto a binary format)

PR #11114, opened 2025-06-23
generalize async fiber abstraction

dicej:
As part of the work implementing the new Component Model async ABI in the `wasip3-prototyping` repo, I've generalized the `FiberFuture` abstraction in `wasmtime::runtime::store::async_` to support fibers which can either retain exclusive access to the store across suspend points or release it.  The latter allows the store to be used by the `component-model-async` event loop and/or other fibers to run before the original fiber resumes, which is the key to allowing multiple fibers to run concurrently, passing control of the store back and forth.

In the case of Pulley, the above generalization means we also need to give each fiber its own `Interpreter` so that multiple concurrent fibers don't clobber each other's state.

Concretely, this moves a lot of the code out of `async_.rs` and into a new `fiber.rs` submodule which will be shared with the `component-model-async` implementation.

This also pulls in a new `StoreToken<T>` utility which has been useful in `wasip3-prototyping` to safely convert from a `&mut dyn VMStore` to a `StoreContextMut<'a, T>` when we previously witnessed a conversion in the other direction.

Note that I've added a `'static` bound to the `VMStore` trait, which simplifies use of `&mut dyn VMStore`, avoiding thorny lifetime issues.

### DIFF
--- a/crates/wasmtime/Cargo.toml
+++ b/crates/wasmtime/Cargo.toml
@@ -194,8 +194,10 @@ async = [
   "dep:wasmtime-fiber",
   "dep:async-trait",
   "dep:trait-variant",
+  "dep:futures",
   "wasmtime-component-macro?/async",
   "runtime",
+  "futures/std",
 ]
 
 # Enables support for the pooling instance allocation strategy
--- a/crates/wasmtime/Cargo.toml
+++ b/crates/wasmtime/Cargo.toml
@@ -194,7 +194,6 @@ async = [
   "dep:wasmtime-fiber",
   "dep:async-trait",
   "dep:trait-variant",
-  "dep:futures",
   "wasmtime-component-macro?/async",
   "runtime",
 ]
--- a/crates/wasmtime/Cargo.toml
+++ b/crates/wasmtime/Cargo.toml
@@ -197,7 +197,6 @@ async = [
   "dep:futures",
   "wasmtime-component-macro?/async",
   "runtime",
-  "futures/std",
 ]
 
 # Enables support for the pooling instance allocation strategy
--- a/crates/wasmtime/src/runtime/fiber.rs
+++ b/crates/wasmtime/src/runtime/fiber.rs
@@ -94,7 +94,6 @@ unsafe impl Sync for AsyncState {}
 /// repeatedly in a loop until the future completes.
 pub(crate) struct AsyncCx {
     current_suspend: *mut *mut Suspend<Result<()>, StoreFiberYield, ()>,
-    current_stack_limit: *mut usize,
     current_poll_cx: *mut PollContext,
 }
 
@@ -118,7 +117,6 @@ impl AsyncCx {
         } else {
             Some(Self {
                 current_suspend: unsafe { &raw mut (*store.async_state()).current_suspend },
-                current_stack_limit: store.vm_store_context().stack_limit.get(),
                 current_poll_cx,
             })
         }
@@ -203,7 +201,6 @@ impl AsyncCx {
         unsafe {
             let reset_suspend = Reset(self.current_suspend, *self.current_suspend);
             *self.current_suspend = ptr::null_mut();
-            let _reset_stack_limit = Reset(self.current_stack_limit, *self.current_stack_limit);
             assert!(!(reset_suspend.1).is_null());
             (*reset_suspend.1).suspend(yield_)
         }
@@ -374,13 +371,26 @@ struct FiberResumeState {
     /// the current mask when this function is called and then restore the mask
     /// when the function returns (aka the fiber suspends).
     mpk: Option<ProtectionMask>,
+
+    /// The current wasm stack limit, if in use.
+    ///
+    /// This field stores the old of `VMStoreContext::stack_limit` that this
+    /// fiber should be using during its execution. This is saved/restored when
+    /// a fiber is suspended/resumed to ensure that when there are multiple
+    /// fibers within the store they all maintain an appropriate fiber-relative
+    /// stack limit.
+    stack_limit: usize,
 }
 
 impl FiberResumeState {
-    unsafe fn replace(self) -> PriorFiberResumeState {
+    unsafe fn replace(self, store: &mut StoreOpaque) -> PriorFiberResumeState {
         let tls = unsafe { self.tls.push() };
         let mpk = swap_mpk_states(self.mpk);
-        PriorFiberResumeState { tls, mpk }
+        PriorFiberResumeState {
+            tls,
+            mpk,
+            stack_limit: store.replace_stack_limit(self.stack_limit),
+        }
     }
 
     fn dispose(self) {
@@ -388,16 +398,33 @@ impl FiberResumeState {
     }
 }
 
+impl StoreOpaque {
+    /// Helper function to swap the `stack_limit` field in the `VMStoreContext`
+    /// within this store.
+    fn replace_stack_limit(&mut self, stack_limit: usize) -> usize {
+        // SAFETY: the `VMStoreContext` points to within this store itself but
+        // is accessed through raw pointers to assist with Miri. The `&mut
+        // StoreOpaque` passed to this function shows that this has permission
+        // to mutate state in the store, however.
+        unsafe { mem::replace(&mut *self.vm_store_context().stack_limit.get(), stack_limit) }
+    }
+}
+
 struct PriorFiberResumeState {
     tls: crate::runtime::vm::PreviousAsyncWasmCallState,
     mpk: Option<ProtectionMask>,
+    stack_limit: usize,
 }
 
 impl PriorFiberResumeState {
-    unsafe fn replace(self) -> FiberResumeState {
+    unsafe fn replace(self, store: &mut StoreOpaque) -> FiberResumeState {
         let tls = unsafe { self.tls.restore() };
         let mpk = swap_mpk_states(self.mpk);
-        FiberResumeState { tls, mpk }
+        FiberResumeState {
+            tls,
+            mpk,
+            stack_limit: store.replace_stack_limit(self.stack_limit),
+        }
     }
 }
 
@@ -431,13 +458,13 @@ pub(crate) fn resume_fiber<'a>(
 
     impl Drop for Restore<'_, '_> {
         fn drop(&mut self) {
-            self.fiber.state = Some(unsafe { self.state.take().unwrap().replace() });
+            self.fiber.state = Some(unsafe { self.state.take().unwrap().replace(self.store) });
             self.store.swap_executor(&mut self.fiber.executor);
         }
     }
     let result = unsafe {
         let _reset_suspend = Reset(fiber.suspend, *fiber.suspend);
-        let prev = fiber.state.take().unwrap().replace();
+        let prev = fiber.state.take().unwrap().replace(store);
         store.swap_executor(&mut fiber.executor);
         let restore = Restore {
             store,
@@ -528,6 +555,7 @@ pub(crate) fn make_fiber<'a>(
             } else {
                 None
             },
+            stack_limit: usize::MAX,
         }),
         engine,
         suspend,
--- a/crates/wasmtime/src/runtime/fiber.rs
+++ b/crates/wasmtime/src/runtime/fiber.rs
@@ -563,31 +563,26 @@ pub(crate) async fn on_fiber<R: Send + Sync>(
     let mut result = None;
     let fiber = make_fiber(store.traitobj_mut(), |store| result = Some(func(store)))?;
 
-    FiberFuture {
-        store,
-        fiber: Some(fiber),
-    }
-    .await;
+    FiberFuture { store, fiber }.await;
 
     Ok(result.unwrap())
 }
 
 struct FiberFuture<'a, 'b> {
     store: &'a mut StoreOpaque,
-    fiber: Option<StoreFiber<'b>>,
+    fiber: StoreFiber<'b>,
 }
 
 impl<'b> Future for FiberFuture<'_, 'b> {
-    type Output = StoreFiber<'b>;
+    type Output = ();
 
     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         let me = self.get_mut();
 
-        let fiber = me.fiber.take().unwrap();
-
         let poll_cx = unsafe { &raw mut (*me.store.async_state()).current_poll_cx };
         let _reset = Reset(poll_cx, unsafe { *poll_cx });
-        let (guard_range_start, guard_range_end) = fiber
+        let (guard_range_start, guard_range_end) = me
+            .fiber
             .fiber
             .as_ref()
             .unwrap()
@@ -622,21 +617,15 @@ impl<'b> Future for FiberFuture<'_, 'b> {
             };
         }
 
-        let mut fiber = Some(fiber);
-        match resume_fiber(me.store, fiber.as_mut().unwrap(), Ok(())) {
-            Ok(()) => Poll::Ready(fiber.take().unwrap()),
-            Err(_) => {
-                me.fiber = fiber;
-                Poll::Pending
-            }
+        match resume_fiber(me.store, &mut me.fiber, Ok(())) {
+            Ok(()) => Poll::Ready(()),
+            Err(_) => Poll::Pending,
         }
     }
 }
 
 impl Drop for FiberFuture<'_, '_> {
     fn drop(&mut self) {
-        if let Some(mut fiber) = self.fiber.take() {
-            fiber.dispose(self.store);
-        }
+        self.fiber.dispose(self.store);
     }
 }
--- a/crates/wasmtime/src/runtime/fiber.rs
+++ b/crates/wasmtime/src/runtime/fiber.rs
@@ -9,11 +9,11 @@ use anyhow::{Result, anyhow};
 use core::mem;
 use core::ops::Range;
 use core::pin::Pin;
-use core::ptr;
+use core::ptr::{self, NonNull};
 use core::task::{Context, Poll};
 use wasmtime_fiber::{Fiber, Suspend};
 
-type WasmtimeResume = Result<*mut Context<'static>>;
+type WasmtimeResume = Result<NonNull<Context<'static>>>;
 type WasmtimeYield = StoreFiberYield;
 type WasmtimeComplete = Result<()>;
 type WasmtimeSuspend = Suspend<WasmtimeResume, WasmtimeYield, WasmtimeComplete>;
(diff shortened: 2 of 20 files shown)


crates/wasmtime/src/runtime/store/token.rs
```diff
@@ -0,0 +1,36 @@
...
+    pub fn new(store: StoreContextMut<T>) -> Self {
+        Self {
+            id: store.0.id(),
+            _phantom: PhantomData,
+        }
+    }
+
+    pub fn as_context_mut<'a>(&self, store: &'a mut dyn VMStore) -> StoreContextMut<'a, T> {
+        assert_eq!(store.store_opaque().id(), self.id);
+        // We know the store with this ID has data type parameter `T` because
```
alexcrichton: Mind prefixing this with `SAFETY:`?

crates/wasmtime/src/runtime/store/token.rs
```diff
@@ -0,0 +1,36 @@
...
+        Self {
+            id: self.id,
+            _phantom: PhantomData,
+        }
+    }
+}
+
+impl<T> Copy for StoreToken<T> {}
+
+impl<T> StoreToken<T> {
```
alexcrichton: In retrospect let's add `'static` here. There's weird things with variance about extending or shortening lifetimes in types (e.g. being able to safely change a `StoreToken<T>` to a `StoreToken<U>` where just the lifetimes differ). Having `'static` means we wouldn't have to worry about it though.

crates/wasmtime/src/runtime/store/token.rs
```diff
@@ -0,0 +1,36 @@
+use crate::runtime::vm::VMStore;
+use crate::{StoreContextMut, store::StoreId};
+use core::marker::PhantomData;
+
+pub struct StoreToken<T> {
```
alexcrichton, line 11: Mind adding some brief comments as to the purpose of this type?

crates/wasmtime/src/runtime/linker.rs
```diff
@@ -1,3 +1,5 @@
+#[cfg(feature = "async")]
+use crate::fiber::AsyncCx;
```
alexcrichton: Could this be directly imported at use-sites below to avoid the #[cfg]'d import?

crates/wasmtime/src/runtime/fiber.rs
```diff
@@ -0,0 +1,700 @@
...
+    func: impl FnOnce(&mut StoreOpaque) -> R + Send,
+) -> Result<R> {
+    on_fiber_raw(store.traitobj_mut(), move |store| {
+        func((*store).store_opaque_mut())
+    })
+    .await
+}
+
+/// Wrap the specified function in a fiber and return it.
+fn prepare_fiber<'a, R: Send + 'static>(
```
alexcrichton: I was surprised how much `'static` bounds cropped up all over the place from this refactoring and this is what I ended up tracing it back to. This `'static` can be replaced with `'a` and with that I believe many transitive callers of this function can also remove `'static`

<details>

<summary>diff</summary>

```diff
diff --git a/crates/wasmtime/src/runtime/component/func/typed.rs b/crates/wasmtime/src/runtime/component/func/typed.rs
index f87447af7..c378742da 100644
--- a/crates/wasmtime/src/runtime/component/func/typed.rs
+++ b/crates/wasmtime/src/runtime/component/func/typed.rs
@@ -179,7 +179,7 @@ where
     ) -> Result<Return>
     where
         Params: Send + Sync,
-        Return: Send + Sync + 'static,
+        Return: Send + Sync,
     {
         let mut store = store.as_context_mut();
         assert!(
diff --git a/crates/wasmtime/src/runtime/fiber.rs b/crates/wasmtime/src/runtime/fiber.rs
index ba34e9e6b..76a89b3e7 100644
--- a/crates/wasmtime/src/runtime/fiber.rs
+++ b/crates/wasmtime/src/runtime/fiber.rs
@@ -585,7 +585,7 @@ unsafe fn suspend_fiber(

 /// Run the specified function on a newly-created fiber and `.await` its
 /// completion.
-pub(crate) async fn on_fiber<R: Send + 'static>(
+pub(crate) async fn on_fiber<R: Send>(
     store: &mut StoreOpaque,
     func: impl FnOnce(&mut StoreOpaque) -> R + Send,
 ) -> Result<R> {
@@ -596,7 +596,7 @@ pub(crate) async fn on_fiber<R: Send + 'static>(
 }

 /// Wrap the specified function in a fiber and return it.
-fn prepare_fiber<'a, R: Send + 'static>(
+fn prepare_fiber<'a, R: Send + 'a>(
     store: &mut dyn VMStore,
     func: impl FnOnce(&mut dyn VMStore) -> R + Send + 'a,
 ) -> Result<(StoreFiber<'a>, oneshot::Receiver<R>)> {
@@ -612,7 +612,7 @@ fn prepare_fiber<'a, R: Send + 'static>(

 /// Run the specified function on a newly-created fiber and `.await` its
 /// completion.
-async fn on_fiber_raw<R: Send + 'static>(
+async fn on_fiber_raw<R: Send>(
     store: &mut StoreOpaque,
     func: impl FnOnce(&mut dyn VMStore) -> R + Send,
 ) -> Result<R> {
diff --git a/crates/wasmtime/src/runtime/func/typed.rs b/crates/wasmtime/src/runtime/func/typed.rs
index e9304b755..0df0dcc40 100644
--- a/crates/wasmtime/src/runtime/func/typed.rs
+++ b/crates/wasmtime/src/runtime/func/typed.rs
@@ -132,10 +132,7 @@ where
         &self,
         mut store: impl AsContextMut<Data: Send>,
         params: Params,
-    ) -> Result<Results>
-    where
-        Results: 'static,
-    {
+    ) -> Result<Results> {
         let mut store = store.as_context_mut();
         assert!(
             store.0.async_support(),
diff --git a/crates/wasmtime/src/runtime/store/async_.rs b/crates/wasmtime/src/runtime/store/async_.rs
index c34d1dac4..f1c8a7454 100644
--- a/crates/wasmtime/src/runtime/store/async_.rs
+++ b/crates/wasmtime/src/runtime/store/async_.rs
@@ -171,7 +171,7 @@ impl StoreOpaque {
     /// This function will convert the synchronous `func` into an asynchronous
     /// future. This is done by running `func` in a fiber on a separate native
     /// stack which can be suspended and resumed from.
-    pub(crate) async fn on_fiber<R: Send + 'static>(
+    pub(crate) async fn on_fiber<R: Send>(
         &mut self,
         func: impl FnOnce(&mut Self) -> R + Send,
     ) -> Result<R> {
@@ -286,7 +286,7 @@ impl StoreOpaque {

 impl<T> StoreContextMut<'_, T> {
     /// Executes a synchronous computation `func` asynchronously on a new fiber.
-    pub(crate) async fn on_fiber<R: Send + 'static>(
+    pub(crate) async fn on_fiber<R: Send>(
         &mut self,
         func: impl FnOnce(&mut StoreContextMut<'_, T>) -> R + Send,
     ) -> Result<R>
```
</summary>

would that be possible to avoid the extra `'static` cropping up? I can sort of see how component-model-async bits may still require `'static`, but for preexisting bits in Wasmtime I'd be surprised if it required `'static`.

crates/wasmtime/Cargo.toml
```diff
@@ -194,8 +194,10 @@ async = [
   "dep:wasmtime-fiber",
   "dep:async-trait",
   "dep:trait-variant",
+  "dep:futures",
   "wasmtime-component-macro?/async",
   "runtime",
+  "futures/std",
```
alexcrichton: This runs a risk of causing problems because the `async` feature previously I believe didn't require the `std` feature, and now it does. Would it be possible to refactor to remove the need for the oneshot bits?

crates/wasmtime/src/runtime/store.rs
```diff
@@ -1926,21 +1932,41 @@ at https://bytecodealliance.org/security.
...
     pub(crate) fn async_guard_range(&self) -> core::ops::Range<*mut u8> {
         core::ptr::null_mut()..core::ptr::null_mut()
     }
 
     pub(crate) fn executor(&mut self) -> ExecutorRef<'_> {
-        match &mut self.executor {
+        #[cfg(feature = "async")]
+        let executor = unsafe { &mut *self.async_state.current_executor };
+        #[cfg(not(feature = "async"))]
+        let executor = &mut self.executor;
```
alexcrichton: I'm not sure how exactly yet, but I'd prefer to avoid this #[cfg] necessity just to access `self.executor` as it's very easy to forget an non-obvious that it's required.

crates/wasmtime/src/runtime/fiber.rs
alexcrichton: If possible I think it'd be good to do the same to the `Send` bounds as well, but I haven't gamed that out yet.

crates/wasmtime/src/runtime/store.rs
```diff
@@ -1926,21 +1932,41 @@ at https://bytecodealliance.org/security.
         )
     }
 
+    #[cfg(feature = "async")]
+    pub(crate) fn async_state(&mut self) -> *mut fiber::AsyncState {
```
alexcrichton: Would it be possible for this to return `&mut fiber::AsyncState` instead of a raw pointer?

crates/wasmtime/src/runtime/fiber.rs
```diff
@@ -0,0 +1,700 @@
+#![deny(unsafe_op_in_unsafe_fn)]
+
+use {
```
alexcrichton: FWIW this style of imports is currently largely nonexistent within Wasmtime. Would you be ok reformatting this in a style more consistent with the rest of Wasmtime?

crates/wasmtime/src/runtime/fiber.rs
```diff
@@ -0,0 +1,700 @@
...
+    Ok(rx.try_recv().unwrap().unwrap())
+}
+
+/// Wrap the specified function in a future which, when polled, will store a
+/// pointer to the `Context` in the `AsyncState::current_poll_cx` field for the
+/// specified store and then call the function.
+///
+/// This is intended for use with functions that resume fibers which may need to
+/// poll futures using the stored `Context` pointer.
+pub(crate) async fn poll_fn<R>(
```
alexcrichton: Where possible I'd like to aggressively remove `pub(crate)` from this file. This function for example in wasip3-prototyping is only used elsewhere in `Instance::resume_fiber` and the implementation looks like a good candidate for moving to this file as a method on `StoreFiber<'static>`, so could this function be made purely internal?

alexcrichton: Actually, on raeding this more, it looks like `Instance::resume_fiber` and `on_fiber_raw` above are more-or-less the same function where the only difference is `Instance::resume_fiber` can have a different type of return value which `on_fiber_raw` would assert never happens. Could the methods all get refactored/merged into this file?

alexcrichton: To expand on this some more, this is a pretty unsafe function and the boundaries of abstraction aren't really that clear to me. Understanding this function in isolation is pretty difficult and I feel it fits best within the context of `on_fiber_raw` or `Instance::resume_fiber`, which is where I'd prefer to keep all these unsafe set/reset/restore/etc bits all centralized in one function where possible so it's clear what's responsible for what.

crates/wasmtime/src/runtime/fiber.rs
```diff
@@ -0,0 +1,700 @@
...
+/// which it was created.
+///
+/// This will return `Ok(result)` if the fiber resolved, where `result` is the
+/// returned value; it will return `Err(yield_)` if the fiber suspended, where
+/// `yield_` indicates whether it released access to the store or not.  See
+/// `StoreFiber::fiber` for details.
+///
+/// SAFETY: The caller must confer exclusive access to the store to the fiber
+/// until the fiber is either dropped, resolved, or forgotten, or until it
+/// releases the store when suspending.
```
alexcrichton: This to me feels like it's a safety contract that's a bit too onerous to uphold since none of the callers of this function actually have a `&mut StoreOpaque` or similar in context. You'd have to travel a layer or two up the stack to actually see that and it feels a bit too "far" to me to find the proof of "ah, yes, ok this is safe". Would it be possible to actually take `&mut StoreOpaque` here as an argument? (that would also simplify the executor handling pieces I believe)

alexcrichton: One of the sticklers for taking an argument is probably `Drop for StoreFiber`, and that's something where we can soup up the `Drop for Store` to call a function over here to avoid relying on `Drop for StoreFiber` (or something like that)

alexcrichton: Well, on second thought, no maybe that's not a good idea. Fibers can't be dropped unless they've finished so that would introduce footguns of forgetting to call the destructor on panics for example (or something like that)...

If the dtor is a "special case" that seems somewhat reasonable

alexcrichton (commented):
I'll keep taking a look at `fiber.rs`, but this is at least an initial round of feedback.

crates/wasmtime/Cargo.toml
dicej: I _think_ you can use `oneshot` without enabling the `futures/std` feature, so we might be able to just remove this line.  I'll play around with it.

crates/wasmtime/src/runtime/store.rs
dicej: I believe we did this to satisfy miri, but I can retest with `&mut` and see what happens.

crates/wasmtime/src/runtime/fiber.rs
dicej: I've applied your patch to remove the `'static` bounds; let me know if you want me to do the same for `Send`.

alexcrichton: Yeah if possible I think we should remove `Send` here too as I think `'static` and `Send` are the root of https://github.com/bytecodealliance/wasip3-prototyping/issues/145 perhaps?

crates/wasmtime/src/runtime/store.rs
dicej: Yeah, miri won't tolerate that:
```
test scenario::round_trip::async_round_trip_stackful ... error: Undefined Behavior: attempting a read access using <3440249> at alloc1440939[0x18], but that tag does not exist in the borrow stack for this location
    --> /home/dicej/p/wasmtime/crates/wasmtime/src/runtime/fiber.rs:444:57
     |
444  |         let _reset_executor = Reset(fiber.executor_ptr, *fiber.executor_ptr);
     |                                                         ^^^^^^^^^^^^^^^^^^^
     |                                                         |
     |                                                         attempting a read access using <3440249> at alloc1440939[0x18], but that tag does not exist in the borrow stack for this location
     |                                                         this error occurs as part of an access at alloc1440939[0x18..0x20]
     |
     = help: this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
     = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
help: <3440249> was created by a SharedReadWrite retag at offsets [0x18..0x20]
    --> /home/dicej/p/wasmtime/crates/wasmtime/src/runtime/fiber.rs:476:18
     |
476  |         unsafe { &raw mut (*store.store_opaque_mut().async_state()).current_executor };
     |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
help: <3440249> was later invalidated at offsets [0x0..0x38] by a Unique retag
    --> /home/dicej/p/wasmtime/crates/wasmtime/src/runtime/store.rs:1947:9
     |
1947 |         &mut self.async_state
     |         ^^^^^^^^^^^^^^^^^^^^^
```

We might be able to address it a different way (e.g. a larger refactoring), though.  Not sure how deep that rabbit hole will go.

dicej: Oof, and it looks like miri is upset for other reasons on `wasip3-prototyping` `main`; I seem to have caused a regression with the fiber unification PR.  Debugging now.

dicej: Okay, just pushed a fix for that regression: https://github.com/bytecodealliance/wasmtime/pull/11114/commits/ae227b658643761ebb3603a428879a826e4f0762

crates/wasmtime/src/runtime/fiber.rs
```diff
@@ -0,0 +1,693 @@
...
+            state: Some(prev),
+        };
+        (*restore.fiber).fiber.as_ref().unwrap().resume(result)
+    }
+}
+
+/// Create a new `StoreFiber` which runs the specified closure.
+pub(crate) fn make_fiber<'a>(
+    store: &mut dyn VMStore,
+    fun: impl FnOnce(&mut dyn VMStore) -> Result<()> + 'a,
```
alexcrichton: Toying around a bit with wasip3-prototyping, for this to be more sound we're going to want to add `Send + Sync` to the bounds of `fun`. That makes the `unsafe impl Send/Sync` above a bit more sound because it at least accounts for all captured values within `fun`. The fallout wasn't too large from what I saw and seemed mostly as-expected

crates/wasmtime/src/runtime/fiber.rs
```diff
@@ -0,0 +1,693 @@
...
+    let executor = if cfg!(feature = "pulley") && engine.target().is_pulley() {
+        Executor::Interpreter(Interpreter::new(&engine))
+    } else {
+        Executor::Native
+    };
+    #[cfg(not(has_host_compiler_backend))]
+    let executor = {
+        debug_assert!(engine.target().is_pulley());
+        Executor::Interpreter(Interpreter::new(&engine))
+    };
```
alexcrichton: Oh I forgot to mention this earlier but FWIW this looks like it's copy/pasted from elsewhere and would be good to deduplicate.

crates/wasmtime/src/runtime/store.rs
dicej: We chatted about this over Zoom and concluded that changing the `*mut` to a `&mut` would require a larger-scale refactoring to make sound and miri-approved, so we're leaving it as-is for now.

crates/wasmtime/src/runtime/fiber.rs
dicej: I've pushed an update which addresses this.

alexcrichton (approved):
@dicej ok these are all the changes I'd like to make, mind reviewing the changes I made as well before merging? Afterwards feel free to go ahead and queue this for merge

crates/wasmtime/src/runtime/fiber.rs
```diff
@@ -526,37 +784,53 @@ pub(crate) fn make_fiber<'a>(
...
+    let fiber = Fiber::new(stack, move |result: WasmtimeResume, suspend| {
+        let future_cx = match result {
+            Ok(cx) => cx,
             // Cancelled before we started? Just return.
-            if result.is_err() {
-                return Ok(());
+            Err(_) => return Ok(()),
+        };
+
+        // SAFETY: Per the documented contract for
```
dicej: This comment is confusing since it refers to nonexistent docs for `resume_fiber`, and in any case we don't need to handwave about the "contract" anymore since `resume_fiber` now takes an explicit `&mut StoreOpaque` parameter.  So maybe just say something like "This fiber will only be resumed using `resume_fiber`, which takes a `&mut StoreOpaque` parameter and has given us exclusive access...".

PR #11115, opened 2025-06-23
x64: migrate many AVX instructions

abrown:
This migration of AVX instructions to use the new assembler is intended to be a large but quite mechanical conversion of all instructions that fit within the `Inst::XmmRmiRVex` variant. Unfortunately, three unimplemented SSE instructions (`vpmaddwd`, `vpmaddubsw`, `vpshufb`) prevent us from fully removing `Inst::XmmRmiRVex` and its ISLE helper, `xmm_rmir_vex`; we can remove those later once the SSE versions are implemented.

Beyond small format-related changes in Cranelift to accept the new VEX shapes, the largest change comes in Winch: because of how `AvxOpcode` was passed around previously, I was forced to create new constructors and refactor various bits. I resisted the temptation to go too far, but I could not resist making some of these constructors more consistent, e.g., using common suffixes (`rrr`, `rri`) and moving like instructions closer to each other.


cranelift/codegen/src/isa/x64/lower.isle
```diff
@@ -1290,22 +1290,22 @@
           (x Xmm x)
           (negated Xmm (x64_psubb (xmm_zero $I8X16) x))
         )
-        (x64_pminub x negated)))
+        (x64_pminub_a x negated)))
 
 (rule 1 (lower (has_type $I16X8 (iabs x)))
         (if-let true (use_ssse3))
-        (x64_pabsw x))
+        (x64_pabsw_a x))
```
alexcrichton: Should this (and various others around here) be using `*_or_avx`?

cranelift/filetests/filetests/isa/x64/fcvt-avx.clif
```diff
@@ -123,11 +123,11 @@ block0(v0: i64x2):
...
 ;   movq %rsp, %rbp
 ; block0:
-;   vpand   %xmm0, const(0), %xmm2
-;   vpor    %xmm2, const(1), %xmm4
-;   vpsrlq  %xmm0, $32, %xmm6
-;   vpor    %xmm6, const(2), %xmm0
-;   vsubpd  %xmm0, const(3), %xmm2
+;   vpand (%rip), %xmm0, %xmm2
+;   vpor (%rip), %xmm2, %xmm4
+;   psrlq $0x20, %xmm0
```
alexcrichton: Accidental regression from `vpsrlq` to `psrlq` I think

cranelift/filetests/filetests/isa/x64/simd-arith-avx.clif
```diff
@@ -1892,7 +1892,7 @@ block0(v0: i16x8):
 ;   pushq %rbp
 ;   movq %rsp, %rbp
 ; block0:
-;   vpabsw %xmm0, %xmm0
+;   pabsw %xmm0, %xmm0
```
alexcrichton: This (and a few below) regressed away from `v*`

tests/disas/winch/x64/f32x4_abs/const_avx.wat
```diff
@@ -14,28 +14,29 @@
 ;;       movq    0x10(%r11), %r11
 ;;       addq    $0x10, %r11
 ;;       cmpq    %rsp, %r11
-;;       ja      0x4f
+;;       ja      0x50
```
alexcrichton, line 17: This looks peculiar! This indicates that we might be placing an unnecessary byte which isn't required? Or that the old encoding for these instructions was 1-byte smaller than the encoding from the assembler?

(or maybe a rebase conflict as one of the PRs I landed recently removed some bytes and this might be adding them back in without having rebased on that? unsure)

tests/disas/winch/x64/v128_ops/not.wat
```diff
@@ -18,7 +18,7 @@
 ;;       movq    %rdi, 8(%rsp)
 ;;       movq    %rsi, (%rsp)
 ;;       movdqu  0x19(%rip), %xmm0
-;;       vpcmpeqd %xmm15, %xmm15, %xmm15
+;;       vpcmpeqb %xmm15, %xmm15, %xmm15
```
alexcrichton: To confirm, changing opcodes here is intentional?

alexcrichton (approved):

tests/disas/winch/x64/v128_ops/not.wat
abrown: It's a "generate all 1s" situation so I just changed it to the lowest lane version, but I'll change it back.

winch/codegen/src/isa/x64/asm.rs
```diff
@@ -2510,67 +2477,123 @@ impl Assembler {
...
+            OperandSize::S32 => asm::inst::vandps_b::new(dst, src1, address).into(),
+            OperandSize::S64 => asm::inst::vandpd_b::new(dst, src1, address).into(),
+            _ => unimplemented!(),
+        };
+        self.emit(Inst::External { inst });
     }
 
     /// Perform an `and` operation on vectors of floats in `src1` and `src2`
     /// and put the results in `dst`.
     pub fn xmm_vandp_rrr(&mut self, src1: Reg, src2: Reg, dst: WritableReg, size: OperandSize) {
```
abrown, line 2516: (referenced above)

tests/disas/winch/x64/f32x4_abs/const_avx.wat
abrown: Ok, looking at this again, I think we should just accept this (like you were pointing out when we talked). This is coming because this PR switched the order of the operands as it lowered the instruction and I think this operand swap makes more sense. Take a look at the constructor I tagged below: we pass `src1` and `src2` in the order one would expect. I'm just going to leave this.

PR #11122, opened 2025-06-24
x64: Optimize branch-with-`band`

alexcrichton:
This commit adds new lowering patterns for the x64 backend to optimize conditional branches with `band` patterns as the conditional. This builds on the infrastructure in the previous commit to funnel `icmp`-equals-zero back up to the top-level `is_nonzero_cmp` and then adding rules to `is_nonzero_cmp` for `band` to convert that to a `test` instruction which performs the `band` and sets the ZF which is what we're interested in.

### DIFF
--- a/cranelift/codegen/src/isa/x64/inst.isle
+++ b/cranelift/codegen/src/isa/x64/inst.isle
@@ -4248,6 +4248,8 @@
 (rule 2 (is_nonzero_cmp (vall_true vec)) (is_vall_true vec))
 (rule 2 (is_nonzero_cmp (vany_true vec)) (is_vany_true vec))
 (rule 2 (is_nonzero_cmp (uextend val)) (is_nonzero_cmp val))
+(rule 2 (is_nonzero_cmp (band a @ (value_type (ty_int (fits_in_64 ty))) b))
+  (CondResult.CC (x64_test ty a b) (CC.NZ)))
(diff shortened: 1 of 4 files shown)


fitzgen:
Yeah, for example given

```wat
    local.get 0
    local.get 1
    i32.or
    if ;; label = @1
      call 0
    else
      call 1
    end
```

we currently emit the comparison as

```asm
00000019    0b d1                             or edx, ecx
0000001b    85 d2                             test edx, edx
0000001d    0f 85 15 00 00 00                 jne 0x38
```

but we should be able to do something like

```asm
or edx, ecx
je 0x38
...
```

alexcrichton:
I think that makes sense to do yeah and was something I was curious about, but doing this robustly I think will require deeper refactorings of ISLE integration with the new assembler to, for example, auto-generate constructors which return `ProducesFlags` instead of `Gpr` as they do today. Instruction-selection for mem/imm/etc then also gets a bit tricky if we don't want to have too too much duplication. Basically `band` turned out to be easy because `test` has the exact semantics of `and` with the main benefit of a register isn't written, only flags.

So given the amount of refactoring/infrastructure I think the other patterns are going to need I'll defer that for a future change.

fitzgen:
I'll make a follow up issue to track this then

PR #11123, opened 2025-06-24
update `component-model-async` plumbing

dicej:
This pulls in the latest Component Model async ABI code from the `wasip3-prototyping` repo, including various API refactors and spec updates.

This includes all the changes to the `wasmtime` crate from `wasip3-prototyping` _except_ that the `concurrent` submodule and child submodules contain only non-functional stubs.  A follow-up PR will contain the real `concurrent` implementation.


crates/wasmtime/src/runtime/component/func/options.rs
```diff
@@ -67,12 +69,15 @@ impl Options {
         memory: Option<NonNull<VMMemoryDefinition>>,
         realloc: Option<NonNull<VMFuncRef>>,
         string_encoding: StringEncoding,
+        async_: bool,
+        _callback: Option<NonNull<VMFuncRef>>,
```
alexcrichton: Did this not end up getting stored in `Options`?

crates/wasmtime/src/runtime/component/values.rs
```diff
@@ -1043,3 +1110,21 @@ fn unexpected<T>(ty: InterfaceType, val: &Val) -> Result<T> {
...
+
+// TODO: This is prone to leaks and infinitely blocked writers given that
+// there's no reference counting and thus no way to know when to close the
+// handle.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct StreamAny(pub(crate) u32);
+
+// TODO: This is prone to leaks and infinitely blocked writers given that
+// there's no reference counting and thus no way to know when to close the
+// handle.
```
alexcrichton: Are these TODOs still applicable? If so mind converting these to doc comments?

crates/wasmtime/src/runtime/component/mod.rs
```diff
@@ -683,3 +699,151 @@ pub mod bindgen_examples;
 #[cfg(not(any(docsrs, test, doctest)))]
 #[doc(hidden)]
 pub mod bindgen_examples {}
+
+#[cfg(not(feature = "component-model-async"))]
+pub(crate) mod concurrent {
+    use {
```
alexcrichton: We'll want to smith this module a bit, mind moving this to a `concurrent_disabled.rs` or similar? Otherwise the big-ish changes we'll want here are:

* Types in this module should somehow contain an uninhabited enum. We've got `crate::runtime::Uninhabited` for this purpose and that'll ensure that we never accidentally create these types.
* Ideally there are no `panic!`s in this module. If anything takes `self` you can `match self.uninhabited {}` which is a compile-time assertion that it's not possible to be in that method.
* For other methods that are panicking they'll want to return an error/trap, at least where possible.

Basically the worry is that if we have substantial #[cfg]'d-out functionality it's basically never tested in this repo so we want to make sure it's as simple as it can be an "obviously correct" for its situation. In that sense my worry here is that I don't want anything to accidentally get into a situation where a runtime panic can happen. In theory many of these methods are dynamically unreachable but it's hard to infer that just from looking at this module, so I'd feel safer transforming things to returning an error to at least know that if we are wrong and these functions are reachable that they'll return some sort of error instead of panicking.

alexcrichton (commented):
These are a few comments so far but I've yet to really get into the meat of this. FWIW I also rebased https://github.com/bytecodealliance/wasmtime/pull/11127 on top of some of the local changes I've made, mostly to ensure that the changes don't break the full build.

alexcrichton:
This may not be plausible as I haven't gotten into the "meat" yet, but I'd like to pursue ideally the ability to land this PR as-is independent of https://github.com/bytecodealliance/wasmtime/pull/11127. Would it be possible to avoid changing `call_async` in this PR and https://github.com/bytecodealliance/wasmtime/pull/11127 would be where the implementation changes?

crates/wasmtime/src/runtime/component/func/options.rs
dicej: Sorry, I don't think I understand the question.  If you're wondering why the field name is prefixed with an underscore, it's because it won't be used until the full async implementation is merged.  I can add a comment and/or `#[cfg(allow(dead_code))]` if that helps.

crates/wasmtime/src/runtime/component/values.rs
dicej: Yes, they're still applicable.  I'll convert them.

crates/wasmtime/src/runtime/component/mod.rs
dicej: I started changing the panics into compile-time assertions or traps, but I quickly realized that in almost all cases the code really should be unreachable, since any attempt to use CM async features should have failed at `wasmparser` validation time.  It feels misleading to me to trap in those cases, because it implies that the code _is_ reachable, i.e. that a guest would somehow sneak use of an async feature past the validator.

I'll go ahead and switch to traps, but also add comments indicating that it's our (Wasmtime's) fault if they're hit, not the guest's.

dicej:
> Would it be possible to avoid changing `call_async` in this PR and #11127 would be where the implementation changes?

I'll take a shot at it.

dicej:
I believe I've addressed all the feedback so far ~~(except for the `Options` question, which I'll need clarification on)~~.

crates/wasmtime/src/runtime/component/func/options.rs
dicej: Nevermind, I understand the confusion now, and just pushed an update.

alexcrichton:
@dicej  ok I've reviewed everything except `func.rs`, `func/host.rs`, and `func/typed.rs` in this PR. I've made quite a few changes myself which I'd appreciate if you could read over as well to confirm you're ok with and such. I think that's enough async for me for today so I'm going to restart reviewing these file tomorrow.

dicej:
Thanks, @alexcrichton!  Your changes so far look reasonable to me.

alexcrichton:
@dicej ok https://github.com/bytecodealliance/wasmtime/pull/11123/commits/cba8a799d9f6584286081fe7b553f3c93d7d1b22 is the culmination of my attempt to keep `Params: 'static` out of this the async impl. That's functionally inert in this PR and won't get activated until the next PR. With that I'm going to call `typed.rs` and `func.rs` done. I'm still concerned about the two different paths through the code but this seems about as good as we can get it for now so I think it's best to tackle that in the next PR, and/or as a follow-up in the future.


I ran out of time to get to the rest of `host.rs` today, but I'll work on benchmarking tomorrow and scanning through `host.rs`.

crates/wasmtime/src/runtime/component/func/typed.rs
```diff
@@ -176,12 +180,47 @@ where
...
+        if false {
+            use crate::runtime::vm::SendSyncPtr;
+            use core::ptr::NonNull;
+
+            let ptr = SendSyncPtr::from(NonNull::from(&params).cast::<u8>());
+            let prepared = self.prepare_call(store.as_context_mut(), move |cx, ty, dst| {
+                // SAFETY: the goal here is to get `Params`, a non-`'static`
+                // value, to live long enough to the lowering of the
+                // parameters. We're guaranteed that `Params` lives in the
+                // future of the oute function (we're in an `async fn`) so it'll
```
dicej:
```suggestion
                // future of the outer function (we're in an `async fn`) so it'll
```

crates/wasmtime/src/runtime/component/func/typed.rs
```diff
@@ -176,12 +180,47 @@ where
...
+                //
+                // Here a pointer to `Params` is smuggled to this location
+                // through a `SendSyncPtr<u8>` to thwart the `'static` check of
+                // rustc and the signature of `prepare_call`.
+                let params = unsafe { ptr.cast::<Params>().as_ref() };
+                Self::lower_args(cx, ty, dst, params)
+            })?;
+
+            // TODO: need to place a dtor on the stack referencing the store
+            // which resmoves `prepared.task` from the store to ensure that the
```
dicej:
```suggestion
            // which removes `prepared.task` from the store to ensure that the
```

crates/wasmtime/src/runtime/component/func/typed.rs
```diff
@@ -176,12 +180,47 @@ where
...
+                // for example, from the signature of `call_concurrent` below.
+                //
+                // Here a pointer to `Params` is smuggled to this location
+                // through a `SendSyncPtr<u8>` to thwart the `'static` check of
+                // rustc and the signature of `prepare_call`.
+                let params = unsafe { ptr.cast::<Params>().as_ref() };
+                Self::lower_args(cx, ty, dst, params)
+            })?;
+
+            // TODO: need to place a dtor on the stack referencing the store
```
dicej, line 216: This part is definitely critical.

alexcrichton: Yeah my thinking is that this'll be easier to write up in the next PR once `concurrent.rs` is filled out

alexcrichton:
Benchmarking update:

As-is this PR yields these numbers in the preexisting `call.rs` benchmarks:

<details>

image

</details>

not amazing, not awful.

I ran `./benches/wasmtime-serve-rps.sh` and saw no measurable difference with before/after.

alexcrichton (approved):
Ok @dicej I'm comfortable with everything here. Mind giving it another once-over on your end to confirm too? If you're ok with it all feel free to enqueue for merge.

alexcrichton:
I'll say I was originally planning to squash everything into one commit but we had some merge commits along the way which makes that slightly annoying so I'm content to let github merges figure out how to squash it instead.

alexcrichton:
(also, to reiterate, thank you again for being patient with my review on this!)

dicej:
Thanks for persisting with this, @alexcrichton!  Your changes look good to me.

PR #11124, opened 2025-06-24
 x64: Refactor emission of LoadExtName 

alexcrichton:
Don't use raw bytes when emitting this instruction but instead use
symbolic assembler directives to avoid needing to hardcode bytes. This
additionally fixes an issue where with `colocated` symbols the offset of
the symbol was not taken into account (but for Wasmtime it's always been
0 so this otherwise hasn't come up so far)

### DIFF
--- a/cranelift/assembler-x64/src/api.rs
+++ b/cranelift/assembler-x64/src/api.rs
@@ -2,9 +2,9 @@
 
 use crate::gpr;
 use crate::xmm;
-use crate::{Amode, GprMem, XmmMem};
+use crate::{Amode, DeferredTarget, GprMem, XmmMem};
 use std::fmt;
-use std::{num::NonZeroU8, ops::Index, vec::Vec};
+use std::{num::NonZeroU8, vec::Vec};
 
 /// Describe how an instruction is emitted into a code buffer.
 pub trait CodeSink {
@@ -24,17 +24,13 @@ pub trait CodeSink {
     /// required for assembling memory accesses.
     fn add_trap(&mut self, code: TrapCode);
 
-    /// Return the byte offset of the current location in the code buffer;
-    /// required for assembling RIP-relative memory accesses.
-    fn current_offset(&self) -> u32;
+    /// Inform the code buffer of a use of `target` is about to happen within
+    /// this buffer at the current offset. After this method is called bytes are
+    /// then placed at this offset.
+    fn use_target(&mut self, target: DeferredTarget);
 
-    /// Inform the code buffer of a use of `label` at `offset`; required for
-    /// assembling RIP-relative memory accesses.
-    fn use_label_at_offset(&mut self, offset: u32, label: Label);
-
-    /// Return the label for a constant `id`; required for assembling
-    /// RIP-relative memory accesses of constants.
-    fn get_label_for_constant(&mut self, id: Constant) -> Label;
+    /// Resolves a `KnownOffset` value to the actual signed offset.
+    fn known_offset(&self, offset: KnownOffset) -> i32;
 }
 
 /// Provide a convenient implementation for testing.
@@ -57,14 +53,10 @@ impl CodeSink for Vec<u8> {
 
     fn add_trap(&mut self, _: TrapCode) {}
 
-    fn current_offset(&self) -> u32 {
-        self.len().try_into().unwrap()
-    }
+    fn use_target(&mut self, _: DeferredTarget) {}
 
-    fn use_label_at_offset(&mut self, _: u32, _: Label) {}
-
-    fn get_label_for_constant(&mut self, c: Constant) -> Label {
-        Label(c.0)
+    fn known_offset(&self, offset: KnownOffset) -> i32 {
+        panic!("unknown offset {offset:?}")
     }
 }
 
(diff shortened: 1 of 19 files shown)


cfallin (approved):
Looks reasonable; thanks! Good catch handling the offsets.

cfallin:
(The `Reg` -> `Gpr` refactoring tripped up in the `cranelift-codegen` tests, it seems)

alexcrichton:
Ok to resolve the test failure I had to push up a semi-substantial change as the final commit here. cc @abrown on the assembler change, and mind taking another look @cfallin?

cranelift/assembler-x64/src/api.rs
```diff
@@ -24,17 +24,13 @@ pub trait CodeSink {
     /// required for assembling memory accesses.
     fn add_trap(&mut self, code: TrapCode);
 
-    /// Return the byte offset of the current location in the code buffer;
-    /// required for assembling RIP-relative memory accesses.
-    fn current_offset(&self) -> u32;
+    /// Inform the code buffer of a use of `target` is about to happen within
```
cfallin: grammar fix -- "inform ... of a use ... is about to happen" -> "inform ... that a use ... is about to happen"

cranelift/assembler-x64/src/api.rs
```diff
@@ -24,17 +24,13 @@ pub trait CodeSink {
...
+    /// this buffer at the current offset. After this method is called bytes are
+    /// then placed at this offset.
```
cfallin: s/the bytes are then placed/the bytes are then expected to be placed/

(it's part of the contract for what the caller must do, rather than a result of this method call, right?)

cranelift/codegen/src/isa/x64/inst/external.rs
```diff
@@ -429,37 +429,59 @@ pub mod offsets {
...
+        }
     }
 
-    fn get_label_for_constant(&mut self, c: asm::Constant) -> asm::Label {
-        self.get_label_for_constant(c.into()).into()
+    fn known_offset(&self, offset: asm::KnownOffset) -> i32 {
+        match offset {
+            offsets::KEY_INCOMING_ARG => self.incoming_arg_offset,
+            offsets::KEY_SLOT_OFFSET => self.slot_offset,
+            _ => unreachable!(),
```
cfallin: Maybe a panic here with a panic message including `offset`? Technically it's unreachable but the unreachability invariant is "remote" (part of another crate) so I'd rather make this easier debug / more verbose if we do miss something later...

cfallin (approved):
New approach looks reasonable, but deferring to @abrown for final opinions on the `CodeSink` refactor...

cfallin:
(Was this auto-added to merge queue? I gave r+ but Andrew should still weigh in and there were a few comment nits)

alexcrichton:
Oops I forgot to disable that and pushes from maintainers doesn't auto-disable it, thanks for catching!

cranelift/codegen/src/isa/x64/inst/external.rs
```diff
@@ -429,37 +429,59 @@ pub mod offsets {
...
-impl asm::CodeSink for MachBuffer<Inst> {
+/// Implementor of the [`asm::CodeSink`] trait.
+pub struct AsmCodeSink<'a> {
+    /// The buffer this is emitting into.
+    pub sink: &'a mut MachBuffer<Inst>,
+    /// The value of `KEY_INCOMING_ARG`.
+    pub incoming_arg_offset: i32,
+    /// The value of `KEY_SLOT_OFFSET`.
+    pub slot_offset: i32,
+}
```
abrown, line 440: Yup, this is better! Thanks for this refactor; I had been worrying a bit about how to get rid of the offset table.

cranelift/codegen/src/isa/x64/inst/external.rs
```diff
@@ -429,37 +429,59 @@ pub mod offsets {
...
     }
 
-    fn get_label_for_constant(&mut self, c: asm::Constant) -> asm::Label {
-        self.get_label_for_constant(c.into()).into()
+    fn known_offset(&self, offset: asm::KnownOffset) -> i32 {
+        match offset {
+            offsets::KEY_INCOMING_ARG => self.incoming_arg_offset,
+            offsets::KEY_SLOT_OFFSET => self.slot_offset,
+            other => panic!("unknown \"known\" offset {other}"),
+        }
```
abrown, line 484: I guess the decision here is to keep `asm::KnownOffset` as an index instead of converting it into an `enum`... I suppose there might be some generic way to pass around a `cranelift-codegen`-specific `KnownOffset` type to be a bit safer. But, honestly, that could come much later. This is better already.

abrown (approved):

cranelift/codegen/src/isa/x64/inst/external.rs
alexcrichton: Yeah I was brainstorming how to get a type in there (also for things like `ExternalName` or relocations) and my conclusion was that we'd have to change `enum Amode<R: AsReg>` to `enum Amode<R: Registers>` and then also have:

```rust
trait Registers {
    type CodeSink: CodeSink;
    // ...
}
```

and then store `<R::CodeSink as CodeSink>::KnownOffset` within an `Amode`.

That all gets pretty hairy and relatively complicated so I decided to not go down that route quite yet...

PR #11126, opened 2025-06-24
Remove duplicate entries from the addrmap section

alexcrichton:
This commit is a minor tweak to the construction of the addrmap section which Wasmtime uses to map binary offsets in compiled code to binary offsets in the original wasm module. The representation of the addrmap section means that if Cranelift reports two adjacent instructions that have the same source location we don't actually have to record the second one, only the first. This commit implements this optimization by skipping encoding duplicate addrmap entries to shrink the size of this section in binaries slightly.


alexcrichton:
As an example this reduces the size of `.wasmtime.addrmap` for `spidermonkey.wasm` from 7886708 bytes to 7082452 bytes, a 10% reduction.

PR #11127, opened 2025-06-24
implement Component Model async ABI

dicej:
This commit replaces the stub functions and types in `wasmtime::runtime::component::concurrent` and its submodules with the working implementation developed in the `wasip3-prototyping` repo.  For ease of review, it does not include any new tests; I'll add those in a follow-up commit.


alexcrichton:
@dicej on this one thing I want to write down before I forget (I've forgotten this many times so far...)

Index allocation for futures/streams/error-context I believe are going to need to be updated here. I believe the current state of this PR is such that there's one indexing namespace for all resources within a component, and then there's another index namespace for futures/streams/error-context. Spec-wise I believe it's required we unify these index namespaces. I realize that this is an artifact of the history of the implementation here, but it's something that I want to be sure to write down and not forget.

Effectively my naive expectation would be that [this type](https://github.com/bytecodealliance/wasmtime/blob/57ba95e92eff1d02f6b05ab86b9e81b35bd69d28/crates/wasmtime/src/runtime/vm/component/resources.rs#L92-L120) grows to include `Stream`/`Future`/`ErrorContext` variants and a table or similar in `ConcurrentState` will end up being removed.

alexcrichton:
Another thought I've now had which I don't want to forget: this is introducing a significant `#[cfg]` in the path of calling a component function with the async style. That means that the non-cm-async implementation will be entirely untested in CI which is not an ideal state to be in. I don't know how best to handle this but I wanted to note it before I forget.

crates/wasmtime/src/runtime/component/concurrent/states.rs
alexcrichton, line 1: IIRC this file is a duplicate of `vm/component/resources.rs` right? (or at least a historic one). IIRC we talked about deferring resolving this duplication to later, but mind filing an issue and registering a FIXME here?

crates/wasmtime/src/runtime/component/concurrent/table.rs
```diff
@@ -0,0 +1,355 @@
+// TODO: This duplicates a lot of resource_table.rs; consider reducing that
+// duplication.
```
alexcrichton: Mind either (a) resolving this TODO in this PR or (b) filing an issue for this TODO to point to and pointing to that?

crates/wasmtime/src/runtime/component/concurrent/table.rs
```diff
@@ -0,0 +1,355 @@
...
+use std::{
+    any::Any,
+    boxed::Box,
+    cmp::Ordering,
+    collections::BTreeSet,
+    fmt,
+    hash::{Hash, Hasher},
+    marker::PhantomData,
+    vec::{self, Vec},
+};
```
alexcrichton: I manually changed the previous PR, but mind updating the import style in the new files to match the preexisting style in Wasmtime?

crates/wasmtime/src/runtime/component/func/typed.rs
```diff
@@ -189,15 +189,26 @@ where
...
+                    self.0.store(ptr::null_mut(), Relaxed);
+                }
+            }
+
+            // TODO: Can we make this more efficient while guaranteeing the
+            // closure we pass to `prepare_call` below never sees a stale
+            // pointer?
+            let mut params = params;
+            let ptr = Arc::new(AtomicPtr::new((&raw mut params).cast::<u8>()));
+            let _clear = ClearOnDrop(ptr.clone());
```
alexcrichton: Could this be done by removing the task on `Drop` instead of allocating an extra `Arc`?

alexcrichton: Also, mind running some of the basic, typed, no-call-hook, etc, benchmarks for host-to-wasm to see the impact of this new code path?

crates/wasmtime/src/runtime/component/func.rs
```diff
@@ -654,6 +667,13 @@ impl Func {
     fn post_return_impl(&self, mut store: impl AsContextMut) -> Result<()> {
         let mut store = store.as_context_mut();
 
+        #[cfg(feature = "component-model-async")]
+        if store.0.async_support() {
+            // In this case, the post-return function will already have been
+            // called.
+            return Ok(());
+        }
+
```
alexcrichton: I'm wary of this in that we don't want to change the public API contract required depending on crate features being enabled or disabled. This implementation looks like it goes from requriing `post_return` to not requiring it, which would be a regression for the public API. Could this be integrated differently to still require post return when `call_async` is used? (it's fine to me that `call_concurrent` is documented as not requiring post-return)

crates/wasmtime/src/runtime/component/func.rs
```diff
@@ -284,9 +284,22 @@ impl Func {
             store.0.async_support(),
             "cannot use `call_async` without enabling async support in the config"
         );
-        store
-            .on_fiber(|store| self.call_impl(store, params, results))
-            .await?
+        #[cfg(feature = "component-model-async")]
+        {
+            let future = self.call_concurrent(store.as_context_mut(), params.to_vec());
```
alexcrichton: Can you file an issue about removing `to_vec` here? I'm remembering now that while I synchronized the async/concurrent signature in `Linker` for dynamic things I think that's something we're going to want to do here as well for `call_concurrent`. The signature will change slightly but I think it will still be doable (and like the linker ones it'd use `Vec` internally for now)

crates/wasmtime/src/runtime/component/concurrent.rs
```diff
@@ -1,163 +1,2535 @@
...
+            self.store.as_context_mut(),
+            self.accessor.clone_for_spawn(),
+            task,
+        )
+    }
+
+    /// Retrieve the component instance of the caller.
+    pub fn instance(&self) -> Instance {
+        self.accessor.instance()
+    }
```
alexcrichton, line 242: Could this be made `pub(crate)`?

crates/wasmtime/src/runtime/component/concurrent.rs
```diff
@@ -1,163 +1,2535 @@
...
+    {
+        let instance = self.instance;
+        let accessor = self.clone_for_spawn();
+        self.with(|mut access| {
+            instance.spawn_with_accessor(access.as_context_mut(), accessor, task)
+        })
+    }
+
+    /// Retrieve the component instance of the caller.
+    pub fn instance(&self) -> Instance {
```
alexcrichton, line 459: Like above, can this be `pub(crate)`

crates/wasmtime/src/runtime/component/concurrent.rs
```diff
@@ -41,123 +124,2330 @@ impl Status {
...
+        &mut self.suspend_reason
+    }
+
+    fn worker(&mut self) -> &mut Option<StoreFiber<'static>> {
+        &mut self.worker
+    }
+
+    fn guest_call(&mut self) -> &mut Option<GuestCall> {
+        &mut self.guest_call
+    }
```
alexcrichton: Are these perhaps relics of a historical implementation? (similar with methods above). I'd otherwise imagine that the method-based accessors could all get removed in favor of accessing the fields directly

crates/wasmtime/src/runtime/component/concurrent.rs
```diff
@@ -284,14 +2858,143 @@ impl Instance {
...
+                caller_instance,
+            }),
+        )
+    }
+
+    /// Implements the `yield` intrinsic.
+    pub(crate) fn yield_(self, store: &mut dyn VMStore, async_: bool) -> Result<bool> {
+        self.waitable_check(store, async_, WaitableCheck::Yield)
+            .map(|_code| {
+                // TODO: plumb cancellation to here
```
alexcrichton: Mind filing an issue for this?

crates/wasmtime/src/runtime/component/concurrent.rs
```diff
@@ -1012,20 +3639,792 @@ unsafe impl<T> VMComponentAsyncStore for StoreInner<T> {
...
+    fn default() -> Self {
+        Self {
+            spawned_tasks: Vec::new(),
+            current_instance: None,
+        }
+    }
+}
+
+unsafe impl Send for AsyncState {}
+unsafe impl Sync for AsyncState {}
```
alexcrichton: We'll want to do something about this (e.g. document, remove, etc).

In toying around locally though I see that the reason for this is `Vec<HostTaskFuture>` which is only `Send`, not also `Sync`. I think we'll want to keep that in the sense that we don't want to require futures to all be both `Sync` and `Send`, only `Send`, if possible.

Also in toying with this though I can remove the `spawned_tasks` field entirely and tests all pass. Given that do you know of a reason to keep this field, and if not can it be removed? (once removed it's possible to remove these `unsafe impl` blocks)

crates/wasmtime/src/runtime/component/concurrent.rs
```diff
@@ -1012,20 +3639,792 @@ unsafe impl<T> VMComponentAsyncStore for StoreInner<T> {
...
+    struct DummyWaker;
+
+    impl Wake for DummyWaker {
+        fn wake(self: Arc<Self>) {}
+    }
+
+    static WAKER: Lazy<Arc<DummyWaker>> = Lazy::new(|| Arc::new(DummyWaker));
+
+    WAKER.clone().into()
+}
```
alexcrichton: Mind removing this and replacing it with `Waker::noop()`?

crates/wasmtime/src/runtime/component/concurrent.rs
dicej: This is intentionally public to allow host function implementations to call `Instance::{future,stream}`, for which they need an `Instance` object.

Given that `Instance` is `Copy`, I suppose we could require the host function closure to close over an `Instance` from the top-level code (i.e. the code that actually creates the instance).  The ergonomics wouldn't be as nice, though, plus it creates the hazard that the host embedder code mixes up its `Instance`s and closes over the wrong one.

crates/wasmtime/src/runtime/component/concurrent.rs
dicej: See my reply above.

crates/wasmtime/src/runtime/component/concurrent.rs
dicej: I think I just got carried away with writing accessor methods; I can get rid of them.

crates/wasmtime/src/runtime/component/concurrent.rs
dicej: > Also in toying with this though I can remove the spawned_tasks field entirely and tests all pass

Sounds like we're missing test coverage, then -- probably because we were leaning on the `wasmtime-wasi[-http]` integration tests to exercise this and never wrote any non-wasi tests for it.  I'm sure some of the wasi tests will fail if you remove that field.

crates/wasmtime/src/runtime/component/func/typed.rs
dicej: > Could this be done by removing the task on `Drop` instead of allocating an extra `Arc`?

I had originally planned to do that, but couldn't see how to avoid shared mutable aliasing of the store.  Revisiting it now, I can see we just need a custom `Future` implementation which has a `StoreContextMut` field, similar to what we did for `FiberFuture`.  I'll do that.

crates/wasmtime/src/runtime/component/concurrent.rs
dicej: BTW, I'm fine with changing the field to a `Mutex<Vec<HostTaskFuture>>` if that's what we need to remove the `unsafe impl`s.

crates/wasmtime/src/runtime/component/func.rs
dicej: Can you elaborate on how you'd like to see this code changed?  Naively, I could imagine changing this to look more like `TypedFunc::call_async`, but that would require the same `unsafe` use of a raw pointer, which I was trying to avoid here.

alexcrichton: Ideally we'd have `fn call_concurrent<'a>(self, store: ..., params: &'a [Val], results: &'a mut [Val]) -> Pin<Box<... + 'a>>` where the returned future would close over `params` and `results`. Internally that'd do something like `Linker::func_new_concurrent` where `Vec` is still used internally to satisfy the `'static` needed to put the future into the store but otherwise the public signature wouldn't change.

crates/wasmtime/src/runtime/component/concurrent.rs
alexcrichton: This sort of API makes it very similar to `Caller` that we have for core wasm which is something I've long wanted to delete from the public API, but it's pretty load-bearing so we probably can't remove that. I've historically wanted to resist any similar temptation to add such a similar call to components as well (e.g. there's no `Caller` on any `component::Linker` methods, it's all just `StoreContextMut`).

I'll try to revisit this when reading over futures and streams code and see if I can't formulate more thoughts about this.

crates/wasmtime/src/runtime/component/concurrent.rs
alexcrichton: Currently in this PR there are no reads of the `spawned_tasks` field which is why tests in the `cm-async-tests` PR (which I realize doesn't include `wasmtime-wasi{,-http}`) pass. Does this mean that some changes from upstream wasip3-prototyping accidentally weren't merged back here?

dicej: > Currently in this PR there are no reads of the `spawned_tasks` field

Maybe I'm misunderstanding, but isn't [this](https://github.com/dicej/wasmtime/blob/c135259dff3ed16d87668c24d7f5698727f85a35/crates/wasmtime/src/runtime/component/concurrent.rs#L3160-L3171) a read of that field?

alexcrichton: Er yes, bad words on my part. Basically nothing ever pushes on that vector so it's always empty

dicej: Ok, then yes, this looks like dead code.  We used to push spawned tasks to a vector in `Access[or]::spawn` but now we call `Instance::spawn_with_accessor` instead, so you're right that the field can be removed.  I'll do that.

crates/wasmtime/src/runtime/component/func.rs
dicej: I've pushed an update to require `post_return_async` for `call_async` and added docs to `call_concurrent` indicating that the `post-return` function is called automatically in that case.

crates/wasmtime/src/runtime/component/func/typed.rs
dicej: > Also, mind running some of the basic, typed, no-call-hook, etc, benchmarks for host-to-wasm to see the impact of this new code path?

Would you mind providing instructions for doing that?  I ran `cargo bench`, but that ran a bunch of other, irrelevant stuff as well.  How did you get your "before vs. after" numbers on the last PR, and how did you specify which ones to run?

alexcrichton: Oh sure yeah, I do:

```
cargo bench --bench call -- --list
```

first, and then this accepts a regex argument to pare down what's being run so I build up a regex to just the benchmarks that are relevant. For example

```
cargo bench --bench call -- --list '^sync.*component'
```

focuses just on sync calls (the fastest ones) and component-related ones. Here we only want host-to-wasm 

```
cargo bench --bench call -- --list '^sync.*component.*host-to-wasm'
```

and eventually what you probably want to run is:

```
cargo bench --bench call -- --list '^sync/no-hook.*component.*host-to-wasm.* typed .*' 
```

alexcrichton: What I'd also recommend is using 

```
$previous_command --save-baseline main
```

on the `main` branch or without the component-model-async feature. Then with this PR you'd run:

```
$previous_command --baseline main
```

which will automate criterion reporting differences to the previously captured baseline

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -20,14 +584,116 @@ use wasmtime_environ::component::{
...
+    ) -> Result<Self> {
+        let Val::Future(FutureAny(rep)) = value else {
+            bail!("expected `future`; got `{}`", value.desc());
+        };
+        let store = store.as_context_mut();
+        instance
+            .concurrent_state_mut(store.0)
+            .get(TableId::<TransmitHandle>::new(*rep))?; // Just make sure it's present
+        Ok(Self::new(*rep, instance))
+    }
```
alexcrichton, line 635: There's no protections here that `T` is correct?

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -39,8 +700,10 @@ unsafe impl<T: Send + Sync> func::ComponentType for HostFuture<T> {
     type Lower = <u32 as func::ComponentType>::Lower;
 
     fn typecheck(ty: &InterfaceType, _types: &InstanceType<'_>) -> Result<()> {
-        _ = ty;
-        todo!()
+        match ty {
+            InterfaceType::Future(_) => Ok(()),
```
alexcrichton, line 706: I think this wants to further type-check that the payload type matches `T`, e.g. `T::typecheck(...)`

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -74,33 +739,226 @@ unsafe impl<T: Send + Sync> func::Lift for HostFuture<T> {
...
+impl<T> FutureReader<T> {
+    fn new(rep: u32, tx: Option<mpsc::Sender<ReadEvent<Option<T>>>>, instance: Instance) -> Self {
+        Self { instance, rep, tx }
+    }
+
+    /// Read the value from this `future`.
+    ///
+    /// The returned `Future` will yield `None` if the guest has trapped
+    /// before it could produce a result or if the write end belonged to the
+    /// host and was dropped without writing a result.
```
alexcrichton: Here "write end dropped without writing" is a documentation-of-eld, right?

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -1,16 +1,575 @@
...
+    /// Write the specified value to this `future`.
+    ///
+    /// The returned `Future` will yield `true` if the read end accepted the
+    /// value; otherwise it will return `false`, meaning the read end was dropped
+    /// before the value could be delivered.
+    ///
+    /// Note that the returned `Future` must be polled from the event loop of
+    /// the component instance from which this `FutureWriter` originated.  See
+    /// [`Instance::run`] for details.
+    pub fn write(mut self, value: T) -> impl Future<Output = bool> + Send + 'static
```
alexcrichton, line 513: Wherever possible I personally find it much easier to reason about/read `async fn` signatures rather than those returning `impl Future`. Would it be possible to change this, and a number of other functions in this file that just return `impl Future` to `async fn`? I believe the `Send`-ness of the future here will be correctly inferred.

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -74,33 +739,226 @@ unsafe impl<T: Send + Sync> func::Lift for HostFuture<T> {
...
+                        Box::pin(me.write_all(buffer)).await
+                    } else {
+                        (Some(me), buffer)
+                    }
+                } else {
+                    (None, buffer)
+                }
+            }),
+        )
+    }
```
alexcrichton, line 933: Along the lines of a comment about `async fn` above, I think this would be much clearer as a loop over the buffer calling `self.write` rather than via recursion.

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -74,33 +739,226 @@ unsafe impl<T: Send + Sync> func::Lift for HostFuture<T> {
...
+    /// number of items the reader accepted).  It will return `(None, _)` if the
+    /// write failed due to the closure of the read end.  In either case, the
+    /// returned buffer will be the same one passed as a parameter, possibly
+    /// mutated to consume any written values.
+    ///
+    /// Note that the returned `Future` must be polled from the event loop of
+    /// the component instance from which this `StreamWriter` originated.  See
+    /// [`Instance::run`] for details.
+    pub fn write(
+        mut self,
```
alexcrichton, line 879: Would it be possible to have this be `&mut self`? I realize that would mean changing how "things are dropped" is modeled but personally I find that style of ownership much easier to work with than the move-in-and-move-out style this is currently using.

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -1,16 +1,575 @@
...
+    /// See [`Instance::run`] for details.
+    pub fn watch_reader(mut self) -> (impl Future<Output = ()> + Send + 'static, Watch<Self>)
+    where
+        T: Send + 'static,
+    {
+        let (tx, rx) = oneshot::channel();
+        send(&mut self.tx.as_mut().unwrap(), WriteEvent::Watch { tx });
+        let instance = self.instance;
+        watch(instance, rx, self)
+    }
```
alexcrichton, line 558: Would it be possible to change all these `watch_*` methods to `async fn watch(&mut self) { ... }`? That I think is a simple signature to model the various interactions here:

* By taking `&mut self` it's effectively "taking ownership", aka preventing anything else happening.
* `Watch::into_inner` is now dropping the returned future.

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -112,14 +970,111 @@ pub struct FutureReader<T> {
...
+        Val::Stream(StreamAny(self.rep))
+    }
+
+    /// Attempt to convert the specified [`Val`] to a `HostStream`.
+    pub fn from_val(
+        mut store: impl AsContextMut<Data: Send>,
+        instance: Instance,
+        value: &Val,
+    ) -> Result<Self> {
+        let Val::Stream(StreamAny(rep)) = value else {
```
alexcrichton, line 1018: Similar to futures above I think this `rep` needs to be type-checked against `T`

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -131,8 +1086,10 @@ unsafe impl<T: Send + Sync> func::ComponentType for HostStream<T> {
     type Lower = <u32 as func::ComponentType>::Lower;
 
     fn typecheck(ty: &InterfaceType, _types: &InstanceType<'_>) -> Result<()> {
-        _ = ty;
-        todo!()
+        match ty {
+            InterfaceType::Stream(_) => Ok(()),
```
alexcrichton, line 1091: Similar to above the payload here needs to be type-checked against `T`

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -166,52 +1125,172 @@ unsafe impl<T: Send + Sync> func::Lift for HostStream<T> {
...
+    /// (possibly with zero items if the write was empty).  It will return
+    /// `(None, _)` if the read failed due to the closure of the write end.  In
+    /// either case, the returned buffer will be the same one passed as a
+    /// parameter, with zero or more items added.
+    ///
+    /// Note that the returned `Future` must be polled from the event loop of
+    /// the component instance from which this `StreamReader` originated.  See
+    /// [`Instance::run`] for details.
+    pub fn read(
+        mut self,
```
alexcrichton, line 1182: Like writers above, could this be an `async fn` with `&mut self` as a receiver?

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams/buffers.rs
```diff
@@ -0,0 +1,292 @@
...
+pub unsafe trait TakeBuffer {
+    /// Take ownership of the specified number of items.
+    ///
+    /// The items are passed to `fun` as a raw pointer which may be cast to the
+    /// type indicated by the specified `TypeId`.
+    fn take(&mut self, count: usize, fun: &mut dyn FnMut(TypeId, *const u8));
+}
+
+/// Trait representing a buffer which may be written to a `StreamWriter`.
+#[doc(hidden)]
```
alexcrichton: This trait, and `ReadBuffer` below, show up in public trait bounds so these need to be public documented traits. Ideally with links to what methods use them and such.

crates/wasmtime/Cargo.toml
```diff
@@ -391,4 +392,7 @@ component-model-async = [
   "component-model",
   "std",
   "wasmtime-component-macro?/component-model-async",
+  "dep:futures",
+  "futures/std",
+  "dep:bytes",
```
alexcrichton, line 402: My understanding is that this is not intrinsic to the feature, is that right? This is only here for the `{Take,Read,Write}Buffer` implementations?

If so can this be made an optional feature to avoid requiring it?

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams/buffers.rs
```diff
@@ -0,0 +1,292 @@
...
+pub trait WriteBuffer<T>: TakeBuffer + Send + Sync + 'static {
+    /// Slice of items remaining to be read.
+    fn remaining(&self) -> &[T];
+    /// Skip and drop the specified number of items.
+    fn skip(&mut self, count: usize);
+}
+
+/// Trait representing a buffer which may be used to read from a `StreamReader`.
+#[doc(hidden)]
+pub trait ReadBuffer<T>: Send + Sync + 'static {
```
alexcrichton, line 111: FWIW I find the terminology here pretty confusing. I understand you pass a `ReadBuffer` to a read operation, but the only thing you can do with a `ReadBuffer` is write to it. I have not spent much time trying to find better names, but I wanted to write this down nonetheless.

crates/wasmtime/src/runtime/component/func/typed.rs
dicej: I've replaced the `Arc` with code that removes the task when the future is resolved or dropped.  I've also opened https://github.com/bytecodealliance/wasmtime/issues/11209 to track future optimization work.

crates/wasmtime/src/runtime/component/func.rs
dicej: Issue filed: https://github.com/bytecodealliance/wasmtime/issues/11218

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
dicej: Issue filed: https://github.com/bytecodealliance/wasmtime/issues/11219

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
dicej: Issue filed: https://github.com/bytecodealliance/wasmtime/issues/11219

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
dicej: Issue filed: https://github.com/bytecodealliance/wasmtime/issues/11219

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
dicej: Issue filed: https://github.com/bytecodealliance/wasmtime/issues/11219

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams/buffers.rs
dicej: Yeah, I struggled with the names of `WriteBuffer` and `ReadBuffer`, too.  What about `SourceBuffer` and `DestinationBuffer` instead?

alexcrichton: I'm going to defer this to https://github.com/bytecodealliance/wasmtime/issues/11223

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
dicej: I've opened an issue for this: https://github.com/bytecodealliance/wasmtime/issues/11222

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
dicej: I've opened an issue for this: https://github.com/bytecodealliance/wasmtime/issues/11222

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
dicej: I've opened an issue for this: https://github.com/bytecodealliance/wasmtime/issues/11222

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
dicej: Issue filed: https://github.com/bytecodealliance/wasmtime/issues/11219

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
dicej: I've opened an issue for this: https://github.com/bytecodealliance/wasmtime/issues/11222

crates/wasmtime/src/runtime/component/concurrent.rs
dicej: Issue filed: https://github.com/bytecodealliance/wasmtime/issues/11226

crates/wasmtime/src/runtime/component/concurrent.rs
dicej: Issue filed: https://github.com/bytecodealliance/wasmtime/issues/11226

PR #11128, opened 2025-06-24
x64: Add lowerings for the `bt` instruction

alexcrichton:
This commit builds on the previous refactorings to leverage the `bt` instruction on x64 which moves a specific bit into the CF bit of EFLAGS which can be useful for comparisons/branches/selects/etc.

### DIFF
--- a/cranelift/codegen/src/isa/x64/lower/isle.rs
+++ b/cranelift/codegen/src/isa/x64/lower/isle.rs
@@ -1080,6 +1080,15 @@ impl Context for IsleContext<'_, '_, MInst, X64Backend> {
         self.emit(&MInst::External { inst: inst.into() });
         ret.to_reg()
     }
+
+    fn bt_imm(&mut self, val: u64) -> Option<u8> {
+        let trailing_zeros = u8::try_from(val.trailing_zeros()).unwrap();
+        if val == 1 << trailing_zeros {
+            Some(trailing_zeros)
+        } else {
+            None
+        }
+    }
 }
 
 impl IsleContext<'_, '_, MInst, X64Backend> {
--- a/cranelift/codegen/src/isa/x64/lower/isle.rs
+++ b/cranelift/codegen/src/isa/x64/lower/isle.rs
@@ -1082,9 +1082,8 @@ impl Context for IsleContext<'_, '_, MInst, X64Backend> {
     }
 
     fn bt_imm(&mut self, val: u64) -> Option<u8> {
-        let trailing_zeros = u8::try_from(val.trailing_zeros()).unwrap();
-        if val == 1 << trailing_zeros {
-            Some(trailing_zeros)
+        if val.count_ones() == 1 {
+            Some(u8::try_from(val.trailing_zeros()).unwrap())
         } else {
             None
         }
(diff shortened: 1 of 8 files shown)


cranelift/filetests/filetests/isa/x64/branches.clif
```diff
@@ -1217,3 +1217,316 @@ block2:
...
+; block2: ; offset 0x10
+;   movl $0xc8, %eax
+;   movq %rbp, %rsp
+;   popq %rbp
+;   retq
+; block3: ; offset 0x1a
+;   movl $0x64, %eax
+;   movq %rbp, %rsp
+;   popq %rbp
+;   retq
```
fitzgen, line 1270: Do we expect this to use a `bt` instruction? If not, then why not?

cranelift/filetests/filetests/isa/x64/branches.clif
```diff
@@ -1217,3 +1217,316 @@ block2:
...
+; block1:
+;   movl $0xc8, %eax
+;   movq %rbp, %rsp
+;   popq %rbp
+;   retq
+; block2:
+;   movl $0x64, %eax
+;   movq %rbp, %rsp
+;   popq %rbp
+;   retq
```
fitzgen, line 1304: Same here?

cranelift/codegen/src/isa/x64/lower/isle.rs
```diff
@@ -1080,6 +1080,15 @@ impl Context for IsleContext<'_, '_, MInst, X64Backend> {
...
         ret.to_reg()
     }
+
+    fn bt_imm(&mut self, val: u64) -> Option<u8> {
+        let trailing_zeros = u8::try_from(val.trailing_zeros()).unwrap();
+        if val == 1 << trailing_zeros {
+            Some(trailing_zeros)
+        } else {
+            None
+        }
```
fitzgen: Slightly easier to understand, for me at least, like this:

```suggestion
        if val.count_ones() == 1 {
            Some(u8::try_from(val.trailing_zeros()).unwrap())
        } else {
            None
        }
```

cranelift/filetests/filetests/isa/x64/branches.clif
alexcrichton: It's intentional yeah, and corresponds to [this lowering rule](https://github.com/bytecodealliance/wasmtime/pull/11128/files#diff-5921dd103c30cf61da9e9befccac790585d04b0d2abe56c8f87f84816eba5676R4295-R4299). My rationale is "LLVM is [probably smarter than I](https://godbolt.org/z/frsh7s1cT)" and I'm assuming LLVM did things like micro-op tests or something to figure out that `test` has a lower latency than `bt` or something like that. No idea if that's actually the case though, just mirroring them.

cfallin: According to [Agner Fog's tables](https://www.agner.org/optimize/instruction_tables.pdf), taking Skylake as a random reasonable baseline µarch, `test` has one cycle latency and a throughput of 4/cycle (four ALU ports can execute it) while `bt` has one cycle latency and a throughput of 2/cycle (two ALU ports can execute it). So probably reasonable to prefer `test` to `bt` where possible.

alexcrichton: Nice that sounds like an excellent reference to put as documentation in the lowering rules so I'll do that.

fitzgen (approved):
Thanks!

Issue #6049, opened 2023-03-17
Port divide-by-constant magic number optimizations from `simple_preopt` to ISLE mid-end


fitzgen:
This is the old simple preopt stuff for some number divided/remaindered by a non-power-of-two constant:

* https://github.com/bytecodealliance/wasmtime/blob/465913eb2c91998c99ae9222e47f8e9f9a88a546/cranelift/codegen/src/divconst_magic_numbers.rs#L41-L217

* https://github.com/bytecodealliance/wasmtime/blob/465913eb2c91998c99ae9222e47f8e9f9a88a546/cranelift/codegen/src/simple_preopt.rs#L206-L243

The ISLE + e-graphs mid-end work replaced `simple_preopt`, but never ported these magic constants over. We should port them over.

HOWEVER, right now we can't replace potentially-trapping instructions in ISLE, and we need that ability to do this.

jameysharp:
Specifically, this depends on #5908.

PR #11129, opened 2025-06-24
Cranelift: Revive division-by-constant magic sequences

fitzgen:
* Revive the old module for generating division-by-constant magic numbers

* Add proptests for divide-by-constant magic number generators, just for a little extra correctness assurances

* Apply division-by-constant magic sequences in the ISLE mid-end.

  The application of these sequences have been ported to ISLE now to better integrate with our existing mid-end optimizations and to allow for recursive rule application in the optimized right-hand sides.

Fixes https://github.com/bytecodealliance/wasmtime/issues/6049

### DIFF
--- a/cranelift/codegen/src/opts/arithmetic.isle
+++ b/cranelift/codegen/src/opts/arithmetic.isle
@@ -123,10 +123,9 @@
       (apply_div_const_magic_s64 (Opcode.Sdiv) x d))
 
 ;; x % 1 == 0
-(rule (simplify_skeleton (urem x (iconst_u ty 1)))
-      (iconst_u ty 0))
-(rule (simplify_skeleton (srem x (iconst_u ty 1)))
-      (iconst_u ty 0))
+(rule (simplify_skeleton (urem x (iconst_u ty 1))) (iconst_u ty 0))
+(rule (simplify_skeleton (srem x (iconst_u ty 1))) (iconst_u ty 0))
+(rule (simplify_skeleton (srem x (iconst_s ty -1))) (iconst_u ty 0))
 
 ;; Unsigned `x % d == x & ((1 << ilog2(d)) - 1)` when `d` is a power of two.
 (rule (simplify_skeleton (urem x (iconst_u ty (u64_extract_power_of_two d))))
@@ -141,6 +140,7 @@
       ;; powers of two.
       (if-let true (i64_is_any_sign_power_of_two d))
       (if-let true (i64_ne d 1))
+      (if-let true (i64_ne d -1))
       (let ((k u32 (i64_trailing_zeros d))
             (t1 Value (sshr ty x (iconst_u ty (u32_sub k 1))))
             (t2 Value (ushr ty t1 (iconst_u ty (u32_sub (ty_bits ty) k))))


cranelift/codegen/src/opts/div_const.rs
```diff
@@ -0,0 +1,1148 @@
...
+        let t2: u64 = y as u64 & 0xffffffffu64;
+        let t3: i64 = y >> 32;
+        let t4: u64 = t0 * t2;
+        let t5: i64 = t1 * t2 as i64 + (t4 >> 32) as i64;
+        let t6: u64 = t5 as u64 & 0xffffffffu64;
+        let t7: i64 = t5 >> 32;
+        let t8: i64 = t0 as i64 * t3 + t6 as i64;
+        let t9: i64 = t1 * t3 + t7 + (t8 >> 32);
+        t9
+    }
```
alexcrichton: FWIW while these were probably copied from C or similar you can probably simplify these greatly with 128-bit integers

alexcrichton: Also this sort of makes me wonder if we could have a monster ISLE rule to pattern match this and generate `smulhi`....

alexcrichton (approved):
Very nice 👍. Review-wise the main thing I did was double-check that the `eval_*` functions match the ISLE lowerings of the instructions to ensure that all the goodness of the proptest/test suite/etc mirrors into ISLE well. Very much appreciate the extensive test suite here!

A few thoughts:

* Could the proptest bits be expanded to `{u,s}rem` as well as division? That way the final bits of ISLE can all additionally be mapped into Rust/proptest as well.
* Is it true that in all cases this expansion is better than whatever the hardware does?
* More of a curiosity, but how do egraphs know to favor these instruction sequences? This is expanding to quite a few more instrunctions than a div so do we have the cost of div/rem set really high?

cfallin:
> Is it true that in all cases this expansion is better than whatever the hardware does?

I guess I'm becoming the resident instruction latency/throughput oracle today but yes definitely -- 64-bit integer divide has a latency of 35-88 cycles (!) on Skylake, with a throughput of 1/6 (so not fully pipelined), while 64-bit integer multiply is 3 cycles and is fully pipelined (1 per cycle). I vaguely recall some issue discussion from when we first switched over to the egraph mid-end where we noticed some benchmarks where this showed up as minor regressions, too. Very happy to see this finally ported over!

cranelift/codegen/src/opts/div_const.rs
fitzgen: The tricky bit with matching this gunk to produce `smulhi` is all of the commutative operations.

fitzgen:
> More of a curiosity, but how do egraphs know to favor these instruction sequences? This is expanding to quite a few more instrunctions than a div so do we have the cost of div/rem set really high?

We don't put the values defined by side-effectful instructions into the egraph, so we have a slightly different path. This is why we use the ISLE `simplify_skeleton` term instead of the usual `simplify`. This path assumes that all RHSes are an improvement over the LHS, and does a shallow/greedy cost comparison for choosing among multiple RHSes (rather than a dynamic programming thing that looks at deep structure).

Concretely, in this case there will be one RHS candidate (the new magic sequence), the one RHS is trivially the best RHS, and for `simplify_skeleton` we always take RHSes over LHSes when available. So basically, "we favor these instructions because we generated them".

cranelift/codegen/src/opts/div_const.rs
jameysharp: I think what you're describing is the draft work I did, and comments I wrote, in https://github.com/bytecodealliance/wasmtime/pull/8653; I was trying to get all the way to wide imul, but I think similar reasoning should work for *mulhi.

PR #11135, opened 2025-06-25
Add shuf and pshuf variants to the new assembler

jlb6740:

### DIFF
--- a/cranelift/assembler-x64/meta/src/instructions/lanes.rs
+++ b/cranelift/assembler-x64/meta/src/instructions/lanes.rs
@@ -62,9 +62,15 @@ pub fn list() -> Vec<Inst> {
         inst("vblendvpd", fmt("RVMR", [w(xmm1), r(xmm2), r(xmm_m128), r(xmm3)]), vex(L128)._66()._0f3a().w0().op(0x4B).r().is4(), _64b | compat | avx),
 
         // Shuffle lanes in various ways.
+        inst("shufpd", fmt("A", [rw(xmm1), r(align(xmm_m128)), r(imm8)]), rex([0x66, 0x0F, 0xC6]).ib(), _64b | compat | sse2).alt(avx, "vshufpd_b"),
+        inst("vshufpd", fmt("B", [w(xmm1), r(xmm2), r(xmm_m128), r(imm8)]), vex(L128)._66()._0f().ib().op(0xC6), _64b | compat | avx),
+        inst("shufps", fmt("A", [rw(xmm1), r(align(xmm_m128)), r(imm8)]), rex([0x0F, 0xC6]).ib(), _64b | compat | sse).alt(avx, "vshufps_b"),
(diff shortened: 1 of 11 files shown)


cranelift/assembler-x64/meta/src/instructions/lanes.rs
```diff
@@ -60,9 +60,15 @@ pub fn list() -> Vec<Inst> {
         inst("vblendvpd", fmt("RVMR", [w(xmm1), r(xmm2), r(xmm_m128), r(xmm3)]), vex(L128)._66()._0f3a().w0().op(0x4B).r().is4(), _64b | compat | avx),
 
         // Shuffle lanes in various ways.
+        inst("shufpd", fmt("A", [rw(xmm1), r(align(xmm_m128)), r(imm8)]), rex([0x66, 0x0F, 0xC6]).ib(), _64b | compat | sse2).alt(avx, "vshufpd_b"),
```
jlb6740, line 65: Note I've added support for shufpd and vshufpd but these instructions did not exist before. Assembler testing passes but there is no filetest and I'd guess this is just dead code. Should I just remove these instructions?

cfallin: Since there is separately the Capstone-based fuzzing for the assembler, I think it's probably fine to fill things out like this where it's not much extra work, and it might be nice to have (this library at least aspires to be a separately usable general-purpose assembler someday, right?).

cfallin (approved):
Thanks!

cfallin:
It looks like there are some merge conflicts -- happy to merge once those are fixed.

cranelift/codegen/src/isa/x64/inst.isle
```diff
@@ -794,6 +792,7 @@
             Vpmaddubsw
             Vinsertps
             Vpshufb
+            Vshufpd
```
abrown: @jlb6740 , somehow this is added now and it shouldn't be. This is never used right?

jlb6740: @abrown I added this because vshufpd is a new instruction that wasn't supported before. Adding it makes it consistent with what is there now. It seems we've kept this AVX enum intact here and here: cranelift/codegen/src/isa/x64/inst/args.rs . Perhaps we are able too delete these avx instructions from both these files but vshufpd wouldn't be the only one .. pack, align, insert, phadd, etc all would need removing and so perhaps we can do that all at once?

abrown: There is no `Vshufpd` on `main`:

https://github.com/bytecodealliance/wasmtime/blob/a6d124480f6bfeed8e9f45d9586362fb01da6928/cranelift/codegen/src/isa/x64/inst.isle#L463-L467

The `args.rs` file is just reflecting this list and it is not there either. I'm not sure I follow why you think it should be added, but we shouldn't, since I am in the process of removing this entire thing.

PR #11143, opened 2025-06-26
x64: migrate FMA instructions

abrown:
This is a very straightforward replacement of all of the FMA instructions that Cranelift might need. Apparently Cranelift never uses the `231` variant but I retained them in the assembler for completeness.


fitzgen (approved):
Nice!

PR #11145, opened 2025-06-27
Remove the check for ':' when setting authority in wasi-http.

rylev:
Authorities can have `:` in them when they are IPV6 addresses. With the current check it is impossible to use ipv6 addresses as authorities for outgoing http requests. 

It's not clear to me why the check was added in the first place, so I might be missing something obvious here why the check is necessary.

### DIFF
--- a/crates/test-programs/src/bin/http_outbound_request_response_build.rs
+++ b/crates/test-programs/src/bin/http_outbound_request_response_build.rs
@@ -48,9 +48,7 @@ fn main() {
                 .is_err()
         );
 
-        assert!(req.set_authority(Some("bad-port:99999")).is_err());
         assert!(req.set_authority(Some("bad-\nhost")).is_err());
-        assert!(req.set_authority(Some("too-many-ports:80:80:80")).is_err());
 
         assert!(
             req.set_scheme(Some(&http_types::Scheme::Other("bad\nscheme".to_string())))
--- a/crates/test-programs/src/bin/http_outbound_request_response_build.rs
+++ b/crates/test-programs/src/bin/http_outbound_request_response_build.rs
@@ -49,6 +49,9 @@ fn main() {
         );
 
         assert!(req.set_authority(Some("bad-\nhost")).is_err());
+        // IPv6 addresses with and without a port should be allowed
+        assert!(req.set_authority(Some("[::]:443")).is_ok());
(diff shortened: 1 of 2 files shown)


dicej:
~~It looks like the check for `:` was an attempt to validate [this](https://github.com/WebAssembly/wasi-http/blob/d97efe470e4d1e3f959162d81eef9666335bc186/wit/types.wit#L370-L373):~~

> The HTTP and HTTPS schemes always require an authority. Fails if the string given is not a syntactically valid URI authority.

~~Presumably, that means it should _not_ contain a port number.  So if we update the check to ensure it doesn't end with a colon followed by zero or more digits (e.g. the regex `:[0-9]*`), that should allow IPv6 addresses but still disallow port numbers.~~  EDIT: Disregard what I wrote.  Reading the code more carefully, I see it was not trying to disallow specifying the port.  I don't know what it was trying to do 🤷

badeend:
From [the PR that introduced this check](https://github.com/bytecodealliance/wasmtime/pull/7451), I can't reverse engineer why this check was added. 

The URI authority  _can_ include the port. As matter of fact, it is the _only_ way in the wasi-http interface to declare the port number.

Taking the code at face value, I _think_ what it tried to check for is that: when a port is provided it shall be a valid u16. For example: `http://bad-port:99999` is syntactically a valid URI authority, but `99999` is not a valid port number.
A better way to write this could be:
```rs
if auth.port().is_some() && auth.port_u16().is_none() {
   return Ok(Err(()));
}
```

Maybe there's still value in keeping such a validation around.

Other than that: I agree with the general sentiment of this PR: `http::uri::Authority::from_str` alone is likely better equipped to validate URI's than ad-hoc validations in wasmtime itself.

Could you add a test that checks a IPv6 address +/- a port number passes the validation? E.g. `"[::]:443"`

rylev:
FYI: the check suggested of...

```rust
if auth.port().is_some() && auth.port_u16().is_none() {
   return Ok(Err(()));
}
```

... won't work as `auth.port()` already parses the port as a u16. So if `auth.port().is_some()` is true - `auth.port_u16()` will also be `Some`. 

I have pushed a test for ipv6 with and without a port.

badeend:
Hmm. I see. In that case, I don't know why that check was there.

Issue #11141, opened 2025-06-26
binary-trees gc benchmark runs really slow in wasmtime


jrmuizel:
### Test Case

[binary-trees-wasi-opt.wasm.zip](https://github.com/user-attachments/files/20929191/binary-trees-wasi-opt.wasm.zip)

### Steps to Reproduce

* Run `time wasmtime run -W gc=y binary-trees-wasi-opt.wasm`

### Actual Results

```
stretch tree of depth 21 check: 4194303
1048576 trees of depth 4 check: 32505856
262144 trees of depth 6 check: 33292288
65536 trees of depth 8 check: 33488896
16384 trees of depth 10 check: 33538048
4096 trees of depth 12 check: 33550336
1024 trees of depth 14 check: 33553408
256 trees of depth 16 check: 33554176
64 trees of depth 18 check: 33554368
16 trees of depth 20 check: 33554416
long lived tree of depth 20 check: 2097151

real	2m13.350s
user	2m12.736s
sys	0m0.454s
```

### Expected Results

It should run much faster.
V8: 1.2s
Firefox: 4.8s
Safari: 5.9s

### Extra Info

The benchmark is:
```java
class TreeNode {
    TreeNode left;
    TreeNode right;
    
    TreeNode bottomUpTree(int depth) {
        if (depth > 0) {
            TreeNode leftChild = this.bottomUpTree(depth - 1);
            TreeNode rightChild = this.bottomUpTree(depth - 1);
            TreeNode node = new TreeNode();
            node.left = leftChild;
            node.right = rightChild;
            return node;
        } else {
            TreeNode node = new TreeNode();
            node.left = null;
            node.right = null;
            return node;
        }
    }
    
    int itemCheck() {
        if (this.left == null) {
            return 1;
        } else {
            return 1 + this.left.itemCheck() + this.right.itemCheck();
        }
    }
}

void main() {
    int minDepth = 4;
    int n = 20;  // Increase n to make it more interesting
    
    int maxDepth;
    if (minDepth + 2 > n) {
        maxDepth = minDepth + 2;
    } else {
        maxDepth = n;
    }
    int stretchDepth = maxDepth + 1;
    
    TreeNode stretchTree = new TreeNode();
    TreeNode temp = stretchTree.bottomUpTree(stretchDepth);
    int check = temp.itemCheck();
    print("stretch tree of depth ");
    print(stretchDepth);
    print(" check: ");
    println(check);
    
[...]

PR #11148, opened 2025-06-27
Avoid cloning and dropping `Arc`s in `LocalMemory::vmmemory`

fitzgen:
Getting the base pointer from the underlying `dyn RuntimeLinearMemory` involved getting a `MemoryBase` which is potentially an `MmapOffset` which itself contains an `Arc<Mmap>`, and we would then call `base.as_non_null()` to turn this into a raw pointer, and then we would drop the `MemoryBase` which ultimately drops the `Arc<Mmap>`. So that's an `Arc` clone and drop just to get a `VMMemoryDefinition`, which is just a pointer and a length, essentially a slice of the linear memory. And, among other places, we call `LocalMemory::vmmemory` to get the GC heap's memory base and bound every time we access a GC object from Rust (for example, during collections).

Altogether, this removes another 30% of runtime from the testcase in https://github.com/bytecodealliance/wasmtime/issues/11141

### DIFF
--- a/crates/wasmtime/src/runtime/trampoline/memory.rs
+++ b/crates/wasmtime/src/runtime/trampoline/memory.rs
@@ -82,6 +82,10 @@ impl RuntimeLinearMemory for LinearMemoryProxy {
     fn base(&self) -> MemoryBase {
         MemoryBase::new_raw(self.mem.as_ptr())
     }
+
+    fn base_non_null(&self) -> core::ptr::NonNull<u8> {
+        core::ptr::NonNull::new(self.mem.as_ptr()).unwrap()
+    }
 }
 
 #[derive(Clone)]
--- a/crates/wasmtime/src/runtime/trampoline/memory.rs
+++ b/crates/wasmtime/src/runtime/trampoline/memory.rs
@@ -83,8 +83,12 @@ impl RuntimeLinearMemory for LinearMemoryProxy {
         MemoryBase::new_raw(self.mem.as_ptr())
     }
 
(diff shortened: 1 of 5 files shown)


fitzgen:
All that said, I don't love how this PR introduces a second version of the "same" method. It might be nicer to instead add a `vmmemory` method to `RuntimeLinearMemory` and push `VMMemoryDefinition` construction down to implementations of that trait, rather than tying together the results of different `RuntimeLinearMemory` methods in `LocalMemory`.

That would actually let us avoid an indirect call too:

* Now:
  * `LocalMemory` indirect-calls the `byte_size` method through a `dyn RuntimeLinearMemory`
  * `LocalMemory` indirect-calls the `base` (now `base_non_null` in this PR) method through a `dyn RuntimeLinearMemory`
  * `LocalMemory` puts them together into a `VMMemoryDefinition`
* Alternative:
  * `LocalMemory` indirect-calls the hypothetical `vmmemory` method through a `dyn RuntimeLinearMemory`
  * The `RuntimeLinearMemory` implementation gets its byte size and base, which can both be done directly/inlined/etc because we aren't crossing a `dyn` boundary anymore
  * The `RuntimeLinearMemory` implementation puts them together into a `VMMemoryDefinition`
  * The `VMMemoryDefinition` is returned to the `LocalMemory`, which returns it to its caller in turn

Note that the "now" sequence has two indirect calls while the alternative has only one.

I'm going to whip this alternative up real quick and then update this PR.

fitzgen:
> I'm going to whip this alternative up real quick and then update this PR.

I've pushed a commit that implements this.

PR #11149, opened 2025-06-27
x64: finish migrating SSE and AVX instructions

abrown:
This completes the migration of SSE and AVX instructions to the new assembler. Each commit has more details, but the overall summary is that a couple of instructions are added (e.g., `insertps`, `pack*`) and many already-migrated instructions allow removing the `SseOpcode` and `AvxOpcode` enumerations. This enables a cascade of dead code removal: mainly dead `Inst` variants but also some now-unused emission code.


cfallin (approved):
Great to see a diffstat like `+114 −1,257` -- thanks for pushing through the last of these!

PR #11150, opened 2025-06-27
x64: Migrate indirect jmp to the new assembler

alexcrichton:
This migrates `JmpUnknown` from ISLE to the new assembler. Other jumps are going to be more complicated so they're deferred to later.


alexcrichton:
Ah yeah this is only created right now during emission so doesn't show up in the `VCode` prints anywhere, so no changes to formatting are expected.

PR #11152, opened 2025-06-27
fix(no_std)!: respect `std` feature when target is windows/unix

salmans:
This PR fixes feature flag handling to properly respect the std feature when targeting Windows/Unix platforms.

Breaking Change: This fix will disable standard library features for dependents that use wasmtime with `default-features = false`.

  Changes:

  - Updated feature dependencies in `Cargo.toml` files to conditionally enable std-dependent features
  - Added proper feature gating in runtime modules for `no_std` compatibility
  - Modified VM system module to respect the std feature flag on Windows/Unix targets

  The fix ensures that when wasmtime is used without default features, it properly builds in `no_std` on all supported platforms.


crates/wasmtime/Cargo.toml
```diff
@@ -333,6 +331,8 @@ std = [
   'pulley-interpreter/std',
   'wasmtime-math/std',
   'addr2line?/std',
+  "dep:rustix",
+  "rustix/mm",
```
salmans: It's not clear what the dependency on `rustix` does here. The code compiles and passes all tests without it. Should we remove these lines?

crates/wasmtime/src/runtime/vm/sys/mod.rs
```diff
@@ -50,6 +50,9 @@ cfg_if::cfg_if! {
     if #[cfg(miri)] {
         mod miri;
         pub use miri::*;
+    } else if #[cfg(not(feature = "std"))] {
+        mod custom;
+        pub use custom::*;
```
salmans, line 55: We could implement and provide `no_std` variations of `mod unix` and `mod windows` instead of leaving it up to the user (These are the existing implementations: https://github.com/bytecodealliance/wasmtime/tree/main/crates/wasmtime/src/runtime/vm/sys).

alexcrichton:
Thanks for the PR! Could you elaborate on the motivation for this change as well? For example I'm not aware of a target that's `cfg(unix)` or `cfg(windows)` and doesn't have `std`.

crates/wasmtime/src/runtime/fiber.rs
```diff
@@ -354,6 +354,7 @@ impl<T> StoreContextMut<'_, T> {
     /// # Panics
     ///
     /// Panics if this is invoked outside the context of a fiber.
+    #[allow(dead_code)]
```
salmans: I made this change to fix `cargo check -p wasmtime --no-default-features --features async,std` (previously `cargo check -p wasmtime --no-default-features --features async`) but I'm not sure if that's an expected behavior change.

salmans:
@alexcrichton I’m experimenting with running Wasmtime (with Pulley) in the Windows kernel. The goal is to use WebAssembly to enable limited changes to kernel logic without requiring a full software update or system restart. I’m exploring Wasm as a potential alternative to eBPF for Windows.

alexcrichton:
Oh nice! Do you know if there's a `#[cfg]` specific to the target that you're working with that this could be keyed on instead of `feature = "std"`? While the breakage here isn't the end of the world and something we could document, I'd prefer to shape the change a bit differently. For example there's no testing on CI that a no_std build works on Linux/Windows so this is something that will likely easily regress over time.

crates/wasmtime/Cargo.toml
bjorn3: mmap and the like is put behind the `mm` feature. It likely passes without because of the

```toml
[target.'cfg(unix)'.dependencies]
rustix = { workspace = true, optional = true, features = ["mm", "param"] }
```

elsewhere in this file.

salmans:
No, I don't think there's any specific target for (windows/unix) kernel. But just to clarify, are you suggesting that turning off `std` should still pull in standard library to avoid issues like the ones with CI? If so, an alternative solution that I can think of is to guard no_std with a new flag like `override-std` or `force-no-std`.

Cargo.toml
```diff
@@ -354,7 +354,7 @@ hashbrown = { version = "0.15", default-features = false }
 capstone = "0.13.0"
 smallvec = { version = "1.6.1", features = ["union"] }
 tracing = "0.1.26"
-bitflags = "2.0"
+bitflags = { version = "2.0", default-features = false }
```
alexcrichton: Can you clarify why this change was necessary? I believe the `std` feature here is off-by-default and never enabled by Wasmtime?

.github/workflows/main.yml
```diff
@@ -322,34 +322,34 @@ jobs:
               -p wasmtime --no-default-features --features wat
               -p wasmtime --no-default-features --features profiling
               -p wasmtime --no-default-features --features cache
-              -p wasmtime --no-default-features --features async
+              -p wasmtime --no-default-features --features async,std
```
alexcrichton: One change I'd like to see in this PR is that the `std` feature should not be necessary to just build the crate, even on Linux/Windows targets. In essence these CI changes shouldn't be necessary.

I tested locally and sketched out https://github.com/salmans/wasmtime/pull/1 which I believe would enable backing out these CI changes and would ensure that `std` is not a necessary feature even on Linux/Windows targets.

crates/wasmtime/src/runtime/fiber.rs
alexcrichton: Mind applying this diff:

```diff
diff --git a/crates/wasmtime/src/runtime/fiber.rs b/crates/wasmtime/src/runtime/fiber.rs
index 4dcd59e532..8aa0033c4f 100644
--- a/crates/wasmtime/src/runtime/fiber.rs
+++ b/crates/wasmtime/src/runtime/fiber.rs
@@ -343,9 +343,7 @@ impl<T> StoreContextMut<'_, T> {
         self,
         f: impl FnOnce(StoreContextMut<'_, T>) -> Pin<Box<dyn Future<Output = R> + Send + '_>>,
     ) -> Result<R> {
-        BlockingContext::with(self.0, |store, cx| {
-            cx.block_on(f(StoreContextMut(store)).as_mut())
-        })
+        self.with_blocking(|store, cx| cx.block_on(f(store).as_mut()))
     }

     /// Creates a `BlockingContext` suitable for blocking on futures or
@@ -354,7 +352,6 @@ impl<T> StoreContextMut<'_, T> {
     /// # Panics
     ///
     /// Panics if this is invoked outside the context of a fiber.
-    #[allow(dead_code)]
     pub(crate) fn with_blocking<R>(
         self,
         f: impl FnOnce(StoreContextMut<'_, T>, &mut BlockingContext<'_, '_>) -> R,
```

to work around the need for the annotation? A bit of a hack but hopefully not too bad.

crates/wasmtime/src/runtime/vm/sys/mod.rs
alexcrichton: While theoretically possible this would primarily require an implementation of TLS in a no_std context which, as far as I know, isn't available?

crates/wasmtime/Cargo.toml
alexcrichton: Yeah it's fine to drop `rustix/mm` here due to our preexisting dep directive and if `dep:rustix` can be dropped that also seems reasonable. It's perhaps already included elsewhere.

.github/workflows/main.yml
salmans: Thanks you!

Cargo.toml
salmans: You're right, we don't need that change. My bad!

.github/workflows/main.yml
salmans: The change resolves almost all CI issues except for stack switching checks: https://github.com/bytecodealliance/wasmtime/actions/runs/15983572956/job/45083266148?pr=11152

salmans:
@alexcrichton after more experimentation on Windows, I realized `win.rs` in `jit-icache-coherence` pulls `std`. I'm happy to explore a no_std implementation for it in a separate PR but for now, I just pushed a new [commit](https://github.com/bytecodealliance/wasmtime/pull/11152/commits/3587cfa95dfcb53cf52ad4667b9159010c7e71ed) to disable icache coherence in Windows no_std. If that change makes sense, I'm going to address the remaining CI issues before finalizing the PR. Thanks again for your help!

alexcrichton:
For the stack switching issues can you update the `stack-switching` cargo feature to require `std`?

crates/wasmtime/src/runtime/code_memory.rs
```diff
@@ -322,23 +322,31 @@ impl CodeMemory {
...
-                        let text = self.text();
-
-                        use wasmtime_jit_icache_coherence as icache_coherence;
-
-                        // Clear the newly allocated code from cache if the processor requires it
-                        //
-                        // Do this before marking the memory as R+X, technically we should be able to do it after
-                        // but there are some CPU's that have had errata about doing this with read only memory.
-                        icache_coherence::clear_cache(text.as_ptr().cast(), text.len())
+                        #[cfg(any(not(windows), feature = "std"))]
```
alexcrichton: For this it's a pretty critical operation here to manage the icache so we'll want to avoid this `#[cfg]` which makes it difficult to see when it's used and when it isn't. I think the best solution here might be to sidestep the problem entirely and update [this line](https://github.com/bytecodealliance/wasmtime/blob/5ac3f799820812b70b88411d9e9ac1833a98069c/crates/wasmtime/build.rs#L12) to take the `std` feature into account. For example that could be `(unix || windows) && cfg!(feature = "std")` or something like that. A comment there would be good to explain that if `std` is disabled then it's not supported in Wasmtime right now to interact with the OS which means that it's no longer supported.

salmans: I tried that change but it doesn't solve the problem: Wasmtime still pulls `std` through `jit-icache-coherence` in `no_std` Windows.
Another possible solution is to use `libc` in `jit-icache-coherence` for `no_std` Windows at this [line](https://github.com/bytecodealliance/wasmtime/blob/5ac3f799820812b70b88411d9e9ac1833a98069c/crates/jit-icache-coherence/src/lib.rs#L82). For that, we could add a new `std` feature flag to `jit-icache-coherence` and enable it if `std` in `wasmtime` is enabled.

salmans: I just pushed a [new commit](https://github.com/bytecodealliance/wasmtime/pull/11152/commits/ceae3ffa2e00abd0aa1086c55c22f4cce10e7019) that implements the change suggested above but happy to implement a different solution if this is problematic.

alexcrichton: Your commit looks reasonable, thanks for doing that! I'll apologize again though about how this is a bit messy. I'd prefer to avoid having the crate be sort of half-std/half-not, so I think I've come up with an idea to sidestep this entirely which is to include the icache crate only if `std` is enabled in Wasmtime itself. That solves the concerns I have about things getting a bit weird sometimes and the only remaining issue is various CI bits and how well this works out. I've pushed up a commit to this PR which implements what I was thinking of and I'll watch CI to see if my strategy won't work due to something I haven't thought of

salmans: Not at all! I appreciate your help and the time you've put into this. ~Okay, I'll wait for CI and merge the changes into the branch as soon as I see green.~ Thanks again!

crates/wasmtime/src/runtime/vm/sys/mod.rs
salmans: Yes, that's right. I'm resolving this for now but will explore that space for a potential future PR.

alexcrichton (approved):

PR #11153, opened 2025-06-27
x64: EVEX Encoding for new assembler

rahulchaphalkar:
Adds evex encoding for new assembler. 
Adds `vaddpd` evex instr.

Following items are not implemented in this PR / left out intentionally -

- Broadcast attribute (`bcst`). Will be implemented in future pr.
- Features: Uses `avx512vl` feature right now, this will be changed to AND of `avx512vl` and `avx512f`, and at a later point to `avx10.1`, `avx10.2` etc.
- `vaddpd` is not lowered yet

@abrown Ready for review. I'll add a few code comments for some discussion soon, apart from above list of omitted items.

### DIFF
--- a/cranelift/assembler-x64/meta/src/dsl/encoding.rs
+++ b/cranelift/assembler-x64/meta/src/dsl/encoding.rs
@@ -23,7 +23,7 @@ use core::fmt;
 pub fn rex(opcode: impl Into<Opcodes>) -> Rex {
     Rex {
         opcodes: opcode.into(),
-        w: false,
+        w: WBit::W0,
         modrm: None,
         imm: Imm::None,
         opcode_mod: None,
@@ -169,7 +169,7 @@ pub struct Rex {
     /// prefix and other optional/mandatory instruction prefixes are discussed
     /// in chapter 2. Note that REX prefixes that promote legacy instructions to
     /// 64-bit behavior are not listed explicitly in the opcode column."
-    pub w: bool,
+    pub w: WBit,
     /// Indicates modifications to the ModR/M byte.
     pub modrm: Option<ModRmKind>,
     /// The number of bits used as an immediate operand to the instruction.
@@ -193,7 +193,10 @@ impl Rex {
     /// Set the `REX.W` bit.
     #[must_use]
     pub fn w(self) -> Self {
-        Self { w: true, ..self }
+        Self {
+            w: WBit::W1,
+            ..self
+        }
     }
 
     /// Set the ModR/M byte to contain a register operand and an r/m operand;
@@ -358,6 +361,8 @@ impl Rex {
                 "the opcode modifier width must match the operand widths"
             );
         }
+
+        assert!(!matches!(self.w, WBit::WIG));
     }
 }
 
@@ -381,7 +386,7 @@ impl fmt::Display for Rex {
         if let Some(group4) = &self.opcodes.prefixes.group4 {
             write!(f, "{group4} + ")?;
         }
-        if self.w {
+        if self.w.as_bool() {
             write!(f, "REX.W + ")?;
         }
         if self.opcodes.escape {
--- a/cranelift/assembler-x64/meta/src/generate/format.rs
+++ b/cranelift/assembler-x64/meta/src/generate/format.rs
@@ -134,7 +134,7 @@ impl dsl::Format {
             |l: &dsl::Location| l.bits() == 8 && matches!(l.kind(), Reg(_) | RegMem(_));
         let uses_8bit = self.locations().any(find_8bit_registers);
         fmtln!(f, "let uses_8bit = {uses_8bit};");
-        fmtln!(f, "let w_bit = {};", rex.w);
+        fmtln!(f, "let w_bit = {};", rex.w.as_bool());
         let bits = "w_bit, uses_8bit";
 
         let style = match self.operands_by_kind().as_slice() {


cranelift/assembler-x64/meta/src/instructions/add.rs
```diff
@@ -79,6 +79,7 @@ pub fn list() -> Vec<Inst> {
         inst("paddusw", fmt("A", [rw(xmm1), r(align(xmm_m128))]), rex([0x66, 0x0F, 0xDD]).r(), _64b | compat | sse2),
         inst("phaddw", fmt("A", [rw(xmm1), r(align(xmm_m128))]), rex([0x66, 0x0F, 0x38, 0x01]).r(), _64b | compat | ssse3),
         inst("phaddd", fmt("A", [rw(xmm1), r(align(xmm_m128))]), rex([0x66, 0x0F, 0x38, 0x02]).r(), _64b | compat | ssse3),
+        inst("vaddpd", fmt("C", [w(xmm1).k(1).z(), r(xmm2), r(xmm_m128)]), evex(EvexLength::L128).r().w1()._66()._0f().op(0x58), _64b | compat | avx512vl),
```
rahulchaphalkar: `EvexLength` and `VexLength` enums have members with same names, causing me to specify which enum. I can either 
- Make all vex/evex look the same form, like `VexLength::L128` and `EvexLength::L128` everywhere
- Introduce constants like `V128`, `E128` etc.
- Use `VexLength` for evex as well or add a common `VectorLength` enum for both vex/evex.
Something else?

cranelift/assembler-x64/meta/src/dsl/format.rs
```diff
@@ -178,6 +178,21 @@ impl Format {
...
+
+    /// Return the operand that uses a mask register
+    pub fn mask_register_operand(&self) -> Option<&Operand> {
+        self.operands.iter().find(|op| op.mask_reg.is_some())
+    }
+
+    /// Returns true if zeroing is used
+    pub fn zeroing(&self) -> bool {
+        self.operands.iter().any(|op| op.zeroing)
+    }
```
abrown: This doesn't make sense because the opmask registers are true operands. The user can specify which register (`k0`-`k7`) to use when the assembler is used at runtime, so we can't predetermine which `mask_reg` the instruction should be using here when instructions are defined.

cranelift/assembler-x64/meta/src/dsl/format.rs
```diff
@@ -229,6 +244,23 @@ pub struct Operand {
     /// Some register operands are implicit: that is, they do not appear in the
     /// disassembled output even though they are used in the instruction.
     pub implicit: bool,
+    /// EVEX opmask register (k1-k7)
+    pub mask_reg: Option<u8>,
+    /// EVEX zeroing-masking flag (z)
+    pub zeroing: bool,
```
abrown: As a further point on my comment above about `Operand::mask_reg`, I think we would also want `Operand::zeroing` to be a runtime decision. In other words, when a user of this assembler is encoding an instruction, they should have the ability to decide not only which opmask register to use (`k0`-`k7`) but also to select if they want the "merging-masking" or "zeroing-masking" behavior (section 2.7.4 in the manual). So this `bool` field would need to be present on the generated instruction `struct`, not here in the DSL.

This raises some questions, though, about how to handle these "options" on an instruction:
- for opmask registers, we probably want some way to use `k0` by default, which does no masking at all; other, than this, opmask registers act pretty much like other operands (no regalloc, though)
- for masking behavior, we do not want them to act like other operands &mdash; this flag isn't an operand at all! I propose we encode the default behavior in `EVEX.z` and open an issue to figure out a good way to expose this kind of flag on instructions. Why? Because...
- for vector lengths, we actually want the ability to select the length as one of these flags (`L128`, `L256`, `L512`). Right now we fix this in the DSL but, if we had a way to expose flags, we would want instruction lengths to parameterize the instruction.

But all of that does not need to be figured out in this PR. I propose we just default to using `k0` and ignore the masking behavior here and open an issue to solve this later.

cranelift/assembler-x64/meta/src/generate/format.rs
```diff
@@ -301,6 +334,104 @@ impl dsl::Format {
...
+                fmtln!(f, "let reg = self.{reg}.enc();");
+                fmtln!(f, "let vvvv = self.{vvvv}.enc();");
+                fmtln!(f, "let rm = self.{rm}.encode_bx_regs();");
+                fmtln!(f, "let evex = EvexPrefix::new(reg, vvvv, rm, {bits});");
+                ModRmStyle::RegMem {
+                    reg: ModRmReg::Reg(*reg),
+                    rm: *rm,
+                }
+            }
+            [Reg(reg), Reg(vvvv), RegMem(rm), Reg(is4)] => {
```
abrown: I'm not sure that we want to copy-paste these yet from the VEX implementation. I suspect they all are the same but I'm not really sure, so we might want to start with only the formats we need and build from there. At some later point we could even deduplicate between VEX and EVEX.

cranelift/assembler-x64/meta/src/generate/inst.rs
```diff
@@ -283,6 +288,26 @@ impl dsl::Inst {
...
+            fmtln!(f, "let trap = self.trap;");
+            if implicit_ops.is_empty() {
+                implicit_ops.push_str(" ;; {trap}");
+            } else {
+                implicit_ops.push_str(", {trap}");
+            }
+        }
+        fmtln!(f, "write!(f, \"{{name}} {ordered_ops}{implicit_ops}\")");
+    }
+
```
abrown: If we ignore the opmask registers and masking behavior then we don't need this, right? Or I guess we still need to print `{k0}` or something like that to match Capstone? Maybe we should implement the "opmask registers as operands" logic after all...

cranelift/assembler-x64/meta/src/instructions/add.rs
```diff
@@ -67,8 +67,8 @@ pub fn list() -> Vec<Inst> {
         inst("addsd", fmt("A", [rw(xmm1), r(xmm_m64)]), rex([0xF2, 0x0F, 0x58]).r(), _64b | compat | sse2),
         inst("addps", fmt("A", [rw(xmm1), r(align(xmm_m128))]), rex([0x0F, 0x58]).r(), _64b | compat | sse).alt(avx, "vaddps_b"),
         inst("vaddps", fmt("B", [w(xmm1), r(xmm2), r(xmm_m128)]), vex(L128)._0f().op(0x58), _64b | compat | avx),
-        inst("addpd", fmt("A", [rw(xmm1), r(align(xmm_m128))]), rex([0x66, 0x0F, 0x58]).r(), _64b | compat | sse2).alt(avx, "vaddpd_b"),
-        inst("vaddpd", fmt("B", [w(xmm1), r(xmm2), r(xmm_m128)]), vex(L128)._66()._0f().op(0x58), _64b | compat | avx),
+        inst("addpd", fmt("A", [rw(xmm1), r(align(xmm_m128))]), rex([0x66, 0x0F, 0x58]), _64b | compat | sse2).alt(avx, "vaddpd_b"),
+        inst("vaddpd", fmt("B", [w(xmm1), r(xmm2), r(xmm_m128)]), vex(L128).r()._66()._0f().op(0x58), _64b | compat | avx),
```
abrown:
```suggestion
        inst("addpd", fmt("A", [rw(xmm1), r(align(xmm_m128))]), rex([0x66, 0x0F, 0x58]).r(), _64b | compat | sse2).alt(avx, "vaddpd_b"),
        inst("vaddpd", fmt("B", [w(xmm1), r(xmm2), r(xmm_m128)]), vex(L128)._66()._0f().op(0x58).r(), _64b | compat | avx),
```

The order doesn't matter to the DSL but we should keep it consistent with the manual.

abrown: Guess the `r()` is missing from a bunch of VEX instructions which goes to show how little it changes!

cranelift/assembler-x64/meta/src/instructions/add.rs
abrown: I think we should just call it `Length` and have it contain all three: `L128`, `L256`, and `L512`. Then, in the validation for VEX instructions we should only allow `L128` and `L256`.

cranelift/assembler-x64/src/evex.rs
```diff
@@ -0,0 +1,88 @@
...
+
+impl EvexPrefix {
+    /// Construct the [`EvexPrefix`] for an instruction.
+    pub fn new(
+        reg: u8,
+        vvvv: u8,
+        (b, x): (Option<u8>, Option<u8>),
+        ll: (u8, u8),
+        pp: u8,
+        mmmmm: u8,
```
abrown:
```suggestion
        mmm: u8,
```

The EVEX field is 3-bit: `EVEX.mmm`. Makes me think we should check that we're using this `u8` correctly elsewhere.

cranelift/assembler-x64/src/evex.rs
```diff
@@ -0,0 +1,88 @@
...
+    ) -> Self {
+        let r = invert_top_bit(reg);
+        let r_prime = invert_top_bit(reg >> 1);
+        let b = invert_top_bit(b.unwrap_or(0));
+        let x = invert_top_bit(x.unwrap_or(0));
+        let vvvv_value = !vvvv & 0b1111;
+        let v_prime = !(vvvv >> 4) & 0b1;
+
+        // byte1
+        debug_assert!(mmmmm >= 0b01 && mmmmm <= 0b11);
```
abrown: E.g., this is true for now but there are two other decoding map IDs, 5 and 6, that would trip over this.

abrown:
I'll probably take over this PR soon since @rahulchaphalkar is out for a bit.

cranelift/assembler-x64/meta/src/dsl/encoding.rs
```diff
@@ -1176,3 +1214,164 @@ impl fmt::Display for Vex {
...
+
+pub struct Evex {
+    /// The vector length of the operand (e.g., 128-bit, 256-bit, or 512-bit).
+    pub length: Length,
+    /// Map the `PP` field encodings.
+    pub pp: Option<VexPrefix>,
+    /// Map the `MMMMM` field encodings.
+    pub mmmmm: Option<VexEscape>,
+    /// The `W` bit.
+    pub w: VexW,
```
abrown: I'll rename this to `WBit` or something like that...

cranelift/assembler-x64/meta/src/instructions/add.rs
```diff
@@ -82,19 +82,20 @@ pub fn list() -> Vec<Inst> {
...
+        inst("vpaddd", fmt("B", [w(xmm1), r(xmm2), r(xmm_m128)]), vex(L128)._66()._0f().op(0xFE).r(), _64b | compat | avx),
+        inst("vpaddq", fmt("B", [w(xmm1), r(xmm2), r(xmm_m128)]), vex(L128)._66()._0f().op(0xD4).r(), _64b | compat | avx),
+        inst("vpaddsb", fmt("B", [w(xmm1), r(xmm2), r(xmm_m128)]), vex(L128)._66()._0f().op(0xEC).r(), _64b | compat | avx),
+        inst("vpaddsw", fmt("B", [w(xmm1), r(xmm2), r(xmm_m128)]), vex(L128)._66()._0f().op(0xED).r(), _64b | compat | avx),
+        inst("vpaddusb", fmt("B", [w(xmm1), r(xmm2), r(xmm_m128)]), vex(L128)._66()._0f().op(0xDC).r(), _64b | compat | avx),
+        inst("vpaddusw", fmt("B", [w(xmm1), r(xmm2), r(xmm_m128)]), vex(L128)._66()._0f().op(0xDD).r(), _64b | compat | avx),
+        inst("vphaddw", fmt("B", [w(xmm1), r(xmm2), r(xmm_m128)]), vex(L128)._66()._0f38().op(0x01).r(), _64b | compat | avx),
+        inst("vphaddd", fmt("B", [w(xmm1), r(xmm2), r(xmm_m128)]), vex(L128)._66()._0f38().op(0x02).r(), _64b | compat | avx),
+        inst("vaddpd", fmt("C", [w(xmm1).k(1).z(), r(xmm2), r(xmm_m128)]), evex(L128)._66()._0f().w1().op(0x58).r(), _64b | compat | avx512vl),
```
abrown: This logic should be `(AVX512VL AND
AVX512F)`. Additionally, since we need the ability to `AND` things together, we need to teach the whole expression to mean: `(_64b | compat) & (...)` (maybe in another PR, since this affects many instructions beyond just this one).

abrown:
@cfallin or @alexcrichton: can one of you review this? I've changed this PR substantially since it was first opened so I think it needs a new reviewer that is not me.

cfallin:
Sure, taking a look!

cfallin (approved):
LGTM. Taking your word (and the fuzzer's) on the actual prefix-encoding details!

cranelift/assembler-x64/meta/src/instructions/add.rs
alexcrichton: I think this may actually work out? Not to say our current modeling of things is correct, but functionally with `| avx512f` here I think it has the net effect of requiring more features. Notably we [take all the features the assembler says](https://github.com/bytecodealliance/wasmtime/blob/da2655157740c4ec85aa329376088c927781428d/cranelift/codegen/src/isa/x64/inst/mod.rs#L119-L142) and flag them all as requirements of each instruction, so by listing more features here it's interpreted as effectively requiring all of them. I don't know if that's an accurate reflection of `_64b` and `compat` though as we otherwise basically ignore those features right now.

Although rereading your comment now I suspect you're talking about the interactions of `_64b` and `compat` with feature flags, which is exactly the gray area for me that iunno much about, so probably just disregard me

PR #11156, opened 2025-06-29
Add new example that uses WIT resource

### LINKED ISSUE #11091: CI Bug: wasip2 examples are not checked
I think this breaking change from #10770 should've broken some examples in CI like https://github.com/bytecodealliance/wasmtime/tree/main/examples/wasip2-async and https://github.com/bytecodealliance/wasmtime/tree/main/examples/wasip2 but these were not checked in CI. Probably we should fix this problem by adding these examples in here?
https://github.com/bytecodealliance/wasmtime/blob/a7d9f3e114fc74889000aeb2c136e9dfd5c374be/Cargo.toml#L147

cc @alexcrichton

ifsheldon:
This adds a new example, demonstrating how to use WIT resources. This also provides an example of how to use the new APIs introduced in #10770, see the line
```
KvDatabase::add_to_linker::<_, HasSelf<_>>(&mut linker, |s| s)?;
```

Closes #11091

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1288,6 +1288,13 @@ name = "example-fib-debug-wasm"
 version = "0.0.0"
 
 [[package]]
+name = "example-resource-component-wasm"
+version = "0.1.0"
+dependencies = [
+ "wit-bindgen",
+]
+
+[[package]]
 name = "example-tokio-wasm"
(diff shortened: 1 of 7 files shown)


alexcrichton (approved):
Thanks!

PR #11166, opened 2025-06-30
 Simplify some conditional testing logic for Winch 

alexcrichton:
Run all tests on all platforms in `#[wasmtime_test]` for Winch without
`#[cfg]` and use the result of `should_fail` to determine whether it
should fail or not (e.g. still expect failure on riscv64/s390x/etc).


alexcrichton:
Oops, sorry there was a wrong commit history which meant the PR description/OP didn't get filled out correctly and I missed that

PR #11168, opened 2025-07-01
Reserve space for outgoing edges we will trace in the DRC collector

fitzgen:
Rather than resizing possibly multiple times as we push.


alexcrichton (approved):

fitzgen:
CI got stuck, retrying.

Issue #11162, opened 2025-06-30
Remove hash sets and bump chunk from DRC collector


fitzgen:
Hashing is expensive and the fact that the bump chunk can fill and needs draining conflates "need GC" with "the bump chunk is full" and "the GC heap is at capacity".

Instead, we should:

* Add what is logically a `Option<VMGcRef>` to the DRC's object header. This forms an intrusive singly-linked list that implements the over-approximation set, which is implemented with a hash set today, but this approach doesn't need any hashing or external allocations.
* Replace the precise hash set with a mark bit in the object header. Again, no more hashing or external-to-the-gc-heap allocation or storage.

With that, DRC sweeping is still logically the same: walk over the over-approximation set, decref any object that is not in the precise set, and reset the over-approximation set to be the precise set. However, this will now be implemented as a mutable traversal of the over-approximation linked list.

This should result in a large speed up for test cases like #11141

PR #11169, opened 2025-07-01
Move the object size out of the reserved bits in the DRC collector

fitzgen:
We are going to use those bits for other things as we implement https://github.com/bytecodealliance/wasmtime/issues/11162.

This also lets us increase the largest supported GC allocation to a full `u32::MAX` bytes.


tests/misc_testsuite/gc/issue-10464.wast
```diff
@@ -1,14 +0,0 @@
...
-  (type $b (array structref))
-  (func (export "f")
-    struct.new_default $a
-    i32.const 536870911
-    array.new $b
-    drop
-  )
-)
-
-(assert_trap (invoke "f") "allocation size too large")
```
alexcrichton, line 14: Is it worth adapting this test to the new limit of 32-bits?

alexcrichton (approved):

tests/misc_testsuite/gc/issue-10464.wast
fitzgen: `tests/misc_testsuite/gc/array-alloc-too-large.wast` already tests the 32-bit limit, so I think we are good.

PR #11175, opened 2025-07-02
Remove hash tables and bump chunk from the DRC collector

fitzgen:
This removes the explicit `HashSet`s used to represent the over-approximated-stack-roots and precise-stack-roots sets and replaces them with an intrusive, singly-linked list and a mark bit in the object headers respectively. The new list implementation also subsumes the old bump chunk that sat in front of the old over-approximated-stack-roots hash set.

This shaves off about 25% of the time it takes to run the test case in https://github.com/bytecodealliance/wasmtime/issues/11141 for me locally.

This also ended up being a nice simplification of the DRC collector, which in turn allowed us to further simplify the `GcHeap` trait, since we no longer ever need to GC before passing GC refs into Wasm.

Fixes https://github.com/bytecodealliance/wasmtime/issues/11162


crates/wasmtime/src/runtime/vm/gc/gc_ref.rs
```diff
@@ -88,8 +88,10 @@ impl VMGcHeader {
     pub fn set_reserved_u27(&mut self, value: u32) {
         assert!(
             VMGcKind::value_fits_in_unused_bits(value),
-            "VMGcHeader::set_reserved_u26 with value using more than 26 bits"
+            "VMGcHeader::set_reserved_u27 with value using more than 27 bits: \
+             {value:#034b} ({value}, {value:#010x})"
         );
+        self.kind &= VMGcKind::MASK;
```
fitzgen, line 94: This was such a silly bug but it made me feel like I was losing my mind :facepalm:

crates/cranelift/src/func_environ/gc/enabled/drc.rs
```diff
@@ -87,35 +87,117 @@ impl DrcCompiler {
...
+        let head = self.load_over_approximated_stack_roots_head(func_env, builder);
+
+        // Load the old first list element, which will be our new next list
+        // element.
+        let next = builder
+            .ins()
+            .load(ir::types::I32, ir::MemFlags::trusted(), head, 0);
+
+        // Store this object as the new head of the list.
+        builder
```
cfallin: Do we want to defer this list-head update until after the code below sets up the object at the head? Otherwise there's a risk the intermediate state is visible -- perhaps fine (we're not a concurrent mutator) but if e.g. a trap occurs or we add some kind of observability mechanism it might present an edge case, especially as we delegate the setup to helpers. Basically it seems cleaner to me to keep the "making it live is the last store" pattern here.

crates/environ/src/gc/drc.rs
```diff
@@ -11,6 +11,13 @@ pub const HEADER_ALIGN: u32 = 8;
 /// The offset of the length field in a `VMDrcArrayHeader`.
 pub const ARRAY_LENGTH_OFFSET: u32 = HEADER_SIZE;
 
+/// The bit within a `VMDrcHeader`'s reserved bits that is the mark bit.
+pub const HEADER_MARK_BIT: u32 = 1 << 0;
```
alexcrichton: Could this expand on what's being marked?

crates/wasmtime/src/runtime/vm/gc/enabled/drc.rs
```diff
@@ -22,21 +22,24 @@
...
+//! precise set. Finally, the over-approximation is reset to the precise set.
 //!
-//! The `VMGcRefActivationsTable` implements the over-approximized set of
-//! `VMGcRef`s referenced by Wasm activations. Calling a Wasm function and
-//! passing it a `VMGcRef` moves the `VMGcRef` into the table, and the compiled
-//! Wasm function logically "borrows" the `VMGcRef` from the table. Similarly,
-//! `global.get` and `table.get` operations clone the gotten `VMGcRef` into the
-//! `VMGcRefActivationsTable` and then "borrow" the reference out of the table.
+//! An intrusive, singly-linked list in the object header implements the
+//! over-approximized set of `VMGcRef`s referenced by Wasm activations. Calling
```
cfallin: s/approximized/approximated/

crates/wasmtime/src/runtime/vm/gc/enabled/drc.rs
```diff
@@ -22,21 +22,24 @@
...
+//! a Wasm function and passing it a `VMGcRef` inserts the `VMGcRef` into that
+//! list if it is not already present, and the compiled Wasm function logically
+//! "borrows" the `VMGcRef` from the table. Similarly, `global.get` and
```
cfallin: s/table/list/ ?

crates/wasmtime/src/runtime/vm/gc/enabled/drc.rs
```diff
@@ -22,21 +22,24 @@
...
+//! `table.get` operations logivally clone the gotten `VMGcRef` into that
```
cfallin: s/logivally/logically/

crates/wasmtime/src/runtime/vm/gc/enabled/drc.rs
```diff
@@ -22,21 +22,24 @@
...
+//! list and then "borrow" the reference out of the table.
```
cfallin: s/table/list/

crates/wasmtime/src/runtime/vm/gc/enabled/drc.rs
```diff
@@ -436,109 +429,103 @@ impl DrcHeap {
...
+
+            // This GC ref was not marked, meaning it is no longer on the stack,
+            // so remove it from the over-approximated-stack-roots list and
+            // decrement its reference count.
+            log::trace!(
+                "  -> {gc_ref:#p} is not marked, removing it from over-approximated-\
+                 stack-roots list and decrementing its ref count"
+            );
+            next = header.next_over_approximated_stack_root();
+            let prev_next = header.next_over_approximated_stack_root();
```
cfallin, line 509: I'm not sure I see why we load this value twice -- can we use `next` below when we patch the prev link?

crates/wasmtime/src/runtime/vm/gc/enabled/drc.rs
```diff
@@ -578,9 +566,92 @@ unsafe impl GcHeapObject for VMDrcHeader {
...
+    #[inline]
+    fn remove_from_over_approximated_stack_roots(&mut self) {
+        debug_assert!(self.is_in_over_approximated_stack_roots());
+        let reserved = self.header.reserved_u27();
+        self.header
+            .set_reserved_u27(reserved & !wasmtime_environ::drc::HEADER_IN_OVER_APPROX_LIST_BIT);
+        // NB: we don't need to set `self.next_over_approximated_stack_root` to
+        // `None` -- that field only has meaning when the bit is set. This lets
+        // us avoid some unnecessary stores.
+    }
```
alexcrichton: Could this method be removed entirely in favor taking a `bit: bool` parameter in `set_in_over_approximated_stack_roots` above? That way the callers would inherit the comment of "no need to update the over approximated set" which I think would make more sense in the context of callers than here.

crates/wasmtime/src/runtime/vm/gc/gc_ref.rs
```diff
@@ -101,6 +103,7 @@ impl VMGcHeader {
     /// be unset.
     pub unsafe fn unchecked_set_reserved_u26(&mut self, value: u32) {
```
cfallin: The function name here says `u26` but the comment, and the function above as well, say `u27` / `27-bit` -- they should match I think?

alexcrichton (approved):
Really like how this worked out, especially how it simplifies passing GC references into wasm

crates/cranelift/src/func_environ/gc/enabled/drc.rs
fitzgen: It shouldn't matter either way -- as you mention, there isn't any concurrency here -- but happy to swap things around.

crates/wasmtime/src/runtime/vm/gc/enabled/drc.rs
fitzgen: `VMGcRef` doesn't implement `Clone` since cloning a `VMGcRef` potentially needs to run barriers, so we have dedicated methods to clone with those barriers if necessary. However, inside the GC implementation that isn't really a concern so we have `unchecked_copy` for these cases that don't need barriers. That said, when the GC ref is wrapped in an `Option` like it is here, it can be a bit of a mouthful, and we would have to do `next.as_ref().map(|r| r.unchecked_copy())`. So I figured I'd just grab it from the header twice and let LLVM clean things up.

PR #11176, opened 2025-07-03
Cranelift: Generate an `InstructionData::map` method

fitzgen:
This allows you to map some functions, described by the given `InstructionMapper`, over each of the entitities in an instruction, producing a new `InstructionData`.

I intend to use this as part of an inliner API for Cranelift that I am developing as part of prototyping [Wasmtime's compile-time builtins](https://github.com/bytecodealliance/rfcs/pull/43).

### DIFF
--- a/cranelift/codegen/meta/Cargo.toml
+++ b/cranelift/codegen/meta/Cargo.toml
@@ -20,6 +20,7 @@ cranelift-srcgen = { workspace = true }
 cranelift-assembler-x64-meta = { path = "../../assembler-x64/meta", version = "0.122.0" }
 cranelift-codegen-shared = { path = "../shared", version = "0.122.0" }
 pulley-interpreter = { workspace = true, optional = true }
+heck = "0.5.0"
 
 [features]
 pulley = ['dep:pulley-interpreter']
--- a/cranelift/codegen/meta/src/gen_inst.rs
+++ b/cranelift/codegen/meta/src/gen_inst.rs
@@ -3,11 +3,11 @@
 use crate::cdsl::camel_case;
 use crate::cdsl::formats::InstructionFormat;
 use crate::cdsl::instructions::{AllInstructions, Instruction};
-use crate::cdsl::operands::Operand;
+use crate::cdsl::operands::{Operand, OperandKindFields};
 use crate::cdsl::typevar::{TypeSet, TypeVar};
 use crate::unique_table::{UniqueSeqTable, UniqueTable};
 use cranelift_codegen_shared::constant_hash;
-use cranelift_srcgen::{Formatter, Language, Match, error, fmtln};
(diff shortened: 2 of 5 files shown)


abrown:
I'll take an in-depth look at this after the weekend?

cranelift/codegen/meta/src/gen_inst.rs
```diff
@@ -159,7 +159,7 @@ fn gen_arguments_method(formats: &[Rc<InstructionFormat>], fmt: &mut Formatter,
 /// - `pub fn eq(&self, &other: Self, &pool) -> bool`
 /// - `pub fn hash<H: Hasher>(&self, state: &mut H, &pool)`
 fn gen_instruction_data_impl(formats: &[Rc<InstructionFormat>], fmt: &mut Formatter) {
-    fmt.add_block("impl InstructionData",|fmt| {
+    fmt.add_block("impl InstructionData", |fmt| {
```
abrown, line 162: It's weird that this missing space didn't trigger a formatting error during CI prior to this; is it possible this crate isn't being checked?

abrown (approved):
Makes sense!

cranelift/codegen/meta/src/gen_inst.rs
fitzgen: I think the whole crate disabled `rustfmt` or something, but I haven't looked into it yet.

cranelift/codegen/meta/Cargo.toml
```diff
@@ -20,6 +20,7 @@ cranelift-srcgen = { workspace = true }
 cranelift-assembler-x64-meta = { path = "../../assembler-x64/meta", version = "0.122.0" }
 cranelift-codegen-shared = { path = "../shared", version = "0.122.0" }
 pulley-interpreter = { workspace = true, optional = true }
+heck = "0.5.0"
```
bjorn3, line 23: Was it necessary to pull in a new dependency for a single function?

fitzgen: It's already in the workspace and vetted, so it didn't seem like a big deal.

bjorn3: It was previously not a Cranelift dependency. Cranelift is slowly getting more and more dependencies.

fitzgen: Sharing common code via dependencies is not bad in and of itself IMO, and is actually generally good, especially when it is already vetted and relied upon in sibling projects.

Of course, there is the potential downside of from-scratch compile times getting a little worse. I'm happy to review a PR replacing the dependency with the hand-rolled equivalent function if you measure a from-scratch compile time regression and want to send such a PR, but I don't personally have the cycles to investigate all that myself at the moment.

Issue #11134, opened 2025-06-25
Cranelift: Avoid string conversion for TargetIsa on every function with the incremental cache


bjorn3:
#### Feature

Avoid the string conversions in https://github.com/bytecodealliance/wasmtime/blob/3632683b38011cd12839c57cbbc4fa03e834b111/cranelift/codegen/src/incremental_cache.rs#L150-L163 that are done for every function. Either by hashing a binary representation of the TargetIsa instead or by caching the hash in TargetIsa.

#### Benefit

This reduces the overhead of computing the caching key, making the incremental cache a bit faster.

PR #11186, opened 2025-07-07
Couple of optimizations to the Cranelift incremental cache

bjorn3:
Fixes https://github.com/bytecodealliance/wasmtime/issues/11134

### DIFF
--- a/cranelift/codegen/src/isa/mod.rs
+++ b/cranelift/codegen/src/isa/mod.rs
@@ -295,7 +295,7 @@ pub trait TargetIsa: fmt::Display + Send + Sync {
     fn isa_flags(&self) -> Vec<settings::Value>;
 
     /// Get the ISA-dependent flag values as raw bytes for hashing.
-    fn isa_flags_hash_key(&self) -> &[u8];
+    fn isa_flags_hash_key(&self) -> IsaFlagsHashKey<'_>;
 
     /// Get a flag indicating whether branch protection is enabled.
     fn is_branch_protection_enabled(&self) -> bool {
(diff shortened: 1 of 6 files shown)


crates/cranelift/src/translate/translation_utils.rs
```diff
@@ -88,8 +89,15 @@ pub fn f64_translation(x: wasmparser::Ieee64) -> ir::immediates::Ieee64 {
     ir::immediates::Ieee64::with_bits(x.bits())
 }
 
-/// Special VMContext value label. It is tracked as 0xffff_fffe label.
+/// Special VMContext value label. It is tracked as 0 label.
 pub fn get_vmctx_value_label() -> ir::ValueLabel {
-    const VMCTX_LABEL: u32 = 0xffff_fffe;
-    ir::ValueLabel::from_u32(VMCTX_LABEL)
+    ir::ValueLabel::from_u32(0)
```
bjorn3: This is necessary to due to the change to using `SecondaryMap` for `ValueLabelsRanges`.

Edit: Removed this commit because I can't figure out how to fix it.

cranelift/codegen/src/isa/mod.rs
```diff
@@ -294,6 +294,9 @@ pub trait TargetIsa: fmt::Display + Send + Sync {
     /// Get the ISA-dependent flag values that were used to make this trait object.
     fn isa_flags(&self) -> Vec<settings::Value>;
 
+    /// Get the ISA-dependent flag values as raw bytes for hashing.
+    fn isa_flags_hash_key(&self) -> &[u8];
```
abrown: I wondered if there was a way to avoid exposing the internal byte representation but `pub(crate)` isn't an option so it's not clear what else would work. If you're happy with this then let's leave it as is.

abrown: But maybe an `IsaFlags([u8])` with `impl Hash` could work...

abrown (approved):
This makes sense. I'll wait to hear if you have any opinions about protecting the internal bit representation before merging.

cranelift/codegen/src/isa/mod.rs
bjorn3: Done

PR #11196, opened 2025-07-07
x64: Add conditional jumps to the new assembler

alexcrichton:
While this doesn't remove any pseudo insts from ISLE (as was thought might be the case with `WinchJmpIf`) this does clean up some emission code to use some helpers and make it more clear what's happening.

### DIFF
--- a/cranelift/assembler-x64/meta/src/instructions/jmp.rs
+++ b/cranelift/assembler-x64/meta/src/instructions/jmp.rs
@@ -1,9 +1,66 @@
 use crate::dsl::{Customization::*, Feature::*, Inst, Location::*};
-use crate::dsl::{fmt, inst, r, rex};
+use crate::dsl::{fmt, inst, r, rex, sxq};
 
 #[rustfmt::skip] // Keeps instructions on a single line.
 pub fn list() -> Vec<Inst> {
     vec![
         inst("jmpq", fmt("M", [r(rm64)]), rex([0xFF]).digit(4), _64b).custom(Display),
+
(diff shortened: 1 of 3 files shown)


cranelift/assembler-x64/meta/src/instructions/jmp.rs
```diff
@@ -1,9 +1,66 @@
...
+        inst("jge", fmt("D32", [r(sxq(imm32))]), rex([0x0F, 0x8D]).id(), _64b | compat).custom(Display),
+        inst("jl", fmt("D8", [r(sxq(imm8))]), rex([0x7C]).ib(), _64b | compat).custom(Display),
+        inst("jl", fmt("D32", [r(sxq(imm32))]), rex([0x0F, 0x8C]).id(), _64b | compat).custom(Display),
+        inst("jle", fmt("D8", [r(sxq(imm8))]), rex([0x7E]).ib(), _64b | compat).custom(Display),
+        inst("jle", fmt("D32", [r(sxq(imm32))]), rex([0x0F, 0x8E]).id(), _64b | compat).custom(Display),
+        // jna == jbe
+        // jnae == jb
+        // jnb == jae
+        // jnbe == ja
+        // jnc == jae
```
fitzgen, line 42: Does it make sense to generate helper functions or something for these aliases? Obviously not necessary in this PR, but something to think about.

alexcrichton: Good point! cc @abrown on this as well about maybe adding "aliases" or such to the assembler. AFAIK it's just anything related to condition codes though which may not be worth the feature too

PR #11198, opened 2025-07-07
Update gimli/object/addr2line crates

alexcrichton:
Keeping up-to-date on their major version tracks


abrown (approved):
Makes sense! Looks like there is a `gimli` duplicate version causing the CI failure.

Issue #10758, opened 2025-05-09
cranelift-object: Support marking data object as used


bjorn3:
#### Feature

See title.

#### Benefit

This is required for implementing `#[used(linker)]` in cg_clif. This attribute is equivalent to `attribute((used))` in C. And on all non-ELF platforms it is the default. when `#[used]` is used. On ELF platforms it may become the default in the future.

#### Implementation

Add a flag to DataDescription and then set the right symbol flag for the given object file format.

bjorn3:
https://github.com/bytecodealliance/wasmtime/pull/11206 implements this on the Cranelift side and https://github.com/rust-lang/rustc_codegen_cranelift/tree/used_static has the cg_clif changes.

PR #11206, opened 2025-07-09
Allow marking data objects as used for the linker

bjorn3:
Fixes https://github.com/bytecodealliance/wasmtime/issues/10758

### DIFF
--- a/cranelift/object/src/backend.rs
+++ b/cranelift/object/src/backend.rs
@@ -15,8 +15,8 @@ use object::write::{
     Object, Relocation, SectionId, StandardSection, Symbol, SymbolId, SymbolSection,
 };
 use object::{
-    RelocationEncoding, RelocationFlags, RelocationKind, SectionKind, SymbolFlags, SymbolKind,
-    SymbolScope,
+    RelocationEncoding, RelocationFlags, RelocationKind, SectionFlags, SectionKind, SymbolFlags,
+    SymbolKind, SymbolScope,
 };
 use std::collections::HashMap;
 use std::collections::hash_map::Entry;
@@ -394,6 +394,7 @@ impl Module for ObjectModule {
             data_relocs: _,
             ref custom_segment_section,
             align,
+            used,
         } = data;
 
         let pointer_reloc = match self.isa.triple().pointer_width().unwrap() {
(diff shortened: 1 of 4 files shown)


cranelift/object/src/backend.rs
```diff
@@ -451,6 +451,34 @@ impl Module for ObjectModule {
...
+                                sh_flags: object::elf::SHF_GNU_RETAIN.into(),
+                            }
+                        }
+                        SectionFlags::Elf { sh_flags } => {
+                            *sh_flags |= u64::from(object::elf::SHF_GNU_RETAIN)
+                        }
+                        _ => unreachable!(),
+                    }
+                }
+                object::BinaryFormat::Coff => {},
```
bjorn3: I couldn't find what flag is necessary for Windows (if any).

bjorn3:
Also could this get a backport to the branch for the next Cranelift release. https://github.com/rust-lang/rust/pull/140525 has removed a workaround for `#[used]` previously defaulting to `#[used(compiler)]` rather than `#[used(linker)]`, so without the Cranelift changes, `.init_array` support in cg_clif is broken with lld.

abrown (approved):
I'm no expert here but this seems reasonable.

cranelift/object/src/backend.rs
```diff
@@ -451,6 +452,34 @@ impl Module for ObjectModule {
...
         };
 
+        if used {
+            match self.object.format() {
+                object::BinaryFormat::Elf => {
+                    let section = self.object.section_mut(section);
+                    match &mut section.flags {
+                        SectionFlags::None => {
+                            section.flags = SectionFlags::Elf {
+                                sh_flags: object::elf::SHF_GNU_RETAIN.into(),
```
bjorn3, line 462: This should have also included `elf::SHF_ALLOC | elf::SHF_WRITE` (or `elf::SHF_ALLOC` only when the data object is neither writable nor contains relocations).

cranelift/object/src/backend.rs
```diff
@@ -451,6 +452,34 @@ impl Module for ObjectModule {
...
+                        }
+                        _ => unreachable!(),
+                    }
+                }
+                object::BinaryFormat::Coff => {}
+                object::BinaryFormat::MachO => {
+                    let symbol = self.object.symbol_mut(symbol);
+                    assert!(matches!(symbol.flags, SymbolFlags::None));
+                    symbol.flags = SymbolFlags::MachO {
+                        n_desc: object::macho::N_NO_DEAD_STRIP,
```
bjorn3, line 476: And I think this is missing `N_WEAK_REF`/`N_WEAK_DEF` for weak symbols.

PR #11207, opened 2025-07-09
Cranelift: Make `ir::{Constant,Immediate}` considered entities

fitzgen:
They reference data in out-of-line pools rather than storing their data inline in the instruction, and when an instruction containing them is moved from one `ir::Function` to another, they need their indices updated accordingly. Therefore, they really are entities rather than immediates.

This recategorization means that they will now be properly mapped in `ir::InstructionData::map` calls.


fitzgen:
(Why we have both `ir::Constant` and `ir::Immediate` is not 100% clear to me, seems like we could probably clean this up so that we just have `ir::Constant`, but that is future work.)

cfallin (approved):

Issue #4127, opened 2022-05-10
Cranelift: develop an inlining pass


cfallin:
In the [2022 roadmap](https://github.com/bytecodealliance/rfcs/blob/main/accepted/cranelift-roadmap-2022.md), we described the need to add [inlining](https://en.wikipedia.org/wiki/Inline_expansion) to Cranelift. This need comes mainly from anticipated future workloads in some Cranelift applications. For example, when used as a Wasm backend, multi-module use-cases will become more common as the [component model](https://github.com/WebAssembly/component-model) becomes a reality. In such use-cases, no inlining would have been done by the Wasm toolchain; execution in Wasmtime/Cranelift is the first time that modules "meet" and can cross-inline calls, or at least interface-type adapter shim functions.

We will likely want to be a little careful about the inlining heuristic and associated costs, in order to preserve our general "fast compilation" focus. Perhaps we want to only inline explicitly-marked-as-inlinable functions. Or perhaps a bottom-up-over-callgraph-SCCs approach (like LLVM) is still the right one, but with a low cutoff threshold for function size to inline.

cfallin:
> Who would do mark functions as inlinable? The Wasm won't have any such annotations and it isn't really something that the Wasm frontend can determine on its own in general (mayyyyyyyyybe by looking at Wasm branch hinting annotation and only inlining for hot call sites in hot blocks? Seems easy to get poor heuristics this way and I'd rather intoduce an inlining hint annotation Wasm proposal than seriously pursue this).

For this use-case I'm mostly anticipating Interface Types-like applications: generated adapter functions/thunks/stubs can be explicitly marked as "please inline". This seems like some nice low-hanging fruit in that (i) the Cranelift user (here Wasmtime) might know that some functions are small, by construction, and (ii) there are relatively few of them and they are only one level deep, so this is not blowing up compile-time by inlining everywhere.

bjorn3:
> However inlining a tiny function from module A into another function in module A doesn't make sense.

For wasm you are mostly right, although someone may have used `#[inline(never)]` for debuginfo reasons or to hide UB from the compiler, in which case inlining at clif ir level should be fine I think. For cg_clif inlining within the same module is absolutely necessary.

FeldrinH:
Has anything come of this? Does cranelift have an inlining pass now?

bjorn3:
No, Cranelift doesn't have an inliner yet.

fitzgen:
[...]

PR #11210, opened 2025-07-09
Cranelift: introduce a function inliner

fitzgen:
This comit adds "inlining as a library" to Cranelift; it does _not_ provide a complete, off-the-shelf inlining solution. Cranelift's compilation context is per-function and does not encompass the full call graph. It does not know which functions are hot and which are cold, which have been marked the equivalent of `#[inline(always)]` versus `#[inline(never)]`, etc... Only the Cranelift user can understand these aspects of the full compilation pipeline, and these things can be very different between (say) Wasmtime and `cg_clif`. Therefore, this infrastructure does not attempt to define hueristics for when inlining a particular call is likely beneficial. This module only provides hooks for the Cranelift user to tell Cranelift whether a given call should be inlined or not, and the mechanics to inline a callee into a particular call site when the user directs Cranelift to do so.

This commit also creates a new kind of filetest that will always inline calls to functions that have already been defined in the file. This lets us exercise the inliner in filetests.

Fixes https://github.com/bytecodealliance/wasmtime/issues/4127

### DIFF
--- a/cranelift/codegen/src/inline.rs
+++ b/cranelift/codegen/src/inline.rs
@@ -1,0 +1,1501 @@
+//! Function inlining infrastructure.
+//!
+//! This module provides "inlining as a library" to Cranelift users; it does
+//! _not_ provide a complete, off-the-shelf inlining solution. Cranelift's
+//! compilation context is per-function and does not encompass the full call
+//! graph. It does not know which functions are hot and which are cold, which
+//! have been marked the equivalent of `#[inline(never)]`, etc... Only the
+//! Cranelift user can understand these aspects of the full compilation
+//! pipeline, and these things can be very different between (say) Wasmtime and
+//! `cg_clif`. Therefore, this module does not attempt to define hueristics for
+//! when inlining a particular call is likely beneficial. This module only
+//! provides hooks for the Cranelift user to define whether a given call should
+//! be inlined or not, and the mechanics to inline a callee into a particular
+//! call site when directed to do so by the Cranelift user.
+//!
+//! The top-level inlining entry point during Cranelift compilation is
+//! [`Context::inline`][crate::Context::inline]. It takes an [`Inline`] trait
+//! implementation, which is authored by the Cranelift user and directs
+//! Cranelift whether to inline a particular call, and, when inlining, gives
+//! Cranelift the body of the callee that is to be inlined.
+
+use crate::cursor::{Cursor as _, FuncCursor};
+use crate::ir::{self, InstBuilder as _};
+use crate::result::CodegenResult;
+use crate::trace;
+use crate::traversals::Dfs;
+use alloc::vec::Vec;
+use cranelift_entity::{SecondaryMap, packed_option::PackedOption};
+use smallvec::SmallVec;
+
+type SmallValueVec = SmallVec<[ir::Value; 8]>;
+type SmallBlockArgVec = SmallVec<[ir::BlockArg; 8]>;
+type SmallBlockCallVec = SmallVec<[ir::BlockCall; 8]>;
+
+/// A command directing Cranelift whether or not to inline a particular call.
+pub enum InlineCommand<'a> {
+    /// Keep the call as-is, out-of-line, and do not inline the callee.
+    KeepCall,
+    /// Inline the call, using this function as the body of the callee.
+    ///
+    /// It is the `Inline` implementor's responsibility to ensure that this
+    /// function is the correct callee. Providing the wrong function may result
+    /// in panics during compilation or incorrect runtime behavior.
+    Inline(&'a ir::Function),
+}
+
+/// A trait for directing Cranelift whether to inline a particular call or not.
+///
+/// Used in combination with the [`Context::inline`][crate::Context::inline]
+/// method.
+pub trait Inline {
+    /// A hook invoked for each direct call instruction in a function, whose
+    /// result determines whether Cranelift should inline a given call.
+    ///
+    /// The Cranelift user is responsible for defining their own hueristics and
+    /// deciding whether inlining the call is beneficial.
+    ///
+    /// It is the `Inline` implementor's responsibility to ensure that, when
+    /// directing Cranelift to inline a call, the provided callee function is
+    /// the correct function. Providing the wrong function may result in panics
+    /// during compilation (e.g. due to signature type mismatches) or incorrect
+    /// runtime behavior.
+    fn inline(
+        &self,
+        caller: &ir::Function,
+        call_inst: ir::Inst,
+        call_opcode: ir::Opcode,
+        callee: ir::FuncRef,
+        call_args: &[ir::Value],
+    ) -> InlineCommand<'_>;
+}
+
+impl<'a, T> Inline for &'a T
+where
+    T: Inline,
+{
+    fn inline(
+        &self,
+        caller: &ir::Function,
+        inst: ir::Inst,
+        opcode: ir::Opcode,
+        callee: ir::FuncRef,
+        args: &[ir::Value],
+    ) -> InlineCommand<'_> {
+        (*self).inline(caller, inst, opcode, callee, args)
+    }
+}
+
+/// Walk the given function, invoke the `Inline` implementation for each call
+/// instruction, and inline the callee when directed to do so.
+///
+/// Returns whether any call was inlined.
+pub(crate) fn do_inlining(func: &mut ir::Function, inliner: impl Inline) -> CodegenResult<bool> {
+    let mut inlined_any = false;
+    let mut allocs = InliningAllocs::default();
+
+    let mut cursor = FuncCursor::new(func);
+    while let Some(block) = cursor.next_block() {
+        // Always keep track of our previous cursor position. After we inline a
+        // call, replacing the current position with an arbitrary sub-CFG, we
+        // back up to this previous position. This makes sure our cursor is
+        // always at a position that is inserted in the layout and also enables
+        // multi-level inlining, if desired by the user, where we consider any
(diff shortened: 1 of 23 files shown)


cranelift/codegen/src/inline.rs
```diff
@@ -0,0 +1,1501 @@
...
+/// A command directing Cranelift whether or not to inline a particular call.
+pub enum InlineCommand<'a> {
+    /// Keep the call as-is, out-of-line, and do not inline the callee.
+    KeepCall,
+    /// Inline the call, using this function as the body of the callee.
+    ///
+    /// It is the `Inline` implementor's responsibility to ensure that this
+    /// function is the correct callee. Providing the wrong function may result
+    /// in panics during compilation or incorrect runtime behavior.
+    Inline(&'a ir::Function),
```
cfallin: Would it make sense to provide a `Cow<'a, ir::Function>` here? I can imagine use-cases that might want to synthesize a body to inline (for example: intrinsics, perhaps even dynamically generated based on parameters) rather than holding ownership of an existing function elsewhere.

cranelift/codegen/src/inline.rs
```diff
@@ -0,0 +1,1501 @@
...
+    /// Map from callee value to inlined caller value.
+    values: SecondaryMap<ir::Value, PackedOption<ir::Value>>,
+
+    /// The set of _callee_ global values that are derived (directly or
+    /// transitively) from the `vmctx`.
+    ///
+    /// Because the `vmctx` of the callee might not be the same as the `vmctx`
+    /// of the caller, we translate these globals differently from others, so we
+    /// must know which globals these are.
+    vmctx_global_values: SecondaryMap<ir::GlobalValue, PackedOption<ir::Value>>,
```
cfallin: A thought -- not sure how I feel about it, but worth noting to see what you think:

The vmctx handling here feels a little awkward, and certainly costs something; what do you think about requiring the inlined function to have been legalized first, and erroring out if we see a `global_value` operator during inlining?

On the one hand, it's a little less user-friendly (not just any function body can be inlined -- one must ensure it's in a particular state), but on the other hand, it feels much more orthogonal and well-factored from a compiler-pass point of view -- the point of legalization is to lower the implicit function-scoped GVs to ordinary (compositional, fully explicit) operators, so we might as well use that logic.

cranelift/codegen/src/inline.rs
```diff
@@ -0,0 +1,1501 @@
...
+        ir::InstructionData::TryCall {
+            opcode: ir::Opcode::TryCall,
+            args: _,
+            func_ref: _,
+            exception,
+        } => {
+            let normal_return = func.dfg.exception_tables[exception].normal_return();
+            let normal_return_block = normal_return.block(&func.dfg.value_lists);
+
+            // Check to see if we can reuse the existing normal-return block.
```
cfallin, line 976: This seems like a reasonable case to optimize here (will be a very frequent case) but it might be worth noting that it could go away, simplifying logic here, if we add a jump-threading optimization pass later...

cranelift/codegen/src/ir/exception_table.rs
```diff
@@ -156,11 +177,32 @@ impl ExceptionTableData {
...
+        self.tags.push(tag.into());
+
+        // We don't want to push the target, because the normal return is always
+        // last. Instead we want to insert it in the second to last spot.
+        //
+        // It is always true that `self.tags.len() == self.targets.len() + 1`
+        // because of the presence of the normal return which does not have an
+        // associated tag *except* for the case where we have cleared the
+        // exception table, in which case both vecs will be empty. We handle the
+        // cleared case by using saturating subtraction here.
```
cfallin: This feels like a somewhat error-prone API -- it imposes an ordering requirement (or rather, the first `push_catch` is not really a catch but the normal return). Perhaps we could have a separate setter for the normal return entry starting from an empty exception table, and then assert here that the normal return is present?

cfallin (approved):
This looks great overall -- thanks a bunch for building this and the care taken in handling all the combinations of kinds of calls!

A few minor thoughts below. `vmctx` handling in particular is interesting and I'm wondering if we should punt some complexity by leaning on legalization instead; could be convinced either way though.

cranelift/codegen/src/inline.rs
```diff
@@ -0,0 +1,1501 @@
...
+                    // exception tables. This can involve rewriting regular `call`s
+                    // into `try_call`s, which requires mutating the CFG because
+                    // `try_call` is a block terminator. However, we can't mutate
+                    // the CFG in the middle of this traversal because we rely on
+                    // the existence of a one-to-one mapping between the callee
+                    // layout and the inlined layout. Instead, we record the set of
+                    // inlined call instructions that will need fixing up, and
+                    // perform that possibly-CFG-mutating exception table merging in
+                    // a follow up pass, when we no longer rely on that one-to-one
+                    // layout mapping.
```
bjorn3, line 395: I don't think we can inline `try_call` at all without frontend specific knowledge. For example for `cg_clif`, a regular call to `_Unwind_Resume` is used to resume unwinding to the caller of the current function. The inliner would need to know to replace `_Unwind_Resume` calls with a jump to the right exception table entry. Just turning it into a `try_call` wouldn't work as `_Unwind_Resume` doesn't call the personality function again for the current function, but rather always unwinds to the caller.

cranelift/codegen/src/inline.rs
fitzgen: Good idea, will do :+1:

cranelift/codegen/src/inline.rs
fitzgen: Yeah I wavered back and forth on this myself, but ultimately it didn't seem too hard to support global values, even with the different-`vmctx` case, so I went ahead and did it. I guess unless you feel strongly that we should require legalization before inlining, I would rather not put effort into changing what is already there, given that I don't think there is a clear winner between the two choices. But I am willing to make this change if you really think it is the better approach.

cranelift/codegen/src/inline.rs
fitzgen: True, will add a comment

cranelift/codegen/src/ir/exception_table.rs
fitzgen: Will add the assert. I don't have any use case for actually building a new exception table from scratch after it has been cleared, and no other place in the code base does either, so I'll avoid the push-normal-return method for now.

I think ideally this type would internally have the equivalent of `enum { Cleared, Full(...) }` so that *if* we have the vecs then we know we aren't cleared and can always rely on their length relationship invariant, but that is a bit outside the scope of this PR, IMO.

cranelift/codegen/src/inline.rs
fitzgen: > without frontend specific knowledge

I think that should be fine because the frontend is in control of deciding whether to inline any given call, and has the information about which function is being called and with which call opcode is being used (`try_call` vs regular `call` vs `return_call`), so the frontend can simply tell Cranelift not to inline any call that would be problematic in the ways you describe, or in any other ways. This inliner just needs to be able to do the actual inlining, given the precondition that it is valid to do said inlining (which is what the frontend is promising by saying "yes inline this, here is a body").

fitzgen: (Furthermore, because there is no "throw" instruction in CLIF, the only way of throwing an exception is via a call into some runtime function (libunwind, the Wasmtime runtime, C++'s `_Unwind*` runtime, etc...), and therefore it is invalid to inline that particular call because then the throw could not happen, which breaks the equivalent-semantics promise that the frontend is making by saying "please inline this call with this callee body." On the other hand, it is fine to inline a `try_call` of a function that does not throw and does not make any calls that could throw; in this case we won't end up using the original `try_call`'s exception edges at all, which is the correct thing to do as those exception edges are unreachable from the inlined call's body -- although the exception edges' target blocks may be reachable from other edges -- as we can statically see that there is no call that could throw and unwind along those exceptional edges.)

cranelift/codegen/src/inline.rs
cfallin: On a little further thought: I do think it would actually be better to remove the vmctx-specific logic here, for a few maintenance- and risk-related reasons:

- We're spreading lowering knowledge around -- we're embodying the mapping between particular kinds of GVs and, say, loads or adds here as well as in the legalization phase. If we add any other sorts of GVs later (I know we want to go in the opposite direction, but...) this adds some more inlining-specific work that needs careful thought.
- The `vmctx`-related GVs in particular are pretty critical for sandboxing logic in Wasmtime, and I'd rather keep that as contained as possible -- i.e., our very well curated `bounds_check.rs` plus the one bit of legalization logic. The risk (however remote) that we mess up something here, now or later, and cross wires with the wrong vmctx, exists; and if/when we try to push something like PCC further, it seems much more difficult if we're doing ad-hoc translation here as well as in the legalization phase.
- If we try to move legalization to ISLE later then we'd also have the question whether to reuse that ISLE context here, or do something else; as above, IMHO it would be better to encapsulate all the knowledge in one place.

Finally I suspect that this requirement wouldn't place too much burden on "typical" use-cases: either one is doing a bottom-up-on-SCC-callgraph general inlining pass, and one will have done optimizations (which start with legalization) already on bodies that are considered for inlining; or one is building a specific compile-time builtin / intrinsification mechanism and has full mutable access to the special function bodies one is inlining as they are prepared. In other words I don't *think* it will be common to be stuck with a `&ir::Function` that's not legalized and have to clone it or whatnot.

All that said: still happy to discuss more if there are things I've missed!

cfallin:
(I think your add-to-queue crossed paths with my comment above -- see discussion on vmctx)

PR #11213, opened 2025-07-10
Add option to use regalloc3 as the register allocator

Amanieu:
This is based on https://github.com/bytecodealliance/regalloc2/pull/230 which adds regalloc3 as a back-end to regalloc2.

Only one change had to be made to codegen: tail calls were reserving `r11` as a scratch register by making it a def on the return instruction. This is not supported by regalloc3, which allows it extends all live ranges to the end of `ret` instruction so that the live range segment can merge with the next one. Since a def on a return instruction is meaningless, I've changed it to a hard-coded register in the emit stage instead.

### DIFF
--- a/cranelift/codegen/src/isa/x64/inst/emit.rs
+++ b/cranelift/codegen/src/isa/x64/inst/emit.rs
@@ -1987,6 +1987,9 @@ fn emit_return_call_common_sequence<T>(
     // Hard-coded register which doesn't conflict with function arguments or
     // callee-saved registers.
     let tmp = Writable::from_reg(regs::r11());
+    for pair in &call_info.uses {
+        debug_assert_ne!(pair.preg, regs::r11());
+    }
 
     for inst in
         X64ABIMachineSpec::gen_clobber_restore(CallConv::Tail, &info.flags, state.frame_layout())
--- a/cranelift/codegen/src/isa/x64/inst/emit.rs
+++ b/cranelift/codegen/src/isa/x64/inst/emit.rs
@@ -1984,7 +1984,9 @@ fn emit_return_call_common_sequence<T>(
                  but the current implementation relies on them being present"
     );
 
-    let tmp = call_info.tmp.to_writable_reg();
+    // Hard-coded register which doesn't conflict with function arguments or
+    // callee-saved registers.
+    let tmp = Writable::from_reg(regs::r11());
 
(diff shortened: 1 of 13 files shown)


cranelift/codegen/src/isa/x64/inst/emit.rs
```diff
@@ -1984,7 +1984,9 @@ fn emit_return_call_common_sequence<T>(
                  but the current implementation relies on them being present"
     );
 
-    let tmp = call_info.tmp.to_writable_reg();
+    // Hard-coded register which doesn't conflict with function arguments or
+    // callee-saved registers.
+    let tmp = Writable::from_reg(regs::r11());
```
fitzgen, line 1989: This might be true for our current calling conventions, but is not guaranteed to continue to hold true in the future. As such, could you assert that `call_info.uses` doesn't contain any uses of `r11` to future proof this code?

fitzgen (approved):
LGTM modulo comment below, removing the `[patch.crates-io]` bits when upstream is ready, and getting CI passing.

cranelift/codegen/src/isa/x64/inst/emit.rs
Amanieu: Actually, I've been over-thinking this. We can keep the existing code and just change it to an early-def, which *is* allowed by regalloc3 on return instructions. It could even be changed to an arbitrary register constraint in the future once regalloc2 support is dropped.

Amanieu: Actually an arbitrary register wouldn't work since it could conflict with callee-saved registers. I've added the assert as you suggested.

Amanieu:
Regarding CI, there are currently 3 issues:
- regalloc3 depends on brie-tree (fast B-Tree implementation) which has an MSRV of 1.88.
- brie-tree uses allocator-api 0.3 but other dependencies of wasmtime use allocator-api 0.2. This fails cargo deny due to the duplicate dependency.
- regalloc3/brie-tree probably need to be vetted. But this can wait until the crate is actually published.

cfallin:
As noted in bytecodealliance/regalloc2#230, Cranelift maintainers have decided not to take regalloc3 at this time due to maintenance-burden concerns, so I'll go ahead and close this. Thank you for all the effort regardless!

Issue #11179, opened 2025-07-03
Management of tables in Wasmtime is excessively `unsafe`


alexcrichton:
The current management of `crate::runtime::vm::Table` is excessively unsafe in the context of `crate::runtime::vm::Instance` in particular at this time. We should clean this up and make this safer.

Some examples of this pattern are:

* [x] [`get_table` returns `*mut Table`](https://github.com/bytecodealliance/wasmtime/blob/58e295ee58026086a91382cffe260d152951bb60/crates/wasmtime/src/runtime/vm/instance.rs#L1298)
* [x] [`get_table` and `get_defined_table`](https://github.com/bytecodealliance/wasmtime/blob/58e295ee58026086a91382cffe260d152951bb60/crates/wasmtime/src/runtime/vm/instance.rs#L1296-L1311) don't share code
* [x] [Helper methods go through unsafe `VMContext` -> `Pin<&mut Instance>` casting](https://github.com/bytecodealliance/wasmtime/blob/58e295ee58026086a91382cffe260d152951bb60/crates/wasmtime/src/runtime/vm/instance.rs#L1313-L1330)
* [x] [Going from `&VMTableDefinition` to `DefinedTableIndex` is not safe](https://github.com/bytecodealliance/wasmtime/blob/58e295ee58026086a91382cffe260d152951bb60/crates/wasmtime/src/runtime/vm/instance.rs#L688-L699)
* [x] [Table initialization requires unsafely simultaneously borrowing `Instance` and `Table` at the same time](https://github.com/bytecodealliance/wasmtime/blob/58e295ee58026086a91382cffe260d152951bb60/crates/wasmtime/src/runtime/vm/instance.rs#L992)
* [x] [Copying between tables is done with raw pointers](https://github.com/bytecodealliance/wasmtime/blob/58e295ee58026086a91382cffe260d152951bb60/crates/wasmtime/src/runtime/vm/table.rs#L911-L918)
* [ ] ~~[Only handing out raw pointers forces `wasmtime::Table` to have excessive `unsafe` blocks](https://github.com/bytecodealliance/wasmtime/blob/58e295ee58026086a91382cffe260d152951bb60/crates/wasmtime/src/runtime/externals/table.rs#L120)~~ (moved to https://github.com/bytecodealliance/wasmtime/issues/11262)

[...]

PR #11215, opened 2025-07-10
Make `vm::Instance::get_defined_table_with_lazy_init` return a reference

alexcrichton:
This commit updates the `get_defined_table_with_lazy_init` method to return a safe reference instead of a raw reference. This then cascaded outwards to updating more locations to handle this new safe reference as well. While this makes some further progress on #11179 there are a number of locations that will require using raw pointers until more refactoring is done.

### DIFF
--- a/crates/wasmtime/src/runtime/externals/table.rs
+++ b/crates/wasmtime/src/runtime/externals/table.rs
@@ -1,5 +1,5 @@
 use crate::prelude::*;
-use crate::runtime::vm::{self as runtime};
+use crate::runtime::vm::{self as runtime, GcStore};
 use crate::store::{AutoAssertNoGc, StoreInstanceId, StoreOpaque};
 use crate::trampoline::generate_table_export;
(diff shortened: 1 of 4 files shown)


crates/wasmtime/src/runtime/externals/table.rs
```diff
@@ -162,48 +169,48 @@ impl Table {
...
-        let table = self.wasmtime_table(&mut store, iter::once(index));
-        let gc_store = store.optional_gc_store_mut();
-        unsafe {
-            match (*table).get(gc_store, index)? {
-                runtime::TableElement::FuncRef(f) => {
-                    let func = f.map(|f| Func::from_vm_func_ref(&store, f));
-                    Some(func.into())
-                }
+        let (table, gc_store) = self.wasmtime_table(&mut store, iter::once(index));
+        // unsafe {
```
pchickey:
```suggestion
```

pchickey (approved):
Sorry, this one fell off my radar.

PR #10063, opened 2025-01-21
feat: add support for `wasi:clocks@0.3.0`

rvolosatovs:
Add support for `wasi:clocks@0.3.0`.

Blocked on https://github.com/bytecodealliance/wasmtime/pull/10061

PR #11221, opened 2025-07-11
feat: begin wasip3 implementation

rvolosatovs:
This starts the process of moving wasip3 implementation from https://github.com/bytecodealliance/wasip3-prototyping/tree/0bcd54e714e70dbee4dd2d440abb3017997a061e/crates/wasi

- implement `wasi:clocks`
    - just like in wasip2, `timezone` interface is not implemented yet
- implement `wasi:random`
- ~add stubs for `Access::get` and `Accessor::with`.~
   - ~for now, these are implemented directly in the file they're used in to avoid an import. There is only one occurrence where we need these in this PR and the hoping is that by the time next PR lands, we'll have them available in the runtime.~
- add tests, ~but ignore them, since async machinery still panics.~ the tests are adapted from analogous p2 tests
- allow specifying a subdirectory in WIT fetch script, since p3 WIT lives in a draft subdirectory
- extract a common `WasiCtxBuilder` to use for both p2 and p3. Unfortunately, inheritance is sort of tricky here - I originally hoped to implement `Deref` on the p2/p3 builders, but that breaks the chaining that users are used to, e.g. `builder.inherit_env().build()` would not work, since `inherit_env` would return a reference to the inner builder, rather than the outer one. Just duplicate the methods for p2 and p3 for now - let's revisit once we have all of the interfaces implemented and more context on how we can handle this. A macro is potential solution. The reason these contexts are different types is primarily stdio - the abstractions are very different for wasip2 and wasip3.

closes #10061 
closes #10063 

refs https://github.com/bytecodealliance/wasip3-prototyping/issues/228


rvolosatovs:
https://github.com/bytecodealliance/wasmtime/pull/11127 has been added to the merge queue, so looks like I'll have to rebase this on Monday, moving back to draft

crates/test-programs/src/p3/mod.rs
```diff
@@ -0,0 +1,19 @@
...
+            export wasi:cli/run@0.3.0;
+        }
+    ",
+    path: "../wasi/src/p3/wit",
+    world: "wasmtime:test/testp3",
+    default_bindings_module: "test_programs::p3",
+    pub_export_macro: true,
+    async: [
+        "wasi:cli/run@0.3.0#run",
+    ],
```
alexcrichton, line 17: Could the `wasi:cli/run#run` function be directly marked as `async` to avoid needing to annotate manually here?

crates/wasi/src/p3/bindings.rs
```diff
@@ -0,0 +1,306 @@
...
+                "wasi:filesystem/types@0.3.0#[async method]descriptor.symlink-at",
+                "wasi:filesystem/types@0.3.0#[async method]descriptor.unlink-file-at",
+                "wasi:filesystem/types@0.3.0#[async method]descriptor.is-same-object",
+                "wasi:filesystem/types@0.3.0#[async method]descriptor.metadata-hash",
+                "wasi:filesystem/types@0.3.0#[async method]descriptor.metadata-hash-at",
+                "wasi:sockets/ip-name-lookup@0.3.0#[async]resolve-addresses",
+                "wasi:sockets/types@0.3.0#[async method]tcp-socket.connect",
+                "wasi:sockets/types@0.3.0#[async method]tcp-socket.send",
+                "wasi:sockets/types@0.3.0#[async method]udp-socket.receive",
+                "wasi:sockets/types@0.3.0#[async method]udp-socket.send",
```
alexcrichton, line 162: For all `[async...]` things, could the original WITs be updated to have `async`?

I realize the is basically the same as my above comment and I also realize that I'm at fault here where I changed the WITs in-repo in wasip3-prototyping, but as we merge upstream this is probably as good a time as any to send PRs to upstream WIT definitions to get `async` in WITs

crates/wasi/src/p3/bindings.rs
```diff
@@ -0,0 +1,306 @@
...
+//!             "wasi:sockets/types@0.3.0#[async method]udp-socket.receive",
+//!             "wasi:sockets/types@0.3.0#[async method]udp-socket.send",
+//!             "wasi:sockets/types@0.3.0#[method]tcp-socket.bind",
+//!             "wasi:sockets/types@0.3.0#[method]tcp-socket.listen",
+//!             "wasi:sockets/types@0.3.0#[method]tcp-socket.receive",
+//!             "wasi:sockets/types@0.3.0#[method]udp-socket.bind",
+//!             "wasi:sockets/types@0.3.0#[method]udp-socket.connect",
+//!         ],
+//!     },
+//! });
```
alexcrichton, line 85: Ideally we're going to want to smith this if we can, the invocation here is quite large and I'd want to dig in to why the `async: { ... }` bits need to be specified (if they even still do, unsure)

crates/wasi/src/p3/clocks/host.rs
```diff
@@ -0,0 +1,114 @@
...
+    fn get(&mut self) -> D::Data<'_> {
+        todo!()
+    }
+}
+
+// TODO: Remove once `Access::get` is implemented
+impl<T, D> AccessGet<D> for wasmtime::component::Access<'_, T, D> where
+    D: wasmtime::component::HasData
+{
+}
```
alexcrichton: I think it should be possible to remove these now (yay!, hopefully at least)

crates/wasi/src/p3/clocks/mod.rs
```diff
@@ -0,0 +1,71 @@
+mod host;
+
+use crate::clocks::{WasiClocksImpl, WasiClocksView};
+use crate::p3::bindings::clocks;
+use wasmtime::component::{HasData, Linker};
+
+/// Add all WASI interfaces from this module into the `linker` provided.
```
alexcrichton, line 7: Could this documentation, and I imagine all other related documentation too for similar functions, also have a clickable link to the "top-level" `add_to_linker` in case folks land on these docs before seeing that?

crates/wasi/tests/all/p3/mod.rs
```diff
@@ -0,0 +1,99 @@
+#![cfg(feature = "p3")]
```
alexcrichton: I think this is a duplicate attribute given the one in `main.rs` too

crates/wasi/tests/all/p3/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+        panic!("should not use wasip2")
+    }
+}
+
+// TODO: Remove once test components are not built for `wasm32-wasip1`
+impl wasmtime_wasi::p2::IoView for Ctx {
+    fn table(&mut self) -> &mut ResourceTable {
+        panic!("should not use wasip2")
+    }
+}
```
alexcrichton: Could these get filled out to assist with debugging? That way `println!` in tests should work and such I believe. It's fine to leave the `TODO` comments though because it's true that yes, ideally, the components don't use WASIp2 at all for tests here (but there's a long way to get there)

crates/wasi/src/p3/filesystem/mod.rs
```diff
@@ -0,0 +1,17 @@
+use crate::{DirPerms, FilePerms, OpenMode};
+
+// TODO: implement
+pub struct Dir;
+
+impl Dir {
+    #[allow(unused)]
```
alexcrichton: Mind changing this to `#[expect(...)]`? (that way it forces removal of the attribute once it's used)

alexcrichton (approved):
A few minor thoughts here and there, but overall looks great, thank you @rvolosatovs! With #11127 merged as well mind removing the `#[ignore]` on the tests to confirm they pass too?

crates/wasi/src/p3/bindings.rs
rvolosatovs: Documented in #11246 and added a TODO with a ref

crates/test-programs/src/p3/mod.rs
rvolosatovs: See #11246

crates/wasi/src/p3/clocks/mod.rs
rvolosatovs: Screenshot 2025-07-16 at 12 43 14

rvolosatovs:
@alexcrichton review comments addressed, mind giving this another look?

alexcrichton (approved):

PR #11228, opened 2025-07-11
Make direct calls to statically-known imported functions

fitzgen:
This will allow for their eventual inlining.

Builds on top of https://github.com/bytecodealliance/wasmtime/pull/11225


crates/environ/src/component/translate.rs
```diff
@@ -492,12 +495,126 @@ impl<'a, 'data> Translator<'a, 'data> {
...
+
+                    // We could in theory inline these trampolines, so it could
+                    // make sense to record that we know this imported function
+                    // is this particular trampoline. However, everything else
+                    // is based around (module, defined-function) pairs and
+                    // trampolines don't fit that paradigm. Also, things get
+                    // really tricky when we consider the stack pointer, frame
+                    // pointer, and return address note-taking that trampolines
+                    // do for the stack walker. Therefore, we don't bother
+                    // turning these into inlinable direct calls.
```
alexcrichton: Mind updating this to say that calling a trampoline directly is totally fine, it's just the inlining part that's probably a no-no.

crates/environ/src/component/translate.rs
```diff
@@ -492,12 +495,126 @@ impl<'a, 'data> Translator<'a, 'data> {
...
+                            // to bottom out the instantiation argument in
+                            // either a definition or an import at the root
+                            // component boundary. In practice, this pattern is
+                            // rare, so following these chains is left for the
+                            // Future.
+                            continue;
+                        };
+
+                        self.static_modules[module].known_imported_functions[imported_func] =
+                            Some((*arg_module, arg_module_def_func));
```
alexcrichton, line 618: Could this assert that whatever was previously there is `None`?

alexcrichton (approved):
Can you additionally add a test that shows the full assembly for a simple A->B call? That way we can guarantee it's exercising the linking code path as well

PR #11229, opened 2025-07-11
Gut `vm::Export` to mostly be `crate::Extern`

alexcrichton:
This series of commits removes all of the `vm::Export*` types from Wasmtime. These were all relatively unsafe types which were schlepping around unsafe pointers from point A to point B. In lieu of this and given prior refactorings this commit replaces all of these items with the top-level crate definitions (e.g. `crate::Func` instead of `ExportFunc`). The goal here is to improve safety and generally reduce `unsafe` functions and blocks necessary. A nice benefit is that from a `vm::Instance` you can now directly load all top-level types such as `crate::Table`.

### DIFF
--- a/crates/wasmtime/src/runtime/vm/instance.rs
+++ b/crates/wasmtime/src/runtime/vm/instance.rs
@@ -573,12 +573,21 @@ impl Instance {
     /// # Panics
     ///
     /// Panics if `index` is out of bounds for this instance.
-    pub fn get_exported_func(
+    ///
+    /// # Safety
+    ///
+    /// The `store` parameter must be the store that owns this instance and the
+    /// functions that this instance can reference.
+    pub unsafe fn get_exported_func(
         self: Pin<&mut Self>,
         store: StoreId,
         index: FuncIndex,
     ) -> crate::Func {
         let func_ref = self.get_func_ref(index).unwrap();
+
+        // SAFETY: the validity of `func_ref` is guaranteed by the validity of
+        // `self`, and the contract that `store` must own `func_ref` is a
+        // contract of this function itself.
         unsafe { crate::Func::from_vm_func_ref(store, func_ref) }
     }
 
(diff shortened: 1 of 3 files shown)


crates/wasmtime/src/runtime/externals/table.rs
```diff
@@ -415,22 +414,6 @@ impl Table {
...
-    pub(crate) unsafe fn from_wasmtime_table(
-        wasmtime_export: crate::runtime::vm::ExportTable,
-        store: &StoreOpaque,
-    ) -> Table {
-        debug_assert!(
-            wasmtime_export
-                .table
-                .ref_type
-                .is_canonicalized_for_runtime_usage()
-        );
```
fitzgen, line 427: yesss

crates/wasmtime/src/runtime/trampoline/global.rs
```diff
@@ -32,7 +32,8 @@ pub fn generate_global_export(
     });
 
     let mut store = AutoAssertNoGc::new(store);
-    let definition = unsafe {
+    // TODO
+    unsafe {
```
fitzgen: Needs safety comment

crates/wasmtime/src/runtime/vm/instance.rs
```diff
@@ -573,9 +573,13 @@ impl Instance {
...
     /// Panics if `index` is out of bounds for this instance.
-    pub fn get_exported_func(self: Pin<&mut Self>, index: FuncIndex) -> ExportFunction {
+    pub fn get_exported_func(
+        self: Pin<&mut Self>,
+        store: StoreId,
+        index: FuncIndex,
+    ) -> crate::Func {
         let func_ref = self.get_func_ref(index).unwrap();
-        ExportFunction { func_ref }
+        unsafe { crate::Func::from_vm_func_ref(store, func_ref) }
```
fitzgen: Can we actually encapsulate the unsafety here? Because `from_vm_func_ref` requires that the store owns the given funcref, but because we take the store as an argument but get the funcref from `self`, we don't technically have the guarantee that the given store owns this funcref. So I think either this function should be `unsafe` because it relies on passing the correct `StoreId` in, or we need to get that `StoreId` from `self` somehow.

fitzgen (approved):
LGTM modulo one safety comment/question below that should be addressed before merging

crates/wasmtime/src/runtime/vm/instance.rs
alexcrichton: Good point yeah, I meant to come back and fix this up but forgot to do so. It's a bit unfortunate how infectious the unsafety is here but I agree it's required to be sound regardless.

PR #11230, opened 2025-07-12
Implement exception objects.

cfallin:
This PR implements *exception objects*: the data model required to support Wasm exceptions, integrated with Wasm GC.

It builds out support for defining exception types (based on function signatures), and a host API to allocate and examine exception objects. It does *not* yet tie this into any exception-throwing or -catching instructions (`throw`, `throw_ref`, `try_table`); that will come next.

The design is a little fiddly because the type system has a nominal aspect to it: tags are *per-instance* entities and exceptions are associated with tags. The stack-switching work already added support for tags analogously to other instance entities (memories, tables, etc.). But, notably, because they are a runtime concept, they are not types per-se. Their signatures are, though, and `TagType` is a thin wrapper around `FuncType` as a result. The Wasm proposal does not define any concrete type definitions for exception objects in the lattice, only providing top (`exnref` / `(ref null exn)`) and bottom (`nullexnref` / `(ref null noexn)`) types. In order to align with the way the rest of the GC system works, I've opted to define our own notion of structural types at the host API and type-interning layer built around the signatures; so all Wasm exception objects with the same func-type / signature are the same "type" as far as GC is concerned. The objects of that type are then associated with particular tag instances via runtime fields. In essence we are refining the Wasm types further to allow our implementation to precompute things (i.e., layouts); this is fine because this refinement is not exposed to Wasm.

### DIFF
--- a/crates/environ/src/gc.rs
+++ b/crates/environ/src/gc.rs
@@ -15,11 +15,11 @@ pub mod drc;
 #[cfg(feature = "gc-null")]
 pub mod null;
 
-use crate::prelude::*;
 use crate::{
     WasmArrayType, WasmCompositeInnerType, WasmCompositeType, WasmStorageType, WasmStructType,
     WasmValType,
 };
+use crate::{WasmExnType, prelude::*};
 use core::alloc::Layout;
 
 /// Discriminant to check whether GC reference is an `i31ref` or not.
@@ -119,17 +119,15 @@ fn common_array_layout(
     }
 }
 
-/// Common code to define a GC struct's layout, given the size and alignment of
-/// the collector's GC header and its expected offset of the array length field.
+/// Shared layout code for structs and exception objects, which are
+/// identical except for the tag field (present in
+/// exceptions). Returns `(size, align, fields)`.
 #[cfg(any(feature = "gc-null", feature = "gc-drc"))]
-fn common_struct_layout(
-    ty: &WasmStructType,
+fn common_struct_or_exn_layout(
+    fields: &[crate::WasmFieldType],
     header_size: u32,
     header_align: u32,
-) -> GcStructLayout {
-    assert!(header_size >= crate::VM_GC_HEADER_SIZE);
-    assert!(header_align >= crate::VM_GC_HEADER_ALIGN);
-
+) -> (u32, u32, Vec<GcStructLayoutField>) {
     // Process each field, aligning it to its natural alignment.
     //
     // We don't try and do any fancy field reordering to minimize padding (yet?)
@@ -138,11 +136,11 @@ fn common_struct_layout(
     // subtyping where we need a subtype's fields to be assigned the same
     // offsets as its supertype's fields. We can come back and improve things
     // here if we find that (a) isn't actually holding true in practice.
+
     let mut size = header_size;
     let mut align = header_align;
 
-    let fields = ty
-        .fields
+    let fields = fields
         .iter()
         .map(|f| {
             let field_size = byte_size_of_wasm_ty_in_gc_heap(&f.element_type);
@@ -157,9 +155,49 @@ fn common_struct_layout(
     let align_size_to = align;
     align_up(&mut size, &mut align, align_size_to);
 
+    (size, align, fields)
+}
+
+/// Common code to define a GC struct's layout, given the size and alignment of
+/// the collector's GC header and its expected offset of the array length field.
+#[cfg(any(feature = "gc-null", feature = "gc-drc"))]
+fn common_struct_layout(
+    ty: &WasmStructType,
+    header_size: u32,
+    header_align: u32,
+) -> GcStructLayout {
+    assert!(header_size >= crate::VM_GC_HEADER_SIZE);
+    assert!(header_align >= crate::VM_GC_HEADER_ALIGN);
+
+    let (size, align, fields) = common_struct_or_exn_layout(&ty.fields, header_size, header_align);
+
     GcStructLayout {
         size,
         align,
+        fields,
+    }
+}
+
+/// Common code to define a GC exception object's layout, given the
+/// size and alignment of the collector's GC header and its expected
+/// offset of the array length field.
+#[cfg(any(feature = "gc-null", feature = "gc-drc"))]
+fn common_exn_layout(ty: &WasmExnType, header_size: u32, header_align: u32) -> GcExceptionLayout {
+    assert!(header_size >= crate::VM_GC_HEADER_SIZE);
+    assert!(header_align >= crate::VM_GC_HEADER_ALIGN);
+
+    // Compute a struct layout, with extra header size for the
+    // `(instance_idx, tag_idx)` fields.
+    let tag_offset = header_size;
+    assert!(header_align >= 8);
+    let header_size = header_size + 2 * u32::try_from(core::mem::size_of::<u32>()).unwrap();
+
+    let (size, align, fields) = common_struct_or_exn_layout(&ty.fields, header_size, header_align);
+
+    GcExceptionLayout {
+        size,
+        align,
+        tag_offset,
         fields,
     }
 }
@@ -186,6 +224,7 @@ pub trait GcTypeLayouts {
             WasmCompositeInnerType::Cont(_) => {
                 unimplemented!("Stack switching feature not compatbile with GC, yet")
             }
+            WasmCompositeInnerType::Exn(ty) => Some(self.exn_layout(ty).into()),
         }
     }
 
@@ -194,6 +233,9 @@ pub trait GcTypeLayouts {
 
     /// Get this collector's layout for the given struct type.
     fn struct_layout(&self, ty: &WasmStructType) -> GcStructLayout;
+
+    /// Get this collector's layout for the given exception type.
+    fn exn_layout(&self, ty: &WasmExnType) -> GcExceptionLayout;
 }
 
 /// The layout of a GC-managed object.
@@ -204,6 +246,9 @@ pub enum GcLayout {
 
     /// The layout of a GC-managed struct object.
     Struct(GcStructLayout),
+
+    /// The layout of a GC-managed exception object.
+    Exception(GcExceptionLayout),
 }
 
 impl From<GcArrayLayout> for GcLayout {
@@ -215,6 +260,12 @@ impl From<GcArrayLayout> for GcLayout {
 impl From<GcStructLayout> for GcLayout {
     fn from(layout: GcStructLayout) -> Self {
         Self::Struct(layout)
+    }
+}
+
+impl From<GcExceptionLayout> for GcLayout {
+    fn from(layout: GcExceptionLayout) -> Self {
+        Self::Exception(layout)
     }
 }
 
@@ -234,6 +285,15 @@ impl GcLayout {
         match self {
             Self::Array(a) => a,
             _ => panic!("GcLayout::unwrap_array on non-array GC layout"),
+        }
+    }
+
+    /// Get the underlying `GcExceptionLayout`, or panic.
+    #[track_caller]
+    pub fn unwrap_exception(&self) -> &GcExceptionLayout {
+        match self {
+            Self::Exception(e) => e,
+            _ => panic!("GcLayout::unwrap_exception on a non-exception GC layout"),
         }
     }
 }
@@ -337,6 +397,41 @@ pub struct GcStructLayoutField {
     pub is_gc_ref: bool,
 }
 
+/// The layout for a GC-managed exception object.
+///
+/// This layout is only valid for use with the GC runtime that created it. It is
+/// not valid to use one GC runtime's layout with another GC runtime, doing so
+/// is memory safe but will lead to general incorrectness like panics and wrong
+/// results.
+///
+/// All offsets are from the start of the object; that is, the size of the GC
+/// header (for example) is included in the offset.
+#[derive(Clone, Debug)]
+pub struct GcExceptionLayout {
+    /// The size (in bytes) of this struct.
+    pub size: u32,
+
+    /// The alignment (in bytes) of this struct.
+    pub align: u32,
+
+    /// The offset of the VMTagImport pointer.
+    pub tag_offset: u32,
+
+    /// The fields of this exception object. The `i`th entry contains
+    /// information about the `i`th parameter in the associated tag
+    /// type.
+    pub fields: Vec<GcStructLayoutField>,
+}
+
+impl GcExceptionLayout {
+    /// Get a `core::alloc::Layout` for an exception object of this type.
+    pub fn layout(&self) -> Layout {
+        let size = usize::try_from(self.size).unwrap();
+        let align = usize::try_from(self.align).unwrap();
+        Layout::from_size_align(size, align).unwrap()
+    }
+}
+
 /// The kind of an object in a GC heap.
 ///
 /// Note that this type is accessed from Wasm JIT code.
@@ -367,11 +462,12 @@ pub struct GcStructLayoutField {
 #[rustfmt::skip]
 #[expect(missing_docs, reason = "self-describing variants")]
 pub enum VMGcKind {
-    ExternRef      = 0b01000 << 27,
-    AnyRef         = 0b10000 << 27,
-    EqRef          = 0b10100 << 27,
-    ArrayRef       = 0b10101 << 27,
-    StructRef      = 0b10110 << 27,
+    ExternRef      = 0b010000 << 26,
+    AnyRef         = 0b100000 << 26,
+    EqRef          = 0b101000 << 26,
+    ArrayRef       = 0b101010 << 26,
+    StructRef      = 0b101100 << 26,
+    ExnRef         = 0b000001 << 26,
 }
 
 /// The size of the `VMGcKind` in bytes.
@@ -381,7 +477,7 @@ const _: () = assert!(VM_GC_KIND_SIZE as usize == core::mem::size_of::<VMGcKind>
 
 impl VMGcKind {
     /// Mask this value with a `u32` to get just the bits that `VMGcKind` uses.
-    pub const MASK: u32 = 0b11111 << 27;
+    pub const MASK: u32 = 0b111111 << 26;
 
     /// Mask this value with a `u32` that potentially contains a `VMGcKind` to
     /// get the bits that `VMGcKind` doesn't use.
@@ -404,6 +500,7 @@ impl VMGcKind {
             x if x == Self::EqRef.as_u32() => Self::EqRef,
             x if x == Self::ArrayRef.as_u32() => Self::ArrayRef,
             x if x == Self::StructRef.as_u32() => Self::StructRef,
+            x if x == Self::ExnRef.as_u32() => Self::ExnRef,
             _ => panic!("invalid `VMGcKind`: {masked:#032b}"),
         }
     }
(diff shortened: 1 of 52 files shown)


cfallin:
A few other missing bits that I will do before marking as ready:

- I haven't implemented the GC-disabled stubs yet.
- We'll need to think about how to hide this appropriately -- at the very least, behind the GC feature, but perhaps behind another feature flag as well, until we have the full exceptions proposal implemented? We don't enable exception support in wasmparser so no one will be accidentally fooled by a failure deep in unimplemented bits; but the presence of a host API may be misleading still.
- No fuzzing yet; I'll build that out once I have the rest of the exceptions proposal done too.

crates/environ/src/gc.rs
```diff
@@ -367,11 +462,12 @@ pub struct GcStructLayoutField {
...
-    AnyRef         = 0b10000 << 27,
-    EqRef          = 0b10100 << 27,
-    ArrayRef       = 0b10101 << 27,
-    StructRef      = 0b10110 << 27,
+    ExternRef      = 0b010000 << 26,
+    AnyRef         = 0b100000 << 26,
+    EqRef          = 0b101000 << 26,
+    ArrayRef       = 0b101010 << 26,
+    StructRef      = 0b101100 << 26,
+    ExnRef         = 0b000001 << 26,
```
alexcrichton, line 470: Could the docs above this referenceing 27 bits be updated to refer to 26?

crates/wasmtime/src/runtime/coredump.rs
```diff
@@ -204,6 +204,11 @@ impl WasmCoreDump {
                     Val::AnyRef(_) => {
                         wasm_encoder::ConstExpr::ref_null(wasm_encoder::HeapType::ANY)
                     }
+                    Val::ExnRef(_) => {
+                        todo!(
+                            "Cannot encode a core dump with an exception object global because the wasm_encoder::HeapType does not exist"
+                        )
+                    }
```
alexcrichton: To avoid panicking during a coredump in the future by accident, could this return an error instead of panicking?

crates/wasmtime/src/runtime/externals/tag.rs
```diff
@@ -28,6 +31,43 @@ impl Tag {
...
+    /// [`Store`](`crate::Store`).
+    #[cfg(feature = "async")]
+    pub async fn new_async(mut store: impl AsContextMut<Data: Send>, ty: &TagType) -> Result<Tag> {
+        let mut store = store.as_context_mut();
+        assert!(
+            store.0.async_support(),
+            "cannot use `new_async` without enabling async support on the config"
+        );
+        store.on_fiber(|store| Tag::_new(store.0, ty)).await?
+    }
```
alexcrichton: It's ok to skip this since there's no limiter being invoked as part of `generate_tag_export`, that's just for memories/tables.

crates/wasmtime/src/runtime/vm/instance.rs
```diff
@@ -1045,6 +1050,9 @@ impl Instance {
                             )
                         }),
                     )?,
+                    WasmHeapTopType::Exn => {
+                        unreachable!("Cannot initialize exception objects from an element segment")
```
alexcrichton: Could this be a `bail!` instead of `unreachable!` in case this ever becomes possible in the future?

crates/wasmtime/src/runtime/gc/enabled/exnref.rs
```diff
@@ -0,0 +1,784 @@
...
+    }
+
+    #[inline]
+    fn dynamic_concrete_type_check(
+        &self,
+        store: &StoreOpaque,
+        _nullable: bool,
+        ty: &HeapType,
+    ) -> Result<()> {
+        self.ensure_matches_ty(store, ty)
```
alexcrichton: Technically this should alwayas be `Ok(())`, right? Given that it's impossible for a wasm module to request a typed exnref?

alexcrichton (commented):
Nice!

> We'll need to think about how to hide this appropriately 

Personally I think it's reasonable to keep this behind the `gc` feature flag for now and leave it at that. This won't bring in any significantly new runtime dependencies that `gc` doesn't already bring, so having it behind the same flag I think is reasonable. In that sense filling out similar-ish GC-disabled stubs for exnref I think is fine.

In the future this means that the hypothetical `Config::wasm_exceptions` knob will only be available with `feature = "gc"` and if the gc feature is disabled then this feature will always be disabled for wasmparser, effectively being where we're at today.

One hypothetical branch point we could reach in the future is to implement wasm exceptions entirely independently of wasm gc. In such a world we'd probably have a separate feature gate for exception-handling but in lieu of that I don't think it's necessary to add a new feature just for exceptions (assuming no new "substantial" runtime dependencies or such are needed, which AFAIK won't be the case)

---

As a small aside, it's a bit regrettable how much duplication in a sense there is necessary for this. Most of this PR looks like a copy/paste of the GC bits for structref but then edited to be for exnref instead. That's not a dealbreaker by any means and rustc more-or-less shoehorns these implementations to exactly what they are, but I also think it'd be reasonable if you see refactoring opportunities to apply them more aggressively too (e.g. what you did here with initialization/reading struct fields and refactoring that so exceptions can use the same path)

alexcrichton:
> The design is a little fiddly because the type system has a nominal aspect to it

Oh one thing I forgot, are you familiar enough with the JS API for exceptions to say whether or not the approach here is roughly analagous? I don't know myself what the JS API looks like so I'm not sure myself.

crates/environ/src/types.rs
```diff
@@ -838,6 +861,79 @@ impl TypeTrace for WasmContType {
...
+/// the type system here.
+///
+/// Because the Wasm type system does not have concrete `exn` types
+/// (i.e., the heap-type lattice has only top `exn` and bottom
+/// `noexn`), we are free to decide what we mean by "concrete type"
+/// here. Thus, we define an "exception type" to refer to the
+/// type-level *signature*. When a particular *exception object* is
+/// created in a store, it can be associated with a particular *tag
+/// instance* also in that store, and the compatibility is checked
+/// (the tag's function type must match teh function type in the
```
bjorn3: *the

cfallin:
> Oh one thing I forgot, are you familiar enough with the JS API for exceptions to say whether or not the approach here is roughly analagous? I don't know myself what the JS API looks like so I'm not sure myself.

As far as I can tell, this is pretty close to both the [Wasm JS API](https://developer.mozilla.org/en-US/docs/WebAssembly/Reference/JavaScript_interface/Exception) and the host API outlined in the [spec](https://webassembly.github.io/exception-handling/core/appendix/embedding.html#mathrm-exn-alloc-xref-exec-runtime-syntax-store-mathit-store-xref-exec-runtime-syntax-tagaddr-mathit-tagaddr-xref-exec-runtime-syntax-val-mathit-val-ast-xref-exec-runtime-syntax-store-mathit-store-xref-exec-runtime-syntax-exnaddr-mathit-exnaddr), modulo requiring the `ExnRefPre` explicitly in our case whereas the above APIs allow one to allocate an exception object solely with a runtime tag reference and payloads. This is more-or-less in line with the way we support allocating e.g. struct objects too -- separating the layout lookup/computation is a reasonable performance optimization IMHO.

cfallin:
> One hypothetical branch point we could reach in the future is to implement wasm exceptions entirely independently of wasm gc. In such a world we'd probably have a separate feature gate for exception-handling but in lieu of that I don't think it's necessary to add a new feature just for exceptions (assuming no new "substantial" runtime dependencies or such are needed, which AFAIK won't be the case)

Yeah, I thought about this quite a bit before building out this part of the proposal: I was wondering whether the two (exceptions and GC) could be kept orthogonal as much as possible, largely to avoid the effort in this PR. An exnref would then be more like an externref; maybe held as an `Option<Box<Exception>>` in a table on the `Store` or something like that. In the end I didn't like that for three main reasons: (i) exception objects can store GC refs as payload, so when both are enabled, one needs special handling anyway to trace edges in both directions; (ii) it's quite a lot of duplicated effort (computing how to lay out values in a struct-like container, handling layouts and interning them, etc.); (iii) if one addresses problem (i) by having a separate exnref implementation for with-GC and without-GC, then one has two code paths to test and that's awful. Plus in practice most users of exceptions will likely want GC too. (Perhaps with the exception of a setjmp/longjmp use-case in wasi-libc?) Or to summarize all the above: GC has enough overlap with exceptions that they just need to interact, so we might as well handle them with common bits.

> As a small aside, it's a bit regrettable how much duplication in a sense there is necessary for this. Most of this PR looks like a copy/paste of the GC bits for structref but then edited to be for exnref instead. That's not a dealbreaker by any means and rustc more-or-less shoehorns these implementations to exactly what they are, but I also think it'd be reasonable if you see refactoring opportunities to apply them more aggressively too (e.g. what you did here with initialization/reading struct fields and refactoring that so exceptions can use the same path)

Yeah, I don't like it too much either; but several alternatives I considered didn't feel great either. In particular one could imagine either using more dynamism internally (having a "struct or exception layout" that knows its kind dynamically and has the appropriate dynamic checks) or using traits or macros to factor out method implementations. In the end the surface bits needed to be separate for clean typesafe API reasons and the rest kind of fell out of that. Happy to take any particulara suggestions for refactors if anyone has any though!

alexcrichton:
> I was wondering whether the two (exceptions and GC) could be kept orthogonal as much as possible

FWIW I agree with your conclusions as well. I mostly brought this up to say that I think it would warrant a separate feature in such a hypothetical alternative world, but given the current implementation I don't think this warrants a separate feature or anything more than gating behind GC.

> In the end the surface bits needed to be separate for clean typesafe API reasons and the rest kind of fell out of that

FWIW this is mostly my takeaway as well. It's wordy, but easy to read and verify it's right. The downside is that if it's got a bug (which would likely be preexisting if any) there's a lot of places to chase down. I think it's best to leave that for the hypothetical future in which we discover this though. Otherwise though I don't have any silver bullet suggestions myself either.

crates/wasmtime/src/runtime/gc/enabled/exnref.rs
cfallin: Good point!

crates/wasmtime/src/runtime/vm/instance.rs
cfallin: Actually, updated to handle `Exn` with the same logic as `Any` (added to the match arm) -- the logic is generic and so if we ever need to support this it's already there.

crates/wasmtime/src/runtime/externals/tag.rs
cfallin: Done!

crates/wasmtime/src/runtime/coredump.rs
cfallin: One better, it's possible to create an exnref type manually (there's just no alias) -- done.

crates/environ/src/gc.rs
cfallin: Done!

crates/environ/src/types.rs
cfallin: Fixed, thanks!

cfallin:
I think this is ready for a proper review now (and I've addressed comments from Alex and bjorn3 already -- thanks!).

crates/wasmtime/src/runtime/externals/tag.rs
```diff
@@ -65,4 +83,30 @@ impl Tag {
...
+
+    /// Get the "index coordinates" for this `Tag`: the raw instance
+    /// ID and defined-tag index within that instance. This can be
+    /// used to "serialize" the tag as safe (tamper-proof,
+    /// bounds-checked) values, e.g. within the GC store for an
+    /// exception object.
+    #[cfg(feature = "gc")]
+    pub(crate) fn to_raw_indices(&self) -> (InstanceId, DefinedTagIndex) {
+        (self.instance.instance(), self.index)
+    }
```
fitzgen, line 95: Should this return a `StoreInstanceId` (i.e. a `(StoreId, InstanceId)` pair) instead of a plain `InstanceId` as the first tuple member here? That would make it fully "tamper-proof".

I am not sure where this method is used yet tho...

crates/wasmtime/src/runtime/vm/gc.rs
```diff
@@ -268,4 +270,24 @@ impl GcStore {
     pub fn array_len(&self, arrayref: &VMArrayRef) -> u32 {
         self.gc_heap.array_len(arrayref)
     }
+
+    /// Allocate an uninitialized exceptino object with the given type
```
fitzgen:
```suggestion
    /// Allocate an uninitialized exception object with the given type
```

"exceptino" would make a good pokemon name

crates/wasmtime/src/runtime/vm/gc/gc_runtime.rs
```diff
@@ -338,6 +342,46 @@ pub unsafe trait GcHeap: 'static + Send + Sync {
...
+    ///
+    /// The `ty` and `layout` must match, and the tag's function type
+    /// must have a matching signature to the exception layout's.
+    ///
+    /// Failure to do either of the above is memory safe, but may result in
+    /// general failures such as panics or incorrect results.
+    ///
+    /// Return values:
+    ///
+    /// * `Ok(Some(_))`: The allocation was successful.
```
fitzgen:
```suggestion
    /// * `Ok(Ok(_))`: The allocation was successful.
```

I guess the copy-paste source probably needs updates too...

fitzgen (approved):
LGTM!

crates/wasmtime/src/runtime/vm/gc.rs
jeffparsons: Isn't that the evolution of [_MissingNo._](https://en.wikipedia.org/wiki/MissingNo.)?

tschneidereit:
> Perhaps with the exception of a setjmp/longjmp use-case in wasi-libc?

I think this is a very load-bearing "perhaps", fwiw: I can very very easily see people wanting to have smallest-possible Wasmtime builds and not use GC. And I can equally easily see wasi-libc switch to using sjlj based on EH.

None of this might be enough to put in the work to support exceptions without GC, of course. But if it's more a lack of signal for a need for it: I would bet good money on us getting that signal before too long

sunfishcode:
wasi-libc does have a [setjmp/longjmp implementation using EH](https://github.com/WebAssembly/wasi-libc/pull/483), and does not support setjmp/longjmp otherwise.

cfallin:
Yes, that's fair, and I did discuss this with Nick early on re: overheads of requiring GC. It seems that the [lazy GC heap allocation](https://github.com/bytecodealliance/wasmtime/blob/64bc3bd982d256bdd608736533e704959bb537b4/crates/wasmtime/src/runtime/store.rs#L1483-L1488) means that we should be able to have GC enabled without paying a large penalty for it when not used. That, put against the huge duplication of code to also have a "GC-less exceptions" implementation (which would necessarily be almost fully separate, and would also require separate codegen paths), convinced me to go this route.

I'll note that the exceptions proposal's structuring unfortunately pushes us in this direction too: it does not hold back `catch_ref` clauses and `throw_ref` to a "when GC enabled" superset, but rather these are always available. It wouldn't be too bad to do an "unboxed exceptions" implementation that directly copies payload values and supports only `catch` clauses and `throw`, and indeed I started designing down that route in the hope that I could make setjmp/longjmp a happy fastpath, but it's just too much duplication to be maintainable, and also once you need to support the full proposal and box into exception objects, it adds dynamic checks to support the full matrix of `{boxed, unboxed} ^ 2`.

alexcrichton:
Personally while I agree with @cfallin's conclusions about this direction being the way to go for now I also believe that this issue won't be put to rest and this is something we're going to have to deal with. One example is that allocating a GC heap is akin to a second linear memory meaning that the per-Store overhead will jump from one linear memory to two linear memories for "lean" languages such as Rust/C/C++, just to support exceptions. (assuming guests eventually get to the point of using exceptions by default). That's then additionally coupled with the fact that it's unlikely embedders will want to interact deeply with exceptions (e.g. make their own, read fields, etc) and nor will Rust/C/C++ use the GC-like behavior of exceptions (e.g. a `catch_ref` is probably immediately always followed by a `throw_ref` later in the control flow).

I think these are all problems we're going to have to tackle at some point in the future. "Just turn on GC" as-is today I don't think is going to cut it in the long-term, but I also don't think that should apply stop-energy here either. This implementation is also going to be required in the fullness of time to support GC languages and the exceptions they throw/use. In essence I think that this implementation is basically inevitable and the only hypothetical missing piece is a leaner implementation of "only the exceptions bits needed for linear memory languages". That's hypothetical in that it's not certain we need it and it's also not certain what exact subset, if any, is needed.

crates/wasmtime/src/runtime/externals/tag.rs
cfallin: I thought about that, but this is stored within an object within the store, and when we reconstitute a `&Tag` we use the store it came from, so I think we should be fine in this case (vs. adding 4 bytes to every exception object).

crates/wasmtime/src/runtime/vm/gc/gc_runtime.rs
cfallin: Fixed!

crates/wasmtime/src/runtime/vm/gc.rs
cfallin: Fixed!

And fwiw, I believe exceptinos are the subatomic particles that convey exception throws at the speed of light...

fitzgen:
I think we will want to keep the implementation of exceptions in terms of GC objects, with possible unboxing optimizations that are really orthogonal to the core discussion here, but we will want to add config knobs dedicated for GC heaps. Right now, GC heaps are always the same configuration as linear memories. For LIME-style deployments, however, they will want to configure the GC heap to be much smaller than 64KiB (effectively using custom-page-sizes for GC heap memories), after which

> allocating a GC heap is akin to a second linear memory

is effectively resolved.

We really don't want to have to maintain two implementations of exceptions: one for LIME Wasm and one for GC Wasm. By allowing one to configure the GC heap so that it is tiny, we are effectively allowing the GC-based implementation to be tuned down into the equivalent side-table that we would implement for the LIME Wasm version.

tschneidereit:
> is effectively resolved.

Does this allow us to set the allocation to 0, or something really very small? Because if it still requires at least one host page size of memory, I'm not sure it really is effectively resolved for all use cases.

Besides memory usage, I'm also concerned about the code size increase we incur by forcing the inclusion of full GC support for exception handling. Do we have a way to quantify / at least very roughly estimate how much larger a size-optimized binary becomes because of that?

sunfishcode:
> It wouldn't be too bad to do an "unboxed exceptions" implementation that directly copies payload values and supports only `catch` clauses and `throw`

In the Lime series, we do have the ability to [subset features](https://github.com/WebAssembly/tool-conventions/blob/main/Lime.md#feature-subsets), for exactly this reason, so perhaps we should to that for EH too.

It seems likely that other engines will want a subset too. And the [setjmp/longjmp](https://github.com/WebAssembly/tool-conventions/blob/main/SetjmpLongjmp.md) used by LLVM and wasi-libc already uses just `catch` and `throw`.

fitzgen:
Meta note: I am going to file a follow up issue for this discussion. Once I do so, let's move all this discussion there.

fitzgen:
> Meta note: I am going to file a follow up issue for this discussion. Once I do so, let's move all this discussion there.

Filed https://github.com/bytecodealliance/wasmtime/issues/11256

PR #11235, opened 2025-07-14
Add `#[inline]` to some small functions

alexcrichton:
This adds `#[inline]` to some functions which are otherwise not available for cross-crate inlining. These functions started being more heavily used after historical refactorings such as #10877 so this helps benchmarks which access linear memory in host functions, for example.


adambratschikaye:
This fixes the regression in https://github.com/bytecodealliance/wasmtime/issues/11236

abrown (approved):

Issue #11224, opened 2025-07-11
Rework `component-model-async` APIs to use `Accessor` where appropriate


dicej:
Alex and I had a discussion and hack session today and came up with more ergonomic, harder to misuse, and easier-to-implement-efficiently APIs for e.g. `TypedFunc::call_concurrent`, `StreamReader::read`, etc. based on a refined version of the `Accessor` API we created for host functions.  Early experiments show that we can adopt those APIs without major refactoring of the runtime internals; later, we can update the internals to take advantage of the efficiencies those new APIs unlock.

PR #11238, opened 2025-07-14
Start to use `&Accessor<T, D>` more in concurrent code

alexcrichton:
After discussion with Joel we've concluded that while `&mut Accessor<T, D>` was originally added to model host functions it is also appropriate to use it to model embedder-rooted invocations of items such as wasm as well. Effectively the conclusion we reached was that `*::call_concurrent` should be taking `&Accessor`, not `StoreContextMut`. This has a number of benefits to it over the previous iteration:

* This makes exports behave more like imports where `Accessor` means "you're in the concurrent world".

* This makes exports have an `async fn` signature which is easier to read and understand.

* This automatically enforces the guarantee that the returned future is only polled within the main event loop because the future is always considered to close over the `&Accessor` provided meaning it statically cannot live outside of the event loop.

* This paves the way forward to future refactorings to avoid storing so much state within a `Store<T>` and instead try to store state directly in futures themselves. This should make cancellation more natural and eventually also remove `'static` bounds on params/results. Furthermore this should make it easier to avoid spawning tasks internally by storing state in futures instead of spawned tasks.

In doing this one of the main questions we were faced with was what to do about `&mut Accessor<T, D>`, namely the `mut` part. With a mutable accessor it would be only possible to call one function concurrently. One option considered was to add combinators like `Accessor::join` and `Accessor::race` but in the end we decided to avoid going that direction and instead switch to `&Accessor<T, D>` everywhere, freely enabling aliasing of the accessor. This has the downside that `Accessor::with` is now a relatively dangerous function in that it can panic, but idiomatic usage of it is not expected to panic as the distinction between the `async` and sync boundary of `Accessor` vs `StoreContextMut` is expected to naturally make the recursive panic condition of `with` rare to come up in practice.

Concrete changes in this commit are:

* `Accessor::with` now requires `&self`.
* `Accessor::spawn` now requires `&self`.
* Host functions are now given `&Accessor`, not `&mut Accessor`.
* `{Typed,}Func::call_concurrent` is now an `async fn` which takes an `&Accessor` instead of `StoreContextMut`.
* Guest bindings generation for concurrent invocations now looks exactly like async bindings generation except for replacing `StoreContextMut` with `Accessor`.

Note that this commit does not yet update the internal implementations of these functions to benefit from the new abilities that taking `&Accessor` implies. Instead that's deferred to a future update as necessary. Instead this is only updating the public API of the `wasmtime` crate to enable these refactorings in the future.

Also note that this does not yet update all functions to take `&Accessor`. Notably futures and streams still need to be updated.

cc #11224

### DIFF
--- a/crates/wit-bindgen/src/lib.rs
+++ b/crates/wit-bindgen/src/lib.rs
@@ -2723,13 +2723,12 @@ impl<'a> InterfaceGenerator<'a> {
         // that get to be optionally async/sync.
         let style = self.generator.opts.call_style();
         let (async_, async__, await_, concurrent) = match &style {
-            CallStyle::Async | CallStyle::Concurrent => {
-                if self.generator.opts.concurrent_exports {
-                    ("async", "_async", ".await", true)
-                } else {
-                    ("async", "_async", ".await", false)
-                }
-            }
+            CallStyle::Async | CallStyle::Concurrent => (
+                "async",
+                "_async",
+                ".await",
+                self.generator.opts.concurrent_exports,
+            ),
             CallStyle::Sync => ("", "", "", false),
         };
 
--- a/crates/wit-bindgen/src/lib.rs
+++ b/crates/wit-bindgen/src/lib.rs
@@ -2453,10 +2453,7 @@ impl<'a> InterfaceGenerator<'a> {
 
         let wt = self.generator.wasmtime_path();
         if let CallStyle::Concurrent = style {
-            uwrite!(
-                self.src,
-                "move |caller: &mut {wt}::component::Accessor::<T>, ("
-            );
+            uwrite!(self.src, "move |caller: &{wt}::component::Accessor::<T>, (");
         } else {
             uwrite!(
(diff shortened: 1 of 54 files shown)


alexcrichton:
One other update: `Instance::run_with` is updated to now take `AyncFnOnce` instead of `FnOnce() -> Box<...>`

crates/wit-bindgen/src/lib.rs
```diff
@@ -2728,7 +2725,7 @@ impl<'a> InterfaceGenerator<'a> {
         let (async_, async__, await_, concurrent) = match &style {
             CallStyle::Async | CallStyle::Concurrent => {
                 if self.generator.opts.concurrent_exports {
-                    ("", "INVALID", "INVALID", true)
+                    ("async", "_async", ".await", true)
```
dicej: Could collapse the `if` expression here into `("async", "_async", ".await", self.generator.opts.concurrent_exports)`

PR #11239, opened 2025-07-15
Implement Tarjan's strongly-connected components algorithm

fitzgen:
This commit implements [Tarjan's algorithm] for finding strongly-connected components.

This algorithm takes `O(V+E)` time and uses `O(V+E)` space.

Tarjan's algorithm is usually presented as a recursive algorithm, but we do not trust the input and cannot recurse over it for fear of blowing the stack. Therefore, this implementation is iterative.

This will be used to do bottom-up inlining in Wasmtime's compilation orchestration.

[Tarjan's algorithm]: https://en.wikipedia.org/wiki/Tarjan%27s_strongly_connected_components_algorithm

### DIFF
--- a/crates/wasmtime/src/compile/scc.rs
+++ b/crates/wasmtime/src/compile/scc.rs
@@ -46,6 +46,8 @@ where
         F: Fn(Node) -> S,
         S: Iterator<Item = Node>,
     {
+        let nodes = nodes.into_iter();
+
         // The resulting components and their nodes.
         let mut component_nodes = vec![];
         let mut components = PrimaryMap::<Component, Range<u32>>::new();
@@ -55,8 +57,10 @@ where
 
         // The DFS index and the earliest on-stack node reachable from each
         // node.
-        let mut indices = SecondaryMap::<Node, Option<NonMaxU32>>::new();
-        let mut lowlinks = SecondaryMap::<Node, Option<NonMaxU32>>::new();
+        let (min, max) = nodes.size_hint();
+        let capacity = max.unwrap_or_else(|| 2 * min);
+        let mut indices = SecondaryMap::<Node, Option<NonMaxU32>>::with_capacity(capacity);
+        let mut lowlinks = SecondaryMap::<Node, Option<NonMaxU32>>::with_capacity(capacity);
 
         // The stack of nodes we are currently finding an SCC for. Not the same
         // as the DFS stack: we only pop from this stack once we find the root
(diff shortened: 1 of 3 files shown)


alexcrichton:
I'm feeling a bit overwhelmed recently so I'm going to re-roll  review.

cfallin:
(I'm happy to cover this one as a "core compiler algorithms PR" unless you really want it, Andrew!)

abrown:
Go for it!

crates/wasmtime/src/compile/scc.rs
```diff
@@ -0,0 +1,528 @@
...
+        let mut component_nodes = vec![];
+        let mut components = PrimaryMap::<Component, Range<u32>>::new();
+
+        // The DFS index counter.
+        let mut index = NonMaxU32::default();
+
+        // The DFS index and the earliest on-stack node reachable from each
+        // node.
+        let mut indices = SecondaryMap::<Node, Option<NonMaxU32>>::new();
+        let mut lowlinks = SecondaryMap::<Node, Option<NonMaxU32>>::new();
```
cfallin: Just a little optimization thing but perhaps we could use a length hint (perhaps from `Iterator::size_hint`) to pre-allocate using `with_capacity`?

crates/wasmtime/src/compile/scc.rs
```diff
@@ -0,0 +1,528 @@
...
+
+    /// Iterate over each strongly-connnected component and the `Node`s that are
+    /// members of it.
+    ///
+    /// Iteration happens in reverse-topological order (successors are visited
+    /// before predecessors in the resulting SCC DAG).
+    pub fn iter(&self) -> impl ExactSizeIterator<Item = (Component, &[Node])> + '_ {
+        self.components.iter().map(|(component, range)| {
+            let start = usize::try_from(range.start).unwrap();
+            let end = usize::try_from(range.end).unwrap();
```
cfallin: Can we reuse `self.nodes(component)` here and below?

crates/wasmtime/src/compile/scc.rs
```diff
@@ -0,0 +1,528 @@
...
+        loop {
+            let event = self.stack.pop()?;
+
+            if let DfsEvent::Pre(node) = event {
+                if seen(node) {
+                    continue;
+                }
+
+                self.stack.push(DfsEvent::Post(node));
+                self.stack.extend(successors(node).flat_map(|v| {
```
cfallin: This is a neat iterator chain but I wonder if it would be clearer in imperative style in this case (vs. the chaining and nested if-expression returning an option) -- something like:

```rust
for successor in successors(node) {
  self.stack.push(DfsEvent::AfterEdge(node, successor));
  if !seen(successor) {
    self.stack.push(DfsEvent::Pre(successor));
  }
}
```

cfallin (approved):
Very nice! A few thoughts below but overall LGTM.

crates/wasmtime/src/compile/scc.rs
fitzgen: That technically implies an extra bounds check unless LLVM optimizes it away, but I can factor out the range conversion and indexing bits into a shared helper.

cfallin: Ah, good point! Not a huge deal, just noticed the duplication.

crates/wasmtime/src/compile/scc.rs
fitzgen: Yeah this is a reflection of how it was originally written and then modified as I found and fixed bugs. Should have moved it over to pushes already.

Issue #10766, opened 2025-05-11
Cranelift impressions from a first-time user


timstr:
Hello,

I tried using Cranelift for a new project JIT-compiling scripts for realtime audio processing. I've done similar work using [Inkwell](https://github.com/TheDan64/inkwell/) which wraps LLVM. I remain very interested in having a modern and competitive Rust-first JIT compiler, and I really want to see the Cranelift project succeed. While I was able to get get started quickly with Cranelift, there have been multiple separate and persistent issues which taken together have made me reconsider using Cranelift over Inkwell. I want to share my experiences and specific feedback because I think such first-time impressions are vital to the health of open source software and again, I love this project and want it to keep growing.

Apologies for putting so many things into a single issue. If it suits you, feel free to chop this up and quote verbatim or paraphrase into separate issues. For now, I'm moving on from Cranelift and returning to Inkwell and LLVM for my own productivity.

### Multiple distracting TODO's in the JIT demo
My early rapid success interacting with Cranelift was from following and adapting the [jit demo](https://github.com/bytecodealliance/cranelift-jit-demo/blob/e829a2a00b76627f0b46240ae4045b3a0f350eea/src/jit.rs). Reading through example code though, there are several distracting `TODO` comments questioning Cranelift's own design and pointing out known pain points. As a first-time user, this gives me the impression that the API hasn't matured and is probably unstable, which makes me hesitant to want to commit to it. What I would love to see here instead is just a self-contained demo where each part of the library plays an obvious role. Possibly some of this commentary is out of date and the demo can be refined already. 

### Scattered documentation due to crate separation
I rely heavily on documentation hosted on docs.rs where it is easily navigable and rapidly searchable. Oftentimes I search for individual types, names, and keywords to quickly and serendipitously discover related functionality. The separation of Cranelift into multiple crates however means that oftentimes, the thing I'm looking for isn't found because it's in a separate crate. For example, `FunctionBuilder` is defined in `cranelift-frontend`, but all of the instruction-inserting methods I use it with are defined in the `InstBuilder` trait in the `cranelift-codegen` crate, where I would have to know to look first. This may be more of a problem for Rust's own documentation system than with Cranelift per se.

### Difficulties with the first block of a function
[...]

PR #11241, opened 2025-07-15
Cranelift: Support block names

masnagam:
This PR implements a feature mentioned in #10766 and allows developers to set meaningful names to blocks.

Block names are used only in Cranelift **textual** IR representation.
`print!("{block}")` does not show the block name.

The implementaion is naive and allocates a `String` for each block name.
So, it's recommend to use block names only for debugging purposes at this point.
Someone may implement more memory efficient mechanism for block names (such as `Twine` in LLVM) in the future.

### DIFF
--- a/cranelift/codegen/src/ir/dfg.rs
+++ b/cranelift/codegen/src/ir/dfg.rs
@@ -21,6 +21,7 @@ use core::ops::{Index, IndexMut};
 use core::u16;
 
 use alloc::collections::BTreeMap;
+use alloc::string::String;
 #[cfg(feature = "enable-serde")]
 use serde_derive::{Deserialize, Serialize};
 use smallvec::SmallVec;
@@ -1460,6 +1461,24 @@ impl DataFlowGraph {
             }
         }
     }
+
+    /// Get the block name.
+    pub fn block_name(&self, block: Block) -> Option<&str> {
+        self.blocks[block].name.as_deref()
+    }
+
+    /// Set the block name which is used only in the textual IR representation.
+    ///
+    /// It's recommended to set the block name only for debugging purposes.
+    /// The current implementation is memory inefficient and allocates a [`String'] for each block name.
+    pub fn set_block_name(&mut self, block: Block, name: String) {
+        self.blocks[block].name = Some(name);
+    }
+
+    /// Clear the block name.
+    pub fn clear_block_name(&mut self, block: Block) {
+        self.blocks[block].name = None;
+    }
 }
 
 /// Contents of a basic block.
@@ -1472,12 +1491,16 @@ impl DataFlowGraph {
 pub struct BlockData {
     /// List of parameters to this block.
     params: ValueList,
+
+    /// Block name.
+    name: Option<String>,
 }
 
 impl BlockData {
     fn new() -> Self {
         Self {
             params: ValueList::new(),
+            name: None,
         }
     }
 
--- a/cranelift/codegen/src/write.rs
+++ b/cranelift/codegen/src/write.rs
@@ -234,7 +234,10 @@ pub fn write_block_header(
     };
 
     // The `indent` is the instruction indentation. block headers are 4 spaces out from that.
-    write!(w, "{1:0$}{2}", indent - 4, "", block)?;
+    match func.dfg.block_name(block) {
+        Some(block_name) => write!(w, "{1:0$}{2}", indent - 4, "", block_name)?,
+        None => write!(w, "{1:0$}{2}", indent - 4, "", block)?,
+    }
 
     let mut args = func.dfg.block_params(block).iter().cloned();
     match args.next() {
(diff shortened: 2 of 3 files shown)


cranelift/codegen/src/ir/dfg.rs
```diff
@@ -1472,12 +1491,16 @@ impl DataFlowGraph {
 pub struct BlockData {
     /// List of parameters to this block.
     params: ValueList,
+
+    /// Block name.
+    name: Option<String>,
```
bjorn3, line 1496: Putting this into a separate `SecondaryMap<Block, String>` would reduce memory usage when the block name functionality isn't used and remove unnecessary padding and improve cpu cache efficiency when it is used.

bjorn3: Also you can represent no name being given as an empty string.

cranelift/frontend/src/frontend.rs
```diff
@@ -353,6 +354,19 @@ impl<'a> FunctionBuilder<'a> {
         self.func.layout.set_cold(block);
     }
 
+    /// Set the block name which is used only in the textual IR representation.
+    ///
+    /// It's recommended to set the block name only for debugging purposes.
+    /// The current implementation is memory inefficient and allocates a [`String'] for each block name.
```
bjorn3, line 360: You need to use backticks both before and after.

cranelift/frontend/src/frontend.rs
```diff
@@ -353,6 +354,19 @@ impl<'a> FunctionBuilder<'a> {
         self.func.layout.set_cold(block);
     }
 
+    /// Set the block name which is used only in the textual IR representation.
+    ///
+    /// It's recommended to set the block name only for debugging purposes.
```
bjorn3, line 359: "It is" reads better to me.

cranelift/codegen/src/write.rs
```diff
@@ -234,7 +234,10 @@ pub fn write_block_header(
     };
 
     // The `indent` is the instruction indentation. block headers are 4 spaces out from that.
-    write!(w, "{1:0$}{2}", indent - 4, "", block)?;
+    match func.dfg.block_name(block) {
+        Some(block_name) => write!(w, "{1:0$}{2}", indent - 4, "", block_name)?,
+        None => write!(w, "{1:0$}{2}", indent - 4, "", block)?,
+    }
```
bjorn3, line 240: The block names should probably also be printed in for example `jump block1`. Also would be nice to support them in the parser too.

cfallin:
Thanks @masnagam for this PR!

Following on from @bjorn3's helpful comments, I think the main high-level thing that we would need to address to merge this is the round-tripping aspect of textual CLIF. Right now, this PR prints CLIF that can't be re-parsed (in addition to not using the user-provided names everywhere). I think we need to think about entity names from a syntax-design perspective first, come up with a consistent answer that represents all the needed information and can be round-tripped, then go from there.

A few considerations: entity numbering is deeply embedded in the way that the parser and the in-memory representation both work. It is what lets us have circular references without a separate name-resolution pass; it is what lets us avoid a symbol table. We'll want to re-think this: if entities can have names, then we need a symtab, we need a way to temporarily represent names during parsing and then resolve the references into numeric entity indices (allocating new indices for each name, and avoiding indices already used by the user), and we need to think about whether we want to print names *and* entity indices in the textual output. (Probably? Otherwise a roundtrip might not preserve them.)

We can sidestep a bunch of this if names are more like annotations and the entity indices are still the primary syntactic elements. For example, we could print comments alongside block definitions and references that give the names for those blocks. (As an aside, this is how I handle block names in [waffle](https://github.com/bytecodealliance/waffle/), another compiler library I designed that uses a Cranelift-like IR.) I'd personally be in favor of this -- it requires a little more care when writing textual CLIF (one still needs to come up with entity indices) but it keeps the language simpler and still allows for clarity when reading.

If we go that route, I think the needed changes are:
* Don't replace the entity name, but rather, print the string name as a comment at the end of the line
* Print a comment with a block name whenever a block is referenced in a terminator, jump table (used in `br_table` instructions), or exception table (used in `try_call` / `try_call_indirect` instructions)

as well as all the other review comments above. What do you think?

masnagam:
@bjorn3 @cfallin 
Thank you for your comments.
At this point, there are many issues in my PR that I should have to consider before sending the PR.
I'd like to close this PR after discussion w/ you and restart investigating Cranelift implementation in depth and reconsidering this feature based on your comments.
I'll create a new PR if I succeeded to implement this feature.
(Or ask someone to check my software design before implementing it).

Currently, I'm trying implementing [a JavaScript engine](https://github.com/bee-browser/bee/tree/main/libs/jsruntime/src/backend/clir/compiler) using Cranelift JIT.
And I want to assign meaningful block names (and variable names) for debugging purposes.
This is my motivation to want to implement this feature.

I need time to understand what you wrote because I don't have enough knowledge about Cranelift implementation at this point (as you may have noticed).
I'll come back here once I've finished investigating relevent code.

cranelift/codegen/src/write.rs
masnagam: oh, right. i completely forgot that...

masnagam:
I spent a few hours for investigating the Cranelift source tree and eventually understood what you wrote in the previous comment.
And I strongly feel the the direction of my change was not good.
Especially, I don't like to modify the lexer for clif files.
Probably, we can handle any block names in clif files by introducing a table for block names.
But the change increases complexity and regresses simplicity and maintainability (and throughputs slightly).

I chekced [waffle](https://github.com/bytecodealliance/waffle/) and got some hints from it.
Currently, I'm thinking of introducing a supplemental separate file to my JavaScript implementation, which contains a map between block entity IDs (`block*`) and the corresponding block names.
We can easily convert block entity IDs in Craneleft IR textual representation (and its CFG in the DOT format) to corresponding block names by using the map.
That's enough for debugging purposes.
Probably a similar map can be introduced for variable names.

Thank you for your kindly support :)
I close this PR.

PR #11244, opened 2025-07-15
fix(wasmtime): fix `--no-default-features` with async

rvolosatovs:
Currently, building `wasmtime` via:

```
$ cargo build -p wasmtime --no-default-features --features component-model-async
```

fails with:

```
error[E0432]: unresolved import `wasmtime_environ::fact`
  --> crates/wasmtime/src/runtime/component/concurrent.rs:88:36
   |
88 | use wasmtime_environ::{PrimaryMap, fact};
   |                                    ^^^^ no `fact` in the root
   |
note: found an item that was configured out
  --> /Users/rvolosatovs/src/github.com/bytecodealliance/wasmtime/crates/environ/src/lib.rs:72:9
   |
72 | pub mod fact;
   |         ^^^^
note: the item is gated here
  --> /Users/rvolosatovs/src/github.com/bytecodealliance/wasmtime/crates/environ/src/lib.rs:71:1
   |
71 | #[cfg(all(feature = "component-model", feature = "compile"))]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

in order to avoid the need for `component-model-async` feature to enable `wasmtime-environ/compile`, promote `START_FLAG_ASYNC_CALLEE` to crate-level const

### DIFF
--- a/crates/environ/src/lib.rs
+++ b/crates/environ/src/lib.rs
@@ -78,10 +78,3 @@ pub use cranelift_entity::*;
 
 /// Version number of this crate.
 pub const VERSION: &str = env!("CARGO_PKG_VERSION");
-
-/// Bit flag for indicating async-lifted exports
-///
-/// This flag may be passed to the `async-start` built-in function (which is
-/// called from both async->async and async->sync adapters) to indicate that the
-/// callee is an async-lifted export.
-pub const START_FLAG_ASYNC_CALLEE: i32 = 1 << 0;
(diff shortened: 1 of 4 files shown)


crates/environ/src/lib.rs
```diff
@@ -78,3 +78,10 @@ pub use cranelift_entity::*;
 
 /// Version number of this crate.
 pub const VERSION: &str = env!("CARGO_PKG_VERSION");
+
+/// Bit flag for indicating async-lifted exports
+///
+/// This flag may be passed to the `async-start` built-in function (which is
+/// called from both async->async and async->sync adapters) to indicate that the
+/// callee is an async-lifted export.
+pub const START_FLAG_ASYNC_CALLEE: i32 = 1 << 0;
```
alexcrichton: Mind moving this to `component.rs`? There's a few other constants in there as well so I think it'd fit well there.

alexcrichton (approved):
Thanks! Mind adding a check for this on CI as well? (e.g. the `micro_checks` job)

PR #11250, opened 2025-07-15
Require `Accessor` on all future/stream functions

alexcrichton:
This is a follow-up to #11238 which adds `&Accessor` arguments to all functions for futures and streams. Like #11238 this is done to make future refactorings easier for the internal implementation but the internal implementations are not updated at this time. Many functions, for example, do not use the argument at all just yet. The purpose of this is to ensure host usage of these functions always provides a store context.

This change required large refactorings of the upcoming wasmtime-wasi-http implementation in the wasip3-prototyping repository. That's all been sorted out now though so the changes are being pulled back here into the Wasmtime repository as well.

This commit additionally changes the `watch_*` functions on the various stream/future types to take `&mut self` instead of `self`-by-value. This is mostly a stylistic change and is more API-driven than anything else. Functionally this behaves the same as before where, while watching, the stream/future cannot be read/written to otherwise.

### DIFF
--- a/crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
+++ b/crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
@@ -7,7 +7,7 @@ use crate::component::concurrent::{ConcurrentState, tls};
 use crate::component::func::{self, LiftContext, LowerContext, Options};
 use crate::component::matching::InstanceType;
 use crate::component::values::{ErrorContextAny, FutureAny, StreamAny};
-use crate::component::{Instance, Lower, Val, WasmList, WasmStr};
+use crate::component::{AsAccessor, Instance, Lower, Val, WasmList, WasmStr};
 use crate::store::{StoreOpaque, StoreToken};
 use crate::vm::{VMFuncRef, VMMemoryDefinition, VMStore};
 use crate::{AsContextMut, StoreContextMut, ValRaw};
@@ -507,13 +507,18 @@ impl<T> FutureWriter<T> {
     /// value; otherwise it will return `false`, meaning the read end was dropped
     /// before the value could be delivered.
     ///
-    /// Note that the returned `Future` must be polled from the event loop of
-    /// the component instance from which this `FutureWriter` originated.  See
-    /// [`Instance::run`] for details.
-    pub fn write(mut self, value: T) -> impl Future<Output = bool> + Send + 'static
+    /// # Panics
+    ///
+    /// Panics if the store that the `accessor` is derived from does not own
+    /// this future.
+    pub async fn write(mut self, accessor: impl AsAccessor, value: T) -> bool
     where
         T: Send + 'static,
     {
+        // FIXME: this is intended to be used in the future to directly
+        // manipulate state for this future within the store without having to
+        // go through an mpsc.
+        let _accessor = accessor.as_accessor();
         let (tx, rx) = oneshot::channel();
         send(
             &mut self.tx.as_mut().unwrap(),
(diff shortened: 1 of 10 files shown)


alexcrichton:
Ah I also added a new `AsAccessor` helper trait here to assist with functions-taking-this-as-a-bound to make those a bit easier to write (in theory at least)

rvolosatovs (approved):

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -507,13 +507,18 @@ impl<T> FutureWriter<T> {
     /// value; otherwise it will return `false`, meaning the read end was dropped
     /// before the value could be delivered.
     ///
-    /// Note that the returned `Future` must be polled from the event loop of
-    /// the component instance from which this `FutureWriter` originated.  See
-    /// [`Instance::run`] for details.
-    pub fn write(mut self, value: T) -> impl Future<Output = bool> + Send + 'static
+    /// # Panics
+    ///
+    /// Panics if the store that the `accessor` is derived from does not own
```
dicej:
```suggestion
    /// Panics if the store that the [`Accessor`] is derived from does not own
```
... just to make it consistent with the corresponding doc comments below

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -903,54 +913,51 @@ impl<B> StreamWriter<B> {
...
-    /// Note that calling `Watch::into_inner` on the returned `Watch` will have
-    /// the side effect of causing the `Future` to resolve immediately if it
-    /// hasn't already.
+    /// # Panics
     ///
-    /// Also note that the returned `Future` must be polled from the event loop
-    /// of the component instance from which this `StreamWriter` originated.
-    /// See [`Instance::run`] for details.
-    pub fn watch_reader(mut self) -> (impl Future<Output = ()> + Send + 'static, Watch<Self>)
+    /// Panics if the store that the `accessor` is derived from does not own
```
dicej:
```suggestion
    /// Panics if the store that the [`Accessor`] is derived from does not own
```

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -1171,53 +1178,55 @@ impl<B> StreamReader<B> {
...
     /// either case, the returned buffer will be the same one passed as a
     /// parameter, with zero or more items added.
     ///
-    /// Note that the returned `Future` must be polled from the event loop of
-    /// the component instance from which this `StreamReader` originated.  See
-    /// [`Instance::run`] for details.
-    pub fn read(
+    /// # Panics
+    ///
+    /// Panics if the store that the `accessor` is derived from does not own
```
dicej:
```suggestion
    /// Panics if the store that the [`Accessor`] is derived from does not own
```

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -1171,53 +1178,55 @@ impl<B> StreamReader<B> {
...
-    /// Note that calling `Watch::into_inner` on the returned `Watch` will have
-    /// the side effect of causing the `Future` to resolve immediately if it
-    /// hasn't already.
+    /// # Panics
     ///
-    /// Also note that the returned `Future` must be polled from the event loop
-    /// of the component instance from which this `StreamReader` originated.
-    /// See [`Instance::run`] for details.
-    pub fn watch_writer(mut self) -> (impl Future<Output = ()> + Send + 'static, Watch<Self>)
+    /// Panics if the store that the `accessor` is derived from does not own
```
dicej:
```suggestion
    /// Panics if the store that the [`Accessor`] is derived from does not own
```

dicej (approved):

PR #11255, opened 2025-07-16
More table safety improvements

### LINKED ISSUE #11179: Management of tables in Wasmtime is excessively `unsafe`

alexcrichton:
This is some more progress on #11179 aimed at improving the safety of management of tables internally within Wasmtime:

* `Instance::table_index` is removed as it can be replaced with data stored directly in the `VMTableImport` now.
* `Instance::get_table` now returns `&mut Table`
* `Instance::get_defined_table_with_lazy_init` now returns `&mut Table`
* `Instance::with_defined_table_index_and_instance` now directly returns `DefinedTableIndex` plus `Pin<&mut Instance>`, codifying the ability to "laterally move" between instances.
* `Instance::table_init_segment` was refactored to "take" the tables during initialization and replace them afterwards, resolving the split borrow issue and removing an `unsafe` block in the function.

Closes #11179

### DIFF
--- a/cranelift/entity/src/primary.rs
+++ b/cranelift/entity/src/primary.rs
@@ -7,7 +7,6 @@ use alloc::boxed::Box;
 use alloc::vec::Vec;
 use core::fmt;
 use core::marker::PhantomData;
-use core::mem;
 use core::ops::{Index, IndexMut};
 use core::slice;
 #[cfg(feature = "enable-serde")]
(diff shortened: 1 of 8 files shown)


pchickey (approved):

alexcrichton:
The Miri violation of stacked borrows is I believe legitimate, so I've added a new commit which improves the safety situation around `Table::copy`. @pchickey mind reviewing that new commit as well?

alexcrichton:
I've also flagged this as closing https://github.com/bytecodealliance/wasmtime/issues/11179 as this resolves all remaining bits I know of there, and I otherwise split out https://github.com/bytecodealliance/wasmtime/issues/11262 from that issue.

PR #11257, opened 2025-07-16
feat(wasip3): implement `wasi:cli`

rvolosatovs:
follow-up on #11221, this implements `wasi:cli`, mostly by moving and cleaning up the implementation from https://github.com/bytecodealliance/wasip3-prototyping


crates/wasi/src/cli.rs
```diff
@@ -0,0 +1,142 @@
...
+    pub initial_cwd: Option<String>,
+    pub stdin: I,
+    pub stdout: O,
+    pub stderr: O,
+}
+
+pub trait IsTerminal {
+    /// Returns whether this stream is backed by a TTY.
+    fn is_terminal(&self) -> bool;
+}
```
alexcrichton, line 17: Would it be possible to use `std::io::IsTerminal` instead?

crates/wasi/src/clocks.rs
```diff
@@ -11,6 +11,12 @@ impl<T: WasiClocksView> WasiClocksView for &mut T {
     }
 }
 
+impl<T: WasiClocksView> WasiClocksView for Box<T> {
+    fn clocks(&mut self) -> &WasiClocksCtx {
+        (**self).clocks()
```
alexcrichton: FWIW a trick I've used in the past is to use `T::clocks(self)` in the body which forcibly ensures that recursion doesn't happen and avoids `**` syntax as well. Just a style thing though, no meaningful difference

crates/wasi/src/p3/cli/mod.rs
```diff
@@ -0,0 +1,263 @@
...
+struct WasiCli<T>(T);
+
+impl<T: 'static> HasData for WasiCli<T> {
+    type Data<'a> = WasiCliImpl<&'a mut T>;
+}
+
+pub struct TerminalInput;
+pub struct TerminalOutput;
+
+pub trait InputStream: IsTerminal {
```
alexcrichton: Could the `InputStream` and `OutputStream` traits directly have a `Send` supertrait to avoid needing `InputStream + Send` for example in the builder and such?

crates/wasi/src/p3/cli/mod.rs
```diff
@@ -0,0 +1,263 @@
...
+    fn reader(&self) -> Box<dyn AsyncRead + Send + Sync + Unpin>;
```
alexcrichton: Would it be possible to drop `Unpin` since this is already in a box anyway?

crates/wasi/src/cli.rs
alexcrichton: My guess is "no" given the litany of implementations we have, but in such a case could documentation be added that this is basically a mirror of what's in libstd? Also is it worth trying to add a blanket `impl<T: std::IsTerminal> wasi::IsTermainal for T` impl?

crates/wasi/src/cli.rs
```diff
@@ -0,0 +1,142 @@
...
+impl<T: ResourceView> ResourceView for WasiCliImpl<T> {
+    fn table(&mut self) -> &mut ResourceTable {
+        self.0.table()
+    }
+}
+
+impl<T: WasiCliView> WasiCliView for WasiCliImpl<T> {
+    type InputStream = T::InputStream;
+    type OutputStream = T::OutputStream;
+
```
alexcrichton: I'm a bit wary of making the implementation more generic than it already is, so would it be possible to drop these associated types and hardwire the trait objects? Or are these needed to multiplex the p2/p3 implementation?

crates/wasi/src/p3/view.rs
```diff
@@ -25,10 +37,20 @@ use crate::p3::ctx::WasiCtx;
...
+
+    #[doc(hidden)]
+    fn as_wasi_impl(&mut self) -> &mut WasiImpl<Self>
+    where
+        Self: Sized,
+    {
+        // TODO: Figure out how to avoid `unsafe`
+        // SAFETY: `WasiImpl` is `repr(transparent)`
+        unsafe { core::mem::transmute(self) }
+    }
```
alexcrichton: Could you explain a bit more as to why this is necessary?

I forget the matrix of types/traits involved here, but I would naively expect that `&mut WasiImpl<Self>` is basically equivalent to `WasiImpl<&mut Self>`

crates/wasi/src/cli.rs
rvolosatovs: We cannot use the `std::io::IsTerminal`, because it's sealed: https://github.com/rust-lang/rust/blob/f8f6997469237299c1d60814c7b9828602a1f8e4/library/std/src/io/stdio.rs#L1197

Unfortunately, largely because of that, blanket implementations for `T` also do not appear to be possible.

For example:
```
error[E0119]: conflicting implementations of trait `cli::IsTerminal` for type `tokio::io::Stderr`
   --> crates/wasi/src/cli.rs:126:1
    |
66  | impl<T: std::io::IsTerminal> IsTerminal for T {
    | --------------------------------------------- first implementation here
...
126 | impl IsTerminal for tokio::io::Stderr {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation for `tokio::io::Stderr`
    |
    = note: upstream crates may add a new impl of trait `std::sealed::Sealed` for type `tokio::io::Stderr` in future versions
    = note: upstream crates may add a new impl of trait `std::io::IsTerminal` for type `tokio::io::Stderr` in future versions
```

`tokio::io::Stderr` cannot implement `std::io::IsTerminal`, because it's sealed and *we* cannot implement `wasmtime_wasi::cli::IsTerminal` for it if we provide a blanket implementation for `std::io::IsTerminal`

crates/wasi/src/cli.rs
rvolosatovs: Depends, if these per-proposal "View" traits are adopted by `p2` implementation, then the associated types would be used for exactly this purpose. Currently, only the p3 implementation uses these traits.

I'll move this trait to `wasmtime_wasi::p3::cli::WasiCliView` without the associated types for now and maybe we can revisit once we have all of WASI implemented in this repo.

crates/wasi/src/p3/view.rs
rvolosatovs: I was hoping for that as well, but I've spent roughly 1h actively working on it and roughly an hour more trying things out while in meetings, but I could not figure it out without significant API changes, so I just gave up and went with the obvious solution requiring `unsafe`. Really, all we need here is to be able to "wrap" the value in the store.

What I'd like to get from this implementation:
- be able to select a proposal to link (`wasi:cli` in this case) and:
   - *only* implement parts relevant for that proposal (`WasiCliView` instead of `WasiView`)
   - add symbols to linker using a familiar API (`wasmtime_wasi::p3::cli::add_to_linker`) without having to resort to lower-level, generated `wasmtime_wasi::p3::bindings::wasi::cli::*` per-interface functions - which is error prone, tedious and introduces additional maintenance burden
- be able to link *all* of WASI, i.e. implement `WasiView` and call `wasmtime_wasi::p3::add_to_linker`

The convention is that we implement interfaces for `WasiCliImpl<T>`, where `T: WasiCliView`. The challenge then in the likes of: https://github.com/bytecodealliance/wasmtime/blob/90ab7916a996a38dc405b1ff861c7c137ed230a7/crates/wasi/src/p3/clocks/mod.rs#L74

We need a `WasiCliImpl(&mut T)`, where `T` implements `WasiCliView`, but I don't see a way to acquire such a `T`, when we start with `WasiView`. I can't do a `|v| &mut WasiImpl(v)`, since that would create an intermediate reference, which would be immediately dropped. I don't see a way to thread the lifetime through in a way that supports two different traits directly here.
We also currently have a separate trait for getting the resource table and we cannot call both `WasiCliView::cli` and `ResourceView::table` inside the closure, since that would make us mutably borrow twice

There are workarounds though:
1. Have `WasiCliView::cli` return a tuple `(&mut WasiCliCtx, &mut ResourceTable)`
2. Store `ResourceTable` in `WasiCliCtx`

This morning I actually got another idea, which I'm going to try now: `WasiCliCtxView<'a>`, which would have to `&mut` borrows, `&mut WasiCliCtx` and `&mut ResourceTable`, so `WasiCliView::cli` would have to return that. A bit awkward to use, but potentially it could work and hopefully it'd only be required in this lower level/advanced use case API

rvolosatovs: It worked! No `WasiCliImpl` and no need for generics in the host implementations anymore

rvolosatovs:
Semantics of `wasi:cli` stdio are not completely clear to me at this point, so I don't want to spend too much time designing the API for it, especially since this interface will likely change soon. There's currently no out-of-the-box way to "capture" stdout/stderr, users will have to implement their own with synchronization (e.g. using a `Mutex`) for now

refs https://github.com/WebAssembly/WASI/issues/757
refs https://github.com/WebAssembly/wasi-cli/issues/64

alexcrichton:
> There's currently no out-of-the-box way to "capture" stdout/stderr

Agreed yeah this is ok for now. For anything else along these lines mind opening an issue for that? I'll open an issue for this one specifically

alexcrichton:
Oh while cargo vet I think was spurious the [failure on Windows](https://github.com/bytecodealliance/wasmtime/actions/runs/16350124789/job/46194454343) I think is valid  which is [this line](https://github.com/bytecodealliance/wasmtime/pull/11257/files#diff-20f1fb49b0d9c0f977b5090e0b0b5df8d606c8200105e41431969a8dbfea8766R14). IIRC we set a few extra env vars for Windows for WASI testing

PR #11259, opened 2025-07-16
Rename `Instance::run_with` to `Instance::run_concurrent`

### LINKED ISSUE #11224: Rework `component-model-async` APIs to use `Accessor` where appropriate

alexcrichton:
This commit renames `run_with` to `run_concurrent` and additionally deletes the `Instance::run` method. With `Accessor`-related changes there's no need to have a "just run with a future" method as almost everything is going to require `run_concurrent` anyway to acquire an accessor.

Closes #11224

### DIFF
--- a/crates/wasmtime/src/runtime/component/func/typed.rs
+++ b/crates/wasmtime/src/runtime/component/func/typed.rs
@@ -232,7 +232,7 @@ where
             let result = concurrent::queue_call(wrapper.store.as_context_mut(), prepared)?;
             self.func
                 .instance
-                .run(wrapper.store.as_context_mut(), result)
+                .run_concurrent(wrapper.store.as_context_mut(), async |_| result.await)
                 .await?
         }
         #[cfg(not(feature = "component-model-async"))]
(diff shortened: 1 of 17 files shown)


crates/wasmtime/src/runtime/component/func/typed.rs
```diff
@@ -250,13 +250,12 @@ where
     /// made using this method may run concurrently with other calls to the same
     /// instance.  In addition, the runtime will call the `post-return` function
     /// (if any) automatically when the guest task completes -- no need to
-    /// explicitly call `Func::post_return` afterward.
+    /// explicitly cll `Func::post_return` afterward.
```
dicej:
```suggestion
    /// explicitly call `Func::post_return` afterward.
```

dicej (approved):

PR #11269, opened 2025-07-17
Stratification of call graphs for parallel bottom-up inlining

fitzgen:
This commit takes a call graph and constructs a strata, which is essentially a parallel execution plan. A strata consists of an ordered sequence of layers, and a layer of an unordered set of functions. The `i`th layer must be processed before the `i + 1`th layer, but functions within the same layer may be processed in any order (and in parallel).

For example, when given the following tree-like call graph:

    +---+   +---+   +---+
    | a |-->| b |-->| c |
    +---+   +---+   +---+
      |       |
      |       |     +---+
      |       '---->| d |
      |             +---+
      |
      |     +---+   +---+
      '---->| e |-->| f |
            +---+   +---+
              |
              |     +---+
              '---->| g |
                    +---+

then stratification will produce these layers:

    [
        {c, d, f, g},
        {b, e},
        {a},
    ]

Our goal in constructing the layers is to maximize potential parallelism at each layer. Logically, we do this by finding the strongly-connected components of the input call graph and peeling off all of the leaves of SCCs' condensation (i.e. the DAG that the SCCs form; see the documentation for the `StronglyConnectedComponents::evaporation` method for details). These leaves become the strata's first layer. The layer's components are removed from the condensation graph, and we repeat the process, so that the condensation's new leaves become the strata's second layer, and etc... until the condensation graph is empty and all components have been processed. In practice we don't actually mutate the condensation graph or remove its nodes but instead count how many unprocessed dependencies each component has, and a component is ready for inclusion in a layer once its unprocessed-dependencies count reaches zero.

This commit also renames the entity type for strongly-connected components from `Component` to `Scc`, as I felt the former was a bit ambiguous given Wasm components.

The next PR will extend Wasmtime's compilation driver code to actually make use of this new infrastructure.

### DIFF
--- a/crates/wasmtime/src/compile/call_graph.rs
+++ b/crates/wasmtime/src/compile/call_graph.rs
@@ -1,0 +1,103 @@
+//! Construction of the call-graph.
+
+#![cfg_attr(not(test), expect(dead_code, reason = "used in upcoming PRs"))]
+
+use super::*;
+use core::{
+    fmt::{self, Debug},
+    ops::Range,
+};
+use wasmtime_environ::{EntityRef, SecondaryMap};
+
+/// A call graph reified into a densely packed and quickly accessible
+/// representation.
+///
+/// In a call graph, nodes are functions, and an edge `f --> g` means that the
+/// function `f` calls the function `g`.
+pub struct CallGraph<Node>
+where
+    Node: EntityRef + Debug,
+{
+    /// A map from each node to the subslice of `self.edge_elems` that are its
+    /// edges.
+    edges: SecondaryMap<Node, Range<u32>>,
+
+    /// Densely packed edge elements for `self.edges`.
+    edge_elems: Vec<Node>,
+}
+
+impl<Node> Debug for CallGraph<Node>
+where
+    Node: EntityRef + Debug,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        struct Edges<'a, Node: EntityRef + Debug>(&'a CallGraph<Node>);
+
+        impl<'a, Node: EntityRef + Debug> Debug for Edges<'a, Node> {
+            fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+                f.debug_map()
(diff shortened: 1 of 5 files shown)


crates/wasmtime/src/compile/call_graph.rs
```diff
@@ -0,0 +1,103 @@
+//! Construction of the call-graph.
```
cfallin: We should clarify the scope and soundness/completeness of the call graph here -- in particular, with respect to imports and indirect calls (one could pick any of several set-points to get e.g. an overapproximation that's sound, or an underapproximation that gives partial but exact answers, and I think inlining wants the latter)

crates/wasmtime/src/compile/call_graph.rs
```diff
@@ -0,0 +1,103 @@
...
+        let mut calls = vec![];
+        for caller in funcs {
+            calls.clear();
+            get_calls(caller, &mut calls)?;
+
+            let start = edge_elems.len();
+            let start = u32::try_from(start).unwrap();
+            edge_elems.extend_from_slice(&calls);
+            let end = edge_elems.len();
+            let end = u32::try_from(end).unwrap();
```
cfallin: Idle thought -- we have this idiom a lot (list-of-lists stored in a flattened manner, with ranges in another list). It might be nice to build an abstraction around it -- something like

```rust
let mut elems = edge_elems.start_range();
elems.extend_from_slice(...); // or elems.push or whatever
edges[caller] = elems.finish();
```

not necessarily in this PR but as a cleanup at some point.

We do have `Ranges` which optimizes for the case that we build the ranges themselves in-order so we can store only one end of the range (and use previous-end as start); here we're processing callers in arbitrary order.

cfallin: (Actually, on reading further down and seeing more instances of this pattern -- I suspect it might be a nice cleanup in this PR!)

cfallin (approved):
This looks great! Very happy to see this coming together. The dependency-based scheduling into layers looks exactly like I'd expect.

One thought below on a potential abstraction opportunity but otherwise LGTM

PR #11272, opened 2025-07-17
x64: handle ISA features more completely

abrown:
This adds full boolean term support for instructions emitted in the new assembler (terms like `(_64b | compat) & avx2`). Despite doing more checks, this may be quicker too: instead of building up a `SmallVec<InstructionSet>` to compare against, this generates Rust code like the following that queries what features are available in the target via the `AvailableFeatures` trait:

```rust
#[must_use] // cranelift/assembler-x64/meta/src/generate/inst.rs:227
pub fn is_available(&self, features: &impl AvailableFeatures) -> bool {
    (features._64b() || features.compat()) && features.avx2() // cranelift/assembler-x64/meta/src/generate/inst.rs:232
}
```

With all of this in place, this PR has a large mechanical translation of all the old, incorrect feature definitions (`_64b | compat | avx2`) into their new, correct definitions (`(_64b | compat) & avx2`). I expect this will see a lot more use of this when using more instructions from AVX512, AVX10, APX, etc.


cranelift/codegen/src/isa/x64/inst/emit.rs
```diff
@@ -167,38 +167,10 @@ pub(crate) fn emit(
...
-
-    // Certain instructions may be present in more than one ISA feature set; we must at least match
-    // one of them in the target CPU.
-    let isa_requirements = inst.available_in_any_isa();
-    if !isa_requirements.is_empty() && !isa_requirements.iter().all(matches_isa_flags) {
-        panic!(
-            "Cannot emit inst '{inst:?}' for target; failed to match ISA requirements: {isa_requirements:?}"
-        )
+    if !inst.is_available(&info) {
+        panic!("Cannot emit inst '{inst:?}' for target; failed to match ISA requirements")
```
abrown: So this is a bit unfortunate at the moment: we've lost the ability to see why this check has failed. I mulled this over for a bit, thinking how we can get this back. One option is to have the assembler just "return the right string" through some other method which we can paste in here. But I'm leaning toward adding an `Inst::features() -> Features`, much like we had before, but that would return a boolean term like the one we have available at the meta level. Though this would duplicate the some of `Features`, it would return an `enum` that is more widely usable than a string while I still have all of this paged in.

Either option, "return a string" or "return a `Features`", is really fine at this point and I'm interested in feedback. Also, this seems like something that could get fixed up in a follow-on PR.

abrown:
I do expect this to fail in Winch: something about how we're using `-Ccranelift-has-avx2` in [`param_av2.wat`](https://github.com/bytecodealliance/wasmtime/blob/3c9305c0bb8d614b37f8d121aa115c2fd822296e/tests/disas/winch/x64/i8x16/splat/param_avx2.wat) doesn't seem to propagate the right features into the ISA flags used here:

https://github.com/bytecodealliance/wasmtime/blob/3c9305c0bb8d614b37f8d121aa115c2fd822296e/winch/codegen/src/isa/x64/masm.rs#L3205-L3208

cranelift/codegen/src/isa/x64/inst/emit.rs
alexcrichton: `Inst::features` sounds reasonable to me. Implementation-wise I might recommend a different representation of `Features` that doesn't use `Box` internally but instead uses `&'static ...` and put everything into `const`s. That way there's no need to actually allocate anything and it's all just pointing at a bunch of static data (which is in theory deduplicated across functions too)

cranelift/codegen/src/isa/x64/inst/mod.rs
```diff
@@ -1551,6 +1526,85 @@ impl EmitInfo {
...
+        self.isa_flags.use_popcnt()
+    }
+
+    fn avx(&self) -> bool {
+        self.isa_flags.use_avx()
+    }
+
+    fn avx2(&self) -> bool {
+        self.isa_flags.use_avx2()
+    }
```
alexcrichton: I think the discrepancy here with Winch is `use_avx2` vs `has_avx2` perhaps? Here I think it'd be reasonable to switch to `has_*` (for other methods too) where `use_*` affects codegen but `has_*` is the literal CPU features (in theory)

abrown: See de476c2.

cranelift/codegen/src/isa/x64/inst/emit.rs
abrown: See f529126.

alexcrichton (approved):

PR #11283, opened 2025-07-18
Wasmtime: Add (optional) bottom-up function inlining to Wasm compilation

fitzgen:
This commit plumbs together two pieces of recently-added infrastructure:

1. function inlining in Cranelift, and
2. the parallel bottom-up inlining scheduler in Wasmtime.

Sprinkle some very simple inlining heuristics on top, and this gives us function
inlining in Wasm compilation.

The default Wasmtime configuration does not enable inlining, and when we do
enable it, we only enable it for cross-component calls by default (since
presumably the toolchain that produced a particular core Wasm module, like LLVM,
already performed any inlining that was beneficial within that module, but that
toolchain couldn't know how that Wasm module would be getting linked together
with other modules via component composition, and so it could not have done any
cross-component inlining). For what it is worth, there is a config knob to
enable intra-module function inlining, but this is primarily for use by our
fuzzers, so that they can easily excercise and explore this new inlining
functionality.

All this plumbing required some changes to the `wasmtime_environ::Compiler`
trait, since Winch cannot do inlining but Cranelift can. This is mostly
encapsulated in the new `wasmtime_environ::InliningCompiler` trait, for the most
part. Additionally, we take care not to construct the call graph, or any other
data structures required only by the inliner and not regular compilation, both
when using Winch and when using Cranelift with inlining disabled.

Finally, we add a `disas` test to verify that we successfully inline a series of
calls from a function in one component, to a cross-component adapter function,
to a function in another component. Most test coverage is expected to come from
our fuzzing, however.

Depends on https://github.com/bytecodealliance/wasmtime/pull/11282

### DIFF
--- a/crates/cli-flags/src/lib.rs
+++ b/crates/cli-flags/src/lib.rs
@@ -111,6 +111,27 @@ wasmtime_option_group! {
         /// dense.
         pub memory_guaranteed_dense_image_size: Option<u64>,
 
+        /// Whether to perform function inlining during compilation.
+        pub compiler_inlining: Option<bool>,
+
+        /// Whether to perform function inlining within the same core Wasm
+        /// module or only for calls from one module into a different
+        /// module. Only exposed for fuzzing.
+        #[doc(hidden)]
+        #[serde(default)]
+        #[serde(deserialize_with = "crate::opt::cli_parse_wrapper")]
+        pub compiler_inlining_intra_module: Option<wasmtime::IntraModuleInlining>,
+
+        /// Configure the "small-callee" size for function inlining
+        /// heuristics. Only exposed for fuzzing.
+        #[doc(hidden)]
+        pub compiler_inlining_small_callee_size: Option<u32>,
+
+        /// Configure the "sum size threshold" for function inlining
+        /// heuristics. Only exposed for fuzzing.
+        #[doc(hidden)]
+        pub compiler_inlining_sum_size_threshold: Option<u32>,
+
         /// The maximum number of WebAssembly instances which can be created
         /// with the pooling allocator.
         pub pooling_total_core_instances: Option<u32>,
@@ -815,6 +836,18 @@ impl CommonOptions {
         }
         if let Some(size) = self.opts.memory_guaranteed_dense_image_size {
             config.memory_guaranteed_dense_image_size(size);
+        }
+        if let Some(enable) = self.opts.compiler_inlining {
+            config.compiler_inlining(enable);
+        }
+        if let Some(cfg) = self.opts.compiler_inlining_intra_module {
+            config.compiler_inlining_intra_module(cfg);
+        }
+        if let Some(size) = self.opts.compiler_inlining_small_callee_size {
+            config.compiler_inlining_small_callee_size(size);
+        }
+        if let Some(size) = self.opts.compiler_inlining_sum_size_threshold {
+            config.compiler_inlining_sum_size_threshold(size);
         }
         if let Some(enable) = self.opts.signals_based_traps {
             config.signals_based_traps(enable);
--- a/crates/cli-flags/src/lib.rs
+++ b/crates/cli-flags/src/lib.rs
@@ -240,14 +240,6 @@ wasmtime_option_group! {
         /// Whether to perform function inlining during compilation.
         pub inlining: Option<bool>,
 
-        /// Whether to perform function inlining within the same core Wasm
-        /// module or only for calls from one module into a different
-        /// module. Only exposed for fuzzing.
-        #[doc(hidden)]
-        #[serde(default)]
-        #[serde(deserialize_with = "crate::opt::cli_parse_wrapper")]
-        pub inlining_intra_module: Option<wasmtime::IntraModuleInlining>,
-
         /// Configure the "small-callee" size for function inlining
         /// heuristics. Only exposed for fuzzing.
         #[doc(hidden)]
@@ -845,15 +837,6 @@ impl CommonOptions {
         }
         if let Some(enable) = self.codegen.inlining {
             config.compiler_inlining(enable);
-        }
-        if let Some(cfg) = self.codegen.inlining_intra_module {
-            config.compiler_inlining_intra_module(cfg);
-        }
-        if let Some(size) = self.codegen.inlining_small_callee_size {
-            config.compiler_inlining_small_callee_size(size);
-        }
-        if let Some(size) = self.codegen.inlining_sum_size_threshold {
-            config.compiler_inlining_sum_size_threshold(size);
         }
 
         // async_stack_size enabled by either async or stack-switching, so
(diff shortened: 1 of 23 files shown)


crates/cli-flags/src/lib.rs
```diff
@@ -111,6 +111,27 @@ wasmtime_option_group! {
...
+
+        /// Configure the "small-callee" size for function inlining
+        /// heuristics. Only exposed for fuzzing.
+        #[doc(hidden)]
+        pub compiler_inlining_small_callee_size: Option<u32>,
+
+        /// Configure the "sum size threshold" for function inlining
+        /// heuristics. Only exposed for fuzzing.
+        #[doc(hidden)]
+        pub compiler_inlining_sum_size_threshold: Option<u32>,
```
alexcrichton: To bikeshed a bit, what about `-C` instead and dropping the `compiler_*` prefix?

crates/cli-flags/src/lib.rs
```diff
@@ -111,6 +111,27 @@ wasmtime_option_group! {
...
+        /// module or only for calls from one module into a different
+        /// module. Only exposed for fuzzing.
+        #[doc(hidden)]
+        #[serde(default)]
+        #[serde(deserialize_with = "crate::opt::cli_parse_wrapper")]
+        pub compiler_inlining_intra_module: Option<wasmtime::IntraModuleInlining>,
+
+        /// Configure the "small-callee" size for function inlining
+        /// heuristics. Only exposed for fuzzing.
+        #[doc(hidden)]
```
alexcrichton: How come these are doc(hidden)?

crates/cranelift/src/compiler.rs
```diff
@@ -671,6 +692,126 @@ impl wasmtime_environ::Compiler for Compiler {
...
+        let func = &cx.codegen_context.func;
+        calls.extend(
+            func.params
+                .user_named_funcs()
+                .values()
+                .filter(|name| name.namespace == crate::NS_WASM_FUNC)
+                .map(|name| FuncIndex::from_u32(name.index)),
+        );
+        Ok(())
+    }
```
alexcrichton: Should this extend into a set instead of a vec?

crates/wasmtime/src/compile.rs
```diff
@@ -643,6 +725,326 @@ the use case.
...
+        caller_size: u32,
+        caller_module: StaticModuleIndex,
+        caller_def_func: DefinedFuncIndex,
+        caller_needs_gc_heap: bool,
+        callee_size: u32,
+        callee_module: StaticModuleIndex,
+        callee_def_func: DefinedFuncIndex,
+        callee_needs_gc_heap: bool,
+    ) -> bool {
+        log::trace!(
```
alexcrichton: Given the quantity of parameters here and how this is probably going to get tuned over time, how about packaging this all up in `struct InlineDecisionParams { .. }` or something like that? That way there's named keys and no need to worry about getting things in the wrong order by accident.

crates/wasmtime/src/config.rs
```diff
@@ -2027,6 +2027,52 @@ impl Config {
...
+    }
+
+    /// Set the "sum size threshold" for the function inlining heuristic.
+    ///
+    /// Only exposed for fuzzing.
+    #[doc(hidden)]
+    pub fn compiler_inlining_sum_size_threshold(&mut self, size: u32) -> &mut Self {
+        self.tunables.inlining_sum_size_threshold = Some(size);
+        self
+    }
```
alexcrichton: An alternative to this perhaps, the `LinkOptions::padding_between_functions` field is a "compiler setting" but layered on top of Cranelift sort of. These sorts of options might make sense there too? That might require pushing more deduction of whether to inline into the `InliningCompiler` trait but that seems reasonable. That way we get `-Ccranelift-...` options for everything and don't have to have custom `Config` methods.

crates/wasmtime/src/engine/serialization.rs
```diff
@@ -355,6 +360,18 @@ impl Metadata<'_> {
...
+            inlining_small_callee_size,
+            other.inlining_small_callee_size,
+            "function inlining small-callee size",
+        )?;
+        Self::check_int(
+            inlining_sum_size_threshold,
+            other.inlining_sum_size_threshold,
+            "function inlining sum-size threshold",
+        )?;
+        Self::check_intra_module_inlining(inlining_intra_module, other.inlining_intra_module)?;
```
alexcrichton, line 374: These can probably all be ignored, right? Regardless of what the host and artifact have configured the artifact should still work the same way? (e.g. inlining doesn't affect abi)

crates/cranelift/src/compiler.rs
fitzgen: That would require sorting just to be safe w.r.t. non-determinism and also propagating a dependency on hashbrown up to wasmtime-environ. Neither is really a problem, but they are both a little annoying. Pair this with the fact that the deduplication shouldn't matter, since we are creating call-graph edges based on the imported funcs arena and not based on call instructions, and we deduplicate the former during Wasm-to-CLIF construction, I don't think it is worth switching this from a vec to a set.

crates/wasmtime/src/engine/serialization.rs
fitzgen: Yeah, this is true but I figured the safest, most conservative thing to do here was to require them to be identical, so that's what I did out of an abundance of caution. If you feel strongly, I can remove these checks and ignore these options.

crates/cranelift/src/compiler.rs
alexcrichton: I think [`IndexSet` in wasmtime-environ](https://github.com/bytecodealliance/wasmtime/blob/1155d6dfbfe859cad85cfe0d5700d4acefd04745/crates/environ/src/prelude.rs#L28) would solve that? Basically I'm worried about big-O factors here of making a very large list of effectively all callsites.

fitzgen: > Basically I'm worried about big-O factors here of making a very large list of effectively all callsites.

We already dedupe imported functions when building the CLIF so we don't need to worry about the big-O factors here. But it would be more future proof, I suppose.

fitzgen:
@alexcrichton okay, this should be ready for a second look

crates/cli-flags/src/lib.rs
```diff
@@ -237,6 +237,19 @@ wasmtime_option_group! {
...
+        pub inlining_small_callee_size: Option<u32>,
+
+        /// Configure the "sum size threshold" for function inlining
+        /// heuristics. Only exposed for fuzzing.
+        #[doc(hidden)]
+        pub inlining_sum_size_threshold: Option<u32>,
```
alexcrichton: I think these can be removed now?

fitzgen: Good eye, thanks

PR #11284, opened 2025-07-18
Record and Replay Support for Component Model

arjunr2:
## Brief

This PR is intended to support deterministic record and replay (RR) of Wasm components intrinsically in Wasmtime, and received an initial round of discussion in the Wasmtime bi-weekly meeting on 07/17

## Motivation

RR is a very useful primitive for improving the debugging story of Wasm in Wasmtime. Bugs that are often encountered in modules during deployment can be deterministically reproduced.  In particular, it provides the foundation for the following (to name a few):

- Reverse-execution (a.k.a, time-travel) debugging
- Offline static/dynamic analysis of prior module executions
- Profiling of module/runtime components
- Automatic extraction of differential unit-tests for system interfaces (e.g. WASI)
- Interposition points for targeted fuzzing of system interfaces and/or modules

## Scope 

This initial PR provides the base primitives for recording and replay events. It supports RR at all import function boundaries and lowering rules for component types. The RR event infrastructure is intended to be easily extensible to new event types as new use-cases emerge.

### Primary Goals

- Enabling low overhead (memory, compute, and trace size) recording and high-performance replay,
- Full determinism during replay that can run outside the embedder context.
- An engine-agnostic trace recording format -- the goal is to purely capture guest-host boundary crossings (import calls and component model interactions) that can be reasonably interpreted by another component model compliant engine. 

### Non-Goals (Subject to Discussion)

- A human readable trace format. This belong better in something like wit-bindgen, and/or as an independent tool over the low-level trace
- Replay support for updated versions of a recorded module -- this requires a much more coordinated effort from the producers as well to make this practically useful.

## Initial Performance Numbers

Some initial runs on compression libraries like `zstd` show a 4-5% overhead on recording logic, excluding the disk I/O. This seems reasonable at the moment and likely doesn't need further optimization unless there are explicit use-cases. 

## Minor Todo

The following (minor) additions will be made in the coming days prior to potential merging:
- Encoding hashes of modules in the recorded trace for validation 
- Generic writers/readers for `RecordBuffer` and `ReplayBuffer`
- Feature gating all of RR

## Questions for Maintainers
- Do we get wasip1 for free by recording/replay at this level?
- What's typically the most idiomatic way to serialize anyhow:Errors?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2645,6 +2645,7 @@ checksum = "1fa7e49bd669d32d7bc2a15ec540a527e7764aec722a45467814005725bcd721"
 dependencies = [
  "ort-sys",
  "smallvec 2.0.0-alpha.10",
+ "tracing",
 ]
 
 [[package]]
--- a/crates/wasi-nn/Cargo.toml
+++ b/crates/wasi-nn/Cargo.toml
@@ -70,6 +70,10 @@ openvino = ["dep:openvino"]
 onnx = ["dep:ort"]
 # Use prebuilt ONNX Runtime binaries from ort.
 onnx-download = ["onnx", "ort/download-binaries"]
+# CUDA execution provider for NVIDIA GPU support (requires CUDA toolkit)
+onnx-cuda = ["onnx", "ort/cuda"]
+# Enable tracing for ONNX Runtime
+ort-tracing = ["onnx", "ort/tracing"]
 # WinML is only available on Windows 10 1809 and later.
 winml = ["dep:windows"]
 # PyTorch is available on all platforms; requires Libtorch to be installed
--- a/crates/wasi-nn/examples/classification-component-onnx/README.md
+++ b/crates/wasi-nn/examples/classification-component-onnx/README.md
@@ -3,35 +3,111 @@
 This example demonstrates how to use the `wasi-nn` crate to run a classification using the
 [ONNX Runtime](https://onnxruntime.ai/) backend from a WebAssembly component.
 
+It supports CPU and GPU (Nvidia CUDA) execution targets.
+
+**Note:**
+GPU execution target only supports Nvidia CUDA (onnx-cuda) as execution provider (EP) for now.
+
 ## Build
+
 In this directory, run the following command to build the WebAssembly component:
 ```console
+# build component for target wasm32-wasip1
 cargo component build
+
+# build component for target wasm32-wasip2
+cargo component build --target wasm32-wasip2
 ```
 
-In the Wasmtime root directory, run the following command to build the Wasmtime CLI and run the WebAssembly component:
-```sh
-# build wasmtime with component-model and WASI-NN with ONNX runtime support
-cargo build --features component-model,wasi-nn,wasmtime-wasi-nn/onnx-download
+## Running the Example
 
-# run the component with wasmtime
+In the Wasmtime root directory, run the following command to build the Wasmtime CLI and run the WebAssembly component:
+
+### Building Wasmtime
+
+#### For CPU-only execution:
+```sh
+cargo build --features component-model,wasi-nn,wasmtime-wasi-nn/onnx-download
+```
+
+#### For GPU (Nvidia CUDA) support:
+```sh
+cargo build --features component-model,wasi-nn,wasmtime-wasi-nn/onnx-cuda,wasmtime-wasi-nn/onnx-download
+```
+
+### Running with Different Execution Targets
+
+The execution target is controlled by passing a single argument to the WASM module.
+
+Arguments:
+- No argument or `cpu` - Use CPU execution
+- `gpu` or `cuda` - Use GPU/CUDA execution
+
+#### CPU Execution (default):
+```sh
 ./target/debug/wasmtime run \
     -Snn \
     --dir ./crates/wasi-nn/examples/classification-component-onnx/fixture/::fixture \
-    ./crates/wasi-nn/examples/classification-component-onnx/target/wasm32-wasip1/debug/classification-component-onnx.wasm
+    ./crates/wasi-nn/examples/classification-component-onnx/target/wasm32-wasip2/debug/classification-component-onnx.wasm
 ```
 
-You should get the following output:
+#### GPU (CUDA) Execution:
+```sh
+# path to `libonnxruntime_providers_cuda.so` downloaded by `ort-sys`
+export LD_LIBRARY_PATH={wasmtime_workspace}/target/debug
+
+./target/debug/wasmtime run \
+    -Snn \
+    --dir ./crates/wasi-nn/examples/classification-component-onnx/fixture/::fixture \
+    ./crates/wasi-nn/examples/classification-component-onnx/target/wasm32-wasip2/debug/classification-component-onnx.wasm \
+    gpu
+
+```
+
+## Expected Output
+
+You should get output similar to:
 ```txt
+No execution target specified, defaulting to CPU
 Read ONNX model, size in bytes: 4956208
-Loaded graph into wasi-nn
+Loaded graph into wasi-nn with Cpu target
 Created wasi-nn execution context.
 Read ONNX Labels, # of labels: 1000
-Set input tensor
 Executed graph inference
-Getting inferencing output
 Retrieved output data with length: 4000
 Index: n02099601 golden retriever - Probability: 0.9948673
 Index: n02088094 Afghan hound, Afghan - Probability: 0.002528982
 Index: n02102318 cocker spaniel, English cocker spaniel, cocker - Probability: 0.0010986356
 ```
+
+When using GPU target, the first line will indicate the selected execution target.
+You can monitor GPU usage using cmd `watch -n 1 nvidia-smi`.
+
+To see trace logs from `wasmtime_wasi_nn` or `ort`, run Wasmtime with `WASMTIME_LOG` enabled, e.g.,
+
+```sh
+WASMTIME_LOG=wasmtime_wasi_nn=warn ./target/debug/wasmtime run ...
+WASMTIME_LOG=ort=warn ./target/debug/wasmtime run ...
+```
+
+## Prerequisites for GPU(CUDA) Support
+- NVIDIA GPU with CUDA support
+- CUDA Toolkit 12.x with cuDNN 9.x
+- Build wasmtime with `wasmtime-wasi-nn/onnx-cuda` feature
+
+## ONNX Runtime's Fallback Behavior
+
+If the GPU execution provider is requested (by passing `gpu`) but the device does not have a GPU or the necessary CUDA drivers are missing, ONNX Runtime will **silently fall back** to the CPU execution provider. The application will continue to run, but inference will happen on the CPU.
+
+To verify if fallback is happening, you can enable ONNX Runtime logging:
+
+1. Build Wasmtime with the additional `wasmtime-wasi-nn/ort-tracing` feature:
+   ```sh
+   cargo build --features component-model,wasi-nn,wasmtime-wasi-nn/onnx-cuda,wasmtime-wasi-nn/ort-tracing
+   ```
+
+2. Run Wasmtime with `WASMTIME_LOG` enabled to see `ort` warnings:
+   ```sh
+   WASMTIME_LOG=ort=warn ./target/debug/wasmtime run ...
+   ```
+   You should see a warning like: `No execution providers from session options registered successfully; may fall back to CPU.`
--- a/crates/wasi-nn/examples/classification-component-onnx/src/main.rs
+++ b/crates/wasi-nn/examples/classification-component-onnx/src/main.rs
@@ -17,14 +17,46 @@ use self::wasi::nn::{
     tensor::{Tensor, TensorData, TensorDimensions, TensorType},
 };
 
+/// Determine execution target from command-line argument
+/// Usage: wasm_module [cpu|gpu|cuda]
+fn get_execution_target() -> ExecutionTarget {
+    let args: Vec<String> = std::env::args().collect();
+
+    // First argument (index 0) is the program name, second (index 1) is the target
+    // Ignore any arguments after index 1
+    if args.len() >= 2 {
+        match args[1].to_lowercase().as_str() {
+            "gpu" | "cuda" => {
+                println!("Using GPU (CUDA) execution target from argument");
+                return ExecutionTarget::Gpu;
+            }
+            "cpu" => {
+                println!("Using CPU execution target from argument");
+                return ExecutionTarget::Cpu;
+            }
+            _ => {
+                println!("Unknown execution target '{}', defaulting to CPU", args[1]);
+            }
+        }
+    } else {
+        println!("No execution target specified, defaulting to CPU");
+        println!("Usage: <program> [cpu|gpu|cuda]");
+    }
+
+    ExecutionTarget::Cpu
+}
+
 fn main() {
     // Load the ONNX model - SqueezeNet 1.1-7
     // Full details: https://github.com/onnx/models/tree/main/vision/classification/squeezenet
     let model: GraphBuilder = fs::read("fixture/models/squeezenet1.1-7.onnx").unwrap();
     println!("Read ONNX model, size in bytes: {}", model.len());
 
-    let graph = load(&[model], GraphEncoding::Onnx, ExecutionTarget::Cpu).unwrap();
-    println!("Loaded graph into wasi-nn");
+    // Determine execution target
+    let execution_target = get_execution_target();
+
+    let graph = load(&[model], GraphEncoding::Onnx, execution_target).unwrap();
+    println!("Loaded graph into wasi-nn with {:?} target", execution_target);
 
     let exec_context = Graph::init_execution_context(&graph).unwrap();
     println!("Created wasi-nn execution context.");
--- a/crates/wasi-nn/src/backend/onnx.rs
+++ b/crates/wasi-nn/src/backend/onnx.rs
@@ -7,12 +7,17 @@ use crate::backend::{Id, read};
 use crate::wit::types::{ExecutionTarget, GraphEncoding, Tensor, TensorType};
 use crate::{ExecutionContext, Graph};
 use ort::{
+    execution_providers::{CPUExecutionProvider, ExecutionProviderDispatch},
     inputs,
     session::{Input, Output},
     session::{Session, SessionInputValue, builder::GraphOptimizationLevel},
     tensor::TensorElementType,
     value::{Tensor as OrtTensor, ValueType},
 };
+
+#[cfg(feature = "onnx-cuda")]
+use ort::execution_providers::CUDAExecutionProvider;
+
 use std::path::Path;
 use std::sync::{Arc, Mutex};
 
@@ -31,7 +36,11 @@ impl BackendInner for OnnxBackend {
             return Err(BackendError::InvalidNumberOfBuilders(1, builders.len()));
         }
 
+        // Configure execution providers based on target
+        let execution_providers = configure_execution_providers(target)?;
+
         let session = Session::builder()?
+            .with_execution_providers(execution_providers)?
             .with_optimization_level(GraphOptimizationLevel::Level3)?
             .commit_from_memory(builders[0])?;
 
@@ -42,6 +51,38 @@ impl BackendInner for OnnxBackend {
 
     fn as_dir_loadable<'a>(&'a mut self) -> Option<&'a mut dyn BackendFromDir> {
         Some(self)
+    }
+}
+
+/// Configure execution providers based on the target
+fn configure_execution_providers(
+    target: ExecutionTarget,
+) -> Result<Vec<ExecutionProviderDispatch>, BackendError> {
+    match target {
+        ExecutionTarget::Cpu => {
+            // Use CPU execution provider with default configuration
+            tracing::debug!("Using CPU execution provider");
+            Ok(vec![CPUExecutionProvider::default().build()])
+        }
+        ExecutionTarget::Gpu => {
+            #[cfg(feature = "onnx-cuda")]
+            {
+                // Use CUDA execution provider for GPU acceleration
+                tracing::debug!("Using Nvidia GPU CUDA execution provider");
+                Ok(vec![CUDAExecutionProvider::default().build()])
+            }
+            #[cfg(not(feature = "onnx-cuda"))]
+            {
+                tracing::warn!("GPU CUDA execution provider is not enabled, falling back to CPU");
+                Ok(vec![CPUExecutionProvider::default().build()])
+            }
+        }
+        ExecutionTarget::Tpu => {
+            tracing::warn!(
+                "TPU execution target is not supported for ONNX backend yet, falling back to CPU"
+            );
+            Ok(vec![CPUExecutionProvider::default().build()])
+        }
     }
 }
 


crates/cli-flags/src/lib.rs
```diff
@@ -528,6 +561,29 @@ pub struct CommonOptions {
...
+    /// Options to enable and configure execution replay, `-P help` to see all.
+    ///
+    /// Run a determinstic, embedding-agnostic replay execution of the Wasm module
+    /// according to a prior recorded execution trace (see the `--record` option).
+    ///
+    /// Note: Minimal configs for deterministic Wasm semantics will be
+    /// enforced during replay by default (NaN canonicalization, deterministic relaxed SIMD)
+    #[arg(short = 'P', long = "replay", value_name = "KEY[=VAL[,..]]")]
+    #[serde(skip)]
+    replay_raw: Vec<opt::CommaSeparated<Replay>>,
```
cfallin: High-level thoughts on command-line options:

- It makes sense to me that record is a *general* option that the user should be able to set on any Wasmtime subcommand that executes WebAssembly; this includes `wasmtime run`, `wasmtime serve`, and maybe others. (Curious: have you tried it on an HTTP server component?)
- However, replay is a *specific and separate* thing: for example, if I am replaying an HTTP server execution, I should not be spawning an actual HTTP server on the host side. Basically, the idea of replay is that it is replacing the "real" host with a trace, so it should be properly seen (I think) as a separate top-level driver.

With that in mind, what do you think about turning replay into a command, rather than an option?

crates/cli-flags/src/lib.rs
```diff
@@ -970,6 +1042,27 @@ impl CommonOptions {
...
+        let record = &self.record;
+        let replay = &self.replay;
+        let rr_cfg = if let Some(path) = record.path.clone() {
+            Some(RRConfig::from(RecordConfig {
+                writer_initializer: Arc::new(move || Box::new(fs::File::create(&path).unwrap())),
+                metadata: RecordMetadata {
+                    add_validation: record.validation_metadata.unwrap_or(true),
+                },
+            }))
+        } else if let Some(path) = replay.path.clone() {
```
cfallin: If the above refactor (replay as a separate subcommand, without an option to itself record) doesn't pan out, separately we should always error if conflicting options are set here (that is, both record and replay together).

crates/wasmtime/Cargo.toml
```diff
@@ -45,7 +45,7 @@ wat = { workspace = true, optional = true }
 serde = { workspace = true }
 serde_derive = { workspace = true }
 serde_json = { workspace = true, optional = true }
-postcard = { workspace = true }
+postcard = { workspace = true, features = ["use-std"] }
```
cfallin: Does this break our no-std build? If we need std for something we should put the feature down in the feature dependencies below (`std = [..., "postcard/use-std"]`).

cfallin: Also, in general, we should make this `optional = true` and then enable the postcard dep itself when rr feature(s) are turned on below.

crates/wasmtime/Cargo.toml
```diff
@@ -393,3 +393,20 @@ component-model-async = [
   "wasmtime-component-macro?/component-model-async",
   "dep:futures"
 ]
+
+# Enables support for record/replay
+rr = ["rr-component", "rr-core"]
+
+# Component model RR support
+rr-component = ["component-model", "std"]
```
cfallin: I'd be curious which parts of our implementation depend on `std` -- I would suspect not too many, once we factor out the filesystem-based trace-writer? It might be nice to separate that part so we can do recording in no-std environments (e.g. embedded use-cases) too.

crates/wasmtime/Cargo.toml
```diff
@@ -393,3 +393,20 @@ component-model-async = [
...
+# Core wasm RR support
+rr-core = ["std"]
+
+# Support for type information of recorded events for replay validation
+rr-type-validation = []
+
+# Support for input values to recorded events for replay validation.
+#
+# This can be used to check whether the entire module is truly deterministic
+rr-args-validation = []
```
cfallin: I definitely appreciate the detailed feature-slicing here but I wonder if it might be a bit too much to be manageable internally -- perhaps we could have just `rr` and then `rr-validate`?

Also: have you measured the impact of validation checks that are always compiled in but disabled dynamically?

crates/wasmtime/src/config.rs
```diff
@@ -99,6 +101,17 @@ impl core::hash::Hash for ModuleVersionStrategy {
     }
 }
 
+impl ModuleVersionStrategy {
+    /// Get the string-encoding version of the module
```
cfallin: style nit: we end doc-comments with periods so `s/module/module./`. Likewise for the rest of the doc-comments in this diff (I see a few instances below too).

crates/wasmtime/src/config.rs
```diff
@@ -219,6 +233,106 @@ impl Default for CompilerConfig {
     }
 }
 
+/// Metadata for specifying recording strategy
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RecordMetadata {
```
cfallin: Is this "metadata" or is "config" a better name for it? Ordinarily I'd expect "metadata" to have something to do with the actual recorded content ("meta" to the "data"), e.g., module versions or something.

crates/wasmtime/src/config.rs
```diff
@@ -219,6 +233,106 @@ impl Default for CompilerConfig {
...
+            add_validation: true,
+        }
+    }
+}
+
+/// A [`Write`] usable for recording in RR
+///
+/// Only constraint is that writer must be Send + Sync
+pub trait RecordWriter: Write + Send + Sync {}
+impl<T: Write + Send + Sync> RecordWriter for T {}
```
cfallin: A thought: right now we have a `RecordWriter` that sinks trace content at the level of pre-serialized bytes (and hence we take a serializer as a dependency). Would it make sense to have a sink that instead consumes Rust-typed records, and is itself in charge of serializing them? That would let us build other sorts of interesting filtering external to Wasmtime; and would also mean that the trace reading and writing backends could be used as libraries for tooling. It would push a `std` dependence out of the core, too.

We could still provide fast default serialization/deserialization methods if we wanted (`Record::to_bytes()` / `Record::from_bytes()` or similar).

crates/wasmtime/src/config.rs
```diff
@@ -2625,6 +2748,51 @@ impl Config {
         self.tunables.signals_based_traps = Some(enable);
         self
     }
+
+    /// Enforce deterministic execution configurations
+    ///
+    /// Required for faithful record/replay execution. Currently, this does the following:
```
cfallin: Let's put "required for faithful record/replay execution" as a new paragraph below this -- it's true and useful information, but it's secondary to what the option actually does.

Also, does "required" here mean that the user is required to configure it, or is it just describing a fact about this kind of configuration (and that is magically enabled some other way)?

(Sorry for writing nits here -- this becomes public documentation so we want to be very clear/unambiguous!)

crates/wasmtime/src/config.rs
```diff
@@ -2625,6 +2748,51 @@ impl Config {
...
+    /// * Enables NaN canonicalization with [`Config::cranelift_nan_canonicalization`]
+    /// * Enables deterministic relaxed SIMD with [`Config::relaxed_simd_deterministic`]
+    #[inline]
+    pub fn enforce_determinism(&mut self) -> &mut Self {
+        self.cranelift_nan_canonicalization(true)
+            .relaxed_simd_deterministic(true);
+        self
+    }
+
+    /// Repeal determinstic execution configurations (opposite
```
cfallin: Rather than "repeal", I might say "disable"?

crates/wasmtime/src/config.rs
```diff
@@ -2625,6 +2748,51 @@ impl Config {
...
+            .relaxed_simd_deterministic(false);
+        self
+    }
+
+    /// Evaluates to true if current configuration must respect
+    /// deterministic execution in its configuration
+    ///
+    /// Required for faithful record/replay execution
+    #[inline]
+    pub fn check_determinism(&mut self) -> bool {
```
cfallin: `check_determinism` sounds like it could be enabling some option on the runtime config -- usually for bool predicates we say `is_*`, so perhaps `is_determinism_enforced()`?

crates/wasmtime/src/config.rs
```diff
@@ -2625,6 +2748,51 @@ impl Config {
...
+        self.rr.is_some()
+    }
+
+    /// Configure the record/replay options for use by the runtime
+    ///
+    /// This method implicitly enforces determinism (see [`Config::enforce_determinism`]
+    /// for details).
+    pub fn rr(&mut self, rr: Option<RRConfig>) -> &mut Self {
```
cfallin: `s/rr/record_replay/`? Or better, separate `enable_record()` and `enable_replay` methods, each taking the data that goes into one arm of `RRConfig`?

crates/wasmtime/src/runtime/vm/vmcontext.rs
```diff
@@ -1533,6 +1533,18 @@ impl ValRaw {
         assert!(cfg!(feature = "gc") || anyref == 0);
         anyref
     }
+
+    /// Get the raw bits of the union
+    #[inline]
+    pub fn as_bytes(&self) -> [u8; mem::size_of::<Self>()] {
+        unsafe { mem::transmute(*self) }
```
cfallin: I like to avoid transmute whenever possible, as it is so easy to get wrong; could we instead add a member to the union that is the size of the whole union (e.g., 128 bits) and read out the content that way? It is still unsafe (because it's a union) but is a whole lot more constrained.

Alternately, if you have to use transmute, an idiom we have in a bunch of places is `transmute::<From, To>(...)`, i.e., explicitly write out the types so we can be sure we're casting what we think we are (and not, e.g., accidentally casting a pointer itself rather than the contents at that pointer).

crates/wasmtime/src/runtime/component/func/host.rs
```diff
@@ -12,11 +15,75 @@ use alloc::sync::Arc;
...
 use core::mem::{self, MaybeUninit};
 use core::ptr::NonNull;
+#[cfg(feature = "rr-type-validation")]
+use wasmtime_environ::component::TypeTuple;
 use wasmtime_environ::component::{
     CanonicalAbiInfo, ComponentTypes, InterfaceType, MAX_FLAT_PARAMS, MAX_FLAT_RESULTS,
     StringEncoding, TypeFuncIndex,
 };
 
+/// Record/replay stubs for host function entry events
```
cfallin: We should add more documention here: what is the macro for, how is it used, generally what does it expand to at a high level?

(It seems to dispatch to either adding a new event when recording, or advancing the trace and validating against it when replaying, right?)

crates/wasmtime/src/runtime/component/func/host.rs
```diff
@@ -12,11 +15,75 @@ use alloc::sync::Arc;
...
+                        $args,
+                        #[cfg(feature = "rr-type-validation")]
+                        Some($param_types),
+                    )
+                },
+            );
+            $store.next_replay_event_if(
+                |_, r| r.add_validation,
+                |_event: HostFuncEntryEvent, _r: &ReplayMetadata| {
+                    #[cfg(feature = "rr-type-validation")]
```
cfallin: Rather than conditionalizing individual statements, generally what we try to do for this sort of thing is conditionalize on function definitions, and write alternate (often empty) ones -- this keeps the flow a bit cleaner inside bodies. So something like:

```
#[cfg(feature = "rr-type-validation")]
fn validate(...) { /* actually do it */ }

#[cfg(not(feature = "rr-type-validation"))]
fn validate(...) { /* nothing */ }
```

on `event`'s impl, then always write `if r.validate { event.validate(...); }` here. This also avoids the awkward unused-variable warnings you're working around with the underscores here.

crates/wasmtime/src/runtime/component/func/host.rs
```diff
@@ -12,11 +15,75 @@ use alloc::sync::Arc;
...
+#[cfg(feature = "rr-type-validation")]
+use wasmtime_environ::component::TypeTuple;
 use wasmtime_environ::component::{
     CanonicalAbiInfo, ComponentTypes, InterfaceType, MAX_FLAT_PARAMS, MAX_FLAT_RESULTS,
     StringEncoding, TypeFuncIndex,
 };
 
+/// Record/replay stubs for host function entry events
+macro_rules! rr_host_func_entry_event {
+    { $args:expr, $param_types:expr => $store:expr } => {
```
cfallin: The "arrow syntax" here is a little too clever for my tastes at least -- it's not clear what is going on when I see that in the macro use-sites. I'd prefer the macro invocations to look as much like normal function calls as possible.

crates/wasmtime/src/runtime/component/func/host.rs
cfallin: And in general actually, it might be good to have an explanatory comment here about record/replay hook points and what the sequences are, and how you're using macros to generalize over all of those.

Something like:
- The sorts of events that need to be recorded and/or validated (e.g. write out a call sequence and note each point in the sequence)
- Somehow describe strategy here: you have two main macros that are invoked at the start and end of a host-function call, and a few different kinds of wrappers to record the important values within that span.

crates/wasmtime/src/runtime/component/func/host.rs
```diff
@@ -225,20 +294,42 @@ where
...
     flags.set_may_leave(true);
 
-    lower.exit_call()?;
+    if !replay_enabled {
+        lower.exit_call()?;
+    }
 
     return Ok(());
 
+    enum ReturnMode<Return: Lower> {
```
cfallin: Can we add a doc-comment here describing why we need a separate `Replay` return-mode? (Basically, the effects of the return are handled by replaying the trace instead, right?)

crates/wasmtime/src/runtime/component/func/host.rs
```diff
@@ -270,21 +361,65 @@ where
             &mut self,
             cx: &mut LowerContext<'_, T>,
             ty: InterfaceType,
-            ret: R,
+            ret: ReturnMode<R>,
```
cfallin: Just a hunch but after reading through this logic, I wonder: would it clean up the flow somewhat if we dispatched once on `ReturnMode` at the top level, then avoided all of the separate conditionals throughout the body? As-is, it's quite difficult to follow the different paths -- it seems somewhat decoupled and overly-abstracted.

crates/wasmtime/src/runtime/component/func/host.rs
```diff
@@ -372,61 +512,119 @@ where
...
-            .map(|ty| Val::lift(&mut cx, *ty, &mut iter))
-            .collect::<Result<Box<[_]>>>()?;
-        ret_index = param_count;
-        assert!(iter.next().is_none());
+
+    let replay_enabled = store.0.replay_enabled();
+
+    rr_host_func_entry_event! { storage, &types[func_ty.params] => store.0 };
+
+    let results = if !replay_enabled {
```
cfallin: Same comment here as above re: dispatching at a higher level.

In general, the sense I'm getting is that this sequence evolved out of a desire to keep the overall shape the same, and add hooks where needed. But it seems to me that there are basically three possible flows at a high level:

- Normal host call;
- Host call, recorded;
- Replay of host call.

Could we instead write three different `lower_results` implementations depending on our mode, and dispatch once on the mode in `call_host`, to avoid the maze of fine-grained conditionals?

crates/wasmtime/src/runtime/component/func/options.rs
```diff
@@ -2,16 +2,111 @@ use crate::component::ResourceType;
...
+/// The above are intended to be the narrow waists for recording changes to guest state, and
+/// should be the **only** interfaces used during lowerng. In particular,
+/// [`get`](LowerContext::get)/[`get_dyn`](LowerContext::get_dyn) return
+/// ([`ConstMemorySliceCell`]/[`MemorySliceCell`]), which implement [`Drop`]
+/// allowing us a hook to just capture the final aggregate changes made to guest memory by the host.
+///
+/// ## Critical Invariants
+///
+/// Typically recording would need to know both when the slice was borrowed AND when it was
+/// dropped, since memory movement with [`realloc`](LowerContext::realloc) can be interleave between
```
cfallin: s/interleave/interleaved/

crates/wasmtime/src/runtime/store.rs
```diff
@@ -1334,6 +1369,93 @@ impl StoreOpaque {
...
+        if let Some(buf) = self.record_buffer_mut() {
+            buf.record_event(f);
+        }
+    }
+
+    /// Conditionally record the given event into the store's record buffer
+    ///
+    /// Convenience wrapper around [`Recorder::record_event_if`]
+    #[inline]
+    pub(crate) fn record_event_if<T, P, F>(&mut self, pred: P, f: F)
```
cfallin: It seems that the `_if` part of this combinator is mainly used to conditionalize running the closure on `add_validation`. Would it make sense to instead have `record_event` and `record_event_validation` entry points, or perhaps even just wrap the relevant callsites in `if r.add_validation { store.record_event(...); }`? In general I'm wary of complex type-generic closure-taking helpers like this unless they pay their way to a significant degree.

cfallin (commented):
Initial review -- this is looking really good overall!

My main high-level thought around the heart of the record/replay (the interaction with the component hostcall logic and lifting/lowering): there are a lot of fine-grained conditionals, and it's somewhat hard to trace, even though there are only three modes (running + recording, replaying, or just running). It feels like somehow we should be able to hoist the conditionals a bit more and then have straight-line sequences for each of those modes.

A few other misc thoughts:
- We'll want to make sure we either plug all soundness holes or document them for the initial landing. We already know about component builtins (and their memory effects); we should ask around to ensure we catch any others.
- For core-wasm recording, I suspect we'll want to validate that there are no exported memories or tables.
- We should validate that the module is exactly the same on replay as on record -- I know you're working on adding hashing for this.
- Some comments below about `std` dependence -- we should ensure the core can work, serializing into memory, in a no-std build.
- We'll want at least @alexcrichton to give this a review as well, maybe after we try some of the refactors mentioned here.

Thanks again!

alexcrichton:
> We'll want at least @alexcrichton to give this a review as well, maybe after we try some of the refactors mentioned here.

Happy to help out! (and agreed I'd like to once-over at some point)

How about scheduling a call when y'all are ready with the 3 of us? That'd probably be best to draw attention to any various areas and for me to ask some questoins in a high-bandwidth way before going off to review on my own.

crates/cli-flags/src/lib.rs
fitzgen: +1 to what @cfallin said about replay being a command, not an option.

crates/wasmtime/src/runtime/component/func/host.rs
```diff
@@ -203,6 +270,8 @@ where
     let param_tys = InterfaceType::Tuple(ty.params);
     let result_tys = InterfaceType::Tuple(ty.results);
 
+    rr_host_func_entry_event! { storage, &types[ty.params] => cx.0 };
```
fitzgen: I'll go a step further than @cfallin did regarding the macros: I don't think that they are paying for themselves at all. I think we could use regular function calls and get very close to the macros in brevity, but end up with something much easier to understand for readers:

```suggestion
    cx.0.host_func_entry_event(storage, &types[ty.params])?;
```

And if we are concerned about the overhead of indexing into `types` here even when we aren't recording, then we could pass `types` and `ty.params` as two different arguments and the method would only index when recording was enabled. (Either way, we would probably want to mark the method as `#[inline]`.)

crates/wasmtime/src/runtime/component/func/options.rs
```diff
@@ -374,6 +551,63 @@ impl<'a, T: 'static> LowerContext<'a, T> {
...
+    ) -> Result<()> {
+        if self.store.0.replay_buffer_mut().is_none() {
+            return Ok(());
+        }
+        let mut complete = false;
+        while !complete {
+            let buf = self.store.0.replay_buffer_mut().unwrap();
+            let event = buf.next_event()?;
+            let _replay_metadata = buf.metadata();
+            let _ = match event {
```
fitzgen: Nitpick: the `let _ =` shouldn't be necessary here (you may also need to remove the semicolon at the end of the `match`).

crates/wasmtime/src/runtime/rr/events/core_wasm.rs
```diff
@@ -0,0 +1,76 @@
+//! Module comprising of core wasm events
+use super::*;
+#[allow(unused_imports)]
```
fitzgen: This should be `#[expect(unused_imports, reason = "<description of why>")]` so that when we fix the warning, the missing warning becomes a warning itself, so that we don't forget to remove the `allow`.

If it is only unused in some `cfg`s, then we want to either `#[cfg(...)]` the imports appropriately, or use `#[cfg_attr(expect(unused_imports, reason = "..."))]`. In some scenarios one or the other makes more sense.

crates/wasmtime/src/runtime/rr/events/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+        ValRaw::from_bytes(value)
+    }
+}
+
+impl ValRawBytesConvertable for MaybeUninit<ValRaw> {
+    #[inline]
+    fn to_valraw_bytes(self) -> ValRawBytes {
+        // Uninitialized data is assumed and serialized, so hence
+        // may contain some undefined values
+        unsafe { self.assume_init() }.to_valraw_bytes()
```
fitzgen: I don't think this is sound. Instead, I think we'll want something like this:

```rust
union SerializedValRaw {
    bytes: ValRawBytes,
    val: ValRaw,
}

impl SerializedValRaw {
    pub fn new(val: ValRaw) -> Self {
        // Zero-initialize `self.bytes` to ensure that there are
        // no undefined bytes in `self` and we don't ever try to
        // read and serialize undefined data.
        let ret = Self {
            // SAFETY: it is safe to zero-initialize `u8` arrays.
            bytes: unsafe { mem::zeroed() },
        };
        ret.val = val;
        ret
    }

    pub fn get_val(&self) -> ValRaw {
        // SAFETY: `self` is always initialized in `new` such
        // that there is a valid `ValRaw` inside.
        unsafe { self.val }
    }

    pub fn get_bytes(&self) -> ValRawBytes {
        // SAFETY: We take care to ensure that `self` has no
        // undefined bytes in the constructor, so accessing the
        // raw bytes representation of `ValRaw` is safe.
        unsafe { self.bytes }
    }
}
```

crates/wasmtime/src/config.rs
fitzgen: Alternatively, if we don't pursue what @cfallin is sketching out, then I think we can just use `std::io::Write` directly rather than defining a new `RecordWriter` trait.

crates/wasmtime/src/runtime/store.rs
```diff
@@ -1334,6 +1369,97 @@ impl StoreOpaque {
...
+        if let Some(buf) = self.replay_buffer_mut() {
+            buf.next_event_if(pred, f)
+        } else {
+            Ok(())
+        }
+    }
+
+    /// Check if recording or replaying is enabled for the Store
+    #[inline]
+    pub fn rr_enabled(&self) -> bool {
```
fitzgen: Do we ever want to check if *either* recording or replaying is enabled and then take some action in both cases, rather than checking for just recording and then doing some recording-specific action or just checking replaying and then doing some replaying-specific action? I might be missing something, but I am not seeing any uses of this method.

crates/wasmtime/Cargo.toml
fitzgen: Yeah, actually we definitely want to enable record-replay in no-std long term, so I take back my previously suggested option of using `std::io::Write` instead of `RecordWriter` and agree that we should do something a little more generic, along the lines of what Chris suggested of having a trait method for appending each kind of event to the trace.

fitzgen (commented):
Super excited for this!

I think we should adjust the `#[cfg(...)]` options / cargo features a little bit. Instead of having a `cfg` feature for just type validation, but always including the general record-replay code in the build, I think we should have a `#[cfg(feature = "rr")]` cargo feature for controlling whether we are building in the ability to do any kind of record-replay at all. (We don't want to force the smallest embedded builds of Wasmtime to include record-replay infrastructure, for example.) And then, when record-replay support was included at compile time, we can have runtime boolean knobs for whether to do type validation and divergence-checking when replaying. 

Finally, in order to make it so that the whole core runtime isn't littered with `#[cfg(feature = "rr")]`, we can do something similar to what we do with `#[cfg(feature = "gc")]` and the `wasmtime::runtime::gc` submodule:

```rust
// crates/wasmtime/src/runtime/rr.rs

#[cfg(feature = "rr")]
mod enabled;
#[cfg(feature = "rr")]
pub use enabled::*;

#[cfg(not(feature = "rr"))]
mod disabled;
#[cfg(not(feature = "rr"))]
pub use disabled::*;
```

The `wasmtime::runtime::rr::enabled` submodule would have the actual record-replay implementation, and the `wasmtime::runtime::rr::disabled` submodule would have a stubbed out version of the same public API but without constructors and with every public type being a newtype of [`wasmtime::runtime::uninhabited::Uninhabited`](https://github.com/bytecodealliance/wasmtime/blob/0f457fad63258ef3354008eb2da188cae0a63686/crates/wasmtime/src/runtime/uninhabited.rs#L5). This pattern lets the core runtime usage of record-replay APIs look the same regardless whether we actually build in the code to do record-replay.

The other thing I think we need before this can land is some kind of testing or fuzzing story. At minimum, we should (1) [make it so that the fuzzer can turn on recording during our existing fuzzing](https://github.com/bytecodealliance/wasmtime/blob/815c10dea9cc130316ac9b58c4aaed3e97eb8760/crates/fuzzing/src/generators/config.rs#L541-L580), and (2) we should have a smoke test that records some kind of Wasm program and then replays it back again and asserts that we get the same result again. As a follow up, I think we should also generalize that smoke test into a new fuzz target where we take an arbitrary Wasm module generated by the fuzzer, record its execution, and then replay that execution and assert that we get the same results (we can mostly rely on enabling the internal type validation and divergence checks for these assertions).

Let me know if any of this isn't clear or if I've overlooked something.

Thanks!

cfallin:
> How about scheduling a call when y'all are ready with the 3 of us? That'd probably be best to draw attention to any various areas and for me to ask some questoins in a high-bandwidth way before going off to review on my own.

That'd be great! FWIW, I'm on PTO next week and the week after; please feel free to talk with Arjun directly before then if you both want, or I can join after Aug 11...

arjunr2:
@alexcrichton @fitzgen I'll take a pass through the comments early next week, and perhaps we can find a time later next week that works

alexcrichton:
Sounds good! Feel free to ping me on Zulip when ready

crates/wasmtime/src/runtime/component/func/host.rs
arjunr2: The reason I opted for macros was to prevent unnecessary duplication across `call_host` and `call_host_dynamic`, which seemed harder to parse at initial glance, at least earlier. I can do a refactor removing them now and take a look

crates/cli-flags/src/lib.rs
arjunr2: Replay in practice operates very similar to the `run` command and shares a lot of the top level command line options with it as well.

The replay command as a result would be just a minor super-set of the `run` setup. It sort of makes sense to do one of two things:
* Have a top-level `replay` command, that just adds additional replay options, and calls `run` under the hood with it
* Add replay options perhaps just for the `run` command (as opposed to general options)

Leaning towards the former since maybe in the future `replay` could support different modes of operation

fitzgen: > Replay in practice operates very similar to the `run` command and shares a lot of the top level command line options with it as well.

Shouldn't all the top-level options pretty much be fixed and effectively identical to whatever they were when the trace was recorded? When does it make sense to specify different options than were used when recording? If the answer is "it doesn't make sense" then I don't think we want to force users to manually provide the same options from before in the replay command.

arjunr2: I guess that's true for most options. I think the ones that might be usable are perhaps just profiling

fitzgen: Ah yeah great point, very cool to profile an execution after the fact and ignore I/O and the external world.

I think we can probably add CLI options one at a time as they make sense to the replay command, and have the default[^0] be "whatever was configured during the recording".

[^0]: I think this is implicit/automatic in that the trace will reflect whatever configuration settings were in effect during recording? At least that is true for anything that affects execution.

crates/wasmtime/src/runtime/store.rs
arjunr2: You're right, this is now obsolete and removed.

SebastienGllmt:
For anybody wondering what's happening with this given the PR was closed, I believe development was moved into a separate repo: https://github.com/bytecodealliance/wasmtime-rr-prototyping/tree/rr-dev

PR #11290, opened 2025-07-21
Add fuzzer integration for gc/mutatis

khagankhan:
- Add integration of `mutatis` wit fuzzer
- remove `serialize/deserialize` for config instead use `u64`
- use `serialize/deserialize` for `fuzz_mutate`


crates/fuzzing/src/oracles.rs
```diff
@@ -1331,6 +1332,22 @@ pub fn call_async(wasm: &[u8], config: &generators::Config, mut poll_amts: &[u32
...
+        *b = rng.r#gen()
+    }
+
+    let u = Unstructured::new(&buf);
+    let config = generators::Config::arbitrary_take_rest(u)
+        .expect("should be able to generate config from seed");
+
+    let _ = crate::oracles::table_ops(config, ops);
+}
+
```
fitzgen: We already have an oracle for fuzzing table ops, which is what this function is ultimately calling. This is really just a small adapter for when we need to use `Arbitrary` to generate a config, and it feels out of place here. The only place we need to actually do that is in the fuzz target itself, so lets just inline it into the `fuzz_target!` invocation in `fuzz/fuzz_targets/table_ops.rs`.

fuzz/Cargo.toml
```diff
@@ -12,6 +12,8 @@ workspace = true
 cargo-fuzz = true
 
 [dependencies]
+mutatis = { workspace = true }
+bincode = "2.0.1"
```
fitzgen: We should use `postcard` instead of `bincode` here, because we already depend on it in the workspace and don't need to worry about auditing or anything.

```suggestion
postcard = { workspace = true }
```

fuzz/fuzz_targets/table_ops.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    // With probability of about 1/8, just use the default mutator.
+    if seed.count_ones() % 8 == 0 {
+        return fuzzer_mutate(data, size, max_size);
+    }
+
+    // Fallback to default on decode failure
+    let mut tuple =
+        bincode::decode_from_slice::<(u64, TableOps), _>(data, bincode::config::standard())
+            .map_or_else(|_err| (0, TableOps::default()), |(tuple, _)| tuple);
```
fitzgen: Slight simplification:

```suggestion
        bincode::decode_from_slice::<(u64, TableOps), _>(data, bincode::config::standard())
            .ok()
            .unwrap_or_default();
```

crates/fuzzing/src/generators/table_ops.rs
```diff
@@ -182,6 +183,10 @@ impl TableOps {
 
         self.ops = new_ops;
     }
+    /// Pops from the vector of the opcodes and returns bool
+    pub fn pop(&mut self) -> bool {
+        self.ops.pop().is_some()
+    }
```
fitzgen: Nitpicks:

* Newline between methods, as elsewhere in the file
* "returns bool" is not very descriptive, the doc comment should describe what the boolean result means.

crates/fuzzing/Cargo.toml
```diff
@@ -15,11 +15,13 @@ workspace = true
 wasmtime-test-util = { workspace = true, features = ['wast'] }
 
 [dependencies]
+bincode = "2.0.1"
```
fitzgen: `postcard` here too:

```suggestion
postcard = { workspace = true }
```

fitzgen (commented):
Looks good, but there are a handful of small things to address before we merge this. Thanks!

crates/fuzzing/src/oracles.rs
khagan-karimov: Yea makes sense! I guess I just updated the previous function which was in this file

fuzz/fuzz_targets/table_ops.rs
khagan-karimov: This is nice!

crates/fuzzing/src/generators/table_ops.rs
khagan-karimov: I just wanted it to pass the clippy haha. Thanks!

khagan-karimov:
Thanks! Working on them!

fuzz/fuzz_targets/table_ops.rs
```diff
@@ -0,0 +1,55 @@
...
+
+use libfuzzer_sys::arbitrary::{Arbitrary, Unstructured};
+use libfuzzer_sys::{fuzz_mutator, fuzz_target, fuzzer_mutate};
+use mutatis::Session;
+use postcard::{from_bytes, to_slice};
+use rand::{Rng, SeedableRng};
+use wasmtime_fuzzing::generators::table_ops::TableOps;
+use wasmtime_fuzzing::oracles::table_ops;
+
+fuzz_target!(|input: (u64, TableOps)| {
```
fitzgen: We don't want the fuzz target to take structured types directly, as those get generated via `Arbitrary`, and we want to do our serialization- and mutation-based thing here.

Instead we should do something like this:

```rust
fuzz_target!(|data: &[u8]| {
    let Ok((seed, ops)) = postcard::from_bytes::<(u64, TableOps)>(data) else {
        return;
    }

    // As before: Make an RNG from the seed, fill the buf with random
    // data, make an `Unstructured`, call `Config::arbitrary_take_rest`,
    // and finally invoke `oracles::table_ops(config, ops)`.
});
```

Aside: once this PR lands, we should probably remove the `Arbitrary` implementation for `TableOps` as we won't be using it anymore and removing it will make usage footguns like this one disappear.

fitzgen (commented):
Very close!

We should also remove `table_ops` from the `misc` fuzz target as part of this PR, as it will be redundant.

With the comment below addressed, this should be good to merge.

Thanks!

fitzgen:
@khagankhan it looks like the `Cargo.lock` is still updating all dependencies. Do you have a tool that is doing that locally?

khagankhan:
I ran them locally and passed. Let us see!!

khagankhan:
@fitzgen Done ✅

fitzgen (approved):
Thanks!

alexcrichton:
Would it be possible to re-integrate the `table_ops` fuzzer back into the main "misc" fuzzer?

fitzgen:
Unfortunately, we can't because it is using mutation-based fuzzing via `mutatis` now, not generative-based fuzzing via `arbitrary`. I think that's fine though because it is going to be gaining more GC-specific features and will start exploring new code paths now, rather than just the same paths it has been executing for a long time.

alexcrichton:
Sounds good to me 👍

PR #11291, opened 2025-07-21
feat(wasip3): implement `wasi:sockets`

rvolosatovs:
Implement `wasip3` `wasi:sockets`.

I've refactored `FutureWriter::write` a little bit to allow for non-blocking send

Refs https://github.com/bytecodealliance/wasip3-prototyping/issues/228
Closes https://github.com/bytecodealliance/wasip3-prototyping/issues/226 (note the addition of yield from the guest)
I was not able to reproduce https://github.com/bytecodealliance/wasip3-prototyping/issues/44 with:
```sh
while cargo test --features p3 p3_sockets_udp_sample_application; do :; done
```
on MacOS


rvolosatovs:
FYI: I'm currently working on reusing most of the socket utilities across p2 and p3 to avoid duplication

crates/wasi/src/p3/sockets/udp.rs
```diff
@@ -0,0 +1,281 @@
...
+    addr: SocketAddr,
+) -> Result<(), ErrorCode> {
+    loop {
+        let n = socket.send_to(buf, addr).await?;
+        if n == buf.len() {
+            return Ok(());
+        }
+        buf = &buf[n..]
+    }
+}
```
alexcrichton: This may be a preexisting bug in the p2 implementation, but I'm under the impression that the equivalent of `write_all` for UDP is inappropriate. I'd expect this to return an error if the entire buffer wasn't sent personally, but I'm also not sure the context these are used within.

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -511,7 +511,7 @@ impl<T> FutureWriter<T> {
     ///
     /// Panics if the store that the [`Accessor`] is derived from does not own
     /// this future.
-    pub async fn write(mut self, accessor: impl AsAccessor, value: T) -> bool
+    pub fn write(mut self, accessor: impl AsAccessor, value: T) -> impl Future<Output = bool>
```
alexcrichton: Could you detail a bit more the rationale for this change? The PR description mentions a non-blocking send but I don't fully grok that. In the context of https://github.com/bytecodealliance/wasip3-prototyping/pull/240 this method is going to change more so if precise control over sending messages is needed that may be a red flag of something we'll need to change at the call-site rather than here

crates/wasi/src/p3/sockets/host/types/tcp.rs
```diff
@@ -0,0 +1,637 @@
...
+                    }
+                }
+            };
+            let socket = store.with(|mut view| {
+                view.get()
+                    .table
+                    .push(TcpSocket::from_state(state, self.family))
+                    .context("failed to push socket to table")
+            })?;
+            self.tx.write(store, Some(socket)).await;
```
alexcrichton: I think this'll want some extra handling, namely:

* If `Some(...)` is returned, that means the other side went away, so the loop can be exited immediately here.
* If `Some(...)` is returned then the resource table needs to be updated here to take out what was previously pushed in

crates/wasi/src/p3/sockets/host/types/tcp.rs
```diff
@@ -0,0 +1,637 @@
...
+    #[cfg(target_os = "macos")]
+    keep_alive_idle_time: Arc<core::sync::atomic::AtomicU64>, // nanoseconds
+}
+
+impl<T> AccessorTask<T, WasiSockets, wasmtime::Result<()>> for ListenTask {
+    async fn run(mut self, store: &Accessor<T, WasiSockets>) -> wasmtime::Result<()> {
+        loop {
+            let Some(res) = ({
+                let mut accept = pin!(self.listener.accept());
+                let mut tx = pin!(self.tx.watch_reader(store));
```
alexcrichton, line 69: In the future we may have to watch out for this as `watch_reader` is a pretty expensive operation right now and each iteration of the loop unconditionally sets this up. The changes at https://github.com/bytecodealliance/wasip3-prototyping/pull/240 may help though.

crates/wasi/src/p3/sockets/host/types/tcp.rs
```diff
@@ -0,0 +1,637 @@
...
+    // on all platforms. So we keep track of which options have been explicitly
+    // set and manually apply those values to newly accepted clients.
+    #[cfg(target_os = "macos")]
+    receive_buffer_size: Arc<core::sync::atomic::AtomicUsize>,
+    #[cfg(target_os = "macos")]
+    send_buffer_size: Arc<core::sync::atomic::AtomicUsize>,
+    #[cfg(target_os = "macos")]
+    hop_limit: Arc<core::sync::atomic::AtomicU8>,
+    #[cfg(target_os = "macos")]
+    keep_alive_idle_time: Arc<core::sync::atomic::AtomicU64>, // nanoseconds
```
alexcrichton: Could this be wrapped up in something like `Arc<NonInheritedOptions>` or something like that? That'll help cut down on the `Arc`-traffic and additionally enable scoping handling of this to a separate file and/or abstraction. Ideally it'd also be shared with the p2 implementation if possible which I suspect has very similar #[cfg].

crates/wasi/src/p3/sockets/host/types/tcp.rs
```diff
@@ -0,0 +1,637 @@
...
+struct ReceiveTask {
+    stream: Arc<TcpStream>,
+    data_tx: StreamWriter<Cursor<BytesMut>>,
+    result_tx: FutureWriter<Result<(), ErrorCode>>,
+}
+
+impl<T> AccessorTask<T, WasiSockets, wasmtime::Result<()>> for ReceiveTask {
+    async fn run(mut self, store: &Accessor<T, WasiSockets>) -> wasmtime::Result<()> {
+        loop {
+            let mut buf = BytesMut::with_capacity(8192);
```
alexcrichton: I think this buf is intended to be outside of the `loop`?

alexcrichton: Also, mind extracting this constant to a `const` at the top of the file (or something like that) to reference?

crates/wasi/src/p3/sockets/host/types/tcp.rs
```diff
@@ -0,0 +1,637 @@
...
+                }
+                Ok(..) => {
+                    buf = self
+                        .data_tx
+                        .write_all(store, Cursor::new(buf))
+                        .await
+                        .into_inner();
+                    if self.data_tx.is_closed() {
+                        _ = self.result_tx.write(store, Ok(()));
+                        break;
```
alexcrichton: All of the `break`s in this loop are write-then-break, so could this be refactored as break-with-value followed by just one write?

crates/wasi/src/p3/sockets/host/types/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+    store.with(|mut view| view.get().ctx.socket_addr_check.clone())
+}
+
+async fn is_addr_allowed<T>(
+    store: &Accessor<T, WasiSockets>,
+    addr: SocketAddr,
+    reason: SocketAddrUse,
+) -> bool {
+    get_socket_addr_check(store)(addr, reason).await
+}
```
alexcrichton, line 24: One thing we may want to keep an eye out for in the future is some sort of fast path here to avoid the `Arc::clone` in the typical case.

crates/wasi/src/p3/sockets/host/types/tcp.rs
```diff
@@ -0,0 +1,637 @@
...
+                        }
+                        Err(err) => {
+                            socket.tcp_state = TcpState::Closed;
+                            Ok(Err(err.into()))
+                        }
+                    }
+                })
+            }
+            Ok(Err(err)) => Ok(Err(err)),
+            Err(err) => Err(err),
```
alexcrichton: This branch I think could be replaced with a `?` on the `with` block.

crates/wasi/src/p3/sockets/host/types/tcp.rs
```diff
@@ -0,0 +1,637 @@
...
+            match mem::replace(&mut socket.tcp_state, TcpState::Connecting) {
+                TcpState::Default(sock) | TcpState::Bound(sock) => Ok(Ok(sock)),
+                tcp_state => {
+                    socket.tcp_state = tcp_state;
+                    Ok(Err(ErrorCode::InvalidState))
+                }
+            }
+        }) {
+            Ok(Ok(sock)) => {
+                let res = sock.connect(remote_address).await;
```
alexcrichton, line 261: It might be worth leaving a `FIXME` here along the lines of handling cancellation. Here we'll ideally want to hook into the drop of the outer future and put the socket back in the store or something like that.

crates/wasi/src/p3/sockets/host/types/tcp.rs
```diff
@@ -0,0 +1,637 @@
...
+        store.with(|mut view| {
+            let instance = view.instance();
+            let (data_tx, data_rx) = instance
+                .stream::<_, _, BytesMut>(&mut view)
+                .context("failed to create stream")?;
+            let (result_tx, result_rx) = instance
+                .future(|| unreachable!(), &mut view)
+                .context("failed to create future")?;
+            let TcpSocket { tcp_state, .. } = get_socket_mut(view.get().table, &socket)?;
+            match mem::replace(tcp_state, TcpState::Closed) {
```
alexcrichton, line 397: Should this be using `Receiving` instead of `Closed`? As-is it looks like you can send-then-receive but you can't receive-then-send.

crates/wasi/src/p3/sockets/udp.rs
badeend: Nice catch. This is indeed not correct for UDP.

The [P2 implementation](https://github.com/bytecodealliance/wasmtime/blob/0e6d9c679f5df5cf712357f147088bbce0c61fe9/crates/wasi/src/p2/host/udp.rs#L454-L458) simply ignores the return value.

I'd have to refresh my mind on this, but I recall that UDP either: returns successfully with input buffer size or fails with EMSGSIZE, but never silently truncates the messages. Could be wrong.

Edit: If ignoring the return value is not desirable, we could add an assert that verifies the returned length matches the input length. Or return a trap.

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
rvolosatovs: Currently, calling `FutureWriter::write` without awaiting the returned future does nothing. This patch makes it so that once `Future::write` returns it is guaranteed that a value is *written* akin to https://docs.rs/tokio/latest/tokio/sync/oneshot/struct.Sender.html#method.send, which is used internally. *async* is only used then to await the *receipt* of the value.

From the WASI implementation perspective this means that we now can write future values without an `async` context, meaning, for example, that we don't need to spawn a task just to send a static value like was done e.g. here: https://github.com/bytecodealliance/wasip3-prototyping/blob/9c837320498adb1f4bda9ad7bbb12769fcdd08b5/crates/wasi/src/p3/sockets/host/types/tcp.rs#L482-L487

crates/wasi/src/p3/sockets/host/types/tcp.rs
rvolosatovs: `Closed` is normally used throughout the codebase as a marker to emulate a *take* operation, the resulting state is never `Closed`, it's either changed (to receiving) or rolled back, if it was not *connected*

crates/wasi/src/p3/sockets/udp.rs
alexcrichton: I think it'd be reasonable to check the return value to double-check it matches the message size on success and synthesizing our trap (or error) on failure.

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
alexcrichton: If possible I'd prefer to avoid this change and perhaps add a new method for this. I'm worried about conflating the guarantees/semantics of Rust-level `async` with the semantics desired here. For example this seems like it would encourage calling `write` and dropping the returned future while still having a guarantee that the item was sent. That the item is sent in this situation is, in my opinion, a bug. Dropping a future should cancel whatever that future's computation is so in my mind the semantics of this function is that you're never guaranteed that the value is sent until the future is `.await`-ed.

I think it's reasonable though to shoot for an ideal of being able to send a value without having to full wait for its receipt. For this an extra host-level function I think would be appropriate. Even that though I think would be best to wait for the refactoring above.

In the interm would you be ok to back out this change and spawn tasks if necessary?

crates/wasi/src/p3/sockets/host/types/tcp.rs
alexcrichton: Oh sorry I somehow missed the `*tcp_state = ..` below. The take/replace pattern seems reasonable to me but I just missed that (not sure how I did...)

crates/wasi/src/p3/sockets/udp.rs
rvolosatovs: I've updated the p3 implementation.

I do want to point out that on Windows at most `i32::MAX` bytes will be written https://github.com/rust-lang/rust/blob/cb6785f73df1aa3f558796a22a4ab9652cf38e26/library/std/src/sys/net/connection/socket.rs#L806 https://github.com/rust-lang/rust/blob/cb6785f73df1aa3f558796a22a4ab9652cf38e26/library/std/src/sys/net/connection/socket/windows.rs#L19

According to the spec, it appears that lists can have up to `u32::MAX` elements (core Wasm https://webassembly.github.io/spec/core/binary/conventions.html#binary-vec and CM https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#container-types). So it appears that at least on Windows a partial write is theoretically possible, but I'm not sure if it could cause an issue in practice for most use cases

badeend: That shouldn't be an issue for UDP. The [spec says](https://github.com/WebAssembly/wasi-sockets/blob/main/wit-0.3.0-draft/types.wit#L605):

>  Implementations may trap if the `data` length exceeds 64 KiB.

alexcrichton (approved):
I had to update the REUSEADDR test a bit more and I'll file a follow-up issue for that, but otherwise this all looks good to me.

PR #11292, opened 2025-07-22
allow configurable forbidden header

xofyarg:
When the HTTP library is not used as a proxy, removing of the forbidden headers by default may not make sense. This change makes it configurable via the `WasiHttpView`, which keeps the default behavior.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -6,6 +6,8 @@ Unreleased.
 
 ### Changed
 
+Users who implemented `WasiHttpView::is_forbidden_header` from `wasmtime-wasi-http` now need to include `DEFAULT_FORBIDDEN_HEADERS`, e.g. `DEFAULT_FORBIDDEN_HEADERS.contains(name) || name.as_str() == "custom-forbidden-header"` #11292
+
 ### Fixed
 
 * Fix a panic in the host caused by preview1 guests using `fd_renumber`.
--- a/crates/wasi-http/src/types.rs
+++ b/crates/wasi-http/src/types.rs
@@ -121,8 +121,8 @@ pub trait WasiHttpView: IoView {
     }
 
     /// Whether a given header should be considered forbidden and not allowed.
-    fn is_forbidden_header(&mut self, _name: &HeaderName) -> bool {
-        false
+    fn is_forbidden_header(&mut self, name: &HeaderName) -> bool {
+        DEFAULT_FORBIDDEN_HEADERS.contains(name)
     }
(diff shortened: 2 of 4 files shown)


xofyarg:
Not sure which is better: [3c9f4c1](https://github.com/bytecodealliance/wasmtime/pull/11292/commits/3c9f4c15ad4a88ba2ff2f4ee2a30e7940cba9c2d) looks clean to me, but it changes previous behavior if user has implemented `is_forbidden_header`; [d183eff](https://github.com/bytecodealliance/wasmtime/pull/11292/commits/d183effc402266006cabc6efbcccb1b4f738a64e) adds an extra method.

rvolosatovs (approved):
In p3 I've fixed this in the following manner: https://github.com/bytecodealliance/wasip3-prototyping/blob/main/crates%2Fwasi-http%2Fsrc%2Fp3%2Fmod.rs#L381-L387

This way users can choose the headers to deny with no breaking changes to the API

That said, IMO this diff is already an improvement

xofyarg:
Should I adjust the change close to p3 then? There are two commits in the PR, I can either drop the 2nd, or squash them together. Please let me know what you think. Thanks.

Edit:

> choose the headers to deny with no breaking changes to the API

The way it works in p2 is, when user override this trait method, the new customized headers as well as the default forbidden headers are denied: https://github.com/bytecodealliance/wasmtime/blob/v34.0.2/crates/wasi-http/src/types.rs#L287

I prefer the way how p3 is implemented, but it means a break change to the current API. Otherwise we have to introduce something new, which will then be dropped when the user upgrade to p3, which also not nice.

xofyarg:
I updated the PR to follow the p3 API, that should give the users a smooth path forward. Although it has a break change for people who override the trait method. The PR is ready to be merged.

rvolosatovs (approved):
LGTM, thank you!
As you correctly point out, this is actually *is* a breaking change for users providing an implementation of the method. Should we record this in release notes? I'm not actually sure about the expectations/procedure here
cc @bytecodealliance/wasmtime-wasi-reviewers

alexcrichton:
Breaking changes are fine yeah, and if you're able to write a note in the release notes that'd be appreciated, but it's not required. When I go back and review merged PRs for the release notes I try to catch issues like this and write them in.

rvolosatovs (approved):

xofyarg:
I put a note in the RELEASES.md file, please take another look.

PR #11304, opened 2025-07-22
Allow compiling debug-builtins in no_std environment to enable gdb jit

dblnz:
## Description
This PR takes care of enabling the compilation of `gdb_jit_int` module in the `jit-debug` crate to ensure the native debugging also works in `no_std` environments.

My aim is to have this in for the LTS version 36.0.0 😃 

This work has been discussed in [#wasmtime > Debugging a wasm module in a hyperlight guest](https://bytecodealliance.zulipchat.com/#narrow/channel/217126-wasmtime/topic/Debugging.20a.20wasm.20module.20in.20a.20hyperlight.20guest).

## Summary
- Make `jit-debug` crate `#![no_std]` when `perf_jitdebug` is not enabled
- Make `gdb_jit_int` module build in `no_std` environment
- Use std when `perf_jitdebug` is enabled because the `perf_jitdebug` module does not run in `no_std` environment
- Split `debug-builtins` feature into `debug-builtins` and `perf-builtins` to individually toggle:
    - the `no_std` `gdb_jit_int` module by using `debug-builtins`
    - the std `perf_jitdebug` module using `perf-builtins`. NOTE: They can be both enabled, but it won't support `no_std`.

### DIFF
--- a/crates/jit-debug/src/gdb_jit_int.rs
+++ b/crates/jit-debug/src/gdb_jit_int.rs
@@ -36,7 +36,7 @@ unsafe extern "C" {
 
 #[cfg(feature = "std")]
 mod gdb_registration {
-    use std::sync::Mutex;
+    use std::sync::{Mutex, MutexGuard};
 
     /// The process controls access to the __jit_debug_descriptor by itself --
     /// the GDB/LLDB accesses this structure and its data at the process startup
@@ -47,25 +47,29 @@ mod gdb_registration {
     pub static GDB_REGISTRATION: Mutex<()> = Mutex::new(());
 
     /// The lock guard for the GDB registration lock.
-    /// The field is not used directly, but it is needed to ensure that the lock
-    /// is held until the `LockGuard` is dropped.
-    #[allow(
+    #[expect(
         dead_code,
         reason = "field used to hold the lock until the end of the scope"
     )]
-    pub struct LockGuard<'a>(std::sync::MutexGuard<'a, ()>);
+    pub struct LockGuard<'a>(MutexGuard<'a, ()>);
 
     pub fn lock() -> LockGuard<'static> {
         LockGuard(GDB_REGISTRATION.lock().unwrap())
     }
 }
 
+/// For no_std there's no access to synchronization primitives so a primitive
+/// fallback for now is to panic-on-contention which is, in theory, rare to come
+/// up as threads are rarer in no_std mode too.
+///
+/// This provides the guarantee that the debugger will see consistent state at
+/// least, but if this panic is hit in practice it'll require some sort of
+/// no_std synchronization mechanism one way or another.
 #[cfg(not(feature = "std"))]
 mod gdb_registration {
     use core::sync::atomic::{AtomicBool, Ordering};
 
-    /// In no_std mode, we use an atomic boolean to control access to the
-    /// __jit_debug_descriptor. This is a simple lock mechanism.
+    /// Whether or not a lock is held or not.
     pub static GDB_REGISTRATION: AtomicBool = AtomicBool::new(false);
 
     /// The lock guard for the GDB registration lock.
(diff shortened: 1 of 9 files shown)


crates/wasmtime/Cargo.toml
```diff
@@ -243,7 +244,8 @@ coredump = ["dep:wasm-encoder", "runtime", "std"]
 
 # Export some symbols from the final binary to assist in debugging
 # Cranelift-generated code with native debuggers like GDB and LLDB.
-debug-builtins = ["dep:wasmtime-jit-debug", "std"]
+debug-builtins = ["wasmtime-jit-debug/gdb_jit_int"]
+perf-builtins = ["wasmtime-jit-debug/perf_jitdump"]
```
alexcrichton: ~~I forget how we ended up at the organization~~

EDIT: hit comment too soon

alexcrichton: I think it'd be best to skip this feature and have this force-enabled as part of the `profiling` feature, this separate feature doesn't actually enable anything otherwise in Wasmtime itself

crates/jit-debug/src/gdb_jit_int.rs
```diff
@@ -2,9 +2,11 @@
...
 //! or unregister generated object images with debuggers.
 
-use std::pin::Pin;
-use std::ptr;
-use std::sync::Mutex;
+use alloc::boxed::Box;
+use alloc::vec::Vec;
+use core::pin::Pin;
+use core::ptr;
+use spin::mutex::Mutex;
```
alexcrichton: This unfortunately is a dependency we cannot use. Spinning on contention is almost never the right concurrency primitive as it's only particularly useful in a kernel setting when interrupts are disabled for example. I also unfortunately do not have a great answer of what to do as an alternative. The best I can think of is "panic on contention" in no_std mode but still use a real `std::sync::Mutex` in std mode.

crates/wasmtime/build.rs
```diff
@@ -27,10 +27,10 @@ fn main() {
     custom_cfg("has_virtual_memory", has_virtual_memory);
     custom_cfg("has_host_compiler_backend", has_host_compiler_backend);
 
-    // If this OS isn't supported or if Cranelift doesn't support the host then
-    // there's no need to build these helpers.
+    // If this OS isn't supported or if Cranelift doesn't support the host or
+    // if no debug-builtins enabled then there's no need to build these helpers.
     #[cfg(feature = "runtime")]
-    if supported_os && has_host_compiler_backend {
+    if (supported_os && has_host_compiler_backend) || cfg!(feature = "debug-builtins") {
```
alexcrichton: Can this condition be massaged a bit to still include the `has_host_compiler_backend` conditional somehow? (or something like that)

Previously it was a nice property that compiling Wasmtime to an unknown/unsupported platform wouldn't involve compiling any C code and I'd ideally like to preserve that. An example situation where C code is not needed is a platform using Pulley even when debug-builtins is enabled because in such a situation the builtins aren't actually used by anything.

crates/jit-debug/src/gdb_jit_int.rs
dblnz: Ok, thanks for the comment. I wasn't sure what the right decision was here.
I think panic in `no_std` mode should be fine, but I'll have another look.

crates/wasmtime/build.rs
dblnz: Hmm, I see, I followed the `debug-builtins` feature and noticed it depends on some symbols from the C, and I thought that for `debug-builtins` this will be needed.
I'll have another look

crates/wasmtime/Cargo.toml
dblnz: The idea I had in mind was that before this change, `debug-builtins` also enabled the perf_jitdump, so when compiling `wasmtime-cli` and providing `--features debug-builtins` now it won't enable the `perf_jitdump` anymore.
I had added a `perf-builtins` so that this option could be enabled from the `wasmtime-cli`.

I have taken your advice and force-enabled `perf_jitdump` as part of `profiling`. Let me know if you'd like to change

alexcrichton:
I've pushed a few commits as feedback here. Mind double-checking them and confirming they're ok for you?

dblnz:
> I've pushed a few commits as feedback here. Mind double-checking them and confirming they're ok for you?

It looks good from my point of view.

Nice touch with the use of optional feature enablement.
```rust
"wasmtime-jit-debug?/std",
```

PR #11318, opened 2025-07-24
Document subtlety of safety conditions for `vm::Instance::sibling_vmctx_mut`

fitzgen:
I initially thought that the documentation was missing a safety invariant that callers must uphold, and then realized that it was a little subtler than I originally thought, so I updated the documentation to clarify this for future readers.


fitzgen:
Follow up nitpick from https://github.com/bytecodealliance/wasmtime/pull/11255

Issue #11285, opened 2025-07-19
Wasmtime exception support needs instance vmctx in each Wasm frame


cfallin:
I have come to a late realization in my quest to implement exception support: we will need to modify the Wasm calling convention to store the vmctx value in each frame. This will unfortunately pessimize common-case execution because it expands each frame by one word (possibly two with padding; 1.5 on average) and adds a store to every prologue.

The reason for this has to do with the nominal aspect of tag identities (see [spec for `throw_ref`](https://webassembly.github.io/exception-handling/core/exec/instructions.html#xref-syntax-instructions-syntax-instr-control-mathsf-throw-ref) for details):

* Tags to identify handlers are *dynamic entities*, instantiated as part of the state of a given instance in the store. The spec refers to "tag addresses" to denote this. Tag instances can be exported and imported just as memories, tables, and globals can be.
* In Wasmtime, we represent tags with `VMTagDefinition`s inline in the vmctx, and `VMTagImport`s that hold pointers to the tag definitions, similarly to memories and tables.
* Exception objects reference tag instances by defining-instance ID and defined tag index in that instance, since we have to ensure that GC heap contents remain untrusted and bounds-checked on use, but this is otherwise equivalent: we are naming the dynamic instance.

When compiling a `try_table`, we emit tag identities for handlers with static `TagIndex`es, and those get serialized into the exception table. My thought has always been that on a throw's stack-walk, we will translate these to dynamic tag instances and compare to the dynamic tag instance in the thrown exception.

The problem is that as we walk the stack, we have PC and FP only; we can map PC to a particular *static module*, but one module may be instantiated multiple times within a store. And each of these instances will have different tag instances (in general) for a given static tag index. The vmctx is saved *somewhere*, but that's up to regalloc, and opaque to our stack-walk. We simply don't have enough information.

In the case where we have a single-instance store, and no imported tags (the former implies the latter actually, because we create dummy instances for host-created tags), we can get around this by comparing static tag indices directly. But that's a subset and we need to support the full spec.

Prior art in other engines seems to be that the instance (vmctx) is available during stackwalking -- e.g., in SpiderMonkey, [caller and callee vmctx are saved in-frame](https://searchfox.org/mozilla-central/rev/0fe32133f12ed0f20d5fde8fe970157f04817982/js/src/wasm/WasmFrame.h#365-366) in every Wasm frame.

[...]

PR #11321, opened 2025-07-24
Cranelift: support dynamic contexts in exception-handler lists.

cfallin:
In #11285, we realized that Wasm semantics require us to match on dynamic instances of exception tags, rather than static tag types. This fundamentally requires the unwinder to be able to resolve the current Wasm instance for each Wasm frame on the stack that has any handlers, and our frame format does not provide this today.

We discussed many options, some of which solve the more general problem (Wasm vmctx for any frame), but ultimately landed on a notion of "dynamic context for evaluating tags", specific to Cranelift's exception-catch metadata; and storing that context and carrying it through to a place that is named in the unwind metadata. The reasoning is fairly straightforward: we cannot afford a more general approach that stores vmctx in every frame (I measured this at 20% overhead for a recursive-Fibonacci benchmark that is call-intensive); and inlining means that we may have *multiple* contexts at any given program point, each associated with a different slice of the handler tags; so we need a mechanism that, *just for a try-call*, intersperses contexts with tags (or puts a context on each tag) and stores these somewhere that the exception-unwind ABI doesn't clobber (e.g., on the stack).

This PR implements "option 4" from that issue, namely, *dynamic exception contexts*. The idea is that this is the dual to exception payload: while payload lets the unwinder communicate state *to* the catching code, context lets the unwinder take state *from* the catching code that lets it decide whether the tag is a match. Because of inlining, we need to either associate (optional) context with every tag, or intersperse context-updates with handler tags. I've opted for the latter for efficiency at the CLIF level (in most cases there will be multiple tags per context), though they are isomorphic.

The new tag-matching semantics are: when walking up the stack, upon reaching a `try_call`, evaluate catch-clauses in listed order. A `context` clause sets the current context. A `tagN: block(...)` clause attempts to match the throwing exception against `tagN`, *evaluated in the current context*, and branches to the named block if it matches. A `default: block(...)` always branches to the named block.

Note that this lets us assume less about tags than before, and this particularly manifests in the changes to the inliner. Whereas before, `tagN` is `tagN` and an inner handler for that tag shadows an outer handler (that is, tags always alias if identical indices); and whereas before, `tagN` is not `tagM` and so we can order the tags arbitrarily (that is, tags never alias if non-identical indices); now any two static tag indices may or may not alias depending on the dynamic context of each. Or, even in the same context, two may alias, because we leave the match-predicate as an unspecified (user-chosen) algorithm during unwinding. (This mirrors the reality that, for example, a Wasm instance may import two tags, and dynamically these tags may be equal or different at runtime, even instantiation-to-instantiation.) Cranelift's only job is to faithfully carry the list of contexts and tags through to the compiled-code metadata; and to ensure that they remain in the order they were specified in the CLIF.

This PR introduces the Cranelift-level feature, and it will be used in a subsequent PR that introduces Wasm exception handling. Because of that, I've opted not to update the clif-utils runtest "runtime" to read out contexts and do something with them -- we will have plenty of test coverage via a bunch of Wasm tests for corner cases such as the above. This PR does include filetests that show that contexts are carried through to spillslots and those appear in the metadata.


cranelift/codegen/src/machinst/buffer.rs
```diff
@@ -1774,16 +1774,79 @@ impl<T: CompilePhase> MachBufferFinalized<T> {
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[cfg_attr(
+    feature = "enable-serde",
+    derive(serde_derive::Serialize, serde_derive::Deserialize)
+)]
+pub enum ExceptionContextLoc {
+    /// An offset from SP at the callsite.
+    SPOffset(u32),
+    /// A GPR at the callsite. The physical register number for the
+    /// GPR register file on the target architecture is used.
```
bjorn3, line 1852: Is this using the same numbering scheme as DWARF debuginfo? If so can this reuse the enum we use for variable location debuginfo? If not what numbering scheme is used?

cfallin: Right now it's a little ad-hoc -- it propagates through whatever numbering scheme we use in `PReg`, which on x86 is the ISA encoding's numbering ([here](https://github.com/bytecodealliance/wasmtime/blob/88079b4f3f47e11c1263bde10b9c30bcbe86cc86/winch/codegen/src/isa/x64/regs.rs#L7)) and on all others mirrors the GPR numbering directly (`N` is `xN` on aarch64, `xN` on riscv64, `rN` on s390x, `xN` on pulley). I was worried about putting an unwind dependency in a core type but perhaps I can pull the enum the other way -- good idea.

cfallin: (err, [here](https://github.com/bytecodealliance/wasmtime/blob/88079b4f3f47e11c1263bde10b9c30bcbe86cc86/cranelift/codegen/src/isa/x64/inst/regs.rs#L15), rather)

cfallin: Hmm, actually, this creates a bunch of new complexity: it requires pulling in the register mapper from the unwind implementation which ordinarily only exists on SysV platforms, conditionalizing this enum arm existence on the `unwind` feature, and some other mess as well. I only added this for genericity based on yesterday's Cranelift meeting (support context in regs/stack in general, assert stack only on the Wasmtime side), but I would rather not build out a bunch of functionality for register-based contexts if it's not actually used; and if it is eventually in Wasmtime, we will likely define our own register mappings. So I am going to go ahead and declare what we have now as Cranelift's "native" dynamic exception context register mapping, and we can refine this later as needed -- or if folks would prefer (@alexcrichton?) I can remove the `GPR` arm here and put the assert and stack-only constraint back into Cranelift.

alexcrichton: Would it be possible to use a literal `PReg` here? That way embedders could remap if they like, but there's a canonical definition for each register (the various backend helpers). Otherwise I'd agree it's not worth too much infrastructure just yet and it's ok to assert with a todo that a register doesn't pop out

cfallin: Perhaps; but we don't export regalloc2 types from the public API anywhere else, nor those backend helpers? The latter especially mean that the `PReg` isn't much more meaningful than the raw `hw_enc` -- that's the integer we are returning currently.

In fullness of time this seems like something an assembler layer should own -- canonical register names and ISA-defined numbering, then we can use this mapping in our public APIs -- but I guess no matter which type we wrap around that index today, the issue is that we don't have any public API surface that defines it.

cfallin: (Except for the unwind stuff, of course, but that has the above platform-dependence and feature-dependence issues)

alexcrichton: Ok makes sense, personally I'm ambivalent about leaving this as `u8` or just removing the variant for now, could go either way.

cfallin: Cool; maybe the deciding factor then is, as long as this PR has the change to rely on the ABI clobbers to force us to the stack case for the tail-call ABI (a good suggestion!), let's keep the register case, so we don't have to fall back to a compile panic or disallow contexts for try-calls in SysV ABI (which are already supported).

cranelift/codegen/src/ir/exception_table.rs
```diff
@@ -41,10 +54,14 @@ pub struct ExceptionTableData {
     /// branch targets for any branch instruction. The last BlockCall
     /// is the normal-return destination, and the rest correspond to
     /// the tags in `tags` below. Thus, we have the invariant that
-    /// `targets.len() == tags.len() + 1`.
+    /// `targets.len() == items.len() + 1`.
```
fitzgen: This length invariant no longer holds because context entries do not have an associated target, right?

cranelift/codegen/src/ir/exception_table.rs
```diff
@@ -127,38 +191,16 @@ impl ExceptionTableData {
...
+    pub fn catches(&self) -> impl Iterator<Item = ExceptionTableItem> + '_ {
+        self.items.iter().map(|item| match item {
+            InternalExceptionTableItem::Tag(tag, target_idx) => {
+                ExceptionTableItem::Tag(*tag, self.targets[usize::try_from(*target_idx).unwrap()])
+            }
+            InternalExceptionTableItem::Default(target_idx) => {
+                ExceptionTableItem::Default(self.targets[usize::try_from(*target_idx).unwrap()])
+            }
+            InternalExceptionTableItem::Context(ctx) => ExceptionTableItem::Context(*ctx),
+        })
```
fitzgen: This method would previously give only exceptional edges, but now gives all kinds of table items. That seems fine if callers were not relying on that property, but we should definitely update the method documentation and probably the method name as well.

cranelift/codegen/src/isa/pulley_shared/inst/emit.rs
```diff
@@ -224,9 +224,9 @@ fn pulley_emit<P>(
             }
 
             if let Some(try_call) = info.try_call_info.as_ref() {
-                sink.add_call_site(&try_call.exception_dests);
+                sink.add_call_site(try_call.exception_handlers(&state.frame_layout));
             } else {
-                sink.add_call_site(&[]);
+                sink.add_call_site(core::iter::empty());
             }
```
fitzgen: Maybe makes sense to define `add_call_site` such that it doesn't take any handlers, and then a new `add_try_call_site` that does take them? Would cut down on the empty iterator argument stuff to clean up all these call sites a little.

Actually, even better would be if we pushed the try call info into the caller, so that call sites just looked like

```rust
sink.add_call_site(info.try_call_info())
```

and then `add_call_site` would internally do the `if let Some(...)`.

cranelift/codegen/src/verifier/mod.rs
```diff
@@ -1413,8 +1414,19 @@ impl<'a> Verifier<'a> {
...
+                        ExceptionTableItem::Tag(_, block_call)
+                        | ExceptionTableItem::Default(block_call) => {
+                            self.typecheck_block_call(
+                                inst,
+                                &block_call,
+                                BlockCallTargetType::Exception,
+                                errors,
+                            )?;
+                        }
+                        ExceptionTableItem::Context(_) => {}
```
fitzgen, line 1428: Double checking my understanding: we don't need to check basic properties of the context value here (like its definition dominates this use) because these checks are handled via the generic methods of getting all uses and defs of a particular instruction elsewhere in the verifier, right?

crates/unwinder/src/exception_table.rs
```diff
@@ -173,6 +214,11 @@ pub struct ExceptionTable<'a> {
     callsites: &'a [U32Bytes<LittleEndian>],
     ranges: &'a [U32Bytes<LittleEndian>],
     tags: &'a [U32Bytes<LittleEndian>],
+    #[allow(
```
fitzgen:
```suggestion
    #[expect(
```

so that it turns into a warning (and CI failure) if we forget to remove this once it is actually used

fitzgen (approved):
Looks great! A couple nitpicks below, r=me with them addressed

cranelift/codegen/src/isa/pulley_shared/inst/emit.rs
cfallin: Went with the former option -- interpreting the `TryCallInfo` to get handlers requires a `FrameLayout` as well, and pushing that into the `MachBuffer` felt like too much coupling (vs. taking an iterator over `MachExceptionHandler`s, which are a buffer-defined concept).

cranelift/codegen/src/verifier/mod.rs
cfallin: Yep, exactly -- the context values are included in the instruction's values iterator (`DataFlowGraph::inst_values()`) now, so are verified to exist and be valid SSA uses that way.

cranelift/codegen/src/ir/exception_table.rs
cfallin: Done!

cranelift/codegen/src/ir/exception_table.rs
cfallin: Updated, thanks!

crates/unwinder/src/exception_table.rs
cfallin: Done!

PR #11323, opened 2025-07-25
wasi-http: Add default-send-request feature flag

brian14708:
This PR adds a `default-send-request` feature flag to the `wasmtime-wasi-http` crate that allows disabling the default HTTP request handler implementation and its associated TLS dependencies.


crates/wasi-http/Cargo.toml
```diff
@@ -11,6 +11,10 @@ description = "Experimental HTTP library for WebAssembly in Wasmtime"
 [lints]
 workspace = true
 
+[features]
+default = ["default-send-request"]
+default-send-request = ["tokio-rustls", "rustls", "webpki-roots"]
```
dicej: Would you mind prefixing each of these with `dep:` (e.g. `"dep:tokio-rustls"` per https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies to avoid possible ambiguity?

crates/wasi-http/src/types.rs
```diff
@@ -117,7 +120,18 @@ pub trait WasiHttpView: IoView {
...
+            Ok(default_send_request(request, config))
+        }
+        #[cfg(not(feature = "default-send-request"))]
+        {
+            let _ = (request, config);
+            Err(crate::bindings::http::types::ErrorCode::InternalError(Some(
+                "default-send-request feature disabled".to_string(),
+            ))
+            .into())
+        }
```
alexcrichton: Personally I don't feel this is the best way to model the behavior here. Could the crate feature be used to control whether this is is a provided function or not? For example without the crate feature this would be an empty function meaning that callers would be required to implement it. With the crate feature though it would be a provided function meaning it has a default implementation.

crates/wasi-http/tests/all/p2.rs
```diff
@@ -67,7 +67,17 @@ impl WasiHttpView for Ctx {
...
+            {
+                Ok(types::default_send_request(request, config))
+            }
+            #[cfg(not(feature = "default-send-request"))]
+            {
+                Err(ErrorCode::InternalError(Some(
+                    "default-send-request feature disabled".to_string(),
+                ))
+                .into())
+            }
```
alexcrichton: FWIW we generally do not test all possible combinations of features, so I think it's reasonable to leave this out and have it be a compile error if the feature is disabled and tested. My thinking is along the lines of my above comment where I'd prefer to avoid reachable always-return-error functions if possible

alexcrichton (commented):
(if you don't mind me doing some review too)

Thanks for this! In addition to the comments below can you add a check on CI that the crate builds without default features? For example somewhere [around here](https://github.com/bytecodealliance/wasmtime/blob/88079b4f3f47e11c1263bde10b9c30bcbe86cc86/.github/workflows/main.yml#L310-L396)

crates/wasi-http/src/types.rs
brian14708: Thanks, missed a `no-default-features` in `wasmtime-cli` otherwise.

PR #11325, opened 2025-07-25
refactor `{Stream,Future}|{Reader,Writer}` APIs and internals

dicej:
This makes a several changes to how `{Stream,Future}|{Reader,Writer}` work to make them more efficient and, in some ways, more ergonomic:

- The background tasks have been removed, allowing reads and writes to complete without task context switching.  We now only allocate and use oneshot channels lazily when the other end is not yet ready; this improves real world performance benchmarks (e.g. wasi-http request handling) considerably.

- Instances of `{Stream,Future}Reader` can now be lifted and lowered directly; no need for `Host{Stream,Future}` anymore.

- The type parameter for `Stream{Reader,Writer}` no longer refers to the buffer type -- just the payload type (i.e. `StreamReader<u8>` instead of `StreamReader<Vec<u8>>`), meaning any buffer type may be used for a given read or write operation.  This also means the compiler needs help with type inference less often when calling `Instance::stream`.

- Instances of `{Stream,Future}|{Reader,Writer}` now require access to the store in order to be disposed of properly.  I've added RAII wrapper structs (`WithAccessor[AndValue]`) to help with this, and also updated `Store::drop` and `Instance::run_concurrent` to ensure the store thread-local is set when dropping futures closing over `&Accessor`s.

- In order to ensure that resources containing `{Stream,Future}|{Reader,Writer}` instances are disposed of properly, I've added `LinkerInstance::resource_concurrent` and have updated `wasmtime-wit-bindgen` to use it.  This gives resource drop functions access to a `StoreContextMut` via an `Accessor`, allowing the stream and future handles to be disposed of.
    - In order to make this work, I had to change `Accessor::instance` from a `Instance` to an `Option<Instance>`, which is awkward but temporary since we're planning to remove `Accessor::instance` entirely once we've moved concurrent state from `ComponentInstance` to `Store`.

That problem of disposal is definitely the most awkward part of all this.  In simple cases, it's easy enough to ensure that read and write handles are disposed of properly, but both `wasmtime-wasi` and `wasmtime-wasi-http` have some pretty complicated functions where handles are passed between tasks and/or stored inside resources, so it can be tricky to ensure proper disposal on all code paths.  I'm open to ideas for improving this, but I suspect we'll need new Rust language features (e.g. linear types) to make it truly ergonomic, robust, and efficient.

While testing the above, I discovered an issue with `Instance::poll_until` such that it would prematurely give up and return a "deadlock" trap error, believing that there was no further work to do, even though the future passed to it was ready to resolve the next time it was polled.  I've fixed this by polling it one last time and only trapping if it returns pending.

Note that I've moved a few associated functions from `ConcurrentState` to `Instance` (e.g. `guest_drop_writable` and others) since they now need access to the store; they're unchanged otherwise.  Apologies for the diff noise.

Finally, I've tweaked how `wasmtime serve` to poll the guest for content before handing the response to Hyper, which helps performance by ensuring the first content chunk can be sent with the same TCP packet as the beginning of the response.

### DIFF
--- a/crates/misc/component-async-tests/src/resource_stream.rs
+++ b/crates/misc/component-async-tests/src/resource_stream.rs
@@ -1,5 +1,7 @@
 use anyhow::Result;
-use wasmtime::component::{Accessor, AccessorTask, HostStream, Resource, StreamWriter};
+use wasmtime::component::{
+    Accessor, AccessorTask, GuardedStreamWriter, Resource, StreamReader, StreamWriter,
+};
 use wasmtime_wasi::p2::IoView;
 
 use super::Ctx;
@@ -36,20 +38,20 @@ impl bindings::local::local::resource_stream::HostWithStore for Ctx {
     async fn foo<T: 'static>(
         accessor: &Accessor<T, Self>,
         count: u32,
-    ) -> wasmtime::Result<HostStream<Resource<ResourceStreamX>>> {
+    ) -> wasmtime::Result<StreamReader<Resource<ResourceStreamX>>> {
         struct Task {
-            tx: StreamWriter<Option<Resource<ResourceStreamX>>>,
+            tx: StreamWriter<Resource<ResourceStreamX>>,
 
             count: u32,
         }
 
         impl<T> AccessorTask<T, Ctx, Result<()>> for Task {
             async fn run(self, accessor: &Accessor<T, Ctx>) -> Result<()> {
-                let mut tx = self.tx;
+                let mut tx = GuardedStreamWriter::new(accessor, self.tx);
                 for _ in 0..self.count {
                     let item =
                         accessor.with(|mut view| view.get().table().push(ResourceStreamX))?;
-                    tx.write_all(accessor, Some(item)).await;
+                    tx.write_all(Some(item)).await;
                 }
                 Ok(())
             }
@@ -57,10 +59,10 @@ impl bindings::local::local::resource_stream::HostWithStore for Ctx {
 
         let (tx, rx) = accessor.with(|mut view| {
             let instance = view.instance();
-            instance.stream::<_, _, Option<_>>(&mut view)
+            instance.stream(&mut view)
         })?;
         accessor.spawn(Task { tx, count });
-        Ok(rx.into())
+        Ok(rx)
     }
 }
 
--- a/crates/wasmtime/src/runtime/component/concurrent.rs
+++ b/crates/wasmtime/src/runtime/component/concurrent.rs
@@ -69,7 +69,8 @@ use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::fmt;
 use std::future::Future;
 use std::marker::PhantomData;
-use std::mem::{self, MaybeUninit};
+use std::mem::{self, ManuallyDrop, MaybeUninit};
+use std::ops::DerefMut;
 use std::pin::{Pin, pin};
 use std::ptr::{self, NonNull};
 use std::slice;
@@ -88,8 +89,9 @@ use wasmtime_environ::component::{
 
 pub use abort::AbortHandle;
 pub use futures_and_streams::{
-    ErrorContext, FutureReader, FutureWriter, HostFuture, HostStream, ReadBuffer, StreamReader,
-    StreamWriter, VecBuffer, Watch, WriteBuffer,
+    ErrorContext, FutureReader, FutureWriter, GuardedFutureReader, GuardedFutureWriter,
+    GuardedStreamReader, GuardedStreamWriter, ReadBuffer, StreamReader, StreamWriter, VecBuffer,
+    WriteBuffer,
 };
 pub(crate) use futures_and_streams::{
     ResourcePair, lower_error_context_to_index, lower_future_to_index, lower_stream_to_index,
@@ -100,7 +102,7 @@ mod error_contexts;
 mod futures_and_streams;
 mod states;
 mod table;
-mod tls;
+pub(crate) mod tls;
 
 /// Constant defined in the Component Model spec to indicate that the async
 /// intrinsic (e.g. `future.write`) has not yet completed.
@@ -228,7 +230,7 @@ where
     where
         T: 'static,
     {
-        self.accessor.instance.spawn_with_accessor(
+        self.accessor.instance.unwrap().spawn_with_accessor(
             self.store.as_context_mut(),
             self.accessor.clone_for_spawn(),
             task,
@@ -328,7 +330,7 @@ where
 {
     token: StoreToken<T>,
     get_data: fn(&mut T) -> D::Data<'_>,
-    instance: Instance,
+    instance: Option<Instance>,
 }
 
 /// A helper trait to take any type of accessor-with-data in functions.
@@ -415,7 +417,7 @@ impl<T> Accessor<T> {
     ///
     /// - `instance`: used to access the `Instance` to which this `Accessor`
     /// (and the future which closes over it) belongs
-    fn new(token: StoreToken<T>, instance: Instance) -> Self {
+    pub(crate) fn new(token: StoreToken<T>, instance: Option<Instance>) -> Self {
         Self {
             token,
             get_data: |x| x,
@@ -506,7 +508,7 @@ where
     where
         T: 'static,
     {
-        let instance = self.instance;
+        let instance = self.instance.unwrap();
         let accessor = self.clone_for_spawn();
         self.with(|mut access| {
             instance.spawn_with_accessor(access.as_context_mut(), accessor, task)
@@ -515,7 +517,7 @@ where
 
     /// Retrieve the component instance of the caller.
     pub fn instance(&self) -> Instance {
-        self.instance
+        self.instance.unwrap()
     }
 
     fn clone_for_spawn(&self) -> Self {
@@ -1291,11 +1293,33 @@ impl Instance {
         let mut store = store.as_context_mut();
         let token = StoreToken::new(store.as_context_mut());
 
-        self.poll_until(store.as_context_mut(), async move {
-            let accessor = Accessor::new(token, self);
-            fun(&accessor).await
-        })
-        .await
+        struct Dropper<'a, T: 'static, V> {
+            store: StoreContextMut<'a, T>,
+            value: ManuallyDrop<V>,
+        }
+
+        impl<'a, T, V> Drop for Dropper<'a, T, V> {
+            fn drop(&mut self) {
+                tls::set(self.store.0.traitobj_mut(), || {
+                    // SAFETY: Here we drop the value without moving it for the
+                    // first and only time -- per the contract for `Drop::drop`,
+                    // this code won't run again, and the `value` field will no
+                    // longer be accessible.
+                    unsafe { ManuallyDrop::drop(&mut self.value) }
+                });
+            }
+        }
+
+        let accessor = &Accessor::new(token, Some(self));
+        let dropper = &mut Dropper {
+            store,
+            value: ManuallyDrop::new(fun(accessor)),
+        };
+        // SAFETY: We never move `dropper` nor its `value` field.
+        let future = unsafe { Pin::new_unchecked(dropper.value.deref_mut()) };
+
+        self.poll_until(dropper.store.as_context_mut(), future)
+            .await
     }
 
     /// Spawn a background task to run as part of this instance's event loop.
@@ -1313,7 +1337,7 @@ impl Instance {
         task: impl AccessorTask<U, HasSelf<U>, Result<()>>,
     ) -> AbortHandle {
         let mut store = store.as_context_mut();
-        let accessor = Accessor::new(StoreToken::new(store.as_context_mut()), self);
+        let accessor = Accessor::new(StoreToken::new(store.as_context_mut()), Some(self));
         self.spawn_with_accessor(store, accessor, task)
     }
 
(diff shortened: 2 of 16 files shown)


crates/wasmtime/src/runtime/component/concurrent.rs
```diff
@@ -1291,11 +1291,32 @@ impl Instance {
...
+        let accessor = &Accessor::new(token, Some(self));
+        let dropper = &mut Dropper {
+            store,
+            value: Some(fun(accessor)),
+        };
+        // SAFETY: `dropper` is a local, non-escaping variable and we do not
+        // move its `value` field until it is dropped.
+        //
+        // TODO: Could/should we make this safe using some combination of `pin!`
+        // and `pin_project!`?
```
alexcrichton: I believe this isn't sound due to the `self.value.take()` in the destructor, notably that moves the future meaning self-poitners are all invalidated, meaning that the `drop` could go wrong. 

Destructors and `Pin` are quite subtle which would make me want to trend more towards a safer version of this rather than having `unsafe` here. The `pin_project_lite!` macro [has support for Drop](https://docs.rs/pin-project-lite/latest/pin_project_lite/macro.pin_project.html#pinned-drop) but unfortunately I'm not sure it would help in this case because that'd just give you `Pin<&mut V>` in the destructor which isn't sufficient for actual destruction.

I know that at least one way to model this is to use `ManuallyDrop` to wrap the field insted of `Option`, meaning that the `Drop for Dropper` would unsafely drop-in-place inside of `tls::set`. That's still unsafe code, however, either with or without `pin_project!`. To me though that's probably the best we can do for now, so how about:

* Add `pin_project_lite` as a dependency of Wasmtime component-model-async and use it here. 
  * Unrelated to this PR we should probably use it for `vm::{Instance, ComponentInstance}` to remove some unsafe code there.
* Use `ManuallyDrop` to store `value` instead of `Option`
* Use `unsafe` code to, in the destructor, run drop-in-place on `Pin<&mut ManuallyDrop<V>>` by unsafely getting access to `&mut ManuallyDrop<V>`.

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -544,79 +623,109 @@ impl<T> FutureWriter<T> {
...
+            self.id,
+            Some(value),
+            TransmitKind::Future,
+        );
+
+        _ = self.just_drop(store);
+
+        result.map(drop)
     }
 }
```
alexcrichton: Something here will need to change, but I'm not exactly sure how and what. I think this is sound but it's panick-y because `host_write` can call `accept_reader` which can lower a value into wasm which isn't valid to do unless we're on a fiber. That runs the risk of suspending Rust code when there's embedder frames on the stack, which ok that's unsound and not just panick-y.

One option is to drop this all entirely and have documentation indicating that a future cannot be closed without sending a value. That's not fantastic, however, as cancellation in Rust has no means of propagating into wasm (e.g. sending a `None` or something like that). 

Another alternative would be to change `host_write` to not execute the lowering immediately but to instead kick of a wasm task in the background of sorts to perform the lowering. Unsure.

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -510,28 +595,22 @@ impl<T> FutureWriter<T> {
...
-        self.default = None;
-        let v = rx.await;
-        drop(self);
-        match v {
+        let accessor = accessor.as_accessor();
+
+        let result = self
+            .instance
+            .host_write_async(accessor, self.id, Some(value), TransmitKind::Future)
+            .await;
```
alexcrichton: This is a bit tricky because this method takes `self`-by-value which means there's no `With*` on the stack that can run the destructor, and as an async cancel point here dropping this future wouldn't actually clean up the future itself.

This sort of relates to the `DropWithStoreAndValue` comment I have below. On one hand this is a fundamental mismatch with "channel semantics" and "that's a future" where things are coming to a clash trying to model WIT futures in Rust in the face of cancellation. On the other hand this is solvable via similar mechanisms to below by retaining the default-value-to-send closure and arranging that to be sent at some point during destruction.

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -544,79 +623,109 @@ impl<T> FutureWriter<T> {
...
-                store.as_context_mut(),
-                self.rep,
-                TransmitKind::Future,
-            )),
+        let accessor = accessor.as_accessor();
+
+        let result = self
+            .instance
+            .host_read_async(accessor, self.id, None, TransmitKind::Future)
+            .await;
```
alexcrichton: Similar to the above, as an async-cancel point this'll need some sort of destructor to run the `self.drop(store)` below.

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -953,78 +989,108 @@ impl<B> StreamWriter<B> {
...
     }
 }
 
-impl<T> Drop for StreamWriter<T> {
-    fn drop(&mut self) {
-        if let Some(mut tx) = self.tx.take() {
-            send(&mut tx, WriteEvent::Drop(None));
-        }
+impl<T> DropWithStore for StreamWriter<T> {
+    fn drop(self, mut store: impl AsContextMut) -> Result<()> {
```
alexcrichton: Could this be surfaced as a `pub fn close(...)` on `StreamWriter` itself? (and the reader, possibly futures too)

Basically I think it'd be good to document this irrespective of the `DropWithStore` impl

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -367,136 +404,184 @@ pub(super) struct FlatAbi {
...
+/// them) which require access to the store in order to be disposed of properly.
+pub trait DropWithStore: Sized {
+    /// Dispose of `self` using the specified store.
+    fn drop(self, store: impl AsContextMut) -> Result<()>;
+
+    /// Dispose of `self` using the specified accessor.
+    fn drop_with(self, accessor: impl AsAccessor) -> Result<()> {
+        accessor.as_accessor().with(move |store| self.drop(store))
+    }
 }
```
alexcrichton: WDYT about changing this to `&mut self`? I feel like that's a bit more idiomatic and maps well to `Drop`?

Also I think this trait will want to be clearly documented in its relation to `Drop`. Specifically this trait implementation does nothing outside of `WithAccessor` and that's required to do anything here.

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -367,136 +404,184 @@ pub(super) struct FlatAbi {
...
-    },
-    /// Drop the read end of the stream or future.
-    Drop,
-    /// Watch the write (i.e. opposite) end of this stream or future, dropping
-    /// the specified sender when it is dropped.
-    Watch { tx: oneshot::Sender<()> },
+/// Trait representing objects (such as the writable end of a future or a struct
+/// containing one) which require access to the store _and_ a value to write in
+/// order to be disposed of properly.
+pub trait DropWithStoreAndValue<T: func::Lower + Send + Sync + 'static>: Sized {
```
alexcrichton: In the abstract I think I personally prefer to keep the closure-to-manufacture-the-value in the constructor of a future rather than having this trait. Would that still work?

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -367,136 +404,184 @@ pub(super) struct FlatAbi {
...
-/// only be used in a context where there is at least one empty spot in the
-/// buffer.  It will silently ignore any other error (e.g. if the `Receiver` has
-/// been dropped).
-fn send<T>(tx: &mut mpsc::Sender<T>, value: T) {
-    if let Err(e) = tx.try_send(value) {
-        if e.is_full() {
-            unreachable!();
+/// This may be used to automatically dispose of the wrapped object when it goes
+/// out of scope.
+pub struct WithAccessor<'a, T: DropWithStore, U: 'static, D: HasData = HasSelf<U>> {
```
alexcrichton: Bikeshedding this a bit, this feels like it should be called `DropWithAccessor` or somehow have "drop" related to the name since "with accessor" is a super generic name that doesn't quite imply what's happening under the hood here. Not that I like the name `DropWithAccessor` though, that also feels bad...

alexcrichton:
This is going to have conflicts with https://github.com/bytecodealliance/wasmtime/pull/11328, and I'll help work through those.

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
dicej: Good point.  Yeah, I think we might need a background task here; I forgot this is one of the reasons I was doing _everything_ on a background task until this PR.

crates/wasmtime/src/runtime/component/concurrent.rs
dicej: I've switched to `ManuallyDrop`.  It turned out `pin-project-lite` didn't buy us much, since we needed to unsafely convert from `Pin<&mut ManuallyDrop<F>>` to `Pin<&mut F>` anyway, so I didn't end up using it.

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
dicej: I've updated this code to use `WithAccessor` and eliminate `DropWithStoreAndValue`.

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
dicej: This is no longer part of the public API, so we can change the name if/when we have a better one, or just not worry about it.

dicej:
I've addressed all the feedback so far except that we're still lowering writes using same stack the host embedder called us on; I'll address that by using a fiber for lowering.

crates/misc/component-async-tests/src/resource_stream.rs
```diff
@@ -36,31 +38,31 @@ impl bindings::local::local::resource_stream::HostWithStore for Ctx {
...
-            tx: StreamWriter<Option<Resource<ResourceStreamX>>>,
+            tx: StreamWriter<Resource<ResourceStreamX>>,
 
             count: u32,
         }
 
         impl<T> AccessorTask<T, Ctx, Result<()>> for Task {
             async fn run(self, accessor: &Accessor<T, Ctx>) -> Result<()> {
-                let mut tx = self.tx;
+                let mut tx = GuardedStreamWriter::new(accessor, self.tx);
```
alexcrichton, line 50: Think about this in the context of rebasing with [wasi:sockets](https://github.com/bytecodealliance/wasmtime/pull/11291) -- is this necessary? Tasks can only be cancelled for two reasons:

1. The store is dropped - in this case there's no need to do further cleanup
2. The `AbortHandle` is triggered, but in this case it's statically known we just forget about that and ignore it.

Given that does WASI actually need `Guarded*`?

dicej: Perhaps we don't _need_ it, but it saves us from having to remember to close the handle explicitly before returning, which isn't too onerous for this simple task, but could become a maintenance hazard later.

`wasmtime-wasi-http` has some pretty complicated control flow, where keeping track of what to close when without RAII would be a real pain.

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
dicej: I just pushed an update to ensure that all host writes use a background task to lower when there may be embedder frames on the stack.

dicej:
@alexcrichton I believe I've addressed all your feedback so far.  Let me know if you have anything else.

BTW, the optimization for lowering flat payloads during host writes is pretty trivial:

```diff
diff --git a/crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs b/crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
index 2809cb6b12..ed56fd7bce 100644
--- a/crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
+++ b/crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
@@ -1777,9 +1777,7 @@ impl Instance {
                     anyhow::Ok(result)
                 };

-                if
-                // TODO: Check if payload is "flat"
-                false {
+                if T::IS_FLAT_TYPE {
                     // Optimize flat payloads (i.e. those which do not require
                     // calling the guest's realloc function) by lowering
                     // directly instead of using a oneshot::channel and
diff --git a/crates/wasmtime/src/runtime/component/func/typed.rs b/crates/wasmtime/src/runtime/component/func/typed.rs
index ad4328fce4..f27120957a 100644
--- a/crates/wasmtime/src/runtime/component/func/typed.rs
+++ b/crates/wasmtime/src/runtime/component/func/typed.rs
@@ -687,6 +687,9 @@ pub unsafe trait ComponentType: Send + Sync {
     #[doc(hidden)]
     const IS_RUST_UNIT_TYPE: bool = false;

+    #[doc(hidden)]
+    const IS_FLAT_TYPE: bool = false;
+
     /// Returns the number of core wasm abi values will be used to represent
     /// this type in its lowered form.
     ///
@@ -1016,6 +1019,8 @@ macro_rules! integers {

             const ABI: CanonicalAbiInfo = CanonicalAbiInfo::$abi;

+            const IS_FLAT_TYPE: bool = true;
+
             fn typecheck(ty: &InterfaceType, _types: &InstanceType<'_>) -> Result<()> {
                 match ty {
                     InterfaceType::$ty => Ok(()),
```

alexcrichton (approved):

PR #11326, opened 2025-07-25
WIP: initial WebAssembly exception-handling support.

cfallin:
This PR introduces support for the [Wasm exception-handling proposal], which introduces a conventional try/catch mechanism to WebAssembly. The PR supports modules that use `try_table` to register handlers for a lexical scope; and provides `throw` and `throw_ref` that allocate (in the first case) and throw exception objects.

This PR builds on top of the work in #10510 for Cranelift-level exception support, #10919 for an unwinder, and #11230 for exception objects built on top of GC, in addition a bunch of smaller fix and enabling PRs around those. It is currently stacked on top of #11321.

This PR does not yet provide host-boundary-crossing exceptions; exceptions that are not caught in a given Wasm activation become traps at the host boundary. That support will come in a subsequent PR.

Because exceptions do not yet cross the host boundary, this also does not yet enable the `assert_exception` wast directive, and so cannot yet support the spec-tests. That will also come in a subsequent PR.

[Wasm exception-handling proposal]: https://github.com/WebAssembly/exception-handling/


cfallin:
Logistical note: I'm posting this as a draft now to get early feedback and because I know folks are waiting to see how it is shaping up. I'm on vacation for two weeks starting now (back Mon Aug 11) and will plan to polish then. I'm hoping to actually get host-boundary integration built as well, if I can, in this PR, to enable spec-tests, but if that turns out to be too much then it will come right after. Following that, fuzzing is the only piece that remains, I think.

crates/cranelift/src/compiler.rs
```diff
@@ -486,6 +494,15 @@ impl wasmtime_environ::Compiler for Compiler {
...
         traps.append_to(obj);
 
+        let exception_section = obj.add_section(
+            obj.segment_name(StandardSegment::Data).to_vec(),
+            ELF_WASMTIME_EXCEPTIONS.as_bytes().to_vec(),
+            SectionKind::ReadOnlyData,
+        );
+        exception_tables.serialize(|bytes| {
+            obj.append_section_data(exception_section, bytes, 1);
+        });
```
alexcrichton, line 582: Could this be skipped if the exception tables section is empty?

src/commands/objdump.rs
```diff
@@ -65,6 +66,10 @@ pub struct ObjdumpCommand {
     /// Whether or not to show information about stack maps.
     #[arg(long, require_equals = true, value_name = "true|false")]
     stack_maps: Option<Option<bool>>,
+
+    /// Whether or not to show information about exception tables.
+    #[arg(long, require_equals = true, value_name = "true|false")]
+    exception_tables: Option<Option<bool>>,
```
alexcrichton, line 72: Mind adding a `tests/disas` test for this showing the exception tables?

crates/wasmtime/src/runtime/vm/throw.rs
```diff
@@ -0,0 +1,131 @@
...
+    vm::{TrapReason, catch_unwind_and_record_trap, raise_preexisting_trap},
+};
+
+/// Implementation of exception throw.
+///
+/// # Safety
+///
+/// Must be invoked when Wasm is in the stack and control has
+/// re-entered the runtime.
+pub unsafe fn throw(nogc: &mut AutoAssertNoGc, exnref: VMExnRef) -> ! {
```
alexcrichton: Could this be made safe and return an action to perform rather than performing the action? I get jittery skipping Rust frames nowadays and I feel it works best when this propagates upwards as far as it can the action to take before actually taking the action. One example is that with Pulley this can't return `!` because it'll need to propagate to the interpreter loop to update interpreter state to perform the pseudo-longjmp

crates/wasmtime/src/runtime/vm/throw.rs
```diff
@@ -0,0 +1,131 @@
...
+                            )
+                            .unwrap()
+                    };
+                    log::trace!("-> read vmctx from frame: {frame_vmctx:x}");
+                    let frame_vmctx =
+                        NonNull::new(frame_vmctx as *mut VMContext).expect("null vmctx in frame");
+
+                    let (handler_tag_instance, handler_tag_index) = unsafe {
+                        InstanceAndStore::from_vmctx(frame_vmctx, |instance| {
+                            let (instance, store) = instance.unpack_mut();
```
alexcrichton: In the long-term I think we'll want to handle this differently because `store` here aliases `nogc` without the compiler knowing. This is probably fine for now but if this sticks around can you file an issue to improve this in the future?

One possible idea would be to move the `InstanceId` into the `VMContext` and to read that out and then lookup through `nogc` the `InstanceId`. Another option would be to add a helper method to `Instance` which takes the store and the `VMContext` and returns `Pin<&mut Instance>` connected to the lifetime of the store, avoiding creating a second store pointer and requiring the store has nothing else borrowed at the same time.

crates/wasmtime/src/runtime/vm/libcalls.rs
```diff
@@ -1543,3 +1543,23 @@ fn cont_new(
...
+    _instance: Pin<&mut Instance>,
+    exnref: u32,
+) -> Result<()> {
+    let exnref = VMGcRef::from_raw_u32(exnref).ok_or_else(|| Trap::NullReference)?;
+    let exnref = exnref
+        .into_exnref(&*store.unwrap_gc_store().gc_heap)
+        .expect("gc ref should be an exception object");
+    unsafe {
+        store.store_opaque_mut().throw_ref(exnref);
+    }
```
alexcrichton: I'm not sure whether it's viable, but personally what I'd ideally like to see is the request-to-throw propagated even through the libcall here through the `Result`. That'd likely require some finesse and refactoring but I suspect such refactoring is going to be required no matter what for Pulley support eventually.

In lieu of that though I'd ideally like to see the "do the throw" action no higher than here, so my comment below about threading the action-to-do would be handled around here. (although I could also be missing other places where it's acted upon)

crates/wasmtime/src/runtime/type_registry.rs
```diff
@@ -1336,6 +1336,23 @@ impl TypeRegistry {
         inner.type_to_gc_layout.get(index).and_then(|l| l.clone())
     }
 
+    /// Borrow the GC layout for teh given index's type.
+    /// Returns `None` for types that do not have GC layouts (i.e. function
+    /// types).
+    pub fn with_layout<R>(
```
alexcrichton: Could you expand the documentation here that this should be used judiciously and specifically mention that the closure `f` is executed under a read lock? That'll help communicate that this can block for awhile waiting for a write lock to finish and additionally this is blocking all writers at the same time.

crates/wasmtime/src/runtime/module.rs
```diff
@@ -1126,6 +1128,13 @@ impl Module {
         let info = self.inner.code.code_memory().stack_map_data();
         wasmtime_environ::StackMap::lookup(text_offset, info)
     }
+
+    /// Obtain an exception-table parser on this module's exception metadata.
+    #[cfg(feature = "gc")]
+    pub(crate) fn exception_table<'a>(&'a self) -> ExceptionTable<'a> {
+        ExceptionTable::parse(self.inner.code.code_memory().exception_tables())
```
alexcrichton, line 1135: To confirm, `parse` here is a pretty cheap operation?

crates/wasmtime/src/runtime/code_memory.rs
```diff
@@ -182,6 +186,11 @@ impl CodeMemory {
         #[cfg(not(has_host_compiler_backend))]
         let _ = &mut unwind;
 
+        // Ensure that the exception table is well-formed. This parser
+        // construction is cheap: it reads the header and validates
+        // ranges but nothing else.
+        let _ = ExceptionTable::parse(&mmap[exception_data.clone()])?;
+
```
alexcrichton: Could this be done with a `#[cfg(debug_assertions)]` perhaps? That'll help avoid paging this in if necessary and by construction this should always be true as a result of compilation such that it shouldn't be required to check here too.

crates/wasmtime/src/config.rs
```diff
@@ -2212,6 +2217,13 @@ impl Config {
             bail!("wmemcheck (memory checker) was requested but is not enabled in this build");
         }
 
+        #[cfg(feature = "gc")]
+        if self.enabled_features.contains(WasmFeatures::EXCEPTIONS)
+            && self.disabled_features.contains(WasmFeatures::GC)
```
alexcrichton: This might be best to switch to `GC_TYPES` instead of `GC` since that's the true reliance here, there's no need to enable structs/arrays to enable exceptions too. (just that under the hood exceptions is implemented with GC types)

crates/environ/src/vmoffsets.rs
```diff
@@ -811,6 +811,18 @@ impl<P: PtrSize> VMOffsets<P> {
...
+    #[inline]
+    pub fn vmtag_import_vmctx(&self) -> u8 {
+        1 * self.pointer_size()
+    }
+
+    /// The offset of the `index` field.
+    #[inline]
+    pub fn vmtag_import_index(&self) -> u8 {
+        2 * self.pointer_size()
+    }
```
alexcrichton, line 825: Mind double-checking there are tests for these new offset methods (and the ones below) in `vmcontext.rs`?

crates/cranelift/src/func_environ/gc/enabled.rs
```diff
@@ -403,6 +403,87 @@ pub fn translate_struct_set(
...
+
+    // Gather accesses first because these require a borrow on
+    // `func_env`, which we later mutate below via
+    // `prepare_gc_ref_access()`.
+    let mut accesses = vec![];
+    for (field_ty, field_layout) in exception_ty.fields.iter().zip(exn_layout.fields.iter()) {
+        accesses.push((field_layout.offset, field_ty.element_type));
+    }
+
+    let mut result = vec![];
```
fitzgen: Might as well `smallvec![]` these temporaries.

crates/cranelift/src/func_environ.rs
```diff
@@ -1644,22 +1710,25 @@ impl<'a, 'func, 'module_env> Call<'a, 'func, 'module_env> {
         (func_addr, callee_vmctx)
     }
 
+    fn caller_vmctx(&self) -> ir::Value {
+        self.builder
+            .func
+            .special_param(ArgumentPurpose::VMContext)
+            .unwrap()
+    }
```
fitzgen, line 2106: I don't think this is correct? I think we mark our own (callee) vmctx as "the" `vmctx` for Cranelift and the caller's vmctx is the second argument that follows afterwards. For example: 

https://github.com/bytecodealliance/wasmtime/blob/1155d6dfbfe859cad85cfe0d5700d4acefd04745/crates/cranelift/src/lib.rs#L145-L160

crates/wasmtime/src/runtime/vm/gc/disabled.rs
```diff
@@ -13,6 +15,28 @@ pub struct VMGcObjectData {
...
+    fn from(_s: VMStructRef) -> VMGcRef {
+        unreachable!()
+    }
+}
+
+impl From<VMExnRef> for VMGcRef {
+    fn from(_e: VMExnRef) -> VMGcRef {
+        unreachable!()
+    }
+}
```
fitzgen: Rather than dynamically assert that these are unreachable at runtime, we can statically assert that they are unreachable at compile time via

```rust
impl From<VMStructRef> for VMGcRef {
    fn from(s: VMStructRef) -> VMGcRef {
        match s {}
    }
}
```

etc...

This is nice because any mistakes are caught at compile time.

crates/wasmtime/src/runtime/vm/throw.rs
fitzgen: > One possible idea would be to move the InstanceId into the VMContext and to read that out and then lookup through nogc the InstanceId. Another option would be to add a helper method to Instance which takes the store and the VMContext and returns Pin<&mut Instance> connected to the lifetime of the store, avoiding creating a second store pointer and requiring the store has nothing else borrowed at the same time.

Another variant on this idea: we could also have a method on the store that takes a vmctx and gives an instance, something like

```rust
impl StoreOpaque {
    // Safety: vmctx must be a valid pointer
    pub unsafe fn instance_from_vmctx(&mut self, vmctx: *mut VMContext) -> Option<Pin<&mut Instance>> {
        // Validate that the vmctx is from this store and all that...
    }
}
```

crates/wasmtime/src/runtime/store.rs
```diff
@@ -2043,6 +2045,18 @@ at https://bytecodealliance.org/security.
         Ok(ptr)
     }
 
+    /// Throws an exception.
+    ///
+    /// # Safety
+    ///
+    /// Must be invoked when Wasm code is on the stack.
```
fitzgen: *and* there are no frames on the stack between the Wasm and here that must run to maintain some safety invariant, and cannot be unwound over.

This is ~impossible to determine in arbitrary code, which is why pushing the command-return all the way out to the libcall is nice, since we know there is ~nothing on the stack before the Wasm frames at that point.

crates/wasmtime/src/runtime/type_registry.rs
fitzgen: Where is this used? I don't actually see it used anywhere and C-f isn't finding anything.

Also, it is probably better, in terms of horizontal scalability, to clone the layout out of the registry than to hold the lock for the duration of whatever code needs to work with it. If that is something we need to do often, we can look into making that cheaper, eg by storing layouts in an `Arc` or something.

fitzgen (commented):
Shaping up nicely!

We had talked elsewhere about removing the exception composite type variant and having tags refer to just their function type (but we would now optionally have a GC struct layout for a function type's parameters for use with exceptions). This would better align us with the Wasm spec and make it so that there are less new additions to the types registry code and also fewer interactions at runtime with its locking and tables and all that. Are you still planning on pursuing this?

cfallin:
> We had talked elsewhere about removing the exception composite type variant and having tags refer to just their function type (but we would now optionally have a GC struct layout for a function type's parameters for use with exceptions). This would better align us with the Wasm spec and make it so that there are less new additions to the types registry code and also fewer interactions at runtime with its locking and tables and all that. Are you still planning on pursuing this?

Ah, sorry, I hadn't made a note in the PR message here, but: I tried and abandoned that path. (Or more precisely, having exception layouts hang off of the function type; `TagType` is already a thin newtype wrapper around `FuncType`.) At a high level, pulling on that string seemed to unwind way too much structure and hit too many places that really still wanted a concrete type for an exception object. If you're curious, my WIP branch is [here](https://github.com/cfallin/wasmtime/tree/wasm-exception-types) (still has many type-errors, mid-refactor). In essence I think that path leads to more complexity, not less, unfortunately.

> also fewer interactions at runtime with its locking and tables and all that

The current implementation performs no locking or accesses to the type registry at runtime; it uses the dynamic context mechanism in Cranelift to get straight to the instance, then look up tags (`VMTagDefinition`s), and compares tag IDs (instance-id/defined-tag-index). In particular, the tag information is at a constant offset in the exception object (similar to array lengths) so we don't need a layout to write the generic path.

(Still on PTO but will respond occasionally to keep review moving)

crates/wasmtime/src/runtime/type_registry.rs
cfallin: Sorry, this is left over from an earlier implementation that fetched layouts dynamically before I had put the tag info at a constant offset in exception objects. (Or rather, it was always at a constant offset per allocator, but I hadn't done the plumbing that array-lengths have to avoid getting it from the layout.)

I'll remove this; but to clarify, we have no need to access layouts or the type registry at all at runtime with the exceptions implementation. The unwinder is completely generic over exception layouts and only compares tag identities.

crates/wasmtime/src/runtime/module.rs
cfallin: Yes -- it reads two `u32`s for lengths of arrays, then builds an `ExceptionTable` with slices over arrays.

crates/cranelift/src/func_environ.rs
cfallin: This is in the context of a `Call` (codegen for a callsite), so the "caller" is the current function; and this change is pulling out an existing expression that was bound as `let caller_vmctx = ...` (see the diff in `unchecked_call_impl` for example). Happy to rename it to something else, but `callee_vmctx` would be inaccurate/misleading in this context, IMHO. `our_vmctx` maybe?

crates/wasmtime/src/runtime/vm/gc/disabled.rs
cfallin: Ah, good call; the first two (methods on `VMGcRef`) can't follow that pattern because `VMGcRef` is not actually an uninhabitable enum in the GC-disabled build, so I followed that in the `From` impls as well, but those can follow the empty-match pattern...

crates/wasmtime/src/runtime/type_registry.rs
cfallin: (Removed in rebase.)

crates/wasmtime/src/config.rs
cfallin: We actually need the GC, though, since we allocate (exception) objects, no?

crates/wasmtime/src/runtime/code_memory.rs
cfallin: Sure, happy to do so. I was following precedent in the rest of this parsing code that returns a clean `Err` on e.g. missing sections; but I agree we don't have to do so, since our overall safety condition is that the loaded code artifact is the exact output that our compile step produced. I'll add a comment noting this.

crates/environ/src/vmoffsets.rs
cfallin: Added!

crates/cranelift/src/func_environ/gc/enabled.rs
cfallin: Done!

crates/cranelift/src/func_environ.rs
cfallin: (resolving but feel free to unresolve if you'd like a different name here)

crates/wasmtime/src/runtime/vm/gc/disabled.rs
cfallin: Done!

crates/wasmtime/src/config.rs
alexcrichton: We need Wasmtime's implementation of a GC yeah, but we don't need the GC-the-wasm-proposal which is what `WasmFeatures::GC` is here. 

Wasmparser's validation is [here](https://github.com/bytecodealliance/wasm-tools/blob/3391156cee997660f992bdd09ca758a429d3a494/crates/wasmparser/src/validator.rs#L288-L292) which effectively equates `GC_TYPES` to "every heap type other than functions/exceptions". That's based on a now-incorrect assumption that exnref wouldn't require a GC in wasmtime. 

On another hand though you can probably delete this entirely. If the exceptions proposal is enabled but `GC` is disabled then we can let the validator figure it out. If `GC_TYPES` is disabled then the program can't use `exnref` anywhere. The only real constraint is that if `EXCEPTIONS` is enabled then `feature = "gc"` should be enabled since our runtime support requires the GC.

crates/wasmtime/src/runtime/vm/throw.rs
cfallin: Addressed in 3c48653a25 -- I went with a variant of `StoreOpaque::instance_from_vmctx`, except with the additional safety requirement that the `vmctx` be from that store. This is valid/satisfiable from the stack walk because all Wasm frames in an activation (up to entry from host) will belong to instances in a single store; and is nice not just for perf reasons but because I was having trouble reasoning about the aliasing implications of creating an instance pointer *first*, and reading out its store pointer, then comparing it to see if we return `Some` or not. (It would be perfectly valid for another thread to own some other `Store`; that thread could mutate the instance pointers freely per Rust's aliasing rules; so I think we can't have the instance pointer then do the check, we have to know by invariant that it belongs to us first; but fortunately we do.)

crates/wasmtime/src/config.rs
cfallin: Ah, right, I had been assuming we separately checked that `feature = "gc"` was enabled when the `GC` feature is enabled at the wasmparser level, then leaning on that as the load-bearing thing to ensure we have `feature = "gc"`. Addressed in a076222f88 -- switched to an explicit check under `not(feature = "gc")` to ensure `EXCEPTIONS` is not enabled.

crates/cranelift/src/compiler.rs
cfallin: It could, but it adds some complication to parsing, and an empty section is 8 bytes of content (plus the section header of 64 bytes) so IMHO not too big a deal?

crates/wasmtime/src/runtime/vm/throw.rs
cfallin: I've been working on this refactor and agree that it's much cleaner, with one caveat: I'm having trouble finding a way to generically provide `catch_unwind_and_record_trap` with the store (which it needs to compute the exception unwind action). It takes a `FnOnce` which ordinarily itself carries the unique ownership of the store. The usual idiom would be to take a `&mut Store` (or moral equivalent) as another arg, and re-pass that into the `FnOnce`; but the various use-sites use either `InstanceAndStore` (libcalls; array-call shim), or `ComponentInstance` (host funcs for the component case), or `Caller` (typed funcrefs). I suppose we could require the closure to return the `HostResult` *and* a mut store borrow of some sort. I'm a little lost with the various wrapper types we have and their invariants, though, and could use some ideas!

My WIP branch for this refactor is [here](https://github.com/cfallin/wasmtime/tree/wasm-exceptions-wip); see last commit.

cfallin: Or said another way -- what I want is I think something like `AsContextMut`, but at the monomorphized level (for `StoreOpaque`). I'm not sure whether that's reasonable to build out, or not. I could use a pairing session to work through this, if someone's up for it (@fitzgen or @alexcrichton?).

EDIT: may have found a way around this, by deferring the exception unwind logic to the actual `CallThreadState::unwind()` point and carrying the exnref from the record-{trap,exception} point until then. Our design *really* doesn't want the store to be accessible from the outermost libcall wrapper but maybe that's fine; save what we have and use the store later...

alexcrichton: I'm not sure I quite follow but sounds like you got it working? Certainly happy to talk in more depth or do a closer review later too

cfallin: *kind* of -- this does mean that we resolve the exception very late, after it is recorded in the `CallThreadState`, because the store is fundamentally owned by the closure so we can't have access to it once we get the `HostResult` (which may be an exception throw). Instead we resolve it when the trampoline re-enters Rust code to call `unwind()`. That implies that we potentially need a new GC root now if a `CallThreadState` exists, though perhaps we can rely on the fact that the "record trap/exception" and "unwind" happen very close to each other and get away with storing a raw `VMExnRef` instead. Happy to hear thoughts from @fitzgen on that one...

fitzgen: Generally it is okay to store a raw GC ref as long as it is within an `AutoAssertNoGc` scope, so that we dynamically catch any errors. But creating one of those requires access to a store, which seems like it is the difficult bit here? I'd have to look more into the exact code to say anything more, maybe we can discuss it in the Cranelift meeting today or something

cfallin: Addressed in 1f9f65ad4f6bd24d1fe15120845689f1e11bbd46!

crates/wasmtime/src/runtime/vm/libcalls.rs
cfallin: Addressed in 1f9f65ad4f6bd24d1fe15120845689f1e11bbd46!

crates/wasmtime/src/runtime/store.rs
cfallin: Addressed in 1f9f65ad4f6bd24d1fe15120845689f1e11bbd46!

src/commands/objdump.rs
cfallin: Done!

cfallin:
@alexcrichton @fitzgen I've now done the refactor we discussed and also added support for Wasm-to-host, host-to-Wasm, and host-to-host (via `Func::new` -> `Func::call`) exception throws. The remaining bit to address is Pulley support (I suspect this should be relatively straightforward now?) and adding `assert_exception` to our WAST runner then enabling spec-tests. I believe the core should be ready for another pass.

crates/environ/src/gc.rs
```diff
@@ -175,30 +175,30 @@ fn common_struct_layout(
         size,
         align,
         fields,
+        is_exception: false,
```
alexcrichton: If it's not too difficult to extract, it might be nice to land the refactoring of how exceptions are allocated/managed separately from this PR which has the other runtime/compiler bits.

crates/environ/src/vmoffsets.rs
```diff
@@ -190,20 +190,26 @@ pub trait PtrSize {
...
+        debug_assert!(offset < self.vmstore_context_last_wasm_exit_trampoline_fp());
         offset
     }
 
+    /// Return the offset of the `last_wasm_exit_trampoline_fp` field
+    /// of `VMStoreContext`.
+    fn vmstore_context_last_wasm_exit_trampoline_fp(&self) -> u8 {
+        self.vmstore_context_gc_heap() + self.size_of_vmmemory_definition()
+    }
+
```
alexcrichton: When updating this, mind updating the comment of the pseudo-struct at the top of this file as well?

Orthogonally, though, I'm a bit concerned about this because it looks like it's basically redundant information in the `VMContext` where we have one field for the fp of the exit trampoline and one field for the fp of the frame before the exit trampoline. That seems like it'd be very easy to confuse the two. Could the previous field be removed to only have the fp of the exit trampoline which can be used to infer the fp of the previous frame of the trampoline?

crates/unwinder/src/stackwalk.rs
```diff
@@ -77,6 +80,18 @@ impl Frame {
...
+    pub fn sp(&self) -> Option<usize> {
+        self.sp
+    }
+
+    /// Read out a machine-word-sized value at the given offset from
+    /// SP in this frame, if the SP is known.
+    pub unsafe fn read_slot(&self, offset: usize) -> Option<usize> {
+        self.sp
+            .map(|sp| unsafe { *(sp.wrapping_add(offset) as *mut usize) })
+    }
```
alexcrichton: Mind adding `# Safety` documentation to the function, as well as `// SAFETY: ...` docs to the `unsafe` block?

crates/wasmtime/src/config.rs
```diff
@@ -1186,7 +1186,12 @@ impl Config {
         self
     }
 
-    #[doc(hidden)] // FIXME(#3427) - if/when implemented then un-hide this
+    /// Configures whether the [Exception-handling proposal][proposal] is enabled or not.
+    ///
+    /// This feature requires GC to be enabled as well.
```
alexcrichton: I think that this isn't quite true any more?

crates/wasmtime/src/runtime/gc/enabled/exnref.rs
```diff
@@ -626,12 +625,20 @@ impl ExnRef {
...
 }
 
+/// Rooted<ExnRef> implements Error so that it can be boxed and
+/// returned in an `anyhow::Result` from host-to-Wasm function calls.
+impl core::error::Error for Rooted<ExnRef> {}
+impl core::fmt::Display for Rooted<ExnRef> {
+    fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+        write!(f, "Wasm exception")
+    }
+}
```
alexcrichton: Instead of implementing these on a public type, could these be implemented on a private internal type? That way we don't have to worry about these details leaking through into the public API

crates/wasmtime/src/runtime/store.rs
```diff
@@ -1309,6 +1320,26 @@ impl StoreOpaque {
...
+    pub unsafe fn instance_from_vmctx(&self, vmctx: NonNull<VMContext>) -> &vm::Instance {
+        // SAFETY: The validity of this `byte_sub` relies on `vmctx`
+        // being a valid allocation which is itself a contract of this
+        // function. Likewise, the `.as_ref()` converts a valid `*mut
+        // Instance` to a `&Instance`.
+        unsafe {
+            vmctx
+                .byte_sub(mem::size_of::<vm::Instance>())
+                .cast::<vm::Instance>()
+                .as_ref()
```
alexcrichton: I think this is copy/pasted from `vm/instance.rs`, could this be encapsulated over there to avoid having the duplication so far from the original source?

crates/wasmtime/src/runtime/store.rs
```diff
@@ -2197,6 +2240,24 @@ at https://bytecodealliance.org/security.
...
         Ok(id)
     }
+
+    /// Set a pending exception. The `exnref` is taken and held on
+    /// this store to be fetched later by an unwind. This method does
+    /// *not* set up an unwind request on the TLS call state; that
+    /// must be done separately.
+    #[cfg(feature = "gc")]
+    pub(crate) fn set_pending_exception(&mut self, exnref: VMExnRef) {
+        self.pending_exception = Some(exnref.into());
```
alexcrichton: Should this `debug_assert!` the previous exception was none?

crates/wasmtime/src/runtime/store.rs
```diff
@@ -2197,6 +2240,24 @@ at https://bytecodealliance.org/security.
...
+    }
+
+    /// Take a pending exception.
+    #[cfg(feature = "gc")]
+    pub(crate) fn take_pending_exception(&mut self) -> VMExnRef {
+        self.pending_exception
+            .take()
+            .expect("No pending exception set")
```
alexcrichton: Could the `expect` here be propagated to the caller? Or otherwise could this have a `# Panics` section of the documentation and be renamed to `unwrap_pending_exception`?

crates/wasmtime/src/runtime/vm/libcalls.rs
```diff
@@ -1548,3 +1548,25 @@ fn cont_new(
...
+unsafe fn get_instance_id(_store: &mut dyn VMStore, instance: Pin<&mut Instance>) -> u32 {
+    instance.id().as_u32()
+}
+
+#[cfg(feature = "gc")]
+unsafe fn throw_ref(
+    mut store: &mut dyn VMStore,
+    _instance: Pin<&mut Instance>,
+    exnref: u32,
+) -> Result<(), TrapReason> {
```
alexcrichton: Could `unsafe` from these two functions be dropped?

crates/wasmtime/src/runtime/vm/libcalls.rs
```diff
@@ -1548,3 +1548,25 @@ fn cont_new(
...
+    use crate::{AsStoreOpaqueMut, vm::ExceptionTombstone};
+
+    let exnref = VMGcRef::from_raw_u32(exnref).ok_or_else(|| Trap::NullReference)?;
+    let exnref = store.unwrap_gc_store_mut().clone_gc_ref(&exnref);
+    let exnref = exnref
+        .into_exnref(&*store.unwrap_gc_store().gc_heap)
+        .expect("gc ref should be an exception object");
+    store.as_store_opaque_mut().set_pending_exception(exnref);
+    Err(TrapReason::User(ExceptionTombstone.into()))
```
alexcrichton: Would it be possible to have `TrapReason::Exception` directly instead of, even at the lowest layers, encoding through the indirection of `anyhow::Error`?

tests/all/main.rs
```diff
@@ -122,6 +123,18 @@ pub(crate) fn gc_store() -> Result<wasmtime::Store<()>> {
...
+    let _ = env_logger::try_init();
+
+    let mut config = wasmtime::Config::new();
+    config.wasm_function_references(true);
+    config.wasm_gc(true);
+    config.wasm_exceptions(true);
+
+    let engine = wasmtime::Engine::new(&config)?;
+    Ok(wasmtime::Store::new(&engine, ()))
+}
```
alexcrichton: Could this be updated to disable `wasm_gc` and `wasm_function_references`, or not explictily enable them, since they shouldn't be required?

crates/wasmtime/src/runtime/vm/traphandlers.rs
```diff
@@ -62,29 +65,37 @@ fn lazy_per_thread_init() {
...
 ///
 /// Only safe to call when wasm code is on the stack, aka `catch_traps` must
 /// have been previously called. Additionally no Rust destructors can be on the
 /// stack. They will be skipped and not executed.
 #[cfg(has_host_compiler_backend)]
-pub(super) unsafe fn raise_preexisting_trap() -> ! {
-    tls::with(|info| unsafe { info.unwrap().unwind() })
+pub(super) unsafe fn raise_preexisting_trap(store: &mut dyn crate::vm::VMStore) -> ! {
+    let mut nogc = AutoAssertNoGc::new(store.store_opaque_mut());
+    tls::with(|info| unsafe { info.unwrap().unwind(&mut nogc) })
```
alexcrichton: I believe this is unsound because there's a rust destructor on the stack (`AutoAssertNoGc`) which is skipped as part of the longjmp and/or unwind resumption.

crates/wasmtime/src/runtime/vm/traphandlers.rs
```diff
@@ -659,33 +781,52 @@ impl CallThreadState {
...
-                    self.capture_coredump(self.vm_store_context.as_ptr(), None),
-                )
+                self.unwind.set(UnwindState::UnwindToHost {
+                    reason: UnwindReason::Trap(trap),
+                    backtrace: self.capture_backtrace(self.vm_store_context.as_ptr(), None),
+                    coredump_stack: self.capture_coredump(self.vm_store_context.as_ptr(), None),
+                });
             }
-        };
-        self.unwind.set(Some((reason, backtrace, coredump)));
```
alexcrichton, line 693: Since `self.unwind` must be set by this function, could this be modeled as:

```rust
let state = match reason {
    // ...
};
self.unwind.set(state);
```

to have compiler checks that every arm sets it?

crates/wasmtime/src/runtime/gc/enabled/exnref.rs
cfallin: The thing is that these *are* necessary for the public API -- the way that a `Func` throws an exception to its caller is to return an `Err` with a boxed `Rooted<ExnRef>`, so we need to implement `Error` (and its required trait `Display` in turn).

Perhaps we could wrap this further, e.g. wrap `Rooted<ExnRef>` in an `Exception` type. The [RFC](https://github.com/bytecodealliance/rfcs/blob/main/accepted/exception-handling.md) describes this as option 4 for the host API -- I had figured that the exnref itself serves that purpose, but we could give it a nicer type for sure. Thoughts?

crates/wasmtime/src/runtime/vm/traphandlers.rs
cfallin: Woah, I'm surprised my tests didn't catch that -- it will leave the no-GC state set, even, keeping the store in a broken state. (I suppose not so surprised, with only short test cases.) Thanks for catching this!

crates/wasmtime/src/runtime/func.rs
```diff
@@ -1265,7 +1269,21 @@ impl Func {
...
+            Ok(()) => {}
+            #[cfg(feature = "gc")]
+            Err(e) if e.is::<Rooted<ExnRef>>() => {
+                let exnref = e.downcast::<Rooted<ExnRef>>().unwrap();
+                let exnref = VMGcRef::from_raw_u32(exnref.to_raw(&mut caller.store.0).unwrap())
+                    .unwrap()
+                    .into_exnref_unchecked();
+                caller.store.0.set_pending_exception(exnref);
+                return Err(ExceptionTombstone.into());
+            }
```
alexcrichton: Structurally something about this is going to need to be adjusted, but I'm not sure how best that would be done. My understanding of this is that anything rooted in `catch_unwind_and_record_trap` effectively needs to have this logic applied. The code here works for `Func::new` and `Linker::func_new`, but there's a lot of other locations that this isn't applied to just yet:

* `Func::new_unchecked` / `Linker::func_new_unchecked`
* `Func::wrap` / `Linker::func_wrap`
* Core wasm libcalls -- unsure how applicable this is, but it's possible to for example have an embedder-defined error throw in `ResourceLimiter` which ostensibly could be an `ExnRef`, so in theory `memory.grow` could be embedder-defined to raise an exception
* Component host functions
* Component libcalls (probably less applicable than core wasm libcalls)

Overall our story for entering/exiting wasm is pretty messy unfortunately. For example one could argue that `CallHook::{CallingHost,ReturningFromHost}` should be invoked for libcalls but they aren't. Otherwise handling that is already duplicated across components and core wasm.

Personally I think the best option is going to be to hook into `catch_unwind_and_record_trap` directly since that's the "narrow waist" for everything anyway. IIRC though you were saying how that's not feasible today since it doesn't have a store nor can easily get a store. I think I see how to do this, so I'll work on a separate PR to thread a store into `catch_unwind_and_record_trap` for this PR to hook into.

crates/environ/src/vmoffsets.rs
cfallin: Great point re: the redundancy -- I was hesitant at first to remove the original Wasm-frame exit FP field, but now I've done that in 3f542bb08e. As a bonus, trampolines are even shorter (no load-from-FP-to-get-Wasm-FP).

Re: comment -- the pseudo-struct comment doesn't describe these fields, only one field for `VMStoreContext`. I did grep for `last_wasm_exit_fp` and try to ensure I updated anything relevant...

crates/wasmtime/src/runtime/func.rs
cfallin: I tried to bottom out all of the ways that host functions could be provided and it did seem to all lead to `HostFunc` -- perhaps I missed somewhere. This is unfortunately the best we can do though unless we do have a store in `catch_unwind_and_record_trap`, as far as I can tell. I agree the whole situation is kind of horrific right now...

crates/unwinder/src/stackwalk.rs
cfallin: Updated!

crates/wasmtime/src/config.rs
cfallin: Updated!

tests/all/main.rs
cfallin: Done!

crates/wasmtime/src/runtime/func.rs
alexcrichton: IMO this required duplication if we don't have a store when handling traps is motivation to push harder on getting a store available while handling traps. I've done this refactoring https://github.com/bytecodealliance/wasmtime/pull/11441 which I believe means that this code here can be scoped to `HostResult` implementations which can now easily access the store.

cfallin:
I've refactored on top of #11441 and this is indeed much nicer; I've removed the `From` impls that would otherwise pass through boxed exceptions from hostcalls without converting to a pending exception, so I'm more confident we won't miss any additional hostcall paths now.

The current challenge is dealing with rooting: the factor has placed the point at which we capture the `Rooted<ExnRef>` (refactor to wrap that up in a `wasmtime::Exception` TBD) outside the root handle scope for the hostcall, so any newly allocated exception from the host becomes unrooted. Pulling this string further may require additional refactors to put the LIFO-scope management in common across all hostcalls; just starting to look into this. More thoughts (or a pairing session if that'd be easier) welcome, @alexcrichton...

crates/wasmtime/src/runtime/func.rs
cfallin: Resolved in refactor (modulo rooting issues below)!

crates/wasmtime/src/runtime/vm/traphandlers.rs
cfallin: Done!

crates/wasmtime/src/runtime/vm/traphandlers.rs
cfallin: Fixed!

crates/wasmtime/src/runtime/vm/libcalls.rs
cfallin: Yep, this makes more sense -- removed now in refactor.

crates/wasmtime/src/runtime/vm/libcalls.rs
cfallin: Done!

crates/wasmtime/src/runtime/store.rs
cfallin: Done!

alexcrichton:
Oh good point! I think it'd be reasonable to move [this management](https://github.com/bytecodealliance/wasmtime/blob/6f6a514b717d323f80d4a810756fb62212d1cc8e/crates/wasmtime/src/runtime/func.rs#L2066-L2079) into the trap-handling bits (or around the `enter_host_from_wasm` bits). That future-proofs this for component-model-gc support and additionally enables eventually cleaning up libcalls that currently create their own scopes as they no longer need to

cfallin:
Actually, I think there are deeper issues with rooting, exceptions, and `Result` / `?` ergonomics that we didn't really address in the RFC, and I'm finding I'm bumping into these design questions now. The current state of this PR has working-but-for-unrooted-ref-panics examples of all directions of exception transfer in `cargo test --test all -- exceptions`, demonstrating a few examples of the issue.

The basic question is: if we return an exception as a GC object boxed in a `Rooted<ExnRef>` (eventually inside a nicely-packaged `wasmtime::Exception` type) in the `Err` arm of a `Result`, how should we expect this to interact with handle scopes?

The most basic example is a host function that allocates an `ExnRef` and returns `Err(exnref.into())`, and has no other scopes of its own. We can fix up this PR by moving the root scope teardown outside of the point that we root a pending exception on the store, so that can work fine.

But in the general case, with user host code creating nested `RootScope`s, we cannot really expect automatic `?` passing-up-the-`Err` to work well. Consider:

```
fn my_hostcall(...) -> Result<()> {
  let scope = RootScope::new(&mut caller);
  helper(&mut scope)?;
  Ok(())
}

fn helper(ctx: impl AsContextMut<'_>, ...) -> Result<()> {
  return Err(ExnRef::new(...).into());
}
```

This will typecheck fine (`ExnRef::new()` returns a `Rooted<ExnRef>`, we currently take that as an `Error` impl directly, we can wrap it up in `wasmtime::Exception` once I do that); but the exception object will be unrooted once the `?` propagates it past the user's own `RootScope`.

More generally, `Err`-carried types should generally expect to *own* their error information. Our lack of lifetimes to denote valid scope is an explicit ergonomic choice, which is great in general, but hides the fact that we will get a dynamic scope-violation (unrooting) error here.

Should we use a `ManuallyRooted` instead? I think that's much worse -- it's way too easy to leak, since its `Drop` impl doesn't unroot (because it can't, because it doesn't have a mut borrow of the `Store` -- again a totally valid choice that is nice in 99% of cases).

It's somewhat tempting to say "don't do that" to all of this -- it's no different than returning any GC ref type at all. If we take this option, we just fix the exact extent of the root scope when calling out to the host, we root any returned exception when it comes back (concretely: put the `HostResult` conversion *inside* the root scope somehow, probably by pulling the scope out to `catch_and_record_traps`). My broader concern here is that I'm not sure that it has the nice error-propagation properties that we expected. Basically, it's a nonlocal/non-composable footgun: anyone creating a scope anywhere inside complex host logic could cause an exception created on deep to blow up dynamically with a panic.

One alternative is to have an explicit `store.throw(exn)` as part of the public API, and then provide a *public-facing* tombstone type that the user can propagate upward through `Result`s. This does make things a little weird in other ways, because it's implicit state (we could have an API around "cancel pending exception" too if someone wants to "catch" it in host code), but it feels the least footgunny in a lot of others. Basically it's extending what we decided to do internally (root the exception on the store, safely) to user code.

The last option does differ from what we agreed to in the RFC (but the RFC is underspecified with respect to rooting behavior in general, on the other hand) -- so I'd want some consensus here, at least, before building that. Any thoughts?

alexcrichton:
Good points!

> Basically, it's a nonlocal/non-composable footgun

To some degree this is inherently true of `Rooted` and `ManuallyRooted`. As you point out there's big downsides to using both. That being said I do agree this is particularly exacerbated due to the nature of "just propagate the error upwards please" which makes it more likely an exception might fly all the way up past whatever scopes are in use.

> One alternative is to have an explicit store.throw(exn) as part of the public API

I like this approach personally. What I might recommend is something like `StoreContextMut::throw(&mut self, &Rooted<ExnRef>) -> wasmtime::ThrownException` so that way the user doesn't have to create the tombstone themselves and can immediately return that error upwards the stack.

cfallin:
> `wasmtime::ThrownException` so that way the user doesn't have to create the tombstone themselves and can immediately return that error upwards the stack.

Indeed, that was my thought as well. I suppose one could actually also have a signature `throw(...) -> Result<Uninhabited, ThrownException>` so that one could do `store.throw(exn)?;` -- or both, `thrown_exception(...) -> ThrownException` and `throw(...)` -- but at this point I'm bikeshedding a bit. I'll prototype this to see how it feels.

alexcrichton:
I'd recommend choosing between:

* `fn throw(&mut self, exn: &Rooted<ExnRef>) -> SomeExceptionType;`
* `fn throw<T>(&mut self, exn: &Rooted<ExnRef>) -> Result<T, SomeExceptionType>;`

where the latter uses a generic `T` instead of `Uninhabited` so that way it can be coerced to any possible type without casts/etc. I'd be on the fence myself about these two signatures, I agree that using `?` would be nice. One possible compromise could be a `Result`-based `throw` for convenience and a method on `ExnRef` for creating `SomeExceptionType`, so that way we have both and whatever's most appropriate can be used?

cfallin:
Alright, I've built this out -- the API on `Store` and friends now has a `throw` as described, and in addition to the lower-level `{take,set,has}_pending_exception()`, also has a `catch` helper so one can do something like

```
store.catch(|store| {
    let exn = ...;
    store.throw(exn)?;
    Ok(())
  },
  |store, exn| {
    Err("caught an exception".into())
  })?;
```

or similar.

---

This just needs
- [x] ~~Pulley support~~
- [x] WAST driver `assert_exception` support
- [x] turn on spec tests

then I'll flip off the "draft" flag.

crates/wasmtime/src/runtime/gc/enabled/exnref.rs
cfallin: Resolved in latest refactor.

crates/wasmtime/src/runtime/store.rs
cfallin: Done!

crates/wasmtime/src/runtime/store.rs
cfallin: Done in 45415e4886!

cfallin:
New final boss appears: Pulley support will, I think, require refactoring how setjmp/longjmp is emulated. Right now Pulley [special-cases the `raise()` libcall](https://github.com/bytecodealliance/wasmtime/blob/f8177c201e466ace97a3ce00be69adae8627ddd1/crates/wasmtime/src/runtime/vm/interpreter.rs#L237-L240) and does an interpreter-register-level longjmp across all Wasm frames.

I believe what we'll ideally want to do instead, to avoid too much duplication in an alternative `raise()` impl, is to actually invoke the real implementation but then indirect the longjmp through a match on the store's executor: either a real longjmp to the real `jmp_buf` in the `CallThreadState`, or a mutation of the interpreter state and a return. Then similarly `wasmtime_unwinder::resume_to_exception_handler` can have another "host" backend that takes the store, fetches the interpreter state, and updates it.

The alternative is to try to get access to the store inside the Pulley `CallIndirectHost` impl and determine its `UnwindState`, but that seems to pull in too much across abstraction boundaries -- for example, it requires Pulley to take the exception payload off the store's root (i.e., have access to the full `Store`, not just some slice of it).

I'm tempted to try to land this first and defer Pulley support to a separate PR, with some help (probably from @alexcrichton) to determine the best path here -- it's somewhat close (the stackwalk works fine!) but definitely not trivial.

fitzgen:
> I'm tempted to try to land this first and defer Pulley support to a separate PR, with some help (probably from @alexcrichton) to determine the best path here -- it's somewhat close (the stackwalk works fine!) but definitely not trivial.

SGTM, we have a history of landing support for new Wasm proposals a backend at a time (e.g. tail calls, ref types) and as long as there isn't a fundamental reason an implementation approach can't be extended to the unsupported backends that would require redoing the supported backends, then I don't see any reason for disallowing incremental PRs.

alexcrichton:
Agreed yeah definitely feel free to defer Pulley, I'm happy to dive in afterwards and/or have a call to discuss how best to get it working

cfallin:
The full test suite is passing now!

There's quite a lot of rebase churn with the ongoing refactors on main -- to avoid too much rebase pain I'm going to re-squash down to one commit then rebase that as a unit (rather than fixing conflicts at each of N steps). Then I'll see how easy it is to pull out the refactor to exception object layouts (to have one tag slot and share more with structs) -- no guarantees on that. Either way, will mark ready once it's static.

crates/environ/src/gc.rs
cfallin: Done (#11467)!

cfallin:
I've now rebased (twice -- #11461 created more conflicts -- hopefully this lands soon!) and marked as ready. This PR is stacked on top of #11467. All spec-tests should be passing (will watch CI to make sure it goes green here). Thanks for the feedback so far!

crates/wasmtime/src/config.rs
```diff
@@ -2234,6 +2244,11 @@ impl Config {
             bail!("support for GC was disabled at compile time")
         }
 
+        #[cfg(not(feature = "gc"))]
+        if self.enabled_features.contains(WasmFeatures::EXCEPTIONS) {
```
alexcrichton: This should test `features`-the-local-variable since that represents the calculation of wasm features after enabled/disabled/defaults/etc.

Also mind folding the `#[cfg]` into the boolean expression with `!cfg!(feature = "gc")`? I personally prefer to favor `cfg!` where possible since it avoids dead code warnings and such (not that such motivations are applicable here)

crates/wasmtime/src/runtime/code_memory.rs
```diff
@@ -182,6 +186,16 @@ impl CodeMemory {
...
         let _ = &mut unwind;
 
+        // Ensure that the exception table is well-formed. This parser
+        // construction is cheap: it reads the header and validates
+        // ranges but nothing else. We do this only in debug-assertion
+        // builds because we otherwise require for safety that the
+        // compiled artifact is as-produced-by this version of
+        // Wasmtime, and we should always produce a correct exception
+        // table (i.e., we are not expecting untrusted data here).
+        #[cfg(debug_assertions)]
```
alexcrichton: Similar to above, could this use `if cfg!(debug_assertions) { ... }` to avoid needing a conditional import on `ExceptionTable`? (this'll show up in CI for release builds I think which aren't enabled by default on PRs)

crates/wasmtime/src/runtime/exception.rs
```diff
@@ -0,0 +1,52 @@
...
+#[derive(Debug)]
+pub struct ThrownException;
+
+/// We need to implement Error for `ThrownException` so it can be boxed up into an `anyhow::Error`.
+impl core::error::Error for ThrownException {}
+
+/// `Error` requires `Display`.
+impl core::fmt::Display for ThrownException {
+    fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+        <Self as core::fmt::Debug>::fmt(self, f)
```
alexcrichton: Personally I'd be in favor of "thrown wasm exception" or something like that as opposed to the generic `ThrownException` this'll otherwise print

crates/wasmtime/src/runtime/store.rs
```diff
@@ -1047,6 +1067,105 @@ impl<T> Store<T> {
...
+                let mut scope = crate::RootScope::new(self);
+                let exn = scope.as_context_mut().take_pending_exception().unwrap();
+                handler(scope.as_context_mut(), exn)
+            }
+            ret @ Ok(_) | ret @ Err(_) => {
+                debug_assert!(!self.has_pending_exception());
+                ret
+            }
+        }
+    }
```
alexcrichton: Given that we need to in theory replicate these methods on `StoreContextMut`, `Store`, `Caller`, and other things that implement `AsContextMut`, and given that in theory we also need to eventually add `catch_async` in addition to `catch`, I'd subjectively say this is probably fine to relegate to examples or documentation but otherwise not live here as a first-class function.

crates/wasmtime/src/runtime/store.rs
```diff
@@ -2264,6 +2424,41 @@ at https://bytecodealliance.org/security.
...
+
+    /// Take a pending exception, if any.
+    #[cfg(feature = "gc")]
+    pub(crate) fn take_pending_exception(&mut self) -> Option<VMExnRef> {
+        self.pending_exception.take()
+    }
+
+    #[cfg(feature = "gc")]
+    fn take_pending_exception_rooted(&mut self) -> Option<Rooted<ExnRef>> {
+        self.take_pending_exception().map(|vmexnref| {
```
alexcrichton: You can use `?` after `self.take_pending_exception()` to avoid indentation if you'd like

crates/wasmtime/src/runtime/store.rs
```diff
@@ -2264,6 +2424,41 @@ at https://bytecodealliance.org/security.
...
+            let mut nogc = AutoAssertNoGc::new(self);
+            let cloned = vmexnref.clone(nogc.require_gc_store_mut().unwrap());
+            ExnRef::from_cloned_gc_ref(&mut nogc, cloned.into())
```
alexcrichton: Is `clone` the right operation to call here? I would have expected we just thread through the `VMGcRef` as-is because we're effectively moving ownership into the return value

crates/wasmtime/src/runtime/vm/traphandlers.rs
```diff
@@ -362,11 +384,32 @@ pub enum TrapReason {
...
+            TrapReason::Exception
+        } else {
+            TrapReason::User(error)
+        }
+    }
+
+    #[cfg(not(feature = "gc"))]
+    fn from(error: Error) -> Self {
+        TrapReason::User(error)
     }
```
alexcrichton: FWIW this is an example of where in my subjective opinion `#[cfg]` is overkill. For example if we have `ThrownException`, a unit struct, in the public API when the `gc` feature is disabled that seems fine to me and worth it to avoid growing more `#[cfg]`

crates/wasmtime/src/runtime/vm/traphandlers.rs
```diff
@@ -387,7 +430,7 @@ pub unsafe fn catch_traps<T, F>(
     store: &mut StoreContextMut<'_, T>,
     old_state: &mut EntryStoreContext,
     mut closure: F,
-) -> Result<(), Box<Trap>>
+) -> Result<()>
```
alexcrichton, line 428: With the changes here and to the `match` below, could you run the `call.rs` benchmark before/after this PR with the `^sync.*no-hook.*host-to-wasm.* typed - nop$` regex? If there's no real change, yay! If there's a regression, mind spot-checking to see if anything jumps out?

crates/wasmtime/src/runtime/vm/traphandlers.rs
```diff
@@ -704,11 +868,43 @@ impl CallThreadState {
     /// called. Additionally this isn't safe as it will skip all Rust
     /// destructors on the stack, if there are any.
     #[cfg(has_host_compiler_backend)]
-    unsafe fn unwind(&self) -> ! {
-        debug_assert!(!self.jmp_buf.get().is_null());
-        debug_assert!(self.jmp_buf.get() != CallThreadState::JMP_BUF_INTERPRETER_SENTINEL);
-        unsafe {
-            traphandlers::wasmtime_longjmp(self.jmp_buf.get());
+    unsafe fn unwind(&self, store: &mut dyn VMStore) -> ! {
```
alexcrichton, line 870: Would it be possible to change this to `Option<usize>` representing the payload? That way `UnwindToHost` would assert it's `None` and `UnwindToWasm` would assert it's `Some`.

crates/wast/src/wast.rs
```diff
@@ -479,6 +479,21 @@ where
...
 
+    fn assert_exception(&mut self, result: Outcome) -> Result<()> {
+        match result {
+            Outcome::Ok(values) => bail!("expected exception, got {values:?}"),
+            Outcome::Trap(err) if err.is::<ThrownException>() => {
+                // Discard the thrown exception.
+                let _ = self
+                    .store
+                    .take_pending_exception()
+                    .expect("there should be a pending exception on the store");
```
alexcrichton, line 490: Reading this over, the semantics of `set_pending_exception` are such that it panics if the exception is already set, right? Would that mean that if an embedder forgot to call `take_pending_exception` and then reentered wasm after an exception, would that panic on another exception?

If not, yay! If so, I might say we should instead allow `set_pending_exception` with whatever and it just overwrites whatever was there previously.

alexcrichton (commented):
Mind updating a few bits and pieces of documentation as well with this? The ones I can think of are:

* [The wasm proposals table](https://github.com/bytecodealliance/wasmtime/blob/main/docs/stability-wasm-proposals.md#wasm-proposals)
* [The architecture table for supported wasm proposals](https://github.com/bytecodealliance/wasmtime/blob/main/docs/stability-tiers.md#x86_64)

Also, apart from Pulley and fuzzing possibly, are there other known entities to record in issues before landing? E.g. anything like an open question about performance, API, etc.

I should also clarify that I so far haven't looked at the wasmtime-cranelift side of things at all, I figured I'd defer to @fitzgen for that, but Nick if you'd like I'm happy to review it here too.

crates/cranelift/src/func_environ/gc/enabled/drc.rs
```diff
@@ -498,6 +496,83 @@ impl GcCompiler for DrcCompiler {
...
+        self.init_field(
+            func_env,
+            builder,
+            instance_id_addr,
+            WasmStorageType::Val(WasmValType::I32),
+            instance_id,
+        )?;
+        let tag_addr = builder
+            .ins()
+            .iadd_imm(raw_ptr_to_exn, i64::from(tag_offset + 4));
```
fitzgen: It is pretty funky and potentially footgun-y that we have `tag_offset` as the offset of the instance id and then do `tag_offset + 4` to get the tag offset. Can we use proper constants and/or `VMOffsets` methods for everything here (and also have tests asserting that they are correct, if we are missing them).

crates/wasmtime/src/runtime/vm/traphandlers.rs
cfallin: Hmm, I pulled on this thread a bit, and adding a `ThrownException` unit struct (or uninhabitable enum) allows this to clean up, but then starts pulling a bunch of other things out of `gc`-only config as well (`TrapReason`, more unwind code depending on where we want to draw the boundary...) and honestly it seems cleaner to me to take the sharp line of types-and-enum-arms-don't-exist-for-exceptions in a GC-disabled build? Otherwise there are a bunch of dynamically unreachable failure points, which seems awkward in its own way...

crates/cranelift/src/func_environ/gc/enabled/drc.rs
fitzgen: I think we will want to have both the existing `GcLayouts::exception_tag_offset` method and a new `GcLayouts::exception_instance_idx_offset` method, and then use those things here (or the collector-specific constants that those things are implemented with here).

crates/cranelift/src/func_environ/gc/enabled.rs
```diff
@@ -409,6 +412,119 @@ pub fn translate_struct_set(
     Ok(())
 }
 
+pub fn translate_exn_unbox(
+    func_env: &mut FuncEnvironment<'_>,
+    builder: &mut FunctionBuilder<'_>,
+    tag_index: TagIndex,
+    exn_ref: ir::Value,
+) -> WasmResult<SmallVec<[ir::Value; 4]>> {
+    log::trace!("translate_exn_unbox({tag_index:?} {exn_ref:?})");
```
fitzgen:
```suggestion
    log::trace!("translate_exn_unbox({tag_index:?}, {exn_ref:?})");
```

crates/wast/src/wast.rs
cfallin: We did indeed have "panic on re-setting pending exception", due to [your review comment here](https://github.com/bytecodealliance/wasmtime/pull/11326#discussion_r2279374762) suggesting that we `debug_assert` to ensure no prior exception. That said, now that we have the "pending exception slot" API, this is user-exposed and clearly a different tradeoff, so I've removed that assert.

crates/wasmtime/src/runtime/store.rs
cfallin: Definitely not, it turns out! I was playing Type Tetris with the APIs and this was the best I had found, but it turns out `Rooted::new` is exactly what I wanted to root an existing raw ref. Updated.

crates/wasmtime/src/runtime/store.rs
cfallin: Done!

crates/wasmtime/src/runtime/store.rs
cfallin: Sure, removed for now -- always possible to add later if it turns out to be a real ergonomic need...

crates/wasmtime/src/runtime/exception.rs
cfallin: Done!

crates/wasmtime/src/runtime/code_memory.rs
cfallin: Done!

crates/wasmtime/src/config.rs
cfallin: Done!

crates/cranelift/src/translate/code_translator.rs
```diff
@@ -4132,3 +4189,74 @@ fn bitcast_wasm_params(
...
+        32 => (wasmparser::ValType::I32, I32),
+        64 => (wasmparser::ValType::I64, I64),
+        _ => panic!("Unsupported pointer width"),
+    };
+    let block = block_with_params(builder, [exn_payload_wasm_ty], environ)?;
+    builder.switch_to_block(block);
+    let exn_ref = builder.func.dfg.block_params(block)[0];
+    if needs_stack_map {
+        builder.declare_value_needs_stack_map(exn_ref);
+    }
```
fitzgen: I think this should be on the maybe-`ireduce`d value, not the direct payload. Don't have a link to the code off-hand but I am pretty sure we assert somewhere that we only have stack maps for `i32` types.

fitzgen: Yeah:

https://github.com/bytecodealliance/wasmtime/blob/c645077df15584651f115ae9c909072cd699ef28/crates/cranelift/src/compiler.rs#L1314-L1329

crates/wasmtime/src/runtime/vm/traphandlers.rs
cfallin: I see a ~10% regression (10.8ns -> 12.1ns) on core Wasm host-to-Wasm calls, and an in-the-noise / "0.4% speedup" on component host-to-Wasm calls. I don't see any obvious reason that would cause this for non-trapping/throwing calls, but given a little more plumbing in general I'm also not too surprised. Happy to try digging in further if needed...

crates/wasmtime/src/runtime/vm/traphandlers.rs
cfallin: This is slightly tricky as it would then imply the caller of `unwind` knows when to provide the payload -- the protocol is that the pending exception remains on the `Store` if we're `UnwindToHost`'ing, and not if we're `UnwindToWasm`'ing. IMHO I'd rather keep the knowledge of `self.unwind` enum arms local to this file, but open to other ideas if you have any.

cfallin:
> Mind updating a few bits and pieces of documentation as well with this? The ones I can think of are:
> 
>     * [The wasm proposals table](https://github.com/bytecodealliance/wasmtime/blob/main/docs/stability-wasm-proposals.md#wasm-proposals)
> 
>     * [The architecture table for supported wasm proposals](https://github.com/bytecodealliance/wasmtime/blob/main/docs/stability-tiers.md#x86_64)
> 
> 
> Also, apart from Pulley and fuzzing possibly, are there other known entities to record in issues before landing? E.g. anything like an open question about performance, API, etc.

Docs updated!

The two major remaining bits are Pulley support and ensuring exception support is fuzzed. I'll work on at least the latter, and I'll file issues for both once this merges.

All comments addressed modulo those still open above -- thanks!

crates/cranelift/src/translate/code_translator.rs
cfallin: Fixed!

crates/cranelift/src/func_environ/gc/enabled.rs
cfallin: Done!

crates/cranelift/src/func_environ/gc/enabled/drc.rs
cfallin: Good point -- fixed!

cfallin:
@alexcrichton I think this is ready for you again on the three open threads above; and a final LGTM or any other issues you see on the runtime side. Thanks!

crates/wasmtime/src/runtime/vm/traphandlers.rs
alexcrichton: No I agree that we should avoid `unreachable!()`-behind-cfg or things like that as agreed it gets tough to reason about "why again isn't this dynamically reachable?" In that sense I agree that `TrapReason::Exception` should stay #[cfg]'d away.

In retrospect I don't think I fully thought this through. My main reaction was duplication of the function signature and fallback to `TrapReason::User`. This could be "solved" with something like:

```rust
fn from(error: Error) -> Self {
    #[cfg(feature = "gc")]
    { /* test for `ThrownException` and return it */ }

    TrapReason::user(error)
}
```

which avoids duplicating the function signature and keeps the gc code "on the side". Not perfect in the sense of it still has #[cfg], but this is all quite minor and not as actionable as I originally thought it was, so I'd say let's conclude "just ignore alex here"

crates/wasmtime/src/runtime/vm/traphandlers.rs
alexcrichton: That's reasonable enough for now IMO. I suspect we might be able to recover that by shuffling around things, maybe adding `#[cold]`, tweaking slow paths, etc. That being said it's still solidly in the realm of "this shouldn't matter and it's still fast", so I think it's fine to defer such theoretical tweaks to a future point in time.

crates/wasmtime/src/runtime/vm/traphandlers.rs
alexcrichton: Nah that seems reasonable, this is a feeble attempt by me to future-proof Pulley support a bit more. Let's just defer and deal with it in total later.

alexcrichton (approved):
All sounds reasonable to me!

I suspect this is likely going to bounce on CI at least once, and for that I'd recommend adding `prtest:full` to the commit messsage to get full CI here on the PR as that should give a much faster cycle time than going through the merge queue. Although that's all moot if it sails through the merge queue on the first try.

crates/wasmtime/src/runtime/vm/traphandlers.rs
cfallin: Ah, cool, the cfg-inside-the-function is at least a little bit cleaner -- switched to that. Thanks!

alexcrichton:
Could you address CI failures in separate commits instead of squashing all back into one? That makes it a bit easier to review after this has merged (and it all gets squashed in the merge queue anyway)

cfallin:
Ah, sorry about that! Bad habit trying to keep a clean commit history. Last diff was

```
diff --git a/crates/wasmtime/src/runtime/vm/libcalls.rs b/crates/wasmtime/src/runtime/vm/libcalls.rs
index de80727fa9..19a18af8ea 100644
--- a/crates/wasmtime/src/runtime/vm/libcalls.rs
+++ b/crates/wasmtime/src/runtime/vm/libcalls.rs
@@ -1612,7 +1612,10 @@ fn raise(store: &mut dyn VMStore, _instance: Pin<&mut Instance>) {
     // When Cranelift isn't in use then this is an unused libcall for Pulley, so
     // just insert a stub to catch bugs if it's accidentally called.
     #[cfg(not(has_host_compiler_backend))]
-    unreachable!()
+    {
+        let _ = store;
+        unreachable!()
+    }
 }

 // Builtins for continuations. These are thin wrappers around the
diff --git a/crates/wasmtime/src/runtime/vm/throw.rs b/crates/wasmtime/src/runtime/vm/throw.rs
index 2cf28c5caa..a6cf333a94 100644
--- a/crates/wasmtime/src/runtime/vm/throw.rs
+++ b/crates/wasmtime/src/runtime/vm/throw.rs
@@ -115,9 +115,10 @@ pub unsafe fn compute_throw_action(store: &mut dyn VMStore) -> ThrowAction {
         }
         None
     };
+    let unwinder = nogc.unwinder();
     let action = unsafe {
         wasmtime_unwinder::compute_throw_action(
-            &wasmtime_unwinder::UnwindHost,
+            unwinder,
             handler_lookup,
             exit_pc,
             exit_trampoline_fp,
```

alexcrichton:
While there may be more than one segfaulting test I can reproduce `exceptions::craneliftnative_basic_throw` failing with this stack trace:

```
#0  0x000002aa022684a8 in wasmtime_environ::module::Module::defined_tag_index () at crates/environ/src/module.rs:528
#1  0x000002aa022768a0 in wasmtime::runtime::vm::instance::Instance::get_exported_tag () at crates/wasmtime/src/runtime/vm/instance.rs:676
#2  0x000002aa020592fe in wasmtime::runtime::vm::throw::compute_throw_action::{closure#0} () at crates/wasmtime/src/runtime/vm/throw.rs:99
#3  0x000002aa02058a6e in wasmtime_internal_unwinder::throw::compute_throw_action::{closure#0}<wasmtime::runtime::vm::throw::compute_throw_action::{closure_env#0}> () at crates/unwinder/src/throw.rs:76
#4  0x000002aa0236f44c in wasmtime_internal_unwinder::stackwalk::visit_frames<wasmtime_internal_unwinder::throw::ThrowAction, wasmtime_internal_unwinder::throw::compute_throw_action::{closure_env#0}<wasmtime::runtime::vm::throw::compute_throw_action::{closure_env#0}>> () at crates/unwinder/src/stackwalk.rs:203
#5  0x000002aa020588b6 in wasmtime_internal_unwinder::throw::compute_throw_action<wasmtime::runtime::vm::throw::compute_throw_action::{closure_env#0}> () at crates/unwinder/src/throw.rs:60
#6  0x000002aa020613f8 in wasmtime::runtime::vm::throw::compute_throw_action () at crates/wasmtime/src/runtime/vm/throw.rs:120
#7  0x000002aa02387efc in wasmtime::runtime::vm::traphandlers::call_thread_state::CallThreadState::record_unwind () at crates/wasmtime/src/runtime/vm/traphandlers.rs:819
#8  0x000002aa0241cca2 in wasmtime::runtime::vm::traphandlers::catch_unwind_and_record_trap::{closure#1}<core::result::Result<(), wasmtime::runtime::vm::traphandlers::TrapReason>, wasmtime::runtime::vm::instance::{impl#0}::enter_host_from_wasm::{closure_env#0}<core::result::Result<(), wasmtime::runtime::vm::traphandlers::TrapReason>, wasmtime::runtime::vm::libcalls::raw::throw_ref::{closure_env#0}>> ()
    at crates/wasmtime/src/runtime/vm/traphandlers.rs:136
#9  0x000002aa0245e3b8 in wasmtime::runtime::vm::traphandlers::tls::with<(), wasmtime::runtime::vm::traphandlers::catch_unwind_and_record_trap::{closure_env#1}<core::result::Result<(), wasmtime::runtime::vm::traphandlers::TrapReason>, wasmtime::runtime::vm::instance::{impl#0}::enter_host_from_wasm::{closure_env#0}<core::result::Result<(), wasmtime::runtime::vm::traphandlers::TrapReason>, wasmtime::runtime::vm::libcalls::raw::throw_ref::{closure_env#0}>>> () at crates/wasmtime/src/runtime/vm/traphandlers.rs:1394
#10 0x000002aa02417ecc in wasmtime::runtime::vm::traphandlers::catch_unwind_and_record_trap<core::result::Result<(), wasmtime::runtime::vm::traphandlers::TrapReason>, wasmtime::runtime::vm::instance::{impl#0}::enter_host_from_wasm::{closure_env#0}<core::result::Result<(), wasmtime::runtime::vm::traphandlers::TrapReason>, wasmtime::runtime::vm::libcalls::raw::throw_ref::{closure_env#0}>> () at crates/wasmtime/src/runtime/vm/traphandlers.rs:136
#11 0x000002aa021466e0 in wasmtime::runtime::vm::instance::Instance::enter_host_from_wasm<core::result::Result<(), wasmtime::runtime::vm::traphandlers::TrapReason>, wasmtime::runtime::vm::libcalls::raw::throw_ref::{closure_env#0}> () at crates/wasmtime/src/runtime/vm/instance.rs:265
#12 0x000002aa02132a9a in wasmtime::runtime::vm::libcalls::raw::throw_ref () at crates/wasmtime/src/runtime/vm/libcalls.rs:125
#13 0x000075fbea7486cc in ?? ()
#14 0x000075fbea7480da in ?? ()
#15 0x000075fbea74813c in ?? ()
#16 0x000075fbea74821c in ?? ()
#17 0x000002aa016c9026 in wasmtime::runtime::vm::vmcontext::VMFuncRef::array_call_native () at crates/wasmtime/src/runtime/vm/vmcontext.rs:961
#18 0x000002aa016c8f66 in wasmtime::runtime::vm::vmcontext::VMFuncRef::array_call () at crates/wasmtime/src/runtime/vm/vmcontext.rs:918
#19 0x000002aa018f2082 in wasmtime::runtime::func::{impl#1}::call_unchecked_raw::{closure#0}<()> () at crates/wasmtime/src/runtime/func.rs:1025
#20 0x000002aa01102302 in wasmtime::runtime::vm::traphandlers::catch_traps::{closure#0}::call_closure<wasmtime::runtime::func::{impl#1}::call_unchecked_raw::{closure_env#0}<()>> ()
    at crates/wasmtime/src/runtime/vm/traphandlers.rs:463
```

Given that it's s390x-only though it's probably an endianness issue, perhaps something with a little-endian load needs to be native-endian? Or vice versa? Or maybe a store on the host needs to be little instead of native endian?

cfallin:
Yep, it's almost certainly endianness -- taking a look!

cfallin:
Actually, it's an issue with dynamic context reads during the stack-walk, it seems -- s390x ABI is a bit different (stackchains rather than FP-chains) so I suspect this is the issue.

cfallin:
s390x turned out to expose a mismatch in the definition of "spillslot offset" for the dynamic context -- the accessor I had exposed returns offset from the fixed storage area, which is ordinarily at `SP+0` *unless* there is an outgoing args area. s390x always has an outgoing args area (per ABI). A win for ISA diversity wrt testing!

PR #11327, opened 2025-07-26
Fix declaration of c-api function wasmtime_component_linker_instance_add_func

bartbes:
I was trying to use the C api and could not figure out how to implement a finalizer that takes no arguments. Turns out, the c declaration simply did not match the rust implementation.

See the rust specification of `finalizer` here:
https://github.com/bytecodealliance/wasmtime/blob/d2f51186ffb9fd2110d8658f215922b8398dd491/crates/c-api/src/component/linker.rs#L105-L112

`extern "C" fn(*mut c_void)>` of course matches `void(*)(void*)`, not `void(*)()`.


alexcrichton (approved):
Thanks!

Issue #11247, opened 2025-07-15
`wit-bindgen` async directives are not composable


rvolosatovs:
#### Feature

For the most part, whether or not a function should be generated as `async` should be dictated by WIT (refs https://github.com/bytecodealliance/wasmtime/issues/11246), however `wasmtime-wasi` might require generating some bindings as `async` even if they are not marked as such in WIT. See https://github.com/bytecodealliance/wasip3-prototyping/blob/3ce9ccb3efd3d15dc3ff55e31d0b85869847cf3f/crates/wasi/src/p3/bindings.rs#L21-L62 for a full list of bindings generated as async in wasip3-prototyping. An example of a use case is e.g. `wasi:sockets` implementation, which relies on a (Rust) `async` socket address check hook https://github.com/bytecodealliance/wasip3-prototyping/blob/3ce9ccb3efd3d15dc3ff55e31d0b85869847cf3f/crates/wasi/src/p3/sockets/mod.rs#L52-L58 https://github.com/bytecodealliance/wasip3-prototyping/blob/3ce9ccb3efd3d15dc3ff55e31d0b85869847cf3f/crates/wasi/src/p3/sockets/host/types/tcp.rs#L220

We need a way to "inherit" `async` directives from the WIT, while being able to "mark" additional ones as `async`

#### Benefit

This will drastically cut down the amount of boilerplate required to generate bindings, since currently passing `only_imports` to `async` `bindgen` argument is the only way to choose imports to be `async`. Note, that at least with current implementation, users of `bindgen` relying on generated bindings using `with`, e.g. embedders extending `wasmtime_wasi` with additional bindings MUST inherit the complete `only_imports` list from bindings depended upon (e.g. https://github.com/rvolosatovs/wasmtime/blob/c822293b26a1f2720d5a4cd035a59a66d4512e90/crates/wasi/src/p3/bindings.rs#L36-L83), otherwise compilation fails with:

```
error[E0277]: the trait bound `D: HostConcurrent` is not satisfied
   --> crates/wasi/src/p3/bindings.rs:20:1
    |
8   | / wasmtime::component::bindgen!({
9   | |     inline: "
10  | |         package example:wasi;
...   |
29  | |     concurrent_imports: true,
30  | | });
    | |__^ unsatisfied trait bound
    |
    = help: the trait `wasmtime_wasi::p3::bindings::sockets::types::HostConcurrent` is not implemented for `D`
[...]

Issue #11246, opened 2025-07-15
Bindgen does not support WIT `async` directives


rvolosatovs:
### Steps to Reproduce

apply:
```diff
diff --git a/crates/wasi/src/p3/bindings.rs b/crates/wasi/src/p3/bindings.rs
index 864d1eb8a4..02651057b8 100644
--- a/crates/wasi/src/p3/bindings.rs
+++ b/crates/wasi/src/p3/bindings.rs
@@ -38,50 +38,6 @@
 //!     },
 //!     concurrent_exports: true,
 //!     concurrent_imports: true,
-//!     async: {
-//!         only_imports: [
-//!             "wasi:cli/stdin@0.3.0#get-stdin",
-//!             "wasi:cli/stdout@0.3.0#set-stdout",
-//!             "wasi:cli/stderr@0.3.0#set-stderr",
-//!             "wasi:clocks/monotonic-clock@0.3.0#[async]wait-for",
-//!             "wasi:clocks/monotonic-clock@0.3.0#[async]wait-until",
-//!             "wasi:filesystem/types@0.3.0#[method]descriptor.read-via-stream",
-//!             "wasi:filesystem/types@0.3.0#[async method]descriptor.write-via-stream",
-//!             "wasi:filesystem/types@0.3.0#[async method]descriptor.append-via-stream",
-//!             "wasi:filesystem/types@0.3.0#[async method]descriptor.advise",
-//!             "wasi:filesystem/types@0.3.0#[async method]descriptor.sync-data",
-//!             "wasi:filesystem/types@0.3.0#[async method]descriptor.get-flags",
-//!             "wasi:filesystem/types@0.3.0#[async method]descriptor.get-type",
-//!             "wasi:filesystem/types@0.3.0#[async method]descriptor.set-size",
-//!             "wasi:filesystem/types@0.3.0#[async method]descriptor.set-times",
-//!             "wasi:filesystem/types@0.3.0#[async method]descriptor.read-directory",
-//!             "wasi:filesystem/types@0.3.0#[async method]descriptor.sync",
-//!             "wasi:filesystem/types@0.3.0#[async method]descriptor.create-directory-at",
-//!             "wasi:filesystem/types@0.3.0#[async method]descriptor.stat",
-//!             "wasi:filesystem/types@0.3.0#[async method]descriptor.stat-at",
-//!             "wasi:filesystem/types@0.3.0#[async method]descriptor.set-times-at",
[...]

PR #11328, opened 2025-07-26
Redesign function configuration in `bindgen!`

alexcrichton:
This commit is a redesign of how function-level configuration works in
Wasmtime's `bindgen!` macro. The main goal of this redesign is to
better support WASIp3 and component model async functions. Prior to this
redesign there was a mish mash of mechanisms to configure behavior of
imports/exports:

* The `async` configuration could turn everything async, nothing async,
  only some imports async, or everything except some imports async.

* The `concurrent_{imports,exports}` keys were required to explicitly
  opt-in to component model async signatures and applied to all
  imports/exports.

* The `trappable_imports` configuration would indicate a list of imports
  allowed to trap and it had special configuration for everything,
  nothing, and only a certain list.

* The `tracing` and `verbose_tracing` keys could be applied to either
  nothing or all functions.

Overall the previous state of configuration in `bindgen!` was clearly a
hodgepodge of systems that organically grew over time. In my personal
opinion it was in dire need of a refresh to take into account how
component-model-async ended up being implemented as well as
consolidating the one-off systems amongst all of these configuration
keys. A major motivation of this redesign, for example, was to inherit
behavior from WIT files by default. An `async` function in WIT should
not require `concurrent_*` keys to be configured, but rather it should
generate correct bindings by default.

In this commit, all of the above keys were removed. All keys have been
replaced with `imports` and `exports` configuration keys. Each behaves
the same way and looks like so:

    bindgen!({
        // ...
        imports: {
            // enable tracing for just this function
            "my:local/interface/func": tracing,

            // enable verbose tracing for just this function
            "my:local/interface/other-func": tracing | verbose_tracing,

            // this is blocking in WIT, but generate async bindings for
            // it
            "my:local/interface/[method]io.block": async,

            // like above, but use "concurrent" bindings which have
            // access to the store.
            "my:local/interface/[method]io.block-again": async | store,

            // everything else is, by default, trappable
            default: trappable,
        },
    });

Effectively all the function-level configuration items are now bitflags.
These bitflags are by default inherited from the WIT files itself (e.g.
`async` functions are `async | store` by default). Further configuration
is then layered on top at the desires of the embedder. Supported keys are:

* `async` - this means that a Rust-level `async` function should be
  generated. This is either `CallStyle::Async` or
  `CallStyle::Concurrent` as it was prior, depending on ...

* `store` - this means that the generated function will have access to
  the store on the host. This is only implemented right now for `async |
  store` functions which map to `CallStyle::Concurrent`. In the future
  I'd like to support just-`store` functions which means that you could
  define a synchronous function with access to the store in addition to
  an asynchronous function.

* `trappable` - this means that the function returns a
  `wasmtime::Result<TheWitBindingType>`. If `trappable_errors` is
  applicable then it means just a `Result<TheWitOkType,
  TrappableErrorType>` is returned (like before)

* `tracing` - this enables `tracing!` integration for this function.

* `verbose_tracing` - this logs all argument values for this function
  (including lists).

* `ignore_wit` - this ignores the WIT-level defaults of the function
  (e.g. ignoring WIT `async`).

The way this then works is all modeled is that for any WIT function
being generated there are a set of flags associated with that function.
To calculate the flags the algorithm looks like:

1. Find the first matching rule in the `imports` or `exports` map
   depending on if the function is imported or exported. If there is no
   matching rule then use the `default` rule if present. This is the
   initial set of flags for the function (or empty if nothing was
   found).

2. If `ignore_wit` is present, return the flags from step 1. Otherwise
   add in `async | store` if the function is `async` in WIT.

The resulting set of flags are then used to control how everything is
generated. For example the same split traits of today are still
generated and it's controlled based on the flags. Note though that the
previous `HostConcurrent` trait was renamed to `HostWithStore` to make
space for synchronous functions in this trait in the future too.

The end result of all these changes is that configuring imports/exports
now uses the exact same selection system as the `with` replacement map,
meaning there's only one system of selecting functions instead of 3.
WIT-level `async` is now respected by default meaning that bindings work
by default without further need to configure anything (unless more
functionality is desired).

One final minor change made here as well is that auto-generated
`instantiate` methods are now always synchronous and an
`instantiate_async` method is unconditionally generated for async mode.
This means that bindings always generate both functions and it's up to
the embedder to choose the appropriate one.

Closes https://github.com/bytecodealliance/wasmtime/issues/11246
Closes https://github.com/bytecodealliance/wasmtime/issues/11247


alexcrichton:
One other large-ish change now here, `*WithStore` is always generated no matter what. That'll make the generated docs slightly more confusing but I think it's worth the tradeoff of enabling bindings to always assume the trait is there.

crates/wasi-io/src/bindings.rs
```diff
@@ -1,27 +1,25 @@
...
+    imports: {
+        "wasi:io/poll/poll": async | trappable,
+        "wasi:io/poll/[method]pollable.block": async | trappable,
+        "wasi:io/poll/[method]pollable.ready": async | trappable,
+        "wasi:io/streams/[method]input-stream.blocking-read": async | trappable,
+        "wasi:io/streams/[method]input-stream.blocking-skip": async | trappable,
+        "wasi:io/streams/[drop]input-stream": async | trappable,
+        "wasi:io/streams/[method]output-stream.blocking-splice": async | trappable,
+        "wasi:io/streams/[method]output-stream.blocking-flush": async | trappable,
+        // "wasi:io/streams/[method]output-stream.blocking-write": async | trappable,
```
dicej: Why is this one commented out?

crates/wit-bindgen/src/lib.rs
```diff
@@ -2936,79 +2845,82 @@ impl<'a> InterfaceGenerator<'a> {
...
+        ret.with_store_name = Some(format!("{trait_name}WithStore"));
+        for (func, flags) in partition.with_store.iter() {
+            self.generate_function_trait_sig(func, *flags);
+            self.push_str(";\n");
         }
+        uwriteln!(self.src, "}}");
 
-        if is_maybe_async {
+        // If `*WithStore` is empty, generate a blanket impl for the trait since
+        // it's otherwies not necessary to implement it manually.
```
dicej:
```suggestion
        // it's otherwise not necessary to implement it manually.
```

crates/wit-bindgen/src/lib.rs
```diff
@@ -2936,79 +2845,82 @@ impl<'a> InterfaceGenerator<'a> {
...
-                        .drop_call_style(self.qualifier().as_deref(), name)
-                    {
-                        uwrite!(self.src, "async ");
-                    }
+                    let flags = self.import_resource_drop_flags(name);
+                    ret.all_func_flags |= flags;
                     uwrite!(
                         self.src,
-                        "fn drop(&mut self, rep: {wt}::component::Resource<{camel}>) -> {wt}::Result<()>;"
+                        "fn drop(&mut self, rep: {wt}::component::Resource<{camel}>) -> " // {wt}::Result<()>;"
```
dicej: Did you intend to keep this comment-ed out code fragment?

crates/wasi-io/src/bindings.rs
alexcrichton: Oops, meant to come back and dig in, but it's just unused so needed to be deleted.

crates/wit-bindgen/src/lib.rs
alexcrichton: Oops nah just a forgotten deletion.

Issue #11348, opened 2025-07-30
`LowerContext` is not always used on a fiber in component-model-async


alexcrichton:
Lowering into WebAssembly may require calling the `realloc` component model abi option to allocate space for a returned list, for example. WebAssembly in async mode must always be invoked on a fiber to support suspension at any time, for example with epochs or fuel injecting yields. Currently the component-model-async implementation does not always create and use a `LowerContext` on a fiber which will result in panics that fiber-related contextual information is unavailable.

This was discovered in review of https://github.com/bytecodealliance/wasmtime/pull/11325 and while that has an attempted solution for futures and streams it was also discovered that the solution isn't complete. I'm opening issue to keep track of this and take notes over time as necessary. 

I shared with @dicej a small patch:

```diff
diff --git a/crates/wasmtime/src/runtime/component/func/options.rs b/crates/wasmtime/src/runtime/component/func/options.rs
index 620e5a71d5..a839772824 100644
--- a/crates/wasmtime/src/runtime/component/func/options.rs
+++ b/crates/wasmtime/src/runtime/component/func/options.rs
@@ -241,6 +241,11 @@ impl<'a, T: 'static> LowerContext<'a, T> {
         types: &'a ComponentTypes,
         instance: Instance,
     ) -> LowerContext<'a, T> {
+        if cfg!(debug_assertions) {
+            if store.engine().config().async_support {
+                store.0.with_blocking(|_, _| {});
+            }
+        }
         LowerContext {
             store,
             options,
```

which can be used to assert that `LowerContext` is created on a fiber. This currently panics many tests in the `component-async-tests` crate (and probably others). The current plan is to fix these panics then perhaps move this assertion to `realloc` instead to have a fast path where "flat" types can skip the fiber if they don't need realloc.

dicej:
Fixed by #11353

PR #11353, opened 2025-07-30
ensure component value lowerings are run on a worker fiber

dicej:
This is necessary because lowering a component value may require calling realloc, which may involve epoch interruption, which may require suspending the current fiber.  Which obviously doesn't work if we're not running in a fiber. Also, we need to make sure there are no host frames on the stack.

Note the use of `Mutex` for `WorkItem::WorkerFunction` and `WorkerItem::Function`.  We never lock the mutex -- it's only used to plumb through the inner, non-`Sync` value while satisfying the compiler that `Store` is `Sync`.

### DIFF
--- a/crates/wasmtime/src/runtime/component/concurrent.rs
+++ b/crates/wasmtime/src/runtime/component/concurrent.rs
@@ -674,6 +674,12 @@ impl GuestCall {
     }
 }
 
+/// Job to be run on a worker fiber.
+enum WorkerItem {
+    GuestCall(GuestCall),
+    Function(Mutex<Box<dyn FnOnce(&mut dyn VMStore, Instance) -> Result<()> + Send>>),
+}
+
 /// Represents state related to an in-progress poll operation (e.g. `task.poll`
 /// or `CallbackCode.POLL`).
 #[derive(Debug)]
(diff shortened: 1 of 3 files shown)


dicej:
This should fix https://github.com/bytecodealliance/wasmtime/issues/11348 (and I've verified tests pass with Alex's assertion).

alexcrichton (approved):
Nice! That was easier to fix than expected. Happy to see this merged whenever, but a few thoughts too:

* For `Mutex` I've been wondering if we should have a wrapper of some kind that sort of looks like a `Mutex` but is just a newtype wrapper. Basically it would disallow access to the underlying data with `&self` and would exclusively provide access via `&mut self`. That would support `unsafe impl<T: Send> Sync for TheWrapper<T>`. There's another example of this in the codebase [here](https://github.com/bytecodealliance/wasmtime/blob/69fecf4e7d914156ff2a2bf3e105fed5640c2a82/crates/wasmtime/src/runtime/store/func_refs.rs#L91-L94) too. The only benefit of this would be avoiding the cost of creating the mutex, which I don't think is too costly nowadays, so no need to get to this any time soon (but would be good to keep in our back pocket I think).
* Could the assertion be added in this PR too? I'm imagining something like where do leave it in `LowerContext::new` and then for the "fast path" case we provide an alternate constructor which is something like `LowerContext::new_assert_no_realloc` which would panic if `realloc` is called (and avoid the debug assertion). Or something like that, basically still have the debug assertion as much as we can but provide a hatch for the fast-path we're thinking of adding.

dicej:
I just added the assertion.  I had to gate it on the `component-model-async` feature because `StoreOpaque::with_blocking` is gated on that.  Alternatively, we could ungate them both.

dicej:
Agreed about using something simpler than a `Mutex` to make stuff like this `Sync` -- performance aside, it's confusing to readers to have `Mutex`s all over the place that are never locked.

alexcrichton:
It's ok to leave it gated for now, eventually I think we're going to want to lift more out of the feature gate but it's ok to defer that to later.

PR #11354, opened 2025-07-30
 Refine `Guard*` APIs and implementation 

alexcrichton:
This is a follow-up to https://github.com/bytecodealliance/wasmtime/pull/11325 with a number of cosmetic changes about
the shape of the API and structure of the internals:

* `{Stream,Future}{Reader,Writer}::guard` is now an alternative
  constructor to `Guard*::new` (import fewer types).
* Internally `WithAccessor` and `DropWithStore` are removed in favor of
  direct `Drop for Guard*` impls.
* An `Option` is used to replace `ManuallyDrop` and `unsafe` code.
* `{Stream,Future}{Reader,Writer}::close{,_with}` now take `&mut self`
  instead of `self` to be more composable with `&mut self` arguments
  during `Drop` for other structures (e.g. build-your-own
  drop-with-store).
* The type parameters on `Guard*` are simplified to just `T`, the future
  or stream payload, and `A: AsAccessor`. This helps cut down on the
  complexity of signatures.
* `Guard*` types now have `into_{stream,future}` as an alternative to
  `.into()` which doesn't require type annotations.


alexcrichton:
~~I'll note the first commit here is from https://github.com/bytecodealliance/wasmtime/pull/11351, so only the second commit is for this PR~~ EDIT: prereqs landed

dicej (approved):

PR #11356, opened 2025-07-30
Create `ir::Signature`s on a per-`ModuleInternedTypeIndex` basis

fitzgen:
Not on a per `TypeIndex` basis. The former is deduplicated, the latter is not. This way we don't end up creating duplicate `ir::Signature`s in the CLIF.

### DIFF
--- a/tests/disas/duplicate-function-types.wat
+++ b/tests/disas/duplicate-function-types.wat
@@ -1,0 +1,101 @@
+;;! target = "x86_64"
+
+(module
+  ;; These two types should be deduped to the same `ir::Signature` in the
+  ;; translated CLIF.
+  (type (func (result i32)))
+  (type (func (result i32)))
+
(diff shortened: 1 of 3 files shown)


tests/disas/duplicate-function-types.wat
```diff
@@ -0,0 +1,101 @@
+;;! target = "x86_64"
+
+(module
+  ;; These two types should be deduped to the same `ir::Signature` in the
+  ;; translated CLIF.
+  (type (func (result i32)))
+  (type (func (result i32)))
```
fitzgen, line 7: FWIW, I did check that without the rest of this commit, this test case produces duplicate `ir::Signature`s given this input.

fitzgen: That is, this test fails without the rest of the commit.

alexcrichton (approved):

PR #11361, opened 2025-07-31
fix sending default value when closing host-owned future writer

dicej:
My earlier commit didn't handle all the cases, which caused a regression for the wasi-http tests in the `wasip3-prototyping` repo.


alexcrichton:
Would it be possible to write a custom small test for this in this repo? (e.g. wast or test-programs)

dicej:
> Would it be possible to write a custom small test for this in this repo? (e.g. wast or test-programs)

Done.

Issue #11346, opened 2025-07-30
table_ops fuzzer: Panic exceeding table limits


alexcrichton:
An issue was opened at https://oss-fuzz.com/testcase-detail/6690130233720832 and the reproduction is [input.gz](https://github.com/user-attachments/files/21512410/input.gz). I'll note though that this is flagged as "fixed on main" right now but I'm not sure why. OSS-Fuzz claims this deterministically fails on https://github.com/bytecodealliance/wasmtime/commit/6047d27e1cbe3b8d168bd823da8b4e1dd1db25e0 but no longer fails on main, and that's probably also worth investigating.

cc @fitzgen @khagankhan 

```
thread '<unnamed>' panicked at /src/wasmtime/crates/fuzzing/src/oracles.rs:917:64:
	called `Result::unwrap()` on an `Err` value: table minimum size of 258809871 elements exceeds table limits
	note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
	AddressSanitizer:DEADLYSIGNAL
	=================================================================
	==411==ERROR: AddressSanitizer: ABRT on unknown address 0x05390000019b (pc 0x7f144979a00b bp 0x7ffe92c99540 sp 0x7ffe92c992f0 T0)
	SCARINESS: 10 (signal)
	    #0 0x7f144979a00b in raise /build/glibc-LcI20x/glibc-2.31/sysdeps/unix/sysv/linux/raise.c:51:1
	    #1 0x7f1449779858 in abort /build/glibc-LcI20x/glibc-2.31/stdlib/abort.c:79:7
	    #2 0x58cf1a0a4069 in std::sys::pal::unix::abort_internal::h96857ca33d9110f4 /rustc/3014e79f9c8d5510ea7b3a3b70d171d0948b1e96/library/std/src/sys/pal/unix/mod.rs:366:14
	    #3 0x58cf1a0a39a8 in std::process::abort::hbb0da5b195767e3b /rustc/3014e79f9c8d5510ea7b3a3b70d171d0948b1e96/library/std/src/process.rs:2499:5
	    #4 0x58cf1a09ee74 in libfuzzer_sys::initialize::_$u7b$$u7b$closure$u7d$$u7d$::he6c18c4a427ce4b8 /rust/registry/src/index.crates.io-1949cf8c6b5b557f/libfuzzer-sys-0.4.8/src/lib.rs:95:9
[...]

PR #11363, opened 2025-07-31
Silently skip when generation fails from EmptyChoose seed

khagankhan:
This resolves one of the mentioned bugs in #11346 by replacing an `expect`
call with a simple `return`, avoiding a panic when `EmptyChoose` is selected.


khagankhan:
@fitzgen should we return error or is this okay?

PR #11364, opened 2025-07-31
optimize host stream/future writes for flat payloads

dicej:
When we know statically that a payload will not require any guest realloc calls to lower, there's no need to defer the lowering to a fiber -- we can just do it immediately.

### DIFF
--- a/crates/wasmtime/src/runtime/component/func/typed.rs
+++ b/crates/wasmtime/src/runtime/component/func/typed.rs
@@ -687,6 +687,9 @@ pub unsafe trait ComponentType: Send + Sync {
     #[doc(hidden)]
     const IS_RUST_UNIT_TYPE: bool = false;
 
+    #[doc(hidden)]
+    const IS_FLAT_TYPE: bool = false;
+
     /// Returns the number of core wasm abi values will be used to represent
     /// this type in its lowered form.
     ///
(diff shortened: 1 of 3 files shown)


dicej:
One possible hole in this is that `LowerContext::options` is a public field, so someone could call `options.realloc` directly without going through `LowerContext::realloc` if they wanted (e.g. in a custom `Lower` implementation).  Do we need to address that?  If so, I could refactor some code so that we clone the options and null out the realloc function before passing it to `LowerContext`.

crates/wasmtime/src/runtime/component/func/typed.rs
```diff
@@ -687,6 +687,9 @@ pub unsafe trait ComponentType: Send + Sync {
     #[doc(hidden)]
     const IS_RUST_UNIT_TYPE: bool = false;
 
+    #[doc(hidden)]
+    const IS_FLAT_TYPE: bool = false;
```
alexcrichton: Could this be renamed to `MAY_REQUIRE_REALLOC`? This otherwise isn't a guarantee that can be relied on as it's only set for integers, but it's not set for aggregates of other flat types (e.g. option-of-enum, etc).

Also can you add documentation that getting this correct is required for not panicking? Basically how other parts of the codebase will, for example, skip fibers during lowering if "may require realloc" is `false`

PR #11365, opened 2025-07-31
Simplify WASI internal implementations

alexcrichton:
This commit migrates the WASIp2 implementation to be closer to the upcoming WASIp3 implementation in terms of how things are implemented internally. Previously the way things worked with WASIp2 is:

* Embedders call `add_to_linker` with `T: WasiView`
* Internally `add_to_linker` is called which creates `WasiImpl<&mut T>`
* All internal implementations were `impl<T> Host for WasiImpl<T> where T: WasiView`
* A forwarding impl of `impl<T: WasiView> WasiView for &mut T` was required

While this all worked it's a bit complicated for a few reasons:

1. Dealing with generically named structures like `WasiImpl` (or `IoImpl` or `WasiHttpImpl`) is a bit baroque and not always obvious as to what's going on.
2. The extra layer of generics in `impl<T> Host for WasiImpl<T>` adds a layer of conceptual indirection which is non-obvious.
3. Other WASI proposal implementations do not use this strategy and instead use "view" types or `impl Host for TheType` for example.
4. Internal incantations of `add_to_linker` had to deal with mixtures of `IoImpl` and `WasiImpl` and aligning everything just right.
5. An extra layer of generics on all impls meant that everything was generic meaning that `wasmtime-wasi`-the-crate didn't generate much code, causing longer codegen times for consumers.

The goal of this commit is to migrate towards the style of what WASIp3 is prototyping for how impls are modeled. This is done to increase the amount of code that can be shared between WASIp2 and WASIp3. This has a number of benefits such as being easier to understand and also being more modular where `wasi:clocks` implementations of traits don't require filesystem context to be present (as is the case today). This in theory helps a more mix-and-match paradigm of blending together various bits and pieces of `wasmtime-wasi` implementations.

Concretely the changes made here are:

* `WasiView` no longer inherits from `IoView`, they're unrelated traits now.
* `WasiView` now returns `WasiViewCtx<'a>` which has `ctx: &'a mut WasiCtx` and `table: &'a mut ResourceTable`. That means it basically does the same thing before but in a slightly different fashion.
* Implementations of `Host` traits are now directly for `WasiCtxView<'_>` and don't involve any generics at all. These are hopefully easier to understand and also better from a codegen/compile-time perspective.
* Embedders no longer need to implement `IoView` directly and instead fold that functionality into `WasiView`.
* `WasiHttpView` no longer inherits from `IoView` and instead has a direct `fn table` method. Additionally `WasiHttpImpl` no longer embeds `IoImpl` inside of it.
* Host traits for `wasi:io` are now implemented directly for `ResourceTable` instead of `IoImpl<T>`.

The immediate goal of this refactoring is to enable more sharing along the lines of #11362. This was not possible prior because WASIp3 requires a simultaneous borrow on the table/ctx while the trait hierarchy previously gave you one-or-the-other. With this new organization it will be possible to get both at the same time meaning more structure/contexts/etc can be shared between implementations.

prtest:full


alexcrichton:
I can sort of preemptively apologize to embedders about this. I feel like every few versions of Wasmtime we rejigger how WASI traits work and how the impls are set up. I feel like we continue to march towards something that fits for all use cases and feels right but this is inevitably an iterative progress where we can't leapfrog to the end.

If anyone runs across this and has difficulty updating I'm happy to help out in updates, feel free to ping me on Zulip or here or an issue.

pchickey (approved):

pchickey:
There are still some more places to thread through this change (CI found wasi-keyvalue) but the design change looks good. Glad this is possible now that we got rid of get_host for HasData.

PR #11369, opened 2025-08-01
Fix preview1 fd_seek when whence == Whence::End

catamorphism:
Previously, `fd_seek` was adding the offset to the file size when called with `Whence::End`, rather than subtracting it.


catamorphism:
Disregard, I misunderstood the man page for `fseek()`.

PR #11370, opened 2025-08-01
 Unify WASIp{2,3} context structures 

alexcrichton:
This removes `wasmtime_wasi::p{2,3}::{WasiCtx, WasiCtxBuilder,
WasiView}` in favor of only having `wasmtime_wasi::{WasiCtx,
WasiCtxBuilder, WasiView}` instead. Conceptually these revisions of WASI
all provide the same functionality just with a different veneer that the
component model offers, so having only one way to configure host-side
behavior will make it easier to both organize implementations internally
(e.g. more sharing of code) as well as for embedders to configure (only
one context to create/manage).


alexcrichton:
Note that this is currently based on https://github.com/bytecodealliance/wasmtime/pull/11368

PR #11371, opened 2025-08-01
Add `CompiledCodeStencil` to public export in machinst, for fixing crate `TargetIsa`

suprohub:
Crate `TargetIsa` have method `compile_function`, what return type is `CodegenResult<CompiledCodeStencil>`.
`CompiledCodeStencil` is not public exported, so i just cant use this crate for custom external isa..

### DIFF
--- a/cranelift/codegen/src/isa/mod.rs
+++ b/cranelift/codegen/src/isa/mod.rs
@@ -420,7 +420,7 @@ pub trait TargetIsa: fmt::Display + Send + Sync {
 
 /// A wrapper around the ISA-dependent flags types which only implements `Hash`.
 #[derive(Hash)]
-pub struct IsaFlagsHashKey<'a>(&'a [u8]);
+pub struct IsaFlagsHashKey<'a>(pub &'a [u8]);
 
 /// Function alignment specifications as required by an ISA, returned by
 /// [`TargetIsa::function_alignment`].
--- a/cranelift/codegen/src/lib.rs
+++ b/cranelift/codegen/src/lib.rs
@@ -68,7 +68,7 @@ pub use crate::machinst::buffer::{
     OpenPatchRegion, PatchRegion,
 };
 pub use crate::machinst::{
-    ABIMachineSpec, ArgPair, ArgsAccumulator, ArgsOrRets, CallInfo, CompiledCode,
+    ABIMachineSpec, ArgPair, ArgsAccumulator, ArgsOrRets, CallInfo, Callee, CompiledCode,
     CompiledCodeStencil, Final, FrameLayout, MachBuffer, MachBufferFinalized, MachInst,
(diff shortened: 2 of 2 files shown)


bjorn3:
Do you actually need to implement `TargetIsa`? Most of the backend framework is still private with this PR. So I think you did either want to fork Cranelift and use the backend framework, or you avoid `TargetIsa` entirely.

suprohub:
> Do you actually need to implement `TargetIsa`? Most of the backend framework is still private with this PR. So I think you did either want to fork Cranelift and use the backend framework, or you avoid `TargetIsa` entirely.

Yes, i actually need to implement `TargetIsa`, and i need implement this externally.
&dyn TargetIsa often used in many places. Also, `Module` needs TargetIsa too.

bjorn3:
You are slowly exposing basically the entire backend framework to end users. The backend framework is highly unstable and should definitively not be re-exported in the root of the cranelift-codegen crate.

bjorn3:
Does the `[patch.crates-io]` section in `Cargo.toml` for replacing cranelift-codegen with a version you patched yourself to include your extra backend not work? You are going to have to change your backend every Cranelift version anyway if you are depending on these unstable implementation details.

suprohub:
> You are slowly exposing basically the entire backend framework to end users. The backend framework is highly unstable and should definitively not be re-exported in the root of the cranelift-codegen crate.

Yeah, you are right. This is not cool

suprohub:
> Does the `[patch.crates-io]` section in `Cargo.toml` for replacing cranelift-codegen with a version you patched yourself to include your extra backend not work? You are going to have to change your backend every Cranelift version anyway if you are depending on these unstable implementation details.

But how i can use this unstable api? I very need custom implementation for `TargetIsa` and maybe `MachInst`.
I need use it as external crate, because i dont want all features from cranelift, only codegen api.

suprohub:
Ok, i make only one field public and export `CompiledCodeStencil` for implement `TargetIsa`. But i very need `MachBuffer` what requires `MachInst` implementation

cfallin:
@suprohub Thanks for the PR; unfortunately, as bjorn3 has said, these are all internal compiler implementation details and will not be exposed. Aside from the types you would need to export to make this work, there is the ISLE DSL, which doesn't have a cross-crate linking story right now; it uses libraries (the "prelude") via inclusion, so we would need to bundle the prelude source code in `cranelift-codegen` and export it as text, or factor it to a separate crate (which is much more awkward for work within `cranelift-codegen`), or something like that. But in general this is exposing a bunch of internals, and we don't make any public promises about the stability of those details across releases; nor do we make any promises about safety or correctness if they are used in the wrong way. Maintaining this freedom to refactor, this standard of correctness, and a clean coherent public interface, are all reasons why we have chosen not to expose these things.

If you want to maintain your own Cranelift backend, the canonical path is to fork; at some point if you would like to upstream a backend, if it is of general interest and the maintenance support is there, we could discuss it.

PR #11376, opened 2025-08-04
CacheConfigs are always enabled; clarify directory validation

jder:
This makes two small changes to the CacheConfig/Cache system:

First, the comments referred to enabled/disabled caches, which is no longer a thing after https://github.com/bytecodealliance/wasmtime/pull/10665, so I've removed this. There were also some docs about panics in these cases (which are also now impossible) which I removed.

Second, `CacheConfig::default()`/`CacheConfig::new()` produced configs which would panic if you asked for their `directory` without first calling `validate`, but `validate` is not public API. So instead we return an Option from `CacheConfig::directory()` and move the panic to `Cache::directory()` and `WorkerThread::directory()` which should be impossible to trigger (since `validate` is called by `Cache::new`). We could alternatively make a ValidatedCacheConfig or something which is returned from `validate` and store/pass that around, but it seemed a bit overkill to me.

cc @benbrandt who was working on this recently in case I'm missing something.

### DIFF
--- a/crates/cache/src/config.rs
+++ b/crates/cache/src/config.rs
@@ -305,14 +305,9 @@ generate_deserializer!(deserialize_percent(num: u8, unit: &str) -> u8 {
     }
 });
 
-static CACHE_IMPROPER_CONFIG_ERROR_MSG: &str =
-    "Cache system should be enabled and all settings must be validated or defaulted";
-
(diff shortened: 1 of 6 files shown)


alexcrichton (approved):
Thanks!

PR #11383, opened 2025-08-05
Build C API release artifacts with unwind tables

### LINKED ISSUE #11344: Why is my host app's crash stack trace overwritten/lost when Wasmtime is embedded?
# Description

I'm embedding wasmtime as a library in my host application to run WebAssembly plugins. However, when my host program encounters a fatal signal (e.g., SIGSEGV) and should generate a core dump, the signal is being intercepted by wasmtime's internal signal handling mechanism. This prevents the default core dump generation and makes it difficult to capture the actual crash stack trace for debugging.

How can I properly configure signal handling so that different signals (especially fatal ones like SIGSEGV, SIGABRT) are handled correctly—allowing the host application to generate core dumps and print stack information as expected, while still allowing wasmtime to handle WebAssembly-related traps?

# wasmtime version

```
wasmtime-v30.0.2-x86_64-linux-c-api.tar.xz
```

# linux system

```
# uname -a
Linux 384a4bbb-d3bb-4a96-ae4e-5529fd44a702 3.10.0-1062.9.1.el7.x86_64 #1 SMP Fri Dec 6 15:49:49 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

# cat /etc/redhat-release
CentOS Linux release 7.3.1611 (Core)
```

# Code initialization process

```c++
int main() {
....
// init wasmtime

// Capture exception signals and print stack information
#ifndef _WIN32
  ::google::InstallFailureSignalHandler();
  ::google::InstallFailureWriter([](const char* data, int size) {
    LOG(ERROR) << "Failure:" << std::string(data, size);
    google::FlushLogFiles(google::INFO);
    google::FlushLogFiles(google::WARNING);
    google::FlushLogFiles(google::ERROR);
    google::FlushLogFiles(google::FATAL);
  });
#endif
}
```

# The host app's coredump stack information is incorrect.

```sh
E0728 10:17:07.475342 108776 main.cpp:124] Failure:*** Aborted at 1753669027 (unix time) try "date -d @1753669027" if you are using GNU date ***
E0728 10:17:07.477213 108776 main.cpp:124] Failure:PC: @                0x0 (unknown)
E0728 10:17:07.477321 108776 main.cpp:124] Failure:*** SIGSEGV (@0xffffffffffffffef) received by PID 108417 (TID 0x7fa9931ff700) from PID 18446744073709551599; stack trace: ***
E0728 10:17:07.479700 108776 main.cpp:124] Failure:    @           0xf311f0 wasmtime::runtime::vm::sys::unix::signals::trap_handler::hfe532cb8288e83d6
```

alexcrichton:
In #11344 it was found that if Wasmtime had a frame on the stack then an application's previous unwinding was broken. This was due to the fact that Wasmtime's C API artifacts do not have unwind information built-in due to being build with `-Cpanic=abort`. This change updates to building the C API artifacts with `-Cforce-unwind-tables` even though Rust itself won't use them to assist with embedders that want to unwind. These should in theory be easily strippable if desired and additionally embedders always have the option to build their own version of the C API too.

Closes #11344

### DIFF
--- a/ci/build-release-artifacts.sh
+++ b/ci/build-release-artifacts.sh
@@ -71,6 +71,12 @@ fi
 
 cargo build --release $flags --target $target -p wasmtime-cli $bin_flags --features run
 
+# For the C API force unwind tables to be emitted to make the generated objects
+# more flexible. Embedders can always build without this but this enables
+# libunwind to produce better backtraces by default when Wasmtime is linked into
+# a different project that wants to unwind.
+export RUSTFLAGS="$RUSTFLAGS -C force-unwind-tables"
+
 mkdir -p target/c-api-build
 cd target/c-api-build
 cmake \


bjorn3:
By the way https://github.com/rust-lang/rust/pull/143613 will make `-Cforce-unwind-tables` the default on Linux. It was already force enabled on Windows. macOS would still need `-Cforce-unwind-tables` though.

alexcrichton:
Oh nice I wasn't aware of that!

PR #11384, opened 2025-08-05
Use the same `UdpSocket` in WASIp{2,3}

alexcrichton:
This commit refactors the implementation of `wasi:sockets` for WASIp2 and WASIp3 to use the same underlying host data structure for the `UdpSocket` resource in WIT. Previously each version of WASI had its own socket which resulted in duplicated code. There's some minor differences between WASIp2 and WASIp3 but it's easy enough to paper over with the same socket type. This is intended to help with the maintainability of this going forward to only have one type to operate on rather than two (which also ensures that bugfixes for one should affect the other).

One other change made in this commit is that sprinkled checks for whether or not UDP is allowed are all removed and canonicalized during UDP socket creation. This means that UDP socket creation is the only location that checks for whether UDP is allowed. Once a UDP socket is created it can be used freely regardless of whether the UDP setting is enabled or disabled. This is not intended to have a large practical effect but it does mean the behavior of hosts that deny UDP but manually give access to a UDP socket resource to a component may behave subtly differently.


crates/wasi/src/p2/host/udp.rs
```diff
@@ -28,51 +22,20 @@ impl udp::HostUdpSocket for WasiSocketsCtxView<'_> {
...
-
-            // Perform the OS bind call.
-            udp_bind(socket.udp_socket(), local_address)?;
-        }
+        check.check(local_address, SocketAddrUse::UdpBind).await?;
 
         let socket = self.table.get_mut(&this)?;
-        socket.udp_state = UdpState::BindStarted;
+        socket.bind(local_address)?;
+        socket.set_socket_addr_check(Some(check.clone()));
```
rvolosatovs:
```suggestion
        socket.set_socket_addr_check(Some(check));
```

crates/wasi/src/sockets/udp.rs
```diff
@@ -77,6 +83,7 @@ impl UdpSocket {
             socket: Arc::new(socket),
             udp_state: UdpState::Default,
             family: socket_address_family,
+            socket_addr_check: None,
```
rvolosatovs, line 89: Could we clone the "check" from context here and avoid having to `Arc::clone` on each p3 usage of it? (https://github.com/bytecodealliance/wasmtime/pull/11291#discussion_r2223901764)

rvolosatovs (approved):

crates/wasi/src/sockets/udp.rs
alexcrichton: Unfortunately no because the check will live in the store no matter what and we can't borrow the store across awaits, so it's got to get moved out.  This is mostly just here to handle how in WASIp2 the socket check is inherited from the `Network` type rather than unconditionally from `WasiCtx`. While esoteric it's possible for embedders to push their own `Network` without using `WasiCtx` which has a distinct check than the one in `WasiCtx` itself, although we might want to remove such a feature eventually (which would remove the need for this field here)

PR #11386, opened 2025-08-06
Use the same `TcpSocket` in WASIp{2,3}

alexcrichton:
This is the same as #11384 except for `TcpSocket` to share the implementation of TCP across WASIp{2,3} and avoid duplicating code between the two.

### DIFF
--- a/crates/wasi/src/p3/sockets/host/types/tcp.rs
+++ b/crates/wasi/src/p3/sockets/host/types/tcp.rs
@@ -22,10 +22,6 @@ use wasmtime::component::{
     Resource, ResourceTable, StreamReader, StreamWriter,
 };
 
-fn is_tcp_allowed<T>(store: &Accessor<T, WasiSockets>) -> bool {
-    store.with(|mut view| view.get().ctx.allowed_network_uses.tcp)
-}
-
 fn get_socket<'a>(
     table: &'a ResourceTable,
     socket: &'a Resource<TcpSocket>,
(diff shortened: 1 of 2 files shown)


crates/wasi/src/p3/sockets/host/types/tcp.rs
```diff
@@ -341,15 +244,12 @@ impl HostTcpSocketWithStore for WasiSockets {
...
             let sock = get_socket(view.get().table, &socket)?;
-            if let TcpState::Connected(stream) | TcpState::Receiving(stream) = &sock.tcp_state {
-                Ok((Arc::clone(&stream), data))
-            } else {
-                Err(ErrorCode::InvalidState.into())
-            }
+            let stream = sock.tcp_stream_arc()?;
+            Ok(Arc::clone(stream))
         })?;
         let mut buf = Vec::with_capacity(8096);
```
rvolosatovs:
```suggestion
        let mut buf = Vec::with_capacity(DEFAULT_BUFFER_CAPACITY);
```

not introduced in this PR, but since we're changing all of these anyway, perhaps good to fix this one

rvolosatovs (approved):
It's a bit hard to keep track of all the changes, but I checked out the code locally to do a little overview and changes look good to me.

Is changing the wasip2 socket behavior a concern at all? (e.g. `TcpSocket::new` now returning "access denied")

alexcrichton:
It's an intentional change yeah as I think it makes more sense to gate creation of sockets rather than gate operations on the sockets since it consolidates the check into just one place instead of having it spread out. It's also non-standard behavior in the sense that it's not specified in WASI itself, so it should be ok to tweak it's behavior over time. While the error is shuffled around disabling TCP and/or UDP still has the same effect where you can't do anything TCP/UDP-related once the options are disabled.

PR #11388, opened 2025-08-06
Update nightly rust used in CI

alexcrichton:
In addition to keeping things up-to-date this enables testing the `tail_loop` interpreter mode of Pulley using the `become` keyword and native Rust syntax. This still isn't ready for prime-time so it's just as gated as before, but we can promote a `cargo check` to a `cargo test` in CI.


fitzgen (approved):
nice!

PR #11389, opened 2025-08-06
Cranelift: Translate `ir::UserExternalNameRef`s into callers when inlining

fitzgen:
This was an entity that we forgot to translate from the callee into the caller. Note that we do not use the `EntityMap` offset approach for these entities because `ir::Function` hash-conses them.


fitzgen:
Yes, I did not mean to imply this fixes all the fuzz bugs. Just one bug. More to come

PR #11392, opened 2025-08-07
Fix Out-of-memory in table-ops

khagankhan:
- Decoupled `TableOps` and `TableOpsLimits,` which are now passed explicitly to `TableOp::fixup`
- Updated the macro to align with the new struct separation
- `TableOps::fixup` now processes the entire sequence of operations instead of starting from specific index

I placed clamping logic at the beginning of the `to_wasm_binary` method since the OOM issues originate in t`o_wasm_binary` This relies on `TableOp::fixup` to ensure that values respect those clamped limits to avoid potential traps.

I previously added clamping at the start of `TableOps::fixup` but the OOM still occurred there.

I let it fuzz for a while, and it ran fine but I sense something might be missing.

### DIFF
--- a/crates/fuzzing/src/generators/table_ops.rs
+++ b/crates/fuzzing/src/generators/table_ops.rs
@@ -10,13 +10,19 @@ use wasm_encoder::{
     TypeSection, ValType,
 };
 
+/// Limits controlling the structure of a generated Wasm module.
+#[derive(Debug, Default, Serialize, Deserialize)]
+pub struct TableOpsLimits {
+    pub(crate) num_params: u32,
+    pub(crate) num_globals: u32,
+    pub(crate) table_size: i32,
+}
+
 /// A description of a Wasm module that makes a series of `externref` table
 /// operations.
 #[derive(Debug, Default, Serialize, Deserialize)]
 pub struct TableOps {
-    pub(crate) num_params: u32,
-    pub(crate) num_globals: u32,
-    pub(crate) table_size: i32,
+    pub(crate) limits: TableOpsLimits,
     pub(crate) ops: Vec<TableOp>,
 }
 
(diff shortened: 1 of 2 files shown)


crates/fuzzing/src/generators/table_ops.rs
```diff
@@ -503,6 +541,7 @@ mod tests {
             let wasm = res.to_wasm_binary();
             let mut validator = Validator::new();
             let wat = wasmprinter::print_bytes(&wasm).expect("[-] Failed .print_bytes(&wasm).");
+            println!("{wat}");
```
fitzgen: This should be `log::debug!` and not a `println!`.

khagankhan:
Thanks! Yes that makes sense. Initially, I did that. Calling fixup in to_wasm_binary(). It hit assertion failure at for limit > 0. I guess after addressing it we can do that. I will make another PR where clamping happen at the beginning of fixup and fixup is called in encoding.

I forgot to remove println! :/

khagankhan:
I was not merged. @fitzgen do you know why this may happen 🤔?

fitzgen:
It looks like CI failed, you can see this via the "view details" button next to the "[github-merge-queue](https://github.com/apps/github-merge-queue) bot removed this pull request from the [merge queue](https://github.com/bytecodealliance/wasmtime/queue/main) due to failed status checks" notification.

In particular, this job failed: https://github.com/bytecodealliance/wasmtime/actions/runs/16817124055/job/47636482289

It looks like it is old enough that the logs were deleted however, so I will try re-enqueing this PR and if it fails again, you can see the failure via the method described above.

fitzgen:
This failure looks like some spurious networking issue involving docker or something. Retrying once more.

fitzgen:
Looks like this time there was an internal assertion inside the macos linker?? Retrying once more...

khagankhan:
Thanks!🤞

PR #11398, opened 2025-08-07
Further expand differential fuzzing documentation

alexcrichton:
Split out the list of non-deterministic behaviors to its own separate section to both make it more easy to scan over and additionally have more fuzzing-related discussion such as generating programs that bake-in NaN canonicalization. I've also updated this to include WASIp1 APIs to discourage differential fuzzing of WASIp1 as the proposal is effectively not suitable for this. I've noted, though, that WASIp2-and-beyond should be suitable for differential fuzzing.

### DIFF
--- a/docs/contributing-fuzzing.md
+++ b/docs/contributing-fuzzing.md
@@ -54,11 +54,10 @@ The Wasmtime maintainers appreciate bug reports with the following:
   well as your rationale for *why* that behavior is expected. For example, just
   because another Wasm engine or an alternative Wasmtime execution strategy
   produces a different result from default Wasmtime, that is not necessarily a
-  bug. Wasm has a variety of sources of
-  [non-determinism](./examples-deterministic-wasm-execution.md) (e.g. `NaN` bit
-  patterns, maximum call stack depth, and whether growing memory succeeds) and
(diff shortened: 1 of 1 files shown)


docs/contributing-fuzzing.md
```diff
@@ -72,6 +71,57 @@ Including the above information is extra important for bugs discovered
...
+modules but can quickly arise during fuzzing. Some example behaviors are:
+
+* NaN patterns - floating-point operations which produce NaN as a result are
+  allowed to produce any one of a set of patterns of NaN. This means that the
+  exact bit-representation of the result of a floating-point operation may
+  diverge across engines. When fuzzing you can update your source-generation
+  to automatically canonicalize NaN values after all floating point operations.
+  Wasmtime has built-in options to curb this [non-determinism] as well.
+
+* Relaxed SIMD - the `relaxed-simd` proposal to WebAssembly explicitly has
```
fitzgen: Care to bold these for even faster scanning? Something like

```markdown
* **NaN patterns:** ...
* **Relaxed SIMD:** ...
* ...
```

fitzgen (approved):

PR #11402, opened 2025-08-08
Improve garbage collection of DWARF for dead code

philipc:
Checking for the tombstone address during the reachability calculation allows more DWARF entries to be marked as unreachable, and there was already code to handle this for addresses left as 0.


fitzgen:
Thanks @philipc!

PR #11405, opened 2025-08-08
Use `naked_asm!`, delete `asm_func!`

alexcrichton:
This deletes our home-grown `asm_func!` macro in favor of using `#[unsafe(naked)]` functions within Wasmtime. This is needed for fiber-related bits right now where we need tight control over the exact assembly of some functions. This additionally migrates s390x fiber bits to Rust as inline assembly is now stable for s390x.

prtest:full


alexcrichton:
I'll note that this is a draft PR because naked functions are in Rust 1.88 which is not in our MSRV yet, but I wanted to see if things worked out ahead of time.

fitzgen:
> This additionally migrates s390x fiber bits to Rust as inline assembly is now stable for s390x.

Is this also part of 1.88? If not, then it seems worth it to split out into a PR that can land earlier.

alexcrichton:
It's possible yeah, but personally I'm content to wait for a release myself

alexcrichton:
Good to go now!

dicej (approved):

PR #11406, opened 2025-08-08
feat(wasip3): implement `wasi:filesystem`

rvolosatovs:
Refs https://github.com/bytecodealliance/wasip3-prototyping/issues/228

Looks like reads will benefit from quite a bit of refactoring, which I did not manage to get to this week yet.
Keeping this open as draft for now until that's done.


crates/wasi/src/p3/filesystem/host.rs
```diff
@@ -0,0 +1,421 @@
...
+    async fn read_via_stream<U>(
+        store: &Accessor<U, Self>,
+        fd: Resource<Descriptor>,
+        offset: Filesize,
+    ) -> wasmtime::Result<(StreamReader<u8>, FutureReader<Result<(), ErrorCode>>)> {
+        let file = store.get_file(&fd)?;
+        if !file.perms.contains(FilePerms::READ) {
+            return Err(types::ErrorCode::NotPermitted.into());
+        }
+        bail!("TODO: read_via_stream")
```
alexcrichton: mind filing an issue for this if the intention is to leave this before landing?

crates/wasi/src/p3/filesystem/host.rs
```diff
@@ -0,0 +1,421 @@
...
+    #[expect(unused)]
+    async fn read_directory<U>(
+        store: &Accessor<U, Self>,
+        fd: Resource<Descriptor>,
+    ) -> wasmtime::Result<(
+        StreamReader<DirectoryEntry>,
+        FutureReader<Result<(), ErrorCode>>,
+    )> {
+        let dir = store.get_dir(&fd)?;
+        bail!("TODO: read_directory")
```
alexcrichton: Similar to above, mind filing an issue for this? (unless the intention is to fill out in this PR)

crates/wasi/src/p2/host/filesystem.rs
```diff
@@ -658,17 +343,14 @@ impl HostDescriptor for WasiFilesystemCtxView<'_> {
         new_path: String,
     ) -> FsResult<()> {
         let old_dir = self.table.get(&fd)?.dir()?;
+        // NOTE: `Dir::rename_at` will check permissions as well, but we're doing it here early to
+        // preserve legacy behavior
```
alexcrichton: Is this a difference from WASIp3 intentionally? Or just an artifact of a test doing two invalid things at once and which error is coming out is now different?

crates/wasi/src/filesystem.rs
```diff
@@ -54,3 +60,1122 @@ bitflags::bitflags! {
...
+                Ok(meta.file_type().into())
+            }
+            Self::Dir(_) => Ok(DescriptorType::Directory),
+        }
+    }
+
+    pub(crate) async fn set_times(
+        &self,
+        atim: NewTimestamp,
+        mtim: NewTimestamp,
```
alexcrichton: Where it makes sense for types like this I think it'd also be reasonable to use the cap-std types directly and avoid having a layer of abstraction which is wasmtime-wasi specific (e.g. directly convert bindgen-generated types to cap-std types).

I don't think it makes sense for all types, but for some like this it might be reasonable to skip the extra enum here.

alexcrichton (approved):

crates/wasi/src/p2/host/filesystem.rs
rvolosatovs: I've tried to preserve existing behavior as much as possible originally, but have since changed that and documented in https://github.com/bytecodealliance/wasmtime/pull/11406/commits/0418f5432d93be38dad8f8dd71926a911b3a8933

rvolosatovs:
Unfortunately, I'm running into a runtime panic in https://github.com/bytecodealliance/wasmtime/pull/11406/commits/54384721d13fcf5eecdc11a57eeb2c2a1c3e73c9:

```
thread 'p3::p3_filesystem_file_read_write' panicked at crates/wasmtime/src/runtime/component/concurrent/tls.rs:81:5:
attempted to recursively call `tls::get` when the pointer was not present or already taken by a previous call to `tls::get`
```

I've ignored the test for now https://github.com/bytecodealliance/wasmtime/pull/11406/commits/f8dc7e0c088178b3346bab261e2f73720944f29c and ~will file an issue in a bit~ filed and issue: https://github.com/bytecodealliance/wasmtime/issues/11412

crates/wasi/src/runtime.rs
```diff
@@ -187,3 +189,46 @@ where
...
+}
+
+#[cfg(feature = "p3")]
+impl Drop for TaskTable {
+    fn drop(&mut self) {
+        for task in self.tasks.values() {
+            task.abort()
+        }
+    }
+}
```
alexcrichton: Was this necessary to pass tests? Ideally I'd like to avoid abstractions like this and as-is I'm not sure that this is doing much because a strong reference to the `TaskTable` is held in all subtasks meaning that it'll never get dropped until the file is already dropped otherwise. For filesystems this is also problematic because aborting a task does not guarantee its destruction, you'd still have to await the handle in some form (we don't support this right now) due to the fact that the other task still needs to be dropped. For filesystem things that's also hard because it can often involve a `spawn_blocking` which itself contains a strong reference to the file in question and isn't cancellable once it's started.

More-or-less my hypothesis is that this isn't doing anything right now other than adding overhead for task-tracking because it's only dropped when a file is already being dropped. Otherwise I'm not sure we're prepared to give such a guarantee for filesystems that the host descriptor is closed once the WASI descriptor is closed.

alexcrichton (commented):
I'll take a look at this panic

crates/wasi/src/runtime.rs
rvolosatovs: The drop of a descriptor explicitly makes sure to "drain" the task table https://github.com/bytecodealliance/wasmtime/pull/11406/files#diff-0e2974d9a17b9ade0679e01bc627e119e1956a7bc903a9f10530f10b5865c2cfR621

I'm happy to drop https://github.com/bytecodealliance/wasmtime/pull/11406/commits/40415b52f79a33c376a2a875ffecb1d0a01c7903 though and perhaps talk through this later today?

rvolosatovs:
@alexcrichton this should be ready for a repeated review now

Issue #11262, opened 2025-07-16
Internally within Wasmtime synchronous functions are often a lie


alexcrichton:
In the spirit of making Wasmtime more sound internally over time this is a major thorn in our side right now. Many functions in Wasmtime are synchronous but they're actually asynchronous in nature due to the possibility that something could be context-switched away at any moment. These context switches are required to happen on a fiber and while I'm not aware of any issues internally within Wasmtime, this is, in my opinion, unnecessarily surprising. I would ideally like to move towards a world where Wasmtime actually uses `async` internally, not just for the `async` feature of wasm itself, and then in synchronous mode it asserts the future is always ready since there aren't any sources of blocking async.

Concretely sources of surprising async-ness are:

* Async resource limiters
* Async GC
* Async call hooks 
* ... maybe more? this isn't intended to be exhaustive

Concretely one location of unsoundness has to do with growth of tables and memories. For example memory growth requires a [split borrow](https://github.com/bytecodealliance/wasmtime/blob/c3ac18b3e4fc3dacd83bbf9e1b0fefb0837a4607/crates/wasmtime/src/runtime/memory.rs#L602-L615) both of and within the store. The same is [true for tables](https://github.com/bytecodealliance/wasmtime/blob/c3ac18b3e4fc3dacd83bbf9e1b0fefb0837a4607/crates/wasmtime/src/runtime/externals/table.rs#L275-L288). This is due to the fact that the methods of growth require a `&mut self` on the runtime data structure but also require, effectively, `&mut StoreOpaque` to suspend the world. Given that memories/tables live within `&mut StoreOpaque`, this is not a sound way to model this problem.

Another concrete location of unsoundness is that [growth of a shared memory](https://github.com/bytecodealliance/wasmtime/blob/c3ac18b3e4fc3dacd83bbf9e1b0fefb0837a4607/crates/wasmtime/src/runtime/vm/memory/shared_memory.rs#L77) suspends the stack with an rwlock write lock on the stack. This means that `unsafe impl Send for FiberFuture {}` is a lie because that's foundationally built on the premise that Wasmtime's own stack-local variables are all `Send`. This is effectively something that's impossible to guarantee and this hypothetically can result in locking an rwlock on one thread and unlocking it on another (technically UB, tested on glibc and other platforms to at least not behave badly for now). I was able to trivially detect this by migrating to `async fn`, however, as the compiler checks the send-ness of local variables.

[...]

PR #11411, opened 2025-08-09
Minimize lazy allocation of the GC store

alexcrichton:
This commit is an effort to minimize the number of entrypoints which might lazily allocate a GC store. The is currently done through `StoreOpaque::gc_store_mut` but this method is very commonly used meaning that there are many many places to audit for lazily allocating a GC store. The reason that this needs an audit is that lazy allocation is an async operation right now that must be on a fiber and is something I'm looking to fix as part of #11262.

This commit performs a few refactorings to achieve this:

* `gc_store_mut` is renamed to `ensure_gc_store`. This is intended to be an `async` function in the future and clearly demarcates where lazy allocation of a GC store is occurring.

* `require_gc_store{,_mut}` is now added which is a pure accessor of the GC store with no lazy allocation. Most locations previously using `gc_store_mut` are updated to use this instead.

Documentation is added to store methods to clearly indicate which ones are allocating and which ones should only be called in a context where allocation should already have happened.

### DIFF
--- a/crates/environ/src/compile/module_environ.rs
+++ b/crates/environ/src/compile/module_environ.rs
@@ -360,12 +360,7 @@ impl<'a, 'data> ModuleEnvironment<'a, 'data> {
                 for entry in tables {
                     let wasmparser::Table { ty, init } = entry?;
                     let table = self.convert_table_type(&ty)?;
-                    self.result.module.needs_gc_heap |= match table.ref_type.heap_type.top() {
-                        WasmHeapTopType::Extern | WasmHeapTopType::Exn | WasmHeapTopType::Any => {
-                            true
-                        }
-                        WasmHeapTopType::Func | WasmHeapTopType::Cont => false,
-                    };
+                    self.result.module.needs_gc_heap |= table.ref_type.is_vmgcref_type();
                     self.result.module.tables.push(table);
                     let init = match init {
                         wasmparser::TableInit::RefNull => TableInitialValue::Null {
--- a/crates/environ/src/compile/module_environ.rs
+++ b/crates/environ/src/compile/module_environ.rs
@@ -360,6 +360,12 @@ impl<'a, 'data> ModuleEnvironment<'a, 'data> {
                 for entry in tables {
                     let wasmparser::Table { ty, init } = entry?;
                     let table = self.convert_table_type(&ty)?;
+                    self.result.module.needs_gc_heap |= match table.ref_type.heap_type.top() {
+                        WasmHeapTopType::Extern | WasmHeapTopType::Exn | WasmHeapTopType::Any => {
+                            true
+                        }
+                        WasmHeapTopType::Func | WasmHeapTopType::Cont => false,
+                    };
                     self.result.module.tables.push(table);
                     let init = match init {
                         wasmparser::TableInit::RefNull => TableInitialValue::Null {
--- a/crates/wasmtime/src/runtime/gc/enabled/rooting.rs
+++ b/crates/wasmtime/src/runtime/gc/enabled/rooting.rs
@@ -175,7 +175,7 @@ mod sealed {
         /// objects that have been unrooted.
         fn try_clone_gc_ref(&self, store: &mut AutoAssertNoGc<'_>) -> Result<VMGcRef> {
             let gc_ref = self.try_gc_ref(store)?.unchecked_copy();
-            Ok(store.gc_store_mut()?.clone_gc_ref(&gc_ref))
+            Ok(store.require_gc_store_mut()?.clone_gc_ref(&gc_ref))
         }
     }
 }
@@ -263,7 +263,7 @@ impl GcRootIndex {
     /// particular `T: GcRef`.
     pub(crate) fn try_clone_gc_ref(&self, store: &mut AutoAssertNoGc<'_>) -> Result<VMGcRef> {
         let gc_ref = self.try_gc_ref(store)?.unchecked_copy();
-        Ok(store.gc_store_mut()?.clone_gc_ref(&gc_ref))
+        Ok(store.require_gc_store_mut()?.clone_gc_ref(&gc_ref))
     }
 }
 
@@ -1781,7 +1781,7 @@ where
         val_raw: impl Fn(u32) -> ValRaw,
     ) -> Result<()> {
         let gc_ref = self.try_clone_gc_ref(store)?;
-        let raw = store.gc_store_mut()?.expose_gc_ref_to_wasm(gc_ref);
+        let raw = store.require_gc_store_mut()?.expose_gc_ref_to_wasm(gc_ref);
         ptr.write(val_raw(raw.get()));
         Ok(())
     }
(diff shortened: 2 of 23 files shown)


alexcrichton:
@fitzgen ok so turns out there were more places than expected that really do lazily allocate a GC store, notably dealing with i31ref and null references. Best I can think of for handling this is to either:

1. Support allocating a GC store without a GC heap so things like `write_gc_ref` if the reference is i31ref or null. We'd then have `ensure_gc_store` (sync, just allocates some data structures) and `ensure_gc_heap` (async, allocates a heap) where both return `&mut GcStore`. Just `ensure_gc_store` wouldn't allocate a heap so we'd still skip it with i31ref and null references.
2. Change  callers of `write_gc_ref` to internally check for null references and i31ref an directly perform writes. This would basically require that all GC implementations don't need barriers for i31ref and null and would require more runtime checking in the host when performing those writes.

Do you  have other ideas and/or a preference about how might be best to resolve this?

fitzgen:
> Change callers of `write_gc_ref` to internally check for null references and i31ref an directly perform writes. This would basically require that all GC implementations don't need barriers for i31ref and null and would require more runtime checking in the host when performing those writes.

We already bake in the assumption that `i31ref`s don't need barriers in various places, because they aren't actually GC heap objects, so they aren't managed by the GC runtime. Same for null pointers. *Although* writing an `i31ref` or null reference into a field of a non-null object *does* need to call GC barriers.

Anyways, we have some of these checks in the `GcStore::*` barrier methods, and we could pull them out to being `StoreOpaque::*` barrier methods perhaps.

Also, if we have a non-`i31ref`, non-null reference, then we *must* have already allocated the GC store. And if we don't have one of those, then we don't need barriers, so we don't need to allocate a `GcStore` lazily to run barriers.

So I think that the `StoreOpaque::*` barrier methods shouldn't ever need to lazily allocate a `GcStore`. Basically something like

```rust
impl StoreOpaque {
    pub(crate) fn clone_gc_ref(&mut self, gc_ref: &GcRef) -> GcRef {
        if gc_ref.is_i31() {
            gc_ref.copy_i31()
        } else {
            self.gc_store
                .as_mut()
                .expect("non-null, non-i31 gc ref means we must have a gc store")
                .clone_gc_ref(gc_ref)
    }
}
```

crates/wasmtime/src/runtime/store.rs
```diff
@@ -1419,8 +1419,17 @@ impl StoreOpaque {
         &mut self.vm_store_context
     }
 
+    /// Allocates
```
fitzgen: Reminder to finish this doc comment

crates/wasmtime/src/runtime/store.rs
```diff
@@ -1493,24 +1501,36 @@ impl StoreOpaque {
         }
     }
 
+    /// Helper method to require that a `GcStore` was previously allocated for
+    /// this store, failing if it has not yet been allocated.
+    ///
+    /// Note that this should only be used in a context where allocation of a
+    /// `GcStore` is sure to have already happened prior, otherwise this may
+    /// return a confusing error to embedders which is a bug in Wasmtime.
```
fitzgen, line 1515: Perhaps provide some examples of when it should have already happened?

* We have a non-null, non-i31ref gc ref: in order to have allocated this reference's object, we must have allocated a gc heap and associated `GcStore`
* We are working with an instance whose associated `wasmtime_environ::Module::needs_gc_heap` flag is true: we always allocate the gc store during instantiation for these moduels.
* etc

crates/wasmtime/src/runtime/vm/instance/allocator.rs
```diff
@@ -612,7 +612,7 @@ fn initialize_tables(
                 let idx = module.table_index(table);
                 match module.tables[idx].ref_type.heap_type.top() {
                     WasmHeapTopType::Extern => {
-                        store.gc_store_mut()?;
+                        store.ensure_gc_store()?;
```
fitzgen: FWIW, if these lazy-allocation sites in `initialize_tables` are problematic, we could do `require_gc_store` here and require that callers check and initialize the gc heap if necessary. For instances, they should already be allocated by instantiation. For host tables, we would need to add a new check, but that should be straightforward enough to add.

crates/wasmtime/src/runtime/trampoline/global.rs
```diff
@@ -52,23 +52,23 @@ pub fn generate_global_export(
...
             }
             Val::ExnRef(e) => {
                 let new = match e {
                     None => None,
                     Some(e) => Some(e.try_gc_ref(&store)?.unchecked_copy()),
                 };
                 let new = new.as_ref();
-                global.write_gc_ref(store.gc_store_mut()?, new);
+                global.write_gc_ref(store.require_gc_store_mut()?, new);
             }
```
fitzgen: It isn't clear to me why these can `require` rather than `ensure`? Wouldn't we need to `ensure` in `Global::new` or somewhere in order for this `require` to be okay?

alexcrichton: These are the problematic writes yeah and a leading cause of failure in the tests below. There's no such `ensure` in `Global::new` right now which is why these are failing.

crates/wasmtime/src/runtime/vm/instance/allocator.rs
alexcrichton: Agreed yeah, I updated compilation to look at the top type of tables and flag `needs_gc_heap` from that as appropriate.

alexcrichton:
Ok I think everything should be handled now. Mind taking another look @fitzgen?

crates/environ/src/compile/module_environ.rs
```diff
@@ -360,6 +360,12 @@ impl<'a, 'data> ModuleEnvironment<'a, 'data> {
                 for entry in tables {
                     let wasmparser::Table { ty, init } = entry?;
                     let table = self.convert_table_type(&ty)?;
+                    self.result.module.needs_gc_heap |= match table.ref_type.heap_type.top() {
+                        WasmHeapTopType::Extern | WasmHeapTopType::Exn | WasmHeapTopType::Any => {
+                            true
+                        }
+                        WasmHeapTopType::Func | WasmHeapTopType::Cont => false,
+                    };
```
fitzgen:
```suggestion
                    self.result.module.needs_gc_heap |= table
                        .ref_type
                        .is_vmgcref_type_and_points_to_object();
```

crates/wasmtime/src/runtime/gc/enabled/rooting.rs
```diff
@@ -263,14 +263,14 @@ impl GcRootIndex {
...
         let gc_ref = self.try_gc_ref(store)?.unchecked_copy();
-        Ok(store.gc_store_mut()?.clone_gc_ref(&gc_ref))
+        Ok(store.clone_gc_ref(&gc_ref))
     }
 }
 
 /// This is a bit-packed version of
 ///
 /// ```ignore
-/// enema {
```
fitzgen, line 273: LOL how did this sneak by????

crates/environ/src/compile/module_environ.rs
alexcrichton: I ended up going with `is_vmgcref_type` since the method here is on `wasmtime_environ` types and not `wasmtime` types (therefore `is_vmgcref_type_and_points_to_object` not naively available). I tried using `is_vmgcref_type_not_i31` but due to the way table initialization works it requires the GC store to be [present right now](https://github.com/bytecodealliance/wasmtime/blob/8b68f6f193cd745b65cfea91d111582654a2701f/crates/wasmtime/src/runtime/vm/instance/allocator.rs#L630) so I switched it to `is_vmgcref_type`. Should be easy enough to fix in the future if we really want, but I figure it's not too important to be too optimal here and keeping the same code between the externref/exnref/anyref clauses is nice

PR #11414, opened 2025-08-11
Refactor `AbortHandle`, renamed to `JoinHandle`

alexcrichton:
This commit is inspired by recent discussions about providing more guarantees around dropping WASI resources. This commit renames `wasmtime::component::AbortHandle` to `wasmtime::component::JoinHandle` and additionally implements `Future for JoinHandle` to know when the associated task's future has been dropped. The renaming is intended to more closely align with `tokio::task::JoinHandle` where `tokio::task::AbortHandle` is notably different and doesn't have a `Future` implementation. The `Future` implementation helps embedders know exactly when a value has been dropped and synchronize on that.

### DIFF
--- a/crates/wasmtime/src/runtime/component/concurrent/abort.rs
+++ b/crates/wasmtime/src/runtime/component/concurrent/abort.rs
@@ -8,6 +8,10 @@ use std::task::{Context, Poll, Waker};
 /// This represents a handle to a running task which can be cancelled with
 /// [`JoinHandle::abort`]. The final result and drop of the task can be
 /// determined by `await`-ing this handle.
+///
+/// Note that dropping this handle does not affect the running task it's
+/// connected to. A manual invocation of [`JoinHandle::abort`] is required to
+/// affect the task.
 pub struct JoinHandle {
     state: Arc<Mutex<JoinState>>,
 }
--- a/crates/wasmtime/src/runtime/component/concurrent/abort.rs
+++ b/crates/wasmtime/src/runtime/component/concurrent/abort.rs
@@ -1,71 +1,278 @@
-use std::future;
-use std::pin::pin;
+use std::mem::{self, ManuallyDrop};
+use std::pin::Pin;
 use std::sync::{Arc, Mutex};
 use std::task::{Context, Poll, Waker};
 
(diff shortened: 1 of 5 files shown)


crates/wasmtime/src/runtime/component/concurrent/abort.rs
```diff
@@ -1,71 +1,278 @@
...
+impl JoinHandle {
     /// Abort the task.
     ///
     /// This flags the connected task should abort in the near future, but note
     /// that if this is called while the future is being polled then that call
     /// will still complete.
+    ///
+    /// Note that this `JoinHandle` is itself a `Future` and can be used to
+    /// await the result and destruction of the task that this is associated
+    /// with.
```
pchickey, line 52: It would be useful to note here the task can also be aborted by dropping the JoinHandle

crates/wasmtime/src/runtime/component/concurrent/abort.rs
```diff
@@ -1,71 +1,278 @@
...
 use std::task::{Context, Poll, Waker};
 
-/// Handle to a task which may be used to abort it.
+/// Handle to a task which may be used to join on the result of executing it.
 ///
 /// This represents a handle to a running task which can be cancelled with
-/// [`AbortHandle::abort`].
-pub struct AbortHandle {
-    state: Arc<Mutex<AbortState>>,
+/// [`JoinHandle::abort`]. The final result and drop of the task can be
```
pchickey, line 9: Useful to note here that dropping the JoinHandle will also abort.

pchickey (commented):
Abort on drop is an important difference over the tokio JoinHandle (and why we ended up with the mouthful https://docs.rs/wasmtime-wasi/latest/wasmtime_wasi/runtime/struct.AbortOnDropJoinHandle.html) so lets document it prominently

alexcrichton:
Oh the previous `AbortHandle`, nor the new `JoinHandle` name, have abort-on-drop behavior. We'll probably still want to end up with a similar structure for tokio tasks for async tasks, but I figured I'd leave that for the wasmtime-wasi crate.

dicej (approved):

PR #11415, opened 2025-08-11
Make Wasmtime's `FuncKey` one-to-one with Cranelift's `ir::UserExternalName`

fitzgen:
`FuncKey`, which used to be called `CompileKey`, is now one-to-one with `cranelift_codegen::ir::UserExternalName`, and is used for not just identifying compilation objects but also relocations and call-graph edges. This allows us to determine the `StaticModuleIndex` and `DefinedFuncIndex` pair for any `cranelift_codegen::ir::FuncRef`, regardless of inlining depth, which fixes some fuzz bugs on OSS-Fuzz.

This continues pushing on the idea that Wasmtime's compilation orchestration and linking should be relatively agnostic to the kinds of things it is actually compiling and linking, allowing us to tweak, add, and remove new kinds of `FuncKey`s more easily. Adding a new `FuncKey` should not require modifying relocation resolution, for example, just a little bit of code to run the associated compilation and optionally some code to extract metadata into our final artifacts for querying at runtime. Everything in between should Just Continue Working. We still aren't all the way there yet, but this does bring us a little bit closer.

Finally, in Cranelift's inlining pass, this adds a check that a block is inserted in the layout before attempting to remove it from the layout, which would otherwise cause panics. This was triggered by multi-level inlining and now-unreachable blocks in the inner callees.

I'll note that this does update basically all of the disas tests, or at least nearly all of them that make function calls. This is because the namespace/index numbering pair changed slightly to align with `FuncKey`, but that should pretty much be the only changes.

### DIFF
--- a/crates/environ/src/compile/key.rs
+++ b/crates/environ/src/compile/key.rs
@@ -1,0 +1,220 @@
+//! TODO FITZGEN
+
+#[cfg(feature = "component-model")]
+use crate::component;
+use crate::{
+    BuiltinFunctionIndex, DefinedFuncIndex, HostCall, ModuleInternedTypeIndex, StaticModuleIndex,
+};
+
+/// A sortable, comparable function key for compilation output, call graph
+/// edges, and relocations.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum FuncKey {
+    /// A Wasm-defined function.
+    DefinedWasmFunction(StaticModuleIndex, DefinedFuncIndex),
+
+    /// A trampoline from an array-caller to the given Wasm-callee.
+    ArrayToWasmTrampoline(StaticModuleIndex, DefinedFuncIndex),
+
+    /// A trampoline from a Wasm-caller to an array-callee of the given type.
+    WasmToArrayTrampoline(ModuleInternedTypeIndex),
+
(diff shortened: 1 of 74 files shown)


crates/environ/src/compile/key.rs
```diff
@@ -0,0 +1,220 @@
+//! TODO FITZGEN
```
fitzgen: Whoops, I need to fill out this comment

alexcrichton (approved):

tests/disas/call-indirect.wat
```diff
@@ -16,7 +16,7 @@
 ;;     gv5 = load.i64 notrap aligned gv3+56
 ;;     sig0 = (i64 vmctx, i64, i32) -> i32 tail
 ;;     sig1 = (i64 vmctx, i32, i64) -> i64 tail
-;;     fn0 = colocated u1:9 sig1
+;;     fn0 = colocated u1610612736:9 sig1
```
bjorn3, line 19: Maybe print the indexes as hex without `0x` prefix? `format!("u{namespace:x}:{index:x}")`

fitzgen: This would require also updating the CLIF parser, and I'd rather not get into that in this PR, as it is fairly unrelated.

PR #11416, opened 2025-08-12
Refactor internals of table initialization and management

alexcrichton:
This is a pair of commit inspired by semi-related work in Wasmtime to make progress on #11262. I got stuck in one particular area and as I stared more at the code I realized there were deeper issues that needed fixing before I start tackling #11262. Namely this PR fixes bugs such as:

* There were no write barriers on `table.init`, meaning that overwritten values were leaked.
* There were no write barriers on `table.set` via the embedder API, meaning overwritten values were leaked.
* The GC referenced passed to `table.grow` was not properly dropped via the embedder API, meaning that whatever reference that was passed in was leaked.
* Const-expression evaluation disallowed GC, but due to allocation of GC values it's possible to trigger a GC
* When const-expression evaluation triggers a GC it would be problematic during `table.init` because an instances tables were temporarily removed to help work within the borrow checker.

All of the above have now been addressed with a set of new tests that assert no leaks and which trigger GC-during-const-eval. More details can be found in the commit messages but my hope with this refactoring is to close out this class of issues or make it more obvious when they show up in the future. Some algorithms were pessimized here by having to repeatedly re-access things from the store, but it's expected that these are not performance hot-spots and feature-complete-ness and correctness is more paramount.

### DIFF
--- a/crates/wasmtime/src/runtime/vm/libcalls.rs
+++ b/crates/wasmtime/src/runtime/vm/libcalls.rs
@@ -890,10 +890,9 @@ unsafe fn array_new_elem(
             let mut const_evaluator = ConstExprEvaluator::default();
 
             vals.extend(xs.iter().map(|x| unsafe {
-                const_evaluator
+                *const_evaluator
                     .eval(&mut store, &mut const_context, x)
                     .expect("const expr should be valid")
-                    .clone()
             }));
         }
     }
@@ -979,10 +978,9 @@ unsafe fn array_init_elem(
                 .ok_or_else(|| Trap::TableOutOfBounds)?
                 .iter()
                 .map(|x| unsafe {
-                    const_evaluator
+                    *const_evaluator
                         .eval(&mut store, &mut const_context, x)
                         .expect("const expr should be valid")
-                        .clone()
                 })
                 .collect::<Vec<_>>()
         }


alexcrichton:
Looks like instantiation is indeed regressing a bit, so I'll dig in.

Screenshot 2025-08-12 at 11 21 37 AM

crates/wasmtime/src/runtime/externals/table.rs
```diff
@@ -270,12 +283,31 @@ impl Table {
...
+                TableElementType::Func => {
+                    let element = init.into_table_func(store, ty.element())?;
+                    (*table).grow_func(store, delta, element)?
+                }
+                TableElementType::GcRef => {
+                    // FIXME: `grow_gc_ref` shouldn't require the whole store
+                    // and should require `AutoAssertNoGc`. For now though we
+                    // know that table growth doesn't trigger GC so it should be
+                    // ok to create a copy of the GC reference even though it's
+                    // not tracked anywhere.
```
fitzgen, line 300: Why don't we make `grow_gc_ref` take an `AutoAssertNoGc`?

crates/wasmtime/src/runtime/externals/table.rs
```diff
@@ -222,13 +216,32 @@ impl Table {
...
-        let ty = self.ty(&store);
-        let val = val.into_table_element(store, ty.element())?;
-        let (table, _) = self.wasmtime_table(store, iter::empty());
-        table
-            .set(index, val)
-            .map_err(|()| anyhow!("table element index out of bounds"))
+        self.set_(store.as_context_mut().0, index, val)
+    }
+
+    pub(crate) fn set_(&self, store: &mut StoreOpaque, index: u64, val: Ref) -> Result<()> {
```
fitzgen, line 222: Here an elsewhere: we have already been using the convention `foo` for the generic version and `_foo` for the opaque version, not `foo_` for the opaque version. Should be consistent and have underscores at the front.

crates/wasmtime/src/runtime/vm/const_expr.rs
```diff
@@ -160,75 +143,68 @@ impl ConstExprEvaluator {
     /// stack.
     pub unsafe fn eval(
         &mut self,
-        store: &mut StoreOpaque,
+        // Ensure that we don't permanently root any GC references we allocate
+        // during const evaluation, keeping them alive for the duration of the
+        // store's lifetime.
+        //
+        // TODO: move this comment
```
fitzgen: 👀

crates/wasmtime/src/runtime/vm/table.rs
```diff
@@ -16,27 +16,6 @@ use wasmtime_environ::{
...
-
-    /// An uninitialized funcref value. This should never be exposed
-    /// beyond the `wasmtime` crate boundary; the upper-level code
-    /// (which has access to the info needed for lazy initialization)
-    /// will replace it when fetched.
-    UninitFunc,
-
-    /// A `contref`
-    ContRef(Option<VMContObj>),
-}
```
fitzgen, line 38: 🎉

crates/wasmtime/src/runtime/values.rs
```diff
@@ -112,6 +112,16 @@ impl Val {
         Val::AnyRef(None)
     }
 
+    pub(crate) const fn null_top(top: WasmHeapTopType) -> Val {
```
fitzgen, line 115: Aside: we should probably add a `wasmtime::HeapTopType` because the environ version is so useful all over the place and it is really nice to exhaustive match on top types. Then we could expose this helper to embedders too, which seems like it would be useful.

fitzgen (commented):
LGTM but withholding approval until we figure out the perf regression, since I expect that I will probably need to give this another once-over after resolving that

fitzgen:
> Looks like instantiation is indeed regressing a bit, so I'll dig in.

copying a message I sent Alex in private chat here for posterity:

> another idea: basically nothing uses non-constant const-expressions. we could make `ConstExpr` be something like
>
> ```rust
> enum ConstExpr {
>     Const(Val),
>     Expr(Vec<ConstOp>),
> }
> ```

crates/wasmtime/src/runtime/externals/table.rs
alexcrichton: Heh yeah I've done that historically as well (`_foo`) but Joel pointed out recently that if we stop using the function for whatever reason it won't get lints about dead code so I started switching to `foo_` instead. I'll keep it consistent for now internally in Wasmtime.

crates/wasmtime/src/runtime/externals/table.rs
alexcrichton: That can't work right now as growth requires the whole `dyn VMStore` to perform a fiber suspension but `AutoAssertNoGc` only requires the `StoreOpaque`. Soon though...

crates/wasmtime/src/runtime/externals/table.rs
fitzgen: Ah yeah good point we should fix everything at once in a follow up.

alexcrichton:
@fitzgen ok with the most recent commits I've clawed back all of the performance and then some, so should be good to go.

This makes me want to generate a Cranelift-compiled function for wasm instance initialization really.

PR #11418, opened 2025-08-12
Fix the QEMU command in the cross-compilation documentation.

yomaytk:
cargo test command of cross compilation target causes the following error in the many tests because the setting `runner = "qemu-aarch64 -L /usr/aarch64-linux-gnu -E LD_LIBRARY_PATH=/usr/aarch64-linux-gnu/lib -E WASMTIME_TEST_NO_HOG_MEMORY=1"` of .cargo/config.toml cannot pass environment variables to the qemu-user executed by binfmt.

```shell
# aarch64
aarch64-binfmt-P: Could not open '/lib/ld-linux-aarch64.so.1': No such file or directory
```

This PR fixes the error by specifying `env QEMU_LD_PREFIX=...`  to the .cargo/config.toml.

### DIFF
--- a/docs/contributing-cross-compiling.md
+++ b/docs/contributing-cross-compiling.md
@@ -62,15 +62,15 @@ if none already exists).
 ```toml
 [target.aarch64-unknown-linux-gnu]
 linker = "aarch64-linux-gnu-gcc"
-runner = "qemu-aarch64 -L /usr/aarch64-linux-gnu -E LD_LIBRARY_PATH=/usr/aarch64-linux-gnu/lib -E WASMTIME_TEST_NO_HOG_MEMORY=1"
+runner = "env QEMU_LD_PREFIX=/usr/aarch64-linux-gnu WASMTIME_TEST_NO_HOG_MEMORY=1 qemu-aarch64"
 
 [target.riscv64gc-unknown-linux-gnu]
 linker = "riscv64-linux-gnu-gcc"
-runner = "qemu-riscv64 -L /usr/riscv64-linux-gnu -E LD_LIBRARY_PATH=/usr/riscv64-linux-gnu/lib -E WASMTIME_TEST_NO_HOG_MEMORY=1"
+runner = "env QEMU_LD_PREFIX=/usr/riscv64-linux-gnu WASMTIME_TEST_NO_HOG_MEMORY=1 qemu-riscv64"
 
 [target.s390x-unknown-linux-gnu]
 linker = "s390x-linux-gnu-gcc"
-runner = "qemu-s390x -L /usr/s390x-linux-gnu -E LD_LIBRARY_PATH=/usr/s390x-linux-gnu/lib -E WASMTIME_TEST_NO_HOG_MEMORY=1"
+runner = "env QEMU_LD_PREFIX=/usr/s390x-linux-gnu WASMTIME_TEST_NO_HOG_MEMORY=1 qemu-s390x"
 ```
 
 ## Cross-Compile Tests and Run Them!


alexcrichton:
Could you detail a bit more about the differences? I'm mostly curious because I've been using the previously-documented form for as long as I can remember. Additionally sticking to just a `qemu-*` binary feels nice in the sense that it doesn't also require `env` necessarily.

cfallin:
It appears that `-L` and the `QEMU_LD_PREFIX` environment variable should be completely equivalent -- from the output of `qemu-aarch64 -h`, I see

```
Options and associated environment variables:

Argument             Env-variable      Description
[ ... ]
-L path              QEMU_LD_PREFIX    set the elf interpreter prefix to 'path'
```

I'm surprised that the existing command didn't work for you. Do you have the named file (from your error message) at `/usr/aarch64-linux-gnu/lib/ld-linux-aarch64.so.1`?

If so, could you show your complete Cargo config, the command you ran, and share info about your environment (Linux distro, Rust compiler version)?

yomaytk:
@alexcrichton @cfallin Thank you for the review!

According to the error message, the path to `ld-linux-aarch64.so.1` was missing `/usr/aarch64-linux-gnu`, so it seemed that binfmt was being launched without setting `/usr/aarch64-linux-gnu` as the sysroot. Therefore, I thought that during the execution of qemu-aarch64, the binfmt process launched as a other process could not receive the original `-L` option, leaving the sysroot empty, so I specified it via `env` command.

However, since the original configuration passes your tests, I’m not sure if my understanding is correct…

> Do you have the named file (from your error message) at /usr/aarch64-linux-gnu/lib/ld-linux-aarch64.so.1?

Yes, I do.
```shell
$ ls -l /usr/aarch64-linux-gnu/lib/ld-linux-aarch64.so.1
-rwxr-xr-x 1 root root 202784 Apr 15  2024 /usr/aarch64-linux-gnu/lib/ld-linux-aarch64.so.1
```

> If so, could you show your complete Cargo config, the command you ran, and share info about your environment (Linux distro, Rust compiler version)?

Environment is as follows:
```shell
$ uname -m; lsb_release -d
x86_64
No LSB modules are available.
Description:    Ubuntu 24.04.2 LTS
$ rustc --version
rustc 1.89.0 (29483883e 2025-08-04)
```

.cargo/config.toml is as follows (before the changes in this PR).
```shell
[target.aarch64-unknown-linux-gnu]
linker = "aarch64-linux-gnu-gcc"
runner = "qemu-aarch64 -L /usr/aarch64-linux-gnu -E LD_LIBRARY_PATH=/usr/aarch64-linux-gnu/lib -E WASMTIME_TEST_NO_HOG_MEMORY=1"
```

Also, I attached the `Cargo.lock` file (the `Cargo.toml` is the same as in the latest commit) and the log file from running `cargo test --target aarch64-unknown-linux-gnu`.

[aarch64-test.log](https://github.com/user-attachments/files/21747261/aarch64-test.log)
[Cargo.lock.txt](https://github.com/user-attachments/files/21747279/Cargo.lock.txt)

cfallin:
The `aarch64-binfmt-P` is suspect -- that seems to imply to me that some other configuration is causing Linux's binfmt support (i.e., automatic registration of qemu to allow `exec()` of foreign-architecture binaries directly) to run instead.

Question: do you have any other Cargo config, perhaps in your `~/.cargo`? And @alexcrichton would certainly know better than me about how Cargo finds configuration and whether there are any other gotchas here...

yomaytk:
> The aarch64-binfmt-P is suspect -- that seems to imply to me that some other configuration is causing Linux's binfmt support (i.e., automatic registration of qemu to allow exec() of foreign-architecture binaries directly) to run instead.

Yes, I agree. So I disabled the aarch64 binfmt and reran the tests, but they failed in the same tests. The error details have changed, but when I set the environment variables with `env` and enable binfmt, all the tests pass, so it looks like I need to review the configuration related to binfmt.

The results of some tests are as follows.

```shell
$ cargo test --target aarch64-unknown-linux-gnu big_table_in_pooling_allocator
...
running 1 test
test cli_tests::big_table_in_pooling_allocator ... FAILED

failures:

---- cli_tests::big_table_in_pooling_allocator stdout ----
Error: Failed to execute wasmtime with: ["tests/all/cli_tests/big_table.wat"]
status: exit status: 127

$ cargo test --target aarch64-unknown-linux-gnu bad_text_syntax
...
running 1 test
test cli_tests::bad_text_syntax ... FAILED

failures:

---- cli_tests::bad_text_syntax stdout ----

thread 'cli_tests::bad_text_syntax' panicked at tests/all/cli_tests.rs:816:5:
bad stderr: 
```
Just in case, I attached the full log file.

[aarch64-test-nobinfmt.log](https://github.com/user-attachments/files/21748257/aarch64-test-nobinfmt.log)

> do you have any other Cargo config, perhaps in your ~/.cargo? 

No, just .cargo/config.toml.

alexcrichton:
Aha ok this makes sense now. Some tests that we have will execute the `wasmtime` binary itself (e.g. CLI tests). This gets tricky when the test suite has a configure runner (e.g. cargo wrapped it in qemu) because the test doesn't know that it needs to also wrap the invocation of the `wasmtime` binary itself. Effectively the wrapper that Cargo uses may or may not be communicated to the test itself to further execute the cross-architecture binary.

Currently the test itself detecting a runner [only works if the runner is in an env var](https://github.com/bytecodealliance/wasmtime/blob/0e7876a09dc7da2241baa0a98f426d96ae966b7f/crates/test-util/src/lib.rs#L13-L21), not via Cargo configuration. This is what our CI runners configure, for example. If you're using `~/.cargo/config.toml` then Cargo won't set env vars and the test won't know it's running under emulation so it won't know to use the emulation when re-executing `wasmtime`.

Now if you have binfmt support enabled on Linux that's why the env vars appear to fix things for you. With binfmt support the test doesn't need to know to wrap the `wasmtime` invocation in QEMU, the kernel does that. The reason the env vars then work is that they're propagated automatically vs with CLI flags they're not propagated automatically.

So, in essence, I think this is a confluence of events where it's less to do with Cargo and more to do with how our test suite executes the `wasmtime` CLI and tries to do that correctly when under emulation. Unfortunately both before/after this PR suffer from not being 100% accurate:

* Before this PR the documented way of configuring a runner technically doesn't work because the env vars aren't set for the test suite to re-execute `wasmtime`. All parts of our test suite that don't execute commands (most of them) will still work however.
* After this PR the documented way of configuring will only work if binfmt support is configured. While your PR fixes systems with binfmt support setup, those systems without it still won't wrap the `wasmtime` execution with `qemu-*` (as you've seen)

The only "real" way to fix this is to export `CARGO_*_RUNNER` env vars like CI does, but that's a bit of a bummer. Ideally we could detect Cargo's `runner` configuration and/or detect we're under QEMU and just skip the tests.

yomaytk:
Thank you for the detailed analysis—I understand.
As you suggested, exporting `CARGO_TARGET_*_RUNNER` resolves the issue, so I’ve updated the code in that way. Could your review again?

> but that's a bit of a bummer.

Do you mean that it’s not ideal for .cargo/config.toml to depend on `CARGO_TARGET_*_RUNNER`?
`CARGO_TARGET_*_RUNNER` looks like a variable that’s intended to be used only in CI.

alexcrichton:
Would you mind updating this to recommend setting the env vars directly instead of `.cargo/config.toml`? That way `test-util/src/lib.rs` wouldn't need modifications or need to encode system-specific paths like it's currently doing.

Ah by "bummer" I mean that requiring env vars is not fun since they're not shared by shells by default (unlike `.cargo/config.toml`). What you've got here is a bit of a quine-like situation where you want to specify a runner that specifies env vars to specify the runner which would need to specify env vars which ... (so on and so forth). What this really just wants is the env var to be inherited, or some way for the test itself to learn the runner configuration from Cargo's config, but only the former is possible.

yomaytk:
Thanks! Exporting `CARGO_*_RUNNER` is simple and requires no code changes. However, since the current `cargo_test_runner` is implemented to use the first runner in the iterator, when we want to run tests for multiple targets, we have to unset all non-target `CARGO_*_RUNNER` variables each time. Then, I’m considering two approaches:

1. Introduce an architecture-agnostic `CARGO_TARGET_RUNNER` and set it each time we switch targets.:
`$ export CARGO_TARGET_RUNNER=‘qemu-aarch64 -L /usr/aarch64-linux-gnu -E LD_LIBRARY_PATH=/usr/aarch64-linux-gnu/lib -E WASMTIME_TEST_NO_HOG_MEMORY=1’`.

2. Introduce `CARGO_TARGET_ARCH` representing the target, and fix `cargo_test_runner` to detect the appropriate runner based on this environment variable. Run tests like:
`$ CARGO_TARGET_ARCH=aarch64 cargo test --target aarch64-unknown-linux-gnu.`

For reference, the latest changes in this PR implement the second approach.

There may be a better method; I’d appreciate your thoughts.

yomaytk:
Thanks! Exporting `CARGO_*_RUNNER` is simple and requires no code changes. However, since the current `cargo_test_runner` is implemented to use the first runner in the iterator, when we want to run tests for multiple targets, we have to unset all non-target `CARGO_*_RUNNER` variables each time. Then, I’m considering the following approaches:

- Introduce `CARGO_TARGET_ARCH` representing the target, allowing `cargo_test_runner` to detect the appropriate runner. The same variable will also be defined in CI. Run local tests like:
`$ CARGO_TARGET_ARCH=aarch64 cargo test --target aarch64-unknown-linux-gnu`.

The latest changes in this PR implement this approach.

There may be a better method; I’d appreciate your thoughts.

alexcrichton:
Thanks for your work here, and I'm sorry if this is taking longer than expected. This is unfortunately a bit of a minefield where there's bugs that can't be fixed and I also have personal preferences about how we document and structure this. I also ideally want to have minimal impact on various bits of infrastructure since this is a relatively niche thing that shouldn't haev an outside impact. 

This PR as-is adds a new required environment variable that's required to be specified to run tests in QEMU. Additionally it specifies configuration which is redundant in `.cargo/config.toml` where there's non-obvious overlap and surprising behavior.

If you're interested in seeing this PR all the way through, I'd recommend a possible change of removing the `.cargo/config.toml` configuration entirely and documenting only the `CARGO_*_RUNNER` env vars with an example of how to do one. Personally I think that would be sufficient for now and it's not super pressing to get the tests to a working state where you can set all the env vars for all the targets and run tests all at once for many targets.

If you'd like to get the multi-target approach working I'd recommend using `cfg(target_arch = "...")` instead of a new environment variable to figure out which one to filter for and look for.

Orthogonally to this PR, another option is to skip tests that use `Command` by default in cross-compiled situations. That way CI could opt-in to running these tests but others running `cargo test` could use the preexisting `.cargo/config.toml` configuration and it would work as expected.

---

Basically I want to avoid redundant configuration, hardcoding system-specifics in the repo, or adding undue burden on infrastructure for running these tests. It's unfortunately not easy to balance all these concerns though.

yomaytk:
Ah, I see—thank you for the detailed explanation. I understand that the proposed feature is niche and that it wouldn’t be good to place a burden on the infrastructure. Then, I’ll update this PR to document how to set `CARGO_*_RUNNER`.

> documenting only the CARGO_*_RUNNER env vars

Just to confirm, do we also need to document `CARGO_*_LINKER`?

alexcrichton:
The linker bits don't *need* to be in env vars, but they do need to be specified somewhere. If the runner is in an env var I think it'd make sense to follow suit and use that for the linker too (no need to document two similar-and-related configuration systems here)

yomaytk:
Got it. I’ve updated the PR—could you review it again?

PR #11421, opened 2025-08-12
Add a cwasm produced during testing to `.gitignore`

alexcrichton:
Added in wasip3-prototyping but never made its way upstream here, so doing so now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ cranelift/isle/veri/veri_engine/test_output
 crates/explorer/node_modules
 tests/all/pulley_provenance_test.cwasm
 tests/all/pulley_provenance_test_component.cwasm
+tests/all/pulley_provenance_test_async_component.cwasm
 /artifacts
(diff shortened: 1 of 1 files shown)


cfallin (approved):

cfallin:
This seems fine, but would it make sense to denylist all `*.cwasm`s instead, with something like `/**/*.cwasm`? We shouldn't ever check one in to source control, IMHO, because of the version evolution/bootstrapping headaches that would cause...

bjorn3:
Why are the .cwasm files generated inside the source directory rather than the target directory by the way? Having an immutable source directory is not uncommon for distros.

alexcrichton:
It's an artifact of [this test script](https://github.com/bytecodealliance/wasmtime/blob/main/ci/miri-provenance-test.sh) which is only architected as-is out of convenience, it wouldn't be too hard to modify that to put artifacts elsewhere.

PR #11425, opened 2025-08-13
Use the new section/symbol_flags_mut helpers in cranelift-object

bjorn3:
This ensures that the correct default flags are used.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -347,7 +347,7 @@ json-from-wast = "0.236.0"
 arbitrary = "1.4.0"
 mutatis = "0.3.2"
 cc = "1.0"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -347,7 +347,7 @@ json-from-wast = "0.236.0"
 arbitrary = "1.4.0"
 mutatis = "0.3.2"
 cc = "1.0"
-object = { version = "0.37.0", default-features = false, features = ['read_core', 'elf'] }
+object = { version = "0.37.3", default-features = false, features = ['read_core', 'elf'] }
```
bjorn3, line 350: This is the first release introducing these flags. It also contains a fix for emitting weak symbols on Windows.

alexcrichton (approved):

PR #11439, opened 2025-08-15
feat(p3): begin `wasi:http` implementation

rvolosatovs:
Move tests from https://github.com/bytecodealliance/wasip3-prototyping and generate bindings to simplify future development and review iterations

refs https://github.com/bytecodealliance/wasip3-prototyping/issues/228
refs https://github.com/bytecodealliance/wasip3-prototyping/issues/109


crates/wasi-http/src/p3/conv.rs
```diff
@@ -0,0 +1,105 @@
+use crate::p3::bindings::http::types::{ErrorCode, Method, Scheme};
+use core::convert::Infallible;
+
+impl From<Infallible> for ErrorCode {
+    fn from(_: Infallible) -> Self {
+        unreachable!()
```
alexcrichton: You can avoid `unreachable!()` here with `match x {}` by replacing the `_` above with `x`

crates/wasi-http/tests/all/http_server.rs
```diff
@@ -32,7 +33,8 @@ pub struct Server {
 
 impl Server {
     fn new<F>(
-        run: impl FnOnce(TokioIo<tokio::net::TcpStream>) -> F + Send + 'static,
+        run: impl Fn(TokioIo<tokio::net::TcpStream>) -> F + Send + 'static,
+        conns: usize,
```
alexcrichton: Mind swapping the order here? (idiomatically closures come last in arguments)

alexcrichton (commented):
Could the `#[ignore]` annotations be changed to `#[should_panic]` or similar? I'm not sure that works with returning a `Result<_>` though... If not no worries.

Otherwise this feels reasonable to me, land the scaffolding/tests here and slowly get them working one-by-one.

rvolosatovs:
> Could the `#[ignore]` annotations be changed to `#[should_panic]` or similar? I'm not sure that works with returning a `Result<_>` though... If not no worries.

unfortunately, does not look like this will work:
```
error: functions using `#[should_panic]` must return `()`
  --> crates/wasi-http/tests/all/p3/incoming.rs:34:1
   |
34 | #[test_log::test(tokio::test)]
```

crates/wasi-http/tests/all/http_server.rs
```diff
@@ -107,7 +115,7 @@ impl Server {
 
 impl Drop for Server {
     fn drop(&mut self) {
-        tracing::debug!("shutting down http1 server");
+        debug!("shutting down http1 server");
         // Force a connection to happen in case one hasn't happened already.
         let _ = TcpStream::connect(&self.addr);
```
alexcrichton: I think this may need to be done `self.conn` times? (or thread that through here perhaps)

alexcrichton (approved):

PR #11440, opened 2025-08-15
feat(p3): implement `wasi:http`

rvolosatovs:
This PR builds on top of #11221 and will contain the full implementation of `wasi:http` proposal with refcounting (no child resources). This will be made ready-for-review once all currently ignored tests pass

https://github.com/bytecodealliance/wasmtime/pull/11440/commits/4819aade1167a937d6c00f6026eb5451d2f100e3 is currently the only commit here worth looking at


crates/wasi-http/src/p3/host/mod.rs
```diff
@@ -1,2 +1,90 @@
...
+}
+
+fn get_fields_mut<'a>(
+    table: &'a mut ResourceTable,
+    fields: &Resource<Fields>,
+) -> wasmtime::Result<&'a mut Fields> {
+    table
+        .get_mut(&fields)
+        .context("failed to get fields from table")
+}
```
alexcrichton, line 24: One suggestion perhaps for many of these methods is to have a generic extension trait which supports an arbitrary `T` for `Resource<T>` which could help deduplicate these methods?

crates/wasi-http/src/p3/host/types.rs
```diff
@@ -1,84 +1,246 @@
...
+        fields.remove(&name);
+        Ok(Ok(()))
     }
 
     fn get_and_delete(
         &mut self,
-        self_: Resource<Fields>,
+        fields: Resource<Fields>,
         name: FieldName,
     ) -> wasmtime::Result<Result<Vec<FieldValue>, HeaderError>> {
```
alexcrichton, line 289: Could this use `trappable_error` configuration to aovid the double-layer of `Result`? That would also enable using `?` below to propagate errors.

rvolosatovs:
I'm done for the week and will continue on Monday, but this is pretty close to having all test cases passing.

@alexcrichton @dicej feel free to make the necessary changes in the PR to make the rest of the test cases work and get this over the line if there's a desire to still finish `wasi:http` this week.
Alternatively, merging this PR as-is and addressing the rest in follow-ups would work as well.

If we're okay with waiting until next week, this should be complete by Monday evening

crates/wasi-http/tests/all/p3/mod.rs
```diff
@@ -40,3 +66,373 @@ impl WasiHttpView for Ctx {
...
+    run_cli(P3_HTTP_OUTBOUND_REQUEST_INVALID_PORT_COMPONENT, &server).await
+}
+
+#[test_log::test(tokio::test(flavor = "multi_thread"))]
+async fn p3_http_outbound_request_invalid_dnsname() -> anyhow::Result<()> {
+    let server = Server::http1(1)?;
+    run_cli(P3_HTTP_OUTBOUND_REQUEST_INVALID_DNSNAME_COMPONENT, &server).await
+}
+
+#[ignore = "unimplemented"] // TODO: implement
```
alexcrichton, line 219: Debugging this: in my opinion the bug is that dropping a `request` resource does not properly close its stream handles. That's not currently possible, however, because during `drop` there's no access to the store. I made https://github.com/bytecodealliance/wasmtime/pull/11628 to fix this.

crates/wasi-http/tests/all/p3/mod.rs
```diff
@@ -40,3 +66,373 @@ impl WasiHttpView for Ctx {
...
+#[test_log::test(tokio::test(flavor = "multi_thread"))]
+async fn p3_http_outbound_request_large_post() -> anyhow::Result<()> {
+    let server = Server::http1(1)?;
+    run_cli(P3_HTTP_OUTBOUND_REQUEST_LARGE_POST_COMPONENT, &server).await
+}
+
+#[ignore = "unimplemented"] // TODO: implement
```
alexcrichton, line 172: I've debugged these three ignores and it's due to [this code](https://github.com/bytecodealliance/wasmtime/blob/a655cc19550c1381ea263ca3a1ec52c4bde53ff6/crates/test-programs/src/p3/http.rs#L101-L109) needing to happen concurrently with `handler::handle` now instead of synchronously after. It's easy enough to lift that `async` block to the `join!` above except then it runs into the panic described in https://github.com/bytecodealliance/wasmtime/issues/11621 so these will have to wait for that to get fixed

crates/wasi-http/src/p3/response.rs
```diff
@@ -0,0 +1,90 @@
...
+                Box::into_pin(fut).await
+            },
+        )
+    }
+
+    /// Convert [Response] into [http::Response].
+    pub fn into_http<T: WasiHttpView + 'static>(
+        self,
+        store: impl AsContextMut<Data = T>,
+        fut: impl Future<Output = Result<(), ErrorCode>> + Send + 'static,
```
alexcrichton, line 69: Could you comment this function to the purpose of `fut`?

crates/wasi-http/src/p3/response.rs
```diff
@@ -0,0 +1,90 @@
...
+
+impl Response {
+    /// Construct a new [Response]
+    pub fn new(
+        status: StatusCode,
+        headers: impl Into<Arc<HeaderMap>>,
+        body: impl Into<BoxBody<Bytes, ErrorCode>>,
+    ) -> (
+        Self,
+        impl Future<Output = Result<(), ErrorCode>> + Send + 'static,
```
alexcrichton, line 46: Could you comment function as to what this future is?

crates/wasi-http/src/p3/request.rs
```diff
@@ -0,0 +1,289 @@
...
+            },
+        )
+    }
+
+    /// Construct a new [Request] from [http::Request].
+    pub fn from_http<T>(
+        req: http::Request<T>,
+    ) -> (
+        Self,
+        impl Future<Output = Result<(), ErrorCode>> + Send + 'static,
```
alexcrichton, line 80: Like response below, could you comment this returned future here and with `new`?

crates/wasi-http/src/p3/request.rs
```diff
@@ -0,0 +1,289 @@
...
+                        Err(err) => Poll::Ready(Err(err)),
+                    },
+                    Err(err) => Poll::Ready(Err(ErrorCode::from_hyper_request_error(err))),
+                }
+            } else {
+                Poll::Pending
+            }
+        }
+    })
+    .await?;
```
alexcrichton, line 282: This may be preexisting with wasip2 as well, but I'm having a tough time piecing together what's going on here. Mind leaving some comments to what this is doing?

crates/wasi-http/src/p3/proxy.rs
```diff
@@ -0,0 +1,37 @@
...
+use anyhow::Context as _;
+use wasmtime::component::Accessor;
+
+impl Proxy {
+    /// Call `handle` on [Proxy] getting a [Future] back.
+    pub async fn handle(
+        &self,
+        store: &Accessor<impl WasiHttpView>,
+        req: impl Into<Request>,
+    ) -> wasmtime::Result<Result<Response, ErrorCode>> {
```
alexcrichton, line 13: Could this use a `TrappableError` from the `wasmtime_wasi` crate like filesystem & sockets?

crates/wasi-http/src/p3/mod.rs
```diff
@@ -90,3 +184,66 @@ where
     types::add_to_linker::<_, WasiHttp>(linker, T::http)?;
     Ok(())
 }
+
+/// An [Arc], which may be immutable.
```
alexcrichton, line 188: Mind expanding the docs here to explain what this is used for? For example how headers are immutable when read from some locations but mutable in others. Regardless everything uses an `Arc` and exhibits copy-on-write behavior.

crates/wasi-http/src/p3/mod.rs
```diff
@@ -90,3 +184,66 @@ where
...
+
+    pub fn new_immutable(v: impl Into<Arc<T>>) -> Self {
+        Self::Immutable(v.into())
+    }
+
+    fn as_arc(&self) -> &Arc<T> {
+        match self {
+            Self::Mutable(v) | Self::Immutable(v) => v,
+        }
+    }
```
alexcrichton, line 228: Could this and `get` get removed in favor of the `Deref` impl?

crates/wasi-http/src/p3/mod.rs
```diff
@@ -9,24 +9,118 @@
...
+
+    /// Scheme to default to, when not set by the guest.
+    ///
+    /// If [None], `handle` will return [ErrorCode::HttpProtocolError]
+    /// for requests missing a scheme.
+    fn default_scheme(&mut self) -> Option<Scheme> {
+        Some(Scheme::HTTPS)
+    }
+
+    /// Send an outgoing request.
```
alexcrichton, line 71: Could you expand the documentation here to include what the various layers of futures are and what their meaning is?

crates/wasi-http/src/p3/host/handler.rs
```diff
@@ -1,15 +1,118 @@
...
+                path_with_query,
+                headers,
+                options,
+                body,
+            } = delete_request(table, req).map_err(HttpError::trap)?;
+            let mut headers = Arc::unwrap_or_clone(headers);
+            if ctx.set_host_header() {
+                let host = if let Some(authority) = authority.as_ref() {
+                    HeaderValue::try_from(authority.as_str())
+                        .map_err(|err| ErrorCode::InternalError(Some(err.to_string())))?
```
alexcrichton, line 36: Technically this is a slightly dangerous `?` (and everything below before the body is consumed) because if the body is a guest body the resources won't get properly closed.

Would it make sense to lift the body construction all the way to the top here so that happens first?

crates/wasi-http/src/p3/host/handler.rs
```diff
@@ -1,15 +1,118 @@
...
+            HttpResult::Ok((
+                store.get().ctx.send_request(
+                    req,
+                    options.as_deref().copied(),
+                    Box::new(async {
+                        let Ok(fut) = res_result_rx.await else {
+                            return Ok(());
+                        };
+                        Box::into_pin(fut).await
+                    }),
```
alexcrichton, line 88: This is one of the reasons I'm seeking more documentation on the various future parameters. This and below looks to be a very careful and particular dance with respect to various errors but I'm not really sure what's going on here in the abstract. I would hope that we can figure out a way to simplify this, but in lieu of that I'm hoping for more docs as to what's going on.

crates/wasi-http/src/p3/host/handler.rs
```diff
@@ -1,15 +1,118 @@
...
+                ),
+                result_tx,
+            ))
+        })?;
```
alexcrichton, line 92: Is this `HttpResult::Ok(( ... ))?` still needed? (vs just "do the thing")

crates/wasi-http/src/p3/request.rs
```diff
@@ -0,0 +1,289 @@
...
+            return Err(dns_error("address not available".to_string(), 0));
+        }
+        Ok(Err(err))
+            if err
+                .to_string()
+                .starts_with("failed to lookup address information") =>
+        {
+            return Err(dns_error("address not available".to_string(), 0));
+        }
+        Ok(Err(..)) => return Err(ErrorCode::ConnectionRefused),
```
alexcrichton, line 198: I realize this is probably copied from wasip2, but could this `log::warn!` the error that's discarded here?

crates/wasi-http/src/p3/body.rs
```diff
@@ -0,0 +1,257 @@
...
+            }
+            Err(err) => {
+                _ = self.tx.send(Err(err));
+                Ok(())
+            }
+        }
+    }
+}
+
+pub(crate) struct IncomingResponseBody {
```
alexcrichton, line 221: Mind adding a small doc block here explaining what this is doing? The code is pretty self-explanatory but I still think it'd be good to have an intro

crates/wasi-http/src/p3/body.rs
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+    fn size_hint(&self) -> http_body::SizeHint {
+        http_body::SizeHint::with_exact(0)
+    }
+}
+
+pub(crate) struct GuestTrailerConsumer<T> {
+    pub(crate) tx: oneshot::Sender<Result<Option<Arc<HeaderMap>>, ErrorCode>>,
+    pub(crate) getter: for<'a> fn(&'a mut T) -> WasiHttpCtxView<'a>,
```
alexcrichton, line 188: If you'd like you can drop the `for<'a>` here and use `WasiHttpCtxView<'_>` at the end

crates/wasi-http/src/p3/body.rs
```diff
@@ -0,0 +1,257 @@
...
+        self.trailers_rx = None;
+        match res {
+            Ok(Ok(Some(trailers))) => Poll::Ready(Some(Ok(http_body::Frame::trailers(
+                Arc::unwrap_or_clone(trailers),
+            )))),
+            Ok(Ok(None)) => Poll::Ready(None),
+            Ok(Err(err)) => Poll::Ready(Some(Err(err))),
+            Err(..) => Poll::Ready(None),
+        }
+    }
```
alexcrichton, line 140: Mind leaving some comments here (and in `is_end_stream`) about the state machine being managed here? Doesn't need to be too wordy but having headings to grasp on I think would be helpful

crates/wasi-http/src/p3/host/types.rs
```diff
@@ -1,290 +1,807 @@
...
+                                getter,
+                            },
+                        ),
+                        FutureReader::new(instance, &mut store, FutureOneshotProducer(trailers_rx)),
+                    )))
+                }
+                Body::Consumed => Ok(Err(())),
+            }
+        })
     }
```
alexcrichton: Is it possible to share this with `Request::consume_body` since it's more-or-less the same function?

crates/wasi-http/src/p3/host/types.rs
```diff
@@ -1,290 +1,807 @@
...
+    type Item = u8;
+    type Buffer = Cursor<Bytes>;
+
+    fn poll_produce<'a>(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        mut store: StoreContextMut<'a, D>,
+        mut dst: Destination<'a, Self::Item, Self::Buffer>,
+        finish: bool,
+    ) -> Poll<wasmtime::Result<StreamResult>> {
```
alexcrichton, line 131: Mind leaving some small comments in this function body to help navigate what the state machine is doing? (e.g. a ocmment explaining the signficance of each arm here and there)

alexcrichton (approved):
Ok I've gone over this and it looks good to me, thanks again @rvolosatovs!

I've left a bunch of comments below which I'd like to resolve at some point, but doesn't need to block. I'm going to flag this for merge and I'll file follow-ups.

crates/test-programs/src/bin/p3_http_outbound_request_content_length.rs
```diff
@@ -44,84 +44,93 @@ fn make_request() -> (
...
+                    assert!(
+                        remaining.is_empty(),
+                        "{}",
+                        String::from_utf8_lossy(&remaining)
+                    );
+                    drop(contents_tx);
+                    trailers_tx.write(Ok(None)).await.unwrap();
+                },
             );
+            // FIXME(#11631) should have a more precise error like the one
```
rvolosatovs: There's currently no `Content-Length` header handling, so I would not modify this test until that's implemented in the host.

crates/wasi-http/tests/all/p3/mod.rs
rvolosatovs: I forgot to port the test response streaming, which I've done in wasip3-prototyping precisely to fix this. I think it's useful to specifically test the fact that `handle` returns before response body are streamed, so I'd rather not rearrange the `join!`.

This is fixed in https://github.com/bytecodealliance/wasmtime/pull/11636

crates/wasi-http/tests/all/p3/mod.rs
rvolosatovs: Exactly, I've used the feature you've added to fix it in https://github.com/bytecodealliance/wasmtime/pull/11636

crates/wasi-http/src/p3/proxy.rs
rvolosatovs: I'm not sure that's a good idea for this particular one, since that would require exposing the `HttpError` to embedders and they'd need to handle the `trap` case themselves.

While this signature is definitely not ideal, IMO it's clearer and it aligns with what `call_handle` would otherwise return if called directly

PR #11441, opened 2025-08-15
Require a store in `catch_unwind_and_record_trap`

alexcrichton:
This commit does some preparatory refactoring for #11326 to ensure that a store is available when trap information is being processed. Currently this doesn't leverage the new parameter but it should be leverage-able in #11326.

### DIFF
--- a/crates/wasmtime/src/runtime/func.rs
+++ b/crates/wasmtime/src/runtime/func.rs
@@ -2062,8 +2062,6 @@ impl<T> Caller<'_, T> {
             (gc_lifo_scope, ret)
         };
 
-        // Safe to recreate a mutable borrow of the store because `ret`
-        // cannot be borrowing from the store.
         store.0.exit_gc_lifo_scope(gc_lifo_scope);
 
         ret
--- a/crates/wasmtime/src/runtime/trampoline/func.rs
+++ b/crates/wasmtime/src/runtime/trampoline/func.rs
@@ -38,7 +38,7 @@ where
     F: Fn(&mut dyn VMStore, InstanceId, &mut [ValRaw]) -> Result<()> + 'static,
 {
     // SAFETY: this is an entrypoint of wasm calling a host and our parameters
-    // should reflect that making `enter_host_from_wasm` suitable. Futhter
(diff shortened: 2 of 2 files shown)


cfallin:
Fantastic, thank you for this! I'll rebase/refactor on top of it -- will be much cleaner for exception processing, as we've discussed.

crates/wasmtime/src/runtime/trampoline/func.rs
```diff
@@ -32,41 +35,45 @@ unsafe extern "C" fn array_call_shim<F>(
...
-    // boundary, and then otherwise delegate as normal.
-    crate::runtime::vm::catch_unwind_and_record_trap(|| {
-        // SAFETY: this function itself requires that the `vmctx` is valid to
-        // use here.
-        let state = unsafe {
-            let vmctx = VMArrayCallHostFuncContext::from_opaque(vmctx);
-            vmctx.as_ref().host_state()
-        };
+    // SAFETY: this is an entrypoint of wasm calling a host and our parameters
+    // should reflect that making `enter_host_from_wasm` suitable. Futhter
```
fitzgen:
```suggestion
    // should reflect that making `enter_host_from_wasm` suitable. Further
```

crates/wasmtime/src/runtime/func.rs
```diff
@@ -2041,46 +2042,31 @@ impl<T> Caller<'_, T> {
...
-                    (gc_lifo_scope, ret)
-                };
+            (gc_lifo_scope, ret)
+        };
 
-                // Safe to recreate a mutable borrow of the store because `ret`
-                // cannot be borrowing from the store.
-                store.0.exit_gc_lifo_scope(gc_lifo_scope);
+        // Safe to recreate a mutable borrow of the store because `ret`
+        // cannot be borrowing from the store.
```
fitzgen: I think this comment is outdated now that this isn't using any raw pointers anymore?

fitzgen: (That is, the borrow checker can statically check for the double-mutable-borrowing that this comment was previously addressing)

PR #11444, opened 2025-08-15
Delete `StoreOpaque::traitobj_mut`

alexcrichton:
This was a fundamentally unsound function because it is widening a mutable borrow to encompass more than it originally contained. Removal of its usage falls into a few buckets here:

* Many instances of `store.0.traitobj_mut()` were starting from a `&mut StoreInner<T>`, going to `&mut StoreOpaque`, then going back to `&mut dyn VMStore`. These are replaced with `store.0` as `&mut StoreInner<T>` directly coerces into `&mut dyn VMStore`.

* Some fiber-related helpers were updated to take any store-opaque-thing which encapsulates the ability to start/resume a borrow with a part of the store, but you only get that part of the store during the fiber itself. This means there's no widening necessary.

* Some methods previously taking `&mut StoreOpaque` are now appropriately widened to `&mut dyn VMStore`.

This all enables full deletion of this function which prevents accidentally ever tripping over its unsound-ness.

### DIFF
--- a/crates/wasmtime/src/runtime/fiber.rs
+++ b/crates/wasmtime/src/runtime/fiber.rs
@@ -112,11 +112,17 @@ impl AsyncState {
     }
 }
 
-trait AsStoreOpaque {
+pub(crate) trait AsStoreOpaque {
     fn as_store_opaque(&mut self) -> &mut StoreOpaque;
 }
 
 impl AsStoreOpaque for StoreOpaque {
+    fn as_store_opaque(&mut self) -> &mut StoreOpaque {
+        self
+    }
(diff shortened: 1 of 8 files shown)


crates/wasmtime/src/runtime/fiber.rs
```diff
@@ -112,7 +112,7 @@ impl AsyncState {
     }
 }
 
-trait AsStoreOpaque {
+pub(crate) trait AsStoreOpaque {
```
fitzgen: Should this trait and its implementations move to `store.rs`?

PR #11453, opened 2025-08-18
Add a helper always-`Sync` utility to Wasmtime

alexcrichton:
This commit adds a newtype wrapper to Wasmtime, `AlwaysMut`, which is unconditionally `Sync` if the stored type is `Send`. This is similar to a `Mutex<T>` where it promotes a `Send` bound to a `Sync` bound, but it's unlike `Mutex<T>` in that `AlwaysMut<T>` has no synchronization. The reason that this is safe is that `AlwaysMut<T>` completely disallows access to the underlying data through `&self` and requires `&mut self`. This is similar to how `Mutex::get_mut` is safe, for example.

This type cleans up a preexisting `unsafe impl Sync` block in funcref management around `SendSyncBump` (`bumpalo::Bump` is `Send`, not `Sync`, but we only access it through `&mut self`). This then additionally removes `unsafe impl Sync for StoreFiber` which, upon reflection, is not sound because we don't ever constraint the store's `T` type to `Sync`, only `Send`. This is effectively no change throughout Wasmtime, however, as fibers are only accessed with `&mut`.

Overall this is mostly just internal refactoring to reduce the amount of `unsafe` inside of Wasmtime and to add a new utility to use in the future too.

### DIFF
--- a/crates/wasmtime/src/runtime/vm/always_mut.rs
+++ b/crates/wasmtime/src/runtime/vm/always_mut.rs
@@ -33,6 +33,7 @@ impl<T> AlwaysMut<T> {
     }
 
     /// Consume this [`AlwaysMut`], returning the underlying data.
+    #[cfg(feature = "async")]
     pub fn into_inner(self) -> T {
         self.0
     }


crates/wasmtime/src/runtime/vm/traphandlers.rs
```diff
@@ -990,6 +990,9 @@ pub(crate) mod tls {
         state: raw::Ptr,
     }
 
+    #[cfg(feature = "async")]
+    unsafe impl Send for AsyncWasmCallState {}
```
fitzgen, line 1000: Safety comment?

crates/wasmtime/src/runtime/fiber.rs
```diff
@@ -723,25 +732,31 @@ fn resume_fiber<'a>(
 
     impl Drop for Restore<'_, '_> {
         fn drop(&mut self) {
-            self.fiber.state = Some(unsafe { self.state.take().unwrap().replace(self.store) });
+            self.fiber.state = Some(AlwaysMut::new(unsafe {
+                self.state.take().unwrap().replace(self.store)
+            }));
```
fitzgen: Maybe add a `impl From<T> for AlwaysMut<T>` impl so tha this can just `.into()`?

PR #11454, opened 2025-08-18
 Relax the `Send` bound on fiber creation 

alexcrichton:
This commit removes the need for `Send` for all usages of fiber-related
bits. The goal is that the future returned from `async` functions is
`Send` if all the inputs are `Send`, but there's no actual need to
require that in the function signatures themselves. This property was
identified in upcoming work to make more internals of Wasmtime `async`
where `async` functions are going to be used to implement the
synchronous path through Wasmtime where `Send` isn't a bound today, nor
ideally do we want to have it there.

The way this commit works is:

* First `make_fiber` now has a `Send` bound which makes it sound. Before
  https://github.com/bytecodealliance/wasmtime/pull/11444 it took `&mut dyn VMStore` which is never `Send` and the
  refactoring there missed adding this bound. This change required a few
  `Send` bounds in `concurrent.rs` where it's expected to have no
  impact as everything there is basically already `Send`.

* Next `make_fiber_unchecked` is added which is the same as `make_fiber`
  except without a `Send` bound. The `on_fiber` function is then updated
  to use this `*_unchecked` variant. This means that fibers can now be
  used with non-`Send` stores. Crucially though the structure of
  internals in play means that future produced is still only `Send` if
  `T: Send` meaning that there is no loss in safety.

* Next some `Send` bounds were dropped throughout async functions in
  Wasmtime as a proof-of-concept to require these changes to `on_fiber`.
  This means that these entrypoints into Wasmtime no longer require
  `Send`, but still naturally require `Send` if the result future is
  sent to various threads.

* Finally a new doctest is added to `Instance::new_async` with a
  `compile_fail` example to ensure that this fails. This is
  unfortunately a very brittle test because all we can assert is that
  compilation failed, not why compilation failed. That means that this
  is likely to regress over time, but it's the best we can do at this
  time.

Note that the goal here is NOT to remove `Send` bounds throughout
Wasmtime. Many of them are still required, for example all the ones
related to `Linker`. The realization is that we can remove some of the
"edge" bounds on entrypoints where the resulting future can be
conditionally `Send` depending on `T` meaning we don't actually have to
write down any bounds ourselves.

This refactoring will enable future refactorings to have sync-and-async
functions use the same internals and neither entrypoint needs to have a
`Send` bound.

### DIFF
--- a/crates/wasmtime/src/runtime/instance.rs
+++ b/crates/wasmtime/src/runtime/instance.rs
@@ -178,6 +178,9 @@ impl Instance {
     ///     let engine = Engine::new(&config)?;
     ///
     ///     let module = Module::new(&engine, "(module)")?;
+    ///
+    ///     // Note that `Rc<()>` is NOT `Send`, which is what many future
+    ///     // runtimes require and below will cause a failure.
     ///     let mut store = Store::new(&engine, Rc::new(()));
     ///
     ///     // Compile failure because `Store<Rc<()>>` is not `Send`
@@ -898,10 +901,6 @@ impl<T: 'static> InstancePre<T> {
     ///
     /// Panics if any import closed over by this [`InstancePre`] isn't owned by
     /// `store`, or if `store` does not have async support enabled.
-    ///
-    /// ```compile_fail
-    /// what
-    /// ```
     #[cfg(feature = "async")]
     pub async fn instantiate_async(
         &self,


alexcrichton:
Procedurally this is built on https://github.com/bytecodealliance/wasmtime/pull/11444 and https://github.com/bytecodealliance/wasmtime/pull/11453. The goal is to work towards https://github.com/bytecodealliance/wasmtime/pull/11430

fitzgen:
I can review this once it is rebased, now that the things it was based on top of are approved (and presumably will land shortly)

alexcrichton:
I've rebased on the merge commit of https://github.com/bytecodealliance/wasmtime/pull/11444 since I need to head out soon, but if/when that PR merges github won't actually remove the commit from this PR, I think I'll have to open/close it or something like that (it's weird)

crates/wasmtime/src/runtime/instance.rs
```diff
@@ -137,9 +137,60 @@ impl Instance {
...
+    /// use std::rc::Rc;
+    ///
+    /// #[tokio::main]
+    /// async fn main() -> Result<()> {
+    ///     let mut config = Config::new();
+    ///     config.async_support(true);
+    ///     let engine = Engine::new(&config)?;
+    ///
+    ///     let module = Module::new(&engine, "(module)")?;
+    ///     let mut store = Store::new(&engine, Rc::new(()));
```
fitzgen, line 184: Maybe add a comment here that `Rc` is not `Send`, just to make it a little easier to follow the example. Alternatively, below say "Store<Rc<()>> is not Send because Rc<()> is not Send" or something.

crates/wasmtime/src/runtime/instance.rs
```diff
@@ -850,14 +898,15 @@ impl<T: 'static> InstancePre<T> {
     ///
     /// Panics if any import closed over by this [`InstancePre`] isn't owned by
     /// `store`, or if `store` does not have async support enabled.
+    ///
+    /// ```compile_fail
+    /// what
+    /// ```
```
fitzgen: ?

alexcrichton: what

(Forgot to delete after testing)

PR #11455, opened 2025-08-19
Add clif instructions tests (`imul_imm`, `get_return_address`) to riscv64 backend.

yomaytk:
The riscv64 backend seems to be missing tests for some CLIF instructions that are included in other architectures. This PR adds the `imul_imm` and `get_return_address` instructions.


fitzgen (approved):
Thanks!

FWIW, `imul_imm` should be legalized away (and turned into an `imul` with an `iconst` operand) at the start of the pipeline and shouldn't ever directly make it to the backend. That said, adding a clif test for what it ultimately compiles down to doesn't do any harm.

PR #11458, opened 2025-08-19
Make `Tunables::inlining_sum_size_threshold` a hard limit

fitzgen:
This prevents infinitely inlining small recursive functions whose size is less than `Tunables::small_callee_size`.


fitzgen:
Let me turn this fuzz input into a test case too...

alexcrichton (approved):
r=me with a test

PR #11459, opened 2025-08-19
Use RAII guards in a few locations instead of closures

alexcrichton:
This represents no functional difference from what happens today but in the future all of these functions will become `async` functions where RAII guards will then be required, so this goes ahead and switches them to RAII guards to split out the diff.


alexcrichton:
I'll also note that for review a whitespace-ignoring diff is probably best on this one.

fitzgen (approved):
LGTM 👍

Issue #4245, opened 2022-06-08
Tracking issue for shared memory


abrown:
This issue tracks anything incomplete with the effort to add shared memories started in #4187:

 - [x] https://github.com/bytecodealliance/wasm-tools/pull/624
 - [x] https://github.com/bytecodealliance/wasmtime/pull/4266
 - [x]  improve tests (@alexcrichton had a [comment](https://github.com/bytecodealliance/wasmtime/pull/4187#discussion_r891697133) about trying to stress things by doing more in Wasm)
 - [ ] add a `SharedMemoryType` to the `wasmtime` external API (@alexcrichton wanted this to disambiguate entirely between memories and shared memories
 - [x] enable shared memory in the C API (#7940)
 - [ ] #4244
 - [ ] #4240
 - [ ] #4203
 - [x] #4267 
 - [ ] #4268
 - [ ] Growth/creation of a shared memory is not subject to resource limiters 
 - [ ] Shared memories are disabled by default in Wasmtime
 - [ ] Pulley's reconcilation with shared memories may or may not have holes (unsure)
 - [ ] Behavior of wait/notify aren't customizable #12809

Issue #4244, opened 2022-06-08
shared memory: use shared memories with the pooling allocator


abrown:
The pooling allocator creates memories in a specific region that is torn down with the instance. In #4187, shared memories are implemented to own their own memory. It is unclear what changes are needed at this point to make shared memories work with the pooling allocator (e.g., assign shared memories to a store?) but this issue serves to track the resolution of this problem.

Issue #4240, opened 2022-06-07
shared memory: no way to report memory growth failure


abrown:
In certain cases, a shared memory will not have an associated store. A `ResourceLimiter` is attached to a store to not only constrain the resources use but also accept certain error scenarios, e.g., when a memory growth operation fails. If memory growth fails for a shared memory, some code should handle the error but as of #4187 there is no mechanism to do so.

abrown:
@alexcrichton, should the resolution of this just be to return an error if no `Store` (and thus `ResourceLimiter`) are available? E.g.:

```rust
if let Some(store) = store {
  store.memory_grow_failed(&format_err!("Memory maximum size exceeded"));
} else {
  bail!("Memory maximum size exceeded");
}
```

alexcrichton:
I don't think so because that would translate to a trap rather than a `-1` return value from `memory.grow`. Additionally the more worrisome case for me is [this one](https://github.com/bytecodealliance/wasmtime/blob/a2d49ebf2779a88650f8ed624019d63300177804/crates/runtime/src/memory.rs#L103-L108) where there's no embedder-defined way to limit the growth of a shared memory like there is a way to limit the growth of a non-shared memory.

abrown:
Hm. I had originally thought that a) the `ResourceLimiter` would have to be disconnected from a `Store` or b) the shared memory would have to be associated with a store--neither of these seems great... or easy. What do you think?

alexcrichton:
I don't personally know how best to solve this, and I suspect that reusing what we have today probably won't cut it and this'll need some sort of new abstraction or something like that.

PR #11460, opened 2025-08-19
 Make memory growth an async function 

alexcrichton:
This is an analog of https://github.com/bytecodealliance/wasmtime/pull/11442 but for memories. This had a little more
impact due to memories being hooked into GC operations. Further
refactoring of GC operations to make them safer/more-async is deferred
to a future PR and for now it's "no worse than before". This is another
step towards https://github.com/bytecodealliance/wasmtime/pull/11430 and enables removing a longstanding `unsafe` block
in `runtime/memory.rs` which previously could not be removed.

One semantic change from this is that growth of a shared memory no
longer uses an async limiter. This is done to keep growth of a shared
memory consistent with creation of a shared memory where no limits are
applied. This is due to the cross-store nature of shared memories which
means that we can't tie growth to any one particular store. This
additionally fixes an issue where an rwlock write guard was otherwise
held across a `.await` point which creates a non-`Send` future, closing
a possible soundness hole in Wasmtime.

### DIFF
--- a/crates/wasmtime/src/runtime/vm/instance.rs
+++ b/crates/wasmtime/src/runtime/vm/instance.rs
@@ -681,6 +681,9 @@ impl Instance {
     ) -> Result<Option<usize>, Error> {
         let memory = &mut self.as_mut().memories_mut()[idx].1;
 
+        // SAFETY: this is the safe wrapper around `Memory::grow` because it
+        // automatically updates the `VMMemoryDefinition` in this instance after
+        // a growth operation below.
         let result = unsafe { memory.grow(delta, limiter).await };
 
         // Update the state used by a non-shared Wasm memory in case the base
--- a/crates/wasmtime/src/runtime/vm/instance.rs
+++ b/crates/wasmtime/src/runtime/vm/instance.rs
@@ -17,7 +17,7 @@ use crate::runtime::vm::{
     GcStore, HostResult, Imports, ModuleRuntimeInfo, SendSyncPtr, VMGlobalKind, VMStore,
     VMStoreRawPtr, VmPtr, VmSafe, WasmFault, catch_unwind_and_record_trap,
 };
-use crate::store::{InstanceId, StoreId, StoreInstanceId, StoreOpaque};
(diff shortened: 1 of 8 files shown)


alexcrichton:
Note that the first commit here is from https://github.com/bytecodealliance/wasmtime/pull/11459

fitzgen:
> One semantic change from this is that growth of a shared memory no
> longer uses an async limiter. This is done to keep growth of a shared
> memory consistent with creation of a shared memory where no limits are
> applied. This is due to the cross-store nature of shared memories which
> means that we can't tie growth to any one particular store.

Seems like we should call the async limiter of the store that is doing the growth, that way an embedder could still guarantee that all growth is limited from a holistic system point of view, even if one particular store's limiter doesn't handle all the grow calls.

Mind filing a follow up issue for this?

alexcrichton:
That's theoretically possible yeah but it's already preexisting that the resource limiter isn't used when a `SharedMemory` is created through the embedder API. There's also still the issue that the rwlock write guard can't be held across an await point which would be a much more invasive change.

I've added a note to https://github.com/bytecodealliance/wasmtime/issues/4245 to accompany similar issues in https://github.com/bytecodealliance/wasmtime/issues/4240 and https://github.com/bytecodealliance/wasmtime/issues/4244

crates/wasmtime/src/runtime/vm/instance.rs
```diff
@@ -673,15 +673,15 @@ impl Instance {
...
         mut self: Pin<&mut Self>,
-        store: &mut dyn VMStore,
+        limiter: Option<&mut StoreResourceLimiter<'_>>,
         idx: DefinedMemoryIndex,
         delta: u64,
     ) -> Result<Option<usize>, Error> {
         let memory = &mut self.as_mut().memories_mut()[idx].1;
 
-        let result = unsafe { memory.grow(delta, Some(store)) };
+        let result = unsafe { memory.grow(delta, limiter).await };
```
fitzgen, line 687: Mind adding a safety comment while you're here?

Issue #11409, opened 2025-08-08
Implementation of GC is not sound with resource limiters


alexcrichton:
This line of code:

https://github.com/bytecodealliance/wasmtime/blob/b500820efe1576188653f9d710f7d21cdb4e39be/crates/wasmtime/src/runtime/store/gc.rs#L124

expands a borrow of `&mut StoreOpaque` (`self` in this case) into a `&mut dyn VMStore` which is effectively a `&mut StoreInner<T>`. This is a "widening" operation which makes the original mutable borrow wider by encompassing more fields. This is not a sound operation because a caller might simultaneously be borrowing `&mut StoreOpaque` and `&mut T` or similarly.

alexcrichton:
This is also arguably a problem during [instance allocation](https://github.com/bytecodealliance/wasmtime/blob/28f978355509b2a2802cc74388e6feb0be8d0571/crates/wasmtime/src/runtime/store.rs#L2088) and allocation [of the gc heap](https://github.com/bytecodealliance/wasmtime/blob/28f978355509b2a2802cc74388e6feb0be8d0571/crates/wasmtime/src/runtime/store.rs#L1459)

PR #11461, opened 2025-08-19
 Make garbage collection an `async` function 

alexcrichton:
This commit is similar to https://github.com/bytecodealliance/wasmtime/pull/11442 and https://github.com/bytecodealliance/wasmtime/pull/11460 except it's applied to the
garbage collection phase of Wasmtime's GC. Specifically the functions
that actually perform a GC are no longer duplicated across sync, async,
and maybe async versions. There's only one "always async" version and
the root-level crate entrypoints for sync versions assert that async
support is disabled and then use the async version.

Worth noting here is that GC suffers from a preexisting issue described
in https://github.com/bytecodealliance/wasmtime/issues/11409 where it's not sound how a `StoreOpaque` is widened to acquire
a resource limiter. This commit seemingly makes the issue worse by
adding a few more `unsafe` blocks, but they're all fundamentally doing
the same thing as before. Fully solving this issue will require making
memory/table creation an `async` function that takes the limiter as an
argument. Doing this will require further refactoring/code movement so
my goal is to effectively maintain the status quo, but in a slightly
different location, and enable knocking out the `unsafe` in the future.
In the meantime the previous `unsafe` block is "lifted higher up" so
it's not quite so deep and should be easier to remove in the future.

### DIFF
--- a/crates/wasmtime/src/runtime/store/gc.rs
+++ b/crates/wasmtime/src/runtime/store/gc.rs
@@ -45,7 +45,7 @@ impl StoreOpaque {
         root: Option<VMGcRef>,
         bytes_needed: Option<u64>,
     ) -> Option<VMGcRef> {
-        // SAFETY: this isn't safe, see #11409
+        // SAFETY: this isn't sound, see #11409
         let (mut limiter, store) =
             unsafe { self.traitobj().as_mut().resource_limiter_and_store_opaque() };
         store.gc(limiter.as_mut(), root, bytes_needed).await
--- a/crates/wasmtime/src/runtime/store/gc.rs
+++ b/crates/wasmtime/src/runtime/store/gc.rs
@@ -224,11 +224,14 @@ impl StoreOpaque {
                     // SAFETY: FIXME(#11409)
                     unsafe {
                         if self.async_support() {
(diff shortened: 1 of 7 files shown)


alexcrichton:
Note that this is currently built on https://github.com/bytecodealliance/wasmtime/pull/11460 which is in turn built on https://github.com/bytecodealliance/wasmtime/pull/11459

crates/wasmtime/src/runtime/store/gc.rs
```diff
@@ -1,78 +1,77 @@
...
+    /// FIXME(#11409) this method should not need to exist, but performing such
+    /// a refactoring will require making memory creation async.
+    async unsafe fn gc_unsafe_get_limiter(
+        &mut self,
+        root: Option<VMGcRef>,
+        bytes_needed: Option<u64>,
+    ) -> Option<VMGcRef> {
+        // SAFETY: this isn't safe, see #11409
+        let (mut limiter, store) =
+            unsafe { self.traitobj().as_mut().resource_limiter_and_store_opaque() };
```
fitzgen: s/safe/sound/?

fitzgen (approved):
Nice

PR #11463, opened 2025-08-19
Change how wasm features/gc support work in `Config`

### LINKED ISSUE #11450: `wasmtime run GC=no` doesn't seem to disable GC in the host
Thanks for filing a bug report! Please fill out the TODOs below.

**Note: if you want to report a security issue, please read our [security policy](https://bytecodealliance.org/security)!**

### Steps to Reproduce

Write a library that compiles your component with the following features from wasmtime:

```
wasmtime = { version="34", default-features = false, features = ["cranelift", "runtime", "component-model" ] }
```

```
    let mut config = Config::new();
  let engine = Engine::new(&config).unwrap();
  let bytes = std::fs::read(infile).unwrap();
  let serialized = if args.get_flag("component") {
      engine.precompile_component(&bytes).unwrap()
  } else {
      engine.precompile_module(&bytes).unwrap()
  };
```

Then try and run it in in wasmtime, turning off GC in the host, it fails to run:

```
wasmtime run -W gc=no -W component-model-gc=no --target x86_64-unknown-none --allow-precompiled --invoke 'echo("hi")' /home/jstur/projects/hyperlight-wasm/x64/debug/runcomponent.aot
```

```
Caused by:
    module was compiled without GC but GC is enabled in the host
```

### Expected Results

I would expect that I can turn the features off and run them.

### Actual Results

error `module was compiled without GC but GC is enabled in the host`

### Versions and Environment

Wasmtime version or commit: any released version

Operating system: TODO

Architecture: TODO

### Extra Info

I tracked it down to this being always on via build flag even though `-W gc=no -W component-model-gc=no` settings are turned off and causing a failure later on in processing.

https://github.com/bytecodealliance/wasmtime/blob/759ccdfaee65f8c470ccc5b8b63506a1af5cd32d/crates/wasmtime/src/config.rs#L2162C1-L2163C1

I moved this logic to use the gc flag and things worked.  Then I found https://github.com/bytecodealliance/wasmtime/pull/10455#discussion_r2008287118 and wasn't sure what the correct solution is here.

alexcrichton:
This commit aims to address #11450 in complementary but somewhat orthogonal ways. First a new `Config::gc_support` option is added which is hooked up to the CLI as `-Wgc-support`. This option controls the wasmparser-internal `GC_TYPES` feature. Its default value, like before, is `cfg!(feature = "gc")` and additionally enabling it requires the `gc` crate feature to be enabled.

This commit then additionally updates how wasm features are processed during validating a deserialized module to only require enabled features to be enabled in the host. Previously modules that disabled a feature but the feature was enabled in the host would fail to deserialize. All WebAssembly proposals are additive, however, so it's always ok to disable a feature and then load it into a module that enables the wasm proposal.

Closes #11450

### DIFF
--- a/crates/cli-flags/src/lib.rs
+++ b/crates/cli-flags/src/lib.rs
@@ -403,6 +403,8 @@ wasmtime_option_group! {
         pub exceptions: Option<bool>,
         /// DEPRECATED: Configure support for the legacy exceptions proposal.
         pub legacy_exceptions: Option<bool>,
+        /// Whether or not any GC infrastructure in Wasmtime is enabled or not.
+        pub gc_support: Option<bool>,
     }
 
     enum Wasm {
@@ -988,6 +990,10 @@ impl CommonOptions {
             ["cranelift" : self.wasm.wmemcheck]
             enable => config.wmemcheck(enable),
             true => err,
+        }
+
+        if let Some(enable) = self.wasm.gc_support {
+            config.gc_support(enable);
         }
 
         Ok(config)
--- a/crates/wasmtime/src/config.rs
+++ b/crates/wasmtime/src/config.rs
@@ -2226,9 +2226,12 @@ impl Config {
         if self.max_wasm_stack == 0 {
             bail!("max_wasm_stack size cannot be zero");
         }
-        #[cfg(not(feature = "wmemcheck"))]
-        if self.wmemcheck {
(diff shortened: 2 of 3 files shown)


fitzgen (approved):

jsturtevant:
Thanks!

PR #11467, opened 2025-08-19
Refactor exception objects to share more implementation with structs.

cfallin:
This PR updates the implementation of Wasm exception objects to share more layout-computation code with structs, except for a single fixed tag slot. The former is simply a nice refactor, and the latter is in preparation for full exception support, which will require accessing an exception's tag without knowing its layout dynamically at runtime.

### DIFF
--- a/crates/environ/src/compile/module_types.rs
+++ b/crates/environ/src/compile/module_types.rs
@@ -4,7 +4,11 @@ use crate::{
     WasmExnType, WasmFieldType, WasmFuncType, WasmHeapType, WasmResult, WasmStorageType,
     WasmStructType, WasmSubType, wasm_unsupported,
 };
-use std::{borrow::Cow, collections::HashMap, ops::Index};
+use std::{
+    borrow::Cow,
+    collections::{HashMap, hash_map::Entry},
+    ops::Index,
+};
 use wasmparser::{UnpackedIndex, Validator, ValidatorId};
 
 /// A type marking the start of a recursion group's definition.
@@ -38,6 +42,12 @@ pub struct ModuleTypesBuilder {
     /// trampoline type for a given function type if we've already interned one.
     trampoline_types: HashMap<WasmFuncType, ModuleInternedTypeIndex>,
 
+    /// An interning map for exception types corresponding to function
+    /// types used by tags. Tags are nominal, but the underlying
+    /// Wasmtime types describe only the object layout and so are
+    /// structural.
+    exception_types: HashMap<ModuleInternedTypeIndex, ModuleInternedTypeIndex>,
+
     /// A map from already-interned `wasmparser` types to their corresponding
     /// Wasmtime type.
     wasmparser_to_wasmtime: HashMap<wasmparser::types::CoreTypeId, ModuleInternedTypeIndex>,
@@ -57,6 +67,7 @@ impl ModuleTypesBuilder {
             validator_id: validator.id(),
             types: ModuleTypes::default(),
             trampoline_types: HashMap::default(),
+            exception_types: HashMap::default(),
             wasmparser_to_wasmtime: HashMap::default(),
             already_seen: HashMap::default(),
             defining_rec_group: None,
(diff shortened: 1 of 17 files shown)


cfallin:
This has been pulled out of #11326 for separate landing.

crates/environ/src/compile/module_types.rs
```diff
@@ -301,6 +302,37 @@ impl ModuleTypesBuilder {
...
+                    func_ty: EngineOrModuleTypeIndex::Module(for_func_ty),
+                    fields,
+                }),
+                shared: false,
+            },
+        });
+        let next = self.types.next_ty();
+        self.types.push_rec_group(idx..next);
+        idx
+    }
```
fitzgen: This is not doing any hash consing because tags are nominal, correct? Can we add that note in a comment somewhere?

And also something about how this is only for our implementation-specific `ModuleInternedTypeIndex` and does not affect the the Wasm-level types index space. That is, these aren't types at the wasm-level, but we treat them as types for convenience in our implementation and this is in service of that.

crates/environ/src/compile/module_types.rs
```diff
@@ -372,6 +404,50 @@ impl ModuleTypesBuilder {
...
+        if composite_type.shared {
+            return Err(wasm_unsupported!(
+                "shared functions are not yet implemented"
+            ));
+        }
+        match &composite_type.inner {
+            WasmCompositeInnerType::Func(f) => Ok(f),
+            _ => unreachable!(),
+        }
+    }
```
fitzgen: FWIW, if you have appetite for a little clean up, I think these methods and the existing ones above could use the `unwrap_*` methods on `WasmCompositeType` to dedupe a little bit of code and then become something like

```rust
        let comp_ty = &self.types[interned_ty].composite_type;
        if comp_ty.shared {
            return Err(wasm_unsupported!(
                "shared functions are not yet implemented"
            ));
        }
        Ok(comp_ty.unwrap_func())
```

Although I must also say (and maybe this is my fault) I am also confused by the combination of error-returning sometimes and panicking other times. Seems like we could move everything to one or the other, depending on usage.

Feel free to deal with both or one or none of these things in this PR, as you see fit.

crates/environ/src/gc.rs
```diff
@@ -175,30 +175,30 @@ fn common_struct_layout(
...
+        is_exception: false,
     }
 }
 
 /// Common code to define a GC exception object's layout, given the
 /// size and alignment of the collector's GC header and its expected
 /// offset of the array length field.
 #[cfg(any(feature = "gc-null", feature = "gc-drc"))]
-fn common_exn_layout(ty: &WasmExnType, header_size: u32, header_align: u32) -> GcExceptionLayout {
+fn common_exn_layout(ty: &WasmExnType, header_size: u32, header_align: u32) -> GcStructLayout {
```
fitzgen, line 186: Does it make sense to still have `GcExceptionLayout` but have it just be a newtype wrapper over `GcStructLayout`? That way we can still use helpers that want to take a `GcStructLayout` argument but we also have some tidiness with a dedicated type and can remove the `is_exception: bool` field from `GcStructLayout`.

Not sure this is actually a useful suggestion yet, going to look at a little bit more of the PR...

crates/wasmtime/src/runtime/vm/gc/enabled/null.rs
```diff
@@ -286,19 +286,20 @@ unsafe impl GcHeap for NullHeap {
...
-            VMGcHeader::from_kind_and_index(VMGcKind::StructRef, ty),
-            layout.layout(),
-        )
-        .map(|r| r.map(|r| r.into_structref_unchecked()))
+    ) -> Result<Result<VMGcRef, u64>> {
+        let kind = if layout.is_exception {
+            VMGcKind::ExnRef
+        } else {
+            VMGcKind::StructRef
+        };
```
fitzgen, line 298: Okay I guess it doesn't really make sense to remove the `is_exception: bool` field and do that newtype thing. Or else we would have to thread an `is_exception: bool` argument through here which seems like 6 of one and half a dozen of another. Nevermind!

fitzgen (approved):
LGTM, thanks for splitting things up and making that PR a little smaller!

crates/wasmtime/src/runtime/vm/gc/enabled/null.rs
cfallin: Yep, this seemed like the most reasonable way to do it to me...

crates/environ/src/gc.rs
cfallin: (resolving as per below)

crates/environ/src/compile/module_types.rs
cfallin: Did the simple cleanup as suggested at least -- good call.

crates/environ/src/compile/module_types.rs
cfallin: Ah, great point, actually we can hash-cons -- tags are nominal, but that is covered by the dynamic tag-instance reference in an exception object; the Wasmtime exception-object type describes only the layout, which can be structural / shared across all exceptions with the same layout. I've added hash-consing here, as well as comments about this (and about the Wasmtime-specific nature of the exception types).

PR #11468, opened 2025-08-19
Make const-expr evaluation `async`

alexcrichton:
This commit is extracted from #11430 to accurately reflect how const-expr evaluation is an async operation due to GC pauses that may happen. The changes in this commit are:

* Const-expr evaluation is, at its core, now an `async` function.
* To leverage this new `async`-ness all internal operations are switched from `*_maybe_async` to `*_async` meaning all the `*_maybe_async` methods can be removed.
* Some libcalls using `*_maybe_async` are switch to using `*_async` plus the `block_on!` utility to help jettison more `*_maybe_async` methods.
* Instance initialization is now an `async` function. This is temporarily handled with `block_on` during instance initialization to avoid propagating the `async`-ness further upwards. This `block_on` will get deleted in future refactorings.
* Const-expr evaluation has been refactored slightly to enable having a fast path in global initialization which skips an `await` point entirely, achieving performance-parity in benchmarks prior to this commit.

This ended up fixing a niche issue with GC where if a wasm execution was suspended during `table.init`, for example, during a const-expr evaluation triggering a GC then if the wasm execution was cancelled it would panic the host. This panic was because the GC operation returned `Result` but it was `unwrap`'d as part of the const-expr evaluation which can fail not only to invalid-ness but also due to "computation is cancelled" traps.

### DIFF
--- a/crates/wasmtime/src/runtime/instance.rs
+++ b/crates/wasmtime/src/runtime/instance.rs
@@ -377,7 +377,21 @@ impl Instance {
             .features()
             .contains(WasmFeatures::BULK_MEMORY);
 
-        vm::initialize_instance(store, id, compiled_module.module(), bulk_memory)?;
+        if store.async_support() {
+            store.block_on(|store| {
+                let module = compiled_module.module().clone();
+                Box::pin(
+                    async move { vm::initialize_instance(store, id, &module, bulk_memory).await },
+                )
+            })??;
+        } else {
+            vm::assert_ready(vm::initialize_instance(
+                store,
+                id,
+                compiled_module.module(),
+                bulk_memory,
+            ))?;
+        }
 
         Ok((instance, compiled_module.module().start_func))
     }
--- a/crates/wasmtime/src/runtime/instance.rs
+++ b/crates/wasmtime/src/runtime/instance.rs
@@ -378,12 +378,15 @@ impl Instance {
             .contains(WasmFeatures::BULK_MEMORY);
 
         if store.async_support() {
+            #[cfg(feature = "async")]
             store.block_on(|store| {
                 let module = compiled_module.module().clone();
                 Box::pin(
                     async move { vm::initialize_instance(store, id, &module, bulk_memory).await },
                 )
             })??;
+            #[cfg(not(feature = "async"))]
+            unreachable!();
         } else {
             vm::assert_ready(vm::initialize_instance(
                 store,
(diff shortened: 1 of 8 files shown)


alexcrichton:
Actually, thinking more, the issue about panicking on const expr evaluation failure means [allocation failure in GC is a host panic](https://github.com/bytecodealliance/wasmtime/issues/11469) which isn't great. Deferring that to a separate issue.

crates/wasmtime/src/runtime/vm/libcalls.rs
```diff
@@ -674,18 +673,19 @@ unsafe fn gc_alloc_raw(
...
-        store.retry_after_gc_maybe_async((), |store, ()| {
-            store
-                .unwrap_gc_store_mut()
-                .alloc_raw(header, layout)?
-                .map_err(|bytes_needed| crate::GcHeapOutOfMemory::new((), bytes_needed).into())
-        })?
-    };
+    block_on!(store, async |store| {
+        let gc_ref = store
+            .retry_after_gc_async((), |store, ()| {
```
fitzgen, line 678: Are there any non-async `retry_after_gc` calls at this point? I think all of them should be async by the time we finish the asyncification of the internals, but I'm not sure if we are there yet or not. But when we are, we should delete the non-async version and then just rename the async variation to plain `retry_after_gc`.

crates/wasmtime/src/runtime/instance.rs
```diff
@@ -377,7 +377,24 @@ impl Instance {
...
+            #[cfg(not(feature = "async"))]
+            unreachable!();
+        } else {
+            vm::assert_ready(vm::initialize_instance(
+                store,
+                id,
+                compiled_module.module(),
+                bulk_memory,
+            ))?;
+        }
```
fitzgen, line 397: And then `Instance::new_raw` will become an `async` function in another follow up PR?

crates/wasmtime/src/runtime/vm/libcalls.rs
alexcrichton: Soon! (got a PR queued up after #11470 to delete the sync version

crates/wasmtime/src/runtime/instance.rs
alexcrichton: Indeed! That'll be #11470

PR #11470, opened 2025-08-19
 Make table/memory creation async functions 

alexcrichton:
This commit is a large-ish refactor which is made possible by the many
previous refactorings to internals w.r.t. async-in-Wasmtime. The end
goal of this change is that table and memory allocation are both `async`
functions. Achieving this, however, required some refactoring to enable
it to work:

* To work with `Send` neither function can close over `dyn VMStore`.
  This required changing their `Option<&mut dyn VMStore>` arugment to
  `Option<&mut StoreResourceLimiter<'_>>`
* Somehow a `StoreResourceLimiter` needed to be acquired from an
  `InstanceAllocationRequest`. Previously the store was stored here as
  an unsafe raw pointer, but I've refactored this now so
  `InstanceAllocationRequest` directly stores `&StoreOpaque` and
  `Option<&mut StoreResourceLimiter>` meaning it's trivial to acquire
  them. This additionally means no more `unsafe` access of the store
  during instance allocation (yay!).
* Now-redundant fields of `InstanceAllocationRequest` were removed since
  they can be safely inferred from `&StoreOpaque`. For example passing
  around `&Tunables` is now all gone.
* Methods upwards from table/memory allocation to the
  `InstanceAllocator` trait needed to be made `async`. This includes new
  `#[async_trait]` methods for example.
* `StoreOpaque::ensure_gc_store` is now an `async` function. This
  internally carries a new `unsafe` block carried over from before with
  the raw point passed around in `InstanceAllocationRequest`. A future
  PR will delete this `unsafe` block, it's just temporary.

I attempted a few times to split this PR up into separate commits but
everything is relatively intertwined here so this is the smallest
"atomic" unit I could manage to land these changes and refactorings.

cc #11430

### DIFF
--- a/crates/wasmtime/src/runtime/component/instance.rs
+++ b/crates/wasmtime/src/runtime/component/instance.rs
@@ -634,7 +634,7 @@ impl<'a> Instantiator<'a> {
         }
     }
 
-    fn run<T>(&mut self, store: &mut StoreContextMut<'_, T>) -> Result<()> {
+    async fn run<T>(&mut self, store: &mut StoreContextMut<'_, T>) -> Result<()> {
         let env_component = self.component.env_component();
 
         // Before all initializers are processed configure all destructors for
@@ -714,7 +714,7 @@ impl<'a> Instantiator<'a> {
                     // if required.
 
                     let i = unsafe {
-                        crate::Instance::new_started_impl(store, module, imports.as_ref())?
+                        crate::Instance::new_started(store, module, imports.as_ref()).await?
                     };
                     self.instance_mut(store.0).push_instance_id(i.id());
                 }
@@ -991,7 +991,7 @@ impl<T: 'static> InstancePre<T> {
             !store.as_context().async_support(),
             "must use async instantiation when async support is enabled"
         );
-        self.instantiate_impl(store)
+        vm::assert_ready(self._instantiate(store))
     }
     /// Performs the instantiation process into the store specified.
     ///
@@ -999,29 +999,18 @@ impl<T: 'static> InstancePre<T> {
     //
     // TODO: needs more docs
     #[cfg(feature = "async")]
-    pub async fn instantiate_async(
-        &self,
-        mut store: impl AsContextMut<Data = T>,
-    ) -> Result<Instance>
-    where
-        T: Send,
-    {
-        let mut store = store.as_context_mut();
-        assert!(
-            store.0.async_support(),
-            "must use sync instantiation when async support is disabled"
-        );
-        store.on_fiber(|store| self.instantiate_impl(store)).await?
+    pub async fn instantiate_async(&self, store: impl AsContextMut<Data = T>) -> Result<Instance> {
+        self._instantiate(store).await
     }
 
-    fn instantiate_impl(&self, mut store: impl AsContextMut<Data = T>) -> Result<Instance> {
+    async fn _instantiate(&self, mut store: impl AsContextMut<Data = T>) -> Result<Instance> {
         let mut store = store.as_context_mut();
         store
             .engine()
             .allocator()
             .increment_component_instance_count()?;
         let mut instantiator = Instantiator::new(&self.component, store.0, &self.imports);
-        instantiator.run(&mut store).map_err(|e| {
+        instantiator.run(&mut store).await.map_err(|e| {
             store
                 .engine()
                 .allocator()
--- a/crates/wasmtime/src/runtime/externals/table.rs
+++ b/crates/wasmtime/src/runtime/externals/table.rs
@@ -94,7 +94,8 @@ impl Table {
     /// # }
     /// ```
     pub fn new(mut store: impl AsContextMut, ty: TableType, init: Ref) -> Result<Table> {
-        Table::_new(store.as_context_mut().0, ty, init)
+        vm::one_poll(Table::_new(store.as_context_mut().0, ty, init))
+            .expect("must use `new_async` when async resource limiters are in use")
     }
 
     /// Async variant of [`Table::new`]. You must use this variant with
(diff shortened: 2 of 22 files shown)


alexcrichton:
I'll note that I've split this into two commits, the first of which is https://github.com/bytecodealliance/wasmtime/pull/11456 resurrected here. That commit cannot be split out to a second PR due to all the various constraints in play unfortunately, so this is a bit larger than I would have otherwise anticipated.

alexcrichton:
Oh, I should also note, this is a 5% performance penalty overhead to instance instantiation. Benchmarking makes me think it's primarily related to the `Box`-cost of `#[async_trait]`. We can burn down that cost I think if we really need from allocation-per-table-and-memory to once-per-module (aka two for most modules to one), but I don't know how to get away from it entirely. I subjectively concluded a few extra allocations is fine, but others can reasonably differ.

alexcrichton:
I was a bit afraid of this, but tests here will fail until I rebase this on top of #11468. So ready for review, but won't pass tests until #11468 lands first.

crates/wasmtime/src/runtime/externals/table.rs
```diff
@@ -94,7 +94,8 @@ impl Table {
     /// # }
     /// ```
     pub fn new(mut store: impl AsContextMut, ty: TableType, init: Ref) -> Result<Table> {
-        Table::_new(store.as_context_mut().0, ty, init)
+        vm::one_poll(Table::_new(store.as_context_mut().0, ty, init))
+            .expect("must use `new_async` when async resource limiters are in use")
```
fitzgen, line 98: Should we not assert that this is a non-async config before the `.expect(..)` to catch mismatches a little earlier?

crates/wasmtime/src/runtime/component/instance.rs
```diff
@@ -991,37 +991,26 @@ impl<T: 'static> InstancePre<T> {
...
-        T: Send,
-    {
-        let mut store = store.as_context_mut();
-        assert!(
-            store.0.async_support(),
-            "must use sync instantiation when async support is disabled"
-        );
-        store.on_fiber(|store| self.instantiate_impl(store)).await?
+    pub async fn instantiate_async(&self, store: impl AsContextMut<Data = T>) -> Result<Instance> {
+        self._instantiate(store).await
```
fitzgen, line 1003: And similarly assert that the config is async here?

crates/wasmtime/src/runtime/vm/instance/allocator/pooling.rs
```diff
@@ -644,14 +645,29 @@ unsafe impl InstanceAllocator for PoolingInstanceAllocator {
...
+        request: &mut InstanceAllocationRequest<'_>,
         ty: &wasmtime_environ::Memory,
-        tunables: &Tunables,
         memory_index: Option<DefinedMemoryIndex>,
     ) -> Result<(MemoryAllocationIndex, Memory)> {
-        self.with_flush_and_retry(|| self.memories.allocate(request, ty, tunables, memory_index))
+        // FIXME(rust-lang/rust#145127) this should ideally use a version of
+        // `with_flush_and_retry` but adapted for async closures instead of only
+        // sync closures. Right now that won't compile though so this is the
+        // manually expanded version of the method.
```
fitzgen, line 658: Add a note for this to our items to discuss with the lang team, if we don't have one already?

crates/wasmtime/src/runtime/vm/instance/allocator.rs
```diff
@@ -52,79 +50,11 @@ pub struct InstanceAllocationRequest<'a> {
...
-unsafe impl Send for StorePtr {}
-unsafe impl Sync for StorePtr {}
-
-impl StorePtr {
-    /// A pointer to no Store.
-    pub fn empty() -> Self {
-        Self(None)
-    }
+    /// The store that this instance is being allocated into.
+    pub store: &'a StoreOpaque,
```
fitzgen, line 54: This is so nice to clean up

fitzgen (approved):

crates/wasmtime/src/runtime/vm/instance/allocator/pooling.rs
alexcrichton: Oh this one's technically [already fixed](https://github.com/rust-lang/rust/pull/143545) and just waiting on stabilization, and browsing https://github.com/rust-lang/rust/issues/110338 shows this is pretty well-known, so not a major issue.

crates/wasmtime/src/runtime/externals/table.rs
alexcrichton: Oh this is actually replicating the documented behavior where this works with `async_support` so long as an async resource limiter isn't used. In that sense to avoid breaking the semantics here `one_poll` is what's required as opposed to an `assert!(!async_support)` plus `vm::assert_ready`.

crates/wasmtime/src/runtime/component/instance.rs
alexcrichton: As I've been making these changes I've actually been undoing a lot of `assert!(async_support)`-style assertions. Previously that was required because `on_fiber` was immediately used which required `async_support` to be turned on, but now it's just normal Rust async functions so there's no reason to prevent usage when `async_support` is disabled. In that sense it's intentional that the assert here is lost, but how's that sound to you?

alexcrichton:
I'm going to go ahead and land this, but @fitzgen if you have follow up comments I'm happy to address them.

fitzgen:
> I'm going to go ahead and land this, but @fitzgen if you have follow up comments I'm happy to address them.

Nope, looks good

PR #11473, opened 2025-08-20
Add public accessors of sub-contexts of `WasiCtx`

alexcrichton:
This is intended to assist with embedders who aren't necessarily using `WasiView` for everything.


alexcrichton:
I'll note that this is required to update to Wasmtime 36 in Spin, so once this merges I'll be doing a backport + point release.

PR #11480, opened 2025-08-20
 Deduplicate some more GC functions 

alexcrichton:
Use the `async` version from sync API entrypoints to deduplicate the
implementation.

PR #11481, opened 2025-08-20
 Remove unsoundness of widening store borrows 

alexcrichton:
This commit removes preexisting unsoundness in Wasmtime where a
`&mut StoreOpaque` borrow was "widened" into encompassing the limiter on
the `T` in `StoreInner<T>`, for example, by using the self-pointer
located in an instance or the store. This fix is done by threading
`&mut StoreOpaque` as a parameter separately from a
`StoreResourceLimiter`. This means that various callers now take a new
`Option<&mut StoreResourceLimiter<'_>>` parameter in various locations.

Closes https://github.com/bytecodealliance/wasmtime/issues/11409

### DIFF
--- a/crates/wasmtime/src/runtime/vm/const_expr.rs
+++ b/crates/wasmtime/src/runtime/vm/const_expr.rs
@@ -222,6 +222,10 @@ impl ConstExprEvaluator {
 
         self.stack.clear();
 
+        // On GC-less builds ensure that this is always considered used an
+        // needed-mutable.
+        let _ = &mut limiter;
+
         for op in expr.ops() {
             log::trace!("const-evaluating op: {op:?}");
             match op {


alexcrichton:
Currently built on https://github.com/bytecodealliance/wasmtime/pull/11480, which is in turn built on https://github.com/bytecodealliance/wasmtime/pull/11470, so the first three commits aren't part of this PR

crates/wasmtime/src/runtime/component/instance.rs
```diff
@@ -991,37 +991,26 @@ impl<T: 'static> InstancePre<T> {
...
-        T: Send,
-    {
-        let mut store = store.as_context_mut();
-        assert!(
-            store.0.async_support(),
-            "must use sync instantiation when async support is disabled"
-        );
-        store.on_fiber(|store| self.instantiate_impl(store)).await?
+    pub async fn instantiate_async(&self, store: impl AsContextMut<Data = T>) -> Result<Instance> {
+        self._instantiate(store).await
```
dicej: Why do we not need an `assert!(store.0.async_support())` here anymore?

alexcrichton: A good question! That's part of the previous commits and Nick was also curious [here](https://github.com/bytecodealliance/wasmtime/pull/11470#discussion_r2288971332), with some rationale in my response over there

fitzgen (approved):
Nice!

PR #11475, opened 2025-08-20
Test out building the C API artifacts with LTO

alexcrichton:
Trying to shrink the size of compiled artifacts...

cc https://github.com/bytecodealliance/wasmtime-go/issues/250

PR #11483, opened 2025-08-20
Shrink C API release artifacts

### LINKED ISSUE #11476: C API Artifacts getting too large for redistribution
The current motivation for this issue is that the wasmtime-go repository [cannot make a 36.0.0 release](https://github.com/bytecodealliance/wasmtime-go/issues/250) because the C API artifacts have gotten too large. A rough overview of what's happening there is that the only way I know of to distribute the C API in Go is to literally check the binaries themselves (for all architectures no less) into the repository at the published tags. GitHub has a 100M file size limit for files in git repositories Wasmtime has been steadily growing large and larger over time (as GitHub prints warnings when I push the tag on the CLI) and the x86_64 linux C API is now above this threshold.

On one hand this is expected, we slowly add things over time and binaries are expected to get larger. This is why we have conditional compile-time features to reduce the size of artifacts. In the wasmtime-go repository, though, it's trying to reuse the exact artifacts that Wasmtime publishes as I'd really rather avoid duplicating the build process just for a slightly reduced feature set. Thus, this issue.

If we accept this as an issue that should be addressed, then the question remains of what to do about it. We've got a [well-documented series of steps](https://docs.wasmtime.dev/examples-minimal.html) of how to build a minimal embedding of Wasmtime which gives a whole bunch of knobs of how to shrink the size of the compiled artifact, and it's a question of how best to apply them.

## Why is `libwasmtime.a` big?

The current state of the world is that wasmtime-go checks in `libwasmtime.a` static archives. It avoids `*.so` or dynamic library artifacts to avoid adding new dependencies to executables produced (I don't even know how we'd communicate to users how to copy around the library). The [`libwasmtime.a` artifact for x86_64-linux](https://github.com/bytecodealliance/wasmtime/releases/download/v36.0.0/wasmtime-v36.0.0-x86_64-linux-c-api.tar.xz) is 104M. The `libwasmtime.so` artifact, however, is only 26M. I would consider the `libwasmtime.so` artifact the "lower bound" of what we would ideally be able to achieve.

So first off: why is `libwasmtime.so` so much smaller? The main reason for this is that `*.a` is just a huge collection of `*.o` files internally. These object files are not all 100% used in the final artifact (e.g. `--gc-sections` would remove a lot) and they're just raw copied from rlibs. The `*.so` is much smaller as it's a fully linked artifact that has all dead code removed. There's also "linking metadata" such as relocation sections, big symbols tables, llvm bitcode, etc, which is all present in `*.a` but not in `*.so`.

## Shrinking `libwasmtime.a`

Ok that's enough intro, what to do about this

* Turn on LTO. This cuts the `*.a` size in half because it's effectively DCE at compile time. The optimization side of things ends up making it really slow though. Testing in https://github.com/bytecodealliance/wasmtime/pull/11475 the Windows release builds clock in at 20m where they were previously  15, which would make them the slowest builder. (not a problem for Linux/macOS builds though...)
* Turn on ThinLTO. Similar to above but faster as it's more parallelizable. Only shaves ~10-15M locally though so not really enough to buy much runway.
* Turn off some features. This is here for completeness but I don't think this is really viable since the point of these artifacts is to be the "default featureful artifacts".
* Turn on panic=abort. Alas this is already done.
* Turn on `opt-level=s` or `z`. Measured locally to basically have no effect (this is good for micro-optimizing, but bad for macro-optimizing in my experience)
* Use `CARGO_PROFILE_RELEASE_CODEGEN_UNITS=1`. Shaves 20-30M off locally. (possible!)
* Some how use the `-r` flag to `ld`. This is in theory "apply `--gc-sections` ahead of time" AFAIK, but I've never used this option successfully nor have I seen anyone else use it successfully. Trying naively locally requires knowledge of symbols which we do not have.

Well I at least wanted to open an issue about this. I don't think much will come of opening the issue here, but maybe someone can swoop in with That One Easy Trick which makes the binary tiny.

alexcrichton:
This is needed to currently shrink our `*.a` artifacts under 100M in size which is required downstream in the wasmtime-go embedding. Some testing in #11475 of using full LTO or only building Linux/macOS with 1 CGU shows that this'll likely increase cycle time by ~1 minute. Windows builders are already the slowest release build builders and this is adding 1 minute using 1 CGU vs 2-3 extra minutes for full LTO. Linux/macOS are switched to using full LTO which the CI builders currently have plenty of time for.

This shaves 50M off the C API artifacts for Linux which buys a good amount of headroom for future changes.

Closes #11476

### DIFF
--- a/ci/build-release-artifacts.sh
+++ b/ci/build-release-artifacts.sh
@@ -77,6 +77,21 @@ cargo build --release $flags --target $target -p wasmtime-cli $bin_flags --featu
 # a different project that wants to unwind.
 export RUSTFLAGS="$RUSTFLAGS -C force-unwind-tables"
 
+# Shrink the size of `*.a` artifacts without spending too much extra time in CI.
+# See #11476 for some more context. Here Windows builds achieve this with 1 CGU
+# which results in modest size gains, and other platforms use LTO to achieve
+# much more significant size gains. The reason the platforms are different is
+# that CI is extremely slow on Windows, almost 2x slower, so this is an attempt
+# to keep CI cycle time under control.
+case $build in
+  *-mingw* | *-windows*)
+    export CARGO_PROFILE_RELEASE_CODEGEN_UNITS=1
+    ;;
+  *)
+    export CARGO_PROFILE_RELEASE_LTO=true
+    ;;
+esac
+
 mkdir -p target/c-api-build
 cd target/c-api-build
 cmake \


alexcrichton:
I'll also note that assuming everyone else is ok with this I plan on including this in a 36.0.1 point release tomorrow.

PR #11487, opened 2025-08-21
Add initial empty rec groups

khagankhan:
This PR introduces initial support for empty `(rec ...)` groups:

- Adds generation of empty recursive groups within a configurable range (clamping is applied).
- Establishes foundational support for future GC work (e.g., struct definitions, related implementations).
- Updates existing tests and adds new unit tests.
- Introduce a new `Op` that attempts to instantiate a struct from a group. It is designed to be self-dropping due to type incompatibility.

---

Note:

The line:

    StructNew(k: |ops| ops.total_struct_types() => u32) : 0 => 0,

should effectively prevent `struct.new` from being generated when no struct types exist. Depending on the configuration and macro logic, it will either:

- Not be generated at all,
- Trigger an assertion failure during macro expansion,
- Or work as expected by skipping generation.

I will confirm and refine this behavior in a follow-up PR.

### DIFF
--- a/crates/fuzzing/src/generators/table_ops.rs
+++ b/crates/fuzzing/src/generators/table_ops.rs
@@ -5,9 +5,9 @@ use serde::{Deserialize, Serialize};
 use smallvec::SmallVec;
 use std::ops::RangeInclusive;
 use wasm_encoder::{
-    CodeSection, CompositeInnerType, CompositeType, ConstExpr, EntityType, ExportKind,
-    ExportSection, Function, FunctionSection, GlobalSection, ImportSection, Instruction, Module,
-    RefType, StructType, SubType, TableSection, TableType, TypeSection, ValType,
+    CodeSection, ConstExpr, EntityType, ExportKind, ExportSection, Function, FunctionSection,
+    GlobalSection, ImportSection, Instruction, Module, RefType, TableSection, TableType,
+    TypeSection, ValType,
 };
 
 /// RecGroup ID struct definition.
@@ -16,26 +16,21 @@ pub struct RecGroupId(u32);
 
 /// Struct types definition.
 #[derive(Debug, Default, Serialize, Deserialize)]
-pub struct StructTypes {
-    next_group: u32,
+pub struct Types {
     rec_groups: std::collections::BTreeSet<RecGroupId>,
 }
 
-impl StructTypes {
-    /// Create a fresh `StructTypes` allocator with no recursive groups defined yet.
+impl Types {
+    /// Create a fresh `Types` allocator with no recursive groups defined yet.
     pub fn new() -> Self {
         Self {
-            next_group: 0,
             rec_groups: Default::default(),
         }
     }
 
-    /// Allocate a new empty recursive group `(rec ...)` and return its unique `RecGroupId`.
-    pub fn alloc_empty_rec_group(&mut self) -> RecGroupId {
-        let id = RecGroupId(self.next_group);
-        self.next_group += 1;
-        self.rec_groups.insert(id.clone());
-        id
+    /// Insert a rec-group id. Returns true if newly inserted, false if it already existed.
+    pub fn insert_rec_group(&mut self, id: RecGroupId) -> bool {
+        self.rec_groups.insert(id)
     }
 
     /// Iterate over all allocated recursive groups.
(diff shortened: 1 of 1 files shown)


crates/fuzzing/src/generators/table_ops.rs
```diff
@@ -5,17 +5,62 @@ use serde::{Deserialize, Serialize};
...
+    RefType, StructType, SubType, TableSection, TableType, TypeSection, ValType,
 };
 
+/// RecGroup ID struct definition.
+#[derive(Debug, Clone, Eq, PartialOrd, PartialEq, Ord, Hash, Default, Serialize, Deserialize)]
+pub struct RecGroupId(u32);
+
+/// Struct types definition.
+#[derive(Debug, Default, Serialize, Deserialize)]
+pub struct StructTypes {
```
fitzgen: Nitpick: even though we are starting with `struct` types we plan on defining all kinds of types here eventually, so lets future-proof this a little bit and name it simply `Types`.

crates/fuzzing/src/generators/table_ops.rs
```diff
@@ -5,17 +5,62 @@ use serde::{Deserialize, Serialize};
...
+    next_group: u32,
+    rec_groups: std::collections::BTreeSet<RecGroupId>,
```
fitzgen: Because this is going to be serialized and deserialized, and the fuzzer will occasionally mutate those raw bytes, that means that we will get arbitrary values for `next_group` and `rec_groups` and we can't rely on any relationship between the two fields (such as, `rec_groups` does not contain a group with `id == next_group`).

It is probably better to get rid of the `next_group` field and just insert random values, resolving conflicts at insertion time as necessary (or even just ignoring conflicts and letting that "merge" rec groups).

crates/fuzzing/src/generators/table_ops.rs
```diff
@@ -5,17 +5,62 @@ use serde::{Deserialize, Serialize};
...
+}
+
 /// Limits controlling the structure of a generated Wasm module.
 #[derive(Debug, Default, Serialize, Deserialize)]
 pub struct TableOpsLimits {
     pub(crate) num_params: u32,
     pub(crate) num_globals: u32,
     pub(crate) table_size: i32,
+    pub(crate) num_rec_groups: u32,
+    pub(crate) empty_structs_per_group: u32,
```
fitzgen: I think we should delay emitting empty structs in rec groups, and just emit empty rec groups in this PR. Because longer-term, we aren't going to want to special-case empty structs in rec groups like this.

Then, in the next PR, we can extend `Types` something like this:

```rust
struct Types {
    // ...previous stuff from above...

    // The actual type definitions.
    type_defs: BTreeMap<TypeId, SubType>,
}

struct SubType {
    // Eventually, this will have finality and an optional super
    // type; for now it will only have this stuff.

    // The rec group that this type should be defined within.
    rec_group: RecGroupId,

    // The actual type definition.
    composite_type: CompositeType,
}

enum CompositeType {
    Struct(StructType),

    // And eventually, this will have array types too.
}

struct StructType {
    // Empty for now, but eventually will have fields.
}
```

When doing fixups/emitting a Wasm binary, we will create a map from rec group id to set of type ids for the types that are defined within it, since we need that to encode the rec group and its members. And then further in the future, we will have to do an SCC-based thing[^0] to merge rec groups together if there are cycles, since references to types across rec groups need to be acyclic (that is, type references can only be cyclic within the same rec group).

[^0]: FWIW, we have an [existing implementation of Tarjan's SCC algorithm](https://github.com/bytecodealliance/wasmtime/blob/c7308165c6d92d3cb6334e1980800a024ce2d889/crates/wasmtime/src/compile/scc.rs#L31) in tree for our inliner. We could consider reusing that directly, or we could just copy and tweak some code. Will need to think more about this when we get to it.

Anyways, all that should give us a really good skeleton to start filling out everything else on top of.

crates/fuzzing/src/generators/table_ops.rs
```diff
@@ -439,6 +526,8 @@ define_table_ops! {
     LocalGet(local_index: |ops| ops.num_params => u32) : 0 => 1,
     LocalSet(local_index: |ops| ops.num_params => u32) : 1 => 0,
 
+    StructNew(k: |ops| ops.total_struct_types() => u32) : 0 => 0,
```
fitzgen: And then we can add this in that next PR as well.

crates/fuzzing/src/generators/table_ops.rs
```diff
@@ -498,23 +592,32 @@ mod tests {
...
+                num_rec_groups: 3,
+                empty_structs_per_group: 4,
             },
             ops: vec![],
+            struct_types: StructTypes::new(),
+        };
+        for _ in 0..t.limits.num_rec_groups {
+            t.struct_types.alloc_empty_rec_group();
         }
+        t
```
fitzgen: Kinda unexpected to have the "empty" ops contain a non-zero number of rec groups?

khagankhan: Oh... Copy-paste error

khagankhan:
Thanks @fitzgen ! I guess it is ready for another round of review

fitzgen (approved):
Looks great!

khagankhan:
Thank you! Time to make new ones!!!!

Issue #11449, opened 2025-08-18
Expose pooling allocator runtime metrics


lann:
> This is similar to #2640 but is about pull rather than push metrics.

The pooling allocator has [a bunch of configurable limits](https://docs.rs/wasmtime/35.0.0/wasmtime/struct.PoolingAllocationConfig.html). It can be difficult to understand the runtime behavior of a system to decide how to tune those limits, so it would be helpful to expose some runtime metrics to embedders.

For the API, we could crib from [tokio](https://docs.rs/tokio/1.47.1/tokio/runtime/struct.Runtime.html#method.metrics) with something like `Engine::metrics(&self) -> RuntimeMetrics` and `RuntimeMetrics::pooling_allocator_live_core_instances(&self) -> u64`.

I've been looking at the allocator code and I think I have a rough understanding of how to implement what I'm most interested in:

- [`live_core_instances` and `live_component_instances`](https://github.com/bytecodealliance/wasmtime/blob/1292237f809c20d7587bf6471e51fd5edf913ca9/crates/wasmtime/src/runtime/vm/instance/allocator/pooling.rs#L304-L305): these are already atomics so should be straightforward to report
- Used memory slots: I _think_ the used slot count can be derived from the [stripe allocator(s)](https://github.com/bytecodealliance/wasmtime/blob/1292237f809c20d7587bf6471e51fd5edf913ca9/crates/wasmtime/src/runtime/vm/instance/allocator/pooling/memory_pool.rs#L82). That approach would require taking a `Mutex` which seems OK in this context as long as metrics reporting frequency is limited. Alternatively the used slot count could be tracked separately in a new atomic.
- Used table slots: looks like the same story as `MemoryPoll`, potentially tracking via the [allocator](https://github.com/bytecodealliance/wasmtime/blob/1292237f809c20d7587bf6471e51fd5edf913ca9/crates/wasmtime/src/runtime/vm/instance/allocator/pooling/table_pool.rs#L20)

alexcrichton:
Sounds like a very reasonable design to me! The only thing I'd to a minor bikeshed on is something like `Engine::pooling_allocator_metrics(&self) -> Option<PoolingAllocatorMetrics>` to handle the defaults-vs-pooling distinction. A general `Engine::metrics` could focus on generic things like instance count perhaps which makes sense to track for all allocators instead of just pooling.

PR #11490, opened 2025-08-21
Add PoolingAllocatorMetrics

lann:
This exposes some basic runtime metrics derived from the internal state of a `PoolingInstanceAllocator`.

Two new atomics were added to PoolingInstanceAllocator: `live_memories` and `live_tables`. While these counts could be derived from existing state it would require acquiring mutexes on some inner state.

Fixes https://github.com/bytecodealliance/wasmtime/issues/11449

### DIFF
--- a/crates/wasmtime/src/runtime/vm/instance/allocator/pooling.rs
+++ b/crates/wasmtime/src/runtime/vm/instance/allocator/pooling.rs
@@ -21,6 +21,7 @@
 mod decommit_queue;
 mod index_allocator;
 mod memory_pool;
+mod metrics;
 mod table_pool;
 
 #[cfg(feature = "gc")]
@@ -54,6 +55,7 @@ use crate::runtime::vm::{
     mpk::{self, ProtectionKey, ProtectionMask},
     sys::vm::PageMap,
 };
+use core::sync::atomic::AtomicUsize;
 use std::borrow::Cow;
 use std::fmt::Display;
 use std::sync::{Mutex, MutexGuard};
@@ -64,6 +66,8 @@ use std::{
 use wasmtime_environ::{
     DefinedMemoryIndex, DefinedTableIndex, HostPtr, Module, Tunables, VMOffsets,
 };
+
+pub use self::metrics::PoolingAllocatorMetrics;
 
 #[cfg(feature = "gc")]
 use super::GcHeapAllocationIndex;
@@ -305,8 +309,12 @@ pub struct PoolingInstanceAllocator {
     live_component_instances: AtomicU64,
 
     decommit_queue: Mutex<DecommitQueue>,
+
     memories: MemoryPool,
+    live_memories: AtomicUsize,
+
     tables: TablePool,
+    live_tables: AtomicUsize,
 
     #[cfg(feature = "gc")]
     gc_heaps: GcHeapPool,
@@ -335,6 +343,8 @@ impl Drop for PoolingInstanceAllocator {
 
         debug_assert_eq!(self.live_component_instances.load(Ordering::Acquire), 0);
         debug_assert_eq!(self.live_core_instances.load(Ordering::Acquire), 0);
+        debug_assert_eq!(self.live_memories.load(Ordering::Acquire), 0);
+        debug_assert_eq!(self.live_tables.load(Ordering::Acquire), 0);
 
         debug_assert!(self.memories.is_empty());
         debug_assert!(self.tables.is_empty());
@@ -357,7 +367,9 @@ impl PoolingInstanceAllocator {
             live_core_instances: AtomicU64::new(0),
             decommit_queue: Mutex::new(DecommitQueue::default()),
             memories: MemoryPool::new(config, tunables)?,
+            live_memories: AtomicUsize::new(0),
             tables: TablePool::new(config)?,
+            live_tables: AtomicUsize::new(0),
             #[cfg(feature = "gc")]
             gc_heaps: GcHeapPool::new(config)?,
             #[cfg(feature = "async")]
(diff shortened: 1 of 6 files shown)


crates/wasmtime/src/runtime/vm/instance/allocator/pooling.rs
```diff
@@ -712,21 +732,28 @@ unsafe impl InstanceAllocator for PoolingInstanceAllocator {
         ty: &wasmtime_environ::Table,
         _table_index: DefinedTableIndex,
     ) -> Result<(super::TableAllocationIndex, Table)> {
-        // FIXME: see `allocate_memory` above for comments about duplication
-        // with `with_flush_and_retry`.
-        let e = match self.tables.allocate(request, ty).await {
-            Ok(result) => return Ok(result),
-            Err(e) => e,
-        };
+        self.live_tables.fetch_add(1, Ordering::AcqRel);
```
lann: I copied this approach from `increment_component_instance_count` but I could easily be convinced that it would be fine to just increment after successful allocation.

crates/wasmtime/src/runtime/vm/instance/allocator/pooling.rs
```diff
@@ -652,23 +664,30 @@ unsafe impl InstanceAllocator for PoolingInstanceAllocator {
...
     ) -> Result<(MemoryAllocationIndex, Memory)> {
-        // FIXME(rust-lang/rust#145127) this should ideally use a version of
-        // `with_flush_and_retry` but adapted for async closures instead of only
-        // sync closures. Right now that won't compile though so this is the
-        // manually expanded version of the method.
-        let e = match self.memories.allocate(request, ty, memory_index).await {
-            Ok(result) => return Ok(result),
-            Err(e) => e,
-        };
+        self.live_memories.fetch_add(1, Ordering::AcqRel);
```
lann: I was on the fence about using `Ordering::Relaxed` for these, but if someone ends up using metrics in actual business logic (like server backpressure) it would be a pretty nasty surprise for these to underflow...

alexcrichton: I'd actually opt towards using `Relaxed` here since there's explicitly no data dependency with anything else in the system (e.g. this isn't looking for memory effects of another operation or publishing its own memory effects). Could you detail more about the underflow point though? How would atomic ordering affect that?

crates/wasmtime/src/runtime/vm/instance/allocator/pooling.rs
```diff
@@ -652,23 +664,30 @@ unsafe impl InstanceAllocator for PoolingInstanceAllocator {
...
             }
-        }
 
-        Err(e)
+            Err(e)
+        }
+        .await
+        .inspect_err(|_| {
+            self.live_memories.fetch_sub(1, Ordering::AcqRel);
+        })
```
alexcrichton: You'll want to switch this to using a `Drop` implementation because otherwise this won't be correctly accounted for if the future is cancelled/dropped.

crates/wasmtime/src/runtime/vm/instance/allocator/pooling.rs
alexcrichton: Given the need to have a `Drop` to handle the cancelled-the-future case I'd be ok just putting this in the `Ok` path since that would avoid the need for `Drop` entirely

crates/wasmtime/src/runtime/vm/instance/allocator/pooling.rs
```diff
@@ -735,6 +762,7 @@ unsafe impl InstanceAllocator for PoolingInstanceAllocator {
         allocation_index: TableAllocationIndex,
         mut table: Table,
     ) {
+        self.live_tables.fetch_sub(1, Ordering::AcqRel);
```
alexcrichton: Perhaps `debug_assert!` the previous value is `> 0`?

crates/wasmtime/src/runtime/vm/instance/allocator/pooling/metrics.rs
```diff
@@ -0,0 +1,114 @@
...
+impl PoolingAllocatorMetrics {
+    pub(crate) fn new(engine: &Engine) -> Option<Self> {
+        engine.allocator().as_pooling().map(|_| Self {
+            engine: engine.clone(),
+        })
+    }
+
+    /// Returns the number of core (module) instances currently allocated.
+    pub fn core_instances(&self) -> u64 {
+        self.allocator().live_core_instances.load(Ordering::Acquire)
```
alexcrichton: Similar to above, I'd say that using `Relaxed` here is fine since these aren't actually synchronizing with anything other than themselves.

alexcrichton (approved):

crates/wasmtime/src/runtime/vm/instance/allocator/pooling.rs
lann: Ah good point; I didn't think about this hard enough when I rebased on #11470. I'll just switch to incrementing on Ok anyway.

crates/wasmtime/src/runtime/vm/instance/allocator/pooling.rs
lann: Sounds good. I don't really have a good grasp of relaxed atomic semantics so I just pessimistically assume that a concurrent thread would be able to go back in time and kill its own grandparent or whatever.

alexcrichton: Heh nah makes sense. I'm by no means an expert but I'm confident enough in my understanding that if atomics are used for things like metric purposes it's fine to use relaxed. Relaxed ordering, AFAIK, basically means that you get no guarantees about other memory locations in the program when you observe a value. For the memory location in question, though, everything is still just as atomic as you'd expect (e.g. fetch-and-add still does that, it doesn't like switch to non-atomics). This happens because relaxed atomics can be reordered around each other by the compiler, so you can't for example say "I saw N memories in use, therefore I must see N+1 tables in use next", that's not valid.

lann: Yeah makes sense. I guess I'm not entirely clear on what makes `increment_component_instance_count` different here (above) that it would need `AcqRel`?

alexcrichton: Oh it's not, the main difference is I saw this here and no one saw it there most likely. There's no need for that to use `AcqRel` either and IMO it should also be using `Relaxed`

alexcrichton:
While the vet error is spurious, the [miri failure](https://github.com/bytecodealliance/wasmtime/actions/runs/17137044780/job/48615486728) looks legitimate.

lann:
This does indeed fail under miri locally: https://github.com/bytecodealliance/wasmtime/blob/fc3f429d72fabd4fb669f17c977c11752e078313/crates/wasmtime/src/runtime/vm/instance/allocator/pooling/metrics.rs#L76-L77

Presumably I can work around this by overriding some default config, but is it a bug that the defaults don't work with miri?

lann:
Oh well I guess that's pretty straightforward: https://github.com/bytecodealliance/wasmtime/blob/fc3f429d72fabd4fb669f17c977c11752e078313/crates/environ/src/tunables.rs#L149-L150

Any preference on how to fix this?

- Add a `cfg!(miri)` for default `max_memory_size`
- Just set `max_memory_size` in the test

alexcrichton:
The main test suite has a [helper function for a "small pool config"](https://github.com/bytecodealliance/wasmtime/blob/f5af05e25533246f8f0750f3bfff5ba6456072f0/tests/all/main.rs#L94-L113) which could be modeled here as well, but this test isn't too interesting for miri so it's also fine to slap a `#[cfg_attr(miri, ignore)]` on the test

Issue #11489, opened 2025-08-21
Exceptions: dynamic context load during stack walk doesn't account for stack-arg region


vouillon:
### Test Case

[bug.zip](https://github.com/user-attachments/files/21920759/bug.zip)

### Steps to Reproduce

Run the following command:

```
./target/debug/wasmtime  -W=all-proposals=y bug.wasm
```

### Expected Results

It should not crash

### Actual Results

I get a segfault

### Versions and Environment

Wasmtime commit: 2d25f862b38abd484e5418327a9149e69d3274aa

Operating system: Linux

Architecture: x64

### Extra Info

I suspect that the stack unwinder does not deal properly with tail calls.

I get the following stack trace in gdb:
```
#0  wasmtime::runtime::vm::ModuleRuntimeInfo::env_module (self=0x7ffeffffff68)
    at crates/wasmtime/src/runtime/vm.rs:329
#1  0x0000555557fd524f in wasmtime::runtime::vm::instance::Instance::env_module
    (self=0x7ffeffffff60) at crates/wasmtime/src/runtime/vm/instance.rs:357
#2  0x0000555557fd5fe8 in wasmtime::runtime::vm::instance::Instance::get_exported_tag (self=0x7ffeffffff60, store=..., index=...)
    at crates/wasmtime/src/runtime/vm/instance.rs:666
#3  0x0000555557fb4cad in wasmtime::runtime::vm::throw::compute_throw_action::{closure#0} (frame=0x7ffff69f48f0) at crates/wasmtime/src/runtime/vm/throw.rs:98
#4  0x0000555557f12e5e in wasmtime_internal_unwinder::throw::compute_throw_action::{closure#0}<wasmtime::runtime::vm::throw::compute_throw_action::{closure_env#0}> (frame=...) at crates/unwinder/src/throw.rs:76
#5  0x0000555558034d24 in wasmtime_internal_unwinder::stackwalk::visit_frames<wasmtime_internal_unwinder::throw::ThrowAction, wasmtime_internal_unwinder::throw::compute_throw_action::{closure_env#0}<wasmtime::runtime::vm::throw::compute_throw_action::{closure_env#0}>> (unwind=..., pc=140737353797698, 
    fp=140737331026544, trampoline_fp=140737331026560, f=...)
    at crates/unwinder/src/stackwalk.rs:203
[...]

PR #11495, opened 2025-08-21
Support exceptions with Pulley

### LINKED ISSUE #11486: Exceptions: support Pulley
The exception-handling support as implemented in #11326 does not yet support Pulley. It is almost there, but not quite:

- Pulley's use of Cranelift for code generation means that we *do* get exception metadata alongside Pulley bytecode, and a module that uses exceptions can be *compiled*;
- The support for stack-walking of Pulley stacks means that the exception unwinder *can* find handlers in Pulley code;
- However, the `raise` libcall implementation that Pulley uses does not account for `UnwindState::UnwindToWasm`. This means that we'll reach host-to-Wasm entry instead, with an unexpected unwind state, and panic.

In [this comment](https://github.com/bytecodealliance/wasmtime/pull/11326#issuecomment-3201732110) I outline an alternative approach: we should

- invoke the real `raise` libcall, just as we do other libcalls, from Pulley;
- when we enter `unwind`, and either `wasmtime_longjmp()` or `wasmtime_unwinder::resume_to_exception_handler()`, we instead check the store's executor and
  - on native code, do as we do today
  - on Pulley, update the Pulley interpreter state appropriately

alexcrichton:
This implements Chris's suggestion to avoid special-casing the `raise` libcall and instead let it fully execute for Pulley. The switch of whether to do a native longjmp or a Pulley longjmp is done based on the interpreter state of Wasmtime's configured executor.

Closes #11486

### DIFF
--- a/crates/wasmtime/src/runtime/vm/interpreter.rs
+++ b/crates/wasmtime/src/runtime/vm/interpreter.rs
@@ -254,7 +254,7 @@ impl InterpreterRef<'_> {
                 DoneReason::CallIndirectHost { id, resume } => {
                     let state = unsafe { self.call_indirect_host(id) };
 
-                    // After the host as finished take a look at what hostcall
+                    // After the host has finished take a look at what hostcall
                     // was just made. The `raise` hostcall gets special
                     // handling for its non-local transfer of control flow,
                     // notably here we see if it's a longjmp or a resume that
--- a/crates/wasmtime/src/runtime/vm/interpreter.rs
+++ b/crates/wasmtime/src/runtime/vm/interpreter.rs
@@ -1,10 +1,11 @@
 use crate::runtime::vm::vmcontext::VMArrayCallNative;
 use crate::runtime::vm::{
     StoreBox, TrapRegisters, TrapTest, VMContext, VMOpaqueContext, f32x4, f64x2, i8x16, tls,
+    traphandlers,
 };
 use crate::{Engine, ValRaw};
 use core::marker;
-use core::ptr::NonNull;
+use core::ptr::{self, NonNull};
 use pulley_interpreter::interp::{DoneReason, RegType, TrapKind, Val, Vm, XRegVal};
 use pulley_interpreter::{FReg, Reg, XReg};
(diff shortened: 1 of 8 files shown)


cfallin:
Happy to take this one for review (and thanks for doing this!). Working out a fix to #11489 which I'll want to make sure doesn't conflict (probably won't?) then will review this.

crates/wasmtime/src/runtime/vm/traphandlers.rs
```diff
@@ -433,7 +446,7 @@ where
 
     let result = CallThreadState::new(store.0, old_state).with(|cx| match store.0.executor() {
         // In interpreted mode directly invoke the host closure since we won't
-        // be using host-based `setjmp`/`longjmp` as that's not going to save
+        // be using host-based `etjmp`/`longjmp` as that's not going to save
```
fitzgen: 👀

cfallin: I believe [this is an ET jmp](https://i.imgflip.com/8bd001.jpg), for reference

alexcrichton: One day we'll have support for that in Wasmtime...

crates/wasmtime/src/runtime/vm/interpreter.rs
```diff
@@ -228,19 +249,39 @@ impl InterpreterRef<'_> {
...
-                // here based on `sig`. Once that returns we can resume
+                // here based on `id`. Once that returns we typically resume
                 // execution at `resume`.
-                //
-                // Note that the `raise` libcall is handled specially here since
-                // longjmp/setjmp is handled differently than on the host.
                 DoneReason::CallIndirectHost { id, resume } => {
+                    let state = unsafe { self.call_indirect_host(id) };
+
+                    // After the host as finished take a look at what hostcall
```
cfallin: s/as/has/

cfallin (approved):
Looks great, thanks!

PR #11500, opened 2025-08-21
Wasmtime/Cranelift: carry "FP to SP offset" in exception data, and use it in stackwalk.

### LINKED ISSUE #11489: Exceptions: dynamic context load during stack walk doesn't account for stack-arg region

cfallin:
Currently Wasmtime unwinds stack frames to look for exception handlers by walking frames one-by-one, following the FP chain as usual, and assuming that *these frames are contiguous*: that is, that the SP in any given frame (bottom of that frame) is immediately above the FP of the next lower frame, plus the FP/return address pair (e.g. 16 bytes). This allows us to get the SP for any given frame in addition to FP. We need SP for two reasons:

- To look up dynamic context, to match Wasm tag instances for handlers against the thrown tag;
- To actually set SP when we resume, if we do resume to a handler in this frame.

This logic *almost but not quite* worked: I had forgotten that in our tail-call ABI, we need to clean up incoming stack args in the callee (because only the final callee in a parade of tail-calling functions that reuse the same stack frame location knows how many args it has, not the original caller). This implies that there is an "incoming args area" *above* the FP/return address pair. Thus, frames are not necessarily contiguous by the above definition.

In #11489 we see a case where a function of signature `(func)` tail-calls one of `(func (param i32 i32 i32 i32 i32))`, which on x86-64 (with four arg registers left for Wasm) is sufficient to create incoming stack args, which then trips up the unwinder, reading a bogus vmctx and segfaulting.

The most reasonable solution seems to be to embed the SP-to-FP offset in the exception metadata itself, so from only the FP (which is totally robust -- we rely on the FP chain for multiple kinds of stack-walking) we can get the SP, allowing us to read dynamic context and to reset SP during resume.

This PR does just that. Technically, in our ABI, the SP-to-FP offset is constant for an entire function, but it was simpler in the exception metadata to encode this per callsite instead (there is no other notion of "per-function" data, only "per-callsite", so it would be a separate binary search).

Fixes #11489.

prtest:full


cranelift/codegen/src/machinst/buffer.rs
```diff
@@ -2017,6 +2020,13 @@ pub struct MachCallSite {
     /// start of the buffer*.
     pub ret_addr: CodeOffset,
 
+    /// The offset from the FP at this callsite down to the SP when
+    /// the call occurs, if known. In other words, the size of the
+    /// stack frame up to the saved FP slot. Useful to recover the
+    /// start of the stack frame and to look up dynamic contexts
+    /// stored in [`ExceptionContextLoc::SPOffset`].
+    pub frame_offset: Option<u32>,
```
alexcrichton: Could this (and below) expand documentation to explain what `None` means?

crates/wasmtime/src/runtime/vm/throw.rs
```diff
@@ -43,42 +43,39 @@ pub unsafe fn compute_throw_action(store: &mut dyn VMStore) -> ThrowAction {
...
-            frame.sp().unwrap(),
             frame.pc()
         );
         let module = nogc.modules().lookup_module_by_pc(frame.pc())?;
         let base = module.code_object().code_memory().text().as_ptr() as usize;
         let rel_pc = u32::try_from(frame.pc().wrapping_sub(base)).expect("Module larger than 4GiB");
         let et = module.exception_table();
-        for handler in et.lookup_pc(rel_pc) {
+        let (frame_offset, handlers) = et.lookup_pc(rel_pc);
+        let fp_to_sp = -isize::try_from(frame_offset.unwrap_or(0)).unwrap();
```
alexcrichton: Could this have a comment as to why `unwrap_or` is ok here? Or otherwise, if `None` is the same as 0 should this be represented as a u32 instead of a `Option<u32>`?

tests/misc_testsuite/stack-args-exceptions.wast
```diff
@@ -0,0 +1,34 @@
...
+    i32.const 0
+    i32.const 0
+    return_call $throw
+  )
+  (func $throw (param i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 i32)
+    throw $tag
+  )
+)
+
+(invoke "_start")
```
alexcrichton: Would it be worth it to add a disas test for `$throw` to ensure that it's handling the offset correctly?

cfallin: The interesting new bit is actually the offset at the `try_table` site (callsite in `$start`), which is the same as the case in `exceptions.wat`, so I think we're covered already!

crates/wasmtime/src/runtime/vm/throw.rs
cfallin: Updated to explicitly `.expect()` only when there are handlers. The subtlety is that `frame_offset` will be `None` if there are no handlers because we don't encode any metadata in that case.

cranelift/codegen/src/machinst/buffer.rs
cfallin: Updated!

PR #11510, opened 2025-08-22
Refactor resetting memory on `MemoryImageSlot` drop

alexcrichton:
This commit refactors the behavior of dropping a `MemoryImageSlot` to no longer map anonymous memory into the slot. This behavior was implemented previously because if a `MemoryImageSlot` is dropped then the state of the slot is unknown and to prevent any sort of data leakage a reset is performed.

This reset operation, however, is fallible in that it calls `mmap`. Calls to `mmap` can fail due to `ENOMEM`, for example, if the process has reached its VMA limit. This means that if a process is in a near-OOM condition then failing to allocate a memory image could panic the process due to the `unwrap()` in the destructor of `MemoryImageSlot`. The purpose of this commit is to avoid this `unwrap()` and instead move the reset behavior to a location where an error can be propagated.

This commit removes the clear-on-drop behavior of `MemoryImageSlot` slot. This was already disabled everywhere except the pooling allocator. The pooling allocator now maintains an extra bit of state-per-slot where instead of storing `Option<MemoryImageSlot>` it now stores effectively one other variant of "unknown". On reuse of an "unknown" slot the memory is reset back to an anonymous mapping and this is all done in a context where an error can be propagated.

Two tests are added in this commit to confirm all of this behavior:

* The first test is a new test that passes both before and after this commit which performs a failed allocation of a memory slot. A successful allocation is then made to ensure that the previous image is not present and zero memory is present. This test fails before the commit if the clear-on-drop behavior is removed, and it fails with this commit if the clear-on-reusing-unknown behavior is removed. Effectively this test ensures that the clear-on-unknown-state logic is present.

* The second test is a new test that panicked before this commit and passes afterwards. This second test exhausts all VMAs in the current process, or at least most of them, and then tries to allocate some instances with an image. Instance allocation will eventually fail and cause the erroneous path to get executed. This previously unwrapped a `ENOMEM` failure, and now it can be handled gracefully by the embedder.

### DIFF
--- a/crates/wasmtime/src/runtime/vm/instance/allocator/pooling/memory_pool.rs
+++ b/crates/wasmtime/src/runtime/vm/instance/allocator/pooling/memory_pool.rs
@@ -64,6 +64,7 @@ use crate::{
     runtime::vm::mpk::{self, ProtectionKey, ProtectionMask},
     vm::HostAlignedByteCount,
 };
+use std::mem;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex};
 use wasmtime_environ::{DefinedMemoryIndex, Module, Tunables};
@@ -112,7 +113,7 @@ pub struct MemoryPool {
 
     /// If using a copy-on-write allocation scheme, the slot management. We
     /// dynamically transfer ownership of a slot to a Memory when in use.
-    image_slots: Vec<Mutex<Option<MemoryImageSlot>>>,
+    image_slots: Vec<Mutex<ImageSlot>>,
 
     /// A description of the various memory sizes used in allocating the
     /// `mapping` slab.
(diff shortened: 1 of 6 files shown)


alexcrichton:
Some follow-up bits on this:

* Personally I wouldn't consider this a CVE since panicking in near-OOM conditions is not something we can protect against. For example `Vec::with_capacity` can panic in OOM conditions at any time.
* This was discovered in Spin recently, so if amenable to others I'd probably do another point release of 36.0.2 so we can update to a fixed version more quickly.

crates/wasmtime/tests/pooling_alloc_near_oom.rs
fitzgen, line 1: I imagine that these tests will want to return early on `WASMTIME_TEST_NO_HOG_MEMORY` or whatever the env var is for our tests under `qemu`.

alexcrichton: Good point! Let's see what CI says...

alexcrichton:
```
qemu-arm: GLib: ../../../glib/gmem.c:106: failed to allocate 40 bytes
qemu-arm: QEMU internal SIGTRAP {code=128, addr=(nil)}
error: test failed, to rerun pass `-p wasmtime --test pooling_alloc_near_oom`

Caused by:
  process didn't exit successfully: `/opt/hostedtoolcache/qemu/bin/qemu-arm -L /usr/arm-linux-gnueabihf -E LD_LIBRARY_PATH=/usr/arm-linux-gnueabihf/lib /home/runner/work/wasmtime/wasmtime/target/armv7-unknown-linux-gnueabihf/debug/deps/pooling_alloc_near_oom-e44c0c4e75189ad2` (signal: 5, SIGTRAP: trace/breakpoint trap)
```

Whelp looks like I'll need to disable this in QEMU yeah

crates/wasmtime/src/runtime/vm/instance/allocator/pooling/memory_pool.rs
```diff
@@ -503,21 +540,49 @@ impl MemoryPool {
...
+    /// slot cannot be used at this time.
+    fn take_memory_image_slot(
+        &self,
+        allocation_index: MemoryAllocationIndex,
+    ) -> Result<MemoryImageSlot> {
+        let (maybe_slot, needs_reset) = {
+            let mut slot = self.image_slots[allocation_index.index()].lock().unwrap();
+            match mem::replace(&mut *slot, ImageSlot::Unknown) {
+                ImageSlot::Unmapped => (None, false),
+                ImageSlot::Unknown => (None, true),
```
tschneidereit, line 565: I might misunderstand how this works, but shouldn't this return `(Some(state), true)` to actually reuse the slot after resetting it? By my reading, the current code creates a new slot and then resets that one.

tschneidereit: Oh, never mind: I get it now. The reset operates on the memory as identified by `allocation_index`, not the slot itself.

Issue #11505, opened 2025-08-21
Panic unwrapping `None` compiling exceptions-related fuzz test case


alexcrichton:
This input:

```wasm
(module
  (type (;0;) (func))
  (tag (;0;) (type 0))
  (global (;0;) (mut i32) i32.const 1000)
  (export "" (func 0))
  (func (;0;) (type 0)
    global.get 0
    i32.eqz
    if ;; label = @1
      unreachable
    end
    global.get 0
    i32.const 1
    i32.sub
    global.set 0
    loop (type 0) ;; label = @1
      global.get 0
      i32.eqz
      if ;; label = @2
        unreachable
      end
      global.get 0
      i32.const 1
      i32.sub
      global.set 0
      return_call 0
      try_table ;; label = @2
        try_table (type 0) (catch_all 0 (;@2;)) (catch_all 0 (;@2;)) (catch_all 0 (;@2;)) ;; label = @3
        end
      end
    end
  )
)
```

panics with:

```
$ cargo run compile testcase0.wat -W exceptions
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.10s
     Running `target/x86_64-unknown-linux-gnu/debug/wasmtime compile testcase0.wat -W exceptions`

thread '<unnamed>' panicked at crates/cranelift/src/translate/code_translator.rs:3162:45:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

@cfallin I was curious to see what would happen since wasm-smith has support for exceptions. I don't think the support is too useful from a runtime perspective but it's likely interesting from a compile-time perspective (similar with most of our other wasm-smith-generated modules). If you're curious this is the diff I used to get `differential` running and it took a few seconds to find the above case:

<details>

```diff
diff --git a/crates/fuzzing/src/generators/config.rs b/crates/fuzzing/src/generators/config.rs
index 553bbea530..72b9c0fc46 100644
--- a/crates/fuzzing/src/generators/config.rs
+++ b/crates/fuzzing/src/generators/config.rs
@@ -662,6 +662,7 @@ impl WasmtimeConfig {
                 config.config.gc_enabled = false;
                 config.config.tail_call_enabled = false;
                 config.config.reference_types_enabled = false;
+                config.config.exceptions_enabled = false;
                 config.function_references_enabled = false;

                 // Winch's SIMD implementations require AVX and AVX2.
[...]

PR #11511, opened 2025-08-22
Exceptions: add basic support for fuzzing.

### LINKED ISSUE #11485: Exceptions: fuzz against other exceptions-capable engines
We should:

- [ ] ensure that wasm-smith can create modules with `try_table`/`throw`/`throw_ref`
- [ ] ensure that we permit exceptions to be chosen by our fuzzing configuration generation
- [ ] ensure that exceptions are enabled in our oracle engine(s) that support it -- at least V8 and the spec interpreter
- [ ] ensure that we can compare an "exception thrown" result

in order to test that the Wasm exception-handling proposal implementation is robust.

cfallin:
This pulls in Alex's patch from [here](https://github.com/bytecodealliance/wasmtime/issues/11505#issue-3343508228) as the first commit, and an update to `DiffValue` comparison to support exnrefs as the second commit.

I've fuzzed locally with `differential` and (i) found the exnref comparison that led to the second commit, so I know that this is doing "something", and (ii) hit a segfault in V8 on a divide-by-zero, so that's a thing but I think not my thing.

Fixes #11485.

### DIFF
--- a/crates/fuzzing/src/generators/config.rs
+++ b/crates/fuzzing/src/generators/config.rs
@@ -662,6 +662,7 @@ impl WasmtimeConfig {
                 config.config.gc_enabled = false;
                 config.config.tail_call_enabled = false;
                 config.config.reference_types_enabled = false;
+                config.config.exceptions_enabled = false;
                 config.function_references_enabled = false;
 
                 // Winch's SIMD implementations require AVX and AVX2.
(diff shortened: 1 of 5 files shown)


cfallin:
Test failures seem to be unrelated, from a new nightly warning perhaps (?) -- cc @alexcrichton on [this](https://github.com/bytecodealliance/wasmtime/actions/runs/17160152385/job/48687121824?pr=11511) ?

PR #11514, opened 2025-08-22
GC: replace ManuallyRooted with OwnedRooted.

### LINKED ISSUE #11445: Wasm GC host API: a slightly less manual owned `Rooted` type?
Upon pondering the various rooting options in #11326, and reading through the discussion of API tradeoffs [here](https://github.com/bytecodealliance/wasmtime/blob/82f3b2a16c139297684ae0c84a7568a30c762f41/crates/wasmtime/src/runtime/gc/enabled/rooting.rs#L6), it occurs to me that there may be a possible addition to the API, which offers more natural "owned root" semantics in Rust but is still pay-as-you-go, which I'd like to sketch here.

## Background and Motivation

Currently, we have two types, `Rooted<T>` and `ManuallyRooted<T>`. The former has a LIFO discipline, implicitly attaching to the deepest `RootScope` in a stack of scopes on the store, and the latter is completely manual. Notably, the latter does *not* have a `Drop` impl that actually unroots, because doing so would require holding a reference to the `Store` somehow -- implicitly with `Arc`s somehow, because actual borrows of the `Store` would preclude any other operations.

The LIFO discipline of `Rooted` works well when it works, i.e., when the user is aware of their scopes, but in [this comment](https://github.com/bytecodealliance/wasmtime/pull/11326#issuecomment-3192952035) I outline a few of the realizations I've had around ergonomics when it combines with other features -- in particular, Rust's `?` operator, which pushes users to naturally "propagate errors upward without explicit thought". The usual expectation is that the `E` type in the `Result` somehow owns its info. So a dynamic failure (a panic, no less!) when that `E` crosses over a scope is quite surprising. Exceptional returns are the most obvious example to me right now, but I believe some of the surprise here may also occur with users who naively (but understandably) expect that "GC" means they don't have to worry about lifetimes. Said more succinctly, a type named `Rooted` in a language like Rust where types often imply ownership might imply that it keeps a root as long as it exists. I know I certainly had that expectation at first. (The docs are very good, but "least surprise" still applies here!)

We provide the "escape hatch", as the docs describe, of `ManuallyRooted`, and this can certainly work well if the user has extreme discipline -- unfortunately, the requirement of a manual unrooting step means that it is *very* easy to get wrong, again as the docs describe well.

The middle ground, of a type that somehow unregisters itself on `Drop` by keeping enough of a handle on the engine's internal state to do so, is described as impractical: it would require an `Arc<Mutex<...>>` to track internal handle state/registration, as the docs say. This would create synchronization overhead, and potentially pessimize common-case GC operations too. Is there a better way?

## Idea: Pay-as-you-Go "Arc'd liveness flags"

Ideally, I want something that:

- Acts as an "owned root", i.e., keeps the referent alive as long as the root type exists and the store exists
- Has a proper `Drop` such that dropping the root type ensures the referent is not leaked(*)
- Imposes zero overhead on GC if these roots are not used, and imposes only small overhead proportional to the number of such roots if they are

Note the sneaky (*) above: I want to avoid leaks, not to eagerly detect when an unregistration occurs. In other words, let's permit deferring some action from the `Drop` to, say, the next time a GC runs.

Then to avoid synchronization overhead and contention, rather than a large monolithic registry under a single mutex, let's have some small shared state per owned root.

The idea is: keep a "liveness flag" in an `Arc<AtomicBool>`. In steady state, when live, there are *two* references to this `Arc`: from the owned root type, and from a `owned_roots: Vec<(Option<VMGcRef>, Arc<AtomicBool>)>` in the GC roots list. When the owned root drops, it sets its atomic bool to false, and drops its `Arc` reference. When the GC scans roots, it reads out the liveness flags, and removes those roots that the owner has dropped. (E.g. via a `retain` on the `Vec`.)[^1]

[^1]: Slight variant: the tuple above could instead be a single `Arc<OwnedRootData>` with liveness and the `VMGcRef`, and maybe that's cleaner; I haven't thought too much about how this would live alongside `ManuallyRooted` and whether it would want to share a GC-ref root array somewhere else or not...

Two realizations make this more efficient than the `Arc<Mutex<whole root list>>` approach: (i) we *do* have a mut borrow to the store when we *create*, so it's fine to have a normal `Vec` of roots registered -- only `Drop` is "remote" without the store; (ii) we have a *separate* bit of state per root, and it's just an atomic bool, which on common architectures (x86 and Apple Silicon's aarch64 at least) has atomic loads that are exactly as cheap as normal loads.

It's also fully safe Rust (`Arc` makes it so), and is pay-as-you-go: with no such roots existing, GC root scanning has one check of vec-is-empty and a never-taken branch; as close to zero-overhead as we can make it. There is no mutex contention anywhere, because the only "meeting point" is the `Vec` that's mutated under a `&mut Store`. In terms of memory allocation, it's certainly more expensive than a LIFO root (which is just an index into an array!), because there's the separate `Arc` allocation, but I suspect most uses of these roots are likely to be relatively high-level "entry points" or cases like exception returns where scopes don't map well to usage patterns; we can encourage use of LIFO scopes where possible.

## Naming

I've called this an "owned root" and I'd gently suggest considering different names for our current `Rooted`, to more explicitly describe the difference, if we adopt this -- something like `ScopedRooted` vs. `OwnedRooted`?

cfallin:
This implements the ideas from #11445: it replaces `ManuallyRooted`, which requires an explicit unroot action with a mut borrow of the store (making it impossible to implement in a standard `Drop` impl), with `OwnedRooted`, which holds an `Arc` only to a small auxiliary memory allocation (an `Arc<()>`) and uses this externalized "liveness flag" to allow for a `Store`-less drop. These liveness flags are scanned during a "trim" pass, which happens both when new owned roots are created, and just before a GC.

This should greatly increase safety for host-side users of GC: it provides a way to have a handle whose ownership works like any other Rust value, alive until dropped. It is still not quite as efficient as LIFO-scoped handles (by analogy, for the same reason that individually-freed RAII types are not as efficient as arena allocation), so those remain for efficiency-minded users that have a clear picture of reference lifetimes.

At some later time we may wish to use `OwnedRooted` exclusively in our public APIs rather than `Rooted`, and we may wish to rename `Rooted` to `ScopedRooted`, but I haven't done either of those things yet.

I opted to *replace* `ManuallyRooted` rather than add a third kind of root, after discussion with fitzgen. One implication of this is that the C API's `anyref` and `externref` types are now 24 or 20 bytes rather than 16 (because of the `Arc` pointer), and correspondingly the Val union grew to that size. I *believe* this is an acceptable tradeoff, but I'm happy to put `ManuallyRooted` back if not.

Fixes #11445.

### DIFF
--- a/crates/wasmtime/src/runtime/gc/enabled/rooting.rs
+++ b/crates/wasmtime/src/runtime/gc/enabled/rooting.rs
@@ -27,24 +27,70 @@
 //!    not a burden for users. Additionally, the API's types should be `Sync`
 //!    and `Send` so that they work well with async Rust.
 //!
-//! For example, goals (3) and (4) are in conflict when we think about how to
-//! support (2). Ideally, for ergonomics, a root would automatically unroot
-//! itself when dropped. But in the general case that requires holding a
-//! reference to the store's root set, and that root set needs to be held
-//! simultaneously by all GC roots, and they each need to mutate the set to
-//! unroot themselves. That implies `Rc<RefCell<...>>` or `Arc<Mutex<...>>`! The
-//! former makes the store and GC root types not `Send` and not `Sync`. The
-//! latter imposes synchronization and locking overhead. So we instead make GC
-//! roots indirect and require passing in a store context explicitly to unroot
-//! in the general case. This trades worse ergonomics for better performance and
-//! support for moving GC.
+//! The two main design axes that trade off the above goals are:
+//!
+//! - Where the GC reference itself is held. A root object could
+//!   directly hold the underlying GC reference (offset into the GC
+//!   storage area), which would allow more efficient dereferencing
+//!   and access to the referred-to GC object. However, goal (2)
+//!   requires that the GC is able to update references when objects
+//!   move during a GC. Thus, such "direct roots" would need to be
+//!   registered somehow in a global root registry, and would need to
+//!   unregister themselves when dropped.
+//!
+//!   The alternative is to hold some indirect kind of reference to a
+//!   GC reference, with the latter stored directly in the `Store` so
+//!   the GC can update it freely. This adds one pointer-chasing hop
+//!   to accesses, but works much more nicely with ownership
+//!   semantics. Logically, the `Store` "owns" the actual pointers;
+//!   and rooting types own the slots that they are stored in.
+//!
+//!   For the above reasons, all of our rooting types below use
+//!   indirection. This avoids the need for an unsafe
+//!   intrusive-linked-list for global registration, or a shared
+//!   reference to a mutex-protected registry, or some other
+//!   error-prone technique.
+//!
+//! - How unrooting occurs. Ideally, a rooting type would implement
+//!   the Rust `Drop` trait and unroot itself when the Rust value is
+//!   dropped. However, because the rooting state is held in the
+//!   `Store`, this direct approach would imply keeping a shared,
+//!   mutex-protected handle to the registry in every rooting
+//!   object. This would add synchronization overhead to the common
+//!   case, and in general would be a bad tradeoff.
+//!
+//!   However, there are several other approaches:
+//!
+//!   - The user could use an RAII wrapper that *does* own the `Store`,
+//!     and defines a "scope" in which roots are created and then
+//!     bulk-unrooted at the close of the scope.
+//!   - The rooting type could hold a shared reference to some state
+//!     *other* than the full registry, and update a flag in that
+//!     state indicating it has been dropped; the `Store` could then
+//!     later observe that flag and remove the root. This would have
+//!     some allocation cost, but the shared state would be
+//!     independent of the `Store` and specific to each root, so would
+//!     impose no synchronization overhead between different roots or
+//!     the GC itself.
+//!   - The rooting type could provide a fully manual `unroot` method,
+//!     allowing the user to make use of their own knowledge of their
+//!     application's lifetimes and semantics and remove roots when
+//!     appropriate.
+//!
+//!   We provide an implementation of the first two of these
+//!   strategies below in `Rooted` and `OwnedRooted`. The last, fully
+//!   manual, approach is too difficult to use correctly (it cannot
+//!   implement Rust's `Drop` trait, but there is no way in Rust to
+//!   enforce that a value must be consumed rather than dropped) so it
+//!   is not implemented.
 //!
 //! ## Two Flavors of Rooting API
 //!
-//! Okay, with that out of the way, this module provides two flavors of rooting
-//! API. One for the common, scoped lifetime case, and another for the rare case
-//! where we really need a GC root with an arbitrary, non-LIFO/non-scoped
-//! lifetime:
+//! Okay, with that out of the way, this module provides two flavors
+//! of rooting API. One for the common, scoped lifetime case, and one that
+//! carries ownership until dropped, and can work as an RAII handle
+//! that interacts well with Rust ownership semantics (but at a minor
+//! performance cost):
 //!
 //! 1. `RootScope` and `Rooted<T>`: These are used for temporarily rooting GC
 //!    objects for the duration of a scope. The internal implementation takes
@@ -64,27 +110,40 @@
 //!    This supports the common use case for rooting and provides good
 //!    ergonomics.
 //!
-//! 2. `ManuallyRooted<T>`: This is the fully general rooting API used for
-//!    holding onto non-LIFO GC roots with arbitrary lifetimes. However, users
-//!    must manually unroot them. Failure to manually unroot a
-//!    `ManuallyRooted<T>` before it is dropped will result in the GC object
-//!    (and everything it transitively references) leaking for the duration of
-//!    the `Store`'s lifetime.
+//! 2. `OwnedRooted<T>`: This is a root that manages rooting and
+//!    unrooting automatically with its lifetime as a Rust value. In
+//!    other words, the continued existence of the Rust value ensures
+//!    the rooting of the underlying GC reference; and when the Rust
+//!    value is dropped, the underlying GC reference is no longer
+//!    rooted.
+//!
+//!    Internally, this root holds a shared reference to a
+//!    *root-specific* bit of state that is also tracked and observed
+//!    by the `Store`.  This means that there is minor memory
+//!    allocation overhead (an `Arc<()>`) for each such root; this
+//!    memory is shared over all clones of this root. The rooted GC
+//!    reference is *logically* unrooted as soon as the last clone of
+//!    this root is dropped. Internally the root may still exist until
+//!    the next GC, or "root trim" when another `OwnedRooted` is
+//!    created, but that is unobservable externally, and will not
+//!    result in any additional GC object lifetime because it is
+//!    always cleaned up before a gc.
 //!
 //!    This type is roughly similar to SpiderMonkey's [`PersistentRooted<T>`],
-//!    although they avoid the manual-unrooting with internal mutation and
-//!    shared references. (Our constraints mean we can't do those things, as
-//!    mentioned explained above.)
+//!    although they use internal mutation and shared references in a slightly
+//!    different way that produces more contention, because the rooting type
+//!    eagerly unroots itself in shared data structures from its destructor.
 //!
 //!    [`PersistentRooted<T>`]: http://devdoc.net/web/developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/JSAPI_reference/JS::PersistentRooted.html
 //!
-//! At the end of the day, both `Rooted<T>` and `ManuallyRooted<T>` are just
-//! tagged indices into the store's `RootSet`. This indirection allows working
-//! with Rust's borrowing discipline (we use `&mut Store` to represent mutable
-//! access to the GC heap) while still allowing rooted references to be moved
-//! around without tying up the whole store in borrows. Additionally, and
-//! crucially, this indirection allows us to update the *actual* GC pointers in
-//! the `RootSet` and support moving GCs (again, as mentioned above).
+//! At the end of the day, all of the above root types are just tagged
+//! indices into the store's `RootSet`. This indirection allows
+//! working with Rust's borrowing discipline (we use `&mut Store` to
+//! represent mutable access to the GC heap) while still allowing
+//! rooted references to be moved around without tying up the whole
+//! store in borrows. Additionally, and crucially, this indirection
+//! allows us to update the *actual* GC pointers in the `RootSet` and
+//! support moving GCs (again, as mentioned above).
 //!
 //! ## Unrooted References
 //!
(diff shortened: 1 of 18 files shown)


cfallin:
~~I've come up with a way to avoid the size growth on the C-API `ValUnion`, I think -- will update on Monday.~~ (actually perhaps not: the idea was to put the `Weak` directly in the slab of owned GC ref roots, but then the trim pass requires iteration over the whole index space rather than only live roots. The original C API struct tweak is probably better than pessimizing the otherwise fairly good asymptotic behavior here for the common case)

alexcrichton:
Updating C API bits I think is fine, but naively I'm surprised that this is still using a generational index. I thought that was only necessary due to the manual nature of `ManuallyRooted` and for something with a dtor like this we wouldn't need generational indices? I haven't fully thought this through but that feels redundant now.

Considering the C API some more too, this'll want to go over documentation with a fine-tooth comb as well because we've probably documented "you should call unroot, but nothing bad happens if you don't so long as you free the store". Now that needs to be changed to "you must call unroot or host memory will leak". That's a pretty big departure in behavior from before, which while not bad we need to be sure is clearly communicated.

crates/wasmtime/src/runtime/gc/enabled/rooting.rs
```diff
@@ -64,27 +110,40 @@
...
+//!    created, but that is unobservable externally, and will not
+//!    result in any additional GC object lifetime because it is
+//!    always cleaned up before a gc.
 //!
 //!    This type is roughly similar to SpiderMonkey's [`PersistentRooted<T>`],
-//!    although they avoid the manual-unrooting with internal mutation and
-//!    shared references. (Our constraints mean we can't do those things, as
-//!    mentioned explained above.)
+//!    although they use internal mutation and shared references in a slightly
+//!    different way that produces more contention, because the rooting type
```
fitzgen: Nitpick/aside: SM avoids contention/synchronization by essentially having `JSContext` not be `Send`. Our `Store` is `Send`, however -- and _must_ be for Wasmtime to work smoothly within Rust's async ecosystem -- so we would have to deal with that contention/synchronization. In practice, I doubt there would ever really be _contention_, but even the uncontended synchronization costs would be annoyingly high for us, almost definitely too high for serious consideration.

crates/wasmtime/src/runtime/gc/enabled/rooting.rs
```diff
@@ -466,6 +536,84 @@ impl RootSet {
...
+    /// root-trim first: the root-trim should be quite a lot cheaper,
+    /// and it will allow for more garbage to exist.
+    ///
+    /// There is, additionally, nothing stopping us from doing trims
+    /// more often than just before each GC, and there are reasons
+    /// this could be a good idea: for example, a user program that
+    /// creates and removes many roots (perhaps as it accesses the GC
+    /// object graph) but ultimately is dealing with a static graph,
+    /// without further allocation, may need the "root set" to be GC'd
+    /// independently from the actual heap. THus, we could trim before
```
fitzgen:
```suggestion
    /// independently from the actual heap. Thus, we could trim before
```

crates/wasmtime/src/runtime/gc/enabled/rooting.rs
```diff
@@ -466,6 +536,84 @@ impl RootSet {
...
+    /// goes just over the threshold, we might scan all N liveness
+    /// flags for each step, resulting in quadratic behavior overall.
+    ///
+    /// This, we implement an exponentially-growing "high water mark"
+    /// to guard against this latter case: on the add-a-new-root case,
+    /// we trim only if the list is longer than the high water mark,
+    /// and we grow the high water mark each time (by a geometric
+    /// factor).
+    ///
+    /// `eager` chooses whether we eagerly trim roots or pre-filter using the high-water mark.
```
fitzgen: Nitpick: it looks like this one line wasn't wrapped.

crates/wasmtime/src/runtime/gc/enabled/rooting.rs
```diff
@@ -466,6 +536,84 @@ impl RootSet {
...
+    pub(crate) fn trim_liveness_flags(&mut self, gc_store: &mut GcStore, eager: bool) {
+        if !eager {
+            const DEFAULT_HIGH_WATER: usize = 8;
+            const GROWTH_FACTOR: usize = 4;
```
fitzgen: A little surprised to see `4` here instead of `2`. I guess in general I'd prefer to err on the side of smaller growth factors before the next trim to help cut down on floating bloat. But now I am wondering if there is a specific reason you have this higher?

crates/wasmtime/src/runtime/gc/enabled/rooting.rs
```diff
@@ -1277,139 +1423,96 @@ where
...
-/// * Etc...
-///
-/// Wasmtime does *not* assert that a `ManuallyRooted<T>` is unrooted on `Drop`,
-/// or otherwise raise a panic, log a warning, or etc... on failure to manually
-/// unroot. Sometimes leaking is intentional and desirable, particularly when
-/// dealing with short-lived [`Store`][crate::Store]s where unrooting would just
-/// be busy work since the whole store is about to be dropped.
-#[repr(transparent)] // NB: the C API relies on this
-pub struct ManuallyRooted<T>
+pub struct OwnedRooted<T>
```
fitzgen, line 1564: I think we can `#[derive(Clone)]` for `OwnedRooted`, right?

crates/wasmtime/src/runtime/gc/enabled/rooting.rs
```diff
@@ -1277,139 +1423,96 @@ where
...
-impl<T> ManuallyRooted<T>
+impl<T: GcRef> Clone for OwnedRooted<T> {
+    fn clone(&self) -> Self {
+        OwnedRooted {
+            inner: self.inner,
+            liveness_flag: self.liveness_flag.clone(),
+            _phantom: marker::PhantomData,
+        }
+    }
+}
```
fitzgen, line 1616: That is, I think this can be derived now, instead of manually written out.

crates/wasmtime/src/runtime/gc/enabled/rooting.rs
cfallin: Ah, this one is tricky! `#[derive(Clone)]` creates a `T: Clone` constraint implicitly, I think because Rust does not support [perfect derives](https://smallcultfollowing.com/babysteps//blog/2022/04/12/implied-bounds-and-perfect-derive/) (due to some combination of action-at-a-distance hazard at the language design level, and trait-solver complication at the implementation level?). I found this out the hard way then had to add the manual derive below.

crates/wasmtime/src/runtime/gc/enabled/rooting.rs
cfallin: (Addressed above -- the derived `Clone` doesn't work because of language restrictions)

crates/wasmtime/src/runtime/gc/enabled/rooting.rs
cfallin: Updated description; thanks!

crates/wasmtime/src/runtime/gc/enabled/rooting.rs
cfallin: Fixed!

crates/wasmtime/src/runtime/gc/enabled/rooting.rs
cfallin: Fixed!

crates/wasmtime/src/runtime/gc/enabled/rooting.rs
cfallin: No real reason, other than wanting to avoid excessive trims at very small root-set sizes; but you're right, better to trust the exponential growth and stick with a standard factor.

cfallin:
> Updating C API bits I think is fine, but naively I'm surprised that this is still using a generational index. I thought that was only necessary due to the manual nature of `ManuallyRooted` and for something with a dtor like this we wouldn't need generational indices? I haven't fully thought this through but that feels redundant now.

This is a plumbing issue, mostly, I think -- the `GcRootIndex` carries the generation (this is true because of the LIFO case still, at least, and the type is shared between LIFO and owned), and we need to reconstruct the `GcRootIndex` when the raw bits from C are passed back in. We could in theory read them out from the slab, but we would need access to the store to do that (ah, a familiar theme in my recent work!), so it seemed simpler not to touch the way it works other than to add the liveness flag. Happy to hear other thoughts if you or Nick have any (or see this refactored further on followup).

> Considering the C API some more too, this'll want to go over documentation with a fine-tooth comb as well because we've probably documented "you should call unroot, but nothing bad happens if you don't so long as you free the store". Now that needs to be changed to "you must call unroot or host memory will leak". That's a pretty big departure in behavior from before, which while not bad we need to be sure is clearly communicated.

At least in `wasmtime/val.h`, we do have

> ```
>   * Anyref values are required to be explicitly unrooted via
>   * #wasmtime_anyref_unroot to enable them to be garbage-collected.
> ```

I suppose that we don't warn more explicitly than that, but it seems reasonably documented that the usual "free what you allocate" C API rules apply? Happy to iterate here as well in followup!

alexcrichton:
As a minor request, could this use `void*` instead of `usize` (in C and Rust) to carry the "this is a pointer" bit along?

> it seems reasonably documented that the usual "free what you allocate" C API rules apply?

My main worry about this is that it's the only allocated "value" in the C API (as a discriminant of `wasmtime_valunion_t`). All other values are scalars and were previously scalars too. I'm not sure how many users this would affect though in that I'm not sure anyone's using GC right now.

I  know I've had historical issues binding owned values to other languages like Python and Go in the past, but I don't recall the specifics.

cfallin:
@alexcrichton sure thing!

* See 20a929ba938a95ef4cee11ce9c6dc69932407969 for the pointer-rather-than-`usize` change; needs unsafe impls of Send/Sync on the wrapper struct in the C layer but this should be straightforwardly OK because it's an Arc (I think).
* See 8b3bf8fb1d540cd3838632925ac1ba8901c22d2a for some additional warnings in the header docs about the need to unroot explicitly.
* See the other commit loosening the size assert slightly because riscv32imac gotta be different wrt padding (sad).

Let me know what you think...

alexcrichton:
Oh, sorry, I'm just bemoaning the state of managed values. I don't really want to get this working with the C API in other languages but I'm not sure anyone else wants to either. I don't have a great alternative, however, so all I can point out is that it's objectively more dangerous to switch to a "real pointer" from an index because mistakes are segfaults rather than  panics. That doesn't mean we shouldn't do this, so that's sort of me just bemoaning that there's not really any motivation to fully flesh this out on the C side right now

cfallin:
@fitzgen a review of one more bit if you don't mind -- I just pushed a07c3b0ef819db3956c3a0c2030dcda6a507bf5a that switches to a different amortized algorithm for root trimming.

It doesn't grow the high-water mark exponentially; instead it sets it adaptively based on the last post-trim live-root set size. I realized that the former will still result in unbounded liveness-flag-array growth between GCs if one creates and drops many GC roots without mutating the GC graph itself. The new algorithm is "self-equilibriating" -- it provides a strong guarantee that the size is at most double the max true live-set size, and still has amortized constant-time root creation. I wrote a proof in the doc-comment to convince myself :-)

Let me know what you think!

fitzgen (approved):
Ah yes of course. This is actually what I was assuming we were going to do in the original issue discussion and then it fled my mind when I actually reviewed the code. Looks great! And with a proof as well!

cfallin:
(Added one more detail after staring at this again: should not have gated the high-water mark *update* on eager (GC) vs root-creation mode; only the early-out is gated on that.)

Issue #11342, opened 2025-07-29
wasip3: Closing a socket resource does not always close the socket


alexcrichton:
Currently there is liberal use of `Arc<...>` in the WASIp3 `wasi:sockets` implementation which is used for sharing sockets between tasks which are used to implement streams. One example is that a stream of clients can be extracted from a TCP listener, meaning that the underlying socket is now shared between the original TCP listener resource and the stream that is reading clients. The consequence of this implementation is that when the resource is closed it does not mean that the actual socket is itself closed. For example the OS-level port remains reserved until the task for accepted clients also exits. While the tasks will "promptly exit" there is currently no reliable mechanism to await this happening.

This leads to tests for REUSEADDR, for example, having a loop around re-binding a port and asserting it's successful. This test is being added in https://github.com/bytecodealliance/wasmtime/pull/11291.

rvolosatovs:
One thing that is currently missing in the implementation is the receive/listen task cancellation on socket drop, i.e. `receive` and `listen` should store the `AbortHandle` in the socket state and abort these tasks once the socket is dropped.

I spent some time on it just now, but I don't really see a nice way to integrate this with the `start_receive` etc. abstractions without making `TcpState` field public again, @alexcrichton could you handle this?

I've implemented this for filesystem though: https://github.com/bytecodealliance/wasmtime/pull/11406/commits/40415b52f79a33c376a2a875ffecb1d0a01c7903

I don't know if that would necessarily fix this issue, but it's something we should do anyway. For reference, here's the documentation of `wasi:http` for a similar construct: https://github.com/WebAssembly/wasi-http/blob/ad500ba30bdfd57e04bdabdcd0480111681bf017/wit-0.3.0-draft/types.wit#L420-L423

For reference, here's the in-progress diff, if it helps:
```diff
diff --git a/crates/wasi/src/p3/sockets/host/types/tcp.rs b/crates/wasi/src/p3/sockets/host/types/tcp.rs
index 66314d5372..a2bf1ff0c1 100644
--- a/crates/wasi/src/p3/sockets/host/types/tcp.rs
+++ b/crates/wasi/src/p3/sockets/host/types/tcp.rs
@@ -266,11 +266,12 @@ impl HostTcpSocketWithStore for WasiSockets {
                     let (result_tx, result_rx) = instance
                         .future(&mut view, || unreachable!())
                         .context("failed to create future")?;
-                    view.spawn(ReceiveTask {
[...]

PR #11515, opened 2025-08-22
Revamp component model stream/future host API (again)

dicej:
This changes the host APIs for dealing with futures and streams from a "rendezvous"-style API to a callback-oriented one.

Previously you would create e.g. a `StreamReader`/`StreamWriter` pair and call their `read` and `write` methods, respectively, and those methods would return `Future`s that resolved when the operation was matched with a corresponding `write` or `read` operation on the other end.

With the new API, you instead provide a `StreamProducer` trait implementation whe creating the stream, whose `produce` method will be called as soon as a read happens, giving the implementation a chance to respond immediately without making the reader wait for a rendezvous.  Likewise, you can match the read end of a stream to a `StreamConsumer` to respond immediately to writes.  This model should reduce scheduling overhead and make it easier to e.g. pipe items to/from `AsyncWrite`/`AsyncRead` or `Sink`/`Stream` implementations without needing to explicitly spawn background tasks.  In addition, the new API provides direct access to guest read and write buffers for `stream<u8>` operations, enabling zero-copy operations.

Other changes:

- I've removed the `HostTaskOutput`; we were using it to run extra code with access to the store after a host task completes, but we can do that more elegantly inside the future using `tls::get`.  This also allowed me to simplify `Instance::poll_until` a bit.

- I've removed the `watch_{reader,writer}` functionality; it's not needed now given that the runtime will automatically dispose of the producer or consumer when the other end of the stream or future is closed -- no need for embedder code to manage that.

- In order to make `UntypedWriteBuffer` `Send`, I had to wrap its raw pointer `buf` field in a `SendSyncPtr`.

- I've removed `{Future,Stream}Writer` entirely and moved `Instance::{future,stream}` to `{Future,Stream}Reader::new`, respectively.

- I've added a bounds check to the beginnings of `Instance::guest_read` and `Instance::guest_write` so that we need not do it later in `Guest{Source,Destination}::remaining`, meaning those functions can be infallible.

Note that I haven't updated `wasmtime-wasi` yet to match; that will happen in one or more follow-up commits.


crates/misc/component-async-tests/src/util.rs
```diff
@@ -21,3 +29,111 @@ pub async fn sleep(duration: std::time::Duration) {
...
+        } else {
+            self.closed = true;
+        }
+
+        Ok(self.state())
+    }
+
+    async fn when_ready(&mut self, _: &Accessor<D>) -> Result<StreamState> {
+        Ok(self.state())
+    }
```
alexcrichton: Technically this should call `.next()` and buffer up the result if one comes in, right?

crates/misc/component-async-tests/src/util.rs
```diff
@@ -21,3 +29,111 @@ pub async fn sleep(duration: std::time::Duration) {
...
+    ) -> Result<StreamState> {
+        let item = &mut None;
+        accessor.with(|access| source.read(access, item))?;
+        _ = self.tx.send(item.take().unwrap()).await;
+        Ok(self.state())
+    }
+
+    async fn when_ready(&mut self, _: &Accessor<D>) -> Result<StreamState> {
+        Ok(self.state())
+    }
```
alexcrichton: Technically this should use [`poll_ready`](https://docs.rs/futures/latest/futures/channel/mpsc/struct.Sender.html#method.poll_ready), right?

alexcrichton (commented):
Is it worth it to resolve the zero-length read/write TODOs in the code before landing? We don't have many users of that right now so it also seems ok to defer that too.

Otherwise though this all seems reasonable to me, although I'm mostly relying on tests. The interfaces we've talked about historically and I think are ok to land. I'm also happy to help out with the porting of wasmtime-wasi later this afternoon

dicej:
> Is it worth it to resolve the zero-length read/write TODOs in the code before landing? We don't have many users of that right now so it also seems ok to defer that too.

Yup, I'll do that.  Testing it will be mildly tedious, but might as well bite the bullet.

> Otherwise though this all seems reasonable to me, although I'm mostly relying on tests. The interfaces we've talked about historically and I think are ok to land. I'm also happy to help out with the porting of wasmtime-wasi later this afternoon

@rvolosatovs is planning to take a crack at updating `wasmtime-wasi` starting tomorrow, FYI.

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -355,200 +212,473 @@ pub(super) struct FlatAbi {
...
-    fn drop(&mut self) {
-        if let Some(writer) = &mut self.writer {
-            writer.close_with(&self.accessor)
-        }
-    }
+/// Represents a host-owned write end of a future.
+pub trait FutureProducer<D, T>: Send + 'static {
+    /// Handle a host- or guest-initiated read by producing a value.
+    fn produce(self, accessor: &Accessor<D>) -> impl Future<Output = Result<T>> + Send;
+}
```
rvolosatovs: It appears that we'll need to expose the second type parameter to `Accessor` in order to be able to use this API from *within* e.g. `wasmtime_wasi`.
Binding implementations get an opaque `T` passed in, e.g. https://github.com/bytecodealliance/wasmtime/blob/3aa392399e785df414ad6ee6ee7f034989b75301/crates/wasi/src/p3/sockets/host/types/tcp.rs#L252-L253
```suggestion
pub trait FutureProducer<T, U, V>: Send + 'static {
    /// Handle a host- or guest-initiated read by producing a value.
    fn produce(self, accessor: &Accessor<T, U>) -> impl Future<Output = Result<V>> + Send;
}
```

alexcrichton: Oh for that @dicej and I talked about this a bit ago, I'd prefer to avoid the type parameter explosion but I have an alternative solution that should work. Let me sketch that out and push it up here too

rvolosatovs: Awesome, thanks @alexcrichton .
I actually just ran into a very similar issue earlier today in `wasi:http`, trying to reuse the same task implementation both from outside and from within the `wasi:http` crate, which I worked around the following way:
```rust
impl<T, U, Fut> AccessorTask<T, U, wasmtime::Result<()>> for GuestBodyTask<Fut>
where
    T: WasiHttpView,
    U: HasData,
    Fut: Future<Output = Result<(), ErrorCode>> + Send + 'static,
{
    async fn run(self, store: &Accessor<T, U>) -> wasmtime::Result<()> {
        self.run(store, |mut store, trailers| {
            store
                .data_mut()
                .http()
                .table
                .delete(trailers)
                .context("failed to delete trailers")
        })
        .await
    }
}

/// This is a duplicate of [GuestBodyTask], which can be used from within this crate
pub(crate) struct GuestBodyTaskInternal<T>(GuestBodyTask<T>);

impl<T, U, Fut> AccessorTask<T, U, wasmtime::Result<()>> for GuestBodyTaskInternal<Fut>
where
    for<'a> U: HasData<Data<'a> = WasiHttpCtxView<'a>>,
    Fut: Future<Output = Result<(), ErrorCode>> + Send + 'static,
{
    async fn run(self, store: &Accessor<T, U>) -> wasmtime::Result<()> {
        self.0
            .run(store, |mut store, trailers| {
                store
                    .get()
                    .table
                    .delete(trailers)
                    .context("failed to delete trailers")
            })
            .await
    }
}
```

Perhaps this serves as a good data point for the fix

alexcrichton: Ok I pushed an extra commit up with what I was thinking, the rough idea is that the WASI implementations will close over the "getter", the `fn(&mut T) -> D::Data<'_>` projection, themselves. If that's onerous though to schlep with WASI we can look to bake it in.

rvolosatovs: Testing it now, looks like bindgen still needs to be updated for the rename 🤔

rvolosatovs: Unfortunately, I don't think that the refactor fixes the underlying issue, which is that the host interface implementations do not have access to the getter provided by the the call to `add_to_linker`. The getter provided by `Access<T>::getter` will always produce `&mut T`, because the second type parameter defaults to `HasSelf<T>`.

What this means is that for this to work, the getter would have to be passed to the implementation through `WasiSocketsCtxView<'a>`, for example.

- `Accessor::new` creates a getter, which is an identity function https://github.com/dicej/wasmtime/blob/61c5a24ebb6fac0e879077f47a43fea5a55b4af1/crates/wasmtime/src/runtime/component/concurrent.rs#L422, that's the only getter visible to `produce` implementations
- The actual getter implementations need though is only visible to the generated bindings https://github.com/dicej/wasmtime/blob/61c5a24ebb6fac0e879077f47a43fea5a55b4af1/crates/wit-bindgen/src/lib.rs#L2333 and it's later used via `with_getter` to map the 2nd type parameter to a different type https://github.com/dicej/wasmtime/blob/61c5a24ebb6fac0e879077f47a43fea5a55b4af1/crates/wit-bindgen/src/lib.rs#L2493-L2520

I'm happy to jump on a quick call to talk through this today to give more details.

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -355,200 +212,473 @@ pub(super) struct FlatAbi {
...
 
-impl<T> FutureWriter<T> {
-    fn new(default: fn() -> T, id: TableId<TransmitHandle>, instance: Instance) -> Self {
-        Self {
-            default,
-            id,
-            instance,
+impl<T> Destination<T> {
+    /// Deliver zero or more items to the reader.
+    pub async fn write<B, A: AsAccessor>(&mut self, accessor: A, mut buffer: B) -> Result<B>
```
rvolosatovs: I would have expected this function to not be `async`, in fact, shouldn't this function *move* `Destination`, such that it could only be called once with at most the number of elements that reader asked for?

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -355,200 +212,473 @@ pub(super) struct FlatAbi {
...
+/// Represents a guest read from a `stream<u8>`, providing direct access to the
+/// guest's buffer.
+pub struct GuestDestination<'a, D: 'static> {
+    instance: Instance,
+    id: TableId<TransmitState>,
+    store: StoreContextMut<'a, D>,
+}
 
-        let result = self
+impl<D: 'static> GuestDestination<'_, D> {
```
rvolosatovs: Since we're dealing with byte buffers only here, could we reuse https://docs.rs/tokio/latest/tokio/io/struct.ReadBuf.html directly in some way?

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
```diff
@@ -355,200 +212,473 @@ pub(super) struct FlatAbi {
...
+pub trait FutureConsumer<D, T>: Send + 'static {
+    /// Handle a host- or guest-initiated write by consuming a value.
+    fn consume(self, accessor: &Accessor<D>, value: T) -> impl Future<Output = Result<()>> + Send;
 }
 
 /// Represents the readable end of a Component Model `future`.
 ///
-/// Note that `FutureReader` instances must be disposed of using either `read`
+/// Note that `FutureReader` instances must be disposed of using either `pipe`
 /// or `close`; otherwise the in-store representation will leak and the writer
```
rvolosatovs, line 759: Once a `FutureReader` is returned by a host implementation, is it guaranteed that the runtime will call `close` if the guest does not use it? (or e.g. traps)

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
rvolosatovs: nvm, we have figured it out together with @dicej , thanks!

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
rvolosatovs: Resolved offline. Once returned to the guest the handle will be properly disposed off with a *well-behaving* guest. We probably want to introduce some limit on the amount of open future/stream handles that the guest can have

cc @dicej

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
rvolosatovs: Resolved offline, we may need to wait for realloc, which is why this needs to be `async`.
@dicej will add a `remaining` function to this struct

crates/test-programs/src/bin/p3_sockets_tcp_sample_application.rs
```diff
@@ -47,6 +47,7 @@ async fn test_tcp_sample_application(family: IpAddressFamily, bind_address: IpSo
 
             // Check that we sent and received our message!
             assert_eq!(data, first_message); // Not guaranteed to work but should work in practice.
+            drop(data_rx);
```
rvolosatovs: I think this is an important detail of the implementation to call out - in the absence of task, the stream handle returned in `tuple<stream<T>, future<result<_, E>>>` pattern has to be dropped/closed before a value can arrive on the `future`, I suspect that is because in case of a task, the host could independently track the status of the socket and be notified of the socket shutdown by the OS as part of it's own event loop. Without a task, the guest must effectively "signal" the host to "do work", in success case triggering the `Drop` of the producer, I suppose.

I wonder if that's the expected behavior here and whether that would play well with GC languages.

rvolosatovs: One solution could be giving both stream and future producers access to the underlying socket and check for shutdown in the future producer as well. It seems that such an approach would require the future producer to also perform reads from the socket and communicate those to stream producer.

dicej: > Giving it a bit more thought, the guest is getting `StreamResult::Complete` on L46, shouldn't the host eagerly drop the producer _before_ letting the guest know that the stream is done?

Yes, it should; if it's not, I'd consider that a bug.  I'll take a look.

rvolosatovs: I've originally mistaken `Complete` for `Dropped`, but realized that was wrong https://github.com/bytecodealliance/wit-bindgen/blob/f61634fd7905595ba7b83bb8cda42a4f3ee3ac29/crates/guest-rust/src/rt/async_support/stream_support.rs#L225-L235 and deleted the comment.

I think the test was simply incorrect, see [#wasmtime > async API revamp implications](https://bytecodealliance.zulipchat.com/#narrow/channel/217126-wasmtime/topic/async.20API.20revamp.20implications)

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
dicej: I just opened https://github.com/bytecodealliance/wasmtime/issues/11552 to track that.

crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs
dicej: Wasmtime doesn't currently depend on `tokio` except for testing; I'm not sure `ReadBuf` adds enough value to justify pulling it in.

crates/wasi/src/p3/sockets/host/types/tcp.rs
```diff
@@ -39,106 +40,256 @@ fn get_socket_mut<'a>(
...
+                    }
+                    self.stream.try_write(&self.buffer)?
+                };
+                debug_assert!(n > 0);
+                std::io::Result::Ok(Ok(n))
+            }) {
+                Ok(Ok(..)) if self.buffer.is_empty() => return Ok(StreamState::Open),
+                Ok(Ok(n)) => {
+                    let mut buf = &self.buffer[n..];
+                    while !buf.is_empty() {
```
rvolosatovs: It seems that the consumer must loop here to ensure that all of the data already read from the guest has been written - otherwise, if the consumer were to buffer, how would the guest know that a partial write occurred?

It seems that what we really want is to have a way for the consumer to report back the amount of elements actually read, replicating something like https://doc.rust-lang.org/nightly/std/io/trait.Read.html#tymethod.read

Given the existing API, it looks like the consumer should be able to *return* the buffer, just like the `Destination::write` would currently do.

This is, of course, already addressed for byte-buffers originating from the guest, but this seems problematic in general case

rvolosatovs:
All existing WASI tests pass now for me locally. I'm a bit surprised that it looks like CI does not run on this PR anymore?

I've removed the reuseaddr test workaround, since this change set should remove the need for it and close https://github.com/bytecodealliance/wasmtime/issues/11342

rvolosatovs:
The main remaining item left here is the cancellation-safety of `produce` functions, but we've been having some private conversations with @dicej and hopefully a pretty minor API change should suffice to address it.
@dicej please let me know when the new API is available and I will happily fix WASI again.
Could you also rebase this PR on latest `main`, please?

I will meanwhile work on migrating #11440 to this API

crates/wasi/src/p3/filesystem/host.rs
```diff
@@ -131,102 +161,249 @@ impl<T> AccessorTask<T, WasiFilesystem, wasmtime::Result<()>> for ReadFileTask {
...
+                bail!("failed to write entry")
+            }
+            return Ok(StreamState::Open);
+        };
+        self.close(res);
+        Ok(StreamState::Closed)
+    }
+
+    async fn when_ready(&mut self, _: &Accessor<D>) -> wasmtime::Result<StreamState> {
+        Ok(StreamState::Open)
```
rvolosatovs: Is "lying" like this OK, @dicej, or do we have to drive I/O and buffer the result?

crates/wasi/src/p3/filesystem/host.rs
```diff
@@ -131,102 +161,249 @@ impl<T> AccessorTask<T, WasiFilesystem, wasmtime::Result<()>> for ReadFileTask {
...
-        drop(self.dir);
-        drop(data_tx);
-        result_tx.write(res).await;
-        Ok(())
+        self.close(res);
+        Ok(StreamState::Closed)
+    }
+
+    async fn when_ready(&mut self, _: &Accessor<D>) -> wasmtime::Result<StreamState> {
+        Ok(StreamState::Open)
```
rvolosatovs: Same as above, do we need to drive I/O?

crates/wasi/src/p3/filesystem/host.rs
```diff
@@ -131,102 +161,249 @@ impl<T> AccessorTask<T, WasiFilesystem, wasmtime::Result<()>> for ReadFileTask {
...
+                }
+                Err(err) => break 'result Err(err),
+            }
+        };
+        self.close(res);
+        Ok(StreamState::Closed)
+    }
+
+    async fn when_ready(&mut self, _: &Accessor<D>) -> wasmtime::Result<StreamState> {
+        Ok(StreamState::Open)
```
rvolosatovs: same here

rvolosatovs:
On a second thought, I'm guessing that "lying" about the stream being open is probably not the expected behavior, so I'll work on driving I/O in these and buffering first

dicej:
> On a second thought, I'm guessing that "lying" about the stream being open is probably not the expected behavior, so I'll work on driving I/O in these and buffering first

This is a good reminder that we should add tests that involve zero-length reads and writes from the guest, asserting that the following non-zero-length read or write completes immediately.  So far, this hasn't really been relevant, but once we add p3 support to wasi-libc we'll have real-world code that relies on it.

rvolosatovs:
> > On a second thought, I'm guessing that "lying" about the stream being open is probably not the expected behavior, so I'll work on driving I/O in these and buffering first
> 
> This is a good reminder that we should add tests that involve zero-length reads and writes from the guest, asserting that the following non-zero-length read or write completes immediately. So far, this hasn't really been relevant, but once we add p3 support to wasi-libc we'll have real-world code that relies on it.

I was under impression that we wanted to avoid buffering, however buffering seems to be the only way of providing this behavior for writes to e.g. files. It looks like for filesystem we'll need to introduce a construct (similar to p2), where we'll have a thread performing file I/O and `wasi:filesystem` *write* just sends the buffer on something like `mpsc` channel.

Is that the expectation here?

I think that makes sense to me, but it appears we'd also need to introduce `sync`/`flush` or similar to `wasi:filesystem` to be able to ensure that the host worker thread buffer has actually been passed over to the kernel since `stream.write` will no longer guarantee that

rvolosatovs:
hmm, `sync` does exist https://github.com/WebAssembly/wasi-filesystem/blob/d81d6256c271fe1c8937eb8353e2ddc25517c153/wit-0.3.0-draft/types.wit#L434, so would we need to await the flush before returning?

crates/wasi/src/p3/sockets/host/types/tcp.rs
dicej: Yeah, I can see how that could be awkward -- you read items from the `Source` and then try to write them to some kind of sink, but that sink might not be immediately ready to accept all the items.  In that case do you just .await until the sink has accepted them all?  If you do, then you're kind of blocking the original writer more than you should -- you'd rather just say you didn't read all the items and let the writer write them again if and when it wants to.  Also, what if the sink closes before you can write them all?

So yeah, I think we might need to change this API as well as the `StreamProducer` one.

rvolosatovs:
for reference: [#wasi > 0-length stream writes with files and stdio](https://bytecodealliance.zulipchat.com/#narrow/channel/219900-wasi/topic/0-length.20stream.20writes.20with.20files.20and.20stdio)

sunfishcode:
> hmm, `sync` does exist https://github.com/WebAssembly/wasi-filesystem/blob/d81d6256c271fe1c8937eb8353e2ddc25517c153/wit-0.3.0-draft/types.wit#L434, so would we need to await the flush before returning?

There are two kinds of synchronizing here: one is that the data will be visible to independent readers, and the other is that the data will be visible to independent readers even after an abrupt power failure. wasil-filesystem's `sync` is for that second kind, which can be very slow, so we shouldn't use it unless the application has explicitly requested it.

crates/wasi/src/p3/cli/host.rs
```diff
@@ -1,155 +1,119 @@
...
-                    }
+        if let Some(mut dst) = dst.as_direct_destination(store) {
+            let mut buf = ReadBuf::new(dst.remaining());
+            match self.rx.as_mut().poll_read(cx, &mut buf) {
+                Poll::Ready(Ok(())) if buf.capacity() == 0 => {
+                    Poll::Ready(Ok(StreamResult::Completed))
                 }
-            } else {
-                let capacity = destination
-                    .remaining(store)
```
rvolosatovs: this would always return `None`, wouldn't it? (since that's a host read)

dicej: Good point; I didn't revisit that code after I added the `if/else` block around it.

rvolosatovs:
I'm signing off for today and have just finished the WASI crate adaptation a few minutes ago. I have not even really proof-read, for example, the `read_directory` implementation yet.
There's quite a bit of room for improvement, mainly refactoring and cutting down on duplication.
I also noticed that `read_directory` seems to not be tested at all.

That all said, all tests pass (at least locally, will have to wait for CI) and IMO this is "good enough" to get this PR merged. Feel free to clean-up as desired, otherwise I will do that tomorrow.

crates/wasi/src/p3/cli/host.rs
```diff
@@ -8,70 +8,110 @@ use crate::p3::bindings::cli::{
...
+                    Poll::Ready(Ok(())) if buf.filled().is_empty() => {
+                        return Poll::Ready(Ok(StreamResult::Dropped));
+                    }
+                    Poll::Ready(Ok(())) => {
+                        let n = buf.filled().len();
+                        dst.mark_written(n);
+                        return Poll::Ready(Ok(StreamResult::Completed));
+                    }
+                    Poll::Ready(Err(..)) => {
+                        // TODO: Report the error to the guest
```
alexcrichton, line 49: I opened https://github.com/WebAssembly/wasi-cli/issues/81 for this, but for now it might be good to `log::warn!` the error too

crates/wasi/src/p3/cli/host.rs
```diff
@@ -8,70 +8,110 @@ use crate::p3::bindings::cli::{
...
+                        return Poll::Ready(Ok(StreamResult::Dropped));
+                    }
+                    Poll::Pending if finish => return Poll::Ready(Ok(StreamResult::Cancelled)),
+                    Poll::Pending => return Poll::Pending,
                 }
             }
         }
-        Ok(())
+        let mut buf = dst.take_buffer().into_inner();
```
alexcrichton, line 57: Can we skip this fallback and rely on `as_direct_destination` always succeeding?

crates/wasi/src/p3/cli/host.rs
```diff
@@ -8,70 +8,110 @@ use crate::p3::bindings::cli::{
...
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        store: StoreContextMut<D>,
+        src: &mut Source<Self::Item>,
+        finish: bool,
+    ) -> Poll<wasmtime::Result<StreamResult>> {
+        let mut src = src.as_direct_source(store);
+        let buf = src.remaining();
+        match self.tx.as_mut().poll_write(cx, buf) {
+            Poll::Ready(Ok(n)) if buf.is_empty() => {
```
alexcrichton, line 100: How come this needs a special case vs the below case?

alexcrichton:
I talked with @dicej about this and we concluded that let's go ahead and land this. I've got follow-up feedback and Joel's got some follow-up implementation work but we feel it's best to land this and iterate rather than continuing to block this. The hope is that by landing this @rvolosatovs you're more-or-less unblocke to continue to work on wasi-http while we continue to smith some details here in parallel

alexcrichton (approved):

crates/wasi/src/p3/cli/host.rs
alexcrichton: Talked with Joel about this and he's going to refactor things so the fallback isn't necessary (e.g. `as_direct_destination` always returns non-`None`)

crates/wasi/src/p3/cli/host.rs
alexcrichton: (talked with Joel and I'll handle this in a follow-up)

crates/wasi/src/p3/filesystem/host.rs
```diff
@@ -108,125 +116,568 @@ fn systemtimespec_from(t: NewTimestamp) -> Result<Option<fs_set_times::SystemTim
...
+    type Buffer = Cursor<BytesMut>;
+
+    fn poll_produce<'a>(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        store: StoreContextMut<'a, D>,
+        dst: &'a mut Destination<'a, Self::Item, Self::Buffer>,
+        finish: bool,
+    ) -> Poll<wasmtime::Result<StreamResult>> {
+        if let Some(task) = self.task.as_mut() {
```
alexcrichton, line 151: I'm going to look into restructuring this function to hopefully not need duplication in this `match` here and the one far below after the spawn, but that's just refactoring a bit.

crates/wasi/src/p3/cli/host.rs
```diff
@@ -8,70 +8,110 @@ use crate::p3::bindings::cli::{
...
+                debug_assert_eq!(n, 0);
+                Poll::Ready(Ok(StreamResult::Completed))
             }
+            Poll::Ready(Ok(n)) => {
+                src.mark_read(n);
+                Poll::Ready(Ok(StreamResult::Completed))
```
alexcrichton, line 106: For stdio in particular I think we'll additionally want to call `poll_flush` here before returning `Completed` because otherwise tokio only puts this off on some other task and it hasn't actually made its way to stdout yet.

crates/wasi/src/p3/sockets/host/types/tcp.rs
```diff
@@ -39,106 +41,202 @@ fn get_socket_mut<'a>(
...
+    type Buffer = Option<Self::Item>;
+
+    fn poll_produce<'a>(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        mut store: StoreContextMut<'a, D>,
+        dst: &'a mut Destination<'a, Self::Item, Self::Buffer>,
+        finish: bool,
+    ) -> Poll<wasmtime::Result<StreamResult>> {
+        let res = match self.listener.poll_accept(cx) {
```
alexcrichton, line 68: This prompted discussion between Joel and I and led to https://github.com/WebAssembly/component-model/issues/561 but we'll need to carefully handle 0-length reads here

crates/wasi/src/p3/sockets/host/types/tcp.rs
```diff
@@ -39,106 +41,202 @@ fn get_socket_mut<'a>(
...
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        store: StoreContextMut<'a, D>,
+        dst: &'a mut Destination<'a, Self::Item, Self::Buffer>,
+        finish: bool,
+    ) -> Poll<wasmtime::Result<StreamResult>> {
+        let res = 'result: {
+            if let Some(mut dst) = dst.as_direct_destination(store) {
+                let buf = dst.remaining();
+                if !buf.is_empty() {
```
alexcrichton, line 121: Like writes below I think this'll want a special-case of 0-length buffers to do a `poll_read_ready` here

rvolosatovs: With a 0-length read this tries a buffered read and if that would block, it polls readiness.
From what I understand that's the expectation of this API - produce at least 1 (buffered) element.

Is there a reason to poll readiness before optimistically trying to read?

alexcrichton: True! I should clarify my comment more... I'm assuming that with @dicej's queued up changes the `None` case for `as_direct_destination` goes away and so I was assuming that all the non-`as_direct_destination` code was going to be deleted. In that world the 0-length case is missed here, but otherwise you're right the 0-length case is handled through the buffers otherwise. (although that [has its own problems](https://github.com/WebAssembly/component-model/issues/561))

PR #11517, opened 2025-08-23
Update to wasm-tools 0.237.0

badeend:
That version comes with support for [fallible resource constructors](https://github.com/WebAssembly/component-model/pull/550).

`wasmtime-wit-bindgen` doesn't need any additional changes to support this, though I've added a codegen test for it to be sure.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -625,7 +625,7 @@ dependencies = [
  "test-programs-artifacts",
  "tokio",
  "wasm-compose",
- "wasmparser 0.236.0",
+ "wasmparser 0.237.0",
  "wasmtime",
  "wasmtime-wasi",
 ]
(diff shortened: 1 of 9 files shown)


dicej:
@alexcrichton Do you happen to know why these updates aren't covered by the wildcard audits in `supply-chain/audits.toml`?

dicej:
> @alexcrichton Do you happen to know why these updates aren't covered by the wildcard audits in `supply-chain/audits.toml`?

`cargo vet` claims the publisher is `UNKNOWN`, which may be the reason.  Not sure why that would happen, though.

alexcrichton:
I think the problem is due to the fact of wasm-tools now using crates.io trusted publishing workflows. I've added audits in the meantime.

badeend:
There were connectivity issues during the CI run. Maybe try again?

PR #11521, opened 2025-08-25
[Cranelift] `(x - y) u> x --> y u> x`

bongjunj:
This adds an optimization: `(X - Y) u> X --> Y u> X`.
Same proof with crocus is used to verify the correctness. (See my previous PR #11359)


fitzgen (commented):
Looks great, thanks!

Would you mind also adding a runtest for this rule (and for other rules you add, going forward, please) so that we can test the behavior under the non-optimizing interpreted CLIF and make sure it matches the results of the optimized and compiled to native code version? Anywhere under `cranelift/filetests/filetests/runtests/*` would be fine.

Will merge once that test is added. Thanks again!

bongjunj:
Hi, thank you for the review.

I've just added runtests for the new opts.

fitzgen (approved):
Fantastic, thanks again @bongjunj!

PR #11532, opened 2025-08-25
Try to make CI more robust with more retries

alexcrichton:
Try adding `--retry` to a bunch of `curl` commands to see if we can make CI a bit more robust in the face of flaky network failures.

### DIFF
--- a/.github/actions/install-rust/action.yml
+++ b/.github/actions/install-rust/action.yml
@@ -17,6 +17,10 @@ runs:
     - name: Install Rust
       shell: bash
       id: select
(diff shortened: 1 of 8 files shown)


fitzgen (approved):

alexcrichton:
Ok switching to the curl backend for rustup seems to have done the trick. Maybe a CDN blanket banned the IP range for Github actions with the rustup user agent...

alexcrichton:
Nope this just got lucky in PR CI. Still showing issues in the queue

PR #11533, opened 2025-08-25
Pull in new regalloc2 with fastalloc fixes for exceptions, and re-enable and add to testing.

cfallin:
Thanks to @d-sonuga in bytecodealliance/regalloc2#233, the fastalloc (single-pass) register allocator now supports arbitrary numbers of non-register-constrained defs, which we use on call instructions after #10502. This should be sufficient to resolve the issues that had led us to disable fastalloc temporarily in #10554.

This PR reverts that disabling step; adds Cranelift-with-single-pass as a new compiler strategy (alongside Cranelift, Winch, and Cranelift/Pulley) to all of our wast/spec tests and fuzzing; and pulls in the new RA2.

Adding to our compiler strategy list is perhaps the most controversial thing here, but I believe it's probably appropriate to add the extra dimension here to ensure fastalloc stays working (it's a great feature to have!). Note that this surfaces as a new compiler strategy rather than a separate config dimension (like GC backend is currently) because it applies only to Cranelift, not Winch.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2995,9 +2995,9 @@ dependencies = [
 
 [[package]]
 name = "regalloc2"
-version = "0.12.2"
+version = "0.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5216b1837de2149f8bc8e6d5f88a9326b63b8c836ed58ce4a0a29ec736a59734"
+checksum = "68e18e1ef763167dc6718c28a5585e62f907590a21028b8e87be1318f19ef1cb"
 dependencies = [
  "allocator-api2",
  "bumpalo",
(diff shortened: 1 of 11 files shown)


cfallin:
Fair enough! Removed that commit, so now it's just a clean update of RA2 + revert of #10554.

cfallin:
The `cargo vet` CI job failed with a bad download (more Internet weather) and I'm not seeing an option to re-run it; CI itself also appears wedged in that it can't cancel and the "Record the result" job is queued but not starting.

alexcrichton:
The retry button doesn't show up until everything finishes, and CI is overflowing with jobs right now due to a [very large queue](https://github.com/bytecodealliance/wasmtime/queue/main) due to the network failures.

PR #11547, opened 2025-08-26
merge `concurrent::table::Table` back into `ResourceTable`

### LINKED ISSUE #11190: Deduplicate `resource_table.rs` and `concurrent/table.rs`
#11127 adds a `runtime/component/concurrent/table.rs` to the `wasmtime` crate, which is a duplicated and modified version of `runtime/component/resource_table.rs`.  We should deduplicate those files, e.g. by implementing one in terms of the other.

dicej:
When I started implementing component model async support, I copied `resource_table.rs` to `concurrent/table.rs` and made some modifications.  Now I'm unifying the implementations to reduce duplication.

- `concurrent/table.rs` now only contains the `TableId` type and the `TableDebug` trait; the `Table` type has been merged into `ResourceTable`.

- `TableId` can be thought of as a lightweight variation of `Resource` representing handles which can't be borrowed.  It also implements `Copy`, `Hash`, `Debug`, etc. for convenience, and can be converted to and from `Resource`.

- `ResourceTable` now has `enable_debug`, `is_empty`, `add_child`, `remove_child`, and `iter_mut` methods.

- `ConcurrentState::table` is now of type `Mutex<ResourceTable>`; the `Mutex` wrapper is needed to ensure the field is `Sync`, but we never lock it.

Fixes #11190


crates/wasmtime/src/runtime/component/concurrent.rs
```diff
@@ -3970,7 +3979,7 @@ pub struct ConcurrentState {
     /// out, poll it, then put it back to avoid any mutable aliasing hazards.
     futures: Mutex<Option<FuturesUnordered<HostTaskFuture>>>,
     /// The table of waitables, waitable sets, etc.
-    table: Table,
+    table: Mutex<ResourceTable>,
```
alexcrichton: Could this replace `Mutex` with `AlwaysMut` added in https://github.com/bytecodealliance/wasmtime/pull/11453?

crates/wasmtime/src/runtime/component/concurrent.rs
```diff
@@ -4095,36 +4104,45 @@ impl ConcurrentState {
...
+    fn push<V: Send + Sync + 'static>(
+        &mut self,
+        value: V,
+    ) -> Result<TableId<V>, ResourceTableError> {
+        self.table.get_mut().unwrap().push(value).map(TableId::from)
     }
 
-    fn get<V: 'static>(&self, id: TableId<V>) -> Result<&V, TableError> {
-        self.table.get(id)
+    fn get<V: 'static>(&mut self, id: TableId<V>) -> Result<&V, ResourceTableError> {
```
alexcrichton: I'd probably say go ahead and remove this and only have `get_mut` for now

crates/wasmtime/src/runtime/component/resource_table.rs
```diff
@@ -100,18 +102,53 @@ impl TableEntry {
...
             entries: Vec::new(),
             free_head: None,
-        }
+            debug: false,
+        };
+
+        // Reserve 0 as an invalid entry.  This effectively reserves
+        // `Resource::new_own(0)` as a sentinal value meaning "this resource
+        // doesn't exist anymore".
+        me.push(Tombstone).unwrap();
```
alexcrichton: While this is true of indices we hand out to wasm itself, I think this can be removed now? I think that this was required before merging the wasm-facing tables together but nowadays I think this should be unnecessary

crates/wasmtime/src/runtime/component/resource_table.rs
```diff
@@ -32,10 +33,10 @@ impl fmt::Display for ResourceTableError {
 impl core::error::Error for ResourceTableError {}
 
 /// The `ResourceTable` type maps a `Resource<T>` to its `T`.
-#[derive(Debug)]
 pub struct ResourceTable {
     entries: Vec<Entry>,
     free_head: Option<usize>,
+    debug: bool,
```
alexcrichton: Have you used per-table debugging in the past? If not how about moving this to a `const` in this module? That way it can still be easily flipped but won't take up space at runtime

alexcrichton (approved):

crates/wasmtime/src/runtime/component/resource_table.rs
dicej: You're right that this was originally a hack because I was handing these handles directly to guests in an early draft.  Once I stopped doing that, I removed the hack, but later I added it back in -- not because we were handing these handles to guest, but because I found it useful e.g. [here](https://github.com/bytecodealliance/wasmtime/blob/1f6c3aadaa3c018e379310513d70eaebdd6571aa/crates/wasmtime/src/runtime/component/concurrent/futures_and_streams.rs#L680-L681).

In other words, it's not strictly necessary, but it could help us catch our own mistakes more easily.

alexcrichton: Could `u32::MAX` be used there instead? It's effectively a nice property of container-like data structures that creation doesn't require allocation, so that's what I'd like to shoot for here

crates/wasmtime/src/runtime/component/resource_table.rs
dicej: I'm using it in some parts of `concurrent-async-tests` and did it this way because I didn't want a static or const to bleed over into other tests where it wasn't meant to be enabled, but I guess it's not that important.  Maybe I'll just change the `self.debug` checks to `cfg!(debug_assertions)`.

alexcrichton: That seems reasonable to me as well yeah 👍

Issue #8963, opened 2024-07-16
How to get control over filesystem access with `wasmtime_wasi::WasiCtxBuilder`


stevenj:
Our implementation requires filesystems to be fully controlled, and files come from a VFS not any mounted filesystem.
That means we need to control the low level aspects of filesystem access for WasiP2.

I don't see how that can be achieved, because the only option I see is giving Wasi access to a pre-opened directory, which is 100% off the table for our application.  How to use the WaiCtxBuilder, and abstract the necessary filesystem interfaces to the runtime?

alexcrichton:
For a fully virtual filesystem you won't be using `WasiCtxBuilder`. You'll want to instead implement the `Host` traits directly generated in the `wasmtime-wasi` crate such as [this](https://docs.rs/wasmtime-wasi/latest/wasmtime_wasi/bindings/sync/filesystem/types/trait.Host.html). You'd then call [`add_to_linker`](https://docs.rs/wasmtime-wasi/latest/wasmtime_wasi/bindings/filesystem/types/fn.add_to_linker.html) or similarly.

pchickey:
We used to have a more dynamic trait system to make this pluggable when wasmtime-wasi was backed by wasi-common, and when we rewrote wasmtime-wasi to detach it from the legacy wasi-common implementation, we got rid of it because it created significant maintenance burden and complexity for very little benefit over using the linker to do the same.

You can still use the linker to define the filesystem interfaces, and use the rest of wasmtime-wasi's linkers to define the rest of wasi based on WasiCtx, and they will interoperate fine as long as your wasmtime::component::bindgen to generate your filesystem traits has the same `with` settings for the wasi:io package https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasi/src/bindings.rs#L412-L418

Timmmm:
Ah ok thanks. Sounds a little bit beyond me tbh but I'll have a go. So I wonder get multiple definition linker errors or something like that?

pchickey:
Symbols will be duplicated if you call the `wasmtime_wasi::add_to_linker_*` functions directly, because those put all symbols into the linker. If you drop down to using the individual `wasmtime_wasi::bindings::<package>::<interface>::add_to_linker_get_host` (see https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasi/src/lib.rs#L356-L385 for how these are collected into the whole set for add_to_linker_async), you can add symbols to the linker on a per-interface basis.

[...]

PR #11549, opened 2025-08-27
Make `ReaddirIterator` public

Timmmm:
This is required because it is used in the generated `wasmtime_wasi::p2::bindings::async_io::wasi::filesystem::types::HostDescriptor` trait for `read_directory()`. Since the type is currently private, third parties can't implement custom versions of that trait even if they don't actually use the `ReaddirIterator` from wasmtime.

See #8963


rvolosatovs (approved):

Timmmm:
Thanks for the quick merge!

PR #11550, opened 2025-08-27
WASI preview 0,1 set_rights API returns ENOTSUP

wingo:
Instead of silently succeeding, return an error.

See https://github.com/WebAssembly/wasi-testsuite/issues/101#issuecomment-3219290417.


pchickey:
We should pull in an updated wasi-testsuite to wasmtime at some point soonish, but I don't think we need to do so as part of this PR.

PR #11553, opened 2025-08-27
Disallow loading code on `x86_64-unknown-none`

### LINKED ISSUE #11506: Compiling wasmtime for the embedding with a target of `x86_64-unknown-none` doesn't match Cranelift compilation
### Test Case

a module/component with a simple c function that will trigger this: 

```
int RoundToNearestInt()
{
    float c = 1.331* 24.0;
    float r = lrintf(c);
    printf("rounded answer: %f\n", r); // should print 32
    return r;
}
```

### Steps to Reproduce

1. Build wasmtime for an emedding using the `x86_64-unknown-none`.  Similar to how https://github.com/bytecodealliance/wasmtime/tree/release-36.0.0/examples/min-platform does it.  
2.Then pre-compile module/component  https://github.com/bytecodealliance/wasmtime/blob/ebce5d453464d3b5fcc6f9391a9b21fd6307844d/examples/min-platform/src/main.rs#L66-L82

run the wasm

### Expected Results

rounding is completed properly

### Actual Results

incorrect result. Example above prints 31

### Versions and Environment

Wasmtime version or commit: was using 34+

Operating system: Linux

Architecture: x86

### Extra Info

I wasn't able to reproduce this with the `min-platform` example.  I believe this is because its [actually compiling to a platform Linux platform](https://github.com/bytecodealliance/wasmtime/blob/ebce5d453464d3b5fcc6f9391a9b21fd6307844d/examples/min-platform/build.sh#L51-L58)?

alexcrichton:
... And then also add an escape hatch to allow loading code. This commit is the culmination of discussion on #11506 with a proposed resolution for Wasmtime. The resolution being:

* Wasmtime will reject loading code on `x86_64-unknown-none` platforms by default.
* A new `Config::x86_float_abi_ok` escape hatch is added to bypass this check.
* Documentation/errors are updated around `x86_float_abi_ok` to document the situation.
* The `min-platform` example is updated to showcase how this is valid to run in that particular embedding (aka enable more features and sufficiently detect said features).

The basic tl;dr; is that we can't detect in stable Rust what float ABI is being used so therefore we pessimistically assume that `x86_64-unknown-none` is using a soft-float ABI. This is incompatible with libcalls unless they aren't actually called which is only possible when sufficiently many target features are enabled.

The goal of this commit is to be a relatively low-effort way to place a roadblock in the way of "ok ABIs are weird" but at the same time enable getting around the roadblock easily. Additionally the roadblock points to documentation about itself to learn more about what's going on here.

Closes #11506

### DIFF
--- a/crates/wasmtime/src/config.rs
+++ b/crates/wasmtime/src/config.rs
@@ -163,6 +163,7 @@ pub struct Config {
     pub(crate) coredump_on_trap: bool,
     pub(crate) macos_use_mach_ports: bool,
     pub(crate) detect_host_feature: Option<fn(&str) -> Option<bool>>,
+    pub(crate) x86_float_abi_ok: Option<bool>,
 }
 
 /// User-provided configuration for the compiler.
@@ -271,6 +272,7 @@ impl Config {
             detect_host_feature: Some(detect_host_feature),
             #[cfg(not(feature = "std"))]
             detect_host_feature: None,
+            x86_float_abi_ok: None,
         };
         #[cfg(any(feature = "cranelift", feature = "winch"))]
         {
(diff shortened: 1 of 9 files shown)


alexcrichton:
For posterity @fitzgen I tried to get `min-platform` working on aarch64-unknown-none but there's relocation errors about compiling `libembedding.a` (Rust code for `aarch64-unknown-none`) into a Linux shared library. I don't know how to resolve the errors without first going off and memorizing a bunch of relocation/shared-object things about aarch64...

examples/min-platform/src/main.rs
```diff
@@ -76,6 +76,19 @@ fn main() -> Result<()> {
...
+    // For x86_64 targets be sure to enable relevant CPU features to avoid
+    // float-related libcalls which is required for the `x86_64-unknown-none`
+    // target.
+    if cfg!(target_arch = "x86_64") {
+        unsafe {
+            config.cranelift_flag_enable("has_sse3");
+            config.cranelift_flag_enable("has_ssse3");
+            config.cranelift_flag_enable("has_sse41");
+            config.cranelift_flag_enable("has_sse42");
+            config.cranelift_flag_enable("has_fma");
```
cfallin, line 93: Looking this up to verify -- this has existed since 2013 (Haswell) on Intel chips and 2012 (Piledriver) on AMD chips so it's probably fine for anyone wanting to actually run the example on a semi-modern dev machine or server. Perhaps we want to put a comment here ("Note: these features should be present on Intel Haswell (2013) / AMD Piledriver (2012) or later; make sure to run this example on such a platform.") next to the config?

crates/wasmtime/src/config.rs
```diff
@@ -2701,6 +2703,76 @@ impl Config {
     pub fn gc_support(&mut self, enable: bool) -> &mut Self {
         self.wasm_feature(WasmFeatures::GC_TYPES, enable)
     }
+
+    /// Explicitly indicate or not whether the host is using a hardware float
+    /// ABI on x86 targets.
+    ///
+    /// This configuration option is only applicable on the
+    /// `x86_64-unknown-none` Rust target and has no affect on other host
```
cfallin: s/affect/effect/

It does likely have an *affect*, namely that it's mildly sad that it has to exist...

(with apologies for the spelling pedantry!)

cfallin (approved):
Thanks for following up from our Cranelift meeting discussion! A minor documentation thought below in the example but otherwise this looks workable for us at least...

examples/min-platform/embedding/src/lib.rs
```diff
@@ -52,15 +55,55 @@ fn run_result(
...
+        // bound through an external crate.
+        unsafe {
+            config.detect_host_feature(move |feature| {
+                let id = raw_cpuid::CpuId::new();
+                match feature {
+                    "sse3" => Some(id.get_feature_info()?.has_sse3()),
+                    "ssse3" => Some(id.get_feature_info()?.has_sse3()),
+                    "sse4.1" => Some(id.get_feature_info()?.has_sse41()),
+                    "sse4.2" => Some(id.get_feature_info()?.has_sse42()),
+                    "fma" => Some(id.get_feature_info()?.has_fma()),
```
jsturtevant, line 98: All of these features are required to bypass the `libcalls` or just a subset?

alexcrichton: SE4.1 is sufficient if you're not using relaxed-simd, but with relaxed-simd you'll also need FMA.

PR #11558, opened 2025-08-28
Call `fixup` before `to_wasm_binary` in `TableOps` fuzzer

fitzgen:
Rather than just after particular mutations. Because mutating the serialized bytes can also result in invalid and out-of-bounds `TableOps` instances. This just does the fixup in the one place, instead of needing to remember all the different places where we might need to do it.

cc @khagankhan


fitzgen:
Also a few random code clean ups: pulling the fixing up of limits into `TableOpsLimits::fixup`, etc...

alexcrichton (approved):

PR #11560, opened 2025-08-28
Update and refactor how wasi-testsuite is run

alexcrichton:
This commit updates the `WebAssembly/wasi-testuite` submodule in this repository which hasn't been updated once in the past 2 years. This then additionally refactors how tests are run:

* A new top-level test is added at `tests/wasi.rs` similar to `tests/wast.rs`.
* The top-level test uses `libtest_mimic` to enable running tests in parallel.
* Using `libtest_mimic` enables running a test-at-a-time.
* Using `libtest_mimic` enables seeing a list of failures instead of just the first failure.
* Component versions of tests are run in addition to core wasm versions of tests.
* Minor updates are made to the adapter and wasmtime-wasi to get some tests passing.
* The list of allowed failures is updated with a `FIXME`-per-test of what's remaining.

prtest:full


alexcrichton:
This test suite hasn't been running on Windows so this is a draft until I get CI passing.

alexcrichton:
Ok all good now (just needed a huge allow-list of failures on Windows...)

crates/test-programs/src/bin/preview1_stdio.rs
```diff
@@ -5,7 +5,8 @@ use test_programs::preview1::{STDERR_FD, STDIN_FD, STDOUT_FD};
 unsafe fn test_stdio() {
     for fd in &[STDIN_FD, STDOUT_FD, STDERR_FD] {
         wasip1::fd_fdstat_get(*fd).expect("fd_fdstat_get on stdio");
-        wasip1::fd_renumber(*fd, *fd + 100).expect("renumbering stdio");
+        assert_eq!(wasip1::fd_renumber(*fd, *fd + 100), Err(wasip1::ERRNO_BADF));
```
abrown, line 8: I don't know much about this implementation, but shouldn't this succeed as it did previously? Doesn't the logic in [`descriptor.rs`](https://github.com/bytecodealliance/wasmtime/pull/11560/files#diff-dbbe655bc4c663438bc5b239e2fc7f27ae063d649a74078701d1446f0f8aa654R366) below expand the table so this _should_ work?

abrown (approved):
Thanks for moving and improving the old test. My only question has to do with the change to the WASI renumbering semantics.

crates/test-programs/src/bin/preview1_stdio.rs
alexcrichton: Part of the change here is to explicitly reject `fd_renumber` into an invalid file descriptor (aka one un-opened like `*fd + 100` here). The wasi-testsuite upstream specifically asserts that's a failure, where previously we explicitly supported that.

abrown: Ok, so perhaps we no longer should expand the table? Probably someone more familiar with this should have an opinion here... Feel free to merge if I'm nitpicking here.

alexcrichton: This is something we can eventually clean up in the future but the extend-the-table code shouldn't ever get hit any more because both to/from are guaranteed to be allocated. In that sense it's just dynamically dead code right now but otherwise harmless

PR #11526, opened 2025-08-25
[Cranelift] `(C ? X : Y) + Z --> (C ? X + Z : Y + Z)`

bongjunj:
This adds a new optimization `(C ? X : Y) + Z --> (C ? X + Z : Y +Z)`.
Same proof with crocus is used to verify the correctness. (See my previous PR https://github.com/bytecodealliance/wasmtime/pull/11359)

PR #11562, opened 2025-08-28
Fix a verifier error generated during optimizations

alexcrichton:
This commit fixes an accidental regression from #11526 detected on OSS-Fuzz. The `imm64` constants created needed to be masked off to the type width to avoid having the upper bits set.


cfallin:
cc @mmcloughlin @avanhatt -- this one is kind of interesting: the initial contribution had a Crocus proof but because I guess the prelude used there wasn't modeling the upper-bits-zeroed constraint on `iconst`s, this slipped through. More fuel on the fire of modeling IR constraints precisely!

PR #11564, opened 2025-08-28
Exceptions: global.get from external or table init: null exnref is not an anyref.

### LINKED ISSUE #11561: Panic instantiating table with exnref initializer via global.get
This input:

```wasm
(module
  (table $t 10 exnref)
  (global $g exnref (ref.null exn))
  (elem (table $t) (i32.const 0) exnref (global.get $g))
)
```

fails with:

```
$ cargo run -- -Wexceptions,gc,function-references testcase0.wat
...
thread 'main' panicked at crates/wasmtime/src/runtime/vm/instance.rs:971:72:
called `Result::unwrap()` on an `Err` value: type mismatch: value does not match table element type

Caused by:
    type mismatch: expected (ref null exn), found (ref null none)
```

cc @cfallin

### LINKED ISSUE #11563: Failure to instantiate a table-with-exnref-from-global
This input:

```wasm
(module
  (import "" "" (global exnref))
  (table 1 exnref (global.get 0))
)
```

is valid:

```
$ wasm-tools validate testcase0.wat
```

but cannot be instantiated:

```
$ cargo run -- -Wexceptions,function-references -W unknown-imports-default testcase0.wasm
Error: failed to run main module `testcase0.wasm`

Caused by:
    0: failed to instantiate "testcase0.wasm"
    1: type mismatch: value does not match table element type
    2: type mismatch: expected (ref null exn), found (ref null none)
```

Possibly related to https://github.com/bytecodealliance/wasmtime/issues/11561, but also feels like it may be distinct? Unsure.

cc @cfallin

cfallin:
Previously, the `Global::get()` implementation lumped exnref types in with other anyrefs and wrapped the raw GC ref in an anyref type. This isn't appropriate for exnrefs as they are not anyrefs.

Found via OSS-Fuzz.

Fixes #11561.
Fixes #11563.

### DIFF
--- a/crates/wasmtime/src/runtime/externals/global.rs
+++ b/crates/wasmtime/src/runtime/externals/global.rs
@@ -1,8 +1,8 @@
 use crate::prelude::*;
 use crate::runtime::vm::{self, VMGlobalDefinition, VMGlobalKind, VMOpaqueContext};
 use crate::{
-    AnyRef, AsContext, AsContextMut, ExternRef, Func, GlobalType, HeapType, Mutability, Ref,
-    RootedGcRefImpl, Val, ValType,
+    AnyRef, AsContext, AsContextMut, ExnRef, ExternRef, Func, GlobalType, HeapType, Mutability,
+    Ref, RootedGcRefImpl, Val, ValType,
     store::{AutoAssertNoGc, InstanceId, StoreId, StoreInstanceId, StoreOpaque},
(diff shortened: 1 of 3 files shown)


cfallin:
Added test from [here](https://github.com/bytecodealliance/wasmtime/issues/11563#issuecomment-3235246006) as this also fixes #11563.

PR #11566, opened 2025-08-29
Simplify Miri-compatible sleep in tests

alexcrichton:
This test occasionally gets stuck infinitely on CI and I think it's due to a race with the manual state management so this commit simplifies the sleep implementation to use more standard primitives.


alexcrichton:
example timeout is https://github.com/bytecodealliance/wasmtime/actions/runs/17310409135/job/49143235969

pchickey (approved):

PR #11569, opened 2025-08-29
[docs] Add C++ chapter and missing examples

tschneidereit:
This adds a chapter for C++ embedding. The chapter intro is rudimentary, but there are a bunch more examples now: I added C++ ports for everything that already had both a Rust and a C version available, matching the Rust version's behavior where applicable.

I also, in separate commits, added docs for some additional Rust and C examples, fixed small issues, and removed some outdated information from the README.


examples/wasip1/main.cc
```diff
@@ -0,0 +1,72 @@
...
+  if (!file.is_open()) {
+    throw std::runtime_error(std::string("failed to open wasm file: ") + path);
+  }
+  std::vector<uint8_t> data((std::istreambuf_iterator<char>(file)),
+                            std::istreambuf_iterator<char>());
+  return data;
+}
+
+int main() {
+  try {
```
alexcrichton: C++ stuff in theory shouldn't use exceptions? Would it be possible to remove these bits?

docs/examples-c-embed.md
fitzgen, line 1: Was this already unreachable? In general, deleting docs files is risky since it will create 404s for anything that previously linked to it.

docs/SUMMARY.md
```diff
@@ -11,19 +11,43 @@
...
+    - [Using Linear Memory](./examples-cpp-memory.md)
+    - [WASIp1](./examples-cpp-wasi.md)
+    - [Linking Modules](./examples-cpp-linking.md)
+    - [Using Multi-Value](./examples-cpp-multi-value.md)
+    - [`anyref` Reference Type](./examples-cpp-anyref.md)
+    - [`externref` Reference Type](./examples-cpp-externref.md)
+    - [Multiple Memories](./examples-cpp-multimemory.md)
+    - [Serializing Modules](./examples-cpp-serialize.md)
+    - [Wasm Threads](./examples-cpp-threads.md)
+    - [Async Host Functions](./examples-cpp-async.md)
```
fitzgen: Maybe put C above C++ since it is a bit more of a lingua franca?

docs/examples-c-serialize.md
```diff
@@ -0,0 +1,14 @@
+# Serializing and Deserializing Modules
+
+You can also [browse this source code online][code] and clone the wasmtime
+repository to run the example locally.
+
+[code]: https://github.com/bytecodealliance/wasmtime/blob/main/examples/serialize.c
+
+This example shows how to serialize a compiled module to disk and later deserialize it to skip compilation.
```
fitzgen:
```suggestion
This example shows how to serialize a compiled module to disk and later deserialize it to skip compilation on the critical path.
```

or something?

docs/examples-cpp-serialize.md
```diff
@@ -0,0 +1,14 @@
+# Serializing and Deserializing Modules
+
+You can also [browse this source code online][code] and clone the wasmtime
+repository to run the example locally.
+
+[code]: https://github.com/bytecodealliance/wasmtime/blob/main/examples/serialize.cc
+
+This example shows how to serialize a compiled module to disk and later deserialize it to skip compilation.
```
fitzgen: ditto

docs/examples-rust-serialize.md
```diff
@@ -0,0 +1,14 @@
+# Serializing and Deserializing Modules
+
+You can also [browse this source code online][code] and clone the wasmtime
+repository to run the example locally.
+
+[code]: https://github.com/bytecodealliance/wasmtime/blob/main/examples/serialize.rs
+
+This example shows how to compile a module once and serialize its compiled representation to disk and later deserialize it to skip compilation. See also the pre-compilation example for ahead-of-time compilation.
```
fitzgen: ditto

and also the "see also" bit should be a link probably

docs/examples-rust-threads.md
fitzgen, line 1: All these `examples-*-threads.md` files should probably be called `examples-*-threaded-embeddings.md` so that we can use the former for showing off Wasm threads (or even if we don't use exactly the former, the two won't be ambiguous with each other).

docs/examples-rust-wasip2.md
fitzgen, line 1: This is going to break any existing links that people have to to these docs. Can we avoid the rename? A policy we could take: `examples-*-wasi.md` always describes the latest wasi release, and we can add `examples-*-legacy-wasi.md` with instructions for p1, and p2 when p3 comes out, etc.

README.md
```diff
@@ -139,8 +139,8 @@ the implementation.
 Languages supported by the Bytecode Alliance:
 
 * **[Rust]** - the [`wasmtime` crate]
-* **[C]** - the [`wasm.h`, `wasi.h`, and `wasmtime.h` headers][c-headers], [CMake](crates/c-api/CMakeLists.txt) or [`wasmtime` Conan package]
-* **C++** - the [`wasmtime.hh` header][c-headers] or the [`wasmtime-cpp` Conan package]
+* **[C]** - the [`wasm.h`, `wasi.h`, and `wasmtime.h` headers][c-headers]
+* **C++** - the [`wasmtime.hh` header][c-headers]
```
fitzgen: What is the motivation behind removing the cmake and conan stuff? I have no idea if it is worth listing those things, as I do very little C/C++ stuff, but it seems like something we should change with intention.

docs/examples-rust-wasip2.md
tschneidereit: I like that as a policy, yes. Will change things accordingly

docs/SUMMARY.md
tschneidereit: My hunch was that more people would want to use the C++ bindings, because they're much nicer than the C ones. But I don't feel strongly about this, and am entirely happy with moving this down.

README.md
tschneidereit: I explained this in a pretty well-hidden place: the commit message for the specific patch removing this: the Conan packages are heavily outdated, and I don't think we have a way to update them. Though maybe we should try to acquire such a way instead of just removing the links? Not sure how we'd go about that, nor whether it's worth the effort to set up automation for this and such.

(Removing the `CMake` link was an accident, thanks for catching it!)

tschneidereit:
> FWIW I pushed up a few small commits to help with some warnings/etc, mostly just enough to get CI passing

Thank you for that! I tried too hard to make the behavior exactly match the Rust example's, with no real need to do so.

docs/examples-c-embed.md
tschneidereit: It was already unreachable, yes. Given that it was also empty except for `# C`, removing it seems prudent :)

docs/examples-rust-wasip2.md
tschneidereit: Thinking about this more, I'm a bit less sure. We could also use mdbook's [redirect support](https://rust-lang.github.io/mdBook/format/configuration/renderers.html#outputhtmlredirect) to avoid breaking links.

Alternatives could be to either have WASI have sub-chapters for p{1,2,3,N}, or organize them into sections on a single page: I think it'll be quite a while (until well after WASI 1.0) that people will stop wanting to have embeddings for some of these versions.

redradist:
@tschneidereit @alexcrichton Why mention of C and C++ Conan packages was removed ?

alexcrichton:
There was some discussion of that [here](https://github.com/bytecodealliance/wasmtime/pull/11569#discussion_r2311268468), but it looks like they're outdated?

tschneidereit:
@redradist indeed, the packages are heavily outdated. My apologies for not pinging you about this and instead just removing them. I had meant to reach out and then unfortunately forgot :(

I think for us to have these links in our official documentation, we'd want to be able to ensure that they're always up-to-date. That probably would mean we'd have automation for pushing new versions from our CI setup. If that's something you'd be interested in helping with, that'd be fantastic!

PR #11570, opened 2025-08-29
aarch64: Add support for "near" in `LoadExtName`

alexcrichton:
The current aarch64 backend does not support `symbol_value` to get the value of a function, for example, with a "near" relocation using a relative relocation. Currently it uses an `Abs8` relocation which means that it's not suitable in Wasmtime, for example.

This commit refactors relocation/external name support in the aarch64 backend to support this mode of relocation. The previous `LoadExtName` was split into `LoadExtName{Got,Near,Far}` where the "near" bit is what's new to the backend. The preexisting `symbol-value.clif`-style tests were updated to match the x64 backend which has a more comprehensive suite of examples of what it looks like to refer to various symbols.

The goal of this commit is to enable Wasmtime to generate code which refers to a relative point elsewhere in the code (e.g. an exception handler) and load the value into a register. This part isn't filled out yet, but it seemed good to at least in the meantime fill out these missing relocations in the backend.

### DIFF
--- a/cranelift/filetests/filetests/isa/aarch64/symbol-value-pic.clif
+++ b/cranelift/filetests/filetests/isa/aarch64/symbol-value-pic.clif
@@ -3,22 +3,185 @@ set unwind_info=false
 set is_pic
 target aarch64
 
-function %f() -> i64 {
-  gv0 = symbol %my_global
+function %func_addr() -> i64 {
+    fn0 = %func0(i64) -> i64
 
 block0:
-  v0 = symbol_value.i64 gv0
-  return v0
+    v0 = func_addr.i64 fn0
+    return v0
 }
 
 ; VCode:
+;   stp fp, lr, [sp, #-16]!
+;   mov fp, sp
 ; block0:
-;   load_ext_name x0, TestCase(%my_global)+0
+;   load_ext_name_got x0, TestCase(%func0)
+;   ldp fp, lr, [sp], #16
 ;   ret
 ;
 ; Disassembled:
 ; block0: ; offset 0x0
-;   adrp x0, #0 ; reloc_external Aarch64AdrGotPage21 %my_global 0
-;   ldr x0, [x0] ; reloc_external Aarch64AdrGotLo12Nc %my_global 0
+;   stp x29, x30, [sp, #-0x10]!
+;   mov x29, sp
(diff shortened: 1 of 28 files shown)


cranelift/filetests/filetests/isa/aarch64/symbol-value-pic.clif
```diff
@@ -3,22 +3,185 @@ set unwind_info=false
...
 block0:
-  v0 = symbol_value.i64 gv0
-  return v0
+    v0 = func_addr.i64 fn0
+    return v0
 }
 
 ; VCode:
+;   stp fp, lr, [sp, #-16]!
+;   mov fp, sp
```
cfallin, line 16: Sort of curious that the leaf-function optimization is no longer happening here, and we're getting a frame now -- won't matter for Wasmtime since we always force frames, but is there a reason you're aware of that this is happening?

cfallin: On looking a bit further, it seems that [`Function::is_leaf`](https://github.com/bytecodealliance/wasmtime/blob/0744262f79e235b172331f5ba1a2c43b5aedcbfb/cranelift/codegen/src/ir/function.rs#L325) only looks to see if there are signatures -- so `func_addr` triggers the "not a leaf function" mode and forces the frame.

Mind filing an issue that we should probably determine leaf-ness by scanning VCode instead for any instructions that claim to be calls? It should be machine-dependent anyway since libcalls can happen even in functions without IR-level calls. (I suppose that `is_leaf` and the no-frame ABI optimization isn't even right, in that case -- but we get away with it because the aarch64 backend doesn't fall back to any libcalls for FP stuff, and also because this configuration isn't exposed to Wasmtime?)

cfallin (approved):
Thanks! This definitely makes sense to have.

cranelift/filetests/filetests/isa/aarch64/symbol-value-pic.clif
alexcrichton: Sure thing! https://github.com/bytecodealliance/wasmtime/issues/11573

And yeah AFAIK we're ok in Wasmtime, but I''m not 100% sure in that judgment. Your reasoning sounds reasonable to me, however

PR #11574, opened 2025-08-29
Change macOS binary link to aarch64 version

bashor:
It’s more likely that a modern macOS user needs an AArch64 version rather than an x86_64 one.


alexcrichton (approved):
Thanks!

